PR #299, opened 2017-08-25
Remove duplicate open operation for swss.rec file

JipanYanga:
swss.rec has been opened in main(), should not be opened again here.


JipanYanga:
Fixed in https://github.com/Azure/sonic-swss/pull/298

PR #301, opened 2017-08-29
PG buffer profile based on port configuration

andriymoroz-mlnx:


andriymoroz-mlnx:
related PR https://github.com/Azure/sonic-buildimage/pull/926

debian/swss.install
```diff
@@ -1,8 +1,7 @@
 swssconfig/sample/netbouncer.json etc/swss/config.d
 swssconfig/sample/00-copp.config.json etc/swss/config.d
-swssconfig/sample/msn2700.32ports.buffers.json etc/swss/config.d
+swssconfig/sample/msn27xx.32ports.buffers.json.j2 usr/share/sonic/templates
```
stcheng: i'm also wondering if we could still leave the file inside the `/etc/swss/config.d` folder. i think to have all swss related configurations (including templates) inside this folder is better than putting the files somewhere around.

andriymoroz-mlnx:
buffers json file generated from the template in this review:
[msn27xx-32ports-buffers-json](https://gist.github.com/andriymoroz-mlnx/e36d2cc39007a217e90437263f880273#file-msn27xx-32ports-buffers-json)
I've uploaded it as the "gist". It allows commenting but not per line. But at least it shows the line numbers. If you know any better way to publish/comment files - please let me know

debian/swss.install
andriymoroz-mlnx: probably you are right...
also the source directory looks not very nice
"production" configs should be under some `sonic-buildimage/platform/<vendor>/config` and `sonic-buildimage/platform/generic/config`
I think this should be done globally for all configuration files

stcheng: i agree with your idea here. i created the `sample` folder but haven't imagined that it is no longer 'sample' folder. we could fix it in the long run. but right now, i'd suggest it is okay to put it into the 'sample' folder.

stcheng:
thanks for hadding the output. gist is nice. Ying and I will take a look!

swssconfig/sample/msn27xx.32ports.buffers.json.j2
```diff
@@ -0,0 +1,284 @@
...
+    {
+        "BUFFER_POOL_TABLE:ingress_lossless_pg_pool": {
+            "size": "{{ ingress_lossless_pg_pool_size | sum }}",
+            "type": "ingress",
+            "mode": "dynamic"
+        },
+        "OP": "SET"
+    },
+    {
+        "BUFFER_PG_TABLE:{{ ethernet_interfaces_names }}:0-1": {
```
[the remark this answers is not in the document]
andriymoroz-mlnx: The other table which previously linked one profile to all ports and queues 3-4 is now split into many tables (one per port). These tables generated in j2 template in [line 219](https://github.com/Azure/sonic-swss/pull/301/files#diff-210587b996c1cb954589dfc9fbc96644R219)
In the [example of generated buffers config json](https://gist.github.com/andriymoroz-mlnx/e36d2cc39007a217e90437263f880273#file-msn27xx-32ports-buffers-json-L115) they are in lines 115-337

stcheng:
Hello @andriymoroz-mlnx could you close this pull request and generate a new one against the master?

PR #303, opened 2017-08-31
[AclOrch]: refactor doAclTableTask and doAclRuleTask

sihuihan88:
After refactoring, users can directly call add/removeAclTable, add/removeAclRule to do the corresponding action. This flexibility is needed for PFC WD, and will be helpful for others as well.

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1060,6 +1060,108 @@ void AclOrch::doTask(Consumer &consumer)
     }
 }
 
+void AclOrch::addAclTable(AclTable &newTable, string table_id)
+{
+    sai_object_id_t table_oid = getTableById(table_id);
+
+    if (table_oid != SAI_NULL_OBJECT_ID)
+    {
+        // table already exists, delete it first
+        if (deleteUnbindAclTable(table_oid) == SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_INFO("Successfully deleted ACL table %s", table_id.c_str());
+            m_AclTables.erase(table_oid);
+        }
+    }
+    if (createBindAclTable(newTable, table_oid) == SAI_STATUS_SUCCESS)
+    {
+        m_AclTables[table_oid] = newTable;
+        SWSS_LOG_INFO("Successfully created ACL table %s, oid: %lX", newTable.description.c_str(), table_oid);
+    }
+    else
+    {
+        SWSS_LOG_ERROR("Failed to create table %s", table_id.c_str());
+    }
+
+}
+
+void AclOrch::removeAclTable(string table_id)
+{
+    sai_object_id_t table_oid = getTableById(table_id);
+    if (table_oid != SAI_NULL_OBJECT_ID)
+    {
+        if (deleteUnbindAclTable(table_oid) == SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_INFO("Successfully deleted ACL table %s", table_id.c_str());
+            m_AclTables.erase(table_oid);
+        }
+        else
+        {
+            SWSS_LOG_ERROR("Failed to delete ACL table %s", table_id.c_str());
+        }
+    }
+    else
+    {
+        SWSS_LOG_ERROR("Failed to delete ACL table. Table %s does not exist", table_id.c_str());
+    }
+
+}
+
+void AclOrch::addAclRule(shared_ptr<AclRule> newRule, string table_id, string rule_id)
+{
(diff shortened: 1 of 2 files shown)


orchagent/aclorch.cpp
```diff
@@ -1060,6 +1060,108 @@ void AclOrch::doTask(Consumer &consumer)
...
+        else
+        {
+            SWSS_LOG_ERROR("Failed to delete ACL table %s", table_id.c_str());
+        }
+    }
+    else
+    {
+        SWSS_LOG_ERROR("Failed to delete ACL table. Table %s does not exist", table_id.c_str());
+    }
+
```
andriymoroz-mlnx: extra newline...

orchagent/aclorch.cpp
```diff
@@ -1060,6 +1060,108 @@ void AclOrch::doTask(Consumer &consumer)
...
+    if (newRule->create())
+    {
+        m_AclTables[table_oid].rules[rule_id] = newRule;
+        SWSS_LOG_INFO("Successfully created ACL rule %s in table %s", rule_id.c_str(), table_id.c_str());
+    }
+    else
+    {
+        SWSS_LOG_ERROR("Failed to create rule in table %s", table_id.c_str());
+    }
+
```
andriymoroz-mlnx: extra newline

orchagent/aclorch.cpp
```diff
@@ -1060,6 +1060,108 @@ void AclOrch::doTask(Consumer &consumer)
...
+        else
+        {
+            SWSS_LOG_ERROR("Failed to delete ACL rule. Unknown rule %s", rule_id.c_str());
+        }
+    }
+    else
+    {
+        SWSS_LOG_ERROR("Failed to delete rule %s from ACL table %s. Table does not exist", rule_id.c_str(), table_id.c_str());
+    }
+
```
andriymoroz-mlnx: extra newline

orchagent/aclorch.cpp
```diff
@@ -1060,6 +1060,108 @@ void AclOrch::doTask(Consumer &consumer)
     }
 }
 
+void AclOrch::addAclTable(AclTable &newTable, string table_id)
```
andriymoroz-mlnx: in this and other functions I would suggest to return some status and not void
it will be useful for the PFC WD or some other caller

orchagent/aclorch.cpp
```diff
@@ -1060,6 +1060,108 @@ void AclOrch::doTask(Consumer &consumer)
...
+            }
+        }
+        else
+        {
+            SWSS_LOG_ERROR("Failed to delete ACL rule. Unknown rule %s", rule_id.c_str());
+        }
+    }
+    else
+    {
+        SWSS_LOG_ERROR("Failed to delete rule %s from ACL table %s. Table does not exist", rule_id.c_str(), table_id.c_str());
```
stcheng: same as above.

orchagent/aclorch.cpp
```diff
@@ -1060,6 +1060,108 @@ void AclOrch::doTask(Consumer &consumer)
     }
 }
 
+void AclOrch::addAclTable(AclTable &newTable, string table_id)
+{
+    sai_object_id_t table_oid = getTableById(table_id);
+
+    if (table_oid != SAI_NULL_OBJECT_ID)
+    {
+        // table already exists, delete it first
```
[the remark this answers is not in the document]
sihuihan88: This depends on what the init design is. If the design is to remove first if the table exists, then this is needed. Since I am not sure about the initial design. I would rather leave it here.

orchagent/aclorch.cpp
sihuihan88: Returning the status is not necessary for now, since in aclorch we are not checking these status. We could refactor and add this later if we want to check the status in aclorch.

stcheng: @andriymoroz-mlnx I think it is okay that we update this in the future pull request. More refactorings are needed.

stcheng:
Hi, @sihuihan88 could you close this one and reopen a new one against the master? Thanks.

orchagent/aclorch.cpp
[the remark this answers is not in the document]
zhenggen-xu: The initial design won't work, especially if the table has rules associated with it.  This will trigger SAI level failure due to the resource busy.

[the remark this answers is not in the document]
sihuihan88: I agree the initial design will cause SAI level failure. That will require more in-depth change, which is beyond the scope of this pr. We need to change the AclOrch design.

PR #307, opened 2017-09-11
[switchorch]: Add SwitchOrch to deal with switch attributes changes

stcheng:
Right now, three switch attributes are supported: FDB unicast/multicast/
broadcast miss packet action. The attribute could be set to drop, forward,
or trap.

The current implementation adds switch configuration to the APPL_DB.
In the future, it would be migrated to the CONF_DB.

### DIFF
--- a/orchagent/switchorch.cpp
+++ b/orchagent/switchorch.cpp
@@ -1,0 +1,85 @@
+#include <map>
+
+#include "switchorch.h"
+
+using namespace std;
+using namespace swss;
+
+extern sai_object_id_t gSwitchId;
+extern sai_switch_api_t *sai_switch_api;
+
+const map<string, sai_switch_attr_t> switch_attribute_map =
+{
+    {"fdb_unicast_miss_packet_action",      SAI_SWITCH_ATTR_FDB_UNICAST_MISS_PACKET_ACTION},
+    {"fdb_broadcast_miss_packet_action",    SAI_SWITCH_ATTR_FDB_BROADCAST_MISS_PACKET_ACTION},
+    {"fdb_multicast_miss_packet_action",    SAI_SWITCH_ATTR_FDB_MULTICAST_MISS_PACKET_ACTION}
+};
+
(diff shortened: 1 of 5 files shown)


gongjianLhr:
Great it can meet our requirement about setting mac aging time dynamically when it is done and after a little update.

orchagent/switchorch.cpp
```diff
@@ -0,0 +1,85 @@
+#include <map>
+
+#include "switchorch.h"
+
+using namespace std;
+using namespace swss;
+
+extern sai_object_id_t gSwitchId;
```
andriymoroz-mlnx: since we are going to have SwitchOrch, I suggest to replace global gSwitchId variable to 
static member + some static getSwitchId() method in SwitchOrch class

stcheng: I agree. I would suggest to move it to a separate pull request and I'm thinking about creating a Switch class to move all the Orch instances into this class as singletons. What do you think about this idea?

andriymoroz-mlnx: sure

stcheng:
@gongjianLhr @andriymoroz-mlnx  could you review the new pull request? if it looks fine, could you approve that? thanks!

orchagent/switchorch.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+                if (switch_attribute_map.find(attribute) == switch_attribute_map.end())
+                {
+                    SWSS_LOG_ERROR("Unsupported switch attribute %s", attribute.c_str());
+                    it = consumer.m_toSync.erase(it);
+                    continue;
+                }
+
+                auto value = fvValue(i);
+                if (packet_action_map.find(value) == packet_action_map.end())
```
gongjianLhr: It is not so generic here. What if I want to change a switch attribute whose value is not about pack_action?

stcheng: good point. I think we could patch that in a more general way. thanks for the feedback.

PR #308, opened 2017-09-11
[buffer]: Remove failed tasks from the m_toSync queue

stcheng:
- Remove the task with erase(it) to avoid looping inside the while
- Align the code and refactor the error messages
- Remove table verification code since it is unnecessary


stcheng:
very similar to this pull request: https://github.com/Azure/sonic-swss/pull/306

PR #312, opened 2017-09-16
Optimize orch performance by pops()

qiluo-msft:
For route sync performance, original implementation will pop route one-by-one, and go through m_toSync each time. If using pops(), the times of loop m_toSync will decrease largely.

### DIFF
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -83,54 +83,58 @@ bool Orch::execute(string tableName)
     }
     Consumer& consumer = consumer_it->second;
 
-    KeyOpFieldsValuesTuple new_data;
-    consumer.m_consumer->pop(new_data);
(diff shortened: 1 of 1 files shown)


orchagent/orch.cpp
```diff
@@ -83,54 +83,58 @@ bool Orch::execute(string tableName)
...
-    if (op.empty())
+    /* Nothing popped */
+    if (new_data.empty())
     {
         return true;
     }
 
-    /* Record incoming tasks */
-    if (gSwssRecord)
+    for (auto new_data1: new_data)
```
lguohan: >new_data [](start = 25, length = 8)

entries

lguohan: >new_data1 [](start = 14, length = 9)

entry?

new_data1 does not have any meaning.

PR #314, opened 2017-09-19
[switchorch]: Add SwitchOrch to deal with switch attributes changes

stcheng:
Right now, three switch attributes are supported: FDB unicast/multicast/
broadcast miss packet action. The attribute could be set to drop, forward,
or trap.

The current implementation adds switch configuration to the APPL_DB.
In the future, it would be migrated to the CONF_DB.

**What I did**
Add the functionality of updating switch attributes. This architecture also allows future expansion for richer attributes other than FDB related ones.

**Why I did it**
It's a feature request from various sources.

**How I verified it**
Use `swssconfig` to load a switch attribute set JSON file and check in the database if this attribute is set.

### DIFF
--- a/orchagent/Makefile.am
+++ b/orchagent/Makefile.am
@@ -10,7 +10,7 @@ else
 DBGFLAGS = -g -DNDEBUG
 endif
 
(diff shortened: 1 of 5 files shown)


stcheng:
retest this please

PR #320, opened 2017-09-25
[portsorch]: Add support of cable breakout feature

volodymyrsamotiy:

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -5,12 +5,14 @@
 #include <sstream>
 #include <set>
 #include <algorithm>
+#include <tuple>
 
 #include <netinet/if_ether.h>
 #include "net/if.h"
 
 #include "logger.h"
 #include "schema.h"
+#include "converter.h"
 
 extern sai_switch_api_t *sai_switch_api;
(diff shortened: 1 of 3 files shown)


stcheng:
@nikos-li do you have some concerns about this pull request? Let's discuss them here.

orchagent/portsorch.cpp
```diff
@@ -470,6 +472,107 @@ void PortsOrch::updateDbPortOperStatus(sai_object_id_t id, sai_port_oper_status_
...
+            SWSS_LOG_INFO("Port has already been initialized before alias:%s", alias.c_str());
+        }
+        else
+        {
+            Port p(alias, Port::PHY);
+
+            p.m_index = static_cast<int32_t>(m_portList.size()); // TODO: Assume no deletion of physical port
+            p.m_port_id = id;
+
+            /* Initialize the port and create router interface and host interface */
```
lguohan: >create router interface [](start = 39, length = 23)

this comment here is actualy incorrect, the initializePort no longer crate router interface.

stcheng: correct. let me modify it.

PR #323, opened 2017-09-27
Added queue stat counters map

andrii-savka:
SWSS part of SNMP queue stat counters feature.

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -15,6 +15,8 @@
 #include "converter.h"
 #include "saiserialize.h"
 
+#include "saiserialize.h"
+
 extern sai_switch_api_t *sai_switch_api;
 extern sai_bridge_api_t *sai_bridge_api;
 extern sai_port_api_t *sai_port_api;
@@ -22,6 +24,7 @@ extern sai_vlan_api_t *sai_vlan_api;
 extern sai_lag_api_t *sai_lag_api;
 extern sai_hostif_api_t* sai_hostif_api;
 extern sai_acl_api_t* sai_acl_api;
+extern sai_queue_api_t *sai_queue_api;
 extern sai_object_id_t gSwitchId;
 
 #define VLAN_PREFIX         "Vlan"
@@ -51,6 +54,9 @@ PortsOrch::PortsOrch(DBConnector *db, vector<string> tableNames) :
 
     /* Initialize port table */
     m_portTable = unique_ptr<Table>(new Table(m_db, APP_PORT_TABLE_NAME));
+
+    /* Initialize queue table */
+    m_queueTable = unique_ptr<Table>(new Table(counter_db, COUNTERS_QUEUE_NAME_MAP));
 
     uint32_t i, j;
     sai_status_t status;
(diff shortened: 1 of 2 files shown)


orchagent/portsorch.cpp
```diff
@@ -49,6 +52,9 @@ PortsOrch::PortsOrch(DBConnector *db, vector<string> tableNames) :
     /* Initialize port table */
     m_portTable = unique_ptr<Table>(new Table(m_db, APP_PORT_TABLE_NAME));
 
+    /* Initialize queue table */
+    m_queueTable = unique_ptr<Table>(new Table(counter_db, COUNTERS_QUEUE_NAME_MAP));
```
qiluo-msft: try make_unique?
If it works, please help modify existing code above.

orchagent/portsorch.cpp
```diff
@@ -1045,7 +1051,21 @@ void PortsOrch::initializeQueues(Port &port)
         SWSS_LOG_ERROR("Failed to get queue list for port %s rv:%d", port.m_alias.c_str(), status);
         throw runtime_error("PortsOrch initialization failure.");
     }
+
     SWSS_LOG_INFO("Get queues for port %s", port.m_alias.c_str());
+
+    /* Create the Queue map in the Counter DB */
+    vector<FieldValueTuple> vector;
```
qiluo-msft: >vector [](start = 28, length = 6)

vector as variable name is really confusing.

orchagent/portsorch.cpp
```diff
@@ -1045,7 +1051,21 @@ void PortsOrch::initializeQueues(Port &port)
...
+
+    for (size_t queue_index = 0; queue_index < port.m_queue_ids.size(); ++queue_index)
+    {
+        std::stringstream name;
```
qiluo-msft: ostringstream

orchagent/portsorch.cpp
```diff
@@ -1045,7 +1051,21 @@ void PortsOrch::initializeQueues(Port &port)
...
+        name << port.m_alias << ":" << queue_index;
+        FieldValueTuple tuple(name.str(), sai_serialize_object_id(port.m_queue_ids[queue_index]));
+        vector.push_back(tuple);
+    }
+
+    m_queueTable->set("", vector);
```
qiluo-msft: design question: why serialize all pairs into 'vector'? why not store them as key-value in the table?

orchagent/portsorch.cpp
marian-pritsak: Why do we need to mix refactoring with new functionality? This line is written in the same style as above. If there's a need for this kind of change, it should be done separately.

qiluo-msft: I don't know @andriis@mellanox.com 's idea. If you think the code review comment is not acceptable, I am ok with new iteration only on the new code. I prefer better code quality than consistent code style.

Let me handle existing code refactoring if you don't want to.

---
In reply to: [141885713](https://github.com/Azure/sonic-swss/pull/323#discussion_r141885713) [](ancestors = 141885713)

orchagent/portsorch.cpp
andrii-savka: It's written in the same style as port counters

andrii-savka:
Fixed review remarks.

PR #324, opened 2017-09-27
[switchorch]: Add support of ECMP and LAG hash seed attribute

volodymyrsamotiy:

### DIFF
--- a/orchagent/switchorch.cpp
+++ b/orchagent/switchorch.cpp
@@ -1,6 +1,7 @@
 #include <map>
 
 #include "switchorch.h"
+#include "converter.h"
 
 using namespace std;
 using namespace swss;
@@ -12,7 +13,9 @@ const map<string, sai_switch_attr_t> switch_attribute_map =
 {
     {"fdb_unicast_miss_packet_action",      SAI_SWITCH_ATTR_FDB_UNICAST_MISS_PACKET_ACTION},
     {"fdb_broadcast_miss_packet_action",    SAI_SWITCH_ATTR_FDB_BROADCAST_MISS_PACKET_ACTION},
-    {"fdb_multicast_miss_packet_action",    SAI_SWITCH_ATTR_FDB_MULTICAST_MISS_PACKET_ACTION}
+    {"fdb_multicast_miss_packet_action",    SAI_SWITCH_ATTR_FDB_MULTICAST_MISS_PACKET_ACTION},
+    {"ecmp_hash_seed",                      SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_SEED},
+    {"lag_hash_seed",                       SAI_SWITCH_ATTR_LAG_DEFAULT_HASH_SEED}
 };
 
 const map<string, sai_packet_action_t> packet_action_map =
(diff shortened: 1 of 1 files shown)


orchagent/switchorch.cpp
```diff
@@ -42,38 +46,67 @@ void SwitchOrch::doTask(Consumer &consumer)
         {
             for (auto i : kfvFieldsValues(t))
             {
-                auto attribute = fvField(i);
+                vector<string> attributes;
 
-                if (switch_attribute_map.find(attribute) == switch_attribute_map.end())
+                if (fvField(i) == "hash_seed")
```
stcheng: i wonder if we don't need such logics here. if we would like to set the two values with the same value, it is okay that we feed the `swssconfig` with the JSON file that contains both attributes with the same value. Thus the `for` loop below is not necessary.

volodymyrsamotiy: I agree that it would be better and logical to set hash seed attributes separately. I will update the pull request with appropriate change.
Since changes were done according to requirements document, I believe it should  be updated as well. (https://github.com/Azure/SONiC/wiki/ECMP-and-LAG-Hash-Seed)

orchagent/switchorch.cpp
```diff
@@ -12,7 +13,9 @@ const map<string, sai_switch_attr_t> switch_attribute_map =
 {
     {"fdb_unicast_miss_packet_action",      SAI_SWITCH_ATTR_FDB_UNICAST_MISS_PACKET_ACTION},
     {"fdb_broadcast_miss_packet_action",    SAI_SWITCH_ATTR_FDB_BROADCAST_MISS_PACKET_ACTION},
-    {"fdb_multicast_miss_packet_action",    SAI_SWITCH_ATTR_FDB_MULTICAST_MISS_PACKET_ACTION}
+    {"fdb_multicast_miss_packet_action",    SAI_SWITCH_ATTR_FDB_MULTICAST_MISS_PACKET_ACTION},
+    {"ecmp_default_hash_seed",              SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_SEED},
```
lguohan: -> ecmp_hash_seed

orchagent/switchorch.cpp
```diff
@@ -12,7 +13,9 @@ const map<string, sai_switch_attr_t> switch_attribute_map =
...
+    {"lag_default_hash_seed",               SAI_SWITCH_ATTR_LAG_DEFAULT_HASH_SEED}
```
lguohan: -> lag_hash_seed

orchagent/switchorch.cpp
```diff
@@ -52,16 +55,32 @@ void SwitchOrch::doTask(Consumer &consumer)
...
+                            SWSS_LOG_ERROR("Unsupported packet action %s", value.c_str());
+                            it = consumer.m_toSync.erase(it);
+                            continue;
+                        }
+                        attr.value.s32 = packet_action_map.at(value);
+                        break;
+
+                    case SAI_SWITCH_ATTR_ECMP_DEFAULT_HASH_SEED:
+                    case SAI_SWITCH_ATTR_LAG_DEFAULT_HASH_SEED:
+                        attr.value.u32 = to_uint<uint32_t>(value);
```
lguohan: do we have test cases for different hash seeds?

orchagent/switchorch.cpp
stcheng: agree. default shall not be in the attribute name. default is a value not an attribute. the SAI attribute name is not perfect.

orchagent/switchorch.cpp
volodymyrsamotiy: No, as far as I know, we don't have.

PR #326, opened 2017-09-28
[pfcwdactionhandler]: Add PG setting

marian-pritsak:
PFC storm drop handler that applies zero size buffer profile to a queue, also
has to apply zero size profile to ingress PG to drop RX traffic too.
Code that creates profiles and pools was refactored to always create
only one mode of pools/profiles (because it actually doesn't matter which mode
is configured - traffic is dropped anyway).

### DIFF
--- a/orchagent/pfcactionhandler.cpp
+++ b/orchagent/pfcactionhandler.cpp
@@ -1,6 +1,7 @@
 #include "pfcactionhandler.h"
 #include "logger.h"
 #include "saiserialize.h"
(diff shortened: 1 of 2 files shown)


marian-pritsak:
@sihuihan88 please review

sihuihan88:
Are we going to create profile at the init stage

marian-pritsak:
Yes, let me do it in a separate pull request. Since it is platform-dependent, I will add an interface to initialize resources at the beginning.

PR #327, opened 2017-09-29
orchagent: Support port/LAG in multiple VLANs

JipanYanga:
One port/LAG may join multi-VLAN in tagged/untagged mode.
pvid on LAG is not supported in SAI v1.0, as workaround we deal with each LAG member explicitly for pvid.
bridge port has to be created first for port/LAG to join 1Q VLAN. The same bridge port is used to join all VLANs.
For each VLAN that port/LAG has joined, there is one vlan member id which is the handle for VLAN member attribute get/set and remove.

**What I did**
Support port/LAG in multiple VLANs
**Why I did it**
VLAN trunk feature implementation
**How I verified it**
Passed VLAN trunk test cases

**Details if related**


stcheng:
retest this please

orchagent/port.h
```diff
@@ -8,11 +8,26 @@ extern "C" {
...
 
 namespace swss {
 
+struct VlanMemberEntry
+{
+    sai_object_id_t            vlan_member_id;
+    sai_vlan_tagging_mode_t    vlan_mode;
+};
+
+typedef std::map<sai_vlan_id_t, VlanMemberEntry> port_vlan_members_t;
```
lguohan: >port_vlan_members_t [](start = 49, length = 19)

rename to vlan_members_t

orchagent/portsorch.cpp
```diff
@@ -345,6 +345,74 @@ bool PortsOrch::setPortMtu(sai_object_id_t id, sai_uint32_t mtu)
...
+
+    if (port.m_vlan_members.size() == 0)
+    {
+        pvid = DEFAULT_PORT_VLAN_ID;
+        return true;
+    }
+
+    for (auto &vme: port.m_vlan_members)
+    {
+        if(vme.second.vlan_mode == SAI_VLAN_TAGGING_MODE_UNTAGGED) {
```
lguohan: >{ [](start = 67, length = 1)

'{' in a separate line.

lguohan: >f( [](start = 9, length = 2)

keep the space

orchagent/portsorch.cpp
```diff
@@ -345,6 +345,74 @@ bool PortsOrch::setPortMtu(sai_object_id_t id, sai_uint32_t mtu)
     return true;
 }
 
+bool PortsOrch::setPortPvid (Port &port, sai_uint32_t pvid)
```
lguohan: >  [](start = 27, length = 1)

no space here.

orchagent/portsorch.cpp
```diff
@@ -534,6 +603,10 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 if (fvField(i) == "mtu")
                     mtu = (uint32_t)stoul(fvValue(i));
 
+                /* Set port pvid */
+                if (fvField(i) == "pvid")
+                    pvid = (uint32_t)stoul(fvValue(i));
```
lguohan: why do we need this? we can keep the assumption that the untagged mode vlan is the pvid of the port.

orchagent/portsorch.cpp
```diff
@@ -691,24 +782,43 @@ void PortsOrch::doVlanTask(Consumer &consumer)
...
+                // TODO: UNKNOWN_UNICAST_FLOOD_GROUP, UNKNOWN_MULTICAST_FLOOD_GROUP, BROADCAST_FLOOD_GROUP
+            }
+
+            if (mtu != 0)
+            {
+                Port v;
+                getPort(vlan_alias, v);
+                v.m_mtu = mtu;
+                // TODO: update MTU for the IP interface of the VLAN
+            }
```
lguohan: since it is todo, I suggest to remove it. The rif MTU should be configured under interface_table.

orchagent/portsorch.cpp
```diff
@@ -1155,6 +1261,9 @@ bool PortsOrch::addBridgePort(Port &port)
 {
     SWSS_LOG_ENTER();
 
+    if (port.m_bridge_port_id)
+        return true;
```
lguohan: use {} even for one line statement.

lguohan: I think we should print some warning here, since we should not add a port to a bridge if the port is in a bridge already.

---
In reply to: [142000240](https://github.com/Azure/sonic-swss/pull/327#discussion_r142000240) [](ancestors = 142000240)

orchagent/portsorch.cpp
```diff
@@ -1252,82 +1387,79 @@ bool PortsOrch::removeVlan(Port vlan)
...
+                port.m_alias.c_str(), vlan.m_alias.c_str(), vlan.m_vlan_info.vlan_id, port.m_port_id);
         return false;
     }
     SWSS_LOG_NOTICE("Add member %s to VLAN %s vid:%hu pid%lx",
-            port.m_alias.c_str(), vlan.m_alias.c_str(), vlan.m_vlan_id, port.m_port_id);
+            port.m_alias.c_str(), vlan.m_alias.c_str(), vlan.m_vlan_info.vlan_id, port.m_port_id);
 
-    if (tagging_mode == "untagged") // set pvlan id for untagged port only
+    /* Use untagged VLAN as pvid of the member port when no pvid has been configured explictly */
+    if (port.m_port_vlan_id == DEFAULT_PORT_VLAN_ID && sai_tagging_mode == SAI_VLAN_TAGGING_MODE_UNTAGGED)
```
lguohan: >port.m_port_vlan_id == DEFAULT_PORT_VLAN_ID  [](start = 8, length = 44)

not sure why we need this condition, can you explain?

orchagent/portsorch.cpp
```diff
@@ -1338,37 +1470,39 @@ bool PortsOrch::addVlanMember(Port vlan, Port port, string& tagging_mode)
...
-
-    sai_attribute_t attr;
-    attr.id = SAI_PORT_ATTR_PORT_VLAN_ID;
-    attr.value.u16 = DEFAULT_PORT_VLAN_ID;
+            port.m_alias.c_str(), vlan.m_alias.c_str(), vlan.m_vlan_info.vlan_id, vlan_member_id);
 
-    status = sai_port_api->set_port_attribute(port.m_port_id, &attr);
-    if (status != SAI_STATUS_SUCCESS)
+    /* Restore to default pvid if no pvid has been configured explictly and this port was in untagged mode */
+    if (port.m_port_vlan_id == DEFAULT_PORT_VLAN_ID && sai_tagging_mode == SAI_VLAN_TAGGING_MODE_UNTAGGED)
```
lguohan: >port.m_port_vlan_id == DEFAULT_PORT_VLAN_ID &&  [](start = 8, length = 47)

not sure if we need this.

lguohan:
can you also submit the vlan trunk test you mentioned in the PR.

orchagent/portsorch.h
```diff
@@ -88,20 +88,22 @@ class PortsOrch : public Orch, public Subject
...
     bool removeLag(Port lag);
-    bool addLagMember(Port lag, Port port);
-    bool removeLagMember(Port lag, Port port);
-
+    bool addLagMember(Port &lag, Port &port);
+    bool removeLagMember(Port &lag, Port &port);
+    void getLagMember(Port &lag, vector<Port> &portv);
     bool setPortAdminStatus(sai_object_id_t id, bool up);
     bool setPortMtu(sai_object_id_t id, sai_uint32_t mtu);
+    bool setPortPvid (Port &port, sai_uint32_t pvid);
```
lguohan: >setPortPvid [](start = 9, length = 11)

It is better to rename it as setPvid, no need to give a port. actually, I think we should also remove the Port from the functions names the setPortAdminStatus/setPortMtu/setPortSpeed/getPortSpeed

orchagent/portsorch.cpp
```diff
@@ -1462,13 +1616,17 @@ bool PortsOrch::addLagMember(Port lag, Port port)
 
     m_portList[lag.m_alias] = lag;
 
+    sai_uint32_t pvid = DEFAULT_PORT_VLAN_ID;
```
lguohan: > = DEFAULT_PORT_VLAN_ID [](start = 21, length = 23)

not needed.

orchagent/portsorch.cpp
```diff
@@ -1462,13 +1616,17 @@ bool PortsOrch::addLagMember(Port lag, Port port)
...
+    if (getPortPvid(lag, pvid))
+        setPortPvid (port, pvid);
```
lguohan: use {}

orchagent/portsorch.cpp
```diff
@@ -345,6 +345,74 @@ bool PortsOrch::setPortMtu(sai_object_id_t id, sai_uint32_t mtu)
...
+    }
+
+    for (auto &vme: port.m_vlan_members)
+    {
+        if(vme.second.vlan_mode == SAI_VLAN_TAGGING_MODE_UNTAGGED) {
+            pvid =  vme.first;
+            return true;
+        }
+    }
+    pvid = DEFAULT_PORT_VLAN_ID;
```
lguohan: this is complicated, when the port is not a router interface, we should just return port.m_port_vlan_id, and we should make the m_port_vlan_id correct.

orchagent/portsorch.cpp
```diff
@@ -345,6 +345,74 @@ bool PortsOrch::setPortMtu(sai_object_id_t id, sai_uint32_t mtu)
     return true;
 }
 
+bool PortsOrch::setPortPvid (Port &port, sai_uint32_t pvid)
+{
+    SWSS_LOG_ENTER();
+    vector<Port> portv;
+
+    if (port.m_type == Port::PHY)
```
lguohan: >   if (port.m_type == Port::PHY) [](start = 1, length = 32)

if the port.m_rid_id is not null, we should return false. since we do not allow set pvid for router interface.

orchagent/portsorch.cpp
```diff
@@ -345,6 +345,74 @@ bool PortsOrch::setPortMtu(sai_object_id_t id, sai_uint32_t mtu)
...
+        attr.value.u32 = pvid;
+
+        sai_status_t status = sai_port_api->set_port_attribute(p.m_port_id, &attr);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("Failed to set pvid %u to port: %s", attr.value.u32, p.m_alias.c_str());
+            return false;
+        }
+        SWSS_LOG_NOTICE("Set pvid %u to port: %s", attr.value.u32, p.m_alias.c_str());
+    }
```
lguohan: m_port_vlan_id = pvid

orchagent/port.h
```diff
@@ -53,17 +68,17 @@ class Port
     Type                m_type;
     int                 m_index = 0;    // PHY_PORT: index
     int                 m_ifindex = 0;
+    sai_uint32_t        m_mtu;
```
lguohan: this is irrelevant to the pr. consider to remove.

orchagent/port.h
JipanYanga: done

orchagent/portsorch.cpp
JipanYanga: done

JipanYanga: done

orchagent/portsorch.cpp
JipanYanga: done

orchagent/portsorch.cpp
JipanYanga: The original design is to support pvid setting on port and lag from admin interface,  as provided at SAI api.  Since it is causing a lot of confusion and no actual use case for it yet,  removing support for explicit pvid configuration.

orchagent/portsorch.cpp
JipanYanga: done.  
btw,  VLAN_TABLE schema might needs some update to remove mtu field.
mtu                 = 1*4DIGIT             ; MTU for the IP interface of the VLAN

orchagent/portsorch.cpp
JipanYanga: done

orchagent/portsorch.cpp
JipanYanga: As explained previously, it was for supporting pvid configuration from admin interface.  
removed.

orchagent/portsorch.cpp
JipanYanga: As explained previously, it was for supporting pvid configuration from admin interface.
removed.

orchagent/portsorch.cpp
JipanYanga: removed.

orchagent/portsorch.cpp
JipanYanga: done

orchagent/portsorch.cpp
JipanYanga: As explained previously, it was for supporting pvid configuration from admin interface.

Updated.

orchagent/portsorch.cpp
JipanYanga: done.

orchagent/portsorch.cpp
JipanYanga: done.

orchagent/port.h
JipanYanga: done.

JipanYanga:
vlan trunk test mentioned in this PR is the VLAN trunk test plan reviewed at SONiC community meeting few weeks.  
Developer of the VLAN trunk test plan will push the document to github soon.

stcheng:
hi could you resolve the conflicts?

orchagent/portsorch.cpp
```diff
@@ -1179,15 +1251,25 @@ bool PortsOrch::addBridgePort(Port &port)
...
 {
     SWSS_LOG_ENTER();
 
+    if (port.m_vlan_members.size() > 0)
+    {
+        for (auto &vme: port.m_vlan_members)
+        {
+            SWSS_LOG_DEBUG("Port %s still in VLAN %d", port.m_alias.c_str(), vme.first);
+        }
+        return true;
```
lguohan: this should be false.

orchagent/portsorch.cpp
```diff
@@ -1179,15 +1251,25 @@ bool PortsOrch::addBridgePort(Port &port)
...
-bool PortsOrch::removeBridgePort(Port port)
+bool PortsOrch::removeBridgePort(Port &port)
 {
     SWSS_LOG_ENTER();
 
+    if (port.m_vlan_members.size() > 0)
+    {
+        for (auto &vme: port.m_vlan_members)
+        {
+            SWSS_LOG_DEBUG("Port %s still in VLAN %d", port.m_alias.c_str(), vme.first);
```
lguohan: SWSS_LOG_ERROR, this should not happen at orchagent level, the dependency resolution should be handled by the mgr level.

orchagent/portsorch.cpp
JipanYanga: each time  a port/lag leaves a VLAN,  " if (removeVlanMember(vlan, port) && removeBridgePort(port))"  is called,    port.m_vlan_members.erase(vlan_member) is called in removeVlanMember(vlan, port), and here it is to check whether the port/lag is still in any vlan.  It is ok to be non-empty.

orchagent/portsorch.cpp
JipanYanga: As previous explanation.    
We may argue there could be different way of implementation if it is too confusing.

orchagent/portsorch.cpp
lguohan: this make the return value confusing, according to the function name the return value true means you have removed the bridge port, in this case you should return false.

orchagent/portsorch.cpp
```diff
@@ -1213,9 +1295,11 @@ bool PortsOrch::removeBridgePort(Port port)
             port.m_alias.c_str(), status);
         return false;
     }
+    port.m_bridge_port_id = 0;
```
lguohan: 0 -> SAI_NULL_OBJECT_ID

orchagent/portsorch.cpp
lguohan: the return value of the function needs to be consistent, true mean removed, false means not removed.

orchagent/portsorch.cpp
```diff
@@ -817,10 +871,7 @@ void PortsOrch::doVlanMemberTask(Consumer &consumer)
         {
             if (vlan.m_members.find(port_alias) != vlan.m_members.end())
             {
-                /* Assert the port belongs the a VLAN */
-                assert(port.m_vlan_id && port.m_vlan_member_id);
-
-                if (removeVlanMember(vlan, port))
+                if (removeVlanMember(vlan, port) && removeBridgePort(port))
```
lguohan: if there is no vlan member, then we can removeBridgePort.

orchagent/portsorch.cpp
JipanYanga: Ok,  it is a little confusing.  Will do the m_vlan_members set empty check outside of this function.

orchagent/portsorch.cpp
```diff
@@ -1252,82 +1336,79 @@ bool PortsOrch::removeVlan(Port vlan)
...
     {
-        SWSS_LOG_ERROR("Failed to remove VLAN %s vid:%hu", vlan.m_alias.c_str(), vlan.m_vlan_id);
+        SWSS_LOG_ERROR("Failed to remove VLAN %s vid:%hu",
+                vlan.m_alias.c_str(), vlan.m_vlan_info.vlan_id);
         return false;
     }
 
-    SWSS_LOG_NOTICE("Remove VLAN %s vid:%hu", vlan.m_alias.c_str(), vlan.m_vlan_id);
+    SWSS_LOG_NOTICE("Remove VLAN %s vid:%hu", vlan.m_alias.c_str(),
+		vlan.m_vlan_info.vlan_id);
```
lguohan: >		 [](start = 0, length = 2)

make this as space.

orchagent/portsorch.cpp
```diff
@@ -1179,15 +1251,25 @@ bool PortsOrch::addBridgePort(Port &port)
...
     }
 
+    m_portList[port.m_alias] = port;
     SWSS_LOG_NOTICE("Add bridge port %s to default 1Q bridge", port.m_alias.c_str());
 
     return true;
 }
 
-bool PortsOrch::removeBridgePort(Port port)
+bool PortsOrch::removeBridgePort(Port &port)
```
lguohan: >BridgePo [](start = 22, length = 8)

symmetric to addBridgePort, we can first check if the port is already in bridge or not, if the port is not in bridge, we can return true.

orchagent/portsorch.cpp
```diff
@@ -1155,6 +1208,12 @@ bool PortsOrch::addBridgePort(Port &port)
 {
     SWSS_LOG_ENTER();
 
+    if (port.m_bridge_port_id)
+    {
+        SWSS_LOG_WARN("port %s already in 1Q bridge", port.m_alias.c_str());
```
lguohan: can be DEBUG LEVEL.

orchagent/portsorch.cpp
```diff
@@ -805,10 +862,7 @@ void PortsOrch::doVlanMemberTask(Consumer &consumer)
                 continue;
             }
 
-            /* Assert the port doesn't belong to any VLAN */
-            assert(!port.m_vlan_id && !port.m_vlan_member_id);
-
-            if (addBridgePort(port) && addVlanMember(vlan, port, tagging_mode))
+            if ((port.m_bridge_port_id || addBridgePort(port)) && addVlanMember(vlan, port, tagging_mode))
```
lguohan: >port.m_bridge_port_id ||  [](start = 17, length = 25)

since you check this in addBridgePort, you do not need to check the bridge_port_id here.

orchagent/portsorch.cpp
JipanYanga: done.

orchagent/portsorch.cpp
JipanYanga: done.

orchagent/portsorch.cpp
JipanYanga: done.

orchagent/portsorch.cpp
JipanYanga: there is back and forth for addBridgePort() and removeBridgePort().  
Hope the following implementation is ok:
"Calling addBridgePort() only when bridge port not created for port/lag, and calling removeBridgePort() only when port/lag not in any VLAN"

orchagent/portsorch.cpp
JipanYanga: there is back and forth for addBridgePort() and removeBridgePort().
Hope the following implementation is ok:
"Calling addBridgePort() only when bridge port not created for port/lag, and calling removeBridgePort() only when port/lag not in any VLAN"

orchagent/portsorch.cpp
JipanYanga: As explained before.

JipanYanga:
It looks "bool PortsOrch::removePort(sai_object_id_t port_id)" for port break out feature is triggering syncd crash.  I'll hold from updating orchagent/portsorch.h for the conflict until there is conclusion for the issue: https://github.com/Azure/sonic-sairedis/issues/235

orchagent/portsorch.cpp
lguohan: there is no back and forth, what I meant is that

for addBridgePort() if it is already a bridge port, you can return true. For removeBridgePort, if it is not a bridge, you can also return true.

they are different than when the port is still in vlan, you should return false.

JipanYanga: Ok, let me change it again, hope the new code will match what you want.

orchagent/portsorch.cpp
```diff
@@ -1576,7 +1664,22 @@ bool PortsOrch::removeLag(Port lag)
...
-bool PortsOrch::addLagMember(Port lag, Port port)
+void PortsOrch::getLagMember(Port &lag, vector<Port> &portv)
+{
+    Port member;
+
+    for (auto &name: lag.m_members)
+    {
+        if (!getPort(name, member))
+        {
+            SWSS_LOG_ERROR("Failed to get port for %s alias\n", name.c_str());
```
stcheng: Remove trailing `\n`

orchagent/portsorch.cpp
```diff
@@ -952,13 +1004,18 @@ void PortsOrch::doVlanMemberTask(Consumer &consumer)
         {
             if (vlan.m_members.find(port_alias) != vlan.m_members.end())
             {
-                /* Assert the port belongs the a VLAN */
-                assert(port.m_vlan_id && port.m_vlan_member_id);
-
                 if (removeVlanMember(vlan, port))
```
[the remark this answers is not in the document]
JipanYanga: That was the original implementation, changed to current format to address comment of below: 
"lguohan 3 days ago Contributor
if there is no vlan member, then we can removeBridgePort."

orchagent/portsorch.cpp
JipanYanga: ok

orchagent/portsorch.cpp
```diff
@@ -826,24 +882,23 @@ void PortsOrch::doVlanTask(Consumer &consumer)
 
         if (op == SET_COMMAND)
         {
-            /* Duplicate entry */
-            if (m_portList.find(vlan_alias) != m_portList.end())
+            if (m_portList.find(vlan_alias) == m_portList.end())
```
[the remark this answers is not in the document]
JipanYanga: According to SAI specification, there are some VLAN attributes like mac learning enable/disable, flood control which are configurable though we don't support them for now. 
We may add more comments here.

orchagent/portsorch.cpp
```diff
@@ -354,6 +354,62 @@ bool PortsOrch::setPortMtu(sai_object_id_t id, sai_uint32_t mtu)
...
+    if (port.m_rif_id)
+    {
+        SWSS_LOG_DEBUG("Router interface exists on %s, don't set pvid",
+                      port.m_alias.c_str());
+        return false;
+    }
+
+    pvid = port.m_port_vlan_id;
+    return true;
+}
```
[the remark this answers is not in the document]
JipanYanga: empty line before "return true;"?  It might be good if SONiC has clear specification about the coding style.

orchagent/port.h
```diff
@@ -8,11 +8,26 @@ extern "C" {
...
 
 namespace swss {
 
+struct VlanMemberEntry
+{
+    sai_object_id_t            vlan_member_id;
+    sai_vlan_tagging_mode_t    vlan_mode;
+};
+
+typedef std::map<sai_vlan_id_t, VlanMemberEntry> vlan_members_t;
```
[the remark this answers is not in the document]
JipanYanga: it was changed to this name according to comment:
"
lguohan 6 days ago Contributor
port_vlan_members_t [](start = 49, length = 19)
rename to vlan_members_t
"

orchagent/portsorch.cpp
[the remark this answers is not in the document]
stcheng: I mean between the `}` and the next function.

orchagent/port.h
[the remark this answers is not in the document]
stcheng: ok

JipanYanga:
VLAN trunk test plan:  https://github.com/Azure/SONiC/wiki/VLAN-trunk-test-plan

PR #328, opened 2017-09-29
Orchagent: Host interface tag processing

JipanYanga:
Before SAI_HOSTIF_VLAN_TAG_ORIGINAL is supported by libsai from all asic vendors, the VLAN tag on hostif is explicitly controlled with SAI_HOSTIF_VLAN_TAG_STRIP & SAI_HOSTIF_VLAN_TAG_KEEP attributes.

When a port/LAG is in 1Q bridge, SAI_HOSTIF_VLAN_TAG_KEEP is set for SAI_HOSTIF_ATTR_VLAN_TAG, for all other cases SAI_HOSTIF_VLAN_TAG_STRIP will be used.

**What I did**
Add support for hostif VLAN tag processing support.

**Why I did it**
With VLAN trunk feature,  it is necessary to have packet trapped to host with right VLAN tag.

**How I verified it**
Tested with whole VLAN trunk feature, please refer to VLAN trunk test cases.

**Details if related**


JipanYanga:
The PR has dependency on https://github.com/Azure/sonic-swss/pull/327

orchagent/portsorch.cpp
```diff
@@ -1462,6 +1531,17 @@ bool PortsOrch::addLagMember(Port lag, Port port)
 
     m_portList[lag.m_alias] = lag;
 
+    if (lag.m_vlan_members.size() > 0)
```
lguohan: >if (lag.m_vlan_members.size() > 0) [](start = 4, length = 34)

this criteria is a little bit indirect, it is better to check if lag is a bridge port or not (do we have an attribute to indicate if the lag is a bridge port or not), if it is then set to keep.

stcheng: to check if a lag is a bridge port or not, could we check the m_bridge_port_id variable? we're ensuring that once a port is added to a bridge port, we assign this id, once the bridge port is removed, we reset this id.

orchagent/portsorch.cpp
```diff
@@ -345,6 +350,50 @@ bool PortsOrch::setPortMtu(sai_object_id_t id, sai_uint32_t mtu)
...
+    {
+        sai_attribute_t attr;
+        attr.id = SAI_HOSTIF_ATTR_VLAN_TAG;
+        attr.value.s32 = strip;
+
+        sai_status_t status = sai_hostif_api->set_hostif_attribute(p.m_hif_id, &attr);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("Failed to set %s to host interface %s",
+                        hostif_vlan_tag[strip], p.m_alias.c_str());
```
stcheng: align the spaces, or just use 8 spaces. same below.

orchagent/portsorch.cpp
```diff
@@ -1179,6 +1233,13 @@ bool PortsOrch::addBridgePort(Port &port)
         return false;
     }
 
+    bool rv = setHostIntfsStripTag(port, SAI_HOSTIF_VLAN_TAG_KEEP);
```
stcheng: just use `if (!setHostIntfsStripTag)` is fine. same below.

orchagent/portsorch.cpp
JipanYanga: done

JipanYanga: done

lguohan:
can you rebase your code and resolve the conflict?

JipanYanga:
I did a rebase against master branch.  Before the possible merge for this PR is done, the libsai from ASIC vendors should support SAI_HOSTIF_VLAN_TAG_STRIP and SAI_HOSTIF_VLAN_TAG_KEEP 
 for attribute SAI_HOSTIF_ATTR_VLAN_TAG  in both sai_create_hostif_fn() and sai_set_hostif_attribute_fn().   

The broadcom libsai I tested has support for them from 3.0.3.3.  Current version of broadcom libsai is sonic-buildimage repo is 3.0.3.2-6 .  https://github.com/Azure/sonic-buildimage/blob/master/platform/broadcom/sai.mk

orchagent/portsorch.cpp
```diff
@@ -413,6 +418,50 @@ bool PortsOrch::getPortPvid(Port &port, sai_uint32_t &pvid)
...
+    {
+        portv.push_back(port);
+    }
+    else if (port.m_type == Port::LAG)
+    {
+        getLagMember(port, portv);
+    }
+    else
+    {
+        SWSS_LOG_ERROR("PortsOrch::setPortPvid port type %d not supported", port.m_type);
```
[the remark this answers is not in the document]
JipanYanga: fixed.

lguohan:
@JipanYanga , 3.0.3.2-6 does not support STRIP/KEEP attribute value, it will cause the T0 to break. For normal VLAN scenario, can we still use STRIP option and setup the pvid in the kernel to make it work? That seems to be the only way I can merge this PR right now.

JipanYanga:
So the official libsai from some ASIC vendors don't have support for  attributes "SAI_HOSTIF_VLAN_TAG_STRIP and SAI_HOSTIF_VLAN_TAG_KEE" in sai_hostif_vlan_tag_t. 
While the two attributes or the better attribute SAI_HOSTIF_VLAN_TAG_ORIGINAL is mandatory for VLAN trunk support.  
We may put "return true;" at beginning of bool PortsOrch::setHostIntfsStripTag(Port &port, sai_hostif_vlan_tag_t strip),  the existing access vlan function won't be affected by the libsai issue.
We may just remove that line of code from the function once libsai support is ready.

orchagent/portsorch.cpp
```diff
@@ -1412,6 +1461,11 @@ bool PortsOrch::addHostIntfs(sai_object_id_t id, string alias, sai_object_id_t &
         return false;
     }
 
+    /*
+     * Port has been removed from 1q bridge at PortsOrch constructor,
+     * also start stripping off VLAN tag.
+     */
+    setHostIntfsStripTag(port, SAI_HOSTIF_VLAN_TAG_STRIP);
```
lguohan: by default, it is stripped, so you do not call this.

orchagent/portsorch.cpp
```diff
@@ -1463,6 +1517,12 @@ bool PortsOrch::addBridgePort(Port &port)
         return false;
     }
 
+    if (!setHostIntfsStripTag(port, SAI_HOSTIF_VLAN_TAG_KEEP))
+    {
+        SWSS_LOG_ERROR("Failed to set %s for hostif of port %s",
+                hostif_vlan_tag[SAI_HOSTIF_VLAN_TAG_KEEP], port.m_alias.c_str());
+        return false;
+    }
```
lguohan: let's move this to addVlanMember(), in the untagged mode, we do not need to call KEEP.

orchagent/portsorch.cpp
```diff
@@ -1490,6 +1550,13 @@ bool PortsOrch::removeBridgePort(Port &port)
         return false;
     }
 
+    if (!setHostIntfsStripTag(port, SAI_HOSTIF_VLAN_TAG_STRIP))
+    {
+        SWSS_LOG_ERROR("Failed to set %s for hostif of port %s",
+                hostif_vlan_tag[SAI_HOSTIF_VLAN_TAG_STRIP], port.m_alias.c_str());
+        return false;
+    }
```
lguohan: move this to removeVlanMember()

orchagent/portsorch.cpp
```diff
@@ -1772,6 +1839,15 @@ bool PortsOrch::addLagMember(Port &lag, Port &port)
 
     m_portList[lag.m_alias] = lag;
 
+    if (lag.m_bridge_port_id > 0)
```
lguohan: also check if the port belongs to multiple vlans or not, if yes then add keep, otherwise do not need to do anything.

lguohan:
retest this please

PR #331, opened 2017-09-30
[portsorch]: Use sai_serialize api to write to DB

marian-pritsak:
Keep object IDs serializeable in database,
so that they can be reused later by client code


marian-pritsak:
Along with https://github.com/Azure/sonic-sairedis/pull/234

stcheng:

lguohan:
can you resolve the conflict.

marian-pritsak:
Done

PR #335, opened 2017-10-04
[pfcwdorch]: Change SW watchdog to notification

marian-pritsak:
Software PFC WD does not implement polling anymore
It registers .lua plugins in syncd and subscribes to
notifications from plugins


marian-pritsak:
Depends on https://github.com/Azure/sonic-swss-common/pull/134

marian-pritsak:
retest this please

lguohan:
what is the relation with https://github.com/Azure/sonic-sairedis/pull/229/files

orchagent/pfcwdorch.cpp
```diff
@@ -320,15 +302,61 @@ void PfcWdOrch<DropHandler, ForwardHandler>::unregisterFromWdDb(const Port& port
...
+    PfcWdOrch<DropHandler,
+    ForwardHandler>(db, tableNames),
+    m_pfcWdDb(new DBConnector(PFC_WD_DB, DBConnector::DEFAULT_UNIXSOCKET, 0)),
+    m_pfcWdTable(new ProducerStateTable(m_pfcWdDb.get(), PFC_WD_STATE_TABLE)),
+    c_portStatIds(portStatIds),
+    c_queueStatIds(queueStatIds)
 {
     SWSS_LOG_ENTER();
+
+    string platform = getenv("platform") ? getenv("platform") : "";
```
lguohan: in /usr/bin/start.sh swss docker

export platform=`sonic-cfggen -v platform`

platform = $onie_platform, not sure if this work.

marian-pritsak:
They don't depend on each other
I still see missing defines from swss-common. Should something be triggered in jenkins to use latest version?

orchagent/pfcwdorch.cpp
marian-pritsak: https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-orchagent/orchagent.sh#L7

marian-pritsak: In orchagent.sh it's exported as onie_switch_asic

orchagent/pfcwdorch.cpp
```diff
@@ -320,15 +302,61 @@ void PfcWdOrch<DropHandler, ForwardHandler>::unregisterFromWdDb(const Port& port
...
+                PfcWdOrch<DropHandler, ForwardHandler>::getCountersDb().get(),
+                detectLuaScript);
+
+        string restoreLuaScript = swss::loadLuaScript(restorePluginName);
+        restoreSha = swss::loadRedisScript(
+                PfcWdOrch<DropHandler, ForwardHandler>::getCountersDb().get(),
+                restoreLuaScript);
+
+        vector<FieldValueTuple> fieldValues;
+        fieldValues.emplace_back(SAI_OBJECT_TYPE, sai_serialize_object_type(SAI_OBJECT_TYPE_QUEUE));
```
lguohan: why SAI_OBJECT_TYPE_QUEUE? why this plugin table is related to SAI_OBJECT_TYPE_QUEUE?

marian-pritsak: Client registers to a specific type of objects. It will receive a list of queue oids for which thread updated counters.

lguohan: see my comments on sonic-sairedis repo, this does not have to reuse SAI OBJECT TYPE, it is plugin table, you can define your own attribute and value.

lguohan:
retest this please

orchagent/pfcwdorch.cpp
```diff
@@ -446,173 +392,109 @@ bool PfcWdSwOrch<DropHandler, ForwardHandler>::startWdOnQueue(sai_object_id_t qu
...
-    }
+    sai_object_id_t queueId = SAI_NULL_OBJECT_ID;
+    sai_deserialize_object_id(queueIdStr, queueId);
 
-    if (!stormedQueues.empty())
+    auto entry = m_entryMap.find(queueId);
+    if (entry == m_entryMap.end())
     {
-        restoreCheckReply = runRedisScript(db, restoreSha, stormedQueues, argv);
+        SWSS_LOG_ERROR("Queue %s is not registered", queueIdStr.c_str());
```
lguohan: should return here since entry is null now.

orchagent/pfc_detect_mlnx.lua
```diff
@@ -0,0 +1,86 @@
...
+
+-- Iterate through each queue
+local n = table.getn(KEYS)
+for i = n, 1, -1 do
+    local counter_keys = redis.call('HKEYS', counters_table_name .. ':' .. KEYS[i])
+    local counter_num = 0
+    local old_counter_num = 0
+    local is_deadlock = false
+    local pfc_wd_status = redis.call('HGET', counters_table_name .. ':' .. KEYS[i], 'PFC_WD_STATUS')
+    if pfc_wd_status == 'operational' then
```
lguohan: why do we have two lua script? It looks like both script iterate through the queue, depending on the pfc_wd_status (operational or not), it does either detection or check if storm is stopped. Why not just have one lua script?

orchagent/pfcwdorch.cpp
marian-pritsak: Done

orchagent/pfcwdorch.cpp
marian-pritsak: I will change in a follow-up PR, since it requires change in swss-common schema

orchagent/pfc_detect_mlnx.lua
marian-pritsak: You are correct. It's left from the previous implementation.

marian-pritsak:
We don't need separate plugins. We needed them before, but not anymore. Is it ok if I'll merge them in a folow-up PR?

lguohan:
ok for me. let's merge the lua script into one in the next pr.

orchagent/pfcwdorch.cpp
```diff
@@ -273,6 +243,17 @@ void PfcWdOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port)
         return;
     }
 
+    if (!c_portStatIds.empty())
+    {
+        vector<FieldValueTuple> fieldValues;
+        string str = counterIdsToStr(c_portStatIds, &sai_serialize_port_stat);
+        fieldValues.emplace_back(PFC_WD_PORT_COUNTER_ID_LIST, str);
+
```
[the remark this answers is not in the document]
lguohan: you have one get_stat to get all port counters

orchagent/pfcwdorch.cpp
```diff
@@ -282,26 +263,27 @@ void PfcWdOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port)
         }
 
         sai_object_id_t queueId = port.m_queue_ids[i];
+        string queueIdStr = sai_serialize_object_id(queueId);
+
+        // Store detection and restoration time for plugins
+        vector<FieldValueTuple> countersFieldValues;
+        countersFieldValues.emplace_back("PFC_WD_DETECTION_TIME", to_string(detectionTime * 1000));
```
sihuihan88: Can we also put "action" in the table?

orchagent/pfcwdorch.cpp
```diff
@@ -253,13 +224,12 @@ void PfcWdOrch<DropHandler, ForwardHandler>::deleteEntry(const string& name)
...
     }
 
-    unregisterFromWdDb(port);
-
     SWSS_LOG_NOTICE("Stopped PFC Watchdog on port %s", name.c_str());
 }
 
 template <typename DropHandler, typename ForwardHandler>
-void PfcWdOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port)
+void PfcWdSwOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port,
```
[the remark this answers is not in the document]
marian-pritsak: Why does HW watchdog need to query counters? It has a native notification. It was a mistake to put it in a base class. Or did I miss something?

orchagent/pfcwdorch.cpp
marian-pritsak: Detection and restoration time is written to DB for plugins so that they can use it as input.
I suspect that it is required for later use to know which action to apply. Maybe it would be better to move PfcWdQueueEntry to the base class and reuse it?

orchagent/pfcwdorch.cpp
```diff
@@ -446,173 +392,110 @@ bool PfcWdSwOrch<DropHandler, ForwardHandler>::startWdOnQueue(sai_object_id_t qu
...
 {
+    SWSS_LOG_ENTER();
+
     DBConnector db(COUNTERS_DB, DBConnector::DEFAULT_UNIXSOCKET, 0);
 
-    // Load script for storm detection
-    string detectScriptName = getStormDetectionCriteria();
-    string detectLuaScript = loadLuaScript(detectScriptName);
-    string detectSha = loadRedisScript(&db, detectLuaScript);
+    swss::Select s;
```
stcheng: add `using namespace swss;` in *.cpp.

orchagent/pfcwdorch.cpp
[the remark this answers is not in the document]
sihuihan88: As I remembered, we need to register SAI_QUEUE_STAT_PACKETS and SAI_QUEUE_STAT_DROPPED_PACKETS to get the forward/drop packets count during recovery for all kinds of pfcwd. Is this part changed?

orchagent/pfcwdorch.cpp
sihuihan88: It's a good idea to move PfcWdQueueEntry to base class. 
We need action for alert-only mode, in which case we need to continuously run detect_plugin even when queue is stormed. It's OK that we don't add it now but we definitely need it in future.

orchagent/orchdaemon.cpp
```diff
@@ -97,7 +97,38 @@ bool OrchDaemon::init()
...
+            SAI_PORT_STAT_PFC_3_RX_PKTS,
+            SAI_PORT_STAT_PFC_4_RX_PKTS,
+            SAI_PORT_STAT_PFC_5_RX_PKTS,
+            SAI_PORT_STAT_PFC_6_RX_PKTS,
+            SAI_PORT_STAT_PFC_7_RX_PKTS,
+        };
+
+        static const vector<sai_queue_stat_t> queueStatIds =
+        {
+            SAI_QUEUE_STAT_PACKETS,
```
sihuihan88: So we don't poll SAI_QUEUE_STAT_PACKETS and SAI_QUEUE_STAT_DROPPED_PACKETS now? Are we going to handle the stats in phase 3?

orchagent/pfcwdorch.cpp
```diff
@@ -446,173 +392,110 @@ bool PfcWdSwOrch<DropHandler, ForwardHandler>::startWdOnQueue(sai_object_id_t qu
...
-            {
-                throw runtime_error("Invalid PFC WD Action");
-            }
-
-            queueEntry.pollTimeLeft = queueEntry.c_restorationTime;
+            entry->second.handler = make_shared<DropHandler>(
+                    entry->second.portId,
+                    entry->first,
+                    entry->second.index,
+                    PfcWdOrch<DropHandler, ForwardHandler>::getCountersTable());
```
sihuihan88: Can we write all PFCWD state back to PFC_WD_STATE_TABLE (including queue status, drop_pkts etc) in phase 3?

orchagent/orchdaemon.cpp
marian-pritsak: We don't poll them for the purpose of gathering statistics because statistics also depends on platform implementation of handlers. Yes, they will be changed in P3 with a common interface to let different handlers gather statistics in a way that suits them.

marian-pritsak:
@sihuihan88 Counters will be moved to P3. First, they will not rely on counters registration, but instead, I'll implement get_stats support. Also as I understand different handlers can have a different way to get counters, like, for example, ACLs have their own counters, so it will require some common interface for getting counters in the future.

marian-pritsak:
Also requires https://github.com/Azure/sonic-swss/pull/340

marian-pritsak:
@lguohan can we merge this?

orchagent/pfc_detect_mlnx.lua
```diff
@@ -0,0 +1,86 @@
...
+            packets_last = tonumber(packets_last)
+            pfc_rx_packets_last = tonumber(pfc_rx_packets_last)
+            pfc_duration_last = tonumber(pfc_duration_last)
+
+            -- Check actual condition of queue being in PFC storm
+            if (occupancy_bytes > 0 and packets - packets_last == 0 and pfc_rx_packets - pfc_rx_packets_last > 0) or
+                -- DEBUG CODE START. Uncomment to enable
+                (debug_storm == "enabled") or
+                -- DEBUG CODE END.
+                (occupancy_bytes == 0 and packets - packets_last == 0 and (pfc_duration - pfc_duration_last) > poll_time * 0.8) then
```
wendani: @marian-pritsak Looks like the value of 'SAI_PORT_STAT_PFC_#_RX_PAUSE_DURATION' return from get_stat is in milliseconds as it is directly compared with poll_time, which is 200 millisecond. Does it comply with the SAI definition of 'SAI_PORT_STAT_PFC_#_RX_PAUSE_DURATION'

PR #338, opened 2017-10-05
Limit FlexPort feature to only Mellanox Platform

lguohan:
For other platform that do not support SAI create/remove_port API,
generate a log message instead of crash.

**What I did**
Limit FlexPort feature (port breakout feature) to only Mellanox Platform

**Why I did it**
Only mellanox platform currently support this feature. Other platform would crash if we do not do this limit.

**How I verified it**
Change port_config.ini and tested the new syslog messages.

**Details if related**

```
Oct  5 05:53:16.002653 str-a7050-acs-1 NOTICE orchagent: :- doPortTask: Failed to remove Port 1000000000022 due to missing SAI remove_port API.
Oct  5 05:53:16.011418 str-a7050-acs-1 NOTICE orchagent: :- doPortTask: Failed to create Port Ethernet124 due to missing SAI create_port API.
```

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -671,10 +671,19 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 {
                     if (m_lanesAliasSpeedMap.find(it->first) == m_lanesAliasSpeedMap.end())
                     {
-                        if (!removePort(it->second))
+                        char *platform = getenv("platform");
+                        if (platform && strstr(platform, MLNX_PLATFORM_SUBSTRING))
                         {
-                            throw runtime_error("PortsOrch initialization failure.");
+                            if (!removePort(it->second))
+                            {
+                                throw runtime_error("PortsOrch initialization failure.");
+                            }
                         }
+                        else
+                        {
+                            SWSS_LOG_NOTICE("Failed to remove Port %lx due to missing SAI remove_port API.", it->second);
+                        }
+
                         it = m_portListLaneMap.erase(it);
                     }
                     else
@@ -685,21 +694,48 @@ void PortsOrch::doPortTask(Consumer &consumer)
 
                 for (auto it = m_lanesAliasSpeedMap.begin(); it != m_lanesAliasSpeedMap.end();)
                 {
+                    bool port_created = false;
+
                     if (m_portListLaneMap.find(it->first) == m_portListLaneMap.end())
                     {
-                        if (!addPort(it->first, get<1>(it->second)))
+                        // work around to avoid syncd termination on SAI error due missing create_port SAI API
+                        // can be removed when SAI redis return NotImplemented error
+                        char *platform = getenv("platform");
+                        if (platform && strstr(platform, MLNX_PLATFORM_SUBSTRING))
+                        {
+                            if (!addPort(it->first, get<1>(it->second)))
+                            {
+                                throw runtime_error("PortsOrch initialization failure.");
+                            }
+
+                            port_created = true;
+                        }
+                        else
+                        {
+                            SWSS_LOG_NOTICE("Failed to create Port %s due to missing SAI create_port API.", get<0>(it->second).c_str());
+                        }
+                    }
+                    else
+                    {
+                        port_created = true;
+                    }
+
+                    if (port_created)
+                    {
+                        if (!initPort(get<0>(it->second), it->first))
                         {
                             throw runtime_error("PortsOrch initialization failure.");
                         }
                     }
 
-                    if (!initPort(get<0>(it->second), it->first))
-                    {
-                        throw runtime_error("PortsOrch initialization failure.");
-                    }
-
                     it = m_lanesAliasSpeedMap.erase(it);
                 }
+            }
+
+            if (!m_portConfigDone)
+            {
+                it = consumer.m_toSync.erase(it);
+                continue;
             }
 
             Port p;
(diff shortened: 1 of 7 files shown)


lguohan:
it is a temp workaround

lguohan:
will plan accordingly.

lguohan:
The flexport feature is defined here, https://github.com/Azure/SONiC/wiki/Port-Breakout-and-Speed 

The design is also available here. https://github.com/Azure/SONiC/wiki/Port-Breakout-High-Level-Design

Please take a look at the feature and let us know your concern.

To your first question, create_port/remove_port is a basic SAI API. It should be supported by every SAI library. Once vendor SAI supports this feature in a couple of month we can remove the check.

JipanYanga:
A general question, is there mechanism at SAI layer to check whether an API and an attribute is supported at running time?   if not it might be good to add that support so we don't need to add platform dependent check in PI code in the future.

As a temp workaround,  I'm ok with the change to quickly get around the syncd crash problem.

lguohan:
yes, sai can check the api, but swss is calling sairedis. we need feedback from sairedis, which needs more time to do it.

lguohan:
I do not understand your point that this commit id does not solve the crash issue. it does solve the crash issue. 

I agree this is not the ideal way of doing it as I mentioned in my reply and the code comment, this is a temp workaround until we find time to do it in a right way.

As for plan for the right fix, I think we can discuss with PM on the timeline and priority. This FlexPort feature does solve the problem that you do not need to have multiple sai profile for the same platform in case of port breakout. I do not see why we need to revert the original PR.

zhenggen-xu:
The crash happened in syncd when calling the API function with null pointer.  We should either force all the SAI library to have valid pointers for all functions (including stub functions in case no implementation of the real functions.), or we have to check the API function pointers validness before calling it. 
PD code in PI functions is not ideal, maybe we can address this by having a valid API map at swss level by query ASIC SAI lib during the init, then we decide whether we should call the API or not.

Also, fundamentally, we do need address the case where we got failure from SAI lib call as well ( resource failure etc...)

lguohan:
@zhenggen-xu , the SAI api in swss level is the sairedis function pointer.  currently swss does not have a way to get the actually SAI API function pointer. 

@nikos-li, let's focus on this specific issue first. Let's plan to the longer term fix. The feature is useful, and it does not cause crash on other platforms. Do we really want to block it?

zhenggen-xu:
@lguohan Understood the swss using sairedis function pointer.  what I meant was to have a map which will be filled in during the initialization so we know whether the ASIC SAI API is valid or not. The other problem is, even if the SAI API function pointer is valid but not implemented, syncd would not be able to feedback to swss for more action currently. 
Given that we may not be able to address this in a short time, I think we can keep this change, but at the same time, I suggest to log a few bugs:
1,  SAI API function pointer validness should be checked before calling sairedis API. This might be done by some query at init time.
2, SAI API failure feedback need to be done, we can not just crash swss if SAI API fails.
3, A bug should be logged so that we don't forget to remove the PD code in PI area later.

orchagent/portsorch.cpp
```diff
@@ -685,23 +694,50 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
                     if (m_portListLaneMap.find(it->first) == m_portListLaneMap.end())
                     {
-                        if (!addPort(it->first, get<1>(it->second)))
+                        // work around to avoid syncd termination on SAI error due missing create_port SAI API
+                        // can be removed when SAI redis return NotImplemented error
+                        char *platform = getenv("platform");
+                        if (platform && strstr(platform, MLNX_PLATFORM_SUBSTRING))
                         {
-                            throw runtime_error("PortsOrch initialization failure.");
+                            if (!addPort(it->first, get<1>(it->second)))
```
JipanYanga: addPort() should be called for all platforms as before. Or I missed something here?

JipanYanga: Never mind, I mistook it with addHostIntfs().

orchagent/portsorch.cpp
```diff
@@ -685,23 +694,50 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
 
                     it = m_lanesAliasSpeedMap.erase(it);
                 }
             }
 
+            if (!m_portConfigDone)
+            {
+                it = consumer.m_toSync.erase(it);
+                continue;
+            }
```
JipanYanga: Not sure about this change.   So all port speed/admin_status/mtu set request for any specific port will be dropped before m_portConfigDone is set?

lguohan: this solves another bug introduced by previous PR. we do not process mtu, admin_status, speed before m_portConfigDone is set to true. It only make sense after initPort() is called.

lguohan:
address issue https://github.com/Azure/sonic-sairedis/issues/235

lguohan:
@zhenggen-xu , issues created. https://github.com/Azure/sonic-sairedis/issues/239 and https://github.com/Azure/sonic-sairedis/issues/238

orchagent/portsorch.cpp
nikos-li: Please add a comment to that effect.

JipanYanga: There might be issue here,  currently port speed/admin_status/mtu "SET" operation in orchagent is triggered by Linux netlink message. If the request is dropped with "consumer.m_toSync.erase(it)",   it may never get set again?    There are multiple ports,  some ports may get netlink message before other ports get created in some extreme scenarios.

lguohan: it won't. the linux netlink messages will only come apter m_portConfigDone.

JipanYanga:
If we are ok with the workaround and the plan of next step,  could we merge the code change asap to avoid syncd segmentation fault crash under this specific scenario?  thanks!

lguohan:
Let me know if I can merge now?

nikos-li:
@lguohan I am not blocking this on the assumption that the PD checks you introduced will be removed by end of year.

PR #340, opened 2017-10-09
[portsorch]: Add queue-port and queue-index map

marian-pritsak:


marian-pritsak:
This change is required for PFC WD. It will allow a quick lookup with queue OID as a key.

PR #341, opened 2017-10-10
[pfcwdorch]: Allow unlimited restoration time

marian-pritsak:
In case if user does not pass restoration time, it is assumed that queue
will be mitigated indefinitely in an event of PFC storm until user disables
watchdog on that port.


lguohan:
@marian-pritsak , can you resolve the conflict?

PR #342, opened 2017-10-10
[pfcwdorch]: Add alert mode

marian-pritsak:
Alert mode tells watchdog that whenever it detects PFC storm, it should
only write message to the log. It is implemented as a separate action
along with drop and forward.

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -97,6 +97,7 @@ PfcWdAction PfcWdOrch<DropHandler, ForwardHandler>::deserializeAction(const stri
     {
         { "forward", PfcWdAction::PFC_WD_ACTION_FORWARD },
         { "drop", PfcWdAction::PFC_WD_ACTION_DROP },
+        { "alert", PfcWdAction::PFC_WD_ACTION_ALERT },
     };
 
     if (actionMap.find(key) == actionMap.end())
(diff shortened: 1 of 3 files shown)


orchagent/pfcwdorch.cpp
```diff
@@ -97,6 +97,7 @@ PfcWdAction PfcWdOrch<DropHandler, ForwardHandler>::deserializeAction(const stri
     {
         { "forward", PfcWdAction::PFC_WD_ACTION_FORWARD },
         { "drop", PfcWdAction::PFC_WD_ACTION_DROP },
+        { "alert", PfcWdAction::PFC_WD_ACTION_ALERT },
```
stcheng: is it better to move this const into the header file?

marian-pritsak: It is not intended to be used by anyone else. Enum is in header.

sihuihan88:
In this pr, in alert mode, it will run restore plugins for stormed queues, which is not correct. The detect plugin needs to check stormed queues in alert mode. https://github.com/Azure/sonic-swss/pull/343 includes the enhancement.

orchagent/pfcwdorch.cpp
stcheng: then a const static is better? do you know if the compiler will optimize here to only initialize the variable once?

qiluo-msft: const static will be better here. Move out of this function, and make it a cpp global is even better.

stcheng: @qiluo-msft thanks for the explanation.

PR #343, opened 2017-10-10
[PFCWD]: Add alert mode broadcom support

sihuihan88:

**What I did**
1. Enhance alert mode for PFCWD: in alert mode, pfcwd will keep using PFC storm detection criteria to see whether storm happens. If storm happens, pfcwd sends the syslog message, and keeps using detection criteria to monitor the queue status, and send syslog message when the storm stops.

2. Add initial support for broadcom PFCWD: current broadcom PFCWD will only support alert mode and rely on software implementation to detect storm. Will add support for drop and storm recovery in future.

 
**Why I did it**
Enhance PFCWD feature.
**How I verified it**
Test on the broadcom DUT, it behaves as desired. 
**Details if related**


sihuihan88:
Depends on https://github.com/Azure/sonic-swss-common/pull/141 and https://github.com/Azure/sonic-sairedis/pull/241

orchagent/pfc_detect_broadcom.lua
```diff
@@ -0,0 +1,91 @@
...
+                (occupancy_bytes == 0 and pfc_rx_packets - pfc_rx_packets_last > 0 and pfc_on2off - pfc_on2off_last == 0 and queue_pause_status_last == 'true' and queue_pause_status == 'true') then
+                if time_left <= poll_time then
+                    redis.call('PUBLISH', 'PFC_WD', '["' .. KEYS[i] .. '","storm"]')
+                    is_deadlock = true
+                    time_left = detection_time
+                else
+                    time_left = time_left - poll_time
+                end
+            else
+                if pfc_wd_action == 'alert' and pfc_wd_status ~= 'operational' then
```
marian-pritsak: Plugins do not distinguish between actions. Two types of notifications are "storm" and "restore". Orchagent decides what to do based on configuration.

orchagent/orchdaemon.cpp
```diff
@@ -124,11 +124,54 @@ bool OrchDaemon::init()
...
+            SAI_QUEUE_STAT_PACKETS,
+            SAI_QUEUE_STAT_CURR_OCCUPANCY_BYTES,
+        };
+
+        static const vector<sai_queue_attr_t> queueAttrIds =
+        {
+            SAI_QUEUE_ATTR_PAUSE_STATUS,
+        };
+
+        m_orchList.push_back(new PfcWdSwOrch<PfcWdActionHandler, PfcWdActionHandler>(
```
marian-pritsak: I think you can reuse forward handler

orchagent/pfc_restore_broadcom.lua
```diff
@@ -0,0 +1,66 @@
+-- KEYS - queue IDs
```
marian-pritsak: Is there any difference from mlnx plugin? If not, maybe it's better to rename the original to some general name and create symlinks?

orchagent/pfcwdorch.cpp
```diff
@@ -108,6 +111,27 @@ PfcWdAction PfcWdOrch<DropHandler, ForwardHandler>::deserializeAction(const stri
 }
 
 template <typename DropHandler, typename ForwardHandler>
+string PfcWdOrch<DropHandler, ForwardHandler>::serializeAction(const PfcWdAction& action)
```
marian-pritsak: Plugins don't have to know which action type it is. It's orchagent's responsibility to make such decision.

orchagent/pfcwdorch.cpp
```diff
@@ -312,51 +344,50 @@ PfcWdSwOrch<DropHandler, ForwardHandler>::PfcWdSwOrch(
...
     string platform = getenv("platform") ? getenv("platform") : "";
 
-    if (platform == "")
+    if (platform == MLNX_PLATFORM_SUBSTRING)
     {
-        SWSS_LOG_ERROR("Platform environment variable is not defined");
-        return;
+        SWSS_LOG_NOTICE("Current platform is mlnx");
+    }
+    else
```
marian-pritsak: Why else? We must not default to any platform.
And I don't really understand why we need this kind of notice message.

orchagent/pfcwdorch.cpp
```diff
@@ -312,51 +344,50 @@ PfcWdSwOrch<DropHandler, ForwardHandler>::PfcWdSwOrch(
...
+    {
+        platform = BROADCOM_PLATFORM;
+        SWSS_LOG_NOTICE("Current platform is broadcom");
     }
 
     string detectSha, restoreSha;
     string detectPluginName = "pfc_detect_" + platform + ".lua";
     string restorePluginName = "pfc_restore_" + platform + ".lua";
 
-    try
```
marian-pritsak: Do you want not to handle it?
If we'll fail to load .lua script due to, for example, syntax error orchagent will crash.

marian-pritsak:
@sihuihan88 In general, if PFCWD does not support some action for a specific platform, you can just use the base class which does nothing except logging.

orchagent/orchdaemon.cpp
```diff
@@ -124,11 +124,54 @@ bool OrchDaemon::init()
...
+
         m_orchList.push_back(new PfcWdSwOrch<PfcWdZeroBufferHandler, PfcWdLossyHandler>(
                     m_applDb,
                     pfc_wd_tables,
                     portStatIds,
-                    queueStatIds));
+                    queueStatIds,
+                    queueAttrIds));
+    }
+    else
```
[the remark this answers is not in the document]
sihuihan88: getenv("platform") is empty for broadcom platform. Before we fix that issue, let's do this way.

orchagent/pfc_detect_broadcom.lua
sihuihan88: Currently, the detect plugin checks queues in operational status, and restore plugin checks queues in stormed status. This is not true for the alert mode. In alert mode, the detect plugin will check all the queues regardless of the queue status, and restore plugin never check anything. Currently the logic of when/which queue to check is included in the plugin itself. That's the reason why plugin also needs to know the action  type to decide which queue to check for alert-mode and normal case. In https://github.com/Azure/sonic-swss/pull/342, it still runs restore plugins for stormed queues in alert mode, which is not correct.

orchagent/pfc_restore_broadcom.lua
sihuihan88: I think we will combine the detect/restore plugin into a single file in future. It might be better to separate the lua scripts for different platform for now.

orchagent/pfcwdorch.cpp
sihuihan88: As above comments, detect plugin need to detect stormed queues in alert mode instead of restore plugin. This logic need to be added into lua script based on current implementation.

orchagent/pfcwdorch.cpp
sihuihan88: Removed notice message.

orchagent/orchdaemon.cpp
sihuihan88: Currently both drop and forward uses base class for simplicity. I agree we might reuse forward handler. This PR is for alert mode. So will update drop/forward handler together in a separate PR.

orchagent/pfcwdorch.cpp
sihuihan88: I agree it might be better to catch the exception to prevent orchagent from crashing. Update the change.

orchagent/orchdaemon.cpp
marian-pritsak: ok

orchagent/orchdaemon.cpp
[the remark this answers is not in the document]
marian-pritsak: I see. In this case, maybe some kind of "TODO" comment would be appropriate with the explanation why Broadcom is a default?

orchagent/pfc_restore_broadcom.lua
marian-pritsak: Well, now that we see that restoration criteria are common, it might be a good idea to leave them apart. @lguohan what do you think?

marian-pritsak:
@stcheng are you aware of any github bot that will do such style checks for every new PR? It might significantly simplify a reviewer's job.

orchagent/pfc_detect_broadcom.lua
marian-pritsak: According to the document, definition of alert mode is following:
PFC watchdog logs when the action of dropping packets when pause storm occurs and resume forwarding traffic when pause storm stops. With the alert mode, PFC watchdog only need to log that pause storm detected and pause storm stopped.

So as I understand, alert mode is same as other actions, except it doesn't apply any configuration, but only logs events.

@lguohan can you please take a look? Maybe I'm missing something?

stcheng:
@marian-pritsak we might need to add some checks during the build time. e.g. cpplint

orchagent/pfc_detect_broadcom.lua
sihuihan88: The restore criteria is that we don't receive PFC any more. This actually bases on the assumption that we don't send any packets because of the recovery action we have taken. In alert mode, we are not doing anything but log the events. In such case, restore criteria is too strong as it's OK to occasionally receive PFC in normal case. That is also the reason why we need to use the detection criteria to see whether the storm is stopped in such case.

orchagent/orchdaemon.cpp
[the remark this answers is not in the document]
sihuihan88: Will do

orchagent/pfc_restore_broadcom.lua
sihuihan88: I see your point. Will use pfc_restore.lua for both platforms.

orchagent/pfc_detect_broadcom.lua
marian-pritsak: ok

sihuihan88:
retest this please

PR #325, opened 2017-09-28
Move ACL and Everflow configuration to Config DB.

oleksandrivantsiv:
**What I did**
- Add possibility to subscribe *Orch to both APP DB and Config DB
- Rename ACL and Everflow tables.
- Fix issue with DBConnector obbject multiple deletion.

**Why I did it**
All changes are done according to ACL dynamic configuration design.

**How I verified it**
- Run BGP and FIB tests.
- Run ACL tables and Rule manual tests with "acl-loader" utility.
- Run Everflow manual tests.

PR #349, opened 2017-10-13
[acl]: Allow ACL table bind to LAGs and VLANs.

oleksandrivantsiv:
**What I did**
Allowed to bind ACL table to LAGs and VLANs.
Denied to bind ACL table to LAG members.
Fixed issue with ACL table type and ports validation.
**Why I did it**
All changes are done in the scope of ACL dynamic configuration feature.
**How I verified it**
Manual tests
Run ACL testbed tests

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -878,10 +878,39 @@ bool AclTable::create()
           SAI_ACL_RANGE_TYPE_L4_SRC_PORT_RANGE
         };
 
-    attr.id = SAI_ACL_TABLE_ATTR_ACL_BIND_POINT_TYPE_LIST;
+    set<sai_acl_bind_point_type_t> binds;
+    for (const auto& portid_pair : ports)
+    {
+        Port port;
+        if (!gPortsOrch->getPort(portid_pair.first, port))
+        {
+         continue;
+        }
+
+        switch (port.m_type)
+        {
+        case Port::PHY:
+            binds.insert(SAI_ACL_BIND_POINT_TYPE_PORT);
+            break;
+        case Port::VLAN:
+            binds.insert(SAI_ACL_BIND_POINT_TYPE_VLAN);
+            break;
+        case Port::LAG:
+            binds.insert(SAI_ACL_BIND_POINT_TYPE_LAG);
+            break;
+        default:
+            return SAI_STATUS_FAILURE;
+        }
+    }
+
     vector<int32_t> bpoint_list;
-    bpoint_list.push_back(SAI_ACL_BIND_POINT_TYPE_PORT);
-    attr.value.s32list.count = 1;
+    for (auto bind : binds)
+    {
+        bpoint_list.push_back(bind);
+    }
+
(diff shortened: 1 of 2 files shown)


oleksandrivantsiv:
Please do not merge. Depends on https://github.com/Azure/sonic-swss/pull/325

orchagent/port.cpp
```diff
@@ -18,12 +20,15 @@ namespace swss {
...
+
     sai_object_id_t groupOid;
 
     // If port ACL table group does not exist, create one
     if (m_acl_table_group_id == 0)
     {
-        sai_object_id_t bp_list[] = { SAI_ACL_BIND_POINT_TYPE_PORT };
+        int32_t bp_list[] = {SAI_ACL_BIND_POINT_TYPE_PORT,
+                SAI_ACL_BIND_POINT_TYPE_LAG,
+                SAI_ACL_BIND_POINT_TYPE_VLAN};
```
lguohan: depending on the port type (phy/lag/vlan), we can select the BIND_POINT_TYPE here. We do not need to add all bind point type in the SAI_ACL_TABLE_GROUP_ATTR_ACL_BIND_POINT_TYPE_LIST. It is not necessary. On some platforms, it can cost us to use more resources.

orchagent/port.cpp
```diff
@@ -33,8 +38,8 @@ sai_status_t Port::bindAclTable(sai_object_id_t& group_member_oid, sai_object_id
         group_attrs.push_back(group_attr);
 
         group_attr.id = SAI_ACL_TABLE_GROUP_ATTR_ACL_BIND_POINT_TYPE_LIST;
-        group_attr.value.objlist.count = 1;
-        group_attr.value.objlist.list = bp_list;
+        group_attr.value.s32list.count = sizeof(bp_list)/sizeof(*bp_list);
+        group_attr.value.s32list.list = bp_list;
```
lguohan: I am not sure it is necessary to add all bind point types when you creating acl table group for this port. Since the port has only one type, we should add appropriate port type to the bind point type list. Only 1 type is needed.

nikos-li:
@xinliu-seattle @lguohan 
Have you added to the existing test framework for ACLs, test cases for this enhancement?
If so, can you point to the review of those test cases?

oleksandrivantsiv:
@nikos-li https://github.com/Azure/sonic-mgmt/pull/370

orchagent/aclorch.cpp
```diff
@@ -1585,13 +1614,26 @@ bool AclOrch::processPorts(string portsList, std::function<void (sai_object_id_t
...
+            break;
+        case Port::LAG:
+            inserter(port.m_lag_id);
+            break;
+        case Port::VLAN:
+            inserter(port.m_vlan_info.vlan_oid);
+            break;
+        default:
+          SWSS_LOG_ERROR("Failed to process port. Incorrect port %s type %d", alias.c_str(), port.m_type);
+          return false;
```
lguohan: minor alignment issue?

lguohan:
@oleksandrivantsiv , can you resolve the conflict here?

PR #350, opened 2017-10-13
[Pfcwd]: Add pfcAclHandler to drop ingress packets by acl rules

sihuihan88:

**What I did**
Add pfcAclHandler to drop ingress packets by acl rules
**Why I did it**
To drop ingress packets during PFC storm recovery with acl rules.
**How I verified it**
Test on DUT.
**Details if related**

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -19,6 +19,8 @@ extern sai_object_id_t             gSwitchId;
 PortsOrch *gPortsOrch;
 /* Global variable gFdbOrch declared */
 FdbOrch *gFdbOrch;
+/*Global variable gAclOrch declared*/
+AclOrch *gAclOrch;
 
 OrchDaemon::OrchDaemon(DBConnector *applDb) :
         m_applDb(applDb)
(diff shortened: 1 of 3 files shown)


lguohan:
@qiluo-msft to review

lguohan:
should create two acl table each one block one TC, and when deadlock is detected, then apply the table to the corresponding port.

orchagent/pfcactionhandler.cpp
```diff
@@ -133,6 +133,59 @@ void PfcWdActionHandler::updateWdCounters(const string& queueIdStr, const PfcWdQ
     m_countersTable->set(queueIdStr, resultFvValues);
 }
 
+PfcWdAclHandler::PfcWdAclHandler(sai_object_id_t port, sai_object_id_t queue,
+        uint8_t queueId, shared_ptr<Table> countersTable):
+    PfcWdActionHandler(port, queue, queueId, countersTable)
+{
+    acl_table_type_t table_type = ACL_TABLE_L3;
+
+    m_aclTable = "Table_" + sai_serialize_object_id(port) +"_" + sai_serialize_object_id(queue);
```
qiluo-msft: >sai_serialize_object_id [](start = 28, length = 23)

serialize the same thing twice.

qiluo-msft: inconsistent blank style

orchagent/pfcactionhandler.cpp
```diff
@@ -133,6 +133,59 @@ void PfcWdActionHandler::updateWdCounters(const string& queueIdStr, const PfcWdQ
...
+    newTable.id = m_aclTable;
+    gAclOrch->addAclTable(newTable, m_aclTable);
+}
+
+void PfcWdAclHandler::createPfcAclRule(shared_ptr<AclRuleL3> rule, uint8_t queueId)
+{
+    string attr_name, attr_value;
+
+    attr_name = "RULE_PRIORITY";
+    attr_value = "999";
```
qiluo-msft: Use ctor directly

qiluo-msft:
@lguohan Could you clarify "should create two acl table each one block one TC"?
In my understanding, there should be one ACL table per queue. Then bind or unbind ACL table with the queue.

lguohan:
you cannot bind acl tabl to the queue. there  is no sai api for that, you can only bind acl table to the port.

orchagent/orchdaemon.cpp
```diff
@@ -19,6 +19,8 @@ extern sai_object_id_t             gSwitchId;
 PortsOrch *gPortsOrch;
 /* Global variable gFdbOrch declared */
 FdbOrch *gFdbOrch;
+/*Global variable gAclOrch declared*/
+AclOrch *gAclOrch;
```
andriymoroz-mlnx: no need to make it global
if PFC needs AclOrch, it can receive it via constructor as all other components do
For example, RouteOrch receives pointer to NeighborOrch:
`new RouteOrch(m_applDb, APP_ROUTE_TABLE_NAME, neigh_orch);`

Pfc is created in lines 129 and 169

qiluo-msft: I think it is ok. gAclOrch is used by PfcWdAclHandler, whose objects are short-living.

qiluo-msft:
Followed up with another PR.

PR #352, opened 2017-10-17
[switchorch]:support changing mac aging time dynamically.

gongjianLhr:
Signed-off-by: Haoran Liu gongjian.lhr@alibaba-inc.com

**Add codes in switchorch to support changing mac aging time dynamically**


lguohan:
do you have test for mac aging feature?

gongjianLhr:
Hi lguohan @lguohan this pull request is only about the consumer side of setting the mac aging time(reading from APP_DB).  The cli (writing to CONFIG_DB) and the enforcer(reading from CONFIG_DB and writing to APP_DB) are still absent.
I have developed the cli and tested it in our lab.
But the enforcer is based on liuqu's pull request of hostCFGD of aaa cli.
Since you haven't accept the hostCFGD design. It just blocked me from pulling the cli of setting mac aging time.
Considing that there is no cli to set the mac aging time, there is no way to test it.
So should I wait for liuqu's pull request to be accepted ? Or you accept without test and later I add the test after the hostCFGD is accessiable?

PR #355, opened 2017-10-21
Add cfgOrch class

JipanYanga:
Signed-off-by: Jipan Yang <jipan.yang@alibaba-inc.com>

**What I did**
Add cfgOrch class to be used by cfgDB enforcers

**Why I did it**
Common function for  cfgDB enforcers

**How I verified it**
jipan@27bcdfdac3ca:/sonic/src/sonic-swss/tests$ ./tests 
Running main() from gtest_main.cc
[==========] Running 9 tests from 2 test cases.
[----------] Global test environment set-up.
[----------] 8 tests from swssnet
[ RUN      ] swssnet.copy1_v6
[       OK ] swssnet.copy1_v6 (0 ms)
[ RUN      ] swssnet.copy1_v4
[       OK ] swssnet.copy1_v4 (0 ms)
[ RUN      ] swssnet.copy2_v6
[       OK ] swssnet.copy2_v6 (0 ms)
[ RUN      ] swssnet.copy2_v4
[       OK ] swssnet.copy2_v4 (0 ms)
[ RUN      ] swssnet.copy3_v6
[       OK ] swssnet.copy3_v6 (0 ms)
[ RUN      ] swssnet.copy3_v4
[       OK ] swssnet.copy3_v4 (0 ms)
[ RUN      ] swssnet.subnet_v6
[       OK ] swssnet.subnet_v6 (0 ms)
[ RUN      ] swssnet.subnet_v4
[       OK ] swssnet.subnet_v4 (0 ms)
[----------] 8 tests from swssnet (0 ms total)

[----------] 1 test from CfgOrch
[ RUN      ] CfgOrch.test
Starting cfgOrch testing
- Step 1. Provision TEST_CONFIG_DB
- Step 2. Verify TEST_APP_DB content
- Step 3. Flush TEST_APP_DB
- Step 4. Sync from TEST_CONFIG_DB
- Step 5. Verify TEST_APP_DB content
- Step 6. Clean TEST_CONFIG_DB
- Step 7. Verify TEST_APP_DB content is empty
Done.
[       OK ] CfgOrch.test (11023 ms)
[----------] 1 test from CfgOrch (11023 ms total)

[----------] Global test environment tear-down
[==========] 9 tests from 2 test cases ran. (11023 ms total)
[  PASSED  ] 9 tests.
jipan@27bcdfdac3ca:/sonic/src/sonic-swss/tests$ 

**Details if related**

PR #344, opened 2017-10-10
Add support for lag state in stateDB

JipanYanga:
Create an entry in stateDB for each LAG created, and delete the entry when lag removed

Signed-off-by: Jipan Yang <jipan.yang@alibaba-inc.com>

PR #336, opened 2017-10-04
[portsyncd]: Add producer for state database updates

stcheng:
- Add m_stateTableProducer to write to the state database once the ports are ready
- Remove suffix Consumer in some variables since they are not 'consumers'

Signed-off-by: Shu0T1an ChenG <shuche@microsoft.com>

PR #356, opened 2017-10-21
configDB enforcer for VLAN

JipanYanga:

**What I did**
Configuration enforcer for VLAN

**Why I did it**
Support vlan configuration via configDB for both initialization and incremental update
**How I verified it**
VLAN trunk test plan: https://github.com/Azure/SONiC/wiki/VLAN-trunk-test-plan
**Details if related**


JipanYanga:
Has dependency on:
https://github.com/Azure/sonic-swss-common/pull/146
https://github.com/Azure/sonic-swss/pull/355
https://github.com/Azure/sonic-swss/pull/336
https://github.com/Azure/sonic-swss/pull/344

lguohan:
@taoyl-ms , @stcheng to review

Makefile.am
```diff
@@ -1,4 +1,4 @@
-SUBDIRS = fpmsyncd neighsyncd intfsyncd portsyncd orchagent swssconfig
+SUBDIRS = fpmsyncd neighsyncd intfsyncd portsyncd orchagent swssconfig vlanconfd
```
lguohan: i prefer vlanmgr than vlanconfd

JipanYanga: There is already bgpconfd.  We may change to vlanmgr from vlanconfd, then it introduces name inconsistency.

As suggested by @qiluo-msft in https://github.com/Azure/sonic-swss/pull/357,  probably we may merge the files of vlan configuration enforcer and intf configuration enforcer into one flat folder while keeping them in separate processes,  other future configuration enforcer (in c++) could follow the same practice. 

I don't have strong opinion about the naming of configuration enforcer folder and configuration enforcers themselves, but to avoid multiple iterations of name change and folder rearrangement, could @lguohan @taoyl-ms @qiluo-msft  @stcheng agree on the following names:

folder name:  cfgagent
vlan configuration enforcer:  vlanconfd
intf configuration enforcer:    intfconfd
 
Or please help give names (and folder structure) you all agree on, thanks!

lguohan:
can you seperate the vlanmgr and switchmgr into two PR?

cfgagent/vlanconf.cpp
```diff
@@ -0,0 +1,475 @@
...
+    string cmd, res;
+
+    cmd = "bridge vlan add vid " + to_string(vlan_id) + " dev "
+            + DOT1Q_BRIDGE_NAME + " self";
+    swss::exec(cmd, res);
+    cmd = "ip link add link ";
+    cmd += DOT1Q_BRIDGE_NAME;
+    cmd += " name ";
+    cmd += VLAN_PREFIX + to_string(vlan_id)
+            + " type vlan id " + to_string(vlan_id);
```
lguohan: can we contruct the command in one line, easier to ready.

cfgagent/vlanconf.cpp
```diff
@@ -0,0 +1,475 @@
...
+    {
+        cmd = "bridge vlan add vid " + to_string(vlan_id) + " dev "
+            + port_alias;
+    }
+    swss::exec(cmd, res);
+    // Apply switch level flood control to this port
+    gSwtichConfVlan->updateHostFloodControl(port_alias);
+
+    // Bring up vlan member port and set MTU to 9100 by default
+    cmd = "ip link set " + port_alias + " up mtu 9100";
```
lguohan: 9100 hard coded? do we have that value in config db?

cfgagent/vlanconf.cpp
JipanYanga: I had problem compiling it if put them in one line. Will check again to see how to around the problem.

cfgagent/vlanconf.cpp
JipanYanga: It is hard coded in interfaces.j2,  not in config DB.  Probably port schema in configDB should be updated to have MTU field.

127.0.0.1:6379[4]> hgetall "PORT|Ethernet20"
1) "alias"
2) "tenGigE20"
127.0.0.1:6379[4]> hgetall "VLAN_MEMBER|Vlan1000|Ethernet20"
1) "tagging_mode"
2) "untagged"

JipanYanga:
@lguohan Ok, will take switch related configuration to a separate PR

cfgagent/vlanconf.cpp
prsunny: I think "stringstream" would be a better option here and it takes different types of input so you wudn't have to convert integers to_string

cfgagent/vlanconf.h
```diff
@@ -0,0 +1,44 @@
...
+
+namespace swss {
+
+class VlanConf : public OrchBase
+{
+public:
+    VlanConf(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, vector<string> tableNames);
+    void syncCfgDB();
+private:
+	ProducerStateTable m_appVlanTableProducer, m_appVlanMemberTableProducer;
```
lguohan: >	P [](start = 0, length = 2)

can you change all tab to spaces?

cfgagent/vlanconfd.cpp
```diff
@@ -0,0 +1,88 @@
...
+int main(int argc, char **argv)
+{
+    Logger::linkToDbNative("vlanconfd");
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_NOTICE("--- Starting vlanconfd ---");
+
+    string mac_str;
+    swss::exec("ip link show eth0 | grep ether | awk '{print $2}'", mac_str);
+    gMacAddress = mac_str;
```
lguohan: can we make this as an parameter instead of hard code the vlan mac address into the code logic, see -m option in orchagent?

cfgagent/vlanconf.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+bool VlanConf::isMemberStateOk(string &alias)
+{
+    vector<FieldValueTuple> temp;
+
+    if (!alias.compare(0, strlen(LAG_PREFIX), LAG_PREFIX))
+    {
+        if (m_stateLagTable.get(alias, temp))
+        {
+            SWSS_LOG_DEBUG("%s is ready\n", alias.c_str());
```
lguohan: >\n [](start = 39, length = 2)

remove this '\n', does dot look like we need it.

cfgagent/vlanconf.h
JipanYanga: ok

cfgagent/vlanconfd.cpp
JipanYanga: SWITCH_ATTR in SWITCH table has switch_mac field, to implement mac fetching from configDB SWITCH.  mac could be be taken from there if configured, otherwise eth0 mac is used by default.  
mac address here is to set on VLAN.

cfgagent/vlanconf.cpp
JipanYanga: ok

cfgagent/vlanconf.cpp
JipanYanga: Using stringstream looks like a good idea.

cfgagent/vlanconf.cpp
```diff
@@ -0,0 +1,475 @@
...
+bool VlanConf::removeHostVlanMember(int vlan_id, string &port_alias)
+{
+    string cmd, res;
+
+    cmd = "bridge vlan del vid " + to_string(vlan_id) + " dev "
+            + port_alias;
+    swss::exec(cmd, res);
+
+    // When port is not member of any VLAN, it shall be detached from Dot1Q bridge!
+    cmd = "bridge vlan show dev " + port_alias + " | grep None";
```
lguohan: we need to use the full path for all the commands, otherwise those commands could be override by users.

cfgagent/vlanmgr.cpp
```diff
@@ -0,0 +1,470 @@
...
+                SWSS_LOG_DEBUG("%s not ready, delaying", kfvKey(t).c_str());
+                it++;
+                continue;
+            }
+            string tagging_mode = "untagged";
+
+            for (auto i : kfvFieldsValues(t))
+            {
+                if (fvField(i) == "tagging_mode")
+                    tagging_mode = fvValue(i);
```
lguohan: always have {}

cfgagent/vlanmgr.cpp
```diff
@@ -0,0 +1,470 @@
...
+void VlanMgr::doTask(Consumer &consumer)
+{
+    SWSS_LOG_ENTER();
+
+    string table_name = consumer.m_consumer->getTableName();
+
+    if (table_name == CFG_VLAN_TABLE_NAME)
+        doVlanTask(consumer);
+    else if (table_name == CFG_VLAN_MEMBER_TABLE_NAME)
+        doVlanMemberTask(consumer);
```
lguohan: always have {}

cfgagent/vlanconf.cpp
JipanYanga: right.  I'm thinking of having a common header file with all the commands macro defined,  full path is specified in the macro definition.

cfgagent/vlanmgr.cpp
JipanYanga: ok

cfgagent/vlanmgr.cpp
JipanYanga: ok

cfgagent/vlanmgr.cpp
```diff
@@ -0,0 +1,470 @@
...
+        if (consumer.m_toSync.find(member_key) == consumer.m_toSync.end())
+        {
+            vector<FieldValueTuple> fvVector;
+            FieldValueTuple t("tagging_mode", "untagged");
+            fvVector.push_back(t);
+            consumer.m_toSync[member_key] = make_tuple(member_key, SET_COMMAND, fvVector);
+            SWSS_LOG_DEBUG("%s", (dumpTuple(consumer, consumer.m_toSync[member_key])).c_str());
+        }
+        /*
+         * There is pending task from consumber pipe, in this case just skip it.
```
lguohan: consumer

cfgagent/vlanmgr.cpp
```diff
@@ -0,0 +1,494 @@
...
+    cmd.str("");
+    cmd << ECHO_CMD << " 1 > /sys/class/net/" << DOT1Q_BRIDGE_NAME << "/bridge/vlan_filtering";
+    EXEC_WITH_ERROR_THROW(cmd.str(), res);
+
+    cmd.str("");
+    cmd << BRIDGE_CMD << " vlan del vid " << DEFAULT_VLAN_ID << " dev " << DOT1Q_BRIDGE_NAME << " self";
+    EXEC_WITH_ERROR_THROW(cmd.str(), res);
+}
+
+void VlanMgr::syncCfgDB()
```
lguohan: >syncCfgDB [](start = 14, length = 9)

where is function gets called?

JipanYanga:
has dependency on https://github.com/Azure/sonic-swss-common/pull/156

lguohan:
retest this please

lguohan:
retest this please

lguohan:
Azure/sonic-swss-common#156 is merged, but build still failed.

JipanYanga:
Sorry, too many iterations of name change and folder restructuring,  I missed OrchBase to Orch name change in few places for last commit. Please check it again.

lguohan:
> const string VLAN_PREFIX = "Vlan";

this can be removed. minor.

---
Refers to: portsyncd/linksync.cpp:26 in f67e288. [](commit_id = f67e2883879fe9673822b33789c15b32a784fed3, deletion_comment = False)

JipanYanga:
Removed const string VLAN_PREFIX = "Vlan"; from portsyncd/linksync.cpp
Also the unused g_vlanMap variable form linksync.cpp and portsyncd.cpp

cfgmgr/vlanmgrd.cpp
```diff
@@ -0,0 +1,107 @@
...
+int main(int argc, char **argv)
+{
+    Logger::linkToDbNative("vlanmgrd");
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_NOTICE("--- Starting vlanmgrd ---");
+
+    string mac_str;
+    stringstream cmd;
+    cmd << IP_CMD << " link show eth0 | " << GREP_CMD << " ether | " << AWK_CMD << " '{print $2}'";
```
lguohan: we need to use the -m option to get the mac address since it is complicated. 

You can check here. [https://github.com/Azure/sonic-buildimage/blob/37dc7bd4786c793796530e53da304b5e68b7ed16/files/image_config/interfaces/interfaces-config.sh#L7](https://github.com/Azure/sonic-buildimage/blob/37dc7bd4786c793796530e53da304b5e68b7ed16/files/image_config/interfaces/interfaces-config.sh#L7) 

For some asic, the switch mac must match certain pattern. 
Here for the VLAN mac, you will want it to be the same as the switch mac. Therefore, we need to follow the same procedure.

cfgmgr/vlanmgr.cpp
```diff
@@ -0,0 +1,488 @@
...
+                    fvVector.push_back(i);
+                }
+                /* Set vlan mtu */
+                else if (fvField(i) == "mtu")
+                {
+                    mtu = (uint32_t)stoul(fvValue(i));
+                    /*
+                     * TODO: support host VLAN mtu setting.
+                     * Host VLAN mtu should be set only after member configured
+                     * and VLAN state is not UNKOWN.
```
lguohan: UNKOWN-> UNKNOWN

cfgmgr/vlanmgr.cpp
```diff
@@ -0,0 +1,488 @@
...
+                     */
+                    SWSS_LOG_DEBUG("%s mtu %u: Host VLAN mtu setting to be supported.", key.c_str(), mtu);
+                    fvVector.push_back(i);
+                }
+                else if (fvField(i) == "members@") {
```
lguohan: in the new configdb schema, we have VLAN_MEMBER table, we should remove the members@ attribute in VLAN table, otherwise, there cloud potentially be two sources of truth.

We do not need to do this change now. But, please do it in later PR.

cfgmgr/vlanmgr.cpp
```diff
@@ -0,0 +1,488 @@
...
+        if (op == SET_COMMAND)
+        {
+            /* Don't proceed if member port/lag is not ready yet */
+            if (!isMemberStateOk(port_alias) || !isVlanStateOk(vlan_alias))
+            {
+                SWSS_LOG_DEBUG("%s not ready, delaying", kfvKey(t).c_str());
+                it++;
+                continue;
+            }
+            string tagging_mode = "untagged";
```
lguohan: can you also update the VLAN_MEMBER scheme in https://github.com/Azure/SONiC/wiki/Configuration ?

cfgmgr/vlanmgrd.cpp
JipanYanga: Ok, I'm think of using the switch table in configDB with has switch mac field

127.0.0.1:6379[4]> hgetall "SWITCH|SWITCH_ATTR"
1) "switch_mac"
2) "00:05:64:30:73:c0"

In interfaces-config.sh,  we set the switch_mac to configDB if it is not there already, then use the configDB data in vlanMgr.   This way it give people option to specify switch mac other than that of eth0.

cfgmgr/vlanmgr.cpp
JipanYanga: ok

cfgmgr/vlanmgr.cpp
JipanYanga: Ok, will remove it once minigraph.py update for the member processing is committed. 

https://github.com/Azure/sonic-buildimage/pull/1063/files#diff-18b32e4dc045fa3d1c5941a732124c89

cfgmgr/vlanmgrd.cpp
lguohan: that is fine, but use eth0 mac is not an option. need to remove that part.

JipanYanga: done.

cfgmgr/vlanmgr.cpp
```diff
@@ -0,0 +1,488 @@
...
+                    members = fvValue(i);
+                }
+            }
+            /* fvVector should not be empty */
+            if (fvVector.empty())
+            {
+                FieldValueTuple a("admin_status",  "up");
+                fvVector.push_back(a);
+            }
+            m_appVlanTableProducer.set(key, fvVector);
```
zhenggen-xu: Is there any plan to deal with the operational state of Vlan interface if not yet done?  We may need that to disable routing if vlan interface is down.
If we are doing that, I assume some other entity (like netlink message in linksync.cpp) will update the app-db?  In that case, do we have a timing issue here? (e,g, netlink messages came before we put the table to app-db)?

We can address this in future PR if needed.

lguohan: we need to define the operational state of vlan first. if it mean all members are down, then we have way to know the member oper status already, we do not need to rely on the netlink to get the vlan op status, then. This is going to be another feature request and thus a separate PR.

PR #361, opened 2017-10-26
[Pfcwd]: Add pfcAclHandler to drop ingress packets by acl rules

qiluo-msft:
This is following https://github.com/Azure/sonic-swss/pull/350. The original one will be closed.

**What I did**
Add pfcAclHandler to drop ingress packets by acl rules.
**Why I did it**
To drop ingress packets during PFC storm recovery with acl rules.
**How I verified it**
Test on DUT.
**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -981,6 +981,73 @@ void AclTable::link(sai_object_id_t portOid)
     ports.emplace(portOid, SAI_NULL_OBJECT_ID);
 }
 
+bool AclTable::add(shared_ptr<AclRule> newRule)
+{
+    string rule_id = newRule->getId();
+    auto ruleIter = rules.find(rule_id);
+    if (ruleIter != rules.end())
+    {
+        // If ACL rule already exists, delete it first
+        if (ruleIter->second->remove())
+        {
+            rules.erase(ruleIter);
+            SWSS_LOG_NOTICE("Successfully deleted ACL rule: %s", rule_id.c_str());
+        }
+    }
+
+    if (newRule->create())
+    {
+        rules[rule_id] = newRule;
+        SWSS_LOG_NOTICE("Successfully created ACL rule %s in table %s", rule_id.c_str(), id.c_str());
+        return true;
+    }
+    else
+    {
+        SWSS_LOG_ERROR("Failed to create rule in table %s", id.c_str());
+        return false;
+    }
+}
+
+bool AclTable::remove(string rule_id)
+{
+    auto ruleIter = rules.find(rule_id);
+    if (ruleIter != rules.end())
+    {
+        if (ruleIter->second->remove())
+        {
+            rules.erase(ruleIter);
+            SWSS_LOG_NOTICE("Successfully deleted ACL rule %s", rule_id.c_str());
+            return true;
+        }
+        else
+        {
+            SWSS_LOG_ERROR("Failed to delete ACL rule: %s", rule_id.c_str());
+            return false;
+        }
+    }
+    else
+    {
+        SWSS_LOG_WARN("Skip deleting ACL rule. Unknown rule %s", rule_id.c_str());
+        return true;
+    }
+}
+
+bool AclTable::clear()
+{
+    for (auto& rulepair: rules)
+    {
+        auto& rule = *rulepair.second;
+        bool suc = rule.remove();
+        if (!suc)
+        {
+            SWSS_LOG_ERROR("Failed to delete existing ACL rule %s when removing the ACL table %s", rule.getId().c_str(), id.c_str());
+            return false;
+        }
+    }
+    rules.clear();
+    return true;
+}
+
 AclRuleCounters AclRuleMirror::getCounters()
 {
     AclRuleCounters cnt(counters);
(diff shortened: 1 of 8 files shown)


orchagent/pfcactionhandler.cpp
```diff
@@ -133,6 +133,86 @@ void PfcWdActionHandler::updateWdCounters(const string& queueIdStr, const PfcWdQ
...
 
+PfcWdAclHandler::PfcWdAclHandler(sai_object_id_t port, sai_object_id_t queue,
+        uint8_t queueId, shared_ptr<Table> countersTable):
+    PfcWdActionHandler(port, queue, queueId, countersTable)
+{
+    acl_table_type_t table_type = ACL_TABLE_L3;
+
+    string portstr = sai_serialize_object_id(port);
+    string queuestr = sai_serialize_object_id(queue);
+    m_strTable = "Table_" + portstr + "_" + queuestr;
```
lguohan: the Table name should be PFC_INGRESS_DROP_TC_X. It does not need to have port name, since this table will be able to bind to multiple ports.

orchagent/pfcactionhandler.cpp
```diff
@@ -133,6 +133,86 @@ void PfcWdActionHandler::updateWdCounters(const string& queueIdStr, const PfcWdQ
...
+    m_strRule = "Rule_" + portstr + "_" + queuestr;
```
lguohan: >portstr [](start = 26, length = 7)

rule should not have port name either.

orchagent/pfcactionhandler.cpp
```diff
@@ -133,6 +133,86 @@ void PfcWdActionHandler::updateWdCounters(const string& queueIdStr, const PfcWdQ
...
+
+    attr_name = "MATCH_TC";
+    attr_value = to_string(queueId);
+    rule->validateAddMatch(attr_name, attr_value);
+
+    attr_name = ACTION_PACKET_ACTION;
+    attr_value = PACKET_ACTION_DROP;
+    rule->validateAddAction(attr_name, attr_value);
+
+    gAclOrch->addAclRule(rule, m_strTable);
```
marian-pritsak: When is the rule removed?

marian-pritsak:
Function entries are missing SWSS_LOG_ENTER();

Refactoring code for ACL could be done in a separate PR.

marian-pritsak:
What about egress packets? How are they dropped?

orchagent/aclorch.cpp
```diff
@@ -981,6 +981,73 @@ void AclTable::link(sai_object_id_t portOid)
...
+    if (newRule->create())
+    {
+        rules[rule_id] = newRule;
+        SWSS_LOG_NOTICE("Successfully created ACL rule %s in table %s", rule_id.c_str(), id.c_str());
+        return true;
+    }
+    else
+    {
+        SWSS_LOG_ERROR("Failed to create rule in table %s", id.c_str());
+        return false;
```
JipanYanga: Since you are doing the refactoring,  I'm curious to see how you handle the error return code all the levels up the stack?  I have similar issues in other PR and would like to get some hint here.

orchagent/pfcactionhandler.cpp
qiluo-msft: Not clear in requirement, so I assume them be long living. I created a function clear() to remove all the tables and rules, and it could be called if we make different decision.

Let me know if you have any other idea?

orchagent/aclorch.cpp
qiluo-msft: If you are talking specific to this function, the call stacks is like below.
```
bool AclTable::add(shared_ptr<AclRule> newRule)
bool AclOrch::addAclRule(shared_ptr<AclRule> newRule, string table_id)
void AclOrch::doAclRuleTask(Consumer &consumer)
```
You can see if any lower-level function is possible to fail, it returns a false result, and upper-level function should check the result and response to the failure. Upper-level function should never ignore the returned result. That is the point when I asked you to 'handle the error'.

qiluo-msft:
We plan to add egress ACL in a separate PR.

---
In reply to: [340288153](https://github.com/Azure/sonic-swss/pull/361#issuecomment-340288153) [](ancestors = 340288153)

orchagent/aclorch.cpp
JipanYanga: What happens in AclTable::add() upon error?   Just log the error without quit or recover?

qiluo-msft:
The logs are fixed.
Not going to separate the refactor, if you are not insist on:)

---
In reply to: [340274189](https://github.com/Azure/sonic-swss/pull/361#issuecomment-340274189) [](ancestors = 340274189)

orchagent/aclorch.cpp
qiluo-msft: Yes, you are right. Base on current implementation, this is the best we could do.

marian-pritsak: Understood

orchagent/aclorch.h
```diff
@@ -214,14 +213,43 @@ class AclRuleMirror: public AclRule
...
+    // Bind the ACL table to a port which is alread linked
+    bool bind(sai_object_id_t portOid);
+    // Unbind the ACL table to a port which is alread linked
+    bool unbind(sai_object_id_t portOid);
+    // Bind the ACL table to all ports linked
+    bool bind();
+    // Unbind the ACL table to all ports linked
+    bool unbind();
+    // Link the ACL table with a port, for future bind or unbind
+    void link(sai_object_id_t portOid);
```
lguohan: >link [](start = 9, length = 4)

maybe word attach is better here.

PR #363, opened 2017-10-30
[pfcwd]: Rx/Tx/Drop counters

marian-pritsak:


marian-pritsak:
Please see https://github.com/Azure/SONiC/wiki/PFC-Watchdog-Design#23-counters-db
It has description of counters.

Code under #if 0 is legitimate and can be compiled and run. ifdef will be removed as soon as SAI will be updated.

marian-pritsak:
Depends on

https://github.com/Azure/sonic-sairedis/pull/247

lguohan:
@qiluo-msft to review

orchagent/pfcwdorch.cpp
```diff
@@ -491,9 +491,11 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::handleWdNotification(swss::Notifi
         {
             throw runtime_error("Unknown PFC WD action");
         }
+        entry->second.handler->initCounters();
```
qiluo-msft: >initCounters [](start = 31, length = 12)

What is the purpose of this PR? refactor or fix?
Is the behavior same as before?

orchagent/pfcactionhandler.h
```diff
@@ -39,15 +47,28 @@ class PfcWdActionHandler
         }
 
         static void initWdCounters(shared_ptr<Table> countersTable, const string &queueIdStr);
+        void initCounters(void);
+        void commitCounters(void);
+
+        virtual bool getHwCounters(PfcWdHwStats& counters)
```
qiluo-msft: >getHwCounters [](start = 21, length = 13)

What is Hw counter? Is there any Sw counter?

orchagent/pfcactionhandler.h
```diff
@@ -58,6 +79,7 @@ class PfcWdActionHandler
         uint8_t m_queueId = 0;
         shared_ptr<Table> m_countersTable = nullptr;
         PfcWdQueueStats m_stats;
```
qiluo-msft: >m_stats [](start = 24, length = 7)

call it m_queuestats?

orchagent/pfcactionhandler.h
```diff
@@ -39,15 +47,28 @@ class PfcWdActionHandler
...
+        void initCounters(void);
+        void commitCounters(void);
+
+        virtual bool getHwCounters(PfcWdHwStats& counters)
+        {
+            return true;
+        };
 
     private:
         struct PfcWdQueueStats
```
qiluo-msft: >PfcWdQueueStats [](start = 15, length = 15)

inherit PfcWdHwStats?

orchagent/pfcactionhandler.h
marian-pritsak: HW counters are actual counters from ASIC, They can be either queue/PG counters or ACL counters (depending on handler). PFC WD counters calculate an overall traffic that was transmitted or dropped during PFC storm based on HW counters values.

orchagent/pfcactionhandler.h
marian-pritsak: No need to keep it as a class member anymore. Thanks for the suggestion.

orchagent/pfcactionhandler.h
marian-pritsak: I'd rather not do it. Values in every struct have different meaning and value. counters in PfcWdQueueStats represent PFC WD stats, whereas values in PfcWdHwStats are snapshots from hardware based on which first ones are calculated.

orchagent/pfcwdorch.cpp
marian-pritsak: It's actually a fix. First assumption was that counters will be the same for all handlers. But for example ACL handler needs to count dropped packets in corresponding rule.

orchagent/pfcactionhandler.h
qiluo-msft: >counters [](start = 49, length = 8)

If base class has valid implementation for getHwCounters(), and you intend to return a zero counter, you should set it to zero explicitly.

orchagent/pfcactionhandler.h
```diff
@@ -13,6 +13,14 @@ extern "C" {
 using namespace std;
 using namespace swss;
 
+struct PfcWdHwStats
+{
+    uint64_t txPkt     = 0;
```
qiluo-msft: >= 0 [](start = 23, length = 3)

Normally you don't need to provide a struct a ctor, or init values.

orchagent/pfcactionhandler.h
```diff
@@ -39,15 +47,28 @@ class PfcWdActionHandler
         }
 
         static void initWdCounters(shared_ptr<Table> countersTable, const string &queueIdStr);
+        void initCounters(void);
```
sihuihan88: There are initWdCounters and initCounters in this class, which might be a little difficult to understand. Consider rename to be more specific?

orchagent/pfcactionhandler.cpp
```diff
@@ -34,30 +41,60 @@ PfcWdActionHandler::PfcWdActionHandler(sai_object_id_t port, sai_object_id_t que
...
+void PfcWdActionHandler::initCounters(void)
+{
+    SWSS_LOG_ENTER();
+
+    if (!getHwCounters(m_hwStats))
+    {
+        return;
+    }
+
+    auto wdQueueStats = getQueueStats(m_countersTable, sai_serialize_object_id(m_queue));
```
sihuihan88: minor, getQueueStats and updateWdCounters seem to get/set pfcwd counters. Shall we name them follow same routine, which might make it more understandable?

orchagent/pfcactionhandler.h
marian-pritsak: In a C-style code, I'd normally write something like this
```
PfcWdHwStats stats = {};
```
To initialize all values.
Do you prefer this way over current implementation?

qiluo-msft: Yes.

orchagent/pfcactionhandler.cpp
```diff
@@ -271,6 +329,69 @@ PfcWdLossyHandler::~PfcWdLossyHandler(void)
...
+{
+    SWSS_LOG_ENTER();
+
+    static const vector<sai_queue_stat_t> queueStatIds =
+    {
+        SAI_QUEUE_STAT_PACKETS,
+        SAI_QUEUE_STAT_DROPPED_PACKETS,
+    };
+
+    static const vector<sai_ingress_priority_group_stat_t> pgStatIds =
```
sihuihan88: Can we separate the counters for drop/forward action, and only query the needed counters depends on the pfcwd action?

orchagent/pfcactionhandler.h
marian-pritsak: Done

orchagent/pfcactionhandler.h
marian-pritsak: Done

orchagent/pfcactionhandler.cpp
marian-pritsak: These functions are not introduced in this PR. But I agree that renaming should be done. Can I suggest to do it in a separate PR?

orchagent/pfcactionhandler.h
marian-pritsak: Same as above

orchagent/pfcactionhandler.cpp
marian-pritsak: I cannot guarantee that all handlers have a correct implementation of drop/forward handlers, and not getting all counters can hide bugs with traffic not being dropped/forwarded. Besides, it's not polling but a one time action. What's the benefit of not reading 2 counters?

orchagent/pfcactionhandler.cpp
sihuihan88: Sure.

orchagent/pfcactionhandler.cpp
sihuihan88: It is possible that some vendors might only support one action type eg. drop action, and don't support the counters needed for forward action, though I can also see your point.

marian-pritsak: If some vendor does not support any of counters, then it should be set to 0. It's common practice to set counters to zero. It's easier to extend counters if needed in the future just by extending struct definition instead of declaring new functions.

sihuihan88: We could set counters to zero, though I still don't think we need to query all the hardware counters for both drop/forward actions regardless of the action type.

qiluo-msft:
I closes all my comments. Ok to me.

PR #364, opened 2017-10-30
[everflow]: Use MLNX_PLATFORM_SUBSTRING define to get platform.

oleksandrivantsiv:

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -3,6 +3,7 @@
 #include <utility>
 #include <exception>
 
+#include "orch.h"
 #include "logger.h"
 #include "swssnet.h"
 #include "converter.h"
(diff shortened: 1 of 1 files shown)


orchagent/mirrororch.cpp
```diff
@@ -45,7 +44,7 @@ MirrorEntry::MirrorEntry(const string& platform) :
     nexthopInfo.resolved = false;
     neighborInfo.resolved = false;
 
-    if (platform == MLNX_PLATFORM)
+    if (platform == MLNX_PLATFORM_SUBSTRING)
```
nikos-li: Rather than keep changing names to platform string inside PI code, you should come up with a design that removes PD checks from PI code.

oleksandrivantsiv: This change is done according to ACL dynamic configuration design

oleksandrivantsiv: And come from requirements https://github.com/Azure/SONiC/wiki/ACL-Configuration-Requirement-Description

lguohan: I agree with what Nikos said in principle. For this PR, the code is already there. I think we should address the problem in future PR, and we need to more discussion on how to address it properly.

PR #366, opened 2017-11-01
[orchagent]: Remove the global variable gPortsOrch in orch.cpp file

stcheng:
This change is the first step of refactoring Orch class.
The global variable gPortsOrch is removed from that orch.cpp file
and this class no longer requires such external variable.

The current solution is to move the global variable into each different
places that need this variable. Later, a better design could be
introduced to further remove all the global variables from the whole
project.

**What I did**
Remove the gPortsOrch from the orch.cpp file.

**Why I did it**
Free the orch.cpp file.

**How I verified it**
Compile and run the code.

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -605,7 +605,7 @@ sai_object_id_t AclRuleL3::getRedirectObjectId(const string& redirect_value)
 
     // Try to parse physical port and LAG first
     Port port;
-    if(m_pAclOrch->m_portOrch->getPort(target, port))
+    if (gPortsOrch->getPort(target, port))
     {
         if (port.m_type == Port::PHY)
         {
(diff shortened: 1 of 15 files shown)


stcheng:
@JipanYanga could you check this pull request? thanks.

JipanYanga:
@stcheng  As the first step of eliminating global variables from Orch class, the change looks good to me.   
We should come up the new design asap.  With "gPortsOrch->isInitDone()" check in more places, 
it is a little like the spread of germs.

andriymoroz-mlnx:
If global variable gPortsOrch is removed from orch.cpp, why 
`extern PortsOrch *gPortsOrch;`
is still there? ;)

stcheng:
@andriymoroz-mlnx ah thanks, i'll remove it.

PR #370, opened 2017-11-04
IntfsOrch uses ConsumerTable instead of ConsumerStateTable

qiluo-msft:
Signed-off-by: Qi Luo <qiluo-msft@users.noreply.github.com>

**What I did**
IntfsOrch uses ConsumerTable instead of ConsumerStateTable

**Why I did it**
Sample Intfs events:
```
Nov  4 18:53:52.393657 str-msn2700-03 WARNING orchagent: :- doTask: IntfsOrch::doTask receives Ethernet96:10.0.0.48/31 DEL
Nov  4 18:53:52.403133 str-msn2700-03 WARNING orchagent: :- doTask: IntfsOrch::doTask receives Ethernet96:10.0.1.48/8 SET
Nov  4 18:53:52.407662 str-msn2700-03 WARNING orchagent: :- doTask: IntfsOrch::doTask receives Ethernet96:10.0.1.48/24 SET
Nov  4 18:53:52.407999 str-msn2700-03 WARNING orchagent: :- doTask: IntfsOrch::doTask receives Ethernet96:10.0.1.48/8 DEL
```
The order of events are important here, so use original ConsumerTable to keep the event order. ConsumerStateTable will not keep the order.

**How I verified it**
Test it on DUT, use 'sudo ifconfig Ethernet96 10.0.1.48/24‘ to change interface IP address

**Details if related**

### DIFF
--- a/intfsyncd/intfsync.cpp
+++ b/intfsyncd/intfsync.cpp
@@ -6,7 +6,6 @@
 #include "logger.h"
 #include "netmsg.h"
 #include "dbconnector.h"
-#include "producerstatetable.h"
 #include "linkcache.h"
 #include "intfsyncd/intfsync.h"
 
--- a/intfsyncd/intfsync.h
+++ b/intfsyncd/intfsync.h
@@ -2,7 +2,7 @@
 #define __INTFSYNC__
 
 #include "dbconnector.h"
-#include "producerstatetable.h"
+#include "producertable.h"
 #include "netmsg.h"
 
 namespace swss {
@@ -17,7 +17,7 @@ public:
     virtual void onMsg(int nlmsg_type, struct nl_object *obj);
 
 private:
-    ProducerStateTable m_intfTable;
+    ProducerTable m_intfTable;
 };
 
 }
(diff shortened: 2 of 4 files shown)


lguohan:
I do not understand the sequence here. if you set 10.0.0.148/31, at end of the day, you will have an interface in the appdb which is 10.0.0.148/31.

qiluo-msft:
(sorry, corrected original text)

lguohan:
in the end, your log still suggest it creates the Ethernet96:10.0.1.48/24, so I am not sure if change to the state table what is exactly the issue. Can you show the original issue?

qiluo-msft:
There should be no issue if keep using ConsumerStateTable, or not intensively test. I submitted this PR because in concept, IntfsOrch processes events which are order sensitive. So no point to use ConsumerStateTable here.

zhenggen-xu:
Can we get this code in if we don't have objections?  This is the right way to handle interface messages.  This can solve the weirdness of the ifconfig problem work-arounded in intfsorch.cpp.   We want to get rid of those code to pave the road for ipv6 link-local use cases or even any type of overlapping subnet/ip use cases.

PR #374, opened 2017-11-07
[pfcwd]: Switch to Config DB

marian-pritsak:


qiluo-msft:
swssconfig still writes everything into APPL_DB. So this change will break (manual) PFC WD test.
Please double check.

PR #375, opened 2017-11-08
[mtu] Update the mtu on host interfaces if config DB changes

hzheng5:
Signed-off-by: Haiyang Zheng <haiyang.z@alibaba-inc.com>

**What I did**

Now, portsyncd is subscribed to config DB, and use handlePortConfig to copy port config from config DB to APP DB. Here, we also need to update the host interfaces MTU in linux kernel if the port config changes.

**Why I did it**

Otherwise, there will be discrepancy between linux kernel MTU and port MTU in config DB/app DB. 

**How I verified it**

Update the port inside config_db.json file with new mtu field, and verify both kernel MTU and ASIC port MTU has been updated properly.

**Details if related**

This pull request require the following pull request
https://github.com/Azure/sonic-buildimage/pull/1128

With both changes, the MTU from config DB will be the only source for Kernel MTU and port MTU on ASIC.

### DIFF
--- a/portsyncd/portsyncd.cpp
+++ b/portsyncd/portsyncd.cpp
@@ -13,6 +13,7 @@
 #include "producerstatetable.h"
 #include "portsyncd/linksync.h"
 #include "subscriberstatetable.h"
+#include "exec.h"
 
 #define DEFAULT_PORT_CONFIG_FILE     "port_config.ini"
 
@@ -250,6 +251,19 @@ void handlePortConfig(ProducerStateTable &p, map<string, KeyOpFieldsValuesTuple>
             if (op == SET_COMMAND)
             {
                 p.set(key, values);
+                for (auto fv : values)
+                {
+                    string field = fvField(fv);
+                    string value = fvValue(fv);
+
+                    /* Update the mtu field on host interface */
+                    if (field == "mtu")
+                    {
+                        string cmd, res;
+                        cmd = "ip link set " + key + " mtu " + value;
+                        swss::exec(cmd, res);
+                     }
+                }
             }
 
             it = port_cfg_map.erase(it);


portsyncd/portsyncd.cpp
```diff
@@ -250,6 +250,24 @@ void handlePortConfig(ProducerStateTable &p, map<string, KeyOpFieldsValuesTuple>
...
+                {
+                    string field = fvField(fv);
+                    string value = fvValue(fv);
+
+                    /* Update the mtu field on host interface */
+                    if (field == "mtu")
+                    {
+                        char system_cmd[1024];
+                        snprintf(system_cmd, sizeof(system_cmd), "ip link set %s mtu %s", key.c_str(), value.c_str());
+                        int ret_val = system(system_cmd);
```
prsunny: Generally, I've seen the usage of swss::exec to execute commands instead of direct system call. Can you check? Also, you could use stringstream for the command which would be convenient instead of allocating a buffer and then use snprintf

You could refer: IntfMgr::setIntfIp in intfmgr.cpp

hzheng5: Thanks Sunny. 

Will check.

hzheng5: Hi @prsunny ,

Just updated the diff with exec. Thanks for your suggestion.

Please help review the latest diff.

Cheers,
Haiyang

hzheng5:
Hi @prsunny ,

Just updated the diff with exec. Thanks for your suggestion.

Please help review the latest diff.

Cheers,
Haiyang

prsunny:
I'm fine with the code changes.

portsyncd/portsyncd.cpp
```diff
@@ -250,6 +251,19 @@ void handlePortConfig(ProducerStateTable &p, map<string, KeyOpFieldsValuesTuple>
...
+                {
+                    string field = fvField(fv);
+                    string value = fvValue(fv);
+
+                    /* Update the mtu field on host interface */
+                    if (field == "mtu")
+                    {
+                        string cmd, res;
+                        cmd = "ip link set " + key + " mtu " + value;
+                        swss::exec(cmd, res);
```
prsunny: Do you want to check for return value?.

hzheng5: Hi Sunny @prsunny ,

The return value will be checked inside swss::exec, if failed swss::exec will throw an error. 

Thanks,
Haiyang

PR #376, opened 2017-11-09
Refine syntax: expose overloaded inherited functions

qiluo-msft:


cfgmgr/vlanmgr.cpp
```diff
@@ -167,12 +167,11 @@ bool VlanMgr::removeHostVlanMember(int vlan_id, const string &port_alias)
 
 bool VlanMgr::isVlanMacOk()
 {
-    return !(!gMacAddress);
+    return !!gMacAddress;
```
rodnymolina: I know i'm late for this one, but i just noticed it now. Why would you change from the previous readable notation to this unintuitive one?

qiluo-msft: Unnecessary parentheses make the previous code less readable, and less intuitive.

PR #382, opened 2017-11-12
Remove the threads inside PfcWdSwOrch and AclOrch

qiluo-msft:
**What I did**
Remove the threads inside PfcWdSwOrch and AclOrch, instead, use Selectable objects to represent the notification event or timer event, and process these events similar to ConsumerTable events.

**Why I did it**
One design principle of orchagent is single thread. libsairedis is not thread-safe.

**How I verified it**
Manual PFC WD test on DUT

**Details if related**
Depends on https://github.com/Azure/sonic-swss-common/pull/160

### DIFF
--- a/cfgmgr/vlanmgr.cpp
+++ b/cfgmgr/vlanmgr.cpp
@@ -341,7 +341,7 @@ void VlanMgr::processUntaggedVlanMembers(string vlan, const string &members)
         SWSS_LOG_ERROR("Failed to find tableName:%s", CFG_VLAN_MEMBER_TABLE_NAME);
         return;
     }
-    Consumer& consumer = consumer_it->second;
+    auto& consumer = (Consumer &)*consumer_it->second;
 
     vector<string> vlanMembers = tokenize(members, ',');
 
@@ -469,7 +469,7 @@ void VlanMgr::doTask(Consumer &consumer)
 {
     SWSS_LOG_ENTER();
 
-    string table_name = consumer.m_consumer->getTableName();
+    string table_name = consumer.getTableName();
 
     if (table_name == CFG_VLAN_TABLE_NAME)
     {
--- a/cfgmgr/vlanmgr.cpp
+++ b/cfgmgr/vlanmgr.cpp
@@ -341,7 +341,7 @@ void VlanMgr::processUntaggedVlanMembers(string vlan, const string &members)
         SWSS_LOG_ERROR("Failed to find tableName:%s", CFG_VLAN_MEMBER_TABLE_NAME);
         return;
     }
-    auto& consumer = (Consumer &)*consumer_it->second;
+    auto& consumer = static_cast<Consumer &>(*consumer_it->second);
 
     vector<string> vlanMembers = tokenize(members, ',');
 
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -45,9 +45,6 @@ Orch::Orch(const vector<TableConnector>& tables)
 
 Orch::~Orch()
 {
-    for(auto &it : m_consumerMap)
-        delete it.second.m_consumer;
-
     if (gRecordOfs.is_open())
     {
         gRecordOfs.close();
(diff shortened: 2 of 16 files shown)


orchagent/orch.cpp
```diff
@@ -373,10 +356,17 @@ void Orch::addConsumer(DBConnector *db, string tableName)
 {
     if (db->getDB() == CONFIG_DB)
     {
-        Consumer consumer(new SubscriberStateTable(db, tableName));
-        m_consumerMap.insert(ConsumerMapPair(tableName, consumer));
-    } else {
-        Consumer consumer(new ConsumerStateTable(db, tableName, gBatchSize));
-        m_consumerMap.insert(ConsumerMapPair(tableName, consumer));
+        addExecutor(tableName, new Consumer(new SubscriberStateTable(db, tableName), this));
```
prsunny: Just to understand, since there is no return pointer to object here, the free must be happening in the callee, right?. Also will it be good to use smart pointers (unique_ptr) to avoid any memory leaks?

orchagent/timer.h
```diff
@@ -0,0 +1,28 @@
...
+
+class ExecutableTimer : public Executor
+{
+public:
+    ExecutableTimer(SelectableTimer *timer, Orch *orch)
+        : Executor(timer, orch)
+    {
+    }
+
+    SelectableTimer *getSelectableTimer()
```
prsunny: Noticed in rest of your code changes, the usage of "const" for class functions like this. Do you want to add it here as well?

orchagent/timer.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
```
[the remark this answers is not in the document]
qiluo-msft: I prefer move to pragma since it is much more clean and less easy mistaken. It is nearly a practical 'standard'.

[https://en.wikipedia.org/wiki/Pragma_once](https://en.wikipedia.org/wiki/Pragma_once)

---
In reply to: [150723282](https://github.com/Azure/sonic-swss/pull/382#discussion_r150723282) [](ancestors = 150723282)

orchagent/timer.h
qiluo-msft: This wrapper class doesn't care about non const access to timer, so it will keep it as raw as possible. Do you find any harm for ExecutableTimer class?

---
In reply to: [150728600](https://github.com/Azure/sonic-swss/pull/382#discussion_r150728600) [](ancestors = 150728600)

cfgmgr/vlanmgr.cpp
```diff
@@ -341,7 +341,7 @@ void VlanMgr::processUntaggedVlanMembers(string vlan, const string &members)
         SWSS_LOG_ERROR("Failed to find tableName:%s", CFG_VLAN_MEMBER_TABLE_NAME);
         return;
     }
-    Consumer& consumer = consumer_it->second;
+    auto& consumer = (Consumer &)*consumer_it->second;
```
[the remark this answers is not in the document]
qiluo-msft: original 
consumer_it->second :: Consumer
now
consumer_it->second :: Executor*

The cast is valid, since there is only Consumers in this application. I can change it to static_cast.

---
In reply to: [150726106](https://github.com/Azure/sonic-swss/pull/382#discussion_r150726106) [](ancestors = 150726106)

orchagent/orch.cpp
qiluo-msft: The Executor classes (include Consumer) will hold and responsible to release the memory by design. You may check the comment there.

The Orch classes will hold and responsible to release the memory by original design.

I would not use unique_ptr here, since the lifecycle of the new objects is larger than the function-local.

---
In reply to: [150727650](https://github.com/Azure/sonic-swss/pull/382#discussion_r150727650) [](ancestors = 150727650)

lguohan:
retest this please

orchagent/timer.h
prsunny: I don't see any harm. Just noticed the change between the two class defines

orchagent/timer.h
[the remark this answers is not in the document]
prsunny: pragma has its advantage.. I was just mentioning on the consistency with other header files

orchagent/orch.cpp
prsunny: ok

lguohan:
retest this please

PR #386, opened 2017-11-14
[acl]: Fix issue with bind ACL table group to port.

oleksandrivantsiv:
Port object creates ACL table group, binds it to particular port
and stores group ID. AclOrch class was operating with copies of
Port objects. ID of created ACL table group was stored in object copy.
Which caused resoures leak after copy was destroyed. For each ACL table new
ACL table group was created. Only one ACL table was bind to
group at the same time.


lguohan:
retest this please

lguohan:
retest this please

qiluo-msft:
Sorry the PR description is not clear to me. What are you blaming? Where is the bug? How do you fix it.
Checking the code, do you actually mean gPortsOrch->getPort() is not working as expected? If yes could you just fix that function?

lguohan:
I think getPort() is returning a copy of port object. I think that API may or may not by design. That is why the fix is not fix getPort(), instead it is using getAllPorts() to get the real port object to prevent the resource leak.

lguohan:
@oleksandrivantsiv , do you have a step to reproduce the problem? once we get your step, we can later add a swss test on the virtual switch to test it.

qiluo-msft:
@lguohan I agree there may be a bug in getPort. If this is the case, @oleksandrivantsiv could you add a // TODO comment to the code snippet for 'getPort', and we will revert this change after fixing the real bug.

oleksandrivantsiv:
To reproduce the issue it is enough to create two or more ACL tables. Only one table will be bind to ports. Other won't work.

getPort function was implemented to work with stateless port object. The issue was introduced when AclOrch was refactored and port object became stateful in https://github.com/Azure/sonic-swss/commit/0c98cc36b420fd33fc8066b1f140f6515bb00a76 commit. getPort wasn't adopted to that changes.

PR #387, opened 2017-11-15
[cfgmgr] Don't throw when grep cmd didn't match

ishidawataru:
**What I did**

When a port belongs to several vlans, it is possible that
grep doesn't match with " None".

Just call exec and stop throwing in case of the failure

**Why I did it**

`cfgmgr` crashes when removing a VLAN assignment on a port which is configured with multiple VLANs

**How I verified it**

1. Configure multiple VLAN on a specific port like the configuration below.

```json
{
    "VLAN": {
        "Vlan1000": {
            "members": [
                "Ethernet1",
                "Ethernet9_1"
            ],
            "vlanid": "1000"
        },
        "Vlan2000": {
            "members": [
                "Ethernet1",
                "Ethernet9_2"
            ],
            "vlanid": "2000"
        }
    },
    "VLAN_MEMBER": {
        "Vlan1000|Ethernet1": {
            "tagging_mode": "tagged"
        },
        "Vlan2000|Ethernet1": {
            "tagging_mode": "tagged"
        },
        "Vlan1000|Ethernet9_1": {
            "tagging_mode": "untagged"
        },
        "Vlan2000|Ethernet9_2": {
            "tagging_mode": "untagged"
        }
    }
}
```
2. Remove one VLAN assignment

```bash
$  redis-cli -s /var/run/redis/redis.sock -n 4 del "VLAN_MEMBER|Vlan2000|Ethernet1"
```

**Details if related**

### DIFF
--- a/cfgmgr/vlanmgr.cpp
+++ b/cfgmgr/vlanmgr.cpp
@@ -172,17 +172,22 @@ bool VlanMgr::removeHostVlanMember(int vlan_id, const string &port_alias)
     SWSS_LOG_ENTER();
 
     // The command should be generated as:
-    // /bin/bash -c "/sbin/bridge vlan del vid {{vlan_id}} dev {{port_alias}} &&
-    //               /sbin/bridge vlan show dev {{port_alias}} | /bin/grep -q None &&
-    //               /sbin/ip link set {{port_alias}} nomaster"
+    // /bin/bash -c '/sbin/bridge vlan del vid {{vlan_id}} dev {{port_alias}} &&
+    //               ( /sbin/bridge vlan show dev {{port_alias}} | /bin/grep -q None;
+    //               ret=$?; if [ $ret -eq 0 ]; then
+    //               /sbin/ip link set {{port_alias}} nomaster;
+    //               elif [ $ret -eq 1 ]; then exit 0;
+    //               else exit $ret; fi )'
 
     // When port is not member of any VLAN, it shall be detached from Dot1Q bridge!
     const std::string cmds = std::string("")
-      + BASH_CMD + " -c \""
-      + BRIDGE_CMD + " vlan del vid " + std::to_string(vlan_id) + " dev " + port_alias + " && "
+      + BASH_CMD + " -c \'"
+      + BRIDGE_CMD + " vlan del vid " + std::to_string(vlan_id) + " dev " + port_alias + " && ( "
       + BRIDGE_CMD + " vlan show dev " + port_alias + " | "
-      + GREP_CMD + " -q None && "
-      + IP_CMD + " link set " + port_alias + " nomaster\"";
+      + GREP_CMD + " -q None; ret=$?; if [ $ret -eq 0 ]; then "
+      + IP_CMD + " link set " + port_alias + " nomaster; "
+      + "elif [ $ret -eq 1 ]; then exit 0; "
+      + "else exit $ret; fi )\'";
 
     std::string res;
     EXEC_WITH_ERROR_THROW(cmds, res);


JipanYanga:
Thanks for fixing it!  I missed this one at initial commit.     you are right, I think we may just check the command result for this case.

cfgmgr/vlanmgr.cpp
```diff
@@ -154,7 +154,7 @@ bool VlanMgr::removeHostVlanMember(int vlan_id, const string &port_alias)
     cmd.str("");
     // When port is not member of any VLAN, it shall be detached from Dot1Q bridge!
     cmd << BRIDGE_CMD << " vlan show dev " << port_alias << " | " << GREP_CMD << " None";
-    EXEC_WITH_ERROR_THROW(cmd.str(), res);
+    swss::exec(cmd.str(), res);
```
lguohan: in this case, you can use the return value to decide.

Normally the exit status is 0 if a line is selected, 1 if no lines were selected, and 2 if an error occurred.  However, if the -q or --quiet or --silent is used and a line is selected, the exit status is 0 even  if        an error occurred.

ishidawataru:
@JipanYanga  @lguohan Thanks for your review. Reflected your comments. PTAL

cfgmgr/vlanmgr.cpp
```diff
@@ -147,19 +147,26 @@ bool VlanMgr::removeHostVlanMember(int vlan_id, const string &port_alias)
...
     // When port is not member of any VLAN, it shall be detached from Dot1Q bridge!
     cmd << BRIDGE_CMD << " vlan show dev " << port_alias << " | " << GREP_CMD << " None";
-    EXEC_WITH_ERROR_THROW(cmd.str(), res);
-    if (!res.empty())
-    {
-        cmd.str("");
-        cmd << IP_CMD << " link set " << port_alias << " nomaster";
-        EXEC_WITH_ERROR_THROW(cmd.str(), res);
+    ret = swss::exec(cmd.str(), res);
+    ret = WEXITSTATUS(ret);
```
lguohan: @JipanYanga , It looks like this WEXITSTATUS is better to be hided in the swss::exec(). 

Inside swss::exec(). The pclose() function waits for the associated process to terminate and returns the exit status of the command as returned by wait4(2).

pclose() returns -1 if wait4(2) returns an error.

ishidawataru: @lguohan putting `WEXITSTATUS` in `swss:exec()` doesn't break anything? `WEXITSTATUS` loses some information about the child process (e.g. received signal ). If this is not the case, I will put it in `swss:exec()`.

JipanYanga: @ishidawataru @lguohan For shell command execution in current implementation,  for most of the cases  the return value of zero is expected, signal interruption should trigger exception.   But for this case and https://github.com/Azure/sonic-swss/blob/master/cfgmgr/vlanmgr.cpp#L38 ,  we know the return value may be non-zero.    It looks to me  user of swss:exec() should have the original returned value and process it accordingly, though sometimes it is safe to ignore the return value.   If we don't put  WEXITSTATUS in swss:exec() ,  probably more comments including how to use WEXITSTATUS could be added in the header file?

ishidawataru: @JipanYanga Thanks for your comment. Yes, I also think keeping returning value of `swss::exec()` as is and adding comments would be better.

@lguohan WDYT?

jipanyang:
@ishidawataru @stcheng @lguohan  Not sure why this fix is missed.  The problem is still there and vlanmgrd will exit at this scenario.  
@ishidawataru do you mind updating your fix again with the latest code?

ishidawataru:
Hi @jipanyang. updated the code.
PTAL @stcheng @lguohan

jipanyang:
@ishidawataru  thanks for the fix, it looks good to me.  Also did the test:
# before this change:
config vlan add 1200
config vlan member add 1200 Ethernet33
config vlan add 2000
config vlan member add 2000 Ethernet33

config vlan member del  2000 Ethernet33    <----- Vlanmgr will exit

```
May 22 05:44:23.021050 sonic ERR vlanmgrd: :- exec: /bin/bash -c "/sbin/bridge vlan del vid 2000 dev Ethernet33 && /sbin/bridge vlan show dev Ethernet33 | /bin/grep -q None && /sbin/ip link set Ethernet33 nomaster": Success
May 22 05:44:23.021157 sonic ERR vlanmgrd: :- main: Runtime error: /bin/bash -c "/sbin/bridge vlan del vid 2000 dev Ethernet33 && /sbin/bridge vlan show dev Ethernet33 | /bin/grep -q None && /sbin/ip link set Ethernet33 nomaster" : 
May 22 05:44:23.022464 sonic INFO swss.sh[8122]: 2018-05-22 05:44:23,021 INFO exited: vlanmgrd (exit status 255; not expected)
May 22 05:44:27.905887 sonic INFO supervisord 2018-05-22 05:44:23,021 INFO exited: vlanmgrd (exit status 255; not expected)
```

# After this change:   it works without problem.

root@sonic:/var/log# config vlan member add   2000 Ethernet33
root@sonic:/var/log# config vlan member del  2000 Ethernet33

```
May 22 05:51:50.483804 sonic NOTICE syncd: :- processEvent: key: SAI_OBJECT_TYPE_VLAN_MEMBER:oid:0x27000000000a38 op: remove
May 22 05:51:50.483804 sonic INFO syncd: brcm_sai_remove_vlan_member:551 SAI Enter brcm_sai_remove_vlan_member
May 22 05:51:50.484129 sonic DEBUG syncd: brcm_sai_remove_vlan_member:635 Removing member type 1 val 51 from vlan 2000
May 22 05:51:50.484129 sonic NOTICE orchagent: :- removeVlanMember: Remove member Ethernet33 from VLAN Vlan2000 lid:7d0 vmid:27000000000a38
May 22 05:51:50.484645 sonic INFO syncd: brcm_sai_remove_vlan_member:703 SAI Exit brcm_sai_remove_vlan_member
```

PR #389, opened 2017-11-16
Ignore netlink messages for old interface during swss restart

prsunny:
**What I did**
In netlink process for link events, a new interface creation always comes without a _master_ or _admin_ status set to iff_up.  These are set after the interface is created. In this way, if the first netlink message comes with master set,  add it to `old-interface-map` and ignore this and all subsequent netlink message for this `ifIndex`. 

**Why I did it**
Without this fix, portsynd publishes the old interface events to stateDB. This resulted in failure of bridge creation and member port addition. Also, the newly created interfaces were not made admin-up
Related bug - https://github.com/Azure/sonic-buildimage/issues/1153

**How I verified it**
Restart swss in t0/t1-lag topology
Execute 'sudo brctl show', 'ip link show' and ensure bridge is added with member ports and links are coming up. 

**Details if related**
N/A

### DIFF
--- a/portsyncd/linksync.cpp
+++ b/portsyncd/linksync.cpp
@@ -90,14 +90,42 @@ void LinkSync::onMsg(int nlmsg_type, struct nl_object *obj)
                        nlmsg_type, key.c_str(), admin, oper, addrStr, ifindex, master);
     }
 
-    /* Insert or update the ifindex to key map */
-    m_ifindexNameMap[ifindex] = key;
-
     /* teamd instances are dealt in teamsyncd */
     if (type && !strcmp(type, TEAM_DRV_NAME))
     {
         return;
     }
+
+    /* In the event of swss restart, it is possible to get netlink messages during bridge
+     * delete, interface delete etc which are part of cleanup. These netlink messages for
+     * the front-panel interface must not be published or it will update the statedb with
+     * old interface info and result in subsequent failures. A new interface creation shall
+     * not have master or admin status iff_up. So if the first netlink message comes with these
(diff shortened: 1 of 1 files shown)


lguohan:
retest this please

lguohan:
retest this please

lguohan:
retest this please

lguohan:
retest this please

lguohan:
retest this please

portsyncd/linksync.cpp
```diff
@@ -90,15 +90,43 @@ void LinkSync::onMsg(int nlmsg_type, struct nl_object *obj)
...
+    /* In the event of swss restart, it is possible to get netlink messages during bridge
+     * delete, interface delete etc which are part of cleanup. These netlink messages for
+     * the front-panel interface must not be published or it will update the statedb with
+     * old interface info and result in subsequent failures. A new interface creation shall
+     * not have master or admin status iff_up. So if the first netlink message comes with these
+     * values set, it is considered to be happening during a cleanup process.
+     * Fix to ignore this and any further messages for this ifindex
+     */
+
+    static std::map<unsigned int, std::string> m_ifindexOldNameMap;
```
stcheng: no need for `std::`

yxieca: Why use static variable, can you add it as a member attribute of LinkSync?

When would m_ifindexOldNameMap be cleared? If it never gets cleared, will it stop working for the second restart?

I suggest you use unordered_map to get O(1) look up time.

PR #392, opened 2017-11-17
[cfgmgr] Use sonic-swss-common instead of exec redis-cli

ishidawataru:
this enforces portablity of vlanmgrd

Signed-off-by: Wataru Ishida <ishida.wataru@lab.ntt.co.jp>

**What I did**

Use `sonic-swss-common` instead of executing `redis-cli` to get MAC address of the device.

**Why I did it**

This is kinda refactoring but I'm actually facing a problem with current implementation.

I'm running `redis-server` in a diffrent docker container from a container which runs sonic's daemons.
Two containers shares a redis unix socket file and doesn't share a network namespace.
In this case, `vlanmgrd` crashes since current implementation executes `redis-cli` directly to get MAC address.

Since we are passing `DBConnector::DEFAULT_UNIXSOCKET` to `DBConnector `, using `sonic-swss-common` can solve this issue.
Also, we can easily add an option to `vlanmgrd` for choosing how to connect to the redis server in the future.

**How I verified it**

**Details if related**

### DIFF
--- a/cfgmgr/vlanmgrd.cpp
+++ b/cfgmgr/vlanmgrd.cpp
@@ -4,6 +4,7 @@
 #include <fstream>
 #include <iostream>
 #include <mutex>
+#include <algorithm>
 #include "dbconnector.h"
 #include "select.h"
 #include "exec.h"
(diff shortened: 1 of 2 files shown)


cfgmgr/vlanmgrd.cpp
```diff
@@ -66,6 +56,20 @@ int main(int argc, char **argv)
         DBConnector appDb(APPL_DB, DBConnector::DEFAULT_UNIXSOCKET, 0);
         DBConnector stateDb(STATE_DB, DBConnector::DEFAULT_UNIXSOCKET, 0);
 
+        /*
+         * swss service starts after interfaces-config.service which will have
+         * switch_mac set.
+         * Dynamic switch_mac update is not supported for now.
+         */
+        Table table(&cfgDb, "DEVICE_METADATA", CONFIGDB_TABLE_NAME_SEPARATOR);
+        std::vector<FieldValueTuple> ovalues;
```
stcheng: no need `std::`

PR #397, opened 2017-11-22
[PfcAclHanlder]: Egress drop by creating egress acl table and rule

sihuihan88:


lguohan:
@qiluo-msft  to review

orchagent/aclorch.cpp
```diff
@@ -912,18 +915,31 @@ bool AclTable::create()
...
+        attr.id = SAI_ACL_TABLE_ATTR_ACL_STAGE;
+        attr.value.s32 = SAI_ACL_STAGE_INGRESS;
+        table_attrs.push_back(attr);
+
+        attr.id = SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE;
+        attr.value.s32list.count = (uint32_t)(sizeof(range_types_list) / sizeof(range_types_list[0]));
+        attr.value.s32list.list = range_types_list;
+        table_attrs.push_back(attr);
+    }
+    else
```
qiluo-msft: else if (stage == ACL_STAGE_EGRESS)
How about others (including ACL_STAGE_UNKNOWN)?

orchagent/aclorch.cpp
```diff
@@ -912,18 +915,31 @@ bool AclTable::create()
...
+    if(stage == ACL_STAGE_INGRESS)
+    {
+        attr.id = SAI_ACL_TABLE_ATTR_ACL_STAGE;
+        attr.value.s32 = SAI_ACL_STAGE_INGRESS;
+        table_attrs.push_back(attr);
+
+        attr.id = SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE;
+        attr.value.s32list.count = (uint32_t)(sizeof(range_types_list) / sizeof(range_types_list[0]));
+        attr.value.s32list.list = range_types_list;
+        table_attrs.push_back(attr);
```
qiluo-msft: Is this attr only for ingress ACL?

orchagent/aclorch.cpp
```diff
@@ -912,18 +915,31 @@ bool AclTable::create()
     attr.value.booldata = true;
     table_attrs.push_back(attr);
 
+    if(stage == ACL_STAGE_INGRESS)
+    {
+        attr.id = SAI_ACL_TABLE_ATTR_ACL_STAGE;
+        attr.value.s32 = SAI_ACL_STAGE_INGRESS;
```
qiluo-msft: >attr.value.s32 [](start = 8, length = 14)

You could write a smaller scope if-else only for s32 field. Others are the same.

orchagent/aclorch.cpp
```diff
@@ -1424,6 +1443,13 @@ void AclOrch::doAclTableTask(Consumer &consumer)
                         SWSS_LOG_ERROR("Failed to process table ports for table %s", table_id.c_str());
                     }
                 }
+                else if (attr_name == TABLE_STAGE)
+                {
+                   if(!processAclTableStage(attr_value, newTable.stage))
```
qiluo-msft: add one more blank after 'if'

orchagent/aclorch.cpp
```diff
@@ -1613,6 +1639,24 @@ bool AclOrch::processAclTableType(string type, acl_table_type_t &table_type)
     return true;
 }
 
+bool AclOrch::processAclTableStage(string stage, acl_stage_type_t &acl_stage)
+{
+    SWSS_LOG_ENTER();
+
+    auto tt = aclStageLookUp.find(toUpper(stage));
```
qiluo-msft: >tt [](start = 9, length = 2)

tt stands for 'table type'. You should get your own name.

orchagent/aclorch.h
```diff
@@ -227,6 +241,7 @@ class AclTable {
     AclTable()
         : type(ACL_TABLE_UNKNOWN)
         , m_oid(SAI_NULL_OBJECT_ID)
+        , stage(ACL_STAGE_INGRESS)
```
qiluo-msft: >ACL_STAGE_INGRESS [](start = 16, length = 17)

default as ACL_STAGE_UNKNOWN?

orchagent/aclorch.cpp
```diff
@@ -1613,6 +1639,24 @@ bool AclOrch::processAclTableType(string type, acl_table_type_t &table_type)
...
+
+    if (tt == aclStageLookUp.end())
+    {
```
qiluo-msft: set acl_stage to ACL_STAGE_UNKNOWN? that is super safe.

orchagent/aclorch.cpp
```diff
@@ -935,7 +951,10 @@ bool AclTable::bind(sai_object_id_t portOid)
     assert(ports.find(portOid) != ports.end());
 
     sai_object_id_t group_member_oid;
-    if (!gPortsOrch->bindAclTable(portOid, m_oid, group_member_oid))
+
+    bool ingress = (stage == ACL_STAGE_INGRESS) ? true : false;
+
+    if (!gPortsOrch->bindAclTable(portOid, m_oid, group_member_oid, ingress))
```
qiluo-msft: >ingress [](start = 68, length = 7)

It's better to pass original acl_stage_type_t

orchagent/pfcactionhandler.h
```diff
@@ -73,10 +73,12 @@ class PfcWdAclHandler: public PfcWdActionHandler
         // class shared dict: ACL table name -> ACL table
         static std::map<std::string, AclTable> m_aclTables;
```
qiluo-msft: >m_aclTables [](start = 47, length = 11)

You may use 2 maps for acl tables. There is no benefit to mix ingress and egress into one.

orchagent/pfcactionhandler.cpp
```diff
@@ -143,16 +143,32 @@ PfcWdAclHandler::PfcWdAclHandler(sai_object_id_t port, sai_object_id_t queue,
 
     // There is one handler instance per queue ID
     string queuestr = to_string(queueId);
-    m_strTable = "Table_PfcWdAclHandler_" + queuestr;
-    m_strRule = "Rule_PfcWdAclHandler_" + queuestr;
+    m_strIngressTable = "IngressTable_PfcWdAclHandler_" + queuestr;
```
qiluo-msft: If there are 2 maps, you can still use original strings.

orchagent/aclorch.h
sihuihan88: It would be better to default as ACL_STAGE_INGRESS. As the default behavior is to create ingress ACL is no stage is claimed.

orchagent/aclorch.cpp
sihuihan88: Yes. This attr is only for Ingress ACL.

orchagent/aclorch.cpp
sihuihan88: done

orchagent/pfcactionhandler.h
sihuihan88: It would better to maintain one map for all aclTables from my perspective. The process of handling ingress/egress table are the same. Maintaining two maps every time looks redundant.

orchagent/pfcactionhandler.cpp
sihuihan88: AclOrch requires unique acl table name, so we need different names for ingress/egress acl table. Separating into two maps won't help. To reduce complexity a little bit, I use the same acl Rule name for both acl tables.

orchagent/aclorch.cpp
sihuihan88: I would prefer to pass boolean in this function. bindAclTable is a member function of PortsOrch. It doesn't have to know the detail of the existence acl_stage_stype_t which is more specific for aclOrch. Instead, a boolean to indicate ingress/egress should be enough for this interface.

orchagent/aclorch.cpp
sihuihan88: good catch. Updated.

sihuihan88:
Updated as reviews.

orchagent/portsorch.cpp
```diff
@@ -409,16 +417,25 @@ bool PortsOrch::bindAclTable(sai_object_id_t id, sai_object_id_t table_oid, sai_
...
+        bool ingress = acl_stage == ACL_STAGE_INGRESS ? true : false;
+        // If port ACL table group does not exist, create one
         sai_object_id_t bp_list[] = { SAI_ACL_BIND_POINT_TYPE_PORT };
 
         vector<sai_attribute_t> group_attrs;
         sai_attribute_t group_attr;
 
         group_attr.id = SAI_ACL_TABLE_GROUP_ATTR_ACL_STAGE;
-        group_attr.value.s32 = SAI_ACL_STAGE_INGRESS; // TODO: double check
+        group_attr.value.s32 = ingress ? SAI_ACL_STAGE_INGRESS : SAI_ACL_STAGE_EGRESS; // TODO: double check
```
qiluo-msft: >// TODO: double check [](start = 87, length = 21)

You already double checked. So remove this comment.

lguohan:
@sihuihan88 , can you add virtual switch based test?

sihuihan88:
Will add the test later.

PR #400, opened 2017-11-30
Fix to prevent OA crashes when configuring interfaces with overlapping subnets

rodnymolina:
To prevent orchagent from crashing with a simple ifconfig/ip command, i added some sanity-checking logic to verify that the new ip/mask being associated with an interface, does not overlap with any of the existing interfaces in the system. If an overlap is detected, the corresponding netlink message is discarded while an error log is dumped in syslog to alert user of their mistake.

When facing a subnet's overlapping scenario, OA will identify the offending prefix (meta_sai_validate_route_entry) and will throw an exception that will eventually terminate OA. See an example below...

Original interfaces -- Ethernet112 = 10.1.2.2/24, Ethernet116 = 10.2.2.2/24

I go ahead and configure an overlapping subnet by doing:

admin@lnos-x1-a-asw02:~$ sudo ifconfig Ethernet112 10.2.2.3 netmask 255.255.255.0

Nov 27 21:44:32.0 lnos-x1-a-asw02 NOTICE orchagent: :- removeSubnetRoute: Remove subnet route to 10.1.2.2/24 from Ethernet112
Nov 27 21:44:32.0 lnos-x1-a-asw02 NOTICE orchagent: :- removeIp2MeRoute: Remove packet action trap route ip:10.1.2.2
Nov 27 21:44:32.0 lnos-x1-a-asw02 ERR orchagent: :- meta_sai_validate_route_entry: object key SAI_OBJECT_TYPE_ROUTE_ENTRY:{"dest":"10.2.2.0/24","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000072"} already exists
Nov 27 21:44:32.0 lnos-x1-a-asw02 ERR orchagent: :- addSubnetRoute: Failed to create subnet route to 10.2.2.3/24 from Ethernet112, rv:-6
Nov 27 21:44:32.0 lnos-x1-a-asw02 ERR orchagent: :- main: Failed due to exception: Failed to create subnet route.
Nov 27 21:44:32.0 lnos-x1-a-asw02 INFO supervisord: orchagent terminate called without an active exception
Nov 27 21:44:33.170506 lnos-x1-a-asw02 INFO swss.sh[29508]: 2017-11-27
21:44:33,169 INFO exited: orchagent (terminated by SIGABRT (core dumped); not expected)
Nov 27 21:44:33.0 lnos-x1-a-asw02 INFO supervisord 2017-11-27 21:44:33,169 INFO exited: orchagent (terminated by SIGABRT (core
dumped); not expected)

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -143,6 +143,17 @@ void IntfsOrch::doTask(Consumer &consumer)
                 continue;
             }
 
+            /*
+             * Doing some sanity-checking to prevent interfaces with overlapping
+             * subnets from being able to be configured in the system. Notice
+             * that previous 'overlap' logic only takes care of 'ifconfig' special
+             * behavior -- no system-wide overlap verification is done above.
+             */
+            if (!validIntfAddress(ip_prefix, alias)) {
+                it = consumer.m_toSync.erase(it);
+                continue;
+            }
+
             addSubnetRoute(port, ip_prefix);
             addIp2MeRoute(ip_prefix);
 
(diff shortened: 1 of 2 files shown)


lguohan:
hi rodny, 

1. the code is very similar to code from line 127-144. I guess there might be some missing in previous code. If yes, can we combine them into one check? 

2. do we also observe this in virtual switch? Can you try it on virtual switch and add a test case for this?

rodnymolina:
Guohan, true, the code that you are referring to (127-144) looks similar to the one i added, but the problem we are solving there is different. In (127-144) case, the goal is to address the particular behavior that 'ifconfig' displays, and as such, the overlapping detection-logic only compares the new subnet being proposed with the one already configured in the interface being changed.

On the other hand, my code is addressing a different problem: how to prevent overlaps with all the other interfaces configured in the system. Also, and this is even more important, in (127-144) case, we want to temporarily skip the overlapping state, but we are not discarding it, whereas with my code, we are eliminating the offending state.

Writing a single function that combines both cases wouldn't be a good idea in my opinion, as we would be heavily overloading its semantics.

And no, i haven't checked VS yet, will give it a try.

orchagent/intfsorch.cpp
```diff
@@ -398,3 +409,83 @@ void IntfsOrch::removeIp2MeRoute(const IpPrefix &ip_prefix)
...
+ * interfaces:
+ *
+ * 1) Obviously, a configuration error caused by a user creating overlapping
+ *    interfaces.
+ *
+ * 2) A configuration change made through 'ifconfig' tool. In this case, the
+ *    following set of netlink messages is always generated. The exception to
+ *    this message sequence is the case in which the new ip-addr contains a /8
+ *    netmask -- in that case only 2.1) and 2.2) are expected. In all other
+ *    cases, this is the expected behavior:
```
lguohan: I am a little confused, because this netlink message sequencing is what previous code handled.

rodnymolina: Right, the previous code (127-144) addresses 'ifconfig' special msg sequencing, and by doing that it verifies if there's a subnet that overlaps with the _same_ interface that is being configured. However, the goal there is not to detect overlaps, but to ensure that the route-deletion is consumed before the final route-addition. It happens to use overlap-detection logic coz it serves the purpose of identifying the 'ifconfig' case, but that's not the goal there.

On the other, my new code truly takes care of preventing overlaps with the rest of the system interfaces. Now, due to the unfortunate 'ifconfig' behavior, i have to address that special case too. See an example below:

Eth1: 10.1.1.0/24
Eth2: 11.1.1.0/24

Now you do: ifconfig Eth2 10.2.2.0 netmask 255.255.255.0

There's no real overlap above, but 'ifconfig' behavior demands more thinking from our side. With today's code, upon arrival of netlink route-add msg for 10/8, we will go ahead and push that route down to asic. With this fix's code, we will see the overlap between 10/8 and 10.1.1.0/24, and we will skip this netlink message. Eventually, the proper route-add msg for 10.2.2.0/24 will arrive, and only then we will push it down to asic.

So as you can see, my code has to deal with this 'ifconfig' behavior too, that's why i went ahead and wrote a full description of that issue, and tested all possible combinations i could think of.

rodnymolina:
Closing this PR as it's a subset of larger-scope PR#437.

PR #401, opened 2017-11-30
use IP cmd to setup vlan_filter as a backup method

lguohan:
**What I did**
use IP cmd to setup vlan_filter as a backup method

**Why I did it**
echo will fail in virtual switch since /sys directory is read-only.
need to use ip command to setup the vlan_filtering which is not available in debian 8.
Once we move sonic to debian 9, we can use IP command by default 

**How I verified it**
build virtual switch and verified.

**Details if related**
Will generate an PR to add test for virtual switch. The test will be in another PR since we need to build the new virtual switch docker first.


lguohan:
@JipanYanga to review

PR #402, opened 2017-11-30
Ip mtu settings

hzheng5:
**What I did**
Get the mtu attribute value from linkcache when creating a new router interfaces. And pass it with as SAI_ROUTER_INTERFACE_ATTR_MTU to SAI.

**Why I did it**

As discussed in SONiC community meeting and followup meetings, the linux host interface MTU, port MTU on ASIC and IP MTU on ASIC should be consistent (Kernel MTU = IP MTU = Port MTU – 22 Bytes) 

Config DB Schema
'''
    "PORT": {
        "Ethernet8": {
            "alias": "Ethernet9",
             "mtu": "1500"
        },
        …
     },

    "PORTCHANNEL": {
        "PortChannel10": {
            "fallback": "true", 
            "members": [
                "Ethernet10"
            ], 
             "mtu": "1500"
        },
        …
     },
'''
**How I verified it**

Need libsai to support SAI_ROUTER_INTERFACE_ATTR_MTU, then check the ASIC dump of ip_mtu

**Details if related**

This commit requires https://github.com/Azure/sonic-swss-common/pull/166 as prerequisites


prsunny:
Changes looks ok to me. Just curious, are we handling mtu change? I think if mtu value is changed by user for a port, we set this to kernel. Are we also updating the corresponding SAI router interface mtu?

hzheng5:
Thanks Prince for the review.
No. We are unable to handling the if mtu changes for a router interfaces. We need SAI support, current the attribute can only be passed down to ASIC during the router interface create. There is no SAI API available to update the attribute of a router interfaces.
Thanks,Haiyang------------------------------------------------------------------From:Prince Sunny <notifications@github.com>Time:2017 Nov 30 (Thu) 17:08To:Azure/sonic-swss <sonic-swss@noreply.github.com>Cc:Haiyang <haiyang.z@alibaba-inc.com>; Author <author@noreply.github.com>Subject:Re: [Azure/sonic-swss] Ip mtu settings (#402)
Changes looks ok to me. Just curious, are we handling mtu change? I think if mtu value is changed by user for a port, we set this to kernel. Are we also updating the corresponding SAI router interface mtu?

lguohan:
retest this please

lguohan:
the PR name is IP MTU, but the ConfigDB schema is PORT MTU.

Since it is IP MTU, I think we should add in the INTERFACE:xxx: object. 

Check below
``` 
127.0.0.1:6379[4]> keys VLAN_INTER*
1) "VLAN_INTERFACE|Vlan2|172.0.0.1/26"
127.0.0.1:6379[4]> hgetall "VLAN_INTERFACE|Vlan2|172.0.0.1/26"
1) "mtu"
2) "9000"
```

Also do you have the code to listen to configdb?

lguohan:
@JipanYanga  to review.

I think we should modify the intfmgrd to listen to configdb, meanwhile write to APPDB directly. We should not listen to netlink message any more.

hzheng5:
@lguohan 

Thanks for the code review. 

As we discussed earlier, there should be only one source of the MTU settings in config DB  for linux kernel host interface, port/LAG, and router interface created on the port/LAG. So with my previous merged pull request (https://github.com/Azure/sonic-swss/pull/402/commits/51158b3e7b68b37f454bbeccee3d2242d8d9aaa7), only port syncd listens to CONFIG DB, and update the port MTU and Kernel MTU. Then when a router interfaces is created on the port, it get the MTU from netlink, and pass it down to router_interface_create.

For vlan/vlan interfaces MTU settings, let me discuss with Jipan offline first and then update with you.

BTW, how to restart the test? The other pull request has been merged, and the build should be fixed now.

Thanks,
Haiyang

intfsyncd/intfsync.cpp
```diff
@@ -46,6 +51,12 @@ void IntfSync::onMsg(int nlmsg_type, struct nl_object *obj)
         return;
 
     key = LinkCache::getInstance().ifindexToName(rtnl_addr_get_ifindex(addr));
+    /* Get mtu from link */
+    name = key.c_str();
+    link = LinkCache::getInstance().getLinkByName(name);
+    if (link != NULL)
+        if_mtu = rtnl_link_get_mtu(link);
```
lguohan: use {} even for a single line

intfsyncd/intfsync.cpp
```diff
@@ -58,7 +69,9 @@ void IntfSync::onMsg(int nlmsg_type, struct nl_object *obj)
     std::vector<FieldValueTuple> fvVector;
     FieldValueTuple f("family", family);
     FieldValueTuple s("scope", scope);
+    FieldValueTuple m("if_mtu", to_string(if_mtu));
```
lguohan: can we use mtu instead of if_mtu?

orchagent/intfsorch.cpp
```diff
@@ -95,6 +95,13 @@ void IntfsOrch::doTask(Consumer &consumer)
             auto it_intfs = m_syncdIntfses.find(alias);
             if (it_intfs == m_syncdIntfses.end())
             {
+                for (auto i : kfvFieldsValues(t))
+                {
+                    /* Set router interface mtu */
+                    if (fvField(i) == "if_mtu")
+                        port.m_if_mtu = stoul(fvValue(i));
```
lguohan: use {} even for single line.

orchagent/port.h
```diff
@@ -64,6 +65,7 @@ class Port
     Type                m_type;
     int                 m_index = 0;    // PHY_PORT: index
     int                 m_ifindex = 0;
+    uint32_t            m_if_mtu = DEFAULT_IF_MTU;
```
lguohan: using m_ip_mtu instead of m_if_mtu will be more clear.

lguohan:
@hzheng5 , even though SAI does not support router interface MTU set, our orchagent code should still support that MTU change in my opinion, then later we can close it by adding SAI support for this MTU change which I think should not be difficult.

intfsyncd/intfsync.cpp
```diff
@@ -46,6 +51,12 @@ void IntfSync::onMsg(int nlmsg_type, struct nl_object *obj)
         return;
 
     key = LinkCache::getInstance().ifindexToName(rtnl_addr_get_ifindex(addr));
+    /* Get mtu from link */
+    name = key.c_str();
+    link = LinkCache::getInstance().getLinkByName(name);
+    if (link != NULL)
```
lguohan: when will this condition happen, by the time you get an interface, you should have the link?

orchagent/port.h
```diff
@@ -11,6 +11,7 @@ extern "C" {
 #include <map>
 
 #define DEFAULT_PORT_VLAN_ID    1
+#define DEFAULT_IF_MTU        9100
```
lguohan: why do we have DEFAULT_IF_MTU in two places? can we have them in one place?

lguohan:
@prsunny , I do not see router interface mtu support in broadcom sai, can you confirm? It looks like this PR will break the broadcom platform if we check in.

lguohan:
in this PR, you introduced a new field (if_mtu) in INTERFACE object, which is not necessary. We already have an MTU field in the PORT_TABLE. I think we can use that. 

in the intforch.cpp, when it tries to create router interface, it should check the port_mtu, and setup the router interface mtu accordingly. then the mtu changes, the portsorch will get the mtu notification change, and if it is a router interface, it should also change the router interface mtu. Then, we do not need to change intfsyncd.cpp at all.

lguohan:
also, can you add swss integration tests in the next iteration for the mtu changes. Examples are https://github.com/Azure/sonic-swss/tree/master/tests

lguohan:
retest this please

lguohan:
@hzheng5 , to restart the test, just add a comment "retest this please"

prsunny:
@lguohan, @hzheng5, from what I've observed, there is NO Broadcom support for SAI_ROUTER_INTERFACE_ATTR_MTU in the current code. This check-in will break the Broadcom platforms.

intfsyncd/intfsync.cpp
hzheng5: The if_mtu is defined in the following schema, please me know if it is OK to update the schema for this
https://github.com/Azure/sonic-swss/blob/d1fdf399e1e0e2e0d71fad1bce2ea7babee2eb48/doc/swss-schema.md

intfsyncd/intfsync.cpp
hzheng5: Yes. The router interface is attached to the link, by the time we create the router interface, we should have the link ready.

intfsyncd/intfsync.cpp
hzheng5: sure. will do.

orchagent/intfsorch.cpp
hzheng5: sure. will do. thanks

orchagent/port.h
hzheng5: sure. will combine these two definitions in one header file. Thanks

orchagent/port.h
hzheng5: The if_mtu is defined in the following schema
https://github.com/Azure/sonic-swss/blob/d1fdf399e1e0e2e0d71fad1bce2ea7babee2eb48/doc/swss-schema.md

So I'm trying to keep it consistent from intfsyncd->app_db->intfsorch->sai by using "if_mtu" instead of "ip_mtu".
Please let me know if you think it's better to change "if_mtu" to "ip_mtu" in both swss schema and everywhere else.

hzheng5:
Thanks @prsunny  any idea when SAI_ROUTER_INTERFACE_ATTR_MTU will be added to the brcm libsai? Will it be available before Dec 15 release?

hzheng5:
retest this please

hzheng5:
@lguohan  Thanks for your code review.  I have updated the diff and tried to answer your comments. Please see inline.

For your last comment:
"in this PR, you introduced a new field (if_mtu) in INTERFACE object, which is not necessary. We already have an MTU field in the PORT_TABLE. I think we can use that.

in the intforch.cpp, when it tries to create router interface, it should check the port_mtu, and setup the router interface mtu accordingly. then the mtu changes, the portsorch will get the mtu notification change, and if it is a router interface, it should also change the router interface mtu. Then, we do not need to change intfsyncd.cpp at all."

I'm adding the changes in intfsyncd because I want to follow the complete config path as other router interfaces attributes as from intfsyncd->app_db->intfsorch->asic_db. And the if_mtu is already defined in swss schema for router interfaces.  If we bypass intfsyncd, app_db won't have the if_mtu value for  router_interfaces, and it's not consistent with other router interfaces attributes.

Sure. I will add some swss intergration test shortly.

Thanks,
Haiyang

stcheng:
@hzheng5 Hello Haiyang. Yes. The SAI_ROUTER_INTERFACE_ATTR_MTU attribute is not supported right now and will be available AFTER Dec 15 when the SAI header version moves to 1.2.

hzheng5:
OK. I see. Thanks Shuotian for the update.------------------------------------------------------------------From:Shuotian Cheng <notifications@github.com>Time:2017 Dec 7 (Thu) 22:41To:Azure/sonic-swss <sonic-swss@noreply.github.com>Cc:Haiyang <haiyang.z@alibaba-inc.com>; Mention <mention@noreply.github.com>Subject:Re: [Azure/sonic-swss] Ip mtu settings (#402)
@hzheng5 Hello Haiyang. Yes. The SAI_ROUTER_INTERFACE_ATTR_MTU attribute is not supported right now and will be available AFTER Dec 15 when the SAI header version moves to 1.2.

intfsyncd/intfsync.cpp
lguohan: I prefer to use "mtu" in this context it is "ip_mtu", but we do not need to prepend ip before the mtu.

lguohan:
@hzheng5 , it is not neccessary to have two mtu attribute for the same physical interface, one for the port mtu and one for the router interface. If you look at the other router configuration (like cisco), you won't see it. 

Is there any flaw in the workflow I suggested? it will be simpler for people to understand. since there is only one mtu for the port. If the port becomes a port-based router interface, the same mtu value will be applied to the router interface mtu.

hzheng5:
@lguohan Sure. The workflow you suggested sounds good to me. Will remove the intfsyncd and app_DB changes. 
Thanks, Haiyang

PR #403, opened 2017-12-01
Output recording start time into the record file

pavel-shirshov:
**What I did**
Write start recording time into the record file.

**Why I did it**
Otherwise it's hard to find where every record was started.

**How I verified it**
Output in a recording file

**Details if related**

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -12,6 +12,9 @@ extern "C" {
 #include <getopt.h>
 #include <unistd.h>
 
+#include <sys/time.h>
+#include "timestamp.h"
+
 #include <sairedis.h>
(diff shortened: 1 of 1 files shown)


orchagent/main.cpp
```diff
@@ -170,6 +173,7 @@ int main(int argc, char **argv)
             SWSS_LOG_ERROR("Failed to open SwSS recording file %s", gRecordFile.c_str());
             exit(EXIT_FAILURE);
         }
+        gRecordOfs << getTimestamp() << "|" << "recording on" << endl;
```
jleveque: Suggest changing message to "recording started"

pavel-shirshov:
retest this please

PR #406, opened 2017-12-04
Save fdb entries inside of fdborch until ports are ready

pavel-shirshov:
**What I did**
Add a logic to save FDB entries inside of the FDBOrch when then can't be inserted right away. When corresponding port becomes a VLAN member FDBOrch will try to insert them again.

**Why I did it**
Currently FDBOrch will try to insert FDB entries continuously until they inserted. It consumes 100% cpu.

**How I verified it**
Check syslog output comparing timings for addVlanMember and addFdbEntry entries.

**Details if related**


lguohan:
can you add a swss integration test for this? https://github.com/Azure/sonic-swss/blob/master/tests/README.md

you can add an fdb entry then create the vlan member, and check the asic db.

pavel-shirshov:
Yes, I can. But the test you're describing will pass either with or without my patch.

lguohan:
I mena add fdb entry before vlan member creation, verify in asic db that fdb entry is not inserted, then create vlan member and verify the fdb entry is added.

orchagent/fdborch.cpp
```diff
@@ -157,6 +179,21 @@ void FdbOrch::doTask(Consumer& consumer)
...
 }
 
+void FdbOrch::updateVlanMember(const VlanMemberUpdate& update)
+{
+    SWSS_LOG_ENTER();
+
+    if (!update.add) return; // we need only additions
+
+    string port_name = update.member.m_alias;
+    auto fdb_list = std::move(saved_fdb_entries[port_name]);
```
prsunny: Just thinking, lets say port 1 is member of Vlan 1 and Vlan 2 and fdb_entries are saved per port. Now when` port 1` is added to` Vlan 1`, is it the case that fdb entries on` Vlan 2`, `port 1` is also removed and added even-though we still didn't get a member add of `port 1` to `Vlan 2`?

pavel-shirshov: Valid point. Thanks Prince. I'll add a fix ASAP.

orchagent/fdborch.cpp
```diff
@@ -157,6 +179,21 @@ void FdbOrch::doTask(Consumer& consumer)
     }
 }
 
+void FdbOrch::updateVlanMember(const VlanMemberUpdate& update)
+{
+    SWSS_LOG_ENTER();
+
+    if (!update.add) return; // we need only additions
```
lguohan: please use {} to quote.

orchagent/fdborch.cpp
pavel-shirshov: So I've double checked this case.
I programmed following logic here. As soon as we get vlan update we try to reinsert fdb entries on the port. If something wrong with some entry, it will be reinserted back to the hash. I'll add comments describing the logic.

lguohan:
retest this please

tests/test_fdb.py
```diff
@@ -0,0 +1,149 @@
...
+def create_entry_tbl(db, table, separator, key, pairs):
+    tbl = swsscommon.Table(db, table, separator)
+    create_entry(tbl, key, pairs)
+
+def create_entry_pst(db, table, separator, key, pairs):
+    tbl = swsscommon.ProducerStateTable(db, table)
+    create_entry(tbl, key, pairs)
+
+
+def get_map_port_id_2_iface_name(asic_db):
```
lguohan: >get_map_port_id_2_iface_name [](start = 4, length = 28)

this is already done in conftest.py

dvs.asicdb.portoidmap

lguohan:
looks good to me, only a minor issue in the test code.

PR #409, opened 2017-12-06
[portsyncd] support read port configuration from configDB

ishidawataru:
**What I did**

read port configuration from configDB when `-p` is not specified.

**Why I did it**

configDB is the master data of port configuration.
We can simply use it instead of passing dedicated configuration file for `portsyncd`

**How I verified it**

Run `portsyncd` without `-p` option.

**Details if related**

### DIFF
--- a/portsyncd/portsyncd.cpp
+++ b/portsyncd/portsyncd.cpp
@@ -15,8 +15,6 @@
 #include "subscriberstatetable.h"
 #include "exec.h"
 
-#define DEFAULT_PORT_CONFIG_FILE     "port_config.ini"
-
 using namespace std;
 using namespace swss;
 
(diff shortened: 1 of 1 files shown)


portsyncd/portsyncd.cpp
```diff
@@ -85,7 +86,11 @@ int main(int argc, char **argv)
         netlink.registerGroup(RTNLGRP_LINK);
         cout << "Listen to link messages..." << endl;
 
-        handlePortConfigFile(p, port_config_file);
+        if (port_config_file.empty()) {
+            handlePortConfigFile(p, port_config_file);
```
lguohan: use the port_config_file when it is empty?

portsyncd/portsyncd.cpp
```diff
@@ -153,6 +158,35 @@ int main(int argc, char **argv)
...
+    cout << "Get port configuration from ConfigDB..." << endl;
+
+    Table table(&cfgDb, CFG_PORT_TABLE_NAME, CONFIGDB_TABLE_NAME_SEPARATOR);
+    std::vector<FieldValueTuple> ovalues;
+    std::vector<string> keys;
+    table.getKeys(keys);
+    for ( auto &k : keys ) {
+        table.get(k, ovalues);
+        vector<FieldValueTuple> attrs;
+        for ( auto &v : ovalues ) {
```
lguohan: please make all bracket into separate lines.

portsyncd/portsyncd.cpp
ishidawataru: `string::empty()` returns 0 if the string is empty

http://www.cplusplus.com/reference/string/string/empty/

portsyncd/portsyncd.cpp
ishidawataru: done

portsyncd/portsyncd.cpp
```diff
@@ -85,7 +86,12 @@ int main(int argc, char **argv)
         netlink.registerGroup(RTNLGRP_LINK);
         cout << "Listen to link messages..." << endl;
 
-        handlePortConfigFile(p, port_config_file);
+        if (port_config_file.empty())
```
lguohan: true if the string length is 0, false otherwise.

ishidawataru: my bad. fixed

PR #413, opened 2017-12-11
[pfcwd]: create queue pfcwd handler only when it doesn't exist

sihuihan88:

**What I did**
Create queue pfcwd handler only when it doesn't exist.
**Why I did it**
Redis pub/sub might have latency and it's possible that orchagent receives multiple storm/restore notification continuously. This PR is to prevent orchagent keeps creating new handler smart pointer and removing existing one in such case. Instead, it will only honor the first notification and do the corresponding operation.
**How I verified it**
Verified on DUT
**Details if related**


lguohan:
can you fill those fields required by the PR template?

PR #417, opened 2017-12-15
Add Buffer Config Manager

andriymoroz-mlnx:

**What I did**
PG Buffers update on port speed change

**Why I did it**
Added buffer manager to track port speed and update PG profiles

**How I verified it**
Checked PG profiles on port on different speed set
Manually tested typical setups configuration (t1, t1-lag, t0)

**Details if related**
See: https://github.com/Azure/SONiC/wiki/Run-Time-Buffers-Configuration-Update-design

**NOTE1:** 
Dynamic profile deletion if not used is not implemented (requires more tests and investigation)
Does not affect main functionality. Suggest to implement it as an improvement.

**NOTE2:**
speed+cable-to-profile mapping implemented in .ini file not in an additional DB as described in the design. (will be in sonic-buildimage PR)


lguohan:
@yxieca to review

cfgmgr/buffermgr.cpp
```diff
@@ -0,0 +1,201 @@
...
+        {
+            for (auto i : kfvFieldsValues(t))
+            {
+                if (table_name == CFG_PORT_CABLE_LEN_TABLE_NAME)
+                {
+                    // receive and cache cable length table
+                    doCableTask(fvField(i), fvValue(i));
+                }
+                if (table_name == CFG_PORT_TABLE_NAME && fvField(i) == "speed")
+                {
```
yxieca: Will the field ever != "speed"?

If yes, why ignoring them is correct? Please at least add comment to state reason why it is safe to ignore them.

cfgmgr/buffermgr.h
```diff
@@ -0,0 +1,54 @@
...
+#include "dbconnector.h"
+#include "producerstatetable.h"
+#include "orch.h"
+
+#include <map>
+#include <string>
+
+namespace swss {
+
+#define INGRESS_LOSSLESS_PG_POOL_NAME "ingress_lossless_pool"
```
yxieca: Is assuming ingress lossless pg pool name the right thing to do?

Can this code support both shared ingress pg pool and separate pg pools for lossless and lossy traffic?

cfgmgr/buffermgr.h
```diff
@@ -0,0 +1,54 @@
...
+#define LOSSLESS_PGS "3-4"
```
yxieca: By defining lossless pgs to 3-4 in code, do we lose the flexibility of choosing any pg as lossless?

I know that in practice we are always have been using 3-4 as lossless pgs, but losing the capability of changing is a bit concerning.

cfgmgr/buffermgrd.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+            ret = s.select(&sel, &fd, SELECT_TIMEOUT);
+            if (ret == Select::ERROR)
+            {
+                SWSS_LOG_NOTICE("Error: %s!", strerror(errno));
+                continue;
+            }
+            if (ret == Select::TIMEOUT)
+            {
+                buffmgr.doTask();
```
yxieca: Do you think calling drain() makes more sense?

cfgmgr/buffermgr.cpp
andriymoroz-mlnx: yes, we receive updates for other fields as well. Actually every time at least one field in the table gets updated, all subscribers receive notifications about all fields. buffer manager is interested in speed only. I'll add a comment.

cfgmgr/buffermgrd.cpp
andriymoroz-mlnx: Probably, I'll check. Just used the solution similar to `initmgrd`.

cfgmgr/buffermgr.h
andriymoroz-mlnx: currently we support buffers update for profiles which use ingress lossless pools only

cfgmgr/buffermgr.h
andriymoroz-mlnx: theoretically PGs numbers can be fetched from the PORT_QOS_MAP:pfc_enable.
@lguohan, can we rely on the fact that queues specified in PORT_QOS_MAP:pfc_enable ("3,4") will always be the ones for which we need to update ingress lossless profile on speed change?

lguohan:
I would really like to see swss integration tests on this PR.

cfgmgr/buffermgrd.cpp
andriymoroz-mlnx: doTask() in base class calls drain:
```
void Orch::doTask()
{
    for(auto &it : m_consumerMap)
    {
        it.second->drain();
    }
}
```

cfgmgr/buffermgr.cpp
yxieca: Thanks for the explanation!

cfgmgr/buffermgrd.cpp
yxieca: I see. Thanks for digging it up!

cfgmgr/buffermgr.h
yxieca: Does it prevent shared pg pool?

cfgmgr/buffermgrd.cpp
yxieca: I think doTask is virtual function and the call will be dispatched to derived class. but calling doTask() is fine.

cfgmgr/buffermgr.h
yxieca: I just had a discussion with Guohan, this change will affect Broadcom platform in negative way.

On some Broadcom platforms, due to the MMU memory size limitation, we have to allocate a shared buffer pool for all pgs (lossy and lossless), and create 2 buffer profiles on top of the shared pg pool.

I think by hard coding the pg pool name, here is a potential risk for Broadcom platforms that requires shared buffer pool.

There is a buffers.json file checked in for Broadcom TH chip: https://github.com/Azure/sonic-swss/blob/master/swssconfig/sample/th.64ports.buffers.json

Please make sure your code works on this scenario as well.

Thanks,
Ying

andriymoroz-mlnx: ok, pool_name is yet another candidate to get to the new table with QOS/BUFFER configuration (besides pg/queue index to generate profile for)
But existing implementation will still work with this platform.
Since the configuration is static no new profiles will be generated on speed change.

andriymoroz-mlnx: I've updated config for Force10-S6100 (previously by mistake used `td2.32ports.buffers.json` for it)

andriymoroz-mlnx:
VS test requires changes in VS docker (https://github.com/Azure/sonic-buildimage/pull/1250/commits/ff8332feea5a8372833d7cdd2beb0893706acd55) which require changes from this PR
probably need to be reverted and added later to avoid deadlock
so it is for preview only

andriymoroz-mlnx:
Changes with the test reverted and moved to the separate PR
https://github.com/Azure/sonic-swss/pull/432

cfgmgr/buffermgr.h
yxieca: I see. thanks for the update.

cfgmgr/buffermgr.h
yxieca: We had a discussion, we decided to move forward with this change for now and revisit this area and remove the hard coded pg numbers in the future.

cfgmgr/buffermgrd.cpp
yxieca: I see. Thanks :-)

yxieca:
It appears that this review might need to be updated before it can be merged?

yxieca:
retest please

yxieca:
please retest

yxieca:
retest this please

PR #420, opened 2017-12-18
[Tests]: Add extra options to the test runner

pavel-shirshov:
**What I did**
Add two extra options to the test runner:
1. Change orchagent log level
2. Save logs after each test to a host machine

**Why I did it**
It allows us to get reason why tests fails faster

**How I verified it**
Run it manually

**Details if related**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import os
+import os.path
 import re
 import time
 import docker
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -206,9 +211,28 @@ def init_asicdb_validator(self):
...
     name = request.config.getoption("--dvsname")
+    get_logs = request.config.getoption("--get_logs")
+    orch_log_level = request.config.getoption("--orchagent_loglevel")
+
     dvs = DockerVirtualSwitch(name)
+    dvs.ctn.exec_run("swssloglevel -l %s -c orchagent" % orch_log_level)
+
     yield dvs
+
+    if get_logs:
```
lguohan: this is getting all logs. Usually, when you done with an appdb input, you just want to get new logs. I think we need to provide api to do that.

pavel-shirshov: What do you mean by "new logs"? Do you mean we want to see only logs which were created during a test?

lguohan: I mean new log entries generated after you run certain commands.

PR #424, opened 2018-01-10
[vrf orchagent]: Add Request parser class and VRF orchagent based on it

pavel-shirshov:
**What I did**
Implemented VRF orchagent.

**Why I did it**
To support virtual routers configurations on SONiC.

**How I verified it**
Usings tests/tests_vrf.py

**Details if related**
I've added a request parser (see orchagent/request_parser.cpp, orchagent/request_parser.h and tests/request_parser-ut.cpp). When using this class for any orchagent the class can parse and validate 
request parameters for an orchagent. It allows us to create new orchagents faster with less errors.

### DIFF
--- a/orchagent/request_parser.h
+++ b/orchagent/request_parser.h
@@ -98,13 +98,13 @@ private:
     const request_description_t& request_description_;
     char key_separator_;
     bool is_parsed_;
+    size_t number_of_key_items_;
 
     std::string operation_;
-    size_t number_of_key_items_;
     std::string full_key_;
-    std::unordered_set<std::string> attr_names_;
     std::unordered_map<int, std::string> key_item_strings_;
     std::unordered_map<int, MacAddress> key_item_mac_addresses_;
+    std::unordered_set<std::string> attr_names_;
     std::unordered_map<std::string, std::string> attr_item_strings_;
     std::unordered_map<std::string, bool> attr_item_bools_;
     std::unordered_map<std::string, MacAddress> attr_item_mac_addresses_;
--- a/orchagent/request_parser.h
+++ b/orchagent/request_parser.h
@@ -3,11 +3,11 @@
 
 typedef enum _request_types_t
 {
+    REQ_T_NOT_USED,
     REQ_T_BOOL,
     REQ_T_STRING,
     REQ_T_MAC_ADDRESS,
     REQ_T_PACKET_ACTION,
-    REQ_T_NOT_USED,
 } request_types_t;
 
 typedef struct _request_description
--- a/orchagent/request_parser.h
+++ b/orchagent/request_parser.h
@@ -19,22 +19,8 @@ typedef struct _request_description
 class Request
 {
 public:
-    Request(const request_description_t& request_description, const char key_separator)
-        : request_description_(request_description),
-          key_separator_(key_separator),
-          is_parsed_(false),
-          number_of_key_items_(request_description.key_item_types.size())
-    {
-    }
-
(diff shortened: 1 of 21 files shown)


orchagent/vrforch.cpp
```diff
@@ -0,0 +1,174 @@
...
+        else
+        {
+            SWSS_LOG_ERROR("Logic error: Unknown attribute: %s", name.c_str());
+            continue;
+        }
+        attrs.push_back(attr);
+    }
+
+    sai_object_id_t router_id;
+    sai_status_t status = sai_virtual_router_api->create_virtual_router(&router_id,
```
lguohan: >create_virtual_router [](start = 50, length = 21)

if the vrf exists, then we need to do a set operation, i think.

orchagent/request_parser.h
```diff
@@ -0,0 +1,114 @@
...
+    REQ_T_MAC_ADDRESS,
+    REQ_T_PACKET_ACTION,
+    REQ_T_NOT_USED,
+} request_types_t;
+
+typedef struct _request_description
+{
+    std::vector<request_types_t> key_item_types;
+    std::unordered_map<std::string, request_types_t> attr_item_types;
+    std::vector<std::string> mandatory_attr_items;
```
lguohan: >std::vector<std::string> mandatory_attr_items; [](start = 4, length = 46)

I am ok with current implementation. 

but, it would be more extensible to add different property for the attribute item, for example, the type, whether it is mandatory or not, whether it depends on other attribute item of not.

tests/test_vrf.py
```diff
@@ -0,0 +1,180 @@
...
+    # delete the created vrf entry
+    delete_entry_tbl(
+        conf_db,
+        "VRF", '|', "vrf1",
+    )
+
+    # check that the vrf entry was removed
+    assert how_many_entries_exist(asic_db, "ASIC_STATE:SAI_OBJECT_TYPE_VIRTUAL_ROUTER") == 1, "The vrf wasn't removed"
+
+    # check that the correct vrv entry was removed
```
lguohan: >vrv [](start = 29, length = 3)

vrf

orchagent/vrforch.cpp
pavel-shirshov: You mean 'update' operation? Don't create a new vrf, but update existing one?

lguohan: yes

pavel-shirshov: ok. In this case 'update' operation might have two semantics:
1. remove old attributes and add the new one from the request. Currently impossible in SAI (no function to remove attributes or setting them into default value).
2. merge old attributes with new ones from the request..

I'll implement the second approach.

tests/request_parser_ut.cpp
```diff
@@ -0,0 +1,663 @@
...
+    EXPECT_STREQ(request.getFullKey().c_str(), "key1:key2");
+    EXPECT_STREQ(request.getKeyString(0).c_str(), "key1");
+    EXPECT_STREQ(request.getKeyString(1).c_str(), "key2");
+    EXPECT_TRUE(request.getAttrFieldNames() == (std::unordered_set<std::string>{"v4", "v6"}));
+    EXPECT_FALSE(request.getAttrBool("v4"));
+    EXPECT_FALSE(request.getAttrBool("v6"));
+}
+// FIXME: add a test to test an unsupported key type and attr type REQ_T_NOT_USED
+// FIXME: expect_no_throw for constructors
+// FIXME: Add a test to test 'empry' attribute
```
lguohan: empty

PR #425, opened 2018-01-11
[portsorch]: Only add port stat IDs that are needed to query

sihuihan88:

**What I did**
Only query needed port stat sai attributes.
**Why I did it**
The hardcoded for loop was not the good option to query all
IDs defined in the SAI header. While only limited partial of
the counters are needed, we don't need to query each of them
**How I verified it**
Verified on DUT
**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -39,7 +39,7 @@ static map<string, sai_port_fec_mode_t> fec_mode_map =
     { "fc", SAI_PORT_FEC_MODE_FC }
 };
 
-static const vector<sai_port_stat_t> portStatIds =
+const vector<sai_port_stat_t> portStatIds =
 {
     SAI_PORT_STAT_IF_IN_OCTETS,
     SAI_PORT_STAT_IF_IN_UCAST_PKTS,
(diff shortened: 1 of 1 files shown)


orchagent/portsorch.cpp
```diff
@@ -39,6 +39,37 @@ static map<string, sai_port_fec_mode_t> fec_mode_map =
     { "fc", SAI_PORT_FEC_MODE_FC }
 };
 
+static const vector<sai_port_stat_t> portStatIds =
```
pavel-shirshov: static is not required here.
you may check the links below for explanations
https://stackoverflow.com/questions/11055802/static-and-extern-global-variables-in-c-and-c
http://en.cppreference.com/w/cpp/language/storage_duration

orchagent/portsorch.cpp
```diff
@@ -928,9 +959,9 @@ bool PortsOrch::initPort(const string &alias, const set<int> &lane_set)
 
                 std::string delimiter = "";
                 std::ostringstream counters_stream;
-                for (int cntr = SAI_PORT_STAT_IF_IN_OCTETS; cntr <= SAI_PORT_STAT_PFC_7_ON2OFF_RX_PKTS; ++cntr)
+                for (auto it = portStatIds.begin(); it !=  portStatIds.end(); it++)
```
pavel-shirshov: for (const auto& id: portStatIds) // Shorter version
{
}
if you have a container with a primitive type we could write it shorter
for(auto id: portStatIds)
{
}

Another suggestion is never use postfix increment/decrement. In these cases on every operation the object will be copied by its copy constructor.

P.S.: My both comments doesn't mean you must change your code. They're only suggestions.

lguohan: yes, it is much easier to read.

PR #426, opened 2018-01-12
[fdborch]Add the support for a clear fdb cli

keboliu:
**What I did**
revise the FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_object_id_t bridge_port_id) function to support a clear fdb cli

**Why I did it**
the current implementation is assuming that SAI will send all the flushed fdb entries to upper layers, but actually SAI only send one event to indicate the flush type, like flush all, flush fdb entries in certain port or certain vlan. upper layer need to find out all the flushed fdb entries by itself and then update it.

**How I verified it**

tested on MLNX plat form with various flush fdb cases.

**Details if related**
design doc see [https://github.com/Azure/SONiC/wiki/SONiC-Clear-FDB-CLI-Design](url)


orchagent/fdborch.cpp
```diff
@@ -34,21 +34,66 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
 
         (void)m_entries.insert(update.entry);
         SWSS_LOG_DEBUG("FdbOrch notification: mac %s was inserted into vlan %d", update.entry.mac.to_string().c_str(), entry->vlan_id);
+        
+        for (auto observer: m_observers)
+        {
+            observer->update(SUBJECT_TYPE_FDB_CHANGE, static_cast<void *>(&update));
```
pavel-shirshov: you can cast to void pointer without any cast mechanisms.
Check https://stackoverflow.com/questions/18929225/casting-class-pointer-to-void-pointer
So
observer->update(SUBJECT_TYPE_FDB_CHANGE, &update); is enough

orchagent/fdborch.cpp
```diff
@@ -34,21 +34,66 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
...
-    case SAI_FDB_EVENT_FLUSHED:
     case SAI_FDB_EVENT_MOVE:
         update.add = false;
 
         (void)m_entries.erase(update.entry);
         SWSS_LOG_DEBUG("FdbOrch notification: mac %s was removed from vlan %d", update.entry.mac.to_string().c_str(), entry->vlan_id);
+        
+        for (auto observer: m_observers)
+        {
+            observer->update(SUBJECT_TYPE_FDB_CHANGE, static_cast<void *>(&update));
```
pavel-shirshov: you can cast to void pointer without any cast mechanisms.
Check https://stackoverflow.com/questions/18929225/casting-class-pointer-to-void-pointer
So
observer->update(SUBJECT_TYPE_FDB_CHANGE, &update); is enough

orchagent/fdborch.cpp
```diff
@@ -34,21 +34,66 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
...
+        }
+      
+        break;
+        
+    case SAI_FDB_EVENT_FLUSHED:
+        if( !bridge_port_id && !(entry->vlan_id))
```
pavel-shirshov: Parenthesis around entry->vlan_id are redundant

orchagent/fdborch.cpp
```diff
@@ -34,21 +34,66 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
...
+                update.entry.vlan = itr->vlan;
+                update.add = false;
+                
+                m_entries.erase(itr);
+                
+                SWSS_LOG_DEBUG("FdbOrch notification: mac %s was removed", update.entry.mac.to_string().c_str());
+                
+                for (auto observer: m_observers)
+                {
+                    observer->update(SUBJECT_TYPE_FDB_CHANGE, static_cast<void *>(&update));
```
pavel-shirshov: you can cast to void pointer without any cast mechanisms.
Check https://stackoverflow.com/questions/18929225/casting-class-pointer-to-void-pointer
So
observer->update(SUBJECT_TYPE_FDB_CHANGE, &update); is enough

orchagent/fdborch.cpp
```diff
@@ -34,21 +34,66 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
...
+                }
+            }
+        }
+        else if(bridge_port_id && !(entry->vlan_id))
```
pavel-shirshov: redundand parenthesis around entry->vlan_id

orchagent/fdborch.cpp
```diff
@@ -34,21 +34,66 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
...
+        {
+            /*this is a placeholder for flush port fdb case, not supported yet.*/
+            SWSS_LOG_ERROR("FdbOrch notification: not supported flush type");
+        }
+        else if(!bridge_port_id && (entry->vlan_id))
```
pavel-shirshov: redundand parenthesis around entry->vlan_id

orchagent/fdborch.cpp
```diff
@@ -34,21 +34,66 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
...
+        {
+            observer->update(SUBJECT_TYPE_FDB_CHANGE, static_cast<void *>(&update));
+        }
+      
+        break;
+        
+    case SAI_FDB_EVENT_FLUSHED:
+        if( !bridge_port_id && !(entry->vlan_id))
+        {
+            for(set<FdbEntry>::iterator itr = m_entries.begin(); itr != m_entries.end(); ++itr )
```
pavel-shirshov: for (auto itr = ....)

orchagent/fdborch.cpp
keboliu: revised.

orchagent/fdborch.cpp
keboliu: revised.

orchagent/fdborch.cpp
keboliu: revised.

orchagent/fdborch.cpp
keboliu: revised.

orchagent/fdborch.cpp
keboliu: revised.

orchagent/fdborch.cpp
keboliu: revised.

orchagent/fdborch.cpp
keboliu: revised.

orchagent/fdborch.cpp
```diff
@@ -34,21 +34,66 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
...
+        
+        for (auto observer: m_observers)
+        {
+            observer->update(SUBJECT_TYPE_FDB_CHANGE, &update);
+        }
+      
+        break;
+        
+    case SAI_FDB_EVENT_FLUSHED:
+        if( !bridge_port_id && !entry->vlan_id)
```
lguohan: please follow existing code convention for the whitespaces.

orchagent/fdborch.cpp
```diff
@@ -34,21 +34,66 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
 
         (void)m_entries.insert(update.entry);
         SWSS_LOG_DEBUG("FdbOrch notification: mac %s was inserted into vlan %d", update.entry.mac.to_string().c_str(), entry->vlan_id);
+        
+        for (auto observer: m_observers)
+        {
+            observer->update(SUBJECT_TYPE_FDB_CHANGE, &update);
+        }
```
lguohan: this is about the fdb flush, why touch both LEARNED, AGED, MOVE events?

orchagent/fdborch.cpp
lguohan: change to bridge_port_id == SAI_NULL_OBJECT_ID

orchagent/fdborch.cpp
```diff
@@ -34,21 +34,66 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
...
+      
+        break;
+        
+    case SAI_FDB_EVENT_FLUSHED:
+        if( !bridge_port_id && !entry->vlan_id)
+        {
+            for(auto itr = m_entries.begin(); itr != m_entries.end(); ++itr )
```
lguohan: keep white space convention as existing code, thanks.

orchagent/fdborch.cpp
```diff
@@ -34,21 +34,66 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
...
+        {
+            /*this is a placeholder for flush port fdb case, not supported yet.*/
+            SWSS_LOG_ERROR("FdbOrch notification: not supported flush port fdb action, port_id = %lu, vlan_id = %d.", bridge_port_id, entry->vlan_id);
+        }
+        else if(!bridge_port_id && entry->vlan_id)
+        {
+            /*this is a placeholder for flush vlan fdb case, not supported yet.*/
+            SWSS_LOG_ERROR("FdbOrch notification: not supported flush vlan fdb action, port_id = %lu, vlan_id = %d.", bridge_port_id, entry->vlan_id);
+        }
+        else
```
lguohan: @prsunny , in case the fdb flush will send notification of deleted mac one-by-one, does it fall into this category?

orchagent/fdborch.cpp
```diff
@@ -34,21 +34,66 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
...
+        if( !bridge_port_id && !entry->vlan_id)
+        {
+            for(auto itr = m_entries.begin(); itr != m_entries.end(); ++itr )
+            {
+                /*This is a flush all case, need to clear up all the fdb entries*/
+                update.entry.mac = itr->mac;
+                update.entry.vlan = itr->vlan;
+                update.add = false;
+                
+                m_entries.erase(itr);
```
prsunny: Looks to me as a non-standard way of deleting the elements of a set in for-loop as it invalidates the iterator after erase. It might have worked in test. Can you check this?

orchagent/fdborch.cpp
prsunny: Observed in testing that per-entry delete is coming with event type `SAI_FDB_EVENT_AGED` instead of event FLUSHED, which is handled in the above case.  Also, I think, deleting mac one-by-one doesn't fall into this category as there could be multiple mac entries learnt on the same port-vlan combination

orchagent/fdborch.cpp
keboliu: Will check. and I am revising the solution as Guohan suggested, will update all of these with the new solution.

orchagent/fdborch.cpp
```diff
@@ -179,6 +241,59 @@ void FdbOrch::doTask(Consumer& consumer)
...
+        /*
+         * so far only support flush all the FDB entris
+         * flush per port and flush per vlan will be added later.
+         */
+        sai_status_t status;
+        sai_attribute_t attr;
+        attr.id = SAI_FDB_FLUSH_ATTR_ENTRY_TYPE;
+        attr.value.s32 = SAI_FDB_FLUSH_ENTRY_TYPE_DYNAMIC;
+
+        status = sai_fdb_api->flush_fdb_entries(gSwitchId, 1, &attr);
```
kcudnik: when flushing all, there should be 0 attributes passed, all static and dynamic entries should be cleared

keboliu: revised.

orchagent/fdborch.cpp
keboliu: separate the handling FLUSH from AGED and MOVE, at the same time remove unnecessary type cast.  No impaction to the handling of other events.

PR #431, opened 2018-01-18
[pfcwd]: Clean up auxilary fields, port ID

marian-pritsak:

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -7,6 +7,7 @@
 #include "redisapi.h"
 #include "select.h"
 #include "notifier.h"
+#include "redisclient.h"
 
 #define PFC_WD_ACTION                   "action"
 #define PFC_WD_DETECTION_TIME           "detection_time"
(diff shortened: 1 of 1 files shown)


marian-pritsak:
Depends on https://github.com/Azure/sonic-sairedis/pull/285

orchagent/pfcwdorch.cpp
```diff
@@ -447,6 +451,14 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::unregisterFromWdDb(const Port& po
         // Unregister in syncd
         m_pfcWdTable->del(key);
         m_entryMap.erase(queueId);
+
+        // Clean up
+        RedisClient redisClient(PfcWdOrch<DropHandler, ForwardHandler>::getCountersDb().get());
+        string countersKey = COUNTERS_TABLE ":" + sai_serialize_object_id(queueId);
+        redisClient.hdel(countersKey, "PFC_WD_DETECTION_TIME");
```
sihuihan88: Can we use macro PFC_WD_RESTORATION_TIME etc here? Also what about using for loop to remove all needed fields?

lguohan: @sihuihan88 , maybe I am missing something, but the macro PFC_WD_DETECTION_TIME is different from "PFC_WD_DETECTION_TIME". From the code, I cannot find PFC_WD_DETECTION_TIME used anywhere.

PR #432, opened 2018-01-19
Add speed and buffer set test

andriymoroz-mlnx:

**What I did**
Added test for port speed change and PG buffer profile update

**Why I did it**

**How I verified it**
Ran the test and checked it passes

**Details if related**
should pass after
https://github.com/Azure/sonic-swss/pull/417 and
https://github.com/Azure/sonic-buildimage/pull/1250
are merged

### DIFF
--- a/tests/test_speed.py
+++ b/tests/test_speed.py
@@ -1,0 +1,66 @@
+from swsscommon import swsscommon
+import time
+import re
+import json
+import os
+
+class TestSpeedSet(object):
+    num_ports = 32
+    def test_SpeedAndBufferSet(self, dvs):
+        speed_list = ['50000', '25000', '40000', '10000', '100000']
+
+        cdb = swsscommon.DBConnector(4, dvs.redis_sock, 0)
+        adb = swsscommon.DBConnector(1, dvs.redis_sock, 0)
+        cfg_port_table = swsscommon.Table(cdb, "PORT", '|')
+        cfg_buffer_profile_table = swsscommon.Table(cdb, "BUFFER_PROFILE", '|')
+        cfg_buffer_pg_table = swsscommon.Table(cdb, "BUFFER_PG", '|')
+        asic_port_table = swsscommon.Table(adb, "ASIC_STATE:SAI_OBJECT_TYPE_PORT")
+        asic_profile_table = swsscommon.Table(adb, "ASIC_STATE:SAI_OBJECT_TYPE_BUFFER_PROFILE")
+
+        buffer_profiles = cfg_buffer_profile_table.getKeys()
+        expected_buffer_profiles_num = len(buffer_profiles)
+        assert expected_buffer_profiles_num == 7
+        assert len(asic_profile_table.getKeys()) == expected_buffer_profiles_num
+
+        for speed in speed_list:
+            fvs = swsscommon.FieldValuePairs([("speed", speed)])
+            # set same speed on all ports
+            for i in range(0, self.num_ports):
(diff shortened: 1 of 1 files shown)


yxieca:
retest please

tests/test_speed.py
```diff
@@ -0,0 +1,66 @@
...
+        adb = swsscommon.DBConnector(1, dvs.redis_sock, 0)
+        cfg_port_table = swsscommon.Table(cdb, "PORT", '|')
+        cfg_buffer_profile_table = swsscommon.Table(cdb, "BUFFER_PROFILE", '|')
+        cfg_buffer_pg_table = swsscommon.Table(cdb, "BUFFER_PG", '|')
+        asic_port_table = swsscommon.Table(adb, "ASIC_STATE:SAI_OBJECT_TYPE_PORT")
+        asic_profile_table = swsscommon.Table(adb, "ASIC_STATE:SAI_OBJECT_TYPE_BUFFER_PROFILE")
+
+        buffer_profiles = cfg_buffer_profile_table.getKeys()
+        expected_buffer_profiles_num = len(buffer_profiles)
+        assert expected_buffer_profiles_num == 7
```
yxieca: Using number 7 here seems to be a recipe for long investigation when test fails. How can we make it clearer where this number came from and when test failed here, what to look for?

A comment will be a good start.

tests/test_speed.py
```diff
@@ -0,0 +1,66 @@
...
+        assert len(asic_profile_table.getKeys()) == expected_buffer_profiles_num
+
+        for speed in speed_list:
+            fvs = swsscommon.FieldValuePairs([("speed", speed)])
+            # set same speed on all ports
+            for i in range(0, self.num_ports):
+                cfg_port_table.set("Ethernet%d" % (i*4), fvs)
```
yxieca: I suspect this change won't work for all devices.

It all depends on what is self.num_ports?

For instance, Arista7260CX3 has 64 physical ports, if self.num_ports is 64 for this device, then it is fine. However, when Arista7260CX3 is used for T0, it will have 56 out of 64 ports break into 2 50G ports, bring the total number of ports to 112 + 8 = 120. Most of these ports are Etherenet(x * 4) and Ethernet(x * 4 + 2) except the 8 100G ports. So if self.num_ports is 120 for this device, then the code here is incorrect.

lguohan:
the test does not pass, please check

tests/test_speed.py
andriymoroz-mlnx: it comes from the buffers profile used for the test
I think it can be calculates (something like `cat ... | grep "PROFILE" | wc -l`)
I'll check and update this comment

andriymoroz-mlnx:
It will pass after VS docker is updated https://github.com/Azure/sonic-buildimage/commit/ff8332feea5a8372833d7cdd2beb0893706acd55 (it is part of buffers config PR https://github.com/Azure/sonic-buildimage/pull/1250)
vicious circle...

tests/test_speed.py
andriymoroz-mlnx: It is designed to run on `PLATFORM=vs`

yxieca: I see. Can you at least add a comment at beginning of this file?

Thanks,
Ying

yxieca:
retest please

yxieca:
retest this please

yxieca:
retest this please

yxieca:
retest this please

yxieca:
retest this please

yxieca:
retest this please

PR #433, opened 2018-01-19
[pfcwd]: Assign different names to executors.

marian-pritsak:
Having executors with same names results in only first being registered.


orchagent/pfcwdorch.cpp
```diff
@@ -35,7 +35,7 @@ PfcWdOrch<DropHandler, ForwardHandler>::PfcWdOrch(DBConnector *db, vector<string
     auto interv = timespec { .tv_sec = PFC_WD_LOSSY_POLL_TIMEOUT_SEC, .tv_nsec = 0 };
     auto timer = new SelectableTimer(interv);
     auto executor = new ExecutableTimer(timer, this);
-    Orch::addExecutor("", executor);
+    Orch::addExecutor("COUNTERS_POLL", executor);
```
sihuihan88: change to more specific name?

lguohan: any suggestion?

sihuihan88: PFC_WD_LOSSY?

marian-pritsak: This executor will be extended with more counters (e.g. for multicast queues). I'd want to keep it more general.

PR #435, opened 2018-01-25
[PfcWd]: Make polling interval configurable

sihuihan88:

**What I did**
Make the polling interval configurable
**Why I did it**
Extend the feature.
**How I verified it**
Test on DUT
**Details if related**

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -9,6 +9,8 @@
 #include "notifier.h"
 #include "redisclient.h"
 
+#define PFC_WD_FLEX_COUNTER_GROUP       "PFC_WD"
+#define PFC_WD_GLOBAL                   "GLOBAL"
 #define PFC_WD_ACTION                   "action"
 #define PFC_WD_DETECTION_TIME           "detection_time"
 #define PFC_WD_RESTORATION_TIME         "restoration_time"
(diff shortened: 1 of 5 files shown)


sihuihan88:
depends on https://github.com/Azure/sonic-swss-common/pull/182 and https://github.com/Azure/sonic-sairedis/pull/289

lguohan:
retest this please

orchagent/pfcwdorch.h
```diff
@@ -108,7 +109,7 @@ class PfcWdSwOrch: public PfcWdOrch<DropHandler, ForwardHandler>
     const vector<sai_queue_attr_t> c_queueAttrIds;
 
     shared_ptr<DBConnector> m_pfcWdDb = nullptr;
-    shared_ptr<ProducerStateTable> m_pfcWdTable = nullptr;
+    shared_ptr<ProducerTable> m_pfcWdTable = nullptr;
```
marian-pritsak: Why switch to ProducerTable?

orchagent/pfcwdorch.cpp
```diff
@@ -356,6 +356,32 @@ void PfcWdOrch<DropHandler, ForwardHandler>::deleteEntry(const string& name)
...
+        const vector<FieldValueTuple>& data)
+{
+    if (key == GENERAL)
+    {
+        for (auto valuePair: data)
+        {
+            const auto &field = fvField(valuePair);
+            const auto &value = fvValue(valuePair);
+
+            if (field == POLL_INTERVAL_FIELD)
```
marian-pritsak: Is there a benefit of making this field global and not per port? If yes, what will happen in case if I will enable WD on some port(s) without setting global polling interval?

orchagent/pfcwdorch.h
lguohan: same questions here.

sihuihan88: Using statetable, when we want to change a single field, the consumer will receive the state of all the fields and set them one more time. Eg. We set F1 v1. first, and then when set F2 v2 later, consumer can still receive F1 V1 when handling F2 V2 set. This is not ideal as we have the use case to only change the value of a single field. Using statetable will trigger the current syncd error that it only allow to add the plugin with same sha once. 
 
Since we do not frequently change the field value for a single key for flexcounter, there is no benefit to use statetable. So switch to producertable/cosumertable

orchagent/pfcwdorch.cpp
sihuihan88: Polling interval is global by design. It will use the default polling interval if not set specifically.

sihuihan88:
retest this please

PR #436, opened 2018-01-27
[vlanmgr]: Run a sequence of shell commands using bash -c

pavel-shirshov:
Use bash -c to minimize numbers of fork-execs in the Vlan Manager

**What I did**
I rewrote vlanmgr to minimize number of fork-execs in the code. Now vlanmgr should work faster and generate less tasks.

**Why I did it**
Because I found that vlanmgr used a lot of resources when SONiC started. It slowed down SONiC start time.

**How I verified it**
Manually checked output of ip and bridge commands.

**Details if related**
It's better to avoid using ip and bridge command completely. But I found that currently libnl3 doesn't support vlan_filtering feature and also knows nothing about bridge vlans. Also I found that ioctls don't support everything I need. Later it's possible to rewrite the code using libmnl library which is recommended to be used by creators of iproute2.

### DIFF
--- a/cfgmgr/vlanmgr.cpp
+++ b/cfgmgr/vlanmgr.cpp
@@ -62,8 +62,6 @@ VlanMgr::VlanMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
         const std::string echo_cmd_backup = std::string("")
           + IP_CMD + " link set " + DOT1Q_BRIDGE_NAME + " type bridge vlan_filtering 1";
 
-        SWSS_LOG_DEBUG("shell backup cmd: %s", echo_cmd_backup.c_str());
-
         EXEC_WITH_ERROR_THROW(echo_cmd_backup, res);
     }
 }
@@ -84,8 +82,6 @@ bool VlanMgr::addHostVlan(int vlan_id)
                + " address " + gMacAddress.to_string()
                + " type vlan id " + std::to_string(vlan_id) + "\"";
 
-    SWSS_LOG_DEBUG("shell cmds: %s", cmds.c_str());
-
     std::string res;
     EXEC_WITH_ERROR_THROW(cmds, res);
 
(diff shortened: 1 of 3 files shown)


cfgmgr/vlanmgr.cpp
```diff
@@ -32,144 +32,162 @@ VlanMgr::VlanMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
...
-    EXEC_WITH_ERROR_THROW(cmd.str(), res);
+     * Once we move sonic to debian 9, we can use IP command by default
+     * ip command to create bridge with filtering:
+     * /sbin/ip link add Bridge up type bridge vlan_filtering 1
+     */
+    const std::string cmds = std::string("")
+      + BASH_CMD + " -c \""
+      + IP_CMD + " link del " + DOT1Q_BRIDGE_NAME + " 2>/dev/null; "
+      + IP_CMD + " link add " + DOT1Q_BRIDGE_NAME + " up type bridge && "
+      + "echo 1 > /sys/class/net/" + DOT1Q_BRIDGE_NAME + "/bridge/vlan_filtering && "
```
lguohan: as explained in the comment, echo will fail in virtual switch since /sys directory is read-only. therefore, you need keep the ip link as the fallback, because in virtual switch you are using the kernel of host, that can support vlan_filtering command.

pavel-shirshov: Debian 8 doesn't have support of ip link set Bridge vlan_filtering command. The ip route 2 sources doesn't have anything related to vlan_filtering.
Debian 9 contains updated iproute2 which allows set vlan_filtering from command line without echoing to /sys.
So I left a comment
Once we move sonic to debian 9, we can use IP command by default
ip command to create bridge with filtering:                     
``` /sbin/ip link add Bridge up type bridge vlan_filtering 1```

lguohan: as I said, "echo 1 > ...bridge/vlan_filtering" does not work for virtual switch docker as the /sys is read only. The only way to make it work is to use ip link command. In virtual switch docker, I installed the latest iproute2 command so it works. remove the fallback will break virtual switch.

lguohan: I specifically added this fallback method for virtual switch. 

https://github.com/Azure/sonic-swss/commit/57a98acb893c73063792bf6d7adf8bca71f8b9d6

ishidawataru: I'm not familiar with virtual switch but I'm workarounding by executing `$mount -o remount,rw /sys` to make sysfs writable in privileged docker container.

Maybe you could do the same to remove this fallback method?

lguohan: the fallback logic will work for debian 9 without code change. it is better to keep it. i tried many methods in virtual switch env, even the one you suggested, none of them works.

ishidawataru: oh, I got it. sorry for the noise.

pavel-shirshov:
Guohan
Can you please check it?

PR #437, opened 2018-01-27
Fixes for various OA crashes derived from intf-overlapping and activation of link-local forwarding

rodnymolina:
Fixes for various OA crashes derived from intf-overlapping and activation of link-local forwarding.

From a functional standpoint, these are the changes made in this PR:

 * Add sanity-checking logic to OA to prevent overlapping interface prefixes from being pushed to hardware.
 * Differentiated processing/injection of "subnet" / "ip2me" interface-routes to cope with scenarios requiring independent handling (e.g. link-local)
 * Contributed to reduce kernel<->swss inconsistency by allowing OA to cope with overlapped subnets, and enabling these ones to be resurrected should the primary goes down (as kernel does).
 * Enabled local-scope-based forwarding (link-local), which requires the overlapping prevention logic included in this PR.

From a coding perspective, the following changes were made as part of this patch:

 * Creation of a hashmap to store all interface-routes in the system (both "subnet" and "ip2me" entries are tracked).
 * Move subnet/ip2me's route-creation logic away from m_syncdIntfses struct, as this one doesn't track ip2me state, nor serves our purpose of enabling interface-route tracking for overlap prevention.
 * Adjusting neighOrch::doTask() by moving invokation of sai-route-creation routines to a more convenient/intuitive location.
 * Allow intfsyncd to process/pass interface addresses with local-scope.
 * Do the same on neighsyncd to allow link-local neighbors to be passed to OA.

 Note-1: This PR supersedes previous PR#400.
 Note-2: This PR relies on PR#184 (sonic-swss-common).

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


lguohan:
retest this please

rodnymolina:
Code re-tested with latest images.

lguohan:
retest this please

rodnymolina:
Code is failing to build due to a dependency with PR/187: https://github.com/Azure/sonic-swss-common/pull/187

lguohan:
retest this please

rodnymolina:
All checks are passing now since dependencies are met.

lguohan:
As we discussed, to address ipv6 link-local issue, we need ask vendor to implement NO_HOST_ENTRY option. I think that is a prerequsite to merge this PR, right?

rodnymolina:
The problematic scenario that we discussed is potentially presented when we are dealing with link-local + VLAN subinterfaces. In this code we are preventing that problem by skipping the creation of 'conflicting' neighbor-entries in VLAN scenarios. Traffic directed to this neighbor-entry will still get forwarded when initiated by the local router.

Problem shouldn't affect non-VLAN/regular setups given that link-local addresses are derived from each switch's mac-address, so there's no room for overlapping ipv6 neighbor-entries (unless the user deliberately configure a bogus/duplicated entry).

We will remove this link-local-over-VLAN limitation from our code once that we get vendor support for "NO_HOST_ENTRY" feature.

zhenggen-xu:
Let's split this PR to two parts,  one is to handle overlapping case for same or different interfaces.  The other is to let the link local messages going to swss and HW. 
We should get this fix in also: https://github.com/Azure/sonic-swss/pull/370  so we can completely remote the logic in intfsorch.cpp for dealing with the ifconfig weirdness.

rodnymolina:
Yes, i'm planning to send a new patch on this one to deactivate link-local processing for the time being, which will be re-enabled later on once that we have the missing sai support. This way we could have one uniform logic to deal with all interface overlaps in the system.

Now, #370 by itself doesn't seem to be fixing all the issues the current "ifconfig" logic is dealing with. Need to do a bit more research in this area.

orchagent/intfsorch.cpp
```diff
@@ -402,3 +431,273 @@ void IntfsOrch::removeIp2MeRoute(const IpPrefix &ip_prefix)
...
+        if (curIfRoute.prefix == ifRoute.prefix)
+        {
+            SWSS_LOG_ERROR("New %s route %s for interface %s overlaps with "
+                           "existing route %s for interface %s. "
+                           "Skipping...",
+                           ifRoute.type.c_str(),
+                           ifRouteStr.c_str(),
+                           ifRoute.ifName.c_str(),
+                           curIfRoute.prefix.to_string().c_str(),
+                           curIfRoute.ifName.c_str());
```
jipanyang: same link local address may exist on multiple interfaces,  should the log level be changed to NOTICE instead of error?


366: Vlan10@Bridge: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state LOWERLAYERDOWN group default qlen 1000
    link/ether 02:42:ac:11:00:02 brd ff:ff:ff:ff:ff:ff
    inet 10.10.10.100/24 scope global Vlan10
       valid_lft forever preferred_lft forever
    inet6 **fe80::42:acff:fe11:2/64** scope link 
       valid_lft forever preferred_lft forever
110: eth6@if111: <BROADCAST,MULTICAST,PROMISC,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 9e:10:98:c4:5a:10 brd ff:ff:ff:ff:ff:ff link-netnsid 6
367: Vlan20@Bridge: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state LOWERLAYERDOWN group default qlen 1000
    link/ether 02:42:ac:11:00:02 brd ff:ff:ff:ff:ff:ff
    inet 20.20.20.100/24 scope global Vlan20
       valid_lft forever preferred_lft forever
    inet6 **fe80::42:acff:fe11:2/64** scope link 
       valid_lft forever preferred_lft forever

jipanyang:
This pr is needed to fix the link local overlap error (?): 
 
```
Jan 18 01:35:12.569456 ERR #orchagent: :- : object key SAI_OBJECT_TYPE_ROUTE_ENTRY:{"dest":"fe80::42:acff:fe11:2/128","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000022"} already exists
Jan 18 01:35:12.569470  ERR #orchagent: :- Failed to create IP2me route ip:fe80::42:acff:fe11:2, rv:-6
Jan 18 01:35:12.569562  ERR #orchagent: :- Failed due to exception: Failed to create IP2me route.
```

rodnymolina:
@jipanyang Finishing dealing with code-conflicts. Will re-test when done and generate a patch-update.

nikos-github:
@rodnymolina @jipanyang Is this PR getting updated?

rodnymolina:
@nikos-github yes, code is mostly ready (there were many conflicts with the original patch), i'm dealing with a few failing UTs. PR will be updated within the next few days.

rodnymolina:
@lguohan @jipanyang can you guys please look into this one asap? Thanks.

tests/test_interface.py
```diff
@@ -461,3 +461,762 @@ def test_InterfacesCreateRemove(self, dvs, testlog):
...
+    def test_IPv6LinkLocalInterfacePartialOverlapReversed(self, dvs, testlog):
+
+        self.setup_db(dvs)
+
+        # Add a regular ipv4 address to interface #1
+        self.add_ip_address("Ethernet8", "fe80:1:1::1/64", "IPv6", "local")
+
+        self.verify_appdb_ip_intf_addr_presence("Ethernet8", "fe80:1:1::1/64", "IPv6", "local")
+
+        self.verify_asicdb_ip_intf_addr_presence("fe80:1:1::/64", "fe80:1:1::1/128", 1)
```
nikos-github: The prefixes you are using for IPv6 link-local are illegal prefixes. Can you please correct?

tests/test_interface.py
```diff
@@ -461,3 +461,762 @@ def test_InterfacesCreateRemove(self, dvs, testlog):
...
+    #   interface with the primary IPsubnet route ("fe80:1:1::/64"), our implementation
+    #   must 'resurrect' the equivalent subnet-route associated to the secondary
+    #   intf.
+    #
+    def test_IPv6LinkLocalInterfacePartialOverlapReversed(self, dvs, testlog):
+
+        self.setup_db(dvs)
+
+        # Add a regular ipv4 address to interface #1
+        self.add_ip_address("Ethernet8", "fe80:1:1::1/64", "IPv6", "local")
```
nikos-github: The prefixes you are using for IPv6 link-local are illegal prefixes. Can you please correct?

rodnymolina: @nikos-github why are these prefixes illegal? They fall within fe80/10 subnet. Please let me know.

orchagent/intfsorch.cpp
```diff
@@ -193,19 +202,348 @@ bool IntfsOrch::setIntf(const string& alias, sai_object_id_t vrf_id, const IpPre
...
+        }
+    }
+}
+
+/*
+ * Method's goal is to track/record any potential overlap between the interfaces
+ * configured in the system, and alert caller of such an incident.
+ */
+bool IntfsOrch::trackIntfRouteOverlap(const IntfRouteEntry &ifRoute)
+{
```
jipanyang: If interfaces are in different VRF,  overlapping addresses should be allowed?

rodnymolina: Good point. Do you know if we are already supporting that as part of our new VRF implementation?

cfgmgr/intfmgr.cpp
```diff
@@ -204,8 +204,18 @@ bool IntfMgr::doIntfAddrTask(const vector<string>& keys,
...
+
+        if (ip_prefix.getIp().getAddrScope() == IpAddress::AddrScope::LINK_SCOPE)
+        {
+            FieldValueTuple s("scope", "local");
+            fvVector.push_back(s);
+        }
+        else
+        {
+            FieldValueTuple s("scope", "global");
+            fvVector.push_back(s);
```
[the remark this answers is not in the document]
rodnymolina: @jipanyang thanks for the review!

There's a need to differentiate between GLOBAL and LINK/LOCAL scope, but why would you want to differentiate between GLOBAL and HOST scope here? With the existing code, there's no backend process that would benefit from this...

orchagent/neighorch.cpp
```diff
@@ -321,6 +321,23 @@ void NeighOrch::doTask(Consumer &consumer)
...
         IpAddress ip_address(key.substr(found+1));
 
+        /*
+         * When dealing with link-scope addresses we will not create nor remove
+         * neighbor-entries. This is motivated by a SAI implementation constrain,
+         * that prevents the creation of more than one host-route to the same
+         * ip-address through different intfs. This if-statement should be
+         * eliminated the moment this issue is fixed in SAI implementations.
+         */
+        if (ip_address.getAddrScope() == IpAddress::AddrScope::LINK_SCOPE)
```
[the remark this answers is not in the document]
rodnymolina: @jipanyang IPv6 link-local neighbors are useful when they are utilized to resolve global-scope routes over link-local NHs, e.g. bgp-path (fc00::1/64) pointing to link-local NH (fe80::1/128). But you may just want to turn on link-local to allow your switches to generate/respond to link-local traffic for network-autoconfiguration/discovery or troubleshooting purposes, i would say that this is the most typical case.

orchagent/intfsorch.cpp
jipanyang: Yes,  vrforch has the VRF implementation with multiple virtual routers.

zhenggen-xu:
retest this please

orchagent/intfsorch.h
```diff
@@ -12,20 +12,66 @@
...
+ * Indexed by the string associated to each interface-route (either ip2me or
+ * subnet or bcast). Values are formed by a list of elements that keep track
+ * of each route IpPrefix, as well as the interface on which it was configured.
+ *
+ * Example:
+ *
+ *    Key                                        Value
+ * ----------             -------------------------------------------------------
+ * 10.1.1.0/24 (subnet)   10.1.1.0/24 eth1, 10.1.1.0/24 eth2, 10.1.1.0/24 eth3
+ * 10.1.1.1/32 (ip2me)    10.1.1.1/32 eth1, 10.1.1.10/32 eth2, 10.1.1.255/32 eth3
```
lguohan: >10.1.1.10/32 eth2, 10.1.1.255/32 eth3 [](start = 44, length = 37)

10.1.1.10 or 10.1.1.1?

rodnymolina: @lguohan you right, there's a typo here, it should be '10.1.1.1/32' in that second line, will adjust this.

MalteJ:
what is the status here?
would be great to have the feature!

PR #440, opened 2018-01-29
Don't generate an error message when alias == 'ConfigPortDone'

pavel-shirshov:
**What I did**
Add a code to avoid not informational error message
"ERR orchagent: :- doPortTask: Failed to get port id by alias:PortConfigDone"

**Why I did it**
Because log supposed to be clean and easy to read

**How I verified it**
Run on a DUT

**Details if related**


pavel-shirshov:
close for now. I'll reopen it later

PR #452, opened 2018-03-14
Support for platforms based on Barefoot Networks' device

rsunkad:
**- Incorporate support for platforms based on Barefoot Networks' device **

**- Added port config files and related scripts for the platforms **

**- Verified using SONiC tests  **


orchagent/portsorch.cpp
```diff
@@ -728,6 +726,33 @@ bool PortsOrch::getQueueType(sai_object_id_t queue_id, string &type)
...
+    switch(an) {
+      case 0: // leave it as is?
+        return true;
+      case 1:
+        attr.value.booldata = true;
+        break;
+      case 2:
+        attr.value.booldata = false;
+        break;
+    }
```
lguohan: why not just bool? 0, 1 here?

krambn: The original idea was to support default/on/off. Default would imply SDK defaults. But SAI does not support that. The use case is to default to different settings based on the type of cable/transcievers/etc. I can revert this back to boolean, if you feel it won't be supported in the near future.

configure.ac
```diff
@@ -39,6 +39,11 @@ esac],[gtest=false])
 AM_CONDITIONAL(GTEST, test x$gtest = xtrue)
 
 CFLAGS_COMMON="-std=c++11 -Wall -fPIC -Wno-write-strings -I/usr/include/libnl3 -I/usr/include/swss"
+
+AM_CONDITIONAL(sonic_asic_platform_barefoot,   test x$CONFIGURED_PLATFORM = xbarefoot)
+AM_COND_IF([sonic_asic_platform_barefoot],
+           [CFLAGS_COMMON+=" -I/opt/bfn/install/include/switchsai"])
```
lguohan: @rsunkad , why is this needed?

portsyncd/portsyncd.cpp
```diff
@@ -199,7 +199,7 @@ void handlePortConfigFile(ProducerStateTable &p, string file)
         throw "Port configuration file not found!";
     }
 
-    list<string> header = {"name", "lanes", "alias", "speed"};
+    list<string> header = {"name", "lanes", "alias", "speed", "autoneg", "fec"};
```
lguohan: in the master branch, fec is coming from config. If we are going to support autoneg in master, it should also come from config db.

orchagent/portsorch.cpp
lguohan: according to SAI, it seems is bool value, and the default value is false. If you feel that the default value is incorrect, then we should change in the sai spec. 

I think it will be clear here to follow SAI standard.

orchagent/portsorch.cpp
```diff
@@ -789,6 +814,18 @@ bool PortsOrch::addPort(const set<int> &lane_set, uint32_t speed)
     attr.value.u32list.count = static_cast<uint32_t>(lanes.size());
     attrs.push_back(attr);
 
+    if(an == true) {
+        attr.id = SAI_PORT_ATTR_AUTO_NEG_MODE;
+        attr.value.booldata = true;
+        attrs.push_back(attr);
+    }
+
+    if(fec_mode != "") {
```
lguohan: can you make brace in a separate line and leave a space between if and parenthesis.

portsyncd/portsyncd.cpp
krambn: Isn't there a problem with config DB for initial port states - ports will come up as default and then potentially "flap" when this configuration from DB is applied? I thought the portconfig.ini is first played into the SAI library and then modified when the config DB is evaluated and it notices a change. I may be wrong on this.

configure.ac
krambn: Below is the compilation without the change: sai,h is always included from the install dir in our current debs.

libtool: link: g++ -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wl,-z -Wl,relro -o portsyncd portsyncd-portsyncd.o portsyncd-linksync.o  -lnl-3 -lnl-route-3 -lswsscommon -lnl-genl-3 -lhiredis
make[4]: Leaving directory '/sonic/src/sonic-swss/portsyncd'
Making all in orchagent
make[4]: Entering directory '/sonic/src/sonic-swss/orchagent'
g++ -DHAVE_CONFIG_H -I. -I.. -I .. -g -DNDEBUG  -std=c++11 -Wall -fPIC -Wno-write-strings -I/usr/include/libnl3 -I/usr/include/swss -Werror -Wno-reorder -Wcast-align -Wcast-qual -Wconversion -Wdisabled-optimization -Wextra -Wfloat-equal -Wformat=2 -Wformat-nonliteral -Wformat-security -Wformat-y2k -Wimport -Winit-self -Winvalid-pch -Wlong-long -Wmissing-field-initializers -Wmissing-format-attribute -Wno-aggregate-return -Wno-padded -Wno-switch-enum -Wno-unused-parameter -Wpacked -Wpointer-arith -Wredundant-decls -Wstack-protector -Wstrict-aliasing=3 -Wswitch -Wswitch-default -Wunreachable-code -Wunused -Wvariadic-macros -Wno-switch-default -Wno-long-long -Wno-redundant-decls -I /usr/include/sai -D_FORTIFY_SOURCE=2  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -c -o orchagent-main.o `test -f 'main.cpp' || echo './'`main.cpp
main.cpp:2:17: fatal error: sai.h: No such file or directory
 #include "sai.h"
                 ^
compilation terminated.
Makefile:550: recipe for target 'orchagent-main.o' failed

lguohan: I see, I think it is because on other platforms, they have a libsaixxx-dev packages which installs SAI headers under /usr/include/sai. They install the packets before compile swss, so they do not need this.

portsyncd/portsyncd.cpp
lguohan: ultimately, all the configurations should come from one place. we plan to consolidate all configurations into config db. 

I think since we are using create port API with the attributes coming from config db, there should be no flap.

krambn: So, we can remove the portconfig.ini file and get the breakout, speeds and hostinterface name, index associations from the configuration DB. The problem we had was that at boot up - we come up with default configurations - if the default matches the portconfig.ini no changes are observed. If changes come later it causes the remote nodes to see another flap. If we can eliminate this it will be great. Currently, you can see that some of the device configurations use breakout, fec and AN in the default config files. If you want us to remove fec and autoneg in  the 201712 branch, we can do it. But the devices will need to apply a patch to build for their platforms. Some of them have fixed breakout configuration and default to fec and AN settings (sonic-buildimage pull request). Let me know if you want me to revert the changes and I will do so.

lguohan: sorry for the confusion, I am ok with this on 201712 branch. For master branch, we need to take a different approach.

PR #456, opened 2018-03-23
[countercheck]: move pfc counter alerting to countercheckorch

sihuihan88:

**What I did**
Simple refactor: move lossy class PFC pause alerting to countercheckOrch
**Why I did it**
The logic by design align with countercheckOrch. Refactor to make it clearer
**How I verified it**
Tested on DUT
**Details if related**

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -22,6 +22,7 @@
 #define PFC_WD_POLL_TIMEOUT             5000
 #define PFC_WD_LOSSY_POLL_TIMEOUT_SEC   (5 * 60)
 #define SAI_PORT_STAT_PFC_PREFIX        "SAI_PORT_STAT_PFC_"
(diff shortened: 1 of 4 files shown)


orchagent/pfcwdorch.cpp
```diff
@@ -400,24 +304,33 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port,
         return;
     }
 
+    set<uint8_t> losslessTc;
```
marian-pritsak: Isn't it a code from your previous PR?

sihuihan88: Thanks for the review.

orchagent/pfcwdorch.h
```diff
@@ -9,8 +9,6 @@
 #include "timer.h"
 #include <array>
```
lguohan: can this be removed?

PR #458, opened 2018-03-27
[pfcwd]: only init counter when creating action handler

sihuihan88:

**What I did**
1. Only init counter when handler is created
2. Update lua script to support infinite restoration time
3. Clean up no need code
**Why I did it**

**How I verified it**
Tested on DUT
**Details if related**

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -20,7 +20,6 @@
 #define PFC_WD_RESTORATION_TIME_MAX     (60 * 1000)
 #define PFC_WD_RESTORATION_TIME_MIN     100
 #define PFC_WD_POLL_TIMEOUT             5000
-#define PFC_WD_LOSSY_POLL_TIMEOUT_SEC   (5 * 60)
 #define SAI_PORT_STAT_PFC_PREFIX        "SAI_PORT_STAT_PFC_"
 #define PFC_WD_TC_MAX 8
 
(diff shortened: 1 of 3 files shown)


orchagent/pfcwdorch.cpp
```diff
@@ -596,13 +592,13 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::doTask(swss::NotificationConsumer
...
                         entry->second.index,
                         PfcWdOrch<DropHandler, ForwardHandler>::getCountersTable());
+                entry->second.handler->initCounters();
             }
         }
         else
         {
             throw runtime_error("Unknown PFC WD action");
         }
-        entry->second.handler->initCounters();
```
marian-pritsak: What is an actual problem here? If the handler is not created, we throw an exception.

sihuihan88: If receiving multiple same storm notifications it init counters multiple times

marian-pritsak: Don't you think it is a workaround rather than solution?
What's the reason that we get multiple notifications for same port/queue?

sihuihan88: I would think this is an allowed behavior for our design. swss is a single thread, and queues all the notification and handle them one by one. If swss cannot handle pfcw storm/restore notification from syncd in time, it's possible that syncd send multiple notifications until swss handles it. This is possible to happen when during swss start time or swss busy updating large amount of routes. Therefore, in the pfcwdorch we should keep this in mind and be robust to handle this kind of case as well. It's not a workaround, unless changing the whole design.

marian-pritsak: ok

PR #459, opened 2018-03-27
[pfcwd]: update counters of queues associated with handler when pfcwd stops

sihuihan88:

**What I did**
Update counters of queues associated with handler when pfcwd stops
**Why I did it**
If the queue is in drop mode and pfcwd stops, queue are restored and the counters should be updated as well.
**How I verified it**
Tested on DUT.
**Details if related**

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -432,6 +432,17 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::unregisterFromWdDb(const Port& po
 
         // Unregister in syncd
         m_flexCounterTable->del(key);
+
+        auto entry = m_entryMap.find(queueId);
+        if (entry != m_entryMap.end())
+        {
+            if (entry->second.handler != nullptr)
+            {
+                entry->second.handler->commitCounters();
+                entry->second.handler = nullptr;
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


orchagent/pfcwdorch.cpp
```diff
@@ -432,6 +432,17 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::unregisterFromWdDb(const Port& po
...
         // Unregister in syncd
         m_flexCounterTable->del(key);
+
+        auto entry = m_entryMap.find(queueId);
+        if (entry != m_entryMap.end())
+        {
+            if (entry->second.handler != nullptr)
+            {
+                entry->second.handler->commitCounters();
+                entry->second.handler = nullptr;
```
marian-pritsak: I think this is not required. Entry is erased in code below.

sihuihan88: Please be aware even if entry is erased counter are not commited

sihuihan88: If there are existing action handler， commitcounter are needed when entry are deleted.

marian-pritsak: Yes, I mean that line setting it to nullptr is not required
I agree with commitCounters

PR #460, opened 2018-04-02
[portsorch]: Add asymmetric PFC support

volodymyrsamotiy:
* Implement asymmetric PFC feature
* Implement unit test on top of virtual switch

**What I did**
Added asymmetric PFC support
**Why I did it**
Feature request
**How I verified it**
Performed manual tests, also implemented and ran unit test on top of virtual switch


orchagent/portsorch.cpp
```diff
@@ -463,6 +469,121 @@ bool PortsOrch::setPortFec(sai_object_id_t id, sai_port_fec_mode_t mode)
...
+    attr.value.s32 = (int32_t) port.m_pfc_asym;
+
+    sai_status_t status = sai_port_api->set_port_attribute(port.m_port_id, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to set PFC mode %d to port id 0x%lx (rc:%d)", port.m_pfc_asym, port.m_port_id, status);
+        return false;
+    }
+
+    if (!setPortPfc(port.m_port_id, pfc))
```
sihuihan88: Can you help me understand why set port pfc=0 here?

orchagent/portsorch.cpp
```diff
@@ -463,6 +469,121 @@ bool PortsOrch::setPortFec(sai_object_id_t id, sai_port_fec_mode_t mode)
...
+    {
+        return false;
+    }
+
+    if (port.m_pfc_asym == SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_SEPARATE)
+    {
+        attr.id = SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL_RX;
```
sihuihan88: To enable asym mode, SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL_RX is set to FF, what will the value of SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL_TX be? Will it automatically inherit the same value of SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL without any action?

orchagent/portsorch.cpp
```diff
@@ -463,6 +469,121 @@ bool PortsOrch::setPortFec(sai_object_id_t id, sai_port_fec_mode_t mode)
...
+    {
+        SWSS_LOG_ERROR("Failed to get port object for port id 0x%lx", portId);
+        return false;
+    }
+
+    if (p.m_pfc_asym == SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_COMBINED)
+    {
+        attr.id = SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL;
+    }
+    else
```
sihuihan88: I would recommend to use else if  p.m_pfc_asym == SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_SEPARATE the below case, and log error if m_pfc_asym doesn't match any.

orchagent/portsorch.cpp
```diff
@@ -463,6 +469,121 @@ bool PortsOrch::setPortFec(sai_object_id_t id, sai_port_fec_mode_t mode)
...
+
+    sai_attribute_t attr;
+    uint8_t pfc = 0;
+
+    if (!getPortPfc(port.m_port_id, &pfc))
+    {
+        return false;
+    }
+
+    port.m_pfc_asym = pfc_asym_map.at(pfc_asym);
```
sihuihan88: It will break if pfc_asym is not entered as “on/off”, please consider to handle this to make it more robust.

orchagent/portsorch.cpp
```diff
@@ -1311,6 +1437,20 @@ void PortsOrch::doPortTask(Consumer &consumer)
                     }
 
                 }
+
+                if (pfc_asym != "")
+                {
+                    if (setPortPfcAsym(p, pfc_asym))
+                    {
```
sihuihan88: Consider to move "m_portList[port.m_alias] = port;" within setPortPfcAsym out of the function to here. As setPortPfcAsym should focus on set port asym pfc mode.

orchagent/portsorch.h
```diff
@@ -57,6 +57,9 @@ class PortsOrch : public Orch, public Subject
     bool setHostIntfsOperStatus(sai_object_id_t id, bool up);
     void updateDbPortOperStatus(sai_object_id_t id, sai_port_oper_status_t status);
     bool bindAclTable(sai_object_id_t id, sai_object_id_t table_oid, sai_object_id_t &group_member_oid, acl_stage_type_t acl_stage = ACL_STAGE_INGRESS);
+
+    bool getPortPfc(sai_object_id_t portId, uint8_t *pfc_bitmask);
```
sihuihan88: In getPortPfc, it queries sai attribute each time which is very insufficient. Can we cache the value and let getPortPfc only obtain the cached value? The cached value is only updated every time value changes

tests/test_pfc.py
```diff
@@ -0,0 +1,101 @@
...
+
+    time.sleep(1)
+
+
+def setPortPfcAsym(dvs, port_name, pfc_asym):
+
+    cfg_db = swsscommon.DBConnector(swsscommon.CONFIG_DB, dvs.redis_sock, 0)
+
+    port_tbl = swsscommon.Table(cfg_db, 'PORT', '|')
+    fvs = swsscommon.FieldValuePairs([('pfc_asym', pfc_asym)])
```
sihuihan88: I am a little confused here. setPortPfcAsym align with doporttask in portsorch which subscribes app db instead of config db from my understanding. How come it can subscribe change in config db?

sihuihan88:
Do you have plan to implement cli to enable/disable asym pfc and show the current asym pfc status?

orchagent/portsorch.cpp
volodymyrsamotiy: "pfc" isn't equal to 0, it's just initial value, then we get current value and set it to "pfc" in line 546

orchagent/portsorch.cpp
volodymyrsamotiy: "SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL_TX" is set to current PFC value (which is retrived in line 546) in line 564

orchagent/portsorch.cpp
volodymyrsamotiy: "m_portList[port.m_alias] = port" is inside "setPortPfcAsym" intentionally.
It is because before updating "asym_mode" we need to get current value (line 546) and only then update.

orchagent/portsorch.h
volodymyrsamotiy: Implemented required changes

tests/test_pfc.py
volodymyrsamotiy: That is correct, portsorch subscribes for a port table in APP_DB.
Portsyncd subscribes for a port table in CFG_DB, handles config and applies it to APP_DB.
So then portsorch got a notification and processes new config values.

volodymyrsamotiy:
CLI is already implemented:  https://github.com/Azure/sonic-utilities/pull/229

sihuihan88:
Can you also update https://github.com/Azure/sonic-swss/blob/master/orchagent/pfcwdorch.cpp#L355 to use the new getPfc api?

sihuihan88:
I am a little confused in sonic-utilities, the pfc_asym config is written to config_db, while the portsorch only consumes app_db. Did I miss something?

sihuihan88:
Can you also verify pfcwd test passes with asym pfc enable?

volodymyrsamotiy:
Merged latest changes from 'master' and updated https://github.com/Azure/sonic-swss/blob/master/orchagent/pfcwdorch.cpp#L355

volodymyrsamotiy:
Regarding CFG_DB and APP_DB subscription question please see my previous answer:
> volodymyrsamotiy 12 days ago  Contributor
> That is correct, portsorch subscribes for a port table in APP_DB.
> Portsyncd subscribes for a port table in CFG_DB, handles config and applies it to APP_DB.
> So then portsorch got a notification and processes new config values.

volodymyrsamotiy:
As far as I know PFCWD test doesn't pass currently. Is that correct?
If so, I will be able to run PFCWD test with asymmentric PFC only once it is passed.

liatgrozovik:
@volodymyrsamotiy due to recent changes PFC WD is now passing. you can run the Asymmetric PFC feature with most recent changes and check no degradation.

sihuihan88:
@volodymyrsamotiy Can you share if PFCWD passes with asym PFC on/off?

sihuihan88:
Please resolve the conflict so that we can merge it.

PR #462, opened 2018-04-06
[orchagent]: Add internal loopback port logic

stcheng:
While doing service interfaces-config restart, we notice that
loopback interface tear down message won't be generated. Thus,
orchagent will receive duplicate loopback interface generation
messages, which will cause crash.

This commit reuses the logic for regular interfaces to ignore
duplicate messages so that orchagent won't crash when multiple
loopback interface bring up messages are received.

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -80,13 +80,6 @@ void IntfsOrch::doTask(Consumer &consumer)
         string op = kfvOp(t);
         if (op == SET_COMMAND)
         {
-            if (alias == "lo")
-            {
-                addIp2MeRoute(ip_prefix);
-                it = consumer.m_toSync.erase(it);
-                continue;
-            }
-
             Port port;
             if (!gPortsOrch->getPort(alias, port))
             {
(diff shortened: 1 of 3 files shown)


lguohan:
can you add virtual switch test?

orchagent/intfsorch.cpp
```diff
@@ -146,9 +140,16 @@ void IntfsOrch::doTask(Consumer &consumer)
...
             }
 
-            addSubnetRoute(port, ip_prefix);
             addIp2MeRoute(ip_prefix);
-            if(port.m_type == Port::VLAN && ip_prefix.isV4())
+
+            /* Add subnet route for non-loopback interfaces */
+            if (port.m_type != Port::LOOPBACK)
+            {
+                addSubnetRoute(port, ip_prefix);
```
nikos-li: This condition in sonic didn't seem correct before and doesn't seem correct now either. There is no restriction that loopbacks are /32 or /128 addresses. Seems to me if we don't install the subnet route, we can cause a routing loop with a default in place.

So what's the reason for skipping addSubnetRoute for loopbacks?

stcheng: what about adding the logic of checking if it's a /32 or /128 addresses in the addSubnetRoute function. if it's a subnet address, then we add the subnet route.

nikos-li: Can you explain what happens if the loopback configuration has changed? How will that be handled? At first glance it seems we won't reflect the right ip2meroute.

zhenggen-xu:
I have similar concerns,  why do we treat loopback differently than other interfaces?  Loopback can have no full-masked subnet. 

Also, if interface-config does not tear down the loopback, we may end up with having the old information in SONiC if things changed, we need add logic in stopping service logic to tear it down manually, like using ifdown.

stcheng:
close this pull request.
this issue will be handled more gracefully in an alternative way

PR #467, opened 2018-04-10
[pfcwd]: support BIG_RED_SWITCH mode

sihuihan88:

**What I did**
Enable support for BIG_RED_SWITCH mode:
The Expect behavior:
1. When BIG_RED_SWITCH mode is enabled:
- Regardless of PFCWD configuration or PFC storm status, PFCWD starts to drop all lossless TC packets on all ports. It's a big red button to drop all lossless packets "blindly".
- PFCWD will not do storm check/restore once BIG_RED_SWITCH is enabled. If there is some active action handler handling ongoing PFC storm when enabling BIG_RED_SWITCH, stop these handlers and rely on BIG_RED_SWITCH to drop all the packets.
- Lossy traffic should not be affected.

2. When BIG_RED_SWITCH mode is disabled:
    PFCWD goes back to normal, and start to check/restore pfc storm.

3. BIG_RED_SWITCH mode can be enabled and disabled anytime even if PFCWD was not enabled on any port before.

**Why I did it**
Per request, BIG_RED_SWITCH mode provides a button to drop all lossless packets on a switch.
**How I verified it**
Tested on DUT
**Details if related**

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -14,6 +14,7 @@
 #define PFC_WD_ACTION                   "action"
 #define PFC_WD_DETECTION_TIME           "detection_time"
 #define PFC_WD_RESTORATION_TIME         "restoration_time"
+#define BIG_RED_SWITCH_FIELD            "BIG_RED_SWITCH"
 
 #define PFC_WD_DETECTION_TIME_MAX       (5 * 1000)
 #define PFC_WD_DETECTION_TIME_MIN       100
@@ -261,6 +262,8 @@ template <typename DropHandler, typename ForwardHandler>
 void PfcWdSwOrch<DropHandler, ForwardHandler>::createEntry(const string& key,
         const vector<FieldValueTuple>& data)
 {
+    SWSS_LOG_ENTER();
+
     if (key == PFC_WD_GLOBAL)
     {
         for (auto valuePair: data)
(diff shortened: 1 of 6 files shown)


sihuihan88:
Please retest

sihuihan88:
cli https://github.com/Azure/sonic-utilities/pull/237

sihuihan88:
retest this please

orchagent/pfcwdorch.cpp
```diff
@@ -604,13 +809,20 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::doTask(swss::NotificationConsumer
...
         {
-            throw runtime_error("Unknown PFC WD action");
+            SWSS_LOG_ERROR("Unknown PFC WD action");
         }
     }
     else if (event == "restore")
     {
         if (entry->second.handler != nullptr)
         {
+            SWSS_LOG_NOTICE(
```
marian-pritsak: Why move logs from handler to this orch?

orchagent/pfcwdorch.cpp
```diff
@@ -282,6 +290,176 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::createEntry(const string& key,
...
+            entry.second.handler->commitCounters();
+            entry.second.handler = nullptr;
+        }
+    }
+
+    // Create pfcwdaction hanlder on all the ports.
+    for (auto & it: allPorts)
+    {
+        Port port = it.second;
+        if (port.m_alias != "Ethernet8")
```
marian-pritsak: What's this magic ifname?

orchagent/pfcwdorch.h
```diff
@@ -95,7 +97,13 @@ class PfcWdSwOrch: public PfcWdOrch<DropHandler, ForwardHandler>
 
     string filterPfcCounters(string counters, set<uint8_t>& losslessTc);
     string getFlexCounterTableKey(string s);
+
+    void disableBigRedSwitchMode();
+    void enableBigRedSwitchMode();
+    void setBigRedSwitchMode(string value);
+
     map<sai_object_id_t, PfcWdQueueEntry> m_entryMap;
+    map<sai_object_id_t, PfcWdQueueEntry> m_brsEntryMap;
```
marian-pritsak: Why separate map for BRS?

orchagent/pfcwdorch.cpp
```diff
@@ -282,6 +290,176 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::createEntry(const string& key,
...
+            sai_object_id_t queueId = port.m_queue_ids[i];
+            if ((pfcMask & (1 << i)) == 0 && m_entryMap.find(queueId) == m_entryMap.end())
+            {
+                continue;
+            }
+
+            string queueIdStr = sai_serialize_object_id(queueId);
+
+            vector<FieldValueTuple> countersFieldValues;
+            countersFieldValues.emplace_back("BIG_RED_SWITCH_MODE", "enable");
```
marian-pritsak: I don't think that plugins should know about BRS. Since you chose to keep them running, you might as well ignore notifications from plugins by checking a flag in orch.

orchagent/pfcwdorch.cpp
```diff
@@ -282,6 +290,176 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::createEntry(const string& key,
...
+
+        // use portorch api to get lossless tc in future.
+        sai_status_t status = sai_port_api->get_port_attribute(port.m_port_id, 1, &attr);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("Failed to get PFC mask on port %s: %d", port.m_alias.c_str(), status);
+            return;
+        }
+
+        uint8_t pfcMask = attr.value.u8;
```
marian-pritsak: This is not aligned with asymmetric PFC

orchagent/pfcwdorch.cpp
sihuihan88: Since in big-red-switch mode, we don't want to have log relates to storm detect/restore. Moving log out of handler class provides flexibility for users to input various logs depending on the use case when they create action handler

orchagent/pfcwdorch.h
sihuihan88: After disable big-red-switch, we need to keep the PFC WD original configuration. This information actually is stored in m_entryMap. It would be better if we don't mix this information with brs.

orchagent/pfcwdorch.cpp
sihuihan88: I would prefer to let plugins aware of BRS. Otherwise, syncd will runplugin and keeps sending notification every 100ms. Even though swss can ingore, it still consumes a lot unnecessary resources.

orchagent/pfcwdorch.cpp
sihuihan88: This need to be updated once asym PFC is supported. I added one comment to update this part once asy is enabled.

orchagent/pfcwdorch.cpp
marian-pritsak: Maybe then it is better to unregister plugins or have some field in DB along with poll interval that will globally turn plugins on/off?

orchagent/pfcwdorch.cpp
marian-pritsak: Oh, I thought it already was. My bad.

orchagent/pfcwdorch.cpp
marian-pritsak: ok

orchagent/pfcwdorch.cpp
sihuihan88: I think both method are mostly identical, and the only difference is at which part to skip plugins. The benefit to achieve this in lua script is it's lightweight and only affect pfcwd, also it don't need modification of syncd and makes syncd logic simpler. If you would prefer to change the logic to syncd, I can add it to backlog and change it later.

sihuihan88: Have added it to backlog. https://github.com/Azure/sonic-sairedis/issues/314

marian-pritsak: ok

orchagent/pfcwdorch.h
marian-pritsak: In line 398 in .cpp file entry is removed from the regular map when BRS is enabled, but it is not added back when BRS is disabled.
In case we had some handler, and BRS is flipped on/off, will we end up with no handler after BRS off?

sihuihan88: That’s true. The previous active entries mark the storm happens when BRS was enabled. After BRS runs for a while, the previous state is out of date so it’s unecessary to create those handlers. Instead we just need to recover the PWCWD configuration and let it detect storm. we need is just the configuration which is the entrymap

marian-pritsak: ok

PR #472, opened 2018-04-16
[crmorch]: neighbor used counter increased twice

tylerlinp:
**What I did**
Modified crmorch stats for neighbor used counter. Increase it just after create neighbor entry success. Decrease it when remove neighbor entry success in case of add nexthop failure.

**Why I did it**
Test found bug.

**How I verified it**
By code walkthrough I found neighbor used counter increased both before and after add nexthop.

**Details if related**
before add ipv4 neighbor,the value of crm_stats_ipv4_neighbor_used is 8
3) "crm_stats_ipv4_neighbor_used" 
4) "8"
after add one ipv4 neighbor,the value of crm_stats_ipv4_neighbor_used is 10
3) "crm_stats_ipv4_neighbor_used" 
4) "10"

### DIFF
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -395,16 +395,16 @@ bool NeighOrch::addNeighbor(NeighborEntry neighborEntry, MacAddress macAddress)
             }
             m_intfsOrch->decreaseRouterIntfsRefCount(alias);
 
-            return false;
-        }
+            if (neighbor_entry.ip_address.addr_family == SAI_IP_ADDR_FAMILY_IPV4)
+            {
+                gCrmOrch->decCrmResUsedCounter(CrmResourceType::CRM_IPV4_NEIGHBOR);
+            }
+            else
+            {
+                gCrmOrch->decCrmResUsedCounter(CrmResourceType::CRM_IPV6_NEIGHBOR);
+            }
 
-        if (neighbor_entry.ip_address.addr_family == SAI_IP_ADDR_FAMILY_IPV4)
-        {
-            gCrmOrch->incCrmResUsedCounter(CrmResourceType::CRM_IPV4_NEIGHBOR);
-        }
-        else
-        {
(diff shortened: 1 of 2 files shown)


lguohan:
can you sign the cla?

orchagent/neighorch.cpp
```diff
@@ -395,16 +395,16 @@ bool NeighOrch::addNeighbor(NeighborEntry neighborEntry, MacAddress macAddress)
...
+            if (neighbor_entry.ip_address.addr_family == SAI_IP_ADDR_FAMILY_IPV4)
+            {
+                gCrmOrch->decCrmResUsedCounter(CrmResourceType::CRM_IPV4_NEIGHBOR);
+            }
+            else
+            {
+                gCrmOrch->decCrmResUsedCounter(CrmResourceType::CRM_IPV6_NEIGHBOR);
+            }
 
-        if (neighbor_entry.ip_address.addr_family == SAI_IP_ADDR_FAMILY_IPV4)
```
prsunny: Looks like a previous review comment on this got missed half-way. 
Suggest to keep this here and remove the increment counter above addNextHop. You need not decrement the counter if addNextHop failed.

prsunny:
We have a testcase to verify the used counter but since the check was `">="`, this bug was not caught.
sonic-mgmt/ansible/roles/test/tasks/crm/crm_test_ipv4_neighbor.yml

```
    - name: Verify "crm_stats_ipv4_neighbor_used" counter was incremented
      assert: {that: "{{new_crm_stats_ipv4_neighbor_used|int - crm_stats_ipv4_neighbor_used|int >= 1}}"}
```

orchagent/neighorch.cpp
tylerlinp: I thought same as you said at the beginning, but then found when failed both addNextHop and remove_neighbor_entry we should add increment there, that is less readability.

prsunny:
@lguohan, VS test also checks `>=`. I'll fix the test cases.

PR #473, opened 2018-04-16
[PFCWD]: Periodically poll WD counters

marian-pritsak:
For every action handler call commitCounters on a periodic basis

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -22,6 +22,7 @@
 #define PFC_WD_POLL_TIMEOUT             5000
 #define SAI_PORT_STAT_PFC_PREFIX        "SAI_PORT_STAT_PFC_"
 #define PFC_WD_TC_MAX 8
+#define COUNTER_CHECK_POLL_TIMEOUT_SEC  5
 
 extern sai_port_api_t *sai_port_api;
 extern sai_queue_api_t *sai_queue_api;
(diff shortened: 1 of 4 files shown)


orchagent/pfcwdorch.cpp
```diff
@@ -22,6 +22,7 @@
 #define PFC_WD_POLL_TIMEOUT             5000
 #define SAI_PORT_STAT_PFC_PREFIX        "SAI_PORT_STAT_PFC_"
 #define PFC_WD_TC_MAX 8
+#define COUNTER_CHECK_POLL_TIMEOUT_SEC  5
```
lguohan: prefer 1 second

nikos-li: Is the polling interval controlled via cli or via config_db.json? If yes can you please explain how?

If not this needs fixing prior to committing.

lguohan: this is not controllable from the cli, this is only polling queue when the queue in pfc recovery state, in this case, we need to periodically poll it to get dropped packets. I do not see a generic need to make it controllable from cli.

nikos-li: @lguohan All polling intervals need to be controlled via the cli and should provide the ability to be disabled.

lguohan:
retest this please

lguohan:
@nikos-li , this is a internal feature and the user should not care about it and does not need to control that.

PR #475, opened 2018-04-17
[orchagent]: Use priorities in orchagent

pavel-shirshov:
**What I did**
I've added support of Selectable priorities for orchagent.
I've put following priorities to orchagent tables:
```
APP_PORT_TABLE_NAME = 45
APP_VLAN_TABLE_NAME = 42
APP_LAG_TABLE_NAME = 42
APP_VLAN_MEMBER_TABLE_NAME = 40
APP_LAG_MEMBER_TABLE_NAME = 40
APP_INTF_TABLE_NAME = 35
APP_NEIGH_TABLE_NAME= 30
APP_FDB_TABLE_NAME = 20
APP_ROUTE_TABLE_NAME = 5
```
All other orchagent run with lowest priority which is 0.

**Why I did it**
To prioritize messages inside of the orchagent message queue.

**How I verified it**
Build swss and run it

**Details if related**

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -16,8 +16,10 @@ extern sai_object_id_t  gSwitchId;
 extern PortsOrch*       gPortsOrch;
 extern CrmOrch *        gCrmOrch;
 
+const int fdborch_pri = 20;
+
 FdbOrch::FdbOrch(DBConnector *db, string tableName, PortsOrch *port) :
-    Orch(db, tableName),
+    Orch(db, tableName, fdborch_pri),
     m_portsOrch(port),
     m_table(Table(db, tableName))
 {
--- a/orchagent/orch.h
+++ b/orchagent/orch.h
@@ -5,6 +5,7 @@
 #include <unordered_set>
 #include <map>
 #include <memory>
+#include <utility>
 
 extern "C" {
 #include "sai.h"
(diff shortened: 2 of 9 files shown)


orchagent/fdborch.cpp
```diff
@@ -16,8 +16,10 @@ extern sai_object_id_t  gSwitchId;
 extern PortsOrch*       gPortsOrch;
 extern CrmOrch *        gCrmOrch;
 
+const int fdborch_pri = 20;
```
qiluo-msft: >fdborch_pri [](start = 10, length = 11)

It's inconvenient to scatter all the similar priority settings to many files.

orchagent/orch.h
```diff
@@ -36,6 +37,8 @@ const char config_db_key_delimiter = '|';
 #define CONFIGDB_KEY_SEPARATOR "|"
 #define DEFAULT_KEY_SEPARATOR  ":"
 
+const int default_orch_pri = 0;
```
qiluo-msft: >default_orch_pri [](start = 10, length = 16)

Move it to class local?

pavel-shirshov:
Hi Qi,

I'm going to refactor the whole orchdaemon class to make it readable including priorities. I didn't want to refactor in this PR, otherwise this PR would be stuck for another week.

PR #476, opened 2018-04-17
Updated CRM test case to check for exact used counter

prsunny:
**What I did**
Updated CRM test case to check for exact used counter for IPv4/v6 neighbor tables

**Why I did it**
https://github.com/Azure/sonic-swss/pull/472 fixed the double increment for neighbor counters.
This PR is to check for the usage based on exact used counter.

**How I verified it**
Run CRM VS test case (In-progress)

**Details if related**
N/A


prsunny:
@volodymyrsamotiy , Please review

PR #478, opened 2018-04-18
[orchagent]: Remove global lock caused by notifications running in another thread

pavel-shirshov:
**What I did**
I removed the notification handlers which were controlled by libsairedis and added direct support of notifications from ASIC_DB. 

**Why I did it**
To remove global lock from orchagent.

**How I verified it**
Install into the switch and run it.
Also /var/log/syslog shows notifications from port status change.

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -355,10 +355,6 @@ void FdbOrch::doTask(NotificationConsumer& consumer, const std::string &name)
             sai_deserialize_free_fdb_event_ntf(count, fdbevent);
         }
     }
-    else
-    {
-        SWSS_LOG_ERROR("Wrong NotificationConsumer name: '%s'", name.c_str());
-    }
 }
 
 void FdbOrch::updateVlanMember(const VlanMemberUpdate& update)
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -9,6 +9,7 @@
 #include "fdborch.h"
 #include "crmorch.h"
 #include "notifier.h"
+#include "sai_serialize.h"
 
 extern sai_fdb_api_t    *sai_fdb_api;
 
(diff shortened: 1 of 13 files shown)


orchagent/orch.h
```diff
@@ -152,7 +152,7 @@ class Orch
 
     /* Run doTask against a specific executor */
     virtual void doTask(Consumer &consumer) = 0;
-    virtual void doTask(NotificationConsumer &consumer) { }
+    virtual void doTask(NotificationConsumer &consumer, const std::string &name) { }
```
lguohan: >const std::string &name [](start = 56, length = 23)

can you add a comment on the name, what does this name use for?

orchagent/fdborch.cpp
```diff
@@ -25,8 +26,14 @@ FdbOrch::FdbOrch(DBConnector *db, string tableName, PortsOrch *port) :
 {
     m_portsOrch->attach(this);
     auto consumer = new NotificationConsumer(db, "FLUSHFDBREQUEST");
-    auto fdbNotification = new Notifier(consumer, this);
+    auto fdbNotification = new Notifier(consumer, this, "FLUSHFDBREQUEST");
```
qiluo-msft: >fdbNotification [](start = 9, length = 15)

Confused about fdbNotification vs fdb_notification. What is the diff?

orchagent/notifier.h
```diff
@@ -2,8 +2,8 @@
 
 class Notifier : public Executor {
 public:
-    Notifier(NotificationConsumer *select, Orch *orch)
-        : Executor(select, orch)
+    Notifier(NotificationConsumer *select, Orch *orch, const std::string& notifier_name)
```
qiluo-msft: >notifier_name [](start = 74, length = 13)

I notice there is a similar name at NotificationConsumer.m_channel. I prefer using that one instread of a new parameter.

orchagent/fdborch.cpp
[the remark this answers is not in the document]
pavel-shirshov: With enumerations we will have hardcoded enumerations.
I think in this case it's better to develop better abstraction than NotificationConsumer(). Currently it's connected too deep into Orch class (doTask methods for NotificationConsumer). We should decouple them from each other and implement doTask methods as a base class for Orch classes.

orchagent/orch.h
pavel-shirshov: Comments added

pavel-shirshov: I've also renamed this parameter to more clear name

orchagent/fdborch.cpp
pavel-shirshov: Renamed all the variable names for notification related stuff

orchagent/notifier.h
pavel-shirshov: Renamed to m_channel

orchagent/fdborch.cpp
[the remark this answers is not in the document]
zhenggen-xu: I agree more changes is needed for a better abstraction than what we have today. but I am not sure if you are going to do it in this PR.
My point was, at least we should put the message types in a common place and reuse then on different places so you don't end up with mistype something and get run-time problem.  enumerations will detect the errors during the build time not run time if you mistyped, also by putting things in the same place improves manageability and extendability.

PR #479, opened 2018-04-19
[pfcwd]: create PFCWD acl instead of L3 ACL

sihuihan88:

**What I did**
Create a new ACL type called for PFCWD
**Why I did it**
L3 ACL has many qualifiers that are not needed for PFCWD. Creating L3 ACL consumes more resources than needed, which will cause issue when the ACL resources is limited.
**How I verified it**
Tested on DUT
**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -498,6 +498,11 @@ shared_ptr<AclRule> AclRule::makeShared(acl_table_type_t type, AclOrch *acl, Mir
     {
         return make_shared<AclRuleL3V6>(acl, rule, table, type);
     }
+    /* Pfcwd rules can exist only in PFCWD table */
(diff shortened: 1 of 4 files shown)


orchagent/aclorch.cpp
```diff
@@ -971,6 +994,26 @@ bool AclTable::create()
     attr.value.s32list.list = bpoint_list.data();
     table_attrs.push_back(attr);
 
+    if (type == ACL_TABLE_PFCWD)
+    {
+        attr.id = SAI_ACL_TABLE_ATTR_FIELD_TC;
```
prsunny: Just a thought.. we can have TC and STAGE outside the _if_ so, no need to duplicate the code below.

sihuihan88:
Retest this please

sihuihan88:
Retest this please

PR #481, opened 2018-04-20
[acl]: test different subnet masks

simone-dell:
**What I did**
Added a test case that checks that different subnet mask for dest/src ip addresses in ACL rules are programmed correctly in the ASICDB

**Why I did it**
Part of the ACL test cases propsed by Dell and approved by Microsoft

**How I verified it**
Ran the python test script in the repo

**Details if related**
Output of test attatched
[subnet_test_logs.txt](https://github.com/Azure/sonic-swss/files/1930588/subnet_test_logs.txt)

### DIFF
--- a/tests/test_acl.py
+++ b/tests/test_acl.py
@@ -13,7 +13,7 @@ class TestAcl(object):
         assert len(acl_tables) == 1 
 
         return acl_tables[0]
- 
+
     def test_AclTableCreation(self, dvs):
     
         db = swsscommon.DBConnector(4, dvs.redis_sock, 0)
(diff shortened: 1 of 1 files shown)


tests/test_acl.py
```diff
@@ -779,3 +779,105 @@ def test_V6AclTableDeletion(self, dvs):
...
+
+        acl_entry = [k for k in keys if k not in dvs.asicdb.default_acl_entries]
+        assert len(acl_entry) == subnet_mask_rules
+
+        matched_masks = 0
+        for entry in acl_entry:
+           (status, fvs) = atbl.get(entry)
+           assert status == True
+           assert len(fvs) == 6
+           if ('SAI_ACL_ENTRY_ATTR_FIELD_SRC_IP', '23.103.0.0&mask:255.255.192.0') in fvs:
```
lguohan: there is code redundancy in the block as well.

can you define a function as below as use the function to check all entries in acl_entry?

check_rule_existence(entry, rules):
      for rule in rules:
          check if src_ip in the rule match the entry
          check if action of the rule match the entry
          check if priority of the rule match the entry
          if all match:
              then return the True
    if none of the rule match, return false

PR #482, opened 2018-04-20
[acl]: Insert new acl rule with priority in between other rules

simone-dell:
**What I did**
I added a new acl test case: Four ACL rules are configured with priorities 10, 20, 30, 40.  Then I added another rule with priority 21.  I made sure all five rules were pushed to the ASICDB and with the proper priorities.

**Why I did it**
Part of the ACL test cases suggested by Dell and approved by Microsoft

**How I verified it**
Ran the python script in the repo

**Details if related**
Logs are attached
[insert_acl_rule_logs.txt](https://github.com/Azure/sonic-swss/files/1930605/insert_acl_rule_logs.txt)

### DIFF
--- a/test_acl.py
+++ b/test_acl.py
@@ -1,0 +1,894 @@
+from swsscommon import swsscommon
+import time
+import re
+import json
+
+class TestAcl(object):
+    def get_acl_table_id(self, dvs, adb):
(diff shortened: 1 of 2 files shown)


simone-dell:
Thanks for pointing it out.  I have removed the extra file

simone-dell:
@lguohan  I used the same logic in pull #481

PR #483, opened 2018-04-21
[counters]: separate query of port/queue counters

sihuihan88:

**What I did**
Separate port and queue query. By default query port counters every 1s, and queue counters every 10s
**Why I did it**
Querying all queue counters every 1s is cpu consuming. Reducing queue counter query rate helps to reduce cpu overhead. 
**How I verified it**
Tested on DUT.
**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -34,8 +34,10 @@ extern CrmOrch *gCrmOrch;
 
 #define VLAN_PREFIX         "Vlan"
 #define DEFAULT_VLAN_ID     1
-#define FLEX_STAT_COUNTER_POLL_MSECS "1000"
-#define STAT_COUNTER_FLEX_COUNTER_GROUP "STAT_COUNTER"
+#define PORT_FLEX_STAT_COUNTER_POLL_MSECS "1000"
+#define QUEUE_FLEX_STAT_COUNTER_POLL_MSECS "10000"
+#define PORT_STAT_COUNTER_FLEX_COUNTER_GROUP "PORT_STAT_COUNTER"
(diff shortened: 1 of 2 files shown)


nikos-li:
Are both polling intervals controlled by cli?

orchagent/portsorch.cpp
```diff
@@ -34,8 +34,10 @@ extern CrmOrch *gCrmOrch;
 
 #define VLAN_PREFIX         "Vlan"
 #define DEFAULT_VLAN_ID     1
-#define FLEX_STAT_COUNTER_POLL_MSECS "1000"
-#define STAT_COUNTER_FLEX_COUNTER_GROUP "STAT_COUNTER"
+#define PORT_FLEX_STAT_COUNTER_POLL_MSECS "1000"
```
nikos-li: Are both polling intervals controlled by cli or through fields in config_db.json?
If yes, can you please tell me how?

If not, then those intervals need to have more control.

nikos-li: Also queue counters polling should not be enabled by default.

lguohan: @nikos-li , please create feature request in the issue, we'll triage later.

nikos-li: @lguohan This isn't a feature request, nor a proper solution to the CPU increase we have observed. This needs to be discussed in the 3-way meeting. Please include it in the agenda.

nikos-li: @lguohan @sihuihan88 Please also provide CPU util data of before and after.

nikos-li: @lguohan @sihuihan88 Do not merge this without proper RCA and explicit approval and discussion in the 3-way meeting.

PR #484, opened 2018-04-23
Fix tables handling race condition in buffermgr

andriymoroz-mlnx:

**What I did**
 Fixed issue which may leave switch not fully configured when speed is applied before cable length table

**Why I did it**
Bug fix

**How I verified it**
Forced CABLE_LENGTH to be applied the last and checked switch completes config after all input i available.

**Details if related**

### DIFF
--- a/cfgmgr/buffermgr.cpp
+++ b/cfgmgr/buffermgr.cpp
@@ -72,9 +72,10 @@ void BufferMgr::readPgProfileLookupFile(string file)
     infile.close();
 }
 
-void BufferMgr::doCableTask(string port, string cable_length)
+task_process_status BufferMgr::doCableTask(string port, string cable_length)
 {
     m_cableLenLookup[port] = cable_length;
+    return task_process_status::task_success;
 }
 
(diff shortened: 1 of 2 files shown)


cfgmgr/buffermgr.cpp
```diff
@@ -198,25 +200,37 @@ void BufferMgr::doTask(Consumer &consumer)
...
                     // receive and cache cable length table
-                    doCableTask(fvField(i), fvValue(i));
+                    task_status = doCableTask(fvField(i), fvValue(i));
                 }
                 // In case of PORT table update, Buffer Manager is interested in speed update only
                 if (table_name == CFG_PORT_TABLE_NAME && fvField(i) == "speed")
                 {
                     // create/update profile for port
-                    doSpeedUpdateTask(port, fvValue(i));
+                    task_status = doSpeedUpdateTask(port, fvValue(i));
```
lguohan: task_status is going to get overridden in the for loop.

andriymoroz-mlnx: right... fixed

PR #487, opened 2018-04-24
 Add Cable and Buffer test

chitra-raghavan:
**What I did**
Adding Testcase for verifying dynamic buffer_profile depending upon cable and speed length

**Why I did it**
No existing script to check BufferProfile depending upon cable length

**How I verified it**
Running the script using vs in vs environment

### DIFF
--- a/tests/test_cableBuffer.py
+++ b/tests/test_cableBuffer.py
@@ -1,0 +1,157 @@
+"""
+    test_Cablebuffer.py implements list of tests to check  whether the approprotae buffer profile
+    is selected for an interface depending upon cable and Speed values .
+
+    These tests need to be run in prepared environment and with the
+    SONiC version compiled for PLATFORM=vs
+
+    See README.md for details
+"""
+
+import os
+import re
+import time
+import docker
+import pytest
+import commands
+from swsscommon import swsscommon
+
+import os 
+import pytest
+class TestCableBufferProfile(object):
+    cable_list = ['5m' , '40m' , '300m']
+    speed_list = ['50000', '25000', '40000', '10000', '100000']
+    num_ports=32
+
+    def getProfileDetails(self):
+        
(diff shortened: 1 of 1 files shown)


tests/test_cableBuffer.py
```diff
@@ -0,0 +1,134 @@
...
+import time
+import docker
+import pytest
+import commands
+from swsscommon import swsscommon
+import pytest
+
+
+class TestCableBufferProfile(object):
+    cable_list = ['5m', '40m', '300m']
```
yxieca: Can you rename this set to cable_len_list or cable_length_list?

tests/test_cableBuffer.py
```diff
@@ -0,0 +1,134 @@
...
+                # set same speed on all ports
+                for i in range(0, self.num_ports):
+                    cfg_port_table.set("Ethernet%d" % (i*4), fvs)
+                time.sleep(1)
+
+                # check the speed was set
+                asic_port_records = asic_port_table.getKeys()
+                assert len(asic_port_records) == (self.num_ports + 1)  # +CPU port
+                num_set = 0
+                for k in asic_port_records:
```
prsunny: Is the CPU port assumed to be at the end of records? Else I think the below logic would fail right?

chitra-raghavan: There is an  conditional check for getting the speed Fv for all the port records .For cpu port case,  conditional check will get skipped and it will not  evaluated. So this has been handled in the below code.

                    (status, fvs) = asic_port_table.get(k)
                    assert status == True
                    for fv in fvs:
                        if fv[0] == "SAI_PORT_ATTR_SPEED":
                            assert fv[1] == speed
                            num_set += 1

chitra-raghavan:
@prsunny , Any update upon when it will be merged ?

prsunny:
@yxieca, this is approved from my side. Can you sign-off?

PR #493, opened 2018-05-02
[acl]: fixed & refactored bind ACL table method.

nazarii-gnydyn:
**What I did**
Fixed bind ACL table method.

**Why I did it**
Applying ACL Table Group object without members created results in SAI error.

**How I verified it**
Created ACL table and binded it to port.

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -484,27 +484,25 @@ bool PortsOrch::bindAclTable(sai_object_id_t id, sai_object_id_t table_oid, sai_
 
     auto &port = m_portList.find(p.m_alias)->second;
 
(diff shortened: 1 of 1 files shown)


lguohan:
we should be able to create empty acl table group and the bind to the port, I do not understand why sai won't allow us to do it.

NStetskovych:
@lguohan yes, we can close it. But here is still the an issue with CRM counters:
https://github.com/Azure/sonic-swss/blob/201803/orchagent/portsorch.cpp#L556
(group_attr.value.s32  look at the group type but not to the stage)
We can provide separate PR or take this as it does not break anything.

prsunny:
@NStetskovych , can you provide a separate PR for the CRM counter fix since without it, the counter value would be unpredictable.

nazarii-gnydyn:
@lguohan You are right. Declining this because changes are redundant.

PR #496, opened 2018-05-07
[crm] Fix Crm Acl used counter update

NStetskovych:
Signed-off-by: Nadiya Stetskovych <Nadiya.Stetskovych@cavium.com>

**What I did**
Fix type when updating ACl CRM Counters

PR #498, opened 2018-05-09
 Fix issue with update ACL TABLE CRM error

JaredLius:
**What I did**
 Fix issue with update ACL TABLE CRM error

**Why I did it**
when execute command "crm show resource all", it display wrong used count on acl_table. the ingress used acl_table was counted to egress used. from the code, when it call function gCrmOrch->incCrmAclUsedCounter, the second parameter is the acl stage, but the code use a variable attr.value.s32 which has been modified, it is not equal to acl stage any more.

**How I verified it**
I make a new swss_1.0.0_amd64.deb, and update it to switch. after reboot, when execute "crm show resource all", it can display the acl_table used count correctly.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -553,7 +553,9 @@ bool PortsOrch::bindAclTable(sai_object_id_t id, sai_object_id_t table_oid, sai_
             port.m_egress_acl_table_group_id = groupOid;
         }
 
-        gCrmOrch->incCrmAclUsedCounter(CrmResourceType::CRM_ACL_GROUP, (sai_acl_stage_t) group_attr.value.s32, SAI_ACL_BIND_POINT_TYPE_PORT);
+        sai_acl_stage_t acl_stage = ingress ? SAI_ACL_STAGE_INGRESS : SAI_ACL_STAGE_EGRESS;
+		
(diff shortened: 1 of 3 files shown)


orchagent/portsorch.cpp
```diff
@@ -553,7 +553,9 @@ bool PortsOrch::bindAclTable(sai_object_id_t id, sai_object_id_t table_oid, sai_
             port.m_egress_acl_table_group_id = groupOid;
         }
 
-        gCrmOrch->incCrmAclUsedCounter(CrmResourceType::CRM_ACL_GROUP, (sai_acl_stage_t) group_attr.value.s32, SAI_ACL_BIND_POINT_TYPE_PORT);
+        sai_acl_stage_t acl_stage = ingress ? SAI_ACL_STAGE_INGRESS : SAI_ACL_STAGE_EGRESS;
```
prsunny: This was a bug that was fixed and merged recently. Can you check the latest code?

JaredLius: I have checked the latest code on master branch, but did not see the fix code. Could you please tell me in which commit it merged? thank you.

prsunny: https://github.com/Azure/sonic-swss/pull/496

lguohan:
retest this please

lguohan:
test failure, please fix.

stcheng:
retest this please

PR #499, opened 2018-05-10
[flexcounter]: add support to change port/queue counter poll interval

sihuihan88:

**What I did**
Add support to change port/queue counter poll interval
**Why I did it**

**How I verified it**
Tested on DUT.
**Details if related**

### DIFF
--- a/orchagent/flexcounterorch.cpp
+++ b/orchagent/flexcounterorch.cpp
@@ -1,0 +1,76 @@
+#include <unordered_map>
+#include "flexcounterorch.h"
+#include "portsorch.h"
+#include "select.h"
+#include "notifier.h"
+#include "redisclient.h"
+#include "sai_serialize.h"
+
+extern sai_port_api_t *sai_port_api;
+
+extern PortsOrch *gPortsOrch;
+
+unordered_map<string, string> flexCounterGroupMap =
+{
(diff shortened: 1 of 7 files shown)


orchagent/flexcounterorch.cpp
```diff
@@ -0,0 +1,76 @@
...
+        auto data = kfvFieldsValues(t);
+
+        if (!flexCounterGroupMap.count(key))
+        {
+            SWSS_LOG_NOTICE("Invalid flex counter group input, %s", key.c_str());
+            consumer.m_toSync.erase(it++);
+            continue;
+        }
+
+        if (op == SET_COMMAND)
```
marian-pritsak: IFAIK if nothing is set, the default value of 1 second is used.
Then I think we also have to set the default back if the entry is deleted from DB.

sihuihan88:
retest this please

orchagent/flexcounterorch.cpp
sihuihan88: This entry only has set_command and no del command. We can always use set command to change the interval value.

lguohan:
retest this please

orchagent/flexcounterorch.cpp
marian-pritsak: ok

PR #500, opened 2018-05-10
[teamsyncd]: Add team_ifindex2ifname return value check

stcheng:
Without the return value check, it is possible that ifname is
empty due to interface not found.

### DIFF
--- a/teamsyncd/teamsync.cpp
+++ b/teamsyncd/teamsync.cpp
@@ -160,11 +160,19 @@ int TeamSync::TeamPortSync::onChange()
         bool enabled;
 
         ifindex = team_get_port_ifindex(port);
-        team_ifindex2ifname(m_team, ifindex, ifname, MAX_IFNAME);
+
+        /* Skip if interface is not found */
+        if (!team_ifindex2ifname(m_team, ifindex, ifname, MAX_IFNAME))
+        {
+            SWSS_LOG_INFO("Interface ifindex(%u) is not found", ifindex);
+            continue;
+        }
 
         /* Skip the member that is removed from the LAG */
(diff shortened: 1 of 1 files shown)


teamsyncd/teamsync.cpp
```diff
@@ -160,7 +160,10 @@ int TeamSync::TeamPortSync::onChange()
         bool enabled;
 
         ifindex = team_get_port_ifindex(port);
-        team_ifindex2ifname(m_team, ifindex, ifname, MAX_IFNAME);
+
+        /* Skip if interface is not found */
+        if (!team_ifindex2ifname(m_team, ifindex, ifname, MAX_IFNAME))
+            continue;
```
SAI-shared-user: print out info message saying I cannot find the ifname for this index.

teamsyncd/teamsync.cpp
```diff
@@ -160,11 +160,19 @@ int TeamSync::TeamPortSync::onChange()
         bool enabled;
 
         ifindex = team_get_port_ifindex(port);
-        team_ifindex2ifname(m_team, ifindex, ifname, MAX_IFNAME);
+
+        /* Skip if interface is not found */
+        if (!team_ifindex2ifname(m_team, ifindex, ifname, MAX_IFNAME))
+        {
+            SWSS_LOG_INFO("Interface ifindex:%d is not found", ifindex);
```
lguohan: %u

lguohan: team member interface ifindex(%u) cannot be found.

stcheng: updated

PR #503, opened 2018-05-14
[portsorch]: Don't print error when alias equal to PortConfigDone

pavel-shirshov:
Don't print 'orchagent: :- doPortTask: Failed to get port id by alias PortConfigDone error'

**What I did**
Don't print 'orchagent: :- doPortTask: Failed to get port id by alias PortConfigDone error'

**Why I did it**
Because this error message was an error

**How I verified it**
Build swss and run it

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1220,7 +1220,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
             }
 
             Port p;
-            if (!getPort(alias, p))
+            if (!getPort(alias, p) && alias != "PortConfigDone")
             {
                 SWSS_LOG_ERROR("Failed to get port id by alias:%s", alias.c_str());
             }
@@ -1332,7 +1332,9 @@ void PortsOrch::doPortTask(Consumer &consumer)
             }
         }
         else
+        {
             SWSS_LOG_ERROR("Unknown operation type %s", op.c_str());
+        }
 
         it = consumer.m_toSync.erase(it);
     }


orchagent/portsorch.cpp
```diff
@@ -1220,7 +1220,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
             }
 
             Port p;
-            if (!getPort(alias, p))
+            if (!getPort(alias, p) && alias != "PortConfigDone")
```
jleveque: The PORT_TABLE contains two non-port "flag" entries: `PortConfigDone` and `PortInitDone`. You will probably want to ignore both.

We really need to devise a way to remove them from the table. They don't belong there.

pavel-shirshov: Joe, I agree with you, but in this PR I want to make a fast fix of the wrong error message.
I don't want to change the swss architecture in this PR

jleveque: Understood. For this PR I suggest you also ignore `PortInitDone`.

Changing the architecture is a TODO.

pavel-shirshov: We don't see `PortInitDone` because the code exits from the method as soon as it sees alias 'PortInitDone'.

pavel-shirshov: https://github.com/Azure/sonic-swss/blob/master/orchagent/portsorch.cpp#L1095

rodnymolina: Agree with you both, we should find a new location in redisDB for all this metadata, or identify a different channel to transfer this state. PortInitDone and PortConfigDone are not the only cases, we also have "CONFIG_DB_INITIALIZED" and i'm sure i'm missing one or two examples more.

zhenggen-xu: Looks like it is late in the review since it is "merged", but to not calling the getPort unnecessarily , we should use below logic:

            if (alias == "PortConfigDone" || !getPort(alias, p))
            {
                if (alias != "PortConfigDone")
                {
                    SWSS_LOG_ERROR("Failed to get port id by alias:%s", alias.c_str());
                }
            }

PR #505, opened 2018-05-16
[portsorch]Fix port queue index init bug

yangbashuang:
**What I did**
get port queue index from vender through attribute "SAI_QUEUE_ATTR_INDEX"
**Why I did it**
should get queue index from attribute "SAI_QUEUE_ATTR_INDEX", which is vender specific
for unicast queue index may start from 0, and mc queue the same

together with another fix on queuestats in sonic-utilities:
https://github.com/Azure/sonic-utilities/pull/255

**How I verified it**
root@switch1:/home/admin# show queue counters Ethernet1
Command: queuestat -p Ethernet1
     Port    TxQ    Counter/pkts    Counter/bytes    Drop/pkts    Drop/bytes
---------  -----  --------------  ---------------  -----------  ------------
Ethernet1    UC0               0                0            0             0
Ethernet1    UC1               0                0            0             0
Ethernet1    UC2               0                0            0             0
Ethernet1    UC3               0                0            0             0
Ethernet1    UC4               0                0            0             0
Ethernet1    UC5               0                0            0             0
Ethernet1    UC6               0                0            0             0
Ethernet1    UC7             106            23728            0             0
Ethernet1    MC0               0                0            0             0
Ethernet1    MC1               0                0            0             0
Ethernet1    MC2               0                0            0             0
Ethernet1    MC3               0                0            0             0

root@switch1:/home/admin# 
**Details if related**


orchagent/portsorch.cpp
```diff
@@ -868,10 +869,12 @@ bool PortsOrch::getQueueType(sai_object_id_t queue_id, string &type)
         type = "SAI_QUEUE_TYPE_MULTICAST";
         break;
     default:
-        SWSS_LOG_ERROR("Got unsupported queue type %d for %lu queue", attr.value.s32, queue_id);
+        SWSS_LOG_ERROR("Got unsupported queue type %d for %lu queue", attr[0].value.s32, queue_id);
         throw runtime_error("Got unsupported queue type");
     }
 
+    index = attr[1].value.s8;
```
qiluo-msft: >s8 [](start = 26, length = 2)

why assign signed to unsigned?

orchagent/portsorch.cpp
```diff
@@ -1768,19 +1771,19 @@ void PortsOrch::initializeQueues(Port &port)
...
-        FieldValueTuple queueIndexTuple(
-                sai_serialize_object_id(port.m_queue_ids[queueIndex]),
-                to_string(queueIndex));
-        queueIndexVector.push_back(queueIndexTuple);
-
-
         string queueType;
-        if (getQueueType(port.m_queue_ids[queueIndex], queueType))
+        uint8_t queueRealIndex = 0;
+        if (getQueueTypeAndIndex(port.m_queue_ids[queueIndex], queueType, queueRealIndex))
```
qiluo-msft: >getQueueTypeAndIndex [](start = 12, length = 20)

handle return code?

yangbashuang: Thanks for review, will have error syslog in "getQueueTypeAndIndex" if failed, so here use the same logic as code in left before. i think "getQueueTypeAndIndex" fail is not that critical, the initialize could continue, we can check syslog about this error
what do you suggest?

yangbashuang:
@hui-ma 
yes, "SAI_QUEUE_ATTR_INDEX"is vender specific attribute, both UC and MC 'index should start from 0, queueRealIndex will be in COUNTERS_QUEUE_INDEX_MAP

lguohan:
this is introducing to use a new sai attribute, we need to make sure the existing SAI supports this. @qiluo-msft , @hui-ma .

yangbashuang:
@lguohan "SAI_QUEUE_ATTR_INDEX" is supported in sai 1.2.4 which used in Azure/sonic-buildimage

stcheng:

yangbashuang:
> 
> 
> could you resolve the conflicts?

@stcheng resolved

stcheng:
retest this please

PR #511, opened 2018-05-24
[counter]: add support to enable/disable counter query

sihuihan88:

**What I did**
Add support to enable/disable counter query
**Why I did it**
Add the flexibility to enable or disable counter query.
**How I verified it**
Tested on DUT
**Details if related**

### DIFF
--- a/orchagent/flexcounterorch.cpp
+++ b/orchagent/flexcounterorch.cpp
@@ -16,6 +16,7 @@ unordered_map<string, string> flexCounterGroupMap =
     {"QUEUE", QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP},
 };
 
+
 FlexCounterOrch::FlexCounterOrch(DBConnector *db, vector<string> &tableNames):
     Orch(db, tableNames),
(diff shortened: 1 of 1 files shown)


sihuihan88:
depends on https://github.com/Azure/sonic-swss-common/pull/202 and https://github.com/Azure/sonic-sairedis/pull/322

lguohan:
retest this please

orchagent/flexcounterorch.cpp
```diff
@@ -61,13 +62,22 @@ void FlexCounterOrch::doTask(Consumer &consumer)
                 const auto &field = fvField(valuePair);
                 const auto &value = fvValue(valuePair);
 
-                if (field ==  POLL_INTERVAL_FIELD)
+                if (field == POLL_INTERVAL_FIELD)
```
lguohan: I do not recall we need the FIELD suffix now, when did we introduce them?

sihuihan88: They were introduced before. I can have a separate PR to clean it up later if that is preferred

PR #516, opened 2018-06-07
[Aclorch] fix Acl bind point type issue

keboliu:
**What I did**
Change the previous way to add ACL table bind point type. previously it will check all the port type in the ACL table to decide which type of bind point will be added to this table. Now by default, an ACL table will support both PORT and LAG bind point, no need to go through all the port and check the port type. 

**Why I did it**
The original way will cause an issue when ACL member LAG is not configured yet but the table already created, then the LAG bind point will not be added and result in failure later when the LAG configured and bind the ACL table to this LAG.

**How I verified it**
Run ACL/Everflow and CRM test on the testbed.

**Details if related**


qiluo-msft:
Please also cherry pick this PR to 201803 after merged into master.

orchagent/aclorch.cpp
```diff
@@ -952,43 +952,7 @@ bool AclTable::create()
...
-        }
-    }
-
-    vector<int32_t> bpoint_list;
-    for (auto bind : binds)
-    {
-        bpoint_list.push_back(bind);
-    }
-
+    vector<int32_t> bpoint_list = { SAI_ACL_BIND_POINT_TYPE_PORT, SAI_ACL_BIND_POINT_TYPE_LAG };
```
qiluo-msft: Also add SAI_ACL_BIND_POINT_TYPE_VLAN?

keboliu: yes, can add this, then we don't need to change the code in the future when deciding to support VLAN.

keboliu:
well, let's remove it.

qiluo-msft:
Offline discussed with @prsunny. Our goal is:
1. merge https://github.com/Azure/sonic-buildimage/pull/1712 into 201803 branch
2. merge this PR into master and 201803 branch

However one ASIC SAI implementation is not ready right now. The ETA for unblocking is 7/4.

PR #517, opened 2018-06-07
[Aclorch] hard code table name separator to make it compatible with 201803 branch

keboliu:
**What I did**

Hard code the table name separator with a Macro instead of getting it by calling swss-common functions.

**Why I did it**

On the 201803 branch, the swss-common submodule have an issue when the table obeject constructed without a table separator, to make the fix #https://github.com/Azure/sonic-buildimage/pull/1712 also work on the branch, change it.

**How I verified it**
Run ACL/Everflow/CRM test.

**Details if related**
If decided to update the swss-common submodule on the branch, then this PR not needed anymore.


qiluo-msft:
Not against master, it should be a quick bug fix for 201803 branch.

PR #518, opened 2018-06-07
[Aclorch] hard code table name separator to make it compatible with 201803 branch

keboliu:
**What I did**
Hard code the table name separator with a Macro instead of getting it by calling swss-common functions.

**Why I did it**
On the 201803 branch, the swss-common submodule have an issue when the table obeject constructed without a table separator, to make the fix  #https://github.com/Azure/sonic-buildimage/pull/1712 also work on the branch, so make this change.

**How I verified it**
Run ACL/Everflow/CRM test.

**Details if related**
If decided to update the swss-common submodule on the branch, then this PR not needed anymore.

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -12,6 +12,8 @@
 using namespace std;
 using namespace swss;
 
+#define DB_TABLE_NAME_SEPARATOR_VBAR '|'
+
 mutex AclOrch::m_countersMutex;
 map<acl_range_properties_t, AclRange*> AclRange::m_ranges;
 condition_variable AclOrch::m_sleepGuard;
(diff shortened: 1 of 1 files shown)


jleveque:
@zhenggen-xu: I created a non-backwards-compatible change to the way swss-common deals with table name separators. I removed many instances of hard-coding so as to allow for multiple table name separators (since we are currently transitioning from ":" to "|" and both are used). The table name separator is now an attribute of each DB.

However, these commits were not merged into the 201803 branch. We are only merging small, crucial changes into the branch as we deem necessary. However, @keboliu added features in the master branch that we wanted to pull into 201803, but it relied on my new table name separator changes. This PR is to modify his changes to work with the old swss-common which is present in the 201803 branch.

zhenggen-xu:
Thanks, assuming the risk has been assessed, it would always be good to move things forward.

keboliu:
@zhenggen-xu if the related change in master will be merged to 201803 branch,  then this PR will not be needed anymore.

PR #521, opened 2018-06-08
swss: Dataplane telemetry support in SONiC

shruthi9:
Adding dataplane telemetry configuration support in SONiC

Pending: new VS tests to be added for this feature
A separate PR will be created for new community tests that will added for this feature

Design for the feature is at:
https://github.com/Azure/SONiC/pull/182/


lguohan:
Can you make the AN and FEC as a separate PR?

kuomarcel:
Look forward for this feature to be in SONiC. We have many customer waiting for this feature to validate their SONiC plan.

orchagent/portsorch.cpp
```diff
@@ -1031,6 +1077,9 @@ bool PortsOrch::initPort(const string &alias, const set<int> &lane_set)
 
                 m_flexCounterTable->set(key, fields);
 
+    		PortUpdate update = {p, true };
+    		notify(SUBJECT_TYPE_PORT_CHANGE, static_cast<void *>(&update));
```
lguohan: what is the purpose to notify the port change here?

shruthi9: INT sink port configuration and queue report configuration depend on port add/remove events. Dtelorch agent receives the configuration for sink ports and queue reports and keeps a copy of it locally. When the corresponding port is added/removed, the config is pushed/torn down.

orchagent/dtelorch.cpp
```diff
@@ -0,0 +1,1384 @@
...
+
+    if (type != SUBJECT_TYPE_PORT_CHANGE)
+    {
+        return;
+    }
+
+    PortUpdate *update = static_cast<PortUpdate *>(cntx);
+
+    /* Check if sink ports need to be updated */
+    if (update->add)
```
dproc: Hi,
Is this intended to handle the dynamic port breakup case. For example Ethernet128 is configured in the sink port link but not created at startup. Later when the port is created you want to download it to sink port link in hardware. If so I think the logic here doesn't work. From DTelOrch::doDtelTableTask, I see that if port configured in the sink port link doesn't exist, it is not event add into the cache, as a result here the update will not update the newly created port to hardware.
If you are not handling the dynamic port breakup, then all the ports are created at startup, then you just need to wait for the initdone event in APPDB before processing any message, just like what AclOrch::doTask does. In this way you don't need to handle the port update event. 
In short, if here the logic is for handling dynamic port breakup(I am not sure it's already implementd or will be there in future sonic release), then the logic is flawy. If here the log is not for dynamic port breakup, then the logic here seems needless.

orchagent/dtelorch.cpp
```diff
@@ -0,0 +1,1384 @@
...
+    }
+
+    dTelPortQueueTable_t qTable = port_entry_iter->second.queueTable;
+
+    for (auto it = qTable.begin(); it != qTable.end(); it++ )
+    {
+        DTelQueueReportEntry qreport = it->second;
+        if (update->add)
+        {
+            if (qreport.queueReportOid != 0)
```
dproc: Hi,
Besides the port creation/deletion, does the queue can be created/deleted dynamically(I am not sure about this, I appreciate if you can help provide the answer)? If so do we also need to update the queue report table according to the queue creation/deletion?

orchagent/dtelorch.cpp
```diff
@@ -0,0 +1,1384 @@
...
+	    /* If report session already exists, delete it first */
+	    if (reportSessionExists(report_session_id))
+	    {
+	        if (!deleteReportSession(report_session_id))
+                {
+                    goto report_session_table_continue;
+                }
+	    }
+
+            for (auto i : kfvFieldsValues(t))
```
dproc: minor issue. The indentation here doesn't look consistent

orchagent/dtelorch.cpp
```diff
@@ -0,0 +1,1384 @@
...
+            {
+                if (fvField(i) == REPORT_TAIL_DROP)
+                {
+                    qr_attr.id = SAI_DTEL_QUEUE_REPORT_ATTR_TAIL_DROP;
+                    qr_attr.value.booldata = (fvValue(i) == ENABLED) ? true : false;
+                    qreport->queue_report_attr.push_back(qr_attr);
+                }
+                else if (fvField(i) == QUEUE_DEPTH_THRESHOLD)
+                {
+                    qr_attr.id = SAI_DTEL_QUEUE_REPORT_ATTR_DEPTH_THRESHOLD;
```
dproc: Hi
What are the units of qdepth and latency

kuomarcel:
Please merge this feature soon. We look forward to run this with customer as many are expecting.

orchagent/dtelorch.cpp
shruthi9: qdepth is expressed in number of cells and qlatency is in nanoseconds.

orchagent/dtelorch.cpp
shruthi9: As far as I see in swss code, only port creation/deletion results in new OID allocation for the respective queues

orchagent/dtelorch.cpp
shruthi9: The case we are handling here is the dynamic port breakup case. I have modified the code to not return an error is the port is not present when the sink ports are configured. The configuration is cached and later played to syncd when the port is added.

orchagent/aclorch.cpp
```diff
@@ -1240,6 +1303,242 @@ AclRuleCounters AclRuleMirror::getCounters()
...
+        {
+            SWSS_LOG_ERROR("Could not decrement INT session %s reference count", m_intSessionId.c_str());
+            return false;
+        }
+    }
+
+    return true;
+}
+
+void AclRuleDTelFlowWatchList::update(SubjectType type, void *cntx)
```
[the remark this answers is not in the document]
shruthi9: I have fixed the notification mechanism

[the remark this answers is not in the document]
shruthi9: I have fixed the notification mechanism. Now, if the INT session is not present, ACL rule is not configured in AsicDB, but the config is cached in the agent. Once the INT session is created and notification is received, the cached config is played to sairedis.

orchagent/aclorch.h
```diff
@@ -236,7 +261,34 @@ class AclRuleMirror: public AclRule
     MirrorOrch *m_pMirrorOrch;
 };
 
-class AclTable {
+class AclRuleDTelFlowWatchList: public AclRule
```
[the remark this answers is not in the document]
shruthi9: Thanks for taking the time to review. I have fixed the names.

orchagent/dtelorch.cpp
shruthi9: Fixed

lguohan:
please rebase

orchagent/aclorch.h
```diff
@@ -236,7 +261,36 @@ class AclRuleMirror: public AclRule
...
+    AclRuleDTelDropWatchListEntry(AclOrch *m_pAclOrch, DTelOrch *m_pDTelOrch, string rule, string table, acl_table_type_t type);
+    bool validateAddAction(string attr_name, string attr_value);
+    bool validate();
+    void update(SubjectType, void *);
+
+protected:
+    DTelOrch *m_pDTelOrch;
+};
+
+struct AclTable {
```
lguohan: why change AclTable from class to struct?

orchagent/aclorch.cpp
```diff
@@ -465,20 +505,25 @@ shared_ptr<AclRule> AclRule::makeShared(acl_table_type_t type, AclOrch *acl, Mir
...
+            attr_name == ACTION_DTEL_TAIL_DROP_REPORT_ENABLE || 
+            attr_name == ACTION_DTEL_FLOW_SAMPLE_PERCENT ||
+            attr_name == ACTION_DTEL_REPORT_ALL_PACKETS)
         {
             action_found = true;
             action = attr_name;
             break;
         }
-    }
+    }
```
lguohan: please remove all training whitespaces

orchagent/aclorch.cpp
```diff
@@ -1257,7 +1586,7 @@ AclRange *AclRange::create(sai_acl_range_type_t type, int min, int max)
 
         // work around to avoid syncd termination on SAI error due to max count of ranges reached
         // can be removed when syncd start passing errors to the SAI callers
-        char *platform = getenv("platform");
+        char *platform = getenv("onie_platform");
```
lguohan: this needs to be kept as platform, we need to support non onie-base platform.

tests/test_dtel.py
```diff
@@ -0,0 +1,289 @@
+from swsscommon import swsscommon
+import time
+import re
+import json
+
+class TestDtel(object):
```
lguohan: can you include a test case that test port breakout case?

tests/test_dtel.py
```diff
@@ -0,0 +1,289 @@
...
+        atbl = swsscommon.Table(adb, "ASIC_STATE:SAI_OBJECT_TYPE_DTEL")
+        keys = atbl.getKeys()
+        
+        for k in keys:
+            (status, fvs) = atbl.get(k)
+            assert status == True
+            
+            for fv in fvs:
+            if fv[0] == "SAI_DTEL_ATTR_SWITCH_ID":
+            assert fv[1] == "1"
```
lguohan: the syntax here looks strange to me, can you check?

tests/test_dtel.py
```diff
@@ -0,0 +1,289 @@
...
+        fvs = swsscommon.FieldValuePairs([("POSTCARD", "TRUE")])
+        tbl.set("POSTCARD", fvs)
+
+        fvs = swsscommon.FieldValuePairs([("DROP_REPORT", "TRUE")])
+        tbl.set("DROP_REPORT", fvs)
+
+        fvs = swsscommon.FieldValuePairs([("QUEUE_REPORT", "TRUE")])
+        tbl.set("QUEUE_REPORT", fvs)
+
+        fvs = swsscommon.FieldValuePairs([("Ethernet0", "Ethernet0"), ("Ethernet4", "Ethernet4")])
```
lguohan: field name is strange, both field and value are port name?

orchagent/dtelorch.cpp
lguohan: I agree with dproc the update->add logic is flawed. sink port list should be explicitly specified by user, the orchagent cannot arbitrarily add any breakout port into sink port list.

orchagent/dtelorch.cpp
lguohan: I doubt the assumption hold here. whether or not queue id changes during port creation and deletion depends on SAI implementation. I doubt SAI has this assumption. To be on the safe side, once port is created, the queue id should be re-queried and added.

similiar to above port update logic, whether a queue should be reported or not should come from user input. My assumption here is that if a port is newly create, the port should come up clean, meaning has only default configuration, in this case sink port, queue report should be explicitly added by user. We should not implicitly add them to sink port or queue report.

orchagent/dtelorch.cpp
```diff
@@ -0,0 +1,1407 @@
...
+                    event_attr.push_back(e_attr);
+                }
+            }
+
+            status = sai_dtel_api->create_dtel_event(&event_oid, 
+                gSwitchId, (uint32_t)event_attr.size(), event_attr.data());
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("DTEL ERROR: Failed to create event %s", event.c_str());
+        goto event_table_continue;
```
lguohan: alignment here

orchagent/dtelorch.cpp
shruthi9: Sink port is explicitly specified by the user and not arbitrarily added on update->add. We check to see if a prior configuration exists for the newly added port and only then push the config down.

tests/test_dtel.py
shruthi9: Converting tabs to spaces messed this up. Will fix it.

tests/test_dtel.py
shruthi9: Since we need to take multiple values and we can only use hash maps as value for key, we are using this approach to take in the configuration. SINK_PORT_LIST is the key and all the sink ports are specified in the hmap as value

tests/test_dtel.py
shruthi9: This would require some time to get port breakout tested on VS platform. I did not see an existing VS test for port breakout. Can we add this test after upstreaming the current code?

orchagent/dtelorch.cpp
shruthi9: Will fix it

orchagent/aclorch.cpp
shruthi9: Will fix it

orchagent/aclorch.h
shruthi9: Fixed

orchagent/aclorch.cpp
shruthi9: Fixed

orchagent/dtelorch.cpp
lguohan: maybe that is true for your implementation, but certainly it is not the assumption for SAI. The object id is opaque to the user, each SAI implementation can decide how to assign object id as long as it is unique.

shruthi9: We are not exposing the queue OIDs to the users. Users configure using relative queue numbers for the port: HMSET DTEL_QUEUE_REPORT|Ethernet8|0 ...
Here 0 is the first queue for Ethernet8. DtelOrch agent will query portsorch agent to get the actual OID for this queue and then configure the queue report.

lguohan:
retest this please

tests/test_dtel.py
lguohan: ok

lguohan:
please fix the vs test, it is broken.

shruthi9:
Fixed the VS test

orchagent/dtelorch.cpp
lguohan: this is an interesting design. one can argue that if we user add sink ports to config db, all sink ports (physical ports) should exist in the system. If the sink port does not corresponds to a physical port, the configuration should reject that sink port. basically, you cannot configure a sink port without mapping it to a physical port. In this case, whenever you do break out new port will be created and those new ports will not in the sink port list and you won't have above logic. Then after break out, user can then explicitly add the physical port to the sink port list.

shruthi9: That was exactly how we had designed this originally. Ports should be created before sink port config. But, as part of design review there were several comments to handle the dynamic port OID change. So, we added this logic. This, in any case, future-proofs the code.

orchagent/orchdaemon.cpp
```diff
@@ -108,9 +108,54 @@ bool OrchDaemon::init()
...
+        CFG_DTEL_REPORT_SESSION_TABLE_NAME,
+        CFG_DTEL_INT_SESSION_TABLE_NAME,
+        CFG_DTEL_QUEUE_REPORT_TABLE_NAME,
+        CFG_DTEL_EVENT_TABLE_NAME
+    };
+
+    m_orchList = { switch_orch, gCrmOrch, gPortsOrch, intfs_orch, gNeighOrch, gRouteOrch, copp_orch, tunnel_decap_orch, qos_orch, gBufferOrch, mirror_orch };
+
+    bool initialize_dtel = false;
+    if (platform == BFN_PLATFORM_SUBSTRING)
```
lguohan: in vs env, the platform is not BFN, then initialize_dtel is false and dtel_orch is not initialized, how can dtel vs test passing? Where do I miss here?

tests/test_dtel.py
```diff
@@ -0,0 +1,289 @@
...
+
+        fvs = swsscommon.FieldValuePairs([("INT_L4_DSCP_VALUE", "128"), ("INT_L4_DSCP_MASK", "255")])  
+        tbl.set("INT_L4_DSCP", fvs)
+
+        time.sleep(1)
+
+        atbl = swsscommon.Table(adb, "ASIC_STATE:SAI_OBJECT_TYPE_DTEL")
+        keys = atbl.getKeys()
+        
+        for k in keys:
```
lguohan: i see, here the keys is null then, no assert is checked, that is why it is passing.

shruthi9: Yes

lguohan: need to assert len(keys) > 1

shruthi9: The check would fail if platform is not barefoot. Is there a "vs platform" check that we can add to orchdaemon.cpp to initialize dtel, in addition to barefoot?

orchagent/orchdaemon.cpp
lguohan: I will add platform vs for this env, and you can check if platform == "vs" or platform == BFN_PLATFORM_SUBSTRING

orchagent/orch.h
```diff
@@ -33,6 +33,7 @@ const char config_db_key_delimiter = '|';
 
 #define MLNX_PLATFORM_SUBSTRING "mellanox"
 #define BRCM_PLATFORM_SUBSTRING "broadcom"
+#define BFN_PLATFORM_SUBSTRING  "barefoot"
```
lguohan: add BFN_PLATFORM_SUBSTRING "vs" here.

I will export the env in the vs.

shruthi9: Will do. Thanks!

lguohan: check here. 

https://github.com/Azure/sonic-buildimage/pull/1860

lguohan: I mean #define VS_PLATFORM_SUBSTRING "vs"

tests/test_dtel.py
```diff
@@ -0,0 +1,289 @@
...
+                                          ("VRF", "default"), 
+                                          ("TRUNCATE_SIZE", "256"), 
+                                          ("UDP_DEST_PORT", "2000")])  
+        tbl.set("RS-1", fvs)
+
+        time.sleep(1)
+
+        atbl = swsscommon.Table(adb, "ASIC_STATE:SAI_OBJECT_TYPE_DTEL_REPORT_SESSION")
+        keys = atbl.getKeys()
+
```
lguohan: check keys is not empty.

tests/test_dtel.py
```diff
@@ -0,0 +1,289 @@
...
+                                          ("COLLECT_SWITCH_PORTS", "TRUE"), 
+                                          ("COLLECT_QUEUE_INFO", "TRUE")])  
+        tbl.set("INT-1", fvs)
+
+        time.sleep(1)
+
+        atbl = swsscommon.Table(adb, "ASIC_STATE:SAI_OBJECT_TYPE_DTEL_INT_SESSION")
+        keys = atbl.getKeys()
+        
+        for k in keys:
```
lguohan: check keys is not empty.

tests/test_dtel.py
```diff
@@ -0,0 +1,289 @@
...
+                                          ("THRESHOLD_BREACH_QUOTA", "3000"), 
+                                          ("REPORT_TAIL_DROP", "TRUE")])  
+        tbl.set("Ethernet0|0", fvs)
+
+        time.sleep(1)
+
+        atbl = swsscommon.Table(adb, "ASIC_STATE:SAI_OBJECT_TYPE_DTEL_QUEUE_REPORT")
+        keys = atbl.getKeys()
+        
+        for k in keys:
```
lguohan: check keys is not empty.

tests/test_dtel.py
```diff
@@ -0,0 +1,289 @@
...
+        fvs = swsscommon.FieldValuePairs([("EVENT_REPORT_SESSION", "RS-1"), 
+                                          ("EVENT_DSCP_VALUE", "60")])  
+        tbl.set("EVENT_TYPE_DROP_REPORT", fvs)
+
+        time.sleep(1)
+
+        atbl = swsscommon.Table(adb, "ASIC_STATE:SAI_OBJECT_TYPE_DTEL_EVENT")
+        keys = atbl.getKeys()
+        
+        for k in keys:
```
lguohan: check keys is not empty.

orchagent/orch.h
shruthi9: Fixed

shruthi9:
Enabled DTel for VS platform and fixed the test

orchagent/dtelorch.cpp
lguohan: it is arguable since if the physical port is never created, then we have mis-match between the config db and the actual asic configuation. I'll defer this to future discussion. For now, we can keep as it is.

lguohan:
retest this please

shruthi9:
Looks like platform env variable is not set to "vs" and dtel test fails due to that in the newly added assert. sonic-buildimage seems to be at commit daf590e6777b81fc06abd19baa10f2b52670267e. How do we fix this to pick the latest?

lguohan:
retest this please

shruthi9:
buildimage-vs-all build has been failing last couple of times. VS tests here are not passing because the requisite change in buildimage is not present

lguohan:
retest this please

lguohan:
retest this please

lguohan:
platform vs is added. the dtel orch tests are passing, but it looks like it is causing mirror test to fail. 

it looks like it is causing two tests to fail.
```
test_mirror.py::TestMirror::test_AclRuleDscpWithoutMask FAILED           [ 70%]
test_mirror.py::TestMirror::test_AclRuleDscpWithMask FAILED              [ 72%]
```

@stcheng , can you also take a look?

lguohan:
thanks

tests/test_mirror.py
```diff
@@ -13,9 +13,19 @@ def get_acl_table_id(self, dvs):
         assert dvs.asicdb.default_acl_table in keys
 
         acl_tables = [k for k in keys if k not in dvs.asicdb.default_acl_table]
-        assert len(acl_tables) == 1
+        assert len(acl_tables) >= 1
```
stcheng: what is the exact number of the ACL tables?

tests/test_mirror.py
```diff
@@ -13,9 +13,19 @@ def get_acl_table_id(self, dvs):
...
 
-        return acl_tables[0]
+        # Filter out DTel Acl tables
```
stcheng: If this ACL table is there by default, you could add the table information in the `conftest.py` file.

tests/test_mirror.py
shruthi9: DTel adds 2 additional ACL tables in addition to the default one

PR #523, opened 2018-06-11
Allocate buffer for 256 entries for ACL resource query

prsunny:
**What I did**
Allocate an initial buffer instead of NULL during ACL table/group query 

**Why I did it**
During every CRM polling interval, the following `WARNING `message is observed. This fix will suppress the frequent WARNINGs 

`Jun 11 20:23:07.092436 str-a7060cx-acs-1 WARNING syncd: :- processEvent: get API for key: SAI_OBJECT_TYPE_SWITCH:oid:0x21000000000000 op: get returned status: SAI_STATUS_BUFFER_OVERFLOW`

**How I verified it**
Load the changes and verify for WARNING message.
RUN CRM test and also the following commands:

```
admin@str-a7060cx-acs-1:~$ crm show resources acl table
Table ID         Resource Name      Used Count    Available Count
---------------  ---------------  ------------  -----------------
0x700000000093c  acl_entry                   2               1278

admin@str-a7060cx-acs-1:~$ crm show resources acl group
Stage    Bind Point    Resource Name      Used Count    Available Count
-------  ------------  ---------------  ------------  -----------------
INGRESS  PORT          acl_group                  56                968
INGRESS  PORT          acl_table                   1                  2
INGRESS  LAG           acl_group                   0                968
INGRESS  LAG           acl_table                   0                  4
...
```

**Details if related**

### DIFF
--- a/orchagent/crmorch.cpp
+++ b/orchagent/crmorch.cpp
@@ -436,20 +436,19 @@ void CrmOrch::getResAvailableCounters()
             case SAI_SWITCH_ATTR_AVAILABLE_ACL_TABLE:
             case SAI_SWITCH_ATTR_AVAILABLE_ACL_TABLE_GROUP:
             {
-                attr.value.aclresource.count = 0;
-                attr.value.aclresource.list = NULL;
+                const int aclResourceCount = 256;
+                vector<sai_acl_resource_t> resources(aclResourceCount);
 
+                attr.value.aclresource.count = aclResourceCount;
+                attr.value.aclresource.list = resources.data();
                 sai_status_t status = sai_switch_api->get_switch_attribute(gSwitchId, 1, &attr);
-                if ((status != SAI_STATUS_SUCCESS) && (status != SAI_STATUS_BUFFER_OVERFLOW))
+                if (status == SAI_STATUS_BUFFER_OVERFLOW)
                 {
-                    SWSS_LOG_ERROR("Failed to get switch attribute %u , rv:%d", attr.id, status);
-                    break;
+                    resources.resize(attr.value.aclresource.count);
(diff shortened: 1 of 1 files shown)


orchagent/crmorch.cpp
```diff
@@ -436,20 +436,19 @@ void CrmOrch::getResAvailableCounters()
...
 
+                attr.value.aclresource.count = aclResourceCount;
+                attr.value.aclresource.list = resources.data();
                 sai_status_t status = sai_switch_api->get_switch_attribute(gSwitchId, 1, &attr);
-                if ((status != SAI_STATUS_SUCCESS) && (status != SAI_STATUS_BUFFER_OVERFLOW))
+                if (status == SAI_STATUS_BUFFER_OVERFLOW)
                 {
-                    SWSS_LOG_ERROR("Failed to get switch attribute %u , rv:%d", attr.id, status);
-                    break;
+                    resources.resize(attr.value.aclresource.count);
```
jleveque: `attr.value.aclresource.count` is set to `aclResourceCount` on line 442, which is the same value used when initializing `resources` on line 440. It seems to me this resize will have no effect.

orchagent/crmorch.cpp
```diff
@@ -436,20 +436,19 @@ void CrmOrch::getResAvailableCounters()
             case SAI_SWITCH_ATTR_AVAILABLE_ACL_TABLE:
             case SAI_SWITCH_ATTR_AVAILABLE_ACL_TABLE_GROUP:
             {
-                attr.value.aclresource.count = 0;
-                attr.value.aclresource.list = NULL;
+                const int aclResourceCount = 256;
```
zhenggen-xu: Can we define this number as a macro?

PR #525, opened 2018-06-22
[fdborch]: Fix 'double free' bug in fdb event handler

pavel-shirshov:
**What I did**
Fix a bug which could cause to free the same memory twice.

**Why I did it**
To fix a segmentation fault in case we got 'count' variable greater than 1.

**How I verified it**
Build and run on DUT.

**Details if related**


pavel-shirshov:
Thanks to @volodymyrsamotiy and @stepanblyschak for the report.

PR #527, opened 2018-06-25
Postpone QueueMap initialization until activation of the counters

pavel-shirshov:
**What I did**
I postponed generation of the QueueMaps until the counters are activated. 

**Why I did it**
We don't need QueueMaps until the counters are active

**How I verified it**
Build an image and run on DUT

**Details if related**

### DIFF
--- a/orchagent/flexcounterorch.cpp
+++ b/orchagent/flexcounterorch.cpp
@@ -70,6 +70,10 @@ void FlexCounterOrch::doTask(Consumer &consumer)
                 }
                 else if(field == FLEX_COUNTER_STATUS_FIELD)
                 {
+                    // Currently the counters are disabled by default
+                    // The queue maps will be generated as soon as counters are enabled
+                    gPortsOrch->generateQueueMap();
+
                     vector<FieldValueTuple> fieldValues;
                     fieldValues.emplace_back(FLEX_COUNTER_STATUS_FIELD, value);
                     m_flexCounterGroupTable->set(flexCounterGroupMap[key], fieldValues);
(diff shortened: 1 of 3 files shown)


orchagent/flexcounterorch.cpp
```diff
@@ -70,6 +70,10 @@ void FlexCounterOrch::doTask(Consumer &consumer)
                 }
                 else if(field == FLEX_COUNTER_STATUS_FIELD)
                 {
+                    // Currently the counters are disabled by default
+                    // The queue maps will be generated as soon as counters are enabled
+                    gPortsOrch->generateQueueMap();
```
sihuihan88: Is the map generated after queue counters are enabled or all the counters are enabled?

orchagent/portsorch.cpp
```diff
@@ -1817,61 +1817,6 @@ void PortsOrch::initializeQueues(Port &port)
     }
 
     SWSS_LOG_INFO("Get queues for port %s", port.m_alias.c_str());
-
-    /* Create the Queue map in the Counter DB */
-    /* Add stat counters to flex_counter */
```
lguohan: do we need queue map for applying qos configuration?

orchagent/flexcounterorch.cpp
pavel-shirshov: The map is generated after activation of the flex counters. As I understand the maps are related to port queues. So they will be used with queue counters.

orchagent/portsorch.cpp
pavel-shirshov: No. We use only Port::m_queue_ids vector. We use the maps only:
1. swss: pfc handler
2. swss: countercheckorch
3. queuestat
4. snmp

PR #530, opened 2018-07-02
Check m_port_ready_list_ref value for presense before using the value

pavel-shirshov:
**What I did**
Check that m_port_ready_list_ref item is exist before using it.

**Why I did it**
Otherwise it could be a race condition if portsorch receive events before we initialize bufferorch

**How I verified it**
Build an image and run on DUT.

**Details if related**


pavel-shirshov:
retest this please

pavel-shirshov:
retest this please

PR #535, opened 2018-07-05
Create empty ready lists for buffers by default

pavel-shirshov:
**What I did**
Add additional logic to initialize ready list as empty by default.

**Why I did it**
Otherwise the port would be stopped if it doesn't have any buffer configuration

**How I verified it**
Build an image and run on your DUT

**Details if related**

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -58,9 +58,20 @@ void BufferOrch::initBufferReadyList(Table& table)
 {
     SWSS_LOG_ENTER();
 
+    // init all ports with an empty list
+    for (const auto& it: gPortsOrch->getAllPorts())
+    {
+        if (it.second.m_type == Port::PHY)
+        {
+            const auto& port_name = it.first;
+            m_port_ready_list_ref[port_name] = {};
+        }
+    }
+
     std::vector<std::string> keys;
     table.getKeys(keys);
 
+    // populate the lists with buffer configuration information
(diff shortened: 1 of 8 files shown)


lguohan:
can you explain why test_port_buffer_rel.py pass in the first place?

lguohan:
do not worry about broadcom and mellanox build failure, it is because sai package header mismatch for now.

orchagent/bufferorch.cpp
```diff
@@ -58,9 +58,20 @@ void BufferOrch::initBufferReadyList(Table& table)
 {
     SWSS_LOG_ENTER();
 
+    // init all ports with an empty list
+    for (const auto& it: gPortsOrch->getAllPorts())
```
qiluo-msft: I think gPortsOrch->getAllPorts() will return only one port, the CPU port. Is it what you expected?

orchagent/bufferorch.cpp
```diff
@@ -58,9 +58,20 @@ void BufferOrch::initBufferReadyList(Table& table)
 {
     SWSS_LOG_ENTER();
```
qiluo-msft: This function will be called twice, so m_port_ready_list_ref will be overwritten again. Is it a bug?

qiluo-msft: I try to fix this in https://github.com/Azure/sonic-swss/pull/590, please help review.

PR #538, opened 2018-07-18
VS: ProducerStateTable should not be used to set stateDb table

jipanyang:

**What I did**
Use Table instead of ProducerStateTable for setting stateDB table.    

**Why I did it**
The consumers of stateDB lagTable is SubscriberStateTable,  they don't clear "LAG_TABLE_KEY_SET"

**How I verified it**

VS test.

**Details if related**


pavel-shirshov:
retest this please

PR #541, opened 2018-07-20
[test]: Fix test_mirror with correct number of ACL tables

stcheng:
After creation of the ACL table, check both the number of ACL tables
and the number of ACL table groups.

After removal of the ACL table, correct the number of remaining
ACL tables.

Right now, after the ACL table is removed, the ACL table groups are
not removed correspondingly. This should be fixed in the future.


stcheng:
will close this pull request, since https://github.com/Azure/sonic-swss/pull/521 has changed part of the code here. i'll reevaluate all the changes.

PR #545, opened 2018-07-25
[portmgrd]: Add portmgrd to monitor port MTU configurations

stcheng:
In order to get rid of /etc/network/interfaces file and move
all the configurations to the configuration database, MTU configurations
are required to be put into the database and monitored by a specific
daemon. This daemon is right now to achieve this goal.

Currently, this daemon will only listen to the port MTU configurations
if existed in the database and then call the 'ip link' command to
configure the kernel netdevs. Then, portsyncd will catch the messages
via libnl and send the message to orchagent.

### DIFF
--- a/cfgmgr/Makefile.am
+++ b/cfgmgr/Makefile.am
@@ -1,7 +1,7 @@
 INCLUDES = -I $(top_srcdir) -I $(top_srcdir)/orchagent
 CFLAGS_SAI = -I /usr/include/sai
 
-bin_PROGRAMS = vlanmgrd intfmgrd buffermgrd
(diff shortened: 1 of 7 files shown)


stcheng:
@jipanyang thanks for pointing this out. I have removed this part of the code. In the next stage, all the configuration part of the code will be moved to portmgrd and then we could finally remove the dependency of libnl.

stcheng:
@jipanyang Yes. I'll update this pull request to handle this case.

stcheng:
update this pull request to support admin status configuration as well.

PR #546, opened 2018-07-25
Warm reboot: Extract the addToSync() function from Consumer::execute(), and new Or…

jipanyang:
…ch::addExistingData()

**What I did**
Add support to explicitly read existing data from a table and add them to the m_toSync of consumer in this Orch

**Why I did it**
Needed for state restore of orchagent during warm restart.

**How I verified it**
VS testing.

**Details if related**


orchagent/orch.cpp.orig
```diff
@@ -0,0 +1,495 @@
+#include <fstream>
```
prsunny: Why this file? orch.cpp.orig?

jipanyang:
Most of the changes have been covered in https://github.com/Azure/sonic-swss/pull/551,   the remaining part will be incorporated to https://github.com/Azure/sonic-swss/pull/554.

Close it for now.

PR #547, opened 2018-07-25
Add common warm start functions to be used by all SWSS processes

jipanyang:

**What I did**
Add common support functions for SWSS warm restart check, restart count  and state log.
The functions are supposed to be used by all processes in SWSS docker.
 
**Why I did it**
support functions for warm restart feature.

**How I verified it**

admin@sonic:~$ redis-cli
127.0.0.1:6379> keys WAR*
1) "WARM_RESTART_TABLE:neighsyncd"
2) "WARM_RESTART_TABLE:orchagent"
3) "WARM_RESTART_TABLE:vlanmgrd"
4) "WARM_RESTART_TABLE:portsyncd"

127.0.0.1:6379> hgetall  "WARM_RESTART_TABLE:orchagent"
1) "restart_count"
2) "8"
3) "state_restored"
4) "true"
127.0.0.1:6379> hgetall  "WARM_RESTART_TABLE:portsyncd"
1) "restart_count"
2) "8"

**Details if related**


warmrestart/warm_restart.cpp
```diff
@@ -0,0 +1,105 @@
...
+    warmStart.m_stateRedisClient = std::make_shared<swss::RedisClient>(warmStart.m_stateDb.get());
+
+    warmStart.m_cfgDb          = std::make_shared<swss::DBConnector>(CONFIG_DB, swss::DBConnector::DEFAULT_UNIXSOCKET, 0);
+    warmStart.m_cfgRedisClient = std::make_shared<swss::RedisClient>(warmStart.m_cfgDb.get());
+
+    bool system_warm_start = false;
+    warmStart.enabled = false;
+
+    // Check system level warm restart config first
+    auto pvalue = warmStart.m_cfgRedisClient->hget(CFG_WARM_RESTART_TABLE_NAME"|system", "enable");
```
lguohan: | should not be hard coded.

lguohan: use table api?

jipanyang: Probably table should be expanded with member functions like getEntry(const std::string &key, const std::string &field) and setEntry(const std::string &key, const std::string &field, const std::string &value)

I switched to using RedisClient to avoid the clumsy key level only operations.

Will give it a try.

PR #551, opened 2018-07-27
Warm reboot for PortsOrch

qiluo-msft:
The non-warm reboot behavior is backward compatible, and tested in lab.

The idea is best effort warm reboot based on left over entries in PORT_TABLE.
1. During a cold reboot, the whole table is empty, so keep original behavior
2. During a warm reboot, the previous port entries are left in the table, together with event entries such as PortConfigDone, PortInitDone. So PortsOrch will add them into PortsOrch's consumer queue to propagate to downstream syncd and ASIC.
3. During a warm reboot, if any corruption found in the table, clean up the table and fallback to cold reboot.

The warm reboot is not end-to-end tested.

### DIFF
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -141,25 +141,21 @@ void Consumer::drain()
 }
 
 // TODO: Table should be const
-void Orch::addExistingData(Table *table)
+bool Orch::addExistingData(Table *table)
 {
     string tableName = table->getTableName();
-    Consumer* consumer;
-    auto it = m_consumerMap.begin();
-
-    while (it != m_consumerMap.end())
+    auto found = m_consumerMap.find(tableName);
+    if (found == m_consumerMap.end())
     {
-        consumer = (Consumer*)(it->second.get());
-        if (tableName == consumer->getTableName())
-        {
-            break;
-        }
-        it++;
+        SWSS_LOG_ERROR("No table %s in Orch", tableName.c_str());
+        return false;
     }
 
-    if (it == m_consumerMap.end())
+    Consumer* consumer = dynamic_cast<Consumer *>(found->second.get());
+    if (consumer == NULL)
     {
-        return;
+        SWSS_LOG_ERROR("Executor is not a Consumer: %s", tableName.c_str());
+        return false;
     }
 
     std::deque<KeyOpFieldsValuesTuple> entries;
@@ -180,6 +176,7 @@ void Orch::addExistingData(Table *table)
     }
 
     consumer->addToSync(entries);
+    return true;
 }
 
 /*
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -66,12 +66,9 @@ vector<Selectable *> Orch::getSelectables()
     return selectables;
 }
 
-void Consumer::execute()
+void Consumer::addToSync(std::deque<KeyOpFieldsValuesTuple> &entries)
 {
     SWSS_LOG_ENTER();
-
-    std::deque<KeyOpFieldsValuesTuple> entries;
-    getConsumerTable()->pops(entries);
 
     /* Nothing popped */
     if (entries.empty())
@@ -123,6 +120,16 @@ void Consumer::execute()
             m_toSync[key] = KeyOpFieldsValuesTuple(key, op, existing_values);
         }
     }
+}
+
+void Consumer::execute()
+{
+    SWSS_LOG_ENTER();
+
+    std::deque<KeyOpFieldsValuesTuple> entries;
+    getConsumerTable()->pops(entries);
+
+    addToSync(entries);
 
     drain();
 }
@@ -131,6 +138,48 @@ void Consumer::drain()
 {
     if (!m_toSync.empty())
         m_orch->doTask(*this);
+}
+
+// TODO: Table should be const
+void Orch::addExistingData(Table *table)
+{
+    string tableName = table->getTableName();
+    Consumer* consumer;
+    auto it = m_consumerMap.begin();
+
+    while (it != m_consumerMap.end())
+    {
+        consumer = (Consumer*)(it->second.get());
+        if (tableName == consumer->getTableName())
+        {
+            break;
+        }
+        it++;
+    }
+
+    if (it == m_consumerMap.end())
+    {
+        return;
+    }
+
+    std::deque<KeyOpFieldsValuesTuple> entries;
+    vector<string> keys;
+    table->getKeys(keys);
+    for (const auto &key: keys)
+    {
+        KeyOpFieldsValuesTuple kco;
+
+        kfvKey(kco) = key;
+        kfvOp(kco) = SET_COMMAND;
+
+        if (!table->get(key, kfvFieldsValues(kco)))
+        {
+            continue;
+        }
+        entries.push_back(kco);
+    }
+
+    consumer->addToSync(entries);
 }
 
 /*
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -122,6 +122,37 @@ void Consumer::addToSync(std::deque<KeyOpFieldsValuesTuple> &entries)
     }
 }
 
+// TODO: Table should be const
+void Consumer::refillToSync(Table* table)
+{
+    std::deque<KeyOpFieldsValuesTuple> entries;
+    vector<string> keys;
+    table->getKeys(keys);
+    for (const auto &key: keys)
+    {
+        KeyOpFieldsValuesTuple kco;
+
+        kfvKey(kco) = key;
+        kfvOp(kco) = SET_COMMAND;
+
+        if (!table->get(key, kfvFieldsValues(kco)))
+        {
+            continue;
+        }
+        entries.push_back(kco);
+    }
+
+    addToSync(entries);
+}
+
+void Consumer::refillToSync()
+{
+    auto db = getConsumerTable()->getDbConnector();
+    string tableName = getConsumerTable()->getTableName();
+    auto table = Table(db, tableName);
+    refillToSync(&table);
+}
+
 void Consumer::execute()
 {
     SWSS_LOG_ENTER();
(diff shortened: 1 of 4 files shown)


lguohan:
please fix vs test failure.

lguohan:
can you explain the idea in the commit message, it will be better for us to understand and do the review

portsyncd/portsyncd.cpp
```diff
@@ -69,7 +71,7 @@ int main(int argc, char **argv)
     DBConnector cfgDb(CONFIG_DB, DBConnector::DEFAULT_UNIXSOCKET, 0);
     DBConnector appl_db(APPL_DB, DBConnector::DEFAULT_UNIXSOCKET, 0);
     DBConnector state_db(STATE_DB, DBConnector::DEFAULT_UNIXSOCKET, 0);
-    ProducerStateTable p(&appl_db, APP_PORT_TABLE_NAME);
```
lguohan: why change to producer state table?

qiluo-msft: Actually change to ProducerTable. The entries in APP_PORT_TABLE_NAME are sequence sensitive because there are event entries.


---
In reply to: [205855555](https://github.com/Azure/sonic-swss/pull/551#discussion_r205855555) [](ancestors = 205855555)

lguohan: what event entries?

orchagent/orch.cpp
```diff
@@ -133,6 +140,48 @@ void Consumer::drain()
         m_orch->doTask(*this);
 }
 
+// TODO: Table should be const
+void Orch::addExistingData(Table *table)
```
lguohan: can you add one line description what addExistingData do?

lguohan:
>     		notify(SUBJECT_TYPE_PORT_CHANGE, static_cast<void *>(&update));

change tab to spaces.

---
Refers to: orchagent/portsorch.cpp:1163 in d363262. [](commit_id = d36326228a318da7667e33481414af7e18481fc8, deletion_comment = False)

portsyncd/portsyncd.cpp
jipanyang: Could you do the ProducerStateTable  to ProducerTable  change in a separate PR?  It is not strictly related to warm reboot.

orchagent/orch.cpp
```diff
@@ -133,6 +140,48 @@ void Consumer::drain()
...
+{
+    string tableName = table->getTableName();
+    Consumer* consumer;
+    auto it = m_consumerMap.begin();
+
+    while (it != m_consumerMap.end())
+    {
+        consumer = (Consumer*)(it->second.get());
```
jipanyang: The change overlap with part of #547.  One of the problem here is that other types of class may be in m_consumerMap too and they don't  have getTableName() methods.  casting them to Consumer may cause problem. 
The other reason to add getName()/setName() for Executor is to facilitate debugging.

orchagent/portsorch.cpp
```diff
@@ -1177,6 +1180,64 @@ bool PortsOrch::initPort(const string &alias, const set<int> &lane_set)
...
+    SWSS_LOG_NOTICE("foundPortInitDone = %d", foundPortInitDone);
+
+    vector<string> keys;
+    m_portTable->getKeys(keys);
+    SWSS_LOG_NOTICE("m_portTable->getKeys %zd", keys.size());
+
+    if (!foundPortConfigDone || !foundPortInitDone)
+    {
+        SWSS_LOG_NOTICE("No port table, fallback to cold start");
+        cleanPortTable(keys);
```
jipanyang: There is #547 to have common function for all swss processes to check whether a process is doing warm start . 
We agreed to discuss the schema design.  The basic idea there is to assume that for cold start, the corresponding WARM_RESTART_TABLE entry cleared, then we create the entry and set restart_count to 0.
Every time the process is warm restarted, restart_count is incremented by 1.   

Not sure about the purpose of cleanPortTable() here.

orchagent/orch.cpp
qiluo-msft: Actually it is in the header file.

---
In reply to: [205862889](https://github.com/Azure/sonic-swss/pull/551#discussion_r205862889) [](ancestors = 205862889)

orchagent/portsorch.cpp
qiluo-msft: We could still discuss that design. No confliction here.

>> Not sure about the purpose of cleanPortTable() here.
If the redis leftover data for warm reboot is corrupted, clean up everything left and fallback to cold start.

---
In reply to: [205869453](https://github.com/Azure/sonic-swss/pull/551#discussion_r205869453) [](ancestors = 205869453)

qiluo-msft:
Retest this please

orchagent/orch.h
```diff
@@ -149,6 +150,9 @@ class Orch
 
     vector<Selectable*> getSelectables();
 
+    // add the existing table data (left by warm reboot) to the consumer todo task list.
+    bool addExistingData(Table *table);
```
jipanyang: There are quite a few places to use addExistingData() where only table name is available.   It should be better to use table name directly.

orchagent/portsorch.cpp
```diff
@@ -1177,6 +1180,64 @@ bool PortsOrch::initPort(const string &alias, const set<int> &lane_set)
...
+        return false;
+    }
+
+    doPortConfigDoneTask(tuples);
+    if (m_portCount != keys.size() - 2)
+    {
+        // Invalid port table
+        SWSS_LOG_ERROR("Invalid port table: m_portCount");
```
jipanyang: In what scenario will this case be hit?

orchagent/portsorch.cpp
```diff
@@ -245,6 +245,9 @@ PortsOrch::PortsOrch(DBConnector *db, vector<table_name_with_pri_t> &tableNames)
     m_portStatusNotificationConsumer = new swss::NotificationConsumer(notificationsDb, "NOTIFICATIONS");
     auto portStatusNotificatier = new Notifier(m_portStatusNotificationConsumer, this);
     Orch::addExecutor("PORT_STATUS_NOTIFICATIONS", portStatusNotificatier);
+
+    // Try warm start
+    bake();
```
jipanyang: Within  PortsOrch::initializePort(Port &p),  for warm start,  oper_status should not be changed.

qiluo-msft: I checked code in void PortsOrch::doPortTask(Consumer &consumer).

During warm start, m_portListLaneMap will have all existing ports from SAI query. So there is no new port created, and no new port initialized.

The function PortsOrch::initializePort(Port &p) will not be called.

---
In reply to: [206311005](https://github.com/Azure/sonic-swss/pull/551#discussion_r206311005) [](ancestors = 206311005)

qiluo-msft:
Retest this please

orchagent/portsorch.cpp
jipanyang: Not sure if there is misunderstanding here,  the warm restart testing I have done so far didn't skip  PortsOrch::initializePort(Port &p).    Even for cold start, m_portListLaneMap will have all the exiting physical ports.

qiluo-msft:
Retest this please

orchagent/portsorch.cpp
qiluo-msft: I am talking about this code snippet:
```
for (auto it = m_lanesAliasSpeedMap.begin(); it != m_lanesAliasSpeedMap.end();)
{
    bool port_created = false;

    if (m_portListLaneMap.find(it->first) == m_portListLaneMap.end())
    {
```

---
In reply to: [206363163](https://github.com/Azure/sonic-swss/pull/551#discussion_r206363163) [](ancestors = 206363163)

jipanyang: This is for port breakout.  PortsOrch::initPort() still will be called.

orchagent/portsorch.cpp
jipanyang: If redis database is corrupted, should not a whole system cold boot be performed instead of doing partial fix here and putting the system into unknown state?   Supporting Individual process cold boot alone is another area.

orchagent/portsorch.cpp
qiluo-msft: Thanks! Got your idea, and it is really a bug. I am using another PR to fix it. Please help [https://github.com/Azure/sonic-swss/pull/555](https://github.com/Azure/sonic-swss/pull/555)

---
In reply to: [206644401](https://github.com/Azure/sonic-swss/pull/551#discussion_r206644401) [](ancestors = 206644401)

portsyncd/portsyncd.cpp
qiluo-msft: Let's start another PR for it. Please help [https://github.com/Azure/sonic-swss/pull/556](https://github.com/Azure/sonic-swss/pull/556)

---
In reply to: [205866896](https://github.com/Azure/sonic-swss/pull/551#discussion_r205866896) [](ancestors = 205866896)

lguohan:
retest this please

qiluo-msft:
Retest this please.

orchagent/portsorch.cpp
qiluo-msft: It is another design choice, and I am ok with it.

---
In reply to: [206646838](https://github.com/Azure/sonic-swss/pull/551#discussion_r206646838) [](ancestors = 206646838)

orchagent/orch.cpp
```diff
@@ -123,6 +120,47 @@ void Consumer::execute()
             m_toSync[key] = KeyOpFieldsValuesTuple(key, op, existing_values);
         }
     }
+}
+
+// TODO: Table should be const
+void Consumer::refillToSync(Table* table)
```
jipanyang: It looks to me if we remove "bool Orch::addExistingData(Table *table)" and use bool Orch::addExistingData(const string& tableName) only,  
void Consumer::refillToSync(Table* table) is not needed and quite a few line of redundant code function may be eliminated.

orchagent/portsorch.cpp
```diff
@@ -1177,6 +1180,69 @@ bool PortsOrch::initPort(const string &alias, const set<int> &lane_set)
...
+
+    if (!foundPortConfigDone || !foundPortInitDone)
+    {
+        SWSS_LOG_NOTICE("No port table, fallback to cold start");
+        cleanPortTable(keys);
+        return false;
+    }
+
+    doPortConfigDoneTask(tuples);
+    if (m_portCount != keys.size() - 2)
```
jipanyang: After "PortConfigDone",   for some platform, port may be removed with  removePort(it->second),  could it cause this check invalid in case of warm start?

     if (platform && (strstr(platform, BFN_PLATFORM_SUBSTRING) || strstr(platform, MLNX_PLATFORM_SUBSTRING)))
                        {
                            if (!removePort(it->second))
                            {
                            }
                        }

orchagent/orch.cpp
qiluo-msft:
```
void Consumer::refillToSync(Table* table)
```
is used for optimization if we have a table object already.

---
In reply to: [207438102](https://github.com/Azure/sonic-swss/pull/551#discussion_r207438102) [](ancestors = 207438102)

orchagent/portsorch.cpp
qiluo-msft: removePort() will never del entry in PORT_TABLE. So the answer is no.

---
In reply to: [207439672](https://github.com/Azure/sonic-swss/pull/551#discussion_r207439672) [](ancestors = 207439672)

jipanyang: In your change, doPortConfigDoneTask() is also called at construct phase,   
having m_portConfigDone set prematurely will change the processing flow of void PortsOrch::doPortTask(Consumer &consumer).

```
            /* Once all ports received, go through the each port and perform appropriate actions:
             * 1. Remove ports which don't exist anymore
             * 2. Create new ports
             * 3. Initialize all ports
             */
            if (m_portConfigDone && (m_lanesAliasSpeedMap.size() == m_portCount))     <--- will be false before all of the ports handled.
            {
            }
            
            if (!m_portConfigDone)              <------- m_portConfigDone is true. it continue the flow.
            {
                it = consumer.m_toSync.erase(it);      <----- While if m_portConfigDone is false, the request will be deleted, not good for the one phase warm restore.
                continue;
            }
            
            ........
            Port p;
            if (!getPort(alias, p))
            {
                SWSS_LOG_ERROR("Failed to get port id by alias:%s", alias.c_str());   <--- will hit here, since initPort not called yet.
            }
```
            
Why do you want to mess with the port init orders?  We might hit more subtle sequence problems.             

This also reminds me that the previous switching from ProducerState/ConsumerState to Producer/Consumer for portTable might have some potential issue for cold boot,
 the processing after "PortConfigDone" will not get all fields of the port and make the config like fec, AN void.  The old  ProducerState/ConsumerState happened to get all
 fields of a port key every time.

qiluo-msft: You are correct. I removed calling doPortConfigDoneTask in ctor

---
In reply to: [207479060](https://github.com/Azure/sonic-swss/pull/551#discussion_r207479060) [](ancestors = 207479060)

portsyncd/portsyncd.cpp
qiluo-msft: Your suggestion is good. Let's keep ProducerStateTable in this PR.

---
In reply to: [206734165](https://github.com/Azure/sonic-swss/pull/551#discussion_r206734165) [](ancestors = 206734165,205866896)

PR #552, opened 2018-07-27
Add test for IPv4/IPv6 tunnel creation

stepanblyschak:
Signed-off-by: stepanb <stepanb@mellanox.com>

**What I did**
I added VS test for IPv4/IPv6 tunnel creation

**Why I did it**
There was no VS test for tunnel creation, so I added it

**How I verified it**
Run VS tests

### DIFF
--- a/tests/test_tunnel.py
+++ b/tests/test_tunnel.py
@@ -1,0 +1,157 @@
+from swsscommon import swsscommon
+import time
+
+def create_fvs(**kwargs):
+    return swsscommon.FieldValuePairs(kwargs.items())
+
+class TestTunnel(object):
+    APP_TUNNEL_DECAP_TABLE_NAME = "TUNNEL_DECAP_TABLE"
+    ASIC_TUNNEL_TABLE           = "ASIC_STATE:SAI_OBJECT_TYPE_TUNNEL"
+    ASIC_TUNNEL_TERM_ENTRIES    = "ASIC_STATE:SAI_OBJECT_TYPE_TUNNEL_TERM_TABLE_ENTRY"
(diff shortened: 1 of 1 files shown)


tests/test_tunnel.py
```diff
@@ -0,0 +1,151 @@
...
+
+        # check asic db table
+        tunnel_table = swsscommon.Table(asicdb, self.ASIC_TUNNEL_TABLE)
+
+        tunnels = tunnel_table.getKeys()
+        assert len(tunnels) == 1
+
+        tunnel_sai_obj = tunnels[0]
+
+        status, fvs = tunnel_table.get(tunnel_sai_obj)
```
lguohan: we probably need to assert status == true here

tests/test_tunnel.py
```diff
@@ -0,0 +1,151 @@
...
+
+        expected_ecn_mode = self.ecn_modes_map[ecn_mode]
+        expected_dscp_mode = self.dscp_modes_map[dscp_mode]
+        expected_ttl_mode = self.ttl_modes_map[ttl_mode]
+
+        for field, value in fvs:
```
lguohan: we need to access the len(fvs) since some attribute might be missing.

tests/test_tunnel.py
stepanblyschak: Done

tests/test_tunnel.py
stepanblyschak: Added assert len(fvs) == expected number of fields

lguohan:
Retest this please

PR #553, opened 2018-07-30
Warm reboot: Add basic schema for warm start in configDB and stateDB.

jipanyang:

**What I did**
Basic schema for warm start 
Configuration:   Mainly to decide whether to flush the DB data and some timer setting for arp/route sync.

AppDB:   Store basic warm restore and sync up state and restart count for each process.

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/doc/swss-schema.md
+++ b/doc/swss-schema.md
@@ -622,15 +622,6 @@ Equivalent RedisDB entry:
     12) "0"
     127.0.0.1:6379>
 
----------------------------------------------
-### WARM\_RESTART\_TABLE
-    ;Stores application and orchdameon warm start status
-    ;Status: work in progress
-    key             = WARM_RESTART_TABLE:process_name         ; process_name is unique process identifier
-    restart_count   = 1*10DIGIT                               ; a number between 0 and 2147483647,
-                                                              ; count of warm start times.
-    state           = "init" / "restored" / "synced"
-
 
 ## Configuration DB schema
 
@@ -641,6 +632,23 @@ Equivalent RedisDB entry:
     key             = WARM_RESTART:name         ; name is the name of SONiC docker or "system" for global configuration
     enable          = "true" / "false"          ; default value as false
 
+
+## State DB schema
+
+### WARM\_RESTART\_TABLE
+    ;Stores application and orchdameon warm start status
+    ;Status: work in progress
+
+    key             = WARM_RESTART_TABLE:process_name         ; process_name is a unique process identifier.
+    restart_count   = 1*10DIGIT                               ; a number between 0 and 2147483647,
+                                                              ; count of warm start times.
+
+    state           = "init" / "restored" / "reconciled"      ; init: process init with warm start enabled.
+                                                              ; restored: process restored to the previous
+                                                              ; state using saved data.
+                                                              ; reconciled: process reconciled with up to date
+                                                              ; dynanic data like port state, neighbor, routes
+                                                              ; and so on.
 
 ## Configuration files
 What configuration files should we have?  Do apps, orch agent each need separate files?
--- a/doc/swss-schema.md
+++ b/doc/swss-schema.md
@@ -635,10 +635,12 @@ Equivalent RedisDB entry:
                                                 ; If "system" warm start knob is true, docker level knob will be ignored.
                                                 ; If "system" warm start knob is false, docker level knob takes effect.
 
-    timer_name      = 1*255VCHAR                ; timer_name is the name of warm start timer for the whole system or the specific docker,
-                                                ; Ex. "neighbor_timer", "bgp_timer".
-                                                ; There could be more than one timer in a docker and the system.
-    timer_duration  = 1*4DIGIT                  ; timer duration, in seconds.
+    timer_name      = 1*255VCHAR,               ; timer_name is the name of warm start timer for the whole system or the specific docker,
+                                                ; Ex. "neighbor_timer", "bgp_timer". The name should not contain "," character.
+                                                ; There could be more than one timer in a docker or the system, separated by ",".
+
+    timer_duration  = 1*4DIGIT,                 ; timer duration, in seconds. Separated by "," if there are more than on timer.
+
 
 ## State DB schema
 
--- a/doc/swss-schema.md
+++ b/doc/swss-schema.md
@@ -629,10 +629,16 @@ Equivalent RedisDB entry:
     ;Stores system warm start configuration
     ;Status: work in progress
 
-    key             = WARM_RESTART:name         ; name is the name of SONiC docker or "system" for global configuration
-    enable          = "true" / "false"          ; default value as false
-    timer           = 1*4DIGIT                  ; timer to determine when reconciliation is done, in seconds.
+    key             = WARM_RESTART:name         ; name is the name of SONiC docker or "system" for global configuration.
 
+    enable          = "true" / "false"          ; Default value as false.
+                                                ; If "system" warm start knob is true, docker level knob will be ignored.
+                                                ; If "system" warm start knob is false, docker level knob takes effect.
+
+    timer_name      = 1*255VCHAR                ; timer_name is the name of warm start timer for the whole system or the specific docker,
+                                                ; Ex. "neighbor_timer", "bgp_timer".
+                                                ; There could be more than one timer in a docker and the system.
+    timer_duration  = 1*4DIGIT                  ; timer duration, in seconds.
 
 ## State DB schema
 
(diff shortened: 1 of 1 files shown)


doc/swss-schema.md
```diff
@@ -620,7 +622,27 @@ Equivalent RedisDB entry:
...
 
-### Configuration files
+---------------------------------------------
+### WARM\_RESTART\_TABLE
+    ;Stores application and orchdameon warm start status
+    ;Status: work in progress
+    key             = WARM_RESTART_TABLE:process_name         ; process_name is unique process identifier
+    restart_count   = 1*10DIGIT                               ; a number between 0 and 2147483647,
+                                                              ; count of warm start times.
+    state           = "init" / "restored" / "synced"
```
lguohan: can you add definition for each phase?

lguohan: restored: restore the previous
synced-> reconciled-> reconcile the previous view and new view

doc/swss-schema.md
```diff
@@ -620,7 +622,27 @@ Equivalent RedisDB entry:
...
+
+
+## Configuration DB schema
+
+### WARM\_RESTART
+    ;Stores system warm start configuration
+    ;Status: work in progress
+
+    key             = WARM_RESTART:name         ; name is the name of SONiC docker or "system" for global configuration
+    enable          = "true" / "false"          ; default value as false
```
lguohan: timer to determine when to do the reconciliation.

doc/swss-schema.md
```diff
@@ -620,7 +622,27 @@ Equivalent RedisDB entry:
     12) "0"
     127.0.0.1:6379>
 
-### Configuration files
+---------------------------------------------
+### WARM\_RESTART\_TABLE
```
lguohan: should be put into state db.

jipanyang:
@zhenggen-xu @rodnymolina Please help review it.

doc/swss-schema.md
```diff
@@ -620,7 +622,36 @@ Equivalent RedisDB entry:
...
+
+## Configuration DB schema
+
+### WARM\_RESTART
+    ;Stores system warm start configuration
+    ;Status: work in progress
+
+    key             = WARM_RESTART:name         ; name is the name of SONiC docker or "system" for global configuration
+    enable          = "true" / "false"          ; default value as false
+    timer           = 1*4DIGIT                  ; timer to determine when reconciliation is done, in seconds.
```
lguohan: add comment that "not all docker will need the reconciliation timer, for example if BGP docker use GR then as soon as the EOR message is received, the reconciliation is finished."

doc/swss-schema.md
rodnymolina: I thought we had agreed to remove this docker-specific knob to turn on/off this feature, and make use of a system-wide knob instead, right?

doc/swss-schema.md
rodnymolina: A couple of points here...

* The bgp-restartability solution that i've been working on does not rely on EOR functionality -- neither FRR nor Quagga are capable today of generating this message from BGP down to Zebra.

* Even if we had an EOR message coming to fpmSyncd, we would still need a timer to make sure that the reconciliation process is always executed after a reasonable amount of time has elapsed -- the EOR message could get lost or not being generated due to a bug or diverse reasons.

rodnymolina: Btw, as discussed, i'm fine with this docker/feature-specific timer.

doc/swss-schema.md
jipanyang: I think the agreement is to keep both docker-specific knob and system-wide knob. 
The docker-specific knob will mainly be used by developer.

zhenggen-xu: I agree to have both knobs which give convenience .  Can we be more specific in the document how system wide knob work on conjunction with docker knob?

doc/swss-schema.md
zhenggen-xu: neighborsyncd will need the timer as well,  this timer is not necessarily docker specific, but more process specific.

BTW: Should we keep the table key/fields definition in some file, so we know what exactly the tables are supported?

doc/swss-schema.md
rodnymolina: Agree, I like the idea of keeping a docker-specific knob for tshooting purposes.

jipanyang: The expect behavior:
<1> If system wide knob  is enabled, then docker level know will be ignored. 
There is no point doing system warm reboot while specific docker warm restart is disabled.
<2> If system wide knob  is disabled,  docker level knob will be checked.
The use case is that we may want to docker level upgrade for dockers like bgp and swss.
Will put more info in the schema comment section.

doc/swss-schema.md
jipanyang: If the timer is more process specific,  some update here is necessary since the key here is either "system" or docker name.  For now, we don't want to have support for process level warm restart knob(?) .

I guess this swss-schema.md file is a good place to keep the key/fields definitions for all the DBs?

zhenggen-xu: Looks good to me.

I think we need put the table key/fields definitions in the code (e,g, .h file) as well,  so it is synced by all code. This can be done later.

doc/swss-schema.md
```diff
@@ -620,7 +622,42 @@ Equivalent RedisDB entry:
...
+
+### WARM\_RESTART
+    ;Stores system warm start configuration
+    ;Status: work in progress
+
+    key             = WARM_RESTART:name         ; name is the name of SONiC docker or "system" for global configuration.
+
+    enable          = "true" / "false"          ; Default value as false.
+                                                ; If "system" warm start knob is true, docker level knob will be ignored.
+                                                ; If "system" warm start knob is false, docker level knob takes effect.
```
lguohan: docker knob always override system level knob, that was the use cases we dicussed.

jipanyang: The tricky point is the default value is false.   If docker knob always override system level knob,  when warm restart is enabled at system level, each docker has to be enabled too. 

What is the use case of having warm restart enabled at system level, while disabled for certain docker?

lguohan: if docker knob does not have this attribute, then use system level.

the use case was rodney's, snmp docker snmp warm reboot is available but experimental, so in production he wants to explicitly disable it although system level is enabled.

jipanyang: @rodnymolina  for snmp case, probably it should alway do cold boot  as other dockers like lldp, dhcp_relay until its warm reboot support is stable enough.  

If the existence of certain attribute is treated as one condition, then it becomes tri-state variable:  true/false/not-exist,  to me, it seems error prone. We don't have such case in sonic yet(?).

zhenggen-xu: In general, more specific knobs take precedence over global configuration. I think we can use this principle here as well.
If we don't like the not-exist case, we need this knob for every docker. For the time-being, we can define the knobs for dockers which have code to support warm reboot, for the ones do not support warm reboot, they do not look at the knobs anyways.

jipanyang: If docker knob always takes precedence over system knob,  the not-exist case is asked implicitly, otherwise why system knob is needed if it gets preempted all the time. That means the default value for enable is not-exist.

The knobs are also used to control whether to flush DB.  
Ex. based on the current schema,  if system knob or one docker knob enabled, we keep the appDB and some other DBs data. 

But if docker knob always takes precedence,  and it is disabled,  should the DB be flushed or not.   If DB is not flushed, the specific docker will have to handle the scenarios of DB with existing data though it is supposed to do a cold start.   If DB is flushed, how about other dockers?

Better not to support the case of system knob enabled, while docker knob disabled and takes effect.  That is one of the reasons for  `If "system" warm start knob is true, docker level knob will be ignored.`

zhenggen-xu: @jipanyang  As talked, this "system" knob was meant for whole system warm reboot, not a global setting of docker restart, that would lead to a different discussion.

Could you please help to provide (the current) work flow of using these knobs during the shutdown and bringing up of the dockers/system? We can then think whether the whole work flow make sense and keeps everybody on the same page.

jipanyang: After discussion with @zhenggen-xu ,  a little more clarification about the use cases for system knob and docker knobs.
<1> Upon system reboot, the system enable knob will be checked.  If enabled, database data will be preserved, if not, database will be flushed.
No need to check docker level knobs in this case since the whole system is being rebooted .

<2> Upon docker service start,  first to check system knob, if enabled, docker warm start should be performed, otherwise system warm reboot will be ruined.  If system knob disabled, while docker knob enabled, this is likely an individual docker warm restart request. 

Within each application, when the system knob or docker knob enabled, we do further check on the actual warm start state ( restart_count),  if no warm start state data available, the database has been flushed, do cold start. Otherwise warm start.

The exact logic has been implemented in WarmStart::checkWarmStart():
https://github.com/Azure/sonic-swss/pull/547/files

zhenggen-xu: I am fine with this logic.

PR #554, opened 2018-07-30
Warm reboot: Orchagent state restore

jipanyang:

**What I did**
State restore for Orchagent warm start 

It includes basic after restore orchagent state validation: No un-met object dependency in orchagent.  

**Why I did it**

**How I verified it**
VS testing.  All existing vs test cases pass.

**Details if related**

Has dependency on:
https://github.com/Azure/sonic-swss-common/pull/211
https://github.com/Azure/sonic-swss-common/pull/214
https://github.com/Azure/sonic-swss/pull/546
https://github.com/Azure/sonic-swss/pull/547
https://github.com/Azure/sonic-swss/pull/551

Follow up PRs planned: 
<1> Orchagent  pre-warm-restart validation.
This is to have a deterministic known state before warm restart,  otherwise state restore will be a best effort processing.

<2> state sync up
dynamic data like port state, fdb, arp may change during the restart window.

<3> Warm restart phase state data.
This is to store basic warm restart state data as in  https://github.com/Azure/sonic-swss/pull/553
And more debug information should be provided like exact list of postponed tasks/executor during restore.

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1291,7 +1291,7 @@ bool AclRuleDTelFlowWatchListEntry::validateAddAction(string attr_name, string a
     sai_object_id_t session_oid;
 
     if (!m_pDTelOrch ||
-        (attr_name != ACTION_DTEL_FLOW_OP && 
+        (attr_name != ACTION_DTEL_FLOW_OP &&
         attr_name != ACTION_DTEL_INT_SESSION &&
         attr_name != ACTION_DTEL_FLOW_SAMPLE_PERCENT &&
         attr_name != ACTION_DTEL_REPORT_ALL_PACKETS))
@@ -1356,7 +1356,7 @@ bool AclRuleDTelFlowWatchListEntry::validateAddAction(string attr_name, string a
         value.aclaction.parameter.booldata = (attr_value == DTEL_ENABLED) ? true : false;
         value.aclaction.enable = (attr_value == DTEL_ENABLED) ? true : false;
     }
-    
+
     m_actions[aclDTelActionLookup[attr_name]] = value;
 
     return true;
@@ -1514,7 +1514,7 @@ bool AclRuleDTelDropWatchListEntry::validateAddAction(string attr_name, string a
 
     value.aclaction.parameter.booldata = (attr_value == DTEL_ENABLED) ? true : false;
     value.aclaction.enable = (attr_value == DTEL_ENABLED) ? true : false;
-    
+
     m_actions[aclDTelActionLookup[attr_name]] = value;
 
     return true;
@@ -1693,9 +1693,8 @@ void AclOrch::init(vector<TableConnector>& connectors, PortsOrch *portOrch, Mirr
     // initialized before thread start.
     auto interv = timespec { .tv_sec = COUNTERS_READ_INTERVAL, .tv_nsec = 0 };
     auto timer = new SelectableTimer(interv);
-    auto executor = new ExecutableTimer(timer, this);
-    executor->setName("ACL_POLL_TIMER");
-    Orch::addExecutor("", executor);
+    auto executor = new ExecutableTimer(timer, this, "ACL_POLL_TIMER");
+    Orch::addExecutor(executor);
     timer->start();
 }
 
@@ -2321,7 +2320,7 @@ sai_status_t AclOrch::bindAclTable(sai_object_id_t table_oid, AclTable &aclTable
     sai_status_t status = SAI_STATUS_SUCCESS;
 
     SWSS_LOG_INFO("%s table %s to ports", bind ? "Bind" : "Unbind", aclTable.id.c_str());
-    
+
     if (aclTable.ports.empty())
     {
         if (bind)
@@ -2432,7 +2431,7 @@ sai_status_t AclOrch::createDTelWatchListTables()
     SWSS_LOG_INFO("Successfully created ACL table %s, oid: %lX", flowWLTable.description.c_str(), table_oid);
 
     /* Create Drop watchlist ACL table */
-    
+
     table_attrs.clear();
 
     dropWLTable.id = TABLE_TYPE_DTEL_DROP_WATCHLIST;
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1694,6 +1694,7 @@ void AclOrch::init(vector<TableConnector>& connectors, PortsOrch *portOrch, Mirr
     auto interv = timespec { .tv_sec = COUNTERS_READ_INTERVAL, .tv_nsec = 0 };
     auto timer = new SelectableTimer(interv);
     auto executor = new ExecutableTimer(timer, this);
+    executor->setName("ACL_POLL_TIMER");
     Orch::addExecutor("", executor);
     timer->start();
 }
--- a/orchagent/crmorch.cpp
+++ b/orchagent/crmorch.cpp
@@ -165,9 +165,8 @@ CrmOrch::CrmOrch(DBConnector *db, string tableName):
     // The CRM stats needs to be populated again
     m_countersCrmTable->del(CRM_COUNTERS_TABLE_KEY);
 
-    auto executor = new ExecutableTimer(m_timer.get(), this);
-    executor->setName("CRM_COUNTERS_POLL");
-    Orch::addExecutor("CRM_COUNTERS_POLL", executor);
+    auto executor = new ExecutableTimer(m_timer.get(), this, "CRM_COUNTERS_POLL");
+    Orch::addExecutor(executor);
     m_timer->start();
 }
 
--- a/orchagent/crmorch.cpp
+++ b/orchagent/crmorch.cpp
@@ -162,7 +162,11 @@ CrmOrch::CrmOrch(DBConnector *db, string tableName):
         m_resourcesMap.emplace(res.first, CrmResourceEntry(res.second, CRM_THRESHOLD_TYPE_DEFAULT, CRM_THRESHOLD_LOW_DEFAULT, CRM_THRESHOLD_HIGH_DEFAULT));
     }
 
+    // The CRM stats needs to be populated again
+    m_countersCrmTable->del(CRM_COUNTERS_TABLE_KEY);
+
     auto executor = new ExecutableTimer(m_timer.get(), this);
+    executor->setName("CRM_COUNTERS_POLL");
     Orch::addExecutor("CRM_COUNTERS_POLL", executor);
     m_timer->start();
 }
@@ -333,7 +337,7 @@ void CrmOrch::decCrmAclUsedCounter(CrmResourceType resource, sai_acl_stage_t sta
     {
         m_resourcesMap.at(resource).countersMap[getCrmAclKey(stage, point)].usedCounter--;
 
-        // Remove ACL table related counters 
+        // Remove ACL table related counters
         if (resource == CrmResourceType::CRM_ACL_TABLE)
         {
             auto & cntMap = m_resourcesMap.at(CrmResourceType::CRM_ACL_TABLE).countersMap;
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -54,7 +54,7 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
         }
 
         // we already have such entries
-        if (m_entries.count(update.entry) != 0)
+        if (m_entries.find(update.entry) != m_entries.end())
         {
              SWSS_LOG_INFO("FdbOrch notification: mac %s is already in bv_id 0x%lx",
                     update.entry.mac.to_string().c_str(), entry->bv_id);
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -53,7 +53,8 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
             return;
         }
 
-        if (m_entries.count(update.entry) != 0) // we already have such entries
+        // we already have such entries
+        if (m_entries.count(update.entry) != 0)
         {
              SWSS_LOG_INFO("FdbOrch notification: mac %s is already in bv_id 0x%lx",
                     update.entry.mac.to_string().c_str(), entry->bv_id);
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -27,13 +27,17 @@ FdbOrch::FdbOrch(DBConnector *db, string tableName, PortsOrch *port) :
     m_portsOrch->attach(this);
     m_flushNotificationsConsumer = new NotificationConsumer(db, "FLUSHFDBREQUEST");
     auto flushNotifier = new Notifier(m_flushNotificationsConsumer, this);
+    flushNotifier->setName("FLUSHFDBREQUEST");
     Orch::addExecutor("", flushNotifier);
 
     /* Add FDB notifications support from ASIC */
     DBConnector *notificationsDb = new DBConnector(ASIC_DB, DBConnector::DEFAULT_UNIXSOCKET, 0);
     m_fdbNotificationConsumer = new swss::NotificationConsumer(notificationsDb, "NOTIFICATIONS");
     auto fdbNotifier = new Notifier(m_fdbNotificationConsumer, this);
+    fdbNotifier->setName("FDB_NOTIFICATIONS");
     Orch::addExecutor("FDB_NOTIFICATIONS", fdbNotifier);
+
+    addExistingData(tableName);
 }
 
 void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_object_id_t bridge_port_id)
@@ -51,6 +55,13 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
         {
             SWSS_LOG_ERROR("Failed to get port by bridge port ID 0x%lx", bridge_port_id);
             return;
+        }
+
+        if (m_entries.count(update.entry) != 0) // we already have such entries
+        {
+             SWSS_LOG_INFO("FdbOrch notification: mac %s is already in bv_id 0x%lx",
+                    update.entry.mac.to_string().c_str(), entry->bv_id);
+             break;
         }
 
         update.add = true;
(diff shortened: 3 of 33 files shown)


lguohan:
need new vs test to test warm boot scenario

orchagent/aclorch.cpp
```diff
@@ -1694,6 +1694,7 @@ void AclOrch::init(vector<TableConnector>& connectors, PortsOrch *portOrch, Mirr
     auto interv = timespec { .tv_sec = COUNTERS_READ_INTERVAL, .tv_nsec = 0 };
     auto timer = new SelectableTimer(interv);
     auto executor = new ExecutableTimer(timer, this);
+    executor->setName("ACL_POLL_TIMER");
```
lguohan: why not just change the constructor to set the name. 

not sure if it is useful to introduce setName method, since we do not (probably should not) allow change the name.

orchagent/orch.cpp
```diff
@@ -171,6 +171,23 @@ void Consumer::drain()
...
+{
+    for (auto &tm :m_toSync)
+    {
+        KeyOpFieldsValuesTuple& tuple = tm.second;
+
+        string s = getTableName() + ":" + kfvKey(tuple)
+               + "|" + kfvOp(tuple);
+        for (auto i = kfvFieldsValues(tuple).begin(); i != kfvFieldsValues(tuple).end(); i++)
+        {
+            s += "|" + fvField(*i) + ":" + fvValue(*i);
```
lguohan: do not hardcode the separator.

orchagent/orchdaemon.cpp
```diff
@@ -313,5 +358,81 @@ void OrchDaemon::start()
...
+{
+    /*
+     * No pending task should exist for any of the consumer at this point.
+     * All the prexisting data in appDB and configDb have been read and processed.
+     */
+    vector<string> ts;
+    getTaskToSync(ts);
+    if (ts.size() != 0)
+    {
+        // TODO: change it to fatal once staged ProducerStateTable/ConsumerStateTable change and
```
lguohan: remove TODO here?

orchagent/orchdaemon.cpp
```diff
@@ -152,13 +162,13 @@ bool OrchDaemon::init()
         gAclOrch = new AclOrch(acl_table_connectors, gPortsOrch, mirror_orch, gNeighOrch, gRouteOrch);
     }
 
-    m_orchList.push_back(gAclOrch);
     m_orchList.push_back(gFdbOrch);
+    m_orchList.push_back(mirror_orch);
```
lguohan: why move mirror orch and acl orch after fdb orch?

orchagent/orchdaemon.cpp
```diff
@@ -313,5 +358,81 @@ void OrchDaemon::start()
...
+                o->doTask();
+            }
+
+            warmRestoreValidation();
+            SWSS_LOG_NOTICE("Orchagent state restore done");
+            restored = true;
+            syncd_apply_view();
+
+            // TODO: should be set after port/fdb/arp sync up
+            WarmStart::setWarmStartState("orchagent", WarmStart::RECONCILED);
```
lguohan: how do we sync the port status and fdb notifications?

orchagent/orchdaemon.cpp
jipanyang: For planned warm restart,  there is one PR https://github.com/Azure/sonic-swss/pull/562/files     for enforcing orchagent to be in certain state before stopping orchagent/swss docker.  After orchagent is restored, it should gain back the same state.

But there is open issue about the exact criteria of warm restart readiness. Currently we want to make sure that there is no un-met object dependency in orchagent which means the "SyncMap m_toSync;" of each Consumer is empty.   @zhenggen-xu  ever expressed concern over this check.

orchagent/orchdaemon.cpp
jipanyang: This is to reduce the possible turbulence caused by the dependency relaxation processing during state restore.
mirror orch may use fdb mac/vlan info. 
   ``` if (!m_portsOrch->getPort(session.neighborInfo.neighbor.alias, session.neighborInfo.port))```
and acl orch may check mirror session
      ``` "if (!m_pMirrorOrch->getSessionState(m_sessionName, state))"```

orchagent/orchdaemon.cpp
jipanyang: Port state sync: https://github.com/Azure/sonic-swss/pull/557
FDB state sync: https://github.com/Azure/sonic-swss/pull/558
I think @zhenggen-xu  is to propose PR for arp sync up.

orchagent/orchdaemon.cpp
zhenggen-xu: This check is not necessary reach the desired result.  The problem is if some dependencies were never met (e,g, the nexthop is not resolved for some reason for a particular route), which was the case for pre-reboot,  but may kill the whole warm reboot process. Looks like we may want to handle those entries differently or we simply do a loose check (print warning etc) and move forward.

orchagent/orchdaemon.cpp
lguohan: I feel we should take a different approach to solve port oper status and fdb sync in orchagent. 

Since we have information in asic db, we should just send syncd a command and ask syncd to enumerate all state in asic db and regenerate all the events back to orchagent. 

The ground truth is in asic_db.

jipanyang: We may have more in-depth discussion of port state sync up and fdb sync up in the corresponding PRs.  Here is some quick recap:

<1> Current SONiC use NotificationProducer to inform port state change and fdb,  which means orchagent will miss the signal if it happens between orchagent shutdown and startup.

Syncd doens't know whether the signal has been lost or not.  Also that notification event is triggered from libsai/SDK.

<2> In current port sync up implementation,  we do a get SAI_PORT_ATTR_OPER_STATUS via sai API to have the latest oper status, then update hostif and db accordingly.  

<3> For FDB sync up,   since syncd does put the latest FDB entry into ASIC DB,  yes, we are retrieving the latest FDB from ASIC.

Surely there is some other alternative, like change the Notification mechanism to some db data persistent channel. Under current framework, the above approaches have proven to be working and stable.  I should have put more comments to the port state and fdb sync up implementation.

syncd_apply_view() is called after state restore,  I guess that could make the life of syncd view comparison a little easier, but it is up to how you want view comparison to work.  The libsai redis idempotent API solution doesn't have dependency on  syncd_apply_view().

lguohan: we can ask syncd to pause sending notification when we are upgrading.

it is better to ask syncd to re-send the previous view to us. we do not orchagent to scan the asic view because the format might change and it can break orchagent code. orchagent should not look at the asic db.

Can you modify you fdb sync and port sync to syncd?

jipanyang: For port state sync, I don't think there is need to make the solution so complex.
FDB format could be a valid concern, but we should alway be careful when making format change.

My understanding of SONiC design about docker is to decouple the services and make them more independent from each other. But I'm seeing the intent to couple swss and syncd tightly together, what is the real purpose behind this?  Why were syncd and orchagent put into different dockers at the first place?

orchagent/orchdaemon.cpp
jipanyang: Do we want to allow this scenario (there is unfulfilled dependency) for planned warm restart?

The possible alternative is to add a knob to have two levels of pre-reboot check:
1. Allow unfulfilled dependency.  
2. disallow unfulfilled dependency.  
After warm restore, we do the check accordingly.

With no intention to make it a hot topic here and block everything else,  I'll change the log level to NOTICE and continue for now.

Please go to https://github.com/Azure/sonic-swss/pull/562 for further discussion.

orchagent/orchdaemon.cpp
```diff
@@ -313,5 +358,81 @@ void OrchDaemon::start()
...
+            /*
+             * drain remaining data that are out of order like LAG_MEMBER_TABLE and VLAN_MEMBER_TABLE
+             * since they were checked before LAG_TABLE and VLAN_TABLE.
+             */
+            for (Orch *o : m_orchList)
+            {
+                o->doTask();
+            }
+
+            warmRestoreValidation();
```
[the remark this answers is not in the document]
jipanyang: Yes,  once we reach agreement what should be checked before reboot, and enforce the validation after restore.

orchagent/orchdaemon.cpp
```diff
@@ -304,7 +324,17 @@ void OrchDaemon::start()
         }
 
         auto *c = (Executor *)s;
-        c->execute();
+
+        if (restored)
```
qiluo-msft: >if [](start = 8, length = 2)

select() is used for normal processing, and it is not for warm start, because all the data is already added during bake() -> addExistingData().

You can extract the warm start logic and move before the normal processing while-loop.

jipanyang: Right.  Previously configDb data was retrieved here,  it is not needed any more.  pre-shutdown  warm restart check may need to be processed in select(), and possibly arp sync up event too,  but that could be done in separate PRs.

orchagent/orch.h
```diff
@@ -68,9 +68,10 @@ class Orch;
 class Executor : public Selectable
 {
 public:
-    Executor(Selectable *selectable, Orch *orch)
+    Executor(Selectable *selectable, Orch *orch, const string &name)
         : m_selectable(selectable)
         , m_orch(orch)
+        , m_name(name)
```
qiluo-msft: >m_name [](start = 10, length = 6)

I believe the new name field is not necessary, and you could always get it from m_consumerMap key. Could you help explorer the possibility?

orchagent/orch.h
```diff
@@ -116,15 +125,23 @@ class Consumer : public Executor {
...
     }
 
+    int getDbId() const
+    {
+        return getConsumerTable()->getDbId();
+    }
+
+    string dumpTuple(KeyOpFieldsValuesTuple &tuple);
+    void dumpToSyncTasks(vector<string> &ts);
+    void execute();
```
qiluo-msft: >void execute(); [](start = 4, length = 15)

No need to move this line.

orchagent/orch.h
jipanyang: Isn't it a good practice to give executor a fixed unique name? For now it is mainly for debugging, later it may also be used for state validation log.  
I noticed that there are even cases where empty string given to to map as key in current code,  all those should be changed.

orchagent/orch.cpp
```diff
@@ -185,6 +185,30 @@ void Consumer::drain()
         m_orch->doTask(*this);
 }
 
+string Consumer::dumpTuple(KeyOpFieldsValuesTuple &tuple)
+{
+    string s = getTableName() + getConsumerTable()->getTableNameSeparator() + kfvKey(tuple)
+               + "|" + kfvOp(tuple);
```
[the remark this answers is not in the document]
jipanyang: This a good question.
We had it fixed to ":" previously.  Changed it to getConsumerTable()->getTableNameSeparator() upon comment.  I think one of he benefits of doing this is we'll have the same key format as that in redis DB.  Any other idea?

orchagent/orch.cpp
```diff
@@ -185,6 +185,30 @@ void Consumer::drain()
...
+    for (auto i = kfvFieldsValues(tuple).begin(); i != kfvFieldsValues(tuple).end(); i++)
+    {
+        s += "|" + fvField(*i) + ":" + fvValue(*i);
+    }
+
+    return s;
+}
+
+void Consumer::dumpToSyncTasks(vector<string> &ts)
```
qiluo-msft: >dumpToSyncTasks [](start = 15, length = 15)

dumpTuple() and dumpToSyncTasks() are not well designed interfaces.

I agree the observability is important.

orchagent/main.cpp
```diff
@@ -22,6 +22,7 @@ extern "C" {
 #include "saihelper.h"
 #include "notifications.h"
 #include <signal.h>
+#include <warm_restart.h>
```
qiluo-msft: #include "warm_restart.h"

orchagent/intfsorch.cpp
```diff
@@ -86,7 +86,16 @@ void IntfsOrch::doTask(Consumer &consumer)
         {
             if (alias == "lo")
             {
-                addIp2MeRoute(ip_prefix);
+                // set request for lo may come after warm start restore
```
qiluo-msft: Does it related to warm start? If not, let's do it in another PR.

orchagent/intfsorch.cpp
```diff
@@ -171,6 +180,7 @@ void IntfsOrch::doTask(Consumer &consumer)
         {
             if (alias == "lo")
             {
+                m_syncdIntfses.erase(alias);
```
qiluo-msft: The same.

orchagent/fdborch.cpp
```diff
@@ -53,6 +53,14 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
             return;
         }
 
+        // we already have such entries
+        if (m_entries.count(update.entry) != 0)
```
qiluo-msft: >count [](start = 22, length = 5)

find()

orchagent/crmorch.cpp
```diff
@@ -162,8 +162,11 @@ CrmOrch::CrmOrch(DBConnector *db, string tableName):
         m_resourcesMap.emplace(res.first, CrmResourceEntry(res.second, CRM_THRESHOLD_TYPE_DEFAULT, CRM_THRESHOLD_LOW_DEFAULT, CRM_THRESHOLD_HIGH_DEFAULT));
     }
 
-    auto executor = new ExecutableTimer(m_timer.get(), this);
-    Orch::addExecutor("CRM_COUNTERS_POLL", executor);
+    // The CRM stats needs to be populated again
+    m_countersCrmTable->del(CRM_COUNTERS_TABLE_KEY);
```
qiluo-msft: >m_countersCrmTable->del(CRM_COUNTERS_TABLE_KEY); [](start = 4, length = 48)

If it is applied to cold start, let's do it in another PR.

orchagent/orch.h
qiluo-msft: I agree on "even cases where empty string given to to map as key in current code, all those should be changed." and I will fix them.

orchagent/orch.h
jipanyang: ok.

orchagent/orch.cpp
```diff
@@ -185,6 +185,30 @@ void Consumer::drain()
...
+{
+    for (auto &tm :m_toSync)
```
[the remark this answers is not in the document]
jipanyang: ok

orchagent/crmorch.cpp
jipanyang: For cold start,  no data exists for CRM stats table, so  it has no effect.

orchagent/fdborch.cpp
jipanyang: ok.

orchagent/intfsorch.cpp
jipanyang: It is related to warm start.  It is noticed that sometime IP data for lo comes after warm start.

orchagent/main.cpp
jipanyang: ok.

orchagent/orch.cpp
jipanyang: dumpTuple() is existing function name.  It is refactored so the common function may be used in multiple places. Any specific issue/suggestion?

orchagent/orch.cpp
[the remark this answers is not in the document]
jipanyang: ok

orchagent/orch.h
jipanyang: ok

orchagent/intfsorch.cpp
qiluo-msft: The logic is also applied to cold start. Are you fixing a bug also exists in cold start? if yes, we prefer a separate PR just for it.

---
In reply to: [209787586](https://github.com/Azure/sonic-swss/pull/554#discussion_r209787586) [](ancestors = 209787586)

orchagent/orch.h
qiluo-msft: Fixed by [https://github.com/Azure/sonic-swss/pull/575](https://github.com/Azure/sonic-swss/pull/575)

---
In reply to: [209784061](https://github.com/Azure/sonic-swss/pull/554#discussion_r209784061) [](ancestors = 209784061)

jipanyang: I didn't say there was bug there, at least it is working. But it is a bad design putting empty string as key.
The purpose of name related change in this PR is to facilitate debugging and logging via a meaningful name for each excecutor including those timers and Notifiers.  I found it especially useful during the warm restart testing since very often you need to trace specific executor whenever problem happened.

orchagent/orch.cpp
qiluo-msft: I will investigate and hopefully submit a PR

---
In reply to: [209787965](https://github.com/Azure/sonic-swss/pull/554#discussion_r209787965) [](ancestors = 209787965)

jipanyang:
What exactly are you doing? Reviewing code? or copying code and making a little change then push your own pr and we have to merge it again and again?





来自钉钉专属商务邮箱------------------------------------------------------------------
Sender:Qi Luo<notifications@github.com>
Date:2018/08/15 02:07:03
Recipient:Azure/sonic-swss<sonic-swss@noreply.github.com>
Cc:YANG, Jipan(继攀)<jipan.yang@alibaba-inc.com>; Author<author@noreply.github.com>
Theme:Re: [Azure/sonic-swss] Warm reboot: Orchagent state restore (#554)

@qiluo-msft commented on this pull request. 
In orchagent/orch.cpp:
> @@ -185,6 +185,30 @@ void Consumer::drain()
         m_orch->doTask(*this);
 }
 
+string Consumer::dumpTuple(KeyOpFieldsValuesTuple &tuple)
+{
+    string s = getTableName() + getConsumerTable()->getTableNameSeparator() + kfvKey(tuple)
+               + "|" + kfvOp(tuple);
+    for (auto i = kfvFieldsValues(tuple).begin(); i != kfvFieldsValues(tuple).end(); i++)
+    {
+        s += "|" + fvField(*i) + ":" + fvValue(*i);
+    }
+
+    return s;
+}
+
+void Consumer::dumpToSyncTasks(vector<string> &ts)

I will investigate and hopefully submit a PR 
In reply to: 209787965 [](ancestors = 209787965)

lguohan:
test vs fails, can you fix?

lguohan:
retest this please

PR #557, opened 2018-08-01
Warm reboot: port state sync up

jipanyang:

**What I did**
During orchagent warm start window,  port state may be changed.  PortsOrch::syncUpPortState() is to be called after the whole orchagent state has been restored.

**Why I did it**

**How I verified it**

VS test case to be provided once  https://github.com/Azure/sonic-swss/pull/554 is settled.

**Details if related**

Has dependency on https://github.com/Azure/sonic-swss/pull/554


orchagent/portsorch.cpp
```diff
@@ -1029,6 +1029,30 @@ void PortsOrch::updateDbPortOperStatus(sai_object_id_t id, sai_port_oper_status_
     }
 }
 
+sai_port_oper_status_t PortsOrch::getDbPortOperStatus(sai_object_id_t id)
```
lguohan: where does this one get used?

jipanyang: This to get the last know port oper status saved in appDB.  Baseline for status change check.

orchagent/portsorch.cpp
```diff
@@ -2873,3 +2873,49 @@ void PortsOrch::doTask(NotificationConsumer &consumer)
...
+ *
+ * Currently NotificationProducer is used by syncd to inform port state change,
+ * which means orchagent will miss the signal if it happens between orchagent shutdown and startup.
+ * Syncd doesn't know whether the signal has been lost or not.
+ * Also the source of notification event is from libsai/SDK.
+ *
+ * Latest oper status for each port is retrieved via SAI_PORT_ATTR_OPER_STATUS sai API,
+ * the hostif and db are updated accordingly.
+ */
+void PortsOrch::syncUpPortState()
```
qiluo-msft: >syncUpPortState [](start = 16, length = 15)

syncUpPortStatus

orchagent/portsorch.cpp
```diff
@@ -2873,3 +2873,49 @@ void PortsOrch::doTask(NotificationConsumer &consumer)
...
+{
+    SWSS_LOG_ENTER();
+
+    for (auto &it: m_portList)
```
qiluo-msft: By design, it does the same with PortsOrch::doTask(NotificationConsumer &consumer). Is it possible to reuse code?

For example, feed status into NotificationConsumer, and let doTask process them? Or refactor a common function.

orchagent/portsorch.cpp
qiluo-msft: I have the same question. It is not called by anywhere.

---
In reply to: [209355244](https://github.com/Azure/sonic-swss/pull/557#discussion_r209355244) [](ancestors = 209355244)

orchagent/portsorch.cpp
jipanyang: The way current implementation works doesn't require timer to do a cutoff as that for arp and route.  I don't see the benefit of transforming status fetched from ASIC into notification message. 

Yes, they both call updateDbPortOperStatus() and setHostIntfsOperStatus(), all other processing are different due to the difference in port state source and object scope.  

In the warm start sync up phase, a little more check is done to compare with current port oper status so updateDbPortOperStatus() and setHostIntfsOperStatus() are called only when needed.  Similar check might be desired for port state change notification, in that case common function may be good.

orchagent/portsorch.cpp
jipanyang: For warm start, we have restore phase and sync up phase, that is one of reasons the name of syncUpPortState() is used.   My focus is on the function part.  It looks you have strong opinion on the name,  I could change it to  refreshPortStatus().

orchagent/portsorch.cpp
jipanyang: it was in old implementation, which had been superseded by m_oper_status in port class.

qiluo-msft: Sorry I think there is miscommunication. I mean in this PR, the new function is added and seems good, but not be called by any place, so it takes no effect.

qiluo-msft:
Retest this please

qiluo-msft:
@lguohan please help review. Do we need add vs test here?

orchagent/portsorch.cpp
```diff
@@ -2866,10 +2866,64 @@ void PortsOrch::doTask(NotificationConsumer &consumer)
...
+        auto &p = it.second;
+        if (p.m_type == Port::PHY)
+        {
+            sai_attribute_t attr;
+            attr.id = SAI_PORT_ATTR_OPER_STATUS;
+
+            sai_status_t ret = sai_port_api->get_port_attribute(p.m_port_id, 1, &attr);
+            if (ret != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to get oper status for %s", p.m_alias.c_str());
```
nikos-github: If you fail here, shouldn't you revert to cold start? An error is not sufficient and system will be in inconsistent state post warm reboot.

jipanyang: I guess throwing exception is the option here. There is no such magic of automatically reverting to cold start yet.

qiluo-msft:
Retest this please

PR #558, opened 2018-08-01
Warm reboot: FDB  sync up

jipanyang:

**What I did**
During orchagent warm start window,  FDB may be changed. void FdbOrch::syncUpFdb() is to be called after the whole orchagent state has been restored.

**Why I did it**

**How I verified it**

VS test case to be provided once #554 is settled.

**Details if related**
Has dependency on #554

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -10,9 +10,6 @@
 #include "crmorch.h"
 #include "notifier.h"
 #include "sai_serialize.h"
-#include "json.hpp"
-
-using json = nlohmann::json;
 
 extern sai_fdb_api_t    *sai_fdb_api;
 
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -10,6 +10,9 @@
 #include "crmorch.h"
 #include "notifier.h"
 #include "sai_serialize.h"
+#include "json.hpp"
+
+using json = nlohmann::json;
 
 extern sai_fdb_api_t    *sai_fdb_api;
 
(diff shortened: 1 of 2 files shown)


orchagent/fdborch.cpp
```diff
@@ -36,6 +39,59 @@ FdbOrch::FdbOrch(DBConnector *db, string tableName, PortsOrch *port) :
     Orch::addExecutor("FDB_NOTIFICATIONS", fdbNotifier);
 }
 
+
+void FdbOrch::syncUpFdb()
```
zhenggen-xu: Where this syncUpFdb() is (going to be) called?

orchagent/fdborch.cpp
```diff
@@ -36,6 +39,59 @@ FdbOrch::FdbOrch(DBConnector *db, string tableName, PortsOrch *port) :
...
+    for (const auto &key: keys)
+    {
+        std::vector<FieldValueTuple> values;
+
+        if (!table.get(key, values))
+        {
+            continue;
+        }
+
+        string s = tableName + ":" + key
```
zhenggen-xu: It would be better to not hardcode the separator.

orchagent/fdborch.cpp
```diff
@@ -36,6 +36,56 @@ FdbOrch::FdbOrch(DBConnector *db, string tableName, PortsOrch *port) :
...
+    string tableName = "ASIC_STATE:SAI_OBJECT_TYPE_FDB_ENTRY";
+    DBConnector *db = new DBConnector(ASIC_DB, DBConnector::DEFAULT_UNIXSOCKET, 0);
+    swss::Table table(db, tableName);
+    vector<string> keys;
+
+    table.getKeys(keys);
+    for (const auto &key: keys)
+    {
+        sai_object_id_t bridge_port_id = SAI_NULL_OBJECT_ID;
+        sai_fdb_entry_type_t entryType = SAI_FDB_ENTRY_TYPE_STATIC;
```
prsunny: IMO, the default can by DYNAMIC and check for STATIC in _if_ statement down considering most of the FDBs learnt are dynamic

orchagent/fdborch.cpp
```diff
@@ -36,6 +36,56 @@ FdbOrch::FdbOrch(DBConnector *db, string tableName, PortsOrch *port) :
...
+        std::string value;
+
+        table.hget(key, "SAI_FDB_ENTRY_ATTR_TYPE", value);
+        if (value ==  "SAI_FDB_ENTRY_TYPE_DYNAMIC")
```
prsunny: Extra space after ==

jipanyang:
@prsunny thanks for the review!   FDB sync up will require more sophisticated handling.  This version of change doesn't take FDB restore into account and assumes that no previous dynamic FDB   exists.   Besides new FDB entries,  there might be stale entries as well. 

Will prepare a new version to handle the case of stale FDB entries.

qiluo-msft:
Closed, in favor of https://github.com/Azure/sonic-swss/pull/615
Thanks for the inspiration.

PR #561, opened 2018-08-02
[aclorch]: only bind to port for ACL_TABLE_PFCWD type

sihuihan88:

**What I did**
Only bind to port for ACL_TABLE_PFCWD type
**Why I did it**
PFCWD only requires to bind to port regardless if port is a LAG member. And LAG binding is not supported for egress acl table which will be used for PFCWD.
**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1022,7 +1022,12 @@ bool AclTable::create()
 
     sai_attribute_t attr;
     vector<sai_attribute_t> table_attrs;
-    vector<int32_t> bpoint_list = type == ACL_TABLE_PFCWD ? { SAI_ACL_BIND_POINT_TYPE_PORT} : { SAI_ACL_BIND_POINT_TYPE_PORT, SAI_ACL_BIND_POINT_TYPE_LAG };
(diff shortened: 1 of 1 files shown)


orchagent/aclorch.cpp
```diff
@@ -1022,7 +1022,7 @@ bool AclTable::create()
 
     sai_attribute_t attr;
     vector<sai_attribute_t> table_attrs;
-    vector<int32_t> bpoint_list = { SAI_ACL_BIND_POINT_TYPE_PORT, SAI_ACL_BIND_POINT_TYPE_LAG };
+    vector<int32_t> bpoint_list = type == ACL_TABLE_PFCWD ? { SAI_ACL_BIND_POINT_TYPE_PORT} : { SAI_ACL_BIND_POINT_TYPE_PORT, SAI_ACL_BIND_POINT_TYPE_LAG };
```
prsunny: If LAG binding is not supported on egress, will it be better to check the stage and select the binding point?

sihuihan88: offline discussed with guohan, currently only brcm pfcwd use egress acl and only brcm might have egress lag bind issue. Other platforms might still be available to use egress LAG binding. So it's better to restrict port binding for PFCWD which is also by design for this use case.

PR #562, opened 2018-08-03
Warm reboot: Add support for orchagent pre-shutdown warm-restart state check

jipanyang:

**What I did**
Before stopping orchagent for warm restart, basic state check is preferred to ensure orchagent is not in transient state,  so a deterministic state may be restored after restart.

Here is to implement orchagent_restart_check binary which may talk to orchagent and ask it to do self-check, return  "READY " signal and freeze if everything is ok, otherwise "NOT_READY" signal should be returned.

The exact condition that is treated as restart ready need further discussion. For now,  checking no un-met dependency in orchagent is used as example to verify the end to end flow. The other possible check is to make sure all config data to ASIC DB has been flushed.

**Why I did it**

**How I verified it**
VS test:
```
jipan@sonic-build-2:~/warm_reboot/sonic-buildimage/src/sonic-swss/tests$ sudo pytest -v --dvsname=vs  test_warm_reboot.py
[sudo] password for jipan: 
======================================================================= test session starts =======================================================================
cachedir: .cache
rootdir: /home/jipan/warm_reboot/sonic-buildimage/src/sonic-swss/tests, inifile:
collected 1 item                                                                                                                                                  

test_warm_reboot.py::test_OrchagentWarmRestartReadyCheck PASSED                                                                                             [100%]

```
**Details if related**


lguohan:
can you resolve conflict?

qiluo-msft:
I believe current implementation and roadmap will allow unplanned orchagent warm start, so this PR may be not necessary. I agree it is conservative and super safe condition for a warm start.

jipanyang:
@qiluo-msft  With unplanned restart, the program shutdown could happen at any point of orchagent execution,  if you are confident that current implementation is able to handle all the scenarios,  it is great.  Supporting unplanned warm restart is the ultimate goal for each component of SONiC as we have discussed at the very beginning of warm reboot development.

I'll be happy to see everything works without the planned shutdown, or at least all the potential problems could be exposed and fixed accordingly  :)

orchagent/orchdaemon.cpp
```diff
@@ -4,6 +4,7 @@
 #include "logger.h"
 #include <sairedis.h>
 #include "warm_restart.h"
+#include <limits.h>
```
qiluo-msft: Move to above since it is a standard header.

orchagent/orchdaemon.cpp
```diff
@@ -329,6 +330,26 @@ void OrchDaemon::start()
...
+         * the existence of finer granularity ExecutableTimer with select
+         */
+        if (gSwitchOrch->checkRestartReady())
+        {
+            bool ret = warmRestartCheck();
+            if (ret)
+            {
+                // Orchagent is ready to perform warm restart, stop processing any new db data.
+                // Should sleep here or continue handling timers and etc.??
+                if (!gSwitchOrch->checkRestartNoFreeze())
```
qiluo-msft: >checkRestartNoFreeze [](start = 34, length = 20)

What the use case for gSwitchOrch->checkRestartReady() == true and gSwitchOrch->checkRestartNoFreeze() == true?
I think one checker is good enough?

orchagent/orchdaemon.cpp
```diff
@@ -329,6 +330,26 @@ void OrchDaemon::start()
...
+                {
+                    SWSS_LOG_WARN("Orchagent is frozen for warm restart!");
+                    sleep(UINT_MAX);
```
qiluo-msft: >sleep(UINT_MAX); [](start = 20, length = 16)

Can we just exit with some error code?

orchagent/orchdaemon.cpp
lguohan: checkrestartready to tell orchagent to do warmRestartCheck.

checkrestartnofreeze to tell orchagent to do freeze or not.

lguohan:
code looks good to me. but i would like to model where the tool send request, if orchagent not ready, it will wait till orchagent is ready and then the tool is unblock.

orchagent/orchdaemon.cpp
jipanyang: done

orchagent/orchdaemon.cpp
jipanyang: Exiting orchagent may trigger unnecessary supervisord warning.  Keeping sleep() behavior to avoid that.

orchagent/orchdaemon.cpp
qiluo-msft: I can understand the name. I am asking why need check warm start but not freeze?

orchagent/orchdaemon.cpp
qiluo-msft: It is easy to add supervisor config with some expected return code. I am asking whether this is the right thing to do.

jipanyang: What is the problem for orchagent to sleep() and wait for official shutdown signal with other processes?
I don't see the need to perform supervisor change here, which is not necessary.

tests/test_warm_reboot.py
```diff
@@ -683,3 +682,58 @@ def test_swss_neighbor_syncup(dvs):
...
+    time.sleep(1)
+    (exitcode, result) =  dvs.runcmd("/usr/bin/orchagent_restart_check")
+    assert result == "RESTARTCHECK succeeded\n"
+
+    # Should fail since orchagent has been frozen at last step.
+    (exitcode, result) =  dvs.runcmd("/usr/bin/orchagent_restart_check -n -s -w 500")
+    assert result == "RESTARTCHECK failed\n"
+
+    # recover for test cases after this one.
+    stop_swss(dvs)
```
qiluo-msft: >stop_swss [](start = 4, length = 9)

Not defined. It will break vs test.

jipanyang:
@qiluo-msft if possible, could https://github.com/Azure/sonic-swss/pull/557 be merged first.  The vs test for the PR uses start_swss(dvs) & stop_swss(dvs) defined there.

I could replicate  start_swss(dvs) & stop_swss(dvs) to this PR too if this is preferred.

qiluo-msft:
Sure. Check https://github.com/Azure/sonic-swss/pull/557 first

qiluo-msft:
Help solve conflict?

jipanyang:
done.  It looks VS test env has run mad.

qiluo-msft:
Retest this please

orchagent/orchagent_restart_check.cpp
```diff
@@ -0,0 +1,145 @@
...
+ *            if --noFreeze option is provided, orchagent won't freeze.
+ *            if --skipPendingTaskCheck option is provided, orchagent won't use
+ *                 whether there is pending task existing as state check criterion.
+ */
+int main(int argc, char **argv)
+{
+    swss::Logger::getInstance().setMinPrio(swss::Logger::SWSS_INFO);
+    SWSS_LOG_ENTER();
+
+    std::string skipPendingTaskCheck = "fasle";
```
qiluo-msft: Could you please fix the typo?

jipanyang: Ok.   Fortunately it is not causing problem due to value "true" is checked 
 https://github.com/Azure/sonic-swss/blob/master/orchagent/switchorch.cpp#L179

PR #556, opened 2018-08-01
Port table should be ProducerTable

qiluo-msft:
The entries in APP_PORT_TABLE_NAME are sequence sensitive because there are event entries (PortConfigDone, PortInitDone). Change it back to ProducerTable.

PR #564, opened 2018-08-04
Revert "Port table should be ProducerTable"

qiluo-msft:
Reverts Azure/sonic-swss#556


qiluo-msft:
Discussed with @zhenggen-xu , this fix is not complete, m_toSync in orchagent will also merge operations.

PR #565, opened 2018-08-06
[portsyncd]: Skip empty lines

stcheng:
If the length of the line is 0, skip parsing this line.

### DIFF
--- a/portsyncd/portsyncd.cpp
+++ b/portsyncd/portsyncd.cpp
@@ -205,6 +205,11 @@ void handlePortConfigFile(ProducerTable &p, string file)
     string line;
     while (getline(infile, line))
     {
+        if (line.size() == 0)
+        {
+            continue;
+        }
+
         if (line.at(0) == '#')
         {
             // Take this line as column header line


portsyncd/portsyncd.cpp
```diff
@@ -205,6 +205,11 @@ void handlePortConfigFile(ProducerTable &p, string file)
     string line;
     while (getline(infile, line))
     {
+        if (line.size() == 0)
```
pavel-shirshov: what if line has a space in it?

jleveque: Every line in port_config.ini should contain spaces, as they are space-delimited. What is your concern? This code should only skip empty lines. Lines containing spaces should have a length > 0.

Are you asking about lines which *only* contain whitespace?

pavel-shirshov: If a line has a length equal to 1 will that line work for us?
Should we better skip a line which contains from 0 to many spaces?

qiluo-msft: This is a feature/fix requested by me. Normally I prefer a config file end with a new line character.

stcheng:
closed

PR #566, opened 2018-08-07
Improve PortsOrch::doPortTask()

qiluo-msft:
(Reduce the scope of this PR, please check the updated comment below)

I tested with virtual switch (vs) and Mellanox platform. The initial config of 'fec' is added in ```/usr/share/sonic/device/*/*/port_config.ini```. However this manual test is not materialized as a unit test because of some difficulty.

Original implementation of PortsOrch::doPortTask() has some issues:

1. ~~It was using a ConsumerStateTable, which makes the message coming out of order. The complex logic was used to handle the order. After replace ConsumerStateTable by ConsumerTable, we can simplify the logic.~~

2. The original ConsumerStateTable implementation will always pop all entry attributes even there is one attribute coming. This behavior is not by design and was changed after https://github.com/Azure/sonic-swss-common/pull/204 merged. The  PortsOrch::doPortTask() takes advantage of this specific behavior and doesn't handle some port attributes correctly, if they come before PortConfigDone. These attributes are m_autoneg, speed, mtu, admin_status, fec_mode.

This PR solve these issues, and add vs test case.

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1177,6 +1177,96 @@ bool PortsOrch::initPort(const string &alias, const set<int> &lane_set)
     return true;
 }
 
+void PortsOrch::doPortConfigDoneTask(const vector<FieldValueTuple>& tuples)
+{
+    SWSS_LOG_ENTER();
+
+    m_portConfigDone = true;
+
+    for (auto i : tuples)
+    {
+        if (fvField(i) == "count")
+        {
+            m_portCount = to_uint<uint32_t>(fvValue(i));
+        }
+    }
+
+    if (m_lanesAliasSpeedMap.size() != m_portCount)
+    {
+        SWSS_LOG_ERROR("Unexpected number of PORTs received before PortConfigDone: %zu != %u", m_lanesAliasSpeedMap.size(), m_portCount);
+        return;
+    }
+
+    /* Once all ports received, go through the each port and perform appropriate actions:
+     * 1. Remove ports which don't exist anymore
+     * 2. Create new ports
+     * 3. Initialize all ports
+     */
+    for (auto it = m_portListLaneMap.begin(); it != m_portListLaneMap.end();)
+    {
+        if (m_lanesAliasSpeedMap.find(it->first) == m_lanesAliasSpeedMap.end())
+        {
+            char *platform = getenv("platform");
+            if (platform && (strstr(platform, BFN_PLATFORM_SUBSTRING) || strstr(platform, MLNX_PLATFORM_SUBSTRING)))
+            {
+                if (!removePort(it->second))
+                {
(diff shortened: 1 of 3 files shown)


jipanyang:
If my understanding is correct,  with the change, no initPort() will be called until PortConfigDone is received.  The code logic below will block the au, speed, fec setting from being applied before PortConfigDone.  Not sure how they will be applied later.

Also gBufferOrch->isPortReady(alias) brings more dynamics here.

```c++
            if (!gBufferOrch->isPortReady(alias))
            {
                // buffer configuration hasn't been applied yet. save it for future retry
                it++;
                continue;
            }

            Port p;
            if (!getPort(alias, p))
            {
                SWSS_LOG_ERROR("Failed to get port id by alias:%s", alias.c_str());
            }
```

Did you test on hardware platform with some au, fec setting in the initial config?

qiluo-msft:
I could confirm "with the change, no initPort() will be called until PortConfigDone is received". Is there an issue here?

"gBufferOrch->isPortReady(alias)" will block these settings. The 'it' pointed entry is not erased and remain in consumer.m_toSync. Anytime unblocked, that entry will be applied. It is a kind of postponed.

I tested with vs. The initial config of 'fec' is added in /usr/share/sonic/device/x86_64-dell_s6000_s1220-r0/Force10-S6000/port_config.ini.

---
In reply to: [410928210](https://github.com/Azure/sonic-swss/pull/566#issuecomment-410928210) [](ancestors = 410928210)

jipanyang:
What if there is no buffer pg and queue configuration in configDB,  will "gBufferOrch->isPortReady(alias)" be true and not block it?

jipanyang:
Another question is:   previously the hostif will be created whenever the port config request arrives, linux kernel generates netlink message after that, and portsyncd responds to the netlink message accordingly.  

With this change all the ports config requests are buffered together then hostifs are created in batch.  Since this happens at system init, have you evaluated the impact like boot up speed, possible netlink throttling?

qiluo-msft:
The behavior about "gBufferOrch->isPortReady(alias)" is unchanged.

---
In reply to: [411042467](https://github.com/Azure/sonic-swss/pull/566#issuecomment-411042467) [](ancestors = 411042467)

jipanyang:
The existing port config logic doesn't have hard dependency on gBufferOrch->isPortReady(alias), but more relying on PortConfigDone check. Now your whole solution relies on this gBufferOrch->isPortReady(alias) check,  at least a clear explanation and understanding of the logic is needed, more comments are needed to record the complex interdependency of PortInitDone, isPortReady and port configuration. 

It looks to me, one line change in existing code will have the same effect of solving the problems you mentioned :

```c++
            if (!m_portConfigDone)
            {
                // it = consumer.m_toSync.erase(it);
                it++;
                continue;
            }
```

qiluo-msft:
Temporarily close

PR #218, opened 2017-05-18
Add MTU set on port (issue #77)

andriymoroz-mlnx:
portsyncd synchronizes net device settings with the PORT_TABLE
Added code to set MTU to hardware on PORT_TABLE update

PR #570, opened 2018-08-09
swss: flush g_asicState after each event is done

dzhangalibaba:
* add flush() after event is handled in case some entries are still in buffer, don't wait
* with the changes in sairedis and swss-common, route performance improved by 200~300 routes/sec

  Signed-off-by: Dong Zhang d.zhang@alibaba-inc.com

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -313,5 +313,6 @@ void OrchDaemon::start()
         for (Orch *o : m_orchList)
             o->doTask();
 
+        flush(); //flush after each event is handled, don't wait
     }
 }
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -313,6 +313,7 @@ void OrchDaemon::start()
         for (Orch *o : m_orchList)
             o->doTask();
 
-        flush(); //flush after each event is handled, don't wait
+        //flush after each event is handled, don't wait timeout
+        flush();
     }
 }
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -294,12 +294,6 @@ void OrchDaemon::start()
 
         if (ret == Select::TIMEOUT)
         {
-            /* Let sairedis to flush all SAI function call to ASIC DB.
-             * Normally the redis pipeline will flush when enough request
-             * accumulated. Still it is possible that small amount of
-             * requests live in it. When the daemon has nothing to do, it
-             * is a good chance to flush the pipeline  */
-            flush();
             continue;
         }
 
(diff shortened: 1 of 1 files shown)


lguohan:
you have a few PR submitted in sairedis, common, can you give dependencies on these PR? are they all independent or some pr requires other pr to be merge first?

dzhangalibaba:
swss-common: add empty() member func in ComsumerTableBase class which will be used in syncd #218 ,   this one should be merged first.  sairedis#335 depends on swss-common#218. This two is one feeature.  The other three, swss#570 , swss-common#219  and sairedis#336  are one feature, but they are independent in compiling point of view.

orchagent/orchdaemon.cpp
```diff
@@ -313,5 +313,6 @@ void OrchDaemon::start()
         for (Orch *o : m_orchList)
             o->doTask();
 
+        flush(); //flush after each event is handled, don't wait
```
qiluo-msft: >//flush after e [](start = 17, length = 15)

Please help follow the coding style of other comments.

qiluo-msft:
>             flush();

So the flush() is not needed at all?

---
Refers to: orchagent/orchdaemon.cpp:302 in 8fa353c. [](commit_id = 8fa353c6a60ec6f0964293224259ee3ad9038404, deletion_comment = False)

orchagent/orchdaemon.cpp
qiluo-msft: >flush [](start = 8, length = 5)

About the performance "route performance improved by 200~300 routes/sec". Could you provide more details?
1. What is the test environment and test steps?
2. What is the performance before this PR and after?
3. If it is a small scope test, could you add a unit test or vs test case to automate it?

dzhangalibaba: Sure. Will do

dzhangalibaba:
Also , the flush() is needed, currently it will flush() when there is nothing to select, the SELECT_WAIT time is 1000ms, this is too long to routes, we need to flush them right away, don't wait until timeout

qiluo-msft:
I mean the above flush() for timeout, since you already flush() after execute().

dzhangalibaba:
got it, then that flush is not necessary, we can remove it.

orchagent/orchdaemon.cpp
qiluo-msft: OK. I just want to know performance before this PR and after? You only mentioned improved by 200~300 routes/sec. Rough numbers are ok.

dzhangalibaba: before changes it is 1300 routes/sec, platform is brcm and SAI is 3.1, CPU is Intel(R) Atom(TM) CPU  C2558  @ 2.40GHz. After this enabling pipeline changes only, it is about 1500-1600 routes/sec, if plus the syncd buffer changes in other PR, it is about 1700-1800 routes/sec

PR #571, opened 2018-08-09
Expose the capability to config wred drop probability

wendani:
Expose the capability to config wred drop probability of all packet colors (green, yellow, and red)

**What I did**

**Why I did it**

**How I verified it**
Tested on dut 3 cases:
1) drop probabilities all specified in qos.json
2) drop probabilities partly specified in qos.json
3) no drop probabilities specified in qos.json

**Details if related**

### DIFF
--- a/orchagent/qosorch.cpp
+++ b/orchagent/qosorch.cpp
@@ -329,6 +329,24 @@ bool WredMapHandler::convertFieldValuesToAttributes(KeyOpFieldsValuesTuple &tupl
             attr.value.s32 = stoi(fvValue(*i));
             attribs.push_back(attr);
         }
+        else if (fvField(*i) == green_drop_probability_field_name)
+        {
+            attr.id = SAI_WRED_ATTR_GREEN_DROP_PROBABILITY;
+            attr.value.s32 = stoi(fvValue(*i));
+            attribs.push_back(attr);
+        }
+        else if (fvField(*i) == yellow_drop_probability_field_name)
+        {
+            attr.id = SAI_WRED_ATTR_YELLOW_DROP_PROBABILITY;
+            attr.value.s32 = stoi(fvValue(*i));
+            attribs.push_back(attr);
+        }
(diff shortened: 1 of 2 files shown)


orchagent/qosorch.cpp
```diff
@@ -395,14 +413,6 @@ sai_object_id_t WredMapHandler::addQosItem(const vector<sai_attribute_t> &attrib
...
     vector<sai_attribute_t> attrs;
 
-    attr.id = SAI_WRED_ATTR_GREEN_DROP_PROBABILITY;
-    attr.value.s32 = 100;
-    attrs.push_back(attr);
-
-    attr.id = SAI_WRED_ATTR_YELLOW_DROP_PROBABILITY;
-    attr.value.s32 = 100;
-    attrs.push_back(attr);
-
```
yxieca: This chunk of code looks like providing some default values for these settings. By removing them, you also removed the 'default' value.

I don't see other enforcement to make sure we pass down the configuration all the time. Maybe that belongs to a different PR?

If 'default' value is desirable, then you can create 2 attr objects for them, set the default value before going through the if - else lists. But don't add them until the whole if-else block is done. like following:

type attr_green_drop;
attr_green_drop.s32 = 100;

... ...
if ...
else if ...
...

attrs.push(attr_green_drops);

@lguohan do we need to do this?

wendani: Hardware has a default value if you do not set

lguohan: I agree with @yxieca , the pr can be modified as follows. if the user does not provide drop probability, we should set the default value in swss to be 100. In this case, your other pr is not needed.

wendani:
retest this please

wendani:
the vs check is broken. It used to succeed before retest.

wendani:
retest this please

wendani:
retest this please

lguohan:
retest this please

lguohan:
retest this please

PR #574, opened 2018-08-10
[intfsorch]: Add setRouterIntfsMtu function

stcheng:
This function allows the propagation of MTU from port MTU to router
interface MTU if the router interface is created based on this port.

Once the port MTU gets changes, the router interface MTU will also
get changed.

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -18,8 +18,8 @@ extern sai_router_interface_api_t*  sai_router_intfs_api;
 extern sai_route_api_t*             sai_route_api;
 extern sai_neighbor_api_t*          sai_neighbor_api;
 
-extern PortsOrch *gPortsOrch;
 extern sai_object_id_t gSwitchId;
+extern PortsOrch *gPortsOrch;
 extern CrmOrch *gCrmOrch;
 extern BufferOrch *gBufferOrch;
 
@@ -55,6 +55,27 @@ void IntfsOrch::decreaseRouterIntfsRefCount(const string &alias)
     m_syncdIntfses[alias].ref_count--;
     SWSS_LOG_DEBUG("Router interface %s ref count is decreased to %d",
                   alias.c_str(), m_syncdIntfses[alias].ref_count);
+}
+
+bool IntfsOrch::setRouterIntfsMtu(Port &port)
+{
+    SWSS_LOG_ENTER();
+
+    sai_attribute_t attr;
+    attr.id = SAI_ROUTER_INTERFACE_ATTR_MTU;
+    attr.value.u32 = port.m_mtu;
+
+    sai_status_t status = sai_router_intfs_api->
+            set_router_interface_attribute(port.m_rif_id, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to set router interface %s MTU to %u, rv:%d",
+                port.m_alias.c_str(), port.m_mtu, status);
+        return false;
+    }
+    SWSS_LOG_NOTICE("Set router interface %s MTU to %u",
+            port.m_alias.c_str(), port.m_mtu);
+    return true;
 }
 
 void IntfsOrch::doTask(Consumer &consumer)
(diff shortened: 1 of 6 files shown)


orchagent/intfsorch.cpp
```diff
@@ -57,6 +57,27 @@ void IntfsOrch::decreaseRouterIntfsRefCount(const string &alias)
...
+
+    sai_status_t status = sai_router_intfs_api->
+            set_router_interface_attribute(port.m_rif_id, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to set router interface %s MTU to %u, rv:%d",
+                port.m_alias.c_str(), mtu, status);
+        return true;
+    }
+    SWSS_LOG_NOTICE("Set router interface %s MTU to %u",
```
prsunny: I think this could be INFO

orchagent/intfsorch.cpp
```diff
@@ -57,6 +57,27 @@ void IntfsOrch::decreaseRouterIntfsRefCount(const string &alias)
...
+    attr.id = SAI_ROUTER_INTERFACE_ATTR_MTU;
+    attr.value.u32 = mtu;
+
+    sai_status_t status = sai_router_intfs_api->
+            set_router_interface_attribute(port.m_rif_id, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to set router interface %s MTU to %u, rv:%d",
+                port.m_alias.c_str(), mtu, status);
+        return true;
```
prsunny: return False.

orchagent/intfsorch.cpp
```diff
@@ -57,6 +57,27 @@ void IntfsOrch::decreaseRouterIntfsRefCount(const string &alias)
...
+    }
+    SWSS_LOG_NOTICE("Set router interface %s MTU to %u",
+            port.m_alias.c_str(), mtu);
+    return false;
```
prsunny: return true

orchagent/portsorch.cpp
```diff
@@ -1532,6 +1534,10 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         p.m_mtu = mtu;
                         m_portList[alias] = p;
                         SWSS_LOG_NOTICE("Set port %s MTU to %u", alias.c_str(), mtu);
+                        if (!p.m_rif_id)
+                        {
+                            gIntfsOrch->setRouterIntfsMtu(p, mtu);
```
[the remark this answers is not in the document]
stcheng: thanks for catching this. i'll update it.

orchagent/intfsorch.cpp
stcheng: i'd like to keep it as NOTICE since it's possible to change port channel MTUs or VLAN mtus and we'd like to know the corresponding MTU changes for the router interfaces.

orchagent/portsorch.cpp
```diff
@@ -1532,6 +1534,10 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         p.m_mtu = mtu;
                         m_portList[alias] = p;
                         SWSS_LOG_NOTICE("Set port %s MTU to %u", alias.c_str(), mtu);
+                        if (!p.m_rif_id)
```
jipanyang: if (p.m_rif_id)    ?

stcheng: thanks. updated.

lguohan:
code looks good. please add vs test.

jipanyang:
Change looks good to me, now the ethernet port router intf mtu setting should work as before during system init.
 
Just curious, since in https://github.com/Azure/sonic-swss/pull/545/files you also have lag mtu setting,  are you going to address that in orchagent for lag mtu in separate PR?

stcheng:
@jipanyang yes. since the current implementation also doesn't support LAG MTU changes. I'll do it in the separate PRs.

tests/test_interface.py
```diff
@@ -43,12 +60,40 @@ def test_InterfaceAddIpv4Address(self, dvs):
...
         assert subnet_found and ip2me_found
 
+    def test_InterfaceChangeMtu(self, dvs):
+        pdb = swsscommon.DBConnector(0, dvs.redis_sock, 0)
+        adb = swsscommon.DBConnector(1, dvs.redis_sock, 0)
+        cdb = swsscommon.DBConnector(4, dvs.redis_sock, 0)
+
+        tbl = swsscommon.ProducerStateTable(pdb, "PORT_TABLE")
+        fvs = swsscommon.FieldValuePairs([("mtu", "8888")])
+        tbl.set("Ethernet8", fvs)
```
lguohan: do we need to configure ethernet8 into rif or it is already in rif.

stcheng: in the first test case when the ip is assigned, the associated rif is created

PR #578, opened 2018-08-15
[test]: Change the default router interface MTU to 9100

stcheng:
With the recent pull requests, if the router interface is created
without specifying the MTU in the configuration database, the
default MTU would be 9100 instead of 1500.

Note:
9100 comes from the orchagent/port.h; 1500 comes from the kernel.


lguohan:
retest this please

PR #579, opened 2018-08-16
Warm reboot: Fix race condition and avoid unnecessary create/delete nexthop_group_…

jipanyang:
…member

**What I did**

**Why I did it**
<1> Fix the possible race condition when port is down, but previously buffered neigbor and next hop group events are processed later.

<2> Remove the unnecessary create_next_hop_group_member()/remove_next_hop_group_member() operations for nexthop on down port which may cause traffic interruption.

<3> The change also happens to fix the state restore for ecmp acceleration during warm reboot.
The scenario:
   -  Port is down,  NHFLAGS_IFDOWN is set on the next hop and the next_hop_group_member_id has been removed from the next hop group.    
   - system/orchagent shutdown for warm start.
   - orchagent state restore with certain port oper status down, nexthop(neighbor), nexthop group will be restored from appDB.

**How I verified it**
VS test.
**Details if related**

### DIFF
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -30,6 +30,14 @@ bool NeighOrch::addNextHop(IpAddress ipAddress, string alias)
 {
     SWSS_LOG_ENTER();
 
+    Port p;
+    if (!gPortsOrch->getPort(alias, p))
+    {
+        SWSS_LOG_ERROR("Neighbor %s seen on port %s which doesn't exist",
+                        ipAddress.to_string().c_str(), alias.c_str());
+        return false;
+    }
+
     assert(!hasNextHop(ipAddress));
     sai_object_id_t rif_id = m_intfsOrch->getRouterIntfsId(alias);
 
@@ -78,6 +86,18 @@ bool NeighOrch::addNextHop(IpAddress ipAddress, string alias)
         gCrmOrch->incCrmResUsedCounter(CrmResourceType::CRM_IPV6_NEXTHOP);
     }
 
+    // For nexthop with incoming port which has down oper status, NHFLAGS_IFDOWN
+    // flag Should be set on it.
+    // This scenario may happen under race condition where buffered neighbor event
+    // is processed after incoming port is down.
+    if (p.m_oper_status == SAI_PORT_OPER_STATUS_DOWN)
+    {
+        if (setNextHopFlag(ipAddress, NHFLAGS_IFDOWN) == false)
+        {
+            SWSS_LOG_WARN("Failed to set NHFLAGS_IFDOWN on nexthop %s for interface %s",
+                ipAddress.to_string().c_str(), alias.c_str());
+        }
+    }
     return true;
 }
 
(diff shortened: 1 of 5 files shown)


lguohan:
retest this please

orchagent/neighorch.cpp
```diff
@@ -30,6 +30,14 @@ bool NeighOrch::addNextHop(IpAddress ipAddress, string alias)
 {
     SWSS_LOG_ENTER();
 
+    Port p;
+    if (!gPortsOrch->getPort(alias, p))
```
nikos-github: The port doesn't exist because...?

orchagent/routeorch.cpp
```diff
@@ -564,6 +563,11 @@ bool RouteOrch::addNextHopGroup(IpAddresses ipAddresses)
 
     for (auto nhid: next_hop_ids)
     {
+        // skip next hop group member create for neighbor from down port
+        if (m_neighOrch->isNextHopFlagSet(nhopgroup_members_set[nhid], NHFLAGS_IFDOWN)) {
```
nikos-github: The code was put intentionally at the end and not here. I'd like to understand the issue you were seeing more and I will also explain the logic why the code is at the end so lets discuss offline.

orchagent/neighorch.cpp
jipanyang: Since PortsOrch::getPort() has return value, it is a safety check.

orchagent/routeorch.cpp
jipanyang: Right,  I went through the current implementation.  The next_hop_id was created then removed, but the value is saved which is only used again when the removeNextHopGroup().  orch agent needs to have a copy of nexthop which it already has, but probably not the obsoleted next_hop_id.

It is possible I missed some of the key point. We may have offline discussion, then provide more details here.
btw, I'm glad you are back :)

orchagent/neighorch.cpp
nikos-github: I would recommend to assert then.

orchagent/routeorch.cpp
nikos-github: Unless I missed some other change in your diffs, with this one here I believe we will be leaking nexthop groups.

orchagent/portsorch.cpp
```diff
@@ -2098,6 +2099,11 @@ bool PortsOrch::initializePort(Port &p)
         FieldValueTuple tuple("oper_status", "down");
         vector.push_back(tuple);
         m_portTable->set(p.m_alias, vector);
+        p.m_oper_status = SAI_PORT_OPER_STATUS_DOWN;
+    }
+    else
+    {
+        p.m_oper_status = SAI_PORT_OPER_STATUS_UP;
```
nikos-github: Does this match what the db has? Any concerns about the possibility of a mismatch?

jipanyang: Yes,  operStatus is the value read from appDB.

https://github.com/Azure/sonic-swss/blob/8bfdea086f14bbe420846d42c607760999cbdcc3/orchagent/portsorch.cpp#L2093
p.m_oper_status is a mirror of what set in DB.

PR #581, opened 2018-08-17
[portsorch]: Support LAG MTU configuration

stcheng:
- LAG MTU configuration is now on stage

- By default, the MTU of port channel based router interface is 9100.

- If a new MTU is applied to the port channel via configuration
  database, this new MTU will be used 1) to the netdev in the kernel
  via ip command; 2) to all the member netdevs in the kernel via teamd;
  3) to the port channel based router interface MTU; 4) to all the
  member port MTU.

- Remove the libnl MTU from teamsyncd

- Refactor test_interface.py tests. Each one test is independent from
  the other test. The switch will be cleaned up at the end of the test.

### DIFF
--- a/cfgmgr/portmgr.cpp
+++ b/cfgmgr/portmgr.cpp
@@ -1,4 +1,5 @@
 #include <string>
+
 #include "logger.h"
 #include "dbconnector.h"
 #include "producerstatetable.h"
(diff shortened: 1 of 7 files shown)


stcheng:
note: port channel member manipulation is not part of this pull request. MTU changes associated with membership changes will be in the future pull requests.

cfgmgr/portmgr.cpp
```diff
@@ -16,17 +17,57 @@ PortMgr::PortMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
...
+        FieldValueTuple fv("mtu", mtu);
+        fvs.push_back(fv);
+        m_appLagTable.set(alias, fvs);
+
+        m_cfgLagTable.get(alias, fvs);
+        for (auto fv: fvs)
+        {
+            // Set the port channel members MTU in application database
+            // to update the port MTU in orchagent
+            if (fvField(fv) == "members")
```
jipanyang: Just a suggestion, using Table::hset() and Table::hget could reduce the lines of code greatly.

lguohan:
retest this please

lguohan:
retest this please

lguohan:
can you fix the test failure?

stcheng:
the test failure is due to the old docker-vs
the old docker-vs doesn't have the portmgrd causing this test to fail
since https://github.com/Azure/sonic-buildimage/pull/1908 is merged, the new docker-vs should be able to run and pass all the tests

cfgmgr/portmgr.cpp
```diff
@@ -16,17 +17,57 @@ PortMgr::PortMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
...
+            {
+                for (auto member : tokenize(fvValue(fv), ','))
+                {
+                    vector<FieldValueTuple> member_fvs;
+                    FieldValueTuple member_fv("mtu", mtu);
+                    member_fvs.push_back(member_fv);
+                    m_appPortTable.set(member, member_fvs);
```
prsunny: If the port mtu is overwritten by the lag mtu value, what happens when the port is removed from lag later? Is it handled?

tests/test_interface.py
```diff
@@ -85,77 +118,181 @@ def test_InterfaceChangeMtu(self, dvs):
...
+    # database updates after the future changes of the lagmgrd
+    def create_port_channel(self, dvs, interface):
+        dvs.runcmd("teamd -d -t" + interface)
+        time.sleep(1)
+
+    def remove_port_channel(self, dvs, interface):
+        dvs.runcmd("teamd -k -t" + interface)
+        time.sleep(1)
+
+    def add_port_channel_members(self, dvs, interface, members):
```
prsunny: Can we also have a test case to remove_port_channel_members?

cfgmgr/portmgr.cpp
stcheng: my earlier notes in this pull request says that this won't cover the MTU changes associated with the port channel membership changes. i'll go and cover those cases when i add the support for configuring port channel memberships

tests/test_interface.py
stcheng: will add this along with port channel membership configuration changes

cfgmgr/portmgr.cpp
prsunny: Oh ok, didn't notice the notes.

lguohan:
retest this please

lguohan:
retest this please

PR #587, opened 2018-08-21
Init OrchDaemon with optional warm starting

qiluo-msft:
We don't call syncd_apply_view() at 2 places.

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -281,10 +281,7 @@ int main(int argc, char **argv)
             exit(EXIT_FAILURE);
         }
 
-        if (!WarmStart::isWarmStart())
-        {
-            syncd_apply_view();
-        }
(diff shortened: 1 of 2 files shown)


jipanyang:
Just curious,  are you gonna send apply view signal to syncd after state sync up?

qiluo-msft:
Yes.
This PR doesn't change behaviors and it is a refactoring.

jipanyang:
Current logic is to perform apply view after state restore and before state sync up. With the new change, apply view will be done after state sync up.    I'm not familiar with the latest progress of view comparison solution. Probably that is by design.

Anyway, state sync up is still at TODO status in existing code.  The change here doesn't have real impact, it looks ok to me.

qiluo-msft:
Got your idea. If needed, I will revisit this case after TODO done.

PR #588, opened 2018-08-22
[VS] update dvs.runcmd() return value check

jipanyang:

**What I did**

Fix the return value check error for dvs.runcmd().

**Why I did it**

**How I verified it**
VS test
**Details if related**


lguohan:
retest this please

PR #589, opened 2018-08-22
Add a common function to get the warm-start timer

zhenggen-xu:
**What I did**
Simplify the timer schema
Add a common function to get the warm-start timer

**Why I did it**
Use the same function for different timers.

**How I verified it**

**Details if related**

### DIFF
--- a/doc/swss-schema.md
+++ b/doc/swss-schema.md
@@ -634,17 +634,18 @@ Equivalent RedisDB entry:
     ;Stores system warm start configuration
     ;Status: work in progress
 
-    key             = WARM_RESTART:name         ; name is the name of SONiC docker or "system" for global configuration.
+    key                 = WARM_RESTART:name ; name is the name of SONiC docker or "system" for global configuration.
 
-    enable          = "true" / "false"          ; Default value as false.
-                                                ; If "system" warm start knob is true, docker level knob will be ignored.
-                                                ; If "system" warm start knob is false, docker level knob takes effect.
+    enable              = "true" / "false"  ; Default value as false.
+                                            ; If "system" warm start knob is true, docker level knob will be ignored.
+                                            ; If "system" warm start knob is false, docker level knob takes effect.
 
-    app_timer       = 1*10DIGIT,                ; app_timer is the abstract name of warm start
-                                                ; timer for the whole system or the specific docker.
-                                                ; Real name is "neighsyncd_timer", "bgp_timer" etc.
-                                                ; There could be more than one timer in a docker or the system.
-                                                ; Multiple app_timer fields are created for multiple timers.
+    neighsyncd_timer    = 1*4DIGIT          ; neighsyncd_timer is the timer used for neighsyncd during the warm restart.
+                                            ; Timer is started after we restored the neighborTable to internal data structures.
+                                            ; neighborsyncd then starts to read all linux kernel entries and mark the entries in
+                                            ; the data structures accordingly. Once the timer is expired, we will do reconciliation
+                                            ; and push the delta to appDB
+                                            ; Valid value is 1-9999. 0 is invalid.
 
 
 ## State DB schema
--- a/doc/swss-schema.md
+++ b/doc/swss-schema.md
@@ -640,11 +640,11 @@ Equivalent RedisDB entry:
                                                 ; If "system" warm start knob is true, docker level knob will be ignored.
                                                 ; If "system" warm start knob is false, docker level knob takes effect.
 
-    timer_name      = 1*255VCHAR,               ; timer_name is the name of warm start timer for the whole system or the specific docker,
-                                                ; Ex. "neighbor_timer", "bgp_timer". The name should not contain "," character.
-                                                ; There could be more than one timer in a docker or the system, separated by ",".
-
-    timer_duration  = 1*4DIGIT,                 ; timer duration, in seconds. Separated by "," if there are more than on timer.
+    app_timer       = 1*10DIGIT,                ; app_timer is the abstract name of warm start
+                                                ; timer for the whole system or the specific docker.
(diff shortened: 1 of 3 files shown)


doc/swss-schema.md
```diff
@@ -640,11 +640,11 @@ Equivalent RedisDB entry:
...
 
-    timer_name      = 1*255VCHAR,               ; timer_name is the name of warm start timer for the whole system or the specific docker,
-                                                ; Ex. "neighbor_timer", "bgp_timer". The name should not contain "," character.
-                                                ; There could be more than one timer in a docker or the system, separated by ",".
-
-    timer_duration  = 1*4DIGIT,                 ; timer duration, in seconds. Separated by "," if there are more than on timer.
+    app_timer       = 1*10DIGIT,                ; app_timer is the abstract name of warm start
+                                                ; timer for the whole system or the specific docker.
+                                                ; Real name is "neighsyncd_timer", "bgp_timer" etc.
+                                                ; There could be more than one timer in a docker or the system.
```
jipanyang: This seems contradictory to the existing practice in swss-schema.md. For all other cases, the field name is fixed and value section is dynamic.  

We need to make sure the pros outweigh the cons and it complies with ABNF spec.

lguohan:
retest this please

doc/swss-schema.md
zhenggen-xu: This is not meant for dynamic field name, it will be a fixed one if we know the application name which requires warm-start timer.  Maybe I should just put "neighsyncd_timer" and "bgp_timer" here, and we can add more later if needed, I was trying to describe what the name typically looks like.

warmrestart/warm_restart.cpp
```diff
@@ -96,6 +97,34 @@ bool WarmStart::checkWarmStart(const std::string &app_name, const std::string &d
...
+/*
+ * Get warmStartTimer for an application in a docker (default to be swss)
+ * return value 0 = OK,  -1 = failure
+ * timer value is saved to timer_value if succeed. timer value 0 is invalid.
+ */
+int WarmStart::getWarmStartTimer(uint32_t &timer_value, const std::string &app_name,
+    const std::string &docker_name)
+{
+    auto& warmStart = getInstance();
+    std::string timer_name = app_name + "_timer";
```
jipanyang: So it is assumed that there is only one timer for each application like neighsyncd, bgp and etc..  Is there exception?

warmrestart/warm_restart.cpp
```diff
@@ -96,6 +97,34 @@ bool WarmStart::checkWarmStart(const std::string &app_name, const std::string &d
...
+    unsigned long int temp_value = strtoul(timer_value_str.c_str(), NULL, 0);
+    if (temp_value != 0 && temp_value != LONG_MAX)
+    {
+        timer_value = (uint32_t)temp_value;
+        return 0;
+    }
+    else
+    {
+        SWSS_LOG_WARN("Error in getting warmStartTimer for docker: %s, app: %s",
+                docker_name.c_str(), app_name.c_str());
```
jipanyang: If the log level is warning,  Is there some code logic to put the default timer value in DB?

warmrestart/warm_restart.cpp
```diff
@@ -96,6 +97,34 @@ bool WarmStart::checkWarmStart(const std::string &app_name, const std::string &d
...
+    const std::string &docker_name)
+{
+    auto& warmStart = getInstance();
+    std::string timer_name = app_name + "_timer";
+    std::string timer_value_str;
+
+    warmStart.m_cfgWarmRestartTable->hget(docker_name, timer_name, timer_value_str);
+
+    unsigned long int temp_value = strtoul(timer_value_str.c_str(), NULL, 0);
+    if (temp_value != 0 && temp_value != LONG_MAX)
```
nikos-github: Is there error checking done when configuring timer_value_str? If not, a value with a - will result in wraparound and give a valid integer when in fact the value is invalid. Also, LONG_MAX is quite big a restarting value. You should consider limiting the range.

warmrestart/warm_restart.cpp
nikos-github: You should set timer_value to 0 or to default. Do we allow different defaults per application?

doc/swss-schema.md
nikos-github: Can you please point me to a design document?

warmrestart/warm_restart.cpp
zhenggen-xu: Yes,  the CLI part will do the error checking when write to the configDB. 
strtoul return 0 or ULONG_MAX as invalid values.  Range check is added.

warmrestart/warm_restart.cpp
zhenggen-xu: No, one default per application.

doc/swss-schema.md
```diff
@@ -634,17 +634,18 @@ Equivalent RedisDB entry:
...
-                                                ; There could be more than one timer in a docker or the system, separated by ",".
-
-    timer_duration  = 1*4DIGIT,                 ; timer duration, in seconds. Separated by "," if there are more than on timer.
+    key                 = WARM_RESTART:name ; name is the name of SONiC docker or "system" for global configuration.
+
+    enable              = "true" / "false"  ; Default value as false.
+                                            ; If "system" warm start knob is true, docker level knob will be ignored.
+                                            ; If "system" warm start knob is false, docker level knob takes effect.
+
+    neighsyncd_timer    = 1*4DIGIT          ; neighsyncd_timer is the timer used for neighsyncd during the warm restart.
```
jipanyang: This is a little annoying that neighsyncd_timer seems possible to exist under key of WARM_RESTART:bgp.

lguohan: agree

PR #590, opened 2018-08-23
Tune consumer order in BufferOrch::doTask() based on dependency for warm start

qiluo-msft:
Tested in VS:
1. cold start
2. change config DB
```
127.0.0.1:6379[4]> hset WARM_RESTART|swss enable true
```
3. restart orchagent process only
4. observe syslog about "There are pending consumer tasks after restore". There will be no buffer related tasks.

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -643,7 +643,7 @@ void BufferOrch::doTask()
     //      3.1	    Ingress priority group (PG) configuration
     //      3.2.1	Buffer profile configuration
     //
-    // buffer poll
+    // buffer pool
     // └── buffer profile
     //     ├── buffer port ingress profile list
     //     ├── buffer port egress profile list
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -29,7 +29,7 @@ BufferOrch::BufferOrch(DBConnector *db, vector<string> &tableNames) : Orch(db, t
 {
     SWSS_LOG_ENTER();
     initTableHandlers();
-    initBufferReadyLists();
+    initBufferReadyLists(db);
 };
 
 void BufferOrch::initTableHandlers()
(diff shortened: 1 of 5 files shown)


orchagent/orchdaemon.cpp
```diff
@@ -264,7 +264,11 @@ bool OrchDaemon::init()
 
     if (WarmStart::isWarmStart())
     {
-        warmRestoreAndSyncUp();
+        bool suc = warmRestoreAndSyncUp();
+        if (!suc)
+        {
+            return false;
+        }
```
stcheng: from line 265 - 274 it could be ine one line `return !WarmStart::isWarmStart || warmRestoreAndSyncUp();`

qiluo-msft: It is equivalent. I prefer no change for easy reading.

orchagent/bufferorch.cpp
```diff
@@ -645,6 +645,25 @@ task_process_status BufferOrch::processEgressBufferProfileList(Consumer &consume
...
+
+    for(auto &it : m_consumerMap)
+    {
+        auto consumer = it.second.get();
+        if (consumer == profile_consumer)
+            continue;
+        if (consumer == pool_consumer)
+            continue;
+        consumer->drain();
+    }
```
jipanyang: Do you mean the configuration for other buffer tables like CFG_BUFFER_QUEUE_TABLE_NAME, CFG_BUFFER_PG_TABLE_NAME, CFG_BUFFER_PORT_INGRESS_PROFILE_LIST_NAME and CFG_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME have dependency on CFG_BUFFER_PROFILE_TABLE_NAME & CFG_BUFFER_POOL_TABLE_NAME?    CFG_BUFFER_PROFILE_TABLE_NAME has dependency on CFG_BUFFER_POOL_TABLE_NAME?

I didn't see existing code checking that dependency.

qiluo-msft: Exactly, so I added comments. The ground truth is inside SAI design doc.
For code, task_process_status::task_need_retry possibly indicates a dependency missing. They don't show the dependency relationship as specific as SAI design doc.

---
In reply to: [212513136](https://github.com/Azure/sonic-swss/pull/590#discussion_r212513136) [](ancestors = 212513136)

jipanyang: If there is dependency,  the exact dependency chain should be clarified and the problem should be fixed in normal code logic which applies to cold boot too.

orchagent/bufferorch.cpp
```diff
@@ -635,6 +635,38 @@ task_process_status BufferOrch::processEgressBufferProfileList(Consumer &consume
...
 
+void BufferOrch::doTask()
+{
+    // The hidden dependency tree:
+    // ref: https://github.com/opencomputeproject/SAI/blob/master/doc/QOS/SAI-Proposal-buffers-Ver4.docx
+    //      2	    SAI model
+    //      3.1	    Ingress priority group (PG) configuration
+    //      3.2.1	Buffer profile configuration
+    //
+    // buffer poll
```
lguohan: buffer pool

orchagent/bufferorch.cpp
lguohan: it is difficult to do this in the normal boot since the input sequence is unknown. we need to get the dependency at the entry level and then dispatched to task. in warmboot, since we have all the data, we can just solve the dependency at the table level.

qiluo-msft: During cold start, the task_process_status::task_need_retry handles dependency correctly, ie. anything missing will postpone task processing. So there is no problem to fix.
I agree this code is not explicit in dependency relationship. This should be improved in future.

---
In reply to: [212527615](https://github.com/Azure/sonic-swss/pull/590#discussion_r212527615) [](ancestors = 212527615)

Issue #592, opened 2018-08-24
Need to create a dummy interface in the Bridge


stcheng:
The Bridge interface needs to be up all the time. Otherwise, the command `bridge vlan` will fail due to the 'NO-CARRIER' flag.

The [solution ](https://docs.cumulusnetworks.com/display/DOCS/Ethernet+Bridging+-+VLANs)is to create a dummy interface to make the Bridge interface UP all the time.

```
root@b9987e019a19:/# ip link add Bridge up type bridge
root@b9987e019a19:/# ip link show Bridge
2378: Bridge: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN mode DEFAULT group default
    link/ether ae:bc:cb:d0:de:de brd ff:ff:ff:ff:ff:ff
root@b9987e019a19:/# ip link set Ethernet40 master Bridge
root@b9987e019a19:/# ip link show Bridge
2378: Bridge: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN mode DEFAULT group default
    link/ether 02:42:ac:11:00:02 brd ff:ff:ff:ff:ff:ff
root@b9987e019a19:/# ip link set Ethernet40 up
root@b9987e019a19:/# ip link show Bridge
2378: Bridge: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default
    link/ether 02:42:ac:11:00:02 brd ff:ff:ff:ff:ff:ff
root@b9987e019a19:/# ip link set Ethernet40 nomaster
root@b9987e019a19:/# ip link show Bridge
2378: Bridge: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN mode DEFAULT group default
    link/ether 00:00:00:00:00:00 brd ff:ff:ff:ff:ff:ff
root@b9987e019a19:/# ip link add dummy type dummy
root@b9987e019a19:/# ip link set dummy master Bridge
root@b9987e019a19:/# ip link show Bridge
2378: Bridge: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN mode DEFAULT group default
    link/ether 92:37:a5:ce:07:78 brd ff:ff:ff:ff:ff:ff
root@b9987e019a19:/# ip link set dummy up
[...]

PR #593, opened 2018-08-24
[test]: Add test_MultipleVlan test to test creation/removal

stcheng:
[test]: Add test_MultipleVlan test to test creation/removal

Refactored test_vlan.py so that each test is independent from
each other and intermediate state will be cleaned up afterwards.

Add test_MultipleVlan test to cover scenarios with VLAN and
VLAN members change. This test would fail on earlier kernel version
.e.g 3.16.0-4 when the bridge interface cannot remain UP all
the time. This would cause new VLAN failed to be created later on.

The current Debian stretch doesn't have this issue and the test
shall pass with current implementation.

### DIFF
--- a/tests/test_vlan.py
+++ b/tests/test_vlan.py
@@ -3,91 +3,181 @@ import time
 import re
 import json
 
-def test_VlanMemberCreation(dvs):
+class TestVlan(object):
+    def setup_db(self, dvs):
+        self.pdb = swsscommon.DBConnector(0, dvs.redis_sock, 0)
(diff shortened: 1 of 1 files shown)


stcheng:
to resolve: https://github.com/Azure/sonic-swss/issues/592

stcheng:
only update the test, the previous issue was no longer there in Debian stretch

stcheng:
**NOTE: THIS TEST WILL FAIL ON DEBIAN JESSIE KERNEL**

PR #594, opened 2018-08-24
Ignore testcase in Debian Jessie

qiluo-msft:


stcheng:
I'm okay with this. But I'm not sure how long do we want to keep the code/test compatible with the old kernel.

prsunny:
Just thinking, should we skip or should we just fail the test? IMO, this is a valid use-case.

lguohan:
@qiluo-msft  will remove it after upgrading to stretch?

qiluo-msft:
I will remove it either:
1. we have fix on Debian Jessie
2. we don't have development tasks on Debian Jessie branches, ie. the old branches are almost frozen for dev.

PR #595, opened 2018-08-24
[mirrororch]: Enable mirror session destination to point into VLAN

stcheng:
- Add FdbUpdate notification when manually inserting/removing FDB
  entries

- Refactor MirrorOrch so that it supports:
  1. create mirror session pointing to VLAN
  2. destination port move to VLAN/non-VALN
  3. destination port move to LAG
  Remove unnecessary variables in MirrorEntry

- Add destination IP to the NextHopUpdate message

- Add subnset routes for route observers

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -191,7 +191,7 @@ void MirrorOrch::createEntry(const string& key, const vector<FieldValueTuple>& d
     auto session = m_syncdMirrors.find(key);
     if (session != m_syncdMirrors.end())
     {
-        SWSS_LOG_ERROR("Failed to create session, session %s already exists",
+        SWSS_LOG_NOTICE("Failed to create session, session %s already exists",
                 key.c_str());
         return;
     }
@@ -316,11 +316,14 @@ bool MirrorOrch::getNeighborInfo(const string& name, MirrorEntry& session)
 {
     SWSS_LOG_ENTER();
 
-    // 1) Check if the destination IP is directly connected and get neighbor
-    // 2) Check if the session has next hop and neighbor information is available
+    // 1) If session destination IP is directly connected, and the neighbor
+    //    information is retrieved successfully, then continue.
+    // 2) If session has next hop, and the next hop's neighbor information is
+    //    retrieved successfully, then continue.
+    // 3) Otherwise, return false.
     if (!m_neighOrch->getNeighborEntry(session.dstIp,
                 session.neighborInfo.neighbor, session.neighborInfo.mac) &&
-            (session.nexthopInfo.nexthop == IpAddress() ||
+            (session.nexthopInfo.nexthop.isZero() ||
             !m_neighOrch->getNeighborEntry(session.nexthopInfo.nexthop,
                 session.neighborInfo.neighbor, session.neighborInfo.mac)))
     {
@@ -397,7 +400,7 @@ bool MirrorOrch::updateSession(const string& name, MirrorEntry& session)
             if (old_session.neighborInfo.port.m_type !=
                     session.neighborInfo.port.m_type)
             {
-                ret &= updateSessionVlan(name, session);
+                ret &= updateSessionType(name, session);
             }
 
             if (old_session.neighborInfo.mac !=
@@ -440,8 +443,8 @@ bool MirrorOrch::activateSession(const string& name, MirrorEntry& session)
 
     assert(!session.status);
 
-    /* Some platforms don't support SAI_MIRROR_SESSION_ATTR_TC and only
-     * support global mirror session traffic class. */
+    // Some platforms don't support SAI_MIRROR_SESSION_ATTR_TC and only
+    // support global mirror session traffic class.
     if (session.queue != 0)
     {
         attr.id = SAI_MIRROR_SESSION_ATTR_TC;
@@ -449,7 +452,7 @@ bool MirrorOrch::activateSession(const string& name, MirrorEntry& session)
         attrs.push_back(attr);
     }
 
-    attr.id =  SAI_MIRROR_SESSION_ATTR_MONITOR_PORT;
+    attr.id = SAI_MIRROR_SESSION_ATTR_MONITOR_PORT;
     attr.value.oid = session.neighborInfo.portId;
     attrs.push_back(attr);
 
@@ -457,18 +460,18 @@ bool MirrorOrch::activateSession(const string& name, MirrorEntry& session)
     attr.value.s32 = SAI_MIRROR_SESSION_TYPE_ENHANCED_REMOTE;
     attrs.push_back(attr);
 
-    /* Add the VLAN header when the packet is sent out from a VLAN */
+    // Add the VLAN header when the packet is sent out from a VLAN
     if (session.neighborInfo.port.m_type == Port::VLAN)
     {
         attr.id = SAI_MIRROR_SESSION_ATTR_VLAN_HEADER_VALID;
         attr.value.booldata = true;
         attrs.push_back(attr);
 
-        attr.id =SAI_MIRROR_SESSION_ATTR_VLAN_TPID;
+        attr.id = SAI_MIRROR_SESSION_ATTR_VLAN_TPID;
         attr.value.u16 = ETH_P_8021Q;
         attrs.push_back(attr);
 
-        attr.id =SAI_MIRROR_SESSION_ATTR_VLAN_ID;
+        attr.id = SAI_MIRROR_SESSION_ATTR_VLAN_ID;
         attr.value.u16 = session.neighborInfo.port.m_vlan_info.vlan_id;
         attrs.push_back(attr);
 
@@ -485,37 +488,37 @@ bool MirrorOrch::activateSession(const string& name, MirrorEntry& session)
     attr.value.s32 = SAI_ERSPAN_ENCAPSULATION_TYPE_MIRROR_L3_GRE_TUNNEL;
     attrs.push_back(attr);
 
-    attr.id =SAI_MIRROR_SESSION_ATTR_IPHDR_VERSION;
+    attr.id = SAI_MIRROR_SESSION_ATTR_IPHDR_VERSION;
     attr.value.u8 = MIRROR_SESSION_DEFAULT_IP_HDR_VER;
     attrs.push_back(attr);
 
     // TOS value format is the following:
     // DSCP 6 bits | ECN 2 bits
-    attr.id =SAI_MIRROR_SESSION_ATTR_TOS;
+    attr.id = SAI_MIRROR_SESSION_ATTR_TOS;
     attr.value.u16 = (uint16_t)(session.dscp << MIRROR_SESSION_DSCP_SHIFT);
     attrs.push_back(attr);
 
-    attr.id =SAI_MIRROR_SESSION_ATTR_TTL;
+    attr.id = SAI_MIRROR_SESSION_ATTR_TTL;
     attr.value.u8 = session.ttl;
     attrs.push_back(attr);
 
-    attr.id =SAI_MIRROR_SESSION_ATTR_SRC_IP_ADDRESS;
+    attr.id = SAI_MIRROR_SESSION_ATTR_SRC_IP_ADDRESS;
     copy(attr.value.ipaddr, session.srcIp);
     attrs.push_back(attr);
 
-    attr.id =SAI_MIRROR_SESSION_ATTR_DST_IP_ADDRESS;
+    attr.id = SAI_MIRROR_SESSION_ATTR_DST_IP_ADDRESS;
     copy(attr.value.ipaddr, session.dstIp);
     attrs.push_back(attr);
 
-    attr.id =SAI_MIRROR_SESSION_ATTR_SRC_MAC_ADDRESS;
+    attr.id = SAI_MIRROR_SESSION_ATTR_SRC_MAC_ADDRESS;
     memcpy(attr.value.mac, gMacAddress.getMac(), sizeof(sai_mac_t));
     attrs.push_back(attr);
 
-    attr.id =SAI_MIRROR_SESSION_ATTR_DST_MAC_ADDRESS;
+    attr.id = SAI_MIRROR_SESSION_ATTR_DST_MAC_ADDRESS;
     memcpy(attr.value.mac, session.neighborInfo.mac.getMac(), sizeof(sai_mac_t));
     attrs.push_back(attr);
 
-    attr.id =SAI_MIRROR_SESSION_ATTR_GRE_PROTOCOL_TYPE;
+    attr.id = SAI_MIRROR_SESSION_ATTR_GRE_PROTOCOL_TYPE;
     attr.value.u16 = session.greType;
     attrs.push_back(attr);
 
@@ -578,7 +581,7 @@ bool MirrorOrch::updateSessionDstMac(const string& name, MirrorEntry& session)
     assert(session.sessionId != SAI_NULL_OBJECT_ID);
 
     sai_attribute_t attr;
-    attr.id =SAI_MIRROR_SESSION_ATTR_DST_MAC_ADDRESS;
+    attr.id = SAI_MIRROR_SESSION_ATTR_DST_MAC_ADDRESS;
     memcpy(attr.value.mac, session.neighborInfo.mac.getMac(), sizeof(sai_mac_t));
 
     sai_status_t status = sai_mirror_api->set_mirror_session_attribute(session.sessionId, &attr);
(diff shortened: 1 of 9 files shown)


orchagent/mirrororch.cpp
```diff
@@ -194,6 +188,14 @@ void MirrorOrch::createEntry(const string& key, const vector<FieldValueTuple>& d
 {
     SWSS_LOG_ENTER();
 
+    auto session = m_syncdMirrors.find(key);
+    if (session != m_syncdMirrors.end())
+    {
+        SWSS_LOG_ERROR("Failed to create session, session %s already exists",
```
lguohan: >SWSS_LOG_ERROR [](start = 8, length = 14)

is this error, or just NOTICE

orchagent/mirrororch.cpp
```diff
@@ -320,91 +316,128 @@ bool MirrorOrch::getNeighborInfo(const string& name, MirrorEntry& session)
...
-    assert(session.nexthopInfo.resolved);
-
-    SWSS_LOG_INFO("Getting neighbor info for %s session\n", name.c_str());
-
-    if (!m_neighOrch->getNeighborEntry(session.nexthopInfo.nexthop, neighbor, mac))
+    // 1) Check if the destination IP is directly connected and get neighbor
+    // 2) Check if the session has next hop and neighbor information is available
+    if (!m_neighOrch->getNeighborEntry(session.dstIp,
+                session.neighborInfo.neighbor, session.neighborInfo.mac) &&
+            (session.nexthopInfo.nexthop == IpAddress() ||
```
lguohan: >session.nexthopInfo.nexthop == IpAddress()  [](start = 13, length = 43)

is there a function saying check nexthop.empty() so that it is more clear what you are checking here?

orchagent/mirrororch.cpp
```diff
@@ -320,91 +316,128 @@ bool MirrorOrch::getNeighborInfo(const string& name, MirrorEntry& session)
...
-    NeighborEntry neighbor;
-    MacAddress mac;
-
-    assert(session.nexthopInfo.resolved);
-
-    SWSS_LOG_INFO("Getting neighbor info for %s session\n", name.c_str());
-
-    if (!m_neighOrch->getNeighborEntry(session.nexthopInfo.nexthop, neighbor, mac))
+    // 1) Check if the destination IP is directly connected and get neighbor
+    // 2) Check if the session has next hop and neighbor information is available
```
lguohan: rewrite for clarity

1. if session destination IP is directly connected, then get the neighbor info.
2. if session has next hop, then the get neighbor info of the next hop.
3. otherwise, return false.

orchagent/mirrororch.cpp
```diff
@@ -562,175 +601,165 @@ bool MirrorOrch::updateSessionDstPort(const string& name, MirrorEntry& session)
...
+    }
+    else
+    {
+        attr.id = SAI_MIRROR_SESSION_ATTR_VLAN_HEADER_VALID;
+        attr.value.booldata = false;
+        attrs.push_back(attr);
+    }
+
+    sai_status_t status = sai_mirror_api->
+        set_mirror_session_attribute(session.sessionId, &attr);
```
lguohan: >attr [](start = 57, length = 4)

this is not correct. you need to iterate all attributes and update one-by-one.

orchagent/neighorch.cpp
```diff
@@ -516,11 +513,15 @@ bool NeighOrch::removeNeighbor(NeighborEntry neighborEntry)
         gCrmOrch->decCrmResUsedCounter(CrmResourceType::CRM_IPV6_NEIGHBOR);
     }
 
-    NeighborUpdate update = { neighborEntry, MacAddress(), false };
-    notify(SUBJECT_TYPE_NEIGH_CHANGE, static_cast<void *>(&update));
+    SWSS_LOG_NOTICE("Removed neighbor %s on %s",
+            m_syncdNeighbors[neighborEntry].to_string().c_str(), alias.c_str());
 
     m_syncdNeighbors.erase(neighborEntry);
     m_intfsOrch->decreaseRouterIntfsRefCount(alias);
```
lguohan: why do we have this code here? is this an existing bug?

tests/test_mirror.py
```diff
@@ -6,228 +6,762 @@
...
+        self.remove_ip_address("Vlan9", "8.8.8.0/24")
+        assert self.get_mirror_session_status(session) == "active"
+
+        # check monitor port
+        tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_MIRROR_SESSION")
+        assert len(tbl.getKeys()) == 1
+        (status, fvs) = tbl.get(tbl.getKeys()[0])
+        assert status == True
+        for fv in fvs:
+            if fv[0] == "SAI_MIRROR_SESSION_ATTR_MONITOR_PORT":
```
lguohan: can you assert SAI_MIRROR_SESSION_ATTR_VLAN_HEADER_VALID == false here?

tests/test_mirror.py
```diff
@@ -6,228 +6,762 @@
...
-        adb = swsscommon.DBConnector(1, dvs.redis_sock, 0)
-        cdb = swsscommon.DBConnector(4, dvs.redis_sock, 0)
+        # remove mirror session
+        self.remove_mirror_session(session)
 
-        tbl = swsscommon.Table(cdb, "ACL_TABLE")
-        # remove the ACL table
-        tbl._del("EVERFLOW_TABLE")
+
+    def test_MirrorDestMoveVlan(self, dvs):
```
lguohan: Do you have test case for MirrorDestMoveNonVlan to test move from vlan to non-vlan case?

orchagent/mirrororch.cpp
```diff
@@ -739,14 +768,12 @@ void MirrorOrch::updateFdb(const FdbUpdate& update)
...
-                //activate session
-                session.neighborInfo.resolved = true;
-                session.neighborInfo.mac = update.entry.mac;
+                // Activate session
                 session.neighborInfo.portId = update.port.m_port_id;
-
                 activateSession(name, session);
             }
         }
+        // Remvoe the monitor port
```
lguohan: >Remvoe [](start = 11, length = 6)

remove

orchagent/mirrororch.cpp
```diff
@@ -562,175 +601,165 @@ bool MirrorOrch::updateSessionDstPort(const string& name, MirrorEntry& session)
...
+        auto& session = it->second;
 
-        // It is possible to have few session that points to one FDB entry
-        if (sessionIter->second.neighborInfo.mac != update.entry.mac ||
-                sessionIter->second.neighborInfo.vlanOid != update.entry.bv_id)
+        // Check the following three conditions:
+        // 1) mirror session is pointing to a VLAN
+        // 2) the VLAN matches the FDB notification VLAN ID
+        // 3) the destination MAC matches the FDB notifaction MAC
+        if (session.neighborInfo.port.m_type != Port::VLAN ||
```
lguohan: >ighborInfo.port.m_type != Port::VLAN || [](start = 22, length = 39)

this is not correct. even when port type is non vlan,

orchagent/mirrororch.cpp
```diff
@@ -562,175 +601,165 @@ bool MirrorOrch::updateSessionDstPort(const string& name, MirrorEntry& session)
...
-        else if (session.status)
-        {
-            deactivateSession(name, session);
-            session.neighborInfo.resolved = false;
-        }
+        updateSession(name, session);
     }
 }
 
 void MirrorOrch::updateFdb(const FdbUpdate& update)
```
lguohan: >updateFdb [](start = 17, length = 9)

can you add a note to this function. what does it try to achieve?

when we receive a fdb notification:
1 mac move: then we could protentially update the dst port of the session
2. mac remove: make the session inactive.

orchagent/mirrororch.cpp
lguohan: >session.neighborInfo.port.m_type != Port::VLAN || [](start = 12, length = 49)

the condition is wrong here, even for non-vlan case, you need to de-activate the session.

orchagent/mirrororch.cpp
```diff
@@ -320,91 +316,128 @@ bool MirrorOrch::getNeighborInfo(const string& name, MirrorEntry& session)
...
+    {
+        // Update corresponding attributes
+        if (session.status)
         {
-            return false;
-        }
+            if (old_session.neighborInfo.port.m_type !=
+                    session.neighborInfo.port.m_type)
+            {
+                ret &= updateSessionVlan(name, session);
```
lguohan: rename to updateSessionType

orchagent/mirrororch.cpp
```diff
@@ -562,175 +601,165 @@ bool MirrorOrch::updateSessionDstPort(const string& name, MirrorEntry& session)
...
         return false;
     }
 
+    SWSS_LOG_NOTICE("Update mirror session %s VLAN to %s",
+            name.c_str(), session.neighborInfo.port.m_alias.c_str());
+
     return true;
 }
 
 void MirrorOrch::updateNextHop(const NextHopUpdate& update)
```
lguohan: can you add a comment to describe what case does updaetNextHop handle?

orchagent/mirrororch.cpp
```diff
@@ -562,175 +601,165 @@ bool MirrorOrch::updateSessionDstPort(const string& name, MirrorEntry& session)
...
-    for (; sessionIter != m_syncdMirrors.end(); ++sessionIter)
+    for (auto it = m_syncdMirrors.begin(); it != m_syncdMirrors.end(); it++)
     {
-        if (!update.prefix.isAddressInSubnet(sessionIter->second.dstIp))
+        const auto& name = it->first;
+        auto& session = it->second;
+
+        // Check if mirror session's destination IP drops into the subnet
+        // of the next hop update's prefix
+        if (session.dstIp != update.destination)
```
lguohan: the description does not match the code, please correct.

orchagent/mirrororch.cpp
```diff
@@ -562,175 +601,165 @@ bool MirrorOrch::updateSessionDstPort(const string& name, MirrorEntry& session)
...
+        if (update.nexthopGroup != IpAddresses() &&
+                update.nexthopGroup.getIpAddresses().count(session.nexthopInfo.nexthop))
 
-        if (session.nexthopInfo.resolved)
         {
-            // Check for ECMP route next hop update. If route prefix is the same
-            // and current next hop is still in next hop group - do nothing.
-            if (session.nexthopInfo.prefix == update.prefix && update.nexthopGroup.getIpAddresses().count(session.nexthopInfo.nexthop))
-            {
                 continue;
```
lguohan: >         [](start = 8, length = 8)

4 spaces.

orchagent/mirrororch.cpp
stcheng: yes. there is a function `isZero`. I will use that.

orchagent/mirrororch.cpp
```diff
@@ -562,175 +601,165 @@ bool MirrorOrch::updateSessionDstPort(const string& name, MirrorEntry& session)
...
+        auto& session = it->second;
 
-        SWSS_LOG_INFO("Updating neighbor info %s %s\n", update.entry.ip_address.to_string().c_str(), update.entry.alias.c_str());
+        SWSS_LOG_NOTICE("neighbor %s : %s", update.entry.ip_address.to_string().c_str(), update.entry.alias.c_str());
 
-        const auto& name = sessionIter->first;
-        auto& session = sessionIter->second;
 
-        if (update.add)
+        // XXX: add comments!
```
lguohan: >// XXX: add comments! [](start = 8, length = 21)

can you add comments?

orchagent/mirrororch.cpp
```diff
@@ -562,175 +601,165 @@ bool MirrorOrch::updateSessionDstPort(const string& name, MirrorEntry& session)
...
+    }
+
+    SWSS_LOG_NOTICE("Update mirror session %s monitor port to %s",
+            name.c_str(), port.m_alias.c_str());
+
+
+    return true;
+}
+
+bool MirrorOrch::updateSessionVlan(const string& name, MirrorEntry& session)
```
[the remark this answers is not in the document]
stcheng: Because this function is mainly used for setting/removing VLAN attribute, so what about the name UpdateSessionVlanType?

orchagent/mirrororch.cpp
```diff
@@ -562,175 +601,165 @@ bool MirrorOrch::updateSessionDstPort(const string& name, MirrorEntry& session)
...
-                continue;
-            }
+            session.nexthopInfo.nexthop = IpAddress();
         }
+
+        updateSession(name, session);
     }
 }
 
 void MirrorOrch::updateNeighbor(const NeighborUpdate& update)
```
lguohan: what is the case for updateNeighbor, can you have a comment?

orchagent/mirrororch.cpp
stcheng: I'll fix this

orchagent/mirrororch.cpp
```diff
@@ -758,72 +785,56 @@ void MirrorOrch::updateFdb(const FdbUpdate& update)
...
+    for (auto it = m_syncdMirrors.begin(); it != m_syncdMirrors.end(); it++)
+    {
+        const auto& name = it->first;
+        auto& session = it->second;
 
-        // It is possible to have few session that points to one LAG member
-        if (sessionIter->second.neighborInfo.port != update.lag || sessionIter->second.neighborInfo.portId != update.member.m_port_id)
+        // Check the following two conditions:
+        // 1) the neighbor is LAG
+        // 2) the neighbor LAG matches the update LAG
```
lguohan: >the neighbor LAG matches the update LAG [](start = 14, length = 39)

the second condition does not the code, please check.

orchagent/mirrororch.cpp
stcheng: will there be a non-vlan case?

orchagent/mirrororch.cpp
```diff
@@ -758,72 +785,56 @@ void MirrorOrch::updateFdb(const FdbUpdate& update)
...
+        if (session.neighborInfo.port.m_type != Port::LAG ||
```
lguohan: is this port LAG or LAG_MEMBER?

orchagent/neighorch.cpp
stcheng: So if the update is sent before the `m_syncdNeighbors.erase`, when the observer wants to query the NeighOrch to get the information, the deprecated entry is still in the m_syncdNeighbors data structure, which will cause issues. That is why I mvoe the update code position.

tests/test_mirror.py
stcheng: this test covers both cases: vlan to non-vlan and vice versa

tests/test_mirror.py
stcheng: sure i'll add this

orchagent/mirrororch.cpp
```diff
@@ -758,72 +785,56 @@ void MirrorOrch::updateFdb(const FdbUpdate& update)
...
+                session.neighborInfo.port != update.lag)
```
lguohan: >eighborInfo.port != update.lag [](start = 25, length = 30)

similar to vlan case, we only need to deal with lag member remove case. Even when we add new lag member, we do not need to change existing monitor port.

orchagent/mirrororch.cpp
stcheng: LAG. The neighbor will be the router interface. So it cannot be a LAG_MEMBER.

orchagent/mirrororch.cpp
stcheng: I have thought about this. The issue is that it might be easier to debug if the behavior of always sending to the first member of the LAG to be consistent. That is something different from VLAN where we cannot send the packet to random members. I am fine if we don't do anything for the LAG member add case, which makes the code simpler.

orchagent/mirrororch.cpp
stcheng: added

stcheng:
updated

tests/test_mirror.py
```diff
@@ -6,228 +6,766 @@
...
+        self.add_route(dvs, "8.8.0.0/16", "80.0.0.1")
+        assert self.get_mirror_session_status(session) == "active"
+
+        # check monitor port
+        tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_MIRROR_SESSION")
+        assert len(tbl.getKeys()) == 1
+        (status, fvs) = tbl.get(tbl.getKeys()[0])
+        assert status == True
+        for fv in fvs:
+            if fv[0] == "SAI_MIRROR_SESSION_ATTR_MONITOR_PORT":
```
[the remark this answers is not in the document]
stcheng: added

orchagent/mirrororch.cpp
```diff
@@ -452,49 +488,50 @@ bool MirrorOrch::activateSession(const string& name, MirrorEntry& session)
...
 
-    session.status = true;
-
-    status = sai_mirror_api->create_mirror_session(&session.sessionId, gSwitchId, (uint32_t)attrs.size(), attrs.data());
+    status = sai_mirror_api->
+        create_mirror_session(&session.sessionId, gSwitchId, (uint32_t)attrs.size(), attrs.data());
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to activate mirroring session %s\n", name.c_str());
         session.status = false;
```
qiluo-msft: >session.status = false; [](start = 8, length = 23)

I believe you need to return immediately. Otherwise status is assigned true.

Issue #596, opened 2018-08-25
warm restart test left neighbors on vEthernetXX


qiluo-msft:
After running warm restart test, there are unexpected neighbors on vEthernetXX. It will keep in STALE state, and then later impact neighsync. By design, vEthernetXX should not be visible inside vs docker. 

```
$ sudo python -m py.test -v --dvsname=vs -k "test_PortSyncdWarmRestart"
========================================================================================= test session starts ==========================================================================================
platform linux2 -- Python 2.7.9, pytest-3.3.0, py-1.5.3, pluggy-0.6.0 -- /usr/bin/python
cachedir: .cache
rootdir: /home/qiluo/work/sonic/sonic-buildimage/src/sonic-swss, inifile:
collected 68 items

tests/test_warm_reboot.py::test_PortSyncdWarmRestart PASSED                                                                                                                                      [100%]

========================================================================================= 67 tests deselected ==========================================================================================

$ docker exec -it vs bash
root@346360b6760c:/# ip neigh show
11.0.0.2 dev vEthernet16 lladdr 02:60:00:a7:3d:da STALE
11.0.0.11 dev Ethernet20 lladdr d6:dc:56:7c:ed:7d STALE
11.0.0.2 dev Ethernet16 lladdr 02:60:00:a7:3d:da STALE
11.0.0.10 dev Ethernet20 lladdr d6:dc:56:7c:ed:7d STALE

```

qiluo-msft:
I am working on a workaround.

PR #597, opened 2018-08-25
Fix warm restart test left neighbors on vEthernet(s)

qiluo-msft:
Fixed https://github.com/Azure/sonic-swss/issues/596

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -83,7 +83,7 @@ class VirtualServer(object):
             # create vpeer link
             ensure_system("ip link add %s type veth peer name %s" % (self.nsname[0:12], self.vifname))
             ensure_system("ip link set %s netns %s" % (self.nsname[0:12], self.nsname))
-            ensure_system("ip link set %s netns %s" % (self.vifname, self.nsname))
+            ensure_system("ip link set %s netns %d" % (self.vifname, pid))
 
             # bring up link in the virtual server
             ensure_system("ip netns exec %s ip link set dev %s name eth0" % (self.nsname, self.nsname[0:12]))
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,6 +10,11 @@ import StringIO
 import subprocess
 from swsscommon import swsscommon
 
+def ensure_system(cmd):
+    rc = os.WEXITSTATUS(os.system(cmd))
+    if rc:
+        raise RuntimeError('Failed to run command: %s' % cmd)
+
 def pytest_addoption(parser):
     parser.addoption("--dvsname", action="store", default=None,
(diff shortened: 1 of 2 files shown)


lguohan:
why does warm reboot test left such arp entries in the first place? do you know?

tests/test_warm_reboot.py
```diff
@@ -146,6 +146,11 @@ def test_PortSyncdWarmRestart(dvs):
     (status, fvs) = neighTbl.get("Ethernet20:11.0.0.10")
     assert status == True
 
+    # Make sure no neighbors on vEthernet
+    keys = neighTbl.getKeys();
+    for key in keys:
+        assert not key.startswith("vEthernet")
```
lguohan: this can be added to a appdb_validator in conftest.py, it is generic.

qiluo-msft:
The root cause is not clear to me.

tests/test_warm_reboot.py
qiluo-msft: Resolved

qiluo-msft:
Retest this please

lguohan:
retest this please

tests/conftest.py
```diff
@@ -73,20 +89,24 @@ def __init__(self, ctn_name, pid, i):
...
+            ensure_system("ip netns exec %s ip link set dev eth0 up" % (self.nsname))
+            ensure_system("ip netns exec %s ethtool -K eth0 tx off" % (self.nsname))
 
             # bring up link in the virtual switch
-            os.system("nsenter -t %d -n ip link set dev %s up" % (pid, self.vifname))
+            ensure_system("nsenter -t %d -n ip link set dev %s up" % (pid, self.vifname))
+
+        # disable arp, so no neigh on vEthernet(s)
+        # Note: outside the if-else, so existing VS container could be fixed
+        ensure_system("nsenter -t %d -n ip link set arp off dev %s" % (pid, self.vifname))
```
jipanyang: I think arp is needed for neighborsyncd and fdb function test?  The vEthernet interfaces here are actually mapped to Ethernet interfaces in vs (thought I'm not clear how that is done with the docker --network container:sw magic)

jipanyang:
Maybe the problem is because in the test, vs and sw dockers instances are re-used.  Those vEthernet interfaces are created at the very beginning using create_vnet.sh script.  The stale arp entries could be cleared before starting new round of test run.

tests/conftest.py
qiluo-msft: I disabled arp on only vEthernetXX. ARP and neighbors on EthernetXX are not affected during my experiment.

PR #599, opened 2018-08-26
Support warm-restart for neighsyncd

zhenggen-xu:
A new common class is introduced for the applications' table reconciliation, this could be shared for different applications and tables if the table is using format of key -> <f1/v1, f2/v2 ...>. 

- No function changes if warm-restart is disabled for neighsyncd.

In case warm-restart is enabled, below is done in the process to make sure the ARP/NDP is up-to-date:
- ARP/NDP will be restored from App-DB marked with stale.
- new/changed ARP/NDP during the warm-restart marked as new
- deleted ARP/NDP during the warm-restart marked as delete
- same ARP/NDP entries marked as same
- reconciliation is done after timer-out of a timer
- delta (including stale, new, delete) during the warm-restart would be pushed to App-DB and then to HW.

Add vs test cases to cover:
- No changes during the warm-restart
- ADD/DELETE/CHANGE/SAME neighbors during the warm-restart
- verify the timer changes taking effect for neighsyncd

**What I did**
Add support for warm-restart for neighbor table

**Why I did it**
To fully support warm-restart for SWSS and make sure neighbor table (ARP/NDP) are reconciled after warm-restart.

**How I verified it**
vs test cases and HW tests.

sudo pytest -v --dvsname=vs test_warm_reboot.py                                                                                                 
============================================================================================== test session starts ==============================================================================================
cachedir: .cache

rootdir: ~/SONIC/sonic-buildimage/src/sonic-swss/tests, inifile:
collected 3 items                                                                                                                                                                                               

test_warm_reboot.py::test_PortSyncdWarmRestart PASSED                                                                                                                                                     [ 33%]
test_warm_reboot.py::test_VlanMgrdWarmRestart PASSED                                                                                                                                                      [ 66%]
test_warm_reboot.py::test_swss_neighbor_syncup PASSED           

**Details if related**


neighsyncd/neighsync.h
```diff
@@ -4,20 +4,71 @@
...
     enum { MAX_ADDR_SIZE = 64 };
 
-    NeighSync(DBConnector *db);
+    NeighSync(RedisPipeline *pipelineAppDB,
+      RedisPipeline *pipelineConfDB);
 
     virtual void onMsg(int nlmsg_type, struct nl_object *obj);
-
-private:
     ProducerStateTable m_neighTable;
```
qiluo-msft: >m_neighTable [](start = 23, length = 12)

Why make m_neighTable public?

neighsyncd/neighsyncd.cpp
```diff
@@ -4,15 +4,34 @@
 #include "netdispatcher.h"
 #include "netlink.h"
 #include "neighsyncd/neighsync.h"
+#include "warm_restart.h"
+#include <time.h>
```
qiluo-msft: Move to top?

neighsyncd/neighsync.cpp
```diff
@@ -52,18 +284,32 @@ void NeighSync::onMsg(int nlmsg_type, struct nl_object *obj)
     key+= ipStr;
 
     int state = rtnl_neigh_get_state(neigh);
+    bool delete_key = false;
     if ((nlmsg_type == RTM_DELNEIGH) || (state == NUD_INCOMPLETE) ||
         (state == NUD_FAILED))
     {
-        m_neighTable.del(key);
-        return;
+	   delete_key = true;
```
qiluo-msft: Indent one more blank.

neighsyncd/neighsync.h
```diff
@@ -4,20 +4,71 @@
...
+        UNKNOWN = 4
+    };
+    typedef std::map<cache_state_t, std::string> cache_state_map;
+    static const cache_state_map cacheStateMap;
+    void startReconcileTimer();
+    bool checkReconcileTimer();
+    void readTableToMap();
+    void insertToMap(std::string key, std::vector<FieldValueTuple> fvVector, bool delete_key);
+    void reconcile(ProducerStateTable &ps_table);
+    bool m_warmStartInProgress = true;
```
qiluo-msft: >m_warmStartInProgress [](start = 9, length = 21)

It's not good to make data member public

neighsyncd/neighsync.h
```diff
@@ -4,20 +4,71 @@
...
+    enum cache_state_t
+    {
+        STALE	= 0,
+        SAME 	= 1,
+        NEW 	= 2,
+        DELETE  = 3,
+        UNKNOWN = 4
+    };
+    typedef std::map<cache_state_t, std::string> cache_state_map;
+    static const cache_state_map cacheStateMap;
```
qiluo-msft: >cacheStateMap [](start = 33, length = 13)

It's not good to make data member public

neighsyncd/neighsyncd.cpp
```diff
@@ -4,15 +4,34 @@
...
+
+    if (sync.m_warmStartInProgress)
+    {
+        WarmStart::setWarmStartState("neighsyncd", WarmStart::INIT);
+        // To-be-discussed: Do we need flush the producestate table pipeline to
+        // make sure the relevant Table has the up-to-date data before this daemon restarted?
+        // (e,g, pending messages were not writen to the Table)
+        // On the other hand, in case docker restart, we need make sure the flush
+        // won't break anything for orchagent
+    }
```
qiluo-msft: You may this if-block into NeighRestartAssit::NeighRestartAssit()

neighsyncd/neighsync.cpp
```diff
@@ -11,12 +11,244 @@
...
+    vector<string> keys;
+
+    m_appTable.getKeys(keys);
+    FieldValueTuple state(CACHE_STATE_FIELD, "");
+
+    for (const auto &key: keys)
+    {
+        vector<FieldValueTuple> fv;
+
+	    /* if the fieldvlaue empty, skip */
```
qiluo-msft: >fieldvlaue [](start = 15, length = 10)

typo

neighsyncd/neighsync.cpp
```diff
@@ -11,12 +11,244 @@
...
+            ps_table.set(iter->first, iter->second);
+        }
+        else
+        {
+            assert("unknown type cache state" == NULL);
+        }
+    }
+    // clear the map
+    neighborCacheMap.clear();
+    return;
```
qiluo-msft: >return [](start = 4, length = 6)

Assign true to m_warmStartInProgress?

neighsyncd/neighsync.cpp
```diff
@@ -11,12 +11,244 @@
...
+{
+    string s;
+    for (const auto &temps : fv )
+    {
+	   s += temps.first + ":" + temps.second + ", ";
+    }
+    return s;
+}
+
+void NeighRestartAssit::setCacheEntryState(std::vector<FieldValueTuple> &fvVector,
```
qiluo-msft: >setCacheEntryState [](start = 24, length = 18)

setCacheLastEntryState

neighsyncd/neighsync.cpp
```diff
@@ -11,12 +11,244 @@
...
+{
+    for (auto &iter : cacheStateMap)
+    {
+        if (fvVector.back().second == iter.second)
+        {
+            return iter.first;
+        }
+    }
+    assert("Cache Entry State is corrupted" == NULL);
+    return UNKNOWN;
```
qiluo-msft: >return [](start = 4, length = 6)

throw a logic_error/runtime exception?

neighsyncd/neighsync.h
```diff
@@ -4,20 +4,71 @@
...
+    bool m_warmStartInProgress = true;
+
+private:
+    uint32_t m_reconcileTimer = DEFAULT_RECONCILE_TIMER;
+    time_t m_startTime;
+    double m_secondsPast = 0;
+    NeighborTableMap neighborCacheMap;
+    std::string m_appTableName;
+    Table m_appTable;
+    Table m_cfgTable;
```
qiluo-msft: >m_cfgTable [](start = 10, length = 10)

not used

qiluo-msft:
Design discussion:
1. The orchagent/neighorch.cpp has warm starting feature now. Its responsibility is propagate all the neigh entries in ApplDb to downstream syncd
2. The design principle of neighsyncd is just sync netlink messages to downstream orchagent, there is no state concept in neighsyncd
3. APP_NEIGH_TABLE is the implementation detail of orchagent, and neighsyncd should not read it or assume anything there.

I proposed another design choice, similiar to INIT_VIEW and APPLY_VIEW between orchagent and syncd. The warm start process of neighsyncd:
1. orchagent/neighorch.cpp warm starts, not processing data in ProducerStateTable
2. neighsyncd sends a STALE event to orchagent through ProducerStateTable, so orchagent marks existing neigh entries stale
3. neighsyncd collects all the neigh from netlink, and propagate to orchagent, and then propagate to syncd
4. neighsyncd sends a APPLY event to orchagent, so orchagent del all left stale neigh entries, and then propagte to syncd
   
This choice has some advantages:
1. The warm start features of neighsyncd and orchagent are independent, ie. we could restart only neighsyncd anytime
2. There is no timer logic

neighsyncd/neighsync.cpp
```diff
@@ -11,12 +11,244 @@
...
+        vector<FieldValueTuple> fv;
+
+	    /* if the fieldvlaue empty, skip */
+        if (!m_appTable.get(key, fv))
+        {
+            continue;
+        }
+
+        fv.push_back(state);
+        setCacheEntryState(fv, STALE);
```
jipanyang: What is the reason it is split into two steps? 
it looks to me "fv.emplace_back(CACHE_STATE_FIELD, cacheStateMap.at(STALE))"  should work well?

neighsyncd/neighsync.cpp
```diff
@@ -11,12 +11,244 @@
...
+        if (found != neighborCacheMap.end())
+        {
+            setCacheEntryState(found->second, DELETE);
+        }
+    }
+    else if (found != neighborCacheMap.end())
+    {
+        /* check only the original vector range (exclude cache-state field/value) */
+        if(!equal(fvVector.begin(), fvVector.end(), found->second.begin()))
+        {
```
jipanyang: Is it possible that  the order of field/value has been changed though the internal content is the same?  ex.  f1/v1,f2/v2,  got changed to f2/v2, f1/v1.   equal algorithm will take them as different fvvector.

zhenggen-xu:
The design was proposed a while back but I am happy to discucss more if needed here. 

I disagree with a few points in your design discussion:
- It is fine that orchagent/neighorch.cpp can propagate the neighbor entries in AppDB to syncd, but that has nothing to do with who should do the reconciliation. 
- neighsyncd is the owner/producer of the APP_NEIGH_TABLE, it will definitely need understand the table and should be the one to do reconciliation of the table.

In principle, neighsyncd should not be part of swss at all and should be in another system.

The design you proposed is just making both systems to be coupled together rather than to decouple them.  It makes the design complicated without any benefit:
 - With the design of the current implementation, you can restart neighsyncd at any time. 
 - Timer was the signal to tell system to do the reconciliation,  it plays the same role on either design choice. e,g, we use it to send APPLY_VIEW signal in your choice. 
    In case, there is discussion to eliminate the timer all together, that is not something to do with this design choice but the choice of using timer or not. 

Also, in your design choice, rather than providing benefit, it confuses responsibility of the producer or consumer of tables.  The consumer (orchagent) should not manipulate the table like neighbor_table by any means.

neighsyncd/neighsyncd.cpp
```diff
@@ -4,15 +4,34 @@
...
+    NeighSync sync(&pipelineAppDB, &pipelineConfDB);
+
+    if (sync.m_warmStartInProgress)
+    {
+        WarmStart::setWarmStartState("neighsyncd", WarmStart::INIT);
+        // To-be-discussed: Do we need flush the producestate table pipeline to
+        // make sure the relevant Table has the up-to-date data before this daemon restarted?
+        // (e,g, pending messages were not writen to the Table)
+        // On the other hand, in case docker restart, we need make sure the flush
+        // won't break anything for orchagent
```
jipanyang: For planned warm restart, maybe pipleline flush could be done before shutdown.  What is the pipeline batch number for neighsyncd?   If not 1,  it looks neighsyncd should be process the pre-shutdown state check signal too.

The one that is more tricky is how to deal with those entries which have been created by producestate but not read by consumerstate yet.   Maybe before reading table from APPDB,  no pending data should exist in producestate buffer.

neighsyncd/neighsync.cpp
```diff
@@ -11,12 +11,244 @@
...
+        }
+    }
+}
+
+NeighRestartAssit::~NeighRestartAssit()
+{
+}
+
+/* join the field-value strings for straight printing */
+string NeighRestartAssit::joinVectorString(const vector<FieldValueTuple> &fv)
```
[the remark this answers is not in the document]
zhenggen-xu: joinFieldValues function is in sai_redis_record.cpp, it looks similar with a different format.  We can look into it see if we should make them into swss-common.

neighsyncd/neighsyncd.cpp
jipanyang: fpmsyncd has the same issues to deal with during bgp docker warm restart:  pipeline flush and buffered set/del operation in producerstate set. 

It looks we need to make orchagent_restart_check program in https://github.com/Azure/sonic-swss/pull/562 a common tool beyond orchagent.

lguohan:
@zhenggen-xu , it has benefit to let swss to handle state reconciliation, in this case, we only have service to handle different application state reconciliation. we can handle neighbor sync and bgp sync in same way.

jipanyang:
The possible solution is to check the temporary table used by producerState,  if it is empty then all buffered data has been consumed by orchagent.  

Applications(neighsyncd, fpmsyncd,..) make sure the specific temporary table  is empty before reading existing data from regular table for reconciliation task.  

Implementation wise, it should be very simple to add a isEmpty() method to ProducerStateTable class.

zhenggen-xu:
@jipanyang Yes, we we can do pre-shutdown flush and empty check for producerState table.

jipanyang:
@zhenggen-xu  Just to clarify,  empty check for producerState table probably should be done after re-start.  applications don't know when orchagent will stop consuming the data and go shutdown.

zhenggen-xu:
@jipanyang Yes, that was that I meant and wrote. :-)

lguohan:
my proposal is to have a table level resync operation. The operation has three stages, 1) prepare resync. 2) populate the full table. 3) finish resync. All these three operations are initiated by the application.

In the first step, the application send prepare resyncd to orchagent, and orchagent just freeze its read operation, and also mark all the data stale. In the second step, application populates the full table. In the third step, application sends finish resync signal, and orchagent does the state reconciliation.

Such common table-level semantic simplify application development, we can simplify the fpmsync and neighsync, they do not need to implement comparison logic themselves. 

@rodnymolina , the signal is still give by fpmsyncd, orchagent cannot/do not need to figure out the reconnect event between zebra and fpmsyncd

@rodnymolina , can you explain kernel-routing-reconciliation, I am missing the context.

@zhenggen-xu , in this proposal, the owner is still application, the applications when to start the state reconciliation, swss just make it simply for the application to dump the table and resync.

@jipanyang , such service can be used for many different table where the user just want to dump a new table to the db without worrying about doing diff himself. For example, this can be use to handle those init json configuration cases.

jipanyang:
No matter where the comparison and reconciliation is done for arp and route, fpmsyncd/neigsyncd or orchagent,  we could make them work.
The concerns with doing that in orchagent are:
<1> Orchagent has various dependency processing, once the data is in orchagent it builds up connection with other related objects,
doing comparison and reconciliation for arp and route in orchagent would be more complicated than doing the same task in fpmsyncd or neigsyncd.

<2> Currently the interface between orchagent and application is simple and straightforward, producerstate/consumerstate. Once the phase concept is introduced between them,
the interaction will be become rather complex, race condition may be introduced, I expect seeing a series bugs with it, which could be avoided if we just let neighsyncd and fpmsyncd do their job alone.   We have got a lot headache with the simple port configDone/InitDone signals.

But I agree that there are some common logic with the comparison and reconciliation for different objects, we should try to extract out the common functions for reuse (among route, arp, fdb and etc.)

jipanyang:
For the case of NEW and Delete arp messages,  probably the write operation could be done immediately.  It seems there is no reason to delay them to the very end?   They don't affect the marking of stale entries.

neighsyncd/neighsyncd.cpp
zhenggen-xu: just to conclude here, pipeline flush is not necessary for pre-shutdown, as long as the consumer data and database plain table data are in sync, we should be fine as we are going to get the new table from application/kernel again. 

After daemon is restarted, we need make sure the producerstate table is empty or cleared, see:
https://github.com/Azure/sonic-swss-common/pull/227

neighsyncd/neighsyncd.cpp
```diff
@@ -29,10 +48,24 @@ int main(int argc, char **argv)
...
+            {
+                sync.readTableToMap();
+                WarmStart::setWarmStartState("neighsyncd", WarmStart::RESTORED);
+                sync.startReconcileTimer();
+            }
             while (true)
             {
                 Selectable *temps;
-                s.select(&temps);
+                s.select(&temps, SELECT_TIMEOUT);
```
rodnymolina: I would suggest to make use of SelectableTimer class by pushing an instance of it into the Select/Epoll loop. This logic is unnecessarily stopping the main loop every second, even when warm-reboot feature is disabled.

neighsyncd/neighsync.cpp
zhenggen-xu: The state was meant for the entry not the last,  just that we keep the state at the tail of the entry.

neighsyncd/neighsync.cpp
zhenggen-xu: This is to capture the code level issue not really for run time issue like IO/resource related.

zhenggen-xu:
@jipanyang Regarding the question about write DELETE/NEW entries to AppDB directly, I think it should be fine to wait a little amount of time instead of pushing right away.  The idea was to expand this logic to more tables, but some application could flap entries in the beginning of restart, e,g have delete and add for the entry,  by waiting to things to be settled, we could eliminate any potential traffic interruption.

neighsyncd/neighsyncd.cpp
zhenggen-xu: This is timeout setting for select function, which only takes effect if nothing is active in the file descriptor or no processing for the signals.  This won't impact the performance when warm-reboot feature is enabled or disabled.

jipanyang:
As discussed,  common library functions for "compare and push" or "mark and sweep" should be extracted. The common library functions may be used for arp, route, and other types of objects which have the need to identify delta data.      

Well defined common library functions should not block possible local optimization like the incremental push of arp entries, but we may have a working library first then see how to improve from there.

neighsyncd/neighsync.cpp
qiluo-msft: Then 'logic_error' may fit the purpose?

---
In reply to: [214507371](https://github.com/Azure/sonic-swss/pull/599#discussion_r214507371) [](ancestors = 214507371)

zhenggen-xu: This particular error should never happen as it is just a map from string to enum. If it ever happened due to internal code error, there is nothing for the caller to do, so I think assert is the right approach here.

neighsyncd/neighsync.cpp
```diff
@@ -11,12 +11,255 @@
...
+            SWSS_LOG_NOTICE("%s NEW, key: %s, %s",
+                    m_appTableName.c_str(), iter->first.c_str(), s.c_str());
+
+            //add to appDB, exclude the state
+            iter->second.pop_back();
+            m_psTable->set(iter->first, iter->second);
+        }
+        else
+        {
+            assert("unknown type cache state" == NULL);
```
qiluo-msft: >assert [](start = 12, length = 6)

This is related to assert inside getCacheEntryState.

neighsyncd/neighsync.cpp
qiluo-msft: No, assert is never the right approach for runtime code. In release build, assert will be optimized to nothing, and the code will silently ignore the 'logic error', and keep continue.
If you think 'there is nothing for the caller to do', you should crash or report error and exit.

neighsyncd/neighsyncd.cpp
rodnymolina: Ass discussed, i'm not expecting any performance hit, but this approach doesn't sound right. We shouldn't rely on a permanent 1-second timer when we only need it a single time. In these cases we should add a 'selectable' timer within the 'select' object, and remove it when we don't need it anymore.

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,273 @@
...
+       if has "STALE/DELETE" flag, delete it from appDB.
+       else if "NEW" flag,  add it to appDB
+       else, assert (should not happen)
+    */
+    SWSS_LOG_NOTICE("Hit reconcile function");
+    for (auto iter = neighborCacheMap.begin(); iter != neighborCacheMap.end(); ++iter )
+    {
+        string s = "";
+        s = joinVectorString(iter->second);
+        if (getCacheEntryState(iter->second) == SAME)
```
qiluo-msft: >getCacheEntryState [](start = 12, length = 18)

could you getCacheEntryState() once?

neighsyncd/neighsyncd.cpp
```diff
@@ -29,10 +33,22 @@ int main(int argc, char **argv)
             netlink.dumpRequest(RTM_GETNEIGH);
 
             s.addSelectable(&netlink);
+            if (sync.getRestartAssist()->isWarmStartInProgress())
+            {
```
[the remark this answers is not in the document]
zhenggen-xu: It was reserved in the class construct code, and since now the other PR is merged, code is added.

neighsyncd/neighsync.cpp
zhenggen-xu: In my case, the assert is only for debug purpose for development, it is ok to remove them in release build.  But do a throw should work fine too.

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,276 @@
...
+        }
+
+        fv.push_back(state);
+        setCacheEntryState(fv, STALE);
+
+        string s = "";
+        s = joinVectorString(fv);
+
+        SWSS_LOG_INFO("write to cachemap: %s, key: %s, "
+               "%s", m_appTableName.c_str(), key.c_str(), s.c_str());
```
jipanyang: s = joinVectorString(fv); is used for LOG_INFO only.
s._c_str() could be replaced with joinVectorString(fv).c_str(),

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,276 @@
...
+    return;
+}
+
+void AppRestartAssist::insertToMap(string key, vector<FieldValueTuple> fvVector, bool delete_key)
+{
+    string s;
+    s = joinVectorString(fvVector);
+
+    SWSS_LOG_INFO("Received message %s, key: %s, "
+            "%s, delete = %d", m_appTableName.c_str(), key.c_str(), s.c_str(), delete_key);
```
jipanyang: s = joinVectorString(fv); is used for LOG_INFO only.
s._c_str() could be replaced with joinVectorString(fv).c_str(),

neighsyncd/neighsync.cpp
zhenggen-xu: Yeah,  thought I would need to have it printed in a few places. 
Done.

tests/test_warm_reboot.py
```diff
@@ -285,3 +300,390 @@ def test_VlanMgrdWarmRestart(dvs):
...
+
+def test_swss_neighbor_syncup(dvs):
+
+    appl_db = swsscommon.DBConnector(swsscommon.APPL_DB, dvs.redis_sock, 0)
+    conf_db = swsscommon.DBConnector(swsscommon.CONFIG_DB, dvs.redis_sock, 0)
+    state_db = swsscommon.DBConnector(swsscommon.STATE_DB, dvs.redis_sock, 0)
+
+    # warm start with temp view not supported yet
+    #if dvs.tmpview == True:
+    #    return
```
jipanyang: remove the comment. neigsyncd warm restart doesn't have dependency on syncd.

rodnymolina:
As discussed, i will be replacing the common portion of this reconciliation logic with my own code, which will be flexible enough to satisfy neighborSyncd as well as other applications such as fpmSyncd/routing-stack.

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,273 @@
...
+    m_dockerName(docker_name),
+    m_psTable(ps_table),
+    warmStartTimer(timespec{0, 0})
+{
+    WarmStart::checkWarmStart(m_appName, m_dockerName);
+
+    m_appTableName = m_appTable.getTableName();
+
+    // set the default timer value
+    if (defaultWarmStartTimerValue != 0)
```
qiluo-msft: >if (defaultWarmStartTimerValue != 0) [](start = 4, length = 36)

if (defaultWarmStartTimerValue != 0)
then m_reconcileTimer is uninitialized.

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,273 @@
...
+{
+    {STALE,     "STALE"},
+    {SAME,      "SAME"},
+    {NEW,       "NEW"},
+    {DELETE,    "DELETE"},
+    {UNKNOWN,   "UNKNOWN"}
+};
+
+AppRestartAssist::AppRestartAssist(RedisPipeline *pipelineAppDB,
+    const std::string &app_name, const std::string &docker_name,
```
qiluo-msft: >app_name [](start = 23, length = 8)

you mix 2 naming conventions for parameter variables

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,273 @@
...
+        WarmStart::setWarmStartState(m_appName, WarmStart::INIT);
+    }
+}
+
+AppRestartAssist::~AppRestartAssist()
+{
+}
+
+/* join the field-value strings for straight printing */
+string AppRestartAssist::joinVectorString(const vector<FieldValueTuple> &fv)
```
qiluo-msft: >joinVectorString [](start = 25, length = 16)

You could write a general join function and move it to swss-common. There is one more ", " at the end.

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,273 @@
...
+{
+    for (auto &iter : cacheStateMap)
+    {
+        if (fvVector.back().second == iter.second)
+        {
+            return iter.first;
+        }
+    }
+    throw std::logic_error("cache entry state is invalid");
+    return UNKNOWN;
```
qiluo-msft: Not reachable. I believe you could remove 'UNKNOWN' completely.

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,273 @@
...
+        if (!m_appTable.get(key, fv))
+        {
+            continue;
+        }
+
+        fv.push_back(state);
+        setCacheEntryState(fv, STALE);
+
+        string s = "";
+        s = joinVectorString(fv);
```
qiluo-msft: string s = joinVectorString(fv);

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,273 @@
...
+    {
+        /* check only the original vector range (exclude cache-state field/value) */
+        if(!equal(fvVector.begin(), fvVector.end(), found->second.begin()))
+        {
+            SWSS_LOG_NOTICE("%s, found key: %s, new value ", m_appTableName.c_str(), key.c_str());
+
+            FieldValueTuple state(CACHE_STATE_FIELD, "");
+            fvVector.push_back(state);
+
+            //mark as NEW flag
```
qiluo-msft: add one blank after //

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,273 @@
...
+            setCacheEntryState(fvVector, NEW);
+            appTableCacheMap[key] = fvVector;
+        }
+        else
+        {
+            SWSS_LOG_INFO("%s, found key: %s, same value", m_appTableName.c_str(), key.c_str());
+
+            //mark as SAME flag
```
qiluo-msft: add one blank after //

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,273 @@
...
+void AppRestartAssist::reconcile()
+{
+    /*
+       iterate throught the table
+       if the entry has "SAME" flag, do nothing
+       if has "STALE/DELETE" flag, delete it from appDB.
+       else if "NEW" flag,  add it to appDB
+       else, throw (should not happen)
+    */
+    SWSS_LOG_NOTICE("Hit reconcile function");
```
qiluo-msft: >SWSS_LOG_NOTICE [](start = 4, length = 15)

SWSS_LOG_ENTER() is for the exact purpose.

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,273 @@
...
+    for (auto iter = appTableCacheMap.begin(); iter != appTableCacheMap.end(); ++iter )
+    {
+        string s = "";
+        s = joinVectorString(iter->second);
```
qiluo-msft: string s = joinVectorString(iter->second);

neighsyncd/neighsync.cpp
```diff
@@ -9,14 +9,273 @@
...
+        }
+    }
+    // clear the map
+    appTableCacheMap.clear();
+    WarmStart::setWarmStartState(m_appName, WarmStart::RECONCILED);
+    m_warmStartInProgress = false;
+    return;
+}
+
+//start the timer, take Select class "s" to add the timer.
```
qiluo-msft: add one blank after //

neighsyncd/neighsyncd.cpp
```diff
@@ -29,10 +33,23 @@ int main(int argc, char **argv)
...
             while (true)
             {
                 Selectable *temps;
                 s.select(&temps);
+                if (sync.getRestartAssist()->isWarmStartInProgress())
+                {
+                    if (sync.getRestartAssist()->checkReconcileTimer(temps))
+                    {
+                        sync.getRestartAssist()->reconcile();
+                        sync.getRestartAssist()->stopReconcileTimer(s);
```
qiluo-msft: >stopReconcileTimer [](start = 49, length = 18)

First stopReconcileTimer, then reconcile?

neighsyncd/neighsyncd.cpp
```diff
@@ -29,10 +32,23 @@ int main(int argc, char **argv)
...
```
qiluo-msft: >stopReconcileTimer [](start = 49, length = 18)

I believe what you need is a oneshot selectable timer, and it is totally independent with RestartAssist. Just a refactoring suggestion here.

neighsyncd/neighsync.cpp
zhenggen-xu: This was already addressed in the new commit where there is a internal default timer.

neighsyncd/neighsync.cpp
zhenggen-xu: Yes, it was discussed, let's move this to later PR. thx

neighsyncd/neighsyncd.cpp
zhenggen-xu: It should not matter as there was a warmstartinprogress flag check before, but it is fine to move the other way around.

neighsyncd/neighsyncd.cpp
zhenggen-xu: Let's address this in later PR if needed, thx.

warmrestart/warmRestartAssist.cpp
```diff
@@ -0,0 +1,270 @@
...
+using namespace std;
+using namespace swss;
+
+const AppRestartAssist::cache_state_map AppRestartAssist::cacheStateMap =
+{
+    {STALE,     "STALE"},
+    {SAME,      "SAME"},
+    {NEW,       "NEW"},
+    {DELETE,    "DELETE"},
+    {UNKNOWN,   "UNKNOWN"}
```
qiluo-msft: >UNKNOWN [](start = 5, length = 7)

I mean you can remove all "UNKNOWN" in your code.

warmrestart/warmRestartAssist.h
```diff
@@ -0,0 +1,70 @@
...
+        ProducerStateTable *psTable, const uint32_t defaultWarmStartTimerValue = 0);
+    virtual ~AppRestartAssist();
+
+    enum cache_state_t
+    {
+        STALE	= 0,
+        SAME 	= 1,
+        NEW 	= 2,
+        DELETE  = 3,
+        UNKNOWN = 4
```
qiluo-msft: >UNKNOWN [](start = 8, length = 7)

I mean you can remove all "UNKNOWN" in your code.

neighsyncd/neighsync.cpp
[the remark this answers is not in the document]
zhenggen-xu: refactor in later PR.

PR #600, opened 2018-08-27
WarmRestart class changes

rodnymolina:
The current WarmStart class assumes that restarting applications will invoke checkWarmStart() method during initialization. Currently, this method serves two main purposes: 1) a static-constructor and 2) checks if WarmRestart feature has been enabled for a given application.

In this PR i'm proposing to split these different logics in two separated methods. The main goal here is to allow applications such as fpmSyncd to call checkWarmStart whenever it please (i.e. fpmSyncd restart event or zebra<->fpmSyncd tcp re-establishment). There could be other applications in the future interested in this differentiated approach.

Other than that i'm making small adjustments in the code to enhance readibility.

Signed-off-by: Rodny Molina <rmolina@linkedin.com>

### DIFF
--- a/warmrestart/warm_restart.cpp
+++ b/warmrestart/warm_restart.cpp
@@ -55,13 +55,14 @@ void WarmStart::initialize(const std::string &app_name,
  * No need to check docker level knobs in this case since the whole system is being rebooted .
 
  * <2> Upon docker service start, first to check system knob.
- * if enabled, docker warm start should be performed, otherwise system warm reboot will be ruined.
- * If system knob disabled, while docker knob enabled, this is likely an individual docker warm restart request.
+ * if enabled, docker warm-start should be performed, otherwise system warm-reboot will be ruined.
+ * If system knob disabled, while docker knob enabled, this is likely an individual docker
+ * warm-restart request.
 
  * Within each application which should take care warm start case,
  * when the system knob or docker knob enabled, we do further check on the
- * actual warm start state ( restart_count), if no warm start state data available,
- * the database has been flushed, do cold start. Otherwise warm start.
+ * actual warm-start state ( restore_count), if no warm-start state data available,
+ * the database has been flushed, do cold start. Otherwise warm-start.
  */
 
 /*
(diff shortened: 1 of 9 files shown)


warmrestart/warm_restart.cpp
```diff
@@ -1,24 +1,54 @@
...
 #include "schema.h"
-#include <climits>
 #include "warm_restart.h"
 
 namespace swss {
 
 const WarmStart::WarmStartStateNameMap WarmStart::warmStartStateNameMap =
 {
-    {INIT,          "init"},
+    {INITIALIZED,   "initialized"},
```
jipanyang: If the state name is changed here,  swss-schema.md should be updated too.

lguohan:
test failure

lguohan:
what is the scenario for zebra <-> fpmsyncd re-establish? they are in the same docker, should be restarted as the same time.

rodnymolina:
@lguohan i'm adding a certain level of flexibility in my routing warm-reboot code. My reconciliation code is dealing with all these scenarios without having to add any extra complexity: 1) routing-stack crash, or 2) fpmsyncd crash, or 3) docker restart. If we have a process-manager with the docker container (supervisord or watchfrr) that detects the crash and re-instantiate the process, then there's no need to restart the entire bgp-service from the host side.

warmrestart/warm_restart.cpp
```diff
@@ -1,108 +1,139 @@
...
+ * the database has been flushed, do cold start. Otherwise warm-start.
  */
 
-// Check warm start flag at the very begining of application, do it once for each process.
-bool WarmStart::checkWarmStart(const std::string &app_name, const std::string &docker_name)
+/*
+ * Method to verify/obtain the state of Warm-Restart feature for any warm-reboot
+ * capable component. Typically, this function will be called during initialization of
+ * SONiC modules; however, method could be invoked at any given point to verify the
+ * latest state of Warm-Restart functionality.
```
jipanyang: The comment here is a little missleading.
 ```
however, method could be invoked at any given point to verify the
 * latest state of Warm-Restart functionality.
```
The function also updates the restart_count whenever called, not just perform check.

Comment (rodnymolina) [reply warmrestart/warm_restart.cpp]: Ok, will adjust that comment.

Comment (rodnymolina) [reply warmrestart/warm_restart.cpp]: done

File: warmrestart/warm_restart.cpp (hunk 1 above)
Comment (rodnymolina) [reply warmrestart/warm_restart.cpp]: done

File: warmrestart/warm_restart.cpp [hunk 3]
```diff
@@ -1,108 +1,139 @@
...
+    warmStart.m_stateWarmRestartTable =
+        std::unique_ptr<Table>(new Table(warmStart.m_stateDb.get(), STATE_WARM_RESTART_TABLE_NAME));
+
+    warmStart.m_cfgDb =
+        std::make_shared<swss::DBConnector>(CONFIG_DB, swss::DBConnector::DEFAULT_UNIXSOCKET, 0);
+
+    warmStart.m_cfgWarmRestartTable =
+        std::unique_ptr<Table>(new Table(warmStart.m_cfgDb.get(), CFG_WARM_RESTART_TABLE_NAME));
+
+    setWarmStartState(app_name, WarmStart::INITIALIZED);
```
jipanyang: For portsyncd and vlanmgrd,  since no reconciliation is needed,  "setWarmStartState("orchagent", WarmStart::INITIALIZED);" was not called explicitly like that for orchagent.
Here setWarmStartState() is called implicitly, VS test check for portsyncd and vlanmgrd will fail.  

I think you may either call setWarmStartState() for portsyncd and vlanmgrd at appropriate places
```
    if (WarmStart::isWarmStart())
    {
           WarmStart::setWarmStartState("portsyncd", WarmStart::RECONCILED);
    }
```
or take other fixes.

rodnymolina: Are you saying that you want to transition these two modules to RECONCILED state even though there's no reconciliation phase for them? That sounds a bit strange, isn't it?

jipanyang: Current vs test for portsyncd and vlanmgrd warm restart is failing due to changes here,  it should be fixed.  There are multiple alternatives to the problem, please take the one you think the most appropriate.

lguohan:
retest this please

PR #601, opened 2018-08-28
Fix bug: NeighOrch waits for router interface ready

qiluo-msft:
The router interface creation may be delayed, so neighorch should leave task in m_toSync instead of erasing. Tested in both cold start and warm start.

### DIFF
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -293,7 +293,7 @@ void NeighOrch::doTask(Consumer &consumer)
         if (!p.m_rif_id)
         {
             SWSS_LOG_INFO("Router interface doesn't exist on %s", alias.c_str());
-            it = consumer.m_toSync.erase(it);
+            it++;
             continue;
         }
 


jipanyang:
How did you hit this problem?  For warm start, we have tried to load and process them in order.

If it is because of some race condition in regular operation,  say,  route interface deleted but arp message comes late to orchagent,  it  might cause a persistent entry left there with this change.

qiluo-msft:
It can be easily repro in warm start. I believe there is no race condition. The starting process is:
1. BufferOrch: blocked by gPortsOrch->getPort
2. PortsOrch: blocked by gBufferOrch->isPortReady
3. IntfsOrch: blocked by gPortsOrch->isInitDone
4. NeighOrch: Since router interface creation is delayed in previous step, the task in m_toSync is erased (This is the bug)
5. ...
6. BufferOrch: finish tasks
7. PortsOrch: finish tasks
8. IntfsOrch: finish tasks
9. NeighOrch: no tasks at all
10. RouteOrch: blocked by m_neighOrch->hasNextHop
So the warm start will leave many routes unprocessed.

I don't have idea of your proposed situation, and never met it. In my investigation, this is really a bug.

jipanyang:
neighsyncd and intfsyncd are two different processes, there is no guarantee of processing order, also the messages go through DB interface.
It is possible that arp entry for an interface gets processed after the interface has been deleted, then orchagent will try to process this arp entry for ever.  It is unlikely to happen, could we exclude such possibility in current code flow?    

It looks all the problems are caused by the inter-dependency between portorch and bufforch.   
instead of checking  gPortsOrch->isInitDone() only,  
```
    if (!gPortsOrch->isInitDone())
    {
        return;
    }
```

both gPortsOrch->isInitDone() and gBufferOrch->isInitDone()(`TODO`) should be checked for doTask() of other orch.
```
    if (!gPortsOrch->isInitDone() || !!gBufferOrch->isInitDone())
    {
        return;
    }
```
This way, let portorch and bufforch finish their internal fight,  don't bring others down with them.

qiluo-msft:
> If it is because of some race condition in regular operation, say, route interface deleted but arp message comes late to orchagent, it might cause a persistent entry left there with this change.

Under this assumption with my PR, a new neighbor will try to created on a non router interface port, it will be left in m_toSync. However, netlink will soon update with a deleting message, and Consumer::addToSync() will make overwrite the previous one with this new DEL one.

> It looks all the problems are caused by the inter-dependency between portorch and bufforch.

This is not the root cause. The root cause is risk between neighbor messages and router interface messages. In future, the router interface may be configurable, so the risk may happen more frequently.

jipanyang:
I agree with you that if arp delete message always comes after arp new message,   then no need to worry about persistent neigh event in orchagent.

The issue with warm reboot PortInitDone is that it is available immediately in the DB read and gPortsOrch->isInitDone() is set to true.   while actually the port level processing is pending on finish of bufferorch. 

While most other orches only check gPortsOrch->isInitDone() and assume everything about port has been ready, this brings too much dynamics and potential issues into warm start, and even cold boot.

PR #602, opened 2018-08-29
Routing-stack warm-reboot feature.

rodnymolina:
[ dependencies with PR/600: https://github.com/Azure/sonic-swss/pull/600 ]

Please refer to the corresponding design document for more details: https://github.com/Azure/SONiC/pull/239

The following manual UT plan has been successfully executed. UT automation pending.

Physical topology formed by various BGP peers connecting to the DUT. Both non-ecmp and ecmp prefixes are learned/tested.

```
Testcase Suite 1: Making use of “pkill -9 bgpd && pkill -9 zebra” as trigger.
=============

IPv4 prefixes
==========

    * Restart zebra/bgpd:
            - Verify that all prefixes are properly stale-marked and that no
    change is pushed to AppDB during reconciliation.
            - Result: PASSED

    * Restart zebra/bgpd and add one new non-ecmp IPv4 prefix.
            - To produce a route-state-inconsistency, add prefix in adjacent
    node before bgp sessions are re-established.
            - Verify that new-prefix msg is NOT pushed down to AppDB till
    reconciliation takes place.
            - Result: PASSED

    * Restart zebra/bgpd and withdraw one non-ecmp IPv4 prefix.
            - To produce a route-state-inconsistency, remove prefix in adjacent
    node before bgp sessions are re-established.
            - Verify that deleted-prefix msg is NOT pushed down to AppDB till
    reconciliation takes place.
            - Result: PASSED

    * Restart zebra/bgpd and add one new path to an IPv4 ecmp-prefix.
            - To produce a route-state-inconsistency, add prefix-path in
    adjacent node before bgp sessions are re-established.
            - Verify that new prefix-path msg is NOT pushed down to AppDB till
    reconciliation takes place.
            - Result: PASSED

    * Restart zebra/bgpd and withdraw one ecmp-path from an existing ecmp IPv4
      prefix.
              - To produce a route-state-inconsistency, remove prefix-path in
      adjacent node before bgp sessions are re-established.
              - Verify that deleted-prefix-path msg is NOT pushed down to AppDB
      till reconciliation takes place.
              - Eventually, during reconciliation, this path will be eliminated
      as it’s not being refreshed by remote-peer.
              - Result: PASSED

IPv6 prefixes
==========

    * Restart zebra/bgpd:
            - Verify that all prefixes are properly stale-marked and that no
    change is pushed to AppDB during reconciliation.
            - Result: PASSED

    * Restart zebra/bgpd and add one new non-ecmp IPv6 prefix.
            - To produce a route-state-inconsistency, add prefix in adjacent
    node before bgp sessions are re-established.
            - Verify that new-prefix msg is NOT pushed down to AppDB till
    reconciliation takes place.
            - Result: PASSED

    * Restart zebra/bgpd and withdraw one non-ecmp IPv6 prefix.
            - To produce a route-state-inconsistency, remove prefix in adjacent
    node before bgp sessions are re-established.
            - Verify that deleted-prefix msg is NOT pushed down to AppDB till
    reconciliation takes place.
            - Result: PASSED

    * Restart zebra/bgpd and add one new path to an IPv6 ecmp-prefix.
            - To produce a route-state-inconsistency, add prefix-path in
    adjacent node before bgp sessions are re-established.
            - Verify that new prefix-path msg is NOT pushed down to AppDB till
    reconciliation takes place.
            - Result: PASSED

    * Restart zebra/bgpd and withdraw one ecmp-path from an existing ecmp IPv6
      prefix.
              - To produce a route-state-inconsistency, remove prefix-path in
      adjacent node before bgp sessions are re-established.
              - Verify that deleted-prefix-path msg is NOT pushed down to AppDB
      till reconciliation takes place.
              - Eventually, during reconciliation, this path will be eliminated
      as it’s not being refreshed by remote-peer.
              - Result: PASSED

Testcase Suite 2: Making use of sudo service bgp restart” as trigger.
=============

IPv4 prefixes
==========

    * Restart bgp service/docker:
            - Verify that all prefixes are properly stale-marked and that no
    change is pushed to AppDB during reconciliation.
            - Result: PASSED

    * Restart bgp service/docker and add one new non-ecmp IPv4 prefix.
            - To produce a route-state-inconsistency, add prefix in adjacent
    node before bgp sessions are re-established.
            - Verify that new-prefix msg is NOT pushed down to AppDB till
    reconciliation takes place.
            - Result: PASSED

    * Restart bgp service/docker and withdraw one non-ecmp IPv4 prefix.
            - To produce a route-state-inconsistency, remove prefix in adjacent
    node before bgp sessions are re-established.
            - Verify that deleted-prefix msg is NOT pushed down to AppDB till
    reconciliation takes place.
            - Result: PASSED

    * Restart bgp service/docker and add one new path to an IPv4 ecmp-prefix.
            - To produce a route-state-inconsistency, add prefix-path in
    adjacent node before bgp sessions are re-established.
            - Verify that new prefix-path msg is NOT pushed down to AppDB till
    reconciliation takes place.
            - Result: PASSED

    * Restart bgp service/docker and withdraw one ecmp-path from an existing
      ecmp IPv4 prefix.
              - To produce a route-state-inconsistency, remove prefix-path in
      adjacent node before bgp sessions are re-established.
              - Verify that deleted-prefix-path msg is NOT pushed down to AppDB
      till reconciliation takes place.
              - Eventually, during reconciliation, this path will be eliminated
      as it’s not being refreshed by remote-peer.
              - Result: PASSED

IPv6 prefixes
==========

    * Restart bgp service/docker:
            - Verify that all prefixes are properly stale-marked and that no
    change is pushed to AppDB during reconciliation.
            - Result: PASSED

    * Restart bgp service/docker and add one new non-ecmp IPv6 prefix.
            - To produce a route-state-inconsistency, add prefix in adjacent
    node before bgp sessions are re-established.
            - Verify that new-prefix msg is NOT pushed down to AppDB till
    reconciliation takes place.
            - Result: PASSED

    * Restart bgp service/docker and withdraw one non-ecmp IPv6 prefix.
            - To produce a route-state-inconsistency, remove prefix in adjacent
    node before bgp sessions are re-established.
            - Verify that deleted-prefix msg is NOT pushed down to AppDB till
    reconciliation takes place.
            - Result: PASSED

    * Restart bgp service/docker and add one new path to an IPv6 ecmp-prefix.
            - To produce a route-state-inconsistency, add prefix-path in
    adjacent node before bgp sessions are re-established.
            - Verify that new prefix-path msg is NOT pushed down to AppDB till
    reconciliation takes place.
            - Result: PASSED

    * Restart bgp service/docker and withdraw one ecmp-path from an existing
      ecmp IPv6 prefix.
              - To produce a route-state-inconsistency, remove prefix-path in
      adjacent node before bgp sessions are re-established.
              - Verify that deleted-prefix-path msg is NOT pushed down to AppDB
      till reconciliation takes place.
              - Eventually, during reconciliation, this path will be eliminated
      as it’s not being refreshed by remote-peer.
              - Result: PASSED
```

Signed-off-by: Rodny Molina <rmolina@linkedin.com>

### DIFF
--- a/fpmsyncd/fpmsyncd.cpp
+++ b/fpmsyncd/fpmsyncd.cpp
@@ -35,8 +35,13 @@ int main(int argc, char **argv)
         {
             FpmLink fpm;
             Select s;
-
             SelectableTimer warmStartTimer(timespec{0, 0});
+
+            /*
+             * Pipeline should be flushed right away to deal with state pending
+             * from previous try/catch iterations.
+             */
+            pipeline.flush();
 
             cout << "Waiting for fpm-client connection..." << endl;
             fpm.accept();
@@ -44,7 +49,7 @@ int main(int argc, char **argv)
 
             s.addSelectable(&fpm);
 
-            /* Initialize warm-restart logic if this one is enabled */
+            /* If warm-restart feature is enabled, execute 'restoration' logic */
             bool warmStartEnabled = sync.m_warmStartHelper.isEnabled();
             if (warmStartEnabled)
             {
@@ -59,8 +64,8 @@ int main(int argc, char **argv)
                     warmStartTimer.setInterval(timespec{warmRestartIval, 0});
                 }
 
-                /* Execute recovery instruction and kick off warm-restart timer */
-                if (sync.m_warmStartHelper.runRecovery())
+                /* Execute restoration instruction and kick off warm-restart timer */
+                if (sync.m_warmStartHelper.runRestoration())
                 {
                     warmStartTimer.start();
                     s.addSelectable(&warmStartTimer);
@@ -82,16 +87,16 @@ int main(int argc, char **argv)
                 if (warmStartEnabled && temps == &warmStartTimer)
                 {
                     SWSS_LOG_NOTICE("Warm-Restart timer expired.");
-                    sync.m_warmStartHelper.reconciliate();
+                    sync.m_warmStartHelper.reconcile();
                     s.removeSelectable(&warmStartTimer);
 
                     pipeline.flush();
-                    SWSS_LOG_NOTICE("Pipeline flushed");
+                    SWSS_LOG_DEBUG("Pipeline flushed");
                 }
                 else if (!warmStartEnabled || sync.m_warmStartHelper.isReconciled())
                 {
                     pipeline.flush();
-                    SWSS_LOG_NOTICE("Pipeline flushed");
+                    SWSS_LOG_DEBUG("Pipeline flushed");
                 }
             }
         }
--- a/fpmsyncd/fpmsyncd.cpp
+++ b/fpmsyncd/fpmsyncd.cpp
@@ -1,12 +1,23 @@
 #include <iostream>
 #include "logger.h"
 #include "select.h"
+#include "selectabletimer.h"
 #include "netdispatcher.h"
+#include "warmRestartHelper.h"
 #include "fpmsyncd/fpmlink.h"
 #include "fpmsyncd/routesync.h"
 
+
 using namespace std;
 using namespace swss;
+
+
+/*
+ * Default warm-restart timer interval for routing-stack app. To be used only if
+ * no explicit value has been defined in configuration.
+ */
+const uint32_t DEFAULT_ROUTING_RESTART_INTERVAL = 120;
+
 
 int main(int argc, char **argv)
 {
@@ -18,25 +29,70 @@ int main(int argc, char **argv)
     NetDispatcher::getInstance().registerMessageHandler(RTM_NEWROUTE, &sync);
     NetDispatcher::getInstance().registerMessageHandler(RTM_DELROUTE, &sync);
 
-    while (1)
+    while (true)
     {
         try
         {
             FpmLink fpm;
             Select s;
 
-            cout << "Waiting for connection..." << endl;
+            SelectableTimer warmStartTimer(timespec{0, 0});
+
+            cout << "Waiting for fpm-client connection..." << endl;
             fpm.accept();
             cout << "Connected!" << endl;
 
             s.addSelectable(&fpm);
+
+            /* Initialize warm-restart logic if this one is enabled */
+            bool warmStartEnabled = sync.m_warmStartHelper.isEnabled();
+            if (warmStartEnabled)
+            {
+                /* Obtain warm-restart timer defined for routing application */
+                uint32_t warmRestartIval = sync.m_warmStartHelper.getRestartTimer();
+                if (!warmRestartIval)
+                {
+                    warmStartTimer.setInterval(timespec{DEFAULT_ROUTING_RESTART_INTERVAL, 0});
+                }
+                else
+                {
+                    warmStartTimer.setInterval(timespec{warmRestartIval, 0});
+                }
+
+                /* Execute recovery instruction and kick off warm-restart timer */
+                if (sync.m_warmStartHelper.runRecovery())
+                {
+                    warmStartTimer.start();
+                    s.addSelectable(&warmStartTimer);
+                }
+            }
+
             while (true)
             {
                 Selectable *temps;
-                /* Reading FPM messages forever (and calling "readData" to read them) */
+
+                /* Reading FPM messages forever (and calling "readMe" to read them) */
                 s.select(&temps);
-                pipeline.flush();
-                SWSS_LOG_DEBUG("Pipeline flushed");
+
+                /*
+                 * Upon expiration of the warm-restart timer, proceed to run the
+                 * reconciliation process and remove warm-restart timer from
+                 * select() loop.
+                 */
+                if (warmStartEnabled && temps == &warmStartTimer)
+                {
+                    SWSS_LOG_NOTICE("Warm-Restart timer expired.");
+                    sync.m_warmStartHelper.reconciliate();
+                    s.removeSelectable(&warmStartTimer);
+
+                    pipeline.flush();
+                    SWSS_LOG_NOTICE("Pipeline flushed");
+                }
+                else if (!warmStartEnabled || sync.m_warmStartHelper.isReconciled())
+                {
+                    pipeline.flush();
+                    SWSS_LOG_NOTICE("Pipeline flushed");
+                }
             }
         }
         catch (FpmLink::FpmConnectionClosedException &e)
--- a/fpmsyncd/fpmsyncd.cpp
+++ b/fpmsyncd/fpmsyncd.cpp
@@ -50,7 +50,7 @@ int main(int argc, char **argv)
             s.addSelectable(&fpm);
 
             /* If warm-restart feature is enabled, execute 'restoration' logic */
-            bool warmStartEnabled = sync.m_warmStartHelper.isEnabled();
+            bool warmStartEnabled = sync.m_warmStartHelper.checkAndStart();
             if (warmStartEnabled)
             {
                 /* Obtain warm-restart timer defined for routing application */
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -15,7 +15,7 @@ using namespace swss;
 
 RouteSync::RouteSync(RedisPipeline *pipeline) :
     m_routeTable(pipeline, APP_ROUTE_TABLE_NAME, true),
-    m_warmStartHelper(pipeline, &m_routeTable, "bgp", "bgp")
+    m_warmStartHelper(pipeline, &m_routeTable, APP_ROUTE_TABLE_NAME, "bgp", "bgp")
 {
     m_nl_sock = nl_socket_alloc();
     nl_connect(m_nl_sock, NETLINK_ROUTE);
@@ -41,26 +41,24 @@ void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
 
     /*
      * Upon arrival of a delete msg we could either push the change right away,
-     * or we could opt to defer it if we are in the middle of a warm-reboot
-     * process. The goal here is to avoid unnecessary churn in swss/syncd layers.
+     * or we could opt to defer it if we are going through a warm-reboot cycle.
      */
-    auto warmState = m_warmStartHelper.getState();
+    bool warmRestartInProgress = m_warmStartHelper.inProgress();
 
     if (nlmsg_type == RTM_DELROUTE)
     {
-        if (warmState == WarmStart::INITIALIZED ||
-            warmState == WarmStart::RECONCILED)
+        if (!warmRestartInProgress)
         {
             m_routeTable.del(destipprefix);
             return;
         }
         else
         {
-            SWSS_LOG_INFO("Warm-Restart: Receiving delete msg: %s\n", destipprefix);
+            SWSS_LOG_INFO("Warm-Restart mode: Receiving delete msg: %s\n", destipprefix);
 
             vector<FieldValueTuple> fvVector;
             const KeyOpFieldsValuesTuple kfv = std::make_tuple(destipprefix, "", fvVector);
-            m_warmStartHelper.removeRecoveryMap(kfv, WarmStartHelper::DELETE);
+            m_warmStartHelper.removeRestorationMap(kfv, WarmStartHelper::DELETE);
             return;
         }
     }
@@ -143,8 +141,7 @@ void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
     fvVector.push_back(nh);
     fvVector.push_back(idx);
 
-    if (warmState == WarmStart::INITIALIZED ||
-        warmState == WarmStart::RECONCILED)
+    if (!warmRestartInProgress)
     {
         m_routeTable.set(destipprefix, fvVector);
         SWSS_LOG_DEBUG("RouteTable set msg: %s %s %s\n",
@@ -157,10 +154,10 @@ void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
      */
     else
     {
-        SWSS_LOG_INFO("Warm-Restart: RouteTable set msg: %s %s %s\n",
+        SWSS_LOG_INFO("Warm-Restart mode: RouteTable set msg: %s %s %s\n",
                       destipprefix, nexthops.c_str(), ifnames.c_str());
 
         const KeyOpFieldsValuesTuple kfv = std::make_tuple(destipprefix, "", fvVector);
-        m_warmStartHelper.insertRecoveryMap(kfv, WarmStartHelper::CLEAN);
+        m_warmStartHelper.insertRestorationMap(kfv, WarmStartHelper::CLEAN);
     }
 }
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -14,7 +14,8 @@ using namespace std;
 using namespace swss;
 
 RouteSync::RouteSync(RedisPipeline *pipeline) :
-    m_routeTable(pipeline, APP_ROUTE_TABLE_NAME, true)
+    m_routeTable(pipeline, APP_ROUTE_TABLE_NAME, true),
+    m_warmStartHelper(pipeline, &m_routeTable, "bgp", "bgp")
 {
     m_nl_sock = nl_socket_alloc();
     nl_connect(m_nl_sock, NETLINK_ROUTE);
@@ -38,10 +39,30 @@ void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
         return;
     }
 
+    /*
+     * Upon arrival of a delete msg we could either push the change right away,
+     * or we could opt to defer it if we are in the middle of a warm-reboot
+     * process. The goal here is to avoid unnecessary churn in swss/syncd layers.
+     */
+    auto warmState = m_warmStartHelper.getState();
+
     if (nlmsg_type == RTM_DELROUTE)
     {
-        m_routeTable.del(destipprefix);
-        return;
+        if (warmState == WarmStart::INITIALIZED ||
+            warmState == WarmStart::RECONCILED)
+        {
+            m_routeTable.del(destipprefix);
+            return;
+        }
+        else
+        {
+            SWSS_LOG_INFO("Warm-Restart: Receiving delete msg: %s\n", destipprefix);
+
+            vector<FieldValueTuple> fvVector;
+            const KeyOpFieldsValuesTuple kfv = std::make_tuple(destipprefix, "", fvVector);
+            m_warmStartHelper.removeRecoveryMap(kfv, WarmStartHelper::DELETE);
+            return;
+        }
     }
     else if (nlmsg_type != RTM_NEWROUTE)
     {
@@ -118,8 +139,28 @@ void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
     vector<FieldValueTuple> fvVector;
     FieldValueTuple nh("nexthop", nexthops);
     FieldValueTuple idx("ifname", ifnames);
+
     fvVector.push_back(nh);
     fvVector.push_back(idx);
-    m_routeTable.set(destipprefix, fvVector);
-    SWSS_LOG_DEBUG("RoutTable set: %s %s %s\n", destipprefix, nexthops.c_str(), ifnames.c_str());
+
+    if (warmState == WarmStart::INITIALIZED ||
+        warmState == WarmStart::RECONCILED)
+    {
+        m_routeTable.set(destipprefix, fvVector);
+        SWSS_LOG_DEBUG("RouteTable set msg: %s %s %s\n",
+                       destipprefix, nexthops.c_str(), ifnames.c_str());
+    }
+
+    /*
+     * During routing-stack restarting scenarios route-updates will be temporarily
+     * put on hold by warm-reboot logic.
+     */
+    else
+    {
+        SWSS_LOG_INFO("Warm-Restart: RouteTable set msg: %s %s %s\n",
+                      destipprefix, nexthops.c_str(), ifnames.c_str());
+
+        const KeyOpFieldsValuesTuple kfv = std::make_tuple(destipprefix, "", fvVector);
+        m_warmStartHelper.insertRecoveryMap(kfv, WarmStartHelper::CLEAN);
+    }
 }
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -54,11 +54,14 @@ void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
         }
         else
         {
-            SWSS_LOG_INFO("Warm-Restart mode: Receiving delete msg: %s\n", destipprefix);
+            SWSS_LOG_INFO("Warm-Restart mode: Receiving delete msg: %s\n",
+                          destipprefix);
 
             vector<FieldValueTuple> fvVector;
-            const KeyOpFieldsValuesTuple kfv = std::make_tuple(destipprefix, "", fvVector);
-            m_warmStartHelper.removeRestorationMap(kfv, WarmStartHelper::DELETE);
+            const KeyOpFieldsValuesTuple kfv = std::make_tuple(destipprefix,
+                                                               DEL_COMMAND,
+                                                               fvVector);
+            m_warmStartHelper.insertRefreshMap(kfv);
             return;
         }
     }
@@ -157,7 +160,9 @@ void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
         SWSS_LOG_INFO("Warm-Restart mode: RouteTable set msg: %s %s %s\n",
                       destipprefix, nexthops.c_str(), ifnames.c_str());
 
-        const KeyOpFieldsValuesTuple kfv = std::make_tuple(destipprefix, "", fvVector);
-        m_warmStartHelper.insertRestorationMap(kfv, WarmStartHelper::CLEAN);
+        const KeyOpFieldsValuesTuple kfv = std::make_tuple(destipprefix,
+                                                           SET_COMMAND,
+                                                           fvVector);
+        m_warmStartHelper.insertRefreshMap(kfv);
     }
 }
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -365,7 +365,7 @@ class DockerVirtualSwitch(object):
         return marker
 
     def SubscribeAppDbObject(self, objpfx):
-        r = redis.Redis(unix_socket_path=self.redis_sock, db=swsscommon.APP_DB)
+        r = redis.Redis(unix_socket_path=self.redis_sock, db=swsscommon.APPL_DB)
         pubsub = r.pubsub()
         pubsub.psubscribe("__keyspace@0__:%s*" % objpfx)
         return pubsub
@@ -400,26 +400,40 @@ class DockerVirtualSwitch(object):
         return (nadd, ndel)
 
     def GetSubscribedAppDbObjects(self, pubsub, ignore=None, timeout=10):
-        r = redis.Redis(unix_socket_path=self.redis_sock, db=swsscommon.APP_DB)
+        r = redis.Redis(unix_socket_path=self.redis_sock, db=swsscommon.APPL_DB)
 
         addobjs = []
         delobjs = []
         idle = 0
+        prev_key = None
 
         while True and idle < timeout:
             message = pubsub.get_message()
             if message:
                 print message
                 key = message['channel'].split(':', 1)[1]
+                # In producer/consumer_state_table scenarios, every entry will
+                # show up twice for every push/pop operation, so skip the second
+                # one to avoid double counting.
+                if key != None and key == prev_key:
+                    continue
+                # Skip instructions with meaningless keys. To be extended in the
+                # future to other undesired keys.
+                if key == "ROUTE_TABLE_KEY_SET" or key == "ROUTE_TABLE_DEL_SET":
+                    continue
                 if ignore:
                     fds = message['channel'].split(':')
                     if fds[2] in ignore:
                         continue
+
                 if message['data'] == 'hset':
+                    (_, k) = key.split(':', 1)
                     value=r.hgetall(key)
-                    addobjs.append({'key':k, 'vals':value})
+                    addobjs.append({'key':json.dumps(k), 'vals':json.dumps(value)})
+                    prev_key = key
                 elif message['data'] == 'del':
-                    delobjs.append(key)
+                    (_, k) = key.split(':', 1)
+                    delobjs.append({'key':json.dumps(k)})
                 idle = 0
             else:
                 time.sleep(1)
@@ -449,7 +463,8 @@ class DockerVirtualSwitch(object):
                     (_, t, k) = key.split(':', 2)
                     addobjs.append({'type':t, 'key':k, 'vals':value})
                 elif message['data'] == 'del':
-                    delobjs.append(key)
+                    (_, t, k) = key.split(':', 2)
+                    delobjs.append({'key':k})
                 idle = 0
             else:
                 time.sleep(1)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -290,6 +290,26 @@ class DockerVirtualSwitch(object):
             cmd += "supervisorctl stop {}; ".format(pname)
         self.runcmd(['sh', '-c', cmd])
 
+    def start_zebra(dvs):
+        dvs.runcmd(['sh', '-c', 'supervisorctl start zebra'])
+
+        # Let's give zebra a chance to connect to FPM.
+        time.sleep(5)
+
+    def stop_zebra(dvs):
+        dvs.runcmd(['sh', '-c', 'pkill -x zebra'])
+        time.sleep(1)
+
+    def start_fpmsyncd(dvs):
+        dvs.runcmd(['sh', '-c', 'supervisorctl start fpmsyncd'])
+
+        # Let's give fpmsyncd a chance to connect to Zebra.
+        time.sleep(5)
+
+    def stop_fpmsyncd(dvs):
+        dvs.runcmd(['sh', '-c', 'pkill -x fpmsyncd'])
+        time.sleep(1)
+
     def init_asicdb_validator(self):
         self.asicdb = AsicDbValidator(self)
 
@@ -334,9 +354,14 @@ class DockerVirtualSwitch(object):
             raise RuntimeError("Failed to unpack the archive.")
         os.system("chmod a+r -R log")
 
-    def add_log_marker(self):
+    def add_log_marker(self, file=None):
         marker = "=== start marker {} ===".format(datetime.now().isoformat())
-        self.ctn.exec_run("logger {}".format(marker))
+
+        if file:
+            self.runcmd(['sh', '-c', "echo \"{}\" >> {}".format(marker, file)])
+        else:
+            self.ctn.exec_run("logger {}".format(marker))
+
         return marker
 
     def SubscribeAppDbObject(self, objpfx):
--- a/tests/test_warm_reboot.py
+++ b/tests/test_warm_reboot.py
@@ -708,6 +708,9 @@ def test_OrchagentWarmRestartReadyCheck(dvs, testlog):
     (exitcode, result) =  dvs.runcmd("/usr/bin/orchagent_restart_check -n -s -w 500")
     assert result == "RESTARTCHECK failed\n"
 
+    # Cleaning previously pushed route-entry to ease life of subsequent testcases.
+    del_entry_tbl(appl_db, swsscommon.APP_ROUTE_TABLE_NAME, "2.2.2.0/24")
+
     # recover for test cases after this one.
     dvs.stop_swss()
     dvs.start_swss()
@@ -823,26 +826,6 @@ def stop_fpmsyncd(dvs):
     time.sleep(1)
 
 
-def enable_warmrestart(dvs, db, app_name):
-    create_entry_tbl(
-        db,
-        swsscommon.CFG_WARM_RESTART_TABLE_NAME, app_name,
-        [
-            ("enable", "true"),
-        ]
-    )
-
-
-def disable_warmrestart(dvs, db, app_name):
-    create_entry_tbl(
-        db,
-        swsscommon.CFG_WARM_RESTART_TABLE_NAME, app_name,
-        [
-            ("enable", "false"),
-        ]
-    )
-
-
 def set_restart_timer(dvs, db, app_name, value):
     create_entry_tbl(
         db,
@@ -956,6 +939,18 @@ def check_restart_timer(dvs, timer_value):
     assert num.strip() == timer_value
 
 
+# Temporary instruction to activate warm_restart. To be deleted once equivalent CLI
+# function is pushed to sonic-utils.
+def enable_warmrestart(dvs, db, app_name):
+    create_entry_tbl(
+        db,
+        swsscommon.CFG_WARM_RESTART_TABLE_NAME, app_name,
+        [
+            ("enable", "true"),
+        ]
+    )
+
+
 ################################################################################
 #
 # Routing warm-restart testcases
@@ -967,12 +962,6 @@ def test_routing_WarmRestart(dvs):
     appl_db = swsscommon.DBConnector(swsscommon.APPL_DB, dvs.redis_sock, 0)
     conf_db = swsscommon.DBConnector(swsscommon.CONFIG_DB, dvs.redis_sock, 0)
     state_db = swsscommon.DBConnector(swsscommon.STATE_DB, dvs.redis_sock, 0)
-
-    # Cleaning AppDB entry introduced in a previous test-case (no need to
-    # clean anything else).
-    ps = swsscommon.ProducerStateTable(appl_db, swsscommon.APP_ROUTE_TABLE_NAME)
-    ps._del("2.2.2.0/24")
-    time.sleep(1)
 
     # Restart-timer to utilize during the following testcases
     restart_timer = 10
@@ -1057,7 +1046,7 @@ def test_routing_WarmRestart(dvs):
     #
     #############################################################################
 
-    # Enabling bgp warmrestart and setting restart timer.
+    # Clear logs and restart zebra
     clear_logs(dvs)
     stop_zebra(dvs)
     start_zebra(dvs)
@@ -1080,8 +1069,13 @@ def test_routing_WarmRestart(dvs):
 
 
     # Enabling bgp warmrestart and setting restart timer.
+    # The following two instructions will be substituted by the commented ones
+    # once the later ones are added to sonic-utilities repo.
     enable_warmrestart(dvs, conf_db, "bgp")
     set_restart_timer(dvs, conf_db, "bgp", str(restart_timer))
+    #dvs.runcmd("config warm_restart enable bgp")
+    #dvs.runcmd("config warm_restart bgp_timer {}".format(restart_timer))
+
     time.sleep(1)
 
     # Clear logs and restart zebra
--- a/tests/test_warm_reboot.py
+++ b/tests/test_warm_reboot.py
@@ -494,7 +494,7 @@ def test_swss_neighbor_syncup(dvs, testlog):
 
     #
     # Testcase 4:
-    # Stop neighsyncd, add even nummber of ipv4/ipv6 neighbor entries to each interface again, 
+    # Stop neighsyncd, add even nummber of ipv4/ipv6 neighbor entries to each interface again,
     # Start neighsyncd
     # The neighsyncd is supposed to sync up the entries from kernel after warm restart
     # Check the timer is not retrieved from configDB since it is not configured
@@ -802,30 +802,6 @@ def clear_logs(dvs):
     dvs.runcmd(['sh', '-c', '> /var/log/swss/sairedis.rec'])
 
 
-def start_zebra(dvs):
-    dvs.runcmd(['sh', '-c', 'supervisorctl start zebra'])
-
-    # Let's give zebra a chance to connect to FPM.
-    time.sleep(5)
-
-
-def stop_zebra(dvs):
-    dvs.runcmd(['sh', '-c', 'pkill -x zebra'])
-    time.sleep(1)
-
-
-def start_fpmsyncd(dvs):
-    dvs.runcmd(['sh', '-c', 'supervisorctl start fpmsyncd'])
-
-    # Let's give zebra a chance to connect to Zebra.
-    time.sleep(5)
-
-
-def stop_fpmsyncd(dvs):
-    dvs.runcmd(['sh', '-c', 'pkill -x fpmsyncd'])
-    time.sleep(1)
-
-
 def set_restart_timer(dvs, db, app_name, value):
     create_entry_tbl(
         db,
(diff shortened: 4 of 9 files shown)


fpmsyncd/fpmsyncd.cpp
```diff
@@ -18,25 +29,70 @@ int main(int argc, char **argv)
...
+                }
+
+                /* Execute recovery instruction and kick off warm-restart timer */
+                if (sync.m_warmStartHelper.runRecovery())
+                {
+                    warmStartTimer.start();
+                    s.addSelectable(&warmStartTimer);
+                }
+            }
+
```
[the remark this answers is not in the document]
rodnymolina: @nikos-github thanks for your comments.

You have a point here. For the above to happen we need an unlikely sequence of events to take place:

   1) two or more zebra crashes happening in a small window of time, and
   2) a user deactivating routing-warm-reboot feature **right** between the two zebra crashes

Unlikely but possible. I'll adjust my code to deal with this scenario.

fpmsyncd/fpmsyncd.cpp
```diff
@@ -18,25 +29,70 @@ int main(int argc, char **argv)
...
+                {
+                    warmStartTimer.setInterval(timespec{DEFAULT_ROUTING_RESTART_INTERVAL, 0});
+                }
+                else
+                {
+                    warmStartTimer.setInterval(timespec{warmRestartIval, 0});
+                }
+
+                /* Execute recovery instruction and kick off warm-restart timer */
+                if (sync.m_warmStartHelper.runRecovery())
```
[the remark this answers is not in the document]
rodnymolina: Yep, already took care of that, my next patch will have it.

fpmsyncd/fpmsyncd.cpp
```diff
@@ -18,25 +29,70 @@ int main(int argc, char **argv)
...
+                    sync.m_warmStartHelper.reconciliate();
+                    s.removeSelectable(&warmStartTimer);
+
+                    pipeline.flush();
+                    SWSS_LOG_NOTICE("Pipeline flushed");
+                }
+                else if (!warmStartEnabled || sync.m_warmStartHelper.isReconciled())
+                {
+                    pipeline.flush();
+                    SWSS_LOG_NOTICE("Pipeline flushed");
```
jipanyang: Will there be too many logs for pipeline flush in normal running?

fpmsyncd/routesync.cpp
```diff
@@ -38,10 +39,30 @@ void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
...
+     * process. The goal here is to avoid unnecessary churn in swss/syncd layers.
+     */
+    auto warmState = m_warmStartHelper.getState();
+
     if (nlmsg_type == RTM_DELROUTE)
     {
-        m_routeTable.del(destipprefix);
-        return;
+        if (warmState == WarmStart::INITIALIZED ||
+            warmState == WarmStart::RECONCILED)
```
jipanyang: "warmState == WarmStart::INITIALIZED",  It is confusing,  while warm state is INITIALIZED and route is injected to appDB directly.   Do you mean warm start has been disabled?

rodnymolina: My understanding, based on a recent discussion where we discussed/agreed on this issue, is that INITIALIZED state will be utilized for applications with warm-reboot feature disabled -- so these ones will permanently display INITIALIZED state. Based on that discussion i had to re-adjust my code and add this INITIALIZED checkpoint. My original code was setting all applications to RECONCILED state, regardless of WarmReboot being enable or not, as this is the last FSM stage. Looks like our problem here is simply a consequence of INITIALIZED state being slightly overloaded. The simplest solution is to provided another FSM state (DEACTIVATED). Will ping you offline to discuss FSM semantics.

jipanyang: Maybe I missed some of the discussions.  My understanding for  WarmStart state is that it is only meaningful when warm start has been enabled.  If warm start is disabled,  people won't care about WarmStart state.
Putting too much state change seems to make the problem unnecessary complicated.

rodnymolina: As you know, applications like fpmSyncd need to be able to invoke WarmRestart logic on-demand, and not only when when it boots-up. So the initialization of WarmRestart logic is not necessarily connected to its activation: you could have WarmRestart being turned on at any given point in the future without having to restart fpmSyncd. To fix this problem there are two possible solutions: 1) we overload one of the current FSM states (INITIALIZED), or 2) we create a new UNINITIALIZED/DEACTIVATED FSM state.

jipanyang: For fpmsyncd case, whenever warm restore/reconcile is needed and warm start has been enabled, the INITIALIZED state could be set before starting restore to indicate that warm start    is initiated, after restore, RESTORED;  then RECONCILED.    

I don't see the need to overload INITIALIZED or add new DEACTIVATED state.

nikos-github: I suggest you use RECONCILED always. I agree with Jipan that INITIALIZED is confusing and at the same time UNINITIALIZED/DEACTIVATED will be equally confusing if warm restart is turned on after fpmsyncd is up because the state won't match with what's configured.

nikos-github: I'm trying to point out a distinction between FSM state and whether the feature is enabled or not. Trying to infer what's the feature's state (on or off) from the FSM isn't the right thing in my opinion. If you want to convey both then you either need 2 variables, no overloading of FSM state and introduce another state in the FSM, OR you need 1 variable that has states plus feature enabled/disabled (via a bitmask), no overloading of FSM state or use reconciled. If you don't want any of the above, then from my point of view reconciled is the right state whether the feature is disabled or later enabled.

fpmsyncd/fpmsyncd.cpp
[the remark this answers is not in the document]
nikos-github: If warm restart is found to be disabled, we should still recover and cleanup. There should be no requirement to disrupt the lower sonic layers now that we are adding such infra/functionality.

warmrestart/warmRestartHelper.cpp
```diff
@@ -0,0 +1,472 @@
...
+ * - An element in the recoveryMap with some of its entries in the fvRecMap
+ *   showing as NEW, will have these new entries, along with any CLEAN one,
+ *   being pushed to AppDB.
+ *
+ * - An element in the recoveryMap with some/all of its entries in the
+ *   fvRecMap showing as DELETE, will have these entries being eliminated
+ *   from AppDB.
+ *
+ */
+void WarmStartHelper::reconciliate(void)
```
nikos-github: reconcile not reconciliate

warmrestart/warmRestartHelper.cpp
```diff
@@ -0,0 +1,472 @@
...
+            m_syncTable->del(key);
+            SWSS_LOG_NOTICE("Route reconciliation: deleting withdrawn prefix %s\n",
+                            key.c_str());
+        }
+
+        it = m_recoveryMap.erase(it);
+    }
+
+    /* Recovery map should be entirely empty by now */
+    assert(m_recoveryMap.size() == 0);
```
nikos-github: We should recover gracefully here by falling back to cold restart.

fpmsyncd/fpmsyncd.cpp
rodnymolina: @jipanyang I'm just adding one extra "Pipeline flushed" log, which will be executed only once upon restart-timer expiration. During normal operation we will continue generating this log the same way we are doing it today -- see that i'm not adding this.

fpmsyncd/routesync.cpp
rodnymolina: @nikos-github just to close the loop here. As discussed with the rest of the folks working in warm-reboot, we will rely on the assumption that no FSM state implicitly indicates the lack of warm-reboot functionality, or the fact that this one hasn't been turned on. The three FSM states (initialized, restored, reconciled) will be exclusively used for apps that are warm-reboot capable.

warmrestart/warmRestartHelper.cpp
rodnymolina: done

warmrestart/warmRestartHelper.cpp
rodnymolina: Not sure what do you mean by 'gracefully'. I understand that we should do something about it, but that's something the process-manager (supervisord) should take care of upon the app being cold-booted after the failed assertion.

nikos-github: How will supervisord achieve a cold boot since warm restart is enabled? In any case, asserting is shouldn't be here in my opinion - that's way too harsh. Flushing the db and re-establishing the connection will cause the routes to re-download. That's why I mean by gracefully.

fpmsyncd/routesync.cpp
nikos-github: I have no problem with that. The state should be reconciled though when the app is warm reboot capable but warm reboot isn't enabled or is enabled later.

lguohan:
depend on #602

fpmsyncd/fpmsyncd.cpp
rodnymolina: Sure, i can set all "Pipeline flushed" logs to debug.

fpmsyncd/fpmsyncd.cpp
```diff
@@ -18,25 +29,75 @@ int main(int argc, char **argv)
...
+                if (warmStartEnabled && temps == &warmStartTimer)
+                {
+                    SWSS_LOG_NOTICE("Warm-Restart timer expired.");
+                    sync.m_warmStartHelper.reconcile();
+                    s.removeSelectable(&warmStartTimer);
+
+                    pipeline.flush();
+                    SWSS_LOG_NOTICE("Pipeline flushed");
+                }
+                else if (!warmStartEnabled || sync.m_warmStartHelper.isReconciled())
```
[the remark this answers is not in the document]
rodnymolina: Which other conditions are you referring to? There are only three conditions that we care about here: 1) warmStartTimer expired, which triggers reconciliation, 2) warm-restart disabled -- to deal with legacy/current scenarios, and 3) scenarios with warm-restart reconciliation completed. In 2) and 3) the actions are exactly the same, that's why i'm bundling them together. What would be the use case for the 'else'?

rodnymolina:
I'm working on VS test-cases right now.

Answering your other question: the goal is to replace the existing reconciliation logic with the one i'm presenting here (which offers more flexibility). That will require some refactoring work on neighsyncd which i'm planning to carry out once done with the routing portion. At the end there will be a single reconciliation class.

warmrestart/warmRestartHelper.h
```diff
@@ -0,0 +1,132 @@
...
+struct fvComparator
+{
+    bool operator()(const std::vector<FieldValueTuple> &left,
+                    const std::vector<FieldValueTuple> &right) const
+    {
+        /*
+         * The sizes of both tuple-vectors should always match within any given
+         * application, otherwise we are running into some form of bug.
+         */
+        assert(left.size() == right.size());
```
jipanyang: When comparing two vectors,  how to ensure they are of same size? 
Say, a route has two fields in usual case: nexthop and intf,   what if for some routes, the third fields "blackhole" is set?

```
### ROUTE_TABLE
    ;Stores a list of routes
    ;Status: Mandatory
    key           = ROUTE_TABLE:prefix
    nexthop       = *prefix, ;IP addresses separated “,” (empty indicates no gateway)
    intf          = ifindex? PORT_TABLE.key  ; zero or more separated by “,” (zero indicates no interface)
    blackhole     = BIT ; Set to 1 if this route is a blackhole (or null0)
```

fpmsyncd/fpmsyncd.cpp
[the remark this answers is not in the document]
jipanyang: I was not checking the internal logic.  If the "else if" covers everything,  why don't you just change it to else {}.

Based on your current implementation, it looks some default processing is missing. 
```
if (condition 1)
{
}
else if (condition2)
{
}
else {
// default processing;
}
```

warmrestart/warmRestartHelper.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+            if (recElemState == STALE)
+            {
+                itMap = fvMap.erase(itMap);
+                staleRecElems++;
+            }
+            else if (recElemState == CLEAN)
+            {
+                cleanRecElems++;
+                transformKFV(recElem, fvVector);
```
jipanyang: Is the order of fields in fvRestorationMap guaranteed? 
I saw you put comment in fvComparator
```
/*
             * Notice that we are forced to iterate through all the tuples in the
             * right-vector given that the order of fields within a tuple is not
             * fully deterministic (i.e. 'left' could hold 'nh: 1.1.1.1 / if: eth0'
             * and 'right' could be 'if: eth0, nh: 1.1.1.1').
             */
```

fpmsyncd/fpmsyncd.cpp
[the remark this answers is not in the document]
rodnymolina: Got your point now. The "else if" is there to make it more explicit to the reader that those two conditions could take place. Performance-wise there's no doubt the "else" is slightly cheaper. Have no strong opinion here, i can make use of an "else" if everyone agrees.

warmrestart/warmRestartHelper.h
rodnymolina: Good point. For routing-app there's this "blackhole" corner-case which brakes the uniformity for routing schema. Basically we are allowing a route to be defined as: "prefix/len | blackhole = true". No other app/scenarios that i'm aware of allow this inconsistency. In my humble opinion, we should keep uniformity by allowing the common schema to be re-utilized as this: "prefix/len | nh: blackhole | if: null". This approach will also simplify the addition of new attributes in the future (AFI, SAFI, etc).

Also, if you look at orchagent you'll see that we are not passing down "blackhole" routes -- they are being eliminated right there in routeOrch. The only blackhole route being pushed down to ASIC_DB is the default route, and this one is not even exposed to AppDB. So in reality, no one is really consuming these black-hole schema.

In my opinion, I think this assert() should be left here as it prevents real problems down the road. And concerning the blackhole routes, i would like to make a very minor adjustment in its representation to avoid this uniformity i talked about.

warmrestart/warmRestartHelper.h
```diff
@@ -0,0 +1,132 @@
...
+             * Notice that we are forced to iterate through all the tuples in the
+             * right-vector given that the order of fields within a tuple is not
+             * fully deterministic (i.e. 'left' could hold 'nh: 1.1.1.1 / if: eth0'
+             * and 'right' could be 'if: eth0, nh: 1.1.1.1').
+             */
+            for (auto &fvRight : right)
+            {
+                if (fvField(fvRight) == fvField(fvLeft))
+                {
+                    return fvLeft < fvRight;
```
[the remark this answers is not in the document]
rodnymolina: It doesn't really matter how many fields we have as long as the schema they utilize keeps its consistency. The purpose of this function is to enforce an order within the map, so technically there's no real need to compare all the elements within a tuple. For example, if you are comparing these two tuples corresponding to a particular prefix...

left ->  nh: 10.1.1.1, if: eth0
right-> nh: 10.1.1.2, if: eth1

... you can exit() the 'comparator' method as long as you have found a common-field within each tuple that can be used for comparison. Please let me know if you can think of a use-case that would require the complete iteration of the tuple.

As i said before, the above can only work if the schema is kept uniform, which is the case for all apps so far (with the exception of the 'blackhole' route).

And yes, i'm planning to add some comments to document what this function does.

jipanyang:
My general feeling about the implementation is that it might have been focusing too much on the specific syntax of multiple sub-value of  kfv: 1.1.1.1/30, vector{nexthop: 10.1.1.1, 10.1.1.2, ifname: eth1, eth2}, it added a lot of logic processing complexity.

Some wild thoughts, could we simplify the logic like this:

1.  Read old data from appDB table.

2. Start with accumulating KFV data from Applications
    Keep them separate from the old data, do touch the old data.
    If same key arrives for the new data, just overwrite the previous one. 

3.  Timer expires.
4  Do the comparison in one shot.
     yes,  we should compare field sub values,  but there is no need to mark state for each of them.  

What we care about is the whole entry.

rodnymolina:
@jipanyang thanks for your comments.

There are good reasons for this added complexity, which may also provide extra flexibility to other applications in the future:

- We need to deal with multi-field values (e.g. 1.1.1.1/30 nh: 10.1.1.1, 10.2.2.2, if: eth1, eth2)
- We need to deal with different orders in multi-field values (e.g. 1.1.1.1/30 nh: 10.2.2.2, 10.1.1.1, if: eth2, eth1)
- And we need to deal with partial updates coming from the north-app (e.g. first we get "1.1.1.1/30 nh: 10.1.1.1, if: eth1", and then we get the complete prefix: "1.1.1.1/30 nh: 10.1.1.1, 10.2.2.2, if: eth1, eth2")

I'm not sure how the solution that you proposed would simplify my code while addressing the above requirements. Let's say you keep new/refreshed state in a separated location from the one collected from AppDB. How would you compare these two without doing any splitting:

AppDB state:   1.1.1.1/30 nh: 10.1.1.1, 10.2.2.2, if: eth1, eth2)
Zebra state:     1.1.1.1/30 nh: 10.2.2.2, 10.1.1.1, if: eth2, eth1)

I do understand though, that what we care about is the whole entry. But given that we have to split the tuples anyway, why assigning a STATE to them represents an issue? That gives us the possibility of making use of a single data-structure to run the entire reconciliation logic.

jipanyang:
@rodnymolina  What is really needed to compare 
"
AppDB state: 1.1.1.1/30 nh: 10.1.1.1, 10.2.2.2, if: eth1, eth2)
Zebra state: 1.1.1.1/30 nh: 10.2.2.2, 10.1.1.1, if: eth2, eth1)
"
and Also handle the case  of different number of fields.

That is different from spitting the fields, assigning state to each of them, and making a lot of hard to understand processing.

If it makes it clear, I could create a temporary pr to your branch which will be based on your current framework and major logic flow but with reduced complexity to show how it works.

nikos-github:
@rodnymolina 

One way to compare

AppDB state: 1.1.1.1/30 nh: 10.1.1.1, 10.2.2.2, if: eth1, eth2)
and
Zebra state: 1.1.1.1/30 nh: 10.2.2.2, 10.1.1.1, if: eth2, eth1)

is to create a nexthopgroup object and you have a pointer from each prefix to a nexthopgroup. For 2 identical prefixes, one recovered from the appdb and the other created after recovery, all you need to do is compare if they point to the same object.

nikos-github:
@rodnymolina @jipanyang @lguohan 

**Design Comments**

I always viewed sonic's *syncd parts (fpmsyncd, neighoborsyncd etc.) as thin stateless layers that provide intermediate yet very simple functionality and means of communication, between major sonic parts. Their functionality only goes as far as knowing the format of the data they are manipulating but not necessarily the meaning of it. This view is also reinforced if one looks at the current code and understands the functionality.

With this and other PRs I see that we are diverging from that paradigm. At the same time I haven't seen compelling technical justification to do so or some existing technical debt that needs to be addressed which pushes us in that direction.

We seem to be converting *syncd parts into a stateful layer with additional needless complexity, code duplication and exploding memory requirements just to satisfy the needs of warm restart.

A divergence from the current design will allow us to get the functionality we want without the drawbacks exhibited here. I will point out the issues I see in the current design and then I will outline solutions to counter them in order of preference.

**Current Design**

- Keeping state per path is needless and results in convoluted comparison code. Routes always get implicitly updated rather than incremental per path updates. Eliminating this will probably result in a much simpler design with likely half the code changes.
- Path sharing is not happening. This means every prefix will point to its own set of paths duplicating information on a per prefix basis. For a 4-way ecmp prefix, the memory increase will be roughly 8x. 12K prefixes in fpmsyncd will feel like 100K when doing warm restart in terms of memory consumption. This is too much to ignore.
- We have now turned fpmsyncd into a stateful part of sonic. Adding state always increases interaction and complexity by definition.

**Alternative Designs**

1. Keep fpmsyncd stateless
This personally is my preferred approach and aligns best with the existing sonic design. During warm restart, fpmsyncd clears the appdb queue, starts a timer and informs orchagent (through appdb). Upon timer expiration it tells orchagent again. The orchagent will be the one responsible for marking prefixes as stale and cleaning up when the timer expires. Currently orchagent has code that suppresses routes where nothing has changed. So it has all the comparison logic in place to achieve that already. Quagga and FRR don't keep quiet upon restarts and churn the prefixes as they are learned. The appdb queue can suppress that if we want to or we can fix the issue in frr with minimum effort and quagga with moderate effort.
Pros:
- fpmsyncd remains stateless and with no memory increase.
- orchagent is the one doing the route reconciliation (which is where it should be IMO).
- With new path attributes, only the format knowledge part of fpmsyncd changes so that it understands what it's reading and how to push that info to the db. So fpmsyncd's warm restart part doesn't need to be retested.
Cons:
- Slight increase in complexity from the signalling of the start/stop timer through the appdb.

2. Stateful fpmsyncd and with path sharing.
Read routes from appdb and use the code from routeorch to reconstruct prefixes, ecmp groups etc. and share ecmp groups across prefixes. Read from fpm socket and construct a different set of prefixes but reuse the ecmp groups. On timer expiration, go through the prefixes and compare if they point to the same ecmp group and send only the delta down. Alternatively we can make this even simpler and just download the whole table paying attention only to the stale prefixes which need to be deleted. As I said above, routeorch already has code that suppresses prefixes where nothing has changed so sai/asic won't be notified needlessly.
Pros:
- Can minimize to only the delta on what is sent to the db.
- Share code with routeorch for reconstructing view. This is important, otherwise for every new path attribute, we need to change and test 2 different parts of sonic.
- Path sharing within this solution.
Cons:
- fpmsyncd is now stateful and a bit more complex with increased memory needs during a restart.

3. Stateful fpmsyncd and no path sharing.
Similar to the above but without any path sharing.
Pros:
- Can minimize to only the delta on what is sent to the db.
Cons:
- fpmsyncd is now stateful and more complex.
- Exploding memory consumption requirements.
- No code sharing with orchagent for reconstructing view.
- Need to change 2 different components and retest both when new path attributes are introduced (orchagent and fpmsyncd) and retest warm restart.

warmrestart/warmRestartHelper.h
```diff
@@ -0,0 +1,82 @@
...
+    bool runRestoration(void);
+
+    void insertRefreshMap(const KeyOpFieldsValuesTuple &kfv);
+
+    void reconcile(void);
+
+    bool compareAllFV(const std::vector<FieldValueTuple> &left,
+                      const std::vector<FieldValueTuple> &right);
+
+    bool compareOneFV(const std::string &v1, const std::string &v2);
```
jipanyang: Should they be private?

warmrestart/warmRestartHelper.cpp
```diff
@@ -0,0 +1,352 @@
...
+    for (auto &v2fv : v2)
+    {
+        auto v1Iter = v1Map.find(v2fv.first);
+        /*
+         * The sizes of both tuple-vectors should always match within any
+         * given application. In other words, all fields within v1 should be
+         * also present within v2. Otherwise we are running into some form of
+         * bug or an unsupported functionality.
+         */
+        assert(v1Iter != v1Map.end());
```
jipanyang: What if v1 is superset of v2?

jipanyang:
So the current reconciliation implementation has the assumption that application will always have fix number of FieldValueTuple.  Ex. for route BGP always provide two fields "nexthop" and "ifname",   for neighbor, they are "family" and "neigh".   

If there is any violation of this assumption,  the reconciliation won't work.   It is also related to the ProducerStateTable issue: https://github.com/Azure/sonic-swss-common/issues/236  

I think it is ok for now.  Further update could be done when more flexible cases emerge.  Could you put a little more information about this  in the code comment field?

fpmsyncd/fpmsyncd.cpp
```diff
@@ -18,25 +29,70 @@ int main(int argc, char **argv)
...
             cout << "Connected!" << endl;
 
             s.addSelectable(&fpm);
+
+            /* Initialize warm-restart logic if this one is enabled */
+            bool warmStartEnabled = sync.m_warmStartHelper.isEnabled();
+            if (warmStartEnabled)
+            {
+                /* Obtain warm-restart timer defined for routing application */
+                uint32_t warmRestartIval = sync.m_warmStartHelper.getRestartTimer();
```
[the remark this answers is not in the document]
nikos-github: Please address the comment.

fpmsyncd/fpmsyncd.cpp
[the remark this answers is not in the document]
nikos-github: Please amend the code to address initial comment.

warmrestart/warmRestartHelper.h
rodnymolina: sure, i can make them private.

warmrestart/warmRestartHelper.cpp
rodnymolina: If by 'superset' you're referring to v1 having more fields than v2, that shouldn't ever happen as you commented further below. I'll add more details to my code-comments about this expectation.

tests/test_warm_reboot.py
```diff
@@ -789,3 +789,670 @@ def test_swss_port_state_syncup(dvs):
...
+
+    elif action == 'CREATE':
+        (exitcode, num) = dvs.runcmd(['sh', '-c', "egrep \"{}\" /var/log/swss/swss.rec | wc -l".
+                                      format(table + ".*" + entry + ".*" + "CREATE\|" + attribs)])
+
+    elif action == 'DELETE':
+        (exitcode, num) = dvs.runcmd(['sh', '-c', "egrep \"{}\" /var/log/swss/swss.rec | wc -l".
+                                      format(table + ".*" + entry + ".*" + "DEL")])
+
+    print exitcode
```
rodnymolina: Yep, forgot to remove these two, will send another patch...

tests/test_warm_reboot.py
```diff
@@ -789,3 +795,674 @@ def test_swss_port_state_syncup(dvs):
...
+    dvs.runcmd(['sh', '-c', 'pkill -x fpmsyncd'])
+    time.sleep(1)
+
+
+def enable_warmrestart(dvs, db, app_name):
+    create_entry_tbl(
+        db,
+        swsscommon.CFG_WARM_RESTART_TABLE_NAME, app_name,
+        [
+            ("enable", "true"),
```
jipanyang: Could you use CLI to config it as other test cases?
config warm_restart enable/disable [APP]

tests/test_warm_reboot.py
```diff
@@ -789,3 +795,674 @@ def test_swss_port_state_syncup(dvs):
...
+        ]
+    )
+
+
+def set_restart_timer(dvs, db, app_name, value):
+    create_entry_tbl(
+        db,
+        swsscommon.CFG_WARM_RESTART_TABLE_NAME, app_name,
+        [
+            (app_name + "_timer", value),
```
jipanyang: "config warm_restart bgp_timer" should be prepared as that for arp:

```
root@sonic:/home/admin# config warm_restart --help
Usage: config warm_restart [OPTIONS] COMMAND [ARGS]...

  warm_restart-related configuration tasks

Options:
  -s, --redis-unix-socket-path TEXT
                                  unix socket path for redis connection
  --help                          Show this message and exit.

Commands:
  disable
  enable
  neighsyncd_timer
root@sonic:/home/admin# 
```

warmrestart/warmRestartHelper.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+using namespace swss;
+
+
+WarmStartHelper::WarmStartHelper(RedisPipeline      *pipeline,
+                                 ProducerStateTable *syncTable,
+                                 const std::string  &syncTableName,
+                                 const std::string  &dockerName,
+                                 const std::string  &appName) :
+    m_restorationTable(pipeline, syncTableName, false),
```
jipanyang: What is the reason for using external pipeline for m_restorationTable here? m_restorationTable is for get only.  Asking the caller to provide pipeline brings no benefit.

warmrestart/warmRestartHelper.cpp
```diff
@@ -0,0 +1,358 @@
...
+{
+    return m_state;
+}
+
+
+/*
+ * To be called by each application to obtain the active/inactive state of
+ * warm-restart functionality, and proceed to initialize the FSM accordingly.
+ */
+bool WarmStartHelper::isEnabled(void)
```
jipanyang: I could not think of a good name for this function.  isEnabled() makes people feel it is safe to call it anytime they like to check the state, similar to isReconciled(void) or inProgress(void) , which is actually not true.   How about bool WarmStartHelper::checkAndStart(void)?

fpmsyncd/fpmsyncd.cpp
```diff
@@ -18,25 +29,70 @@ int main(int argc, char **argv)
...
+                if (!warmRestartIval)
+                {
+                    warmStartTimer.setInterval(timespec{DEFAULT_ROUTING_RESTART_INTERVAL, 0});
+                }
+                else
+                {
+                    warmStartTimer.setInterval(timespec{warmRestartIval, 0});
+                }
+
```
heidinet2007: I have concern on assigning a static value to the warmStartTimer, because it's hard for an admin to identify a good number for such timer:
-  if it's too short, obviously some routes will be missing.
-  if it's too long, black hole may occur, as shown in the following example:  a new prefix is advertised to BGP peers,  peers have the route/path installed immediately and start to forward traffic to the warm rebooted box. However, the prefix is NOT yet installed in warm rebooted box's forwarding table due to the  running warmStartTimer (not expire yet), hence packet drop.  It looks worse, if the peers do have other paths to reach the prefix, which is true in most of DC setup with ECMP.  I think loop may also occur, as long as the rebooted box's control plane is inconsistent with its forwarding plane.  The longer the inconsistency, the higher chance the problems happen.
In order to reduce the chance, the forwarding path needs to coordinate with routing/zebra stack  and set timer value accordingly.

rodnymolina: @heidinet2007 absolutely, we covered these cases that you are describing in my design doc (find link below -- search for Known-Limitations section and also look at the CLI portion).

As mentioned in this doc, for the time being user will need to set both timers (bgp and warm-restart one) to similar/same values to avoid inconsistencies. And yes, i'm already working in an FRR extension that will allow bgp/zebra to notify fpmsyncd whenever all BGP EoR's are received. That should address all these issues.

https://github.com/Azure/SONiC/blob/dcac72377f23521a394694214678ea4450f6f70a/doc/routing-warm-reboot/Routing_Warm_Reboot.md)

warmrestart/warmRestartHelper.cpp
rodnymolina: Not sure i got this one. How can we create the table without its pipeline? You mean, using the other table constructor? -- Table(const DBConnector *db, const std::string &tableName). Even if that's what you imply, we would need to create a DBConnector if we don't want the caller to provide it, right?

Please clarify...

warmrestart/warmRestartHelper.cpp
rodnymolina: Yes, i see your point. CheckAndStart() sounds good to me, will change.

tests/test_warm_reboot.py
rodnymolina: To avoid deferring this PR merge, i'll make this small adjustment right after i write this addition to sonic-utils. Then i'll come back to this UT code and will adjust accordingly.

tests/test_warm_reboot.py
rodnymolina: Agree, that's the way to go.

fpmsyncd/fpmsyncd.cpp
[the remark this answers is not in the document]
rodnymolina: @nikos-github i missed this comment. Yes, i'm about to send one (hopefully final) update that explicitly clears the state of the data-structures used within WarmStartHelper class. That should take care of your previous concerns for the bgp/zebra restarting case.

fpmsyncd/fpmsyncd.cpp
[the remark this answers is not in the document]
rodnymolina: Yes, i'm taking care of it, see my other comment on this matter.

warmrestart/warmRestartHelper.cpp
jipanyang: Sorry, my previous comment is confusing.    WarmStartHelper() has both pipeline and syncTableName as parameters,  the only purpose of them is to prepare a Table handle in WarmStartHelper class.  Could a Table handle be prepared by the callers directly and remove pipeline/syncTableName, just like what has been done for "ProducerStateTable *syncTable"?

fpmsyncd/fpmsyncd.cpp
heidinet2007: It would be hard to deploy routing warm reboot without the auto sync between zebra and fpmsyncd (even though  the update-delay timer and warmstarttimer can be tuned),  because admins have alternative ways to achieve zero packet loss during routing stack upgrade.   When will the auto sync be done? Is it possible to include it in this release?

jipanyang: Second to this request.

Once we have current reconciliation processing merged, the channel for notification of EoR (all) received from BGP to fpmsyncd is necessary for a production quality BGP docker warm reboot feature.

rodnymolina: @heidinet2007 Can you please elaborate on this: "because admins have alternative ways to achieve zero packet loss during routing stack upgrade".

I do realize that there's room for suboptimal routing (or even drops) in the interval comprehended between the bgp-peers sending their EoRs, and the warmrestart-timer kicking off, but this is expected to be a very short interval if the timers are properly set.

I have already started to prototype the EoR-to-fpmsyncd solution but that won't be ready for this coming release (only two weeks).

nikos-github: EoR won't work even if routing application could send it since there is no notion of a table AF in sonic and all prefixes are in one bucket.

jipanyang: It is not EoR from BGP to fpmsyncd directly,  the path will be BGP-->Zebra-->fpmsyncd. 
With proper processing zebra should make sure the EoR(zebra) is sent to fpmsyncd after all neighbors have PEER_STATUS_EOR_RECEIVED and routes pushed to fpmsyncd.

rodnymolina: @nikos-github Zooming out a little bit, the fundamental problem that we are trying to fix here is how to close the 'gap' between sonic's control and fwd plane to expedite the reconciliation process. So in my mind, given that SONiC is AFI/SAFI agnostic, there's no real need to treat each AFI/SAFI independently; as long as there's one missing EoR for a given AFI/SAFI, we know that no reconciliation process should take place (obviously, within a certain time boundary).

Of course, this imply that we will be penalizing faster AFI/SAFIs, but the alternative to this approach would be to introduce AFI/SAFI semantics in SONiC, which as you know, is a longer-term project.

And yep, i'm aware of EoR being used outside GR context, it's just a requirement.

nikos-github: @jipanyang I believe you are confusing things. There is no notion of an implicit EoR through the expiration of update-delay. The update-delay timer (not processing) is a catch all mechanism for BGP that controls the maximum time BGP will wait after the first neighbor is established and until it starts calculating bestpaths and sending out updates. You can certainly implement an EoR message based on stopping or upon expiration of the update-delay timer and while it is simple, it would also be wrong since you are not guaranteed that bestpath selection is finished, RIB installation is done, etc. In other words sending an EoR to RIB when the update-delay timer expires or is stopped, can most certainly be premature.

@rodnymolina What you are describing can potentially lead you to wait until the update-delay timer expires. Per my comment above, this has a good chance it will be premature in those cases.

jipanyang:
@rodnymolina While we should continue the discussion,  the implementation in this PR needs to be closed as soon as possible. It is the base for further optimization.

@nikos-github We have some work to be done to make use of the explicit EoR/Implicit EoR. I'm not saying that update-delay should be used, but the implementation of it could be a reference point. There is more processing needed at zebra.  Having more to be done should not be the reason stating it is wrong. I prefer seeing actual effort trying the change and refine it step by step.

```
All the configured peers, except the shutdown peers, have sent explicit EOR
      (End-Of-RIB) or an implicit-EOR. The first keep-alive after BGP has reached
      Established is considered an implicit-EOR.
```

nikos-github:
@jipanyang "We have some work to be done to make use of the explicit EoR/Implicit EoR. I'm not saying that update-delay should be used, but the implementation of it could be a reference point." i

It can not - that's precisely the point I was making. You can not infer anything from the explicit EoR/implicit EoR that the nbrs sent as to when the EoR to zebra should be sent. All that tell you is that you can not send earlier than them but that is useless anyway since we don't want that.

"I prefer seeing actual effort trying the change and refine it step by step."

First you will need to write this up so that it's clear what you proposing. There is no processing needed at zebra other than to make sure the EoR goes through. So for zebra this will be a passthrough because it doesn't hold on to the BGP routes.

We can discuss more offline.

rodnymolina:
@nikos-github @jipanyang i'll have a design-doc ready next week. Let's discuss this EoR feature outside this PR.

fpmsyncd/fpmsyncd.cpp
heidinet2007: It might be confusing to use EoR here, as it's more like a BGP term. What we need here is a RIB/ZEBRA CONVERGED signal to fpm. RIB convergence time has dependancy on routing protocols (here we consider only BGP), current BGP implementation already has a scheme to tell RIB when it receives EoR from all its neighbours. We should be able to utilize it.

rodnymolina: Agree, I vote for using something more generic, like FRR's existent EOIU terminology, which can be potentially extended to other protocols in the future too.

lguohan:
@rodnymolina , can you rebase your pr against the latest master and fix the vstest issue?

rodnymolina:
@lguohan, i don't see that VS failure in my setup, which btw is unrelated to my feature. I just rebased and re-tested everything locally, and again all testcases passed. Will send another patch to force our VS tests to run again.

tests/test_warm_reboot.py
```diff
@@ -780,3 +789,665 @@ def test_swss_port_state_syncup(dvs, testlog):
...
+    )
+
+
+################################################################################
+#
+# Routing warm-restart testcases
+#
+################################################################################
+
+def test_routing_WarmRestart(dvs):
```
lguohan: >tart(dvs): [](start = 24, length = 10)

add (testlog) param, it will print out separator for this test in the syslog.

tests/test_warm_reboot.py
```diff
@@ -780,3 +789,665 @@ def test_swss_port_state_syncup(dvs, testlog):
...
+def start_fpmsyncd(dvs):
+    dvs.runcmd(['sh', '-c', 'supervisorctl start fpmsyncd'])
+
+    # Let's give zebra a chance to connect to Zebra.
+    time.sleep(5)
+
+
+def stop_fpmsyncd(dvs):
+    dvs.runcmd(['sh', '-c', 'pkill -x fpmsyncd'])
+    time.sleep(1)
```
lguohan: these start/stop process should go to conftesst.py for e-use purpose.

tests/test_warm_reboot.py
```diff
@@ -780,3 +789,665 @@ def test_swss_port_state_syncup(dvs, testlog):
...
+        db,
+        swsscommon.CFG_WARM_RESTART_TABLE_NAME, app_name,
+        [
+            (app_name + "_timer", value),
+        ]
+    )
+
+
+# function to check neighbor entry reconciliation status written in syslog
+def check_syslog_change(dvs, entry, action):#new_cnt, delete_cnt, iptype):
```
lguohan: >check_syslog_change [](start = 4, length = 19)

not used, please remove.

lguohan:
@rodnymolina , do not worry about the test_interfacesetmtu, in this case, teamsyncd crashed due to a race condition, we are investigating it.

tests/test_warm_reboot.py
```diff
@@ -780,3 +789,665 @@ def test_swss_port_state_syncup(dvs, testlog):
...
+
+
+#############################################################################
+#                                                                           #
+#                        Routing Warm-Reboot Testing                        #
+#                                                                           #
+#############################################################################
+
+
+def clear_logs(dvs):
```
lguohan: >clear_logs [](start = 4, length = 10)

we must not use the this clear log, use marker, this remove all previous logs, difficult to debug for previous test.

tests/test_warm_reboot.py
rodnymolina: done

tests/test_warm_reboot.py
rodnymolina: done

tests/test_warm_reboot.py
rodnymolina: done

tests/test_warm_reboot.py
rodnymolina: done

tests/test_warm_reboot.py
```diff
@@ -780,3 +789,687 @@ def test_swss_port_state_syncup(dvs, testlog):
...
+    #############################################################################
+
+
+    # Defining create neighbor entries (4 ipv4 and 4 ip6, two each on each interface) in linux kernel
+    intfs = ["Ethernet0", "Ethernet4", "Ethernet8"]
+
+    # Enable ipv6 on docker
+    dvs.runcmd("sysctl net.ipv6.conf.all.disable_ipv6=0")
+
+    dvs.runcmd("ifconfig {} 111.0.0.1/24 up".format(intfs[0]))
```
nikos-github: Do not use ifconfig. It's been deprecated and it's the source of many problems. Unit tests shouldn't break when sonic decides to remove net-tools.

tests/test_warm_reboot.py
```diff
@@ -780,3 +789,687 @@ def test_swss_port_state_syncup(dvs, testlog):
...
+    #############################################################################
+
+    # Restart zebra
+    dvs.stop_zebra()
+    marker1 = dvs.add_log_marker("/var/log/swss/swss.rec")
+    marker2 = dvs.add_log_marker("/var/log/swss/sairedis.rec")
+    dvs.start_zebra()
+
+    time.sleep(5)
+
```
nikos-github: You need to make sure the state is reconciled per your diffs.

warmrestart/warmRestartHelper.cpp
```diff
@@ -0,0 +1,362 @@
...
+     */
+    for (auto &kfv : m_refreshMap)
+    {
+        auto refreshedKey = kfvKey(kfv.second);
+        auto refreshedFV  = kfvFieldsValues(kfv.second);
+
+        SWSS_LOG_NOTICE("Warm-Restart reconciliation: introducing new entry %s",
+                        printKFV(refreshedKey, refreshedFV).c_str());
+
+        m_syncTable->set(refreshedKey, refreshedFV);
```
nikos-github: Seems to me that if you received a delete for a prefix not in the db, you will be creating a stale prefix.

nikos-github:
Unit tests are missing for some very important cases:
- Withdrawals generated for prefixes not in the db.
- Withdrawals and then changes for prefixes not in the db.

tests/test_warm_reboot.py
rodnymolina: See that in this initial testcase, warm-reboot is not enabled yet, so there's no warm-reboot FSM state per-se, as those are only applicable to apps where warm-reboot has been activated.

warmrestart/warmRestartHelper.cpp
rodnymolina: @nikos-github This route 'delete', which could only arrive to fpmsyncd if sent right after an 'add', would be discarded right away at orchagent level as it's missing nh attributes, so it wouldn't ever make it to asicdb, but i agree that we should stop it at fpmsyncd level. Will make a small correction in this logic. Thanks for pointing this out.

jipanyang: If the request is a delete operation,  `kfvOp(iter->second) == DEL_COMMAND`,  it won't reach here.

rodnymolina: @jipanyang , see that Niko's point is for the scenario in which the deleted-route is not present in AppDB, so we won't hit this code that you're highlighting.

nikos-github: @rodnymolina @jipanyang Please read carefully the case I have described. If the prefix is not in the db (and hence not restored), a route delete for which the route add was missed since the connection was severed, will result in a stale prefix being added and never removed. The logic and code are wrong at this point.

tests/test_warm_reboot.py
nikos-github: You shouldn't introduce and knowingly use functionality that has been deprecated. That's the whole point of the review. Besides, you are using ip for IPv6 so I'd rather you don't use ifconfig and you use ip  for IPv4 to match what your tests are doing for IPv6.

warmrestart/warmRestartHelper.cpp
jipanyang: I see.   Actually it is not an issue with current sonic.  consumerStateTable will automatically transform key without field/value into a del operation, no stale prefix will be added. 

But it is better to do the check in WarmStartHelper library instead of relying on consumerStateTable.

tests/test_warm_reboot.py
nikos-github: Shouldn't you then clear any leftover state in checkAndStart if warm-restart is not activated and check against that?

warmrestart/warmRestartHelper.cpp
rodnymolina: done

tests/test_warm_reboot.py
rodnymolina: done

tests/test_warm_reboot.py
rodnymolina: done

warmrestart/warmRestartHelper.cpp
```diff
@@ -0,0 +1,378 @@
...
+        auto refreshedOp  = kfvOp(kfv.second);
+        auto refreshedFV  = kfvFieldsValues(kfv.second);
+
+        /*
+         * During warm-reboot, apps could receive an 'add' and a 'delete' for an
+         * entry that does not exist in AppDB. In these cases we must prevent the
+         * 'delete' from being pushed down to AppDB, so we are handling this case
+         * differently than the 'add' one.
+         */
+        if(refreshedOp == DEL_COMMAND)
```
nikos-github: Minor thing - please add a space after if and before the (.

tests/conftest.py
```diff
@@ -322,9 +342,14 @@ def get_logs(self, modname=None):
             raise RuntimeError("Failed to unpack the archive.")
         os.system("chmod a+r -R log")
 
-    def add_log_marker(self):
+    def add_log_marker(self, file=None):
         marker = "=== start marker {} ===".format(datetime.now().isoformat())
-        self.ctn.exec_run("logger {}".format(marker))
+
+        if file:
+            self.runcmd(['sh', '-c', "echo \"{}\" >> {}".format(marker, file)])
```
lguohan: there is no guarantee that this will not cut in the middle of a line. I prefer not to add log marker for swss or sairedis file. Instead, we should use SubscribeAsicDbObject to detect changes on asic db or app db instead of relying on the log file.

lguohan:
please refer to https://github.com/Azure/sonic-swss/pull/664/files

rodnymolina:
@lguohan I just finished refactoring my UTs as per your suggestion. Please take a look at this one when have a chance.

lguohan:
retest this please

PR #609, opened 2018-09-06
 Store mirror session state in StateDB

qiluo-msft:
The old implementation store only session status in a plain Table in ApplDB, which is not correct databese.
1. Store mirror session state in StateDB
2. Store many session fields besides status

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -11,18 +11,20 @@
 #include "converter.h"
 #include "mirrororch.h"
 
-#define MIRROR_SESSION_STATUS           "status"
-#define MIRROR_SESSION_STATUS_ACTIVE    "active"
-#define MIRROR_SESSION_STATUS_INACTIVE  "inactive"
-#define MIRROR_SESSION_SRC_IP           "src_ip"
-#define MIRROR_SESSION_DST_IP           "dst_ip"
-#define MIRROR_SESSION_GRE_TYPE         "gre_type"
-#define MIRROR_SESSION_DSCP             "dscp"
-#define MIRROR_SESSION_TTL              "ttl"
-#define MIRROR_SESSION_QUEUE            "queue"
-#define MIRROR_SESSION_DST_MAC_ADDRESS  "dst_mac"
-#define MIRROR_SESSION_MONITOR_PORT     "monitor_port"
-#define MIRROR_SESSION_ROUTE_PREFIX     "route_prefix"
+#define MIRROR_SESSION_STATUS               "status"
+#define MIRROR_SESSION_STATUS_ACTIVE        "active"
+#define MIRROR_SESSION_STATUS_INACTIVE      "inactive"
+#define MIRROR_SESSION_SRC_IP               "src_ip"
+#define MIRROR_SESSION_DST_IP               "dst_ip"
+#define MIRROR_SESSION_GRE_TYPE             "gre_type"
+#define MIRROR_SESSION_DSCP                 "dscp"
+#define MIRROR_SESSION_TTL                  "ttl"
+#define MIRROR_SESSION_QUEUE                "queue"
+#define MIRROR_SESSION_DST_MAC_ADDRESS      "dst_mac"
+#define MIRROR_SESSION_MONITOR_PORT         "monitor_port"
+#define MIRROR_SESSION_ROUTE_PREFIX         "route_prefix"
+#define MIRROR_SESSION_VLAN_HEADER_VALID    "vlan_header_valid"
+
 
 #define MIRROR_SESSION_DEFAULT_VLAN_PRI 0
 #define MIRROR_SESSION_DEFAULT_VLAN_CFI 0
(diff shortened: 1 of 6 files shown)


orchagent/mirrororch.cpp
```diff
@@ -294,22 +300,55 @@ void MirrorOrch::deleteEntry(const string& name)
     SWSS_LOG_NOTICE("Removed mirror session %s", name.c_str());
 }
 
-bool MirrorOrch::setSessionState(const string& name, MirrorEntry& session)
+void MirrorOrch::setSessionStatus(const string& name, MirrorEntry& session)
```
stcheng: I think this is confusing to have both `setSessionStatus` and `setSessionState` functions. I would keep only the `setSessionStatus` function since status would refer to the 'final state' of the mirror session.

orchagent/mirrororch.cpp
```diff
@@ -565,11 +600,7 @@ bool MirrorOrch::deactivateSession(const string& name, MirrorEntry& session)
     }
 
     session.status = false;
-
-    if (!setSessionState(name, session))
-    {
-        throw runtime_error("Failed to test session state");
-    }
+    setSessionStatus(name, session);
```
stcheng: For the deactivation part, it is still okay to set all the related information other than the activate or not information.

orchagent/mirrororch.cpp
```diff
@@ -294,22 +300,55 @@ void MirrorOrch::deleteEntry(const string& name)
...
+}
+
+void MirrorOrch::setSessionState(const string& name, const MirrorEntry& session, const string& attr)
 {
     SWSS_LOG_ENTER();
 
     SWSS_LOG_INFO("Setting mirroring sessions %s state\n", name.c_str());
 
     vector<FieldValueTuple> fvVector;
+    string value;
```
stcheng: In this function, I think you could just check all the related information in the object `session` and save all related information that you need. I can't find a reason that you only need to store partial information based on the attribute from the input argument.

orchagent/mirrororch.h
```diff
@@ -93,7 +93,13 @@ class MirrorOrch : public Orch, public Observer, public Subject
     bool updateSessionDstMac(const string&, MirrorEntry&);
     bool updateSessionDstPort(const string&, MirrorEntry&);
     bool updateSessionType(const string&, MirrorEntry&);
-    bool setSessionState(const string&, MirrorEntry&);
+    void setSessionStatus(const string&, MirrorEntry&);
```
stcheng: I think only the `setSessionStatus` function is needed while it will store all the related fields.

tests/test_mirror.py
```diff
@@ -85,17 +85,16 @@ def remove_mirror_session(self, name):
...
-        assert len(fvs) == 1
+        assert len(fvs) > 0
         for fv in fvs:
             if fv[0] == "status":
                 status = fv[1]
-            else:
-                assert False
-        return status
+                return status
+        else:
```
stcheng: why this `else` is kept?

orchagent/mirrororch.cpp
```diff
@@ -4,21 +4,27 @@
...
+#define MIRROR_SESSION_SRC_IP               "src_ip"
+#define MIRROR_SESSION_DST_IP               "dst_ip"
+#define MIRROR_SESSION_GRE_TYPE             "gre_type"
+#define MIRROR_SESSION_DSCP                 "dscp"
+#define MIRROR_SESSION_TTL                  "ttl"
+#define MIRROR_SESSION_QUEUE                "queue"
+#define MIRROR_SESSION_DST_MAC_ADDRESS      "dst_mac"
+#define MIRROR_SESSION_MONITOR_PORT         "monitor_port"
+#define MIRROR_SESSION_ROUTE_PREFIX         "route_prefix"
+#define MIRROR_SESSION_VLAN_HEADER_VALID    "vlan_header_valid"
```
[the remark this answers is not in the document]
qiluo-msft: I don't check detailed inter dependencies, and my criterial is recording every field going through SAI API. Let me know if you prefer removing any.

---
In reply to: [215802906](https://github.com/Azure/sonic-swss/pull/609#discussion_r215802906) [](ancestors = 215802906)

orchagent/mirrororch.cpp
qiluo-msft: 'Status' is actually from 'session.status'. State is the all the states including Status.
Let me know if it makes any sense.

---
In reply to: [215803365](https://github.com/Azure/sonic-swss/pull/609#discussion_r215803365) [](ancestors = 215803365)

tests/test_mirror.py
qiluo-msft: This is part of for-else block. [http://book.pythontips.com/en/latest/for_-_else.html](http://book.pythontips.com/en/latest/for_-_else.html)
If there is no status field, then it is a logic error.

---
In reply to: [215804273](https://github.com/Azure/sonic-swss/pull/609#discussion_r215804273) [](ancestors = 215804273)

jleveque: The for/else isn't really necessary, since there is no way to break out of the loop. If the loop succeeds, the function will return, thus simply having the `assert False` line after the loop body would produce the same behavior.

orchagent/mirrororch.cpp
qiluo-msft: Offline discussed. It's for efficiency purpose. Any data inconsistency should be treat as a future bug.

---
In reply to: [215803903](https://github.com/Azure/sonic-swss/pull/609#discussion_r215803903) [](ancestors = 215803903)

tests/test_mirror.py
qiluo-msft: You're right. Fixed.

---
In reply to: [215820363](https://github.com/Azure/sonic-swss/pull/609#discussion_r215820363) [](ancestors = 215820363)

stcheng:
My only concern is that the current test cases do not cover all the changes of this pull request, e.g. the rest of the fields' value in the state database. Would be glad if they are added in the near future.

PR #613, opened 2018-09-11
[vxlan tunnel]: Don't create encap mapper. Currently not supported

pavel-shirshov:
**What I did**
Removed vxlan tunnel encap mapper creation.

**Why I did it**
Currently SAI don't support this functionality. Also the implementation we had is incorrect. We used the same mapper for both directions, but we need to have two mirror/complementary mappers.

**How I verified it**
Build swss and run on your DUT

**Details if related**


lguohan:
check vs test failure?

test_vxlan_tunnel.py::test_vxlan_term_orch FAILED                        [ 95%]

pavel-shirshov:
>           check vs test failure?
> test_vxlan_tunnel.py::test_vxlan_term_orch FAILED                        [ 95%]

Yes. I'm rebuilding vs version of the image now to check what's going on there

PR #618, opened 2018-09-14
[aclorch] Remove error message when match not found

stepanblyschak:
Fail to find a match criteria is a completely valid flow (orchagent/aclorch.cpp:L2076-L2093)
This caused ACL test failure in loganalysis phase.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -164,8 +164,6 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
     {
         if (aclMatchLookup.find(attr_name) == aclMatchLookup.end())
         {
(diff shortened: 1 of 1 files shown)


stcheng:
Could you list the attribute that falls into this condition? If the flow is valid, is `return false` still expected?

jleveque:
Instead of removing the message, maybe we should just downgrade the log level? It might be useful for catching an attribute that falls into this case in the event it shouldn't have. How noisy is this message?

lguohan:
retest this please

lguohan:
agree with @stepanblyschak , any objections @stcheng , @jleveque ?

stcheng:
Approved. Thanks!

PR #621, opened 2018-09-16
[cfgmgr]: Add vrfmgrd

marian-pritsak:
vrfmgrd is responsible for VRF configuration in Linux
It creates VRF-Lite device for every VRF entry in Config DB

### DIFF
--- a/cfgmgr/Makefile.am
+++ b/cfgmgr/Makefile.am
@@ -1,7 +1,7 @@
 INCLUDES = -I $(top_srcdir) -I $(top_srcdir)/orchagent -I $(top_srcdir)/warmrestart
 CFLAGS_SAI = -I /usr/include/sai
 
-bin_PROGRAMS = vlanmgrd portmgrd intfmgrd buffermgrd
+bin_PROGRAMS = vlanmgrd portmgrd intfmgrd buffermgrd vrfmgrd
 
 if DEBUG
 DBGFLAGS = -ggdb -DDEBUG
@@ -30,3 +30,7 @@ buffermgrd_CFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_SAI)
 buffermgrd_CPPFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_SAI)
 buffermgrd_LDADD = -lswsscommon
 
+vrfmgrd_SOURCES = vrfmgrd.cpp vrfmgr.cpp $(top_srcdir)/orchagent/orch.cpp $(top_srcdir)/orchagent/request_parser.cpp shellcmd.h
+vrfmgrd_CFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_SAI)
+vrfmgrd_CPPFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_SAI)
+vrfmgrd_LDADD = -lswsscommon
--- a/cfgmgr/vrfmgr.cpp
+++ b/cfgmgr/vrfmgr.cpp
@@ -1,0 +1,146 @@
+#include <string.h>
+#include "logger.h"
+#include "dbconnector.h"
+#include "producerstatetable.h"
+#include "tokenize.h"
+#include "ipprefix.h"
+#include "vrfmgr.h"
+#include "exec.h"
+#include "shellcmd.h"
+
+#define VRF_TABLE_START 100
+#define VRF_TABLE_END 164
+
+using namespace swss;
+
+VrfMgr::VrfMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, const vector<string> &tableNames) :
+        Orch(cfgDb, tableNames)
+{
+    for (uint32_t i = VRF_TABLE_START; i < VRF_TABLE_END; i++)
+    {
+        m_freeTables.emplace(i);
+    }
+}
+
+uint32_t VrfMgr::getFreeTable(void)
+{
+    SWSS_LOG_ENTER();
+
+    if (m_freeTables.empty())
+    {
+        return 0;
+    }
+
+    uint32_t table = *m_freeTables.begin();
+    m_freeTables.erase(table);
+
+    return table;
+}
+
+void VrfMgr::returnTable(uint32_t table)
+{
+    SWSS_LOG_ENTER();
+
+    m_freeTables.emplace(table);
+}
+
+bool VrfMgr::delLink(const string& vrfName)
+{
+    SWSS_LOG_ENTER();
+
+    stringstream cmd;
+    string res;
+
+    if (m_vrfTableMap.find(vrfName) == m_vrfTableMap.end())
+    {
+        return false;
+    }
+
+    returnTable(m_vrfTableMap[vrfName]);
+
+    cmd << IP_CMD << " link del " << vrfName;
+    int ret = swss::exec(cmd.str(), res);
+    if (ret)
+    {
+        SWSS_LOG_ERROR("Failed command %s, rc:%d", cmd.str().c_str(), ret);
+        return false;
+    }
+
+    m_vrfTableMap.erase(vrfName);
+
+    return true;
+}
+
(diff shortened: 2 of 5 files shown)


marian-pritsak:
Retest this please

cfgmgr/vrfmgr.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include "dbconnector.h"
+#include "producerstatetable.h"
+#include "tokenize.h"
+#include "ipprefix.h"
+#include "vrfmgr.h"
+#include "exec.h"
+#include "shellcmd.h"
+
+#define VRF_TABLE_START 100
+#define VRF_TABLE_END 164
```
lguohan: what are these two numbers? does this limit to 64 vrfs?

cfgmgr/vrfmgr.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    stringstream cmd;
+    string res;
+
+    uint32_t table = getFreeTable();
+    if (table == 0)
+    {
+        return false;
+    }
+
```
lguohan: can we check the vrfTableMap, if it is there means the vrf has been created, then we can just return true. Instead of creating a new one.

cfgmgr/vrfmgr.h
```diff
@@ -0,0 +1,34 @@
...
+class VrfMgr : public Orch
+{
+public:
+    VrfMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, const vector<string> &tableNames);
+    using Orch::doTask;
+
+private:
+    bool delLink(const string& vrfName);
+    bool setLink(const string& vrfName);
+    void returnTable(uint32_t table);
```
lguohan: can you rename to recycleTableId?

cfgmgr/vrfmgr.cpp
```diff
@@ -0,0 +1,146 @@
...
+    stringstream cmd;
+    string res;
+
+    if (m_vrfTableMap.find(vrfName) == m_vrfTableMap.end())
+    {
+        return false;
+    }
+
+    returnTable(m_vrfTableMap[vrfName]);
```
lguohan: i feel we should move this after line 69 since we should only recycle the id when it is free.

cfgmgr/vrfmgr.cpp
```diff
@@ -0,0 +1,146 @@
...
+    string res;
+
+    uint32_t table = getFreeTable();
+    if (table == 0)
+    {
+        return false;
+    }
+
+    cmd << IP_CMD << " link add " << vrfName << " type vrf table " << table;
+    int ret = swss::exec(cmd.str(), res);
```
jipanyang: I'm think about warm restart support for vrfmgrd.   If the vrf table has been configured on linux,  adding it again would return error.  Could it be changed to warm restart agnostic, like checking the existence of vrf table before adding it?
```
root@sonic:/home/admin# ip link add vrftest1 type vrf table 120
root@sonic:/home/admin# ip link add vrftest1 type vrf table 120
RTNETLINK answers: File exists
root@sonic:/home/admin# echo $?
2
```

nikos-github:
@marian-pritsak @lguohan Is there a plan on how sonic will handles IP addresses for interfaces enslaved to vrfs? Will they be handled the same way kernel does and also be checking the sysctl variables?

cfgmgr/vrfmgr.cpp
marian-pritsak: Linux routing table numbers. Do we need more?

cfgmgr/vrfmgr.cpp
marian-pritsak: Done

cfgmgr/vrfmgr.cpp
marian-pritsak: Not sure if I follow. Warm restart reboots kernel. VRF netdevs will be gone after reboot.

cfgmgr/vrfmgr.h
marian-pritsak: Done

cfgmgr/vrfmgr.cpp
marian-pritsak: Done

cfgmgr/vrfmgr.cpp
jipanyang: There is SWSS docker warm restart. In that case, Linux env is kept.

cfgmgr/vrfmgr.cpp
lguohan: maybe can change to 1k initially.

cfgmgr/vrfmgr.cpp
lguohan: there is no problem to get all vrfs using following command. 
```
admin@str-s6000-on-4:~$ ip -br link show type vrf
vrftest1         DOWN           c2:31:e9:01:c8:f4 <NOARP,MASTER> 
```

the problem is to get the route table id.

lguohan: do we need the table id? I can see the ip route add can use vrfname directly. @marian-pritsak , do you know who is going to use that table id?

lguohan: from what I can see, it seems we can get form netlink messages.

https://github.com/FRRouting/frr/blob/master/zebra/if_netlink.c#L305

I do not know if there is a command line that can get this.

marian-pritsak: Done

marian-pritsak: ip -d shows table ID

cfgmgr/vrfmgr.cpp
marian-pritsak: Done
Used 1001..2000 since there are a few reserved values in the first 256 tables.

cfgmgr/vrfmgr.cpp
lguohan: thanks

lguohan:
@prsunny  and @jipanyang , can you check the pr, any more comments?

cfgmgr/vrfmgr.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+    /* Get existing VRFs from Linux */
+    stringstream cmd;
+    string res;
+
+    cmd << IP_CMD << " -d link show type vrf";
+    int ret = swss::exec(cmd.str(), res);
+    if (ret)
+    {
+        SWSS_LOG_ERROR("Command %s failed, rc: %i", res.c_str(), ret);
```
jipanyang: Probably EXEC_WITH_ERROR_THROW(cmd, res) is better here.

cfgmgr/Makefile.am
```diff
@@ -30,3 +30,7 @@ buffermgrd_CFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_SAI)
 buffermgrd_CPPFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_SAI)
 buffermgrd_LDADD = -lswsscommon
 
+vrfmgrd_SOURCES = vrfmgrd.cpp vrfmgr.cpp $(top_srcdir)/orchagent/orch.cpp $(top_srcdir)/orchagent/request_parser.cpp shellcmd.h
```
jipanyang: Since every daemon here is taking  "$(top_srcdir)/orchagent/orch.cpp $(top_srcdir)/orchagent/request_parser.cpp shellcmd.h",  would it make sense to have a macro for them like

```
CFGMGR_COMMON_SOURCES = $(top_srcdir)/orchagent/orch.cpp $(top_srcdir)/orchagent/request_parser.cpp shellcmd.h
```

cfgmgr/vrfmgr.cpp
marian-pritsak: Done

cfgmgr/Makefile.am
marian-pritsak: Maybe, but I don't like mixing that kind of refactoring with new features in a single PR

lguohan:
retest this please

prsunny:
Looks good. Based on the Vxlan discussion today, we've to update the STATE_DB/APP_DB after the VRF is created in kernel. But this can be done in the next commit.

PR #629, opened 2018-09-26
[watermarkorch] add watermarkorch, extend queue and pg counters with wat…

mykolaf:
…ermark stats

**What I did**
 - Add new orch - watermarkorch
 - add new flex counter groups - QUEUE_WATERMARK, PG_WATERMARK
 - add new queue & pg flex plugins - watermark_queue.lua, watermark_pg.lua
**Why I did it**
This is new code forthe watermarks feature
[High Level Design Document](https://github.com/Azure/SONiC/blob/694754ac31c77724f43458e2779b552223791974/doc/watermarks_HLD.md)
**How I verified it**
Manually verified on mlnx switch.
**Details if related**

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -477,6 +477,7 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port,
         // Only register lossless tc counters in database.
         string str = counterIdsToStr(c_portStatIds, &sai_serialize_port_stat);
         string filteredStr = filterPfcCounters(str, losslessTc);
+        fieldValues.emplace_back(STATS_MODE_FIELD, STATS_MODE_READ);
         fieldValues.emplace_back(PORT_COUNTER_ID_LIST, filteredStr);
 
         m_flexCounterTable->set(key, fieldValues);
@@ -505,6 +506,7 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port,
         if (!c_queueStatIds.empty())
         {
             string str = counterIdsToStr(c_queueStatIds, sai_serialize_queue_stat);
+            queueFieldValues.emplace_back(STATS_MODE_FIELD, STATS_MODE_READ);
             queueFieldValues.emplace_back(QUEUE_COUNTER_ID_LIST, str);
         }
 
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -477,7 +477,6 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port,
         // Only register lossless tc counters in database.
         string str = counterIdsToStr(c_portStatIds, &sai_serialize_port_stat);
         string filteredStr = filterPfcCounters(str, losslessTc);
-        fieldValues.emplace_back(STATS_MODE_FIELD, STATS_MODE_READ);
         fieldValues.emplace_back(PORT_COUNTER_ID_LIST, filteredStr);
 
         m_flexCounterTable->set(key, fieldValues);
@@ -506,7 +505,6 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port,
         if (!c_queueStatIds.empty())
         {
             string str = counterIdsToStr(c_queueStatIds, sai_serialize_queue_stat);
-            queueFieldValues.emplace_back(STATS_MODE_FIELD, STATS_MODE_READ);
             queueFieldValues.emplace_back(QUEUE_COUNTER_ID_LIST, str);
         }
 
@@ -652,6 +650,7 @@ PfcWdSwOrch<DropHandler, ForwardHandler>::PfcWdSwOrch(
         vector<FieldValueTuple> fieldValues;
         fieldValues.emplace_back(QUEUE_PLUGIN_FIELD, detectSha + "," + restoreSha);
         fieldValues.emplace_back(POLL_INTERVAL_FIELD, to_string(m_pollInterval));
+        fieldValues.emplace_back(STATS_MODE_FIELD, STATS_MODE_READ);
         m_flexCounterGroupTable->set(PFC_WD_FLEX_COUNTER_GROUP, fieldValues);
     }
     catch (...)
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -16,6 +16,7 @@
 
 #include "logger.h"
 #include "schema.h"
+#include "redisapi.h"
 #include "converter.h"
 #include "sai_serialize.h"
 #include "crmorch.h"
@@ -40,6 +41,9 @@ extern BufferOrch *gBufferOrch;
 #define DEFAULT_VLAN_ID     1
 #define PORT_FLEX_STAT_COUNTER_POLL_MSECS "1000"
 #define QUEUE_FLEX_STAT_COUNTER_POLL_MSECS "10000"
+#define QUEUE_WATERMARK_FLEX_STAT_COUNTER_POLL_MSECS "1000"
+#define PG_WATERMARK_FLEX_STAT_COUNTER_POLL_MSECS "1000"
+
 
 static map<string, sai_port_fec_mode_t> fec_mode_map =
 {
@@ -105,6 +109,17 @@ static const vector<sai_queue_stat_t> queueStatIds =
     SAI_QUEUE_STAT_DROPPED_BYTES,
 };
 
+static const vector<sai_queue_stat_t> queueWatermarkStatIds =
+{
+    SAI_QUEUE_STAT_SHARED_WATERMARK_BYTES,
+};
+
+static const vector<sai_ingress_priority_group_stat_t> ingressPriorityGroupWatermarkStatIds =
+{
+    SAI_INGRESS_PRIORITY_GROUP_STAT_XOFF_ROOM_WATERMARK_BYTES,
+    SAI_INGRESS_PRIORITY_GROUP_STAT_SHARED_WATERMARK_BYTES,
+};
+
 static char* hostif_vlan_tag[] = {
     [SAI_HOSTIF_VLAN_TAG_STRIP]     = "SAI_HOSTIF_VLAN_TAG_STRIP",
     [SAI_HOSTIF_VLAN_TAG_KEEP]      = "SAI_HOSTIF_VLAN_TAG_KEEP",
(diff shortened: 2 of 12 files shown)


lguohan:
retest this please

mykolaf:
retest this please

mykolaf:
Depends on sonic-utilities https://github.com/Azure/sonic-utilities/pull/327

orchagent/portsorch.cpp
```diff
@@ -2847,6 +2925,67 @@ void PortsOrch::generateQueueMapPerPort(const Port& port)
...
+        string key = getPriorityGroupWatermarkFlexCounterTableKey(id);
+
+        std::string delimiter = "";
+        std::ostringstream counters_stream;
+        /* Add watermark counters to flex_counter */
+        for (const auto& it: ingressPriorityGroupWatermarkStatIds)
+        {
+            counters_stream << delimiter << sai_serialize_ingress_priority_group_stat(it);
+            delimiter = comma;
+        }
```
wendani: If we add the pg watermark stats here, they will be enabled if the corresponding "FLEX_COUNTER_TABLE|FLEX_COUNTER_STATUS" is "enabled" in the CONFIG_DB. We may take the similar approach as in the PFCWD that we listen to the SET_COMMAND and DEL_COMMAND to be able to enable and disable polling watermark stats at run time. Similar case for the queue watermark stats above.

orchagent/pfcwdorch.cpp
```diff
@@ -505,6 +506,7 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port,
         if (!c_queueStatIds.empty())
         {
             string str = counterIdsToStr(c_queueStatIds, sai_serialize_queue_stat);
+            queueFieldValues.emplace_back(STATS_MODE_FIELD, STATS_MODE_READ);
```
wendani: for PORT_STATS, QUEUE_STATS, PG_WATERMARK_STATS, and QUEUE_WATERMARK_STATS, STATS_MODE is set to the corresponding GROUP_TABLE. For PFCWD, it is set to FLEX_COUNTER_TABLE. Why the difference?

orchagent/portsorch.cpp
mykolaf: The syncd has the logic for listening to FC STATUS in place:
https://github.com/Azure/sonic-sairedis/blob/master/syncd/syncd.cpp#L2767
There is also the ability to enable/disable WM polling via counterpoll utility:
https://github.com/Azure/sonic-utilities/blob/614dbeec6a0bb99c0020e276f95cfe60bc6a1a60/counterpoll/main.py#L99

orchagent/watermarkorch.h
```diff
@@ -0,0 +1,57 @@
...
+{
+public:
+    WatermarkOrch(DBConnector *db, const std::string tableName);
+    virtual ~WatermarkOrch(void);
+
+    void doTask(Consumer &consumer);
+    void doTask(NotificationConsumer& consumer);
+    void doTask(SelectableTimer &timer);
+
+    void init_pg_ids();
```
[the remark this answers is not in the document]
mykolaf: Sure

orchagent/portsorch.cpp
```diff
@@ -141,17 +156,53 @@ PortsOrch::PortsOrch(DBConnector *db, vector<table_name_with_pri_t> &tableNames)
...
+    {
+        string queueLuaScript = swss::loadLuaScript(queueWmPluginName);
+        queueWmSha = swss::loadRedisScript(m_counter_db.get(), queueLuaScript);
+
+        string pgLuaScript = swss::loadLuaScript(pgWmPluginName);
+        pgWmSha = swss::loadRedisScript(m_counter_db.get(), pgLuaScript);
+
+        vector<FieldValueTuple> fieldValues;
+        fieldValues.emplace_back(QUEUE_PLUGIN_FIELD, queueWmSha);
+        fieldValues.emplace_back(POLL_INTERVAL_FIELD, QUEUE_WATERMARK_FLEX_STAT_COUNTER_POLL_MSECS);
```
wendani: I will look into the flex counter thread in syncd. Before I get more acquainted, a general question in my mind is that each flex counter table has its own polling interval. How does the flex counter thread in syncd make sure each one gets polled and serviced just on time the interval expires?

orchagent/portsorch.cpp
```diff
@@ -141,17 +156,53 @@ PortsOrch::PortsOrch(DBConnector *db, vector<table_name_with_pri_t> &tableNames)
...
     m_queueTypeTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_QUEUE_TYPE_MAP));
 
+    /* Initialize ingress priority group tables */
+    m_pgTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_PG_NAME_MAP));
+    m_pgPortTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_PG_PORT_MAP));
+    m_pgIndexTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_PG_INDEX_MAP));
+
     m_flex_db = shared_ptr<DBConnector>(new DBConnector(FLEX_COUNTER_DB, DBConnector::DEFAULT_UNIXSOCKET, 0));
     m_flexCounterTable = unique_ptr<ProducerTable>(new ProducerTable(m_flex_db.get(), FLEX_COUNTER_TABLE));
     m_flexCounterGroupTable = unique_ptr<ProducerTable>(new ProducerTable(m_flex_db.get(), FLEX_COUNTER_GROUP_TABLE));
```
wendani: A general question: why the flex counter table and the flex counter group table need to be ProducerTable? Qi mentioned that ProducerTable is a FIFO queue. So it is used when we care about the sequences the redis request is queued.

orchagent/portsorch.cpp
wendani: I checked the code in sycnd. flex counter thread is per flex counter table. So it makes sense to me now.

orchagent/portsorch.cpp
wendani: @mykolaf We should use ProducerTable for flex counter table and flex counter group table. This is what used in the pfcwd. Just curious about the reason behind it.

mykolaf: @wenda Flex counter uses ConsumerTable, so we use the related producer. There is no specific reason for using it, it was just implemented that way.

qiluo-msft:
retest this please

lguohan:
retest this please

mykolaf:
VS test dependencies:
1. "set counter" infra  (sonic-sairedis): 
https://github.com/Azure/sonic-sairedis/pull/366
2. watermark CLI 
https://github.com/Azure/sonic-utilities/pull/327
@wendani could you take a look? 

Probably sonic-utilities pointer also needs to be updated in order to support WM CLI in the docker-vs)

mykolaf:
retest this please

lguohan:
retest this please

mykolaf:
retest this please

lguohan:
retest this please

lguohan:
@mykolaf , can you sign the cla?

mykolaf:
@lguohan That was some bug, I already signed it 8 month ago.
Anyway, now it shows that it's signed.

orchagent/portsorch.cpp
```diff
@@ -2847,6 +2925,67 @@ void PortsOrch::generateQueueMapPerPort(const Port& port)
...
+
+void PortsOrch::generatePriorityGroupMapPerPort(const Port& port)
+{
+    /* Create the PG map in the Counter DB */
+    /* Add stat counters to flex_counter */
+    vector<FieldValueTuple> pgVector;
+    vector<FieldValueTuple> pgPortVector;
+    vector<FieldValueTuple> pgIndexVector;
+
+    for (size_t pgIndex = 0; pgIndex < port.m_priority_group_ids.size(); ++pgIndex)
```
wendani: Lossy PG does not have SAI_INGRESS_PRIORITY_GROUP_STAT_XOFF_ROOM_WATERMARK_BYTES. So the query value will always return 0?

mykolaf: I guess it depends on the SAI implementation of it. If the SAI returns some status other than success, the flex counter will mark the counter as unsupported and stop polling. For this cases we will show the value as 'N/A' in the output. 
But having made a little test, I see that I get a 0 for lossy pg. Is this different on some other platform?
From my point of view, it seems logical  that even lossy PG has a counter for headroom watermark, even if it has lossy profile applied. Looks like the structure of PG's counters doesn't care of pg being lossy/lossless.

lguohan:
retest this please

lguohan:
retest this please

lguohan:
retest this please

orchagent/watermarkorch.cpp
```diff
@@ -0,0 +1,224 @@
...
+    m_countersTable = make_shared<Table>(m_countersDb.get(), COUNTERS_TABLE);
+    m_periodicWatermarkTable = make_shared<Table>(m_countersDb.get(), PERIODIC_WATERMARKS_TABLE);
+    m_persistentWatermarkTable = make_shared<Table>(m_countersDb.get(), PERSISTENT_WATERMARKS_TABLE);
+    m_userWatermarkTable = make_shared<Table>(m_countersDb.get(), USER_WATERMARKS_TABLE);
+
+    m_clearNotificationConsumer = new swss::NotificationConsumer(
+            m_appDb.get(),
+            "WATERMARK_CLEAR_REQUEST");
+    auto clearNotifier = new Notifier(m_clearNotificationConsumer, this, "WM_CLEAR_NOTIFIER");
+    Orch::addExecutor(clearNotifier);
```
wendani: For user issued clear, why not just directly clear through the CLI but signal through an APPL_DB event channel, and delegate to WatermarkOrch to do so? What is the benefit here?

orchagent/watermark_pg.lua
```diff
@@ -0,0 +1,49 @@
...
+    local persistent_headroom_wm = redis.call('HGET', persistent_table_name .. ':' .. KEYS[i], 'SAI_INGRESS_PRIORITY_GROUP_STAT_XOFF_ROOM_WATERMARK_BYTES')
+    local user_headroom_wm = redis.call('HGET', user_table_name .. ':' .. KEYS[i], 'SAI_INGRESS_PRIORITY_GROUP_STAT_XOFF_ROOM_WATERMARK_BYTES')
+
+    -- Set the values into the other tables. Make comparioson, if th evalue was absent in COUNTERS, set to N/A
+    if (pg_shared_wm) then
+        redis.call('HSET', periodic_table_name .. ':' .. KEYS[i], 'SAI_INGRESS_PRIORITY_GROUP_STAT_SHARED_WATERMARK_BYTES',  periodic_shared_wm and math.max(tonumber(pg_shared_wm), tonumber(periodic_shared_wm)) or pg_shared_wm)
+        redis.call('HSET', persistent_table_name .. ':' .. KEYS[i], 'SAI_INGRESS_PRIORITY_GROUP_STAT_SHARED_WATERMARK_BYTES', persistent_shared_wm and math.max(tonumber(pg_shared_wm), tonumber(persistent_shared_wm)) or pg_shared_wm)
+        redis.call('HSET', user_table_name .. ':' .. KEYS[i], 'SAI_INGRESS_PRIORITY_GROUP_STAT_SHARED_WATERMARK_BYTES', user_shared_wm and math.max(tonumber(pg_shared_wm), tonumber(user_shared_wm)) or pg_shared_wm)
```
wendani: The use of ternary operator is fancy, but we do not need to issue a redis hset command if the new value is no greater than the historic high, which is an expense operation in the context of a syncd flex counter thread.

orchagent/watermarkorch.cpp
mykolaf: This is the design we agreed on.
https://github.com/Azure/SONiC/blob/gh-pages/doc/watermarks_HLD.md

orchagent/watermark_pg.lua
mykolaf: I agree

orchagent/watermarkorch.cpp
mykolaf: This was needed in the draft design of the feature, in current implementation it doesn't give any advantages.

PR #636, opened 2018-10-08
vs test: Disable warm_restart at the end of test case

qiluo-msft:


tests/test_fdb_warm.py
```diff
@@ -181,6 +181,6 @@ def test_fdb_notifications(dvs):
 
     finally:
         # enable warm restart
```
jipanyang: disable.

qiluo-msft: Thanks and fixed.

Issue #567, opened 2018-08-08
Warm reboot:  physical port count may not match configured port count


jipanyang:
On platforms other than BFN_PLATFORM and MLNX_PLATFORM, as of today,  remove port will be skpped and message logged at notice level: 

https://github.com/Azure/sonic-swss/blob/bfce363ef306f45b4dc59f407cb281b8735641c2/orchagent/portsorch.cpp#L1361

When trying warm restart :
the m_portCount got from ASIC won't match that in appDB and cause warm restart failure based on current logic.

https://github.com/Azure/sonic-swss/blob/bfce363ef306f45b4dc59f407cb281b8735641c2/orchagent/portsorch.cpp#L1209

Aug  8 23:15:04.807579 sonic ERR swss/orchagent: :- bake: Invalid port table: m_portCount

Record the issue here for future fix.

lguohan:
if remove_port is not supported in the first place, why m_portCount will deviate from appdb port count?

jipanyang:
This is the current implementation for portorch warm start.

PR #645, opened 2018-10-18
Fix portorch bake: get port count from db

qiluo-msft:
Fixed https://github.com/Azure/sonic-swss/issues/567

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1312,8 +1312,10 @@ bool PortsOrch::bake()
 
     // Check the APP_DB port table for warm reboot
     vector<FieldValueTuple> tuples;
-    bool foundPortConfigDone = m_portTable->get("PortConfigDone", tuples);
-    SWSS_LOG_NOTICE("foundPortConfigDone = %d", foundPortConfigDone);
+    string value;
+    bool foundPortConfigDone = m_portTable->hget("PortConfigDone", "count", value);
+    unsigned long portCount = stoul(value);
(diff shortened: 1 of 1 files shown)


orchagent/portsorch.cpp
```diff
@@ -1312,8 +1312,10 @@ bool PortsOrch::bake()
 
     // Check the APP_DB port table for warm reboot
     vector<FieldValueTuple> tuples;
-    bool foundPortConfigDone = m_portTable->get("PortConfigDone", tuples);
-    SWSS_LOG_NOTICE("foundPortConfigDone = %d", foundPortConfigDone);
+    string value;
+    bool foundPortConfigDone = m_portTable->hget("PortConfigDone", "count", value);
+    unsigned long portCount = stoul(value);
+    SWSS_LOG_NOTICE("foundPortConfigDone = %d, portCount = %lu", foundPortConfigDone, portCount);
```
lguohan: is foundPortConfigDone already true, what if it is false, in this case can we skip the check in 1334?

qiluo-msft: If foundPortConfigDone == false, then return at line 1331

---
In reply to: [226145134](https://github.com/Azure/sonic-swss/pull/645#discussion_r226145134) [](ancestors = 226145134)

jipanyang: It might be helpful if m_portCount which is the port count in ASIC gets logged together here.

qiluo-msft:
Retest this please

PR #661, opened 2018-10-30
Support neighsyncd system warmreboot.

zhenggen-xu:
Support neighsyncd system warmreboot.

neighsyncd will waits for kernel restore process to be done
before reconciliation

Add vs testcases to cover kernel neighbor table restore process
and neignsyncd process upon system warm reboot

**What I did**
Support neighsyncd system warmreboot.

neighsyncd will waits for kernel restore process to be done
before reconciliation

Add vs testcases to cover kernel neighbor table restore process
and neignsyncd process upon system warm reboot

**Why I did it**
Support system warm reboot

**How I verified it**
vs test cases and on box manual tests

**Details if related**
This is dependent on below PRs:
https://github.com/Azure/sonic-buildimage/pull/2213
https://github.com/Azure/sonic-swss-common/pull/243
https://github.com/Azure/sonic-swss-common/pull/246

### DIFF
--- a/neighsyncd/neighsync.cpp
+++ b/neighsyncd/neighsync.cpp
@@ -16,10 +16,24 @@
 using namespace std;
 using namespace swss;
 
-NeighSync::NeighSync(RedisPipeline *pipelineAppDB) :
+NeighSync::NeighSync(RedisPipeline *pipelineAppDB, DBConnector *stateDb) :
     m_neighTable(pipelineAppDB, APP_NEIGH_TABLE_NAME),
+    m_stateNeighRestoreTable(stateDb, STATE_NEIGH_RESTORE_TABLE_NAME),
     m_AppRestartAssist(pipelineAppDB, "neighsyncd", "swss", &m_neighTable, DEFAULT_NEIGHSYNC_WARMSTART_TIMER)
 {
+}
+
+// Check if neighbor table is restored in kernel
+bool NeighSync::isNeighRestoreDone()
+{
+    vector<FieldValueTuple> temp;
+
+    if (m_stateNeighRestoreTable.get("Flags", temp))
+    {
+        SWSS_LOG_NOTICE("neighbor table restore to kernel is done");
+        return true;
+    }
+    return false;
 }
 
 void NeighSync::onMsg(int nlmsg_type, struct nl_object *obj)
--- a/neighsyncd/neighsync.cpp
+++ b/neighsyncd/neighsync.cpp
@@ -26,19 +26,13 @@ NeighSync::NeighSync(RedisPipeline *pipelineAppDB, DBConnector *stateDb) :
 // Check if neighbor table is restored in kernel
 bool NeighSync::isNeighRestoreDone()
 {
-    vector<FieldValueTuple> temp;
+    string value;
 
-    if (m_stateNeighRestoreTable.get("Flags", temp))
+    m_stateNeighRestoreTable.hget("Flags", "restored", value);
+    if (value == "true")
     {
-        for (auto it : temp)
-        {
-            if (it.first == "restored" && it.second == "true")
-            {
-                SWSS_LOG_NOTICE("neighbor table restore to kernel is done");
-                return true;
-            }
-        }
-
+        SWSS_LOG_NOTICE("neighbor table restore to kernel is done");
+        return true;
     }
     return false;
 }
--- a/neighsyncd/neighsync.cpp
+++ b/neighsyncd/neighsync.cpp
@@ -30,8 +30,15 @@ bool NeighSync::isNeighRestoreDone()
 
     if (m_stateNeighRestoreTable.get("Flags", temp))
     {
-        SWSS_LOG_NOTICE("neighbor table restore to kernel is done");
-        return true;
+        for (auto it : temp)
+        {
+            if (it.first == "restored" && it.second == "true")
+            {
+                SWSS_LOG_NOTICE("neighbor table restore to kernel is done");
+                return true;
+            }
+        }
+
     }
     return false;
 }
--- a/neighsyncd/neighsync.h
+++ b/neighsyncd/neighsync.h
@@ -15,9 +15,11 @@ class NeighSync : public NetMsg
 public:
     enum { MAX_ADDR_SIZE = 64 };
 
-    NeighSync(RedisPipeline *pipelineAppDB);
+    NeighSync(RedisPipeline *pipelineAppDB, DBConnector *stateDb);
 
     virtual void onMsg(int nlmsg_type, struct nl_object *obj);
+
+    bool isNeighRestoreDone();
 
     AppRestartAssist *getRestartAssist()
     {
@@ -25,6 +27,7 @@ public:
     }
 
 private:
+    Table m_stateNeighRestoreTable;
     ProducerStateTable m_neighTable;
     AppRestartAssist m_AppRestartAssist;
 };
--- a/neighsyncd/neighsync.h
+++ b/neighsyncd/neighsync.h
@@ -8,6 +8,11 @@
 
 #define DEFAULT_NEIGHSYNC_WARMSTART_TIMER 5
 
+//This is the timer value (in seconds) that the neighsyncd waiting for restore_neighbors
+//service to finish, should be longer than the restore_neighbors timeout value (300)
+//This should not happen, if happens, system is in a unknown state, we should exit.
+#define RESTORE_NEIGH_WAIT_TIME_OUT 600
+
 namespace swss {
 
 class NeighSync : public NetMsg
--- a/neighsyncd/neighsync.h
+++ b/neighsyncd/neighsync.h
@@ -9,9 +9,9 @@
 #define DEFAULT_NEIGHSYNC_WARMSTART_TIMER 5
 
 //This is the timer value (in seconds) that the neighsyncd waiting for restore_neighbors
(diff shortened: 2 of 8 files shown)


lguohan:
retest this please

zhenggen-xu:
Updated the swss-schema.md and made the state check function more accurate. 

BTW: the vs tests failing was due to the dependency on https://github.com/Azure/sonic-buildimage/pull/2213

stcheng:
retest this please

tests/test_warm_reboot.py
```diff
@@ -289,6 +289,9 @@ def stop_neighsyncd(dvs):
 def start_neighsyncd(dvs):
     dvs.runcmd(['sh', '-c', 'supervisorctl start neighsyncd'])
 
+def start_restore_neighbors(dvs):
+    dvs.runcmd(['sh', '-c', 'supervisorctl start restore_neighbors'])
+
```
jipanyang: Should a "stop_restore_neighbors(dvs)" function be prepared and used as well?

tests/test_warm_reboot.py
```diff
@@ -780,3 +821,281 @@ def test_swss_port_state_syncup(dvs, testlog):
...
+    check_kernel_reachable_v4_neigh_num(dvs, NUM_OF_NEIGHS)
+    check_kernel_reachable_v6_neigh_num(dvs, NUM_OF_NEIGHS)
+
+    # check syslog and sairedis.rec file for activities
+    check_syslog_for_neighbor_entry(dvs, marker, 0, 0, "ipv4")
+    check_syslog_for_neighbor_entry(dvs, marker, 0, 0, "ipv6")
+    (nadd, ndel) = dvs.CountSubscribedObjects(pubsub)
+    assert nadd == 0
+    assert ndel == 0
```
jipanyang: If the neighbors had IP changes,  why no add/del activities?  arpping(restore_neighbors) result not reflected to kernel?

Also should the pubsub be activated just before start_neighsyncd(dvs)?  Or you are trying to accumulate the nadd/ndel throughout the whole test_system_warmreboot_neighbor_syncup?

tests/test_warm_reboot.py
zhenggen-xu: Since restore_neighbors program is one time only and will exit by itself, so I didn't have the stop functions, but for sure we can prepare that function. will add it.

tests/test_warm_reboot.py
zhenggen-xu: We delete some IPs  and added some IPs at server side, the stale neighbors will take time to be aged out, so no delete activities at this point.  We have later test cases to simulate the age activities.
The new added IPs at servers side were not populated to kernel yet, it requires ping or other ways to get them into kernel, that is also done in later test cases.

My understanding about the pubsub is that, when we count the objects, it will read the messages since last count not from when you subscribed. In my case, the table/objects are the same and we only need pubsub once. The test results match this expectation.  

Maybe @lguohan can comment on this. Also, do we need undo subscribe to the table? that does not seem to be available now.

neighsyncd/neighsyncd.cpp
```diff
@@ -27,16 +29,21 @@ int main(int argc, char **argv)
             NetLink netlink;
             Select s;
 
-            netlink.registerGroup(RTNLGRP_NEIGH);
-            cout << "Listens to neigh messages..." << endl;
-            netlink.dumpRequest(RTM_GETNEIGH);
-
-            s.addSelectable(&netlink);
             if (sync.getRestartAssist()->isWarmStartInProgress())
```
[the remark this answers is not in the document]
zhenggen-xu: Yes, the process for that is the same. only difference is that the restore_neighbors service will restore the table to kernel in case of warm reboot case.

neighsyncd/neighsyncd.cpp
```diff
@@ -27,16 +29,21 @@ int main(int argc, char **argv)
...
             {
                 sync.getRestartAssist()->readTableToMap();
+                while (!sync.isNeighRestoreDone()) {
```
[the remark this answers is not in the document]
zhenggen-xu: Added a timeout, unless restore_neighbors service is crashed or hit exceptions, this timeout should not happen.

neighsyncd/neighsyncd.cpp
```diff
@@ -26,17 +30,32 @@ int main(int argc, char **argv)
...
 
-            netlink.registerGroup(RTNLGRP_NEIGH);
-            cout << "Listens to neigh messages..." << endl;
-            netlink.dumpRequest(RTM_GETNEIGH);
-
-            s.addSelectable(&netlink);
             if (sync.getRestartAssist()->isWarmStartInProgress())
             {
                 sync.getRestartAssist()->readTableToMap();
+                starttime = time(NULL);
```
nikos-github: I think you want to use clock_gettime in case ntp skews the clock during warm reboot.

pavel-shirshov: Or better use std::chrono::steady_clock

neighsyncd/neighsync.cpp
```diff
@@ -16,12 +16,33 @@
...
+bool NeighSync::isNeighRestoreDone()
+{
+    vector<FieldValueTuple> temp;
+
+    if (m_stateNeighRestoreTable.get("Flags", temp))
+    {
+        for (auto it : temp)
+        {
+            if (it.first == "restored" && it.second == "true")
+            {
```
jipanyang: it is optional, but Table::hget() may be used to simplify the code.

neighsyncd/neighsyncd.cpp
```diff
@@ -27,16 +31,36 @@ int main(int argc, char **argv)
...
+            using namespace std::chrono;
 
-            s.addSelectable(&netlink);
             if (sync.getRestartAssist()->isWarmStartInProgress())
             {
                 sync.getRestartAssist()->readTableToMap();
+
+                steady_clock::time_point starttime = steady_clock::now();
+                while (!sync.isNeighRestoreDone())
+                {
```
jipanyang: Just some thought on current  restore_neighbors.py implementation.  For system warm reboot, the stale entries won't be filtered out and has to wait for a new cycle of arp age out time interval,  then the whole reconciliation logic in neighsyncd becomes redundant (skippable).  neighorch simply drops the duplicate entries and accept new entries.

neighsyncd/neighsync.h
```diff
@@ -8,23 +8,31 @@
 
 #define DEFAULT_NEIGHSYNC_WARMSTART_TIMER 5
 
+//This is the timer value (in seconds) that the neighsyncd waiting for restore_neighbors
+//service to finish, should be longer than the restore_neighbors timeout value (300)
+//This should not happen, if happens, system is in a unknown state, we should exit.
+#define RESTORE_NEIGH_WAIT_TIME_OUT 600
```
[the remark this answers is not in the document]
zhenggen-xu: This is just a conservative value that prevent neighsyncd to spin forever. The timeout value in restore_neighbors is the one that need to be considered more. It was the timeout for waiting interfaces to be created and up.  I think it was too conservative (5mins) now, will adjust it.

neighsyncd/neighsyncd.cpp
zhenggen-xu: It is not exactly true, in case the neighbor ports were done during the warm reboot, the neighbor entries won't be inserted to kernel so we need reconciliation logic to remove them. 

Also it is good to use the same logic/code path to handle both scenarios. 

On top of that, we are still exploring the solutions to only get alive entries into kernel during the restoring phase if possible in later PRs.

So overall, we should use the same reconciliation logic.

neighsyncd/neighsync.cpp
zhenggen-xu: good suggestion, changed it.

lguohan:
retest this please

lguohan:
retest this please

lguohan:
after merge bgp warm boot, the test fails, can you check?

zhenggen-xu:
> after merge bgp warm boot, the test fails, can you check?

The merge/rebase on github was not correct, my commit id and content were altered, some necessary code was removed.  I will fix locally and recommit to PR.

zhenggen-xu:
Merge is done, still some failure, does not seem to relate to the PR.  can you check if everything passed without this PR ?

lguohan:
@zhenggen-xu , build looks fine with this pr. https://sonic-jenkins.westus2.cloudapp.azure.com/job/vs/job/sonic-swss-build/19/

lguohan:
retest this please

tests/test_warm_reboot.py
```diff
@@ -301,13 +307,37 @@ def check_neighsyncd_timer(dvs, timer_value):
     (exitcode, num) = dvs.runcmd(['sh', '-c', "grep getWarmStartTimer /var/log/syslog | grep neighsyncd | tail -n 1 | rev | cut -d ' ' -f 1 | rev"])
     assert num.strip() == timer_value
 
+def check_redis_neigh_entries(dvs, neigh_tbl, number):
+    (exitcode, lb_output) = dvs.runcmd(['sh', '-c', "redis-cli keys NEIGH_TABLE:lo* | grep NEI | wc -l"])
```
qiluo-msft: >redis-cli keys NEIGH_TABLE:lo* [](start = 53, length = 30)

Try prevent usage `redis-cli keys *` because if will join keys by blanks without escaping. Since you are already write python code, can you use redis-py or swsscommon ?

PR #666, opened 2018-11-01
[test]: Clean up LAGs after finishing the test

stcheng:
After the creation of the LAG and the LAG member, clean
the application database and verify that the LAG and member
are removed from the ASIC database.

### DIFF
--- a/tests/test_portchannel.py
+++ b/tests/test_portchannel.py
@@ -20,16 +20,14 @@ def test_PortChannel(dvs, testlog):
 
     time.sleep(1)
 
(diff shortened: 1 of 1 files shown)


tests/test_portchannel.py
```diff
@@ -41,3 +39,20 @@ def test_PortChannel(dvs, testlog):
             assert dvs.asicdb.portoidmap[fv[1]] == "Ethernet0"
         else:
             assert False
+
+    # remvoe port channel member
```
lguohan: remove

PR #667, opened 2018-11-01
[vlanmgrd]: Do not bring up VLAN members by default

stcheng:
VLAN members' admin status and MTU are controlled by portmdrd.
vlanmgrd shall not bring up and set MTU automatically by default.


stcheng:
@jipanyang please check

stcheng:
yes, the pull request https://github.com/Azure/sonic-buildimage/pull/2209 ensures that all VLAN members will have admin status set as up by default

PR #669, opened 2018-11-01
[teammgrd]: Add retry logic when enslaving member port into team

stcheng:
When a port is not set as admin down, teamd will prevent the port
from being added into the port channel. This could happen when some
other processes or users are executing commands to bring up the port.
This commit will try three times of enslaving before return a false.

Example error message when adding a port which is admin up:

libteamdctl: cli_usock_process_msg:
usock: Error message received: "PortAddFail"
libteamdctl: cli_usock_process_msg: usock: Error message content: "Failed to add port."
command call failed (Invalid argument)


lguohan:
use global retry please

cfgmgr/teammgr.cpp
```diff
@@ -373,7 +375,34 @@ bool TeamMgr::addLagMember(const string &lag, const string &member)
...
+
+        sleep(1);
+        retry++;
+    }
+
+    if (retry == 3)
+    {
+        SWSS_LOG_ERROR("Failed to enslave %s to %s",
+                member.c_str(), lag.c_str());
+        return false;
```
nikos-github: Shouldn't you be undoing this if you changed the state: cmd << IP_CMD << " link set dev " << member << " down; "; since you are returning false?

cfgmgr/teammgr.cpp
```diff
@@ -373,7 +417,25 @@ bool TeamMgr::addLagMember(const string &lag, const string &member)
...
+        // member port while adding this port into the port channel. This piece
+        // of code will check if the port is set to admin status up. If yes,
+        // it will retry to add the port into the port channel.
+        if (checkPortIffUp(member))
+        {
+            return task_need_retry;
+        }
+        else
+        {
+            return task_failed;
```
lguohan: >task_failed [](start = 19, length = 11)

in case of task_failed, I think we should print error message as it mean a port member is not able to be added to port channel.

cfgmgr/teammgr.cpp
```diff
@@ -207,6 +216,49 @@ void TeamMgr::doLagMemberTask(Consumer &consumer)
...
+bool TeamMgr::findPortMaster(string &master, const string &port)
+{
+    SWSS_LOG_ENTER();
+
+    vector<string> keys;
+    m_cfgLagMemberTable.getKeys(keys);
+
+    for (auto key: keys)
+    {
+        auto tokens = tokenize(key, '|');
```
lguohan: >'|'); [](start = 36, length = 5)

i think we can a separator per db? instead of hardcode the separator?

cfgmgr/teammgr.cpp
```diff
@@ -373,7 +417,25 @@ bool TeamMgr::addLagMember(const string &lag, const string &member)
     // ip link set dev <member> down;
     // teamdctl <port_channel_name> port add <member>;
     cmd << IP_CMD << " link set dev " << member << " down; ";
-    cmd << TEAMDCTL_CMD << " " << lag << " port add " << member << "; ";
+    cmd << TEAMDCTL_CMD << " " << lag << " port add " << member;
+
+    if (exec(cmd.str(), res) != 0)
+    {
+        // teamdctl port add command will fail when the member port is not
+        // set to admin status down; it is possibel that some other processes
```
lguohan: >possibel [](start = 43, length = 8)

possible

lguohan:
some minor issues, please take a look.

PR #670, opened 2018-11-02
Orchagent warm restart data restore requires three iterations

jipanyang:

**What I did**
Fix issue of "pending consumer tasks exist after restore even when the data is consistent"

**Why I did it**
Three iterations of `o->doTask();` are needed for orchagent state restore.

**How I verified it**

**Details if related**


jipanyang:
@yxieca @qiluo-msft  Please help check.

lguohan:
retest this please

PR #674, opened 2018-11-06
[aclorch]: Support both IPv4 and IPv6 for ACL table for mirror

stcheng:
When the ACL table type is mirror, both IPv4 and IPv6 IPs are to
be supported.

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -290,7 +290,6 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
         else if (attr_name == MATCH_SRC_IP || attr_name == MATCH_DST_IP)
         {
             IpPrefix ip(attr_value);
-
             if (!ip.isV4())
             {
                 SWSS_LOG_ERROR("IP type is not v4 type");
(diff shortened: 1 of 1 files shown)


orchagent/aclorch.cpp
```diff
@@ -1124,6 +1123,25 @@ bool AclTable::create()
         attr.value.booldata = true;
         table_attrs.push_back(attr);
     }
+    // ACL table for mirror session supports both IPv4 and IPv6
```
andriymoroz-mlnx: it duplicates sip/dip and sip6/dip6
and several lines below there is another if (type == ACL_TABLE_MIRROR) adding dscp attribute
could you please combine them in one place?
something like this as:
```
if (type != L3v6)
  add SRC_IP and DST_IP

if (type != L3v6 or type==MIRROR)
  add SRC_IPv6 and DST_IPv6

if (type==MIRROR)
  add DSCP
```

I'll verify if it works on MLNX switch

stcheng: so if the type == MIRROR it shall support both v4 and v6, that's why both are needed
I could change the logic to be
```
if type != v6
add SRC_IP and DST_IP
if type != v4
add SRC_IPv6 and DST_IPv6
if type == mirror
add DSCP
```
what do you think?

andriymoroz-mlnx: maybe it will work too but we also have other table types (ACL_TABLE_CTRLPLANE and ACL_TABLE_DTEL_***) and in your case SRC_IPv6 and DST_IPv6 will be applied to them as well
not sure if it is needed

found a typo in my example. should be 
```
if (type != L3v6)
  add SRC_IP and DST_IP

if (type == L3v6 or type==MIRROR)
  add SRC_IPv6 and DST_IPv6

if (type==MIRROR)
  add DSCP
```

stcheng: i will move the DSCP part into my initial code change's if condition

andriymoroz-mlnx: got a table creation failure on a MLNX switch...
will discuss with SAI/SDK

PR #678, opened 2018-11-08
Add warm-reboot mode for teammgrd

pavel-shirshov:
**What I did**
Add new functionality to support warm-reboot mode for teammgrd
1. Now teamd will be always started with option -L
2. teamd will be started with -w option when the system started in warm-reboot mode.

**Why I did it**
To support warm-reboot mode

**How I verified it**
1. Build swss with the changes. 
2. Install into teamd container. 
3. Restart your DUT.
4. Check that all teamd were started with -L option
5. enable warm-reboot mode by 'config warm_restart enable'
6. Restart teamd container
7. Check that all teamd were started with -w option

**Details if related**


pavel-shirshov:
Hi Jipan,

I'm currently testing teamd graceful shutdown integration. I hope I'll add the PR today.

About your question. Current implementation of teamd will save latest received pdu, which is not right. I found this and other issues in my teamd patch recently. I'm going to fix it ASAP.

cfgmgr/teammgr.cpp
```diff
@@ -380,7 +381,15 @@ bool TeamMgr::addLag(const string &alias, int min_links, bool fallback)
...
+    string warmstart_flag = WarmStart::isWarmStart() ? " -w " : "";
+    const string dump_path = "/var/warmboot/teamd/";
+
+    cmd << TEAMD_CMD
+        << warmstart_flag
+        << " -r -t " << alias
+        << " -c " << conf.str()
+        << " -L " << dump_path
+        << " -d";
     EXEC_WITH_ERROR_THROW(cmd.str(), res);
```
qiluo-msft: >EXEC_WITH_ERROR_THROW [](start = 4, length = 21)

For old code: execve() will be better then system(), in case of user input attach.
For example, 'alias' may contain malicious content with blanks inside.

PR #679, opened 2018-11-12
[portsorch]: Move status comparison outside updatePortOperStatus function

stcheng:
updatePortOperStatus will update the port operation status regardless of the
previous operation status. The comparison logic is moved into the notification
receiving part.


stcheng:
@jipanyang @zhenggen-xu @yxieca

PR #680, opened 2018-11-12
View comparison needs port oper state applied first.

jipanyang:

**What I did**
Push port oper state to syncd for view comparison.

**Why I did it**

View comparison logic needs the port oper state to properly work.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -445,10 +445,10 @@ bool OrchDaemon::warmRestoreAndSyncUp()
 
     SWSS_LOG_NOTICE("Orchagent state restore done");
 
-    syncd_apply_view();
-
     /* Start dynamic state sync up */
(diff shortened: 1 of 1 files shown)


qiluo-msft:
Alternatively, we can ask 'comparison logic' ignore oper_status from orchagent. The ground truth of oper status is from ASIC.

jipanyang:
> Alternatively, we can ask 'comparison logic' ignore oper_status from orchagent. The ground truth of oper status is from ASIC.

Yes, the ground truth of oper status is from ASIC.   Currently syncd view comparison logic is restoring the port oper_status from asicDB for old view,  it is not getting the oper_status from ASIC.  What stored in asicDB is from orchagent too, so it could be good to check the data again.

But I'm ok with either approach.

jipanyang:
Current solution has flaw in triggering nexthop change. Closing it.

PR #681, opened 2018-11-12
[teammgrd]: Add kernel master check for port before enslaving

stcheng:


stcheng:
@jipanyang @zhenggen-xu

cfgmgr/teammgr.cpp
```diff
@@ -237,6 +237,16 @@ bool TeamMgr::checkPortIffUp(const string &port)
     return ifr.ifr_flags & IFF_UP;
 }
 
+bool TeamMgr::checkKernelPortMaster(const string &port)
+{
+    SWSS_LOG_ENTER();
+
+    string cmd = "ls /sys/class/net/" + port + "/master";
```
qiluo-msft: Do you intend to check the file existing?

jipanyang: It looks people recommend stat  system call for checking existence of directory/file: 

https://stackoverflow.com/questions/18100097/portable-way-to-check-if-directory-exists-windows-linux-c,

https://stackoverflow.com/questions/3828192/checking-if-a-directory-exists-in-unix-system-call

stcheng:
retest this please

stcheng:
thanks for the comments, i decided to use `lstat` function as it meets my need best.

cfgmgr/teammgr.cpp
```diff
@@ -237,6 +238,16 @@ bool TeamMgr::checkPortIffUp(const string &port)
     return ifr.ifr_flags & IFF_UP;
 }
 
+bool TeamMgr::checkKernelPortMaster(const string &port)
```
svc-acs: the name is so technical and it does not reflect the purpose. the purpose here is to find if the port is already enslave or not. so, better to rename the function as isPortEnslaved()

stcheng:
@lguohan updated

PR #683, opened 2018-11-13
Restore port oper status form appDB port table during restore phase o…

jipanyang:
…f warm restart

**What I did**

Restore port oper-status from appDB port table during restore phase of warm restart/reboot.

**Why I did it**
oper-status is needed for libsairedis interface to complete the picture.
nexthop state won't be affected in this case since the port oper-status had been applied before shutdown.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2914,11 +2914,7 @@ void PortsOrch::doTask(NotificationConsumer &consumer)
                 SWSS_LOG_ERROR("Failed to get port object for port id 0x%lx", id);
                 continue;
             }
-
-            if (status != port.m_oper_status)
-            {
-                updatePortOperStatus(port, status);
-            }
+            updatePortOperStatus(port, status);
         }
 
         sai_deserialize_free_port_oper_status_ntf(count, portoperstatus);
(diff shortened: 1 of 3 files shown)


jipanyang:
Actually there is an issue here.   

Suppose there is no change to the oper-status to a port which is up,
View comparison or other logic will decide that no need to push the SAI_HOSTIF_ATTR_OPER_STATUS set request to libsai/SDK, while during system warm reboot the default oper-status in hostif is down.  We have to put some exception logic.

Why didn't libsai change the hostif oper status automatically upon physical port oper status up?

lguohan:
@jipanyang , can you add vstest?

qiluo-msft:
Please wait @stcheng comments.

orchagent/portsorch.cpp
```diff
@@ -2304,6 +2304,24 @@ bool PortsOrch::initializePort(Port &p)
         p.m_oper_status = SAI_PORT_OPER_STATUS_UP;
     }
 
+    /*
+     * If oper_status is not empty, orchagent is doing warm start, restore hostif oper status.
+     */
+    if (!operStatus.empty())
```
stcheng: is this condition good enough? i wonder if there would be some other flags that could check it is a warm start or not

stcheng: okay i get it. thanks

PR #685, opened 2018-11-13
[WarmStart]: Use right docker name for the teammgrd checkWarmStart

pavel-shirshov:
**What I did**
Fix usage of WarmStart.

**Why I did it**
Otherwise teammgrd didn't see WarmStart flag

**How I verified it**
Build it, install into teamd container. Stop all teamd, then enable WarmStart mode and start teammgrd.

**Details if related**


cfgmgr/teammgr.cpp
```diff
@@ -381,12 +381,12 @@ bool TeamMgr::addLag(const string &alias, int min_links, bool fallback)
     SWSS_LOG_INFO("Port channel %s teamd configuration: %s",
             alias.c_str(), conf.str().c_str());
 
-    string warmstart_flag = WarmStart::isWarmStart() ? " -w " : "";
+    string warmstart_flag = WarmStart::isWarmStart() ? " -w -o " : " -r ";
```
stcheng: remove the trailing white space in the flag:
`" -w -o" : " -r"`

PR #686, opened 2018-11-13
Process pending ports for aclOrch at end of db data replay for warm r…

jipanyang:
…estore

**What I did**
Process pending ports for aclOrch at the end of db data replay for warm restore.

**Why I did it**
Current aclOrch implementation put port into AclTable pendingPortSet if port not available, and wait for lag state flag from stateDB.

During warm restore, no state flag will be checked actively. Here we process the pending ports expicitly.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -2127,7 +2127,7 @@ void AclOrch::processPendingPorts()
         auto table = itmap.second;
         for (auto port_alias: table.pendingPortSet)
         {
-            SWSS_LOG_INFO("found the port: %s in ACL table: %s pending port list, bind it to ACL table.", port_alias.c_str(), table.description.c_str());
+            SWSS_LOG_NOTICE("found the port: %s in ACL table: %s pending port list, bind it to ACL table.", port_alias.c_str(), table.description.c_str());
 
             bool suc = processPendingPort(table, port_alias, [&](sai_object_id_t portOid) {
                 table.link(portOid);
@@ -2145,6 +2145,7 @@ void AclOrch::processPendingPorts()
         }
     }
 }
+
 void AclOrch::doAclTablePortUpdateTask(Consumer &consumer)
 {
     SWSS_LOG_ENTER();
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -2120,6 +2120,31 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
     }
 }
 
+void AclOrch::processPendingPorts()
+{
+    for (auto itmap : m_AclTables)
+    {
+        auto table = itmap.second;
+        for (auto port_alias: table.pendingPortSet)
+        {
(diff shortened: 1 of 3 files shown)


orchagent/aclorch.cpp
```diff
@@ -2120,6 +2120,46 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
...
+    for (auto& pair: gPortsOrch->getAllPorts())
+    {
+        string port_alias = pair.first;
+        processPendingPort(port_alias);
+    }
+}
+
+void AclOrch::processPendingPort(string &port_alias)
+{
+    for (auto& pair: gPortsOrch->getAllPorts())
```
zhenggen-xu: Don't think we need this loop.

orchagent/aclorch.cpp
```diff
@@ -2120,6 +2120,46 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
     }
 }
 
+void AclOrch::processPendingPorts()
+{
```
zhenggen-xu: Is it more effective to just go through the pending list (table.pendingPortSet)?

orchagent/aclorch.cpp
```diff
@@ -2120,6 +2120,31 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
...
+                SWSS_LOG_ERROR("Failed to bind the ACL table: %s to port: %s", table.description.c_str(), port_alias.c_str());
+            }
+            else
+            {
+                table.pendingPortSet.erase(port_alias);
+                SWSS_LOG_DEBUG("port: %s bound to ACL table table: %s, remove it from pending list", port_alias.c_str(), table.description.c_str());
+            }
+        }
+    }
+}
```
qiluo-msft: add one empty line

orchagent/orchdaemon.cpp
```diff
@@ -431,6 +431,13 @@ bool OrchDaemon::warmRestoreAndSyncUp()
         }
     }
 
+    /*
+     * There could be pending ports for ACL.
```
qiluo-msft: >There could be pending ports for ACL. [](start = 7, length = 37)

Is this PR a quick fix and will be replaced? please add TODO if yes.

orchagent/aclorch.cpp
```diff
@@ -2120,6 +2120,32 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
...
 }
 
+void AclOrch::processPendingPorts()
+{
+    for (auto itmap : m_AclTables)
+    {
+        auto table = itmap.second;
+        for (auto port_alias: table.pendingPortSet)
+        {
+            SWSS_LOG_INFO("found the port: %s in ACL table: %s pending port list, bind it to ACL table.", port_alias.c_str(), table.description.c_str());
```
zhenggen-xu: would it be better to make this NOTICE level so we have visibility for pending ports?

jipanyang:
Original problem fixed in https://github.com/Azure/sonic-swss/pull/689

stcheng:
thanks for the temporary fix!

PR #687, opened 2018-11-14
[teammgrd]: Check if port is enslaved on all conditions

stcheng:
[teammgrd]: Check if port is enslaved on all conditions

    During the warm restart, port is already enslaved in the kernel.
    This change is to avoid teammgrd to re-configure those member
    ports which causes link flap.


qiluo-msft:
Write some meaningful commit message?

lguohan:
full description of the fix?

stcheng:
@qiluo-msft @lguohan commit message and comments are updated

PR #689, opened 2018-11-16
[aclorch]: Use the Observer class to listen to port changes

stcheng:
The previous implementation was using state database to check LAG
creation/removal. However, state database only reflects the changes
in the kernel.

This pull request changes the logic so that LAG changes will directly
notify the AclOrch to update pending ports per table. This simplifies
the logic in AclOrch and make it reliable on port change events.

Right now, only LAG creation is support, which aligns the legacy codes.
An extra pull request will be provided to ensure both LAG creation/removal
will be supported.

The unit test is provided to cover the two scenarios:
1. creating ACL table before creating LAG
2. creating LAG before creating ACL table


stcheng:
@jipanyang @zhenggen-xu @qiluo-msft

orchagent/aclorch.cpp
```diff
@@ -1177,6 +1177,47 @@ bool AclTable::create()
...
+
+    // Only interested in port change
+    if (type != SUBJECT_TYPE_PORT_CHANGE)
+    {
+        return;
+    }
+
+    PortUpdate *update = static_cast<PortUpdate *>(cntx);
+
+    Port port = update->port;
```
jipanyang: Using port reference might be better.

stcheng:
retest this please

orchagent/aclorch.cpp
```diff
@@ -1177,6 +1177,47 @@ bool AclTable::create()
...
+    {
+        if (pendingPortSet.find(port.m_alias) != pendingPortSet.end())
+        {
+            sai_object_id_t bind_port_id;
+            if (gPortsOrch->getAclBindPortId(port.m_alias, bind_port_id))
+            {
+                link(bind_port_id);
+                bind(bind_port_id);
+
+                SWSS_LOG_NOTICE("Bound port %s to ACL table %s",
```
zhenggen-xu: Should we remove the port from pendingPortSet after successfully bind?

orchagent/aclorch.cpp
```diff
@@ -2120,140 +2175,40 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
...
 
-    for (const auto& alias : strList)
+    for (auto alias : ports)
     {
-        sai_object_id_t port_id;
         Port port;
         if (!gPortsOrch->getPort(alias, port))
         {
-            SWSS_LOG_INFO("Port %s not configured yet, add it to ACL table %s pending list", alias.c_str(), aclTable.description.c_str());
+            SWSS_LOG_NOTICE("Add unready port %s to pending list for ACL table %s",
```
lguohan: INFO level.

orchagent/aclorch.cpp
```diff
@@ -2120,140 +2175,40 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
...
-    {
-        SWSS_LOG_ERROR("Failed to process port list. Duplicate port entry");
-        return false;
-    }
-
-    if (strList.empty())
+    if (ports.empty())
     {
-        SWSS_LOG_ERROR("Failed to process port list. List is empty");
+        SWSS_LOG_ERROR("Failed to process empty port list");
```
lguohan: >SWSS_LOG_ERROR [](start = 8, length = 14)

why error level? it mean the acl table currently does not bind to any port. It should be a valid configuration. We probably should not print out anything.

orchagent/aclorch.cpp
lguohan: line 1203

---
In reply to: [234106802](https://github.com/Azure/sonic-swss/pull/689#discussion_r234106802) [](ancestors = 234106802)

tests/test_acl_portchannel.py
```diff
@@ -0,0 +1,190 @@
...
+        time.sleep(1)
+
+    def remove_acl_rule(self, dvs, table_name, rule_name):
+        tbl = swsscommon.Table(self.cdb, "ACL_RULE")
+        tbl._del(table_name + "|" + rule_name, fvs)
+        time.sleep(1)
+
+    # Frist create port channel
+    # Second create ACL table
+    def test_PortChannelAcl(self, dvs):
```
lguohan: >test_PortChannelAcl [](start = 8, length = 19)

test_PortChannelBeforeAcl

and 

test_PortChannelAfterAcl

try to give meaningful name to function

tests/test_acl_portchannel.py
```diff
@@ -0,0 +1,190 @@
...
+
+        # create port channel
+        self.create_port_channel(dvs, "PortChannel01")
+
+        # create ACL table
+        self.create_acl_table(dvs, "LAG_ACL_TABLE", "PortChannel01")
+
+        # check ASIC table
+        tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_LAG")
+        lag = tbl.getKeys()[0]
```
lguohan: sleep here.

tests/test_acl_portchannel.py
```diff
@@ -0,0 +1,190 @@
...
+    def test_PortChannelAcl2(self, dvs):
+        self.setup_db(dvs)
+
+        # create ACL table
+        self.create_acl_table(dvs, "LAG_ACL_TABLE", "PortChannel01")
+
+        # create port channel
+        self.create_port_channel(dvs, "PortChannel01")
+
+        # check ASIC table
```
lguohan: sleep here.

tests/test_acl_portchannel.py
```diff
@@ -0,0 +1,190 @@
...
+            elif fv[0] == "SAI_ACL_TABLE_GROUP_MEMBER_ATTR_ACL_TABLE_ID":
+                table_id = fv[1]
+            elif fv[0] == "SAI_ACL_TABLE_GROUP_MEMBER_ATTR_PRIORITY":
+                assert fv[1] == "100"
+            else:
+                assert False
+
+        tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_ACL_TABLE")
+        (status, fvs) = tbl.get(table_id)
+        assert status == True
```
lguohan: check asic db is duplicated. can you share the code with the previous function. need to avoid code duplication.

orchagent/aclorch.cpp
stcheng: Right now adding ports is not supported when creating the ACL table. This log means that the ACL table entry doesn't have any port associated with this. If we want to add ports to this ACL table, the current behavior is to remove old ACL table and create a new one. I have added TODO here.

tests/test_acl_portchannel.py
stcheng: check operations do not need sleep

tests/test_acl_portchannel.py
stcheng: check operations do not need sleep

stcheng:
@jipanyang @zhenggen-xu @lguohan comments have been addressed

orchagent/aclorch.cpp
zhenggen-xu: It was not there in the original code, but now looks good.

PR #690, opened 2018-11-16
[portsorch]: Remove duplicate local variables - port

stcheng:

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -412,6 +412,7 @@ bool PortsOrch::getPortByBridgePortId(sai_object_id_t bridge_port_id, Port &port
     return false;
 }
 
+// TODO: move this into AclOrch
 bool PortsOrch::getAclBindPortId(string alias, sai_object_id_t &port_id)
 {
     SWSS_LOG_ENTER();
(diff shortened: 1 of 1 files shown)


orchagent/portsorch.cpp
```diff
@@ -739,56 +735,59 @@ bool PortsOrch::bindAclTable(sai_object_id_t id, sai_object_id_t table_oid, sai_
...
         {
-        case Port::PHY:
-        {
-            // Bind this ACL group to physical port
-            sai_attribute_t port_attr;
-            port_attr.id = ingress ? SAI_PORT_ATTR_INGRESS_ACL : SAI_PORT_ATTR_EGRESS_ACL;
-            port_attr.value.oid = groupOid;
-
-            status = sai_port_api->set_port_attribute(port.m_port_id, &port_attr);
-            if (status != SAI_STATUS_SUCCESS)
```
stcheng: Fix the indentation here in this switch. No code changes.

orchagent/portsorch.cpp
```diff
@@ -735,60 +731,65 @@ bool PortsOrch::bindAclTable(sai_object_id_t id, sai_object_id_t table_oid, sai_
             port.m_egress_acl_table_group_id = groupOid;
         }
 
+        setPort(port.m_alias, port);
```
zhenggen-xu: Should we save the information after binding successfully?

zhenggen-xu: ok,  we can save the information here.  But we may need some way to indicate whether the binding was successful or not.  Currently, in many places we ignored the bind() function results.  We need fix those places.  We should address it if not in this PR.

Issue #623, opened 2018-09-18
Ports with IP addresses configured are placed in admin shut state


nikos-github:
Upgraded switch from July release to latest image. Used previous working configuration and on reload, ports are placed in admin shut state even though IP addresses are configured. The reason is config_db.json does not have admin_status set to up which is a requirement in latest code. This will break all existing configurations when upgrading to latest versions. The admin_status filed change is also not backwards compatible.

lguohan:
@stcheng to comment

PR #691, opened 2018-11-16
Fix setting default port admin status and mtu

stepanblyschak:
This commit moves setting of default admin status and mtu
to portmgrd to be aligned with other managers (e.g. teammgrd).

It also fixes "N/A" admin status in "show interface status" output
for ports which have no 'admin_status' field in config DB.

**What I did**
Move defaults port settings to portmgrd, remove settings admin status down in portsorch in initializePort

**Why I did it**
Remove setting default admin status to down in case of cold boot because of two reasons:
1. SAI default is admin status down
2. Defaults port settings should come from portmgrd

**How I verified it**
Verify on DUT that port settings are applied correctly. Check that iin case admin/mtu is missing for port in config db defaults are applied and no N/A in 'show interface status'

**Details if related**

### DIFF
--- a/cfgmgr/portmgr.cpp
+++ b/cfgmgr/portmgr.cpp
@@ -10,6 +10,10 @@
 using namespace std;
 using namespace swss;
 
+/* Port default admin status is down */
+#define DEFAULT_ADMIN_STATUS_STR    "down"
+#define DEFAULT_MTU_STR             "9100"
+
 PortMgr::PortMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, const vector<string> &tableNames) :
         Orch(cfgDb, tableNames),
         m_cfgPortTable(cfgDb, CFG_PORT_TABLE_NAME),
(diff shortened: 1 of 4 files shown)


cfgmgr/portmgr.cpp
```diff
@@ -10,6 +10,10 @@
 using namespace std;
 using namespace swss;
 
+/* Port default admin status is down */
+#define DEFAULT_ADMIN_STATUS_STR    "down"
```
nikos-github: The default port admin status must be up (see also https://github.com/Azure/sonic-swss/issues/623)

prsunny: I think, the default must be down.

cfgmgr/portmgr.cpp
```diff
@@ -91,23 +95,44 @@ void PortMgr::doTask(Consumer &consumer)
...
+            if (!mtu.empty())
+            {
+                setPortMtu(alias, mtu);
+                SWSS_LOG_NOTICE("Configure %s MTU to %s", alias.c_str(), mtu.c_str());
+            }
+
+            if (!admin_status.empty())
+            {
+                setPortAdminStatus(alias, admin_status == "up");
+                SWSS_LOG_NOTICE("Configure %s %s", alias.c_str(), admin_status.c_str());
```
pavel-shirshov: can you please specify "admin status" in the log output

cfgmgr/portmgr.cpp
Blyschak: Currently default is down anyway. IMO if not specified explicitly in config_db.json it should be 'down'.

cfgmgr/portmgr.cpp
Blyschak: Done

cfgmgr/portmgr.cpp
stepanblyschak: Currently default is 'down'. IMO if it is not specified explicitely in config_db.json it should be 'down'

cfgmgr/portmgr.cpp
stepanblyschak: Done

stepanblyschak:
retest this please

cfgmgr/portmgr.cpp
nikos-github: @stepanblyschak The default was up if it wasn't specified in the config_db.json and the port had an IP (as of July release). Recent change broke/changed this and also caused older configs to not work and be backwards incompatible. It is not acceptable for customers to move from one release to another and all of a sudden have their configs not working.

lguohan:
@stepanblyschak , can you rebase and test?

cfgmgr/portmgr.cpp
```diff
@@ -91,23 +95,44 @@ void PortMgr::doTask(Consumer &consumer)
                 continue;
             }
 
+            string admin_status, mtu;
+
+            bool configured = (m_portList.find(alias) != m_portList.end());
+
+            /* If this is the first time we set port settings
+             * asign default admin status and mtu
```
stcheng: assign

cfgmgr/portmgr.cpp
stcheng: @nikos-github i think this pull request doesn't change the current behavior. the admin status 'up' is set by converting the information from the minigraph to the configuration database, where the admin status is set to up for all ports that have IP addresses.

nikos-github:
@stcheng @stepanblyschak Please provide testing results with the same config_db.json on an image with this fix and on an image from the July release. If the results regarding the port admin status show no discrepancies and the configs work without changes in both releases, I have no problem in withdrawing my objection. Test should include ports without admin status config, with and without IP address, with admin status config (up and down) and with and without IP address.

cfgmgr/portmgr.cpp
stepanblyschak: thanks, fixed

stcheng:
@nikos-github I get your point that if the port that configured with IP on it without having 'admin_status' set to UP in the configuration database, then the original behavior was to bring up the port. We consider the current behavior is the correct while in order to solve the backward compatibility issue. I will provide a separate solution for configuration database migration from older version to newer ones with a better general solution. Since this pull request is not the original one that causes the backward compatibility issue, could you unblock this pull request and we will address the issue separately?

stcheng:
retest this please

PR #692, opened 2018-11-16
Remove default docker name value of swss.

jipanyang:
The default docker name caused a lot of confusion when the libraries used in other dockers

**What I did**

Adapt to the change with WarmStart library,  default docker name of swss has been removed.

**Why I did it**
The default docker name caused a lot of confusion when the libraries used in other dockers
**How I verified it**
VS test.
**Details if related**


stcheng:
check the VS test failure?

jipanyang:
> check the VS test failure?

How to access the syslog of vs test?

PR #693, opened 2018-11-16
Support ConfigDB neighbor configuration, introduce nbrmgr daemon

prsunny:
**What I did**
Initial commit to support ConfigDB neighbor table configurations

**Why I did it**
Requirement to resolve mac address for a configured neighbor IP. Also support static neighbor configuration

**How I verified it**
Logs:

Resolve case:
```
{
    "NEIGH_TABLE": {
        "Vlan100|100.1.1.3": { 
            "family": "IPv4" 
        }
    }
}

19:01:02.216254 ARP, Request who-has 100.1.1.3 tell 100.1.1.1, length 28
```

Static Nbr:
```
{
    "NEIGH_TABLE": {
        "Vlan100|100.1.1.5": { 
            "neigh": "00:02:02:03:04:05",
            "family": "IPv4" 
        }
    }
}

[admin@str-a7170-acs-1 ~ 08:10 PM]$ ip neigh show 
100.1.1.3 dev Vlan100  FAILED
100.1.1.5 dev Vlan100 lladdr 00:02:02:03:04:05 PERMANENT
```

**Details if related**
This is the initial commit for phase 1.
For phase 2:
1. Combine nbrmgrd and neighsync. 
2. Address any priority items from https://github.com/Azure/sonic-buildimage/pull/1904#issuecomment-412168903

### DIFF
--- a/cfgmgr/nbrmgr.cpp
+++ b/cfgmgr/nbrmgr.cpp
@@ -1,0 +1,236 @@
+#include <arpa/inet.h>
+#include <netinet/in.h>
+#include <net/if.h>
+#include <unistd.h>
+
+#include "logger.h"
+#include "tokenize.h"
+#include "ipprefix.h"
+#include "macaddress.h"
+#include "nbrmgr.h"
+#include "exec.h"
+#include "shellcmd.h"
+
+using namespace swss;
+
+const int IP4_ADDR_LEN  = 4;
+const int IP6_ADDR_LEN  = 16;
+const int NL_MSG_BUF_SZ = 512;
+
+#define VLAN_PREFIX "Vlan"
+#define LAG_PREFIX  "PortChannel"
+
+NbrMgr::NbrMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, const vector<string> &tableNames) :
+        Orch(cfgDb, tableNames),
+        m_statePortTable(stateDb, STATE_PORT_TABLE_NAME),
+        m_stateLagTable(stateDb, STATE_LAG_TABLE_NAME),
+        m_stateVlanTable(stateDb, STATE_VLAN_TABLE_NAME),
+        m_stateIntfTable(stateDb, STATE_INTERFACE_TABLE_NAME)
+{
+
+}
+
+bool NbrMgr::isIntfStateOk(const string &alias)
+{
+    vector<FieldValueTuple> temp;
+
+    if (!alias.compare(0, strlen(VLAN_PREFIX), VLAN_PREFIX))
+    {
+        if (m_stateVlanTable.get(alias, temp))
+        {
+            SWSS_LOG_DEBUG("Vlan %s is ready", alias.c_str());
+            return true;
+        }
+    }
+    else if (!alias.compare(0, strlen(LAG_PREFIX), LAG_PREFIX))
+    {
+        if (m_stateLagTable.get(alias, temp))
+        {
+            SWSS_LOG_DEBUG("Lag %s is ready", alias.c_str());
+            return true;
+        }
+    }
+    else if (m_statePortTable.get(alias, temp))
+    {
+        SWSS_LOG_DEBUG("Port %s is ready", alias.c_str());
+        return true;
+    }
+
+    return false;
+}
+
+bool NbrMgr::sendMsg(struct nlmsghdr *msg)
+{
+    SWSS_LOG_ENTER();
+
+    int sk = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
+    if (sk < 0)
+    {
+        SWSS_LOG_ERROR("Socket create error");
+        return false;
+    }
+
+    struct sockaddr_nl nl_addr ;
+    memset(&nl_addr, 0, sizeof(nl_addr));
+
+    nl_addr.nl_family = AF_NETLINK;
+
+    struct iovec iov[1] =
+    {
+        {
+          .iov_base = msg,
+          .iov_len = msg->nlmsg_len
+        }
+    };
+
+    struct msghdr msg_hdr =
+    {
+        .msg_name = &nl_addr,
+        .msg_namelen = sizeof(nl_addr),
+        .msg_iov = iov,
+        .msg_iovlen = 1,
+        .msg_control = NULL,
+        .msg_controllen = 0,
+        .msg_flags =0
+    };
+
+    if (sendmsg(sk, &msg_hdr, 0) != (msg->nlmsg_len))
+    {
+        SWSS_LOG_ERROR("Error sending message to kernel");
+        close(sk);
+        return false;
+    }
+
+    close(sk);
+
+    SWSS_LOG_INFO("Neighbor add msg sent to kernel");
+    return true;
+}
+
+bool NbrMgr::setNeighbor(const string& alias, IpAddress& ip, MacAddress& mac)
+{
+    SWSS_LOG_ENTER();
+
+    char m_buf[NL_MSG_BUF_SZ];
+    memset(m_buf, 0, sizeof(struct nlmsghdr));
+
+    auto alloc_fn = [&] (struct nlmsghdr *msg, uint32_t len) {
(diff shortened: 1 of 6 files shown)


cfgmgr/nbrmgr.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include "ipprefix.h"
+#include "macaddress.h"
+#include "nbrmgr.h"
+#include "exec.h"
+#include "shellcmd.h"
+
+using namespace swss;
+
+const int IP4_ADDR_LEN  = 4;
+const int IP6_ADDR_LEN  = 16;
```
pavel-shirshov: = sizeof(struct in_addr);
 = sizeof(struct in6_addr);

cfgmgr/nbrmgr.cpp
```diff
@@ -0,0 +1,236 @@
...
+        return false;
+    }
+
+    close(sk);
+
+    SWSS_LOG_INFO("Neighbor add msg sent to kernel");
+    return true;
+}
+
+bool NbrMgr::setNeighbor(const string& alias, IpAddress& ip, MacAddress& mac)
```
pavel-shirshov: ip and mac should be const too?

cfgmgr/nbrmgr.cpp
prsunny: Addressed

cfgmgr/nbrmgr.cpp
prsunny: Done

cfgmgr/nbrmgr.cpp
```diff
@@ -0,0 +1,210 @@
...
+static bool send_message(struct nl_msg *msg)
+{
+    struct nl_sock *sk = nl_socket_alloc();
+    if (!sk)
+    {
+        SWSS_LOG_ERROR("Netlink socket alloc failed");
+        return false;
+    }
+
+    nl_connect(sk, NETLINK_ROUTE);
```
pavel-shirshov: can you please check result of nl_connect?

cfgmgr/nbrmgr.cpp
```diff
@@ -0,0 +1,210 @@
...
+    bool rc = true;
+
+    if (nl_send_auto(sk, msg) < 0)
+    {
+        SWSS_LOG_ERROR("Netlink send message failed");
```
pavel-shirshov: Can you please use nl_geterror to get netlink error text? Can you please add the decrypting of nl error message on every nl library call?

cfgmgr/nbrmgr.cpp
```diff
@@ -0,0 +1,210 @@
...
+    struct nl_msg *msg = nlmsg_alloc();
+    if (!msg)
+    {
+        SWSS_LOG_ERROR("Netlink message alloc failed for '%s'", ip.to_string().c_str());
+        return false;
+    }
+
+    auto flags = (NLM_F_REQUEST | NLM_F_ACK | NLM_F_CREATE | NLM_F_REPLACE);
+
+    nlmsg_put(msg, NL_AUTO_PORT, NL_AUTO_SEQ, RTM_NEWNEIGH, 0, flags);
```
pavel-shirshov: check for return value?

cfgmgr/nbrmgr.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+    struct ndmsg *nd_msg = static_cast<struct ndmsg *>
+                           (nlmsg_reserve(msg, sizeof(struct ndmsg), NLMSG_ALIGNTO));
```
pavel-shirshov: check for return value?

cfgmgr/nbrmgr.cpp
```diff
@@ -0,0 +1,210 @@
...
+    memset(nd_msg, 0, sizeof(struct ndmsg));
+
+    nd_msg->ndm_ifindex = if_nametoindex(alias.c_str());
+
+    auto addr_len = ip.isV4()? sizeof(struct in_addr) : sizeof(struct in6_addr);
+
+    struct rtattr *rta = static_cast<struct rtattr *>
+                         (nlmsg_reserve(msg, sizeof(struct rtattr) + addr_len, NLMSG_ALIGNTO));
```
pavel-shirshov: check the return value?

cfgmgr/nbrmgr.cpp
```diff
@@ -0,0 +1,210 @@
...
+    {
+        SWSS_LOG_INFO("Set mac address '%s'", mac.to_string().c_str());
+
+        nd_msg->ndm_state = NUD_PERMANENT;
+
+        auto mac_len = ETHER_ADDR_LEN;
+        auto mac_addr = mac.getMac();
+
+        struct rtattr *rta = static_cast<struct rtattr *>
+                             (nlmsg_reserve(msg, sizeof(struct rtattr) + mac_len, NLMSG_ALIGNTO));
```
pavel-shirshov: return value?

cfgmgr/nbrmgr.cpp
pavel-shirshov: Is it possible to connect to the socket just once and then keep it opened?

cfgmgr/nbrmgr.cpp
prsunny: Done

cfgmgr/nbrmgr.cpp
prsunny: Done

cfgmgr/nbrmgr.cpp
prsunny: yes

cfgmgr/nbrmgr.cpp
```diff
@@ -0,0 +1,247 @@
...
+        string op = kfvOp(t);
+        MacAddress mac;
+
+        for (auto idx : data)
+        {
+            const auto &field = fvField(idx);
+            const auto &value = fvValue(idx);
+            if (field == "neigh")
+            {
+                mac = value;
```
jipanyang: This is a common issue: If the value field is not a valid mac address string,  this operation will cause nbrmgr to throw and exit.  Not sure if that is the desired behavior or we want to put some check and log error here?

cfgmgr/nbrmgr.cpp
```diff
@@ -0,0 +1,247 @@
...
+    if (!nd_msg)
+    {
+        SWSS_LOG_ERROR("Netlink ndmsg reserve failed for '%s'", ip.to_string().c_str());
+        nlmsg_free(msg);
+        return false;
+    }
+
+    memset(nd_msg, 0, sizeof(struct ndmsg));
+
+    nd_msg->ndm_ifindex = if_nametoindex(alias.c_str());
```
pavel-shirshov: This guys could return 0, if there is no such interface

cfgmgr/nbrmgr.cpp
```diff
@@ -0,0 +1,247 @@
...
+        memcpy(RTA_DATA(rta), &ip_addr.ip_addr.ipv6_addr, addr_len);
+    }
+
+    if (!mac)
+    {
+        /*
+         * If mac is not provided, expected to resolve the MAC
+         */
+        nd_msg->ndm_state = NUD_DELAY;
+        nd_msg->ndm_flags = NTF_USE;
```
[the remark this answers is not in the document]
prsunny: The max I saw is 5minutes. Essentially it is anytime between <0-5mts>

cfgmgr/nbrmgr.cpp
prsunny: Yes, but I thought since we have a previous check on interface status, expecting this to be valid at this point.

cfgmgr/nbrmgr.cpp
prsunny: It is not a desired behavior. We would want to continue even if there is an incorrect mac configuration. I'll fix this by catching the exception.

prsunny: @jipanyang , fixed this.

PR #694, opened 2018-11-17
ASIC temperature sensors support

padmanarayana:
**Please do not merge yet.**

Recently (https://github.com/opencomputeproject/SAI/pull/880), new switch attributes were added to retrieve the temperature readings from the ASIC's internal sensors. 

This is a preliminary commit (pending SAI support from vendors) for temperature monitoring. The max, average and the entire list of temperatures are now added to the flex counters DB so that platform sensors scripts may query and use these values in thermal control algorithms.

**What I did**

1. Created a SWITCH_SENSORS_FLEX_GROUP to track sensors in the ASIC
2. Added SAI_SWITCH_ATTR_MAX_TEMP, SAI_SWITCH_ATTR_AVERAGE_TEMP, SAI_SWITCH_ATTR_TEMP_LIST to the SWITCH sensors flex counters.

**Why I did it**

To provide a mechanism for platform sensor scripts to have a generic way of querying the ASIC sensors via SAI.

**How I verified it**

Currently checked using stubs. Awaiting SAI implementation from vendors.

**Details if related**

### DIFF
--- a/orchagent/switchorch.cpp
+++ b/orchagent/switchorch.cpp
@@ -1,7 +1,10 @@
 #include <map>
+#include <tuple>
 
 #include "switchorch.h"
+#include "schema.h"
 #include "converter.h"
+#include "sai_serialize.h"
 #include "notifier.h"
 #include "notificationproducer.h"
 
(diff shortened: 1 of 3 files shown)


orchagent/switchorch.cpp
```diff
@@ -28,19 +31,76 @@ const map<string, sai_packet_action_t> packet_action_map =
...
+    }
+
+    catch (...)
+    {
+        SWSS_LOG_WARN("SWITCH Sensors flex counter groups was not set");
+    }
+
+    /* Add switch to flex_counter for updating sensors */
+    const auto id = sai_serialize_object_id(gSwitchId);
+    string key = getSwitchSensorsFlexCounterTableKey(id);
```
stcheng: `string key = SWITCH_SENSORS_FLEX_GROUP + id;`
i think it would be clearer

orchagent/switchorch.cpp
```diff
@@ -28,19 +31,76 @@ const map<string, sai_packet_action_t> packet_action_map =
...
+    std::string delimiter = "";
```
stcheng: remove the `std::`

PR #696, opened 2018-11-19
[aclorch]: Remove unnecessary warning message

stcheng:
Because AclOrch listens to all notifications from PortsOrch,
every time LAG member changes the change message will be sent to
AclOrch which is unncessary. This warning message is not wanted.


stcheng:
when LAG member changes, this line will be hit
it is expected and shall not be treated as a warning

stcheng:
@jipanyang for debugging purpose, it would be useful to add the logs in the observer.h file i think.

jipanyang:
@stcheng  Yes,  it might be helpful adding logs in observer.h.  Since I have not met any problem with observer related functions (yet), I don't have any strong opinion regarding the log.

PR #704, opened 2018-11-22
Move ports operational status initial sync in PortInitDone handling code

stepanblyschak:
**What I did**
Move ports operational status initial sync in PortInitDone handling code
Trow "runtime_error" instead of "const char*"

**Why I did it**
The original motivation was that in mlnx fast-fast boot solution orchagent starts in cold mode, so orchagent assumes port operational status is down, however on HW it could be up and SDK will not generate operational status update event.
Also, tried to unify the flow for warm, cold boot.

**How I verified it**
Run vs tests, installed swss debian package on DUT, verified oper status is in sync with HW

**Details if related**


lguohan:
retest this please

jipany:
During the restore phase of warm restart,  the data including port oper status should be restored from db, otherwise changing the oper status only may cause data inconsistencies and disturb data plane traffic.  Port oper status sync up will be done after data restore has finished.

jipanyang:
> @jipanyang , is this breaking warm reboot?

We had the discussion last week when syncd view comparison logic kept bringing down the port oper status during restore.  The original attempted fix is to get port oper status from ASIC then finish the restore, we realized that the asic port oper status may have changed compared with that saved before shutdown @qiluo-msft ,  port oper status change usually will cause changes in related objects like neighbor, router, lag and so on.  changing port oper status only will prevent the restore processing from reaching the pre-shutdown state. 

For warm reboot, we should perform state sync up only after restore (apply view).

stepanblyschak:
@jipanyang @qiluo-msft In my understanding, during warm start, port init done is handled before apply view, so sairedis will return oper status from restored ASIC DB(?), which should be the same as in restored APPL DB as I understand. At this point portsorch does not inform neighorch about oper status change.
After apply view refreshPortStatus will be called anyway to sync with real HW oper status and inform neighorch about oper status change.

VS tests passed on my vm, including warm restart tests.
I have no ability to verify system level warm reboot. Could you please test this with system level warm reboot?

jipanyang:
@https://github.com/stepanblyschak sairedis get operation always reaches asic to get latest data.  We do need to add virtual switch test cases to cover this scenario and other sad path scenarios like link flapping, lag/lag members down during restore phase.

lguohan:
@jipanyang , agree with you, can you add such vstest to prevent future regression? @stepanblyschak , can you make changes based on jipan's comments?

orchagent/portsorch.cpp
```diff
@@ -2932,12 +2889,19 @@ void PortsOrch::doTask(NotificationConsumer &consumer)
 
             SWSS_LOG_NOTICE("Get port state change notification id:%lx status:%d", id, status);
 
-            Port port;
-            if (!getPort(id, port))
```
qiluo-msft: >getPort [](start = 17, length = 7)

anything wrong with existing getPort? if any, fix inside?

orchagent/portsorch.cpp
```diff
@@ -2950,13 +2914,17 @@ void PortsOrch::updatePortOperStatus(Port &port, sai_port_oper_status_t status)
...
-        if (status == SAI_PORT_OPER_STATUS_UP || port.m_oper_status == SAI_PORT_OPER_STATUS_UP)
-        {
-            this->setHostIntfsOperStatus(port.m_port_id, status == SAI_PORT_OPER_STATUS_UP);
-        }
+        return ;
+    }
+
+    updateDbPortOperStatus(port, status);
+    bool isUp = status == SAI_PORT_OPER_STATUS_UP;
+    setHostIntfsOperStatus(port, isUp);
```
qiluo-msft: >setHostIntfsOperStatus [](start = 4, length = 22)

Don't ignore return value

qiluo-msft:
>             m_alias(alias), m_type(type) {};

Could you help remove this extra ';' ?

---
Refers to: orchagent/port.h:47 in 5f7a619. [](commit_id = 5f7a6192dc9c634e2606194a428b637a94fef605, deletion_comment = True)

qiluo-msft:
>     Port() {};

Could you help remove this extra ';' ?

---
Refers to: orchagent/port.h:45 in 5f7a619. [](commit_id = 5f7a6192dc9c634e2606194a428b637a94fef605, deletion_comment = True)

stcheng:
agree with @qiluo-msft on separating this pull request into smaller tasks

jipanyang:
@lguohan Yes, I could try to come up a vs test case for port oper change check during swss warm restart.

PR #705, opened 2018-11-22
Align default MTU value as SAI default

prsunny:
**What I did**
Set default MTU in orchagent to SAI default value (`1514`)

**Why I did it**
To fix issue https://github.com/Azure/sonic-buildimage/issues/2293

**How I verified it**

_**Before fix:**_

```
admin@str-a7060cx-acs-1:~$ show interfaces status 
  Interface            Lanes    Speed    MTU         Alias    Oper    Admin
-----------  ---------------  -------  -----  ------------  ------  -------
  Ethernet0            33,34      50G   9100   Ethernet1/1      up       up

admin@str-a7060cx-acs-1:~$ bcmcmd 'ps'
xe12( 34)  up     2   50G  FD   SW  No   Forward  TX RX   None   FA    KR2  **9412**    No 
```

**_After fix:_**

```
admin@str-a7060cx-acs-1:~$ bcmcmd 'ps xe12'
                 ena/        speed/ link auto    STP                  lrn  inter   max   cut   loop
           port  link  Lns   duplex scan neg?   state   pause  discrd ops   face frame  thru?  back
xe12( 34)  up     2   50G  FD   SW  No   Forward  TX RX   None   FA    KR2  **9122**    No 
```

**Details if related**

### DIFF
--- a/orchagent/port.h
+++ b/orchagent/port.h
@@ -11,7 +11,7 @@ extern "C" {
 #include <map>
 
 #define DEFAULT_PORT_VLAN_ID    1
-#define DEFAULT_MTU              9100
+#define DEFAULT_MTU             1492
 
 namespace swss {
 
--- a/orchagent/port.h
+++ b/orchagent/port.h
@@ -11,6 +11,11 @@ extern "C" {
 #include <map>
 
 #define DEFAULT_PORT_VLAN_ID    1
+/*
+ * Default MTU is derived from SAI_PORT_ATTR_MTU (1514)
+ * Orchagent adds extra 22 bytes for Ethernet header and FCS,
+ * hence setting to 1492 (1514 - 22)
+ */
 #define DEFAULT_MTU             1492
 
 namespace swss {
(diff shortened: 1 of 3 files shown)


orchagent/port.h
```diff
@@ -11,7 +11,7 @@ extern "C" {
 #include <map>
 
 #define DEFAULT_PORT_VLAN_ID    1
-#define DEFAULT_MTU              9100
+#define DEFAULT_MTU             1492
```
qiluo-msft: >1492 [](start = 32, length = 4)

Is it derived from SAI_PORT_ATTR_MTU? Can we use a const expression here? Or add some comment.

qiluo-msft: >DEFAULT_MTU [](start = 8, length = 11)

Also fix DEFAULT_MTU_STR if has the same bug?

qiluo-msft: >define [](start = 1, length = 6)

If you are fixing a bug, could you add a VS test?

lguohan:
retest this please

lguohan:
@prsunny , looks like existing mtu vstest fails.

orchagent/port.h
stcheng: that's not the same issue here i think. those are used to ensure configurations set to 9100.

stcheng: agree

prsunny: Prefer to keep it as before. Added comment for clarify.

prsunny: Yes, DEFAULT_MTU_STR is not the same as this.

prsunny: Had a discussion with Guohan, VS test case would be added in the next commit since it might take some time. This change is mainly to unblock the MTU test failures.

PR #706, opened 2018-11-24
Only update appDB port oper_status when it is empty, alway initialize…

jipanyang:
… Port SAI_HOSTIF_ATTR_OPER_STATUS based on oper_status value in appDB.

Add vs test for port oper_status handling validation during warm restart restore phase.

**What I did**

 Only update appDB port oper_status when it is empty, alway initialize Port SAI_HOSTIF_ATTR_OPER_STATUS based on oper_status value in appDB.
    
    Add vs test for port oper_status handling validation during warm restart restore phase.
    

**Why I did it**

VS test case for port oper_status handling at warm restart restore phase is missing.

**How I verified it**

VS test.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2288,17 +2288,21 @@ bool PortsOrch::initializePort(Port &p)
 
     /**
      * Create database port oper status as DOWN if attr missing
-     * This status will be updated when receiving port_oper_status_notification.
+     * This status will be updated upon receiving port_oper_status_notification.
      */
-    p.m_oper_status = SAI_PORT_OPER_STATUS_DOWN;
     if (operStatus == "up")
     {
         p.m_oper_status = SAI_PORT_OPER_STATUS_UP;
     }
     else if (operStatus.empty())
     {
+        p.m_oper_status = SAI_PORT_OPER_STATUS_DOWN;
         /* Fill oper_status in db with default value "down" */
         m_portTable->hset(p.m_alias, "oper_status", "down");
+    }
+    else
+    {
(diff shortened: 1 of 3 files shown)


qiluo-msft:
Is it a bug fix or an improvement? If a bug fix, could you explain the scenario in previous implementation?

orchagent/portsorch.cpp
```diff
@@ -2290,35 +2290,30 @@ bool PortsOrch::initializePort(Port &p)
...
-        p.m_oper_status = SAI_PORT_OPER_STATUS_DOWN;
+        p.m_oper_status = SAI_PORT_OPER_STATUS_UP;
     }
-    else
+    else if (operStatus.empty())
     {
-        p.m_oper_status = SAI_PORT_OPER_STATUS_UP;
+        /* Fill oper_status in db with default value "down" */
+        m_portTable->hset(p.m_alias, "oper_status", "down");
     }
```
qiluo-msft: Maybe more readable this way.
```
    if (operStatus == "up")
    {
        p.m_oper_status = SAI_PORT_OPER_STATUS_UP;
    }
    else if (operStatus.empty())
    {
        p.m_oper_status = SAI_PORT_OPER_STATUS_DOWN;
        /* Fill oper_status in db with default value "down" */
        m_portTable->hset(p.m_alias, "oper_status", "down");
    }
    else
    {
        p.m_oper_status = SAI_PORT_OPER_STATUS_DOWN;
    }
```

jipanyang: OK.

nikos-github:
I don't suppose you could add a check a few lines up to validate the return value from the addHostIntfs(p, p.m_alias, p.m_hif_id); call and return false while you're here.

PR #708, opened 2018-11-27
[aclorch]: Store control plane ACL tables in orchagent

stcheng:
To prevent control plane ACL rules from being stored in the pending
task set, store the control plane ACL table in orchagent first and
check if the ACL rule is associated with control plane ACL table and
ignore it.

The next step is to combine the data structure of control plane ACL
tables and data plane ACL tables to make the code more efficient.

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1051,10 +1051,14 @@ void AclRuleMirror::update(SubjectType type, void *cntx)
 
 bool AclTable::validate()
 {
-    // Control plane ACLs are handled by a separate process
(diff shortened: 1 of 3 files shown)


lguohan:
vstest?

stcheng:
yes I'll add vs test, but it does not make too much sense since the APP database and ASIC database do not change.

lguohan:
the vstest verifies no new acl rule in the asic db.

stcheng:
@lguohan updated the unit test

stcheng:
retest this please

PR #710, opened 2018-11-27
Revert "Align default MTU value as SAI default"

lguohan:
Reverts Azure/sonic-swss#705


lguohan:
should keep the default mtu to 9100.

PR #713, opened 2018-11-28
Align orchagent MTU value as SAI default

prsunny:
**What I did**
https://github.com/Azure/sonic-swss/pull/705 +  fix for LAG and VLAN router interface MTU

**Why I did it**
Fix for issue Azure/sonic-buildimage#2293

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1870,6 +1870,16 @@ void PortsOrch::doVlanTask(Consumer &consumer)
 
         if (op == SET_COMMAND)
         {
+            // Retrieve attributes
+            uint32_t mtu = 0;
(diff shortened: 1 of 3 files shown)


lguohan:
retest this please

orchagent/portsorch.cpp
```diff
@@ -1884,6 +1894,25 @@ void PortsOrch::doVlanTask(Consumer &consumer)
                 }
             }
 
+            // Process attributes
+            Port l;
```
stcheng: could you rename the variable `l` to `vlan`?

prsunny: Changed to vl

PR #716, opened 2018-11-30
[portsorch]: Change set port speed internal logic

volodymyrsamotiy:

**What I did**
Changed internal logic for setting speed in orchagent
**Why I did it**
It is required for Mellanox FFB implementation, during fastfast boot, on init stage, we need to always set the speed to HW because ```get speed``` returns operational speed and still need to pass config to HW
**How I verified it**
Verify set speed logic is working and no errors observed
**Details if related**
N/A

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1726,25 +1726,23 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         {
                             if (speed != current_speed)
                             {
-                                if (setPortAdminStatus(p.m_port_id, false))
-                                {
-                                    if (setPortSpeed(p.m_port_id, speed))
-                                    {
-                                        SWSS_LOG_NOTICE("Set port %s speed to %u", alias.c_str(), speed);
-                                    }
-                                    else
-                                    {
-                                        SWSS_LOG_ERROR("Failed to set port %s speed to %u", alias.c_str(), speed);
-                                        it++;
-                                        continue;
-                                    }
-                                }
-                                else
+                                if (!setPortAdminStatus(p.m_port_id, false))
                                 {
                                     SWSS_LOG_ERROR("Failed to set port admin status DOWN to set speed");
                                     it++;
                                     continue;
                                 }
+                            }
+
+                            if (setPortSpeed(p.m_port_id, speed))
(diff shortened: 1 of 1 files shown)


orchagent/portsorch.cpp
```diff
@@ -1726,26 +1726,24 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
-                                else
+                                if (!setPortAdminStatus(p.m_port_id, false))
                                 {
                                     SWSS_LOG_ERROR("Failed to set port admin status DOWN to set speed");
                                     it++;
                                     continue;
                                 }
                             }
+
+                            if (setPortSpeed(p.m_port_id, speed))
```
jipanyang: With the previous logic, port admin status has to be down in order to set port speed.  But here the port speed may be set unconditionally,  is it correct?

volodymyrsamotiy: Yes, that is correct. We need to set admin down and then up only ```if (speed != current_speed)```

jipanyang: "We need to set admin down and then up only if (speed != current_speed)",   is this mellanox specific requirement or generic to all ASICS?  Why does the admin status need to be down before changing speed?

My concern is that some asic might need admin status to be down whenever sai_port_api->set_port_attribute() for SAI_PORT_ATTR_SPEED is called,  but I didn't find any explicit statement in SAI spec.

volodymyrsamotiy: I am not sure about other platforms, for Mellanox when speed is changed, port needs to be set down and then up in order to apply new speed.
Anyway there is real reason to put port down and up during speed set only if it indeed changed, so new value should be applied.
Regarding SAI spec, yes, looks like there is no explicit statement for this.

volodymyrsamotiy:
retest this please

lguohan:
the real concern here is that if speed == cur_speed, previously we do not need to set the speed. Now if we set the speed, then it can cause link to flap.

volodymyrsamotiy:
@lguohan, could you please elaborate why it can cause link flap?
I didn't observe any link flap during testing (on Mellanox platform). Are you talking about different behavior on some other platforms?

PR #719, opened 2018-12-03
[portsorch] move ports operational status initial sync to PortInitDone handling code

stepanblyschak:
**What I did**
Move ports operational status sync to PortInitDone handling code.

in this approach oper status is synced on portInitDone. When there is a oper_status in DB it is restored as in DB, when there is no oper_status in DB (cold restart) - read oper status from HW

**Why I did it**
The original motivation was that in mlnx fast-fast boot solution orchagent starts in cold mode, so orchagent assumes port operational status is down, however on HW it could be up and SDK will not generate operational status update event.

**How I verified it**
Verified Mellanox fastfast boot flow & vs test

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1456,6 +1456,12 @@ void PortsOrch::doPortTask(Consumer &consumer)
              */
             if (!m_initDone)
             {
+
+                if (!portStatusInitialSync())
+                {
+                    throw runtime_error("Portsorch: failed to sync port oper state");
+                }
+
                 m_initDone = true;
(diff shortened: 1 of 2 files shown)


qiluo-msft:
Could you please resolve the merge conflict?

stepanblyschak:
> Could you please resolve the merge conflict?

Rebased on master

stepanblyschak:
Do not review. We are going to use orchagent warm restart logic in mellanox fastfast boot solution, so we will not face issue with oper status. We consider to close this PR.

PR #635, opened 2018-10-02
[intfmgrd]: Merge intfsyncd into intfmgrd

marian-pritsak:
Move intfsyncd functionality to intfmgrd, add VRF membership support.

Signed-off-by: Marian Pritsak <marianp@mellanox.com>

PR #720, opened 2018-12-04
[VS]: remove incorrect vlan interface create operation

jipanyang:

**What I did**
remove incorrect vlan interface create operation
**Why I did it**

Triggering intfmgrd error:
```
Dec  4 09:20:33.044836 8719a7e62ab4 ERR #intfmgrd: :- doIntfTask:126: Invalid key Vlan100
```
**How I verified it**
VS test
**Details if related**

### DIFF
--- a/tests/test_vnet.py
+++ b/tests/test_vnet.py
@@ -165,15 +165,6 @@ def create_vlan_interface(dvs, vlan_name, ifname, vnet_name, ipaddr):
 
     time.sleep(1)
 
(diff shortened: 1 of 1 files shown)


prsunny:
@jipanyang , this change is required for VNet and will be handled once https://github.com/Azure/sonic-swss/pull/635 is merged.

For current code, there is a check that skips this table, `if (keys.size() != 2)` with the error log. But this is expected for now.

PR #722, opened 2018-12-06
Optimize neigh restore process during warm-reboot

zhenggen-xu:
During the neigh restore process, we set neighbor entries to stale
instead of reachable to kernel. The subsuquent arp/ns packets will bring the
active neighbors to reachable. And stale entries will be expired after
stale timer (default 60 seconds) or kernel will probe and delete the stale
entires if garbage collection is activated at any point.

**What I did**
Optimize the neighbor restore process during the warm-reboot

**Why I did it**
This is to avoid the stale entries to stay in kernel/SONiC for unnecessarily too long.

**How I verified it**
vs tests

**Details if related**

### DIFF
--- a/tests/test_warm_reboot.py
+++ b/tests/test_warm_reboot.py
@@ -1562,10 +1562,13 @@ def test_system_warmreboot_neighbor_syncup(dvs, testlog):
     macs = []
     for i in range(8, 8+NUM_INTF):
         # set timeout to be the same as real HW
+        # set stale timer bigger to avoid testbed difference related timing issues.
         # set ip on server facing interfaces
         # bring servers' interface up, save the macs
         dvs.runcmd("sysctl -w net.ipv4.neigh.Ethernet{}.base_reachable_time_ms=1800000".format(i*4))
         dvs.runcmd("sysctl -w net.ipv6.neigh.Ethernet{}.base_reachable_time_ms=1800000".format(i*4))
+        dvs.runcmd("sysctl -w net.ipv4.neigh.Ethernet{}.gc_stale_time=180".format(i*4))
+        dvs.runcmd("sysctl -w net.ipv6.neigh.Ethernet{}.gc_stale_time=180".format(i*4))
         dvs.runcmd("ip addr flush dev Ethernet{}".format(i*4))
         dvs.runcmd("ifconfig Ethernet{} {}.0.0.1/24 up".format(i*4, i*4))
         dvs.runcmd("ip -6 addr add {}00::1/64 dev Ethernet{}".format(i*4,i*4))
--- a/tests/test_warm_reboot.py
+++ b/tests/test_warm_reboot.py
@@ -327,6 +327,22 @@ def check_kernel_reachable_v6_neigh_num(dvs, number):
     neigh_num = int(output.strip())
     assert neigh_num == number
 
+def check_kernel_stale_neigh_num(dvs, number):
+    (exitcode, output) = dvs.runcmd(['sh', '-c', "ip neigh show nud stale | grep -v 'dev lo' | wc -l"])
+    neigh_num = int(output.strip())
+    assert neigh_num == number
+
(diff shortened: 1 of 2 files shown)


tests/test_warm_reboot.py
```diff
@@ -1651,8 +1682,11 @@ def test_system_warmreboot_neighbor_syncup(dvs, testlog):
     # should finish the store within 10 seconds
     time.sleep(10)
 
-    check_kernel_reachable_v4_neigh_num(dvs, NUM_OF_NEIGHS)
-    check_kernel_reachable_v6_neigh_num(dvs, NUM_OF_NEIGHS)
+    check_kernel_reachable_v4_neigh_num(dvs, NUM_OF_NEIGHS/2)
+    check_kernel_reachable_v6_neigh_num(dvs, NUM_OF_NEIGHS/2)
+
+    check_kernel_stale_v4_neigh_num(dvs, NUM_OF_NEIGHS/2)
+    check_kernel_stale_v6_neigh_num(dvs, NUM_OF_NEIGHS/2)
```
jipanyang: With half of the neighbor entries in stale state, it seems del operations be triggered against appDB/asicDB?  
  ```
# check syslog and sairedis.rec file for activities
    check_syslog_for_neighbor_entry(dvs, marker, 0, 0, "ipv4")
     check_syslog_for_neighbor_entry(dvs, marker, 0, 0, "ipv6")
     (nadd, ndel) = dvs.CountSubscribedObjects(pubsub)
     assert nadd == 0
     assert ndel == 0
```

zhenggen-xu: No, "stale" means the entry was not removed from kernel.  only when the entry got expired/deleted, then we will see the netlink messages and trigger appDB/asicDB changes.

jipanyang: With https://github.com/Azure/sonic-swss/blob/master/neighsyncd/neighsyncd.cpp#L61 , only valid entries will be taken into account.  I guess the reconciliation logic will get rid of the stale entries?  That should be the expected behavior.

jipanyang: I see.  
To confirm my understanding of the behavior:  The original state for arp/nd neighbor injected to kernel is stale, there is arp/ns packet generated immediately for each entry, upon receiving reply from neighbor the state will move to reachable.  For entries without reply,  they stay in stale state until gc_stale_time times out.  It seems 60 seconds is large enough, but it was changed to 180 seconds in your test cases.  Did you see time taken longer than that for the operations of kernel entry injection and arp/ns packet sending?  I also saw you reduced the NUM_OF_NEIGHS to 128 from 512.

zhenggen-xu: A few things here:
1, 60 seconds stale timer is fine for vs docker on my machine.  However, for slow machines we don't know how long it takes for the process between neigh insert and the neigh verification logic in the test code.  The time took for neigh insertion and arp/ns packets was small (a few seconds) even for 1K entries, the time spent was more for the the neighbor verification logic, the sleep, and ping etc...in the testing code.  And to make the vs testing stable, I increased the stale timer. 

2, In terms of NUM_OF_NEIGHS,  to support higher number of neighbors at vs docker and have expected behavior, we need tune the kernel parameters "gc_thresh1/2/3" at host side of the vs docker.  
The thresholds will determine when the gc task to run,  the gc task could probe and delete the stale entries by kernel.  If gc didn't run and stale timer could kick in and expire the stale entries.
Also, if the number of neighs is bigger than the hardware limit (gc_thresh3), system is going to a crazy state (vs test will hang).
It worked fine on my setup for 512 and 1K entries after tuning them.  but to avoid doing the tuning on different machines with even different kernels(I am not sure how to tune them in the test code itself), I simply reduced the number for vs test cases

Anyways,  The critical time of the restore process is small.  The timing issues and behavior issue we are dealing here are pretty much relevant to vs docker test cases.  And if we need more entries to run on SONiC (more than 1K per family), we will have to tune those kernel variables as well on the box.

PR #723, opened 2018-12-06
If available, read port settings from Config DB first

andriymoroz-mlnx:

**What I did**
Implemented #2 from [this comment](https://github.com/Azure/sonic-buildimage/issues/2047#issuecomment-423681870) to the issue

**Why I did it**
Fixes issue with the custom port split reset after SONiC-to-SONiC update

**How I verified it**
Executed SONiC-to-SONiC update on one of setups where we use custom port split. Verified split survived.

**Details if related**

### DIFF
--- a/portsyncd/portsyncd.cpp
+++ b/portsyncd/portsyncd.cpp
@@ -41,7 +41,7 @@ void usage()
 }
 
 void handlePortConfigFile(ProducerStateTable &p, string file, bool warm);
-void handlePortConfigFromConfigDB(ProducerStateTable &p, DBConnector &cfgDb, bool warm);
+bool handlePortConfigFromConfigDB(ProducerStateTable &p, DBConnector &cfgDb, bool warm);
 void handleVlanIntfFile(string file);
 void handlePortConfig(ProducerStateTable &p, map<string, KeyOpFieldsValuesTuple> &port_cfg_map);
 void checkPortInitDone(DBConnector *appl_db);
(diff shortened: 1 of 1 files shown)


prsunny:
retest this please

portsyncd/portsyncd.cpp
```diff
@@ -92,11 +92,14 @@ int main(int argc, char **argv)
...
         {
-            handlePortConfigFile(p, port_config_file, warm);
-        } else {
-            handlePortConfigFromConfigDB(p, cfgDb, warm);
+            // if port config is missing in ConfigDB
+            // attempt to use port_config.ini
+            if (!port_config_file.empty())
+            {
+                handlePortConfigFile(p, port_config_file, warm);
+            }
```
zhenggen-xu: should we "throw" if port_config_file.empty() is true in this else branch?

andriymoroz-mlnx: I tried to keep the old logic as it was before

andriymoroz-mlnx:
retest this please

andriymoroz-mlnx:
retest this please

andriymoroz-mlnx:
@nikos-github 
since the initial configuration always comes from port_config.ini, ports will always have alias (ini file parser uses port name if alias not specified)
the only scenario which will be different after this change is when you intentionally remove aliases from the config_db.json. In that case after s2s upgrade ports config will remain without aliases (previously aliases were always recreated from ini file).

nikos-github:
@andriymoroz-mlnx Thank you. I understand how it works. I'm just requesting explicit testing of the various scenarios including the ones where the alias is explicitly specified but then it's not and is derived from the port name and vice versa.

andriymoroz-mlnx:
Firs time (e.g. install from ONIE) alias will be used from port_config.ini (or generated if not specified). It means DB will always have it.
Next boot will use DB (no matter s2s or reboot)
To achieve the case "then it's not" after s2s you need to use port_config.ini without aliases, remove config DB and regenerate it. In this case data from port_config.ini will be used which leads to the "first time case"
If this is not what you meant, could you please write steps?

nikos-github:
@andriymoroz-mlnx If I have a config_db.json file with an alias, where will the code look for the alias? Similarly if I have a config_db.json without an alias, will the code go looking in port_config.ini for one? If the answer to this is yes, then that's wrong.

andriymoroz-mlnx:
@nikos-github 
code always uses the DB.
port_config.ini is used only once on start and only when there is no ports info in the DB

PR #725, opened 2018-12-06
[WR]: Add reconciliation logic for teamsyncd

pavel-shirshov:
**What I did**
Added reconciliation logic for teamsyncd.

**Why I did it**
To avoid sending wrong updates to swss from teamsyncd.

**How I verified it**
Install changes to teamd docker. Restart dut in cold mode, then restart dut in wr mode.

**Details if related**


teamsyncd/teamsyncd.cpp
```diff
@@ -34,7 +42,13 @@ int main(int argc, char **argv)
             while (true)
             {
                 Selectable *temps;
-                s.select(&temps);
+                s.select(&temps, 1000); // check every second
+
+                auto diff = duration_cast<seconds>(steady_clock::now() - start_time);
+                if(diff.count() > WR_PENDING_TIME)
```
jipanyang: Like those for bgp and neighsyncd,   will making the pending time configurable be helpful?
```
root@sonic:~# config warm_restart 
Usage: config warm_restart [OPTIONS] COMMAND [ARGS]...

  warm_restart-related configuration tasks

Options:
  -s, --redis-unix-socket-path TEXT
                                  unix socket path for redis connection
  --help                          Show this message and exit.

Commands:
  bgp_timer
  disable
  enable
  neighsyncd_timer
```

teamsyncd/teamsyncd.cpp
```diff
@@ -34,7 +42,13 @@ int main(int argc, char **argv)
             while (true)
             {
                 Selectable *temps;
-                s.select(&temps);
+                s.select(&temps, 1000); // check every second
+
+                auto diff = duration_cast<seconds>(steady_clock::now() - start_time);
```
jipanyang: Maybe do time check within 
```
if(m_warmstart)
{
....
}
```

teamsyncd/teamsyncd.cpp
```diff
@@ -34,7 +42,13 @@ int main(int argc, char **argv)
...
+                if(diff.count() > WR_PENDING_TIME)
+                {
+                    sync.applyState();
```
qiluo-msft: >applyState [](start = 25, length = 10)

Do you need applyState every second after WR_PENDING_TIME? or just once?

teamsyncd/teamsync.cpp
```diff
@@ -90,7 +121,14 @@ void TeamSync::addLag(const string &lagName, int ifindex, bool admin_state,
     fvVector.clear();
     FieldValueTuple s("state", "ok");
     fvVector.push_back(s);
-    m_stateLagTable.set(lagName, fvVector);
+    if(m_warmstart)
```
qiluo-msft: >if [](start = 4, length = 2)

Add one blank after 'if'. Many times in this PR.

teamsyncd/teamsyncd.cpp
```diff
@@ -1,17 +1,23 @@
...
+#include <chrono>
 #include <team.h>
 #include "logger.h"
 #include "select.h"
 #include "netdispatcher.h"
 #include "netlink.h"
+#include "warm_restart.h"
 #include "teamsync.h"
 
+
```
qiluo-msft: Remove the extra line

teamsyncd/teamsyncd.cpp
qiluo-msft: >WR_PENDING_TIME [](start = 34, length = 15)

If we dynamically change WR_PENDING_TIME to 0 in cold reboot, you may have unified logic in teamsync

pavel-shirshov: Good idea. Done

teamsyncd/teamsyncd.cpp
pavel-shirshov: Refactored

teamsyncd/teamsyncd.cpp
pavel-shirshov: Refactored to make it more straight.

teamsyncd/teamsyncd.cpp
pavel-shirshov: currently it the time is 0, we use default timer

qiluo-msft: You code is good now. Actually I mean you can treat the cold reboot as a special warm reboot with a timer 0. So it just create view and apply view in 1 second.

pavel-shirshov: https://github.com/Azure/sonic-utilities/pull/409

pavel-shirshov:
The schema was added

teamsyncd/teamsync.h
```diff
@@ -11,19 +11,23 @@
 #include "netmsg.h"
 #include <team.h>
 
+// seconds
+const uint32_t DEFAULT_WR_PENDING_TIMEOUT = 70;
```
jipanyang: Would like to understand the logic behind this default value of 70 seconds here.

With lacp slow mode, 30 second interval * 3 times = 90 seconds,  all the teamd related processing has to be within that range from teamd shutdown to the lacp connection reestablished.   
During warm reboot, are you expecting it will take about 20 seconds from teamd shutdown to teamd restarted,  then from teamd started to lacp connected, it will take x*10 seconds?

qiluo-msft:
Retest this please

PR #727, opened 2018-12-12
[neighorch] return back original condition for port oper status

stepanblyschak:

**What I did**
Return back original condition for port operational status when adding a next hop.

**Why I did it**
port::m_oper_status is by default SAI_PORT_OPER_STATUS_UNKNOWN.
If port is LAG or VLAN m_oper_status is not overridden with value in state DB, so
(p.m_oper_status != SAI_PORT_OPER_STATUS_UP) is true and next hop flag IFDOWN is set.

This is workaround, need to listen to state db to update oper status in port struct

**How I verified it**
Built and verified that next hops are added to ecmp group

**Details if related**


prsunny:
retest this please

qiluo-msft:
Retest this please

qiluo-msft:
retest this please

PR #729, opened 2018-12-13
[watermark] fix watermark clear not working for PG

mykolaf:

**What I did**
Made the logic deciding which watermark to clear more deterministic. 
Add some more warning logs
**Why I did it**
There was a bug, PG would not clear
**How I verified it**
  show priority-group persistent-watermark shared
  sudo sonic-clear priority-group persistent-watermark shared
  show priority-group persistent-watermark shared

**Details if related**

### DIFF
--- a/orchagent/watermarkorch.cpp
+++ b/orchagent/watermarkorch.cpp
@@ -107,41 +107,50 @@ void WatermarkOrch::doTask(NotificationConsumer &consumer)
 
     consumer.pop(op, data, values);
 
-    if(&consumer == m_clearNotificationConsumer){
-        string wm_name = "";
-        vector<sai_object_id_t> &obj_ids = m_pg_ids;
-        if(data == CLEAR_PG_HEADROOM_REQUEST)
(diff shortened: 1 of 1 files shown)


orchagent/watermarkorch.cpp
```diff
@@ -107,41 +107,50 @@ void WatermarkOrch::doTask(NotificationConsumer &consumer)
 
     consumer.pop(op, data, values);
 
-    if(&consumer == m_clearNotificationConsumer){
-        string wm_name = "";
-        vector<sai_object_id_t> &obj_ids = m_pg_ids;
```
wendani: The root cause of the bug is the use of reference, and rebind it (obj_ids = m_pg_ids; obj_ids = m_unicast_queue_ids;)?

mykolaf: Yes, the clearSingleWm() was called with wrong vector of oids.

mykolaf:
retest this please

PR #736, opened 2018-12-21
 Vlan status enhancement

jihaix:
* Make Vlanxx's oper_status be relavent to its admin_status and member ports' oper_status.

Signed-off-by: xiongjihai@baidu.com

### DIFF
--- a/cfgmgr/vlanmgr.cpp
+++ b/cfgmgr/vlanmgr.cpp
@@ -29,6 +29,7 @@ VlanMgr::VlanMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
         m_stateVlanTable(stateDb, STATE_VLAN_TABLE_NAME),
         m_stateVlanMemberTable(stateDb, STATE_VLAN_MEMBER_TABLE_NAME),
         m_appVlanTableProducer(appDb, APP_VLAN_TABLE_NAME),
+        m_appVlanTable(appDb, APP_VLAN_TABLE_NAME),
         m_appVlanMemberTableProducer(appDb, APP_VLAN_MEMBER_TABLE_NAME)
 {
     SWSS_LOG_ENTER();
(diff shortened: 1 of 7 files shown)


cfgmgr/vlanmgr.h
```diff
@@ -31,6 +34,10 @@ class VlanMgr : public Orch
 
     bool addHostVlan(int vlan_id);
     bool removeHostVlan(int vlan_id);
+
+	bool updateVlanStatusInit(vector<FieldValueTuple> &fvVector, const int vlan_id);
```
prsunny: This doesn't seem to be aligned

orchagent/portsorch.cpp
```diff
@@ -261,7 +266,7 @@ PortsOrch::PortsOrch(DBConnector *db, vector<table_name_with_pri_t> &tableNames)
         if (status != SAI_STATUS_SUCCESS)
         {
             SWSS_LOG_ERROR("Failed to get hardware lane list pid:%lx", port_list[i]);
-            throw runtime_error("PortsOrch initialization failure");
+            throw "PortsOrch initialization failure";
```
prsunny: This is a new change. You may want to resolve conflicts

cfgmgr/vlanmgr.cpp
```diff
@@ -122,6 +123,94 @@ bool VlanMgr::removeHostVlan(int vlan_id)
     return true;
 }
 
+
+bool VlanMgr::updateVlanStatusInit(vector<FieldValueTuple> &fvVector, const int vlan_id)
+{
+    fvVector.push_back(make_pair("admin_status", "up"));
+    fvVector.push_back(make_pair("member_status", "down"));
+    fvVector.push_back(make_pair("real_status", "down"));
```
prsunny: why not "oper_status" instead of "real_status"

prsunny:
Can you explain in detail why is this required? AFAIK, there is no SAI attribute for this Vlan attribute. So which component would be requiring this and the action the agent is performing? 
Also, there are conflicts in this PR to be resolved.

orchagent/portsorch.cpp
```diff
@@ -2553,6 +2581,221 @@ bool PortsOrch::removeBridgePort(Port &port)
     return true;
 }
 
+bool PortsOrch::setHostVlanAdminState(int vlan_id, const string &admin_status)
```
stcheng: My understanding is that all the command line settings shall be moved into vlanmgrd instead of in the orchagent code.

orchagent/portsorch.cpp
```diff
@@ -2553,6 +2581,221 @@ bool PortsOrch::removeBridgePort(Port &port)
...
+        if (fvField(i) == "oper_status")
+        {
+            port_oper_status = fvValue(i);
+            break;
+        }
+    }
+    return port_oper_status;
+}
+
+string PortsOrch::updateLagStatus(Port &lag)
```
stcheng: why this function is called 'updateLagStatus'?

lguohan:
can you describe the problem you are trying to solve clearly?

cfgmgr/vlanmgr.h
caesarkof97: Thanks，this is a mistake and will be fixed

orchagent/portsorch.cpp
caesarkof97: OK，I'll resolve the conflicts

cfgmgr/vlanmgr.cpp
caesarkof97: thanks，I'll change "real_status" to "oper_status"

orchagent/portsorch.cpp
caesarkof97: member port of vlan may be type of Lag，this function update status of lag port by check all its memer port's oper_status

caesarkof97:
This is the design of vlan status enhancement
https://github.com/Azure/SONiC/pull/316/files

orchagent/portsorch.cpp
caesarkof97: This is design file
https://github.com/Azure/SONiC/pull/316/files

caesarkof97:
> Can you explain in detail why is this required? AFAIK, there is no SAI attribute for this Vlan attribute. So which component would be requiring this and the action the agent is performing?
> Also, there are conflicts in this PR to be resolved.

This is the design file

orchagent/portsorch.cpp
caesarkof97: OK, I'll move it all the command line settings into vlanmgrd in next version

caesarkof97:
> Can you explain in detail why is this required? AFAIK, there is no SAI attribute for this Vlan attribute. So which component would be requiring this and the action the agent is performing?
> Also, there are conflicts in this PR to be resolved.
This is current design file 
https://github.com/Azure/SONiC/pull/316/files

caesarkof97:
> can you describe the problem you are trying to solve clearly?

this is current design file
https://github.com/Azure/SONiC/pull/316/files

orchagent/portsorch.cpp
```diff
@@ -1168,36 +1173,55 @@ bool PortsOrch::setPortAutoNeg(sai_object_id_t id, int an)
     return true;
 }
 
-bool PortsOrch::setHostIntfsOperStatus(const Port& port, bool isUp) const
+bool PortsOrch::setHostIntfsOperStatus(sai_object_id_t port_id, bool up)
```
jipanyang: Why changing  from port to port_id ?    It was changed to port for efficiency.

orchagent/portsorch.cpp
```diff
@@ -2553,6 +2581,221 @@ bool PortsOrch::removeBridgePort(Port &port)
...
+    const std::string cmds = std::string("")
+      + IP_CMD + " link set " + VLAN_PREFIX + std::to_string(vlan_id) + " " + admin_status;
+
+    std::string res;
+    EXEC_WITH_ERROR_THROW(cmds, res);
+
+    return true;
+}
+
+string PortsOrch::getPortOperStatus(string port_alias)
```
jipanyang: Why  changing  "bool PortsOrch::getPortOperStatus(const Port& port, sai_port_oper_status_t& status) const"?

you return both boolean false and string,   could it pass build?

jihaix: This is a mistake，I'll fix it. Interestingly, it passed the build

orchagent/portsorch.cpp
jihaix: this is a new change， and I'll solve the conflict

jihaix: This is a new change， I'll solve the conflict

jihaix:
> Please update https://github.com/Azure/sonic-swss/blob/master/doc/swss-schema.md for vlan schema change.
> 
> the vlan oper status are updated in both portmgr and orchagent in this design, how to avoid conflict?

I'll update the vlan schema change and move all the command line into vlanmgr

PR #740, opened 2019-01-02
Improve comments for neighbor warmrestart related functions 

zhenggen-xu:
Improve comments for neighsyncd functions and warmRestartAssist class

**What I did**
Improve comments for neighbor warmrestart related functions and warmRestartAssist class

**Why I did it**
To have better inline documents.

**How I verified it**
No real function changes in this PR.

**Details if related**


jipanyang:
For AppRestartAssist part,  would it be better to switch to ProducerStateTable temp view solution?

zhenggen-xu:
@jipanyang This PR is to improve comments for existing functions only,  other changes will be done later.

PR #741, opened 2019-01-09
[aclorch] Remove  L4 port range support limitation on egress ACL table and add new SWSS virtual test.

keboliu:
**What I did**
1. Remove the restriction of only support L4 port ACL range in ingress table 
   In function AclTable::create() there is code preventing support of ACL range in egress ACL table, should be removed.

3. New SWSS virtual test Added to cover the egress ACL support
  Dedicated test cases for egress ACL table creation/deletion and various egress ACL rule verification

more detail info can be found in this PR #https://github.com/Azure/SONiC/pull/322 which describe the change for the bug fix
**Why I did it**
To have an integrated ACL feature support.

**How I verified it**
Run sonic-mgmt ACL  test case to make sure exiting function not break, manually test the egress ACL  function to make sure it works.

**Details if related**


keboliu:
retest this please

keboliu:
@stcheng "stage" is not a new attribute to the table, it already there,  since if we don't explicitly set the stage, by default the table will be an ingress table, if we want to create an egress table, we have to explicitly set the stage to egress.  

regarding the new attribute to the AclRuleMirror class, this new attribute "m_tableStage" is derived from the table stage which associated with this mirror rule, with this new member, the AclRuleMirror can know it needs to set an ingress mirror action or an egress mirror action to SAI, because in SAI level there is different mirror action for ingress and egress.

keboliu:
@stcheng Thanks for your comments. For the second Item, I doubt whether the case you described is achievable. I consulted internally and the comments is that egress mirror performed at egress ACL rule,  ingress mirror performed at ingress ACL rule, so would you please also double check the feasibility?

keboliu:
@stcheng any update on Item 2?

liat-grozovik:
@keboliu in some ASIC vendor ingress mirroring can be on egress ACL. this is why @stcheng asked not to limit the solution for it.
Now the question is how we can know which vendor supports this option or not. Please check for each what action are supported and allow it. Please verify the get API for each egress/ingress if it supported the relevant action from SAI_ACL_TABLE_ATTR_ACL_ACTION_TYPE_LIST.
for example: SAI_ACL_ACTION_TYPE_MIRROR_EGRESS and SAI_ACL_ACTION_TYPE_MIRROR_INGRESS

stcheng:
Thanks @liat-grozovik for clarifying this. @keboliu could you first generate a separate pull request for the first item?

keboliu:
@stcheng @liat-grozovik  thanks for the clarification, I'll create a separate PR for Item1 first.

keboliu:
@stcheng  I have removed the code change related to Item2, only have Item1 left in this PR, please help to review.

PR #742, opened 2019-01-09
[intfmgrd]: Support loopback

marian-pritsak:
In #635, the intfmgrd does not read loopback address from config db at all and then it does not configure the loopback address in the asic. The PR fix the issue by reading from the config db. Since loopback address is configured via /etc/network/interface so it is skipped in intfmgrd. 

### DIFF
--- a/cfgmgr/intfmgr.cpp
+++ b/cfgmgr/intfmgr.cpp
@@ -13,6 +13,7 @@ using namespace swss;
 
 #define VLAN_PREFIX         "Vlan"
 #define LAG_PREFIX          "PortChannel"
+#define LOOPBACK_PREFIX     "Loopback"
 #define VNET_PREFIX         "Vnet"
 
 IntfMgr::IntfMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, const vector<string> &tableNames) :
@@ -98,6 +99,10 @@ bool IntfMgr::isIntfStateOk(const string &alias)
         SWSS_LOG_DEBUG("Port %s is ready", alias.c_str());
         return true;
     }
+    else if (!alias.compare(0, strlen(LOOPBACK_PREFIX), LOOPBACK_PREFIX))
+    {
+        return true;
+    }
 
     return false;
 }
@@ -110,6 +115,7 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
 
     string alias(keys[0]);
     string vrf_name = "";
+    bool is_lo = !alias.compare(0, strlen(LOOPBACK_PREFIX), LOOPBACK_PREFIX);
 
     for (auto idx : data)
     {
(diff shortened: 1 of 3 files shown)


marian-pritsak:
retest this please

cfgmgr/intfmgr.cpp
```diff
@@ -135,13 +141,29 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
             return false;
         }
 
-        setIntfVrf(alias, vrf_name);
-        m_appIntfTableProducer.set(alias, data);
+        // Set Interface VRF except for lo
+        if (!is_lo)
```
nikos-github: Why are you excluding the loopbacks?

cfgmgr/intfmgr.cpp
```diff
@@ -173,7 +196,11 @@ bool IntfMgr::doIntfAddrTask(const vector<string>& keys,
             return false;
         }
 
-        setIntfIp(alias, "add", ip_prefix.to_string(), ip_prefix.isV4());
+        // Set Interface IP except for lo
+        if (!is_lo)
```
nikos-github: Please explain how your code is dealing with multiple IP addresses on the loopback which is a valid config and must continue to work.

cfgmgr/intfmgr.cpp
```diff
@@ -186,7 +213,11 @@ bool IntfMgr::doIntfAddrTask(const vector<string>& keys,
     }
     else if (op == DEL_COMMAND)
     {
-        setIntfIp(alias, "del", ip_prefix.to_string(), ip_prefix.isV4());
+        // Set Interface IP except for lo
```
nikos-github: This set of changes require unit tests.

cfgmgr/intfmgr.cpp
marian-pritsak: Ip address for lo is configured in /etc/network/interfaces

cfgmgr/intfmgr.cpp
marian-pritsak: What's the issue with multiple IPs?
It's just different keys in APP DB

cfgmgr/intfmgr.cpp
nikos-github: Not sure I follow your response. It is valid for a loopback to be in a vrf.

cfgmgr/intfmgr.cpp
nikos-github: Today the loopback in sonic has more than one address. So with your changes post https://github.com/Azure/sonic-swss/pull/635, will all the loopback addresses make it to the asic and correctly? I'm just asking you to explain and also test.

jipanyang:
Is it expected to support incremental IP config for loopback interface?  

if yes,  lo config should be removed from /etc/network/interfaces and lo needs to be treated as other regular interface as to address config.

nikos-github:
@jipanyang Ideally it should be treated like any other interface when it comes to its incremental config, etc. however the loopback interface is kind of special since it's the anchoring point for 127.0.0.1. So I'm not sure it's a simple yes/no answer to your question but certainly today's loopback may have more than one address.

yxieca:
This change indeed solves the issue where ptf cannot ping lo interface.

cfgmgr/intfmgr.cpp
lguohan: @nikos-github , your comment is valid, but currently we do not support to have loopback in the vrf. We can add it in the future PR.

lguohan:
in long run, we should create lo interface and assign IP address to it, so that we can move those interfaces to the vrf, but current we do not support this.

prsunny:
@marian-pritsak , I don't think we need to check for `if (!is_lo) ` in code. For `doIntfGeneralTask`, it is invoked only if the interface is configured in config_db as part of VRF which is currently not the case. For `doIntfAddrTask`, it will be applying the same IP address again which can produce an error log. If you need to skip this error, its fine to have the check but I feel `doIntfGeneralTask `doesn't need the check. What do you think?

yxieca:
This change is required by warm reboot test. Let's keep the discussion offline and make changes accordingly.

PR #744, opened 2019-01-10
Revert PR#718 and PR#727

yxieca:
**What I did**
Reverting PR #718 and PR #727.

**Why I did it**
These 2 changes caused VLAN interface becoming ready for about 20 seconds later. This delay caused fast-reboot to fail, and triggered a failure in warm-reboot test.

**How I verified it**
After reverting the change, fast-reboot test is passing. And warm-reboot test don't see data plane disruption anymore.

### DIFF
--- a/orchagent/port.h
+++ b/orchagent/port.h
@@ -85,7 +85,7 @@ public:
     sai_object_id_t     m_ingress_acl_table_group_id = 0;
     sai_object_id_t     m_egress_acl_table_group_id = 0;
     vlan_members_t      m_vlan_members;
-    sai_port_oper_status_t m_oper_status = SAI_PORT_OPER_STATUS_UNKNOWN;
+    sai_port_oper_status_t m_oper_status;
     std::set<std::string> m_members;
     std::vector<sai_object_id_t> m_queue_ids;
     std::vector<sai_object_id_t> m_priority_group_ids;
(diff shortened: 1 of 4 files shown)


jipanyang:
What is the root cause of ready delay for VLAN interface?

yxieca:
retest this please

yxieca:
Not sure. I only isolated to these 2 changes. Didn't have time to look deeper yet.

stcheng:
If this PR https://github.com/Azure/sonic-swss/pull/748 is merged, can we avoid reverting?

orchagent/port.h
```diff
@@ -85,7 +85,7 @@ class Port
     sai_object_id_t     m_ingress_acl_table_group_id = 0;
     sai_object_id_t     m_egress_acl_table_group_id = 0;
     vlan_members_t      m_vlan_members;
-    sai_port_oper_status_t m_oper_status = SAI_PORT_OPER_STATUS_UNKNOWN;
+    sai_port_oper_status_t m_oper_status;
```
stepanblyschak: This brings back undefined behavior, e.g:
For LAG, VLAN m_oper_status field is never set, so there is a chance that NeighOrch will invalidate neighbor on LAG or VLAN in next hop group in case m_oper_status equals SAI_PORT_OPER_STATUS_DOWN.

PR #745, opened 2019-01-10
[neighsyncd] increase neighbor syncd restore timeout to 110 seconds

yxieca:
**What I did**
Neighbor syncd is restoring important information for teamd and BGP. therefore, our timeout should not be shorter than the down stream service.

**Why I did it**
If some interface becomes ready too slow, restore_neighbors script will timeout and leave some neighbors not restored. That will cause sonic to relearn these neighbors and re-create these neighbor entries. And thus that cause data plane disruption.

**How I verified it**
Without the change, when vlan interface becomes ready 20 seconds later, data plane disruption is noticed by test. With the change, there is no data plane disruption.

### DIFF
--- a/neighsyncd/neighsync.h
+++ b/neighsyncd/neighsync.h
@@ -11,10 +11,10 @@
 
 /*
  * This is the timer value (in seconds) that the neighsyncd waits for restore_neighbors
- * service to finish, should be longer than the restore_neighbors timeout value (120)
+ * service to finish, should be longer than the restore_neighbors timeout value (110)
  * This should not happen, if happens, system is in a unknown state, we should exit.
  */
-#define RESTORE_NEIGH_WAIT_TIME_OUT 180
+#define RESTORE_NEIGH_WAIT_TIME_OUT 120
 
 namespace swss {
 
--- a/neighsyncd/neighsync.h
+++ b/neighsyncd/neighsync.h
@@ -11,10 +11,10 @@
 
 /*
  * This is the timer value (in seconds) that the neighsyncd waits for restore_neighbors
- * service to finish, should be longer than the restore_neighbors timeout value (60)
+ * service to finish, should be longer than the restore_neighbors timeout value (120)
  * This should not happen, if happens, system is in a unknown state, we should exit.
  */
-#define RESTORE_NEIGH_WAIT_TIME_OUT 70
+#define RESTORE_NEIGH_WAIT_TIME_OUT 130
 
 namespace swss {
 
--- a/neighsyncd/neighsync.h
+++ b/neighsyncd/neighsync.h
@@ -14,7 +14,7 @@
  * service to finish, should be longer than the restore_neighbors timeout value (120)
  * This should not happen, if happens, system is in a unknown state, we should exit.
  */
-#define RESTORE_NEIGH_WAIT_TIME_OUT 130
+#define RESTORE_NEIGH_WAIT_TIME_OUT 180
 
 namespace swss {
 
--- a/neighsyncd/restore_neighbors.py
+++ b/neighsyncd/restore_neighbors.py
@@ -35,7 +35,7 @@ logger.addHandler(logging.NullHandler())
 # and this process is started by supervisord in swss docker.
 # It would be good to keep that time consistent routing reconciliation time-out, here
 # we are upstream, we shouldn't give up before down stream.
-TIME_OUT = 120
+TIME_OUT = 110
 
 # every 5 seconds to check interfaces states
 CHECK_INTERVAL = 5
@@ -191,12 +191,12 @@ def set_statedb_neigh_restore_done():
 # The interfaces' states were checked in a loop with an interval (CHECK_INTERVAL)
 # The function will timeout in case interfaces' states never meet the condition
 # after some time (TIME_OUT).
-def restore_update_kernel_neighbors(intf_neigh_map):
+def restore_update_kernel_neighbors(intf_neigh_map, timeout=TIME_OUT):
     # create object for netlink calls to kernel
     ipclass = IPRoute()
     mtime = monotonic.time.time
     start_time = mtime()
-    while (mtime() - start_time) < TIME_OUT:
+    while (mtime() - start_time) < timeout:
         for intf, family_neigh_map in intf_neigh_map.items():
             # only try to restore to kernel when link is up
             if is_intf_oper_state_up(intf):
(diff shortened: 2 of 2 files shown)


neighsyncd/restore_neighbors.py
```diff
@@ -33,8 +33,9 @@
 # timeout the restore process in 1 min if not finished
 # This is mostly to wait for interfaces to be created and up after system warm-reboot
 # and this process is started by supervisord in swss docker.
-# It would be good to keep that time below routing reconciliation time-out.
-TIME_OUT = 60
+# It would be good to keep that time consistent routing reconciliation time-out, here
+# we are upstream, we shouldn't give up before down stream.
+TIME_OUT = 120
```
stepanblyschak: If neighbor restore should run when all interfaces becomeready can we subscribe to state db and listen untill all interfaces become ready instead of relying on the timer?

yxieca: @zhenggen-xu can you chime in?

zhenggen-xu: The code check the interfaces state in the kernel before we put neighbors onto the interface, there was a check interval, so not relying on timer if interface are in a right state.  It would be better to check the kernel state instead of statedb since we need to make sure IP addresses are configured and operationally up, which do not exist in stateDB today.

Anyway, the timer is required whether we are using polling mode or subscribe mode.   This is to make sure we don't stuck there forever, e,g, the port never come up due to physical down.

yxieca: I thought waiting for *all* interfaces could delay restoring some of the entries. Do we have the consideration of restore as soon as they ready thoughts here too?

zhenggen-xu: The restoring process was done per interface state.  But we only have one signal to tell neighsyncd that the restore is done. To have per interface signal would complicate the solution.

yxieca: I didn't meant to ask for per interface signal. I was saying that current approach is trying best to bring back neighbors as interfaces individually getting ready.

zhenggen-xu: right, that is the current behavior.

neighsyncd/neighsync.h
```diff
@@ -11,10 +11,10 @@
 
 /*
  * This is the timer value (in seconds) that the neighsyncd waits for restore_neighbors
- * service to finish, should be longer than the restore_neighbors timeout value (60)
+ * service to finish, should be longer than the restore_neighbors timeout value (120)
  * This should not happen, if happens, system is in a unknown state, we should exit.
  */
-#define RESTORE_NEIGH_WAIT_TIME_OUT 70
+#define RESTORE_NEIGH_WAIT_TIME_OUT 130
```
zhenggen-xu: This time-out is the last resort, we can put some worst case number here to avoid changes in the future.

neighsyncd/restore_neighbors.py
zhenggen-xu: @yxieca  In normal case, if the interface wasn't down or unconfigured during the warm reboot,  the neighbor should be restored right after the interfaces are ready ( a few seconds later), and it should usually be faster than the routing reconciliation since protocol takes time to converge.

The timer here is just to make sure we don't stuck if interfaces state was changed (down) during the warm reboot.  So we could wait as much as routing stack timer configured.
 
I think this timer should be slightly smaller than the routing timer, so when the routing reconciliation is done, it already has the reconciled neighbors in place at SONiC orchagent.  Set this value same or higher than routing timer means the neighbor could be reconciled after routing is reconciled, it should work just not optimal in the processing pipe of orchagent.

Since you are changing the value here,  do you think you can read the routing timer or if not configured use the default one, then set this accordingly here?

yxieca: Resolved other conversations. Leaving this one for final confirmation.

@zhenggen-xu  I made change according to our last conversation. Please take a look again.

Thanks,
Ying

neighsyncd/neighsync.h
```diff
@@ -11,10 +11,10 @@
 
 /*
  * This is the timer value (in seconds) that the neighsyncd waits for restore_neighbors
- * service to finish, should be longer than the restore_neighbors timeout value (60)
+ * service to finish, should be longer than the restore_neighbors timeout value (120)
  * This should not happen, if happens, system is in a unknown state, we should exit.
  */
-#define RESTORE_NEIGH_WAIT_TIME_OUT 70
+#define RESTORE_NEIGH_WAIT_TIME_OUT 180
```
qiluo-msft: >180 [](start = 36, length = 3)

Could you comment what is the range for valid value? This one is also magic to me.

neighsyncd/restore_neighbors.py
```diff
@@ -33,8 +33,9 @@
 # timeout the restore process in 1 min if not finished
 # This is mostly to wait for interfaces to be created and up after system warm-reboot
 # and this process is started by supervisord in swss docker.
-# It would be good to keep that time below routing reconciliation time-out.
-TIME_OUT = 60
+# It would be good to keep that time consistent routing reconciliation time-out, here
+# we are upstream, we shouldn't give up before down stream.
+TIME_OUT = 110
```
qiluo-msft: >110 [](start = 11, length = 3)

The same. Are they the same thing?

yxieca: The BGP timeout is 120 seconds, Zhenggen want this value to be slightly smaller than the BGP timeout. And he also requested to check if the bgp timeout is set in the state DB. I added the code below.

neighsyncd/neighsync.h
yxieca: This is also Zhenggen's request. Essentially, this value only need to be bigger than the restore_neighbor.py timeout. But since we are reading the potential DB settings, so he wants this value to be something big, like 180 seconds.

neighsyncd/restore_neighbors.py
```diff
@@ -239,6 +240,10 @@ def main():
     warmstart.initialize("neighsyncd", "swss")
     warmstart.checkWarmStart("neighsyncd", "swss", False)
 
+    timeout = warmstart.getWarmStartTimer("bgp", "bgp")
+    if timeout <= 0:
+        timeout = TIME_OUT
+
```
lguohan: the neighbor timeout should not be driven by bgp timeout as if we force neighbor restore to timeout, then we could end up with unresolved entries and once we do reconciliation, we will have delete the neighbor in the asic and disrupt the data plane.

yxieca:
Updated the PR according to the discussion in the warm reboot meeting:

- Restore_neighbors.py should timeout at 110 seconds.
- Neighbor syncd should timeout waiting for restore neighbors for 120 seconds (longer than 110 seconds).
- restore_neighbors.py should not read BGP reconciliation timeout.

neighsyncd/restore_neighbors.py
yxieca: Done.

neighsyncd/restore_neighbors.py
yxieca: Updated comments.

neighsyncd/neighsync.h
yxieca: Updated comments.

yxieca:
retest this please

yxieca:
retest this please

qiluo-msft:
Unblocked. Please fix the vs test.

Issue #750, opened 2019-01-11
acl_table_group created, but not removed,  memory and asic resource leak.


jipanyang:
acl_table_group is created upon acl table bind:
https://github.com/Azure/sonic-swss/blob/master/orchagent/portsorch.cpp#L769

but it never gets removed even when no one uses it.  May hit it in LAG scenario,  create/delete/create lag/LAG_ACL_TABLE .

```
        # create port channel
        self.create_port_channel(dvs, "PortChannel01")

        # create ACL table
        self.create_acl_table(dvs, "LAG_ACL_TABLE", "PortChannel01")

        time.sleep(1)

        # check ASIC table
        self.check_asic_table_existed(dvs)

        # remove ACL table
        self.remove_acl_table(dvs, "LAG_ACL_TABLE")

        # remove port channel
        self.remove_port_channel(dvs, "PortChannel01")


        # create ACL table
        self.create_acl_table(dvs, "LAG_ACL_TABLE", "PortChannel01")

        # create port channel
        self.create_port_channel(dvs, "PortChannel01")

```

jipanyang:
Fixed with https://github.com/Azure/sonic-swss/pull/751

PR #751, opened 2019-01-12
Remove AclTableGroup upon removal of port/lag/vlan

jipanyang:

**What I did**

Remove AclTableGroup upon removal of port/lag/vlan
**Why I did it**
Try to address issue https://github.com/Azure/sonic-swss/issues/750

**How I verified it**
VS

**Details if related**
based on current implementation, each acl table group will be used by one port only?

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -784,7 +784,7 @@ bool PortsOrch::bindAclTable(sai_object_id_t id, sai_object_id_t table_oid, sai_
 
         setPort(port.m_alias, port);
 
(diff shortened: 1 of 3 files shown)


orchagent/portsorch.cpp
```diff
@@ -1258,13 +1258,14 @@ bool PortsOrch::removePort(sai_object_id_t port_id)
         notify(SUBJECT_TYPE_PORT_CHANGE, static_cast<void *>(&update));
     }
 
+
```
stcheng: extra line

jipanyang:
@stcheng  right, if there is reference on ACL table group, the remove_acl_table_group() call will fail. That is also the reason that port/log/vlan have to be removed first,  the acl groupOid was set as an attribute on it. 

https://github.com/Azure/sonic-swss/blob/master/orchagent/aclorch.cpp#L1243   ACL table group member has been removed here.

yxieca:
Made to 201811 branch on 2/4/2019

PR #761, opened 2019-01-17
[vnetorch] Change logic for adding VNet interface

marian-pritsak:

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -163,6 +163,7 @@ bool IntfsOrch::setIntf(const string& alias, sai_object_id_t vrf_id, const IpPre
         addDirectedBroadcast(port, ip_prefix->getBroadcastIp());
     }
 
+    m_syncdIntfses[alias].ip_addresses.insert(*ip_prefix);
     return true;
 }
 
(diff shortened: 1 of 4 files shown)


marian-pritsak:
retest this please

prsunny:
retest this please

orchagent/intfsorch.cpp
```diff
@@ -225,66 +281,29 @@ void IntfsOrch::doTask(Consumer &consumer)
...
                     it++;
                     continue;
                 }
             }
-
-            vrf_id = port.m_vr_id;
-            if (!ip_prefix_in_key || m_syncdIntfses[alias].ip_addresses.count(ip_prefix))
-            {
-                /* Request to create router interface, no prefix present or Duplicate entry */
-                it = consumer.m_toSync.erase(it);
```
prsunny: This case is now differently handled by the new code. It just erased and continued but with new code, since the function is returning true,  it is inserted to `m_syncdIntfses` at line 307. This will insert a default `IpPrefix ` when there is no prefix present in key (VRF/VNET) case

marian-pritsak: You're correct
Fixed

prsunny:
retest this please

prsunny:
retest this please

PR #764, opened 2019-01-21
[portsorch] Don't flap port when setting speed if port is down

stepanblyschak:
**What I did**
When port is already down during speed configuration don't put port down again.
Cache admin status in Port structure, do not set admin port status again if already set.

**Why I did it**
Avoid unnessesary SAI calls. It also had implications on Mellanox fastfast boot solution causing port flapping.

**How I verified it**
Currently VS fails (Check https://github.com/Azure/sonic-sairedis/pull/419)
Verified it is passing all VS tests.
Mannual testing.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -542,6 +542,25 @@ bool PortsOrch::setPortAdminStatus(sai_object_id_t id, bool up)
     return true;
 }
 
+bool PortsOrch::getPortAdminStatus(sai_object_id_t id, bool &up)
+{
+    SWSS_LOG_ENTER();
+
+    sai_attribute_t attr;
+    attr.id = SAI_PORT_ATTR_ADMIN_STATE;
+
+    sai_status_t status = sai_port_api->get_port_attribute(id, 1, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to get admin status for port pid:%lx", id);
+        return false;
+    }
+
+    up = attr.value.booldata;
+
+    return true;
+}
+
 bool PortsOrch::setPortMtu(sai_object_id_t id, sai_uint32_t mtu)
 {
     SWSS_LOG_ENTER();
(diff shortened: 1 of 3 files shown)


orchagent/portsorch.cpp
```diff
@@ -1759,10 +1783,12 @@ void PortsOrch::doPortTask(Consumer &consumer)
                     }
                 }
 
-                if (!admin_status.empty())
+                if (!admin_status.empty() && p.m_admin_state != (admin_status == "up"))
                 {
                     if (setPortAdminStatus(p.m_port_id, admin_status == "up"))
                     {
+                        p.m_admin_state = admin_status == "up";
```
prsunny: Please use brackets to avoid ambiguity

orchagent/portsorch.cpp
```diff
@@ -1704,22 +1704,24 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
-                                    {
-                                        SWSS_LOG_NOTICE("Set port %s speed to %u", alias.c_str(), speed);
-                                    }
-                                    else
-                                    {
-                                        SWSS_LOG_ERROR("Failed to set port %s speed to %u", alias.c_str(), speed);
-                                        it++;
-                                        continue;
-                                    }
+                                    p.m_admin_state = false;
```
prsunny: Do we need to set this and copy ` p `and reset it later below since this is temporary status change?

orchagent/portsorch.cpp
```diff
@@ -1728,12 +1730,34 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
-                            continue;
+
+                            if (setPortSpeed(p.m_port_id, speed))
+                            {
+                                SWSS_LOG_NOTICE("Set port %s speed to %u", alias.c_str(), speed);
+                                p.m_speed = speed;
+                            }
+                            else
+                            {
+                                SWSS_LOG_ERROR("Failed to set port %s speed to %u", alias.c_str(), speed);
```
prsunny: Just thinking if we need to set the admin status to `old_admin_state_up` here if setPortSpeed failed.

stepanblyschak: Previous implementation relied on the fact that on each change the entire PORT table entry(speed, admin, fec etc.) is pushed to orchagent even if only one field changed(e.g. speed).
So, if port speed failed, on next iteration we will set admin status to correct DB status from DB.

orchagent/portsorch.cpp
stepanblyschak: Thanks, fixed

stepanblyschak:
retest this please

lguohan:
retest this please

stepanblyschak:
retest this please

liat-grozovik:
@prsunny , @yxieca this PR requires for Mellanox new SAI implementation. Please provide your feedback so we can merge this one and will be able to have new SAI with additional features required by SONiC.

yxieca:
Made to 201811 branch on 2/21/2019

PR #767, opened 2019-01-24
Incremental config add/del of IPv4/IPv6 addr on Loopback0 intf is not working

kirankella:
In intfmgr, add/del the addresses on "lo" interface too.

**What I did**
Intfmgr needs to push the address config into Linux for "lo"

**Why I did it**
The loopback addresses adding via config command are not pushed to Config DB, Linux Stack and to the Hardware

**How I verified it**

root@sonic:/home/admin# config interface Loopback0 ip add 28.28.28.28/32

root@sonic:/home/admin# ip route show table local | grep "dev 28.28.28.28"
local 28.28.28.28 dev lo proto kernel scope host src 28.28.28.28 

root@sonic:/home/admin# config interface Loopback0 ip add 2037::1/128

root@sonic:/home/admin# ip -6 route show | grep 2037::1
unreachable 2037::1 dev lo proto kernel metric 256  error -101 pref medium

root@sonic:/home/admin# show runningconfiguration all | grep Loopback0
        "Loopback0|28.28.28.28/32": {}, 
        "Loopback0|2037::1/128": {}, 

root@sonic:/home/admin# bcmcmd 'l3 defip show'
l3 defip show
Unit 0, Total Number of DEFIP entries: 8192
#     VRF     Net addr             Next Hop Mac        INTF MODID PORT PRIO CLASS HIT VLAN
.....
514   0        28.28.28.28/32       00:00:00:00:00:00 100003    0     0     0    1 n
.....

root@sonic:/home/admin# bcmcmd 'l3 ip6route show'
l3 ip6route show
Unit 0, Total Number of IPv6 entries: 6144 (IPv6/64 4096, IPv6/128 2048)
Max number of ECMP paths 64
Free IPv6 entries available: 6127 (IPv6/64 4087, IPv6/128 2040)
#     VRF     Net addr                                     Next Hop Mac        INTF MODID PORT PRIO CLASS HIT VLAN
....
7     0        2037:0000:0000:0000:0000:0000:0000:0001/128  00:00:00:00:00:00 100003    0     0     0    1 n
....

root@sonic:/home/admin# config interface Loopback0 ip remove 28.28.28.28/32
root@sonic:/home/admin# ip route show table local | grep "28.28.28.28"

root@sonic:/home/admin# config interface Loopback0 ip remove 2037::1/128
root@sonic:/home/admin# ip -6 route show | grep 2037::1
root@sonic:/home/admin# 

**Details if related**
Along with change in cfgmgr/intfmgr.cpp, this issue needed a change in config/main.py in sonic-utilities repo.

### DIFF
--- a/cfgmgr/intfmgr.cpp
+++ b/cfgmgr/intfmgr.cpp
@@ -196,10 +196,13 @@ bool IntfMgr::doIntfAddrTask(const vector<string>& keys,
             return false;
         }
 
-        // Set Interface IP except for lo
         if (!is_lo)
         {
             setIntfIp(alias, "add", ip_prefix.to_string(), ip_prefix.isV4());
+        }
+        else
+        {
+            setIntfIp("lo", "add", ip_prefix.to_string(), ip_prefix.isV4());
         }
 
         std::vector<FieldValueTuple> fvVector;
@@ -213,10 +216,13 @@ bool IntfMgr::doIntfAddrTask(const vector<string>& keys,
     }
     else if (op == DEL_COMMAND)
     {
-        // Set Interface IP except for lo
         if (!is_lo)
         {
             setIntfIp(alias, "del", ip_prefix.to_string(), ip_prefix.isV4());
+        }
+        else
+        {
+            setIntfIp("lo", "del", ip_prefix.to_string(), ip_prefix.isV4());
         }
         m_appIntfTableProducer.del(appKey);
         m_stateIntfTable.del(keys[0] + state_db_key_delimiter + keys[1]);


cfgmgr/intfmgr.cpp
```diff
@@ -196,11 +196,14 @@ bool IntfMgr::doIntfAddrTask(const vector<string>& keys,
...
 
-        // Set Interface IP except for lo
         if (!is_lo)
         {
             setIntfIp(alias, "add", ip_prefix.to_string(), ip_prefix.isV4());
         }
+        else
+        {
+            setIntfIp("lo", "add", ip_prefix.to_string(), ip_prefix.isV4());
+        }
```
jipanyang: Why not just remove "if (!is_lo)" ?

nikos-github:
@lguohan @jipanyang So the decision was made since https://github.com/Azure/sonic-swss/pull/742 was merged to support incremental loopback config? Would be nice to understand what the goal is here so that changes can be properly reviewed.

cfgmgr/intfmgr.cpp
kirankella: Initially I thought about that :-). But alias value is not 'lo'.
The 'alias' value for loopback interface is "Loopback0" as passed from the config command. The 'is_lo' boolean is set based by comparing the 'alias' name prefix with LOOPBACK_PREFIX (Loopback).
Hence the above change.

kirankella:
> @lguohan @jipanyang So the decision was made since #742 was merged to support incremental loopback config? Would be nice to understand what the goal is here so that changes can be properly reviewed.
The goal of this PR is to support the incremental config of the lo addresses. As I understand, #742 was done to ensure that loopback addresses are going to asic (from config_db) and not for incremental config support. Right?
Any change in config_db for the loopback addresses are pushed to /etc/network/interfaces only via 'config reload', right? I see they are not pushed to /etc/network/interfaces file via 'config load'.

With this PR change in intfmgr.cpp, I agree we will see 'already exist' errors for the config_db loopback addresses (pushed via /etc/network/interfaces), when they are being configured again from doIntfAddrTask.

Hence to complete the support for the lo incremental config, may be we can:
- Either remove the lo support from /etc/network/interfaces, like Jipin mentioned in #742 (or)
- handle the 'already exists' errors for the addresses added in doIntfAddrTask (not just for lo, but for any interface)

kirankella:
> > @lguohan @jipanyang So the decision was made since #742 was merged to support incremental loopback config? Would be nice to understand what the goal is here so that changes can be properly reviewed.
> 
> The goal of this PR is to support the incremental config of the lo addresses. As I understand, #742 was done to ensure that loopback addresses are going to asic (from config_db) and not for incremental config support. Right?
> Any change in config_db for the loopback addresses are pushed to /etc/network/interfaces only via 'config reload', right? I see they are not pushed to /etc/network/interfaces file via 'config load'.
> 
> With this PR change in intfmgr.cpp, I agree we will see 'already exist' errors for the config_db loopback addresses (pushed via /etc/network/interfaces), when they are being configured again from doIntfAddrTask.
> 
> Hence to complete the support for the lo incremental config, may be we can:
> -- Either remove the lo support from /etc/network/interfaces, like Jipin mentioned in #742 (or)
> -- handle the 'already exists' errors for the addresses added in doIntfAddrTask (not just for lo, but for any interface)

@jipanyang @lguohan @nikos-github Like discussed above, only on bootup I see the "File exists" error when intfmgr is adding loopback addresses (as they were already added via /etc/network/interfaces during bootup). I believe the incremental config is good to have on LOOPBACK interface too for consistency purposes, please suggest how to go about this.
Either remove the LOOPBACK interface logic in ./files/image_config/interfaces/interfaces.j2 or ignore "File Exists" error in intfmgr::setIntfIp() only for "lo" interfaces.

PR #770, opened 2019-01-26
Do basic mutual exclusion check on vlan member and router interface c…

jipanyang:
…onfig for same port

**What I did**
Eventually cfgmgr should prevent such configuration being pushed to appDB.  Here just to do basic preventive check in orchagent. 

**Why I did it**
To avoid wrong configuration from messing up orchagent state handling.

**How I verified it**
VS
**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2068,7 +2068,7 @@ void PortsOrch::doVlanMemberTask(Consumer &consumer)
                 addVlanMember(vlan, port, tagging_mode);
             }
             /*
-             * Failure in addBridgePort and addVlanMember is no-recoverable.
+             * Failure in addBridgePort and addVlanMember is non-recoverable.
              * Erase the vlan member task in both success and failure scenarios.
              */
             it = consumer.m_toSync.erase(it);
--- a/tests/test_interface.py
+++ b/tests/test_interface.py
@@ -256,6 +256,17 @@ class TestLagRouterInterfaceIpv4(object):
         tbl.set(interface, fvs)
         time.sleep(1)
 
+    # function to check vlan member and router interface mutual exclusion validation written in syslog
+    def check_syslog_for_rif_vlan_mutex_error(self, dvs, marker, alias, mutextype):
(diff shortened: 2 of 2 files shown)


orchagent/intfsorch.cpp
```diff
@@ -282,6 +282,13 @@ void IntfsOrch::doTask(Consumer &consumer)
                 continue;
             }
 
+            if (port.m_bridge_port_id != SAI_NULL_OBJECT_ID)
+            {
+                SWSS_LOG_ERROR("Router Intfs config on vlan member %s is not allowed", alias.c_str());
+                it = consumer.m_toSync.erase(it);
```
prsunny: I don't think we should block and discard the configuration. Instead, why don't we retry here?

orchagent/portsorch.cpp
```diff
@@ -1972,6 +1972,13 @@ void PortsOrch::doVlanMemberTask(Consumer &consumer)
             continue;
         }
 
+        if (port.m_rif_id)
+        {
+            SWSS_LOG_ERROR("Vlan config on router interface %s is not allowed", port.m_alias.c_str());
```
prsunny: Same as above.

orchagent/intfsorch.cpp
jipanyang: The concern is that performing retry here will introduce too many error logs and extra system overhead if the error is persistent for a long time. While lowering the log level will leave the error unnoticed most likely.

prsunny (dismissed):
Config validation handles this scenario. Approving as this can temporarily resolve the transient issues.

lguohan:
it is strange that this PR changes vnet test cases.

lguohan:
The test case should be added to check the syslog error message for invalid configuration. I do not see such test cases added.

jipanyang:
> it is strange that this PR changes vnet test cases.

Before the change "100.102.1.1/24" was configured on  "Ethernet4" first,   then the second test case tries to add  Ethernet4 to Vlan1002.  It hit the exact problem of one port being put in both rif and vlan.

lguohan:
retest this please

PR #771, opened 2019-01-26
On a routing vlan, the neighbor entry in the /31 subnet is not added to hardware

kirankella:
**What I did**
Don't add broadcast ip specific neighbor entry for the /31 subnet routing vlan.

**Why I did it**
In intfsorch code, it is creating neighbor entry for net directed broadcast address for a Routing VLAN when the routing VLAN interface is created. 
In case of /31 subnets there is no broadcast IP as (last 32nd bit) .0 and .1 corresponding to the only 2 hosts in the subnet. When adding neighbor entry for the .1 host, it is conflicting with the neighbor entry created for the directed broadcast IP and hence it throws the below error. As a result the .1 neighbor entry is not added to the hardware. 

Dec 8 19:15:55.732017 sonic ERR swss#orchagent: :- meta_sai_validate_neighbor_entry: object key SAI_OBJECT_TYPE_NEIGHBOR_ENTRY:{"ip":"12.12.1.1","rif":"oid:0x6000000000606","switch_id":"oid:0x21000000000000"} already exists 
Dec 8 19:15:55.732075 sonic DEBUG swss#orchagent: :< meta_sai_validate_neighbor_entry: exit 
Dec 8 19:15:55.732131 sonic DEBUG swss#orchagent: :< meta_sai_create_neighbor_entry: exit 
Dec 8 19:15:55.732189 sonic DEBUG swss#orchagent: :< redis_create_neighbor_entry: exit 
Dec 8 19:15:55.732245 sonic ERR swss#orchagent: :- addNeighbor: Failed to create neighbor 00:0c:9f:01:02:03 on Vlan900, rv:-6 

**How I verified it**
Have back to back nodes, and configure /31 subnet on both (for eg., 12.1.1.0/31 and 12.1.1.1/31). Ping the /31 peer address (12.1.1.1)
Check that the host entry for 12.1.1.1 is added in the hardware host table.

**Details if related**

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -158,7 +158,7 @@ bool IntfsOrch::setIntf(const string& alias, sai_object_id_t vrf_id, const IpPre
     addSubnetRoute(port, *ip_prefix);
     addIp2MeRoute(vrf_id, *ip_prefix);
 
-    if (port.m_type == Port::VLAN && ip_prefix->isV4())
+    if (port.m_type == Port::VLAN)
     {
         addDirectedBroadcast(port, *ip_prefix);
     }
@@ -345,7 +345,7 @@ void IntfsOrch::doTask(Consumer &consumer)
                     removeSubnetRoute(port, ip_prefix);
                     removeIp2MeRoute(vrf_id, ip_prefix);
 
-                    if(port.m_type == Port::VLAN && ip_prefix.isV4())
+                    if(port.m_type == Port::VLAN)
                     {
                         removeDirectedBroadcast(port, ip_prefix);
                     }
@@ -630,9 +630,9 @@ void IntfsOrch::addDirectedBroadcast(const Port &port, const IpPrefix &ip_prefix
     sai_neighbor_entry_t neighbor_entry;
     IpAddress ip_addr;
 
-    /* For /31 v4 subnets, there is no broadcast address, hence don't
+    /* For /31 and /32 v4 subnets, there is no broadcast address, hence don't
      * add a broadcast route. */
-    if (ip_prefix.getMaskLength() == 31)
+    if ((ip_prefix.isV4()) && (ip_prefix.getMaskLength() > 30))
     {
       return;
     }
@@ -663,8 +663,8 @@ void IntfsOrch::removeDirectedBroadcast(const Port &port, const IpPrefix &ip_pre
     sai_neighbor_entry_t neighbor_entry;
     IpAddress ip_addr;
 
-    /* For /31 v4 subnets, there is no broadcast address */
-    if (ip_prefix.getMaskLength() == 31)
+    /* For /31 and /32 v4 subnets, there is no broadcast address */
+    if ((ip_prefix.isV4()) && (ip_prefix.getMaskLength() > 30))
     {
         return;
     }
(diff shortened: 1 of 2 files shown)


orchagent/intfsorch.cpp
```diff
@@ -160,7 +160,12 @@ bool IntfsOrch::setIntf(const string& alias, sai_object_id_t vrf_id, const IpPre
 
     if (port.m_type == Port::VLAN && ip_prefix->isV4())
     {
-        addDirectedBroadcast(port, ip_prefix->getBroadcastIp());
+        /* For /31 IPv4 subnets, there is no broadcast address, hence don't
+         * add a net directed broadcast route specific neighbor entry */
+        if (ip_prefix->getMaskLength() != 31)
+        {
+            addDirectedBroadcast(port, ip_prefix->getBroadcastIp());
```
nikos-github: We need to fix getBroadcastIp() and have a check inside addDirectedBroadcast(). We shouldn't be explicitly checking for mask lengths in the code. We just need the APIs to do the right thing.

kirankella: I agree in general of APIs doing the right thing. Pushing the /31 checking logic into addDirectedBroadcast(). The getBroadcastIp() just returns broadcast IP irrespective of subnet size. So, i don't think it needs to be changed here.

kirankella: Changed with review comments incorporated are submitted.

nikos-github:
Thank you for fixing this. Personally I would also put 2 more checks to make this stricter. One that the address is ipv4 and the second one that we return for masks >30 (to include both /31 and /32 just in case). I am approving though and up to you if you want to safeguard this further.

kirankella:
> Thank you for fixing this. Personally I would also put 2 more checks to make this stricter inside add/remove DirectedBroadcast functions. One that the address is ipv4 and the second one that we return for masks >30 (to include both /31 and /32 just in case). I am approving though and up to you if you want to safeguard this further.

Changes done. Thanks.

orchagent/intfsorch.cpp
```diff
@@ -158,9 +158,9 @@ bool IntfsOrch::setIntf(const string& alias, sai_object_id_t vrf_id, const IpPre
     addSubnetRoute(port, *ip_prefix);
     addIp2MeRoute(vrf_id, *ip_prefix);
 
-    if (port.m_type == Port::VLAN && ip_prefix->isV4())
```
prsunny: Why this change?. Subnet broadcasting is not supported for IPv6.

orchagent/intfsorch.cpp
```diff
@@ -623,10 +624,20 @@ void IntfsOrch::removeIp2MeRoute(sai_object_id_t vrf_id, const IpPrefix &ip_pref
...
-void IntfsOrch::addDirectedBroadcast(const Port &port, const IpAddress &ip_addr)
+void IntfsOrch::addDirectedBroadcast(const Port &port, const IpPrefix &ip_prefix)
 {
     sai_status_t status;
     sai_neighbor_entry_t neighbor_entry;
+    IpAddress ip_addr;
+
+    /* For /31 and /32 v4 subnets, there is no broadcast address, hence don't
+     * add a broadcast route. */
+    if ((ip_prefix.isV4()) && (ip_prefix.getMaskLength() > 30))
```
prsunny: Another option is to modify this check as` if (!(ip_prefix.isV4()) || (ip_prefix.getMaskLength() > 30))` and edit the comment!.

kirankella: Done.

PR #777, opened 2019-01-31
Increase the watermark polling interval to 10s

wendani:

**What I did**

**Why I did it**
Observe high CPU utilization when watermark polling is enabled with 1s polling interval. Lower the polling resolution to 10s. May reduce the interval later if there will be efforts to improve the watermark read and clear path.

**How I verified it**

**Details if related**


prsunny:
Suggest to provide more details in description especially here for "why I did this" since you now change from 1 sec to 10 sec and what issues were observed with 1sec. This would be good for any future reference

lguohan:
retest this please

PR #779, opened 2019-02-01
[vstest]: fix test_port_an_warm.py test

lguohan:
the test should set the speed and autoneg in config db instead
of application db. As if the speed in config db is not changed,
it will override the value in app db after warm reboot

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


lguohan:
@jipanyang to review

lguohan:
test_speed.py fix in #780

PR #781, opened 2019-02-04
[watermarkorch] only perform periodic clear if the polling is on

mykolaf:
No need to periodically clear watermarks if the polling is disabled.

**What I did**
Changes:
 - make watermarkorch aware of Flex Counter config
 - enable/disable periodic timer based on Flex Counter status of watermarks. 
**Why I did it**

**How I verified it**
```
 # config watermark telemetry interval 10

Apr 25 13:02:21.782115 sonic DEBUG swss#orchagent: :- doTask: Periodic watermark cleared by timer!
Apr 25 13:02:31.762604 sonic DEBUG swss#orchagent: :- doTask: Periodic watermark cleared by timer!
Apr 25 13:02:41.772739 sonic DEBUG swss#orchagent: :- doTask: Periodic watermark cleared by timer!

 # config watermark telemetry interval 15

Apr 25 13:02:51.756292 sonic DEBUG swss#orchagent: :- doTask: Periodic watermark cleared by timer!
Apr 25 13:03:06.765902 sonic DEBUG swss#orchagent: :- doTask: Periodic watermark cleared by timer!

 # counterpoll watermark disable

Apr 25 13:03:21.743069 sonic DEBUG swss#orchagent: :- doTask: Periodic watermark cleared by timer!

 # counterpoll watermark enable

Apr 25 13:07:31.556155 sonic DEBUG swss#orchagent: :- doTask: Periodic watermark cleared by timer!

 # counterpoll watermark enable
 # counterpoll watermark enable
 # counterpoll watermark enable

Apr 25 13:07:46.547644 sonic DEBUG swss#orchagent: :- doTask: Periodic watermark cleared by timer!
```
****

**Details if related**

### DIFF
--- a/orchagent/watermarkorch.cpp
+++ b/orchagent/watermarkorch.cpp
@@ -14,8 +14,8 @@
 extern PortsOrch *gPortsOrch;
 
 
-WatermarkOrch::WatermarkOrch(DBConnector *db, const string tableName):
-    Orch(db, tableName)
+WatermarkOrch::WatermarkOrch(DBConnector *db, const vector<string> &tables):
+    Orch(db, tables)
 {
     SWSS_LOG_ENTER();
 
@@ -36,7 +36,6 @@ WatermarkOrch::WatermarkOrch(DBConnector *db, const string tableName):
     m_telemetryTimer = new SelectableTimer(intervT);
     auto executorT = new ExecutableTimer(m_telemetryTimer, this, "WM_TELEMETRY_TIMER");
     Orch::addExecutor(executorT);
-    m_telemetryTimer->start();
 
     m_telemetryInterval = DEFAULT_TELEMETRY_INTERVAL;
 }
@@ -66,19 +65,13 @@ void WatermarkOrch::doTask(Consumer &consumer)
 
         if (op == SET_COMMAND)
         {
-            if (key == "TELEMETRY_INTERVAL")
+            if (consumer.getTableName() == CFG_WATERMARK_TABLE_NAME)
             {
-                for (std::pair<std::basic_string<char>, std::basic_string<char> > i: fvt)
-                {
-                    if (i.first == "interval")
-                    {
-                        m_telemetryInterval = to_uint<uint32_t>(i.second.c_str());
-                    }
-                    else
-                    {
-                        SWSS_LOG_WARN("Unsupported key: %s", i.first.c_str());
-                    }
-                }
+                handleWmConfigUpdate(key, fvt);
+            }
+            else if (consumer.getTableName() == CFG_FLEX_COUNTER_TABLE_NAME)
+            {
+                handleFcConfigUpdate(key, fvt);
             }
         }
         else if (op == DEL_COMMAND)
@@ -91,6 +84,44 @@ void WatermarkOrch::doTask(Consumer &consumer)
         }
 
         consumer.m_toSync.erase(it++);
+    }
+}
+
+void WatermarkOrch::handleWmConfigUpdate(const std::string &key, const std::vector<FieldValueTuple> &fvt)
+{
+    if (key == "TELEMETRY_INTERVAL")
+    {
+        for (std::pair<std::basic_string<char>, std::basic_string<char> > i: fvt)
+        {
+            if (i.first == "interval")
+            {
+                m_telemetryInterval = to_uint<uint32_t>(i.second.c_str());
+            }
+            else
+            {
+                SWSS_LOG_WARN("Unsupported key: %s", i.first.c_str());
+            }
+        }
+    }
+}
+
+void WatermarkOrch::handleFcConfigUpdate(const std::string &key, const std::vector<FieldValueTuple> &fvt)
+{
+    if (key == "QUEUE_WATERMARK" || key == "PG_WATERMARK")
+    {
+        for (std::pair<std::basic_string<char>, std::basic_string<char> > i: fvt)
+        {
+            if (i.first == "FLEX_COUNTER_STATUS" && i.second == "enable" && m_isTimerEnabled == false)
+            {
+                m_telemetryTimer->start();
+                m_isTimerEnabled = true;
+            }
+            else if (i.first == "FLEX_COUNTER_STATUS" && i.second == "disable" && m_isTimerEnabled == true)
+            {
+                /* no need to stop timer here - it won't be reset after end of this interval */
+                m_isTimerEnabled = false;
+            }
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


orchagent/watermarkorch.cpp
```diff
@@ -170,16 +201,20 @@ void WatermarkOrch::doTask(SelectableTimer &timer)
...
-        /* If the interval was changed */
-        auto intervT = timespec { .tv_sec = m_telemetryInterval , .tv_nsec = 0 };
-        m_telemetryTimer->setInterval(intervT);
-        m_telemetryTimer->reset();
+        /* Timer is only running when the watermark polling is on                         *
+         * if it is disabled we will not restart the timer at the end of current interval */
+        if (m_isTimerEnabled)
+        {
+            auto intervT = timespec { .tv_sec = m_telemetryInterval , .tv_nsec = 0 };
+            m_telemetryTimer->setInterval(intervT);
```
wendani: We do not need to set the firing interval each time. Only when there is a change from CONFIG_DB.

mykolaf:
retest this please

mykolaf:
retest this please

orchagent/watermarkorch.cpp
```diff
@@ -94,6 +87,44 @@ void WatermarkOrch::doTask(Consumer &consumer)
...
+    if (key == "QUEUE_WATERMARK" || key == "PG_WATERMARK")
+    {
+        for (std::pair<std::basic_string<char>, std::basic_string<char> > i: fvt)
+        {
+            if (i.first == "FLEX_COUNTER_STATUS" && i.second == "enable" && m_isTimerEnabled == false)
+            {
+                m_telemetryTimer->start();
+                m_isTimerEnabled = true;
+            }
+            else if (i.first == "FLEX_COUNTER_STATUS" && i.second == "disable" && m_isTimerEnabled == true)
```
wendani: If "FLEX_COUNTER_STATUS" of the first processed key, say PG_WATERMARK, is "enable", while the second "QUEUE_WATERMARK" is "disable", the telemetryTimer will not start.

wendani: @mykolaf can you fix the logic?

mykolaf:
retest this please

orchagent/watermarkorch.h
```diff
@@ -36,8 +36,14 @@ class WatermarkOrch : public Orch
...
 
+    bool isTimerEnabled()
+    {
+        return m_qWmEnabled || m_pgWmEnabled;
+    }
+
 private:
-    bool m_isTimerEnabled = false;
+    bool m_qWmEnabled = false;
+    bool m_pgWmEnabled = false;
```
wendani: Instead of defining one Boolean variable for each watermark category, we may choose to use one uint64_t variable, with each category being a bit set in it.

orchagent/watermarkorch.cpp
```diff
@@ -94,6 +85,52 @@ void WatermarkOrch::doTask(Consumer &consumer)
...
+void WatermarkOrch::handleWmConfigUpdate(const std::string &key, const std::vector<FieldValueTuple> &fvt)
+{
+    if (key == "TELEMETRY_INTERVAL")
+    {
+        for (std::pair<std::basic_string<char>, std::basic_string<char> > i: fvt)
+        {
+            if (i.first == "interval")
+            {
+                auto intervT = timespec { .tv_sec = to_uint<uint32_t>(i.second.c_str()) , .tv_nsec = 0 };
+                m_telemetryTimer->setInterval(intervT);
```
wendani: I just checked the implementation of setInterval(), https://github.com/Azure/sonic-swss-common/blob/master/common/selectabletimer.cpp#L63. It only changes the value of the itimerspec struct, not sync the value to the timer. If the timer is not armed, it will not start the timer. This is good as we may not want to start the timer, but only want to change the interval. If the timer is armed, we will need a call to timerfd_settime(), which is called by SelectableTimer::start() and SelectableTimer::reset(), to sync the actual change to the timer. This means the change will take effect at a later timer when the current timer expires and the SelectableTimer doTask handler is called. If we do not care about the interval to take immediate effect. This is also fine.

orchagent/watermarkorch.cpp
```diff
@@ -94,6 +85,52 @@ void WatermarkOrch::doTask(Consumer &consumer)
...
+                }
+                else if (key == "PG_WATERMARK")
+                {
+                    m_pgWmEnabled = (i.second == "enable");
+                }
+            }
+        }
+        if (isTimerEnabled())
+        {
+            m_telemetryTimer->start();
```
wendani: If the telemetry timer has been started, a call to SelectableTimer::start() will reset the timer settings. This resets the current time to expiration, causing the interval length, where an enable cli issued, to be different from others. We should refrain from calling SelectableTimer::start() if there is no change to the timer interval value itself.

orchagent/watermarkorch.cpp
```diff
@@ -170,16 +207,18 @@ void WatermarkOrch::doTask(SelectableTimer &timer)
```
wendani: For line 198 to 206, how can you be so sure that periodic timers kicks in first before user issues the clear watermark through cli?  That m_pg_ids, m_unicast_queue_ids, and m_multicast_queue_ids is already populated for the latter?

wendani:
This question is not on the PR per se. There are two sources that will set a TABLE in COUNTERS_DB, one from user/timer clear command and the other from the flex counter thread lus script execution. In this case, does redis provide a lock mechanism internally to ensure mutual exclusion on writing to a table? Also @qiluo-msft for insights

qiluo-msft:
@wendani if the Redis written by single transactions, they are mutual exclusive.

orchagent/watermarkorch.cpp
wendani: The idea of introducing m_pg_ids, m_unicast_queue_ids, and m_multicast_queue_ids , I guess, is to have a cached copy of the oids at hand to remove the need of further queries into obtaining these oids. The only opportunity/condition to update this cached copy is when a watermark clear command is issued by timer and the doTask() handler finds that the vector is empty. So we assume here that when this update opportunity comes, the pg, uni_queue, and multi_queue oids are generated complete and will not be subject to any further changes.

The sources of the oids comes from the COUNTERS_DB *MAP table, which is set when FlexCounterOrch receives any signal on "FLEX_COUNTER_STATUS" field to call the generate*Map() function. This signal in the current orchagent initialization sequence, I guess, comes for granted after the port initialization is done, with all pg and queue oids being settled. So we will get a list of pg and queue oids complete and ready at the time the cached oid copy still hits the emptiness condition.

orchagent/watermarkorch.cpp
mykolaf: The timer is expected to be changed after current interval has ended. This is stated in the design.

orchagent/watermarkorch.cpp
mykolaf: Added logic to avoid uneven intervals.

orchagent/watermarkorch.cpp
mykolaf: The main idea of having m_pg_ids, m_unicast_queue_ids, m_multicast_queue_ids is to avoid querying them every time, performing deserialization of oids.  This things are not expected to change.

orchagent/watermarkorch.h
mykolaf: Now status is stored as a uint8_t

orchagent/watermarkorch.cpp
wendani: OK, let's follow the design for now. Then, we need a call to sync change to the actual timer in the doTask(SelectableTimer &timer) if we see there is a change in the telemetry interval

wendani:
Synced offline with Qi. A single transaction is in terms of a hget call or a hset call.

The watermark lua script (https://github.com/Azure/sonic-swss/commit/b750a4b5e4b8531cc13c250c498a2e5c26ef16f3#diff-f8446e5c7632fc3a23e06c9f9419828c) updates a value in the typical 'read-modify-write' fashion as hget-compare-hset.
A user or a timer clear, either directly issued through CLI or delegating to WatermarkOrch via the APPL_DB event channel, runs in a different context from the syncd flex counter thread. If the user clears the value in between the hget and hset, we can have a race condition situation.

One thing we can do in the lua script is to have a value's hset call follow as close as possible to its hget call, rather than hgetting all three values and then hsetting all three values if any update is needed. This may helping reduce the time interval between hget and hset, at least not introducing any more delay between the two operations, and alleviate the problem to some extent.

mykolaf:
@wendani I can enhance the lua script in separate PR.

liat-grozovik:
@wendani the changes in the Lux script should go with the PR or a seperate PR? anything else from this PR is missing?

mykolaf:
@wendani  Just doublechecked, updated the PR message. Ready to merge

PR #787, opened 2019-02-08
Create egress ACL table group during the PFCWD stats list installment

wendani:

**What I did**
On the PFCWD stats list installment, create egress ACL table group for each port and bind to it.

**Why I did it**
In pfc watchdog warm-reboot, we find that brcm sai not supporting ACL table group port unbind yet. If pfc storm ever occurred on a port, egress ACL table group will be created and bound to the port, and persist the association with the port even after the storm is restored. In the situation that there is restored but not ongoing pfc storm before warm-reboot, egress table group will not be created in view after warm-reboot. Syncd comparison logic will see the difference, and remove the egress ACL table group object from the port, the underlying unbind operation of which is not supported in brcm sai yet.

We use the workaround solution that we created egress table group proactively for all ports under PFCWD's surveillance so that syncd will not see the difference post warm-reboot even if pfc storm ever occurs on a port.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -526,9 +526,9 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port,
                 sai_serialize_object_id(queueId));
     }
 
-    // Create egress table group for each port of pfcwd's interest
-    sai_object_id_t group_member_oid;
-    gPortsOrch->bindAclTable(port.m_port_id, SAI_NULL_OBJECT_ID, group_member_oid, ACL_STAGE_EGRESS);
+    // Create egress ACL table group for each port of pfcwd's interest
+    sai_object_id_t group_oid;
+    gPortsOrch->bindAclTableGroup(port.m_port_id, group_oid, ACL_STAGE_EGRESS);
 }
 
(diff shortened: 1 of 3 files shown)


orchagent/pfcwdorch.cpp
```diff
@@ -525,6 +525,10 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port,
                 PfcWdOrch<DropHandler, ForwardHandler>::getCountersTable(),
                 sai_serialize_object_id(queueId));
     }
+
+    // Create egress table group for each port of pfcwd's interest
+    sai_object_id_t group_member_oid;
+    gPortsOrch->bindAclTable(port.m_port_id, SAI_NULL_OBJECT_ID, group_member_oid, ACL_STAGE_EGRESS);
```
lguohan: this is not correct. you want to reuse bindAclTable function, but you return false if table_oid == NULL. 
You want to create an empty table group, the return should be success, not false. 

I suggest you to define a new function createAclTableGroup(), and call gPortsOrch->createAclTableGroup(port.m_port_id, ACL_STAGE_EGRESS)

Then, you rewrite bindAclTable to use createAclTableGroup.

yxieca:
Made to 201811 branch on 2019/2/21

PR #788, opened 2019-02-09
Potential fix for timing issue in warm_reboot's routing UT

rodnymolina:
Goal with this one is to reactivate the unit-tests written as part of the routing warm-reboot effort. There seems to be a timing issue behind the sporadic failure seen during the execution of Testcase-15. I'm adjusting the restart_timer value that i'm using in these testcases to see if that fixes the issue. Notice that i'm unable to reproduce this problem in my testbed, so i'm hoping to be able to run various jenkin's testing-cycles before merging this PR.

### DIFF
--- a/tests/test_warm_reboot.py
+++ b/tests/test_warm_reboot.py
@@ -967,8 +967,7 @@ def enable_warmrestart(dvs, db, app_name):
 #
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


yxieca:
Made to 201811 branch on 2/14/2019 (cherry-pick in case we need to pick future changes in warm reboot test).

Issue #734, opened 2018-12-20
try/catch exception handling in orchagent/main.cpp makes core dump analysis impossible


jipanyang:
Though the try/catch exception handling is supposed to make problem debugging easier  https://github.com/Azure/sonic-swss/blob/master/orchagent/main.cpp#L297

in real practice,  it makes the core file generate doesn't contain the first bug that caused the exception.
One example:
if for any reason the value string is empty,  https://github.com/Azure/sonic-swss/blob/master/orchagent/portsorch.cpp#L1369 
`unsigned long portCount = stoul(value);` will cause orchagent to crash, while  the core file generated will have stack trace like:
```
root@sonic:/etc/sonic/cores# gdb /usr/bin/orchagent orchagent.1479509602.63.core
GNU gdb (Debian 7.7.1+dfsg-5) 7.7.1
Copyright (C) 2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /usr/bin/orchagent...Reading symbols from /usr/lib/debug/.build-id/bd/a86378303f48ea8f616b916761092771482760.debug...done.
done.

warning: exec file is newer than core file.
[New LWP 63]
[New LWP 77]
[New LWP 78]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/usr/bin/orchagent -d /var/log/swss -b 8192 -m 00:05:64:30:73:c0 -l 00:05:64:30'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00000000004a90cc in std::list<Observer*, std::allocator<Observer*> >::remove (this=0x82d460, __value=@0x7ffe595c32e0: 0x841950)
    at /usr/include/c++/4.9/bits/list.tcc:290
[...]

PR #790, opened 2019-02-15
Remove try/catch for correct coredump file

jipanyang:

**What I did**

This is a try to work around the coredump create issue:  https://github.com/Azure/sonic-swss/issues/734

**Why I did it**

**How I verified it**
core dump with right stack trace created.

**Details if related**
Actually It might be caused by incorrect destruct of multiple orch ( inter-dependency, order and etc.), which triggered more segmentation during try/catch stack unwinding, then only the latest segfault got recorded. 

If people have idea about how to fix the possible orch destruct issue, that would be better.

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -275,24 +275,22 @@ int main(int argc, char **argv)
 
     auto orchDaemon = make_shared<OrchDaemon>(&appl_db, &config_db, &state_db);
 
+    if (!orchDaemon->init())
     {
-        if (!orchDaemon->init())
-        {
-            SWSS_LOG_ERROR("Failed to initialize orchstration daemon");
-            exit(EXIT_FAILURE);
-        }
-
-        /*
-        * In syncd view comparison solution, apply view has been sent
-        * immediately after restore is done
-        */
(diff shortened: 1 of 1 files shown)


prsunny:
@qiluo-msft has fixed an issue with stack trace during orchagent crash. Request @qiluo-msft to review this.

qiluo-msft:
Could you elaborate or give an example of below observation?
> Actually It might be caused by incorrect destruct of multiple orch ( inter-dependency, order and etc.), which triggered more segmentation during try/catch stack unwinding, then only the latest segfault got recorded.

I believe it is a bug and we should fix it.

qiluo-msft:
Can we define a function like `SWSS_LOG_STACK` and use it here?

orchagent/main.cpp
```diff
@@ -275,7 +275,6 @@ int main(int argc, char **argv)
 
     auto orchDaemon = make_shared<OrchDaemon>(&appl_db, &config_db, &state_db);
 
-    try
```
stcheng: could you also remove the `try` bracket and the extra indentation as well?

PR #773, opened 2019-01-30
[vnetorch]: Bitmap VNet implementation

marian-pritsak:
Signed-off-by: Marian Pritsak <marianp@mellanox.com>

PR #792, opened 2019-02-15
[vs]: Add vs test support for bitmap VNET

volodymyrsamotiy:

**What I did**
Added ```vs``` test support for bitmap VNET implementation.

**Why I did it**
It is done in order to have ```vs``` tests running for the  bitmap VNET implementation.

**How I verified it**
Executed all ```vs``` tests (including this new test) and verified that all passed.

**Details if related**
Depends on the following PRs:
- https://github.com/Azure/sonic-swss/pull/773
- https://github.com/Azure/sonic-buildimage/pull/2571
- https://github.com/Azure/sonic-sairedis/pull/424


tests/test_vnet_bitmap.py
```diff
@@ -0,0 +1,249 @@
...
+# It is needed in order to enable platform specific "orchagent" code for testing "bitmap" VNET implementation.
+DVS_FAKE_PLATFORM = "mellanox"
+
+
+'''
+Implements "check" APIs for the "bitmap" VNET feature.
+These APIs provides functionality to verify whether specified config is correcly applied to ASIC_DB.
+Such object should be passed to the test class, so it can use valid APIs to check whether config is applied.
+'''
+class VnetBitmapVxlanTunnel(object):
```
prsunny: Can we have this in the same `test_vnet.py` file, considering bitmap and vrf as two different implementation for same feature?.

prsunny:
There are some failures for the Vnet bitmap test, can you check?

volodymyrsamotiy:
There are failures because implementation of "bitmap VNET" is not merged yet and also it depends on some other PRs.
In PR description there is a list of PRs that should be merged first:
> Depends on the following PRs:

* #773
* [Azure/sonic-buildimage#2571](https://github.com/Azure/sonic-buildimage/pull/2571)
* [Azure/sonic-sairedis#424](https://github.com/Azure/sonic-sairedis/pull/424)

tests/test_vnet_bitmap.py
volodymyrsamotiy: We can't, because "vrf" and "bitmap" implementations are under "if platform == ..." condition in orchagent.
So, to enable that code we need to run "vs" container with "platform" environment variable set to different values.
"DVS" fixture, which creates container has scope set to "module", as a result container is started per each pytest module, that is why we need two different modules (files).

prsunny: The class definition? Its fine if you prefer to keep it in this file, but I was thinking more like if we add anything in future, we would need to modify two files with implementation

volodymyrsamotiy: So, what you suggest is to move ```VnetBitmapVxlanTunnel``` class implementation to common ```test_vnet.py``` file, so we have implementation in one file.
And as for the ```TestVnetBitmapOrch``` class, we can still keep it in this file as it is now.
Is my understanding correct?

prsunny: yes, that is my suggestion

volodymyrsamotiy: It makes sense, I already pushed required changes.

prsunny:
retest this please

liat-grozovik:
retest this please

liat-grozovik:
retest this please

volodymyrsamotiy:
retest this please

volodymyrsamotiy:
retest this please

volodymyrsamotiy:
retest this please

volodymyrsamotiy:
retest this please

prsunny:
retest this please

prsunny:
retest this please

PR #794, opened 2019-02-16
Survive pfc watchdog storm action across warm-reboot

wendani:
**What I did**

**Why I did it**
When the PFC storm is ongoing before warm-reboot is issued, make sure the ASIC after warm-reboot is in the consistent state where storm action remains installed as it was in the ASIC until the normal restore criteria runs properly and kicks in to take charge of the dictation.

**How I verified it**
Tested on brcm dut

**Details if related**
Prerequiste: PFC watchdog must be set enabled by default to start running automatically when SONiC boots up. https://github.com/Azure/sonic-swss/pull/795

Depend on: 
schema table name change: https://github.com/Azure/sonic-swss-common/pull/266
multi-fields hdel: https://github.com/Azure/sonic-swss-common/pull/267

Downstream changes:
In accordance with schema table name change: https://github.com/Azure/sonic-utilities/pull/475

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -59,28 +59,37 @@ void PfcWdOrch<DropHandler, ForwardHandler>::doTask(Consumer& consumer)
         return;
     }
 
-    auto it = consumer.m_toSync.begin();
-    while (it != consumer.m_toSync.end())
+    if ((consumer.getDbId() == CONFIG_DB) && (consumer.getTableName() == CFG_PFC_WD_TABLE_NAME))
     {
-        KeyOpFieldsValuesTuple t = it->second;
-
-        string key = kfvKey(t);
-        string op = kfvOp(t);
-
-        if (op == SET_COMMAND)
+        auto it = consumer.m_toSync.begin();
+        while (it != consumer.m_toSync.end())
         {
-            createEntry(key, kfvFieldsValues(t));
-        }
-        else if (op == DEL_COMMAND)
-        {
-            deleteEntry(key);
-        }
-        else
-        {
-            SWSS_LOG_ERROR("Unknown operation type %s\n", op.c_str());
+            KeyOpFieldsValuesTuple t = it->second;
+
+            string key = kfvKey(t);
+            string op = kfvOp(t);
+
+            if (op == SET_COMMAND)
+            {
+                createEntry(key, kfvFieldsValues(t));
+            }
+            else if (op == DEL_COMMAND)
+            {
+                deleteEntry(key);
+            }
+            else
+            {
+                SWSS_LOG_ERROR("Unknown operation type %s\n", op.c_str());
+            }
+
+            consumer.m_toSync.erase(it++);
         }
 
-        consumer.m_toSync.erase(it++);
+        if (consumer.m_toSync.empty())
+        {
+            m_entriesCreated = true;
+            SWSS_LOG_NOTICE("PFC watchdog all entries created");
+        }
     }
 }
 
@@ -202,11 +211,6 @@ void PfcWdOrch<DropHandler, ForwardHandler>::createEntry(const string& key,
                     return;
                 }
             }
-            else if (field == PFC_WD_IN_STORM)
-            {
-                SWSS_LOG_NOTICE("In-storm queues %s on port %s", value.c_str(), key.c_str());
-                queues = tokenize(value, comma);
-            }
             else
             {
                 SWSS_LOG_ERROR(
@@ -249,50 +253,6 @@ void PfcWdOrch<DropHandler, ForwardHandler>::createEntry(const string& key,
     }
 
     SWSS_LOG_NOTICE("Started PFC Watchdog on port %s", port.m_alias.c_str());
-
-    // Start PFC storm action on queues of port
-    if (!queues.empty())
-    {
-        for (const auto &q : queues)
-        {
-            int qIdx = -1;
-            try
-            {
-                qIdx = stoi(q);
-            }
-            catch (const std::invalid_argument &e)
-            {
-                SWSS_LOG_ERROR("Invalid argument %s to %s()", q.c_str(), e.what());
-                continue;
-            }
-            catch (const std::out_of_range &e)
-            {
-                SWSS_LOG_ERROR("Out of range argument %s to %s()", q.c_str(), e.what());
-                continue;
-            }
-            catch (const std::exception &e)
-            {
-                SWSS_LOG_ERROR("Invalid conversion to int from string %s: %s", q.c_str(), e.what());
-                continue;
-            }
-            catch (...)
-            {
-                SWSS_LOG_ERROR("Unknown exception caught in string to int conversion");
-                continue;
-            }
-
-            if ((qIdx < 0) || (static_cast<unsigned int>(qIdx) >= port.m_queue_ids.size()))
-            {
-                SWSS_LOG_ERROR("Invalid queue index %d on port %s", qIdx, key.c_str());
-                continue;
-            }
-            if (!startWdActionOnQueue(PFC_WD_IN_STORM, port.m_queue_ids[qIdx]))
-            {
-                SWSS_LOG_ERROR("Failed to start PFC watchdog %s event action on port %s queue %d", PFC_WD_IN_STORM, key.c_str(), qIdx);
-                continue;
-            }
-        }
-    }
 }
 
 template <typename DropHandler, typename ForwardHandler>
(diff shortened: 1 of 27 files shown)


orchagent/pfcwdorch.h
```diff
@@ -75,6 +80,12 @@ class PfcWdSwOrch: public PfcWdOrch<DropHandler, ForwardHandler>
     void createEntry(const string& key, const vector<FieldValueTuple>& data);
     virtual void doTask(SelectableTimer &timer);
     //XXX Add port/queue state change event handlers
+
+    virtual bool bake() override;
```
qiluo-msft: >override [](start = 24, length = 8)

If 'override', you can remove 'virtual'
ref: [https://stackoverflow.com/a/39932616/2514803](https://stackoverflow.com/a/39932616/2514803)

orchagent/pfcwdorch.h
```diff
@@ -75,6 +80,12 @@ class PfcWdSwOrch: public PfcWdOrch<DropHandler, ForwardHandler>
...
+
+protected:
+    virtual bool startWdActionOnQueue(const string &event, sai_object_id_t queueId) override;
```
qiluo-msft: >override [](start = 84, length = 8)

The same.

orchagent/pfcwdorch.cpp
```diff
@@ -239,6 +248,24 @@ void PfcWdOrch<DropHandler, ForwardHandler>::createEntry(const string& key,
     }
 
     SWSS_LOG_NOTICE("Started PFC Watchdog on port %s", port.m_alias.c_str());
+
+    // Start PFC storm action on queues of port
+    if (!queues.empty())
+    {
+        for (const auto &q : queues)
+        {
+            size_t qIdx = stoi(q);
```
qiluo-msft: >stoi [](start = 26, length = 4)

stoi may throw.

orchagent/pfcwdorch.cpp
```diff
@@ -239,6 +248,24 @@ void PfcWdOrch<DropHandler, ForwardHandler>::createEntry(const string& key,
...
+            if (qIdx >= port.m_queue_ids.size())
+            {
+                SWSS_LOG_ERROR("Invalid queue index %zd on port %s", qIdx, key.c_str());
+                continue;
+            }
+            if (!startWdActionOnQueue(PFC_WD_IN_STORM, port.m_queue_ids[qIdx]))
+            {
+                SWSS_LOG_ERROR("Failed to start PFC watchdog %s event action on port %s queue %zd", PFC_WD_IN_STORM, key.c_str(), qIdx);
```
qiluo-msft: >SWSS_LOG_ERROR [](start = 16, length = 14)

Why not 'continue'?

orchagent/pfcwdorch.cpp
```diff
@@ -818,27 +879,89 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::doTask(swss::NotificationConsumer
...
+
+    return true;
 }
 
 template <typename DropHandler, typename ForwardHandler>
-void PfcWdSwOrch<DropHandler, ForwardHandler>::doTask(SelectableTimer &timer)
+bool PfcWdSwOrch<DropHandler, ForwardHandler>::bake()
 {
-    SWSS_LOG_ENTER();
+    // clean all *_last fields in COUNTERS_TABLE
```
qiluo-msft: >// clean all *_last fields in COUNTERS_TABLE [](start = 4, length = 44)

Is this a new design? Do you have document?

orchagent/pfcwdorch.cpp
```diff
@@ -818,27 +879,89 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::doTask(swss::NotificationConsumer
...
-            handlerPair.second.handler->commitCounters(true);
+            if (fvField(fv).find("_last") != string::npos)
+            {
+                redisClient.hdel(COUNTERS_TABLE ":" + key, fvField(fv));
+            }
         }
     }
 
+    // Re-organize the field-value to "storm" : <lossless queue list> (e.g., "3, 4")
+    Table table(m_applDb.get(), "_" APP_PFC_WD_TABLE_NAME);
```
qiluo-msft: >"_" [](start = 32, length = 3)

Are you using the internal hash of ConsumerStateTable? It is not recommended.

orchagent/pfcwdorch.cpp
wendani: No, I do not have document. I see counter values are still in the COUNTERS_DB after warm-reboot. This makes sure that when PFCWD starts the detect and restore state machines it starts with the same state machine state as cold reboot.

orchagent/pfcwdorch.cpp
wendani: I transformed the syntax of field-value from queue_idx: storm to storm: queue_list, and used _PFC_WD_TABLE:Ethernet100 as a temp table to hold the new field-value (queue list in storm). Once refillToSync takes in the new field-value tuple, the temp table is removed. Not sure if it uses the internal hash of ConsumerStateTable

orchagent/pfcwdorch.cpp
```diff
@@ -818,27 +890,89 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::doTask(swss::NotificationConsumer
...
     {
-        if (handlerPair.second.handler != nullptr)
+        vector<FieldValueTuple> fvTuples;
+        PfcWdSwOrch<DropHandler, ForwardHandler>::getCountersTable()->get(key, fvTuples);
+        for (const auto &fv : fvTuples)
         {
-            handlerPair.second.handler->commitCounters(true);
+            if (fvField(fv).find("_last") != string::npos)
+            {
+                redisClient.hdel(COUNTERS_TABLE ":" + key, fvField(fv));
```
qiluo-msft: >hdel [](start = 28, length = 4)

optimize to 1 redis call. ref: [https://redis.io/commands/HDEL](https://redis.io/commands/HDEL)

orchagent/pfcwdorch.cpp
qiluo-msft: Don't catch all in production code, you can find the specific exceptions types.

---
In reply to: [257485760](https://github.com/Azure/sonic-swss/pull/794#discussion_r257485760) [](ancestors = 257485760)

orchagent/pfcwdorch.cpp
qiluo-msft: Offline discussed, I understand you just create a temporary table and delete immediately later. I suggest you use original table schema, and in DoTask, you can find the table name.

Sample code: [https://github.com/Azure/sonic-swss/blob/a6d60f2ed0b0059be6014e2aba7b57f4dd7387a0/orchagent/portsorch.cpp#L2262](https://github.com/Azure/sonic-swss/blob/a6d60f2ed0b0059be6014e2aba7b57f4dd7387a0/orchagent/portsorch.cpp#L2262)

---
In reply to: [257486025](https://github.com/Azure/sonic-swss/pull/794#discussion_r257486025) [](ancestors = 257486025)

wendani: The schema with the creation of a SubscriberStateTable is not supposed to hear anything anytime other than during the warm-reboot. This means that during the non-warm-reboot period, the storm status must not be updated to the subscribed table keyspace to avoid triggering the doTask, and in other words, this means that the storm status must be kept in a different keyspace, which is not being listened. Only when the warm-reboot is triggered, the subscribed keyspace will hear something, not natively but by explicitly refilled to it in an overrided bake(). While we save the transformation of the field-value  tuples in bake(), we still need to keep two keyspaces, one for warm-reboot doTask and one for tacking the current storm status, and the consumer object we created sits idle almost all the time only to fulfill its role for the flashing short-period of warm-reboot.

orchagent/pfcwdorch.cpp
wendani: This is an infrastructure lack. Added in https://github.com/Azure/sonic-swss-common/pull/267

orchagent/pfcwdorch.cpp
```diff
@@ -714,6 +730,94 @@ bool PfcWdSwOrch<DropHandler, ForwardHandler>::stopWdOnPort(const Port& port)
...
+                catch (const std::exception &e)
+                {
+                    SWSS_LOG_ERROR("Invalid conversion to int from string %s: %s", q.c_str(), e.what());
+                    continue;
+                }
+                catch (...)
+                {
+                    SWSS_LOG_ERROR("Unknown exception caught in string to int conversion");
+                    continue;
+                }
```
qiluo-msft: These catch-blocks are not needed according to library document.

wendani:
retest this please

orchagent/pfcwdorch.cpp
```diff
@@ -600,11 +611,8 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::unregisterFromWdDb(const Port& po
 
         // Clean up
         RedisClient redisClient(PfcWdOrch<DropHandler, ForwardHandler>::getCountersDb().get());
-        string countersKey = COUNTERS_TABLE ":" + sai_serialize_object_id(queueId);
-        redisClient.hdel(countersKey, "PFC_WD_DETECTION_TIME");
-        redisClient.hdel(countersKey, "PFC_WD_RESTORATION_TIME");
-        redisClient.hdel(countersKey, "PFC_WD_ACTION");
-        redisClient.hdel(countersKey, "PFC_WD_STATUS");
+        string countersKey = PfcWdSwOrch<DropHandler, ForwardHandler>::getCountersTable()->getTableName() + PfcWdSwOrch<DropHandler, ForwardHandler>::getCountersTable()->getTableNameSeparator() + sai_serialize_object_id(queueId);
```
qiluo-msft: >PfcWdSwOrch<DropHandler, ForwardHandler>:: [](start = 29, length = 42)

I think you can remove this specifier.

orchagent/pfcwdorch.cpp
```diff
@@ -55,28 +58,36 @@ void PfcWdOrch<DropHandler, ForwardHandler>::doTask(Consumer& consumer)
...
+            }
+
+            consumer.m_toSync.erase(it++);
         }
-        else
+
+        if (consumer.m_toSync.empty())
         {
-            SWSS_LOG_ERROR("Unknown operation type %s\n", op.c_str());
+            m_entriesCreated = true;
```
qiluo-msft: >m_entriesCreated [](start = 12, length = 16)

why you need this?

orchagent/pfcwdorch.cpp
wendani: pfcwdorch.cpp:616:47: error: there are no arguments to 'getCountersTable' that depend on a template parameter, so a declaration of 'getCountersTable' must be available [-fpermissive]
         string countersKey = getCountersTable()->getTableName() + PfcWdSwOrch<DropHandler, ForwardHandler>::getCountersTable()->getTableNameSeparator() + sai_serialize_object_id(queueId);
                                               ^
Compiler does not like the omission

pfcwdorch.cpp:616:47: note: (if you use '-fpermissive', G++ will accept your code, but allowing the use of an undeclared name is deprecated)
Makefile:841: recipe for target 'orchagent-pfcwdorch.o' failed
make[2]: *** [orchagent-pfcwdorch.o] Error 1

orchagent/pfcwdorch.cpp
wendani: This is introduced because we have two separate consumers now (as per your requested change), one in APPL_DB with table name PFC_WD_TABLE as the key in m_consumerMap, and one in COUNTERS_DB with table name PFC_WD as the key.

This introduces dependency between the execution of the two tables. PFC_WD should come first to startWdOnPort() per queue, before we can startWdActionOnQueue, where PFC_WD_TABLE keeps the in-storm queues on the port, where drop actions need to be taken further.

However, because the consumers are maintained as a key-value map, there is no guarantee which consumer comes in first. Introducing m_entriesCreated serves as a barrier in this case to ensure that PFC_WD consumer is processed before that of PFC_WD_TABLE so that we do not waste CPU cycles iterating all in-storm queues  in PFC_WD_TABLE without being able to start the drop action and retry a second time, although executing PFC_WD_TABLE before PFC_WD is totally fine from the code logic perspective (startWdActionOnQueue will return false and continue).


The second case this variable can help is when ports to start watchdog on comes in one batch in PFC_WD. This is the only case as far as I observe during both the cold- and warm-reboot initializations. m_entriesCreated ensures that PFC_WD is processed to its completion before PFC_WD_TABLE is ever processed. This guarantees that when PFC_WD_TABLE is handled, all lossless queues of all ports have watchdog started, and we will succeed starting action on any lossless queue without a doTask retry execution.

It should be totally fine removing the variable, and rely solely on doTask retries to resolve the dependency to converge all the in-storm queues into the drop action state untimately. But for the sake of efficiency, introducing the variable ensures some level of execution order and gives us better chances to successfully start action on all queues in one round of doTask.

PR #795, opened 2019-02-17
Fix PFC watchdog not getting lossless TC

wendani:
Fix PFC watchdog not getting lossless TCs. This will cause the PFC watchdog not getting started properly if enabled by default. A patch that hard-codes the TCs was used. With this fix, the patch can be removed.

Log error and return if no lossless TC is found when registering pfcwd stats on a port

**What I did**

**Why I did it**

**How I verified it**
Tested on brcm dut
**Details if related**


wendani:
One handlePortQosMapTable() call only processes one entry in the PORT_OQS_MAP table. If ports of interest are scattered under different keys (, which is not the current config use case), it is premature to say that the QoS map has been applied to all ports.

PR #796, opened 2019-02-17
DO NOT MERGE: [intfsorch] Wait for interface prior to prefix

marian-pritsak:
Interface table entries must be handled in the following order:
1. INTF_NAME
2. INTF_NAME:PREFIX

We need to wait for interface with VRF information prior to creating IP
addresses

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


orchagent/intfsorch.cpp
```diff
@@ -268,6 +268,14 @@ void IntfsOrch::doTask(Consumer &consumer)
                 continue;
             }
 
+            /* Wait for the Interface entry first */
```
prsunny: Do we need to wait in intfmgr as well?

marian-pritsak: intfmgr does two tasks of setting IP address and enslaving interface to VRF, which are independent from one another

wendani:
:shipit:

PR #800, opened 2019-02-23
Incorrect ACL stage CRM increment with Mirror

prsunny:
**What I did**
CRM counter got incremented for stage `egress `when there is a mirror session. With the mirror session removed, the CRM counter was decremented for stage `ingress`. 

**Why I did it**
Orchagent could crash if the used and available counters add up to `zero `(divide/zero) error. 

**How I verified it**
Tested with the fix and checked for values

**Details if related**


yxieca:
Change made to 201811 branch on 2/22/2019

PR #802, opened 2019-02-26
Fix MIRROR_SESSION table macro name

jipanyang:

**What I did**

"MIRROR_SESSION" table is in stateDB. 

**Why I did it**

**How I verified it**

**Details if related**


jipanyang:
Has dependency on https://github.com/Azure/sonic-swss-common/pull/264

PR #809, opened 2019-03-08
[portsorch]: add support to set mac-address learning attribute on bridge-port

shine4chen:
**What I did**
add support to set mac-address learning attribute on bridge-port

**Why I did it**
Mclag feature need this.  And it is common attribute for bridge-port.

**How I verified it**
compile and test on nephos lab

**Details if related**


lguohan:
learning mode can also be supported in lag, but it looks like the code here does not consider that scenario.

jianjundong:
> learning mode can also be supported in lag, but it looks like the code here does not consider that scenario.

Lag is also supported, please see the function PortsOrch::doLagTask(Consumer &consumer). Thinks.

leoli-nps:
Learning mode has been supported in lag, but the corresponding vs-test requires lag-name-oid-map. [PR#885](https://github.com/Azure/sonic-swss/pull/885) will provide support for lag-name-oid-map, so the vs-test code can only be provided after its modification is completed.

orchagent/portsorch.cpp
```diff
@@ -1813,7 +1820,48 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+                            else
+                            {
+                                SWSS_LOG_ERROR("Failed to set port %s learn mode to %s", alias.c_str(), learn_mode.c_str());
+                                it++;
+                                continue;
+                            }
+                        }
+                        else
+                        {
+                            m_portSetLearnModeList[alias] = learn_mode;
```
stcheng: could you add this to the current `port` class and later check if the variable `learn_mode` has value or not?

orchagent/portsorch.cpp
```diff
@@ -2013,7 +2061,38 @@ void PortsOrch::doVlanMemberTask(Consumer &consumer)
             }
 
             if (addBridgePort(port) && addVlanMember(vlan, port, tagging_mode))
```
stcheng: could you add this portion of the code into `addBridgePort` function and set the attribute during the bridge port creation time?

orchagent/portsorch.cpp
```diff
@@ -2767,7 +2923,7 @@ bool PortsOrch::addLag(string lag_alias)
 
     PortUpdate update = { lag, true };
     notify(SUBJECT_TYPE_PORT_CHANGE, static_cast<void *>(&update));
-
+
```
stcheng: remove

leoli-nps:
The vstest has been added, please review. Thanks!

orchagent/portsorch.cpp
shine4chen: yes, we will move learn_mode to port class. @stcheng

orchagent/portsorch.cpp
shine4chen: Good point. we will refine code per your suggestion.

shine4chen:
@stcheng we has refined code according to your suggest. Pls review thanks.

orchagent/portsorch.cpp
```diff
@@ -2221,6 +2264,42 @@ void PortsOrch::doLagTask(Consumer &consumer)
                         gIntfsOrch->setRouterIntfsMtu(l);
                     }
                 }
+
+                string learn_mode;
+                for (auto i : kfvFieldsValues(t))
+                {
+                     /* Get port fdb learn mode*/
+                    if (fvField(i) == "learn_mode")
```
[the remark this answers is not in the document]
shine4chen: @lisahnguyen  good point. We will revise it.

shine4chen:
retest it please

shine4chen:
retest this please

PR #810, opened 2019-03-08
[aclorch]: add support for acl rule to match out port

shine4chen:
**What I did**
 ACL rule can match out port.  Out port could be port intf , lag or vlan.

**Why I did it**
mclag need acl rule can bind to LAG and acl rule can match out port.

**How I verified it**
test it on nephos lab

**Details if related**


orchagent/aclorch.cpp
```diff
@@ -1168,6 +1200,13 @@ bool AclTable::create()
         table_attrs.push_back(attr);
     }
 
+    if (type == ACL_TABLE_L3)
```
zhenggen-xu: why only  support this on L3 table?

jianjundong: This rule may be supported on all ACL tables, but OUT_PORTS may be not supported by all vendors, so only support this on L3 table currently. ACL_TABLE_L3 can be removed later.

orchagent/aclorch.cpp
```diff
@@ -184,7 +184,23 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
                     SWSS_LOG_ERROR("Failed to locate port %s", alias.c_str());
                     return false;
                 }
-                m_inPorts.push_back(port.m_port_id);
+                if (port.m_type == Port::PHY)
+                {
+                    m_inPorts.push_back(port.m_port_id);
+                }
+                else if (port.m_type == Port::LAG)
```
lguohan: this does not match the SAI spec. only type PORT is support for SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS

https://github.com/opencomputeproject/SAI/blob/master/inc/saiacl.h#L1247

orchagent/aclorch.cpp
```diff
@@ -208,7 +224,23 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
...
                     return false;
                 }
-                m_outPorts.push_back(port.m_port_id);
+                if (port.m_type == Port::PHY)
+                {
+                    m_outPorts.push_back(port.m_port_id);
+                }
+                else if (port.m_type == Port::LAG)
+                {
+                    m_outPorts.push_back(port.m_lag_id);
```
lguohan: does not match SAI spec. 

https://github.com/opencomputeproject/SAI/blob/master/inc/saiacl.h#L1260

same as above.

orchagent/aclorch.cpp
```diff
@@ -1168,6 +1200,13 @@ bool AclTable::create()
         table_attrs.push_back(attr);
     }
 
+    if (type == ACL_TABLE_L3)
+    {
+        attr.id = SAI_ACL_TABLE_ATTR_FIELD_OUT_PORTS;
+        attr.value.booldata = true;
+        table_attrs.push_back(attr);
```
lguohan: I do not think OUT_PORTS is supported in other asic vendors, this code will break other platforms.

orchagent/aclorch.cpp
jianjundong: Thinks, we will delete LAG port.

orchagent/aclorch.cpp
jianjundong: Thinks, we will delete LAG port.

orchagent/aclorch.cpp
jianjundong: OUT_PORTS may be supported by MCLAG, this rule may be supported on all ACL tables, but OUT_PORTS may be not supported by all vendors, so only support this on L3 table currently. What is the best way to fix this issue? Thinks.

leoli-nps:
The corresponding VS test case has been added, please review.

stcheng:

shine4chen:
@stcheng  we have fixed the conflicts.

orchagent/aclorch.cpp
```diff
@@ -1367,6 +1367,13 @@ bool AclTable::create()
         table_attrs.push_back(attr);
     }
 
+    if (type == ACL_TABLE_L3)
```
stcheng: will you need L3V6?

shine4chen: Currently L3 table is enough. But we will need L3IPv6 in future.

lguohan:
where is the mclag design spec? since out_ports match are not supported in various asic vendor. we need a more generic approach to address this need.

shine4chen:
@lguohan  we have discuss it with BRCM folks in sonic-mclag-subgroup.   There is a workaround in SAI layer by combination of ingress acl and egress acl. With this workaround most vendors can support out-ports. 

The following is the isolation logic description copied from MCLAG-HLD
- In ASIC, ACL rule is used to isolate peer link from MC-LAG port. The rule is when the traffic is received from peer link and the output port is MC-LAG member port, the traffic must be dropped. For the chips whose ACL rule can't support out-port, there is a workaround in SAI layer by combination of ingress acl and egress acl. An alternative approach is  to use isolation group. Isolation group still have some weakness, First isolation group can't support tunnel-port and orchagent doesn't support it currently.  secondly isolation group may not be supported by all ASIC vendors. Using ACL is a more generic way to support the isolation function. We can refine this function to use isolation group later if it’s required.

lguohan:
unless Broadcom SAI as well as other vendor SAI has implemented outPorts, merging this PR will immediately break the image.

lguohan:
isolation group is much more generic and different ASIC can choose different way to implement. Not all ASIC verdors have the same workaround as broadcom. I do not see a reason why we cannot add LAG into isolation group, it is a simple metadata change.

shine4chen:
@lguohan Thanks for the feedback. The following is my answer for your concern.
1. Mclag feature is not enabled by default. So even for the vendor who doesn't work around acl-outport it wouldn't break the image by default. If vendor like to enable the mclag , of course it need to fix up.
2. In the existing master branch ,  `saiisolationgroup.h` isn't included and orchagent support is missing too. So needless to mention SAI support. Nephos team plans to include mclag feature on 201910 release. As time is limited could community accept this PR now? We can refine the isolation code on our next release.

adyeung:
BRCM will implement an isolation group solution for ingress filtering instead of outPorts, pkts will be dropped at the ingress instead of queueing for egress drop. Design proposal along with other MCLAG improvements will be shared in the upcoming BRCM enhancement HLD.

shine4chen:
retest this please

shine4chen:
retest this please

orchagent/aclorch.cpp
shine4chen: Yes , we will define new table dedicated for MCLAG. So it wouldn't affect others in case that mclag isn't enabled.

shine4chen: Done

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

PR #812, opened 2019-03-08
[arp] copy arp IO to cpu instead of trap and drop

yxieca:
**What I did**
Switch arp/arp response/neighbor_discovery IO trap action from 'trap' to 'copy'. These IO, instead of trap to CPU and dropped, they will be copy to CPU and forwarded on VLAN.

**Why I did it**
During warm reboot, we need VLAN take care of ARP traffic by itself.

**How I verified it**
Tested with matching test: https://github.com/Azure/sonic-mgmt/pull/824

### DIFF
--- a/swssconfig/sample/00-copp.config.json
+++ b/swssconfig/sample/00-copp.config.json
@@ -22,7 +22,7 @@
     {
         "COPP_TABLE:trap.group.arp": {
             "trap_ids": "arp_req,arp_resp,neigh_discovery",
-            "trap_action":"trap",
+            "trap_action":"copy",
             "trap_priority":"4",
             "queue": "4",
             "meter_type":"packets",


prsunny:
Just to highlight, there will be now two copies of ARP requests forwarded. One flooded by ASIC and other forwarded by kernel. There is no functional impact but a minor behavioral change.

yxieca:
Thanks for the clarification, Prince!

swssconfig/sample/00-copp.config.json
```diff
@@ -22,7 +22,7 @@
     {
         "COPP_TABLE:trap.group.arp": {
             "trap_ids": "arp_req,arp_resp,neigh_discovery",
-            "trap_action":"trap",
+            "trap_action":"copy",
```
andriymoroz-mlnx: this crashes on Mellanox switches
please do not merge yet, will clarify with SAI

lguohan:
this is used for warm reboot only, so we should only do the change before warm reboot, and change it back after warm reboot.

swssconfig/sample/00-copp.config.json
pliviu: I think this change creates a loop for ICMPv6 RS & NS packets in an L2 MC-LAG scenario as i documented here:
https://github.com/sonic-net/SONiC/issues/1253

PR #256, opened 2017-07-18
Fix preallocated bridge_port_list size

qiluo-msft:

PR #814, opened 2019-03-11
[intfsorch]:add support to  change rif mac address

shine4chen:
**What I did**
add support to change rif mac address

**Why I did it**
some apps such as mclag or vrrp need to change rif mac address

**How I verified it**
test it on nephos lab

**Details if related**


orchagent/intfsorch.cpp
```diff
@@ -181,6 +181,10 @@ void IntfsOrch::doTask(Consumer &consumer)
     while (it != consumer.m_toSync.end())
     {
         KeyOpFieldsValuesTuple t = it->second;
+        MacAddress mac;
+        int mac_attr_set;
+
+        mac_attr_set = 0;
```
prsunny: Can you set this in one line?

orchagent/intfsorch.cpp
```diff
@@ -208,6 +212,12 @@ void IntfsOrch::doTask(Consumer &consumer)
             {
                 vnet_name = value;
             }
+            else if (field == "mac_addr")
+            {
+                mac_attr_set = 1;
+
```
prsunny: Remove extra line

lguohan:
need virtual switch test.

lguohan:
@prsunny , I assume this can only be on rif without ip address. We can not set mac address for interface entry with IP address since we only create one rif for all ip addresses on that rif?

orchagent/intfsorch.cpp
jianjundong: I'm not sure in the loop that 'int mac_attr_set = 0' will be execute every time? If do, then can set this in one line.

jianjundong:
> @prsunny , I assume this can only be on rif without ip address. We can not set mac address for interface entry with IP address since we only create one rif for all ip addresses on that rif?

In mclag, the ip address of mclag enabled portchannel of two peers must be the same, thus the mac address of rifs also must be the same.

jianjundong:
> "mac_addr" is not a field currently in the INTF_TABLE/INTERFACE_TABLE. This needs to be defined in the schema and also require corresponding change in interface manager

'mac_addr' is set as a field of INTF_TABLE in APP_DB currently. It is set by iccpd through mclagsyncd. This field may also be set as a field of INTERFACE_TABLE in CFG_DB and processed by intfmgrd? Thanks.

orchagent/intfsorch.cpp
shine4chen: It is no problem to put these codes in one line.

shine4chen:
> need virtual switch test.

sure, we will add it soon

shine4chen:
> need virtual switch test.

@lguohan  Can you give us example or URL how to connect virtual switch test to specific PR

shine4chen:
@lguohan even for the interface associated with the ip address we still can/need to change its src-mac.

shine4chen:
@prsunny Per our discussion on VRF,  src-mac attribute should be put on {INTERFACE:Ethernetx} entry. I suggest we merge this PR to VRF code. How about your opinion?

lguohan:
@shine4chen , thanks for adding vs test, but the added two tests are failing, can you investigate the failure.

orchagent/intfsorch.cpp
stcheng: you could follow the other vairables' logics and check mac is empty or not; you don't need to have this variable

orchagent/intfsorch.cpp
```diff
@@ -299,6 +309,31 @@ void IntfsOrch::doTask(Consumer &consumer)
                 }
             }
 
+            if (mac_attr_set == 1)
+            {
+                auto it_intfs = m_syncdIntfses.find(alias);
+                if (it_intfs != m_syncdIntfses.end())
+                {
+                /* Get mac information and update mac of the interface*/
```
stcheng: Indent 4 more spaces

tests/test_intf_mac.py
```diff
@@ -0,0 +1,248 @@
...
+        tbl.set(interface + "|" + ip, fvs)
+        time.sleep(2)
+        
+    def test_InterfaceSetMac(self, dvs, testlog):
+        self.setup_db(dvs)
+
+        # assign IP to interface
+        self.add_ip_address("Ethernet8", "10.0.0.4/31")
+
+        # check application database
```
stcheng: you could remove this portion of the test code before `set_mac` since this part is already covered by `test_interface.py`.

tests/test_intf_mac.py
```diff
@@ -0,0 +1,248 @@
...
+            elif fv[0] == "mac_addr":
+                assert fv[1] == "6c:ec:5a:11:22:33"
+            else:
+                assert False
+
+        # check ASIC router interface database
+        tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_ROUTER_INTERFACE")
+        intf_entries = tbl.getKeys()
+        # one loopback router interface one port based router interface
+        #assert len(intf_entries) == 2
```
stcheng: why this is commented out?

tests/test_intf_mac.py
```diff
@@ -0,0 +1,248 @@
...
+
+        for key in intf_entries:
+            (status, fvs) = tbl.get(key)
+            assert status == True
+            # a port based router interface has five field/value tuples
+            if len(fvs) == 5:
```
stcheng: this is not correct. if the length is not 5, the test will still pass.
you will need to validate if `SAI_ROUTER_INTERFACE_ATTR_SRC_MAC_ADDRESS` appears in the attribute list.
check the file `test_neighbor.py` and use one more variable `src_mac_address_found` variable to indicate if this attribute appears in the list and assert this variable is true afterwards.

I'll update the similar code in `test_interface.py` file.

tests/test_intf_mac.py
```diff
@@ -0,0 +1,248 @@
...
+    def test_InterfaceSetMac(self, dvs, testlog):
+        self.setup_db(dvs)
+
+        # create port channel
+        self.create_port_channel(dvs, "PortChannel001")
+
+        # assign IP to interface
+        self.add_ip_address("PortChannel001", "30.0.0.4/31")
+
+        # check application database
```
stcheng: remove this portion before `set_mac`; same as above

tests/test_intf_mac.py
```diff
@@ -0,0 +1,248 @@
...
+        tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_ROUTER_INTERFACE")
+        intf_entries = tbl.getKeys()
+        # one loopback router interface one port based router interface
+        #assert len(intf_entries) == 2
+
+        for key in intf_entries:
+            (status, fvs) = tbl.get(key)
+            assert status == True
+            # a port based router interface has five field/value tuples
+            if len(fvs) == 5:
```
stcheng: same suggestions as above

tests/test_intf_mac.py
```diff
@@ -0,0 +1,248 @@
+from swsscommon import swsscommon
+
+import time
+import json
+
+class TestRouterInterfaceMac(object):
```
stcheng: will you support change the source MAC address for the router interfaces? if yes, could you add one more test case to change the source MAC address and validate the result?

shine4chen: > please check the comments.
> 
> one more question, will you support remove the source MAC address once it is set?

@stcheng  Do you mean that set source mac address to default value? Unfortunately the existing producerstate class doesn't  support to del one attribute of the specific entry. You have to explicit set src mac address attribute of the INTERFACE_TABLE entry on appdb to default value(most likely system mac address).

leoli-nps:
The vstest has been updated, please review. Thanks!

shine4chen:
@stcheng  @prsunny  we have refined code according to your suggestion. Pls help to review it. Thanks.

prsunny:
> @prsunny Per our discussion on VRF,  src-mac attribute should be put on {INTERFACE:Ethernetx} entry. I suggest we merge this PR to VRF code. How about your opinion?

@shine4chen , yes I agree. I think you've made the changes based on this assumption currently correct?. If so, the change must be in `doIntfGeneralTask `and not in `doIntfAddrTask `in `intfmgr.cpp`. Can you please check?

tests/test_intf_mac.py
stcheng: @shine4chen I agree. No need to remove the value.

tests/test_intf_mac.py
```diff
@@ -0,0 +1,258 @@
...
+        self.cdb = swsscommon.DBConnector(swsscommon.CONFIG_DB, dvs.redis_sock, 0)
+        self.cntdb = swsscommon.DBConnector(swsscommon.COUNTERS_DB, dvs.redis_sock, 0)
+
+    def add_ip_address(self, interface, ip):
+        tbl = swsscommon.Table(self.cdb, "INTERFACE")
+        fvs = swsscommon.FieldValuePairs([("NULL", "NULL")])
+        tbl.set(interface + "|" + ip, fvs)
+        time.sleep(2) # IPv6 netlink message needs longer time
+
+    def remove_ip_address(self, interface, ip):
```
stcheng: these functions are already defined in `conftest.py` file. you could re-use them.

tests/test_intf_mac.py
```diff
@@ -0,0 +1,258 @@
...
+            tbl._del(lag + "|" + member)
+            time.sleep(1)
+
+    def add_ip_address(self, interface, ip):
+        tbl = swsscommon.Table(self.cdb, "PORTCHANNEL_INTERFACE")
+        fvs = swsscommon.FieldValuePairs([("NULL", "NULL")])
+        tbl.set(interface + "|" + ip, fvs)
+        time.sleep(1)
+
+    def remove_ip_address(self, interface, ip):
```
stcheng: these functions are already defined in conftest.py file. you could re-use them.

tests/test_intf_mac.py
leoli-nps: The vstest has been updated according to your comments, please review again, thanks.

tests/test_intf_mac.py
leoli-nps: @stcheng , the vstest has been updated according to your comments, please review again, thanks.

shine4chen:
> > @prsunny Per our discussion on VRF,  src-mac attribute should be put on {INTERFACE:Ethernetx} entry. I suggest we merge this PR to VRF code. How about your opinion?
> 
> @shine4chen , yes I agree. I think you've made the changes based on this assumption currently correct?. If so, the change must be in `doIntfGeneralTask `and not in `doIntfAddrTask `in `intfmgr.cpp`. Can you please check?
@prsunny  yes, We will move the change to `doIntfGeneralTask` function.

shine4chen:
@prsunny @stcheng  we has refined code and vs test cases according to your opinion.

cfgmgr/intfmgr.cpp
```diff
@@ -125,6 +140,10 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
         {
             vrf_name = value;
         }
+        else  if (field == "mac_addr")
```
stcheng: extra space here

cfgmgr/intfmgr.cpp
```diff
@@ -148,6 +167,11 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
             {
                 setIntfVrf(alias, vrf_name);
             }
+            /*Set the mac of interface*/
+            if(!mac.empty())
```
stcheng: missing a space between `if` and `(`

cfgmgr/intfmgr.cpp
shine4chen: sure, remove it.

cfgmgr/intfmgr.cpp
shine4chen: yes, add it.

lguohan:
retest this please

lguohan:
@stcheng , can you check?

leoli-nps:
4 failed test cases in vstest:
```
test_intf_mac.TestRouterInterfaceMac.test_InterfaceSetMac
test_intf_mac.TestRouterInterfaceMac.test_InterfaceChangeMac
test_intf_mac.TestLagRouterInterfaceMac.test_InterfaceSetMac
test_intf_mac.TestLagRouterInterfaceMac.test_InterfaceChangeMac
```
The reasons for their failure are as follows:
``` 
AttributeError: 'Table' object has no attribute 'hdel'
```
This has been modified in the PR[#256](https://github.com/Azure/sonic-swss-common/pull/256), please check.

prsunny:
@stcheng, can you re-review this? Looks like comments are addressed.

tests/test_intf_mac.py
shine4chen: sure, done.

tests/test_intf_mac.py
shine4chen: Sure,done

orchagent/intfsorch.cpp
shine4chen: Sure,done

shine4chen:
retest this please

rlhui:
@shine4chen , Reminder to resolve conflicts.

prsunny:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

cfgmgr/intfmgr.cpp
```diff
@@ -437,6 +455,12 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
             setIntfVrf(alias, vrf_name);
         }
 
+        /*Set the mac of interface*/
+        if (!mac.empty())
+        {
+            setIntfMac(alias, mac);
+        }
```
wendani: if mac.empty() is true, that is the case of not having "mac_addr" field in CONFIG_DB. In this case, the convention is to push_back field-value tuple `"mac_addr": ""` to `data` to set to APPL_DB so that IntfsOrch does not make any assumption on default values. See the treatment on `mtu` and `adminStatus` on lines 493-497 and 499-504, respectively.

orchagent/intfsorch.cpp
```diff
@@ -382,6 +382,7 @@ void IntfsOrch::doTask(Consumer &consumer)
     while (it != consumer.m_toSync.end())
     {
         KeyOpFieldsValuesTuple t = it->second;
+        MacAddress mac;
```
wendani: This can be moved to lines 408 - 410, where variables used for the same purpose are declared.

shine4chen: good point, done

cfgmgr/intfmgr.cpp
shine4chen: Done

wendani: Thank you for looking into the comment.

Just check the MacAddress class. When IntfOrch gets a value of empty string on line 424, it will use MacAddress::MacAddress(const std::string& macStr) to construct the MacAddress object. An empty string will cause MacAddress::parseMacString to return false, and further throw invalid_argument https://github.com/Azure/sonic-swss-common/blob/master/common/macaddress.cpp#L23

Did you try this change on a vs/a dut?

We may deem an empty string to be valid in constructing MacAddress, or put value "00:00:00:00:00:00" which is deem valid in current parseMacString logic.

shine4chen: Yes, In fact I exactly put "00:00:00:00:00:00" string in app-db

orchagent/intfsorch.cpp
```diff
@@ -419,6 +419,10 @@ void IntfsOrch::doTask(Consumer &consumer)
             {
                 vnet_name = value;
             }
+            else if (field == "mac_addr")
+            {
+                mac = MacAddress(value);
```
wendani: As MacAddress(value) may throw invalid_argument, it is better to put this statement around 
```
try
{
    mac = MacAddress(value);
}
catch (const std::invalid_argument &e)
{
    ...
}
```

shine4chen: Good point, done

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
@stcheng @prsunny  @adyeung I resolved some conflict. It seems systems automatic dismissed your approve. Would you please to  review and approve it?

PR #816, opened 2019-03-15
[vnetorch]: Add VNET/tunnel/route removal flows for Bitmap VNET imple…

volodymyrsamotiy:
…mentation

**What I did**
Added support of  VNET/tunnel/route removal flows for Bitmap VNET implementation.

**Why I did it**
Implemented new logic for Bitmap VNET implementation in order to add support of VNET/tunnel/route removal.

**How I verified it**
1. Verified manually.
2. Ran vnet_vxlan ansible test and verified that it passed.

**Details if related**
N/A

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -264,8 +264,7 @@ std::bitset<VNET_BITMAP_SIZE> VNetBitmapObject::vnetBitmap_;
 std::bitset<VNET_TUNNEL_SIZE> VNetBitmapObject::tunnelOffsets_;
 map<string, uint32_t> VNetBitmapObject::vnetIds_;
 map<uint32_t, VnetBridgeInfo> VNetBitmapObject::bridgeInfoMap_;
-map<tuple<MacAddress, sai_object_id_t>, sai_fdb_entry_t> VNetBitmapObject::fdbMap_;
-map<tuple<MacAddress, sai_object_id_t>, sai_neighbor_entry_t> VNetBitmapObject::neighMap_;
+map<tuple<MacAddress, sai_object_id_t>, VnetNeighInfo> VNetBitmapObject::neighInfoMap_;
 
 VNetBitmapObject::VNetBitmapObject(const std::string& vnet, const VNetInfo& vnetInfo,
                              vector<sai_attribute_t>& attrs) : VNetObject(vnetInfo)
@@ -275,6 +274,7 @@ VNetBitmapObject::VNetBitmapObject(const std::string& vnet, const VNetInfo& vnet
     setVniInfo(vnetInfo.vni);
 
     vnet_id_ = getFreeBitmapId(vnet);
+    vnet_name_ = vnet;
 }
 
 bool VNetBitmapObject::updateObj(vector<sai_attribute_t>&)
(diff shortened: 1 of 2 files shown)


prsunny:
retest this please

orchagent/vnetorch.cpp
```diff
@@ -469,18 +476,115 @@ VnetBridgeInfo VNetBitmapObject::getBridgeInfoByVni(uint32_t vni, string tunnelN
         throw std::runtime_error("vni creation failed");
     }
 
-    tunnel->addEncapMapperEntry(info.bridge_id, vni);
+    // FIXME: Use "createVxlanTunnelMap()" for tunnel mapper creation
+    auto tunnelEncapMapperEntry = tunnel->addEncapMapperEntry(info.bridge_id, vni);
```
prsunny: It is better to call the orch function instead of directly invoking the addEncapMapper and insertMapperEntry. Do you want to fix it in this PR?

orchagent/vnetorch.cpp
```diff
@@ -793,6 +892,70 @@ bool VNetBitmapObject::addTunnelRoute(IpPrefix& ipPrefix, tunnelEndpoint& endp)
         throw std::runtime_error("VNet route creation failed");
     }
 
+    tunnelRouteInfo.vni = endp.vni == 0 ? getVni() : endp.vni;
+    tunnelRouteInfo.mac = mac;
+    tunnelRouteMap_.emplace(ipPrefix, tunnelRouteInfo);
```
prsunny: Do you want to add a DEBUG log here for the create? Please ignore if it is logged elsewhere in the flow.

orchagent/vnetorch.cpp
```diff
@@ -898,9 +1062,56 @@ bool VNetBitmapObject::addRoute(IpPrefix& ipPrefix, nextHop& nh)
...
+}
+
+bool VNetBitmapObject::removeRoute(IpPrefix& ipPrefix)
+{
+    SWSS_LOG_ENTER();
+
+    if (routeMap_.find(ipPrefix) == routeMap_.end())
+    {
+        SWSS_LOG_WARN("VNET route %s doesn't exist", ipPrefix.to_string().c_str());
+        return false;
```
prsunny: if it doesn't exist, may be we can return `true `to avoid further retry.

orchagent/vnetorch.cpp
volodymyrsamotiy: I agree, but currently it cannot be used due to some issue in MLNX SAI. That is why only "encap" mapper is created directly. So, I added a comment here and it will be fixed later on.

orchagent/vnetorch.cpp
volodymyrsamotiy: If it doesn't exist then something went wrong, so I believe we cannot just return ```true```. Also ```false``` is returned here to handle situation when ```add``` and ```remove``` operations are received in wrong order for the same entry.

stcheng:
hello @volodymyrsamotiy , could you address this issue?

```
vnetorch.cpp: In destructor 'virtual VNetBitmapObject::~VNetBitmapObject()':
vnetorch.cpp:1102:55: error: throw will always call terminate() [-Werror=terminate]
         throw std::runtime_error("VNET removal failed");
                                                       ^
vnetorch.cpp:1102:55: note: in C++11 destructors default to noexcept
vnetorch.cpp:1107:55: error: throw will always call terminate() [-Werror=terminate]
         throw std::runtime_error("VNET removal failed");
                                                       ^
vnetorch.cpp:1107:55: note: in C++11 destructors default to noexcept
```

stcheng:
This is blocking https://github.com/Azure/sonic-buildimage/pull/2714

jipanyang:
@volodymyrsamotiy , as @stcheng mentioned, are you addressing this compiling issue with gcc 6.3.0 for swss stretch update?  Since teamd and bgp depend on swss,  those two dockers are also waiting for the fix. thanks!

volodymyrsamotiy:
@stcheng, @jipanyang, I am addressing it, I will provide PR today

PR #818, opened 2019-03-18
Allow ACL entry creation without ACL counter attaching to it

wendani:

**What I did**

**Why I did it**
In pfc watchdog, packet drop stats does not come from the acl counters on brcm platform. In such a case, we do not need to attach the acl counters to the ACL entry. This is also the application case for acl entry creation in QosOrch::initAclEntryForEcn, which does not come with the acl counters attached.

**How I verified it**
On brcm dut
**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -122,7 +122,7 @@ inline string trim(const std::string& str, const std::string& whitespace = " \t"
     return str.substr(strBegin, strRange);
 }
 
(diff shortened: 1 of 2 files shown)


lguohan:
what is the problem for attaching the counter to those acls?

wendani:
PFC watchdog warm-reboot with the presence of dataplane traffic that increments the ACL counters to non-zero. https://github.com/Azure/sonic-swss/pull/818#issuecomment-477410447

lguohan:
we need vs test.

PR #820, opened 2019-03-27
[vnetorch]: Add VNET interface removal flow for Bitmap VNET implement…

volodymyrsamotiy:
…ation

**What I did**
Added support of  VNET interface removal flow for Bitmap VNET implementation.

**Why I did it**
Implemented new logic for Bitmap VNET implementation in order to add support of VNET interface removal.

**How I verified it**
1. Verified manually.
2. Ran vnet_vxlan ansible test and verified that it passed.

**Details if related**
N/A

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -206,6 +206,51 @@ bool IntfsOrch::setIntf(const string& alias, sai_object_id_t vrf_id, const IpPre
     return true;
 }
 
+bool IntfsOrch::removeIntf(const string& alias, sai_object_id_t vrf_id, const IpPrefix *ip_prefix)
+{
+    SWSS_LOG_ENTER();
+
+    Port port;
+    if (!gPortsOrch->getPort(alias, port))
+    {
+        return false;
+    }
+
+    if (m_syncdIntfses.find(alias) == m_syncdIntfses.end())
+    {
+        return false;
+    }
+
+    if (ip_prefix && m_syncdIntfses[alias].ip_addresses.count(*ip_prefix))
+    {
+        removeSubnetRoute(port, *ip_prefix);
+        removeIp2MeRoute(vrf_id, *ip_prefix);
+
+        if(port.m_type == Port::VLAN)
+        {
+            removeDirectedBroadcast(port, *ip_prefix);
+        }
+
+        m_syncdIntfses[alias].ip_addresses.erase(*ip_prefix);
+    }
+
+    /* Remove router interface that no IP addresses are associated with */
+    if (m_syncdIntfses[alias].ip_addresses.size() == 0)
+    {
+        if (removeRouterIntfs(port))
+        {
+            m_syncdIntfses.erase(alias);
+            return true;
+        }
+        else
+        {
+            return false;
+        }
+    }
+
+    return true;
+}
+
 void IntfsOrch::doTask(Consumer &consumer)
 {
     SWSS_LOG_ENTER();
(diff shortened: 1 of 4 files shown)


volodymyrsamotiy:
retest this please

volodymyrsamotiy:
retest this please

volodymyrsamotiy:
retest this please

prsunny:
retest this please

prsunny:
retest this please

orchagent/intfsorch.cpp
```diff
@@ -206,6 +206,51 @@ bool IntfsOrch::setIntf(const string& alias, sai_object_id_t vrf_id, const IpPre
...
+{
+    SWSS_LOG_ENTER();
+
+    Port port;
+    if (!gPortsOrch->getPort(alias, port))
+    {
+        return false;
+    }
+
+    if (m_syncdIntfses.find(alias) == m_syncdIntfses.end())
```
prsunny: This check is redundant.. Its already done before this call.

orchagent/intfsorch.cpp
```diff
@@ -206,6 +206,51 @@ bool IntfsOrch::setIntf(const string& alias, sai_object_id_t vrf_id, const IpPre
...
+        if(port.m_type == Port::VLAN)
+        {
+            removeDirectedBroadcast(port, *ip_prefix);
+        }
+
+        m_syncdIntfses[alias].ip_addresses.erase(*ip_prefix);
+    }
+
+    /* Remove router interface that no IP addresses are associated with */
+    if (m_syncdIntfses[alias].ip_addresses.size() == 0)
```
prsunny: Looks like for VRF implementation, this `if-check` handles the scenario where we get delete of interface without ip_prefix first and later comes the delete of interface with ip_prefix. From the below code, looks like this sequence is handled for Bitmap implementation also. Can you please confirm?

orchagent/vnetorch.cpp
```diff
@@ -715,6 +734,73 @@ bool VNetBitmapObject::addIntf(const string& alias, const IpPrefix *prefix)
...
+
+    auto& intf = intfMap_.at(alias);
+
+    if (prefix)
+    {
+        if (intf.pfxMap.find(*prefix) == intf.pfxMap.end())
+        {
+            SWSS_LOG_ERROR("VNET '%s' interface '%s' prefix '%s' doesn't exist",
+                    getVnetName().c_str(), alias.c_str(), prefix->getIp().to_string().c_str());
+            return false;
```
prsunny: Suggest to `return true` here. We don't want to retry in this case.

orchagent/vnetorch.cpp
```diff
@@ -715,6 +734,73 @@ bool VNetBitmapObject::addIntf(const string& alias, const IpPrefix *prefix)
...
+        gIntfsOrch->removeIp2MeRoute(gVirtualRouterId, *prefix);
+
+        recycleTunnelRouteTableOffset(pfx.offset);
+
+        intf.pfxMap.erase(*prefix);
+    }
+
+    if (intf.pfxMap.size() == 0)
+    {
+        if (gIntfsOrch->getSyncdIntfses().find(alias) == gIntfsOrch->getSyncdIntfses().end())
```
prsunny: This check seems to be redundant. It is already checked before the call. Is there a scenario where this can happen?

orchagent/vnetorch.cpp
```diff
@@ -640,12 +640,18 @@ bool VNetBitmapObject::addIntf(const string& alias, const IpPrefix *prefix)
 
     if (gIntfsOrch->getSyncdIntfses().find(alias) == gIntfsOrch->getSyncdIntfses().end())
     {
+        if (intfMap_.find(alias) != intfMap_.end())
+        {
+            SWSS_LOG_ERROR("VNET '%s' interface '%s' already exists", getVnetName().c_str(), alias.c_str());
+            throw std::runtime_error("VNET interface creation failed");
```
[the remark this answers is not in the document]
volodymyrsamotiy: Fixed.

orchagent/intfsorch.cpp
volodymyrsamotiy: Agree, removed this check.

orchagent/vnetorch.cpp
volodymyrsamotiy: Fixed.

orchagent/vnetorch.cpp
volodymyrsamotiy: Agree, there is no such scenario, remove this check

orchagent/intfsorch.cpp
volodymyrsamotiy: Yes, this sequence is handled for Bitmap implementation (since remove flow is different). Once it is removed in Bitmap ```removeIntf()``` we also  call this ```IntfsOrch::removeIntf``` function but with ip_prefix set to ```nullptr``` so it allows to remove all left SAI object.

PR #821, opened 2019-03-28
Dynamic transceiver tuning support 

dgsudharsan:
**- What I did**
Dynamic transceiver tuning support
**- How I did it**
Dynamically program pre-emphasis and other settings based on media detected in xcvrd. The various supported media are defined in media_settings.json

**- How to verify it**
Perform OIR of different media types. Dump the pre-emphasis values from hardware and verify that they got reflected

**- Description for the changelog**

Support for dynamically programming pre-emphasis

**DO NOT MERGE THIS PULL REQUEST UNTIL BCM SDK SUPPORT FOR SAI PREEMPHASIS ATTRIBUTES HAVE BEEN MERGED**

**- A picture of a cute animal (not mandatory but encouraged)**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1519,6 +1519,9 @@ void PortsOrch::doPortTask(Consumer &consumer)
         if (op == SET_COMMAND)
         {
             set<int> lane_set;
+            vector<uint32_t> pre_emphasis;
+            vector<uint32_t> idriver;
+            vector<uint32_t> ipredriver;
             string admin_status;
             string fec_mode;
             string pfc_asym;
@@ -1574,6 +1577,24 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 if (fvField(i) == "autoneg")
                 {
                     an = (int)stoul(fvValue(i));
+                }
+
+                /* Set port serdes Pre-emphasis */
+                if (fvField(i) == "preemphasis")
+                {
+                    getPortSerdesVal(fvValue(i), pre_emphasis);
+                }
+
+                /* Set port serdes idriver */
+                if (fvField(i) == "idriver")
+                {
+                    getPortSerdesVal(fvValue(i), idriver);
+                }
+
+                /* Set port serdes ipredriver */
+                if (fvField(i) == "ipredriver")
+                {
+                    getPortSerdesVal(fvValue(i), ipredriver);
                 }
             }
 
@@ -1853,6 +1874,51 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         it++;
                         continue;
                     }
+                }
+
+                if (pre_emphasis.size() != 0)
+                {
+                    if (setPortSerdesAttribute(p.m_port_id, SAI_PORT_ATTR_SERDES_PREEMPHASIS, pre_emphasis))
+                    {
+                        SWSS_LOG_NOTICE("Set port %s  preemphasis is success", alias.c_str());
+                    }
+                    else
+                    {
+                        SWSS_LOG_ERROR("Failed to set port %s pre-emphasis", alias.c_str());
+                        it++;
+                        continue;
+                    }
+
+                }
+
+                if (idriver.size() != 0)
+                {
+                    if (setPortSerdesAttribute(p.m_port_id, SAI_PORT_ATTR_SERDES_IDRIVER, idriver))
+                    {
+                        SWSS_LOG_NOTICE("Set port %s idriver is success", alias.c_str());
+                    }
+                    else
+                    {
+                        SWSS_LOG_ERROR("Failed to set port %s idriver", alias.c_str());
+                        it++;
+                        continue;
+                    }
+
+                }
+
+                if (ipredriver.size() != 0)
+                {
+                    if (setPortSerdesAttribute(p.m_port_id, SAI_PORT_ATTR_SERDES_IPREDRIVER, ipredriver))
+                    {
+                        SWSS_LOG_NOTICE("Set port %s ipredriver is success", alias.c_str());
+                    }
+                    else
+                    {
+                        SWSS_LOG_ERROR("Failed to set port %s ipredriver", alias.c_str());
+                        it++;
+                        continue;
+                    }
+
                 }
 
                 /* Last step set port admin status */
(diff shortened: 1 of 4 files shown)


orchagent/portsorch.cpp
```diff
@@ -3166,6 +3171,46 @@ void PortsOrch::doTask(NotificationConsumer &consumer)
...
+            }
+            else if(fvField(i) == "idriver")
+            {
+                attr_id = SAI_PORT_ATTR_SERDES_IDRIVER;
+            }
+            else if(fvField(i) == "ipredriver")
+            {
+                attr_id = SAI_PORT_ATTR_SERDES_IPREDRIVER;
+            } 
+            else if(fvField(i) == "port")
```
wschwartz: Hi Sudharsan,

I am reviewing the SAI piece of this as well at Broadcom and one thing I am wondering is how you handle interface mode.  Along with pre-emphasis settings, for DACs we will want to set either CR/XFI where for copper we would want to program SR/SFI.  The transceiver compliance field is all we would need but that would need to get passed down to SAI to set the interface mode properly.

dgsudharsan: Hi @wschwartz Currently I don't think SONiC has a mechanism to program the interface mode. This pull request is just to handle pre-emphasis related settings alone. The interface mode can later be extended in this logic or can be incorporated in some other different logic. For now without setting interface mode, the pre-emphasis settings alone work good in SONiC and if there is a future requirement we can create a new PR for the related settings.

lguohan: yes, it can. check this attribute. SAI_PORT_ATTR_MEDIA_TYPE

@wschwatz, do we think we can fix this media type in another PR, or it has to be this PR?

wschwartz: Hi @lguohan @dgsudharsan  I think this could be extended in the future to support interface mode in a future PR.  With TH3 coming we have a new set of interface modes to support so its probably good to take time and investigate those.    I think @ben-gale is working to find another reviewer as well.

orchagent/portsorch.cpp
```diff
@@ -3166,6 +3171,46 @@ void PortsOrch::doTask(NotificationConsumer &consumer)
         }
 
         sai_deserialize_free_port_oper_status_ntf(count, portoperstatus);
+    } else if (op == "media_change") {
```
lguohan: media settings should go to standard port app db, like the admin_status in the port. I do not understand why choose the notification consumer here. 

Please make sure your code style is consistent, you need to keep a space for after if.

dgsudharsan: Hi Guohan,
I feel that media event is analogous to oper state notification generated by ASIC. These are from South bound APIs. If we see the oper status implementation, a notification consumer is used which is from syncd to orchagent. I used the same logic to implement this as similar to oper status. On other hand I feel admin status is application triggerd change from user. 
Please let me know your views. If i still want to go ahead and implement it similar to admin status I will go ahead and update the PR.

lguohan:
also since you add new function, you also need to add vs test on this.

dgsudharsan:
Hi @lguohan. I have added vs tests for the attributes I introduced. Below is the output of vs test
sudo pytest -s -v --dvsname=vs test_port.py
========================================================= test session starts =========================================================
platform linux2 -- Python 2.7.15rc1, pytest-3.3.0, py-1.8.0, pluggy-0.6.0 -- /usr/bin/python
cachedir: .cache
rootdir: /home/admin/code/sonic-swss/tests, inifile:
collected 6 items

test_port.py::TestPort::test_PortMtu PASSED                                                                                     [ 16%]
test_port.py::test_PortNotification PASSED                                                                                      [ 33%]
test_port.py::test_PortFec PASSED                                                                                               [ 50%]
test_port.py::test_PortPreemp PASSED                                                                                            [ 66%]
test_port.py::test_PortIdriver PASSED                                                                                           [ 83%]
test_port.py::test_PortIpredriver PASSED                                                                                        [100%]

====================================================== 6 passed in 43.30 seconds ======================================================

dgsudharsan:
> We are calling SAI API to configure serdes tuning values thru orchagent, why not thru syncd?
> How we determine thru which entity to access asic? Do we have a general guidance?

Hi @stevenlu99 , The API called in orchagent actually calls the libsairedis API. This will write into ASIC_DB and which in turn messages the syncd. Syncd then deserializes the parameters and invokes the vendor specific SAI API that programs the NPU.
This is the general design followed in all the SAI calls and i followed the same here too.

orchagent/portsorch.h
```diff
@@ -189,6 +189,12 @@ class PortsOrch : public Orch, public Subject
 
     bool getPortOperStatus(const Port& port, sai_port_oper_status_t& status) const;
     void updatePortOperStatus(Port &port, sai_port_oper_status_t status);
+
+    void sai_get_port_serdes_values (const std::string& s,
```
lguohan: > ( [](start = 35, length = 2)

no space here.

orchagent/portsorch.cpp
```diff
@@ -1575,6 +1581,24 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 {
                     an = (int)stoul(fvValue(i));
                 }
+
+                /* Set port serdes Pre-emphasis */
+                if (fvField(i) == "preemphasis")
+                {
+                    sai_get_port_serdes_values(fvValue(i), pre_emphasis);
```
lguohan: >sai_get_port_serdes_values [](start = 20, length = 26)

app db value do not use json format. Please propose a text format and use that in the app db.

for example: 

value,value

orchagent/portsorch.cpp
```diff
@@ -22,6 +22,9 @@
 #include "crmorch.h"
 #include "countercheckorch.h"
 #include "notifier.h"
+#include "swss/json.hpp"
+
+using json = nlohmann::json;
```
lguohan: do not use json for app db attribute value.

orchagent/portsorch.cpp
lguohan: please add the proposed scheme in here.

[https://github.com/Azure/sonic-swss/blob/master/doc/swss-schema.md](https://github.com/Azure/sonic-swss/blob/master/doc/swss-schema.md)


---
In reply to: [282750165](https://github.com/Azure/sonic-swss/pull/821#discussion_r282750165) [](ancestors = 282750165)

orchagent/portsorch.cpp
```diff
@@ -1855,6 +1879,52 @@ void PortsOrch::doPortTask(Consumer &consumer)
                     }
                 }
 
+                if(pre_emphasis.size() != 0)
```
lguohan: >if(p [](start = 16, length = 4)

leave space here.

orchagent/portsorch.cpp
```diff
@@ -3297,3 +3367,42 @@ bool PortsOrch::removeAclTableGroup(const Port &p)
...
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to set serdes attribute %d to port pid:%lx",
+                       attr_id, port_id);
+        return false;
+    }
+    return true;
+}
+
+void PortsOrch::sai_get_port_serdes_values (const std::string& s,
```
lguohan: >sai_ [](start = 16, length = 4)

this has nothing to do with sai.

dgsudharsan:
Ran the test after the changes.
sudo pytest -s -v --dvsname=vs test_port.py 
===================================================== test session starts =====================================================
platform linux2 -- Python 2.7.15rc1, pytest-3.3.0, py-1.8.0, pluggy-0.6.0 -- /usr/bin/python
cachedir: .cache
rootdir: /home/ubuntu/preemp/sonic-swss/tests, inifile:
collected 6 items                                                                                                             

test_port.py::TestPort::test_PortMtu PASSED                                                                             [ 16%]
test_port.py::test_PortNotification PASSED                                                                              [ 33%]
test_port.py::test_PortFec PASSED                                                                                       [ 50%]
test_port.py::test_PortPreemp PASSED                                                                                    [ 66%]
test_port.py::test_PortIdriver PASSED                                                                                   [ 83%]
test_port.py::test_PortIpredriver PASSED                                                                                [100%]

================================================== 6 passed in 40.25 seconds ==================================================

lguohan:
retest this please

lguohan:
retest this please

lguohan:
retest this please

PR #825, opened 2019-03-29
[aclorch]: Add MIRRORv6 support for ACL table

stcheng:
Enable both IPv4 and IPv6 source and destination IP match and
mirror functionality in SONiC.

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -85,6 +85,7 @@ static acl_table_type_lookup_t aclTableTypeLookUp =
     { TABLE_TYPE_L3,        ACL_TABLE_L3 },
     { TABLE_TYPE_L3V6,      ACL_TABLE_L3V6 },
     { TABLE_TYPE_MIRROR,    ACL_TABLE_MIRROR },
+    { TABLE_TYPE_MIRRORV6,  ACL_TABLE_MIRRORV6 },
     { TABLE_TYPE_CTRLPLANE, ACL_TABLE_CTRLPLANE },
     { TABLE_TYPE_DTEL_FLOW_WATCHLIST, ACL_TABLE_DTEL_FLOW_WATCHLIST },
     { TABLE_TYPE_DTEL_DROP_WATCHLIST, ACL_TABLE_DTEL_DROP_WATCHLIST }
@@ -589,12 +590,17 @@ shared_ptr<AclRule> AclRule::makeShared(acl_table_type_t type, AclOrch *acl, Mir
         throw runtime_error("ACL rule action is not found in rule " + rule);
     }
 
-    if (type != ACL_TABLE_L3 && type != ACL_TABLE_L3V6 && type != ACL_TABLE_MIRROR && type != ACL_TABLE_DTEL_FLOW_WATCHLIST && type != ACL_TABLE_DTEL_DROP_WATCHLIST)
+    if (type != ACL_TABLE_L3 &&
+        type != ACL_TABLE_L3V6 &&
+        type != ACL_TABLE_MIRROR &&
+        type != ACL_TABLE_MIRRORV6 &&
+        type != ACL_TABLE_DTEL_FLOW_WATCHLIST &&
+        type != ACL_TABLE_DTEL_DROP_WATCHLIST)
     {
-        throw runtime_error("Unknown table type.");
+        throw runtime_error("Unknown table type");
     }
 
-    /* Mirror rules can exist in both tables*/
+    /* Mirror rules can exist in both tables */
     if (action == ACTION_MIRROR_ACTION)
     {
         return make_shared<AclRuleMirror>(acl, mirror, rule, table, type);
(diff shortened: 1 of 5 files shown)


orchagent/aclorch.cpp
```diff
@@ -948,6 +956,20 @@ bool AclRuleMirror::validateAddMatch(string attr_name, string attr_value)
         return false;
     }
 
+    if (m_tableType == ACL_TABLE_MIRROR &&
+            (attr_name == MATCH_SRC_IPV6 || attr_name == MATCH_DST_IPV6))
+    {
+        SWSS_LOG_ERROR("shuotian: does not match!!!");
```
jleveque: Did you intend to leave these log messages which contain your name?

tests/test_mirror_ipv6.py
```diff
@@ -0,0 +1,351 @@
...
+        # assert acl rule is created
+        tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_ACL_ENTRY")
+        rule_entries = [k for k in tbl.getKeys() if k not in dvs.asicdb.default_acl_entries]
+        assert len(rule_entries) == 1
+
+        (status, fvs) = tbl.get(rule_entries[0])
+        assert status == True
+        for fv in fvs:
+            if fv[0] == "SAI_ACL_ENTRY_ATTR_FIELD_SRC_IPV6":
+                assert fv[1] == "2777::&mask:ffff:ffff:ffff:ffff::"
```
qiluo-msft: '&mask' is weird. Could you explain?

stcheng: this is the serialization done by sairedis library; swss does not have the control here.

stcheng:
Updated with new design

stcheng:
TODO:
1. will update the unit test both both Broadcom and Mellanox scenarios
2. will remove deprecated code in swss for virtual switch platform

orchagent/aclorch.cpp
```diff
@@ -948,6 +954,43 @@ bool AclRuleMirror::validateAddMatch(string attr_name, string attr_value)
...
+     * | MATCH_DST_IPV6   |      √       |              |       √        |
+     * |-----------------------------------------------------------------|
+     * | MARTCH_ETHERTYPE |      √       |      √       |                |
+     * |-----------------------------------------------------------------|
+     */
+
+    if (m_tableType == ACL_TABLE_MIRROR && !m_pAclOrch->isCombinedMirrorV6Table() &&
+            (attr_name == MATCH_SRC_IPV6 || attr_name == MATCH_DST_IPV6))
+    {
+        return false;
```
lguohan: why not have syslog error message here?

orchagent/aclorch.cpp
```diff
@@ -948,6 +954,43 @@ bool AclRuleMirror::validateAddMatch(string attr_name, string attr_value)
...
+    {
+        SWSS_LOG_ERROR("%s match is not supported for the table of type MIRRORV6",
+                attr_name.c_str());
+        return false;
+    }
+
+    if (m_tableType == ACL_TABLE_MIRRORV6 &&
+            (attr_name == MATCH_SRC_IP || attr_name == MATCH_DST_IP || attr_name == MATCH_ETHER_TYPE))
+    {
+        return false;
```
lguohan: why not have syslog error message here?

orchagent/aclorch.cpp
```diff
@@ -948,6 +954,43 @@ bool AclRuleMirror::validateAddMatch(string attr_name, string attr_value)
...
+    }
```
lguohan: is this duplicated code?

orchagent/aclorch.cpp
```diff
@@ -1780,6 +1860,64 @@ void AclOrch::init(vector<TableConnector>& connectors, PortsOrch *portOrch, Mirr
...
+                fvVector.emplace_back(TABLE_TYPE_MIRROR, value);
+                break;
+            case ACL_TABLE_MIRRORV6:
+                fvVector.emplace_back(TABLE_TYPE_MIRRORV6, value);
+                break;
+            default:
+                break;
+        }
+    }
+    m_switchTable.set("switch", fvVector);
```
lguohan: this code really need to go to syncd code. We will address this in the next release.

orchagent/aclorch.cpp
```diff
@@ -2152,7 +2348,14 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
                 continue;
             }
 
-            newRule = AclRule::makeShared(m_AclTables[table_oid].type, this, m_mirrorOrch, m_dTelOrch, rule_id, table_id, t);
+            auto type = m_AclTables[table_oid].type;
+            if (type == ACL_TABLE_MIRROR || type == ACL_TABLE_MIRRORV6)
+            {
+                    type = table_id == m_mirrorTableId ? ACL_TABLE_MIRROR : ACL_TABLE_MIRRORV6;
```
lguohan: identation

orchagent/aclorch.cpp
stcheng: sure i could add some todo to here

stcheng: One issue about moving this to syncd right now is that we currently cannot guarantee the order of syncd and orchagent starting time. e.g. when orchagent starts and at that time syncd hasn't stored the capability information into the database, we will miss this part of the information.

PR #826, opened 2019-04-04
fix bug in vlan member set

jihaix:
**What I did**
    fix bug in vlan member set

**Why I did it**
    *bool Table::get()* strip  symbol "@" by function  *std::string stripSpecialSym()*,  so field "members@"  in configDB's table "VLAN" will not be handled. Vlan member can take effect only  when both table "VLAN" and "VLAN_MEMBER" are configured. When only table "VLAN" is configured, vlan member can't take effect using default tagging_mod "untagged"

**How I verified it**
    Only table "VLAN" in configDB is configured, and vlan member take effect using default tagging_mod "untagged".

**Details if related**

### DIFF
--- a/tests/test_mirror.py
+++ b/tests/test_mirror.py
@@ -96,7 +96,7 @@ class TestMirror(object):
         return self.get_mirror_session_state(name)["status"]
 
     def get_mirror_session_state(self, name):
-        tbl = swsscommon.Table(self.sdb, "MIRROR_SESSION")
+        tbl = swsscommon.Table(self.sdb, "MIRROR_SESSION_TABLE")
         (status, fvs) = tbl.get(name)
         assert status == True
         assert len(fvs) > 0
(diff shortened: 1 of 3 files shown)


stcheng:
we're not using this field anymore; all the members are having separate entries in the table with the name MASTER|MEMBER

lguohan:
@stcheng , then can we remove the code?

lguohan:
@stcheng , ping?

cfgmgr/vlanmgr.cpp
```diff
@@ -290,7 +290,7 @@ void VlanMgr::doVlanTask(Consumer &consumer)
                      */
                     SWSS_LOG_DEBUG("%s mtu %s: Host VLAN mtu setting to be supported.", key.c_str(), mtu.c_str());
                 }
-                else if (fvField(i) == "members@") {
+                else if (fvField(i) == "members") {
```
prsunny: Is this field used anymore?

PR #828, opened 2019-04-08
modify the string from IPV6_Neighbor to IPV6_NEIGHBOR for crmResTypeNameMap.

iris00522:
1. Modify IPV6_Neighbor to IPV6_NEIGHBOR using upper case of full word.
2. Except this variable, the others all use upper case full word.
For consistency, I modified it to use upper case full word.
3. I test the case by python test. I set the high thresholds of ipv6 neighbor to 90 ,  and  set the thresholds ipv6 neighbor type to "free". Setting the value to 1000 for ipv6 neighbors by function setReadOnlyAttr, and check the syslog . The syslog should display as  "IPV6_NEIGHBOR THRESHOLD_EXCEEDED for TH_FREE" .


iris00522:
Hi @prsunny , 
I update some configuration tests for CRM pytests.
Please help to review, thank you.

PR #849, opened 2019-04-22
[vnetorch]: Fix incorrect creation flow for bitmap VNET

volodymyrsamotiy:
* Initialize VNET ID before creating VNET object

**What I did**
Fixed incorrect creation flow for bitmap VNET: initialize VNET ID before creating VNET object.
**Why I did it**
It is a bug in VNET creation flow and it should to be fixed (it causes random failures for "decap" traffic flow).

**How I verified it**
- Manual tests
- Ansible tests
- Virtual switch tests

**Details if related**
N/A


volodymyrsamotiy:
retest this please

PR #854, opened 2019-04-23
[teamsyncd]: Add retry logic in teamsyncd to avoid team hanlder init failure

stcheng:
team_alloc and team_init fail occationally when they start the same time as
teamd instances. Add the retry logic to avoid such cases.

### DIFF
--- a/teamsyncd/teamsync.cpp
+++ b/teamsyncd/teamsync.cpp
@@ -232,12 +232,14 @@ TeamSync::TeamPortSync::TeamPortSync(const string &lagName, int ifindex,
         }
         catch (const system_error& e)
         {
-            SWSS_LOG_WARN("Caught system_error on %s %d:%s", lagName.c_str(),
-                    e.code().value(), e.what());
-
             if (++count == max_retries)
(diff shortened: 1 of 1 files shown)


teamsyncd/teamsync.cpp
```diff
@@ -194,32 +196,52 @@ TeamSync::TeamPortSync::TeamPortSync(const string &lagName, int ifindex,
...
-    {
-        team_free(m_team);
-        m_team = NULL;
-        SWSS_LOG_ERROR("Unable to register port change event");
-        throw system_error(make_error_code(errc::address_not_available),
-                           "Unable to register port change event");
+            if (++count == max_retries)
+            {
+                throw runtime_error("Failed to initialize team handler");
+            }
```
lguohan: make the error message more clear
if (++count == max_retries)
{ 
   throw
}
else
{
   swss_log_warm("Failed to initialize team handler. %s %d:%s, retry %d", lagName.c_str(),
                    e.code().value(), e.what(), count)
}

sleep 1

Issue #850, opened 2019-04-23
vlanmgrd will crashed if there has incorrect type value on cfg_db.json


EmmaXin:
**Description:**
vlanmgrd will crashed if there has incorrect key on cfg_db.json

**Steps to reproduce the issue:**
1. Create /etc/sonic/cfg_db.json with the following content:
```
{  
    "VLAN": {  
        "Vlann10": {  
            "vlanid": "10"  
        } 
    } 
}
```

2. Check the vlanmgrd is running.
```
root@f72df304122e:/# ps -aux | grep vlan
root       232  0.0  0.1  99940  4052 ?        Sl   03:20   0:00 /usr/bin/vlanmgrd
root       252  0.0  0.0  11132  1016 pts/0    S+   03:20   0:00 grep vlan
```

3. Load the cfg.json via sonic-cfggen.
```
root@f72df304122e:/# sonic-cfggen -j /etc/sonic/cfg_db.json --write-to-db
```

4. The vlanmgrd will crashed.
```
root@f72df304122e:/# ps -aux | grep vlan
root       261  0.0  0.0  11132   960 pts/0    S+   03:20   0:00 grep vlan
root@f72df304122e:/#
```

**Describe the results you received:**
The vlanmgrd crashed.

**Describe the results you expected:**
The vlanmgrd shall not crashed.

**Version**
201811 branch , use virtual switch (Docker)

PR #855, opened 2019-04-24
Fix issue #850 vlanmgrd will crashed if there has incorrect type value on cfg_db.json.

linemma:
**What I did**
Fix the issue #850 vlanmgrd will crashed if there has incorrect type value on cfg_db.json.

**Why I did it**
vlanmgrd will crashed cause by the issue.

**How I verified it**
1. Add test cases in test_vlan.py to make sure the bug fixed.
  - Add vlan with incorrect key prefix
    - shall not succeed
    - shall not cause vlanmgrd crashed
  - Add vlan with incorrect value type
    - shall not succeed
    - shall not cause vlanmgrd crashed

2. Execute the same steps on #850 to make sure the bug fixed.

**Details if related**

Signed-off-by: Emma Lin emma_lin@edge-core.com

### DIFF
--- a/cfgmgr/vlanmgr.cpp
+++ b/cfgmgr/vlanmgr.cpp
@@ -238,7 +238,16 @@ void VlanMgr::doVlanTask(Consumer &consumer)
         }
 
         int vlan_id;
-        vlan_id = stoi(key.substr(4));
+        try
+        {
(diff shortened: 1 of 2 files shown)


lguohan:
@jipanyang to review.

PR #858, opened 2019-04-25
[VLAN] Fix issue: vlanmgrd will crashed if there has incorrect key of "vlan member" on cfg_db.json

linemma:
**What I did**
Fix issue: vlanmgrd will crashed if there has incorrect key of "vlan member" on cfg_db.json
For example, load json file with following content via sonic-cfggen:
```
{
    "VLAN_MEMBER": {
    	"Vlanabc|Ethernet0": {
    		"tagging_mode": "untagged"
    	}
    }
}
```
The "Vlanabc" is incorrect.

**Why I did it**

**How I verified it**
1. Add test cases in test_vlan.py to make sure the bug fixed.
  - Add vlan member with incorrect key prefix
    - shall not succeed
    - shall not cause vlanmgrd crashed
  - Add vlan member with incorrect value type
    - shall not succeed
    - shall not cause vlanmgrd crashed
2. After load json file with incorrect content via sonic-cfggen, the vlanmgrd shall not crashed.

**Details if related**

Signed-off-by: Emma Lin <emma_lin@edge-core.com>

### DIFF
--- a/cfgmgr/vlanmgr.cpp
+++ b/cfgmgr/vlanmgr.cpp
@@ -238,7 +238,12 @@ void VlanMgr::doVlanTask(Consumer &consumer)
         }
 
         int vlan_id;
-        vlan_id = stoi(key.substr(4));
+        if (!vlanIdToInt(key.substr(4), &vlan_id))
+        {
+            SWSS_LOG_ERROR("Invalid key format. Not a number after 'Vlan' prefix: %s", key.c_str());
+            it = consumer.m_toSync.erase(it);
+            continue;
+        }
 
         string vlan_alias, port_alias;
         string op = kfvOp(t);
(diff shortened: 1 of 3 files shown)


prsunny:
I'm not sure if we've to consider such invalid configuration. IMO, this is not a valid use-case and adding more code to catch such invalid use-case would be an overkill.

linemma:
I think it shall consider the invalid case because the invalid configuration cause vlanmgrd crashed.

jleveque:
I think it's better to log an error and handle it gracefully than crash.

linemma:
Hi @lguohan, could you help to review this PR? Thanks.

linemma:
> I'm not sure if we've to consider such invalid configuration. IMO, this is not a valid use-case and adding more code to catch such invalid use-case would be an overkill.

Hi @prsunny, I add the codes because the vlanmgrd will crashed when user loads such invalid configuration via sonic-cfggen. I thought the daemon have to work fine during user operating the system. Log an error to inform user when there are something wrong. Please kindly tell me your suggestations. Thanks a lot.

prsunny:
@linemma , we have a plan to introduce ConfigDB validator and the design discussions are in progress. This module can do the necessary checks and push only the valid ones to swss/orchagent. We wouldn't want to check validity at this place.

linemma:
@prsunny, that's great! I'm looking forward to it. Should I close this PR? For this kind use-case, add test cases after the ConfigDB validator finished (if need)?

prsunny:
@linemma , thanks. You can close this and add test cases when validator is ready

PR #61, opened 2016-08-31
fpmsyncd: Re-allocating link cache when interface name cannot be resolved

stcheng:
- This happens when the host interfaces change and then we need to update
  the link cache.

PR #864, opened 2019-04-28
[fpmsyncd] Fix memory leak bug

baiwei0427:
fpmsyncd: Fix memory leak bug

* Allocate and free Netlink memory in RouteSync::getIfName()

Signed-off-by: Wei Bai baiwei0427@gmail.com

### DIFF
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -23,9 +23,7 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
     m_vnet_tunnelTable(pipeline, APP_VNET_RT_TUNNEL_TABLE_NAME, true),
     m_warmStartHelper(pipeline, &m_routeTable, APP_ROUTE_TABLE_NAME, "bgp", "bgp")
 {
-    m_nl_sock = nl_socket_alloc();
-    nl_connect(m_nl_sock, NETLINK_ROUTE);
-    rtnl_link_alloc_cache(m_nl_sock, AF_UNSPEC, &m_link_cache);
+
 }
 
 void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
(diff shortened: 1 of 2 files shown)


fpmsyncd/routesync.cpp
```diff
@@ -23,9 +23,7 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
     m_vnet_tunnelTable(pipeline, APP_VNET_RT_TUNNEL_TABLE_NAME, true),
     m_warmStartHelper(pipeline, &m_routeTable, APP_ROUTE_TABLE_NAME, "bgp", "bgp")
 {
-    m_nl_sock = nl_socket_alloc();
-    nl_connect(m_nl_sock, NETLINK_ROUTE);
-    rtnl_link_alloc_cache(m_nl_sock, AF_UNSPEC, &m_link_cache);
+
```
stcheng: remove empty line

fpmsyncd/routesync.cpp
```diff
@@ -281,23 +279,67 @@ void RouteSync::onVnetRouteMsg(int nlmsg_type, struct nl_object *obj, string vne
...
+        nl_close(sk);
+        nl_socket_free(sk);
+        return false;
+    }
 
     /* Cannot get interface name. Possibly the interface gets re-created. */
-    if (!rtnl_link_i2name(m_link_cache, if_index, if_name, name_len))
+    if (!rtnl_link_i2name(cache, if_index, if_name, name_len))
     {
-        rtnl_link_alloc_cache(m_nl_sock, AF_UNSPEC, &m_link_cache);
```
15216401116: what about using` nl_cache_refill` method instead of  alloc and and free repeatedly?

baiwei0427:
@jipanyang @stcheng  I guess the memory leak bug is due to the reallocation of link cache when interface name cannot be resolved. Is this bug first introduced on https://github.com/Azure/sonic-swss/commit/142673ad6cddf37286ab097c982b31f25c61358c#diff-4a3ea7752832ebd49f989cc95ae042e1?

fpmsyncd/routesync.cpp
```diff
@@ -23,9 +23,6 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
     m_vnet_tunnelTable(pipeline, APP_VNET_RT_TUNNEL_TABLE_NAME, true),
     m_warmStartHelper(pipeline, &m_routeTable, APP_ROUTE_TABLE_NAME, "bgp", "bgp")
 {
-    m_nl_sock = nl_socket_alloc();
-    nl_connect(m_nl_sock, NETLINK_ROUTE);
-    rtnl_link_alloc_cache(m_nl_sock, AF_UNSPEC, &m_link_cache);
```
jipanyang: Could this netlink connection and cache allocation be kept here and reused at RouteSync::getIfName()?  So no need to  create netlink connection and alloc cache for each route message?

stcheng:
@baiwei0427 yes, this pull request #61 is the place that introduced the function `rtnl_link_alloc_cache` in `onMsg` function. Thanks for narrowing down the issue.

One question I have here is that, since you replaced the function `rtnl_link_alloc_cache` with `nl_cache_refill`, do we need to re-allocate socket and cache every time we want to get the interface name, or could it be done during the initialization time and only do the `nl_cache_refill` in getIfName function?

In this case, the memory leak issue could also be addressed, is that correct?

fpmsyncd/routesync.cpp
baiwei0427: @jipanyang This is a very good idea!

PR #867, opened 2019-04-30
[vxlan]add L2-vxlan [#376]

shine4chen:
**What I did**
Add L2 vxlan feature

**Why I did it**
Current vxlan implementation only support VNET scenario. We add  vxlan support on L2 scenario.

**How I verified it**
Tested it in nephos lab

**Details if related**
[vxlanorch]
- When creating vxlan tunnel, create its bridge port. If VXLAN tunnel
without bridge port, FDB MAC can not be learnt from or set to VXLAN
tunnel in ASIC.
- Add the tunnel name map to counter table, so that the ‘show mac’
command can display the FDB learnt from VXLAN tunnel.
- Send command to Linux kernel to create L2 VXLAN tunnel interface.

[fdborch]
* FDB MAC can be learnt from or set to VXLAN tunnel in ASIC


shine4chen:
@jianjundong

orchagent/vxlanorch.cpp
```diff
@@ -328,6 +335,58 @@ create_tunnel_termination(
...
+    sai_status_t status;
+
+    attr.id = SAI_SWITCH_ATTR_DEFAULT_1Q_BRIDGE_ID;
+    attrs_bridge.push_back(attr);
+
+    status = sai_switch_api->get_switch_attribute(gSwitchId, (uint32_t)attrs_bridge.size(), attrs_bridge.data());
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to get default 1Q bridge, rv:%d", status);
+        throw "PortsOrch initialization failure";
```
prsunny: PortsOrch?

orchagent/vxlanorch.cpp
```diff
@@ -350,6 +409,15 @@ bool VxlanTunnel::createTunnel(MAP_T encap, MAP_T decap)
 
         ids_.tunnel_id = create_tunnel(ids_.tunnel_encap_id, ids_.tunnel_decap_id, ip, gUnderlayIfId);
 
+        shared_ptr<DBConnector> m_counter_db = shared_ptr<DBConnector>(new DBConnector(COUNTERS_DB, DBConnector::DEFAULT_UNIXSOCKET, 0));
```
prsunny: Why add to counter table? Can you capture this in Design Doc if it is needed

orchagent/vxlanorch.cpp
```diff
@@ -731,6 +802,26 @@ bool VxlanTunnelMapOrch::addOperation(const Request& request)
     SWSS_LOG_NOTICE("Vxlan tunnel map entry '%s' for tunnel '%s' was created",
                    tunnel_map_entry_name.c_str(), tunnel_name.c_str());
 
+    //ip link add <vxlan_dev_name> type vxlan id <vni> local <src_ip> remote <dst_ip>  dstport 4789
```
prsunny: This must go to VxlanMgr. In fact, handling the ConfigDB entry for L2 Vxlan is already moved to VxlanMgr.

orchagent/fdborch.cpp
```diff
@@ -440,38 +446,68 @@ void FdbOrch::updateVlanMember(const VlanMemberUpdate& update)
 bool FdbOrch::addFdbEntry(const FdbEntry& entry, const string& port_name, const string& type)
 {
     SWSS_LOG_ENTER();
-
+#if 0
```
prsunny: Is this part of the Mac-Move change? Please remove this.

orchagent/fdborch.cpp
```diff
@@ -440,38 +446,68 @@ void FdbOrch::updateVlanMember(const VlanMemberUpdate& update)
...
     memcpy(fdb_entry.mac_address, entry.mac.getMac(), sizeof(sai_mac_t));
     fdb_entry.bv_id = entry.bv_id;
 
     Port port;
-    /* Retry until port is created */
-    if (!m_portsOrch->getPort(port_name, port))
+    sai_object_id_t bridge_port_id = SAI_NULL_OBJECT_ID;
+    VxlanTunnelOrch* tunnel_orch = gDirectory.get<VxlanTunnelOrch*>();
+    VxlanTunnel *tunnel_obj = NULL;
+    if(strncmp(port_name.c_str(),"VTT",3) == 0)
```
prsunny: What is VTT? We would need to define this in a common place.

orchagent/vxlanorch.cpp
jianjundong: VxlanOrch, mistake in writing.

orchagent/vxlanorch.cpp
jianjundong: In doc/L2-vxlan-hld.md, please see chapter 1.6.1, 'Add the tunnel name map to counter table, so that the 'show mac' command can display the FDB learnt from VXLAN tunnel.'.

orchagent/fdborch.cpp
jianjundong: OK, we will remove this soon.

orchagent/fdborch.cpp
jianjundong: VTT is the abbreviation of vxlan tunnel name, such as VTTNL0001. We will define it in common place.Thanks.

orchagent/vxlanorch.cpp
shine4chen: At the time we wrote L2 vxlan code vxlanMgrd was not there. Sure, we will move these codes to vxlanMgrd.

orchagent/vxlanorch.cpp
```diff
@@ -360,6 +428,9 @@ bool VxlanTunnel::createTunnel(MAP_T encap, MAP_T decap)
         ids_.tunnel_term_id = create_tunnel_termination(ids_.tunnel_id, ips, ip, gVirtualRouterId);
         active_ = true;
         tunnel_map_ = { encap, decap };
+
+        ids_.tunnel_bridge_port_id = create_tunnel_bridge_port(ids_.tunnel_id);
```
srj102: VNetBitmapObject::getBridgeInfoByVni also creates a bridge_port for the tunnel and also calls createTunnel. Isnt It is better to remove the bridge_port creation and consolidate bridge port creations in createTunnel to avoid duplicate bridge port creations ?

orchagent/fdborch.cpp
```diff
@@ -440,38 +446,68 @@ void FdbOrch::updateVlanMember(const VlanMemberUpdate& update)
...
     sai_fdb_entry_t fdb_entry;
 
     fdb_entry.switch_id = gSwitchId;
     memcpy(fdb_entry.mac_address, entry.mac.getMac(), sizeof(sai_mac_t));
     fdb_entry.bv_id = entry.bv_id;
 
     Port port;
-    /* Retry until port is created */
-    if (!m_portsOrch->getPort(port_name, port))
+    sai_object_id_t bridge_port_id = SAI_NULL_OBJECT_ID;
```
srj102: The Sonic VxLAN HLD refers to a separate VXLAN_FDB_TABLE for MACs pointing to remote VTEP.  addFdbEntry is to handle entry additions in APP_FDB_TABLE. The current code changes seem to indicate that the VxLAN MACs are populated in the APP_FDB_TABLE. Is there a change in the design ? 

Either way It is better to handle VxLAN FDB entry add/remove in a different function to keep it clean instead of checking for port_name against VTT in multiple places.

orchagent/fdborch.cpp
```diff
@@ -440,38 +446,68 @@ void FdbOrch::updateVlanMember(const VlanMemberUpdate& update)
 bool FdbOrch::addFdbEntry(const FdbEntry& entry, const string& port_name, const string& type)
```
srj102: In the FdbOrch::update() function when a MAC is learnt, getPortByBridgePortId might fail for the tunnel bridge
port id as it is not added in the m_portList. Tunnels could be added to the port list or the learn event for tunnel could be handled based on the bridge port type.

jianjundong: Yes, getPortByBridgePortId will fail for the tunnel bridge port, FdbOrch::update() function only notify observer. FdbOrch::addFdbEntry() skip observer currently, please see the comment '/*TBD: Vxlan tunnel is not in portlist, here skip observer, refine later*/'. The best way is create tunnel port in m_portlist, this is TBD later.

orchagent/fdborch.cpp
jianjundong: Yes, vxlan tunnel MAC is set in APP_FDB_TABLE currently. Packet forwarding is correctly in lab test.
In Sonic VxLAN HLD, I don't see any description about VXLAN_FDB_TABLE, only 'Add VxlanOrch as a member of FDBOrch. For FDB entries learnt on remote VTEP, app-fdb-table shall be updated and programmed to SAI by getting the BridgeIf/RemoteVTEP mapping from VxlanOrch. (TBD)'.

orchagent/vxlanorch.cpp
jianjundong: class VNetBitmapObject is not defined when we writing the codes. Yes, it is better to consolidate bridge port creations in createTunnel. We wiil remove the bridge_port creation in VNetBitmapObject::getBridgeInfoByVni, only record the bridge port id to info.bridge_port_tunnel_id. Thanks.

orchagent/fdborch.cpp
srj102: https://github.com/Azure/SONiC/blob/master/doc/vxlan/Vxlan_hld.md#22-app-db

VXLAN_FDB_TABLE is defined here.

jianjundong: Thanks. I think that VXLAN_FDB_TABLE is used for L3 vxlan scenario, such as EVPN, entries in the table are exchanged by MBGP. The modification here is only for L2 vxlan scenario, entries are learned by ASIC.

orchagent/vxlanorch.cpp
shine4chen: we has fixed it

orchagent/vxlanorch.cpp
shine4chen: we has refine code according to your suggestion.

orchagent/fdborch.cpp
shine4chen: we has fixed it.

orchagent/fdborch.cpp
shine4chen: we has fixed it.

orchagent/fdborch.cpp
shine4chen: The existing fdborch can't listen to VXLAN_FDB_TABLE. We can refine the code after  fdbOrch support listening to vxlan_fdb_table event.

shine4chen:
retest this please

PR #873, opened 2019-05-02
[test_watermark] avoid watermark clear vs test random failure

mykolaf:

**What I did**
Looks like sometimes the clear request is not finished when the verification of the values starts. This is possible due to clear request being propagated to watermarkorch.
I added a little delay and some additional assertions to the test code. 

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/tests/test_watermark.py
+++ b/tests/test_watermark.py
@@ -188,8 +188,9 @@ class TestWatermark(object):
 
         # clear pg shared watermark, and verify that headroom watermark and persistent watermarks are not affected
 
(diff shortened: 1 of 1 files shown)


mykolaf:
If the clear was performed by CLI, we would not face this issue. 

I remember that since the watermarkorch is supposed to clear the values on telemetry timer, it was decided to reuse the logic for user clear request.

Watermark is the only feature that really assigns zero to fields in the DB on clear event.
This is a lot of operations, and we want this to complete as quickly as possible.
I suspect that C++ will do it faster, but I didn't check)

PR #877, opened 2019-05-05
[fdborch] support mac update

shine4chen:
**What I did**
support mac update on fdborch

**Why I did it**
Some apps such as MCLAG need update mac address

**How I verified it**
test it on nephos lab

**Details if related**
If the FDB entry is already exist when the event is ADD, remove the entry first, then add the entry.


prsunny:
@shine4chen , this is a mac-move case. Can we handle it by setting the attribute `SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID` if there is a change? The observers are able to handle port change, so modifications are only required in `fdborch.cpp`. We don't want `#if 0 `in code.

lguohan:
please add vs test cases here.

jianjundong:
> @shine4chen , this is a mac-move case. Can we handle it by setting the attribute `SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID` if there is a change? The observers are able to handle port change, so modifications are only required in `fdborch.cpp`. We don't want `#if 0 `in code.

If we set the attribute `SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID` when the port is a change, this will be a mac-move case. In function FdbOrch::update(), the event SAI_FDB_EVENT_MOVE is equal to SAI_FDB_EVENT_AGED, this will cause MAC deleting.

prsunny:
@jianjundong , so when were remove and add, we also get SAI_FDB_EVENT_AGED and SAI_FDB_EVENT_LEARNED as well? If so, then i'm thinking we may have some issue since you are now removing and adding and SAI events can come later.

jianjundong:
> @jianjundong , so when were remove and add, we also get SAI_FDB_EVENT_AGED and SAI_FDB_EVENT_LEARNED as well? If so, then i'm thinking we may have some issue since you are now removing and adding and SAI events can come later.

We don't find the result of test is not correctly. Although SAI events may be come later, SAI_FDB_EVENT_LEARNED is the last event, in theory it will add the MAC into ASIC_DB by syncd.

prsunny:
> @jianjundong , so when were remove and add, we also get SAI_FDB_EVENT_AGED and SAI_FDB_EVENT_LEARNED as well? If so, then i'm thinking we may have some issue since you are now removing and adding and SAI events can come later.
> 
> We don't find the result of test is not correctly. Although SAI events may be come later, SAI_FDB_EVENT_LEARNED is the last event, in theory it will add the MAC into ASIC_DB by syncd.

Ok, Can you address the `#if 0` part and also the VS test asked by Guohan.

jianjundong:
> Ok, Can you address the `#if 0` part and also the VS test asked by Guohan.

OK, we will do this ASAP. thanks.

tests/test_fdb_update.py
```diff
@@ -0,0 +1,119 @@
...
+    # Get mapping between interface name and its bridge port_id
+    iface_2_bridge_port_id = dvs.get_map_iface_bridge_port_id(dvs.adb)
+    
+    ok, extra = dvs.is_fdb_entry_exists(dvs.adb, "ASIC_STATE:SAI_OBJECT_TYPE_FDB_ENTRY",
+                    [("mac", "52-54-00-25-06-E9"), ("vlan", "2")],
+                    [("SAI_FDB_ENTRY_ATTR_TYPE", "SAI_FDB_ENTRY_TYPE_DYNAMIC"),
+                     ("SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID", iface_2_bridge_port_id["Ethernet4"]),
+                     ('SAI_FDB_ENTRY_ATTR_PACKET_ACTION', 'SAI_PACKET_ACTION_FORWARD')]
+    )
+    assert ok, str(extra)
```
stcheng: could you tear down all the configurations after the test is finished?

leoli-nps:
The vstest has been updated, please review. Thanks!

shine4chen:
> > @jianjundong , so when were remove and add, we also get SAI_FDB_EVENT_AGED and SAI_FDB_EVENT_LEARNED as well? If so, then i'm thinking we may have some issue since you are now removing and adding and SAI events can come later.
> > We don't find the result of test is not correctly. Although SAI events may be come later, SAI_FDB_EVENT_LEARNED is the last event, in theory it will add the MAC into ASIC_DB by syncd.
> 
> Ok, Can you address the `#if 0` part and also the VS test asked by Guohan.
yes, we have fixed '#if 0' part and add vs test cases. Pls help to review it. Thanks

tests/test_fdb_update.py
shine4chen: yes, we have add it. Pls review

prsunny:
@yxieca to review. I'm not very convinced on the sequencing impact it can have. For e.g, you add entry, insert to cache `m_entries`, get AGE event due to the new `removeFdbEntry ` which will remove from cache and then LEARN event which will re-add to cache. Please test for any impact with a larger scale and verify the cache/`show mac` to see if its as expected. Approving from my side.

jipanyang:
Could you add one VS test case that a `dynamic` FDB entry exists, then perform update on it from appDB.

orchagent/fdborch.cpp
```diff
@@ -483,6 +475,11 @@ bool FdbOrch::addFdbEntry(const FdbEntry& entry, const string& port_name, const
     attr.value.s32 = SAI_PACKET_ACTION_FORWARD;
     attrs.push_back(attr);
 
+    if (m_entries.count(entry) != 0) // we already have such entries
+    {
+        removeFdbEntry(entry);
```
yxieca: What if removeFdbEntry failed to remove the entry from SAI? There is a 'FIXME' in that code path. What will happen for the subsequent create_fdb_entry() call into SAI?

Look like the code flow will stop retry below by returning false at the 'FIXME' line?

jianjundong: Yes,  there is a 'FIXME' in the removeFdbEntry() code path. When sai_fdb_api->remove_fdb_entry() found the entry to be removed is not exist in SAI, it must return SAI_STATUS_SUCCESS. In this scenario, when one entry is removed in ASIC, and there is a time difference to notify the event to ASIC_DB.

jianjundong:
> @shine4chen , this is a mac-move case. Can we handle it by setting the attribute `SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID` if there is a change? The observers are able to handle port change, so modifications are only required in `fdborch.cpp`. We don't want `#if 0 `in code.

@prsunny , If we handle it by setting the attribute SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID if there is a change, this will result in an event SAI_FDB_EVENT_MOVE. In current codes, SAI_FDB_EVENT_MOVE is the same as SAI_FDB_EVENT_MOVE. The entry may be not exist because of the time difference between ASIC_DB and ASIC.

jianjundong:
> Could you add one VS test case that a `dynamic` FDB entry exists, then perform update on it from appDB.

@jipanyang , The FDB entry in this VS test case is already 'dynamic'.

jipanyang:
> > Could you add one VS test case that a `dynamic` FDB entry exists, then perform update on it from appDB.
> 
> @jipanyang , The FDB entry in this VS test case is already 'dynamic'.

@jianjundong  I mean the dynamic FDB which comes from syncd, not the one faked in appDB.  You may find example in this link: https://github.com/Azure/sonic-swss/blob/master/tests/test_fdb_warm.py#L134

FDB processing currently has caveat of race condition, it may result in SAI failure and orchagent to exit. Would like to check this case.

leoli-nps:
> Could you add one VS test case that a `dynamic` FDB entry exists, then perform update on it from appDB.

@jipanyang , The corresponding VS test case has been added, please review.

orchagent/fdborch.cpp
stcheng: I would suggest we still keep the original code location to remove the FDB entry at the entrance of this function. Other than this, the PR looks good to me.

prsunny:
> @shine4chen , this is a mac-move case. Can we handle it by setting the attribute SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID if there is a change? The observers are able to handle port change, so modifications are only required in fdborch.cpp. We don't want #if 0 in code.
> 
> If we set the attribute SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID when the port is a change, this will be a mac-move case. In function FdbOrch::update(), the event SAI_FDB_EVENT_MOVE is equal to SAI_FDB_EVENT_AGED, this will cause MAC deleting.

This would undergo a change based on this [PR.](https://github.com/Azure/SONiC/pull/379) Does that mean, this change needs to be revisited?

jianjundong:
> > @shine4chen , this is a mac-move case. Can we handle it by setting the attribute SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID if there is a change? The observers are able to handle port change, so modifications are only required in fdborch.cpp. We don't want #if 0 in code.
> > If we set the attribute SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID when the port is a change, this will be a mac-move case. In function FdbOrch::update(), the event SAI_FDB_EVENT_MOVE is equal to SAI_FDB_EVENT_AGED, this will cause MAC deleting.
> 
> This would undergo a change based on this [PR.](https://github.com/Azure/SONiC/pull/379) Does that mean, this change needs to be revisited?

I think PR #379 can fix the bug of MAC move. There is no conflict between PR #379 and this modification. If we set the attribute SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID when the port is a change, this MAC item may be not exist in SAI and ASIC, such as MAC is aged, and this will cause fail to update.

shine4chen:
I think  in [PR.](https://github.com/Azure/SONiC/pull/379) mac update logic should be same as this PR. Since the dynamic mac addresses may be aged out anytime in asic/sdk. When you use SAI-mac-set API  to update port , the mac address in asic chip may disappear and set operation would fail. So using the deletion and add combination is more safe operation.

lguohan:
retest this please

PR #878, opened 2019-05-05
change direct-route(AKA subnet-route) issuer from intfsorch to routeorch

shine4chen:
**What I did**
change direct route( AKA subnet-route) issuer from intfsorch to routeorch

**Why I did it**
    intfsOrch add directed route by ip address configuration, but routes from 
fpmsyncd got data from routing protocol, which can refer to the state of 
interfaces. When the interface is down intfsorch still adds subnet route . 
It will result in a wrong situation, route with same prefix from route
protocol will be refused to be added to SAI. 
So we change direct route issuer from infsorch to routeorch.

**How I verified it**
Test it on nephos lab

**Details if related**


shine4chen:
This PR will be included in another PR, so close it

PR #879, opened 2019-05-06
Return 0 for CRM counter instead of None if no match

prsunny:
**What I did**
if stats (used/available) is not present, the function returns None. Modified to return 0 if no stats match.

**Why I did it**
Fixing an issue introduced as part of https://github.com/Azure/sonic-swss/pull/830

**How I verified it**
N/A

**Details if related**


prsunny:
retest this please

PR #880, opened 2019-05-08
[vxlanorch] Ambiguous return code for removeNextHopTunnel

gord1306:
**What I did**
If the tunnel is not existed, it should change to return false when remove next hop for a tunnel 

**Why I did it**
For `VxlanTunnelOrch::removeNextHopTunnel` it both checkt tunnel and nexthop existed or not. But in the checking fail case, the first return true and the second return false   

**How I verified it**
Call r`VxlanTunnelOrch::removeNextHopTunnel` for a non existed tunnel for testing.

### DIFF
--- a/orchagent/vxlanorch.cpp
+++ b/orchagent/vxlanorch.cpp
@@ -522,7 +522,7 @@ VxlanTunnelOrch::removeNextHopTunnel(string tunnelName, IpAddress& ipAddr, MacAd
     if(!isTunnelExists(tunnelName))
     {
         SWSS_LOG_ERROR("Vxlan tunnel '%s' does not exists", tunnelName.c_str());
-        return true;
+        return false;
     }
 
     auto tunnel_obj = getVxlanTunnel(tunnelName);


gord1306:
Hi @prsunny, I have another question about the return value in `addOperation` . e.g. VxlanTunnelMapOrch::addOperation, it returns false means the request will not be erased from the queue and retry in next time. Does the retry mechanism have retry count limit? or it always retry?   Thanks

prsunny:
> Hi @prsunny, I have another question about the return value in addOperation . e.g. VxlanTunnelMapOrch::addOperation, it returns false means the request will not be erased from the queue and retry in next time. Does the retry mechanism have retry count limit? or it always retry?   Thanks

It always retry. That's why some cases, we intentionally return `true `so as to erase from Queue

gord1306:
Thanks for explaining, @prsunny. I’m wondering if the request is always retried, has it the possibility to cause queue full? If yes, how to process this case.

prsunny:
Potentially a queue full can happen if all the request are failing. I think it would be abnormal to have a que full of requests that's failing. Mostly we observed some requests fails and it keeps retrying but other requests coming will be processed and freed from queue.

gord1306:
retest this please

gord1306:
retest this please

gord1306:
retest this please

gord1306:
retest this please

gord1306:
retest this please

gord1306:
retest this please

gord1306:
retest this please

PR #892, opened 2019-05-15
[vxlamgr]: Change default return value to true in doTask() to avoid retry the unknown table request

gord1306:
**What I did**
[vxlamgr]: Change default return value to true in doTask() to avoid retry the unknown table request

**Why I did it**
If the default value is false, the request will be retied next time. But those cases which cause return value be false are unknown table request, I thought this kind of requests should only write LOG instead of doing retry 

**How I verified it**
Send incorrect table name request to doTask for vxlanmgr.cpp

### DIFF
--- a/cfgmgr/vxlanmgr.cpp
+++ b/cfgmgr/vxlanmgr.cpp
@@ -165,7 +165,7 @@ void VxlanMgr::doTask(Consumer &consumer)
     auto it = consumer.m_toSync.begin();
     while (it != consumer.m_toSync.end())
     {
-        bool task_result = false;
+        bool task_result = true;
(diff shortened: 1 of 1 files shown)


prsunny:
Looks to be trivial. Its ok either way. ConfigDB validator is in progress where the expectation is swss only gets correct config.

gord1306:
@prsunny, I saw the Configuration validation be list in the community meeting minute. Have you opened any documents for the design spec now? In fact, I am trying to implement gtest unit test cases for vxlanorch and vxlanmgr, and if those validations are done before orchagent, I will consider whether to skip these kinds of test cases

PR #896, opened 2019-05-16
[vxlanorch]: Allow ipv6 src ip for Vxlan tunnel creation

anish-n:
What I did:
Allow IPv6 source IP for Vxlan tunnel creation by removing associated ipv4 version check

Why I did it:
To enable creation of a vxlan tunnel sourced at ipv6 interface

How I verified it:
IPv6 tunnel creation and enhanced vnet_vxlan ansible test case

Signed-off-by: Anish Narsian anish.narsian@microsoft.com


prsunny:
retest this please

prsunny:
retest this please

orchagent/vxlanorch.cpp
```diff
@@ -626,26 +626,20 @@ bool VxlanTunnelOrch::addOperation(const Request& request)
...
+        if(src_ip.isV4()) {
+            dst_ip = IpAddress("0.0.0.0");
+        } else {
+            dst_ip = IpAddress("::");
+        }
     }
     else
     {
         dst_ip = request.getAttrIP("dst_ip");
-        if (!dst_ip.isV4())
```
marian-pritsak: Need to verify that stc_ip family matches dst_ip family, and also matches endpoint_ip family for every tunnel next hop

prsunny: @marian-pritsak , i'm not sure if we need the endpoint_ip family check. It would be slightly complex and may not be a valid use-case. src_ip and dst_ip check can be added.

anish-n: I have pushed the family check for src_ip and dst_ip. @marian-pritsak, can this be approved if it looks good

prsunny:
retest this please

PR #899, opened 2019-05-20
add inc/decCrmAclUsedCounter for SAI_ACL_BIND_POINT_TYPE_SWITCH case.

iris00522:
What I did
Fix issue: When it creates or deletes DTelWatchListTables successfully, it dosen't inc/decCrmAclUsedCounter.

Why I did it
When it creates or deletes DTelWatchListTables successfully, it should inc/decCrmAclUsedCounter with SAI_ACL_BIND_POINT_TYPE_SWITCH case.

How I verified it
Create AclOrch to check whether DTelWatch add call incCrmAclUsedCounter two times or not, and aslo call decCrmAclUsedCounter two times when it delete AclOrch.

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -2779,6 +2779,7 @@ sai_status_t AclOrch::createDTelWatchListTables()
         return status;
     }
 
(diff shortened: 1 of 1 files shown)


iris00522:
retest this please

PR #901, opened 2019-05-20
[vstest]: skip test_AddMaxVlan as it takes almost two hours to finish

lguohan:

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


stcheng:
I recommend to remove this test since virtual switch does not have any kinds of limitations on the number of VLANs.

PR #910, opened 2019-05-23
[test]: Do not remove vEthernet* interfaces

stcheng:
These interfaces are explicitly created for sending packets between
docker and outisde.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -299,7 +299,7 @@ class DockerVirtualSwitch(object):
             fds = re_space.split(l)
             if len(fds) > 1:
                 pname = fds[1].rstrip(":")
-                m = re.compile("(eth|lo|Bridge|Ethernet)").match(pname)
+                m = re.compile("(eth|lo|Bridge|Ethernet|vEthernet)").match(pname)
                 if not m:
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -299,7 +299,7 @@ def net_cleanup(self):
             fds = re_space.split(l)
             if len(fds) > 1:
                 pname = fds[1].rstrip(":")
-                m = re.compile("(eth|lo|Bridge|Ethernet)").match(pname)
+                m = re.compile("(eth|lo|Bridge|Ethernet|vEthernet)").match(pname)
```
lguohan: check this commit. https://github.com/Azure/sonic-swss/commit/c74dc6079f49814ef290b65ed3e80b9cb27b310b#diff-484462fced51d1a06b1d93b4a44dd535

vEthernet is no longer used in the vs docker.

stcheng: ok. then these interfaces shall not be created?

lguohan: no, they should not. in what cases do you see them?

stcheng:
deprecated; was using an old version of create_vnet.sh script

PR #914, opened 2019-05-28
Ignore neighbor entry with BCAST MAC, check SAI status exists

prsunny:
**What I did**
In certain cases, we get neighbor entries with BCAST MAC. This can be safely ignored from programming the hardware

**Why I did it**
Directed bcast entries are added as part of Interface IP address creation. The neighbor update is redundant. 

**How I verified it**
Configure neighbor entry with BCAST mac. 

```
sudo ip neigh add 10.1.1.255 lladdr ff:ff:ff:ff:ff:ff dev Ethernet124
NOTICE swss#neighsyncd: :- onMsg: Broadcast Mac recieved, ignoring for 10.1.1.255
```

**Details if related**

### DIFF
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -393,9 +393,18 @@ bool NeighOrch::addNeighbor(NeighborEntry neighborEntry, MacAddress macAddress)
         status = sai_neighbor_api->create_neighbor_entry(&neighbor_entry, 1, &neighbor_attr);
         if (status != SAI_STATUS_SUCCESS)
         {
-            SWSS_LOG_ERROR("Failed to create neighbor %s on %s, rv:%d",
+            if (status == SAI_STATUS_ITEM_ALREADY_EXISTS)
+            {
+                SWSS_LOG_ERROR("Entry exists: neighbor %s on %s, rv:%d",
                            macAddress.to_string().c_str(), alias.c_str(), status);
-            return false;
+                return true;
+            }
+            else
+            {
+                SWSS_LOG_ERROR("Failed to create neighbor %s on %s, rv:%d",
+                           macAddress.to_string().c_str(), alias.c_str(), status);
+                return false;
+            }
         }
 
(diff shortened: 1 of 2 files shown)


yxieca:
retest this please

orchagent/neighorch.cpp
```diff
@@ -393,9 +393,18 @@ bool NeighOrch::addNeighbor(NeighborEntry neighborEntry, MacAddress macAddress)
         status = sai_neighbor_api->create_neighbor_entry(&neighbor_entry, 1, &neighbor_attr);
         if (status != SAI_STATUS_SUCCESS)
         {
-            SWSS_LOG_ERROR("Failed to create neighbor %s on %s, rv:%d",
+            if (status == SAI_STATUS_ITEM_ALREADY_EXISTS)
+            {
+                SWSS_LOG_ERROR("Entry exists: neighbor %s on %s, rv:%d",
                            macAddress.to_string().c_str(), alias.c_str(), status);
-            return false;
```
yxieca: function addNeighbor() didn't comment the meaning of false/true. This is an issue that I don't think you have to fix in this PR.

But maybe you could add a comment here to state why returning true when error code is ALREADY_EXISTS?

orchagent/neighorch.cpp
```diff
@@ -393,9 +393,18 @@ bool NeighOrch::addNeighbor(NeighborEntry neighborEntry, MacAddress macAddress)
...
+                return true;
+            }
+            else
+            {
+                SWSS_LOG_ERROR("Failed to create neighbor %s on %s, rv:%d",
+                           macAddress.to_string().c_str(), alias.c_str(), status);
+                return false;
+            }
```
yxieca: Personally, I think the else clause should be indented left and left as main body of the parent if. The reason is so that when we hit ERROR, we always return after certain handling.

prsunny: Rest of the file follows the same pattern.

orchagent/neighorch.cpp
prsunny: Added a comment

PR #921, opened 2019-06-03
[warm restart assist] assume vector values could be reordered

yxieca:
**What I did**

When comparing 2 vectors, assume their elements could be re-ordered.

Signed-off-by: Ying Xie <ying.xie@microsoft.com>

**Why I did it**
Without he change, during warm reboot, some neighbor entries will be mistakenly judged as having new value and recreated.

**How I verified it**
With the change, no neighbor entry got recreated due to having 'new value' after warm reboot.

### DIFF
--- a/warmrestart/warmRestartAssist.cpp
+++ b/warmrestart/warmRestartAssist.cpp
@@ -1,4 +1,5 @@
 #include <string>
+#include <algorithm>
 #include "logger.h"
 #include "schema.h"
 #include "warm_restart.h"
@@ -173,7 +174,7 @@ void AppRestartAssist::insertToMap(string key, vector<FieldValueTuple> fvVector,
     else if (found != appTableCacheMap.end())
     {
         // check only the original vector range (exclude cache-state field/value)
-        if(!equal(fvVector.begin(), fvVector.end(), found->second.begin()))
+        if(! contains(found->second, fvVector))
         {
             SWSS_LOG_NOTICE("%s, found key: %s, new value ", m_appTableName.c_str(), key.c_str());
 
@@ -279,4 +280,19 @@ bool AppRestartAssist::checkReconcileTimer(Selectable *s)
         return true;
     }
     return false;
+}
+
+// check if left vector contains all elements of right vector
+bool AppRestartAssist::contains(const std::vector<FieldValueTuple>& left,
+              const std::vector<FieldValueTuple>& right)
+{
+    for (auto const& rv : right)
+    {
+        if (std::find(left.begin(), left.end(), rv) == left.end())
+        {
+            return false;
+        }
+    }
+
+    return true;
 }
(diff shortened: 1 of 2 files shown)


jipanyang:
I would recommend switching to producerstatetabe view comparison mechanism as that being used in teamsyncd. It helps to reduce the maintenance effort, also producerstatetabe view comparison mechanism is more flexible.

yxieca:
@jipanyang can you point me to an example? I need to get the problem under control first. Can we move on then improve?

yxieca:
@jipanyang I found the code you are talking about. This code is not generic, it is written for team management only. There needs to have some refactoring and abstracting before it can be generally in other cases. Right? And the comparison is similar to what I added here too.

warmrestart/warmRestartAssist.cpp
```diff
@@ -173,7 +174,7 @@ void AppRestartAssist::insertToMap(string key, vector<FieldValueTuple> fvVector,
     else if (found != appTableCacheMap.end())
     {
         // check only the original vector range (exclude cache-state field/value)
-        if(!equal(fvVector.begin(), fvVector.end(), found->second.begin()))
+        if(! contains(found->second, fvVector))
```
zhenggen-xu: Do we want to be exact match instead of "contains"?  We may need sort both vectors (exclude the "state") first. 
The vector (exclude the "state") should be exactly the same as before, even if the new vector was a subset, we should consider it as a "change".

yxieca: @zhenggen-xu in principle, I agree with you. But I lost the repeating system that I had last week and therefore I feel comfortable check in this tested code. Then I don't have objection for the version you proposed. At this point, neighbor sync is the only user of this class, and it exactly expects 2 key value pairs that doesn't overlap. So contains works the same as exact match and exclude the state.

I am sorry that I didn't investigate more over the weekend. I thought this class was a generic class used by many clients. That is why I made this change which doesn't have expectation of the location of the state. and doesn't have expectation of the location of the key/value set needs to be matched. So even if the matching set is a subset of the full list at random places, my change would work.

zhenggen-xu: @yxieca You code should work for the cases in the current SONiC as the FVs in the vector are determined.   I don't see this will break anything in the current SONiC code, so I am fine to use this change as you already tested in the field and feel more comfortable.

jipanyang:
> @jipanyang can you point me to an example? I need to get the problem under control first. Can we move on then improve?

View switch design and implementation:
https://github.com/Azure/SONiC/blob/master/doc/warm-reboot/view_switch.md
https://github.com/Azure/sonic-swss-common/pull/247

Usage example:
https://github.com/Azure/sonic-swss/pull/725

We found view switch mechanism is able to handle FV vectors with non-ordered and variable number of fields.  
Not sure about 210811 branch, but for master branch I think we should move to view switch mechanism.

PR #922, opened 2019-06-03
Set LAG mtu value based on kernel netlink msg

prsunny:
**What I did**
Set LAG `mtu `value based on kernel netlink msg along with `admin `and `oper `status 

**Why I did it**
During initialization, the `mtu `attribute is not always present in `LAG_TABLE`. This change ensures that MTU is always set to APP DB as  the same value as configured in kernel. 

**How I verified it**
Check APP DB entry and verify syslog config

**Details if related**

### DIFF
--- a/teamsyncd/teamsync.cpp
+++ b/teamsyncd/teamsync.cpp
@@ -130,6 +130,7 @@ void TeamSync::addLag(const string &lagName, int ifindex, bool admin_state,
     FieldValueTuple m("mtu", to_string(mtu));
     fvVector.push_back(a);
     fvVector.push_back(o);
+    fvVector.push_back(m);
     m_lagTable.set(lagName, fvVector);
 
     SWSS_LOG_INFO("Add %s admin_status:%s oper_status:%s, mtu: %d",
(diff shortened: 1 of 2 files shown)


teamsyncd/teamsync.cpp
```diff
@@ -114,24 +114,26 @@ void TeamSync::onMsg(int nlmsg_type, struct nl_object *obj)
...
-                      bool oper_state)
+                      bool oper_state, unsigned int mtu)
 {
     /* Set the LAG */
     std::vector<FieldValueTuple> fvVector;
     FieldValueTuple a("admin_status", admin_state ? "up" : "down");
     FieldValueTuple o("oper_status", oper_state ? "up" : "down");
+    FieldValueTuple m("mtu", to_string(mtu));
     fvVector.push_back(a);
     fvVector.push_back(o);
```
yxieca: push_back(m)? :-)

prsunny: Oh.. missed it.. good find and thanks for fixing Ying.

PR #924, opened 2019-06-05
[aclorch] unittest by gtest

yehjunying:
**What I did**
Add unit test for aclorch.

**Why I did it**
Do unit test in development phase. The developer can check each variable in each stack frame. And also do memory leakage detection or buffer overrun test. That is different than pytest, it is black box test. But, this is white box test.

**How I verified it**
The new committed testes are all passed.

**Details if related**
The unit test of aclorch have two scopes. One is cover the whole aclorch behaviors. Second is for internal component like AclTable or AclRule.

The first scope. Each test will call doTask() that simulate configDB was changed. All reaction by AclOrch, AclTable or AclReult will save into libvs via SAI. Then we can verify the results via SAI to make sure every operation is correct.

The second scope using spy function to redirect SAI function pointer to c++ std::function for accessing local variable of test instance. It can verify the result directly without any library.

Because many orch will connect Redis when it is created by constructor, so run the test will need to start Redis first. The later tests will not access Redis.

Signed-off-by: JunYing Yeh junying_yeh@edge-core.com

### DIFF
--- a/tests/.clang-format
+++ b/tests/.clang-format
@@ -1,0 +1,104 @@
+#Language:        Cpp
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+#AlignOperands:   false
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: false
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:
+  AfterClass: true
+  AfterControlStatement: true
+  AfterEnum: true
+  AfterFunction:   true
+  AfterNamespace: true
+  AfterObjCDeclaration: true
+  AfterStruct: true
+  AfterUnion: true
+  BeforeCatch: true
+  BeforeElse: true
+  IndentBraces:    false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
+#BreakBeforeInheritanceComma: false
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializers: AfterColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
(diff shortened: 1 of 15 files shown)


yehjunying:
Console output of Jenkins
```
make[1]: Leaving directory '/sonic/sonic-swss'
   debian/rules override_dh_auto_test
make[1]: Entering directory '/sonic/sonic-swss'
./tests/tests
Running main() from gtest_main.cc
[==========] Running 52 tests from 5 test cases.
...
[----------] 1 test from AclTest
[ RUN      ] AclTest.Create_L3_Acl_Table
[       OK ] AclTest.Create_L3_Acl_Table (1 ms)
[----------] 1 test from AclTest (1 ms total)

[----------] 3 tests from AclOrchTest
[ RUN      ] AclOrchTest.ACL_Creation_and_Destorying
[       OK ] AclOrchTest.ACL_Creation_and_Destorying (1001 ms)
[ RUN      ] AclOrchTest.L3Acl_Matches_Actions
[       OK ] AclOrchTest.L3Acl_Matches_Actions (1000 ms)
[ RUN      ] AclOrchTest.L3V6Acl_Matches_Actions
[       OK ] AclOrchTest.L3V6Acl_Matches_Actions (1001 ms)
[----------] 3 tests from AclOrchTest (3002 ms total)

[----------] 4 tests from SaiSpy
[ RUN      ] SaiSpy.CURD
[       OK ] SaiSpy.CURD (0 ms)
[ RUN      ] SaiSpy.Same_Function_Signature_In_Same_API_Table
[       OK ] SaiSpy.Same_Function_Signature_In_Same_API_Table (0 ms)
[ RUN      ] SaiSpy.Same_Function_Signature_In_Different_API_Table
[       OK ] SaiSpy.Same_Function_Signature_In_Different_API_Table (0 ms)
[ RUN      ] SaiSpy.create_switch_and_acl_table
[       OK ] SaiSpy.create_switch_and_acl_table (0 ms)
[----------] 4 tests from SaiSpy (0 ms total)

[==========] 52 tests from 5 test cases ran. (3013 ms total)
[  PASSED  ] 52 tests.
make[1]: Leaving directory '/sonic/sonic-swss'
   create-stamp debian/debhelper-build-stamp
 fakeroot debian/rules binary
dh binary  --with autotools-dev
```

tests/.clang-format
```diff
@@ -0,0 +1,104 @@
+#Language:        Cpp
+AccessModifierOffset: -4
```
lguohan: this is nice, does it get applied in the build or test process? if not, how should we enable it?

yehjunying: First need install clang-format. Before build process, apply clang-format to committed files (or all files) and save the result into a `tmp` folder. Then compare committed files and files in `tmp` folder. If all files are the same, means submitter do clang-format before send PR.

Current I do it locally to make sure every files have the same style, follow `aclorch.cpp` nearly.
But it shall be another task if we want to enable it in build or test process.

stcheng:
thanks for adding the tests

tests/.clang-format
lguohan: I'd like to add enforce the format check on the pr test. can you add clang-format in the sonic-slave-stretch, and add a small script to perform the check. Then, i can run it in the pr test.

PR #928, opened 2019-06-12
[sonic-swss]: Added flap counter and last flap time functionality.

prprakas:
Number of times a link got reset(up/down) has been added under this change, also the last flap timestamp for the link has also been added, they will be updated once the oper status of any port changes. 2 additional fields have been added to appl_db i.e. "flap_counter" and "last_flap_time" which willbe updated when "oper_status" is updated.

Signed-off-by: Prem Prakash prprakash@linkedin.com

**What I did**
Added flap counter and last flap timestamp feature 

**Why I did it**
This would help in solving operational issues faster. Instead of relying on the logs to figure out a bad flapper we can see this feature under status command.

**How I verified it**
Tested it by building up the **docker-orchagent.gz** after my change, changes were required on **sonic-utilities** repo as well to see the output, I will raise a parallel PR for that.

[flap_output.txt](https://github.com/Azure/sonic-swss/files/3279928/flap_output.txt)

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -10,6 +10,8 @@
 #include <algorithm>
 #include <tuple>
 #include <sstream>
+#include <string>
+#include <ctime>
 
 #include <netinet/if_ether.h>
 #include "net/if.h"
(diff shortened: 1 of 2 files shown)


zhenggen-xu:
retest this please

prprakas:
retest this please

prprakas:
retest this please

prprakas:
retest this please

prprakas:
retest this please

prprakas:
retest this please

prprakas:
Please review the changes.

orchagent/portsorch.cpp
```diff
@@ -1239,14 +1241,71 @@ bool PortsOrch::setHostIntfsOperStatus(const Port& port, bool isUp) const
     return true;
 }
 
+void PortsOrch::updateDbPortFlapCounter(const Port& port, vector<FieldValueTuple>& old_tuples, vector<FieldValueTuple>& new_tuples) const
```
stcheng: store the counter in orchagent and update without querying the current value

orchagent/portsorch.cpp
```diff
@@ -1239,14 +1241,71 @@ bool PortsOrch::setHostIntfsOperStatus(const Port& port, bool isUp) const
...
+  flaps = flap_stream.str();
+  FieldValueTuple flap_tuple("flap_counter", flaps);
+  new_tuples.push_back(flap_tuple);
+}
+
+void PortsOrch::updateDbPortLastFlapTime(vector<FieldValueTuple>& new_tuples) const
+{
+  SWSS_LOG_ENTER();
+
+  time_t now = time(0);
```
stcheng: use the `getTimestamp()` function instead

orchagent/portsorch.cpp
```diff
@@ -1239,14 +1241,71 @@ bool PortsOrch::setHostIntfsOperStatus(const Port& port, bool isUp) const
...
-    vector<FieldValueTuple> tuples;
+    vector<FieldValueTuple> old_tuples;
+    vector<FieldValueTuple> new_tuples;
+
+    bool exist = m_portTable->get(port.m_alias, old_tuples);
+    if (!exist)
+    {
+      return;
+    }
+    updateDbPortLastFlapTime(new_tuples);
```
stcheng: move these two functions into `updatePortOperStatus` function instead of here.
under the two lines:
```
updateDbPortOperStatus(port, status);
port.m_oper_status = status;
```

orchagent/portsorch.cpp
zhenggen-xu: @stcheng Do you see any issues for warm-restart/warm-reboot case where the counters/ts in orchagent would be cleared even though from HW perspective things should stay the same before the warm-restart/reboot?

prprakas: @stcheng may be something to check upon before we make the suggested changes, as there might be some discrepancy then.

stcheng: @zhenggen-xu @prprakas we could query the counter number during the port initialization time to get the base number.

orchagent/portsorch.cpp
```diff
@@ -2500,6 +2559,8 @@ bool PortsOrch::initializePort(Port &port)
     {
         port.m_oper_status = SAI_PORT_OPER_STATUS_DOWN;
     }
+    m_portTable->hset(port.m_alias, "flap_counter", "0");
+    m_portTable->hset(port.m_alias, "last_flap", "N/A");
```
lguohan: should go to state db?

orchagent/portsorch.cpp
```diff
@@ -1239,14 +1241,71 @@ bool PortsOrch::setHostIntfsOperStatus(const Port& port, bool isUp) const
     return true;
 }
 
+void PortsOrch::updateDbPortFlapCounter(const Port& port, vector<FieldValueTuple>& old_tuples, vector<FieldValueTuple>& new_tuples) const
+{
+  SWSS_LOG_ENTER();
```
lguohan: tab = 4

orchagent/portsorch.cpp
lguohan: read and set is better as user can then clear this counter, then it can get reset. however, i suggest to use redis incr command which is atomic.

stcheng:
redis incr is nice but it requires much more changes. i suggest a design review first

lguohan:
how to move forward on this one? schedule a call to discuss this feature? @prprakas to schedule?

prprakas:
@lguohan @stcheng  Yeah we can schedule a call to discuss on this. I am based out of Bangalore, India so let me know what is the good time to talk to you

prprakas:
@lguohan will add a vs-test for this, I think that should be good to move ahead with this as per the communication from @zhenggen-xu

lguohan:
@prprakas , can you fix simply comments we have requested, also can you share the ETA for adding vs test?

lguohan:
retest this please

PR #929, opened 2019-06-12
Handle default VLAN create and delete

santoshdoke:
**What I did**
Handle default VLAN creation/deletion in SWSS orch agent. 

**Why I did it**
The default VLAN is created in SAI/driver  during switch initialization. When user tries to create the default VLAN, OrchAgent detects error from SyncD and terminates. Reason is the VLAN already exists. To avoid that, handling the case of default VLAN creation/deletion and skipped the driver call.

**How I verified it**

1.	Create/Delete VLAN 1 
2.	Add/Remove physical ports from VLAN 1
3.	Tagged/Untagged membership for VLAN 1
4.	Add/Remove PortChannel from VLAN 1
5.	Save/Reload VLAN 1 configuration
6.	Ensure that VLAN membership is updated as per config (in driver/SDK)
7.	L2 MAC learning/flushing on VLAN 1
8.	L3 interface configuration on VLAN 1 (ipv4)
9.	ARP learning on VLAN 1 interface
10.	CPU Tx/Rx with VLAN tag 1

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -296,7 +296,21 @@ PortsOrch::PortsOrch(DBConnector *db, vector<table_name_with_pri_t> &tableNames)
     }
 
     m_default1QBridge = attrs[0].value.oid;
-    m_defaultVlan = attrs[1].value.oid;
+    m_defaultVlan_ObjId = attrs[1].value.oid;
+
+    memset(&attr, 0x00, sizeof(attr));
+    attr.id  = SAI_VLAN_ATTR_VLAN_ID;
+ 
+    /* Get the VLAN ID associated with the default VLAN object */
+    status = sai_vlan_api->get_vlan_attribute(m_defaultVlan_ObjId, 1, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to get default VLAN ID, rv:%d", status);
+        throw runtime_error("PortsOrch initialization failure");
+    }    
+ 
+    m_defaultVlan_Id = attr.value.u16;
+
 
     removeDefaultVlanMembers();
     removeDefaultBridgePorts();
@@ -318,7 +332,7 @@ void PortsOrch::removeDefaultVlanMembers()
     attr.value.objlist.count = (uint32_t)vlan_member_list.size();
     attr.value.objlist.list = vlan_member_list.data();
 
-    sai_status_t status = sai_vlan_api->get_vlan_attribute(m_defaultVlan, 1, &attr);
+    sai_status_t status = sai_vlan_api->get_vlan_attribute(m_defaultVlan_ObjId, 1, &attr);
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to get VLAN member list in default VLAN, rv:%d", status);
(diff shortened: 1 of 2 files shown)


orchagent/portsorch.h
```diff
@@ -110,7 +110,8 @@ class PortsOrch : public Orch, public Subject
     Port m_cpuPort;
     // TODO: Add Bridge/Vlan class
     sai_object_id_t m_default1QBridge;
-    sai_object_id_t m_defaultVlan;
+    sai_object_id_t m_defaultVlan_ObjId;
```
stcheng: Thanks for submitting the PR. The logic looks valid to me. Could you modify the type of the current m_defaultVlan to type `Port`? Then you could use `m_vlan_info` to store both the oid and the id of the VLAN. This would prevent having more private variables.

chenkelly:
retest this please

stcheng:
@santoshdoke @chenkelly could you check my comments on the pull request?

orchagent/portsorch.cpp
```diff
@@ -296,7 +296,21 @@ PortsOrch::PortsOrch(DBConnector *db, vector<table_name_with_pri_t> &tableNames)
...
+    /* Get the VLAN ID associated with the default VLAN object */
+    status = sai_vlan_api->get_vlan_attribute(m_defaultVlan.m_vlan_info.vlan_oid, 1, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to get default VLAN ID, rv:%d", status);
+        throw runtime_error("PortsOrch initialization failure");
+    }    
+ 
+    m_defaultVlan.m_vlan_info.vlan_id = attr.value.u16;
+
```
stcheng: could you put the default VLAN into the m_portList? Then we don't need to handle special case in the function `addVlan`.

orchagent/portsorch.cpp
```diff
@@ -2578,7 +2592,17 @@ bool PortsOrch::addVlan(string vlan_alias)
     sai_attribute_t attr;
     attr.id = SAI_VLAN_ATTR_VLAN_ID;
     attr.value.u16 = vlan_id;
-    sai_status_t status = sai_vlan_api->create_vlan(&vlan_oid, gSwitchId, 1, &attr);
+    sai_status_t status = SAI_STATUS_SUCCESS;
```
stcheng: If the default VLAN is in the `m_portList`, there's no need to modify this function.

orchagent/portsorch.cpp
```diff
@@ -2608,12 +2632,16 @@ bool PortsOrch::removeVlan(Port vlan)
         return false;
     }
 
-    sai_status_t status = sai_vlan_api->remove_vlan(vlan.m_vlan_info.vlan_oid);
-    if (status != SAI_STATUS_SUCCESS)
+    /* Do not delete default VLAN from driver, but clear internal state */
+    if (vlan.m_vlan_info.vlan_id != m_defaultVlan.m_vlan_info.vlan_id)
```
stcheng: Move this check into the function `doVlanTask()` before `removeVlan()` function so that the this function's purpose is not touched. Add add a warning if the to-be-removed VLAN is the default VLAN, since it is not supposed to be removed.

lguohan:
retest this please

lguohan:
can you add vs test to verify the app db to asic db translate is protected for vlan 1 case?

PR #934, opened 2019-06-13
[vxlanmgrd]: Fix for vxlanmgrd cannot correctly work after config reload

Pterosaur:
- What I did
  Clear old vxlan devices that were created at last time

- Why I did it
  1. "config reload" command will kill the swss container, which will make the vxlanmgrd loss the cache information. And to execute "ip link add xxx " command will fail after the vxlanmgrd process restart, because the vxlan device has been created.
  2. The Vnet was deleted by modifying config_db.json, and then execute "config reload" command, which will kill vxlanmgrd process. And when vxlanmgrd restart again, it will cannot receive the delete message, because the redis database also was clear after "config reload". So those vxlan devices will become garbage until the kernel restart.

- How I verified it
  Test it in Chassis topology.

### DIFF
--- a/cfgmgr/vxlanmgr.cpp
+++ b/cfgmgr/vxlanmgr.cpp
@@ -539,7 +539,7 @@ bool VxlanMgr::deleteVxlan(const VxlanInfo & info)
 void VxlanMgr::clearAllVxlanDevices()
 {
     std::string stdout;
-    const std::string cmd = std::string("") + IP_CMD + " addr";
+    const std::string cmd = std::string("") + IP_CMD + " link";
     int ret = EXECUTE(cmd, stdout);
     if (ret != 0)
     {
--- a/cfgmgr/vxlanmgr.cpp
+++ b/cfgmgr/vxlanmgr.cpp
@@ -1,4 +1,7 @@
 #include <algorithm>
+#include <regex>
+#include <sstream>
+#include <string>
 #include <net/if.h>
 
 #include "logger.h"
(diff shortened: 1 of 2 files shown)


Pterosaur:
retest this please

Pterosaur:
retest this please

cfgmgr/vxlanmgr.cpp
```diff
@@ -526,5 +536,38 @@ bool VxlanMgr::deleteVxlan(const VxlanInfo & info)
     return true;
 }
 
-
+void VxlanMgr::clearAllVxlanDevices()
+{
+    std::string stdout;
```
stcheng: no need `std::` in cpp files.

cfgmgr/vxlanmgr.cpp
```diff
@@ -1,4 +1,7 @@
 #include <algorithm>
+#include <regex>
+#include <sstream>
+#include <string>
```
stcheng: already in header file

cfgmgr/vxlanmgr.cpp
Pterosaur: I think an explicit namespace(like std) is a good habit, because it can avoid some name conflict. For example, we use **regex** class which was both existed in **std** and **boost**. An explicit declaration, **std::regex**, will make it to be clearer which library we want to use, I think,

cfgmgr/vxlanmgr.cpp
Pterosaur: I add duplicated header file, because I want to decouple the dependency of this module with other modules, which can avoid the compile error of this module if these header files are removed in other files.

cfgmgr/vxlanmgr.cpp
```diff
@@ -526,5 +536,38 @@ bool VxlanMgr::deleteVxlan(const VxlanInfo & info)
     return true;
 }
 
-
+void VxlanMgr::clearAllVxlanDevices()
+{
+    std::string stdout;
+    const std::string cmd = std::string("") + IP_CMD + " addr";
```
prsunny: I think you may want to use `ip link `which will give you a smaller output as compared to `ip addr` and yet you get all interfaces

Pterosaur: Good idea~

Pterosaur:
retest this please

Pterosaur:
retest this please

PR #941, opened 2019-06-19
[intfsorch]: Add unit test with gtest

shikenghua:
**What I did**
Add unit test for intfsorch with gtest

**Why I did it**
This is a follow-up work of adding unit test for sonic-swss after aclorch

**How I verified it**
Build and run the tests. Verified that all newly added test cases and influenced old test cases are passed.

**Details if related**
 Add test cases in intfsorch_ut.cpp.
* Either call doTask() to simulate applDB changes or call public functions directly
* Verify with data stored in intfsorch and/or SAI

Also modify aclorch_ut.cpp according to the following changes:
* Move common extern declarations to mock_orchagent_main.h
* Move consumer->addToSync() wrapper function to portal.h
* Move SAI initialize/uninitialize function calls to ut_helper


stcheng:
hold this PR
we need to separate the vs tests with mock tests that requires SAI

lguohan:
@stcheng, can you elaborate the reason? what do you mean by "separate the vs tests with mock tests that requires SAI".

stcheng:
@lguohan @shikenghua please check https://github.com/Azure/sonic-swss/pull/950

shikenghua:
@stcheng @lguohan submitted a new PR #971. Will close this one.

stcheng:
thanks!

PR #731, opened 2018-12-17
Fix: crash while destructing crmorch

qiluo-msft:
Fix https://github.com/Azure/sonic-buildimage/issues/1991

PR #944, opened 2019-06-19
stop looping on failure to read from syncd

renukamanavalan:
1) if SAI API fails to fetchport info, break from querying for other ports, as most likely they fail too.
2) crmorch, has a bug of two objects claiming ownership of memory, which leads to double delete during exit.
   Let the base class own and, crmorch does not own.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/crmorch.cpp
+++ b/orchagent/crmorch.cpp
@@ -161,7 +161,8 @@ CrmOrch::CrmOrch(DBConnector *db, string tableName):
         m_resourcesMap.emplace(res.first, CrmResourceEntry(res.second, CRM_THRESHOLD_TYPE_DEFAULT, CRM_THRESHOLD_LOW_DEFAULT, CRM_THRESHOLD_HIGH_DEFAULT));
     }
 
-    auto executor = new ExecutableTimer(m_timer.get(), this);
+    // Note: ExecutableTimer will hold m_timer pointer and release the object later
+    auto executor = new ExecutableTimer(m_timer, this);
     Orch::addExecutor("CRM_COUNTERS_POLL", executor);
     m_timer->start();
 }
@@ -394,10 +395,6 @@ void CrmOrch::doTask(SelectableTimer &timer)
     getResAvailableCounters();
     updateCrmCountersTable();
     checkCrmThresholds();
-
-    auto interv = timespec { .tv_sec = m_pollingInterval.count(), .tv_nsec = 0 };
-    timer.setInterval(interv);
-    timer.reset();
 }
 
 void CrmOrch::getResAvailableCounters()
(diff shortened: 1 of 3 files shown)


orchagent/crmorch.cpp
```diff
@@ -394,10 +395,6 @@ void CrmOrch::doTask(SelectableTimer &timer)
     getResAvailableCounters();
     updateCrmCountersTable();
     checkCrmThresholds();
-
-    auto interv = timespec { .tv_sec = m_pollingInterval.count(), .tv_nsec = 0 };
```
prsunny: Why are these lines removed?. This is currently in master as well.

renukamanavalan: Good you asked, as I was unable to add comments for lines to be deleted. 

In short: redundant.

You need to set interval, only when it changes. The change is handled at CrmOrch::handleSetCommand.

This code CrmOrch::doTask(SelectableTimer &timer) is called, each time the timer is fired. No need to keep setting the same value, on each call.

If you look at other agents, like AclOrch::doTask(SelectableTimer &timer), they don't do that, as it is redundant.

prsunny: Yes, originally, set command didn't have it. It was reset during the next timer expire event. Now that we've it in set, we would need to fix this in master as well and later cherry-pick?

orchagent/crmorch.h
```diff
@@ -52,7 +52,7 @@ class CrmOrch : public Orch
 private:
     shared_ptr<DBConnector> m_countersDb = nullptr;
     shared_ptr<Table> m_countersCrmTable = nullptr;
-    shared_ptr<SelectableTimer> m_timer = nullptr;
+    SelectableTimer *m_timer = nullptr;
```
qiluo-msft: >m_timer [](start = 21, length = 7)

Is it an clean cherry-pick [https://github.com/Azure/sonic-swss/pull/731](https://github.com/Azure/sonic-swss/pull/731)?
If yes, prefer cherry-pick

orchagent/crmorch.cpp
qiluo-msft: m_pollingInterval can be modified. So this snippet is needed.

---
In reply to: [295486739](https://github.com/Azure/sonic-swss/pull/944#discussion_r295486739) [](ancestors = 295486739)

renukamanavalan: CrmOrch::handleSetCommand handles any change to m_pollinginterval.

renukamanavalan: Let me drop this change here and put it in masters, which can be cherry picked into Mar, as suggested by Prince.

lguohan:
retest this please

renukamanavalan:
This may be not needed. Because w/o syncd, orchagent is of no value. So saving time in a loop or not, does not make any difference. The TSG for syncd down is to restart swss, which will restart orchagent.

Plan to drop this PR.

PR #945, opened 2019-06-19
Set timer only when interval changes. Not in each firing of the timer.

renukamanavalan:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/crmorch.cpp
+++ b/orchagent/crmorch.cpp
@@ -398,10 +398,6 @@ void CrmOrch::doTask(SelectableTimer &timer)
     getResAvailableCounters();
     updateCrmCountersTable();
     checkCrmThresholds();
-
-    auto interv = timespec { .tv_sec = m_pollingInterval.count(), .tv_nsec = 0 };
-    timer.setInterval(interv);
-    timer.reset();
 }
 
 void CrmOrch::getResAvailableCounters()


qiluo-msft:
I see one benefit of original code. If the time used in doTask is not trivially short, it will schedule the next timer after everything done. It has no risk to schedule more frequent than what could be handled.

renukamanavalan:
If you see that as a solid benefit, we should for one shot timer, instead of periodic one. That way upon firing, you complete the task and start the next timer run.

lguohan:
@renukamanavalan , can you add more descriptions in the pr?

lguohan:
what if the interval changed, how to you reset the timer interval?

renukamanavalan:
1) When CRMOrch starts, it starts a periodic timer for a default interval of 5 mins and saves it in m_pollinginterval.
2) Whenever polling interval changes, it calls CrmOrch::handleSetCommand, which sets the new interval and restarts timer.
3) Whenever timer firing is sensed (through epoll), the CRMOrch::doTask(timer &) is called.

CRMOrch stops & starts the time in each call to doTask. No other Orch client that uses SelectableTimer does that.

Fix: Remove the redundant restart.

If it is intended that we restart the timer upon each firing, then we should not go for periodic timer, and instead call one shot timer. Upon getting called, do the task and start the next one shot timer.

renukamanavalan:
BTW, this is not a fix for any issue/bug. I just happened to notice, while analyzing SelectableTimer::readData failure.

lguohan:
does this cause readdata failure?

renukamanavalan:
Nope! By theory, it should not.

wendani:
We requested some similar change in watermarkorch https://github.com/Azure/sonic-swss/pull/781

PR #946, opened 2019-06-21
[copporch]: Fix polymorphic type error

stcheng:
copporch.cpp: In member function ‘virtual void CoppOrch::doTask(Consumer&)’:
copporch.cpp:607:34: error: catching polymorphic type ‘const class std::out_of_range’ by value [-Werror=catch-value=]
         catch(const out_of_range e)
                                  ^


jleveque:
Retest this please

PR #947, opened 2019-06-24
Set timer only when interval changes. Not in each firing of the timer…

renukamanavalan:
…. (#945)

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


lguohan:
cherry-picked #945

PR #950, opened 2019-06-25
[mock_test]: Move mock tests into a separate folder to separate them from vs tests

stcheng:
when libsaivs is installed with compile and run the mock tests; otherwise, we will
just run the normal vs tests


stcheng:
retest this please

stcheng:
retest this please

PR #952, opened 2019-06-26
[AclOrch]: Fix the acl mirror counter doubled by inactive mirror and …

RichardWu-Hebut:
…active again

        When disabling the mirror session, the aclorch will record
        the current counter, but it should invoke the base class
        function to get the current acl entry counter, otherwise the
        counters will be doubled by invoke the AclRuleMirror::getCounters()

**What I did**

**Why I did it**

**How I verified it**

**Details if related**
When disable the mirror acl entry, the acl rule counter will be doubled by invoke the AclRuleMirror::getCounters(). For example, the acl mirror entry counter is 1000, it will be 2000 when disable the mirror session and enable it again.  The function of AclRuleMirror::getCounters() will return the counters + the current acl entry counter in hardware.

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1133,7 +1133,7 @@ void AclRuleMirror::update(SubjectType type, void *cntx)
     else
     {
         // Store counters before deactivating ACL rule
-        counters += getCounters();
+        counters += AclRule::getCounters();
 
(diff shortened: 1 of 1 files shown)


stcheng:
retest this please

RichardWu-Hebut:
> retest this please
It is already tested before I submit this PR. What further information is required?

stcheng:
@RichardWu-Hebut there're new fixes merged after the previous check.

PR #953, opened 2019-06-26
[MirrorOrch]: Init the next hop ip with 0 instead of default constructor

RichardWu-Hebut:
Otherwise, in the getgetNeighborInfo function, it will check whether
              the nexthop is zero, but it will return false according to
              IpAddress::isZero(), then it will try to query the
              neighbor orch about this invalid ip address.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**
In current implement， in the getgetNeighborInfo function, it will check whether the nexthop is zero, but it will return false according to  IpAddress::isZero(), then it will try to query the neighbor orch about this invalid ip address.

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -786,7 +786,7 @@ void MirrorOrch::updateNextHop(const NextHopUpdate& update)
         }
         else
         {
-            session.nexthopInfo.nexthop = IpAddress();
+            session.nexthopInfo.nexthop = IpAddress(0);
         }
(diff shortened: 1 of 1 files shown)


stcheng:
thanks for fixing this

stcheng:
retest this please

stcheng:
retest this please

stcheng:
```
test_warm_reboot.py::test_system_warmreboot_neighbor_syncup FAILED       [ 97%]
```

PR #906, opened 2019-05-22
[aclorch]: Add MIRROR_DSCP table type

stcheng:
Add MIRROR_DSCP table to support creating an ACL mirro table
that only matches DSCP value/mask.

Signed-off-by: Shu0T1an ChenG <shuche@microsoft.com>

PR #965, opened 2019-07-02
Backport ACL/MIRROR features to 201811 branch

stcheng:


stcheng:
#906 is missing; it is cherry-picked later

jangidrk:
Hi, 
could you help in explaining if Warm-reboot for Policer is supported and how? 
create_policer(  &policer_id, gSwitchId, (uint32_t)attrs.size(), attrs.data()) 
there could be two differnet policer_id with same attribute.

stcheng:
@jangidrk yes; but that's fine; we will assign back the policers based on which other objects they are attached to; if they are two stand-alone policers with purely same attributes; they will be assigned back randomly after warm reboot.

PR #966, opened 2019-07-03
[aclorch]: Fix matching MIRROR_DSCP throws unnecessary errors

stcheng:
First check if the string to compare is a match action and then
check if the match action is MATCH_DSCP.


stcheng:
```
Jul  3 18:46:22.656607 str-s6100-acs-5 ERR swss#orchagent: :- validateAddMatch: MIRROR_ACTION match is not supported for the table of type MIRROR_DSCP
```
this above message shall not appear since MIRROR_ACTION is not a match action.

PR #971, opened 2019-07-11
[intfsorch]: Add unit test with gtest

shikenghua:
**What I did**
Add unit test for intfsorch with gtest

**Why I did it**
This is a follow-up work of adding unit test for sonic-swss after aclorch

**How I verified it**
Build and run the tests. Verified that all newly added test cases and influenced old test cases are passed.

**Details if related**
Add test cases in intfsorch_ut.cpp.
- Either call doTask() to simulate applDB changes or call public functions directly
- Verify with data stored in intfsorch and/or SAI

Also modify aclorch_ut.cpp according to the following changes:
- Move common extern declarations to mock_orchagent_main.h
- Move consumer->addToSync() wrapper function to portal.h
- Move SAI initialize/uninitialize functions to ut_helper

### DIFF
--- a/tests/mock_tests/Makefile.am
+++ b/tests/mock_tests/Makefile.am
@@ -13,40 +13,41 @@ endif
 CFLAGS_GTEST =
 LDADD_GTEST = -L/usr/src/gtest
 
-tests_SOURCES = swssnet_ut.cpp request_parser_ut.cpp aclorch_ut.cpp saispy_ut.cpp \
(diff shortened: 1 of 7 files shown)


shikenghua:
retest this please

lguohan:
retest this please

qiluo-msft:
Please help rebase it to latest master, and I have some patch for you, which build the subdir, and test it during build process.


[0001-Fix-mock_tests-build.patch.txt](https://github.com/Azure/sonic-swss/files/3786452/0001-Fix-mock_tests-build.patch.txt)

PR #977, opened 2019-07-17
Change nexthop key to ip & ifname

tylerlinp:
**What I did**
Changed nexthop key including ifname.
Made a supplement to vs test of ACL redirect.

**Why I did it**
To implement VRF, single IP cannot indicate nexthop uniquely.

**How I verified it**
vs test

**Details if related**


tylerlinp:
[https://github.com/Azure/sonic-swss/pull/943](url)

orchagent/intfsorch.cpp
```diff
@@ -78,10 +78,22 @@ sai_object_id_t IntfsOrch::getRouterIntfsId(const string &alias)
 {
     Port port;
     gPortsOrch->getPort(alias, port);
-    assert(port.m_rif_id);
     return port.m_rif_id;
 }
 
+string IntfsOrch::getRouterIntfsAlias(const IpAddress &ip)
```
prsunny: This function is ambiguous as same IP can be existing on multiple interfaces in a multi VRF scenario. I don't think we can have a function like this. Is it for some backward compatibility?

orchagent/mirrororch.cpp
```diff
@@ -368,7 +368,7 @@ bool MirrorOrch::getNeighborInfo(const string& name, MirrorEntry& session)
     // 3) Otherwise, return false.
     if (!m_neighOrch->getNeighborEntry(session.dstIp,
                 session.neighborInfo.neighbor, session.neighborInfo.mac) &&
-            (session.nexthopInfo.nexthop.isZero() ||
+            (session.nexthopInfo.nexthop.ip_address.isZero() ||
```
prsunny: As with acl redirect, mirror also would require a schema change to include interface. If it is for backward compatibility, then acl redirect doesn't seem to have backward compatibility

orchagent/neighorch.h
```diff
@@ -7,42 +7,68 @@
 #include "intfsorch.h"
 
 #include "ipaddress.h"
+#include "tokenize.h"
 
 #define NHFLAGS_IFDOWN                  0x1 // nexthop's outbound i/f is down
 
-struct NeighborEntry
+struct NextHopKey
```
prsunny: To me it looks better to be named as `NeighborKey `than `NextHopKey`.

orchagent/neighorch.cpp
```diff
@@ -271,6 +272,18 @@ bool NeighOrch::getNeighborEntry(const IpAddress &ipAddress, NeighborEntry &neig
     return false;
 }
 
+bool NeighOrch::getNeighborEntry(const IpAddress &ipAddress, NeighborEntry &neighborEntry, MacAddress &macAddress)
```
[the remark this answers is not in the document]
tylerlinp: Yes, now only mirror session use it. There is dest ip given, we have to find out ifname. To support vrf for mirror, it would be ip and vrf in configuration, then add a vrf argument.

orchagent/intfsorch.cpp
```diff
@@ -78,10 +78,22 @@ sai_object_id_t IntfsOrch::getRouterIntfsId(const string &alias)
...
+{
+    for (const auto &it_intfs: m_syncdIntfses)
+    {
+        for (const auto &prefixIt: it_intfs.second.ip_addresses)
+        {
+            if (prefixIt.isAddressInSubnet(ip))
+                return it_intfs.first;
```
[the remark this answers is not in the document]
tylerlinp: OK. I will add later.

orchagent/routeorch.cpp
```diff
@@ -341,28 +341,32 @@ void RouteOrch::doTask(Consumer& consumer)
...
 
                 if (fvField(i) == "ifname")
-                    alias = fvValue(i);
+                    aliases = fvValue(i);
             }
+            vector<string> ipv = tokenize(ips, ',');
+            vector<string> alsv = tokenize(aliases, ',');
 
-            // TODO: set to blackhold if nexthop is empty?
-            if (ip_addresses.getSize() == 0)
```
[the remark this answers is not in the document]
tylerlinp: Introduce 0.0.0.0 is to prepare for handling directly connected routes. It is better to keep ips and aliases same size, in general handling tokenize("", ',') will get empty vector. Zero ip would be used to construct NextHopGroupKey.

fpmsyncd/routesync.cpp
```diff
@@ -325,6 +325,17 @@ string RouteSync::getNextHopGw(struct rtnl_route *route_obj)
             nl_addr2str(addr, gw_ip, MAX_ADDR_SIZE);
             result += gw_ip;
         }
+        else
+        {
+            if (rtnl_route_get_family(route_obj) == AF_INET)
```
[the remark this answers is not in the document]
tylerlinp: As above.

orchagent/routeorch.h
```diff
@@ -9,34 +9,155 @@
...
+        return m_nexthops;
+    }
+
+    inline size_t getSize() const
+    {
+        return m_nexthops.size();
+    }
+
+    inline bool operator<(const NextHopGroupKey &o) const
+    {
```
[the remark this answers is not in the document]
tylerlinp: Maybe useless, NextHopGroupKey just extended from IpAddresses.

tests/test_acl.py
```diff
@@ -1206,3 +1206,139 @@ def test_AclRuleIcmpV6(self, dvs, testlog):
...
+        tbl.set("test_redirect", fvs)
+
+        time.sleep(2)
+
+        # create acl rule
+        tbl = swsscommon.Table(self.cdb, "ACL_RULE")
+        fvs = swsscommon.FieldValuePairs([
+                                        ("priority", "100"),
+                                        ("L4_SRC_PORT", "65000"),
+                                        ("PACKET_ACTION", "REDIRECT:10.0.0.2|Ethernet4")])
```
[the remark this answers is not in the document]
tylerlinp: OK. I will update later.

orchagent/intfsorch.cpp
tylerlinp: This function will add a vrf argument to support VRF. The VRF PR includes this. Mirror session is using it now, in getNeighborEntry.

orchagent/mirrororch.cpp
tylerlinp: Now mirror haven't support VRF, so have no schema change. Maybe mirror should add not interface but vrf in future, as the dest ip can be a remote host not neighbor and then to get nexthop from routes.

orchagent/neighorch.h
tylerlinp: Now NeighborEntry is same as NextHopKey, I think NeighborEntry can be named as NeighborKey, but I haven't changed it. It is better to distinguish nexthop and neighbor, in some cases just need nexthop although now they are created together.

prsunny:
@tylerlinp , we need to make sure that an old `config_db.json` with ACL redirect/Mirror session without `ifname ` in schema also shall work with this changes. It is expected to work as if in default VRF.

shine4chen:
> @tylerlinp , we need to make sure that an old `config_db.json` with ACL redirect/Mirror session without `ifname ` in schema also shall work with this changes. It is expected to work as if in default VRF.

@prsunny Sure, we will revise our HLD and code. In fact We doesn't change mirror session schema in current VRF PR. So mirror backward compatibility is not issue. For Acl redirect  we can add support for single ip without interface name.

orchagent/intfsorch.cpp
prsunny: Can you add VRF argument to this function in this PR? It can pass `default vrf` from the caller.

orchagent/neighorch.cpp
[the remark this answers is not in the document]
prsunny: If `getNeighborEntry `is invoked with only IP address and not `Ifname`, then `getRouterIntfsAlias `must only check for interfaces in default VRF.

orchagent/neighorch.h
prsunny: My thought is, it is Neighorch and this is the key structure. So with NeighborKey, it aligns with the neighbor orch functionality. Nexthop is created _from_ the Neighbor. Just a suggestion to think, not a blocker

orchagent/routeorch.h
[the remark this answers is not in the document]
prsunny: Please remove the unused functions

orchagent/intfsorch.cpp
tylerlinp: OK, actually I had removed it just for this PR.

orchagent/neighorch.cpp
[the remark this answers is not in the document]
tylerlinp: That's right.

orchagent/routeorch.h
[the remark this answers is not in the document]
tylerlinp: It is used actually. NextHopGroupKey is used as map key. Compare function is needed for map key. So overload operator< is needed.

prsunny:
retest this please

prsunny:
retest this please

orchagent/neighorch.h
stcheng: I agree with Prince to replace the `NextHopKey` with `NeighborKey` and also remove the `NeighborEntry`. Otherwise it is confusing to have various class names.

stcheng: Also, since the two classes `NextHopKey` and `NextHopGroupKey` are very large. Please move them outside the header and create a separate header file and the implementation file.

orchagent/neighorch.h
```diff
@@ -7,42 +7,83 @@
...
     IpAddress           ip_address;     // neighbor IP address
     string              alias;          // incoming interface alias
 
-    bool operator<(const NeighborEntry &o) const
+    NextHopKey() = default;
+    NextHopKey(const std::string &ipstr, const std::string &alias) : ip_address(ipstr), alias(alias) {}
+    NextHopKey(const IpAddress &ip, const std::string &alias) : ip_address(ip), alias(alias) {}
+    NextHopKey(const std::string &str)
+    {
+        if (str.find(',') != string::npos)
```
stcheng: Why do you need this check?

tests/test_acl.py
```diff
@@ -1215,6 +1215,141 @@ def test_AclRuleIcmpV6(self, dvs, testlog):
 
         self.remove_acl_table(acl_table)
 
+    def set_admin_status(self, interface, status):
```
stcheng: Could you move these helper functions into `conftest.py` file?

tests/test_acl.py
```diff
@@ -1215,6 +1215,141 @@ def test_AclRuleIcmpV6(self, dvs, testlog):
...
+        fvs = swsscommon.FieldValuePairs([("neigh", mac)])
+        tbl.set(interface + "|" + ip, fvs)
+        time.sleep(1)
+
+    def remove_neighbor(self, interface, ip):
+        tbl = swsscommon.Table(self.cdb, "NEIGH")
+        tbl._del(interface + "|" + ip)
+        time.sleep(1)
+
+    def test_AclRuleRedirectToNexthop(self, dvs, testlog):
```
stcheng: Since this is to test a new functionality, could you move this portion into a separate PR?
It could be better to add some unit tests that covers the current changes

orchagent/neighorch.h
shine4chen: I think  `NexthopKey` semantic is more than `NeighborKey`.  A Route has  nexthop but not neigh. Neigh can reuse `NexthopKey` as `NeighborKey`. But route can't use `NeighborKey` as nexthop.
Spliting `Nexthopkey` and `NextHopGroupKey`  is good idea. We will fix it soon.

orchagent/neighorch.h
shine4chen: It originally come from `ipAddress` class. Some apps use multi-ipaddresses string to create `NexthopKey` class. It needs to throw an exception to reminder the apps to use `NexthopGroupKey` class

tests/test_acl.py
shine4chen: Yes, we will do it.

tests/test_acl.py
shine4chen: In fact ACL redirect is the existing function. But no vs test cases can cover it. Since we change nexthop key we add a test case to verify if this function works.

prsunny:
retest this please

debian/changelog
```diff
@@ -1,5 +1,9 @@
 sonic (1.0.0) stable; urgency=medium
 
+  * next-hop key add ifname
```
stcheng: please keep the changelog as it is since right now we're not updating this file routinely.

stcheng:
could you also resolve the conflicts? thanks!

debian/changelog
tylerlinp: New version dpkg-deb use the newest time in chaneglog to set deb files timestamp. There is a problem that new files with same size and timestamp cannot be saved to image in process of PR build, so in this PR fpmsynd cannot be updated and many test cases failed. I add changelog just for building swss.deb with newer timestamp and pass tests.
THE PROBLEM MUST BE RESOLVED, or else the VS test become meaningless, for the running version may not be the codes compiled.

shine4chen:
@prsunny @stcheng we have resolve the conflict and address your comment. It passed all vs test cases.

debian/changelog
stcheng: are you referring to the VS test running under this pull request? and the fpmsyncd is not replaced with the latest binary? if that is the case as you described, I'll check.

shine4chen: Yes，I have forward the email which describes this issue detailly to you. It is common issue, not  just on this PR.

prsunny:
retest this please

tylerlinp:
retest this please

tests/test_acl.py
tylerlinp: I have done this in PR #943.

tests/test_acl.py
tylerlinp: I have removed the test in this PR and revised it in PR #943. So all new unit tests are added in PR #943.

debian/changelog
stcheng: could you check right now if the binaries could overwrite the old ones?
I have added https://github.com/Azure/sonic-build-tools/commit/e9e39576e2f964538cb6648cf3127017b6745f2a to fix this issue.

PR #980, opened 2019-07-18
[vnetorch]: Set default VxLAN encap TTL value

volodymyrsamotiy:

**What I did**
Implemented logic to set default VxLAN encap TTL value
**Why I did it**
On different platforms default VxLAN encap TTL value is different, so we need to set same value for all
**How I verified it**
Executed "VNET" ansible test and verified it passed.
**Details if related**

### DIFF
--- a/orchagent/vxlanorch.cpp
+++ b/orchagent/vxlanorch.cpp
@@ -228,7 +228,8 @@ create_tunnel(
     sai_object_id_t tunnel_encap_id,
     sai_object_id_t tunnel_decap_id,
     sai_ip_address_t *src_ip,
-    sai_object_id_t underlay_rif)
+    sai_object_id_t underlay_rif,
+    sai_uint8_t encap_ttl=0)
 {
     sai_attribute_t attr;
     std::vector<sai_attribute_t> tunnel_attrs;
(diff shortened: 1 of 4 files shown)


volodymyrsamotiy:
https://github.com/Azure/sonic-mgmt/pull/1017

orchagent/vxlanorch.cpp
```diff
@@ -228,7 +228,8 @@ create_tunnel(
     sai_object_id_t tunnel_encap_id,
     sai_object_id_t tunnel_decap_id,
     sai_ip_address_t *src_ip,
-    sai_object_id_t underlay_rif)
+    sai_object_id_t underlay_rif,
+    sai_uint8_t encap_ttl=0)
```
stcheng: no need to specify the default value here, as already specified in the header file

orchagent/vxlanorch.h
```diff
@@ -82,7 +82,7 @@ class VxlanTunnel
         return active_;
     }
 
-    bool createTunnel(MAP_T encap, MAP_T decap);
+    bool createTunnel(MAP_T encap, MAP_T decap, uint8_t encap_ttl=0);
```
stcheng: why do you set the default argument with value 0 instead of `VXLAN_ENCAP_TTL`?

orchagent/vxlanorch.cpp
```diff
@@ -264,6 +265,17 @@ create_tunnel(
         tunnel_attrs.push_back(attr);
     }
 
+    if (encap_ttl != 0)
```
stcheng: since you're not exposing any places for user to change the `encap_ttl` value, why do you need to expose the capability of setting it via different functions while providing all the functions with default value? could you directly set the TTL mode to pipe model and the value to the default value?

prsunny: @stcheng , the VxlanTunnel is created for different scenarios. For the VNET use-case, we would want the Tunnel to have default 128 TTL. So I think it is better to pass this as parameter.

orchagent/vxlanorch.cpp
volodymyrsamotiy: This is ```static``` helper function to create tunnel in SAI, so we need default value here (in header file default value specified for "create tunnel" public APIs)

orchagent/vxlanorch.h
volodymyrsamotiy: As Prince mentioned in another comment, VxLAN tunnel is created for different scenarios and for the VNET use-case we have default TTL set to 128 but for other use-cases all should work as before.

stcheng:
retest this please

PR #990, opened 2019-07-23
Rename Port member function isPortReady() to allPortsReady()

wendani:

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


prsunny:
Why this change?

wendani:
There is some idea to enable port one by one once the buffer profile is configured on it. Just make the current logic approach clear.

stcheng:
🚀

PR #993, opened 2019-07-25
[aclorch]: unbind port from acltable when remove port

ouxlwhu:
**What I did**
1.aclorch unbind port from acltable when process port remove notification
2.AclTable::unbind set ports[portOid] to NULL after remove_acl_table_group_member
3.add AclTable::unlink() to cleanup AclTable::ports.
4.portsorch notify PORTCHANGE when remove vlan and add vlan
5.adjust the execution sequence between m_portList erase and PORT_CHANGE notify

**Why I did it**
When the port(port/lag/vlan) is removed, the port and acltable are not unbound

**How I verified it**
debug print and sai log

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2732,6 +2732,9 @@ bool PortsOrch::addVlan(string vlan_alias)
     vlan.m_members = set<string>();
     m_portList[vlan_alias] = vlan;
     m_port_ref_count[vlan_alias] = 0;
+    /* notify aclorch to bind vlan to acl table */
+    PortUpdate update = { vlan, true };
+    notify(SUBJECT_TYPE_PORT_CHANGE, static_cast<void *>(&update));
 
     return true;
 }
@@ -2753,6 +2756,9 @@ bool PortsOrch::removeVlan(Port vlan)
         SWSS_LOG_ERROR("Failed to remove non-empty VLAN %s", vlan.m_alias.c_str());
         return false;
     }
+    /* notify aclorch to unbind vlan from acl table */
+    PortUpdate update = { vlan, false };
+    notify(SUBJECT_TYPE_PORT_CHANGE, static_cast<void *>(&update));
 
     sai_status_t status = sai_vlan_api->remove_vlan(vlan.m_vlan_info.vlan_oid);
     if (status != SAI_STATUS_SUCCESS)
@@ -2942,6 +2948,9 @@ bool PortsOrch::removeLag(Port lag)
         SWSS_LOG_ERROR("Failed to remove LAG %s, it is still in VLAN", lag.m_alias.c_str());
         return false;
     }
+    /* notify PORTCHANGE before remove port*/
+    PortUpdate update = { lag, false };
+    notify(SUBJECT_TYPE_PORT_CHANGE, static_cast<void *>(&update));
 
     sai_status_t status = sai_lag_api->remove_lag(lag.m_lag_id);
     if (status != SAI_STATUS_SUCCESS)
@@ -2956,9 +2965,6 @@ bool PortsOrch::removeLag(Port lag)
 
     m_portList.erase(lag.m_alias);
     m_port_ref_count.erase(lag.m_alias);
-
-    PortUpdate update = { lag, false };
-    notify(SUBJECT_TYPE_PORT_CHANGE, static_cast<void *>(&update));
 
     return true;
 }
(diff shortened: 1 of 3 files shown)


stcheng:
Could you add some unit test to cover these scenarios?

ouxlwhu:
> Could you add some unit test to cover these scenarios?

OK, I'll try it

orchagent/portsorch.cpp
```diff
@@ -2754,6 +2757,9 @@ bool PortsOrch::removeVlan(Port vlan)
         SWSS_LOG_ERROR("Failed to remove non-empty VLAN %s", vlan.m_alias.c_str());
         return false;
     }
+    /* notify aclorch to unbind vlan from acl table */
```
zhenggen-xu: any reason we need unbind vlan before remove vlan from HW?  What if the remove_vlan API failed later? See LAG case as well.

orchagent/portsorch.cpp
```diff
@@ -2943,6 +2949,9 @@ bool PortsOrch::removeLag(Port lag)
         SWSS_LOG_ERROR("Failed to remove LAG %s, it is still in VLAN", lag.m_alias.c_str());
         return false;
     }
+    /* notify PORTCHANGE before remove port*/
```
zhenggen-xu: Same question, if we notify other classes before we remove_lag on HW, what if the remove_lag failed?

tests/test_vlan.py
```diff
@@ -463,3 +537,28 @@ def test_RemoveVlanWithRouterInterface(self, dvs, testlog):
...
+        # create ACL table
+        self.create_acl_table(dvs, "VLAN_ACL_TABLE", "Vlan1234")
+
+        # create Vlan
+        dvs.create_vlan("1234")
+
+        # check ASIC table
+        self.check_asic_table_existed(dvs,"1234")
+
+        # remove port channel
```
zhenggen-xu: comment should be "remove vlan"

tests/test_vlan.py
```diff
@@ -12,6 +12,80 @@ def check_syslog(self, dvs, marker, process, err_log, vlan_str, expected_cnt):
         (exitcode, num) = dvs.runcmd(['sh', '-c', "awk \'/%s/,ENDFILE {print;}\' /var/log/syslog | grep %s | grep \"%s\" | grep -i %s | wc -l" % (marker, process, err_log, vlan_str)])
         assert num.strip() == str(expected_cnt)
 
+    def create_acl_table(self, dvs, table_name, ports):
```
zhenggen-xu: I think it would be better to move these functions to a new file like test_acl_vlan.py. So that we do all ACL related test cases there.  I suggest we add one more test case where we create VLAN and ACL table after we removed VLAN and ACL Table in the first test case.

orchagent/portsorch.cpp
ouxlwhu: “any reason we need unbind vlan before remove vlan from HW?”
----Sorry, I did not consider the remove_vlan API failure before.
I just thought it was better to notify port change before remove acl table group from HW. In addition, if we want to unbind, it may be better to call set_vlan_attribute API with attribute SAI_VLAN_ATTR_INGRESS_ACL setting to NULL, so it maybe better to unbind before remove vlan from HW.
But, it doesn't call set_vlan_attribute API with attribute SAI_VLAN_ATTR_INGRESS_ACL setting to NULL when unbind acl table, so we need not to unbind vlan before remove vlan from HW. In the future, the set_vlan_attribute may be called for making AclTable::bind and AclTable::unbind symmetric.(When I modified this code, I found a comment "// TODO: make bind/unbind symmetric" before AclTable::bind, I guessed it might need to call set_vlan_attribute API with attribute SAI_VLAN_ATTR_INGRESS_ACL setting to NULL, so I put PORT_CHANGE notification before remove vlan from HW ) 

“What if the remove_vlan API failed later?”
----If the remove_vlan API failed, then return false, doVlanTask() will not erase this task from "consumer.m_toSync". It will execute removeVlan() until return true, so notify(PORT_CHANGE) will call many times. But unbind() Will only be called once because port will be erased from ACLtable::portSet when calling unbind() for the first time.
On the HW, vlan still exists, but there is no acltable bound on this vlan.

So, do I need to put PORT_CHANGE notification before "removeAclTableGroup(vlan);" and after remove vlan from HW ?
I'll do this in the next commit.

orchagent/portsorch.cpp
ouxlwhu: Same as the previous reply

tests/test_vlan.py
ouxlwhu: Yes

tests/test_vlan.py
ouxlwhu: It‘s a good idea, I'll do it in the next commit.Thank you!

PR #994, opened 2019-07-25
[vstest]: report error when team kernel module is not available

lguohan:
make explicit failure if team kernel module is not available on the system

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -158,6 +158,9 @@ class DockerVirtualSwitch(object):
         self.alld  = self.basicd + self.swssd + self.syncd + self.rtd + self.teamd
         self.client = docker.from_env()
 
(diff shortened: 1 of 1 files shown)


lguohan:
sometimes, the kernel driver package is splitted in base module and extra module packages, and team driver is in the extra package, and extra packet is not automatically installed. then, all lag related vs test will not pass. It is hard to debug that, therefore I am adding explicit check for this requirement.

lguohan:
this happened on the worker-2, after I manually installed extra kernel module, everything works.

lguohan:
yes, we want to have that control, but sometime, the server automatically upgrading their kernel controlled by azure. we need to disable it, but it is a separate issue.

PR #997, opened 2019-07-29
Trap DHCPv6 packets for supporting ZTP over in-band interfaces using …

rajendra-dendukuri:
…DHCPv6 discovery

Also increase incoming packet rate on in-band interfaces to support faster
download of large files. SONiC firmware image download over in-band can
take a lot of time if the incoming packet rate is limited to 600pps. This,
change increases it to 6000pps.

**What I did**
Increased packet rate for traffic destined to CPU. 

Added DHCPv6 to trap list to support ZTP over IPv6 transport.

**Why I did it**
This is required for ZTP to download large files over in-band interface.
Support ZTP over IPv6 transport.

**How I verified it**
ZTP over in-band interface

**Details if related**

### DIFF
--- a/swssconfig/sample/00-copp.config.json
+++ b/swssconfig/sample/00-copp.config.json
@@ -34,8 +34,8 @@
         "OP": "SET"
     },
     {
-        "COPP_TABLE:trap.group.lldp.dhcp.udld": {
-            "trap_ids": "lldp,dhcp,udld",
+        "COPP_TABLE:trap.group.lldp.dhcp.dhcpv6.udld": {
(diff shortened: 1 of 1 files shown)


rajendra-dendukuri:
> Could you separate this pull request into two different pull requests? They are serving different purposes.
> Also, changing the CIR/CBS of the current policer might change the current image's behavior and stability. At least it will break the current tests which might not be desired at this moment.

As suggested, I moved the ip2me policer changes to a different PR https://github.com/Azure/sonic-swss/pull/1000

rajendra-dendukuri:
retest vs please

stcheng:
retest this please

PR #1002, opened 2019-07-30
Config DB manual is being moved from Wiki to SWSS repo

kannankvs:
Config DB manual is being moved from Wiki to SWSS repo.
Once if this is merged, the page from Wiki will point to this config DB manual.
After that, whichever developer makes changes on config DB, are expected to make the corresponding changes in this manual.

### DIFF
--- a/doc/Configuration.md
+++ b/doc/Configuration.md
@@ -1,0 +1,1165 @@
+# SONiC Configuration Database Manual
+
+Table of Contents
+=================
+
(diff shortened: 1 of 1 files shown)


xinliu-seattle:
Add the context.  We need to ask the code owner to update the configuration WiKi if he/she introduces any changes. This file should be updated and check in together with the code change that  touches configDB.

stcheng:
retest this please

PR #1003, opened 2019-07-30
[teammgrd]during warm-reboot teamd need to recover system-id from saved lacp-pdu

shine4chen:
Signed-off-by: shine.chen <shine.chen@nephosinc.com>

**What I did**
- When creating new portchannel, teammgrd gets the MAC from the saved LACP PDU and updates the port-channel MAC accordingly.
- Pls refer to [MCLAG-HLD](https://github.com/Azure/SONiC/pull/325) for details.

**Why I did it**
- In MC-LAG scenario, two peer devices form one end point of a LAG, these two devices must have the same MAC address since it’s used for LACP. During warm-reboot, this MAC must not be changed. 

**How I verified it**
- test it on nephos lab. With this approach we observe zero packet loss during mclag warm-reboot procedure.

**Details if related**

### DIFF
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -378,11 +378,13 @@ task_process_status TeamMgr::addLag(const string &alias, int min_links, bool fal
 
     const string dump_path = "/var/warmboot/teamd/";
     MacAddress mac_boot = m_mac;
-    /*set portchannel mac same with mac before warmStart ,when warmStart there is a file written by teamd*/
+
+    // set portchannel mac same with mac before warmStart, when warmStart and there
+    // is a file written by teamd.
     ifstream aliasfile(dump_path + alias);
     if (WarmStart::isWarmStart() && aliasfile.is_open())
     {
-# define PARTNER_SYSTEM_ID_OFFSET 40
+        const int partner_system_id_offset = 40;
         string line;
 
         while (getline(aliasfile, line))
@@ -393,8 +395,8 @@ task_process_status TeamMgr::addLag(const string &alias, int min_links, bool fal
             if (!memberfile.is_open())
                 continue;
 
-            memberfile.seekg(PARTNER_SYSTEM_ID_OFFSET, std::ios::beg);
-            memberfile.read((char *)mac_temp, ETHER_ADDR_LEN);
+            memberfile.seekg(partner_system_id_offset, std::ios::beg);
+            memberfile.read(reinterpret_cast<char*>(mac_temp), ETHER_ADDR_LEN);
             mac_boot = MacAddress(mac_temp);
             break;
         }
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -7,6 +7,8 @@
 #include "portmgr.h"
 
 #include <algorithm>
+#include <iostream>
+#include <fstream>
 #include <sstream>
 #include <thread>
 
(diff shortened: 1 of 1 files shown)


lguohan:
vs test missing.

shine4chen:
@lguohan  Teammgrd runs in teamd docker. Originally we want to add some vs test cases. But we find the existing vs image don't include teamd docker.  We can add vs test cases soon after teamd docker is added in vs image.

adyeung:
BRCM team has completed the review, we have no further request or comment to add, please proceed to the next steps.

pavel-shirshov:
@shine4chen 
How "correct" mac address is going to be set initially, before it is saved to a file?
Why mac address from config db is considered as incorrect?

jianjundong:
> @shine4chen
> How "correct" mac address is going to be set initially, before it is saved to a file?
> Why mac address from config db is considered as incorrect?

@pavel-shirshov : Teamd saved the last PDU from partner per port in /var/warmboot/teamd/, the mac address of this device that saved in the PDU may be different to the mac address in config db. For example, in MCLAG scenario, the standby device will change the local mac address to the mac address of active device. If the standby device is warm rebooted, this mac address must be the same as that before warm-restart, if not, the portchannel will be broken and the data forwarding may be blocked.

cfgmgr/teammgr.cpp
```diff
@@ -373,8 +375,33 @@ task_process_status TeamMgr::addLag(const string &alias, int min_links, bool fal
...
 
     stringstream conf;
+
+    const string dump_path = "/var/warmboot/teamd/";
+    MacAddress mac_boot = m_mac;
+    /*set portchannel mac same with mac before warmStart ,when warmStart there is a file written by teamd*/
+    ifstream aliasfile(dump_path + alias);
+    if (WarmStart::isWarmStart() && aliasfile.is_open())
+    {
+# define PARTNER_SYSTEM_ID_OFFSET 40
```
pavel-shirshov: I'd better to use `const auto` or `const int` here.

cfgmgr/teammgr.cpp
```diff
@@ -373,8 +375,33 @@ task_process_status TeamMgr::addLag(const string &alias, int min_links, bool fal
     string res;
 
     stringstream conf;
+
+    const string dump_path = "/var/warmboot/teamd/";
+    MacAddress mac_boot = m_mac;
+    /*set portchannel mac same with mac before warmStart ,when warmStart there is a file written by teamd*/
```
pavel-shirshov: Please change it to:
`/* set portchannel mac same with mac before warmStart, when warmStart there is a file written by teamd */`

cfgmgr/teammgr.cpp
```diff
@@ -373,8 +375,33 @@ task_process_status TeamMgr::addLag(const string &alias, int min_links, bool fal
...
+        while (getline(aliasfile, line))
+        {
+            ifstream memberfile(dump_path + line, ios::binary);
+            uint8_t mac_temp[ETHER_ADDR_LEN];
+
+            if (!memberfile.is_open())
+                continue;
+
+            memberfile.seekg(PARTNER_SYSTEM_ID_OFFSET, std::ios::beg);
+            memberfile.read((char *)mac_temp, ETHER_ADDR_LEN);
```
pavel-shirshov: instead of `(char *)` it's better to use `static_cast<char*>`

cfgmgr/teammgr.cpp
shine4chen: thanks,done

cfgmgr/teammgr.cpp
shine4chen: Thanks, done

cfgmgr/teammgr.cpp
shine4chen: Thanks, done

pavel-shirshov:
retest all please

pavel-shirshov:
retest this please

shine4chen:
retest this please

pavel-shirshov:
@shine4chen Can you please check why checks are failing?

shine4chen:
@pavel-shirshov I analysised 32 failed test cases and found two root cause.
1. cmd `ip link add pensive_ishi type veth peer name eth1` executes failed which result in the failure of `DockerVirtualSwitch`.
2. in this vs test environment [frr-vrf-patch](https://github.com/Azure/sonic-buildimage/pull/3763) is NOT imported which result in the failure of some vrf related test cases. I know frr-vrf-patch is merged to master. But for some environment issue the build failed, so vs test still gets the obsolete build artifact.

shine4chen:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

pavel-shirshov:
retest this please

shine4chen:
retest this please

shine4chen:
retest this please

PR #1004, opened 2019-07-30
[VNET] Add vnet tunnel route delete protection.

liron-ze:
Signed-off-by: liuze <liuze@asterfusion.com>

**What I did**
Judge if a vnet tunnel route exists before deleting it.

**Why I did it**
Delete a route that does not exist will return an error.

**How I verified it**
test

**Details if related**
in routesyncd
add vnet route:
m_vnet_routeTable.add(vnet_dip) 
del vnet route:
m_vnet_routeTable.del(vnet_dip);
m_vnet_tunnelTable.del(vnet_dip);  (An error occurred in vnetorch)

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -47,6 +47,7 @@ int gBatchSize = DEFAULT_BATCH_SIZE;
 bool gSairedisRecord = true;
 bool gSwssRecord = true;
 bool gLogRotate = false;
+bool gSaiRedisLogRotate = false;
 bool gSyncMode = false;
 
 ofstream gRecordOfs;
@@ -73,15 +74,7 @@ void sighup_handler(int signo)
      * Don't do any logging since they are using mutexes.
      */
     gLogRotate = true;
-
-    sai_attribute_t attr;
-    attr.id = SAI_REDIS_SWITCH_ATTR_PERFORM_LOG_ROTATE;
-    attr.value.booldata = true;
-
-    if (sai_switch_api != NULL)
-    {
-        sai_switch_api->set_switch_attribute(gSwitchId, &attr);
-    }
+    gSaiRedisLogRotate = true;
 }
 
 void syncd_apply_view()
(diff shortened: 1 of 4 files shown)


orchagent/vnetorch.cpp
```diff
@@ -1662,6 +1662,13 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
     }
 
     auto *vrf_obj = vnet_orch_->getTypePtr<VNetVrfObject>(vnet);
+
+    if (op == DEL_COMMAND && !vrf_obj->hasRoute(ipPrefix))
+    {
+        SWSS_LOG_WARN("VNET %s, Tunnel Route %s do not exist", vnet.c_str(), ipPrefix.to_string().c_str());
+        return true;
+    }
+
```
marian-pritsak: Can you elaborate on how it possible to achieve a flow where the Redis DB generates a DEL event if the entry doesn't exist?

liron-ze: I tested this on the machine.

Add sub route to Vnet1:
(App DB)VNET_ROUTE_TABLE:Vnet1|123.0.0.0/24

Then kill orch add del the sub route.
redis-cli 
127.0.0.1:6379> SMEMBERS VNET_ROUTE_TUNNEL_TABLE_DEL_SET
1) "Vnet1|123.0.0.0/24"
127.0.0.1:6379> SMEMBERS VNET_ROUTE_TABLE_DEL_SET
1) "Vnet1|123.0.0.0/24"

prsunny:
I think, this is an invalid use-case. When orchagent starts, it starts with a clean state and all previously created APP tables would be flushed out. Not sure how we can simulate this issue in reality.

ryan44guo:
I know this problem and make the same modification locally. When an ip address removed from one interface which bound to vnet, frr will produce a route delete message to fpmsyncd(routesync),but only ipPrefix provided, no nexthop info. So routeorch produces two action, delete it from VNET ROUTE and delete it from VNET TUNNEL ROUTE, the later (if it is VNET TUNNEL ROUTE, because of  not checking its existence as this modification)will be failed becuase of no such route. And because of it returned false, it will not been removed from m_tosync, so it will try again and again. @prsunny

PR #1005, opened 2019-07-30
[orchagent] Add support for Innovium platform

tonytitus:
Signed-off-by: Tony Titus ttitus@innovium.com

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/Makefile.am
+++ b/orchagent/Makefile.am
@@ -5,6 +5,7 @@ CFLAGS_SAI = -I /usr/include/sai
 swssdir = $(datadir)/swss
 
 dist_swss_DATA = \
+		 pfc_detect_innovium.lua  \
 		 pfc_detect_mellanox.lua  \
 		 pfc_detect_broadcom.lua \
 		 pfc_detect_barefoot.lua \
--- a/orchagent/orch.h
+++ b/orchagent/orch.h
@@ -32,6 +32,7 @@ const char range_specifier     = '-';
 const char config_db_key_delimiter = '|';
 const char state_db_key_delimiter  = '|';
 
+#define INVM_PLATFORM_SUBSTRING "innovium"
 #define MLNX_PLATFORM_SUBSTRING "mellanox"
 #define BRCM_PLATFORM_SUBSTRING "broadcom"
 #define BFN_PLATFORM_SUBSTRING  "barefoot"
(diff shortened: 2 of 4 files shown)


prsunny:
retest this please

yxieca:
@tonytitus Can you explain why you need this change in 201811 branch? I think new feature should stay in master unless there is a strong reason for back porting to 201811 branch.

lguohan:
can you resolve the conflict?

tonytitus:
> can you resolve the conflict?

Done

lguohan:
better to do rebase. merge will show lots of commits in the pr.

tonytitus:
> better to do rebase. merge will show lots of commits in the pr.

Done

orchagent/portsorch.cpp
```diff
@@ -1634,7 +1634,10 @@ void PortsOrch::doPortTask(Consumer &consumer)
             {
                 for (auto it = m_portListLaneMap.begin(); it != m_portListLaneMap.end();)
                 {
-                    if (m_lanesAliasSpeedMap.find(it->first) == m_lanesAliasSpeedMap.end())
+                    sai_uint32_t port_speed;
+                    auto is = m_lanesAliasSpeedMap.find(it->first);
+                    getPortSpeed(it->second, port_speed);
+                    if ((m_lanesAliasSpeedMap.find(it->first) == m_lanesAliasSpeedMap.end()) || (get<1>(is->second) != port_speed))
```
lguohan: why do we need to compare the port speed here?

tonytitus: When number of lanes are same but speed is different in port_config.ini(and config_db.json), we still need to remove port and add port. Without this change the port will not get recreated for new speed.

lguohan: if lanes are the same set, then we just need to set speed. why do we need to remove and add port?

tonytitus: When we breakout a port from one 100G (2lane * 50G PAM4) to two 50G (2lane * 25G NRZ) then we will have to recreate the port as the FEC mode could change. Same is true when we breakout a port from one 200G (4lane * 50G PAM4) to two 100G (4lane * 25G NRZ).

lguohan: sai should handle the fec change accordingly, or let application handle the fec settings. such recreating port should not be expose to application.

tonytitus: I removed this change

lguohan:
retest this please

yxieca:
This PR cannot be cherry-picked into 201811 branch cleanly. Separate PR is required for 201811 branch.

songqh2022:
What I did
Support for platforms based on Clounix Networks' device
Why I did it
Support for platforms based on Clounix Networks' device
How I verified it
Have verified on the hardware device
Details if related

PR #1015, opened 2019-08-01
[build]: Fix compiling warnings using ARM 32 bit compiler

antony-rheneus:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


antony-rheneus:
I couldn't understand the failure results, would you please provide suggestions to resolve it.

fpmsyncd/fpmlink.cpp
```diff
@@ -106,7 +106,7 @@ void FpmLink::readData()
     /* Check for complete messages */
     while (true)
     {
-        hdr = (fpm_msg_hdr_t *)(m_messageBuffer + start);
+        hdr = reinterpret_cast<fpm_msg_hdr_t *>(static_cast<void *>(m_messageBuffer + start));
```
qiluo-msft: >static_cast<void *> [](start = 48, length = 19)

Could you remove static_cast and only use reinterpret_cast? Please also provide the compilation message before and after. We have no easy way to reproduce your problem.

orchagent/watermarkorch.cpp
```diff
@@ -98,7 +98,7 @@ void WatermarkOrch::handleWmConfigUpdate(const std::string &key, const std::vect
         {
             if (i.first == "interval")
             {
-                auto intervT = timespec { .tv_sec = to_uint<uint32_t>(i.second.c_str()) , .tv_nsec = 0 };
+                auto intervT = timespec { .tv_sec = static_cast<time_t>(to_uint<uint32_t>(i.second.c_str())) , .tv_nsec = 0 };
```
qiluo-msft: >static_cast<time_t>(to_uint<uint32_t>(i.second.c_str())) [](start = 52, length = 56)

use mktime()

orchagent/portsorch.cpp
```diff
@@ -1452,7 +1452,7 @@ bool PortsOrch::bake()
     if (portCount != keys.size() - 2)
     {
         // Invalid port table
-        SWSS_LOG_ERROR("Invalid port table: portCount, expecting %" PRIu64 ", got %" PRIu64,
+        SWSS_LOG_ERROR("Invalid port table: portCount, expecting %" PRIu64 ", got %zu",
```
qiluo-msft: >PRIu64 [](start = 68, length = 6)

%lu is the correct format for unsigned long. Then you don't need change type of portCount

orchagent/crmorch.cpp
```diff
@@ -240,7 +240,7 @@ void CrmOrch::handleSetCommand(const string& key, const vector<FieldValueTuple>&
             if (field == CRM_POLLING_INTERVAL)
             {
                 m_pollingInterval = chrono::seconds(to_uint<uint32_t>(value));
-                auto interv = timespec { .tv_sec = m_pollingInterval.count(), .tv_nsec = 0 };
+                auto interv = timespec { .tv_sec = (time_t)m_pollingInterval.count(), .tv_nsec = 0 };
```
qiluo-msft: >(time_t)m_pollingInterval.count() [](start = 51, length = 33)

Use mktime(). time_t is not portable if directly used.

orchagent/pfcactionhandler.cpp
```diff
@@ -404,7 +404,7 @@ bool PfcWdLossyHandler::getHwCounters(PfcWdHwStats& counters)
         return false;
     }
 
-    sai_object_id_t pg = portInstance.m_priority_group_ids[getQueueId()];
+    sai_object_id_t pg = portInstance.m_priority_group_ids[static_cast <size_t> (getQueueId())];
```
qiluo-msft: >static_cast <size_t> (getQueueId())]; [](start = 59, length = 37)

Could you provide the compilation message on original code?

orchagent/watermarkorch.cpp
```diff
@@ -295,7 +295,7 @@ void WatermarkOrch::clearSingleWm(Table *table, string wm_name, vector<sai_objec
 {
     /* Zero-out some WM in some table for some vector of object ids*/
     SWSS_LOG_ENTER();
-    SWSS_LOG_DEBUG("clear WM %s, for %" PRId64 " obj ids", wm_name.c_str(), obj_ids.size());
+    SWSS_LOG_DEBUG("clear WM %s, for %zd obj ids", wm_name.c_str(), obj_ids.size());
```
qiluo-msft: >%zd [](start = 37, length = 3)

%zu

qiluo-msft:
>         ProducerStateTable *psTable, const uint32_t defaultWarmStartTimerValue = 0);

We also need change implementation to use mktime explicitly convert from uint32_t (or any other int type) to time_t

---
Refers to: warmrestart/warmRestartAssist.h:58 in 00c8c93. [](commit_id = 00c8c939f69a8faf857ee2b29f3df7e58b3bd304, deletion_comment = False)

orchagent/portsorch.cpp
antony-rheneus: portCount is declared as size_t and not as ulong.

antony-rheneus: %zu is for (keys.size()-2) which is of type size_t and hence %zu has to be used.

orchagent/crmorch.cpp
antony-rheneus: time_t mktime(struct tm *tm) accepts "tm" and not the seconds.
As per code, m_pollingInterval.count() seems to be secs but was declared as ulong. time_t is also typedef as long in .h file.

orchagent/portsorch.cpp
qiluo-msft: I am talking about `portCount` format string `PRIu64` -> '%lu', and no need to change its original type `unsigned long`

---
In reply to: [309990956](https://github.com/Azure/sonic-swss/pull/1015#discussion_r309990956) [](ancestors = 309990956)

orchagent/crmorch.cpp
qiluo-msft: time_t is 32-bit signed in 32 bit gcc, and 64-bit sign in 64 bit gcc. So time_t is not portable if directly used in casting. You can first construct a `tm` with a integer sec field, and then use `mktime` to get a `time_t`.

---
In reply to: [309991958](https://github.com/Azure/sonic-swss/pull/1015#discussion_r309991958) [](ancestors = 309991958)

warmrestart/warmRestartAssist.h
```diff
@@ -108,7 +108,7 @@ class AppRestartAssist
     std::string m_appTableName;       // application table name
 
     bool m_warmStartInProgress;       // indicate if warm start is in progress
-    uint32_t m_reconcileTimer;        // reconcile timer value
+    time_t m_reconcileTimer;          // reconcile timer value
```
stcheng: please remove the trailing spaces

antony-rheneus: If I remove ";_" trailing space after semicolon, the comment "//" will not start in same column as previous line.

orchagent/portsorch.cpp
antony-rheneus: I have updated the code by changing stoul with strtoumax() API to avoid using ulong.

warmrestart/warmRestartAssist.h
stcheng: ah my bad 😭

fpmsyncd/fpmlink.cpp
antony-rheneus: Original error log:
```
fpmlink.cpp:110:56: error: cast from 'char*' to 'fpm_msg_hdr_t* {aka fpm_msg_hdr_t_*}' increases required alignment of target type [-Werror=cast-align]
         hdr = (fpm_msg_hdr_t *)(m_messageBuffer + start);
```
Removing static cast error log:-
```
fpmlink.cpp:110:72: error: cast from 'char*' to 'fpm_msg_hdr_t* {aka fpm_msg_hdr_t_*}' increases required alignment of target type [-Werror=cast-align]
         hdr = reinterpret_cast<fpm_msg_hdr_t *>(m_messageBuffer + start);
```

orchagent/pfcactionhandler.cpp
antony-rheneus: Error log:
```
pfcactionhandler.cpp:408:70: error: conversion to 'std::vector<long long unsigned int>::size_type {aka unsigned int}' from 'sai_object_id_t {aka long long unsigned int}' may alter its value [-Werror=conversion]
     sai_object_id_t pg = portInstance.m_priority_group_ids[getQueueId()];
                                                            ~~~~~~~~~~^~
```

orchagent/watermarkorch.cpp
antony-rheneus: Done

orchagent/crmorch.cpp
antony-rheneus: Done

orchagent/watermarkorch.cpp
antony-rheneus: Done

antony-rheneus:
> > ```
> >     ProducerStateTable *psTable, const uint32_t defaultWarmStartTimerValue = 0);
> > ```
> 
> We also need change implementation to use mktime explicitly convert from uint32_t (or any other int type) to time_t
> 
> Refers to: warmrestart/warmRestartAssist.h:58 in [00c8c93](https://github.com/Azure/sonic-swss/commit/00c8c939f69a8faf857ee2b29f3df7e58b3bd304). [](commit_id = [00c8c93](https://github.com/Azure/sonic-swss/commit/00c8c939f69a8faf857ee2b29f3df7e58b3bd304), deletion_comment = False)

@qiluo-msft , I addressed as per your comments, changing from typecast to mktime()
however, typecast assigns the same value, whereas mktime() converts the time to different value. This would break the functionality.

antony-rheneus:
> > > ```
> > >     ProducerStateTable *psTable, const uint32_t defaultWarmStartTimerValue = 0);
> > > ```
> > 
> > 
> > We also need change implementation to use mktime explicitly convert from uint32_t (or any other int type) to time_t
> > Refers to: warmrestart/warmRestartAssist.h:58 in [00c8c93](https://github.com/Azure/sonic-swss/commit/00c8c939f69a8faf857ee2b29f3df7e58b3bd304). [](commit_id = [00c8c93](https://github.com/Azure/sonic-swss/commit/00c8c939f69a8faf857ee2b29f3df7e58b3bd304), deletion_comment = False)
> 
> @qiluo-msft , I addressed as per your comments, changing from typecast to mktime()
> however, typecast assigns the same value, whereas mktime() converts the time to different value. This would break the functionality.

@qiluo-msft , I suggest to revert the mktime() changes and go with typecasting approach

orchagent/watermarkorch.cpp
```diff
@@ -98,7 +98,10 @@ void WatermarkOrch::handleWmConfigUpdate(const std::string &key, const std::vect
         {
             if (i.first == "interval")
             {
-                auto intervT = timespec { .tv_sec = to_uint<uint32_t>(i.second.c_str()) , .tv_nsec = 0 };
+                struct tm tmInterval;
+                memset(&tmInterval, 0, sizeof(tmInterval));
+                tmInterval.tm_sec = to_uint<uint32_t>(i.second.c_str());
```
qiluo-msft: >uint32_t [](start = 44, length = 8)

`tm_sec` is `int`, why not convert to `int` directly? You may check the validity after conversion.

antony-rheneus: Should we really need to do mktime(),
For example,
If tm_sec value is 5899989  then mktime returns Time_t 18446744071506376405 which is different from the earlier functionality, where polling count was directly used as value for seconds.

fpmsyncd/fpmlink.cpp
qiluo-msft: The original error does make sense, and original code need improvement to make it portable. casting is bad smell and double casting is even worse.

One possible solution is to refine original code to increase pointer by alignment. I notice [https://github.com/Azure/sonic-swss/blob/master/fpmsyncd/fpm/fpm.h#L116](https://github.com/Azure/sonic-swss/blob/master/fpmsyncd/fpm/fpm.h#L116) indicating alignment is always true.

---
In reply to: [310905179](https://github.com/Azure/sonic-swss/pull/1015#discussion_r310905179) [](ancestors = 310905179)

qiluo-msft:
I could not understand
> typecast assigns the same value, whereas mktime() converts the time to different value. This would break the functionality. 

Could you clarify?

---
In reply to: [518963372](https://github.com/Azure/sonic-swss/pull/1015#issuecomment-518963372) [](ancestors = 518963372)

antony-rheneus:
> I could not understand
> 
> > typecast assigns the same value, whereas mktime() converts the time to different value. This would break the functionality.
> 
> Could you clarify?
> 
> In reply to: [518963372](https://github.com/Azure/sonic-swss/pull/1015#issuecomment-518963372) [](ancestors = 518963372)

Original code:-
auto interv = timespec { .tv_sec = m_pollingInterval.count(), .tv_nsec = 0 };

Modified Code:-
tmInterval.tm_sec = m_pollingInterval.count();
auto interv = timespec { .tv_sec = mktime(&tmInterval), .tv_nsec = 0 };

In the above 2 cases, timespec.tv_sec values are NOT same. as mktime() returns a timestamp which is between given time since Unix Epoch.

orchagent/watermarkorch.cpp
antony-rheneus: Reverted the mktime() changes and test passes.

fpmsyncd/fpmlink.cpp
antony-rheneus: I dont think the gcc warn about pack size of struct, it warns because the pointer to char can have lesser size compared to left-value  pointer reference.

orchagent/crmorch.cpp
antony-rheneus: Reverted the mktime changes, and kept typecasting implementation

qiluo-msft: >(time_t) [](start = 51, length = 8)

Can you remove the `(time_t)`? What is the error message?

fpmsyncd/fpmsyncd.cpp
```diff
@@ -84,7 +84,7 @@ int main(int argc, char **argv)
             if (warmStartEnabled)
             {
                 /* Obtain warm-restart timer defined for routing application */
-                uint32_t warmRestartIval = sync.m_warmStartHelper.getRestartTimer();
+                time_t  warmRestartIval = sync.m_warmStartHelper.getRestartTimer();
```
qiluo-msft: >  [](start = 23, length = 1)

extra blank

orchagent/pfcactionhandler.cpp
```diff
@@ -521,7 +521,7 @@ PfcWdZeroBufferHandler::~PfcWdZeroBufferHandler(void)
         return;
     }
 
-    sai_object_id_t pg = portInstance.m_priority_group_ids[getQueueId()];
+    sai_object_id_t pg = portInstance.m_priority_group_ids[static_cast <size_t> (getQueueId())];
```
qiluo-msft: >static_cast <size_t> (getQueueId()) [](start = 59, length = 35)

`size_t(getQueueId())` may be more clear.

lguohan:
@qiluo-msft , can you approve?

fpmsyncd/fpmsyncd.cpp
antony-rheneus: Done

orchagent/pfcactionhandler.cpp
antony-rheneus: Done

PR #1030, opened 2019-08-16
Update ECMP NHopGroup for Port Channel oper down 

sumukhatv:
What I did
Remove nexthop member from nexthop group immediately when port-channel oper status becomes down

Why I did it
This will reduce the response time for port-channel down event thus helping prevent packet drops

How I verified it

Details if related
Timeline of events on a DUT immediately after bringing down a portchannel:
21:20:51 Updated NextHop group to 3 members
21:20:52 First v4 route update
21:21:13 Last v4 route update
21:21:13 4 member NextHop group deleted (v4)
21:21:13 3 member NextHop group created (v6)
21:21:13 First v6 route update
21:21:16 Last v6 route update

### DIFF
--- a/tests/test_portchannel.py
+++ b/tests/test_portchannel.py
@@ -57,3 +57,130 @@ class TestPortchannel(object):
 
         lagms = lagmtbl.getKeys()
         assert len(lagms) == 0
+
+
+    def test_Portchannel_oper_down(self, dvs, testlog):
+
+        self.adb = swsscommon.DBConnector(1, dvs.redis_sock, 0)
+        self.cdb = swsscommon.DBConnector(4, dvs.redis_sock, 0)
+        self.pdb = swsscommon.DBConnector(0, dvs.redis_sock, 0)
+
+        # Create 4 Ports
+        tbl = swsscommon.Table(self.cdb, "PORT")
+        fvs = swsscommon.FieldValuePairs([("admin_status", "up")])
+
+        tbl.set("Ethernet1", fvs)
+        time.sleep(1)
+        tbl.set("Ethernet2", fvs)
+        time.sleep(1)
+        tbl.set("Ethernet3", fvs)
+        time.sleep(1)
+        tbl.set("Ethernet4", fvs)
+        time.sleep(1)
+
+        # Create 4 PortChannels
+        tbl = swsscommon.Table(self.cdb, "PORTCHANNEL")
+        fvs = swsscommon.FieldValuePairs([("admin_status", "up"),("mtu", "9100")])
+
+        tbl.set("PortChannel001", fvs)
+        time.sleep(1)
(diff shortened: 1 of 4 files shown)


prsunny:
@sumukhatv , would need to add a VS test for this change

tests/test_portchannel.py
```diff
@@ -57,3 +57,130 @@ def test_Portchannel(self, dvs, testlog):
...
+        # check if next hop group consists of 4 members
+        nhg_member_tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_NEXT_HOP_GROUP_MEMBER")
+        keys = nhg_member_tbl.getKeys()
+        assert len(keys) == 4
+
+        for key in keys:
+            (status, fvs) = nhg_member_tbl.get(key)
+            for v in fvs:
+                if v[0] == "SAI_NEXT_HOP_GROUP_MEMBER_ATTR_NEXT_HOP_GROUP_ID":
+                    assert v[1] == nhg_id
```
prsunny: I assume it is a partial commit and you are modifying the test to simulate oper down scenario.

tests/test_portchannel.py
```diff
@@ -57,3 +57,178 @@ def test_Portchannel(self, dvs, testlog):
...
+
+        self.adb = swsscommon.DBConnector(1, dvs.redis_sock, 0)
+        self.cdb = swsscommon.DBConnector(4, dvs.redis_sock, 0)
+        self.pdb = swsscommon.DBConnector(0, dvs.redis_sock, 0)
+
+        # Create 4 PortChannels
+        tbl = swsscommon.Table(self.cdb, "PORTCHANNEL")
+        fvs = swsscommon.FieldValuePairs([("admin_status", "up"),("mtu", "9100"),("oper_status", "up")])
+
+        tbl.set("PortChannel001", fvs)
```
prsunny: I see that there is code repetition. If you can create a list of Portchannels and call function to do set, del operation, it will simplify and reduce the code. Just a suggestion.

tests/test_portchannel.py
```diff
@@ -57,3 +57,178 @@ def test_Portchannel(self, dvs, testlog):
...
+        tbl._del("PortChannel004|Ethernet12")
+        time.sleep(1)
+
+        # remove PortChannel
+        tbl = swsscommon.Table(self.cdb, "PORTCHANNEL")
+        tbl._del("PortChannel001")
+        tbl._del("PortChannel002")
+        tbl._del("PortChannel003")
+        tbl._del("PortChannel004")
+        time.sleep(1)
```
wendani, line 234: Restore eth0 up. If test_port.py follows this test, we got persistent test failure.

prsunny: @wendani , would you please provide the fix?

PR #1031, opened 2019-08-16
Adjust stale timer for warm-reboot neighborsync test cases

zhenggen-xu:
Adjust stale timer for warm-reboot neighborsync test cases

Increase the stale timer to "600" seconds so it won't be aged
out in case the test server is busy or slow

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


stcheng:
retest this please

PR #1032, opened 2019-08-19
[neighorch]: Return false when the next hop IP exists before adding

stcheng:
It is possible that due to events reordering, a different next hop
entry with same IP will be inserted before the previous next hop
entry gets removed. Thus, return false here to prevent the new entry
from replacing the old one automatically.

### DIFF
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -30,6 +30,13 @@ bool NeighOrch::addNextHop(IpAddress ipAddress, string alias)
 {
     SWSS_LOG_ENTER();
 
+    if (hasNextHop(ipAddress))
+    {
+        SWSS_LOG_ERROR("Next hop IP %s already exists",
+                ipAddress.to_string().c_str());
+        return false;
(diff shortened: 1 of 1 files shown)


stcheng:
retest this please4

stcheng:
retest this please

stcheng:
retest this please

orchagent/neighorch.cpp
```diff
@@ -30,6 +30,13 @@ bool NeighOrch::addNextHop(IpAddress ipAddress, string alias)
 {
     SWSS_LOG_ENTER();
 
+    if (hasNextHop(ipAddress))
```
prsunny: Do you want to take care of out-of-sequence in the delete case as well?

lguohan: any update on this @stcheng ?

stcheng:
updated

stcheng:
Since the pull request #977 could fix this issue ultimately, this pull request will be closed.

PR #1035, opened 2019-08-21
[intfsorch] Don't add subnet route for IP addresses with maximum prefix length

shikenghua:
**What I did**
Don't add subnet route for IP addresses with maximum prefix length. 

**Why I did it**
It is a duplicate of IP2Me route and will cause exception "Failed to create IP2me route." after it is added.

**How I verified it**
Test with adding a /32 IP address on interface. Check syslog and whether swss container crash.

**Details if related**
N/A

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -195,8 +195,14 @@ bool IntfsOrch::setIntf(const string& alias, sai_object_id_t vrf_id, const IpPre
         return false;
     }
 
+    /* Don't add subnet route for IP addresses with maximum prefix length.
+     * It is a duplicate of IP2Me route */
+    if ((ip_prefix->isV4() && ip_prefix->getMaskLength() != 32) ||
+        (!ip_prefix->isV4() && ip_prefix->getMaskLength() != 128))
+    {
+        addSubnetRoute(port, *ip_prefix);
+    }
     vrf_id = port.m_vr_id;
-    addSubnetRoute(port, *ip_prefix);
     addIp2MeRoute(vrf_id, *ip_prefix);
 
     if (port.m_type == Port::VLAN)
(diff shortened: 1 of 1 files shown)


orchagent/intfsorch.cpp
```diff
@@ -195,8 +195,14 @@ bool IntfsOrch::setIntf(const string& alias, sai_object_id_t vrf_id, const IpPre
         return false;
     }
 
+    /* Don't add subnet route for IP addresses with maximum prefix length.
+     * It is a duplicate of IP2Me route */
+    if ((ip_prefix->isV4() && ip_prefix->getMaskLength() != 32) ||
```
prsunny: Just thinking, what is the use-case of configuring IPv4 address with /32 mask?

shikenghua:
@stcheng Do you mean in tests/test_interface.py? Ok, I'll add it.

orchagent/intfsorch.cpp
shikenghua: Besides setting it on loopback interfaces, I don't know other use-case either. But Linux kernel does allow user to configure a /32 IPv4 address on netdevice.

stcheng:
@shikenghua thanks!

orchagent/intfsorch.cpp
stcheng: Do we want to allow user on this scenario?

prsunny:
@shikenghua, @stcheng, this doesn't look to be a valid use-case, For lo interface, it is handled differently and we don't add subnet route. IMO, this change is not required.

stcheng:
agree with Prince. it seems to be it's better to prevent '/32' addresses from being added to the interfaces

shikenghua:
Just tested this on a Cisco router and got a "Bad mask /32 for address x.x.x.x" fail message. It seems that Cisco also prevents the users from adding /32 addresses to the interfaces. Then this change is not required and I'll close it. Besides this, adding an IP address to two different interfaces will also crash the swss container. The checks should be added in sonic-utilities/config/main.py. Right?

PR #1041, opened 2019-08-28
[teammgrd]: Do not crash daemon when commands fail

stcheng:
Do not crash the deamon immediately after executing
some Linux commands and get failure responses.


stcheng:
retest this please

lguohan:
can you explain the reason why do not crash? how are we handling those failures? do we do retry?

Issue #1036, opened 2019-08-23
Syslog still print ERR swss#orchagent: :- getResAvailableCounters


chaoskao:
syslog show error message after deleted ACL table

test step:
1. create ACL table by sonic-cfggen
2. add ACL rule by acl-loader
3. remove ACL rule by acl-loader
4. delete ACL table by redis-cli

syslog will print error message every 5 minutes 
ERR swss#orchagent: :- meta_sai_validate_oid: object key SAI_OBJECT_TYPE_ACL_TABLE:oid:0x7000000000bfc doesn't exist" 
ERR swss#orchagent: :- getResAvailableCounters: Failed to get ACL table attribute 4412 , rv:-5", 
ERR swss#orchagent: :- meta_sai_validate_oid: object key  SAI_OBJECT_TYPE_ACL_TABLE:oid:0x7000000000bfc doesn't exist",
ERR swss#orchagent: :- getResAvailableCounters: Failed to get ACL table attribute 4413 , rv:-5",

PR #1044, opened 2019-08-30
[crmOrch]remove acl entry when used counter == 0

chaoskao:
**What I did**
remove acl entry when used counter of acl rule ==0 and add validation code on unit test
**Why I did it**
acl entry still exist after remove acl rules #1036
**How I verified it**
unit test code passed and also passed on switch
**Details if related**


chaoskao:
retest this please

PR #1045, opened 2019-09-03
Single VRF for ingress and egress flows, skip route replication

prsunny:
**What I did**
Change the model from creating two different VRFs (ingress and egress) as captured [here ](https://github.com/opencomputeproject/SAI/blob/master/doc/SAI-Proposal-QinQ-VXLAN.md
)and use only one VRF

**Why I did it**
To avoid potential scaling issues with replicating routes to multiple VRFs.

**How I verified it**
Run ansible test (vnet_vxlan) and the pytest.

**Details if related**

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -1329,7 +1329,7 @@ VNetOrch::VNetOrch(DBConnector *db, const std::string& tableName, VNET_EXEC op)
 
     if (op == VNET_EXEC::VNET_EXEC_VRF)
     {
-        vr_cntxt = { VR_TYPE::ING_VR_VALID, VR_TYPE::EGR_VR_VALID };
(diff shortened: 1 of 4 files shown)


orchagent/vnetorch.cpp
```diff
@@ -1329,7 +1329,7 @@ VNetOrch::VNetOrch(DBConnector *db, const std::string& tableName, VNET_EXEC op)
 
     if (op == VNET_EXEC::VNET_EXEC_VRF)
     {
-        vr_cntxt = { VR_TYPE::ING_VR_VALID, VR_TYPE::EGR_VR_VALID };
+        vr_cntxt = { VR_TYPE::ING_VR_VALID, VR_TYPE::VR_INVALID };
```
lguohan: what is this VR_INVALID context? if we are using single vrf, why not just have one entry in the vr_cntxt?

prsunny: Removed VR_TYPE::VR_INVALID

PR #1047, opened 2019-09-05
[vxlan decap]: Don't remove tunnel when it has tunnel maps configured

pavel-shirshov:
**What I did**
Fix the issue in orchagent: orchagent allowed removing of a tunnel even if the tunnel has tunnel maps configured.

**Why I did it**
Because it was an issue.

**How I verified it**
Add a tunnel and tunnel map, then try to remove tunnel only

**Details if related**


orchagent/vxlanorch.cpp
```diff
@@ -731,7 +750,24 @@ bool VxlanTunnelOrch::delOperation(const Request& request)
...
+        bool first = true;
+        for (auto vni: tunnel->getMapperKeys())
+        {
+            if (!first)
+                vni_list += ", ";
+            vni_list += std::to_string(vni);
+            first = false;
+        }
+        SWSS_LOG_WARN("Vxlan tunnel '%s' has tunnel maps configured [%s]. Can't remove now.", tunnel_name.c_str(), vni_list.c_str());
+        return false;
```
lguohan: do we do retry here if mapper is not empty?

lguohan: if orchagent gets tunnel remove first and then tunnel map, then we should pend the tunnel remove (retry) later to resolve the dependency issue.

pavel-shirshov: 'return false' mean don't erase from the queue. So the operation will be retried later

lguohan: i mean retry.

pavel-shirshov:
Currently master branch doesn't have stable sai implementation for this feature. So I'll have it as a draft until we got vxlan support in master.

lguohan:
retest this please

PR #1048, opened 2019-09-08
change in fpmsyncd to skip the lookup for the Master device name if the route object table value is zero

Arvind-Brcm:


fpmsyncd/routesync.cpp
```diff
@@ -1,3 +1,4 @@
+
```
prsunny: Please remove this extra line

prsunny:
retest this please

PR #1051, opened 2019-09-10
Modify the lanesperport in SWSS

tiantianlv:
**What I did**
      Support 8 lanes for port
**Why I did it**
      In 201811 branch, lanes of port are for 4, it results orchagent exception for 32x400G of TH3
**How I verified it**

**Details if related**


lguohan:
master has this feature. I think there are several commits to pick up to properly support this feature in 201811.

PR #1054, opened 2019-09-11
[MirrorOrch]: Mirror Session Retention across Warm Reboot

stcheng:
After warm reboot, it is expected that the monitor port of
the mirror session is retained - no changing on the monitor
port withint the ECMP group members and the LAG members. This
is due to the general of the sairedis comparison logic and
the minimalization of SAI function calls during reconciliation.

Changes:
1. Add bake() and postBake() functions in MirrorOrch
   bake() function retrieves the state database information
   and get the VLAN + monitor port information.
   postBake() function leverages the information and recovers
   the active mirror sessions the same as before warm reboot.
2. state database format change
   Instead of storing the object ID of the monitor port, store
   the alias of the monitor port.
   Instead of storing true/false of VLAN header, store the VLAN
   ID.

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -15,6 +15,7 @@
 #define MIRROR_SESSION_STATUS               "status"
 #define MIRROR_SESSION_STATUS_ACTIVE        "active"
 #define MIRROR_SESSION_STATUS_INACTIVE      "inactive"
+#define MIRROR_SESSION_NEXT_HOP_IP          "next_hop_ip"
 #define MIRROR_SESSION_SRC_IP               "src_ip"
 #define MIRROR_SESSION_DST_IP               "dst_ip"
 #define MIRROR_SESSION_GRE_TYPE             "gre_type"
@@ -59,6 +60,7 @@ MirrorEntry::MirrorEntry(const string& platform) :
     }
 
     nexthopInfo.prefix = IpPrefix("0.0.0.0/0");
+    nexthopInfo.nexthop = IpAddress("0.0.0.0");
 }
 
 MirrorOrch::MirrorOrch(TableConnector stateDbConnector, TableConnector confDbConnector,
@@ -81,7 +83,7 @@ bool MirrorOrch::bake()
     SWSS_LOG_ENTER();
 
     // Freeze the route update during orchagent restoration
-    m_freezeUpdate = true;
+    m_freeze = true;
 
     deque<KeyOpFieldsValuesTuple> entries;
     vector<string> keys;
@@ -93,7 +95,7 @@ bool MirrorOrch::bake()
 
         bool active = false;
         string monitor_port;
-        string vlan_id;
+        string next_hop_ip;
 
         for (const auto &tuple : tuples)
         {
@@ -105,9 +107,9 @@ bool MirrorOrch::bake()
             {
                 monitor_port = fvValue(tuple);
             }
-            else if (fvField(tuple) == MIRROR_SESSION_VLAN_ID)
+            else if (fvField(tuple) == MIRROR_SESSION_NEXT_HOP_IP)
             {
-                vlan_id = fvValue(tuple);
+                next_hop_ip = fvValue(tuple);
             }
         }
 
@@ -116,10 +118,12 @@ bool MirrorOrch::bake()
             continue;
         }
 
-        SWSS_LOG_NOTICE("Found mirror session %s active before warm reboot");
+        SWSS_LOG_NOTICE("Found mirror session %s active before warm reboot",
+                key.c_str());
 
         // Recover saved active session's monitor port
-        m_recoverySessionPortMap.emplace(key, vlan_id + ":" + monitor_port);
+        m_recoverySessionMap.emplace(
+                key, monitor_port + state_db_key_delimiter + next_hop_ip);
     }
 
     return Orch::bake();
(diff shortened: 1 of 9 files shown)


orchagent/mirrororch.cpp
```diff
@@ -75,10 +76,145 @@ MirrorOrch::MirrorOrch(TableConnector stateDbConnector, TableConnector confDbCon
...
+                m_neighOrch->getNeighborEntry(vlan.m_alias, neighbor_entry, mac);
+
+                SWSS_LOG_NOTICE("Monitor port %s belongs to neighbor %s with IP %s",
+                        port.m_alias.c_str(), vlan.m_alias.c_str(),
+                        neighbor_entry.ip_address.to_string().c_str());
+
+                session.nexthopInfo.nexthop = neighbor_entry.ip_address;
+            }
+            // If the port belongs to a LAG
+            if (port.m_lag_id)
```
lguohan: a lag port can also belong to a vlan, in this case, you will set the nexthop twice, once in vlan, once in lag.

orchagent/mirrororch.cpp
```diff
@@ -75,10 +76,145 @@ MirrorOrch::MirrorOrch(TableConnector stateDbConnector, TableConnector confDbCon
...
+        SWSS_LOG_NOTICE("Found mirror session %s active before warm reboot");
+
+        // Recover saved active session's monitor port
+        m_recoverySessionPortMap.emplace(key, vlan_id + ":" + monitor_port);
+    }
+
+    return Orch::bake();
+}
+
+bool MirrorOrch::postBake()
```
lguohan: this algorithm has a flaw. for example, let's say previous orchagent has a bug which cacluates a wrong session results (wrong port), in this algorithm, the warm reboot will give the same wrong port, and if the route is not update later, the wrong port will not be corrected.

the correct algorithm will need to check if the recorded results is in the newly calculated results or not, if it is in then use it, if it is not, the warm reboot need to correct that error.

lguohan: I found current approach is difficult to achieve above results. 

the best approach is the move mirror orch after route orch in the dotask loop.

Then in the mirror orch dotask loop, caculate the new mirror session results, check if the old results is contained in the new results, if yes, then use old results. It should also results simpler code since the postBake() introduce new logics which can be avoided.

orchagent/mirrororch.cpp
```diff
@@ -75,10 +76,145 @@ MirrorOrch::MirrorOrch(TableConnector stateDbConnector, TableConnector confDbCon
...
+
+                NeighborEntry neighbor_entry;
+                MacAddress mac;
+                m_neighOrch->getNeighborEntry(vlan.m_alias, neighbor_entry, mac);
+
+                SWSS_LOG_NOTICE("Monitor port %s belongs to neighbor %s with IP %s",
+                        port.m_alias.c_str(), vlan.m_alias.c_str(),
+                        neighbor_entry.ip_address.to_string().c_str());
+
+                session.nexthopInfo.nexthop = neighbor_entry.ip_address;
```
lguohan: i found here finding the nexthop here is unnecessary, if you need nexthop, why not put the nexthop ip before warm reboot and just read it. it is much simpler than introducing a new getNeighborEntry and try to reverse the nexthop from mac and port.

however, as my previous comments, neighbor IP could be calculated wrong, and we still check if the previous results are valid or not.

orchagent/mirrororch.cpp
stcheng: currently, the whole MirrorOrch does not support LAG in VLAN scenario. In order to have it supported, a separate pull request will be desired.

orchagent/mirrororch.cpp
stcheng: It will leads back to the original issue that we are facing - if there are more than 1 mirror sessions require destination update, comparison logic will not be able to handle such cases, and the processes will crash. If it is expected to fix the wrong session results during the warm reboot, such logic shall be introduced after the comparison logic.

stcheng:
@lguohan please check the update, thanks!

orchagent/mirrororch.cpp
```diff
@@ -348,10 +420,16 @@ void MirrorOrch::setSessionState(const string& name, const MirrorEntry& session,
...
+        value = to_string(session.neighborInfo.port.m_vlan_info.vlan_id);
+        fvVector.emplace_back(MIRROR_SESSION_VLAN_ID, value);
+    }
+
+    if (attr.empty() || attr == MIRROR_SESSION_NEXT_HOP_IP)
     {
-        value = to_string(session.neighborInfo.port.m_type == Port::VLAN);
-        fvVector.emplace_back(MIRROR_SESSION_VLAN_HEADER_VALID, value);
+     value = session.nexthopInfo.nexthop.to_string();
+     fvVector.emplace_back(MIRROR_SESSION_NEXT_HOP_IP, value);
```
lguohan: spacing issues here

orchagent/mirrororch.cpp
```diff
@@ -782,7 +894,34 @@ void MirrorOrch::updateNextHop(const NextHopUpdate& update)
 
         if (update.nexthopGroup != IpAddresses())
         {
-            session.nexthopInfo.nexthop = *update.nexthopGroup.getIpAddresses().begin();
+            SWSS_LOG_NOTICE("    next hop IPs: %s", update.nexthopGroup.to_string().c_str());
+
+            // Recover the session based on the state database information
+            if (m_recoverySessionMap.find(name) != m_recoverySessionMap.end())
+            {
+                IpAddress nexthop = IpAddress(tokenize(m_recoverySessionMap[name], '|')[1]);
```
lguohan: use m_tableSeparator instead of hard code?

orchagent/mirrororch.cpp
```diff
@@ -396,32 +474,66 @@ bool MirrorOrch::getNeighborInfo(const string& name, MirrorEntry& session)
...
+
+            // Recover the VLAN member monitor port picked before warm reboot
+            // since the FDB entries are not yet learned on the hardware
+            if (m_recoverySessionMap.find(name) != m_recoverySessionMap.end())
             {
-                SWSS_LOG_NOTICE("Waiting to get FDB entry MAC %s under VLAN %s",
-                        session.neighborInfo.mac.to_string().c_str(),
-                        session.neighborInfo.port.m_alias.c_str());
-                return false;
+                string alias = tokenize(m_recoverySessionMap[name], '|')[0];
```
lguohan: >| [](start = 69, length = 1)

use m_tableSeparator instead of hard code?

lguohan:
I think we need vs test for this feature. can you later add vs test?

stcheng:
sure; i'll send a separate pull request to enhance the test

stcheng:
retest this please

stcheng:
retest this please

stcheng:
retest this please

stcheng:
retest this please

PR #1056, opened 2019-09-12
[crm] syslog generation handling for single threshold for different b…

AmitKaushik7:
…indpoints

**What I did**
Updated threshold checking in a way that Exceed Threshold syslog message gets generated when the usage for any of the ACL group(bindpoint) exceeds the configured High Thresholds and gets cleared when usage of all the ACL group falls below configured low Threshold.

**Why I did it**
There is single threshold low and high range for multiple ACL group (similarly for ACL tables as well), but multiple bindpoints. The usage can be low for some bindpoints and higher for other bindpoints. which leads to generation of continuous syslog message generation with THRESHOLD_EXCEEDED and THRESHOLD_CLEAR messages  at the same time.  
Exceeded Threshold message should get generated only when resource count for any acl group exceeds the Max Thresholds and gets cleared when it falls below the lower threshold for all the acl groups.

**How I verified it**
Configure CRM configure thresholds for ACL group.
Create ACL group with different bindpoints such that
 for some bindpoints the usage is higher than the configured threshold and 
 for some bindpoints threshold is lower than the configured high threshold.
Configure crm polling interval
Observe that the Threshold Exceeded syslog messages are getting generated only once for the ACL group. Verify that no Threshold Clear messages are seen.
Now release the ACL resources and observe that the Clear messages gets generated only when the ACL group usage for all bindpoints falls below than the configured low threshold.
 
**Details if related**

### DIFF
--- a/orchagent/crmorch.cpp
+++ b/orchagent/crmorch.cpp
@@ -571,7 +571,6 @@ void CrmOrch::checkCrmThresholds()
                 maxUsedCounter = cnt.usedCounter;
                 maxAvailableCounter = cnt.availableCounter;
             }
-
         } // end of counters loop
         if ((maxUtilization >= res.highThreshold) && (res.exceededLogCounter < CRM_EXCEEDED_MSG_MAX))
         {


AmitKaushik7:
CI:rerun

AmitKaushik7:
Please retest this..

prsunny:
@AmitKaushik7 , as per this doc, "These messages must be suppressed after printing for 10 times.", so you will only see 10 messages since there is a check to suppress this in code. I understand the continuous 'EXCEED' message in this case, but why there is a continuous 'CLEAR'.

AmitKaushik7:
EXCEED Messages must be suppressed after printing 10 times and CLEAR message can appear max once until an EXCEED message appears for a CRM resource.
We will not see any issue with v4/v6 nhop,neigh,routes,ecmp resources as they have single threshold defined for each CRM resource. 

The issue is when multiple Bind Points/table entries share a common High/Low Threshold. For ACL group/table, we have single threshold for multiple bindpoints/ multiple tables.
Here we will observe alternate EXCEED and CLEAR messages continuously. 

Say we have High threshold configured for ACL table as 85% and low threshold as 50% which is common for all bindpoints(for a single CRM resource eg ACL table/group). 
If the usage for One bindpoint is 90% and usage for another bindpoint is 0%, then we will get alternate EXCEED-CLEAR messages (once for each bindpoint/entry)
With a reduced polling interval of 1, every second we will see these messages. 

The fix is to report EXCEED message only if the threshold of any bindpoint exceeds the High Threshold and report CLEAR when the threshold for all bindpoints of that resource are below the Low Threshold for the resource.

PR #1060, opened 2019-09-18
[aclorch]: Fix table name in counter table for mirror rules

stcheng:
In ACL combined mode, v4 and v6 rules are sharing the same
physical table while having separated configuration tables.
The daemon needs to use the configuration table name to store
the counter information.


yxieca:
checked another pr that has policer failure, looks like syncd is slow and is not able to process the request when we end the test.

PR #1071, opened 2019-09-27
[acltable]: Separate class AclTable from AclOrch file

stcheng:
Make the length of each file smaller for better maintenance.


orchagent/acltable.h
```diff
@@ -27,6 +29,78 @@ typedef enum
...
+    ACL_TABLE_PFCWD,
+    ACL_TABLE_CTRLPLANE,
+    ACL_TABLE_DTEL_FLOW_WATCHLIST,
+    ACL_TABLE_DTEL_DROP_WATCHLIST
+} acl_table_type_t;
+
 typedef map<string, acl_stage_type_t> acl_stage_type_lookup_t;
 
-#endif /* SWSS_ACLTABLE_H */
+class AclOrch;
```
prsunny: I think it is better to include 'aclorch.h' in this file as that would a superset for other things as well.

stcheng: I think it's generally not a good practice to mutually include two header files.

prsunny: We have portsorch.h included in most header files. But aside from that, aclorch.h has most of the references for acltable.h.. My point is here as per description, the aim is to reduce the file size and not separate out the functionality.

stcheng: sorry i didn't quite get it. i think there's no duplication between the acltable.h and aclorch.h files.

stcheng:
retest this please

orchagent/acltable.h
daall: I agree Shuotian, I don't think having aclorch include acltable and vice-versa is a good idea. 

From what I can tell the only dependency that AclTable has on AclOrch is `m_mirrorTableCapabilities[ACL_TABLE_MIRRORV6]` and `m_isCombinedMirrorV6Table` in create(). I think it would be cleaner to pass those two variables into create() (or the constructor if you think they might be useful in other methods as well) and then remove the reference to m_pAclOrch entirely. That way there's no cyclical dependency between AclTable and AclOrch.

orchagent/acltable.h
```diff
@@ -27,6 +29,78 @@ typedef enum
...
+    ACL_TABLE_L3,
+    ACL_TABLE_L3V6,
+    ACL_TABLE_MIRROR,
+    ACL_TABLE_MIRRORV6,
+    ACL_TABLE_MIRROR_DSCP,
+    ACL_TABLE_PFCWD,
+    ACL_TABLE_CTRLPLANE,
+    ACL_TABLE_DTEL_FLOW_WATCHLIST,
+    ACL_TABLE_DTEL_DROP_WATCHLIST
+} acl_table_type_t;
```
daall: Can this be an enum class?

orchagent/acltable.h
```diff
@@ -1,14 +1,16 @@
...
+#include <map>
+#include <memory>
 #include <set>
 #include <string>
 #include <vector>
-#include <map>
+
+#include "observer.h"
 
 using namespace std;
```
daall: Can you remove this from the header like we've started doing with the other classes in OA?

orchagent/acltable.h
```diff
@@ -1,14 +1,16 @@
...
```
daall: I think you can drop the TODO below this line.

orchagent/acltable.h
```diff
@@ -27,6 +29,78 @@ typedef enum
     ACL_STAGE_EGRESS
 } acl_stage_type_t;
```
daall: Same thing as below, can we use enum class?

orchagent/acltable.h
```diff
@@ -27,6 +29,78 @@ typedef enum
...
+
 typedef map<string, acl_stage_type_t> acl_stage_type_lookup_t;
 
-#endif /* SWSS_ACLTABLE_H */
+class AclOrch;
+class AclRule;
+
+class AclTable {
+    sai_object_id_t m_oid;
+    AclOrch *m_pAclOrch;
```
daall: Can you explicitly make these private?

orchagent/acltable.h
```diff
@@ -27,6 +29,78 @@ typedef enum
...
+    // Link the ACL table with a port, for future bind or unbind
+    void link(sai_object_id_t portOid);
+    // Add or overwrite a rule into the ACL table
+    bool add(shared_ptr<AclRule> newRule);
+    // Remove a rule from the ACL table
+    bool remove(string rule_id);
+    // Remove all rules from the ACL table
+    bool clear();
+    // Update table subject to changes
+    void update(SubjectType, void *);
```
daall: Can you add variable names to this declaration? It's not clear what's going on here.
```suggestion
    void update(SubjectType type, void *context);
```

orchagent/acltable.h
```diff
@@ -27,6 +29,78 @@ typedef enum
...
-#endif /* SWSS_ACLTABLE_H */
+class AclOrch;
+class AclRule;
+
+class AclTable {
+    sai_object_id_t m_oid;
+    AclOrch *m_pAclOrch;
+public:
+    string id;
+    string description;
```
daall: Does OA need to store the description? I don't think it's used anywhere.

orchagent/acltable.h
```diff
@@ -27,6 +29,78 @@ typedef enum
...
+
+    AclTable(AclOrch *aclOrch)
+        : m_pAclOrch(aclOrch)
+        , type(ACL_TABLE_UNKNOWN)
+        , m_oid(SAI_NULL_OBJECT_ID)
+        , stage(ACL_STAGE_INGRESS)
+    {}
+
+    sai_object_id_t getOid() { return m_oid; }
+    string getId() { return id; }
```
daall: We're not really consistent about using getters vs. public variables here. We should either:
1. Make all the variables public and don't bother with getters, or
2. Make all the variables private and include (inline) getters for them

My personal preference is option 2, but I'm fine with either as long as we're consistent (1 is probably going to be easier as far as code changes go).

orchagent/acltable.h
```diff
@@ -27,6 +29,78 @@ typedef enum
...
+    acl_stage_type_t stage;
+
+    // Map port oid to group member oid
+    std::map<sai_object_id_t, sai_object_id_t> ports;
+    // Map rule name to rule data
+    map<string, shared_ptr<AclRule>> rules;
+    // Set to store the ACL table port alias
+    set<string> portSet;
+    // Set to store the not cofigured ACL table port alias
+    set<string> pendingPortSet;
```
daall: Probably doesn't make a huge difference since we're not storing very much data, but can we use unordered maps/sets here instead of the normal maps/sets?

orchagent/acltable.cpp
```diff
@@ -0,0 +1,436 @@
...
+bool AclTable::add(shared_ptr<AclRule> newRule)
+{
+    SWSS_LOG_ENTER();
+
+    string rule_id = newRule->getId();
+    auto ruleIter = rules.find(rule_id);
+    if (ruleIter != rules.end())
+    {
+        // If ACL rule already exists, delete it first
+        if (ruleIter->second->remove())
```
daall: Can we have logging or something similar to the other methods if remove() fails?

orchagent/acltable.cpp
```diff
@@ -0,0 +1,436 @@
...
+        // TODO: deal with port removal scenario
+    }
+}
+
+// TODO: make bind/unbind symmetric
+bool AclTable::bind(sai_object_id_t portOid)
+{
+    SWSS_LOG_ENTER();
+
+    assert(ports.find(portOid) != ports.end());
```
daall: Is this intentional? I know you didn't originally write AclTable but do you know if there's a reason assert is called here rather than performing some logging and returning false?

orchagent/acltable.cpp
```diff
@@ -0,0 +1,436 @@
...
+    if (type == ACL_TABLE_UNKNOWN || stage == ACL_STAGE_UNKNOWN)
+        return false;
+
+    if (portSet.empty() && pendingPortSet.empty())
+        return false;
+
+    return true;
+}
+
+bool AclTable::create()
```
daall: Just a note, this method is kinda large and there are a lot of conditional checks for different types of tables and it gets a bit confusing in places. I think it would be worth breaking this up into private helpers to handle different table types, but that is also a pretty big change and probably out of the scope of this PR. Can you leave a note about it?

PR #1075, opened 2019-09-30
[orchagent] Adds swss support for drop counters

daall:
Adds swss support for drop counters
- Creates a new orchestrator for managing debug counters
- Adds utility functions for managing debug counters

**How I verified it**
A set of vswitch tests is included.
```
=============================================================================================== test session starts ===============================================================================================
platform linux2 -- Python 2.7.15+, pytest-3.3.0, py-1.8.0, pluggy-0.6.0 -- /usr/bin/python2
cachedir: .cache
rootdir: /home/daall/dev/sonic-swss/tests, inifile:
collected 14 items                                                                                                                                                                                                

test_drop_counters.py::TestDropCounters::test_deviceCapabilitiesTablePopulated remove extra link dummy
PASSED                                                                                                                       [  7%]
test_drop_counters.py::TestDropCounters::test_flexCounterGroupInitialized PASSED                                                                                                                            [ 14%]
test_drop_counters.py::TestDropCounters::test_createAndRemoveDropCounterBasic PASSED                                                                                                                        [ 21%]
test_drop_counters.py::TestDropCounters::test_createAndRemoveDropCounterReversed PASSED                                                                                                                     [ 28%]
test_drop_counters.py::TestDropCounters::test_createCounterWithInvalidCounterType PASSED                                                                                                                    [ 35%]
test_drop_counters.py::TestDropCounters::test_createCounterWithInvalidDropReason PASSED                                                                                                                     [ 42%]
test_drop_counters.py::TestDropCounters::test_addReasonToInitializedCounter PASSED                                                                                                                          [ 50%]
test_drop_counters.py::TestDropCounters::test_removeReasonFromInitializedCounter PASSED                                                                                                                     [ 57%]
test_drop_counters.py::TestDropCounters::test_addDropReasonMultipleTimes PASSED                                                                                                                             [ 64%]
test_drop_counters.py::TestDropCounters::test_addInvalidDropReason PASSED                                                                                                                                   [ 71%]
test_drop_counters.py::TestDropCounters::test_removeDropReasonMultipleTimes PASSED                                                                                                                          [ 78%]
test_drop_counters.py::TestDropCounters::test_removeNonexistentDropReason PASSED                                                                                                                            [ 85%]
test_drop_counters.py::TestDropCounters::test_removeInvalidDropReason PASSED                                                                                                                                [ 92%]
test_drop_counters.py::TestDropCounters::test_createAndDeleteMultipleCounters PASSED                                                                                                                        [100%]

=========================================================================================== 14 passed in 113.65 seconds ===========================================================================================
```

**Details if related**
Depends on:
- ~~Azure/sonic-sairedis#520~~
- ~~Azure/sonic-swss#1093~~
- Azure/sonic-sairedis#531 (for vstests)


utilities/Makefile.am
```diff
@@ -0,0 +1 @@
+SUBDIRS = flex_counter debug_counter
```
qiluo-msft: >flex_counter [](start = 10, length = 12)

Where is this?

configure.ac
```diff
@@ -97,6 +97,9 @@ AC_CONFIG_FILES([
     swssconfig/Makefile
     cfgmgr/Makefile
     tests/Makefile
+    utilities/Makefile
```
qiluo-msft: >utilities [](start = 4, length = 9)

We already have sonic-utilities, another `utilities` is confusing.

orchagent/orchdaemon.cpp
```diff
@@ -185,6 +185,15 @@ bool OrchDaemon::init()
...
+    vector<string> debug_counter_tables = {
+        CFG_DEBUG_COUNTER_TABLE_NAME,
+        CFG_DEBUG_COUNTER_DROP_REASON_TABLE_NAME
+    };
+
+    DebugCounterOrch *debug_counter_orch = new DebugCounterOrch(m_configDb, debug_counter_tables, 1000);
+
+    
+
     /*
```
qiluo-msft: Too many empty lines?

orchagent/Makefile.am
```diff
@@ -53,11 +55,12 @@ orchagent_SOURCES = \
...
             watermarkorch.cpp \
             policerorch.cpp \
-            chassisorch.cpp
+            chassisorch.cpp \
+            debugcounterorch.cpp
 
 orchagent_CFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_SAI)
 orchagent_CPPFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_SAI)
-orchagent_LDADD = -lnl-3 -lnl-route-3 -lpthread -lsairedis -lswsscommon -lsaimeta -lsaimetadata
+orchagent_LDADD = -lnl-3 -lnl-route-3 -lpthread -lsairedis -lswsscommon -lsaimeta -lsaimetadata $(utildir)/flex_counter/libflexcounter.la $(utildir)/debug_counter/libdebugcounter.la
```
qiluo-msft: >libdebugcounter [](start = 163, length = 15)

If it is not shared, we could directly use source code instead of creating a library.

orchagent/debugcounterorch.cpp
```diff
@@ -0,0 +1,577 @@
...
+                task_status = removeDropReason(counter_name, drop_reason);
+            }
+            else
+            {
+                SWSS_LOG_ERROR("Unknown operation type %s\n", op.c_str());
+            }
+        } 
+        else 
+        {
+            SWSS_LOG_WARN("DebugCounterOrch received update from unknown table '%s'", table_name.c_str());
```
qiluo-msft: >SWSS_LOG_WARN [](start = 12, length = 13)

error?

orchagent/debugcounterorch.cpp
```diff
@@ -0,0 +1,577 @@
...
+    for (auto const &counter_type : debug_counter_type_lookup)
+    {
+        string num_counters = std::to_string(getSupportedDebugCounterAmounts(counter_type.second));
+
+        string drop_reasons;
+        if (counter_type.first == PORT_INGRESS_DROPS || counter_type.first == SWITCH_INGRESS_DROPS) 
+        {
+            drop_reasons = supported_ingress_drop_reasons;
+        } 
+        else
```
qiluo-msft: >  [](start = 12, length = 1)

remove trailing spaces.

orchagent/debugcounterorch.cpp
```diff
@@ -0,0 +1,577 @@
...
+    SWSS_LOG_ENTER();
+
+    if (free_drop_counters.find(counter_name) != free_drop_counters.end()) 
+    {
+        SWSS_LOG_DEBUG("Debug counter '%s' is in free counter table", counter_name.c_str());
+        return;
+    }
+
+    SWSS_LOG_DEBUG("Adding debug counter '%s' to free counter table", counter_name.c_str());
+    free_drop_counters.emplace(make_pair(counter_name, counter_type));
```
qiluo-msft: >make_pair [](start = 31, length = 9)

No need to make_pair

orchagent/debugcounterorch.cpp
```diff
@@ -0,0 +1,577 @@
...
+}
+
+void DebugCounterOrch::deleteFreeCounter(const string &counter_name) 
+{
+    SWSS_LOG_ENTER();
+
+    if (free_drop_counters.find(counter_name) == free_drop_counters.end()) 
+    {
+        SWSS_LOG_DEBUG("Debug counter %s does not exist", counter_name.c_str());
```
qiluo-msft: >SWSS_LOG_DEBUG [](start = 8, length = 14)

error?

orchagent/debugcounterorch.cpp
```diff
@@ -0,0 +1,577 @@
...
+    // to either: a) dispatch to different handlers in doTask or b) dispatch to
+    // different helper methods in this method.
+
+    try 
+    {
+        string counter_type = getDebugCounterType(attributes);
+        addFreeCounter(counter_name, counter_type);
+        reconcileFreeDropCounters(counter_name);
+    } 
+    catch (const std::exception& e)
```
qiluo-msft: >exception [](start = 22, length = 9)

Why you need to catch general exception?

orchagent/debugcounterorch.cpp
```diff
@@ -0,0 +1,577 @@
...
+    {
+        SWSS_LOG_DEBUG("Debug counter '%s' is in free counter table", counter_name.c_str());
+        return;
+    }
+
+    SWSS_LOG_DEBUG("Adding debug counter '%s' to free counter table", counter_name.c_str());
+    free_drop_counters.emplace(make_pair(counter_name, counter_type));
+}
+
+void DebugCounterOrch::deleteFreeCounter(const string &counter_name)
```
qiluo-msft: >deleteFreeCounter [](start = 23, length = 17)

Not used?

orchagent/debugcounterorch.cpp
```diff
@@ -0,0 +1,577 @@
...
+}
+
+// createDropCounter creates a new drop counter in the SAI and installs a
+// flex counter to poll the counter data.
+//
+// If SAI initialization fails or flex counter installation fails then this
+// method will throw an exception.
+void DebugCounterOrch::createDropCounter(const string &counter_name, const string &counter_type, const unordered_set<string> &drop_reasons) 
+{
+    std::unique_ptr<DropCounter> counter = std::unique_ptr<DropCounter>(new DropCounter(counter_name, counter_type, drop_reasons));
```
qiluo-msft: >std::unique_ptr<DropCounter> [](start = 4, length = 28)

`auto` is less verbose

utilities/debug_counter/drop_counter.cpp
```diff
@@ -0,0 +1,296 @@
...
+    
+    try
+    {
+        this->drop_reasons.emplace(drop_reason);
+        this->updateDropReasonsInSAI();
+    }
+    catch (const std::exception& e) 
+    {
+        this->drop_reasons.erase(drop_reason);
+        throw e;
```
qiluo-msft: >throw e; [](start = 8, length = 8)

`throw;`

utilities/debug_counter/drop_counter.cpp
```diff
@@ -0,0 +1,296 @@
...
+// Multiple calls to this function are guaranteed to return the same reasons
+// (assuming the device has not been rebooted between calls). The order of
+// the list of reasons is not guaranteed to be the same between calls.
+//
+// If the device does not support querying drop reasons, this method will 
+// return an empty list.
+vector<string> getSupportedDropReasons(sai_debug_counter_attr_t drop_reason_type) 
+{
+    sai_s32_list_t drop_reason_list;
+    int32_t        supported_reasons[100];
```
qiluo-msft: >100 [](start = 37, length = 3)

magic number

utilities/debug_counter/debug_counter.cpp
```diff
@@ -0,0 +1,99 @@
...
+                                                    debug_counter_attributes) != SAI_STATUS_SUCCESS) {
+        SWSS_LOG_ERROR("Failed to create debug counter '%s'", this->name.c_str());
+        throw std::runtime_error("Failed to create debug counter");
+    }
+
+    SWSS_LOG_DEBUG("Created debug counter '%s' with OID=%lu", this->name.c_str(), debug_counter_id);
+    this->counter_id = debug_counter_id;
+}
+
+void DebugCounter::removeDebugCounterFromSAI()
```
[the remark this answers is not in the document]
daall: I'll have to verify but yes, the counters should be reconciled upon warm reboot.

utilities/Makefile.am
daall: Oops, I forgot to remove this when I split that into #1093

configure.ac
daall: Kind of related to your feedback below, I was trying to group related modules/source files rather than dump them all in the orchagent directory.  Do you (or anyone else reviewing this PR) have any suggestions for a better way of organizing this, or do you think I should just place everything in orchagent?

orchagent/debug_counter/debug_counter.cpp
```diff
@@ -0,0 +1,99 @@
...
+DebugCounter::~DebugCounter()
+{
+    SWSS_LOG_ENTER();
+}
+
+// serializeDebugCounterMetadata populates ONLY the type field in
+// serialized_attributes.
+//
+// Behavior is undefined if the length of serialized_attributes is less than 1.
+void DebugCounter::serializeDebugCounterMetadata(sai_attribute_t *serialized_attributes)
```
yxieca: Why not use sai_attribute_t& serialized_attribute for serialization since you are only putting values in the first entry anyways?

orchagent/debug_counter/debug_counter.h
```diff
@@ -0,0 +1,66 @@
...
+#define PORT_EGRESS_DROPS    "PORT_EGRESS_DROPS"
+#define SWITCH_INGRESS_DROPS "SWITCH_INGRESS_DROPS"
+#define SWITCH_EGRESS_DROPS  "SWITCH_EGRESS_DROPS"
+
+static const std::unordered_map<std::string, sai_debug_counter_type_t> debug_counter_type_lookup = {
+    { PORT_INGRESS_DROPS, SAI_DEBUG_COUNTER_TYPE_PORT_IN_DROP_REASONS },
+    { PORT_EGRESS_DROPS,  SAI_DEBUG_COUNTER_TYPE_PORT_OUT_DROP_REASONS },
+    { SWITCH_INGRESS_DROPS, SAI_DEBUG_COUNTER_TYPE_SWITCH_IN_DROP_REASONS },
+    { SWITCH_EGRESS_DROPS, SAI_DEBUG_COUNTER_TYPE_SWITCH_OUT_DROP_REASONS }
+};
```
yxieca: I only see this static variable used in debug_counter.cpp. I think it is best define this variable in the cpp file.

If in the future this static variable needs to be shared, you can add helper get function to provide access.

orchagent/debug_counter/drop_counter.cpp
```diff
@@ -0,0 +1,298 @@
...
+extern sai_debug_counter_api_t *sai_debug_counter_api;
+
+const int maxDropReasons = 100;
+
+// If initialization fails, this constructor will throw a runtime error.
+DropCounter::DropCounter(const string& counter_name, const string& counter_type, const unordered_set<string>& drop_reasons)
+        : DebugCounter(counter_name, counter_type), drop_reasons(drop_reasons)
+{
+    SWSS_LOG_ENTER();
+    this->initializeDropCounterInSAI();
```
yxieca: When add counter exception happens, who is going to catch it?

orchagent/debug_counter/drop_counter.cpp
```diff
@@ -0,0 +1,298 @@
...
+DropCounter::~DropCounter()
+{
+    SWSS_LOG_ENTER();
+    try
+    {
+        DebugCounter::removeDebugCounterFromSAI();
+    }
+    catch (const std::exception& e)
+    {
+        SWSS_LOG_ERROR("Failed to remove drop counter '%s' from SAI", this->name.c_str());
```
yxieca: If exception only mean early return, why not use return code? So that you don't have to worry any caller not catching the exception?

orchagent/debug_counter/debug_counter.h
daall: It's used in debugcounterorch.cpp as well but I like your suggestion better. :)

orchagent/debug_counter/drop_counter.cpp
daall: There are two main reasons I used exceptions instead of return codes:
1) the Drop Counter constructor can fail, and you can't use return codes in constructors, and
2) it made the code simpler in a lot of places, mostly where the drop counters are involved. Returning error codes would involve bubbling up the return code through a series of helper methods whereas throwing the exception allows us to catch it and handle it only at the top-level, which is really what we care about anyways.

I agree though that the exception behavior isn't as explicit as return codes, which can lead to issues like the comment you have above if we're not careful. I think C++ has a "noexcept" qualifier - let me see if that would help here.

daall: It might be helpful for us to have (just to make it easier for us to read :)) but it won't prevent this. I think the simplest solution is to move the try-catch up to doTask, but let me know if you still think that error codes are the way to go.

yxieca: I see. I am not 100% against exceptions.

I've been burnt by exception not caught within API boundaries and reaches other side of the fence. We need to be careful and explicit. I like you thoughts of helping future coders not fall into the trap.

orchagent/debug_counter/drop_counter.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+    sai_attribute_t index_attribute;
+    index_attribute.id = SAI_DEBUG_COUNTER_ATTR_INDEX;
+    if (sai_debug_counter_api->get_debug_counter_attribute(this->counter_id, 1, &index_attribute) != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to get stat for debug counter '%s'", this->name.c_str());
+        throw runtime_error("Failed to get debug counter stat");
+    }
+
+    auto index = index_attribute.value.u32;
```
yxieca: How do we know that index is not out of range?

orchagent/debug_counter/drop_counter.h
```diff
@@ -0,0 +1,107 @@
...
+    { ACL_EGRESS_SWITCH,    SAI_IN_DROP_REASON_ACL_EGRESS_SWITCH }
+};
+
+const std::unordered_map<std::string, sai_out_drop_reason_t> egress_drop_reason_lookup = {
+    { L2_ANY,              SAI_OUT_DROP_REASON_L2_ANY },
+    { EGRESS_VLAN_FILTER,  SAI_OUT_DROP_REASON_EGRESS_VLAN_FILTER },
+    { L3_ANY,              SAI_OUT_DROP_REASON_L3_ANY },
+    { L3_EGRESS_LINK_DOWN, SAI_OUT_DROP_REASON_L3_EGRESS_LINK_DOWN },
+};
+
```
yxieca: Can we move static variables to cpp files and provide look up / comparison helper functions?

orchagent/debug_counter/drop_counter.h
```diff
@@ -0,0 +1,107 @@
...
+    { ACL_INGRESS_LAG,      SAI_IN_DROP_REASON_ACL_INGRESS_LAG },
+    { ACL_INGRESS_VLAN,     SAI_IN_DROP_REASON_ACL_INGRESS_VLAN },
+    { ACL_INGRESS_RIF,      SAI_IN_DROP_REASON_ACL_INGRESS_RIF },
+    { ACL_INGRESS_SWITCH,   SAI_IN_DROP_REASON_ACL_INGRESS_SWITCH },
+    { ACL_EGRESS_PORT,      SAI_IN_DROP_REASON_ACL_EGRESS_PORT },
+    { ACL_EGRESS_LAG,       SAI_IN_DROP_REASON_ACL_EGRESS_LAG },
+    { ACL_EGRESS_VLAN,      SAI_IN_DROP_REASON_ACL_EGRESS_VLAN },
+    { ACL_EGRESS_RIF,       SAI_IN_DROP_REASON_ACL_EGRESS_RIF },
+    { ACL_EGRESS_SWITCH,    SAI_IN_DROP_REASON_ACL_EGRESS_SWITCH }
+};
```
yxieca: Can we move static variables to cpp files and provide look up / comparison helper functions?

orchagent/debugcounterorch.cpp
```diff
@@ -0,0 +1,589 @@
...
+        return;
+    }
+
+    auto it = consumer.m_toSync.begin();
+    while (it != consumer.m_toSync.end())
+    {
+        KeyOpFieldsValuesTuple t = it->second;
+        string key = kfvKey(t);
+        string op = kfvOp(t);
+        std::vector<FieldValueTuple> fvt = kfvFieldsValues(t);
```
yxieca: fvt seems a bit out of league compare to key/op variables naming. Consider val instead?

orchagent/debugcounterorch.h
```diff
@@ -0,0 +1,96 @@
...
+    // we cannot create a drop counter without any drop reasons, we keep track
+    // of these counters in this table.
+    std::unordered_map<std::string, std::string> free_drop_counters;
+
+    // free_drop_reasons are drop reasons that have been added by a user
+    // that do not have a counter associated with them yet. Because we
+    // cannot add drop reasons to a counter that doesn't exist yet,
+    // we keep track of the reasons in this table.
+    std::unordered_map<std::string, std::unordered_set<std::string>> free_drop_reasons;
+};
```
yxieca: Question, looks to me that free_drop_counters and free_drop_reasons are counter and reason waiting to be paired with each other. Is this designed so that user can create them in any order?

What if user add, say 2 drop reasons for a same counter,  and the 2 drop reasons were both added after counter, so the operation would like:

add_counter(c);
add_counter(c):
add_reason(r1);
add_reason(r2);

Can you add 2 identical counters to the free_counters? If not, then would r2 be left never be paired?

tests/test_drop_counters.py
```diff
@@ -0,0 +1,526 @@
...
+        asic_state_table = swsscommon.Table(self.asic_db, ASIC_STATE_TABLE)
+        flex_counter_table = swsscommon.Table(self.flex_db, FLEX_COUNTER_TABLE)
+
+        name = 'ADD_TEST'
+        reason1 = 'L3_ANY'
+        reason2 = 'ACL_ANY'
+
+        self.create_drop_counter(name, SWITCH_INGRESS_DROPS)
+        self.add_drop_reason(name, reason1)
+        self.add_drop_reason(name, reason2)
```
yxieca: Do you expect reason2 to be created? Some comments in your test flow would help a lot.

tests/test_drop_counters.py
```diff
@@ -0,0 +1,526 @@
...
+        self.create_drop_counter(name2, PORT_INGRESS_DROPS)
+        self.add_drop_reason(name2, reason2)
+        time.sleep(5)
+
+        assert len(flex_counter_table.getKeys()) > 0
+
+        asic_keys = asic_state_table.getKeys()
+        assert len(asic_keys) == 2
+        for key in asic_keys:
+            assert (self.checkASICState(key, ASIC_COUNTER_PORT_IN_TYPE, [reason1]) or self.checkASICState(key, ASIC_COUNTER_PORT_IN_TYPE, [reason2]))
```
yxieca: I think your code can deal with either create_counter(), then create_reason(), or create_reason(), then create_counter(). But your test seems only testing the first ordering?

yxieca:
My comments are mostly cosmetic issues. Overall the change looks pretty good.

Thanks for adding these comments in source code to tell tricky choices. These comments helps a lot.

Add some comments to state the purpose of test steps would also help readers.

I found that you used quite a bit 'this->' to specify scope in your code. This is not wrong per-se. Just curious why are you doing compiler's job? :-)

orchagent/debug_counter/drop_counter.cpp
daall: I'm not sure that I understand the feedback here. `index` is just the value returned from the call to the SAI.

orchagent/debugcounterorch.h
daall: Once r1 is added then it will be reconciled against c and a drop counter will be created. When we add r2 we will first check and see that the counter exists, then the reason will be added to the counter directly.

tests/test_drop_counters.py
daall: There is a "createAndRemoveDropCounterBasic" which does the first ordering and "createAndRemoveDropCounterReversed" which does the second ordering. I've added comments to clarify this. :)

orchagent/debug_counter/drop_counter.cpp
```diff
@@ -0,0 +1,370 @@
...
+        SWSS_LOG_DEBUG("Drop reason '%s' already present on '%s'", drop_reason.c_str(), name.c_str());
+        return;
+    }
+
+    try
+    {
+        drop_reasons.emplace(drop_reason);
+        updateDropReasonsInSAI();
+    }
+    catch (const std::exception& e)
```
qiluo-msft: >std::exception [](start = 17, length = 14)

Why you need to catch general exception?

orchagent/debug_counter/drop_counter.cpp
yxieca: The index is used in all the if blocks below. How do we know that the index will come out valid after (BASE+index) operation?

daall: The indices are managed by the SAI. By the time we reach this step, the counter has already been initialized, so we can assume the index is valid. However, there is a limit to the number of enums that are defined in the SAI. I don't believe any of the vendors support enough counters for this to be an issue, but I added a check just to save someone a lot of confusion in the future. :)

daall:
retest this please

yxieca:
retest this please

PR #1078, opened 2019-10-02
[mirrororch]: Toggle the mirror session when moving between VLAN/non-VLAN

stcheng:
As a workaround, this commit prevents SET API get executed when toggling
between a VLAN monitor port and a non-VLAN monitor port.

Until SAI validation is completed, this workaround is needed.

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -816,6 +816,12 @@ bool MirrorOrch::updateSessionType(const string& name, MirrorEntry& session)
 {
     SWSS_LOG_ENTER();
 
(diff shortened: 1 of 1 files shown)


orchagent/mirrororch.cpp
```diff
@@ -816,6 +816,12 @@ bool MirrorOrch::updateSessionType(const string& name, MirrorEntry& session)
 {
     SWSS_LOG_ENTER();
 
+    deactivateSession(name, session);
+    activateSession(name, session);
+
+    return true;
+
+#ifdef 0
```
prsunny: #if 0 ?

PR #1081, opened 2019-10-03
not using namespace in header files

wendani:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


wendani:
Fixed compile error. https://github.com/Azure/sonic-swss/pull/1081#pullrequestreview-297027053

wendani:
retest this please

PR #1085, opened 2019-10-08
[portmgrd/portsyncd] Refactor code so that portsyncd does not deal with configurations at run time

zhenggen-xu:
Refactor code so that portsyncd does not deal with configurations at run time

portmgrd will deal with configurations for both linux side and appDB.
Added vs test cases for port config changes

**What I did**
Refactor code so that portsyncd does not deal with configurations at run time
portmgrd will deal with configurations for both linux side and appDB.
Added vs test cases for port config changes

**Why I did it**
Only one daemon is handling port configurations from cfg_db to app-db/linux.

**How I verified it**
vs test cases

**Details if related**
-- Initial configurations is not moved yet.
-- port_config.ini dependencies is not removed yet.


qiluo-msft:
Please resolve the conflict first

prsunny:
@zhenggen-xu , is this an enhancement or you encountered any issue with current model?

PR #1087, opened 2019-10-10
sonic-swss/orchagent:  Add new protocol trap name support

jpxjlrldgit:
sonic-swss/orchagent:  Add new protocol trap name support in CoPP

Added new protocol name support so these names can be used in copp_config.json file and properly parsed.

This is to support new feature requirement.

Verified Sonic image on Broadcom target by specifying these new traps in copp_config.json file and verified these new protocols are trapped and policed as they are configured in the copp_config.json file.

Signed-off-by: jingping.xie@broadcom.com

### DIFF
--- a/orchagent/copporch.cpp
+++ b/orchagent/copporch.cpp
@@ -69,7 +69,9 @@ static map<string, sai_hostif_trap_type_t> trap_id_map = {
     {"router_custom_range", SAI_HOSTIF_TRAP_TYPE_ROUTER_CUSTOM_RANGE_BASE},
     {"l3_mtu_error", SAI_HOSTIF_TRAP_TYPE_L3_MTU_ERROR},
     {"ttl_error", SAI_HOSTIF_TRAP_TYPE_TTL_ERROR},
-    {"udld", SAI_HOSTIF_TRAP_TYPE_UDLD}
+    {"udld", SAI_HOSTIF_TRAP_TYPE_UDLD},
+    {"bfd", SAI_HOSTIF_TRAP_TYPE_BFD},
+    {"bfdv6", SAI_HOSTIF_TRAP_TYPE_BFDV6}
 };
 
 static map<string, sai_packet_action_t> packet_action_map = {


orchagent/copporch.cpp
```diff
@@ -69,7 +69,11 @@ static map<string, sai_hostif_trap_type_t> trap_id_map = {
     {"router_custom_range", SAI_HOSTIF_TRAP_TYPE_ROUTER_CUSTOM_RANGE_BASE},
     {"l3_mtu_error", SAI_HOSTIF_TRAP_TYPE_L3_MTU_ERROR},
     {"ttl_error", SAI_HOSTIF_TRAP_TYPE_TTL_ERROR},
-    {"udld", SAI_HOSTIF_TRAP_TYPE_UDLD}
+    {"udld", SAI_HOSTIF_TRAP_TYPE_UDLD},
+    {"bfd", SAI_HOSTIF_TRAP_TYPE_BFD},
+    {"bfdv6", SAI_HOSTIF_TRAP_TYPE_BFDV6},
+    {"src_nat_miss", SAI_HOSTIF_TRAP_TYPE_SNAT_MISS},
```
prsunny: Almost all above follows the suffix of SAI type, so shouldn't it be `snat_miss` instead of `src_nat_miss`. Just an observation.

jpxjlrldgit:
Is the 3 failed baseline results or it is due to my new submit?  The code change submitted should not be triggered since there is no change in copp_config.json file, i.e these new trap names are not used or reference.  Any one know what's going on?

Thanks,
Jingping

orchagent/copporch.cpp
jpxjlrldgit: Yes. You are correct that we should stick to the common naming convention.  However, since this is done by other team, I will try to remove these two new trap from the submission if I can make the code change.  Or else I will have to abandon this submit and do a separate submission later.  I will try to abandon this one first.
Thanks,
Jingping

jpxjlrldgit:
Hi,

I have addressed the comment by removing the two new trap with debatable naming and leave them for future submission.
Thanks

prsunny:
retest this please

PR #1088, opened 2019-10-10
Update Route test to check for added entry

prsunny:
**What I did**
Update Route test to check for added entry

**Why I did it**
Sometimes, default routes keeps coming on eth0 during the route add and this can result in test failure

**How I verified it**
Run VS test

**Details if related**
For CRM check log case:
```
marker = '=== start marker 2019-10-10T05:25:38.550022 ==='
err_log = 'IPV6_ROUTE THRESHOLD_EXCEEDED for TH_FREE', expected_cnt = 1

    def check_syslog(dvs, marker, err_log, expected_cnt):
        (exitcode, num) = dvs.runcmd(['sh', '-c', "awk \'/%s/,ENDFILE {print;}\' /var/log/syslog | grep \"%s\" | wc -l" % (marker, err_log)])
>       assert num.strip() == str(expected_cnt)
E       AssertionError: assert '2' == '1'
E         - 2
E         + 1
```

### DIFF
--- a/tests/test_crm.py
+++ b/tests/test_crm.py
@@ -120,10 +120,10 @@ class TestCrm(object):
         assert new_avail_counter == avail_counter
 
         marker = dvs.add_log_marker()
-        dvs.runcmd("crm config polling interval 2")
+        dvs.runcmd("crm config polling interval 3")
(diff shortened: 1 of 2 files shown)


prsunny:
retest this please

PR #1089, opened 2019-10-10
[aclorch]: Check for existing mirror table only when creating a new table

daall:
[aclorch]: Check for existing mirror table only when creating a new table

**What I did**
I modified aclorch to check for existing mirror tables only when creating new tables, not when modifying/replacing an existing table.

**Why I did it**
Configlets may modify the bindings for MIRROR tables, this change makes it easier to do so. It also makes the behavior for updates to MIRROR tables consistent with the behavior with updates for other table types.

**How I verified it**
I updated the bindings for the EVERFLOW ACL table and verified that the group was deleted and created, as expected.

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -2459,7 +2459,7 @@ bool AclOrch::addAclTable(AclTable &newTable, string table_id)
                 mirror_type = TABLE_TYPE_MIRROR;
             }
 
-            if (table_type == ACL_TABLE_MIRRORV6 && !m_mirrorV6TableId.empty()))
+            if (table_type == ACL_TABLE_MIRRORV6 && !m_mirrorV6TableId.empty())
             {
                 mirror_type = TABLE_TYPE_MIRRORV6;
             }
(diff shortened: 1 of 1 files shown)


orchagent/aclorch.cpp
```diff
@@ -2446,6 +2446,21 @@ bool AclOrch::addAclTable(AclTable &newTable, string table_id)
...
+    {
+        // If ACL table is new, check for the existence of current mirror tables
+        // Note: only one table per mirror type can be created
+        auto table_type = newTable.type;
+        if (table_type == ACL_TABLE_MIRROR || table_type == ACL_TABLE_MIRRORV6)
+        {
+            if ((table_type == ACL_TABLE_MIRROR && !m_mirrorTableId.empty()) ||
+                    (table_type == ACL_TABLE_MIRRORV6 && !m_mirrorV6TableId.empty()))
+            {
+                SWSS_LOG_ERROR("Mirror table has already been created");
```
stcheng: could you still print out the mirror table type?
```
WSS_LOG_ERROR("Mirror table %s has already been created",
table_type.c_str());
```

daall:
retest this please

lguohan:
retest this please

PR #1093, opened 2019-10-12
[utilities] Create utility classes for interacting with flex counters

daall:
[utilities] Create utility classes for interacting with flex counters
- Adds utility classes for interacting with flex counters
- Updates makefiles to support new utilities

**Details if related**
Related to/spun off from: Azure/sonic-swss#1075


lguohan:
are you planning to leverage this flex counter manager class for other places using flex counter?

daall:
> are you planning to leverage this flex counter manager class for other places using flex counter?

That's my goal, yes. Eventually I would want to refactor other daemons like the pfcorch and portsorch to use FlexCounterManager instead of re-implementing the same redis messaging schemes. To do that I'll have to extend the functionality to support plugins and such, but I wanted to limit the scope to features needed for the debug counters for this initial PR.

daall:
retest this please

orchagent/flex_counter/dynamic_flex_counter_manager.cpp
```diff
@@ -0,0 +1,88 @@
...
+        const sai_object_id_t object_id,
+        const CounterType counter_type,
+        const string &counter_stat)
+{
+    SWSS_LOG_ENTER();
+
+    auto counter_stats = object_stats.find(object_id);
+    if (counter_stats == object_stats.end()) 
+    {
+        SWSS_LOG_DEBUG("Could not find flex stat '%s' on object '%s'",
```
lguohan: is this warning level? such event should not happen in normal operation.

orchagent/flex_counter/flex_counter_manager.cpp
```diff
@@ -0,0 +1,203 @@
...
+// serializeCounterStats turns a set of stats into a format suitable for FLEX_COUNTER_DB.
+string FlexCounterManager::serializeCounterStats(
+        const unordered_set<string> &counter_stats) const
+{
+    SWSS_LOG_ENTER();
+
+    string stats_string;
+    for (const auto &stat : counter_stats) 
+    {
+        stats_string.append(",");
```
qiluo-msft: >stats_string.append(","); [](start = 8, length = 25)

Don't append before first stat.

orchagent/flex_counter/dynamic_flex_counter_manager.cpp
```diff
@@ -0,0 +1,88 @@
...
+        const CounterType counter_type,
+        const string &counter_stat)
+{
+    SWSS_LOG_ENTER();
+
+    auto counter_stats = object_stats.find(object_id);
+    if (counter_stats == object_stats.end())
+    {
+        unordered_set<string> new_stats = { counter_stat };
+        object_stats.emplace(std::make_pair(object_id, new_stats));
```
qiluo-msft: >std::make_pair [](start = 29, length = 14)

No need to make_pair

orchagent/flex_counter/dynamic_flex_counter_manager.cpp
```diff
@@ -0,0 +1,88 @@
...
+        counter_stats = object_stats.find(object_id);
```
qiluo-msft: >find [](start = 37, length = 4)

No need to `find`, you get an iterator when `emplace`.

orchagent/flex_counter/dynamic_flex_counter_manager.h
```diff
@@ -0,0 +1,33 @@
+#ifndef SWSS_UTIL_DYNAMIC_FLEX_COUNTER_MANAGER_H
+#define SWSS_UTIL_DYNAMIC_FLEX_COUNTER_MANAGER_H
+
+#include "flex_counter_manager.h"
+
+// DynamicFlexCounterManager allows users to manage a group of flex counters
+// where the objects have highly variable sets of stats to track.
+class DynamicFlexCounterManager : public FlexCounterManager
```
qiluo-msft: >DynamicFlexCounterManager [](start = 6, length = 25)

Name is confusing here. Is it dynamic counter or dynamic manager? For my reading, I guess it is `FlexCounterStatManager`

qiluo-msft: Another option is to combine FlexCounterManager with DynamicFlexCounterManager into one class.

---
In reply to: [340903276](https://github.com/Azure/sonic-swss/pull/1093#discussion_r340903276) [](ancestors = 340903276)

orchagent/flex_counter/flex_counter_manager.cpp
```diff
@@ -0,0 +1,203 @@
...
+            group_name.c_str());
+}
+
+string FlexCounterManager::getFlexCounterTableKey(
+        const string &group_name, 
+        const sai_object_id_t object_id) const
+{
+    SWSS_LOG_ENTER();
+
+    return group_name + ':' + sai_serialize_object_id(object_id);
```
lguohan: can you use TableBase::tableNameSeparatorMap:COUNTERS_DB?

orchagent/flex_counter/dynamic_flex_counter_manager.h
daall: I'll go with your FlexCounterStatManager suggestion! The reason they're separate is related to this discussion: https://github.com/Azure/sonic-swss/pull/1093#issuecomment-548103642. Most of the existing daemons only set the stats once, so there's no need to maintain a mapping with all the objects and their stats.

orchagent/flex_counter/flex_counter_manager.cpp
daall: This is actually intentional but may have been unclear. I've added a comment to clarify the purpose.

qiluo-msft: I know your intention. I mean don't append before first component, not a good practice.

daall:
retest this please

yxieca:
Noticed some coding style issue. I think the reference type is usually declared in the format of "type& var" instead of "type &var", especially when var is the only variable defined.

Am I too old school? :-)

orchagent/flex_counter/flex_counter_manager.cpp
```diff
@@ -0,0 +1,209 @@
...
+    flex_counter_group_table->set(group_name, field_values);
+
+    SWSS_LOG_DEBUG("Initialized flex counter group '%s'.", group_name.c_str());
+}
+
+FlexCounterManager::~FlexCounterManager()
+{
+    SWSS_LOG_ENTER();
+
+    flex_counter_group_table->del(group_name);
```
yxieca: Logically, I think the above line should be after the for loop.

daall:
retest vs please

daall:
retest this please

daall:
retest this please

daall:
retest this please

PR #1098, opened 2019-10-17
Quote input strings before constructing a command line

qiluo-msft:
Quote input strings from function arguments before constructing a command line, so prevent of the bash script input injection.

`man dash` shows below section, which indicates the double quotes and escaping is enough for general use.

```
Double Quotes
 Enclosing characters within double quotes preserves the literal meaning of all characters except dollarsign ($), backquote (`), and backslash (\).  The backslash inside double quotes is historically weird, and serves to quote only the following characters:
       $ ` " \ <newline>.
 Otherwise it remains literal.
```

### DIFF
--- a/cfgmgr/shellcmd.h
+++ b/cfgmgr/shellcmd.h
@@ -23,8 +23,8 @@
 
 static inline std::string shellquote(const std::string& str)
 {
-    static const std::regex re("([\\'])");
-    return "$'" + std::regex_replace(str, re, "\\$1") + "'";
+    static const std::regex re("([$`\"\\\n])");
+    return "\"" + std::regex_replace(str, re, "\\$1") + "\"";
 }
 
 #endif /* __SHELLCMD__ */
--- a/cfgmgr/shellcmd.h
+++ b/cfgmgr/shellcmd.h
@@ -1,6 +1,9 @@
 #ifndef __SHELLCMD__
 #define __SHELLCMD__
 
+#include <iomanip>
+#include <regex>
+
 #define IP_CMD               "/sbin/ip"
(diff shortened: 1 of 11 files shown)


qiluo-msft:
Retest this please

pavel-shirshov:
Why?

pavel-shirshov:
```
sonicbld@5de366018e18:~$ cat /sys/class/net/"$(echo hi)"/operstate    
cat: /sys/class/net/hi/operstate: No such file or directory
```
I still can execute commands even with quotes

qiluo-msft:
Retest this please

cfgmgr/shellcmd.h
```diff
@@ -18,4 +21,10 @@
     }                                           \
 })
 
+static inline std::string shellquote(const std::string& str)
+{
+    static const std::regex re("([$`\"\\\n])");
+    return "\"" + std::regex_replace(str, re, "\\$1") + "\"";
```
pavel-shirshov: How can we be sure this regex is enough for everything?
Should we better create functions to check the input parameters and run them?
Our input parameters types are easy:
1. ip address
2. network mask
3. network device
4. number
5. string enum

qiluo-msft: Update the PR comment at the top with answer.
I agree case-by-case protection is also possible, this one is for general purpose.

banagiri:
@qiluo-msft in the api removeHostVlanMember,  is the cmds having cmds as input instead of inner? or is this expected ?
 
`cmds << BASH_CMD " -c " << shellquote(cmds.str());` 
instead of 
`cmds << BASH_CMD " -c " << shellquote(inner.str());`

PR #1099, opened 2019-10-18
Allow buffer profile apply after init

wendani:
**What I did**

**Why I did it**

**How I verified it**
On brcm dut

**Details if related**

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -579,7 +579,7 @@ task_process_status BufferOrch::processQueue(Consumer &consumer)
     }
     else
     {
-        SWSS_LOG_ERROR("Queue profile '%s' was inserted after BufferOrch init", key.c_str());
+        SWSS_LOG_NOTICE("Queue profile '%s' was inserted after BufferOrch init", key.c_str());
     }
 
     return task_process_status::task_success;
@@ -666,7 +666,7 @@ task_process_status BufferOrch::processPriorityGroup(Consumer &consumer)
     }
     else
     {
-        SWSS_LOG_ERROR("PG profile '%s' was inserted after BufferOrch init", key.c_str());
+        SWSS_LOG_NOTICE("PG profile '%s' was inserted after BufferOrch init", key.c_str());
     }
 
     return task_process_status::task_success;
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -579,7 +579,19 @@ task_process_status BufferOrch::processQueue(Consumer &consumer)
     }
     else
     {
-        SWSS_LOG_NOTICE("Queue profile '%s' was inserted after BufferOrch init", key.c_str());
+        // If a buffer queue profile is not in the initial CONFIG_DB BUFFER_QUEUE table
+        // at BufferOrch object instantiation, it is considered being applied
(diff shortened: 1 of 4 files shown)


pavel-shirshov:
Can you please check this PR https://github.com/Azure/sonic-swss/pull/515 ?
I've introduced this change to avoid some deadlock

lguohan:
@pavel-shirshov , he only change the log level, is that a concern?

pavel-shirshov:
@lguohan My concert is about following.
Previously it was considered bad to apply buffer profiles after buffer orch init because it could make a deadlock in the asic. So that messages wasn't supposed to show anytime. Now I see that we want to reduce the syslog level for the message, which looks like it is going to be ok to apply the profiles after init. So the change will not produce any issue, but we might get a deadlock if we apply profiles after bufferorch init time.

orchagent/bufferorch.cpp
```diff
@@ -579,7 +579,19 @@ task_process_status BufferOrch::processQueue(Consumer &consumer)
     }
     else
```
renukamanavalan: Curious!
When will this port get into m_ready_list ?
May be when the port is marked up ?

orchagent/bufferorch.cpp
```diff
@@ -579,7 +579,19 @@ task_process_status BufferOrch::processQueue(Consumer &consumer)
...
+        // at BufferOrch object instantiation, it is considered being applied
+        // at run time, and, in this case, is not tracked in the m_ready_list. It is up to
+        // the application to guarantee the set order that the buffer queue profile
+        // should be applied to a physical port before the physical port is brought up to
+        // carry traffic. Here, we alert to application through syslog when such a wrong
+        // set order is detected.
+        for (const auto &port_name : port_names)
+        {
+            if (gPortsOrch->isPortAdminUp(port_name)) {
+                SWSS_LOG_ERROR("Queue profile '%s' applied after port %s is up", key.c_str(), port_name.c_str());
```
renukamanavalan: In this case, what is the resultant state ?

The port is non functional and assigned buffers go unused ?

orchagent/bufferorch.cpp
```diff
@@ -579,7 +579,19 @@ task_process_status BufferOrch::processQueue(Consumer &consumer)
...
+        // If a buffer queue profile is not in the initial CONFIG_DB BUFFER_QUEUE table
+        // at BufferOrch object instantiation, it is considered being applied
+        // at run time, and, in this case, is not tracked in the m_ready_list. It is up to
+        // the application to guarantee the set order that the buffer queue profile
+        // should be applied to a physical port before the physical port is brought up to
+        // carry traffic. Here, we alert to application through syslog when such a wrong
+        // set order is detected.
+        for (const auto &port_name : port_names)
+        {
+            if (gPortsOrch->isPortAdminUp(port_name)) {
```
renukamanavalan: When PortOrch receives port-up, can the PortOrch query bufferOrch, if port is allotted with buffers or not ? If not, it could fire a self timer and delay the processing of the port UP, until appropriate buffers are allotted. Each time the process is deferred, it could post a WARNING LOG.

wendani:
retest this please

orchagent/bufferorch.cpp
wendani: There will be persistent packet drops on that port

PR #1104, opened 2019-10-23
[mirrororch]: Add retry logic when deleting referenced mirror session

stcheng:
When a mirror session is referenced by an ACL rule, we cannot remove
the mirror session. However, once the ACL rule gets removed, a retry
would help clean up the to-be-removed mirror session.


yxieca:
retest this please

yxieca:
retest this please

PR #1106, opened 2019-10-24
[aclorch] if vendor does not implement ACL action capability quieries…

stepanblyschak:
… - use default capabilities (do not fail)

**What I did**

If ACL max action count query failed - use defaultAclActionsSupported for both INGRESS and EGRESS
If ACL actions query on INGRESS failed use defaultAclActionsSupported for INGRESS
If ACL actions query on EGRESS failed use defaultAclActionsSupported for EGRESS

Warn in any scenario that orchagent will assume defaults.

#ifdef SAI_SUPPORTS_OBJECT_API /* does not compile right now */
if enum values supported for ACL action query failed - assume all are supported
(e.g. action - PACKET_ACTION, values - DROP,FORWARD etc.)
#else
assume all values are supported
#end 

**Why I did it**

Seems to crash on many vendors because of lack of capability query support

**How I verified it**

Verify on system which supports attributes,
Simulate if system does not support and verify defaults are pushed to db

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -115,6 +115,20 @@ static acl_stage_type_lookup_t aclStageLookUp =
     {STAGE_EGRESS,  ACL_STAGE_EGRESS }
 };
 
+static const acl_capabilities_t defaultAclActionsSupported =
+{
+    {
+        ACL_STAGE_INGRESS,
+        {
+            SAI_ACL_ACTION_TYPE_PACKET_ACTION,
+            SAI_ACL_ACTION_TYPE_MIRROR_INGRESS
+        }
+    },
+    {
+        ACL_STAGE_EGRESS, {}
+    }
+};
+
 static acl_ip_type_lookup_t aclIpTypeLookup =
 {
     { IP_TYPE_ANY,         SAI_ACL_IP_TYPE_ANY },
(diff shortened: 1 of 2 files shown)


orchagent/aclorch.cpp
```diff
@@ -2129,69 +2143,71 @@ void AclOrch::queryAclActionCapability()
...
+                               "failed to query supported %s ACL actions", stage_str);
+            }
 
-        fvVector.emplace_back(field, acl_action_value_stream.str());
-        m_switchTable.set("switch", fvVector);
+            // put capabilities in state DB
+            putAclActionCapabilityInDB(stage);
+        }
     }
+    else if (status == SAI_STATUS_NOT_IMPLEMENTED)
```
prsunny: Please have just the `else `case to populate the default table

orchagent/aclorch.cpp
```diff
@@ -2251,15 +2311,28 @@ void AclOrch::queryAclActionAttrEnumValues(const string &action_name,
...
-            SWSS_LOG_THROW("sai_query_attribute_enum_values_capability failed for %s",
-                           action_name.c_str());
+            for (size_t i = 0; i < values.count; i++)
+            {
+                m_aclEnumActionCapabilities[acl_action].insert(values.list[i]);
+            }
         }
-
-        for (size_t i = 0; i < values.count; i++)
+        else if (status == SAI_STATUS_NOT_IMPLEMENTED)
```
prsunny: Same as above comment

orchagent/aclorch.cpp
stepanblyschak: I changed it. But that means SAI does not return valid status code

orchagent/aclorch.cpp
prsunny: Can you address this as well?

stepanblyschak: Fixed

PR #1107, opened 2019-10-24
[portsyncd] Remove the port_config.ini dependency from portsyncd

zhenggen-xu:
[portsyncd] Remove the port_config.ini dependency from portsyncd

portsyncd will rely on configDB port table, not from port_config.ini

port_config.ini is used when we convert minigraph to configDB, then
everything should come from configDB.

Also for the DPB feature, we are going to deprecate port_config.ini

**What I did**
Remove the port_config.ini dependency from portsyncd

**Why I did it**
port_config.ini is used when we convert minigraph to configDB, then
everything should come from configDB.

Also for the DPB feature, we are going to deprecate port_config.ini

**How I verified it**
vs-test

**Details if related**

### DIFF
--- a/portsyncd/portsyncd.cpp
+++ b/portsyncd/portsyncd.cpp
@@ -35,9 +35,9 @@ bool g_init = false;
 
 void usage()
 {
-    cout << "Usage: portsyncd [-p port_config.ini]" << endl;
-    cout << "       -p port_config.ini: import port lane mapping" << endl;
-    cout << "                           use configDB data if not specified" << endl;
+    cout << "Usage: portsyncd" << endl;
+    cout << "       port lane mapping is from configDB" << endl;
+    cout << "       this program will exit if configDB does not contain that info" << endl;
 }
 
 void handlePortConfigFile(ProducerStateTable &p, string file, bool warm);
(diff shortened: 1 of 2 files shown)


lguohan:
@qiluo-msft , do we have port_config.ini dependency with mellanox script?

portsyncd/portsyncd.cpp
```diff
@@ -91,11 +87,9 @@ int main(int argc, char **argv)
         if (!handlePortConfigFromConfigDB(p, cfgDb, warm))
         {
             // if port config is missing in ConfigDB
-            // attempt to use port_config.ini
-            if (!port_config_file.empty())
-            {
-                handlePortConfigFile(p, port_config_file, warm);
-            }
+            // program will exit with failure
+            SWSS_LOG_THROW("ConfigDB does not have port information, existing...");
```
prsunny: exiting - typo

zhenggen-xu:
@lguohan @daall @qiluo-msft  This PR is not intended to remove the port_config.ini file itself. It just to break the dependencies of portsyncd with port_config.ini.  port_config.ini would still exist for other purposes like transceiver show etc.  I have some pending changes for the vs test cases and PR with sonic-buildimage along with this PR.  Will send out soon.

To remove port_config.ini completely,  we are taking graceful steps and keep backwards compatible. We will introduce platform.json which was described in DPB design (https://github.com/Azure/SONiC/pull/450).  In case configDB does not have the information available, we will get from platform.json,  but in the platforms/HWSKUs that still uses port_config.ini, the logic will keep the same.  It will up to the platform vendors to add the platform.json for each platform/HWSKU and completely remove the port_config.ini.

In short, this PR should not break any functions unless you miss the PORT TABLE in configDB.

daall:
Thanks for the explanation @zhenggen-xu! :rocket:

PR #1111, opened 2019-10-29
[fix] Use the same storm detection condition for queue occupancy non-zero case as the zero case

wendani:

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/pfc_detect_broadcom.lua
+++ b/orchagent/pfc_detect_broadcom.lua
@@ -67,7 +67,7 @@ for i = n, 1, -1 do
                     pfc_on2off_last = tonumber(pfc_on2off_last)
 
                     -- Check actual condition of queue being in PFC storm
-                    if (occupancy_bytes > 0 and packets - packets_last == 0 and pfc_rx_packets - pfc_rx_packets_last > 0) or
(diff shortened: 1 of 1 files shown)


orchagent/pfc_detect_broadcom.lua
```diff
@@ -67,11 +67,10 @@ for i = n, 1, -1 do
                     pfc_on2off_last = tonumber(pfc_on2off_last)
 
                     -- Check actual condition of queue being in PFC storm
-                    if (occupancy_bytes > 0 and packets - packets_last == 0 and pfc_rx_packets - pfc_rx_packets_last > 0) or
+                    if (pfc_rx_packets - pfc_rx_packets_last > 0 and pfc_on2off - pfc_on2off_last == 0 and queue_pause_status_last == 'true' and queue_pause_status == 'true') or
                         -- DEBUG CODE START. Uncomment to enable
```
yxieca: It would be safer to remove this comments and the one 2 lines below. They are no longer safe to be here to comment out the change after this change.

abdosi:
@lguohan/@yxieca Can you cherry-pick this in master/202012.

Have cherry-picked in 201911.

cc @neethajohn and @rlhui

qiluo-msft:
Cherry-picked into 202012

qiluo-msft:
Cherry-picked into master. Please submit PR to master in future because PR checkers are better supported on master.

wendani:
> Cherry-picked into master. Please submit PR to master in future because PR checkers are better supported on master.

🙆‍♂️

PR #1113, opened 2019-11-01
Fix traceroute issue

Pterosaur:
**What I did**

Change the MAC address of vxlan bridge to the switch's.

**Why I did it**

The vxlan device will not response the traceroute message if its MAC address isn't same with the switch's.

**How I verified it**

To compare the MAC address of switch's with the Brvxlanxxxx's. We expect that they are exactly same.

Signed-off-by: zegan@microsoft.com

### DIFF
--- a/cfgmgr/vxlanmgr.cpp
+++ b/cfgmgr/vxlanmgr.cpp
@@ -24,6 +24,9 @@ using namespace swss;
 #define VXLAN "vxlan"
 #define VXLAN_IF "vxlan_if"
 
+#define SWITCH "switch"
+#define VXLAN_ROUTER_MAC "vxlan_router_mac"
+
 #define VXLAN_NAME_PREFIX "Vxlan"
 #define VXLAN_IF_NAME_PREFIX "Brvxlan"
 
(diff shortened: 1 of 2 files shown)


cfgmgr/vxlanmgr.cpp
```diff
@@ -432,6 +452,27 @@ bool VxlanMgr::isVxlanStateOk(const std::string & vxlanName)
     return false;
 }
 
+std::string VxlanMgr::getSwitchMacAddress()
```
prsunny: Can you please rename this to` getVxlanRouterMacAddress()`. There is a different `SwitchMacAddress `and it would be confusing with the name.

Pterosaur: Thank you for you suggestion, Done it.

Pterosaur:
retest it please

Pterosaur:
retest this please

cfgmgr/vxlanmgr.cpp
```diff
@@ -432,6 +452,27 @@ bool VxlanMgr::isVxlanStateOk(const std::string & vxlanName)
...
+    {
+        auto itr = std::find_if(
+            temp.begin(),
+            temp.end(),
+            [](const FieldValueTuple &fvt) { return fvt.first == VXLAN_ROUTER_MAC; });
+        if (itr != temp.end() && !(itr->second.empty()))
+        {
+            SWSS_LOG_DEBUG("Mac address %s is ready", itr->second.c_str());
+            return itr->second;
+        }
```
prsunny: So if the switch table is present but vxlan router mac is not present, then I wouldn't recommend retry. In this case the api can be changed to return true and no need to set mac to kernel interface. It will be auto set to switch mac.

Pterosaur: Done

cfgmgr/vxlanmgr.cpp
```diff
@@ -282,6 +296,16 @@ bool VxlanMgr::doVxlanCreateTask(const KeyOpFieldsValuesTuple & t)
         // Suspend this message util the vrf is created
         return false;
     }
+    
+    // If the mac address has been set
+    auto macAddress = getVxlanRouterMacAddress();
+    if (macAddress.first)
```
wendani: if (!macAddress.first) 

Typo? @Pterosaur

Pterosaur: Hi Wenda, Thanks for your comment. This is my fault.
I have solve this problem at : https://github.com/Azure/sonic-swss/pull/1142

PR #1117, opened 2019-11-05
[teammgrd]: Use port-channel id as LACP key so that multiple port-channels from the peer can be distinguished

sandeep-kulambi:
**What I did**
When a member port is added to port-channel, ensure the lacp-key is updated by using the Port-channel id value as the key value.

**Why I did it**
If LACP key is not set, then the peer will not be able to distinguish the ports which are connected to different port-channels, as it will receive the actor key (lacp key) as 0 for all the ports from different port-channels.

Sonic== 2links ===3rd party device
On Sonic 2 POs are created.:
Link1 on Sonic is configured on PO1 
Link2 on Sonic is configured on PO2.
On 3rd party device single PO is created and both links are added to same PO.
All the links are in sync on both POs on Sonic and also in PO on 3rd party device.

**How I verified it**
Configure the devices as mentioned above and see on the 3rd party device only one of the port comes up in the PO and the other one stays down.

### DIFF
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -447,12 +447,26 @@ task_process_status TeamMgr::addLagMember(const string &lag, const string &membe
 
     stringstream cmd;
     string res;
+    string PoId, PoId2;
+    size_t found = lag.find('l');
+    if (found == string::npos)
+    {
+        SWSS_LOG_ERROR("Failed to parse %s", lag.c_str());
+        return task_failed;
+    }
+    PoId = lag.substr(found+1);
+    /* Ignore all 0's in the Port-channel id*/
+    PoId2 = PoId.erase(0, min(PoId.find_first_not_of('0'), PoId.size()-1));
 
     // Set admin down LAG member (required by teamd) and enslave it
     // ip link set dev <member> down;
     // teamdctl <port_channel_name> port add <member>;
     cmd << IP_CMD << " link set dev " << shellquote(member) << " down; ";
-    cmd << TEAMDCTL_CMD << " " << shellquote(lag) << " port add " << shellquote(member);
+    cmd << TEAMDCTL_CMD << " " << shellquote(lag) << " port config update " << shellquote(member) 
+        << " '{\"lacp_key\":"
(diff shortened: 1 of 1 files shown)


prsunny:
In the current use-case which I'm aware, both sides are configured as two different port-channels. In your example, why is one side configured as single PC? Is this any specific use-case like MLAG or something?

sandeep-kulambi:
This is a case of mis-configuration where ports from different PO’s on one device (in this case SONiC) is connected to same PO on the other device (3rd party device). In such a case LACP is expected to identify such errors and not allow the mis-configured port to be part of the LAG.

cfgmgr/teammgr.cpp
```diff
@@ -447,12 +447,26 @@ task_process_status TeamMgr::addLagMember(const string &lag, const string &membe
...
+    string PoId, PoId2;
+    size_t found = lag.find('l');
+    if (found == string::npos)
+    {
+        SWSS_LOG_ERROR("Failed to parse %s", lag.c_str());
+        return task_failed;
+    }
+    PoId = lag.substr(found+1);
+    /* Ignore all 0's in the Port-channel id*/
+    PoId2 = PoId.erase(0, min(PoId.find_first_not_of('0'), PoId.size()-1));
```
lguohan: this part of logic can be improved, it should try to find the first digit, then use stol to get an integert and use it later.

lguohan:
retest this please

pavel-shirshov:
Bug https://github.com/Azure/sonic-buildimage/issues/4009

lmcerbo-ka:
> This approach will not work. Because lacp key_id is 16-bit number. What if someone creates PortChannel with id greater than 65535?
> Also, What if we have PortChannel01 and PortChannel1? It is the same key id.
> 
> I suggest to introduce an additional field to config db for LAG: key_id.
> It allows us to put desired key_id in the configuration.
> Another approach would be to restrict PortChannel format to have a proper format.

We are testing LAG in our lab as well with SONiC and having a configurable key_id as suggested by pavel-shirshov would be more "future" proof (IMHO).

PR #1118, opened 2019-11-05
[vs_test]fix fdb test failed randomly

tylerlinp:
**What I did**
1. Change  test_FdbAddedAfterMemberCreated not to assert the count of entries but to check the existance of entry added by itself, because other entry may be learned after flush. The former test case does't clean the config, that makes more chance to learn fdb entry.
2. Fix the bug is_fdb_entry_exists didn't match key. Add a function get_vlan_oid to change vlanid to bvid. Change mac format to match in asic_db.

**Why I did it**
fdb test failed often

**How I verified it**
pass vs tests

**Details if related**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -548,6 +548,20 @@ class DockerVirtualSwitch(object):
 
         return iface_2_bridge_port_id
 
+    def get_vlan_oid(self, asic_db, vlan_id):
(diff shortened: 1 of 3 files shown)


kcudnik:
thanks for fixing this, issue root cause was that virtual switch was learning other fdb entries during tests, and number of entries could be different each time

PR #1119, opened 2019-11-06
Fix pfc watchdog compile error

wendani:

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


wendani:
@stepanblyschak

PR #1120, opened 2019-11-06
[bufferorch] Fixed buffer and buffer profile attributes types accoring to changes in SAI 1.5

vsenchyshyn:
Signed-off-by: Vitaliy Senchyshyn <vsenchyshyn@barefootnetworks.com>

**What I did**

Changed type of buffer and buffer profile attributes from u32 to u64 as this was done in SAI 1.5

**Why I did it**

As some buffer attributes migrated to from u32 to u64 sonic just initializes u32 field only and the most significant 4 bytes contain garbage. I such case we receive very big and incorrect number for buffer size, threshold, etc.

**How I verified it**

Issued **sudo config qos reload** command on BFN platform and saw that the numbers passed to SDK are the same as in the config. 

**Details if related**


vsenchyshyn:
@stcheng @prsunny Please review this

PR #1059, opened 2019-09-18
Natmgrd changes in sonic-swss sub module to support NAT feature.

AkhileshSamineni:
Added natmgr, and intfmgr changes.

Link to NAT HLD : https://github.com/Azure/SONiC/blob/master/doc/nat/nat_design_spec.md

Depends on:
sonic-swss : https://github.com/Azure/sonic-swss/pull/1126 and https://github.com/Azure/sonic-swss/pull/1125
sonic-swss-common :  https://github.com/Azure/sonic-swss-common/pull/304 
sonic-linux-kernel : https://github.com/Azure/sonic-linux-kernel/pull/100 
sonic-sairedis : https://github.com/Azure/sonic-sairedis/pull/519

 Signed-off-by: Akhilesh Samineni <akhilesh.samineni@broadcom.com>

PR #1125, opened 2019-11-11
Orchagent changes in sonic-swss submodule to support NAT feature.

AkhileshSamineni:
Added orchagent and zone related changes.

Link to NAT HLD : https://github.com/Azure/SONiC/blob/master/doc/nat/nat_design_spec.md

Depends on:
sonic-swss : https://github.com/Azure/sonic-swss/pull/1059/ and https://github.com/Azure/sonic-swss/pull/1126
sonic-swss-common : Azure/sonic-swss-common#304
sonic-linux-kernel : Azure/sonic-linux-kernel#100
sonic-sairedis : Azure/sonic-sairedis#519 and https://github.com/Azure/sonic-sairedis/pull/546
      

orchagent/natorch.cpp
```diff
@@ -0,0 +1,4691 @@
...
+    auto executor   = new ExecutableTimer(m_natQueryTimer, this, "NAT_HITBIT_N_CNTRS_QUERY_TIMER");
+    Orch::addExecutor(executor);
+
+    /* Get the Maximum supported entries */
+    SWSS_LOG_INFO("Get the Maximum supported SNAT entries");
+    sai_status_t     status;
+    sai_attribute_t  attr;
+    memset(&attr, 0, sizeof(attr));
+    attr.id = SAI_SWITCH_ATTR_AVAILABLE_SNAT_ENTRY;
+    maxAllowedNatEntries = 0;
```
arlakshm: please rename this to variable as we are using this to check the number SNAT entries

orchagent/natorch.cpp
```diff
@@ -0,0 +1,4691 @@
...
+
+    status = sai_switch_api->get_switch_attribute(gSwitchId, 1, &attr);
```
arlakshm: please move this SAI call to when NAT is enabled from CLI

orchagent/natorch.cpp
```diff
@@ -0,0 +1,4691 @@
...
+
+bool NatOrch::addNatEntry(const IpAddress &ip_address, const NatEntryValue &entry)
+{
+    SWSS_LOG_ENTER();
+
+    /* Check the entry is present in cache */
+    if (m_natEntries.find(ip_address) != m_natEntries.end())
+    {
+        SWSS_LOG_INFO("Duplicate %s %s NAT entry with ip %s and it's translated ip %s, do nothing",
+                      entry.entry_type.c_str(), entry.nat_type.c_str(), ip_address.to_stri
```
arlakshm: May be can do this check at the beginning of the function

arlakshm: Why are counters being updated here ?

arlakshm: do we need to clean up the entries which are added to HW as well ?

arlakshm: As I understand, this notification comes when a L3 interface is added to NAT ZONE, so every time a interface is added to a NAT_ZONE. I have 2 questions:
 -  since all the conntrack entries are flushed, the entries in asic will also be removed right ? What will happen to the traffic, they will be dropped ?
 - When we need to flush of conntrack entries from natorch can this be done from natmgr instead ?

arlakshm: In the current SONiC architecture, the kernel entries are updated/added/deleted from orchagent. Can you move this functionality to natmgr.

orchagent/copporch.cpp
```diff
@@ -73,6 +73,8 @@ static map<string, sai_hostif_trap_type_t> trap_id_map = {
     {"udld", SAI_HOSTIF_TRAP_TYPE_UDLD},
     {"bfd", SAI_HOSTIF_TRAP_TYPE_BFD},
     {"bfdv6", SAI_HOSTIF_TRAP_TYPE_BFDV6}
+    {"src_nat_miss", SAI_HOSTIF_TRAP_TYPE_SNAT_MISS},
```
arlakshm: Can we make sure that these trap ids are created only on platform which support NAT

AkhileshSamineni: Done based on the feature check.

orchagent/natorch.cpp
AkhileshSamineni: renamed this variable to maxAllowedSNatEntries.

orchagent/natorch.cpp
AkhileshSamineni: Get operation on a unimplemented SAI attribute doesn't result in crash, so retaining in current changes as discussed.

orchagent/natorch.cpp
AkhileshSamineni: NAT entries have to be cached irrespective of NAT is enabled or not.

AkhileshSamineni: The existing dynamic entry is replaced by static entry as Static entry has higher priority, so decremented the dynamic and incremented the static counters.

AkhileshSamineni: Yes, it will also delete the entries from hardware.

AkhileshSamineni: OA is responsible for checking the hardware activity/hitbit of each of the entry in its cache. Since the cache is in OA code, it is the appropriate place to update the timeouts for the corresponding entries in the kernel.

AkhileshSamineni: This notification comes when do "sonic-clear nat translations" is executed. Yes all the entries in the asic are removed. Current traffic flows will see a drop and relearned again and added to asic.

OA is the component that tries to keep the Kernel conntrack entries and the hardware entries in sync. So, it is doing the flush entries.

arlakshm: I understand OA is checking the hitbit, but in the current SONiC architecture the kernel entries are no update from OA. The mgr modules normally updates the kernel entries

orchagent/intfsorch.cpp
```diff
@@ -1015,3 +1087,30 @@ void IntfsOrch::doTask(SelectableTimer &timer)
         }
     }
 }
+
+void IntfsOrch::getNatSupportedInfo()
```
prsunny: I see that you are repeating this same function in multiple places. Can you do this query one time, like in switchorch, and use a single API to check if NAT is supported.

orchagent/intfsorch.cpp
```diff
@@ -550,6 +587,21 @@ void IntfsOrch::doTask(Consumer &consumer)
...
+                if (!nat_zone.empty())
+                {
+                    nat_zone_id = (uint32_t)stoul(nat_zone);
+                    if ((m_nat_zone.find(alias) == m_nat_zone.end()) or
+                        (nat_zone_id != m_nat_zone[alias]))
+                    {
+                        m_nat_zone[alias] = nat_zone_id;
+                        if (isNatSupported)
+                        {
+                            setRouterIntfsNatZoneId(port, nat_zone_id);
```
prsunny: I don't understand this call. The router interface is created only in setIntf. So this API is always going to return `true `after its first check.  You are actually setting the zone in as part of Line 808. Can you please check?

orchagent/intfsorch.cpp
```diff
@@ -739,6 +796,21 @@ bool IntfsOrch::addRouterIntfs(sai_object_id_t vrf_id, Port &port)
     attr.value.u32 = port.m_mtu;
     attrs.push_back(attr);
 
+    if (isNatSupported)
```
prsunny: This is not required.

orchagent/intfsorch.cpp
```diff
@@ -739,6 +796,21 @@ bool IntfsOrch::addRouterIntfs(sai_object_id_t vrf_id, Port &port)
...
+    {
+        attr.id = SAI_ROUTER_INTERFACE_ATTR_NAT_ZONE_ID;
+        if (m_nat_zone.find(port.m_alias) == m_nat_zone.end())
+        {
+            attr.value.u32 = DEFAULT_NAT_ZONE_ID;
```
prsunny: This is not required, its the SAI default

orchagent/intfsorch.cpp
```diff
@@ -739,6 +796,21 @@ bool IntfsOrch::addRouterIntfs(sai_object_id_t vrf_id, Port &port)
...
+        }
+        else
+        {
+            attr.value.u32 = m_nat_zone[port.m_alias];
+        }
```
prsunny: You can revisit the check as if natzone is present in configuration, set the attribute.

orchagent/saihelper.cpp
```diff
@@ -41,9 +41,11 @@ sai_mirror_api_t*           sai_mirror_api;
 sai_fdb_api_t*              sai_fdb_api;
 sai_dtel_api_t*             sai_dtel_api;
 sai_bmtor_api_t*            sai_bmtor_api;
+sai_nat_api_t*              sai_nat_api;
 sai_samplepacket_api_t*     sai_samplepacket_api;
 sai_debug_counter_api_t*    sai_debug_counter_api;
 
+
```
prsunny: Please remove this extra line

orchagent/natorch.cpp
prsunny: Agree with Arvind.

AkhileshSamineni: The natmgr has no idea about the hardware activity of nat entries, so it cannot update the entries timeouts in kernel, if natmgr has to update the timeouts, OA has to inform the natmgr through separate communication channel which is an overhead and has it issues with scaling.

orchagent/Makefile.am
```diff
@@ -54,12 +54,14 @@ orchagent_SOURCES = \
...
             policerorch.cpp \
             sfloworch.cpp \
+            natorch.cpp \
             chassisorch.cpp \
             debugcounterorch.cpp
 
 orchagent_SOURCES += flex_counter/flex_counter_manager.cpp flex_counter/flex_counter_stat_manager.cpp
 orchagent_SOURCES += debug_counter/debug_counter.cpp debug_counter/drop_counter.cpp
 
+
```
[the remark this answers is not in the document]
AkhileshSamineni: Done.

orchagent/intfsorch.cpp
AkhileshSamineni: Ok. Moved this functionality to main.cpp file and updating the global variable 'gIsNatSupported'.

orchagent/copporch.h
```diff
@@ -50,6 +50,7 @@ class CoppOrch : public Orch
 protected:
     object_map m_trap_group_map;
     bool       enable_sflow_trap;
+    bool       isNatSupported;
```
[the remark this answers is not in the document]
AkhileshSamineni: Removed this variable and using 'gIsNatSupported' global variable.

orchagent/intfsorch.h
```diff
@@ -40,8 +41,13 @@ class IntfsOrch : public Orch
     void decreaseRouterIntfsRefCount(const string&);
 
     bool setRouterIntfsMtu(const Port &port);
+    bool setRouterIntfsNatZoneId(Port &port, uint32_t &nat_zone_id);
     bool setRouterIntfsAdminStatus(const Port &port);
+    void getNatSupportedInfo();
+
     std::set<IpPrefix> getSubnetRoutes();
+    std::map<string, uint32_t> m_nat_zone;
+    bool isNatSupported;
```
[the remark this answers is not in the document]
AkhileshSamineni: Removed this variable and using 'gIsNatSupported' global variable.

orchagent/intfsorch.cpp
AkhileshSamineni: Yes, removed it.

orchagent/saihelper.cpp
AkhileshSamineni: Done.

orchagent/intfsorch.cpp
AkhileshSamineni: yes, modified the code to set the attribute if it is present in cache.

orchagent/intfsorch.cpp
AkhileshSamineni: User configures the new 'nat_zone' even after rif is created, this call is used to set the new 'nat_zone'.

orchagent/intfsorch.cpp
AkhileshSamineni: Set operation on a unimplemented SAI attribute (SAI_ROUTER_INTERFACE_ATTR_NAT_ZONE_ID as of SAI 1.5) results in crash, to avoid it added this check here.

kirankella:
The current nexthop tracking in OA works on different SAI implementations (those which track and those which do not track the nexthop changes) though it will be sub-optimal on the SAI that tracks too. We can proceed as-is with the current OA changes till the SAI implementations converge.

AkhileshSamineni:
Retest this please.

orchagent/intfsorch.cpp
prsunny: Got it but the set of RIF attribute is currently handled in setIntf.  It is done for SUB_PORT currently but I think can be modified to set NAT zone for any RIF type. Another follow up is, why do we need m_nat_zone map seperate? Can this nat zone be part of the port structure, like m_rif_id?

orchagent/natorch.cpp
arlakshm: The NAT entries are not updated real time, polling for hit-bit happens every 30 seconds, so adding another hop will not have a big impact on the performance.

orchagent/intfsorch.cpp
AkhileshSamineni: Setting NAT zone for Physical, Vlan and Lag interfaces only, so not handled in setIntf.
Yes, nat zone can be part of port structure, made changes to address it.

orchagent/natorch.cpp
arlakshm: As discussed offline, lets keep the current code for now.  Let's open a issue to track this  change. We can make the code changes in the next release.

arlakshm:
@AkhileshSamineni, @kirankella,
There seems to conflicts with the with the PR. Please resolve the conflicts.

AkhileshSamineni:
> @AkhileshSamineni, @kirankella,
> There seems to conflicts with the with the PR. Please resolve the conflicts.

@arlakshm Resloved merge conflicts.

AkhileshSamineni:
retest this please.

AkhileshSamineni:
Retest this please.

AkhileshSamineni:
Retest this please.

AkhileshSamineni:
Retest this please.

kirankella:
retest vs please

kirankella:
retest vs please

PR #1127, opened 2019-11-11
Add new protocol trapping support and readjust trapping rules based on actual tests and industry practice in copp_config.json file 

jpxjlrldgit:
Component/file touched: sonic-swss/swssconfig/sample/00-copp.config.json

Added new protocols (e.g. STP, PVRST, BFD, etc.) to the CoPP config file for newly supported and community published features.
Modified CPU queues, policing parameters of various protocols based on actual system and scale test results and industry practice.
The changes made in this configuration file have been extensively/thoroughly tested by both development and SQA teams on many Broadcom device based platforms and no issue seen and observed expected behavior for newly supported features.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


lguohan:
this trap needs to be move to config db and let user to configure which trap, we should not add any more traps in this file any more.

PR #1131, opened 2019-11-13
[MultiDB]: vs test conftest.py needs copy config file to mounted path

dzhangalibaba:
* together with https://github.com/Azure/sonic-swss-common/pull/315
*vs test mount host:/var/run/redis-vs/{sw_name} to vs:var/run/redis/, so we need to copy the host:/var/run/redis/sonic-db/database_config.json to host:/var/run/redis-vs/{sw_name}/sonic-db/ and the vs can use this database_config.json file.
* I am not sure why the mount name is different from /var/run/redis/, if it is the same, then we don't need copy.  and  https://github.com/Azure/sonic-swss-common/pull/315 . is enough.

Signed-off-by: Dong Zhang d.zhang@alibaba-inc.com

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -221,6 +221,10 @@ class DockerVirtualSwitch(object):
                     network_mode="container:%s" % self.ctn_sw.name,
                     volumes={ self.mount: { 'bind': '/var/run/redis', 'mode': 'rw' } })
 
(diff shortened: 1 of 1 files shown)


dzhangalibaba:
will do changes in docker-sonic-vs images itself instead of using libswsscommon

PR #1143, opened 2019-12-04
[port/buffer] introduce a sync mechanism to protect port PG/queue from changes under PFC storm

stepanblyschak:
**What I did**
I introduce a sync mechanism to protect port PG/queue from changes under PFC storm.

**Why I did it**

Under warm reboot conditions on Mellanox platform, buffermgrd starts after BufferOrch gets initialized and sets initial state to DB; after reconciliation BufferOrch reacts on those changes and sets PG profile overriding PG profile set by PfcWdOrch.

**How I verified it**
Run PFC WD + warm reboot test cases.
Mannual restart of buffermgrd when port is under PFC storm - verified PG profile was not overriten.
After storm, PG profile was set according to config db.

**Details if related**

### DIFF
--- a/orchagent/pfcactionhandler.cpp
+++ b/orchagent/pfcactionhandler.cpp
@@ -441,7 +441,7 @@ PfcWdZeroBufferHandler::PfcWdZeroBufferHandler(sai_object_id_t port,
         return;
     }
 
-    lockUnlockPriorityGroupAndQueue(portInstance, true);
+    setPriorityGroupAndQueueLockFlag(portInstance, true);
 
     sai_attribute_t attr;
     attr.id = SAI_QUEUE_ATTR_BUFFER_PROFILE_ID;
@@ -536,18 +536,18 @@ PfcWdZeroBufferHandler::~PfcWdZeroBufferHandler(void)
         return;
     }
 
-    lockUnlockPriorityGroupAndQueue(portInstance, false);
+    setPriorityGroupAndQueueLockFlag(portInstance, false);
 }
 
-void PfcWdZeroBufferHandler::lockUnlockPriorityGroupAndQueue(Port& port, bool lock) const
+void PfcWdZeroBufferHandler::setPriorityGroupAndQueueLockFlag(Port& port, bool isLocked) const
 {
     // set lock bits on PG and queue
-    port.m_priority_group_lock[static_cast<size_t>(getQueueId())] = lock;
+    port.m_priority_group_lock[static_cast<size_t>(getQueueId())] = isLocked;
     for (size_t i = 0; i < port.m_queue_ids.size(); ++i)
     {
         if (port.m_queue_ids[i] == getQueue())
         {
-            port.m_queue_lock[i] = lock;
+            port.m_queue_lock[i] = isLocked;
         }
     }
     gPortsOrch->setPort(port.m_alias, port);
--- a/orchagent/pfcactionhandler.cpp
+++ b/orchagent/pfcactionhandler.cpp
@@ -542,7 +542,7 @@ PfcWdZeroBufferHandler::~PfcWdZeroBufferHandler(void)
 void PfcWdZeroBufferHandler::lockUnlockPriorityGroupAndPort(Port& port, bool lock) const
 {
     // set lock bits on PG and queue
-    port.m_priority_group_lock[static_cast<size_t>(getQueueId())] = false;
+    port.m_priority_group_lock[static_cast<size_t>(getQueueId())] = lock;
     for (size_t i = 0; i < port.m_queue_ids.size(); ++i)
     {
         if (port.m_queue_ids[i] == getQueue())
--- a/orchagent/pfcactionhandler.cpp
+++ b/orchagent/pfcactionhandler.cpp
@@ -434,6 +434,15 @@ PfcWdZeroBufferHandler::PfcWdZeroBufferHandler(sai_object_id_t port,
 {
     SWSS_LOG_ENTER();
 
+    Port portInstance;
+    if (!gPortsOrch->getPort(port, portInstance))
+    {
+        SWSS_LOG_ERROR("Cannot get port by ID 0x%" PRIx64, port);
+        return;
+    }
+
+    lockUnlockPriorityGroupAndPort(portInstance, true);
+
     sai_attribute_t attr;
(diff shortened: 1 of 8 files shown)


orchagent/pfcactionhandler.cpp
```diff
@@ -533,6 +535,22 @@ PfcWdZeroBufferHandler::~PfcWdZeroBufferHandler(void)
...
+    // set lock bits on PG and queue
+    port.m_priority_group_lock[static_cast<size_t>(getQueueId())] = lock;
+    for (size_t i = 0; i < port.m_queue_ids.size(); ++i)
+    {
+        if (port.m_queue_ids[i] == getQueue())
+        {
+            port.m_queue_lock[i] = lock;
+        }
+    }
+    gPortsOrch->setPort(port.m_alias, port);
```
qiluo-msft: >gPortsOrch->setPort(port.m_alias, port); [](start = 4, length = 40)

Is it really needed? I think port is already in gPortsOrch->m_portList.

orchagent/port.h
```diff
@@ -96,6 +96,16 @@ class Port
     std::vector<sai_object_id_t> m_priority_group_ids;
     sai_port_priority_flow_control_mode_t m_pfc_asym = SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_COMBINED;
     uint8_t m_pfc_bitmask = 0;
+    /*
+     * Following two bit vectors are used to lock
+     * the PG/queue from beeing changed in BufferOrch.
+     * The use case scenario is when PfcWdZeroBufferHandler
+     * sets zero buffer profile it should protect PG/queue
+     * from beeing overriten in BufferOrch.
```
qiluo-msft: typo

orchagent/port.h
```diff
@@ -96,6 +96,16 @@ class Port
     std::vector<sai_object_id_t> m_priority_group_ids;
     sai_port_priority_flow_control_mode_t m_pfc_asym = SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_COMBINED;
     uint8_t m_pfc_bitmask = 0;
+    /*
+     * Following two bit vectors are used to lock
+     * the PG/queue from beeing changed in BufferOrch.
```
qiluo-msft: >beeing [](start = 25, length = 6)

typo, and one more below.

orchagent/port.h
```diff
@@ -96,6 +96,16 @@ class Port
...
+     * The use case scenario is when PfcWdZeroBufferHandler
+     * sets zero buffer profile it should protect PG/queue
+     * from beeing overriten in BufferOrch.
+     */
+    std::vector<bool> m_queue_lock;
```
qiluo-msft: >m_queue_lock [](start = 22, length = 12)

In theory it is not a lock, it indicates a state.

orchagent/pfcactionhandler.cpp
```diff
@@ -533,6 +535,22 @@ PfcWdZeroBufferHandler::~PfcWdZeroBufferHandler(void)
         SWSS_LOG_ERROR("Failed to set buffer profile ID on queue 0x%" PRIx64 ": %d", getQueue(), status);
         return;
     }
+
+    lockUnlockPriorityGroupAndQueue(portInstance, false);
+}
+
+void PfcWdZeroBufferHandler::lockUnlockPriorityGroupAndQueue(Port& port, bool lock) const
```
qiluo-msft: >lockUnlockPriorityGroupAndQueue [](start = 29, length = 31)

The function name seems weird. I guess you mean 'setSomething'

orchagent/pfcactionhandler.cpp
stepanblyschak: It is needed, gPortsOrch->getPort returns a copy of Port

orchagent/port.h
stepanblyschak: It is not a PFC storm/nostorm state of the queue in case of PfcWdOrch uses acl handler and I think there should be no restriction to change pg or queue during storm when using acl handler?
Could you suggest more suitable stat definition for this case?
In general case it has nothing to do with PFC state of the queue, its just a set of bit flags used to protect the queue.

orchagent/pfcactionhandler.cpp
liat-grozovik: I agree. It seems that you use true to lock and false to unlock. IMO it is better to call the function setLock... and use true to lock and false to unlcok (the opposite of what you just did)

tests/mock_tests/portsorch_ut.cpp
```diff
@@ -310,4 +314,122 @@ namespace portsorch_test
...
+
+        // refill consumer
+        gPortsOrch->addExistingData(&portTable);
+
+        // Apply configuration :
+        //  create ports
+
+        static_cast<Orch *>(gPortsOrch)->doTask();
+
+        // Apply configuration
```
liat-grozovik: could it be this is duplicate as just you do the same in lines 365 to 368?

stepanblyschak: PortsOrch is not processing everything in single doTask.
There are two reasons why single doTask() is not enough:
1. Explicit return from doTask to give other orchs chance perform their tasks that depend on port creation - this is the case for PortsOrch, it skips any port attribute until it processes PortConfigDone, then, since it has all port mapping information it creates/removes ports and returns
2. Out of order tables execution - this is changed for PortsOrch in another PR

The comment above doTask() statement says that on first doTask is creates ports, on second applies port attributes.

orchagent/pfcactionhandler.cpp
stepanblyschak: Let me change to
```c++
void setPriorityGroupAndQueueLockFlag(Port& port, bool isLocked) const
```
As for "and use true to lock and false to unlcok" - this is exactly how it currently is.

stepanblyschak:
All tests failed with:
```
Dec 10 17:02:13.047998 f5cbc3ac3384 ERR #syncd: :- syncd_main: Runtime error: :- processEvent: api 'object_type_get_availability_query' is not implemented
```

Is there some mismatch between sairedis and swss versions?

stepanblyschak:
retest this please

liat-grozovik:
201811 label was removed. PR cannot be cherry-pick. Different PR will be provided soon.

Issue #763, opened 2019-01-18
teamd failed to start the port channel and teamsyncd failed to link to it


stcheng:
```
Jan 18 18:06:08.189263 str-s6100-acs-5 INFO teamd#supervisord: teamsyncd libteam: port_list_init: Failed to get port list.
Jan 18 18:06:08.189263 str-s6100-acs-5 INFO teamd#supervisord: teamsyncd libteam: team_init: Failed to init port list.
Jan 18 18:06:08.189263 str-s6100-acs-5 INFO teamd#supervisord: teamsyncd Exception "Failed to initialize team handler" had been thrown in deamon
Jan 18 18:06:08.194013 str-s6100-acs-5 INFO teamd.sh[1980]: 2019-01-18 18:06:08,192 INFO exited: teamsyncd (exit status 0; expected)
Jan 18 18:06:08.256016 str-s6100-acs-5 INFO swss.sh[2233]: swss
Jan 18 18:06:08.380782 str-s6100-acs-5 INFO teamd#supervisord: teammgrd Using team device "PortChannel63".
Jan 18 18:06:08.387521 str-s6100-acs-5 INFO teamd#supervisord: teammgrd
Jan 18 18:06:08.387521 str-s6100-acs-5 INFO teamd#supervisord: teammgrd Using PID file "/var/run/teamd/PortChannel63.pid"
Jan 18 18:06:08.387521 str-s6100-acs-5 INFO teamd#supervisord: teammgrd
Jan 18 18:06:08.387521 str-s6100-acs-5 INFO teamd#supervisord: teammgrd This program is not intended to be run as root.
Jan 18 18:06:08.387521 str-s6100-acs-5 INFO teamd#supervisord: teammgrd
Jan 18 18:06:08.386166 str-s6100-acs-5 WARNING systemd-udevd[3940]: Could not generate persistent MAC address for PortChannel63: No such file or directory
Jan 18 18:06:08.389354 str-s6100-acs-5 INFO kernel: [   30.778828] PortChannel63: Mode changed to "loadbalance"
Jan 18 18:06:08.409406 str-s6100-acs-5 INFO kernel: [   30.795814] IPv6: ADDRCONF(NETDEV_UP): PortChannel63: link is not ready
Jan 18 18:06:08.418782 str-s6100-acs-5 NOTICE teamd#teammgrd: :- addLag: Start port channel PortChannel63 with teamd
[...]

PR #1144, opened 2019-12-07
teamsyncd::Additional checks put inplace to make sure the netlink message is valid

judyjoseph:
**What I did**

Added additional checks in the netlink message handling of teamsyncd to make sure the message is for a valid interface with ifindex and name. This is using the netlink cache infrastructure to check if the interface name retrieved from cache with the ifindex is matching with the interfaceName in the netlink NEW LINK create message. If the interface name is not same, we drop the message.

**Why I did it**

Have seen random cases where teamsyncd fails and the stack trace was showing interface name as NULL and the ifindex an incorrect one.

**How I verified it**

Could not get the exact reproduction scenario, but induced these errors manually and found the teamsyncd staying up.

**Details if related**

### DIFF
--- a/teamsyncd/teamsync.cpp
+++ b/teamsyncd/teamsync.cpp
@@ -30,17 +30,30 @@ TeamSync::TeamSync(DBConnector *db, DBConnector *stateDb, Select *select) :
     m_nl_sock = nl_socket_alloc();
     if (!m_nl_sock)
     {
-        SWSS_LOG_ERROR("Unable to allocate netlink socket");
-    } else {
-        nl_connect(m_nl_sock, NETLINK_ROUTE);
-        int err = rtnl_link_alloc_cache(m_nl_sock, AF_UNSPEC, &m_link_cache);
-        if (err < 0)
-        {
-            SWSS_LOG_ERROR("Unable to allocate link cache: %s", nl_geterror(err));
-            nl_close(m_nl_sock);
-            nl_socket_free(m_nl_sock);
-            m_nl_sock = NULL;
-        }
+        SWSS_LOG_ERROR("Unable to allocated netlink socket");
+        throw system_error(make_error_code(errc::address_not_available),
+                           "Unable to allocated netlink socket");
+    }
+
+    int err = nl_connect(m_nl_sock, NETLINK_ROUTE);
+    if (err < 0)
+    {
+        SWSS_LOG_ERROR("Unable to connect netlink socket: %s", nl_geterror(err));
+        nl_socket_free(m_nl_sock);
+        m_nl_sock = NULL;
+        throw system_error(make_error_code(errc::address_not_available),
+                           "Unable to connect netlink socket");
+    }
+
+    err = rtnl_link_alloc_cache(m_nl_sock, AF_UNSPEC, &m_link_cache);
+    if (err < 0)
+    {
+        SWSS_LOG_ERROR("Unable to allocate link cache: %s", nl_geterror(err));
+        nl_close(m_nl_sock);
+        nl_socket_free(m_nl_sock);
+        m_nl_sock = NULL;
+        throw system_error(make_error_code(errc::address_not_available),
+                           "Unable to connect netlink socket");
     }
 
     WarmStart::initialize(TEAMSYNCD_APP_NAME, "teamd");
@@ -64,25 +77,29 @@ TeamSync::TeamSync(DBConnector *db, DBConnector *stateDb, Select *select) :
  * Return ifindex in string format if interface don't exist, or no cache
  *        else return the corresponding interface name.
  */
-string TeamSync::ifindexToName(int ifindex)
+bool TeamSync::checkIfindexToName(int ifindex, const string &name)
 {
     char ifName[TeamPortSync::MAX_IFNAME + 1] = {0};
 
-    /* nl_socket, cache create failed earlier */
-    if (!m_nl_sock) {
-        /* Returns ifindex as string / */
-        return to_string(ifindex);
-    }
- 
-    /* Refill cache and check for ifindex */
-    nl_cache_refill(m_nl_sock ,m_link_cache);
-    if (rtnl_link_i2name(m_link_cache, ifindex, ifName, TeamPortSync::MAX_IFNAME) == NULL)
+    /* Refill the cache */
+    int err = nl_cache_refill(m_nl_sock ,m_link_cache);
+    if(err < 0) 
     {
-        /* Returns ifindex as string / */
-        return to_string(ifindex);
+       SWSS_LOG_WARN("Lag interface %s nl_cache_refill failure", name.c_str());
+       return true;
     }
 
-    return string(ifName);
+    /* If interface present in cache, check if it is same as input interface name */
+    if (rtnl_link_i2name(m_link_cache, ifindex, ifName, TeamPortSync::MAX_IFNAME) != NULL) 
+    {
+        if(string(ifName) == name) 
+        {
+            SWSS_LOG_INFO("Lag interface %s exists in cache", ifName);
+            return true;
+        }
+    }
+ 
+    return false;
 }
 
 void TeamSync::periodic()
@@ -144,8 +161,15 @@ void TeamSync::onMsg(int nlmsg_type, struct nl_object *obj)
     if ((nlmsg_type != RTM_NEWLINK) && (nlmsg_type != RTM_DELLINK))
         return;
 
-    string lagName = rtnl_link_get_name(link);
+    /* Introduce check if the interface name is NULL */
+    char *ifName = rtnl_link_get_name(link);
+    if(ifName == NULL) 
+    {
+        SWSS_LOG_WARN("The LAG interface name in Netlink message is empty");
+        return;
+    } 
 
+    string lagName = ifName;
     unsigned int flags = rtnl_link_get_flags(link);
     bool admin = flags & IFF_UP;
     bool oper = flags & IFF_LOWER_UP;
@@ -157,12 +181,6 @@ void TeamSync::onMsg(int nlmsg_type, struct nl_object *obj)
 
     /* Get the interface index */
     unsigned int ifindex = rtnl_link_get_ifindex(link);
-
-    /* Check if the interface name is valid */
-    if(lagName.empty()) {
-        SWSS_LOG_WARN("Lag intf name in Netlink message is empty");
-        return;
-    }
 
     if (type)
     {
@@ -189,10 +207,8 @@ void TeamSync::onMsg(int nlmsg_type, struct nl_object *obj)
      * We are here because this is a NEW link create netlink message from kernel. 
      * Fetch and compare the interface name using ifindex from kernel cache. 
      */
-    string cacheName = ifindexToName(ifindex);
-    if(cacheName.compare(0, cacheName.size(), lagName) != 0) {
-        SWSS_LOG_WARN("Lag intf name ( %s ) doesn't match with name got from kernel( %s )",
-                       lagName.c_str(), cacheName.c_str());
+    if(!checkIfindexToName(ifindex, lagName)) {
+        SWSS_LOG_WARN(" Lag interface ( %s ) don't exist in cache", lagName.c_str());
         return;
     }
 
--- a/teamsyncd/teamsync.cpp
+++ b/teamsyncd/teamsync.cpp
@@ -27,6 +27,22 @@ TeamSync::TeamSync(DBConnector *db, DBConnector *stateDb, Select *select) :
     m_lagMemberTable(db, APP_LAG_MEMBER_TABLE_NAME),
     m_stateLagTable(stateDb, STATE_LAG_TABLE_NAME)
 {
+    m_nl_sock = nl_socket_alloc();
+    if (!m_nl_sock)
+    {
+        SWSS_LOG_ERROR("Unable to allocate netlink socket");
+    } else {
+        nl_connect(m_nl_sock, NETLINK_ROUTE);
+        int err = rtnl_link_alloc_cache(m_nl_sock, AF_UNSPEC, &m_link_cache);
+        if (err < 0)
+        {
+            SWSS_LOG_ERROR("Unable to allocate link cache: %s", nl_geterror(err));
+            nl_close(m_nl_sock);
+            nl_socket_free(m_nl_sock);
+            m_nl_sock = NULL;
+        }
+    }
+
     WarmStart::initialize(TEAMSYNCD_APP_NAME, "teamd");
     WarmStart::checkWarmStart(TEAMSYNCD_APP_NAME, "teamd");
     m_warmstart = WarmStart::isWarmStart();
@@ -41,6 +57,32 @@ TeamSync::TeamSync(DBConnector *db, DBConnector *stateDb, Select *select) :
         WarmStart::setWarmStartState(TEAMSYNCD_APP_NAME, WarmStart::INITIALIZED);
         SWSS_LOG_NOTICE("Starting in warmstart mode");
     }
+}
+
+/*
+ * Get interface name based on interface index
+ * Return ifindex in string format if interface don't exist, or no cache
+ *        else return the corresponding interface name.
+ */
+string TeamSync::ifindexToName(int ifindex)
+{
+    char ifName[TeamPortSync::MAX_IFNAME + 1] = {0};
+
+    /* nl_socket, cache create failed earlier */
+    if (!m_nl_sock) {
+        /* Returns ifindex as string / */
+        return to_string(ifindex);
+    }
+ 
+    /* Refill cache and check for ifindex */
+    nl_cache_refill(m_nl_sock ,m_link_cache);
+    if (rtnl_link_i2name(m_link_cache, ifindex, ifName, TeamPortSync::MAX_IFNAME) == NULL)
+    {
+        /* Returns ifindex as string / */
+        return to_string(ifindex);
+    }
+
+    return string(ifName);
 }
 
 void TeamSync::periodic()
@@ -104,10 +146,34 @@ void TeamSync::onMsg(int nlmsg_type, struct nl_object *obj)
 
     string lagName = rtnl_link_get_name(link);
 
+    unsigned int flags = rtnl_link_get_flags(link);
+    bool admin = flags & IFF_UP;
+    bool oper = flags & IFF_LOWER_UP;
+
     /* Listens to LAG messages */
     char *type = rtnl_link_get_type(link);
     if (!type || (strcmp(type, TEAM_DRV_NAME) != 0))
         return;
+
+    /* Get the interface index */
+    unsigned int ifindex = rtnl_link_get_ifindex(link);
+
+    /* Check if the interface name is valid */
+    if(lagName.empty()) {
+        SWSS_LOG_WARN("Lag intf name in Netlink message is empty");
+        return;
+    }
+
+    if (type)
+    {
+        SWSS_LOG_INFO("nlmsg type:%d key:%s admin:%d oper:%d ifindex:%d type:%s",
+                       nlmsg_type, lagName.c_str(), admin, oper, ifindex, type);
+    }
+    else
+    {
+        SWSS_LOG_INFO("nlmsg type:%d key:%s admin:%d oper:%d ifindex:%d",
+                       nlmsg_type, lagName.c_str(), admin, oper, ifindex);
+    }
 
     if (nlmsg_type == RTM_DELLINK)
     {
@@ -119,10 +185,19 @@ void TeamSync::onMsg(int nlmsg_type, struct nl_object *obj)
         return;
     }
 
+    /* 
+     * We are here because this is a NEW link create netlink message from kernel. 
+     * Fetch and compare the interface name using ifindex from kernel cache. 
+     */
+    string cacheName = ifindexToName(ifindex);
+    if(cacheName.compare(0, cacheName.size(), lagName) != 0) {
+        SWSS_LOG_WARN("Lag intf name ( %s ) doesn't match with name got from kernel( %s )",
+                       lagName.c_str(), cacheName.c_str());
+        return;
+    }
+
     unsigned int mtu = rtnl_link_get_mtu(link);
-    addLag(lagName, rtnl_link_get_ifindex(link),
-           rtnl_link_get_flags(link) & IFF_UP,
-           rtnl_link_get_flags(link) & IFF_LOWER_UP, mtu);
+    addLag(lagName, ifindex, admin, oper, mtu);
 }
 
 void TeamSync::addLag(const string &lagName, int ifindex, bool admin_state,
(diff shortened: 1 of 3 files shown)


madhukar-kamarapu:
Hi judyjoseph,

Could you please share the teamsyncd related logs when the issue is seen (OR when you introduced the issue manually).

{noformat}
root@sonic:/var/log# pwd
/var/log
root@sonic:/var/log# zgrep -ne teamsyncd *
{noformat}

By any chance are the logs similar to the one's reported in https://github.com/Azure/sonic-swss/issues/763 ?

Rgds,
Madhukar

teamsyncd/teamsync.cpp
```diff
@@ -27,6 +27,22 @@ TeamSync::TeamSync(DBConnector *db, DBConnector *stateDb, Select *select) :
     m_lagMemberTable(db, APP_LAG_MEMBER_TABLE_NAME),
     m_stateLagTable(stateDb, STATE_LAG_TABLE_NAME)
 {
+    m_nl_sock = nl_socket_alloc();
+    if (!m_nl_sock)
+    {
+        SWSS_LOG_ERROR("Unable to allocate netlink socket");
+    } else {
+        nl_connect(m_nl_sock, NETLINK_ROUTE);
```
pavel-shirshov: check error code?

teamsyncd/teamsync.cpp
```diff
@@ -27,6 +27,22 @@ TeamSync::TeamSync(DBConnector *db, DBConnector *stateDb, Select *select) :
     m_lagMemberTable(db, APP_LAG_MEMBER_TABLE_NAME),
     m_stateLagTable(stateDb, STATE_LAG_TABLE_NAME)
 {
+    m_nl_sock = nl_socket_alloc();
+    if (!m_nl_sock)
+    {
+        SWSS_LOG_ERROR("Unable to allocate netlink socket");
```
pavel-shirshov: I think it's better to crash it. Otherwise your check will not work

teamsyncd/teamsync.cpp
```diff
@@ -43,6 +59,32 @@ TeamSync::TeamSync(DBConnector *db, DBConnector *stateDb, Select *select) :
...
+    char ifName[TeamPortSync::MAX_IFNAME + 1] = {0};
+
+    /* nl_socket, cache create failed earlier */
+    if (!m_nl_sock) {
+        /* Returns ifindex as string / */
+        return to_string(ifindex);
+    }
+ 
+    /* Refill cache and check for ifindex */
+    nl_cache_refill(m_nl_sock ,m_link_cache);
```
pavel-shirshov: could return not zero as error indication
Also. Would it be to much to refill all cache for every ifindex to name operation?

teamsyncd/teamsync.cpp
```diff
@@ -43,6 +59,32 @@ TeamSync::TeamSync(DBConnector *db, DBConnector *stateDb, Select *select) :
...
+ *        else return the corresponding interface name.
+ */
+string TeamSync::ifindexToName(int ifindex)
+{
+    char ifName[TeamPortSync::MAX_IFNAME + 1] = {0};
+
+    /* nl_socket, cache create failed earlier */
+    if (!m_nl_sock) {
+        /* Returns ifindex as string / */
+        return to_string(ifindex);
```
pavel-shirshov: this is not an interface name

teamsyncd/teamsync.cpp
```diff
@@ -43,6 +59,32 @@ TeamSync::TeamSync(DBConnector *db, DBConnector *stateDb, Select *select) :
...
+    }
+ 
+    /* Refill cache and check for ifindex */
+    nl_cache_refill(m_nl_sock ,m_link_cache);
+    if (rtnl_link_i2name(m_link_cache, ifindex, ifName, TeamPortSync::MAX_IFNAME) == NULL)
+    {
+        /* Returns ifindex as string / */
+        return to_string(ifindex);
```
pavel-shirshov: the same as before

pavel-shirshov:
> Could not get the exact reproduction scenario, but induced these errors manually and found the teamsyncd staying up.
I think it's impossible to verify anything until we can reproduce the issue.
Have you check this change https://github.com/Azure/sonic-swss/commit/b9317511bac3de0f55fa1a7ecdab5742d6e4eabb ?

judyjoseph:
> Hi judyjoseph,
> Could you please share the teamsyncd related logs when the issue is seen (OR when you introduced the issue manually).
> root@sonic:/var/log# pwd
> /var/log
> root@sonic:/var/log# zgrep -ne teamsyncd *
> 
> 
> By any chance are the logs similar to the one's reported in #763 ?
> Rgds,
> Madhukar

Hi Madhukar,
    This fix is to try address a segmentation fault crash in teamsyncd  which was seen when a config reload command is issued and not for the logs you pointed out.

regards,
Judy
image

judyjoseph:
> Could not get the exact reproduction scenario, but induced these errors manually and found the teamsyncd staying up.
> I think it's impossible to verify anything until we can reproduce the issue.
> Have you check this change b931751 ?

Hi Pavel,
    I am not able to reproduce this segmentation fault in teamsyncd with multiple back to back config reloads. This issue occurred earlier after a config reload.

    I have tried the patch b931751 -- but the error message "TeamPortSync: Failed to initialize team handler" comes again randomly on back to back config reloads and now it happens thrice.

thanks,
Judy

teamsyncd/teamsync.cpp
judyjoseph: Sure will add an explicit check.

teamsyncd/teamsync.cpp
judyjoseph: I had thought before about introducing a throw statement and crashing if the socket_alloc() fails. But didn't do so as I have seen these failure messages of "Unable to initialize team socket" with the team_alloc()/team_init() in other scenario's like back to back config reload command issued multiple times. So didn't want teamsyncd to fail when we try to spawn the process because if this additional check we are introducing. 

But I think on a second thought -- I agree with you, it is good to throw error and crash as this is not a good scenario. Shall update, thanks.

teamsyncd/teamsync.cpp
judyjoseph: I will do a bit of profiling and see how much time it takes will all the max number of Port channels created. We need to do a refill before checking -- as I found cache incosistency issues in case when we do config reload command back to back. 

Also I see similar implementations in 
RouteSync::onRouteMsg() --> calling RouteSync::getIfName()
NeighSync::onMsg()  -->  LinkCache::getInstance().ifindexToName()

teamsyncd/teamsync.cpp
judyjoseph: The idea was to give back the ifindex itself as the interface name and hence the caller function match will fail and ignore the NETLINK update, instead of proceeding with the TeamSync::addLag() API.

teamsyncd/teamsync.cpp
judyjoseph: I did more profiling for the refill cache, and found it was done in the same second ( it took around 1ms for refill API to complete execution ) - which looked ok.

teamsyncd/teamsync.cpp
judyjoseph: I changed the API format so that now it checks the interface name in the API checkIfindexToName() and returns true/false depending on the interface name is present in kernel cache or not.

madhukar-kamarapu:
> > Could not get the exact reproduction scenario, but induced these errors manually and found the teamsyncd staying up.
> > I think it's impossible to verify anything until we can reproduce the issue.
> > Have you check this change [b931751](https://github.com/Azure/sonic-swss/commit/b9317511bac3de0f55fa1a7ecdab5742d6e4eabb) ?
> 
> Hi Pavel,
> I am not able to reproduce this segmentation fault in teamsyncd with multiple back to back config reloads. This issue occurred earlier after a config reload.
> 
> ```
> I have tried the patch b931751 -- but the error message "TeamPortSync: Failed to initialize team handler" comes again randomly on back to back config reloads and now it happens thrice.
> ```
> 
> thanks,
> Judy

The reason for the error "TeamPortSync: Failed to initialize team handler" during CONFIG-RELOAD is having the old port-channel netdevices in the kernel. 
Fix: Delete the old port-channel netdevices from the kernel before teamsyncd is started. You may add the logic in TeamSync::TeamSync().
Quick fix to validate: Delete all the port-channel netdevices using the command "ip link del dev PortChannelXXX" before teamd docker starts.

I'll be submitting this PR later with more details; you may use the above fix in the meanwhile.
This fix should also solve your teamsyncd crash.
Note: In sonic-buildimage/dockers/docker-teamd/start.sh, start teamsyncd 1st followed with teammgrd. This helps in addressing few issues.

madhukar-kamarapu:
> > > Could not get the exact reproduction scenario, but induced these errors manually and found the teamsyncd staying up.
> > > I think it's impossible to verify anything until we can reproduce the issue.
> > > Have you check this change [b931751](https://github.com/Azure/sonic-swss/commit/b9317511bac3de0f55fa1a7ecdab5742d6e4eabb) ?
> > 
> > 
> > Hi Pavel,
> > I am not able to reproduce this segmentation fault in teamsyncd with multiple back to back config reloads. This issue occurred earlier after a config reload.
> > ```
> > I have tried the patch b931751 -- but the error message "TeamPortSync: Failed to initialize team handler" comes again randomly on back to back config reloads and now it happens thrice.
> > ```
> > 
> > 
> > thanks,
> > Judy
> 
> The reason for the error "TeamPortSync: Failed to initialize team handler" during CONFIG-RELOAD is having the old port-channel netdevices in the kernel.
> Fix: Delete the old port-channel netdevices from the kernel before teamsyncd is started. You may add the logic in TeamSync::TeamSync().
> Quick fix to validate: Delete all the port-channel netdevices using the command "ip link del dev PortChannelXXX" before teamd docker starts.
> 
> I'll be submitting this PR later with more details; you may use the above fix in the meanwhile.
> This fix should also solve your teamsyncd crash.
> Note: In sonic-buildimage/dockers/docker-teamd/start.sh, start teamsyncd 1st followed with teammgrd. This helps in addressing few issues.

@judyjoseph - can you please let me know if the suggested solution works for you?

teamsyncd/teamsync.cpp
```diff
@@ -102,13 +161,38 @@ void TeamSync::onMsg(int nlmsg_type, struct nl_object *obj)
     if ((nlmsg_type != RTM_NEWLINK) && (nlmsg_type != RTM_DELLINK))
         return;
 
-    string lagName = rtnl_link_get_name(link);
+    /* Introduce check if the interface name is NULL */
+    char *ifName = rtnl_link_get_name(link);
```
lguohan: I do not understand why you change the previous lagName to ifName here. seems ifName is not used anymore afterwards

teamsyncd/teamsync.cpp
```diff
@@ -119,10 +203,17 @@ void TeamSync::onMsg(int nlmsg_type, struct nl_object *obj)
         return;
     }
 
+    /* 
+     * We are here because this is a NEW link create netlink message from kernel. 
+     * Fetch and compare the interface name using ifindex from kernel cache. 
+     */
+    if(!checkIfindexToName(ifindex, lagName)) {
```
lguohan: under what conditions do we see ifindex does not match lagName? if we ignore such messages, what is the consequence here? Are we losing some validate netlink messages?

lguohan:
@madhukar-kamarapu , teammgrd will recreate those port channel devices. is there a race condition between teamsync and teammgrd? maybe we should put some explicit dependency between teamsync and teammgrd?

teamsyncd/teamsync.cpp
judyjoseph: The main reason behind this fix was the observation of a crash in teamsyncd segmentation fault with invalid ifindex and mtu value 

#7  swss::TeamSync::addLag (this=0x3e8, lagName=..., ifindex=-374612960, admin_state=<optimized out>, oper_state=<optimized out>, mtu=3920352592)

So the plan here was to check if the ifName is null … to make sure we don't proceed.

teamsyncd/teamsync.cpp
judyjoseph: From my testing I have seen messages with the older if indices coming to teamsyncd on doing a config reload. So the idea was to make sure we proceed to add Lag only the ifindices --> interface name and the "Name" which came in netlink message is same. This check should not resulting in losing valid messages.

madhukar-kamarapu:
> @madhukar-kamarapu , teammgrd will recreate those port channel devices. is there a race condition between teamsync and teammgrd? maybe we should put some explicit dependency between teamsync and teammgrd?

@judyjoseph - if teammgrd is launched before teamsyncd, I've observed NLE_DUMP_INTR error on the netlink socket opened by teamsyncd. Reason - during config-reload OR cold-reboot, all the netdevices (front-panel-ports, port-channel, VLAN, VTEP, etc) are being created in the kernel. Due to this, the DUMP_REQUEST sent by teamsyncd fails. Note: This is purely a timing issue though.

Upon receiving NLE_DUMP_INTR, we are supposed to retry the DUMP_REQUEST (already done on teamd - https://lists.fedoraproject.org/archives/list/libteam@lists.fedorahosted.org/thread/SMLV7U2FCYL7UUAYVDBGQMHHSI3YWEK7/). 
Currently teamsyncd does not have this infrastructure to retry for NLE_DUMP_INTR.

As long as we delete the stale (old) port-channel netdevices in the kernel during config-reload, I don't see any problem by starting teamsyncd 1st followed with teammgrd.

We can add dependency though - program an entry TeamsyncdInitDone in APP_DB (similar to PortInitDone in PORT_TABLE of APP_DB).

madhukar-kamarapu:
> > Could not get the exact reproduction scenario, but induced these errors manually and found the teamsyncd staying up.
> > I think it's impossible to verify anything until we can reproduce the issue.
> > Have you check this change [b931751](https://github.com/Azure/sonic-swss/commit/b9317511bac3de0f55fa1a7ecdab5742d6e4eabb) ?
> > Hi Pavel,
> > I am not able to reproduce this segmentation fault in teamsyncd with multiple back to back config reloads. This issue occurred earlier after a config reload.
> > I have tried the patch [b931751](https://github.com/Azure/sonic-swss/commit/b9317511bac3de0f55fa1a7ecdab5742d6e4eabb) -- but the error message "TeamPortSync: Failed to initialize team handler" comes again randomly on back to back config reloads and now it happens thrice.
> > thanks,
> > Judy
> > The reason for the error "TeamPortSync: Failed to initialize team handler" during CONFIG-RELOAD is having the old port-channel netdevices in the kernel.
> > Fix: Delete the old port-channel netdevices from the kernel before teamsyncd is started. You may add the logic in TeamSync::TeamSync().
> > Quick fix to validate: Delete all the port-channel netdevices using the command "ip link del dev PortChannelXXX" before teamd docker starts.
> > I'll be submitting this PR later with more details; you may use the above fix in the meanwhile.
> > This fix should also solve your teamsyncd crash.
> > Note: In sonic-buildimage/dockers/docker-teamd/start.sh, start teamsyncd 1st followed with teammgrd. This helps in addressing few issues.
> > @judyjoseph - can you please let me know if the suggested solution works for you?
> 
> @madhukar-kamarapu, Thank you for your comments. From what I understand from code and from testing scenario's - the issues I see of invalid netlink messages is when we do commands like config reload/ load minigraph where we stop start the services. I agree with you that one approach could be to delete all the port-channels in the constructor, but then for this fix to work, we will have to change process startup order to start teamsyncd first and then teammgrd which could have other impacts ( to be analyzed )
> 
> The fix I am working on now is to do a good clean of the interfaces when the teamd docker goes down, like clean up the portchannel interfaces in teammgrd signal handler. Additionally I am looking into checking if we don't wait enough on doing a docker stop. Are we spawning the teamd docker again too quickly ? I find this as the root-cause of invalid netlink messages coming in teamSync::onMsg() -- your thoughts.

@judyjoseph - Adding logic to clean-up the port-channel interfaces in teammgrd signal handlers seems fine. One another thought - how about terminating all the teamd processes; it should take care of removing the port-channel netdevices in the kernel. 
@lguohan  - can you please comment on this.

Spawning teamd docker quickly is fine - all we need to ensure that proper clean-up is done before we bring up the teamd docker. 
As I mentioned earlier, invalid netlink messages is mainly due to the stale (old) port-channel netdevices in the kernel. You may print/dump the interface number when you receive the invalid netlink message, it clearly shows that its the port-channel netdevice created before the config-reload was issued.

judyjoseph:
> Can you please update the questionnaire:
> 
> What I did
> Why I did it
> How I verified it

Thanks for taking a look. I too find difficult to update this PR .. hence I had opened a new one #1159. Can you take a look ? The details are updated already there.

judyjoseph:
> @judyjoseph  - Thanks for taking care of this. Did you get a chance to confirm the PortChannelXXXX netdevices were deleted in the kernel during config-reload/docker-restart/docker-upgrade?

Yes I have verified that the teamd processes exits cleanly and the Port channel interfaces are removed on config reload/ docker restart - I would need to try a docker upgrade though

judyjoseph:
> Could not get the exact reproduction scenario, but induced these errors manually and found the teamsyncd staying up.
> I think it's impossible to verify anything until we can reproduce the issue.
> Have you check this change b931751 ?
> Hi Pavel,
> I am not able to reproduce this segmentation fault in teamsyncd with multiple back to back config reloads. This issue occurred earlier after a config reload.
> I have tried the patch b931751 -- but the error message "TeamPortSync: Failed to initialize team handler" comes again randomly on back to back config reloads and now it happens thrice.
> thanks,
> Judy
> The reason for the error "TeamPortSync: Failed to initialize team handler" during CONFIG-RELOAD is having the old port-channel netdevices in the kernel.
> Fix: Delete the old port-channel netdevices from the kernel before teamsyncd is started. You may add the logic in TeamSync::TeamSync().
> Quick fix to validate: Delete all the port-channel netdevices using the command "ip link del dev PortChannelXXX" before teamd docker starts.
> I'll be submitting this PR later with more details; you may use the above fix in the meanwhile.
> This fix should also solve your teamsyncd crash.
> Note: In sonic-buildimage/dockers/docker-teamd/start.sh, start teamsyncd 1st followed with teammgrd. This helps in addressing few issues.
> @judyjoseph - can you please let me know if the suggested solution works for you?
> 
> @madhukar-kamarapu, Thank you for your comments. From what I understand from code and from testing scenario's - the issues I see of invalid netlink messages is when we do commands like config reload/ load minigraph where we stop start the services. I agree with you that one approach could be to delete all the port-channels in the constructor, but then for this fix to work, we will have to change process startup order to start teamsyncd first and then teammgrd which could have other impacts ( to be analyzed )
> The fix I am working on now is to do a good clean of the interfaces when the teamd docker goes down, like clean up the portchannel interfaces in teammgrd signal handler. Additionally I am looking into checking if we don't wait enough on doing a docker stop. Are we spawning the teamd docker again too quickly ? I find this as the root-cause of invalid netlink messages coming in teamSync::onMsg() -- your thoughts.
> 
> @judyjoseph - Adding logic to clean-up the port-channel interfaces in teammgrd signal handlers seems fine. One another thought - how about terminating all the teamd processes; it should take care of removing the port-channel netdevices in the kernel.
> @lguohan  - can you please comment on this.
> Spawning teamd docker quickly is fine - all we need to ensure that proper clean-up is done before we bring up the teamd docker.
> As I mentioned earlier, invalid netlink messages is mainly due to the stale (old) port-channel netdevices in the kernel. You may print/dump the interface number when you receive the invalid netlink message, it clearly shows that its the port-channel netdevice created before the config-reload was issued.

I did some study on this and found that the teamd processes were not getting the SIGTERM signal from supervisord on issuing the "docker stop". Can you take a look at the PR #1159. This change gets things cleaned and has least impact.

madhukar-kamarapu:
> > @judyjoseph  - Thanks for taking care of this. Did you get a chance to confirm the PortChannelXXXX netdevices were deleted in the kernel during config-reload/docker-restart/docker-upgrade?
> 
> Yes I have verified that the teamd processes exits cleanly and the Port channel interfaces are removed on config reload/ docker restart - I would need to try a docker upgrade though

Thanks. Please update the docker upgrade results in the new PR #1159.
Note: I am sure it would work (since docker-restart worked); but wanted to double confirm.

madhukar-kamarapu:
> > Could not get the exact reproduction scenario, but induced these errors manually and found the teamsyncd staying up.
> > I think it's impossible to verify anything until we can reproduce the issue.
> > Have you check this change [b931751](https://github.com/Azure/sonic-swss/commit/b9317511bac3de0f55fa1a7ecdab5742d6e4eabb) ?
> > Hi Pavel,
> > I am not able to reproduce this segmentation fault in teamsyncd with multiple back to back config reloads. This issue occurred earlier after a config reload.
> > I have tried the patch [b931751](https://github.com/Azure/sonic-swss/commit/b9317511bac3de0f55fa1a7ecdab5742d6e4eabb) -- but the error message "TeamPortSync: Failed to initialize team handler" comes again randomly on back to back config reloads and now it happens thrice.
> > thanks,
> > Judy
> > The reason for the error "TeamPortSync: Failed to initialize team handler" during CONFIG-RELOAD is having the old port-channel netdevices in the kernel.
> > Fix: Delete the old port-channel netdevices from the kernel before teamsyncd is started. You may add the logic in TeamSync::TeamSync().
> > Quick fix to validate: Delete all the port-channel netdevices using the command "ip link del dev PortChannelXXX" before teamd docker starts.
> > I'll be submitting this PR later with more details; you may use the above fix in the meanwhile.
> > This fix should also solve your teamsyncd crash.
> > Note: In sonic-buildimage/dockers/docker-teamd/start.sh, start teamsyncd 1st followed with teammgrd. This helps in addressing few issues.
> > @judyjoseph - can you please let me know if the suggested solution works for you?
> > @madhukar-kamarapu, Thank you for your comments. From what I understand from code and from testing scenario's - the issues I see of invalid netlink messages is when we do commands like config reload/ load minigraph where we stop start the services. I agree with you that one approach could be to delete all the port-channels in the constructor, but then for this fix to work, we will have to change process startup order to start teamsyncd first and then teammgrd which could have other impacts ( to be analyzed )
> > The fix I am working on now is to do a good clean of the interfaces when the teamd docker goes down, like clean up the portchannel interfaces in teammgrd signal handler. Additionally I am looking into checking if we don't wait enough on doing a docker stop. Are we spawning the teamd docker again too quickly ? I find this as the root-cause of invalid netlink messages coming in teamSync::onMsg() -- your thoughts.
> > @judyjoseph - Adding logic to clean-up the port-channel interfaces in teammgrd signal handlers seems fine. One another thought - how about terminating all the teamd processes; it should take care of removing the port-channel netdevices in the kernel.
> > @lguohan  - can you please comment on this.
> > Spawning teamd docker quickly is fine - all we need to ensure that proper clean-up is done before we bring up the teamd docker.
> > As I mentioned earlier, invalid netlink messages is mainly due to the stale (old) port-channel netdevices in the kernel. You may print/dump the interface number when you receive the invalid netlink message, it clearly shows that its the port-channel netdevice created before the config-reload was issued.
> 
> I did some study on this and found that the teamd processes were not getting the SIGTERM signal from supervisord on issuing the "docker stop". Can you take a look at the PR #1159. This change gets things cleaned and has least impact.

@judyjoseph - I guess you're going to abandon the current PR (#1144). Please confirm.
I will give further comments (if any) in the PR#1159

PR #1147, opened 2019-12-13
DPB: Infra code to track dependencies on Port object

vasant17:
**What I did**
Added infrastructure code to track objects dependent on Port. 

**Why I did it**

To support Dynamic PortBreakout feature, we need to support deletion of Port. But before deleting a port, we need to cleanup the objects dependent on Port. For example, VLAN members, ACL polrt list, interfaces, etc.

**How I verified it**

This code has been used with ACL and VLAN dependency handling code for DPB feature. PRs for those will raised soon. I have verified for both those PRs that existing VS test cases and feature related test cases pass.

**Details if related**

Before deleting a port, we will check if any module/feature/object is depending on port. If so, we will skip the port and pict the next one in the queue. That is until the dependency are removed, we keep looping the doPortTask. Once all dependencies are removed, we will delete the port.

### DIFF
--- a/orchagent/port.h
+++ b/orchagent/port.h
@@ -48,6 +48,24 @@ public:
         UNKNOWN
     } ;
 
+    enum Dependency {
+        ACL_DEP,
+        FDB_DEP,
+        INTF_DEP,
+        LAG_DEP,
+        VLAN_DEP,
+        MAX_DEP
+    };
+
+    std::unordered_map<Dependency, std::string>
+    m_dependency_string = {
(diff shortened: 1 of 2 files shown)


vasant17:
retest

zhenggen-xu:
retest it please!

zhenggen-xu:
retest this please!

lguohan:
do we need this?

lguohan:
discussed with @zhenggen-xu , close it.

PR #1112, opened 2019-10-30
[DPB portsyncd/portmgrd/portorch] Support dynamic port add/deletion without dependencies

zhenggen-xu:
**What I did**
Support port add/delete cases in different daemons without dependencies

**Why I did it**
Baseline code for dynamic port breakout code.   

**How I verified it**
vs test cases

**Details if related**

Support port add/delete cases in different daemons
    
Changes were made on portmgrd/portsyncd and orchagent portsorch
so it should be able to remove/add ports in case no configuration
dependencies or runtime depencies (neighbor, mac etc) on them
    
Also skipped the netlink for port add/delete with master in portsyncd
and cleaned up g_init and g_portSet flag and data strcutures usage.
    
Added dynamic portbeakout test cases including the conf_test.py changs

Signed-off-by: Zhenggen Xu <zxu@linkedin.com>
Signed-off-by: Vasant Patil <vapatil@linkedin.com>

PR #1150, opened 2019-12-13
DPB: Handle VLAN dependency

vasant17:
Note: This PR is dependent on https://github.com/Azure/sonic-swss/pull/1112 and https://github.com/Azure/sonic-swss/pull/1147 

* VLAN dependency handling for Dynamic Port Breakout:
    Set port::VLAN_DEP bit when port is added to first VLAN
    Clear port::VLAN_DEP bit when port is removed from all VLANs

**What I did**
Track VLAN dependency on port and set/clear the dependency bit accordingly

**Why I did it**
So, that we can breakout the port when its part of VLAN(s)

**How I verified it**
Wrote VS test cases, and ran existing VS test cases

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2292,9 +2292,14 @@ void PortsOrch::doVlanMemberTask(Consumer &consumer)
             }
 
             if (addBridgePort(port) && addVlanMember(vlan, port, tagging_mode))
+            {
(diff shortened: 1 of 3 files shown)


prsunny:
Is this PR relevant as #1147 is closed?

vasant17:
This is still relevant, but as discussed with Microsoft team, I will update the way we check dependency

vasant17:
Obsolete. As changes have been captured in other PR.

PR #1153, opened 2019-12-17
Move away sairedis logrotate from signal handler

kcudnik:
**What I did**
Move away logrotate from signal handler to main orch loop, to not execute any sairedis library code which potentially can be under mutex and cause deadlock when signal handler occur

**Why I did it**
to not execute sai_switch_api->set_attribute() under signal handler, just to set logrotate flag

**How I verified it**
running orchagent against vs switch and sending HUP signal to orch process

**Details if related**
Since flush method is executed every time that select is returning object from orchdaemon main/start loop, this is good point to check if logrotate is needed

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -47,6 +47,7 @@ int gBatchSize = DEFAULT_BATCH_SIZE;
 bool gSairedisRecord = true;
 bool gSwssRecord = true;
 bool gLogRotate = false;
+bool gSaiRedisLogRotate = false;
 bool gSyncMode = false;
 
 ofstream gRecordOfs;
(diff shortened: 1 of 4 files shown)


kcudnik:
retest this please

orchagent/main.cpp
```diff
@@ -47,6 +47,7 @@ int gBatchSize = DEFAULT_BATCH_SIZE;
 bool gSairedisRecord = true;
 bool gSwssRecord = true;
 bool gLogRotate = false;
+bool gSaiRedisLogRotate = false;
```
qiluo-msft: Add `volatile` ?

orchagent/main.cpp
```diff
@@ -47,6 +47,7 @@ int gBatchSize = DEFAULT_BATCH_SIZE;
 bool gSairedisRecord = true;
 bool gSwssRecord = true;
 bool gLogRotate = false;
```
qiluo-msft: >gLogRotate [](start = 5, length = 10)

The name seems cover gSaiRedisLogRotate. Do you mean `gOrchagentLogRotate` ?

kcudnik: no, this logrotate only covers sairedis library

orchagent/main.cpp
kcudnik: not needed

kcudnik:
not sure why tests fails on VS, seems not related to my change

kcudnik:
retest this please

kcudnik:
test this please

PR #1157, opened 2019-12-18
[201811][port/buffer] introduce a sync mechanism to protect port PG/queue from changes under PFC stormPfcwd bufferorch sync 201811

stepanblyschak:
**What I did**
I introduce a sync mechanism to protect port PG/queue from changes under PFC storm.

**Why I did it**

Under warm reboot conditions on Mellanox platform, buffermgrd starts after BufferOrch gets initialized and sets initial state to DB; after reconciliation BufferOrch reacts on those changes and sets PG profile overriding PG profile set by PfcWdOrch.

**How I verified it**
Run PFC WD + warm reboot test cases.
Mannual restart of buffermgrd when port is under PFC storm - verified PG profile was not overriten.
After storm, PG profile was set according to config db.

**Details if related**


wendani:
How do you find the issue, @stepanblyschak?

PR #1158, opened 2019-12-20
[restore_neighbors.py] build arp packet with correct hwsrc and psrc

stepanblyschak:
Otherwise, we can see that DUT after warm reboot sends ARP
packets with management IP (10.112.71.4)

$ sudo tcpdump -ni PortChannel0003 --direction out
16:33:40.572995 ARP, Request who-has 10.0.0.61 tell 10.112.71.4, length
28

All neighbors than are stil considered as STALE.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/neighsyncd/restore_neighbors.py
+++ b/neighsyncd/restore_neighbors.py
@@ -194,7 +194,7 @@ def set_neigh_in_kernel(ipclass, family, intf_idx, dst_ip, dmac):
 def build_arp_ns_pkt(family, smac, src_ip, dst_ip):
     if family == 'IPv4':
         eth = Ether(src=smac, dst='ff:ff:ff:ff:ff:ff')
-        pkt = eth/ARP(op='who-has', pdst=dst_ip)
+        pkt = eth/ARP(op='who-has', pdst=dst_ip, psrc=src_ip, hwsrc=smac)
     elif family == 'IPv6':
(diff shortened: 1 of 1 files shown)


yxieca:
@zhenggen-xu looks like since day 1 the smac and srcip were not used for IPv4 ARP packet. Do you remember why? Do you have concern for this change?

zhenggen-xu:
What is the test topology and where is PortChannel0003? 

scapy supposed to use the interfaces sip and smac when the arp was sent out.  It would be helpful to understand why it was not.  To add them explicitly should be fine as done in this PR, but might be good to check why,  just don't want to hide any potential real issue somewhere.

stepanblyschak:
@zhenggen-xu test topology is T0, PortChannel0003 is there if you deploy minigraph for t0 from sonic-mgmt.
I remember some python library cached the ip/mac on import time. It might be the same with scapy.
If script starts before PortChannels are created and configured in kernel it might use mgmt as a default. That's my understanding.

liat-grozovik:
retest vs please

liat-grozovik:
retest this please

prsunny:
retest vs please

daall:
retest this please

daall:
retest this please

prsunny:
retest this please

PR #1159, opened 2019-12-23
Teamd :: fix for cleaning up the teamd processes correctly on teamd docker stop

judyjoseph:
**What I did**

The changes implemented in the final fix are, 
   (1)	In teammgrd: While adding LAG, save the mapping of lag <-> teamd PID
   (2)	In teammgrd: While removing LAG, remove the mapping of lag <-> teamd PID
   (3)	Introduce SIGTERM handlers in both teammgrd and teamsyncd

- (a) In teammgrd, send the SIGTERM to all the teamd processes. It takes just a second for teamd processes + port channel interfaces to be cleaned up
- (b) In teamsyncd, clean up the teamd resources like nl_socket etc.

**Why I did it**

    There were multiple Issues seen, in random on doing a config reload/ config loadminigraph

(i)	Teamsyncd segfault 
(ii)	Teamsyncd getting the NETLINK messages with the older IFINDEXs
    (iii) "TeamPortSync: Unable to init team socket" error messages

  Root cause

     With config reload command , the docker teamd stop and start is issued. I find that after "docker stop" the port channel interfaces remains in the linux kernel – it is not getting cleaned. There are indeed signal handlers in the teamd processes. But in this case teamd is started as a “daemon” with the “-d” option. The supervisord is not able to send the SIGTERM to the processes started in a daemon mode (http://supervisord.org/subprocess.html?highlight=daemon)

“docker stop” sends two signals 

1.	SIGTERM to the ENTRYPOINT process in our case it is supervisord process 
2.	After 10 sec send SIGKILL to all processes to stop the container.

     So when docker stop is issued, supervisord send SIGTERM to all processes ( teamsyncd/temmgrd gets the signal correctly – but I find none of the teamd processes get the signal either because they are daemonized and running in background OR because teamd is not started directly by supervisord). 

The teamd processes are killed later with SIGKILL send as a last resort to stop docker. SIGKILL cannot be handled and hence no cleanup – portchanel interfaces remains in kernel.

**How I verified it**

   Tried the config reload back to back multiple times -- verified that the Portchannel interface are cleaned up in kernel, teamd processes are killed correctly.
  When the teamd docker is started again, I don't see random NETLINK create/delete messages with the older Portchannel ifindices. ( which was seen earlier without this fix ) 

**Other possible fixes which I tried out, but decided not to take**

The other possible fix options which was tried were,

   (1)	Use the fork/exec way to spawn the teamd processes
               

- (a) Seeing instabilities in all teamd instances NOT coming up when we do “config reload” back to back.
               

- (b) Even with teamd staying in foreground, those processes were not getting the  SIGTERM on “docker stop”. So doesn’t serve the purpose of trying NOT to  daemonize the teamd.

   
(2)	In the teammgrd Signal handler directly invoke an existing API teammgrd:removeLag() which
        cleans up the teamd processes. This was working fine, but the cleanup was taking longer 
        duration like around 5sec for cleaning up all the teamd processes, which very high 
        considering it is a signal handler.

### DIFF
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -8,11 +8,16 @@
 
 #include <algorithm>
 #include <sstream>
+#include <fstream>
 #include <thread>
 
 #include <net/if.h>
 #include <sys/ioctl.h>
 #include <sys/stat.h>
+#include <sys/types.h>
+#include <signal.h>
+
+#define PID_FILE_PATH "/var/run/teamd/"
 
 using namespace std;
 using namespace swss;
@@ -108,6 +113,77 @@ void TeamMgr::doTask(Consumer &consumer)
     }
 }
 
+
+#define MAX_PID_STRLEN 10
+int TeamMgr::getTeamPid(const string &alias)
+{
+    SWSS_LOG_ENTER();
+    int pid = 0;
+
+    string file = string(PID_FILE_PATH) + alias + string(".pid");
+    ifstream infile(file);
+    if (!infile.is_open())
+    {
+        SWSS_LOG_WARN("The LAG PID file: %s is not readable", file.c_str());
+        return 0;
+    }
+
+    string line;
+    getline(infile, line);
+    if (line.empty())
+    {
+        SWSS_LOG_WARN("The LAG PID file: %s is empty", file.c_str());
+    }
+    else 
+    {
+        /*Store the PID value */
+        pid = stoi(line, nullptr, 10);
+    }
+
+    /* Close the file and return */
+    infile.close();
+
+    return pid;
+}
+
+
+#define MAX_PID_STRLEN 10
+void TeamMgr::addLagPid(const string &alias)
+{
+    SWSS_LOG_ENTER();
+    m_lagPIDList[alias] = getTeamPid(alias);
+}
+
+void TeamMgr::removeLagPid(const string &alias)
+{
+    SWSS_LOG_ENTER();
+    m_lagPIDList.erase(alias);
+}
+
+void TeamMgr::cleanTeamProcesses(int signo)
+{
+    int pid = 0;
+
+    for (const auto& it: m_lagList)
+    {
+        pid = m_lagPIDList[it];
+        if(!pid) {
+            SWSS_LOG_WARN("Invalid PID found for LaG %s ", it.c_str());
+
+            /* Try to get the PID again */
+            pid = getTeamPid(it);
+        }
+
+        if(pid > 0)
+        {
+            SWSS_LOG_WARN("Cleaning Process(PID: %d) for LaG %s ", pid, it.c_str());
+            kill(pid, signo);
+        }
+    }
+
+    return;
+}
+
 void TeamMgr::doLagTask(Consumer &consumer)
 {
     SWSS_LOG_ENTER();
@@ -171,6 +247,7 @@ void TeamMgr::doLagTask(Consumer &consumer)
                 }
 
                 m_lagList.insert(alias);
+                addLagPid(alias);
             }
 
             setLagAdminStatus(alias, admin_status);
@@ -187,6 +264,7 @@ void TeamMgr::doLagTask(Consumer &consumer)
             {
                 removeLag(alias);
                 m_lagList.erase(alias);
+                removeLagPid(alias);
             }
         }
 
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -173,12 +173,12 @@ void TeamMgr::cleanTeamProcesses(int signo)
 
         if(pid > 0)
         {
-            SWSS_LOG_WARN("Sending Signal to (PID: %d) for LaG %s ", pid, it.c_str());
+            SWSS_LOG_INFO("Sending TERM Signal to (PID: %d) for LaG %s ", pid, it.c_str());
             kill(pid, signo);
         }
         else
         {
-            SWSS_LOG_WARN("Cound not find PID corresponding to LaG %s ", it.c_str());
+            SWSS_LOG_ERROR("Can't send TERM signal to LAG %s. PID wasn't found", it.c_str());
         }
     }
 
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -14,7 +14,6 @@
 #include <net/if.h>
 #include <sys/ioctl.h>
 #include <sys/stat.h>
-#include <sys/types.h>
 #include <signal.h>
 
 #define PID_FILE_PATH "/var/run/teamd/"
@@ -114,11 +113,10 @@ void TeamMgr::doTask(Consumer &consumer)
 }
 
 
-#define MAX_PID_STRLEN 10
-int TeamMgr::getTeamPid(const string &alias)
+pid_t TeamMgr::getTeamPid(const string &alias)
 {
     SWSS_LOG_ENTER();
-    int pid = 0;
+    pid_t pid = 0;
 
     string file = string(PID_FILE_PATH) + alias + string(".pid");
     ifstream infile(file);
@@ -147,7 +145,6 @@ int TeamMgr::getTeamPid(const string &alias)
 }
 
 
-#define MAX_PID_STRLEN 10
 void TeamMgr::addLagPid(const string &alias)
 {
     SWSS_LOG_ENTER();
@@ -162,7 +159,7 @@ void TeamMgr::removeLagPid(const string &alias)
 
 void TeamMgr::cleanTeamProcesses(int signo)
 {
-    int pid = 0;
+    pid_t pid = 0;
 
     for (const auto& it: m_lagList)
     {
@@ -176,8 +173,12 @@ void TeamMgr::cleanTeamProcesses(int signo)
 
         if(pid > 0)
         {
-            SWSS_LOG_WARN("Cleaning Process(PID: %d) for LaG %s ", pid, it.c_str());
+            SWSS_LOG_WARN("Sending Signal to (PID: %d) for LaG %s ", pid, it.c_str());
             kill(pid, signo);
+        }
+        else
+        {
+            SWSS_LOG_WARN("Cound not find PID corresponding to LaG %s ", it.c_str());
         }
     }
 
--- a/cfgmgr/teammgr.h
+++ b/cfgmgr/teammgr.h
@@ -49,9 +49,9 @@ private:
     bool setLagMtu(const std::string &alias, const std::string &mtu);
     bool setLagLearnMode(const std::string &alias, const std::string &learn_mode);
  
-    int getTeamPid(const string &alias);
-    void addLagPid(const string &alias);
-    void removeLagPid(const string &alias);
+    int getTeamPid(const std::string &alias);
+    void addLagPid(const std::string &alias);
+    void removeLagPid(const std::string &alias);
 
     bool isPortEnslaved(const std::string &);
     bool findPortMaster(std::string &, const std::string &);
--- a/cfgmgr/teammgr.h
+++ b/cfgmgr/teammgr.h
@@ -7,6 +7,7 @@
 #include "netmsg.h"
 #include "orch.h"
 #include "producerstatetable.h"
+#include <sys/types.h>
 
 namespace swss {
 
@@ -31,7 +32,7 @@ private:
     ProducerStateTable m_appLagTable;
 
     std::set<std::string> m_lagList;
-    std::map<std::string, int> m_lagPIDList;
+    std::map<std::string, pid_t> m_lagPIDList;
 
     MacAddress m_mac;
 
@@ -49,7 +50,7 @@ private:
     bool setLagMtu(const std::string &alias, const std::string &mtu);
     bool setLagLearnMode(const std::string &alias, const std::string &learn_mode);
  
-    int getTeamPid(const std::string &alias);
+    pid_t getTeamPid(const std::string &alias);
     void addLagPid(const std::string &alias);
     void removeLagPid(const std::string &alias);
 
--- a/cfgmgr/teammgr.h
+++ b/cfgmgr/teammgr.h
@@ -17,6 +17,8 @@ public:
             const std::vector<TableConnector> &tables);
 
     using Orch::doTask;
+    void cleanTeamProcesses(int signo);
+
 private:
     Table m_cfgMetadataTable;   // To retrieve MAC address
     Table m_cfgPortTable;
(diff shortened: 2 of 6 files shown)


madhukar-kamarapu:
@judyjoseph  - this change would address bunch of issues we've been seeing in teamsyncd during config-reload and docker-restart. Thanks again.
I have no major comments/concerns on this PR. One minor (log related) comment - can you please look into it.

teamsyncd/teamsync.cpp
```diff
@@ -194,6 +215,41 @@ void TeamSync::removeLag(const string &lagName)
...
+       if (fd > 0)
+       {
+           memset(pid_str, 0, MAX_PID_STRLEN);
+           sz = read(fd, pid_str, MAX_PID_STRLEN-1);
+           if(sz > 0)
+           {
+               pid = (pid_t)strtol(pid_str, &endptr, 10);
+               if(pid > 0) 
+               {
+                   kill(pid, signo);
```
madhukar-kamarapu: Can you please add a log (INFO/NOTICE) here?
Log should contain port-channel name, pid, signo.

teamsyncd/teamsync.cpp
```diff
@@ -194,6 +215,41 @@ void TeamSync::removeLag(const string &lagName)
...
+#define MAX_PID_STRLEN 10
+void TeamSync::cleanTeamProcesses(int signo)
+{
+    ssize_t sz = 0;
+    int fd, pid;
+    char pid_str[MAX_PID_STRLEN] = {0};
+    char *endptr = NULL;
+    string file_path;
+
+    for (auto it=m_teamSelectables.begin(); it!=m_teamSelectables.end(); ++it)
```
pavel-shirshov: Suggest to change it to
```
for (const auto& it: m_teamSelectables)
```

teamsyncd/teamsync.cpp
```diff
@@ -194,6 +215,41 @@ void TeamSync::removeLag(const string &lagName)
...
+    {
+       /* There are PID files created for each teamd process in /var/run/teamd DIR */
+       file_path = string(PID_FILE_PATH) + it->first + string(".pid");
+
+       /* Open the file, get the PID and send signal to the process */
+       fd = open(file_path.c_str(), O_RDONLY);
```
pavel-shirshov: probably it's easier to use c++ streams here.

teamsyncd/teamsync.cpp
```diff
@@ -194,6 +215,41 @@ void TeamSync::removeLag(const string &lagName)
...
+       if (fd > 0)
+       {
```
pavel-shirshov: should we report if fd < 0?

teamsyncd/teamsync.cpp
```diff
@@ -194,6 +215,41 @@ void TeamSync::removeLag(const string &lagName)
...
+           memset(pid_str, 0, MAX_PID_STRLEN);
```
pavel-shirshov: move the following line inside of this code block, and after that we can remove memset from here
```
char pid_str[MAX_PID_STRLEN] = {0};
```

teamsyncd/teamsync.cpp
```diff
@@ -194,6 +215,41 @@ void TeamSync::removeLag(const string &lagName)
...
+           sz = read(fd, pid_str, MAX_PID_STRLEN-1);
+           if(sz > 0)
+           {
```
pavel-shirshov: error message if can't read

teamsyncd/teamsync.cpp
```diff
@@ -194,6 +215,41 @@ void TeamSync::removeLag(const string &lagName)
...
+               pid = (pid_t)strtol(pid_str, &endptr, 10);
+               if(pid > 0)
```
pavel-shirshov: error message if can't convert

judyjoseph:
> @judyjoseph  - this change would address bunch of issues we've been seeing in teamsyncd during config-reload and docker-restart. Thanks again.
> I have no major comments/concerns on this PR. One minor (log related) comment - can you please look into it.



> @judyjoseph  - this change would address bunch of issues we've been seeing in teamsyncd during config-reload and docker-restart. Thanks again.
> I have no major comments/concerns on this PR. One minor (log related) comment - can you please look into it.

I did testing with config reload and config loadminigraph and seen things are ok. These commands does docker stop and docker start explicitly in a sequence starting with swss docker. 

What is the use case of docker restart ? are you referring to "docker restart teamd"? Is this command supported ? Currently I see swss, syncd, teamd tightly bound and think it will go to an inconsistent state if we just restart teamd alone.

teamsyncd/teamsync.cpp
judyjoseph: It was intentional to not add any debug logs in this API as this is a "singal handler". The logging mechanishm will include library calls having buffer manipulations and other resource (blocking) requirements which should be avoided in a signal handler as per Linux system programing guide.

teamsyncd/teamsync.cpp
judyjoseph: Plan was to use the signal safe library calls ( as in Linux Programmer's Manual ) to the maximum extend.

teamsyncd/teamsync.cpp
judyjoseph: Ok shall use this instead, it is fine.

teamsyncd/teamsync.cpp
judyjoseph: Approach I have taken is to avoid ( to the maximum extend ) logging/print mesages in the signal handlers are they are not signal safe.

teamsyncd/teamsync.cpp
judyjoseph: Same comment as above

teamsyncd/teamsync.cpp
judyjoseph: Same as above comment to minimize logging as this function was added as a Signal handler. I will add a Log message in the TeamPortSync destructor, will be helpful in debugging -- logging the cleanup of each Portchannel.

teamsyncd/teamsync.cpp
madhukar-kamarapu: okay.

madhukar-kamarapu:
> the

I was referring to "systemctl restart teamd". Teamd-docker upgrades generally issue this command.
Say, there are NO configurations on the port-channel (VLAN, IP, etc) => docker upgrade can be done (it wouldn't impact applications). 
Note: If there is any application configuration - all applications need to register for port-channel events like (create/delete/up/down/etc) to handle the teamd-docker-upgrade OR teamd-service-restart. Reason - new port-channel netdevices (with new ifindex) are created in the kernel.

madhukar-kamarapu:
> > @judyjoseph  - this change would address bunch of issues we've been seeing in teamsyncd during config-reload and docker-restart. Thanks again.
> > I have no major comments/concerns on this PR. One minor (log related) comment - can you please look into it.
> 
> > @judyjoseph  - this change would address bunch of issues we've been seeing in teamsyncd during config-reload and docker-restart. Thanks again.
> > I have no major comments/concerns on this PR. One minor (log related) comment - can you please look into it.
> 
> I did testing with config reload and config loadminigraph and seen things are ok. These commands does docker stop and docker start explicitly in a sequence starting with swss docker.
> 
> What is the use case of docker restart ? are you referring to "docker restart teamd"? Is this command supported ? Currently I see swss, syncd, teamd tightly bound and think it will go to an inconsistent state if we just restart teamd alone.

I was referring to "systemctl restart teamd". Teamd-docker upgrades generally issue this command.
Say, there are NO configurations on the port-channel (VLAN, IP, etc) => docker upgrade can be done (it wouldn't impact applications).
Note: If there is any application configuration - all applications need to register for port-channel events like (create/delete/up/down/etc) to handle the teamd-docker-upgrade OR teamd-service-restart. Reason - new port-channel netdevices (with new ifindex) are created in the kernel.

teamsyncd/teamsync.cpp
judyjoseph: Added more Logs now.

teamsyncd/teamsync.cpp
judyjoseph: Added more logs now.

teamsyncd/teamsync.cpp
judyjoseph: Added more Logs now. Please take a look

teamsyncd/teamsync.cpp
judyjoseph: Done

teamsyncd/teamsync.cpp
judyjoseph: Added more Logs now.

cfgmgr/teammgr.h
```diff
@@ -29,6 +31,7 @@ class TeamMgr : public Orch
     ProducerStateTable m_appLagTable;
 
     std::set<std::string> m_lagList;
+    std::map<std::string, int> m_lagPIDList;
```
pavel-shirshov: Suggest to use `pid_t` as a type for pid.
http://man7.org/linux/man-pages/man2/kill.2.html

cfgmgr/teammgr.cpp
```diff
@@ -108,6 +113,77 @@ void TeamMgr::doTask(Consumer &consumer)
...
+        if(!pid) {
+            SWSS_LOG_WARN("Invalid PID found for LaG %s ", it.c_str());
+
+            /* Try to get the PID again */
+            pid = getTeamPid(it);
+        }
+
+        if(pid > 0)
+        {
+            SWSS_LOG_WARN("Cleaning Process(PID: %d) for LaG %s ", pid, it.c_str());
```
pavel-shirshov: Suggest to use "Sending signal" instead of "Cleaning Process

cfgmgr/teammgr.cpp
```diff
@@ -108,6 +113,77 @@ void TeamMgr::doTask(Consumer &consumer)
...
+            kill(pid, signo);
+        }
```
pavel-shirshov: Add syslog output, that you didn't find pid for some portchannel and can't send it a signal

cfgmgr/teammgrd.cpp
```diff
@@ -55,6 +67,12 @@ int main(int argc, char **argv)
 
         while (true)
```
pavel-shirshov: I'd replace it with `while (!received_sigterm) {`

cfgmgr/teammgrd.cpp
```diff
@@ -55,6 +67,12 @@ int main(int argc, char **argv)
...
         {
+            if(received_sigterm)
+            {
+                teammgr.cleanTeamProcesses(SIGTERM);
+                received_sigterm = false;
```
pavel-shirshov: should we stop this process too?

teamsyncd/teamsyncd.cpp
```diff
@@ -33,6 +45,12 @@ int main(int argc, char **argv)
             s.addSelectable(&netlink);
             while (true)
             {
+                if(received_sigterm)
+                {
+                  sync.cleanTeamSync();
+                  received_sigterm = false;
+                }
```
pavel-shirshov: should we exit from this process?

cfgmgr/teammgr.h
judyjoseph: Updated from int to pid_t

cfgmgr/teammgr.cpp
judyjoseph: Done

cfgmgr/teammgr.cpp
judyjoseph: Done

cfgmgr/teammgrd.cpp
judyjoseph: Since we need to have Signal Handlers in both the teammgrd and teamsyncd and both have resources to be cleaned up which are "interdependent", I found it is right to let the process continue and get killed by SIGKILL later. If we add explicit process exits, since they are exiting on their on pace, on testing I found some of the PortChannel interfaces remaining in the kernel and not getting cleaned.

teamsyncd/teamsyncd.cpp
judyjoseph: Since we need to have Signal Handlers in both the teammgrd and teamsyncd and both have resources to be cleaned up which are "interdependent", I found it is right to let the process continue and get killed by SIGKILL later. If we add explicit process exits, since they are exiting on their on pace, on testing I found some of the PortChannel interfaces remaining in the kernel and not getting cleaned.

cfgmgr/teammgr.cpp
```diff
@@ -108,6 +112,79 @@ void TeamMgr::doTask(Consumer &consumer)
...
+
+        if(pid > 0)
+        {
+            SWSS_LOG_WARN("Sending Signal to (PID: %d) for LaG %s ", pid, it.c_str());
+            kill(pid, signo);
+        }
+        else
+        {
+            SWSS_LOG_WARN("Cound not find PID corresponding to LaG %s ", it.c_str());
```
pavel-shirshov: Can you please make it as an error?
Also It's better to say "Can't send TERM signal to LAG %s. PID wasn't found"

cfgmgr/teammgrd.cpp
```diff
@@ -17,13 +18,24 @@ bool gLogRotate = false;
 ofstream gRecordOfs;
 string gRecordFile;
 
+bool received_sigterm = false;
+
+void sig_handler(int signo)
```
qiluo-msft: >signo [](start = 21, length = 5)

check signo == SIGTERM first?

teamsyncd/teamsyncd.cpp
```diff
@@ -9,6 +10,14 @@
 using namespace std;
 using namespace swss;
 
+bool received_sigterm = false;
+
+void sig_handler(int signo)
```
qiluo-msft: >signo [](start = 21, length = 5)

the same

teamsyncd/teamsyncd.cpp
```diff
@@ -33,6 +45,12 @@ int main(int argc, char **argv)
             s.addSelectable(&netlink);
             while (true)
             {
+                if(received_sigterm)
+                {
+                  sync.cleanTeamSync();
```
qiluo-msft: >   [](start = 16, length = 2)

Inconsistent indentation.

teamsyncd/teamsyncd.cpp
judyjoseph: The Signal handler is only registered only for SIGTERM now. The handler will be called only on SIGTERM signal.

cfgmgr/teammgrd.cpp
judyjoseph: The Signal handler is only registered only for SIGTERM now. The handler will be called only on SIGTERM signal.

Issue #1016, opened 2019-08-02
Vnet route seems can not support ipv6 route?


ryan44guo:
Hi all:
    Although we do not add ipv6 route to vnet, but linux kernel will create some routes (like ff00::/8 or fe80::/64, etc.), because of the vrf creation. RouteSync distributes these routes with "Vnet" ahead to VNET_ROUTE_TABLE now.
    And VnetOrch use request as its base class, and use ":" as its key_separator_,so when these routes appears, it can not be parsed as we needed.
    Don't you think there should be a special request_parser for VNET_ROUTE?

prsunny:
This is a current limitation. VNET V6 route is not supported. Need some modification on Request Parser to fix this. Let me know if you would like to fix this

ryan44guo:
Can you suffer to use another separator on vnet route? 
It seems modify the public parser is difficult, although we can define private parser for vnetRoute, but the class request only exposes its parser to us(parseKey is private, parseIpPrefix too, I doubt this class is not well defined for subclass use its own parser), we need do many things duplicated to parse it.

PR #1160, opened 2019-12-27
Support ipv6 route in vnet

liron-ze:
**What I did**
Support ipv6 route in vnet.
**Why I did it**
[https://github.com/Azure/sonic-swss/issues/1016](url)
**How I verified it**
tests
**Details if related**

### DIFF
--- a/orchagent/vnetorch.h
+++ b/orchagent/vnetorch.h
@@ -61,7 +61,7 @@ extern std::vector<VR_TYPE> vr_cntxt;
 class VNetRequest : public Request
 {
 public:
-    VNetRequest() : Request(vnet_request_description, ':') { }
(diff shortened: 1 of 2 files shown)


orchagent/vnetorch.h
```diff
@@ -61,7 +61,7 @@ extern std::vector<VR_TYPE> vr_cntxt;
 class VNetRequest : public Request
 {
 public:
-    VNetRequest() : Request(vnet_request_description, ':') { }
+    VNetRequest() : Request(vnet_request_description, '|') { }
```
prsunny: @liron-ze ,` | `is used as a delimiter in config_db. Can you keep the same `:` and try fixing this in `Request `class?

PR #1161, opened 2020-01-02
[vlanmgr] configure mtu while creating the DOT1Q_BRIDGE 

ouxlwhu:
**What I did**
set default mtu while creating the DOT1Q_BRIDGE

**Why I did it**
MTU configuration on VLAN interface has no effect in linux kernel.
like this
```
admin@48Y-oxl:~$ sudo config interface mtu Vlan1000 7000
admin@48Y-oxl:~$ ip link show Vlan1000
12: Vlan1000@Bridge: <NO-CARRIER,BROADCAST,MULTICAST,ALLMULTI,UP> mtu 1500 qdisc noqueue state LOWERLAYERDOWN mode DEFAULT group default qlen 1000
    link/ether 00:a0:c9:00:00:00 brd ff:ff:ff:ff:ff:ff
admin@48Y-oxl:~$ 
```

**How I verified it**
retest it
```
admin@48Y-oxl:~$ sudo config interface mtu Vlan1000 7000
admin@48Y-oxl:~$ ip link show Vlan1000
70: Vlan1000@Bridge: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 7000 qdisc noqueue state LOWERLAYERDOWN mode DEFAULT group default qlen 1000
    link/ether 00:a0:c9:00:00:00 brd ff:ff:ff:ff:ff:ff

```

**Details if related**
related with  bug：Wrong MTU size in ICMP (type 3,code 4) if fragmentation is needed. Reproduced for interface Vlan only. #3942
https://github.com/Azure/sonic-buildimage/issues/3942


liat-grozovik:
will it also work when the MTU is set following the interface up and running? i understand you set the default but what will happen if during system runtime we will change the MTU via the CLI to a different number and it will be saved in the config_db?

cfgmgr/vlanmgr.cpp
```diff
@@ -50,19 +50,19 @@ VlanMgr::VlanMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
...
     //               /sbin/ip link del dummy 2>/dev/null;
-    //               /sbin/ip link add dummy type dummy &&
+    //               /sbin/ip link add dummy mtu 9100 type dummy &&
     //               sbin/ip link set dummy master Bridge"
 
     const std::string cmds = std::string("")
       + BASH_CMD + " -c \""
       + IP_CMD + " link del " + DOT1Q_BRIDGE_NAME + " 2>/dev/null; "
-      + IP_CMD + " link add " + DOT1Q_BRIDGE_NAME + " up type bridge && "
+      + IP_CMD + " link add " + DOT1Q_BRIDGE_NAME + " up mtu " + DEFAULT_MTU_STR + " type bridge && "
```
prsunny: Does kernel changes this configured Vlan interface MTU when you add another interface to Vlan with a lower MTU value?

prsunny:
retest this please

PR #1162, opened 2020-01-02
using PRI instead of %l to support 32 bit arch

banagiri:
Compiling for 32 bit arch causes compilation issue with %l variants

**replace %l variants with PRI**

**to fix compilation for 32 bit arch**

**compile for 32 bit arch (marvell-armhf)**

### DIFF
--- a/orchagent/debug_counter/debug_counter.h
+++ b/orchagent/debug_counter/debug_counter.h
@@ -4,6 +4,9 @@
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
+#define __STDC_FORMAT_MACROS
+#include <inttypes.h>
+
 
 extern "C" {
 #include "sai.h"
--- a/orchagent/debug_counter/debug_counter.h
+++ b/orchagent/debug_counter/debug_counter.h
@@ -4,7 +4,6 @@
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
-#define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
 
--- a/orchagent/copporch.cpp
+++ b/orchagent/copporch.cpp
@@ -415,7 +415,7 @@ bool CoppOrch::removeGenetlinkHostIf(string trap_group_name)
                 sai_status = sai_hostif_api->remove_hostif_table_entry(hostTableEntry->second);
                 if(sai_status != SAI_STATUS_SUCCESS)
                 {
-                    SWSS_LOG_ERROR("Failed to delete hostif table entry %ld \
+                    SWSS_LOG_ERROR("Failed to delete hostif table entry %" PRId64 " \
                                    on trap group %s. rc=%d", hostTableEntry->second,
                                    trap_group_name.c_str(), sai_status);
                     return false;
@@ -431,7 +431,7 @@ bool CoppOrch::removeGenetlinkHostIf(string trap_group_name)
         sai_status = sai_hostif_api->remove_hostif(hostInfo->second);
         if(sai_status != SAI_STATUS_SUCCESS)
         {
-            SWSS_LOG_ERROR("Failed to delete host info %ld on trap group %s. rc=%d",
+            SWSS_LOG_ERROR("Failed to delete host info %" PRId64 " on trap group %s. rc=%d",
                            hostInfo->second, trap_group_name.c_str(), sai_status);
             return false;
         }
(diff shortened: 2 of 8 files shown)


orchagent/debug_counter/debug_counter.h
```diff
@@ -4,6 +4,9 @@
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
+#define __STDC_FORMAT_MACROS
+#include <inttypes.h>
```
jleveque: These lines are not needed in the header file. Also you are defining `__STDC_FORMAT_MACROS` in multiple header files, so there is a chance we will see redefinition warnings. I suggest one of the two alternatives:

1. Add these two lines in a common header file which is included by all cpp files (if there is one), and remove from all other files.
2. Move these two lines into the C++ files which actually use the "PRI" formatter macros.

BONUS: Only add the `#include <inttypes.h>` line to the source files and define `__STDC_FORMAT_MACROS` on the GCC command line (can add `-D__STDC_FORMAT_MACROS` to CFLAGS_COMMON in configure.ac), thus ensuring it is only defined once. Note, however, that with this approach, it is not as clear why we are defining `__STDC_FORMAT_MACROS`.

jleveque: After your most recent change, you have removed all definitions of `__STDC_FORMAT_MACROS`. This needs to be defined somewhere before all inttypes.h includes if you want to use the "PRI" macros in C++ code. In my comment above, I just wanted to see it defined in one place rather than in many.

Also, the `#include <inttypes.h>` lines are still present in header files, even though the "PRI" formatter macros are only used in .cpp files. Thus, this include is only needed in .cpp files, not in .h files.

banagiri: @jleveque 
According to latest include/inttypes.h file, the macro __STDC_FORMAT_MACROS is not being used when including the PRI* definitions, it was an older implementation which is dependent.
Also, for c++ compiler implementations, it is mentioned that compilers skip the usage of the macro.
So, it's no harm in not defining this macro.

Maintaining the #include <inttypes.h> in h files can help other cpp files which depend on the same feature header file not to include the header file again, as it is protected by ```#ifndef _INTTYPES_H``` ```#define _INTTYPES_H	1``` duo.

jleveque: > @jleveque
> According to latest include/inttypes.h file, the macro __STDC_FORMAT_MACROS is not being used when including the PRI* definitions, it was an older implementation which is dependent.
> Also, for c++ compiler implementations, it is mentioned that compilers skip the usage of the macro.
> So, it's no harm in not defining this macro.

Where did you see this? Can you provide a link? All documentation I can find still mentions that  `__STDC_FORMAT_MACROS` must be defined if using a C++ complier.

> Maintaining the #include <inttypes.h> in h files can help other cpp files which depend on the same feature header file not to include the header file again, as it is protected by `#ifndef _INTTYPES_H` `#define _INTTYPES_H 1` duo.

True, but at the same time, it's not as explicit; it's not obvious why something works if it is because of a header file included in a different header. I prefer to include header files at the narrowest scope necessary.

banagiri: https://sourceware.org/bugzilla/show_bug.cgi?id=15366

Also from an answer in a forum
```
However the C++ committee chose to ignore the suggestion: e.g. in the n3242 draft, 27.9.2/3: Note: The macros defined by <cinttypes> are provided unconditionally. In particular, the symbol __STDC_FORMAT_MACROS, mentioned in footnote 182 of the C standard, plays no role in C++. So when the compilers catch up, we won't need __STDC_FORMAT_MACROS in either C or C++. ```

jleveque: Thanks for the info. I think for the time being we should still define `__STDC_FORMAT_MACROS` until we can confirm our version of glibc and/or gcc will work without it, then we can remove it.

banagiri: @jleveque
At present, we are compiling sonic without these flags and compilation is fine

banagiri: @jleveque
Do you have any other comments?

jleveque:
Retest vs please

jleveque:
Retest this please

banagiri:
> Retest this please

@jleveque
we have compiled and tested this on ARM 32 bit variant and x86_64 also.

PR #1167, opened 2020-01-14
[qosorch] converting shaper bandwidth value to unsigned long instead …

Sabareesh-Kumar-Anandan:
…of int

**What I did**
I have changed the conversion of shaper bandwidth str to unsigned long instead of int.
**Why I did it**
For higher shaper rates greater than 20Gbps, integer is not sufficient. So I changed the conversion function to stoul()
**How I verified it**
Verified by using shaper rates greater than 20Gbps.
**Details if related**


lguohan:
retest this please

PR #1170, opened 2020-01-15
[portsorch] Refactor portsorch to use FlexCounterManager to setup port and queue stats

daall:
- Updates portsorch to use FlexCounterManager for port and queue stats instead of direct redis operations
- Updates FlexCounterManager to support standard Port and Queue counters
- Updates FlexCounterManager to allow clients to enable the group in the constructor, for convenience
- Updates the makefile for cfgmgr to include new flex_counter dependency from portsorch

**What I did**
I refactored portsorch to use FlexCounterManager to handle flex counter operations rather than directly interacting with the FLEX_COUNTER_DB.

**Why I did it**
I did it to abstract away some of the details of flex counter setup from portsorch and make portsorch simpler. This is also makes it easier to update the flex counter implementation in the future.

**How I verified it**
I built a sonic image with the changes and verified that the FLEX_COUNTER_DB was the same as before, and that COUNTERS_DB was being updated properly. I also checked that portstat and queuestat were behaving correctly as well.


daall:
retest this please

daall:
retest this please

daall:
retest this please

PR #1173, opened 2020-01-16
[debugcounterorch] Add checks for supported counter types and drop reasons

### LINKED ISSUE #1136: Unsupported drop counter types/drop reasons cause syncd to crash with bcm SAI
Attempting to create a drop counter with an unsupported type or with unsupported drop reasons will cause syncd to crash on Broadcom devices. (Note: This is only possible if you attempt to modify Config DB directly - the CLI will prevent users from pushing a bad config in the first place.)

This is a known issue and we're working with Broadcom to resolve it.

daall:
- Refactor drop reason capability query to trim SAI prefixes
- Store device capabilities in orchagent to perform safety checks

**What I did**
I added a check in orchagent to prevent unsupported drop counter types and drop reasons from crashing orchagent.

**Why I did it**
I did this to prevent devices from crashing due to inadvertent configuration errors.

**How I verified it**
I locally tested by:
1. Inserting a SWITCH_INGRESS counter into config DB on a device where SWITCH_INGRESS_COUNTER was not supported
2. Adding L2_ANY to a counter on a device where L2_ANY was not supported.
3. Created a valid counter on a device to ensure that normal operation was not impacted.

**Details if related**
Fixes #1136 - Rather than depending on each ASIC vendor to follow the same error handling doctrine, this PR validates HW support in orchagent, which should be more reliable.

Related to Azure/sonic-utilities#785 - In order to validate user input, we need to remove the SAI prefixes before we store the results. This removes the need for the CLI to perform these checks.

### DIFF
--- a/orchagent/debug_counter/drop_counter.cpp
+++ b/orchagent/debug_counter/drop_counter.cpp
@@ -3,6 +3,8 @@
 #include "logger.h"
 #include "sai_serialize.h"
 
+#include <vector>
+
 using std::runtime_error;
 using std::string;
 using std::unordered_map;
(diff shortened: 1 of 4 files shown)


daall:
retest this please

PR #1180, opened 2020-01-23
Define lacp key to make possible creation of two lacp channels between couple of switches

pavel-shirshov:
Fix: https://github.com/Azure/sonic-buildimage/issues/4009

**What I did**
Added `teamdctl port config update` command to set lacp key. Also check [why I install configuration before we add interface to a channel](https://lists.fedorahosted.org/archives/list/libteam@lists.fedorahosted.org/message/57SCT3BDPBCWO5ZMJUKHBKYZ37VJFS26/)

**Why I did it**
Otherwise it's impossible to set more than 1 channel between two systems.

**How I verified it**
Build an image and run on my dut. Check tcpdump of lacp packets. Also portchannels must be up.

**Details if related**

### DIFF
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -134,7 +134,7 @@ pid_t TeamMgr::getTeamPid(const string &alias)
     {
         SWSS_LOG_WARN("The LAG PID file: %s is empty", file.c_str());
     }
-    else 
+    else
     {
         /*Store the PID value */
         pid = stoi(line, nullptr, 10);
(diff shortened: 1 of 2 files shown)


cfgmgr/teammgr.cpp
```diff
@@ -541,6 +541,8 @@ task_process_status TeamMgr::addLag(const string &alias, int min_links, bool fal
         return task_need_retry;
     }
 
+    m_lagKeys[alias] = m_lacp_key++;
```
pavel-shirshov: I need to make lacp_key from portchannel name, otherwise after warm reboot we might got wrong lacp_key

Issue #1182, opened 2020-01-27
warm reboot vs tests unstable


daall:
The following warm reboot vs tests are marginal at the moment:
- test_system_warmreboot_neighbor_syncup

This causes the suite to fail ~40-50% of the time across the swss and utilities tests, as well as the nightly swss build.

For now we can mark the test as xfail in pytest so that the failures are not blocking PRs, but it needs to be stabilized and put back into the test suite.

Issue #1181, opened 2020-01-27
VLAN vs tests unstable


daall:
The following VLAN vs tests are marginal at the moment:
- test_RemoveVlanWithRouterInterface
- test_AddPortChannelToVlan

This causes the suite to fail ~40-50% of the time across the swss and utilities tests, as well as the nightly swss build.

For now we can mark the tests as xfail in pytest so that the failures are not blocking PRs, but they need to be stabilized and put back into the test suite.

daall:
Fixed by #1241

PR #1183, opened 2020-01-27
[vs tests] Mark VLAN and warm reboot tests as xfail

daall:

**What I did**
I marked the VLAN and warm reboot vs tests as xfail, which means that test failures in those tests will not cause the entire test suite to fail.

**Why I did it**
I did it to try to reduce the number of test suite failures so that we don't have to spam "retest this please" to unblock PRs.

**Details if related**
Opened #1181 and #1182 to track test stabilization.

### DIFF
--- a/tests/test_warm_reboot.py
+++ b/tests/test_warm_reboot.py
@@ -235,10 +235,6 @@ def ping_new_ips(dvs):
             dvs.runcmd(['sh', '-c', "ping6 -c 1 -W 0 -q {}00::{} > /dev/null 2>&1".format(i*4,j+NUM_NEIGH_PER_INTF+2)])
 
 
-# FIXME: This test is extremely unstable and requires several retries
-# for it to pass - we need to stabilize this test before putting it back
-# into the pipeline.
-@pytest.mark.xfail(reason="warm reboot test unstable")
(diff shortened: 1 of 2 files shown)


tests/test_warm_reboot.py
```diff
@@ -234,6 +234,11 @@ def ping_new_ips(dvs):
             dvs.runcmd(['sh', '-c', "ping -c 1 -W 0 -q {}.0.0.{} > /dev/null 2>&1".format(i*4,j+NUM_NEIGH_PER_INTF+2)])
             dvs.runcmd(['sh', '-c', "ping6 -c 1 -W 0 -q {}00::{} > /dev/null 2>&1".format(i*4,j+NUM_NEIGH_PER_INTF+2)])
 
+
+# FIXME: This test is extremely unstable and requires several retries
+# for it to pass - we need to stabilize this test before putting it back
+# into the pipeline.
+@pytest.mark.xfail(reason="warm reboot test unstable")
```
lguohan: only neighbor syncup test is unstable.

daall: I narrowed down the granularity a little better, we should only have 3 test cases that are being ignored as opposed to ~30 now.

PR #1186, opened 2020-01-28
[vs tests] Activate flaky plugin for virtual switch pytests

daall:

**What I did**
I turned on the "flaky" pytest plugin for the vs tests. This plugin will automatically retry failed test cases one time before failing the test suite altogether.

**Why I did it**
Currently, because of the async nature of the vs tests, any given test case has about a ~5% chance of failing, which will fail the entire build. This forces them to re-run the entire test suite to get a passing result. Rather than making people re-run the entire 2 hour process, this fix will simply retry failed test cases when they fail, which should save all of us a lot of time.

**Details if related**
This is **not** a permanent solution, simply a means to un-block people until we can get all of the virtual switch tests stabilized.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -8,8 +8,9 @@ SWSS Integration tests runs on docker-sonic-vs which runs on top of SAI virtual 
 
 - Install docker and pytest on your dev machine
     ```
-    sudo pip install --system docker==3.5.0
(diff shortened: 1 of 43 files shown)


lguohan:
let's mark all of them as flaky and then get statistics to know which one are really flaky or not.

PR #1192, opened 2020-02-03
[team sync/mgr] Add debug message before cleaning up LAGs

yxieca:
**What I did**

These functions are called only once in the time of service stopping,
add events will help debug. And these messages won't be chatty.

Signed-off-by: Ying Xie <ying.xie@microsoft.com>

**Why I did it**
Once the cleanup code was on radar as a suspect. If we had these debug messages, it would have cleared the suspect very quickly.

### DIFF
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -163,6 +163,7 @@ void TeamMgr::cleanTeamProcesses(int signo)
 {
     pid_t pid = 0;
 
+    SWSS_LOG_ENTER();
     SWSS_LOG_NOTICE("Cleaning up LAGs during shutdown...");
     for (const auto& it: m_lagList)
     {
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -163,6 +163,7 @@ void TeamMgr::cleanTeamProcesses(int signo)
 {
     pid_t pid = 0;
 
+    SWSS_LOG_NOTICE("Cleaning up LAGs during shutdown...");
     for (const auto& it: m_lagList)
     {
         pid = m_lagPIDList[it];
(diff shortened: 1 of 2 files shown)


teamsyncd/teamsync.cpp
```diff
@@ -212,6 +212,8 @@ void TeamSync::removeLag(const string &lagName)
 
 void TeamSync::cleanTeamSync()
 {
+    SWSS_LOG_NOTICE("Cleaning up kernel LAGs ...");
```
daall: Don't you need the `SWSS_LOG_ENTER();` here as well?

cfgmgr/teammgr.cpp
```diff
@@ -163,6 +163,7 @@ void TeamMgr::cleanTeamProcesses(int signo)
 {
     pid_t pid = 0;
 
+    SWSS_LOG_NOTICE("Cleaning up LAGs during shutdown...");
```
daall: I don't think this method has `SWSS_LOG_ENTER();` either.

teamsyncd/teamsync.cpp
judyjoseph: Ideally removeLag() is getting triggered when the kernel LAG interface is deleted. In cleanTeamSync() we are actually cleaning up some teamd resources by calling removeLag() explicitly for all LAG interfaces which are still present.

The Kernel LAG interfaces are removed actually when we do a kill of teamd process in teammgrd process in cleanTeamProcesses()

So the NOTICE message may not be the right wording. If we need to keep it, we could put it as "Cleaning up LAG teamd resources" or so.

yxieca: Thanks, Added.

yxieca: Changed as suggested.

cfgmgr/teammgr.cpp
yxieca: Added. Thanks!

PR #1217, opened 2020-03-05
[vnet]: Fix continues warning messages after config reload

volodymyrsamotiy:

**What I did**
In VNET route remove handler return ```true``` instead of ```false``` if route is not found in the internal vnetorch cache. It is needed because vnetorch might receive double ```remove``` event and it causes endless warning messages since event will be forever in a message queue. In order to remove it from the queue just need to return ```true``` to the caller (Orch2::doTask).
**Why I did it**
To fix continues warning messages (shown below) after config reload with the saved VNETs and VNET routes.
```WARNING swss#orchagent: :- removeTunnelRoute: VNET tunnel route 172.2.10.0/24 doesn't exist```
**How I verified it**
Added VNET config with the different routes to the config_db.json, perform config reload and verified there are no endless warnings in the log.
**Details if related**
N/A


prsunny:
retest this please

PR #1218, opened 2020-03-06
[ACL][VLAN] Add support for ingress and egress ACLs on router interfaces/SVI

jleveque:
Support attaching ingress and egress ACLs to router interfaces/switch virtual interfaces

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1295,7 +1295,7 @@ bool AclTable::create()
     }
     else
     {
-        bpoint_list = { SAI_ACL_BIND_POINT_TYPE_PORT, SAI_ACL_BIND_POINT_TYPE_LAG };
+        bpoint_list = { SAI_ACL_BIND_POINT_TYPE_PORT, SAI_ACL_BIND_POINT_TYPE_LAG, SAI_ACL_BIND_POINT_TYPE_ROUTER_INTERFACE };
     }
 
     attr.id = SAI_ACL_TABLE_ATTR_ACL_BIND_POINT_TYPE_LIST;
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -33,6 +33,7 @@ extern sai_vlan_api_t *sai_vlan_api;
 extern sai_lag_api_t *sai_lag_api;
 extern sai_hostif_api_t* sai_hostif_api;
 extern sai_acl_api_t* sai_acl_api;
+extern sai_router_interface_api_t *sai_router_intfs_api;
 extern sai_queue_api_t *sai_queue_api;
 extern sai_object_id_t gSwitchId;
 extern IntfsOrch *gIntfsOrch;
@@ -479,7 +480,8 @@ bool PortsOrch::getPort(sai_object_id_t id, Port &port)
             }
             break;
         case Port::VLAN:
-            if (portIter.second.m_vlan_info.vlan_oid == id)
+            if (portIter.second.m_vlan_info.vlan_oid == id ||
+                portIter.second.m_rif_id == id)
             {
                 port = portIter.second;
                 return true;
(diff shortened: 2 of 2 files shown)


orchagent/portsorch.cpp
```diff
@@ -910,7 +919,14 @@ bool PortsOrch::createBindAclTableGroup(sai_object_id_t id, sai_object_id_t &gro
                 bind_type = SAI_ACL_BIND_POINT_TYPE_LAG;
                 break;
             case Port::VLAN:
-                bind_type = SAI_ACL_BIND_POINT_TYPE_VLAN;
+                if (port.m_rif_id)
+                {
+                    bind_type = SAI_ACL_BIND_POINT_TYPE_ROUTER_INTERFACE;
```
prsunny: Why is RIF accounted only for Vlan. I mean Port and Lag (cases above) can also have corresponding RIF.

orchagent/portsorch.cpp
```diff
@@ -479,7 +480,8 @@ bool PortsOrch::getPort(sai_object_id_t id, Port &port)
             }
             break;
         case Port::VLAN:
-            if (portIter.second.m_vlan_info.vlan_oid == id)
+            if (portIter.second.m_vlan_info.vlan_oid == id ||
```
prsunny: I don't see a new `getPort `usage in this PR. So where is it called with `rif_id` as `id`? I think this change may not be required

orchagent/portsorch.cpp
prsunny: With this change, if it is a Vlan router interface, it is always BIND_TYPE_ROUTER_INTERFACE.  This is a behavioral change from existing flow and not sure if it breaks for those who use VLAN binding currently (In most cases, Vlan is a RIF). Do we have a case where the `bind_type` can be TYPE_VLAN even if it has a RIF association? @daall

jleveque: This PR is only meant to support VLAN RIFs.

orchagent/portsorch.cpp
anirudhnr: getPort() is called in getAclBindPortId() to get the port information.

prsunny: `getAclBindPortId()` invokes `getPort(alias, port)` with `alias`. Its not this function.

orchagent/aclorch.cpp
```diff
@@ -1295,7 +1295,7 @@ bool AclTable::create()
     }
     else
     {
-        bpoint_list = { SAI_ACL_BIND_POINT_TYPE_PORT, SAI_ACL_BIND_POINT_TYPE_LAG };
+        bpoint_list = { SAI_ACL_BIND_POINT_TYPE_PORT, SAI_ACL_BIND_POINT_TYPE_LAG, SAI_ACL_BIND_POINT_TYPE_ROUTER_INTERFACE };
```
wendani, line 1298: It is better to be conscious of what BIND_POINT_TYPE is indeed needed according to the input field, as BIND_POINT_TYPE is translated into match field/qualifiers allocation in the hardware that will increase the possibility of using double-wide mode TCAM

orchagent/portsorch.cpp
```diff
@@ -910,7 +919,14 @@ bool PortsOrch::createBindAclTableGroup(sai_object_id_t id, sai_object_id_t &gro
                 bind_type = SAI_ACL_BIND_POINT_TYPE_LAG;
                 break;
             case Port::VLAN:
-                bind_type = SAI_ACL_BIND_POINT_TYPE_VLAN;
+                if (port.m_rif_id)
```
wendani, line 922: Vlan interface may not be created when vlan Port object is created. You may need to introduce a structure similar to AclTable::pendingPortSet in AclTable object to hold those rifs that are pending creation, and be an observer to rif creation event, upon which a rif is ready for ACL binding.

Issue #23, opened 2016-05-03
Add VLAN support


stcheng:
In configuration file port_config.ini, VLAN information will be appended. While reading the file, portsyncd will set up the VLANs. VLAN logic will be addressed in portsOrch class.

stcheng:
done.

PR #1219, opened 2020-03-06
[DPB:orchagent:portsorch] Rework port dependency  (#23)

vasant17:
* Use SAI REDIS return code to track dependency on port.

**What I did**
While deleting a port, if SAI call return OBJECT_IN_USE error code, we will continue to wait/loop in orchagent to delete the port. Once all the dependencies are removed, we will go-ahead remove the port. Note that dependency tracking itself is done in SAI REDIS layer. Also note that identifying dependencies and removing the dependencies is done in config mgmt layer.

**Why I did it**
As part Dynamic Port Breakout feature, we need to delete and re-create ports. During deletion of a port, we need to ensure that all dependencies are also removed. For example, port being part of ACL, VLAN, QoS, etc. Initial approach was to track these dependencies in orchagent itself. But after discussion with Microsoft team, we decided that we can use the dependency check that is already being done at the SAI REDIS layer. So, here we are going to check the return error code of SAI call to deduce if the port has any dependency. 

**How I verified it**
Tested Dynamic Port Breakout code with port in VLAN and ACL tables. Verified from syslog that we SAI redis call to remove/delete the port succeeds when the dependencies are removed.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1570,21 +1570,16 @@ bool PortsOrch::removePort(sai_object_id_t port_id)
 {
     SWSS_LOG_ENTER();
 
-    Port p;
-    if (getPort(port_id, p))
-    {
-        PortUpdate update = {p, false };
-        notify(SUBJECT_TYPE_PORT_CHANGE, static_cast<void *>(&update));
-    }
-
     sai_status_t status = sai_port_api->remove_port(port_id);
     if (status != SAI_STATUS_SUCCESS)
     {
-        SWSS_LOG_ERROR("Failed to remove port %" PRIx64 ", rv:%d", port_id, status);
+        if (status != SAI_STATUS_OBJECT_IN_USE)
+        {
+            SWSS_LOG_ERROR("Failed to remove port %" PRIx64 ", rv:%d", port_id, status);
+            throw runtime_error("Delete port failed");
+        }
         return false;
     }
-
-    removeAclTableGroup(p);
 
     m_portCount--;
     SWSS_LOG_NOTICE("Remove port %" PRIx64, port_id);
@@ -1645,6 +1640,8 @@ bool PortsOrch::initPort(const string &alias, const set<int> &lane_set)
 
                 PortUpdate update = { p, true };
                 notify(SUBJECT_TYPE_PORT_CHANGE, static_cast<void *>(&update));
+
+                m_portList[alias].m_init = true;
 
                 SWSS_LOG_NOTICE("Initialized port %s", alias.c_str());
             }
@@ -2264,26 +2261,42 @@ void PortsOrch::doPortTask(Consumer &consumer)
             SWSS_LOG_NOTICE("Deleting Port %s", alias.c_str());
             auto port_id = m_portList[alias].m_port_id;
             auto hif_id = m_portList[alias].m_hif_id;
+            auto bridge_port_oid = m_portList[alias].m_bridge_port_id;
 
-            if (m_portList[alias].has_dependency())
+            if (bridge_port_oid != SAI_NULL_OBJECT_ID)
             {
-                // Port has one or more dependencies, cannot remove
-                SWSS_LOG_WARN("Cannot to remove port because of dependency");
+                // Bridge port OID is set on a port as long as
+                // port is part of at-least one VLAN. 
+                // Ideally this should be tracked by SAI redis. 
+                // Until then, let this snippet be here.
+                SWSS_LOG_NOTICE("Cannot remove port as brodge port OID is present %lx", bridge_port_oid);
+                it++;
                 continue;
-            }
+            } 
 
-            deInitPort(alias, port_id);
-
-            SWSS_LOG_NOTICE("Removing hostif %lx for Port %s", hif_id, alias.c_str());
-            sai_status_t status = sai_hostif_api->remove_hostif(hif_id);
-            if (status != SAI_STATUS_SUCCESS)
+            if (m_portList[alias].m_init)
             {
-                throw runtime_error("Remove hostif for the port failed");
+                deInitPort(alias, port_id);
+                SWSS_LOG_NOTICE("Removing hostif %lx for Port %s", hif_id, alias.c_str());
+                sai_status_t status = sai_hostif_api->remove_hostif(hif_id);
+                if (status != SAI_STATUS_SUCCESS)
+                {
+                    throw runtime_error("Remove hostif for the port failed");
+                }
+                m_portList[alias].m_init = false;
+
+                Port p;
+                if (getPort(port_id, p))
+                {
+                    PortUpdate update = {p, false };
+                    notify(SUBJECT_TYPE_PORT_CHANGE, static_cast<void *>(&update));
+                }
             }
 
             if (!removePort(port_id))
             {
-                throw runtime_error("Delete port failed");
+                it++;
+                continue;
             }
             removePortFromLanesMap(alias);
             removePortFromPortListMap(port_id);
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -982,7 +982,6 @@ bool PortsOrch::unbindRemoveAclTableGroup(sai_object_id_t  port_oid,
         return true;
     }
 
-    port.clear_dependency(Port::ACL_DEP);
     SWSS_LOG_NOTICE("Removing port OID %" PRIx64" ACL table group ID", port_oid);
 
     // Unbind ACL group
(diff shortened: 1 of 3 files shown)


zhenggen-xu:
@vasant17  Please resolve the conflicts, and list this PR https://github.com/Azure/sonic-sairedis/pull/565 as dependency.

vasant17:
retest this please

orchagent/portsorch.cpp
```diff
@@ -2264,26 +2261,41 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
-            if (m_portList[alias].has_dependency())
+            if (bridge_port_oid != SAI_NULL_OBJECT_ID)
             {
-                // Port has one or more dependencies, cannot remove
-                SWSS_LOG_WARN("Cannot to remove port because of dependency");
+                // Bridge port OID is set on a port as long as
+                // port is part of at-least one VLAN. 
+                // Ideally this should be tracked by SAI redis. 
+                // Until then, let this snippet be here.
+                SWSS_LOG_NOTICE("Cannot remove port as brodge port OID is present %lx", bridge_port_oid);
```
qiluo-msft: >brodge [](start = 55, length = 6)

typo

orchagent/portsorch.cpp
```diff
@@ -2264,26 +2261,41 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
             auto hif_id = m_portList[alias].m_hif_id;
+            auto bridge_port_oid = m_portList[alias].m_bridge_port_id;
 
-            if (m_portList[alias].has_dependency())
+            if (bridge_port_oid != SAI_NULL_OBJECT_ID)
             {
-                // Port has one or more dependencies, cannot remove
-                SWSS_LOG_WARN("Cannot to remove port because of dependency");
+                // Bridge port OID is set on a port as long as
+                // port is part of at-least one VLAN.
```
qiluo-msft: >VLAN [](start = 48, length = 4)

how about vxlan?

orchagent/portsorch.cpp
```diff
@@ -2264,26 +2261,41 @@ void PortsOrch::doPortTask(Consumer &consumer)
             SWSS_LOG_NOTICE("Deleting Port %s", alias.c_str());
             auto port_id = m_portList[alias].m_port_id;
             auto hif_id = m_portList[alias].m_hif_id;
+            auto bridge_port_oid = m_portList[alias].m_bridge_port_id;
 
-            if (m_portList[alias].has_dependency())
```
qiluo-msft: >if (m_portList[alias].has_dependency()) [](start = 12, length = 39)

We can still keep this if-block. No harm to check dependency in orchagent first.

orchagent/portsorch.cpp
qiluo-msft: >SWSS_LOG_NOTICE [](start = 16, length = 15)

A warning or error?

qiluo-msft:
>             m_portList.erase(alias);

We should move this part into deInitPort(). Then no need to add m_init field.

---
Refers to: orchagent/portsorch.cpp:2304 in ca4812c. [](commit_id = ca4812c9105f60d15a85903648b4935ad514b700, deletion_comment = False)

qiluo-msft:
> bool PortsOrch::removePort(sai_object_id_t port_id)

The function behavior is weird:
1. if removed, return false
2. if in use, return false
3. if other error, throw.

Suggest return an enum, or sai_status_t. No throw.

---
Refers to: orchagent/portsorch.cpp:1569 in ca4812c. [](commit_id = ca4812c9105f60d15a85903648b4935ad514b700, deletion_comment = False)

orchagent/portsorch.cpp
```diff
@@ -1570,22 +1570,17 @@ bool PortsOrch::removePort(sai_object_id_t port_id)
...
-    }
-
     sai_status_t status = sai_port_api->remove_port(port_id);
     if (status != SAI_STATUS_SUCCESS)
     {
-        SWSS_LOG_ERROR("Failed to remove port %" PRIx64 ", rv:%d", port_id, status);
+        if (status != SAI_STATUS_OBJECT_IN_USE)
+        {
+            SWSS_LOG_ERROR("Failed to remove port %" PRIx64 ", rv:%d", port_id, status);
+            throw runtime_error("Delete port failed");
```
qiluo-msft: >throw runtime_error("Delete port failed"); [](start = 12, length = 42)

Treat it as an error, but not to crash the orchagent.

orchagent/portsorch.cpp
vasant17: Corrected

orchagent/portsorch.cpp
vasant17: Two things:
1. I did NOT want to change the existing behavior. 
2. We need to think if this can ever happen, if our theory says that it can never happen, we should throw a runtime_error.

At this point in time, if a port does NOT have any dependency, I do NOT see any reason why a removal/deletion would fail. If at all it happens, I would prefer to catch it right then and there instead of letting the system go in in-consistent state.

orchagent/portsorch.cpp
vasant17: > > bool PortsOrch::removePort(sai_object_id_t port_id)
> 
> The function behavior is weird:
> 
> 1. if removed, return true
> 2. if in use, return false
> 3. if other error, throw.
> 
> Suggest return an enum, or sai_status_t. No throw.
> 
> Refers to: orchagent/portsorch.cpp:1569 in [ca4812c](https://github.com/Azure/sonic-swss/commit/ca4812c9105f60d15a85903648b4935ad514b700). [](commit_id = [ca4812c](https://github.com/Azure/sonic-swss/commit/ca4812c9105f60d15a85903648b4935ad514b700), deletion_comment = False)

Half heartedly, I agree. I am NOT too happy because, we always handle the sai_status in the inner most function (where we call sai API's) and return only true of false to the calling function. And its pretty consistent throughout the code and I liked it. But here I am making an exception by returning sai_status to the caller. I did think about introducing our own enum, but that seemed like over kill. So, I will stick to this for now.

orchagent/portsorch.cpp
vasant17: Nope. we had a lengthy discussion on this with Guhan and Renuka. Will rely on SAI REDIS return code and do NOT maintain any code in orchagent to track the dependency. In-fact I am removing the dependency tracking code (which went in as part of ACL/DPB code) in this PR which

vasant17:
> > ```
> >         m_portList.erase(alias);
> > ```
> 
> We should move this part into deInitPort(). Then no need to add m_init field. @zhenggen-xu, please check.
> 
> Refers to: orchagent/portsorch.cpp:2304 in [ca4812c](https://github.com/Azure/sonic-swss/commit/ca4812c9105f60d15a85903648b4935ad514b700). [](commit_id = [ca4812c](https://github.com/Azure/sonic-swss/commit/ca4812c9105f60d15a85903648b4935ad514b700), deletion_comment = False)

No. That wont work.  Because, if a port is present in  m_portList, it means an object for it exists in SAI REDIS. Suppose, if you remove it in deInit and subsequent call to port removal fails due to OBJECT_IN_USE, then while removing the dependency, other modules(VLAN, ACL, etc) think the port does NOT exist(because they call getPort()) and they will never remove the dependency.

orchagent/portsorch.cpp
vasant17: Sure. Changed it to warning.

orchagent/portsorch.cpp
```diff
@@ -2264,26 +2253,48 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
             {
-                throw runtime_error("Remove hostif for the port failed");
+                deInitPort(alias, port_id);
+                SWSS_LOG_NOTICE("Removing hostif %lx for Port %s", hif_id, alias.c_str());
+                sai_status_t status = sai_hostif_api->remove_hostif(hif_id);
+                if (status != SAI_STATUS_SUCCESS)
+                {
+                    throw runtime_error("Remove hostif for the port failed");
+                }
+                m_portList[alias].m_init = false;
```
qiluo-msft: >m_portList[alias].m_init = false; [](start = 16, length = 33)

Move this line into deInitPort()?

orchagent/portsorch.cpp
vasant17: > > VLAN [](start = 48, length = 4)
> 
> how about vxlan?

For VxLAN I do NOT see any direct dependency on physical port. If your question is more about handling tunnel interface dependency on physical port, that is NOT in scope of this PR. Because, we expect every dependency to be tracked in SAI REDIS layer. Please let me know if I dint interpret your comment correctly

orchagent/portsorch.cpp
vasant17: Agreed. Done!

qiluo-msft:
Retest vs please

lguohan:
Retest vs please

lguohan:
retest this please

vasant17:
retest this please

PR #1228, opened 2020-03-24
[dvs] Add redis polling interface to dvs fixture

daall:
- Remove standalone db fixtures
- Refactor NAT and ACL tests to use new interface

**What I did**
I removed the redis polling fixtures and moved the polling DB objects into the dvs fixture.

**Why I did it**
I got some feedback that using the new fixtures everywhere was getting messy and that it would be simpler to include the databases as part of dvs. Also, it makes more logical sense for the databases to be part of the switch.

**How I verified it**
NAT and ACL tests pass locally before and after the change.

**Details if related**
There are a few `NOTES` to go back and finish refactoring a few parts of the tests - this was done to keep the PR relatively compact and reviewable. Let me know if you think any of those changes *should* be included in this PR!


tests/test_acl.py
```diff
@@ -560,9 +569,9 @@ def test_AclRuleIcmpV6(self, asic_db, config_db, state_db):
         self.remove_acl_table("test_icmpv6")
         self.verify_no_acl_tables()
 
-    def test_AclRuleRedirectToNextHop(self, dvs, asic_db, config_db, state_db):
+    def test_AclRuleRedirectToNextHop(self, dvs):
         dvs.setup_db()
```
neethajohn: typo?

tests/test_nat.py
```diff
@@ -1,17 +1,12 @@
 import time
 
 class TestNat(object):
-    def setup_db(self, app_db, asic_db, config_db):
```
neethajohn: I would still prefer to keep this one and have wrapper functions in DVS class to return the db handles instead of just accessing them directly (dvs.config_db)

tests/test_acl.py
daall: `set_interface_status` uses `dvs.cdb` under the hood, so it is still needed for now. I can add a comment to make it more clear!

tests/test_nat.py
daall: Why? I can definitely see some benefits to doing it this way but I'm curious what thoughts you have about it. :)

neethajohn: My thoughts weren't written clearly :)
I meant have setup_db(dvs) and DVS class would have methods like get_config_db() etc which setup_db(dvs) would use to initialize self.config_db and use them instead of directly accessing the class DVS vars.

daall: Done!

daall:
retest this please

tests/conftest.py
```diff
@@ -235,20 +235,13 @@ def __init__(self, name=None, imgname=None, keeptb=False, fakeplatform=None):
...
-        self.flex_db = dvs_db.DVSDatabase(self.FLEX_COUNTER_DB_ID, self.redis_sock)
-        self.state_db = dvs_db.DVSDatabase(self.STATE_DB_ID, self.redis_sock)
-
-        # NOTE: In order to limit the scope of this PR only the newly ported
-        # tests are using these variables. I will be refactoring the tests to
-        # all use these variables in a follow-up PR.
-        self.default_acl_tables = self.asicdb.default_acl_tables
-        self.default_acl_entries = self.asicdb.default_acl_entries
-        self.port_name_map = self.asicdb.portnamemap
+        # DB wrappers are declared here, lazy-loaded in the tests
```
neethajohn: I am not sure if there would be some runtime hit due to the lazy loading. if yes, you could still initialize all of them here and just return the handles in the get_*_db wrapper functions

daall: I don't think there's really a big difference either way.

The main reason I opted to lazy-load the DBs is that a lot of the tests are only using a couple of them anyways, so there's no need to create all 6 every time. I think ideally we'll move towards setting up the needed dbs in some sort of module-level fixture or something so that the author doesn't have to remember to include `setup_db` for each test case.

PR #1238, opened 2020-04-04
Bulk routes and next hop group members

qiluo-msft:


qiluo-msft:
Retest vs please

qiluo-msft:
Retest this please

qiluo-msft:
Retest this please

orchagent/routeorch.h
```diff
@@ -54,6 +55,28 @@ struct NextHopObserverEntry
...
+    std::deque<sai_status_t>            object_statuses;    // Bulk statuses
+    NextHopGroupKey                     tmp_next_hop;       // Temporary next hop
+    sai_object_id_t                     vrf_id;
+    IpPrefix                            ip_prefix;
+
+    RouteBulkContext()
+    {
+    }
+
+    // Disable any copy constructors
```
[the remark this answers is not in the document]
qiluo-msft: No need to disable by now. My purpose here is to prevent mistakenly using non-reference in parameter list, because the main usage of this class is for in/out parameters.

There is no harm to assign. I don't have use case to support it or against it right now.

---
In reply to: [408269315](https://github.com/Azure/sonic-swss/pull/1238#discussion_r408269315) [](ancestors = 408269315)

orchagent/bulker.h
```diff
@@ -0,0 +1,656 @@
...
+template <typename T>
+class EntityBulker
+{
+public:
+    using Ts = saitraits<T>;
+    using Te = typename Ts::entry_t;
+
+    EntityBulker(typename Ts::api_t *api)
+    {
+        throw std::logic_error("Not implemented");
```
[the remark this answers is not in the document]
qiluo-msft: The purpose is to enforce the template user to implement explicit ctor specialization, and do not use the general one.

This is one choice to achieve this purpose and I think it is readable.

---
In reply to: [408396144](https://github.com/Azure/sonic-swss/pull/1238#discussion_r408396144) [](ancestors = 408396144)

orchagent/bulker.h
```diff
@@ -0,0 +1,656 @@
...
+    }
+
+    sai_status_t create_entry(
```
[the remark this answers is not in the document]
qiluo-msft: There is no `create` at all. Are you talking about `create_entries`? It is a function type.

And this is a function and not overridden/overload/specialized by others, and used by main logic.

---
In reply to: [408397254](https://github.com/Azure/sonic-swss/pull/1238#discussion_r408397254) [](ancestors = 408397254)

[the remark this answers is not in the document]
daall: Oh! I see the difference now. Thank you for the clarification. :)

orchagent/bulker.h
```diff
@@ -0,0 +1,656 @@
...
+        _Out_ sai_status_t *object_statuses);
+typedef sai_status_t (*sai_bulk_set_fdb_entry_attribute_fn)(
+        _In_ uint32_t object_count,
+        _In_ const sai_fdb_entry_t *fdb_entry,
+        _In_ const sai_attribute_t *attr_list,
+        _In_ sai_bulk_op_error_mode_t mode,
+        _Out_ sai_status_t *object_statuses);
+
+template<typename T>
+struct saitraits { };
```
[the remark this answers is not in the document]
qiluo-msft: So vague that nobody want to reuse it by design. The exact name may be SaiApiFunctionTypesAndFutureMiscUsedByBulker. Open to recommendation.

---
In reply to: [408392024](https://github.com/Azure/sonic-swss/pull/1238#discussion_r408392024) [](ancestors = 408392024)

orchagent/routeorch.cpp
```diff
@@ -373,187 +376,318 @@ void RouteOrch::doTask(Consumer& consumer)
...
+            }
+            else
+            {
+                SWSS_LOG_ERROR("Unknown operation type %s\n", op.c_str());
+                it = consumer.m_toSync.erase(it);
+            }
+        }
+
+        // Flush the route bulker, so routes will be written to syncd and ASIC
+        gRouteBulker.flush();
```
dzhangalibaba, line 591: two questions:
1 when to flush, it depends on how many entries in consumer.m_toSync ?
2 I saw in flush , the order is removing_entries - > creating_entries -> setting_entries, how do you make sure there is no dependence among them.

e.g , for non-bulk case, the order is : add add del set del set del add add, if bulking, it is {add add add add} {set set} {del del del} , need to make sure if any add/set/del must be before others. We need the experiment data.

qiluo-msft: 1 when to flush, it depends on how many entries in consumer.m_toSync ?

Currently there are two trigger points.
- `doTask()` will go through all the `m_toSync` items, after that will trigger one router bulker flush. This is to simplify implementation so no need to remember too much history states
- If there is not more nexthopgroup, and there is some remove_entry queued in the route bulker, will trigger one flush in hope of collecting some nexthopgroup.

2. how do you make sure there is no dependence among them
This is a design choice and after it is fixed, I explicitly merge operation during `create_entry` `remove_entry` and `set_entry_attribute` according this is the sequence.



---
In reply to: [408579209](https://github.com/Azure/sonic-swss/pull/1238#discussion_r408579209) [](ancestors = 408579209)

qiluo-msft:
Retest this please

qiluo-msft:
Retest this please

orchagent/bulker.h
```diff
@@ -0,0 +1,656 @@
...
+#include <vector>
+#include <unordered_map>
+#include <unordered_set>
+#include <stdexcept>
+#include <boost/functional/hash.hpp>
+#include <sairedis.h>
+#include "sai.h"
+#include "logger.h"
+
+static inline bool operator==(const sai_ip_prefix_t& a, const sai_ip_prefix_t& b)
```
kcudnik, line 13: could you not put code in header files ? i guess since inline, then you are trying to gain speed here ? probably not much gain, i think would be better if you put this code in cpp for better readability

orchagent/routeorch.cpp
```diff
@@ -718,6 +852,11 @@ bool RouteOrch::addNextHopGroup(const NextHopGroupKey &nexthops)
             return false;
         }
 
+        // skip next hop group member create for neighbor from down port
+        if (m_neighOrch->isNextHopFlagSet(it, NHFLAGS_IFDOWN)) {
```
prsunny, line 856: move` {` to next line

orchagent/routeorch.cpp
```diff
@@ -997,8 +1154,94 @@ bool RouteOrch::addRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
...
+    }
+    /* The route is pointing to a next hop group */
+    else
+    {
+        if (!hasNextHopGroup(nextHops))
+        {
+            // Previous added an temporary route
+            auto& tmp_next_hop = ctx.tmp_next_hop;
+            addRoutePost(ctx, vrf_id, ipPrefix, tmp_next_hop);
+            return false;
```
prsunny, line 1236: This statement seems redundant as it is recursive.

orchagent/routeorch.cpp
```diff
@@ -997,8 +1154,94 @@ bool RouteOrch::addRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
...
+        NextHopKey nexthop(nextHops.to_string());
+        if (nexthop.ip_address.isZero())
+        {
+            next_hop_id = m_intfsOrch->getRouterIntfsId(nexthop.alias);
+            /* rif is not created yet */
+            if (next_hop_id == SAI_NULL_OBJECT_ID)
+            {
+                SWSS_LOG_INFO("Failed to get next hop %s for %s",
+                        nextHops.to_string().c_str(), ipPrefix.to_string().c_str());
+                return false;
```
prsunny, line 1215: Do you intend to retry in post action or its simply the same behavior as existing?

orchagent/routeorch.cpp
```diff
@@ -1085,18 +1321,21 @@ bool RouteOrch::removeRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix)
...
 
+    sai_route_entry_t route_entry;
+    route_entry.vr_id = vrf_id;
+    route_entry.switch_id = gSwitchId;
+    copy(route_entry.destination, ipPrefix);
+
     auto it_route = it_route_table->second.find(ipPrefix);
-    if (it_route == it_route_table->second.end())
+    size_t creating = gRouteBulker.creating_entries_count(route_entry);
+    if (it_route == it_route_table->second.end() && creating == 0)
```
prsunny, line 1331: So I thought you would want to remove this entry from Bulker if there is a remove Route and `creating `is `True`

orchagent/routeorch.h
```diff
@@ -54,6 +55,28 @@ struct NextHopObserverEntry
     list<Observer *> observers;
 };
 
+struct RouteBulkContext
+{
+    std::deque<sai_status_t>            object_statuses;    // Bulk statuses
+    NextHopGroupKey                     tmp_next_hop;       // Temporary next hop
```
prsunny, line 61: `tmp_next_hop `is only to handle the `addTempRoute `case, right?

qiluo-msft: Yes, any concern?

---
In reply to: [410517262](https://github.com/Azure/sonic-swss/pull/1238#discussion_r410517262) [](ancestors = 410517262)

orchagent/routeorch.cpp
```diff
@@ -373,187 +376,318 @@ void RouteOrch::doTask(Consumer& consumer)
...
-         */
-        if (key == "resync")
+        // Route bulk results will be stored in a map
+        std::map<
+                std::pair<
+                        std::string,            // Key
+                        std::string             // Op
+                >,
+                RouteBulkContext
+        >                                       toBulk;
```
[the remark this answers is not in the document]
qiluo-msft: I made the alignment on purpose. Where do you hate?

---
In reply to: [410425268](https://github.com/Azure/sonic-swss/pull/1238#discussion_r410425268) [](ancestors = 410425268)

orchagent/routeorch.cpp
qiluo-msft: Yes it is recursive.
But it is necessary. Remove it will change behavior.

---
In reply to: [410513353](https://github.com/Azure/sonic-swss/pull/1238#discussion_r410513353) [](ancestors = 410513353)

orchagent/routeorch.cpp
qiluo-msft: It the XxxPost() return false, the corresponding element in `consumer.m_toSync` will be kept, and later doTask() will retry. This behavior is same as existing.

---
In reply to: [410513820](https://github.com/Azure/sonic-swss/pull/1238#discussion_r410513820) [](ancestors = 410513820)

orchagent/routeorch.cpp
qiluo-msft: > if there is a remove Route and creating is True

In this case, this if-block will be skipped, and it may be removed from Bulker if it happen to be there. However upstream merging in consumer.m_toSync will make this SET->DEL impossible to happen.

Did you see a bug?

---
In reply to: [410516175](https://github.com/Azure/sonic-swss/pull/1238#discussion_r410516175) [](ancestors = 410516175)

orchagent/routeorch.cpp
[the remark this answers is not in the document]
prsunny: From the appearence, it didn't appear to be intentional. Would be good to keep the declaration same like [others ](https://github.com/Azure/sonic-swss/blob/2797e760597b9ea705046aa64f890e65aa16da53/orchagent/routeorch.cpp#L831)in file.

orchagent/routeorch.cpp
prsunny: Ok

orchagent/routeorch.h
prsunny: Not really, it was for just clarification

orchagent/routeorch.cpp
```diff
@@ -373,187 +376,318 @@ void RouteOrch::doTask(Consumer& consumer)
...
 
-                    if (j.first != gVirtualRouterId)
+            /* Get notification from application */
+            /* resync application:
+             * When routeorch receives 'resync' message, it marks all current
+             * routes as dirty and waits for 'resync complete' message. For all
+             * newly received routes, if they match current dirty routes, it unmarks
+             * them dirty. After receiving 'resync complete' message, it creates all
+             * newly added routes and removes all dirty routes.
+             */
```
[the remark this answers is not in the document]
qiluo-msft: Actually this is old code. Since the nature of this PR, I need to minimize the changes in favor or easy reviewing. We could focus on refactor later.

---
In reply to: [408379175](https://github.com/Azure/sonic-swss/pull/1238#discussion_r408379175) [](ancestors = 408379175)

orchagent/routeorch.cpp
```diff
@@ -373,187 +376,318 @@ void RouteOrch::doTask(Consumer& consumer)
...
-                SWSS_LOG_NOTICE("Complete resync routes\n");
-                m_resync = false;
+                vrf_id = gVirtualRouterId;
+                ip_prefix = IpPrefix(key);
             }
 
-            it = consumer.m_toSync.erase(it);
-            continue;
-        }
+            if (op == SET_COMMAND)
```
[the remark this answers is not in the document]
qiluo-msft: Same as previous comment.

---
In reply to: [408380613](https://github.com/Azure/sonic-swss/pull/1238#discussion_r408380613) [](ancestors = 408380613)

orchagent/routeorch.cpp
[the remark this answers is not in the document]
daall: Ah, okay! I wasn't aware of the history here. In that case, I agree with you that it's probably best to save the refactoring for later. :)

orchagent/bulker.h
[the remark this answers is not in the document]
daall: SaiBulkerTraits, maybe? Still vague but it's at least clear that it's specific to the bulker.

orchagent/bulker.h
daall: @qiluo-msft I agree with Kamil here. I see that there are a lot of template functions in this file, but have you considered doing something like this: https://isocpp.org/wiki/faq/templates#separate-template-fn-defn-from-decl

qiluo-msft: @daall There is no template function, only template struct or classes.

---
In reply to: [414741883](https://github.com/Azure/sonic-swss/pull/1238#discussion_r414741883) [](ancestors = 414741883)

qiluo-msft:
Retest this please

orchagent/routeorch.cpp
```diff
@@ -1027,8 +1163,97 @@ bool RouteOrch::addRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
         route_attr.value.oid = next_hop_id;
 
         /* Default SAI_ROUTE_ATTR_PACKET_ACTION is SAI_PACKET_ACTION_FORWARD */
-        sai_status_t status = sai_route_api->create_route_entry(&route_entry, 1, &route_attr);
-        if (status != SAI_STATUS_SUCCESS)
+        object_statuses.emplace_back();
+        sai_status_t status = gRouteBulker.create_entry(&object_statuses.back(), &route_entry, 1, &route_attr);
+        if (status == SAI_STATUS_ITEM_ALREADY_EXISTS)
+        {
+            SWSS_LOG_ERROR("Failed to create route %s with next hop(s) %s: already exists in bulker",
```
daall, line 1170: Is this really an ERROR case? I don't have an issue with the logic, but if this is something that could come up commonly in practice then I don't know if we should be printing errors to the syslog.

orchagent/routeorch.cpp
```diff
@@ -1027,8 +1163,97 @@ bool RouteOrch::addRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
...
+    else
+    {
+        /* Set the packet action to forward when there was no next hop (dropped) */
+        if (it_route->second.getSize() == 0)
+        {
+            route_attr.id = SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION;
+            route_attr.value.s32 = SAI_PACKET_ACTION_FORWARD;
+
+            object_statuses.emplace_back();
+            gRouteBulker.set_entry_attribute(&object_statuses.back(), &route_entry, &route_attr);
```
daall, line 1184: Is there supposed to be an `emplace_back` in here somewhere similar to the other if branch?

orchagent/routeorch.cpp
qiluo-msft: It is a logic error. `m_toSync` will not have duplicated SET routes because of its design.

---
In reply to: [417493477](https://github.com/Azure/sonic-swss/pull/1238#discussion_r417493477) [](ancestors = 417493477)

orchagent/routeorch.cpp
qiluo-msft: Seems I already have.

---
In reply to: [417495448](https://github.com/Azure/sonic-swss/pull/1238#discussion_r417495448) [](ancestors = 417495448)

lguohan:
i cannot find any unit test added for this feature. @qiluo-msft , what is your plan?

qiluo-msft:
This is a false alert.

---
In reply to: [622630868](https://github.com/Azure/sonic-swss/pull/1238#issuecomment-622630868) [](ancestors = 622630868)

tests/test_nhg.py
```diff
@@ -138,3 +139,194 @@ def test_route_nhg(self, dvs, testlog):
...
+        last_ipprefix = route_ipprefix
+
+        time.sleep(1)
+
+        # Check ASIC DB on the count of nexthop groups used, and it should not increase
+        nhg_count = len(nhgtbl.getKeys())
+        assert nhg_count == MAX_ECMP_COUNT
+
+        # Check the route points to next hop group
+        k = asic_route_exists("2.2.2.0/24")
```
lguohan, line 293: @daall , is this the right api to use to do the assert?

daall: Nope, @qiluo-msft you'll want to take a look at the methods in `tests/dvs_database.py` and use those to get the keys/entries you need. You can fetch the asic DB/other DBs that you need with `get_asic_db()` and similar methods in the DockerVirtualSwitch in `conftest.py`. Let me know if you have any questions!

tests/test_nhg.py
```diff
@@ -138,3 +139,194 @@ def test_route_nhg(self, dvs, testlog):
...
+            route_ipprefix = gen_ipprefix(base + route_count)
+            ps.set(route_ipprefix, fvs)
+            route_count += 1
+        last_ipprefix = route_ipprefix
+
+        time.sleep(1)
+
+        # Check ASIC DB on the count of nexthop groups used, and it should not increase
+        nhg_count = len(nhgtbl.getKeys())
+        assert nhg_count == MAX_ECMP_COUNT
```
daall, line 290: For example, this should be something like:
```
asic_db.wait_for_n_keys(NEXT_HOP_TABLE, MAX_ECMP_COUNT)
```

tests/test_nhg.py
```diff
@@ -138,3 +139,194 @@ def test_route_nhg(self, dvs, testlog):
...
+
+        def asic_route_exists(ipprefix):
+            keys = rtbl.getKeys()
+            for k in keys:
+                rt_key = json.loads(k)
+
+                if rt_key['dest'] == ipprefix:
+                    return k
+            else:
+                return None
```
daall, line 203: For this you might need to add a custom wait method, there's an example of this in `test_subport_intf`.

tests/test_nhg.py
qiluo-msft: This case should be ok, I already used `wait_for_n_keys` above so everything is available at this time.

---
In reply to: [419621718](https://github.com/Azure/sonic-swss/pull/1238#discussion_r419621718) [](ancestors = 419621718)

tests/test_nhg.py
qiluo-msft: The same reason.

---
In reply to: [419622941](https://github.com/Azure/sonic-swss/pull/1238#discussion_r419622941) [](ancestors = 419622941)

tests/test_nhg.py
qiluo-msft: I follow the new convention on above code, no need to change here since everything is available this time.

---
In reply to: [419608002](https://github.com/Azure/sonic-swss/pull/1238#discussion_r419608002) [](ancestors = 419608002)

daall: This should work! I would prefer if we stuck to the new interface (e.g. using `get` and `set` on the dvs db object) instead of using a mix of the new interface and the swsscommon interface, but I don't think any additional waits are needed.

tests/test_nhg.py
```diff
@@ -138,3 +139,187 @@ def test_route_nhg(self, dvs, testlog):
...
+        MAX_PORT_COUNT = 10
+        IP_INTEGER_BASE = int(ipaddress.IPv4Address(unicode("2.2.2.0")))
+
+        config_db = dvs.get_config_db()
+        intf_tbl = swsscommon.Table(config_db.db_connection, "INTERFACE")
+        fvs = swsscommon.FieldValuePairs([("NULL", "NULL")])
+
+        for i in range(MAX_PORT_COUNT):
+            intf_tbl.set(port_name(i), fvs)
+            intf_tbl.set("{}|{}".format(port_name(i), port_ipprefix(i)), fvs)
```
daall, line 232: For the `set` methods you can use `create_entry` in config_db. Do note that it expects a dict rather than a list of FieldValuePairs so you will need to update a few other parts of the code.

tests/test_nhg.py
```diff
@@ -138,3 +139,187 @@ def test_route_nhg(self, dvs, testlog):
...
+            ps.set(route_ipprefix, fvs)
+            route_count += 1
+
+        asic_db = dvs.get_asic_db()
+        rtbl = swsscommon.Table(asic_db.db_connection, "ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY")
+        nhgtbl = swsscommon.Table(asic_db.db_connection, "ASIC_STATE:SAI_OBJECT_TYPE_NEXT_HOP_GROUP")
+
+        # Wait and check ASIC DB the count of nexthop groups used
+        asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_NEXT_HOP_GROUP", MAX_ECMP_COUNT)
+        asic_routes_count = len(rtbl.getKeys())
```
daall, line 262: For the `getKeys` methods you can use the `get_keys` method in asic_db.

tests/test_nhg.py
```diff
@@ -138,3 +139,187 @@ def test_route_nhg(self, dvs, testlog):
...
+            for k in keys:
+                rt_key = json.loads(k)
+
+                if rt_key['dest'] == ipprefix:
+                    return k
+            else:
+                return None
+
+        def asic_route_nhg_fvs(k):
+            (status, fvs) = rtbl.get(k)
```
daall, line 205: There's a `getEntry` method in dvs database that will handle this. The main differences are that it doesn't return a status and the return type is a dict, so you'll need to update a few other parts of the code as well.

tests/test_nhg.py
```diff
@@ -138,3 +139,187 @@ def test_route_nhg(self, dvs, testlog):
...
+        r = 0
+        while route_count < MAX_ECMP_COUNT:
+            r += 1
+            fmt = '{{0:0{}b}}'.format(MAX_PORT_COUNT)
+            binary = fmt.format(r)
+            # We need at least 2 ports for a nexthop group
+            if binary.count('1') <= 1:
+                continue
+            route_ipprefix = gen_ipprefix(route_count)
+            ps._del(route_ipprefix)
```
daall, line 311: dvs database also has a `delete_entry` method you can use.

tests/test_nhg.py
qiluo-msft: I did not use any db operation here.

---
In reply to: [420330242](https://github.com/Azure/sonic-swss/pull/1238#discussion_r420330242) [](ancestors = 420330242)

tests/test_nhg.py
qiluo-msft: ProducerStateTable is not supported, only Table.

---
In reply to: [420339845](https://github.com/Azure/sonic-swss/pull/1238#discussion_r420339845) [](ancestors = 420339845)

qiluo-msft:
Retest this please

qiluo-msft:
Retest this please

daall:
retest this please

PR #1245, opened 2020-04-07
Refine getDbId() calling to fix build after swss-common change

qiluo-msft:
Fix build after https://github.com/Azure/sonic-swss-common/pull/336/files


qiluo-msft:
@dzhangalibaba please help review

PR #1249, opened 2020-04-08
[dvs] Re-enable RIF tests

daall:

**What I did**
Re-enabled most of the RIF tests.

**Why I did it**
Since we've narrowed down the issue to setting the MTU, we can skip the MTU tests while we implement the fix so that we can at least continue to verify the rest of the RIF behavior.

**How I verified it**
Local test runs

**Details if related**

### DIFF
--- a/tests/test_interface.py
+++ b/tests/test_interface.py
@@ -4,7 +4,6 @@ import pytest
 
 from swsscommon import swsscommon
 
-@pytest.mark.xfail(reason="vs image issue: Azure/sonic-sairedis#574")
 class TestRouterInterface(object):
     def setup_db(self, dvs):
         self.pdb = swsscommon.DBConnector(0, dvs.redis_sock, 0)
@@ -307,6 +306,7 @@ class TestRouterInterface(object):
             if route["dest"] == "10.0.0.4/32":
                 assert False
 
+    @pytest.mark.skip(reason="vs image issue: Azure/sonic-sairedis#574")
     def test_PortInterfaceSetMtu(self, dvs, testlog):
         self.setup_db(dvs)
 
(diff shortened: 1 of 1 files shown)


lguohan:
https://github.com/Azure/sonic-sairedis/issues/574 is fixed.

lguohan:
I have been trying to fight this issue for quite a while. it is was first noticed in this pr. https://github.com/Azure/sonic-buildimage/pull/4342 judy has spend lots of efforts to track this down. And now I noticed that you have found this issue three weeks ago.

lguohan:
I think we need to enable the swss vs tests for sairedis repo. or at least for every sairedis commit, we need to build and build a new swss vs docker and run the test so that we can find the regression.

daall:
> I have been trying to fight this issue for quite a while. it is was first noticed in this pr. [Azure/sonic-buildimage#4342](https://github.com/Azure/sonic-buildimage/pull/4342) judy has spend lots of efforts to track this down. And now I noticed that you have found this issue three weeks ago.

I'll grab the latest changes and take a look. Now that you point it out it looks like it's consistently xfailing on `test_LagInterfaceSetMtu` instead of `test_PortInterfaceSetMtu` now, so it's possible we need to make a similar change for portchannel interfaces as well?

daall:
> I think we need to enable the swss vs tests for sairedis repo. or at least for every sairedis commit, we need to build and build a new swss vs docker and run the test so that we can find the regression.

Agreed.

lguohan:
I think both tests should be fixed.

daall:
I fetched the build from Azure/sonic-buildimage#4379 and the normal port is consistently passing locally but I'm still hitting the crash with the LAG test. I need to check the email thread with @kcudnik to see if it's the same failure path as before or if this is a different one.

PR #1252, opened 2020-04-09
[sflow] Fix ansible test problems caused by enabling sflow

fredyu190011:
**What I did**
To fix the issue that adding sflow packet-sample rule will cause COPP write error log repeatedly.

**Why I did it**
After adding the sflow packet-sample rule in COPP config, COPP will repeatedly write error log when user changes the config each time. Until sflow enable. The error log will cause loganalyzer of pytest-ansible raise the error then all test will be failed. By SAI definition and sflow HLD, the rule (SAI_HOSTIF_TRAP_TYPE_SAMPLEPACKET) used to create genlink for hsflowd not used to trap packet to CPU. So it can be directly set without waiting sflow enable. (By current code logic it doesn't consider to undo when sflow changed from enable to disable. It shall be not correct.)

**How I verified it**
Add the rule in copp json file.
* When sflow is disable, confirm no any packet will be trap to CPU.
* Confirmed no error log when any config changed.
* Run COPP ansible and result is passed.

**Details if related**


prsunny:
This affects sflow functionality. We need a proper fix, @dgsudharsan

fredyu190011:
Hi @prsunny @dgsudharsan, could you please help to review this PR?

dgsudharsan:
> Hi @prsunny @dgsudharsan, could you please help to review this PR?

Hi,
As per discussion with prince i have raised the following pull request to address the problem
https://github.com/Azure/sonic-swss/pull/1256

I was in parallel working on the issue when someone pointed out this PR. however since this would break the functionality i have raised the changes which would not impact the functionality.

fredyu190011:
Hi @dgsudharsan , in #1256, it only remove the code cause log ERR message. But the rule will be still in copp consumer until enable sflow. I think that will be a problem in warm-reboot check. Why not set the rule directly without waiting user enables sflow?

fredyu190011:
Hi @dgsudharsan, is there any progress on it?

dgsudharsan:
> Hi @dgsudharsan, is there any progress on it?

The fix requires a complete redesign of copporch. The design for copporch has been raised by @prsunny . I am not sure about the timelines for the activity.

prsunny:
@fredyu190011 , can you review https://github.com/Azure/SONiC/pull/606

fredyu190011:
Hi @prsunny , so the sflow problem will be reviewed by waiting until copp arch changed?

PR #1256, opened 2020-04-11
Enabling Sflow trap group in CoPP

dgsudharsan:
**What I did**
Enabling SFLOW trap group in COPP json file.

**Why I did it**
Without the trap group being installed SFLOW will not work on hardware platforms. The bug https://github.com/Azure/sonic-buildimage/issues/4386 was raised for the issue.

**How I verified it**
Added sflow trap in file and verified it gets installed only when 'config sflow enable' command is given.

**Details if related**
Apart from including sflow trap in the file, the PR contains fix for an iterator issue in copporch. Without the fix the trap groups after sflow trap will not get installed.

Removed the log for task_needed_retry case. This log will flood the logging file and even when set to debug level it would hinder analyzing other debug logs.

### DIFF
--- a/orchagent/copporch.cpp
+++ b/orchagent/copporch.cpp
@@ -451,14 +451,13 @@ bool CoppOrch::removeGenetlinkHostIf(string trap_group_name)
     return true;
 }
 
-task_process_status CoppOrch::processCoppRule(Consumer& consumer)
+task_process_status CoppOrch::processCoppRule(Consumer& consumer, SyncMap::iterator it)
 {
     SWSS_LOG_ENTER();
 
     sai_status_t sai_status;
     vector<string> trap_id_list;
     string queue_ind;
-    auto it = consumer.m_toSync.begin();
     KeyOpFieldsValuesTuple tuple = it->second;
     string trap_group_name = kfvKey(tuple);
     string op = kfvOp(tuple);
@@ -832,7 +831,7 @@ void CoppOrch::doTask(Consumer &consumer)
 
         try
         {
-            task_status = processCoppRule(consumer);
+            task_status = processCoppRule(consumer, it);
         }
         catch(const out_of_range& e)
         {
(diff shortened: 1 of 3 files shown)


prsunny:
@dgsudharsan , can you address this comment from https://github.com/Azure/sonic-swss/pull/1252 as this is impacting warmboot scenario?

"Hi @dgsudharsan , in #1256, it only remove the code cause log ERR message. But the rule will be still in copp consumer until enable sflow. I think that will be a problem in warm-reboot check. Why not set the rule directly without waiting user enables sflow?"

dgsudharsan:
> @dgsudharsan , can you address this comment from #1252 as this is impacting warmboot scenario?
> 
> "Hi @dgsudharsan , in #1256, it only remove the code cause log ERR message. But the rule will be still in copp consumer until enable sflow. I think that will be a problem in warm-reboot check. Why not set the rule directly without waiting user enables sflow?"

Sure @prsunny . I will check on this.

orchagent/copporch.cpp
```diff
@@ -451,14 +451,13 @@ bool CoppOrch::removeGenetlinkHostIf(string trap_group_name)
     return true;
 }
 
-task_process_status CoppOrch::processCoppRule(Consumer& consumer)
+task_process_status CoppOrch::processCoppRule(Consumer& consumer, SyncMap::iterator it)
```
zhenggen-xu, line 454: It seems this function only need the "tuple" as input, why we don't change it to:
task_process_status CoppOrch::processCoppRule(KeyOpFieldsValuesTuple tuple)

to make it clearer?

swssconfig/sample/00-copp.config.json
```diff
@@ -55,5 +55,21 @@
             "red_action":"drop"
         },
         "OP": "SET"
-    }
+    },
+    {
+        "COPP_TABLE:trap.group.sflow": {
+            "trap_ids": "sample_packet",
+            "trap_action":"trap",
+            "trap_priority":"1",
```
zhenggen-xu, line 63: This is taking the same priority as the ip2me?  I would think we should make it lower than that,  so maybe raise ip2me to "2", and keep this to "1".  to keep consistent with the queue, we can make ip2me to queue 2, and sflow to queue 1.

PR #1260, opened 2020-04-16
[dvs] Update tests to use correct command to bring up interfaces

daall:

**What I did**
Changed the tests over from the Linux command to the SONiC config command.

**Why I did it**
To fix the PR tests.

**How I verified it**
Tests pass after the change locally.

**Details if related**

### DIFF
--- a/tests/test_crm.py
+++ b/tests/test_crm.py
@@ -138,7 +138,7 @@ class TestCrm(object):
         fvs = swsscommon.FieldValuePairs([("NULL","NULL")])
         intf_tbl.set("Ethernet0", fvs)
         intf_tbl.set("Ethernet0|10.0.0.0/31", fvs)
-        dvs.runcmd("ifconfig Ethernet0 up")
+        dvs.runcmd("config interface startup Ethernet0")
 
         dvs.runcmd("crm config polling interval 1")
 
@@ -205,7 +205,7 @@ class TestCrm(object):
         fvs = swsscommon.FieldValuePairs([("NULL","NULL")])
         intf_tbl.set("Ethernet0", fvs)
         intf_tbl.set("Ethernet0|fc00::1/126", fvs)
-        dvs.runcmd("ifconfig Ethernet0 up")
+        dvs.runcmd("config interface startup Ethernet0")
 
         dvs.servers[0].runcmd("ifconfig eth0 inet6 add fc00::2/126")
         dvs.servers[0].runcmd("ip -6 route add default via fc00::1")
@@ -271,7 +271,7 @@ class TestCrm(object):
         fvs = swsscommon.FieldValuePairs([("NULL","NULL")])
         intf_tbl.set("Ethernet0|10.0.0.0/31", fvs)
         intf_tbl.set("Ethernet0", fvs)
-        dvs.runcmd("ifconfig Ethernet0 up")
+        dvs.runcmd("config interface startup Ethernet0")
 
         dvs.runcmd("crm config polling interval 1")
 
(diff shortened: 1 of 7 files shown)


tests/test_fdb_update.py
```diff
@@ -198,9 +198,9 @@ def test_FDBLearnedAndUpdated(self, dvs, testlog):
         dvs.set_interface_status("Ethernet72", "up")
 
         dvs.servers[16].runcmd("ifconfig eth0 hw ether 00:00:00:00:00:11")
-        dvs.servers[16].runcmd("ifconfig eth0 6.6.6.6/24 up")
+        dvs.servers[16].runcmd("config interface startup eth0 6.6.6.6/24")
         dvs.servers[16].runcmd("ip route add default via 6.6.6.1")
-        dvs.servers[17].runcmd("ifconfig eth0 6.6.6.7/24 up")
+        dvs.servers[17].runcmd("config interface startup eth0 6.6.6.7/24")
```
lguohan, line 203: this is server. should not change.

tests/test_port.py
```diff
@@ -27,8 +27,8 @@ def test_PortMtu(self, dvs, testlog):
 
     def test_PortNotification(self, dvs, testlog):
 
-        dvs.runcmd("ifconfig Ethernet0 10.0.0.0/31 up") == 0
-        dvs.runcmd("ifconfig Ethernet4 10.0.0.2/31 up") == 0
+        dvs.runcmd("config interface startup Ethernet0 10.0.0.0/31") == 0
+        dvs.runcmd("config interface startup Ethernet4 10.0.0.2/31") == 0
```
lguohan, line 31: please separate into two commands. config interface startup only takes the interface name. 

use ```config interface ip add Ethernet0 10.0.0.0/31```

tests/test_fdb.py
```diff
@@ -97,22 +97,22 @@ def test_FdbWarmRestartNotifications(self, dvs, testlog):
...
         dvs.servers[19].runcmd("vconfig add eth0 8")
-        dvs.servers[19].runcmd("ifconfig eth0.8 8.8.8.7/24 up")
+        dvs.servers[19].runcmd("config interface startup eth0.8 8.8.8.7/24")
         dvs.servers[19].runcmd("ip route add default via 8.8.8.1")
 
-        dvs.servers[18].runcmd("ifconfig eth0 7.7.7.6/24 up")
+        dvs.servers[18].runcmd("config interface startup eth0 7.7.7.6/24")
         dvs.servers[18].runcmd("ip route add default via 7.7.7.1")
-        dvs.servers[19].runcmd("ifconfig eth0 7.7.7.7/24 up")
+        dvs.servers[19].runcmd("config interface startup eth0 7.7.7.7/24")
```
lguohan, line 115: these are servers, do not change.

tests/test_crm.py
```diff
@@ -138,7 +138,7 @@ def test_CrmIpv4Route(self, dvs, testlog):
         fvs = swsscommon.FieldValuePairs([("NULL","NULL")])
         intf_tbl.set("Ethernet0", fvs)
         intf_tbl.set("Ethernet0|10.0.0.0/31", fvs)
-        dvs.runcmd("ifconfig Ethernet0 up")
```
qiluo-msft, line 141: >ifconfig [](start = 20, length = 8)

Why `ifconfig` suddenly stop working?

lguohan: it hasn't been working for a long time. a sairedis bug hide this issue.

tests/test_port.py
daall: Good catch, thank you!

PR #1261, opened 2020-04-16
 Do not set PG to Buffer porfile mapping again if already exist.

abdosi:
**What I did**
Do not set PG to Buffer profile mapping again if already exist.
This will avoid not to call SAI API
set_ingress_priority_group_attribute() if mapping already exist

**Why I did it**
Calling SAI API set_ingress_priority_group_attribute()  Multiples times result in Buffer getting reduced on Dell-6000

**How I verified it**

a) Load Minigraph
b) Config save
c) config reload

Made sure set of  PG to Buffer profile mapping is not done more than one time via syslog and checking buffer limit in SAI.


lguohan:
retest this please

PR #1262, opened 2020-04-19
Fix obvious syntax errors

qiluo-msft:


qiluo-msft:
@stepanblyschak Please help review

PR #1263, opened 2020-04-20
[routeorch] Handle the empty "nexthop" field for backward compatibility

tylerlinp:
**What I did**
Add handling route entries with abnormal field in RouteOrch.
a. empty "ifname", skip it and do nothing
b. empty "nexthop", fill ips with zeros to match ifnames
c. ip number < interface number, like b, append zeros to ips to match ifnames
d. ip number > interface number,  truncate ips to match ifnames
Ussually none of above would happen, while warmboot from old version b will happen, others only when writing app_db manully.

**Why I did it**
To fix the issue [sonic-buildimage#4399](https://github.com/Azure/sonic-buildimage/issues/4399)

**How I verified it**
test varies of route entries & vs test

**Details if related**

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -475,11 +475,13 @@ void RouteOrch::doTask(Consumer& consumer)
              * as tokenize(",", ',') will miss the last empty segment. */
             if (alsv.size() == 0)
             {
+                SWSS_LOG_WARN("Skip the route %s, for it has an empty ifname field.", key.c_str());
                 it = consumer.m_toSync.erase(it);
                 continue;
             }
             else if (alsv.size() != ipv.size())
             {
+                SWSS_LOG_NOTICE("Route %s: resize ipv to match alsv, %zd -> %zd.", key.c_str(), ipv.size(), alsv.size());
                 ipv.resize(alsv.size());
             }
 
@@ -489,6 +491,7 @@ void RouteOrch::doTask(Consumer& consumer)
             {
                 if (ip.empty())
                 {
+                    SWSS_LOG_NOTICE("Route %s: set the empty nexthop ip to zero.", key.c_str());
                     ip = ip_prefix.isV4() ? "0.0.0.0" : "::";
                 }
             }
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -467,10 +467,12 @@ void RouteOrch::doTask(Consumer& consumer)
             vector<string> alsv = tokenize(aliases, ',');
 
             /*
-             * To check the validity of nexthop and make sure it works even in
-             * abnormal cases. One case is old format - empty "nexthop",
-             * then set ip to zero for backward compatibility.
+             * For backward compatibility, adjust ip string from old format to
+             * new format. Meanwhile it can deal with some abnormal cases.
              */
+
+            /* Resize the ip vector to match ifname vector
+             * as tokenize(",", ',') will miss the last empty segment. */
             if (alsv.size() == 0)
             {
                 it = consumer.m_toSync.erase(it);
@@ -480,6 +482,9 @@ void RouteOrch::doTask(Consumer& consumer)
             {
                 ipv.resize(alsv.size());
             }
+
+            /* Set the empty ip(s) to zero
+             * as IpAddress("") will construst a incorrect ip. */
             for (auto &ip : ipv)
             {
                 if (ip.empty())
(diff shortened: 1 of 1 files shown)


orchagent/routeorch.cpp
```diff
@@ -466,6 +466,21 @@ void RouteOrch::doTask(Consumer& consumer)
...
+             * then set ip to zero for backward compatibility.
+             */
+            if (alsv.size() == 0)
+            {
+                it = consumer.m_toSync.erase(it);
+                continue;
+            }
+            else if (alsv.size() != ipv.size())
+            {
+                ipv.resize(alsv.size(), "0.0.0.0");
```
prsunny, line 481: How about IPv6? Since `nexthop `is a set of entries, will there be a case where we can have a mix of valid ips and empty strings `""`?

tylerlinp:
@prsunny It is a good question. That depends on zebra route types, as I know single nexthop there are 2 types - ip&intf or intf, and for nexthop group I have only seen a set of ip&intf by now. 
If nexthop group contains intf really, actually it maynot work well.
To do exact backward compatibility, some change is needed. One is your case，it is not well to use a ipv4 address in ipv6 route. The other is when empty string not in the tail, that would construct a incorrect ip.

orchagent/routeorch.cpp
```diff
@@ -466,6 +466,28 @@ void RouteOrch::doTask(Consumer& consumer)
...
+            if (alsv.size() == 0)
+            {
+                it = consumer.m_toSync.erase(it);
+                continue;
+            }
+            else if (alsv.size() != ipv.size())
+            {
+                ipv.resize(alsv.size());
+            }
+            for (auto &ip : ipv)
```
lguohan, line 483: please have a empty line above. the for is not part of if else block. have a separate line make it clear.

orchagent/routeorch.cpp
```diff
@@ -466,6 +466,28 @@ void RouteOrch::doTask(Consumer& consumer)
             vector<string> ipv = tokenize(ips, ',');
             vector<string> alsv = tokenize(aliases, ',');
 
+            /*
+             * To check the validity of nexthop and make sure it works even in
+             * abnormal cases. One case is old format - empty "nexthop",
+             * then set ip to zero for backward compatibility.
```
lguohan, line 472: since you are handling two different cases here, please consider to have separate comments. the abnormal case, and back compatibility case. Please move the backware compability case before the for loop.

tylerlinp: All is for back compatibility including `resize() `, so it is difficult to separate cases, and I will separate `for` block and rewrite the comments.

orchagent/routeorch.cpp
```diff
@@ -466,6 +466,33 @@ void RouteOrch::doTask(Consumer& consumer)
...
+            /*
+             * For backward compatibility, adjust ip string from old format to
+             * new format. Meanwhile it can deal with some abnormal cases.
+             */
+
+            /* Resize the ip vector to match ifname vector
+             * as tokenize(",", ',') will miss the last empty segment. */
+            if (alsv.size() == 0)
+            {
+                it = consumer.m_toSync.erase(it);
```
zhenggen-xu, line 478: log it into syslog with WARN level so we know what happened.

orchagent/routeorch.cpp
```diff
@@ -466,6 +466,33 @@ void RouteOrch::doTask(Consumer& consumer)
...
+                ipv.resize(alsv.size());
+            }
+
+            /* Set the empty ip(s) to zero
+             * as IpAddress("") will construst a incorrect ip. */
+            for (auto &ip : ipv)
+            {
+                if (ip.empty())
+                {
+                    ip = ip_prefix.isV4() ? "0.0.0.0" : "::";
```
zhenggen-xu, line 492: log with INFO or NOTICE level here?

zhenggen-xu:
Do we have any vs test cases to cover this fix?

tylerlinp:
@zhenggen-xu I think we have done enough more for backward compatibility just warmboot from old version. In the near future, the codes would have no use. Even now in the process of running, it will never happen as the entries are written by fpmsyncd. So I think there is no need to add new vs test cases or log infos.

prsunny:
@zhenggen-xu , would like to merge this since it is blocking the current tests. As Tyler mentioned, VS test may not be required to cover this but we can plan to add logs in a future iteration if necessary. What do you say?

prsunny:
@tylerlinp , had a discussion with @zhenggen-xu and @yxieca, and we feel it is better to add a log to flag this scenario. Could you please help add the log?

qiluo-msft:
Sorry for late comment. Could you add some gtest or vs test cases to cover the cases you mentioned?

PR #1278, opened 2020-04-29
[aclorch.cpp] Handle all the ACL redirect requests in AclRuleL3::validateAddAction() 

madhanmellanox:
**What I did**
I made changes to methods AclRuleL3()::validateAddAction() and AclRuleL3::getRedirectObject() in order to accomodate all type of ACL redirect commands.

**Why I did it**
Because, the current code had some assumptions on how the ACL redirect request should be parsed in the ACL orchagent. I removed the assumption.

**How I verified it**
I verified it by testing different kinds of ACL redirect commands on my switch.

**Details if related**


tests/test_acl.py
```diff
@@ -560,7 +570,52 @@ def test_AclRuleIcmpV6(self, dvs):
         self.remove_acl_table("test_icmpv6")
         self.verify_no_acl_tables()
 
-    def test_AclRuleRedirectToNextHop(self, dvs):
+    def test_AclpriorityRule(self, dvs):
```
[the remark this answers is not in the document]
madhanmellanox: The test_AclpriorityRule() test case is for the fix of different defect (lower case priority of ACL rule).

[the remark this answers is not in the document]
stepanblyschak: That test should be in different PR together with lower case priority fix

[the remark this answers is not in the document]
stepanblyschak: This test does not cover fix in https://github.com/Azure/sonic-utilities/pull/897. Where does it call 'acl-loader'? What it tests is wether orchagent handles lower case priority string, but not 'acl-loader'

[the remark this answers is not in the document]
madhanmellanox: Removed the test_Aclpriority() rule test case as it does not belong here.

stepanblyschak:
@madhanmellanox please, fix conflicts

liat-grozovik:
retest vs please

orchagent/aclorch.cpp
```diff
@@ -809,8 +809,22 @@ bool AclRuleL3::validateAddAction(string attr_name, string _attr_value)
         // handle PACKET_ACTION_REDIRECT in ACTION_PACKET_ACTION for backward compatibility
         else if (attr_value.find(PACKET_ACTION_REDIRECT) != string::npos)
         {
-            // resize attr_value to remove argument, _attr_value still has the argument
-            attr_value.resize(string(PACKET_ACTION_REDIRECT).length());
+            // check that we have a colon after redirect rule
+            size_t colon_pos = string(PACKET_ACTION_REDIRECT).length();
```
pavel-shirshov, line 813: this could be constexpr

orchagent/aclorch.cpp
```diff
@@ -809,8 +809,22 @@ bool AclRuleL3::validateAddAction(string attr_name, string _attr_value)
...
+
+            if (attr_value.c_str()[colon_pos] != ':')
```
pavel-shirshov, line 815: would `attr_value[colon_pos]` work here? I don't understand why do we need to extract c_str from the string?

stepanblyschak: if size of attr_value is colon_pos (e.g. attr_value == PACKET_ACTION_REDIRECT), then accessing attr_value[colon_pos] is accessing out of bounds, so undefined behaviour. std::string is not required to put '\0' at the end of the string while c_str() method is. So in case size of attr_value is colon_pos then attr_value.c_str()[colon_pos] will return null byte.

orchagent/aclorch.cpp
madhanmellanox: is it like: constexpr size_t colon_pos = string(PACKET_ACTION_REDIRECT).length()?

liat-grozovik:
@pavel-shirshov and @stepanblyschak could you please check the recent changes and approve if possible?

stepanblyschak:
@madhanmellanox Please, fix conflicts

pavel-shirshov:
Danny? Can you please take a look?

daall:
> Danny? Can you please take a look?

ack

orchagent/aclorch.cpp
```diff
@@ -856,21 +870,8 @@ bool AclRuleL3::validateAddAction(string attr_name, string _attr_value)
...
-        return SAI_NULL_OBJECT_ID;
-    }
-
-    if (colon_pos + 1 == redirect_value.length())
-    {
-        SWSS_LOG_ERROR("Redirect action rule must have a target after 'REDIRECT:' action");
-        return SAI_NULL_OBJECT_ID;
-    }
-
-    string target = redirect_value.substr(colon_pos + 1);
```
daall, line 873: I'm curious - why do we need to move this logic from this method to the validate method?

madhanmellanox: We moved this logic to the validate method to accomodate both the redirect ACL rules to be handled in the same place.

daall: Oh, I see that now! It was below the fold. Thanks!

daall:
retest this please

madhanmellanox:
retest vsimage please

madhanmellanox:
retest vsimage please

madhanmellanox:
retest vsimage please

daall:
retest this please

abdosi:
@stepanblyschak Cherry-pick has conflict. Create PR for 201911.

@liat-grozovik @rlhui

abdosi:
merged in 201911 with direct pr

PR #1280, opened 2020-04-30
Using PRI instead of %l to support 32-bit arch

Sabareesh-Kumar-Anandan:
Fixing compilation issue in 32-bit arch(marvell-armhf)

**What I did**
Replaced %l variants with PRI.
Replaced %lu with %zu for size_t

**Why I did it**
fix for compilation issue in bit arch

**How I verified it**
compiled for 32 bit arch (marvell-armhf)

### DIFF
--- a/cfgmgr/natmgr.cpp
+++ b/cfgmgr/natmgr.cpp
@@ -5049,7 +5049,7 @@ void NatMgr::doStaticNatTask(Consumer &consumer)
         /* Ensure the key size is 1 otherwise ignore */
         if (keys.size() != STATIC_NAT_KEY_SIZE)
         {
-            SWSS_LOG_ERROR("Invalid key size %u, skipping %s", keys.size(), key.c_str());
+            SWSS_LOG_ERROR("Invalid key size %" PRIu32 ", skipping %s", keys.size(), key.c_str());
             it = consumer.m_toSync.erase(it);
             continue;
         }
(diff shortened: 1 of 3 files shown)


cfgmgr/natmgr.cpp
```diff
@@ -5049,7 +5049,7 @@ void NatMgr::doStaticNatTask(Consumer &consumer)
         /* Ensure the key size is 1 otherwise ignore */
         if (keys.size() != STATIC_NAT_KEY_SIZE)
         {
-            SWSS_LOG_ERROR("Invalid key size %lu, skipping %s", keys.size(), key.c_str());
+            SWSS_LOG_ERROR("Invalid key size %u, skipping %s", keys.size(), key.c_str());
```
antony-rheneus, line 5052: for size_t or size_type "%z" should be used. Can you please verify it

Sabareesh-Kumar-Anandan: %z is not supported. I changed it to PRIu32.

cfgmgr/natmgr.cpp
```diff
@@ -5049,7 +5049,7 @@ void NatMgr::doStaticNatTask(Consumer &consumer)
         /* Ensure the key size is 1 otherwise ignore */
         if (keys.size() != STATIC_NAT_KEY_SIZE)
         {
-            SWSS_LOG_ERROR("Invalid key size %lu, skipping %s", keys.size(), key.c_str());
+            SWSS_LOG_ERROR("Invalid key size %" PRIu32 ", skipping %s", keys.size(), key.c_str());
```
antony-rheneus, line 5052: Can you confirm it doesn't break for 64bit cpu. Also check %z as it can be used for size_t/size_type

cfgmgr/natmgr.cpp
Sabareesh-Kumar-Anandan: Changed it to %zu.

cfgmgr/natmgr.cpp
Sabareesh-Kumar-Anandan: changed to %zu

PR #1284, opened 2020-04-30
[vnet] Fix IP2ME route creation logic for BITMAP VNET interface

volodymyrsamotiy:

**What I did**
Fixed IP2ME route creation logic for BITMAP VNET interface: added verification if IP2ME route is already added in order to avoid duplication.

**Why I did it**
SONiC failed to start after warm-reboot because there are few events to create interface and as a result it tried to add same IP2ME route twice, which causes orchagent to crash.

**How I verified it**
Executed warm-reboot with the VNET interfaces configured and verified all is up after boot up.

**Details if related**
N/A


orchagent/intfsorch.cpp
```diff
@@ -1094,6 +1094,18 @@ void IntfsOrch::generateInterfaceMap()
     m_updateMapsTimer->start();
 }
 
+bool updateSyncdIntfPfx(const string& alias, const IpPrefix &ip_prefix)
```
prsunny, line 1097: class name - `IntfsOrch  `missing in definition

orchagent/vnetorch.cpp
```diff
@@ -717,7 +717,7 @@ bool VNetBitmapObject::addIntf(const string& alias, const IpPrefix *prefix)
         intfMap_.emplace(alias, intfInfo);
     }
 
-    if (prefix)
+    if (prefix && gIntfsOrch->updateSyncdIntfPfx(alias, *prefix))
```
prsunny, line 720: I don't get this. If it is "vnet" interface, it will call the `VNetOrch::setIntf`. So how will this get added to `syncd `interfaces

orchagent/intfsorch.cpp
volodymyrsamotiy: Yes, mistakanly uploaded wrong patch. Fixed it.

orchagent/vnetorch.cpp
volodymyrsamotiy: Right, ```intfsorch``` calls ```VNetOrch::setIntf -> VNetBitmapObject::addIntf```, and```addIntf``` creates VNET specific objects for the specific VNET interface but before that it also calls ```IntfsOrch::setIntf``` function which adds such interface to ```IntfsOrch::m_syncdIntfses```. So here interface is already in syncd map and we just add IP2ME prefix to the appropriate map in case it is not there.

volodymyrsamotiy:
retest this please

prsunny:
retest this please

PR #1286, opened 2020-05-04
[aclorch] Add support for creating ingress and egress MIRROR tables concurrently

daall:

**What I did**
I made it so that users can create MIRROR tables with ingress ACLs and MIRROR tables with egress ACLs at the same time (and similarly for MIRRORv6 tables).

**Why I did it**
Currently only one mirror table of each type is allowed in order to support the logic for reconciling combined v4/v6 mirror tables. However, we can differentiate based on ACL stage, keeping the reconciliation logic intact while allowing users to mirror with ingress and egress ACLs simultaneously (on devices that support this).

**How I verified it**
WIP. I've provided a new VS test case to check the behavior, I'm in the process of testing the changes on a DUT just to be safe.

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -2717,22 +2717,21 @@ bool AclOrch::addAclTable(AclTable &newTable)
     // Check if a separate mirror table is needed or not based on the platform
     if (newTable.type == ACL_TABLE_MIRROR || newTable.type == ACL_TABLE_MIRRORV6)
     {
-        auto mirror_id = m_mirrorTableId.find(table_stage);
-        auto mirror_v6_id = m_mirrorV6TableId.find(table_stage);
         if (m_isCombinedMirrorV6Table &&
-            (!m_mirrorTableId[table_stage].empty() || !m_mirrorV6TableId[table_stage].empty())) {
+                (!m_mirrorTableId[table_stage].empty() ||
+                !m_mirrorV6TableId[table_stage].empty())) {
             string orig_table_name;
 
             // If v4 table is created, mark v6 table is created
             if (!m_mirrorTableId[table_stage].empty())
             {
-                orig_table_name = mirror_id->second;
+                orig_table_name = m_mirrorTableId[table_stage];
                 m_mirrorV6TableId[table_stage] = newTable.id;
             }
             // If v6 table is created, mark v4 table is created
             else
             {
-                orig_table_name = mirror_v6_id->second;
+                orig_table_name = m_mirrorV6TableId[table_stage];
                 m_mirrorTableId[table_stage] = newTable.id;
             }
 
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -2213,6 +2213,12 @@ void AclOrch::init(vector<TableConnector>& connectors, PortsOrch *portOrch, Mirr
 
     queryAclActionCapability();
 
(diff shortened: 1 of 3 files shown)


daall:
retest this please

daall:
retest this please

prsunny:
lgtm

daall:
@prsunny @lguohan @abdosi please help review. :)

orchagent/aclorch.cpp
```diff
@@ -3212,18 +3224,20 @@ sai_object_id_t AclOrch::getTableById(string table_id)
...
-        else
+    for (auto stage: {ACL_STAGE_INGRESS, ACL_STAGE_EGRESS}) {
+        if (m_isCombinedMirrorV6Table &&
+                (table_id == m_mirrorTableId[stage] || table_id == m_mirrorV6TableId[stage]))
         {
-            return getTableById(m_mirrorTableId);
+            // If the table is v4, the corresponding v6 table is already created
+            if (table_id == m_mirrorTableId[stage])
+            {
+                return getTableById(m_mirrorV6TableId[stage]);
```
abdosi, line 3234: @daall looks this condition need to be swap.
if (table_id == m_mirrorTableId[stage]) shouldn't be return getTableById(m_mirrorTableId[stage]);

orchagent/aclorch.cpp
```diff
@@ -2706,23 +2717,22 @@ bool AclOrch::addAclTable(AclTable &newTable)
     // Check if a separate mirror table is needed or not based on the platform
     if (newTable.type == ACL_TABLE_MIRROR || newTable.type == ACL_TABLE_MIRRORV6)
     {
-
         if (m_isCombinedMirrorV6Table &&
```
abdosi, line 2720: @daall should'nt m_isCombinedMirrorV6Table be also define per stage ? Possible Ingress can be separate and combine on egressor vice-versa

orchagent/aclorch.cpp
prsunny: @abdosi , it looks like the same logic as before and not changed in this PR

abdosi: @prsunny yes even previous logic didn't look correct to me.

orchagent/aclorch.cpp
daall: I'm inclined to leave it alone for now since this variable is hardcoded based on platform: https://github.com/Azure/sonic-swss/blob/727a518159a08e245f6f37d58f796f2b9097774a/orchagent/aclorch.cpp#L2169

orchagent/aclorch.cpp
daall: Now that you point it out, I'm not 100% sure this part is even necessary. If you look above, the sibling table gets assigned the same ID as the original table when it gets created anyways: https://github.com/Azure/sonic-swss/blob/7efc055b1be4830858a5fe364c72e0c823918e4e/orchagent/aclorch.cpp#L2726

So regardless of which ID we use, they should always be the same.

I agree with @prsunny that the logic is the same as before, so I'm inclined to keep it for now and then go back and refactor it in a separate PR (so if we discover some subtlety and this part *is* needed, we can easily revert it). Thoughts?

abdosi: ok, since the table id is same we are fine for now.

orchagent/aclorch.cpp
abdosi: ok

PR #1287, opened 2020-05-04
[portsyncd] fix default select timeout

stepanblyschak:
Default select timeout was set to 1 ms, causing 3-5% CPU usage when
idles.

**What I did**

I changed default select timeout from 1ms to 1000ms.

**Why I did it**

portsyncd is causing 3-5% CPU usage when system idles.

**How I verified it**

Basic sanity verification on T0 topology

**Details if related**

### DIFF
--- a/portsyncd/portsyncd.cpp
+++ b/portsyncd/portsyncd.cpp
@@ -20,6 +20,8 @@
 using namespace std;
 using namespace swss;
 
+#define DEFAULT_SELECT_TIMEOUT 1000 /* ms */
+
 /*
  * This g_portSet contains all the front panel ports that the corresponding
  * host interfaces needed to be created. When this LinkSync class is
@@ -103,16 +105,26 @@ int main(int argc, char **argv)
         {
             Selectable *temps;
             int ret;
-            ret = s.select(&temps, 1);
+            ret = s.select(&temps, DEFAULT_SELECT_TIMEOUT);
 
             if (ret == Select::ERROR)
             {
                 cerr << "Error had been returned in select" << endl;
                 continue;
             }
-
-            if (ret == Select::TIMEOUT)
+            else if (ret == Select::TIMEOUT)
             {
+                continue;
+            }
+            else if (ret != Select::OBJECT)
+            {
+                SWSS_LOG_ERROR("Unknown return value from Select %d", ret);
+                continue;
+            }
+
+            if (temps == static_cast<Selectable*>(&netlink))
+            {
+                /* on netlink message, check if PortInitDone should be sent out */
                 if (!g_init && g_portSet.empty())
                 {
                     /*
@@ -134,8 +146,7 @@ int main(int argc, char **argv)
                     handlePortConfig(p, port_cfg_map);
                 }
             }
-
-            if (temps == (Selectable *)&portCfg)
+            else if (temps == (Selectable *)&portCfg)
             {
                 std::deque<KeyOpFieldsValuesTuple> entries;
                 portCfg.pops(entries);
@@ -152,6 +163,11 @@ int main(int argc, char **argv)
                     port_cfg_map[key] = entry;
                 }
                 handlePortConfig(p, port_cfg_map);
+            }
+            else
+            {
+                SWSS_LOG_ERROR("Unknown object returned by select");
+                continue;
             }
         }
     }
@@ -179,7 +195,9 @@ static void notifyPortConfigDone(ProducerStateTable &p)
 
 bool handlePortConfigFromConfigDB(ProducerStateTable &p, DBConnector &cfgDb, bool warm)
 {
-    cout << "Get port configuration from ConfigDB..." << endl;
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_NOTICE("Getting port configuration from ConfigDB...");
 
     Table table(&cfgDb, CFG_PORT_TABLE_NAME);
     std::vector<FieldValueTuple> ovalues;


stepanblyschak:
Needed for 201811, 201904, 201911 and master

prsunny:
lgtm, @zhenggen-xu to check since you had some changes in portsyncd for dpb?

portsyncd/portsyncd.cpp
```diff
@@ -179,7 +195,9 @@ static void notifyPortConfigDone(ProducerStateTable &p)
 
 bool handlePortConfigFromConfigDB(ProducerStateTable &p, DBConnector &cfgDb, bool warm)
 {
-    cout << "Get port configuration from ConfigDB..." << endl;
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_NOTICE("Getting port configuration from ConfigDB...");
```
qiluo-msft, line 200: Why change this part? This is one time `cout` at the beginning, no harm to cout.

portsyncd/portsyncd.cpp
```diff
@@ -103,16 +105,26 @@ int main(int argc, char **argv)
...
+                continue;
+            }
+            else if (ret != Select::OBJECT)
+            {
+                SWSS_LOG_ERROR("Unknown return value from Select %d", ret);
+                continue;
+            }
 
-            if (ret == Select::TIMEOUT)
+            if (temps == static_cast<Selectable*>(&netlink))
```
qiluo-msft, line 125: >if (temps == static_cast<Selectable*>(&netlink)) [](start = 12, length = 48)

The behavior is changed and suspicious? Could you focus on "fix default select timeout" in this PR?

portsyncd/portsyncd.cpp
```diff
@@ -20,6 +20,8 @@
 using namespace std;
 using namespace swss;
 
+#define DEFAULT_SELECT_TIMEOUT 1000 /* ms */
```
qiluo-msft, line 23: >1000 [](start = 31, length = 4)

Is it too slow? How about CPU usage @200 ms?

portsyncd/portsyncd.cpp
stepanblyschak: SWSS_LOG_* is a usual way to do logging in SONiC, why leave it to be cout?

portsyncd/portsyncd.cpp
stepanblyschak: What is suspicious here? On netlink message, check wether we need to send PortInitDone. Previously this check was done every 1ms.

portsyncd/portsyncd.cpp
stepanblyschak: Any number of ms should be good here, I don't see a problem with 1000 ms, and this is default for other daemons. Why it is to slow? Which tasks portsyncd needs to do every DEFAULT_SELECT_TIMEOUT ms?

portsyncd/portsyncd.cpp
qiluo-msft: It not necessarily mean your change is wrong. I mean the purpose of this PR is to "fix default select timeout".

Let's assume if you don't change the behavior here, this check will be done every 1000ms. Anything wrong? If there is a bug, I prefer fix it in a separate PR.

---
In reply to: [420307407](https://github.com/Azure/sonic-swss/pull/1287#discussion_r420307407) [](ancestors = 420307407)

stepanblyschak: Yes, it is kind of another problem. That's why I assume the select timeout was set to very small value. If we increase select timeout to 1 sec, the check will be performed every 1 sec. This means there might be 1 sec delay in notifying system about PortInitDone. To avoid this unneccesary delay, we check if select was unblocked due to netlink message and do the PortInitDone check. So, I think that issue is strongly related to select timeout and don't see a reason to fix it in a seperate PR. I hope that sounds reasonable to fix in one PR

qiluo-msft: `g_portSet` is decreased by LinkSync::onMsg(). Not sure the relation ship between the timing of that and selecting `netlink`. Is it possible to have time risk?

---
In reply to: [421040960](https://github.com/Azure/sonic-swss/pull/1287#discussion_r421040960) [](ancestors = 421040960,420307407)

stepanblyschak: LinkSync::onMsg is called in NetLink::readData() in select(), so g_portSet will be updated after select()

portsyncd/portsyncd.cpp
zhenggen-xu: From the code, this number is not really determining the frequency of checking the events, since whenever you timeout, you call select again without doing anything, so it is almost always checking the events, thus no slowness concerns.  The code in the review will have similar behavior with bigger timeout even with -1.  But I think setting to 1 is fine.

portsyncd/portsyncd.cpp
zhenggen-xu: It is always good  to improve the code, but I feel we should keep consistency. If we are changing the style in this function that is not touched by your core logic, we should make other places in this file to be consistent.  like line 112 etc.  Or we could leave all these improvements in a separate PR, and address the whole file. This way it would be easier to maintain commits across branches etc.

zhenggen-xu:
Overall, LGTM,  just had two comments.

abdosi:
@lguohan @rlhui 

Do we need this for 201911 ?

yxieca:
This change is an enhancement. I'll leave it out for 201811.

abdosi:
Not taking for 201911 at this stage.

Issue #1294, opened 2020-05-14
[FDB] Doesn't flush fdb entries on port link down


MikeHcChen:
Original issue was posted by [dev-aviz](https://github.com/dev-aviz) on Azure/sonic-buildimage: https://github.com/Azure/sonic-buildimage/issues/4184#issue-569290297
However, this issue should be related with Azure/sonic-swss repository, therefore I copy the issue 
in this repository.

Description
(DUT)Device Under Test does not flush mac entries on port link down.

Expected:
Switch should remove all MAC entries learned on inactive/down ports

Issue observed

Broadcom
Mellnox
Steps to reproduce the issue:

Configure Ethernet10 as trunk port and allow Vlan 10
Send L2 traffic with SMAC = 00:AA:AA:AA:AA:00 -> 00:AA:AA:AA:AA:09
Switch learns MAC address on Ethernet10 and Vlan10 correctly
Shutdown Ethernet10
Switch does not flush MAC entries learned on Ethernet10
Describe the results you received:
Device Under Test does not flush mac entries on port link down.

Describe the results you expected:
Switch should remove all MAC entries learned on inactive/down ports

[Version]
SONiC Software Version: SONiC.master.0-dirty-20191207.003133
Distribution: Debian 9.11
Kernel: 4.9.0-9-2-amd64
Build commit: f345935
Build date: Sat Dec 7 11:52:18 UTC 2019
Built by: dev-aviz@aviznetworks.com

Platform: x86_64-accton_as5712_54x-r0
HwSKU: Accton-AS5712-54X
ASIC: broadcom
Serial Number: 571254X1630035
Uptime: 20:09:57 up 2:53, 1 user, load average: 1.88, 2.05, 2.06

How to validate the results:

[Console]
root@spine1:~# show vlan config
Name VID Member Mode

Vlan10 10 Ethernet0 tagged
Vlan10 10 Ethernet10 tagged
root@spine1:~#

root@spine1:~# show interface status Ethernet10
Interface Lanes Speed MTU Alias Vlan Oper Admin Type Asym PFC

Ethernet10 27 10G 9100 tenGigE10 trunk down down SFP/SFP+/SFP28 N/A
root@spine1:~# show mac
No. Vlan MacAddress Port Type

1      10  00:AA:AA:AA:AA:06  Ethernet10  Dynamic
2      10  00:AA:AA:AA:AA:02  Ethernet10  Dynamic
[...]

PR #1242, opened 2020-04-06
[DPB/VLAN] Add test VS cases and fix FDB flush issues 

vasant17:
**What I did**
Added DPB-VLAN dependency test cases
Fixed FDB flush issues encountered during test

**Why I did it**
When port is a member of VLAN and we do dynamic port breakout, we identify the dependency and remove the port from VLAN and then delete the port.
But, when we remove the port from VLAN, couple of things happen or should happen
1. FDB entries associated with the bridge port ID of the port in all VLANs should be removed.
2. Delete the bridge port.

This will enable/allow us to delete the port.

While testing we hit the following issue:
Bug: Even though we remove the port from all VLANS, we were unable to delete the port, as associated bridge port OID was still present

Root cause: The bridge port OID associated with port was NOT deleted because,  associated FDB entries were NOT flushed. 

Solution: 
We are going to clear FDB entries in two cases:
1. When port is shutdown
2. When port is removed from VLAN.

Firstly, when port is shutdown, we are going to call SAI FDB flush API  by passing bridge port OID. Note that we will do this only if the port has valid (NON NULL) bridge port OID. Thats because if you call SAI FDB flush without any valid bridge port OID or BVID, it will flush all FDB entries.

Secondly, whenever we remove the port from a VLAN, we are going to call SAI FDB flush API with valid <bridge port OID, BVID>

With both these fixes in place, irrespective of the order we will clean the FDB entries, delete bridge port OID and then delete the port. The order I am talking here is, whether the port is shutdown first or its removed from all VLANs.

**How I verified it**
Ran VS test cases

**Details if related**
```
vapatil@server09:~/workspace/DPB/sonic-buildimage/src/sonic-swss/tests$ sudo pytest --pdb -s -v --dvsname=vs-vp test_port_dpb_vlan.py
[sudo] password for vapatil: 
======================================================================= test session starts ========================================================================
platform linux2 -- Python 2.7.15+, pytest-3.3.0, py-1.8.0, pluggy-0.6.0 -- /usr/bin/python
cachedir: .cache
rootdir: /home/vapatil/workspace/DPB/sonic-buildimage/src/sonic-swss/tests, inifile:
collected 4 items                                                                                                                                                  

test_port_dpb_vlan.py::TestPortDPBVlan::test_dependency remove extra link dummy
PASSED                                                                                               [ 25%]
test_port_dpb_vlan.py::TestPortDPBVlan::test_one_port_one_vlan PASSED                                                                                        [ 50%]
[...]

PR #909, opened 2019-05-23
Remove FDB entries when shutting down the interface

InventecJamesHuang:
**What I did**
Add the flushing FDB feature in PortsOrch to handle the situation when the user shut down the interface.

**How I did it**
We create the new function named 'removeFDBEntriesByBridgePortID'. 
This function is responsible for calling the SAI flush_fdb_entries() function to remove the FDB entries.

There are three situations to call the above function,

1. In **doTask** function. 
    When SONiC receives the port_state_change op event, if the interface goes down, it calls removeFDBEntriesByBridgePortID() to remove FDB entries.
2.  In **removeLag** function. 
   When SONiC removes the LAG, it also calls removeFDBEntriesByBridgePortID() to remove FDB entries.
3. In **removeLagMember** function. 
  When SONiC removes one member from the LAG, if it causes the LAG's oper_status goes DOWN ( since the total number of enabled links less than the minimum number of links), it also calls removeFDBEntriesByBridgePortID() to remove FDB entries.

**How I verified it**
We prepared two physical ports (Ethernet44) and one virtual port (PortChannel1) to test.
First, I enabled two ports and each port learned 5 different FDB entries.

```
root@SONiC-Inventec-d6556:/home/admin# show mac
  No.    Vlan  MacAddress         Port           Type
 -----  ------  -----------------  -------------  -------
    1      11  00:10:94:00:00:21  PortChannel01  Dynamic
    2      11  00:10:94:00:00:22  PortChannel01  Dynamic
    3      11  00:10:94:00:00:05  Ethernet44     Dynamic
    4      11  00:10:94:00:00:03  Ethernet44     Dynamic
    5      11  00:10:94:00:00:02  Ethernet44     Dynamic
    6      11  00:10:94:00:00:06  Ethernet44     Dynamic
    7      11  00:10:94:00:00:20  PortChannel01  Dynamic
    8      11  00:10:94:00:00:23  PortChannel01  Dynamic
    9      11  00:10:94:00:00:04  Ethernet44     Dynamic  
   10      11  00:10:94:00:00:24  PortChannel01  Dynamic
Total number of entries 10
```
**Shutting down Ethernet44**

```
root@SONiC-Inventec-d6556:/home/admin# config interface shutdown Ethernet44
root@SONiC-Inventec-d6556:/home/admin# show mac
  No.    Vlan  MacAddress         Port           Type
-----  ------  -----------------  -------------  -------
    1      11  00:10:94:00:00:21  PortChannel01  Dynamic
[...]

PR #1295, opened 2020-05-15
[fdborch]: Flush fdb entries on port link down (Resolves #1063, #1294, resolves Azure/sonic-buildimage#4184)

### LINKED ISSUE #1063: [fdb]fdborch: fdb entries are cleared according to vlan or port or vlan&&port

MikeHcChen:
**What I did**

1. Add a method to inform fdborch to flush FDB while port operation status changed.
2. Implement codes for processing events of the flushed FDB on specified port or vlan notifications of SAI

**Why I did it**
Fix issue: [[FDB] Doesn't flush fdb entries on port link down #1294](https://github.com/Azure/sonic-swss/issues/1294)

**How I verified it**
Please refer test steps in [[FDB] Doesn't flush fdb entries on port link down #1294](https://github.com/Azure/sonic-swss/issues/1294)

Test on broadcom platform 'x86_64-delta_ag9032v1-r0 with SONiC Software Version: 201911 branch commit 2218ed4 on May 1, 2020.

**Details if related**

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -186,17 +186,95 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
         }
         else if (bridge_port_id && entry->bv_id == SAI_NULL_OBJECT_ID)
         {
-            /*this is a placeholder for flush port fdb case, not supported yet.*/
-            SWSS_LOG_ERROR("FdbOrch notification: not supported flush port fdb action, port_id = 0x%" PRIx64 ", bv_id = 0x%" PRIx64 ".", bridge_port_id, entry->bv_id);
+            if (!m_portsOrch->getPortByBridgePortId(bridge_port_id, update.port))
+            {
+                SWSS_LOG_ERROR("Failed to get port by bridge port ID 0x%" PRIx64, bridge_port_id);
+                return;
+            }
+
+            update.entry.port_name = update.port.m_alias;
+
+            /* Remove all FDB entries of specified port by going through m_entries */
+            for (auto itr = m_entries.begin(); itr != m_entries.end();)
+            {
+                if (itr->port_name == update.port.m_alias)
+                {
+                    update.entry.mac = itr->mac;
+                    update.entry.bv_id = itr->bv_id;
+                    update.add = false;
+                    itr++;
+
+                    storeFdbEntryState(update);
+
+                    SWSS_LOG_DEBUG("FdbOrch notification: mac %s , bv_id 0x%" PRIx64 " on port ID 0x%" PRIx64 " was removed", update.entry.mac.to_string().c_str(), update.entry.bv_id, bridge_port_id);
+
+                    for (auto observer: m_observers)
+                    {
+                        observer->update(SUBJECT_TYPE_FDB_CHANGE, &update);
+                    }
+                    continue;
+                }
+                itr++;
+            }
         }
         else if (bridge_port_id == SAI_NULL_OBJECT_ID && entry->bv_id != SAI_NULL_OBJECT_ID)
         {
-            /*this is a placeholder for flush vlan fdb case, not supported yet.*/
-            SWSS_LOG_ERROR("FdbOrch notification: not supported flush vlan fdb action, port_id = 0x%" PRIx64 ", bv_id = 0x%" PRIx64 ".", bridge_port_id, entry->bv_id);
+            /* Remove all FDB entries of specified vlan by going through m_entries */
+            for (auto itr = m_entries.begin(); itr != m_entries.end();)
+            {
+                if (itr->bv_id == entry->bv_id)
+                {
+                    update.entry.mac = itr->mac;
+                    update.entry.bv_id = itr->bv_id;
+                    update.add = false;
+                    itr++;
+
+                    storeFdbEntryState(update);
+
+                    SWSS_LOG_DEBUG("FdbOrch notification: mac %s , bv_id 0x%" PRIx64 " was removed", update.entry.mac.to_string().c_str(), update.entry.bv_id);
+
+                    for (auto observer: m_observers)
+                    {
+                        observer->update(SUBJECT_TYPE_FDB_CHANGE, &update);
+                    }
+                    continue;
+                }
+                itr++;
+            }
         }
         else
         {
-            SWSS_LOG_ERROR("FdbOrch notification: not supported flush fdb action, port_id = 0x%" PRIx64 ", bv_id = 0x%" PRIx64 ".", bridge_port_id, entry->bv_id);
+            if (!m_portsOrch->getPortByBridgePortId(bridge_port_id, update.port))
+            {
+                SWSS_LOG_ERROR("Failed to get port by bridge port ID 0x%" PRIx64, bridge_port_id);
+                return;
+            }
+
+            update.entry.port_name = update.port.m_alias;
+
+            /* Remove all FDB entries of specified port and vlan by going through m_entries */
+            for (auto itr = m_entries.begin(); itr != m_entries.end();)
+            {
+                if ((itr->port_name == update.port.m_alias)
+                    && (itr->bv_id == entry->bv_id))
+                {
+                    update.entry.mac = itr->mac;
+                    update.entry.bv_id = itr->bv_id;
+                    update.add = false;
+                    itr++;
+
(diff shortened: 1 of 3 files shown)


prsunny:
There is a similar PR that handles this use-case differently - https://github.com/Azure/sonic-swss/pull/909. It could be incomplete based on the changes you provided in this PR. One question is when you do "flush" do you also get FDB delete notifications back from SAI for each of the entry that's removed?

MikeHcChen:
> There is a similar PR that handles this use-case differently - #909. It could be incomplete based on the changes you provided in this PR. One question is when you do "flush" do you also get FDB delete notifications back from SAI for each of the entry that's removed?

When I do 'flush'on a port, the ASIC will notify SONiC with sai_fdb_event_t type=SAI_FDB_EVENT_AGED or SAI_FDB_EVENT_FLUSHED. For case SAI_FDB_EVENT_AGED,
 original (FdbOrch::update() in fdborch.cpp) codes can handle well. For case SAI_FDB_EVENT_FLUSHED, I added codes to handle 
the notification of flushed a port in case SAI_FDB_EVENT_FLUSHED of FdbOrch::update() in fdborch.cpp.

orchagent/fdborch.cpp
```diff
@@ -259,6 +337,40 @@ bool FdbOrch::getPort(const MacAddress& mac, uint16_t vlan, Port& port)
     return true;
 }
 
+bool FdbOrch::ifChangeInformFdb(Port &port, bool if_up)
```
prsunny, line 340: I think the name looks better as `flushFdbEntries`

prsunny:
Looks like there is also another PR that handles this - https://github.com/Azure/sonic-swss/pull/1242/files. @vasant17 , can you review this PR?

orchagent/fdborch.cpp
```diff
@@ -186,17 +186,95 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
         }
         else if (bridge_port_id && entry->bv_id == SAI_NULL_OBJECT_ID)
         {
-            /*this is a placeholder for flush port fdb case, not supported yet.*/
-            SWSS_LOG_ERROR("FdbOrch notification: not supported flush port fdb action, port_id = 0x%" PRIx64 ", bv_id = 0x%" PRIx64 ".", bridge_port_id, entry->bv_id);
+            if (!m_portsOrch->getPortByBridgePortId(bridge_port_id, update.port))
```
vasant17, line 189: When we call SAI flush API, with bridge_port_id, we get SAI_FDB_EVENT_AGED event but NOT SAI_FDB_EVENT_FLUSHED. I did verify it by looking at logs. So, here is my understanding
1. If we call SAI flush API with bvid and bridge_port_id both set to SAI_NULL_OBJECT_ID, we get the notification SAI_FDB_EVENT_FLUSHED. So the new code you added wont be exericised as it will enter the first if case 
```
if (bridge_port_id == SAI_NULL_OBJECT_ID && entry->bv_id == SAI_NULL_OBJECT_ID)
```

2. When we call SAI FDB flush API with either bvid or bridge_port_id or both valid, we would end up receiving SAI_FDB_EVENT_AGED. Hence. Again the new code you added here wont be exercised.

orchagent/portsorch.cpp
```diff
@@ -3830,6 +3832,11 @@ void PortsOrch::updatePortOperStatus(Port &port, sai_port_oper_status_t status)
     {
         SWSS_LOG_WARN("Inform nexthop operation failed for interface %s", port.m_alias.c_str());
     }
+
+    if (!gFdbOrch->ifChangeInformFdb(port, isUp))
```
vasant17, line 3836: I agree that we are just following the way ifChangeInformNextHop is done and introducing ifChangeInformFdb. And I was coding up the same solution yesterday and thought of doing it a different way. That is, today at most places if NOT all, we interact between modules using observer pattern. For example, when port gets created or deleted, portsOrch notifies all the observers. I think we should leverage that or follow the pattern to notify observers about port state change. That way we dont have introduce a new function for every module that wants to port status change. With my new code change, we should be able to delete ifChangeInformNextHop and let neighborOrch handle port status change notification. Please refer PR #1242. I will push the code in couple of hours

orchagent/fdborch.cpp
vasant17: > When we call SAI flush API, with bridge_port_id, we get SAI_FDB_EVENT_AGED event but NOT SAI_FDB_EVENT_FLUSHED. I did verify it by looking at logs. So, here is my understanding
> 
> 1. If we call SAI flush API with bvid and bridge_port_id both set to SAI_NULL_OBJECT_ID, we get the notification SAI_FDB_EVENT_FLUSHED. So the new code you added wont be exericised as it will enter the first if case
> 
> ```
> if (bridge_port_id == SAI_NULL_OBJECT_ID && entry->bv_id == SAI_NULL_OBJECT_ID)
> ```
> 
> 1. When we call SAI FDB flush API with either bvid or bridge_port_id or both valid, we would end up receiving SAI_FDB_EVENT_AGED. Hence. Again the new code you added here wont be exercised.

Sorry, I take back my comment on receiving SAI_FDB_EVENT_AGED instead of instead of SAI_FDB_EVENT_FLUSHED. My observation was on one platform. But it was NOT the same on VS container

orchagent/fdborch.h
```diff
@@ -9,6 +9,7 @@ struct FdbEntry
 {
     MacAddress mac;
     sai_object_id_t bv_id;
+    std::string port_name;
```
vasant17, line 12: I liked this. This was much needed!

orchagent/fdborch.cpp
```diff
@@ -467,6 +581,12 @@ bool FdbOrch::addFdbEntry(const FdbEntry& entry, const string& port_name, const
         return true;
     }
 
+    /* Update entry's port name if needed */
+    if (entry.port_name.empty())
+    {
+        entry.port_name = port_name;
```
vasant17, line 587: In which case, the entry is created without port name? because

vasant17: We add an entry to m_entries in two places. One is storeFdbEntryState() and other is addFdbEntry(), Could you please make sure in both places we set port_name in entry before adding to m_entries?

orchagent/fdborch.cpp
```diff
@@ -429,7 +543,7 @@ void FdbOrch::updateVlanMember(const VlanMemberUpdate& update)
     auto fdb_list = std::move(saved_fdb_entries[port_name]);
     if(!fdb_list.empty())
     {
-        for (const auto& fdb: fdb_list)
+        for (auto& fdb: fdb_list)
```
vasant17, line 546: I think we can avoid making the mutable.

orchagent/fdborch.cpp
```diff
@@ -438,7 +552,7 @@ void FdbOrch::updateVlanMember(const VlanMemberUpdate& update)
     }
 }
 
-bool FdbOrch::addFdbEntry(const FdbEntry& entry, const string& port_name, const string& type)
+bool FdbOrch::addFdbEntry(FdbEntry& entry, const string& port_name, const string& type)
```
vasant17, line 555: Same here, I think, we shouldn't make entry mutable. Here is my suggestion:

Change prototype of addFdbEntry() by removing port from parameter list, as it's already part of FDB entry structure. As far as I remember addFdbEntry is called from only two places and you can populate entry.port_name in both places. In fact you don't have to do anything when it's called from  updateVlanMember().

Also note that you have to make some changes in storeFdbStateEntry(). 
It is called from only update() method in three cases. LEARN, AGED/MOVE, FLUSH. 

LEARN is the only case where storeFdbStateEntry() ends up adding the entry to m_entries/saved_fdb_entries. So, populate entry.port_name in LEARN case before calling storeFdbStateEntry(). 

And in the other two cases(AGED/MOVE, FLUSH), anyways we are going to remove the entry from m_entries, hence no need to populate entry.port_name. Also note that you can go ahead and cleanup(delete) few lines of port related code in storeFdbStateEntry().

prsunny:
@MikeHcChen , PR https://github.com/Azure/sonic-swss/pull/1242/files is merged which handles the same functionality. Can you close this PR?

MikeHcChen:
> @MikeHcChen , PR #1242 is merged which handles the same functionality. Can you close this PR?

OK! Thanks.

PR #1300, opened 2020-05-28
 Corrected the copp rule as per NAT HLD

AkhileshSamineni:
Updated the rate-limit value as per the "https://github.com/Azure/SONiC/blob/master/doc/nat/nat_design_spec.md"  in copp json file.

This PR is to address the issue - https://github.com/Azure/SONiC/issues/611

### DIFF
--- a/swssconfig/sample/00-copp.config.json
+++ b/swssconfig/sample/00-copp.config.json
@@ -43,8 +43,8 @@
         "OP": "SET"
     },
     {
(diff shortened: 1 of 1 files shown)


abdosi:
@rlhui with this change SAI-3.7.5.1 is not terminating for NAT policier.

arlakshm:
retest this please

PR #1285, opened 2020-04-30
[proxy_arp] Implement proxy ARP feature

volodymyrsamotiy:
Signed-off-by: Volodymyr Samotiy <volodymyrs@mellanox.com>

**What I did**
Implemented ARP proxy feature.

**Why I did it**
Feature request according to HLD https://github.com/Azure/SONiC/blob/master/doc/arp/Proxy%20Arp.md.

**How I verified it**
- Verified that it is possible to enable and disable proxy ARP on non VNET interfaces
  - Verified kernel configuration via ```proc``` FS
  - Verified VLAN flood SAI configuration for VLAN RIF
- Verified that proxy ARP is enabled by default on VNET interfaces
- Verified that ARP packets are propagated correctly when proxy ARP is enabled 

**Details if related**
N/A

PR #1302, opened 2020-05-28
[proxy_arp] Implement proxy ARP feature

volodymyrsamotiy:

**What I did**
Implemented ARP proxy feature.
(This PR is the same as https://github.com/Azure/sonic-swss/pull/1285, but  for ```201911``` branch)

**Why I did it**
Feature request according to HLD https://github.com/Azure/SONiC/blob/master/doc/arp/Proxy%20Arp.md.

**How I verified it**
- Verified that it is possible to enable and disable proxy ARP on non VNET interfaces
  - Verified kernel configuration via ```proc``` FS
  - Verified VLAN flood SAI configuration for VLAN RIF
- Verified that proxy ARP is enabled by default on VNET interfaces
- Verified that ARP packets are propagated correctly when proxy ARP is enabled 

**Details if related**
N/A

### DIFF
--- a/cfgmgr/intfmgr.cpp
+++ b/cfgmgr/intfmgr.cpp
@@ -344,6 +344,33 @@ void IntfMgr::removeSubIntfState(const string &alias)
     }
 }
 
+bool IntfMgr::setIntfProxyArp(const string &alias, const string &proxy_arp)
+{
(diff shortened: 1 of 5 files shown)


prsunny:
retest this please

PR #1303, opened 2020-05-29
[MultiDB] Use database name instead of database ID

rajendra-dendukuri:
Remove references to hardcoded DB ID in swss as in a MultiDB, the DB Name
should always be used to accurately identify the connector's database role.

**What I did**
Use database name as the key to match a database instead of DB ID. In a MultiDB environment,
the db ID can be changed in database_config.json

**Why I did it**
Modify default DB ID's in database_config.json

**How I verified it**
Swap APPL_DB and CONFIG_DB IDs in database_config.json and reboot. Ensure that switch comes
back up and ports are created.

**Details if related**
 Use database name instead of database ID

### DIFF
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -557,7 +557,7 @@ bool Orch::parseIndexRange(const string &input, sai_uint32_t &range_low, sai_uin
 
 void Orch::addConsumer(DBConnector *db, string tableName, int pri)
 {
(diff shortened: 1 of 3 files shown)


rajendra-dendukuri:
@qiluo-msft @dzhangalibaba Please review

dzhangalibaba:
@qiluo-msft we need this fix as well

PR #1304, opened 2020-05-30
Fix LAG member test case

vasant17:
**What I did**
Fixed LAG memeber test case and ran the VS test case to ensure it passes.
Test was expecting only two fields for LAG member where actual is 4
```
(('SAI_LAG_MEMBER_ATTR_LAG_ID', 'oid:0x20000000005c4'), ('SAI_LAG_MEMBER_ATTR_PORT_ID', 'oid:0x1000000000002'), ('SAI_LAG_MEMBER_ATTR_INGRESS_DISABLE', 'false'), ('SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE', 'false'))
```
**Why I did it**
Test case was broken.
**How I verified it**
By running VS test case
**Details if related**
```

test_vlan.py::TestVlan::test_AddPortChannelToVlan PASSED                                                                                                                                                                 [100%]
 
```

### DIFF
--- a/tests/test_vlan.py
+++ b/tests/test_vlan.py
@@ -447,13 +447,10 @@ class TestVlan(object):
         assert len(lag_member_entries) == 1
 
         (status, fvs) = tbl.get(lag_member_entries[0])
-        for fv in fvs:
-            if fv[0] == "SAI_LAG_MEMBER_ATTR_LAG_ID":
-                assert fv[1] == lag_entries[0]
-            elif fv[0] == "SAI_LAG_MEMBER_ATTR_PORT_ID":
-                assert dvs.asicdb.portoidmap[fv[1]] == "Ethernet0"
-            else:
-                assert False
+        fvs = dict(fvs)
+        assert len(fvs) == 4
+        assert fvs.get("SAI_LAG_MEMBER_ATTR_LAG_ID") == lag_entries[0]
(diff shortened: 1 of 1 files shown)


tests/test_vlan.py
```diff
@@ -447,13 +447,10 @@ def test_AddPortChannelToVlan(self, dvs, testlog):
...
         (status, fvs) = tbl.get(lag_member_entries[0])
-        for fv in fvs:
-            if fv[0] == "SAI_LAG_MEMBER_ATTR_LAG_ID":
-                assert fv[1] == lag_entries[0]
-            elif fv[0] == "SAI_LAG_MEMBER_ATTR_PORT_ID":
-                assert dvs.asicdb.portoidmap[fv[1]] == "Ethernet0"
-            else:
-                assert False
+        fvs = dict(fvs)
+        assert len(fvs) == 4
```
prsunny, line 451: IMO, I dont think you should check for `len(fvs) == 4(hardcoded)` but check for individual attributes in the `if` case. @daall , any suggestions?

daall: It's currently hard-coded in the master as well, but I do agree that it is better to check for individual attributes. I think that can be taken up as a separate PR though since this issue also exists in master.

PR #1305, opened 2020-05-31
Add more log message, fix test code (#1239)

vasant17:
**What I did**
Cherry-picked a test case fix from master
**Why I did it**
To make the test case pass in 201911
**How I verified it**

**Details if related**


qiluo-msft:
I think this one could be easily clean cherry-picked, so no need to submit another PR. I marked it as 'request_for_201911_branch'

PR #1314, opened 2020-06-04
[mirrororch] Port Mirroring implementation

rupesh-k:
Signed-off-by: Rupesh Kumar <rupesh-k.kumar@broadcom.com>

**What I did**

Port mirroring implementation in swss.
HLD @ https://github.com/Azure/SONiC/pull/580

**Why I did it**
To support port mirroring in SONiC

**How I verified it**
Verified using pytests and manual UT. 

**Details if related**
pytests will updated in next commit.


orchagent/mirrororch.cpp
```diff
@@ -613,6 +671,83 @@ bool MirrorOrch::updateSession(const string& name, MirrorEntry& session)
...
+}
+
+bool MirrorOrch::configurePortMirrorSession(const string& name, MirrorEntry& session, bool set)
+{
+    auto ports = tokenize(session.src_port, ',');
+    if (ports.size() != 0)
+    {
+        for (auto alias : ports)
+        {
+            Port port;
```
kperumalbfn, line 729: should we continue or return failure when one of the port/lag mirror session set fails?

orchagent/mirrororch.cpp
```diff
@@ -727,6 +882,18 @@ bool MirrorOrch::activateSession(const string& name, MirrorEntry& session)
     }
 
     session.status = true;
+
+    if (!session.src_port.empty() && !session.direction.empty())
+    {
+        status = configurePortMirrorSession(name, session, true);
+        if (status == false)
+        {
+            SWSS_LOG_ERROR("Failed to activate port mirror session %s\n", name.c_str());
```
kperumalbfn, line 891: port/lag in the log message.

rupesh-k: it is taken care in configurePortMirrorSession. I can add here too.  Port and LAG, LAG member port info is logged.
 SWSS_LOG_ERROR("Failed to configure mirror session port %s: %s, status %d, sessionId %x\n",
                                 port.m_alias.c_str(), p.m_alias.c_str(), status, sessionId);

orchagent/mirrororch.cpp
rupesh-k: Currently it returns when one of the config fails from SAI. The failures from SAI are critical and orchagent terminate is initiated from syncd. I am not sure if continue will help. Please suggest.

orchagent/mirrororch.cpp
```diff
@@ -613,6 +671,83 @@ bool MirrorOrch::updateSession(const string& name, MirrorEntry& session)
...
+    {
+        port_attr.value.objlist.count = 0;
+    }
+
+    if (port.m_type == Port::LAG)
+    {
+        vector<Port> portv;
+        m_portsOrch->getLagMember(port, portv);
+        for (const auto p : portv)
+        {
```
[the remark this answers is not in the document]
rupesh-k: Sure. will take care.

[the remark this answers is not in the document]
rupesh-k: And the port here is coming from internal LAG port . do u see extra validation is needed?
I just saw the other usage of this function in portsorch.cpp and don't see any validation.
So Please suggest. I can take care based on your response.

orchagent/mirrororch.cpp
kperumalbfn: There will be inconsistency for some cases, 

- Mirror on both the directions, when port ingress mirror update succeeds and port egress mirror update fails, there will be inconsistency in the config DB and asic DB.
- Mirror on LAG, when mirror session set/unset fails on some members, there will be inconsistency on some LAG members. But the actual config is for LAG and not the members.

We could go by a common model. 

@lguohan Any thoughts??

rupesh-k: I see similar behaviour in  setHostIntfsStripTag in orchagent/portsorch.cpp 
When the port is LAG and the SAI API (set_hostif_attribute) fails for any port in LAG, 
then the function returns failure to the caller.

rupesh-k:
Hi @daall 

Can u please help with the review. I hope we can make this feature for 202006 release :)

Thanks

orchagent/mirrororch.cpp
kperumalbfn: Yeah, we have similar inconsistency in other modules as well.

daall:
> Hi @daall
> 
> Can u please help with the review. I hope we can make this feature for 202006 release :)
> 
> Thanks

ack, will review this + utilities today :)

orchagent/mirrororch.cpp
```diff
@@ -324,6 +329,48 @@ void MirrorOrch::createEntry(const string& key, const vector<FieldValueTuple>& d
...
+                        if (!gPortsOrch->getPort(alias, port))
+                        {
+                            SWSS_LOG_ERROR("Failed to locate port/LAG %s", alias.c_str());
+                            return;
+                        }
+                    }
+                }
+                entry.src_port = fvValue(i);
+            }
+            else if (fvField(i) == MIRROR_SESSION_DST_PORT)
```
daall, line 349: For both this and SRC_PORT can we verify that the provided ports are either a physical port or LAG (e.g. not a VLAN or VRF)?

orchagent/mirrororch.cpp
daall: @lguohan @prsunny do you have thoughts for this?

orchagent/mirrororch.cpp
```diff
@@ -613,6 +671,83 @@ bool MirrorOrch::updateSession(const string& name, MirrorEntry& session)
...
+            status = sai_port_api->set_port_attribute(p.m_port_id, &port_attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to configure mirror session port %s: %s, status %d, sessionId %x\n",
+                                port.m_alias.c_str(), p.m_alias.c_str(), status, sessionId);
+                return false;
+            }
+        }
+    }
+    else if (port.m_type == Port::PHY)
```
daall, line 709: I asked this up here too - but what happens if the user specifies a PHY port that is a LAG member?

tests/test_mirror_port_erspan.py
```diff
@@ -0,0 +1,951 @@
...
+
+        # remove port channel member; remove port channel
+        self.remove_port_channel_member("008", "Ethernet88")
+        self.remove_port_channel(dvs, "008")
+        self.remove_port_channel_member("001", src_port1)
+        self.remove_port_channel_member("001", src_port2)
+        self.remove_port_channel(dvs, "001")
+
+        # remove mirror session
+        self.remove_mirror_session(session)
```
daall, line 951: Missing newline at the end of this file.

tests/test_mirror_port_erspan.py
```diff
@@ -0,0 +1,951 @@
+# This test suite covers the functionality of mirror feature in SwSS
```
daall, line 1: I think the overall test logic/flow is good, I would just ask if you could use some of the new library functions we have over in `dvslib/dvs_database` and `dvslib/dvs_common` for interacting with the databases. Basically, there are a bunch of functions you can use to poll the database rather than using `time.sleep(1)` throughout. This makes the PR tests a lot more stable and it also speeds up the execution by quite a bit.

There are examples in `test_acl`, `test_vlan`, `test_dpb...`, and `test_sub_port_intf` to name a few. There are also docstrings describing how the methods work as well. I'm also happy to answer any questions if you run into issues. :)

tests/test_mirror_port_span.py
```diff
@@ -0,0 +1,762 @@
+# This test suite covers the functionality of mirror feature in SwSS
```
daall, line 1: See comment for `test_mirror_port_erspan`, take a look at `dvslib` and let me know if you have any questions!

orchagent/mirrororch.cpp
rupesh-k: Have taken care of this

doc/swss-schema.md
```diff
@@ -595,6 +595,10 @@ Stores information about mirror sessions and their properties.
     ttl       = h8                    ; Session TTL
     queue     = h8                    ; Session output queue
     policer   = policer_name          ; Session policer name
+    dst_port  = PORT_TABLE|ifname     ; Session destination PORT
+    src_port  = PORT_TABLE|ifname     ; Session source PORT/LAG list
+    direction = "RX"/"TX"/"BOTH"      ; Session direction
+    type      = "SPAN"/"ERSPAN"       ; Session type. Default is ERSPAN
```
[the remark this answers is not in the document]
rupesh-k: Yes. sorry I didnt get question :) . U want to capture anything here ?

orchagent/mirrororch.h
```diff
@@ -18,6 +18,12 @@
 #include <map>
 #include <inttypes.h>
 
+const string mirror_rx_direction                 = "RX";
+const string mirror_tx_direction                 = "TX";
+const string mirror_both_direction               = "BOTH";
+const string mirror_session_span                 = "SPAN";
+const string mirror_session_erspan               = "ERSPAN";
```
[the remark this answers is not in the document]
rupesh-k: Done.

orchagent/mirrororch.cpp
```diff
@@ -324,6 +329,48 @@ void MirrorOrch::createEntry(const string& key, const vector<FieldValueTuple>& d
                 m_policerOrch->increaseRefCount(fvValue(i));
                 entry.policer = fvValue(i);
             }
+            else if (fvField(i) == MIRROR_SESSION_SRC_PORT)
+            {
+                auto ports = tokenize(fvValue(i), ',');
+                if (ports.size() != 0)
```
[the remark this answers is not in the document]
rupesh-k: No. LAG members cant be part of this list
I handled the cases where LAG port is also in source port list in utilities and swss.

orchagent/mirrororch.cpp
rupesh-k: Done. Added validation.

tests/test_mirror_port_erspan.py
rupesh-k: Done.

orchagent/mirrororch.cpp
rupesh-k: No. LAG members cant be part of this list
I handled the cases where LAG port is also in source port list.. 

In swss:
	1. check port is part of Port channel and in src_port list.
        2. for lag update events, skip programming when the port is already part of mirror session (not possible case, but error check)

tests/test_mirror_port_erspan.py
rupesh-k: Thanks a lot for this suggestion. I have taken care this.
Code looks clean and much faster after this change.

tests/test_mirror_port_span.py
rupesh-k: Done. taken care.

rupesh-k:
> I think overall this looks pretty good, I just have a few clarifying questions about edge cases and some broad suggestions for the test code.

hi @daall ,

Thanks a lot for the review and good suggestions, I have taken care of the changes. Can you please help with review again.

Thanks

orchagent/mirrororch.cpp
daall: So I see both of those checks in the code, but let's say that the src port list only has PHY ports. Do we ever check if any of those PHY ports are LAG members?

rupesh-k: No. we dont check that. 
Its up to user when he wants to mirror only specific ports of LAG, its allowed. 
Do u see any issue with that ?

daall: Ohhh, okay! I think that is OK then. Just to clarify, you can have LAGs OR LAG members, you just can't mix both in the same src port list?

rupesh-k: Yes. thats correct.

tests/dvslib/dvs_acl.py
```diff
@@ -136,6 +136,16 @@ def create_acl_rule(self, table_name, rule_name, qualifiers, action="FORWARD", p
 
         self.config_db.create_entry("ACL_RULE", "{}|{}".format(table_name, rule_name), fvs)
 
+    def create_mirror_acl_rule(self, table_name, rule_name, qualifiers, priority="2020"):
```
daall, line 139: This is a more generic version of create_acl_rule. What I would do is make this method take a mirror stage and then automatically add the MIRROR_ACTION qualifier.

What I want to do next is make this generic method into a helper and then provide `create_l3_acl_rule` (which will replace the existing `create_acl_rule` method) and `create_mirror_acl_rule`, which will both use this helper under the hood. However, this change would touch a lot of files so I'm not going to ask you to do that here. :)

tests/dvslib/dvs_mirror.py
```diff
@@ -0,0 +1,100 @@
...
+        self.app_db = appdb
+
+    def create_span_session(self, name, dst_port, src_ports=None, direction="BOTH", queue=None, policer=None):
+        mirror_entry = {"type": "SPAN"}
+        if dst_port:
+            mirror_entry["dst_port"] = dst_port
+
+        if src_ports:
+            mirror_entry["src_port"] = src_ports
+        # set direction without source port to uncover any swss issues.
```
daall, line 16: Not sure I understand this, what are we trying to uncover?

tests/dvslib/dvs_mirror.py
```diff
@@ -0,0 +1,100 @@
...
+        mirror_entry["direction"] = direction
+
+        if queue:
+            mirror_entry["queue"] = queue
+        if policer:
+            mirror_entry["policer"] = policer
+        self.config_db.create_entry("MIRROR_SESSION", name, mirror_entry)
+
+    def create_erspan_session(self, name, src, dst, gre, dscp, ttl, queue, policer=None, src_ports=None, direction="BOTH"):
+        mirror_entry = {"type": "ERSPAN"}
```
daall, line 26: Do we have a way of testing that orchagent defaults to ERSPAN if no type is specified in the Config DB?

tests/dvslib/dvs_mirror.py
```diff
@@ -0,0 +1,100 @@
...
+            if asic_size:
+                assert asic_size == len(fv_pairs)
+        if state:
+            fv_pairs = dvs.state_db.get_entry("MIRROR_SESSION_TABLE", name)
+            assert all(fv_pairs.get(k) == v for k, v in state.items())
+
+    def verify_session_policer(self, dvs, policer_oid, cir):
+        if cir:
+            entry = dvs.asic_db.get_entry("ASIC_STATE:SAI_OBJECT_TYPE_POLICER", policer_oid)
+            assert entry["SAI_POLICER_ATTR_CIR"] == cir
```
daall, line 81: I would probably still use `wait` methods for these even if the polling is covered by other methods in the existing code. If someone tries to use one of these methods elsewhere it may be difficult to diagnose the failures. :)

tests/test_mirror_port_erspan.py
```diff
@@ -0,0 +1,550 @@
...
+        dvs.set_interface_status("Ethernet4", "down")
+        dvs.set_interface_status(vlan, "down")
+
+        # remove mirror session
+        self.dvs_mirror.remove_mirror_session(session)
+        self.dvs_mirror.verify_no_mirror()
+
+        # remove vlan member; remove vlan
+        self.dvs_vlan.remove_vlan_member(vlan_id, "Ethernet4")
+        self.dvs_vlan.remove_vlan(vlan_id)
```
daall, line 181: In general I recommend checking that there's no vlan/no vlan member at the end of the test case just to ensure that the test case leaves the system in a clean state.

tests/test_mirror_port_erspan.py
```diff
@@ -0,0 +1,550 @@
...
+        self.dvs_mirror.verify_session_status(session, status="inactive")
+
+        # bring down port channel and port channel member
+        dvs.set_interface_status("PortChannel008", "down")
+        dvs.set_interface_status("Ethernet88", "down")
+
+        # remove port channel member; remove port channel
+
+        self.dvs_lag.remove_port_channel_member("008", "Ethernet88")
+        self.dvs_lag.remove_port_channel("008")
```
daall, line 254: Same for LAGs, it's a good idea just to confirm they have been cleaned up at the end of the test case before moving on (although if there is no `verify_no_lag_*` methods like there are for VLANs at the moment I think you can leave this as is for now).

tests/test_mirror_port_span.py
```diff
@@ -0,0 +1,478 @@
...
+        self.dvs_mirror.verify_session(dvs, session1, dst_oid=dst_oid1, asic_db=a1, src_ports=src_asic_ports1, expected = 2, policer=policer1)
+        self.dvs_mirror.verify_session(dvs, session2, dst_oid=dst_oid2, asic_db=a2, src_ports=src_asic_ports2, expected = 2, policer=policer2)
+        self.dvs_mirror.remove_mirror_session(session1)
+        self.dvs_mirror.remove_mirror_session(session2)
+        self.dvs_policer.remove_policer(policer1)
+        self.dvs_policer.remove_policer(policer2)
+        self.dvs_policer.verify_no_policer()
+        self.dvs_mirror.verify_no_mirror()
+
+    def test_LAGMirorrSpanAddRemove(self, dvs, testlog):
```
daall, line 230:
```suggestion
    def test_LAGMirrorSpanAddRemove(self, dvs, testlog):
```

typo

tests/test_mirror_port_span.py
```diff
@@ -0,0 +1,478 @@
...
+        self.dvs_lag.remove_port_channel_member("008", "Ethernet4")
+        self.dvs_vlan.asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_LAG_MEMBER", 1)
+        src_asic_ports = ["Ethernet8"]
+        self.dvs_mirror.verify_session(dvs, session, src_ports=src_asic_ports)
+        
+        #cleanup
+        self.dvs_mirror.remove_mirror_session(session)
+        self.dvs_mirror.verify_no_mirror()
+        self.dvs_lag.remove_port_channel_member("008", "Ethernet8")
+        self.dvs_vlan.asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_LAG_MEMBER", 0)
```
daall, line 475: If there are no `verify_no_lag_members` or `verify_no_lags` methods yet I think these would probably do. :)

tests/dvslib/dvs_mirror.py
rupesh-k: Ok. makes sense. Will do. I thought if the session is active, rest all will be in place, but as u said, if someone else is using then it may cause failures.

tests/dvslib/dvs_mirror.py
rupesh-k: In mirrororch we check both src/direction is set before configure port mirror session.
If direction is only set then it shouldnt do any port mirror config and also session has to be active.(based on other conditions)
This setting just makes sure that setting direction only doesn't break anything :)

orchagent/mirrororch.cpp
```diff
@@ -260,6 +265,76 @@ bool MirrorOrch::decreaseRefCount(const string& name)
     return true;
 }
 
+bool MirrorOrch::validateDstPort(const string& dstPort)
+{
+    Port port;
+    if (!m_portsOrch->getPort(dstPort, port) && port.m_type != Port::PHY)
+    {
+        SWSS_LOG_ERROR("Failed to locate port %s", dstPort.c_str());
```
prsunny, line 273: I think the log message should also specify if type is not `phy`. Can you add that to log and print type?

orchagent/mirrororch.cpp
```diff
@@ -260,6 +265,76 @@ bool MirrorOrch::decreaseRefCount(const string& name)
...
+        }
+    }
+
+    return false;
+}
+
+bool MirrorOrch::validateSrcPort(const string& srcPort)
+{
+    auto ports = tokenize(srcPort, ',');
+
```
prsunny, line 299: Delete this line, to be consistent with above function

orchagent/mirrororch.cpp
```diff
@@ -613,6 +731,92 @@ bool MirrorOrch::updateSession(const string& name, MirrorEntry& session)
...
+                return false;
+            }
+        }
+    }
+    else if (port.m_type == Port::PHY)
+    {
+        status = sai_port_api->set_port_attribute(port.m_port_id, &port_attr);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("Failed to configure %s session on port %s, status %d, sessionId %x\n",
```
prsunny, line 775: You've used `\n` inconsistently. Some places have and some not.` \n` is not required, kindly remove

orchagent/mirrororch.cpp
prsunny: This would be tricky. The actual return value of the function is not checked. One suggestion is, if there is a failure, you can invoke the deactivate session and exit cleanly. Or you can log the error and expect user to delete the session.  Have you checked the flow for "unsetting" session on a port which is not originally "set"?

rupesh-k: Sorry I didn't get which function return value is not handled here. 
configurePortMirrorSession return value is already handled and session becomes inactive when some config fails on any port, and user can delete the session.
Setting a port which is already set gives error SAI_STATUS_ITEM_ALREADY_EXISTS
Unsetting a port which is not set doesnt give any error.

We are already logging error and notifying user about failure.  As i mentioned earlier, there are other codes which follow similar approach as we are doing here.

Please let me know if u have strong preference for calling deactivate session on failure. 
We need to handle both set and unset differently, I will take care in next PR.

prsunny: So typically, there is a retry logic in `orchagent ` that if something fails, it keeps retrying (if you return `false` to `doTask`). What I meant was, this function returns to `activateSession `and subsequent return value is ignored. So essentially, if something fails, there is no retry. Agree with the current approach, you can enhance it in next PR.

rupesh-k: Thank you.

tests/dvslib/dvs_mirror.py
rupesh-k: Modified to not set type to ERSPAN.

tests/dvslib/dvs_mirror.py
rupesh-k: Done. Have taken care.

tests/test_mirror_port_erspan.py
rupesh-k: Done. Added checks.

tests/test_mirror_port_erspan.py
rupesh-k: Done. Added checks.

tests/test_mirror_port_span.py
rupesh-k: Added these new methods.

orchagent/mirrororch.cpp
rupesh-k: Done. Added log

orchagent/mirrororch.cpp
rupesh-k: This is source port list, so modified function name to reflect it.

orchagent/mirrororch.cpp
rupesh-k: Yes. taken care of this. There were some existing code also wrongly doing .fixed those too.

tests/test_mirror_port_span.py
rupesh-k: Done.

amin-alavi:
@prsunny, @kperumalbfn, @daall , Can this PR be merged and any further updates will be made in 202006 branch?

Same for https://github.com/Azure/sonic-utilities/pull/936

Thanks!

rupesh-k:
Thanks @daall @prsunny for review. I have taken care of all comments. Please review.

daall:
retest this please

orchagent/mirrororch.cpp
```diff
@@ -1044,6 +1222,24 @@ void MirrorOrch::updateLagMember(const LagMemberUpdate& update)
         const auto& name = it->first;
         auto& session = it->second;
 
+        // Check the following conditions:
```
FaithRobben, line 1304: if mirror session src port is port+lag,  then lag member update to null, the mirror session status will set to inactive ？

PR #1315, opened 2020-06-04
[fgnhgorch] Changes for Fine Grained ECMP

anish-n:
**What I did**
Added fine grained ECMP changes.

**Why I did it**
To support consistent and layered hashing in SONiC.

**How I verified it**

**Details if related**
HLD: https://github.com/Azure/SONiC/blob/master/doc/ecmp/fine_grained_next_hop_hld.md
SAI: https://github.com/opencomputeproject/SAI/blob/master/doc/ECMP/Ordered_and_Fine_Grained_ECMP.md

### DIFF
--- a/orchagent/fgnhgorch.cpp
+++ b/orchagent/fgnhgorch.cpp
@@ -848,19 +848,19 @@ bool FgNhgOrch::doTaskFgNhg(const KeyOpFieldsValuesTuple & t)
         }
         else 
         {
-        	/* delete all associated FG_NHG and SAI objects */
-        	if (fgNhg_entry->second.prefixes.size() == 0 && fgNhg_entry->second.nextHops.size() == 0)
-        	{
-        		m_FgNhgs.erase(fgNhg_entry);
-        		assert(m_FgNhgs.find(fgNhg_name) == fgNhgPrefixes.end());
-        		SWSS_LOG_INFO("%s: Received delete call for valid entry, deleting all associated FG_NHG and SAI objects %s",
-        				__FUNCTION__, fgNhg_name.c_str());
-        	}
-        	else
-        	{
-				SWSS_LOG_NOTICE("%s: still prefixes and nexthop in FG_NHG group %s", __FUNCTION__, fgNhg_name.c_str());
-				return false;
-        	}
+            /* delete all associated FG_NHG and SAI objects */
+            if (fgNhg_entry->second.prefixes.size() == 0 && fgNhg_entry->second.nextHops.size() == 0)
+            {
+                m_FgNhgs.erase(fgNhg_entry);
+                assert(m_FgNhgs.find(fgNhg_name) == fgNhgPrefixes.end());
+                SWSS_LOG_INFO("%s: Received delete call for valid entry, deleting all associated FG_NHG and SAI objects %s",
+                        __FUNCTION__, fgNhg_name.c_str());
+            }
+            else
+            {
+                SWSS_LOG_NOTICE("%s: still prefixes and nexthop in FG_NHG group %s", __FUNCTION__, fgNhg_name.c_str());
+                return false;
+            }
         }
     }
     return true;
@@ -906,34 +906,35 @@ bool FgNhgOrch::doTaskFgNhg_prefix(const KeyOpFieldsValuesTuple & t)
         }
         else 
         {
-        	/* delete regular ecmp handling for prefix */
-			sai_object_id_t vrf_id = gVirtualRouterId;
-			auto route_table_entry = gRouteOrch->getSyncdRoutes().find(vrf_id);
-			NextHopGroupKey nhg;
-			if (route_table_entry == gRouteOrch->getSyncdRoutes().end())
-			{
-				SWSS_LOG_INFO("Failed to find route table, vrf_id 0x%lx\n", vrf_id);
-				return false;
-			}
+            /* delete regular ecmp handling for prefix */
+            sai_object_id_t vrf_id = gVirtualRouterId;
+            auto route_table_entry = gRouteOrch->getSyncdRoutes().find(vrf_id);
+            NextHopGroupKey nhg;
+            if (route_table_entry == gRouteOrch->getSyncdRoutes().end())
+            {
+                SWSS_LOG_INFO("Failed to find route table, vrf_id 0x%lx\n", vrf_id);
+            }
+            else
+            {
+                /* ipprefix already exist in routeorch */
+                auto it_route = route_table_entry->second.find(ip_prefix);
+                if (it_route != route_table_entry->second.end())
+                {
+                    nhg = it_route->second;
+                    gRouteOrch->removeRoute(vrf_id, ip_prefix);
+                    route_handled = true;
+                }
+            }
 
-			/* ipprefix already exist in routeorch */
-			auto it_route = route_table_entry->second.find(ip_prefix);
-			if (it_route != route_table_entry->second.end())
-			{
-				nhg = it_route->second;
-				gRouteOrch->removeRoute(vrf_id, ip_prefix);
-				route_handled = true;
-			}
-
-        	fgNhg_entry->second.prefixes.push_back(ip_prefix);
+            fgNhg_entry->second.prefixes.push_back(ip_prefix);
             fgNhgPrefixes[ip_prefix] = &(fgNhg_entry->second);
 
             /* add fgnhg route */
             if (route_handled)
             {
-            	addRoute(vrf_id, ip_prefix, nhg);
-            	SWSS_LOG_INFO("%s: Add route with ip prefix %s with nexthop group key %s", __FUNCTION__, ip_prefix.to_string().c_str(),
-            			nhg.to_string().c_str());
+                addRoute(vrf_id, ip_prefix, nhg);
+                SWSS_LOG_INFO("%s: Add route with ip prefix %s with nexthop group key %s", __FUNCTION__, ip_prefix.to_string().c_str(),
+                    nhg.to_string().c_str());
             }
 
             SWSS_LOG_INFO("%s FG_NHG added for group %s, prefix %s",
@@ -950,36 +951,36 @@ bool FgNhgOrch::doTaskFgNhg_prefix(const KeyOpFieldsValuesTuple & t)
         }
         else
         {
-        	/* remove fgnhg route entry */
-        	sai_object_id_t vrf_id = gVirtualRouterId;
-        	NextHopGroupKey nhg;
-        	if (m_syncdFGRouteTables.find(vrf_id) != m_syncdFGRouteTables.end() &&
-        			m_syncdFGRouteTables.at(vrf_id).find(ip_prefix) != m_syncdFGRouteTables.at(vrf_id).end())
-        	{
-        		nhg = m_syncdFGRouteTables.at(vrf_id).at(ip_prefix).nhg_key;
-        		removeRoute(vrf_id, ip_prefix);
-        		route_handled = true;
-        		SWSS_LOG_INFO("%s FG_NHG prefix %s is removed from SAI",
-        				__FUNCTION__, ip_prefix.to_string().c_str());
-        	}
+            /* remove fgnhg route entry */
+            sai_object_id_t vrf_id = gVirtualRouterId;
+            NextHopGroupKey nhg;
+            if (m_syncdFGRouteTables.find(vrf_id) != m_syncdFGRouteTables.end() &&
+                    m_syncdFGRouteTables.at(vrf_id).find(ip_prefix) != m_syncdFGRouteTables.at(vrf_id).end())
+            {
+                nhg = m_syncdFGRouteTables.at(vrf_id).at(ip_prefix).nhg_key;
+                removeRoute(vrf_id, ip_prefix);
+                route_handled = true;
+                SWSS_LOG_INFO("%s FG_NHG prefix %s is removed from SAI",
+                        __FUNCTION__, ip_prefix.to_string().c_str());
+            }
 
-        	/* search and delete local structure */
-        	for (uint32_t i = 0; i < prefix_entry->second->prefixes.size(); i++)
-        	{
-        		if (prefix_entry->second->prefixes[i] == ip_prefix)
-        		{
-        			prefix_entry->second->prefixes.erase(prefix_entry->second->prefixes.begin() + i);
-        			SWSS_LOG_INFO("%s FG_NHG prefix %s is deleted from group %s",
-        					__FUNCTION__, ip_prefix.to_string().c_str(), fgNhgPrefixes[ip_prefix]->fgNhg_name.c_str());
-        			break;
-        		}
-        	}
+            /* search and delete local structure */
+            for (uint32_t i = 0; i < prefix_entry->second->prefixes.size(); i++)
+            {
+                if (prefix_entry->second->prefixes[i] == ip_prefix)
+                {
+                    prefix_entry->second->prefixes.erase(prefix_entry->second->prefixes.begin() + i);
+                    SWSS_LOG_INFO("%s FG_NHG prefix %s is deleted from group %s",
+                            __FUNCTION__, ip_prefix.to_string().c_str(), fgNhgPrefixes[ip_prefix]->fgNhg_name.c_str());
+                    break;
+                }
+            }
             fgNhgPrefixes.erase(ip_prefix);
 
             /* reassign routeorch as the owner of the prefix and call for routeorch to add this route */
             if (route_handled)
             {
-            	gRouteOrch->addRoute(vrf_id, ip_prefix, nhg);
+                gRouteOrch->addRoute(vrf_id, ip_prefix, nhg);
             }
             /* TODO: query next-hop ecmp group and delete if no more prefixes point to it */
         }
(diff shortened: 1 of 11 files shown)


orchagent/fgnhgorch.h
```diff
@@ -0,0 +1,114 @@
...
+typedef std::vector<sai_object_id_t> FGNextHopGroupMembers;
+typedef std::vector<uint32_t> HashBuckets;
+typedef std::map<NextHopKey, HashBuckets> FGNextHopGroupMap;
+typedef std::vector<FGNextHopGroupMap> BankFGNextHopGroupMap;
+typedef std::map<Bank,Bank> InactiveBankMapsToBank;
+
+struct FGNextHopGroupEntry
+{
+    sai_object_id_t         next_hop_group_id;      // next hop group id
+    int                     ref_count;              // reference count
```
abdosi, line 30: can it be less than 0 ? If not better to use uint32_t

anish-n: Thanks for the comment. Don't use ref_count in the code, so removed this.

tests/test_fgnhg.py
```diff
@@ -0,0 +1,947 @@
...
+
+
+def create_entry_tbl(db, table, separator, key, pairs):
+    tbl = swsscommon.Table(db, table)
+    create_entry(tbl, key, pairs)
+
+def remove_entry_tbl(db, table, key):
+    tbl = swsscommon.Table(db, table)
+    tbl._del(key)
+    time.sleep(1)
```
lguohan, line 22: we should not use such sleep. it is causing the test to be unstable.

tests/test_fgnhg.py
```diff
@@ -0,0 +1,947 @@
...
+        fvs = swsscommon.FieldValuePairs([("nexthop","10.0.0.7,10.0.0.11"), ("ifname", "Ethernet12,Ethernet20")])
+        ps.set(fg_nhg_prefix, fvs)
+        time.sleep(1)
+        nh_memb_exp_count = {"10.0.0.7":64,"10.0.0.11":64}
+        verify_programmed_nh_membs(adb,nh_memb_exp_count,nh_oid_map,nhgid,bucket_size)
+
+
+        fvs = swsscommon.FieldValuePairs([("nexthop","10.0.0.5,10.0.0.7,10.0.0.9"), ("ifname", "Ethernet8,Ethernet12,Ethernet16")])
+        ps.set(fg_nhg_prefix, fvs)
+        time.sleep(1)
```
lguohan, line 915: please use the new framework to validate asic db.

tests/test_fgnhg.py
```diff
@@ -0,0 +1,947 @@
...
+        nh_memb_exp_count = {"10.0.0.5":43,"10.0.0.7":43,"10.0.0.9":42}
+        verify_programmed_nh_membs(adb,nh_memb_exp_count,nh_oid_map,nhgid,bucket_size)
+
+
+        fvs = swsscommon.FieldValuePairs([("nexthop","10.0.0.1,10.0.0.3,10.0.0.5,10.0.0.7,10.0.0.9,10.0.0.11"), ("ifname", "Ethernet0,Ethernet4,Ethernet8,Ethernet12,Ethernet16,Ethernet20")])
+        ps.set(fg_nhg_prefix, fvs)
+        time.sleep(1)
```
lguohan, line 922: please use the new framework to validate asic db.

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
+        Orch(db, tableNames),
+        m_neighOrch(neighOrch),
+        m_intfsOrch(intfsOrch),
+        m_vrfOrch(vrfOrch),
+		m_stateWarmRestartRouteTable(stateDb, STATE_FG_ROUTE_TABLE_NAME)
+{
+     SWSS_LOG_ENTER();
+}
+
+void calculate_bank_hash_bucket_start_indices(FgNhgEntry *fgNhgEntry)
```
prsunny, line 31: Please add comments to explain the function and how algorithm works

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
+        fvs.push_back(FieldValueTuple(std::to_string(index), nextHop.to_string()));
+        SWSS_LOG_INFO("Add new next hop entry %s with index %d for ip prefix %s",
+                nextHop.to_string().c_str(), index, ipPrefix.to_string().c_str());
+        m_stateWarmRestartRouteTable.set(key, fvs);
+    }
+
+}
+
+
+bool FgNhgOrch::write_hash_bucket_change_to_sai(FGNextHopGroupEntry *syncd_fg_route_entry, uint32_t index, sai_object_id_t nh_oid,
```
prsunny, line 109: Please rename the function. Suggest not use "sai" in API.

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
+        for( auto &route_table : route_tables.second )
+        {
+            if(!(route_table.second.nhg_key.contains(nexthop)))
+            {
+                continue;
+            }
+
+            FGNextHopGroupEntry *syncd_fg_route_entry = &(route_table.second);
+            auto prefix_entry = fgNhgPrefixes.find(route_table.first);
+            assert(prefix_entry != fgNhgPrefixes.end());
```
prsunny, line 147: can you return failure instead of asserting here?

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
```
prsunny, line 201: Same here

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
+
+    sai_attribute_t nhgm_attr;
+    nhgm_attr.id = SAI_NEXT_HOP_GROUP_MEMBER_ATTR_NEXT_HOP_ID;
+    nhgm_attr.value.oid = nh_oid;
+    sai_status_t status = sai_next_hop_group_api->set_next_hop_group_member_attribute(
+                                                              syncd_fg_route_entry->nhopgroup_members[index],
+                                                              &nhgm_attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to set next hop oid %" PRIx64 " member %" PRIx64 ": %d\n",
```
prsunny, line 122: Remove` \n` from logs and other occurences

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
+        return false;
+    }
+    gCrmOrch->decCrmResUsedCounter(CrmResourceType::CRM_NEXTHOP_GROUP);
+    gRouteOrch->decNextHopGroupCount();
+
+    return true;
+}
+
+
+bool FgNhgOrch::set_active_bank_hash_bucket_changes(FGNextHopGroupEntry *syncd_fg_route_entry, FgNhgEntry *fgNhgEntry,
```
prsunny, line 278: Can you please explain what this function does in few comments?

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
+        {
+            SWSS_LOG_INFO("There were no valid next-hops to add %s:%s", ipPrefix.to_string().c_str(),
+                    nextHops.to_string().c_str());
+            /* Let the route retry logic(upon false rc) take care of this case */
+            return false;
+        }
+
+        if(gRouteOrch->getNextHopGroupCount() >= gRouteOrch->getMaxNextHopGroupCount())
+        {
+            SWSS_LOG_DEBUG("Failed to create new next hop group. \
```
prsunny, line 856: This should be a WARNING..

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
+        nhg_attr.id = SAI_NEXT_HOP_GROUP_ATTR_TYPE;
+        nhg_attr.value.s32 = SAI_NEXT_HOP_GROUP_TYPE_FINE_GRAIN_ECMP;
+        nhg_attrs.push_back(nhg_attr);
+
+        nhg_attr.id = SAI_NEXT_HOP_GROUP_ATTR_CONFIGURED_SIZE;
+        nhg_attr.value.s32 = fgNhgEntry->configured_bucket_size;
+        nhg_attrs.push_back(nhg_attr);
+
+        sai_object_id_t next_hop_group_id;
+        sai_status_t status = sai_next_hop_group_api->create_next_hop_group(&next_hop_group_id,
```
prsunny, line 876: Please move this section of calling sai to a different function. This is already a big function.

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
+
+        /* Now add the route pointing to the fgnhg */
+        sai_route_entry_t route_entry;
+        sai_attribute_t route_attr;
+        route_entry.vr_id = vrf_id;
+        route_entry.switch_id = gSwitchId;
+        copy(route_entry.destination, ipPrefix);
+        route_attr.id = SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID;
+        route_attr.value.oid = syncd_fg_route_entry.next_hop_group_id;
+        status = sai_route_api->create_route_entry(&route_entry, 1, &route_attr);
```
prsunny, line 928: Same for this section as well. Please move to a different function

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
+        }
+        else 
+        {
+            /* delete regular ecmp handling for prefix */
+            sai_object_id_t vrf_id = gVirtualRouterId;
+            auto route_table_entry = gRouteOrch->getSyncdRoutes().find(vrf_id);
+            NextHopGroupKey nhg;
+            if (route_table_entry == gRouteOrch->getSyncdRoutes().end())
+            {
+                SWSS_LOG_INFO("Failed to find route table, vrf_id 0x%lx\n", vrf_id);
```
prsunny, line 1155: This is not a failure, correct.. may be rephrase to "No route table found"

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
+            if (route_handled)
+            {
+                if(!addRoute(vrf_id, ip_prefix, nhg))
+                {
+                    SWSS_LOG_INFO("Failed to add fg route, %s:%s", 
+                            ip_prefix.to_string().c_str(), nhg.to_string().c_str());
+                    return false;
+                }
+                SWSS_LOG_INFO("%s: Add route with ip prefix %s with nexthop group key %s", 
+                        __FUNCTION__, ip_prefix.to_string().c_str(),nhg.to_string().c_str());
```
prsunny, line 1188: you don't need to specify __FUNCTION__.. please remove all occurrences

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
+            if (fvField(i) == "FG_NHG")
+            {
+                fgNhg_name = fvValue(i);
+            }
+            else if(fvField(i) == "bank")
+            {
+                bank = stoi(fvValue(i));
+            }
+        }
+        if(fgNhg_name == "")
```
prsunny, line 1277: check for empty instead. Just a suggestion

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1379 @@
...
+                if (it_route != route_table_entry->second.end())
+                {
+                    nhg = it_route->second;
+                    RouteBulkContext ctx;
+                    ctx.ip_prefix = ip_prefix;
+                    ctx.vrf_id = vrf_id;
+                    ctx.nhg = nhg;
+                    gRouteOrch->removeRoute(ctx);
+                    gRouteOrch->gRouteBulker.flush();
+                    gRouteOrch->removeRoutePost(ctx);
```
prsunny, line 1170: How about the failures here? If this fails, it will not be removed from route_table and you'll end up adding fgnhg prefix as well.

qiluo-msft, line 1170: removeRoute/removeRoutePost are designed to implement bulk operations on routes. If you use this way, there is no bulking at all.

orchagent/routeorch.h
```diff
@@ -107,10 +107,42 @@ class RouteOrch : public Orch, public Subject
...
+    {
+        m_nextHopGroupCount--;
+    }
+
+    int getMaxNextHopGroupCount(void)
+    {
+    	return m_maxNextHopGroupCount;
+    }
+
+    EntityBulker<sai_route_api_t>           gRouteBulker;
```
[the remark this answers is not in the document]
qiluo-msft: The private functions and members are implementation details inside RouteOrch. Others are expected not to use them directly.

---
In reply to: [472580714](https://github.com/Azure/sonic-swss/pull/1315#discussion_r472580714) [](ancestors = 472580714)

tests/test_fgnhg.py
anish-n: I do see the sleep used across the board in the tests today, eg: https://github.com/Azure/sonic-swss/blob/master/tests/test_nhg.py#L117 . Without sleep the test case fails since it may not be enough time for the change to take effect. I do see that it can cause tests to be unstable since there is a possibility that 1s is not enough if the system is exceptionally slow. Is there a good alternative we are using today?

qiluo-msft: You may use one of the `wait_*` functions

---
In reply to: [477487120](https://github.com/Azure/sonic-swss/pull/1315#discussion_r477487120) [](ancestors = 477487120)

anish-n: I see thanks, will use that

tests/test_fgnhg.py
anish-n: Will update to use the new framework

tests/test_fgnhg.py
anish-n: Will do, thanks

orchagent/routeorch.h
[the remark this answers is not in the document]
anish-n: Made bulker and addRoute private again, and used a different methodology to achieve the same behavior.

orchagent/routeorch.cpp
```diff
@@ -1206,6 +1226,15 @@ bool RouteOrch::addRoutePost(const RouteBulkContext& ctx, const NextHopGroupKey
     const sai_object_id_t& vrf_id = ctx.vrf_id;
     const IpPrefix& ipPrefix = ctx.ip_prefix;
 
+    if (m_fgNhgOrch->fgNhgPrefixes.find(ipPrefix) != m_fgNhgOrch->fgNhgPrefixes.end()
+            && vrf_id == gVirtualRouterId)
+    {
+        /* Only support the default vrf for Fine Grained ECMP */
+        SWSS_LOG_INFO("Reroute %s:%s to fgNhgOrch", ipPrefix.to_string().c_str(), 
+                nextHops.to_string().c_str());
+        return m_fgNhgOrch->addRoute(vrf_id, ipPrefix, nextHops);
```
qiluo-msft, line 1235: >m_fgNhgOrch->addRoute [](start = 15, length = 21)

RouteOrch::addRoute() and RouteOrch::addRoutePost will be called once for each route request in RouteOrch::doTask().
So m_fgNhgOrch->addRoute() will be called twice for each route request. Is it intened?

orchagent/routeorch.cpp
```diff
@@ -1334,6 +1363,14 @@ bool RouteOrch::removeRoute(RouteBulkContext& ctx)
     sai_object_id_t& vrf_id = ctx.vrf_id;
     IpPrefix& ipPrefix = ctx.ip_prefix;
 
+    if (m_fgNhgOrch->fgNhgPrefixes.find(ipPrefix) != m_fgNhgOrch->fgNhgPrefixes.end()
+            && vrf_id == gVirtualRouterId)
+    {
+        /* Only support the default vrf for Fine Grained ECMP */
+        SWSS_LOG_INFO("Reroute %s to fgNhgOrch", ipPrefix.to_string().c_str());
+        return m_fgNhgOrch->removeRoute(vrf_id, ipPrefix);
```
qiluo-msft, line 1371: >m_fgNhgOrch->removeRoute [](start = 15, length = 24)

Same as m_fgNhgOrch->addRoute()

orchagent/routeorch.h
```diff
@@ -108,11 +108,47 @@ class RouteOrch : public Orch, public Subject
     bool invalidnexthopinNextHopGroup(const NextHopKey&);
 
     void notifyNextHopChangeObservers(sai_object_id_t, const IpPrefix&, const NextHopGroupKey&, bool);
+    const NextHopGroupKey getSyncdRoute(sai_object_id_t vrf_id, const IpPrefix& ipPrefix)
```
qiluo-msft, line 111: >getSyncdRoute [](start = 26, length = 13)

getSyncdRouteNhgk?

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1500 @@
...
+            syncd_fg_route_entry->nhopgroup_members[index], nh_oid, status);
+        return false;
+    }
+
+    set_state_db_route_entry(ipPrefix, index, nextHop);
+    return true;
+}
+
+
+bool FgNhgOrch::create_fine_grained_next_hop_group(FGNextHopGroupEntry &syncd_fg_route_entry, FgNhgEntry *fgNhgEntry,
```
qiluo-msft, line 140: >create_fine_grained_next_hop_group [](start = 16, length = 34)

Is it possible to move this function into class RouteOrch? Then no need to expose XxxNextHopGroupCount() functions.

orchagent/fgnhgorch.cpp
```diff
@@ -0,0 +1,1500 @@
...
+
+/* calculate_bank_hash_bucket_start_indices: generates the hash_bucket_indices for all banks
+ * and stores it in fgNhgEntry for the group. 
+ * The function will identify the # of next-hops assigned to each bank and 
+ * assign the total number of hash buckets for a bank, based on the proportional
+ * number of next-hops in the bank. 
+ * eg: Bank0: 6 nh, Bank1: 3 nh, total buckets: 30 => 
+ *      calculate_bank_hash_bucket_start_indices: Bank0: Bucket# 0-19, Bank1: Bucket# 20-29
+ */
+void calculate_bank_hash_bucket_start_indices(FgNhgEntry *fgNhgEntry)
```
qiluo-msft, line 39: >calculate_bank_hash_bucket_start_indices [](start = 5, length = 40)

CamelCases ?

orchagent/fgnhgorch.h
```diff
@@ -0,0 +1,128 @@
...
+{
+    uint32_t start_index;
+    uint32_t end_index;
+} bank_index_range;
+
+typedef struct FgNhgEntry
+{
+    string fgNhg_name;                                  // Name of FG NHG group configured by user
+    uint32_t configured_bucket_size;                    // Bucket size configured by user
+    uint32_t real_bucket_size;                          // Real bucket size as queried from SAI
```
qiluo-msft, line 57: >real_bucket_size [](start = 13, length = 16)

camelCases ?

orchagent/routeorch.h
qiluo-msft, line 111: >getSyncdRoute [](start = 26, length = 13)

This function is long and not for inline. Please move it into cpp file.

orchagent/routeorch.cpp
anish-n: Based on my code walk through, 2 calls for the m_fgNhgOrch->addRoute will happen only if the initial fgNhgOrch addRoute failed. So it will function like a retry in failure cases. 

For success case it should return True, and ```consumer.m_toSync.erase(it)``` should occur, as a result it would get skipped when it gets to the main caller of addRoutePost(hence being processed only 1 time):
```
        // Go through the bulker results
        auto it_prev = consumer.m_toSync.begin();
        while (it_prev != it)
```

Do you see this occurring differently/is my understanding off?

qiluo-msft: My understanding is that if the first `m_fgNhgOrch->addRoute` in `RouteOrch::addRoute` succeeded, it will be called again in `RouteOrch::addRoutePost`.

---
In reply to: [487338902](https://github.com/Azure/sonic-swss/pull/1315#discussion_r487338902) [](ancestors = 487338902)

anish-n: That appears to be true if the route is processed by routeorch(because ```RouteOrch::addRoute``` always returns ```FALSE```, hence guaranteeing that it gets processed by addRoutePost too). But if processed by fgNhgOrch it returns ```TRUE``` for success and is not sent to addRoutePost in that case. 

I set loglevel to info and verified that m_fgNhgOrch->addRoute is only being called 1 time in success case.

qiluo-msft: Thanks @anish-n, very clear now.

You mentioned that
> 2 calls for the m_fgNhgOrch->addRoute will happen only if the initial fgNhgOrch addRoute failed. So it will function like a retry in failure cases

This is not the intention of `RouteOrch::addRoutePost`, because it is intended to handle bulk operation results.

So suggest you remove this block in `RouteOrch::addRoutePost`, and the failed route entry in `m_toSync` will be retried in next `doTask()`

---
In reply to: [493819285](https://github.com/Azure/sonic-swss/pull/1315#discussion_r493819285) [](ancestors = 493819285,487338902)

orchagent/fgnhgorch.cpp
anish-n: Moved it to RouteOrch

orchagent/fgnhgorch.h
anish-n: Have changed all function names and data types to be in camelCase. Have left the underscore syntax for the struct data members as-is though. It is now consistent with how these definitions are in routeorch.

orchagent/routeorch.cpp
anish-n: Makes sense, sure, will remove the block.

orchagent/fgnhgorch.cpp
qiluo-msft: Will you also change this to CamelCases?

---
In reply to: [487340495](https://github.com/Azure/sonic-swss/pull/1315#discussion_r487340495) [](ancestors = 487340495)

orchagent/fgnhgorch.h
qiluo-msft: you mix camelCases and under_score conventions for variable names in the same class.

---
In reply to: [487340692](https://github.com/Azure/sonic-swss/pull/1315#discussion_r487340692) [](ancestors = 487340692)

orchagent/fgnhgorch.cpp
anish-n: Oh, missed converting this one to camelCase, will do so

qiluo-msft:
@daall Could you help review tests/test_fgnhg.py. There are some `time.sleep(1)`.

qiluo-msft (approved):
LGTM except `time.sleep(1)` in `tests/test_fgnhg.py`

daall (approved):
All of the `time.sleep(1)` calls have clear comments so I think this is OK as is.

daall:
retest vs please

PR #1326, opened 2020-06-18
[FPMSYNCD] EVPN Type5 prefix handling in FPMSYNCD as described in the PR Azure/SONiC#437

kishorekunal01:
[FPMSYNCD] EVPN Type5 prefix handling in FPMSYNCD.

**What I did**
Add raw netlink msg processing for EVPN routes so rt_nl doesn't provide support to extract custom msg from the netlink msg.

**Why I did it**
Part of overall EVPN-VXLAN Implementation

**How I verified it**
Verified on VS build.

**Details if related**
HLD Location : https://github.com/Azure/SONiC/pull/437/commits


prsunny:
This is dependent on PR - https://github.com/Azure/sonic-buildimage/pull/4806, correct?

kishorekunal01:
> This is dependent on PR - [Azure/sonic-buildimage#4806](https://github.com/Azure/sonic-buildimage/pull/4806), correct?

Yes, that is correct, this changes depend on [Azure/sonic-buildimage#4806](https://github.com/Azure/sonic-buildimage/pull/4806)

fpmsyncd/fpmsyncd.cpp
```diff
@@ -56,6 +56,8 @@ int main(int argc, char **argv)
 
     NetDispatcher::getInstance().registerMessageHandler(RTM_NEWROUTE, &sync);
     NetDispatcher::getInstance().registerMessageHandler(RTM_DELROUTE, &sync);
+    NetDispatcher::getInstance().registerRawMessageHandler(RTM_NEWROUTE, &sync);
+	NetDispatcher::getInstance().registerRawMessageHandler(RTM_DELROUTE, &sync);
```
qiluo-msft, line 60: >	 [](start = 0, length = 1)

prefer spaces instead of tab

fpmsyncd/fpmlink.cpp
```diff
@@ -9,6 +9,81 @@
 using namespace swss;
 using namespace std;
 
+void netlink_parse_rtattr(struct rtattr **tb, int max, struct rtattr *rta,
+        int len)
+{
+    while (RTA_OK(rta, len)) {
```
pavel-shirshov, line 15: Please reformat it in 
```
while ()
{
}
```

fpmsyncd/fpmsyncd.cpp
kishorekunal01: Update as per comment

fpmsyncd/fpmlink.cpp
kishorekunal01: Update the file with reformat

fpmsyncd/routesync.cpp
```diff
@@ -30,6 +54,466 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
...
+
+                    if (subtb[RTA_ENCAP] && subtb[RTA_ENCAP_TYPE]
+                        && (*(uint16_t *)RTA_DATA(subtb[RTA_ENCAP_TYPE]) == NH_ENCAP_VXLAN))
+                    {
+                        parse_encap(subtb[RTA_ENCAP], encap_value, rmac, vlan);
+                    }
+                    SWSS_LOG_DEBUG("Multipath Nexthop:%s encap:%d encap_value:%d rmac:%s vlan:%d",
+                                    nexthopaddr, encap, encap_value, rmac.c_str(), vlan);
+
+                    if (encap_value == 0 || vlan == 0
```
kperumalbfn, line 284: Partial ecmp nexthops are added to route APPDB in this error case. There could be other resolved nhops in the message that are not processed, I think we should process those and add to the DB.

fpmsyncd/routesync.cpp
```diff
@@ -30,6 +54,466 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
...
+                        || MacAddress(rmac) == MacAddress("00:00:00:00:00:00"))
+                        return false;
+
+                    if(ecmp_count)
+                    {
+                        getEvpnNextHopSep(nexthops, vni_list, mac_list, intf_list);
+                    }
+
```
kperumalbfn, line 292: Add this only if the message has RTA_GATEWAY??

fpmsyncd/routesync.cpp
```diff
@@ -30,6 +54,466 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
...
+    int  dst_len = 0;
+    char buf[MAX_ADDR_SIZE];
+    char destipprefix[IFNAMSIZ + MAX_ADDR_SIZE + 2] = {0};
+    int nlmsg_type = h->nlmsg_type;
+    unsigned int vrf_index;
+
+    rtm = (struct rtmsg *)NLMSG_DATA(h);
+
+    /* Parse attributes and extract fields of interest. */
+    memset(tb, 0, sizeof tb);
```
kperumalbfn, line 324: memset syntax

fpmsyncd/fpmlink.cpp
```diff
@@ -9,6 +9,86 @@
...
+        && h->nlmsg_type != RTM_DELROUTE)
+        return 0;
+
+    len = (int)(h->nlmsg_len - NLMSG_LENGTH(sizeof(struct rtmsg)));
+    if (len < 0) 
+    {
+        return 0;
+    }
+
+    memset(tb, 0, sizeof tb);
```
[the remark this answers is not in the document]
kishorekunal01: Updated based on review comment

fpmsyncd/routesync.cpp
kishorekunal01: Update the code based on the comment

fpmsyncd/routesync.cpp
kishorekunal01: Update the code

fpmsyncd/routesync.cpp
kperumalbfn: We shouldn't program nexthop with partial rewrite into AppDB, but this message could contain valid nexthops after one partial nexthop in the list. If this code returns for one partial nexthop, then valid nexthops are also not added to DB. Is this expected??

kishorekunal01: We don't want to program route even when one partial nexthop information is available. 
If this case exist we need to debug why this incorrect nexthop is send for this prefix(even if one nexthop is not correct), else there will be inconsistency between the control plane and data plane.

kperumalbfn: Makes sense to bypass route addition with partial info. When partial nexthop gets valid encap/mac, does kernel notify existing route RTM_NEWROUTE notification with all nexthop list?? I believe VNID encap and rmac will be always valid as that is part of vtep interface configuration. How about vlan? This could be invalid at the time of route addition.

kishorekunal01: These message are coming directly from FRR not via kernel. In FRR function netlink_route_info_encode encode these message and send to fpm which is received to fpmsyncd. Since these route are directly send by FRR it is better to valid all the information before programming the ASIC.

fpmsyncd/fpmlink.cpp
```diff
@@ -9,6 +9,86 @@
...
+                break;
+
+            len -= NLMSG_ALIGN(rtnh->rtnh_len);
+            rtnh = RTNH_NEXT(rtnh);                
+        }
+    }
+
+    SWSS_LOG_INFO("Rx MsgType:%d Encap:%d", h->nlmsg_type, encap_type);
+
+    if (encap_type > 0)
```
prsunny, line 86: Please use `{ `for if

fpmsyncd/fpmlink.cpp
```diff
@@ -9,6 +9,86 @@
...
+        return 1;
```
prsunny, line 87: Why not use true/false?

fpmsyncd/fpmlink.cpp
```diff
@@ -9,6 +9,86 @@
...
+    {
+        /* This is a multipath route */
+        int len;            
+        struct rtnexthop *rtnh = (struct rtnexthop *)RTA_DATA(tb[RTA_MULTIPATH]);
+        len = (int)RTA_PAYLOAD(tb[RTA_MULTIPATH]);
+        struct rtattr *subtb[RTA_MAX + 1];
+        
+        for (;;) 
+        {
+            if (len < (int)sizeof(*rtnh)
```
prsunny, line 60: Suggest to have it in a single line and as a general comment, use open/close braces for `if`

fpmsyncd/routesync.cpp
```diff
@@ -30,6 +54,458 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
     rtnl_link_alloc_cache(m_nl_sock, AF_UNSPEC, &m_link_cache);
 }
 
+char *RouteSync::prefix_mac2str(char *mac, char *buf, int size)
```
prsunny, line 57: Did you check on using `macaddress.h` utility?

fpmsyncd/routesync.cpp
```diff
@@ -30,6 +54,458 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
...
+    struct rtattr *tb_encap[3] = {0};
+    char mac_buf[MAX_ADDR_SIZE+1];
+    char mac_val[MAX_ADDR_SIZE+1];
+
+    netlink_parse_rtattr_nested(tb_encap, 3, tb);
+    encap_value = *(uint32_t *)RTA_DATA(tb_encap[VXLAN_VNI]);
+    memcpy(&mac_buf, RTA_DATA(tb_encap[VXLAN_RMAC]), MAX_ADDR_SIZE);
+    vlan = *(uint32_t *)RTA_DATA(tb_encap[VXLAN_VLAN]);
+
+    SWSS_LOG_DEBUG("Rx MAC %s VNI %d Vlan %d",
```
prsunny, line 103: Please change log level to INFO

fpmsyncd/routesync.cpp
```diff
@@ -30,6 +54,458 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
...
+            /* Need to add the code for multipath */
+            int len;
+            struct rtattr *subtb[RTA_MAX + 1];
+            struct rtnexthop *rtnh = (struct rtnexthop *)RTA_DATA(tb[RTA_MULTIPATH]);
+            len = (int)RTA_PAYLOAD(tb[RTA_MULTIPATH]);
+
+            for (;;) 
+            {
+                uint16_t encap = 0;
+                if (len < (int)sizeof(*rtnh)
```
prsunny, line 216: Use single line as we can have 100 char per line

fpmsyncd/routesync.cpp
```diff
@@ -30,6 +54,458 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
...
+            return;
+        }
+        destipprefix[strlen(destipprefix)] = ':';
+    }
+
+    /* Full mask route append prefix length, or else resync cannot match. */
+    snprintf(destipprefix + strlen(destipprefix), sizeof(destipprefix) - strlen(destipprefix), "%s/%u", 
+                inet_ntop(rtm->rtm_family, dstaddr, buf, MAX_ADDR_SIZE), dst_len);
+
+    SWSS_LOG_DEBUG("Receive route message dest ip prefix: %s Op:%s",
```
prsunny, line 375: Please use INFO level

fpmsyncd/fpmlink.cpp
```diff
@@ -120,12 +200,28 @@ uint64_t FpmLink::readData()
 
         if (hdr->msg_type == FPM_MSG_TYPE_NETLINK)
         {
-            nl_msg *msg = nlmsg_convert((nlmsghdr *)fpm_msg_data(hdr));
+            bool isRaw = false;
+
+            nlmsghdr *nl_hdr = (nlmsghdr *)fpm_msg_data(hdr);
+
+            isRaw = isRawProcessing(nl_hdr);
```
prsunny, line 207: Can you add some comment here on what is Raw processing?

fpmsyncd/routesync.cpp
```diff
@@ -30,6 +54,458 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
...
+
+    SWSS_LOG_DEBUG("Receive route message dest ip prefix: %s Op:%s", 
+                    destipprefix,
+                    nlmsg_type == RTM_NEWROUTE ? "add":"del");
+
+    /*
+     * Upon arrival of a delete msg we could either push the change right away,
+     * or we could opt to defer it if we are going through a warm-reboot cycle.
+     */
+    bool warmRestartInProgress = m_warmStartHelper.inProgress();
```
prsunny, line 383: I'm not sure I understand it here. Why do we want to check for warmboot status in this fpmsyncd flow? How do we hit this case?

fpmsyncd/routesync.h
```diff
@@ -35,12 +41,35 @@ class RouteSync : public NetMsg
     /* Handle regular route (include VRF route) */
     void onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf);
 
+    void parse_encap(struct rtattr *tb, uint32_t &encap_value, string &rmac, uint32_t &vlan);
```
prsunny, line 44: Could you please follow the same naming convention as in the rest of file?

fpmsyncd/routesync.cpp
```diff
@@ -30,6 +54,458 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
...
+            const KeyOpFieldsValuesTuple kfv = std::make_tuple(destipprefix,
+                                                               DEL_COMMAND,
+                                                               fvVector);
+            m_warmStartHelper.insertRefreshMap(kfv);
+            return;
+        }
+    }
+    else if (nlmsg_type != RTM_NEWROUTE)
+    {
+        SWSS_LOG_NOTICE("Unknown message-type: %d for %s", nlmsg_type, destipprefix);
```
prsunny, line 407: I don't think we need to log NOTICE here. Just in case if we get lots of this msg, its better to ignore

fpmsyncd/routesync.cpp
kishorekunal01: This code is similar to onMsg processing(existing code to add/del prefix), when warmStart is in progress then don't delete the entry from the table.

fpmsyncd/fpmlink.cpp
kishorekunal01: Done as suggest in all the files.

fpmsyncd/fpmlink.cpp
kishorekunal01: Add as suggested

fpmsyncd/fpmlink.cpp
kishorekunal01: Done

fpmsyncd/fpmlink.cpp
kishorekunal01: Done as suggested

fpmsyncd/routesync.cpp
kishorekunal01: Done as suggested

fpmsyncd/routesync.cpp
kishorekunal01: Done as suggested

fpmsyncd/routesync.h
kishorekunal01: Done as suggested

fpmsyncd/routesync.cpp
kishorekunal01: Done as suggested

fpmsyncd/fpmlink.cpp
```diff
@@ -120,12 +210,35 @@ uint64_t FpmLink::readData()
...
                 throw system_error(make_error_code(errc::bad_message), "Unable to convert nlmsg");
+            }
 
             nlmsg_set_proto(msg, NETLINK_ROUTE);
-            NetDispatcher::getInstance().onNetlinkMessage(msg);
+
+            if (isRaw)
+            {
+                /* EVPN Type5 Add route processing */
+                NetDispatcher::getInstance().onNetlinkMessageRaw(msg);
```
qiluo-msft, line 236: >NetDispatcher::getInstance().onNetlinkMessageRaw(msg); [](start = 16, length = 54)

Could you call onMsgRaw directly? So there is no need to change sonic-swss-common

fpmsyncd/fpmlink.cpp
```diff
@@ -5,10 +5,103 @@
 #include "netmsg.h"
 #include "netdispatcher.h"
 #include "fpmsyncd/fpmlink.h"
+#include "fpmsyncd/routesync.h"
 
 using namespace swss;
 using namespace std;
 
+extern RouteSync *g_routesync;
```
qiluo-msft, line 13: >g_routesync [](start = 18, length = 11)

You may add it as FpmLink constructor parameter, so removing this global extern variable.
Please also check name convention with nearby code.

qiluo-msft (commented):
Thanks for the improvement! I see one minor issue.

prsunny:
retest vs please

prsunny (approved):

qiluo-msft (approved):

kperumalbfn (dismissed):

prsunny:
retest vs please

kishorekunal01:
Sure, I will retest on VS

prsunny:
Failures: 
test_route.TestRoute.test_RouteAndNexthopInDifferentVrf
test_route.TestRoutePerf.test_PerfAddRemoveRoute
test_warm_reboot.TestWarmReboot.test_routing_WarmRestart
test_warm_reboot.TestWarmReboot.test_VrfMgrdWarmRestart

prsunny:
retest vs please

prsunny:
retest vs please

kishorekunal01:
retest vs please

kishorekunal01:
retest vs please

fpmsyncd/routesync.cpp
```diff
@@ -100,8 +599,14 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
 
     dip = rtnl_route_get_dst(route_obj);
     nl_addr2str(dip, destipprefix + strlen(destipprefix), MAX_ADDR_SIZE);
+#if 0
```
prsunny, line 602: Looks like with this, the VS test is passing, if so, can you please remove this chunk?

kishorekunal01: @prsunny Yes, that is correct this is causing the VS test to fail, Will test it once locally for the side effect of removing this and update by EoD

prsunny (approved):

PR #1327, opened 2020-06-19
Fix array delete and header guard

qiluo-msft:


qiluo-msft:
Retest vs please

qiluo-msft:
Retest this please

PR #1329, opened 2020-06-22
[Vlan]: Added SWSS support to handle new vlan cli OPTIONAL command 

rajkumar38:
vlan type - data/mananagement/control

PR : https://github.com/Azure/sonic-utilities/pull/956
Purpose is to identify the vlan type. This is optional configuration.
If not set, it will be treated as regular ("data") vlan (current behavior).
Note: SAI PR is separately raised and not merged.
Meanwhile, we have hard-coded the values. This can be fixed once SAI PR is merged.

Signed-off-by: Rajkumar Pennadam Ramamoorthy <rpennadamram@marvell.com>

**What I did**
Added SWSS support for new vlan enhancement. type - management/control
**Why I did it**
We have a use-case to differentiate managament and control traffic from regular data-traffic in data path.
**How I verified it**
configure the command from cli. Check CONFIG_DB,APP_DB and ASIC_DB for "vlan_type" value.
**Details if related**
Details of CLI command can be found in PR, https://github.com/Azure/sonic-utilities/pull/956

### DIFF
--- a/cfgmgr/vlanmgr.cpp
+++ b/cfgmgr/vlanmgr.cpp
@@ -266,6 +266,7 @@ void VlanMgr::doVlanTask(Consumer &consumer)
             string mtu = DEFAULT_MTU_STR;
             vector<FieldValueTuple> fvVector;
             string members;
+            string vlan_type = "";
 
             /*
              * If state is already set for this vlan, but it doesn't exist in m_vlans set,
@@ -310,6 +311,12 @@ void VlanMgr::doVlanTask(Consumer &consumer)
                      */
                     SWSS_LOG_DEBUG("%s mtu %s: Host VLAN mtu setting to be supported.", key.c_str(), mtu.c_str());
                 }
+                /* Vlan_type (management/control)*/
+                else if (fvField(i) == "vlan_type")
+                {
+                    vlan_type = fvValue(i);
+                    SWSS_LOG_DEBUG("%s type %s to be supported.", key.c_str(), vlan_type.c_str());
+                }
                 else if (fvField(i) == "members@") {
                     members = fvValue(i);
                 }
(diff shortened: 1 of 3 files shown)


cfgmgr/vlanmgr.cpp
```diff
@@ -324,6 +331,13 @@ void VlanMgr::doVlanTask(Consumer &consumer)
             FieldValueTuple m("mtu", mtu);
             fvVector.push_back(m);
 
+            if(vlan_type != "")
```
prsunny, line 334: please add space after `if`, Also can you check for string `empty` as followed in rest of file?

orchagent/portsorch.cpp
```diff
@@ -3219,17 +3224,47 @@ bool PortsOrch::setBridgePortLearnMode(Port &port, string learn_mode)
...
+    /* SAI support is yet to be added for vlan_type attribute.
+     * For now, we can use SAI_VLAN_ATTR_META_DATA attr_id with
+     * hard-coded values. New enums yet to be introduced in SAI
+     * for values
+     *  0 = data (default)
+     *  1 = management
+     *  2 = control
+     *  Change this code once attribute and enums introduced in saivlan.h
+     *  */
+    if(vlan_type != "")
```
prsunny, line 3249: Add a space after `if` and all cases down.

orchagent/portsorch.cpp
```diff
@@ -3219,17 +3224,47 @@ bool PortsOrch::setBridgePortLearnMode(Port &port, string learn_mode)
...
+    {
+        attr.id = SAI_VLAN_ATTR_META_DATA;
```
prsunny, line 3251: What are the attributes here? Do you have a pointer to SAI proposal?. If it is not yet approved in SAI community, it will be too early for this changes.

orchagent/portsorch.cpp
rajkumar38: Done

orchagent/portsorch.cpp
rajkumar38: Done.

PR #1334, opened 2020-06-27
convert swss vs test to support python3

lguohan:
**What I did**
convert swss vs test to support python3

**Why I did it**
python3 is the future

**How I verified it**
still verifying

**Details if related**


lguohan:
trying to make it work for both python2 and python3

lguohan:
retest this please

PR #1335, opened 2020-06-28
[orchagent] fix vector set error for QUEUE_STAT_COUNTER

JiangboHe:
**What I did**
I found twice set for poll interval of queue stat counter and the first set is wrong. It is the poll interval of port stat counter and should be cleared before setting poll interval of queue stat counter.

**Why I did it**
I clear fields vector before setting poll interval of queue stat counter.

**How I verified it**

**Details if related**


JiangboHe:
It should not have the error in LGTM analysis: C/C++. I do not know why.

PR #1336, opened 2020-06-30
Orchagent not to exit upon shutdown message from syncd

renukamanavalan:
When orchagent receives shutdown message from syncd, it exits, which triggers
some chain of events that result in swss restart. Change this default behavior as
not to exit, but repeatedly log "switcch down" error message, which can raise an
alert for a manual repair.

**Justification:**
When orchagent exits, the chain reaction triggers the entire restart of swss.
But depending on the root-cause of syncd's reason to exit, the swss-restart
may not solve the problem, but put the system in repeated cycle of restart.

Instead if we leave orchagent running with error log, someone can manually
root cause the reason for syncd exit, fix the root cause,  and then restart
swss,  which has higher probability to come up fine. Moreover, until the manual
intervention, the data traffic  would be moving unaffected. The manual fix would
also involve, safety checks and isolate the device, before making any fixes.
Hence the issue can be fixed clean with least disruption.

On the other hand, until manual fix, no new rules can be applied to data-traffic.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/notifications.cpp
+++ b/orchagent/notifications.cpp
@@ -5,7 +5,7 @@ extern "C" {
 #include "logger.h"
 #include "notifications.h"
 
-extern bool gIsSwitchShutdown
+extern bool gIsSwitchShutdown;
 
 void on_fdb_event(uint32_t count, sai_fdb_event_notification_data_t *data)
 {
@@ -23,7 +23,6 @@ void on_switch_shutdown_request()
 {
     SWSS_LOG_ENTER();
 
-    /* TODO: Later a better restart story will be told here */
     SWSS_LOG_ERROR("Syncd stopped");
 
     gIsSwitchShutdown = true;
--- a/orchagent/notifications.cpp
+++ b/orchagent/notifications.cpp
@@ -5,6 +5,8 @@ extern "C" {
 #include "logger.h"
 #include "notifications.h"
 
+extern bool gIsSwitchShutdown
+
 void on_fdb_event(uint32_t count, sai_fdb_event_notification_data_t *data)
 {
     // don't use this event handler, because it runs by libsairedis in a separate thread
@@ -24,5 +26,5 @@ void on_switch_shutdown_request()
     /* TODO: Later a better restart story will be told here */
     SWSS_LOG_ERROR("Syncd stopped");
 
-    exit(EXIT_FAILURE);
+    gIsSwitchShutdown = true;
 }
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -16,6 +16,7 @@ using namespace swss;
 /* select() function timeout retry time */
 #define SELECT_TIMEOUT 1000
 #define PFC_WD_POLL_MSECS 100
+#define SWITCH_SHUTDOWN_LOG_INTERVAL_IN_SECS 60
 
 extern sai_switch_api_t*           sai_switch_api;
 extern sai_object_id_t             gSwitchId;
@@ -38,8 +39,6 @@ Directory<Orch*> gDirectory;
 NatOrch *gNatOrch;
 
 bool gIsNatSupported = false;
-
-#define SWITCH_SHUTDOWN_LOG_INTERVAL_IN_SECS 60
 bool gIsSwitchShutdown = false;
 
 
@@ -521,8 +520,6 @@ void OrchDaemon::start()
             static time_t last_log = 0;
 
             if ((time(NULL) - last_log) > SWITCH_SHUTDOWN_LOG_INTERVAL_IN_SECS) {
-                SWSS_LOG_ENTER();
-
                 SWSS_LOG_ERROR("Syncd stopped");
 
                 last_log = time(NULL)
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -39,6 +39,10 @@ NatOrch *gNatOrch;
 
 bool gIsNatSupported = false;
 
+#define SWITCH_SHUTDOWN_LOG_INTERVAL_IN_SECS 60
+bool gIsSwitchShutdown = false;
+
+
 OrchDaemon::OrchDaemon(DBConnector *applDb, DBConnector *configDb, DBConnector *stateDb) :
         m_applDb(applDb),
         m_configDb(configDb),
@@ -513,6 +517,19 @@ void OrchDaemon::start()
                 }
             }
         }
+        if (gIsSwitchShutdown) {
+            static time_t last_log = 0;
+
+            if ((time(NULL) - last_log) > SWITCH_SHUTDOWN_LOG_INTERVAL_IN_SECS) {
+                SWSS_LOG_ENTER();
+
+                SWSS_LOG_ERROR("Syncd stopped");
+
+                last_log = time(NULL)
+            }
+
+        }
+
     }
 }
 
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -522,7 +522,7 @@ void OrchDaemon::start()
             if ((time(NULL) - last_log) > SWITCH_SHUTDOWN_LOG_INTERVAL_IN_SECS) {
                 SWSS_LOG_ERROR("Syncd stopped");
 
-                last_log = time(NULL)
+                last_log = time(NULL);
             }
 
         }
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -5,7 +5,7 @@
 #include "logger.h"
 #include <sairedis.h>
 #include "warm_restart.h"
-#include "subscriberstatetable.h"
+#include "table.h"
 
 #define SAI_SWITCH_ATTR_CUSTOM_RANGE_BASE SAI_SWITCH_ATTR_CUSTOM_RANGE_START
 #include "sairedis.h"
@@ -456,7 +456,7 @@ void OrchDaemon::start()
 
         if (gIsSwitchShutdown)
         {
-            handle_switch_shutdown();
+            handleSwitchShutdown();
         }
 
         ret = m_select->select(&s, SELECT_TIMEOUT);
@@ -665,45 +665,28 @@ bool OrchDaemon::warmRestartCheck()
 }
 
 void
-OrchDaemon::check_and_exit()
+OrchDaemon::checkAndExit()
 {
     SWSS_LOG_ENTER();
 
-    swss::SubscriberStateTable tbl(m_configDb, "CONTAINER_FEATURE");
-    std::deque<KeyOpFieldsValuesTuple> entries;
+    swss::Table tbl(m_configDb, "CONTAINER_FEATURE");
+    string val;
 
-    tbl.pops(entries);
-
-    for (auto& entry: entries) {
-        string key = kfvKey(entry);
-
-        if (key == "swss") {
-            auto values = kfvFieldsValues(entry);
-
-            for (auto it : values)
-            {
-                string field = fvField(it);
-                string value = fvValue(it);
-
-                if (field == "orchagent_exit_on_switch_shutdown") {
-                    if (value == "enable") {
-                        SWSS_LOG_ERROR("Exiting....");
-                        exit(EXIT_FAILURE);
-                    }
-                }
-            }
-            break;
+    if (tbl.hget("swss", "orchagent_exit_on_switch_shutdown", val)) {
+        if (val == "enable") {
+            SWSS_LOG_ERROR("Exiting....");
+            exit(EXIT_FAILURE);
         }
     }
 }
 
 void
-OrchDaemon::handle_switch_shutdown()
+OrchDaemon::handleSwitchShutdown()
 {
     SWSS_LOG_ENTER();
 
     while(true) {
-        check_and_exit();
+        checkAndExit();
         SWSS_LOG_ERROR("Syncd stopped");
         sleep(SWITCH_SHUTDOWN_LOG_INTERVAL_IN_SECS);
     }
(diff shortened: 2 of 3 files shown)


orchagent/orchdaemon.cpp
```diff
@@ -39,6 +39,10 @@ NatOrch *gNatOrch;
 
 bool gIsNatSupported = false;
 
+#define SWITCH_SHUTDOWN_LOG_INTERVAL_IN_SECS 60
```
jleveque, line 42: Suggest to place this `#define` near the others (lines 17-18)

orchagent/orchdaemon.cpp
```diff
@@ -39,6 +39,10 @@ NatOrch *gNatOrch;
...
+bool gIsSwitchShutdown = false;
```
qiluo-msft, line 43: Suggest to place this new bool near line 40.

orchagent/orchdaemon.cpp
```diff
@@ -513,6 +517,19 @@ void OrchDaemon::start()
                 }
             }
         }
+        if (gIsSwitchShutdown) {
+            static time_t last_log = 0;
+
+            if ((time(NULL) - last_log) > SWITCH_SHUTDOWN_LOG_INTERVAL_IN_SECS) {
+                SWSS_LOG_ENTER();
```
qiluo-msft, line 524: >SWSS_LOG_ENTER(); [](start = 16, length = 17)

LOG_ENTER is for function body. What is the purpose here?

orchagent/orchdaemon.cpp
```diff
@@ -513,6 +517,19 @@ void OrchDaemon::start()
                 }
             }
         }
+        if (gIsSwitchShutdown) {
+            static time_t last_log = 0;
```
qiluo-msft, line 521: >static time_t last_log = 0; [](start = 12, length = 27)

define it outside while loop?

orchagent/notifications.cpp
```diff
@@ -24,5 +26,5 @@ void on_switch_shutdown_request()
     /* TODO: Later a better restart story will be told here */
     SWSS_LOG_ERROR("Syncd stopped");
```
qiluo-msft, line 27: Should we remove this TODO and dummy log?

orchagent/notifications.cpp
```diff
@@ -5,6 +5,8 @@ extern "C" {
 #include "logger.h"
 #include "notifications.h"
 
+extern bool gIsSwitchShutdown
```
qiluo-msft, line 8: >gIsSwitchShutdown [](start = 12, length = 17)

Missing `;`

orchagent/orchdaemon.cpp
renukamanavalan: done

orchagent/orchdaemon.cpp
renukamanavalan: done

orchagent/orchdaemon.cpp
renukamanavalan: True. Dropped

orchagent/orchdaemon.cpp
renukamanavalan: Prefer to keep it inside the scope, where it is used. Gives better readability.

orchagent/notifications.cpp
renukamanavalan: TODO: Hmmm!! Not sure. May be. Let me drop.

renukamanavalan: I will keep the log, which is at the point of receiving shutdown. The new one added, appears at the end of the loop.

orchagent/notifications.cpp
renukamanavalan: done.

orchagent/orchdaemon.cpp
qiluo-msft: In theory, `static` is shared by all instances of this class, so better to remove `static`

---
In reply to: [448016011](https://github.com/Azure/sonic-swss/pull/1336#discussion_r448016011) [](ancestors = 448016011)

daall: I believe when used in this context `static` persists this variable to this particular function call, which lets us define it inside the if-check. I'm not 100% sure what would happen if we had another OrchDaemon, but I think it would define its own `static last_log`.  We could investigate further, but I think it might be simpler just to define this variable outside of the loop as Qi suggested.

jleveque: Inside of a plain C function, a `static` variable is scoped to that function only. However, inside a C++ *class method*, a `static` variable is actually scoped to the class, as Qi stated. Therefore, multiple instances of the class will share the same variable, as if it was defined as a static member variable. Will we ever have more than one OrchDaemon? Probably not, but if we want to be safe, it's better not use `static` here.

orchagent/orchdaemon.cpp
```diff
@@ -513,6 +516,17 @@ void OrchDaemon::start()
...
+
+            if ((time(NULL) - last_log) > SWITCH_SHUTDOWN_LOG_INTERVAL_IN_SECS) {
```
prsunny, line 522: Minor comment - orchagent files follow the style of having opening bracket `{` in a new line. Could you please change?

qiluo-msft: @daall Wish this could be captured by linters instead of wasting @prsunny 's time.

renukamanavalan: Done. Thanks Prince!

orchagent/orchdaemon.cpp
renukamanavalan: 1) A static variable is declared once and the single instance is shared among all callers within its scope.
2) An instance method, having a static variable, is shared among all instances of the class.
   In our case, there is only one instance of orchdaemon. But even if there are multiple instances, they all share a single instance of this static variable.
   There is no diff between declaring it global vs local, when it is used only locally.
3) Same with static method too.

In short, if there is a static variable, it is only a single instance within its scope.

To help prove, I wrote a small code. I pasted the code and the runtime o/p.

If you need any more clarifications, I will be more than happy to dig deeper.

`
#include <stdio.h>
#include <string>

#define PRINT_ADDRESS(m, i) \
    printf("%s:%d %s: addrof = %p val=%d\n", __FUNCTION__, __LINE__, m, &(i), i)

static int k = 1;

void test_fn()
{   
    static int i = 2;
    PRINT_ADDRESS("global function", i);
}   

class x {
    public:
        void test_instance(const char *m)
        {   
            static int i = 3;
            PRINT_ADDRESS((std::string(m) + " instance method").c_str(), i);
        }
        static void test_static()
        {   
            static int i = 4;                
            PRINT_ADDRESS("class method", i);
        }   
};      

int main()
{   
    x xx, yy, zz;
    int j = 5;

    test_fn();   
    xx.test_instance("xx");
    yy.test_instance("yy");
    zz.test_instance("zz");
    x::test_static();
    xx.test_static()

    PRINT_ADDRESS("file static var", k);
    PRINT_ADDRESS("stack var", j);
    return 0;
}   

`
`
localadmin@remanava-dev-1:~/try/c$ ./t
test_fn:14 global function: addrof = 0x564678ba8014 val=2
test_instance:23 xx instance method: addrof = 0x564678ba8018 val=3
test_instance:23 yy instance method: addrof = 0x564678ba8018 val=3
test_instance:23 zz instance method: addrof = 0x564678ba8018 val=3
test_static:30 class method: addrof = 0x564678ba801c val=4
test_static:30 class method: addrof = 0x564678ba801c val=4
main:47 file static var: addrof = 0x564678ba8010 val=1
main:49 stack var: addrof = 0x7ffdfa949d84 val=5
localadmin@remanava-dev-1:~/try/c$ 
`

renukamanavalan: If you guys still prefer, pushing the local static to outside as file-local, for **readability**, I will go ahead and change it.

From code perspective alone, this is good. If you are not agreeing, please throw me the challenges, as this is the ***best*** way to learn.

lguohan:
how did you verify it?

orchagent/orchdaemon.cpp
```diff
@@ -513,6 +516,17 @@ void OrchDaemon::start()
                 }
             }
         }
+        if (gIsSwitchShutdown) {
```
lguohan, line 519: in normal executing, does the code reach here? i saw if select timeout, it just go to continue.

lguohan:
can you also check with kamil, if we are still sending request to syncd, what is the response, does it continue to process the data or blocked? if blocked, what happens to orchagent?

kcudnik:
syncd shutdown is done via separate process/bin syncd_request_shutdown, OA is not sending this request

orchagent/orchdaemon.cpp
renukamanavalan: Very good catch. I just found out, that this notification comes in a different thread and hence all the selectable orchs, could be idling causing timeout.

I just moved it up in the loop.

renukamanavalan:
After 

> how did you verify it?

Here is the plan:

- Install 201811 image with updated orchagent (with log period cut down to every 10s)
- Update syncd_request_shutdown to send switch_shutdown_request in notifications channel, like syncd does
- Freeze syncd with kill SIGSTOP
- Run  send switch_shutdown_request tool

Watch for logs ...

Please let me know, for any tips/changes/better-suggestions.

lguohan:
suggest to get a mellanox box, install ipv6 rules and see, and also remove the ipv6 rules after that and see what happens.

orchagent/notifications.cpp
renukamanavalan: TODO dropped.

orchagent/orchdaemon.cpp
renukamanavalan: No more static.

orchagent/orchdaemon.cpp
```diff
@@ -654,3 +663,50 @@ bool OrchDaemon::warmRestartCheck()
     gSwitchOrch->restartCheckReply(op,  data, values);
     return ret;
 }
+
+void
+OrchDaemon::check_and_exit()
+{
+    SWSS_LOG_ENTER();
+
+    swss::SubscriberStateTable tbl(m_configDb, "CONTAINER_FEATURE");
```
kcudnik, line 672: make a constant fro that container feature?

orchagent/orchdaemon.cpp
```diff
@@ -654,3 +663,50 @@ bool OrchDaemon::warmRestartCheck()
...
+
+        if (key == "swss") {
+            auto values = kfvFieldsValues(entry);
+
+            for (auto it : values)
+            {
+                string field = fvField(it);
+                string value = fvValue(it);
+
+                if (field == "orchagent_exit_on_switch_shutdown") {
```
kcudnik, line 688: same for "orchagent_exit_on_switch_shutdown" and "swss" if you are expecting those to be received, then someone had to send those, so you could put those constants in some header and then reuse same header in sender?

orchagent/orchdaemon.cpp
renukamanavalan: I don't get this comment. Can you please elaborate ?

orchagent/orchdaemon.cpp
renukamanavalan: Yes/No. Need your help to decide.

This is *only* a temp knob, more for an easy way to revert back to legacy behavior, if the freeze-mode was found to be impacting or causing regression. If not for this knob, we will be forced to build a new image.

Said that,  we either have to record this "orchagent_exit_on_switch_shutdown" and "swss" or remove this check. Being in transient phase, would it be OK to raise a github issue, so we get to it, in sometime ?

Or should we record it in code in some form ?

Can you please suggest?

orchagent/orchdaemon.cpp
```diff
@@ -654,3 +663,50 @@ bool OrchDaemon::warmRestartCheck()
     gSwitchOrch->restartCheckReply(op,  data, values);
     return ret;
 }
+
+void
+OrchDaemon::check_and_exit()
```
qiluo-msft, line 668: >check_and_exit [](start = 12, length = 14)

follow the "camelCase"

orchagent/orchdaemon.cpp
```diff
@@ -654,3 +663,50 @@ bool OrchDaemon::warmRestartCheck()
...
+{
+    SWSS_LOG_ENTER();
+
+    swss::SubscriberStateTable tbl(m_configDb, "CONTAINER_FEATURE");
+    std::deque<KeyOpFieldsValuesTuple> entries;
+
+    tbl.pops(entries);
```
qiluo-msft, line 675: You may use simply Table if you don't subscribe.

renukamanavalan: Thanks. This helped cut down lines of code. 
BTW, I have not tested my changes yet. After it passes review, I will do one more final test.

orchagent/orchdaemon.cpp
```diff
@@ -665,45 +665,28 @@ bool OrchDaemon::warmRestartCheck()
...
-
-                if (field == "orchagent_exit_on_switch_shutdown") {
-                    if (value == "enable") {
-                        SWSS_LOG_ERROR("Exiting....");
-                        exit(EXIT_FAILURE);
-                    }
-                }
-            }
-            break;
+    if (tbl.hget("swss", "orchagent_exit_on_switch_shutdown", val)) {
```
prsunny, line 675: @renukamanavalan , `{ `to new line :). Same as below

orchagent/orchdaemon.cpp
renukamanavalan: done

orchagent/orchdaemon.cpp
renukamanavalan: Thanks. In my last places, the style requirement is to place the brace in the same line. Hard to shake that. :-)

renukamanavalan: done.

renukamanavalan:
retest this please

PR #1338, opened 2020-07-01
[Dynamic buffer calc]  Support dynamic buffer calculation

stephenxs:
**What I did**

***Support dynamic buffer calculation***

1. Extend the CLI options for buffermgrd:
   - -a: asic_table provided,
   - -p: peripheral_table provided

   The `buffermgrd` will start the dynamic headroom calculation mode with -a provided.
   Otherwise, it will start the legacy mode (pg_headroom_profile looking up)
2. A new class is provided for dynamic buffer calculation while the old one remains.
   The daemon will instantiate the corresponding class according to the CLI option when it starts.
3. In both modes, the `buffermgrd` will copy BUFFER_XXX tables from CONFIG_DB to APPL_DB and the `bufferorch` will consume BUFFER_XXX tables from APPL_DB

***Backward compatibility***
For legacy mode, the backward compatibility is provided. As mentioned above, `buffermgrd` will check whether the json file representing the `ASIC_TABLE` exists when it starts.
- If yes it will start the dynamic buffer calculating mode
- Otherwise, it will start the compatible mode which is the old looking up mode in the new code committed in this PR.
This logic is in `cfgmgr/buffermgrd.cpp`.

The logic of buffer handling in `buffermgrd` isn't changed in the legacy mode. The differences are:
- in legacy mode which is the old code, there isn't any buffer related table in `APPL_DB`. All tables are in `CONFIG_DB`.
  - `buffermgrd` listens to `PORT` and `CABLE_LENGTH` tables in `CONFIG_DB` and inserts the buffer profiles into `BUFFER_PROFILE` table.
  - `bufferorch` listens to buffer related tables in `CONFIG_DB` and call SAI API correspondingly.
- In the compatible mode, `buffermgrd` listens to tables in `CONFIG_DB` and copies them into `APPL_DB` 
  - `buffermgrd`
    - listens to `PORT` and `CABLE_LENGTH` tables in `CONFIG_DB` and inserts the buffer profiles into `BUFFER_PROFILE` table in `CONFIG_DB` (not changed)
    - listens to buffer related tables in `CONFIG_DB` and copies them into `APPL_DB`
  - `bufferorch` listens to `APPL_DB` and call SAI API correspondingly. (the difference is the db it listens to).
  - `db_migrator` is responsible to copy the buffer related tables from `CONFIG_DB` to `APPL_DB` when system is warmbooted from the old image to the new image for the first time.

The compatible code is in `cfgmgr/buffermgr.cpp`, `orchagent/bufferorch.cpp` and `db_migrator` (in the [sonic-utilities PR](https://github.com/Azure/sonic-utilities/pull/973)).

**Why I did it**

**How I verified it**

1. vs test
2. regression test [PR: [Dynamic buffer calc] Test cases for dynamic buffer calculation](https://github.com/Azure/sonic-mgmt/pull/1971)

**Dynamic buffer details**

1. In the dynamic buffer calculation mode, there are 3 lua plugins are provided for vendor-specific operations:
   - buffer_headroom_<vendor>.lua, for calculating headroom size.
   - buffer_pool_<vendor>.lua, for calculating buffer pool size.
   - buffer_check_headroom_<vendor>.lua, for checking whether headroom exceeds the limit
2. During initialization, The daemon will:
   - load asic_table and peripheral_table from the given json file, parse them and push them into STATE_DB.ASIC_TABLE and STATE_DB.PERIPHERAL_TABLE respectively
   - load all plugins
   - try to load the STATE_DB.BUFFER_MAX_PARAM.mmu_size which is used for updating buffer pool size
   - a timer will be started for periodic buffer pool size audit
3. The daemon will listen to and handle the following tables from CONFIG_DB
   The tables will be cached internally in the daemon for the purpose of saving access time
   - BUFFER_POOL:
     - if the size is provided: insert the entry to APPL_DB
     - otherwise: cache them and push to APPL_DB after the size is calculated by lua plugin
   - BUFFER_PROFILE and BUFFER_PG:
     - items for ingress lossless headroom need to be cached and handled (according to the design)
     - other items will be inserted to the APPL_DB directly
   - PORT_TABLE, for ports' speed and MTU update
   - CABLE_LENGTH, for ports' cable length
4. Other tables will be copied to APPL_DB directly:
   - BUFFER_QUEUE
   - BUFFER_PORT_INGRESS_PROFILE_LIST
   - BUFFER_PORT_EGRESS_PROFILE_LIST
5. BufferOrch modified accordingly:
   - Consume buffer relevant tables from APPL_DB instead of CONFIG_DB
   - For BUFFER_POOL, don't set ingress/egress and static/dynamic to sai if the pool has already existed because they are create-only
   - For BUFFER_PROFILE, don't set pool for the same reason
6. Warm reboot:
   - db_migrator is responsible for copying the data from CONFIG_DB to APPL_DB if the switch is warm-rebooted from an old image to the new image for the first time
   - no specific handling in the daemon side
7. Provide vstest script

### DIFF
--- a/cfgmgr/buffermgr.cpp
+++ b/cfgmgr/buffermgr.cpp
@@ -6,20 +6,28 @@
 #include "producerstatetable.h"
 #include "tokenize.h"
 #include "ipprefix.h"
+#include "timer.h"
 #include "buffermgr.h"
 #include "exec.h"
 #include "shellcmd.h"
+#include "warm_restart.h"
 
 using namespace std;
 using namespace swss;
 
-BufferMgr::BufferMgr(DBConnector *cfgDb, DBConnector *stateDb, string pg_lookup_file, const vector<string> &tableNames) :
+BufferMgr::BufferMgr(DBConnector *cfgDb, DBConnector *applDb, string pg_lookup_file, const vector<string> &tableNames) :
         Orch(cfgDb, tableNames),
         m_cfgPortTable(cfgDb, CFG_PORT_TABLE_NAME),
         m_cfgCableLenTable(cfgDb, CFG_PORT_CABLE_LEN_TABLE_NAME),
         m_cfgBufferProfileTable(cfgDb, CFG_BUFFER_PROFILE_TABLE_NAME),
         m_cfgBufferPgTable(cfgDb, CFG_BUFFER_PG_TABLE_NAME),
-        m_cfgLosslessPgPoolTable(cfgDb, CFG_BUFFER_POOL_TABLE_NAME)
+        m_cfgLosslessPgPoolTable(cfgDb, CFG_BUFFER_POOL_TABLE_NAME),
+        m_applBufferPoolTable(applDb, APP_BUFFER_POOL_TABLE_NAME),
+        m_applBufferProfileTable(applDb, APP_BUFFER_PROFILE_TABLE_NAME),
+        m_applBufferPgTable(applDb, APP_BUFFER_PG_TABLE_NAME),
+        m_applBufferQueueTable(applDb, APP_BUFFER_QUEUE_TABLE_NAME),
+        m_applBufferIngressProfileListTable(applDb, APP_BUFFER_PORT_INGRESS_PROFILE_LIST_NAME),
+        m_applBufferEgressProfileListTable(applDb, APP_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME)
 {
     readPgProfileLookupFile(pg_lookup_file);
 }
@@ -183,9 +191,8 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, string speed)
                          "]";
 
     /* Check if PG Mapping is already then log message and return. */
-
     m_cfgBufferPgTable.get(buffer_pg_key, fvVector);
-    
+
     for (auto& prop : fvVector)
     {
         if ((fvField(prop) == "profile") && (profile_ref == fvValue(prop)))
@@ -194,12 +201,109 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, string speed)
             return task_process_status::task_success;
         }
     }
-    
+
     fvVector.clear();
- 
+
     fvVector.push_back(make_pair("profile", profile_ref));
     m_cfgBufferPgTable.set(buffer_pg_key, fvVector);
     return task_process_status::task_success;
+}
+
+void BufferMgr::transformSeperator(string &name)
+{
+    size_t pos;
+    while ((pos = name.find("|")) != string::npos)
+        name.replace(pos, 1, ":");
+}
+
+void BufferMgr::transformReference(string &name)
+{
+    auto references = tokenize(name, list_item_delimiter);
+    int ref_index = 0;
+
+    name = "";
+
+    for (auto &reference : references)
+    {
+        if (ref_index != 0)
+            name += list_item_delimiter;
+        ref_index ++;
+
+        auto keys = tokenize(reference, config_db_key_delimiter);
+        int key_index = 0;
+        for (auto &key : keys)
+        {
+            if (key_index == 0)
+                name += key + "_TABLE";
+            else
+                name += delimiter + key;
+            key_index ++;
+        }
+    }
+}
+
+/*
+ * This function copies the data from tables in CONFIG_DB to APPL_DB.
+ * With dynamically buffer calculation supported, the following tables
+ * will be moved to APPL_DB from CONFIG_DB because the CONFIG_DB contains
+ * confgured entries only while APPL_DB contains dynamically generated entries
+ *  - BUFFER_POOL
+ *  - BUFFER_PROFILE
+ *  - BUFFER_PG
+ * The following tables have to be moved to APPL_DB because they reference
+ * some entries that have been moved to APPL_DB
+ *  - BUFFER_QUEUE
+ *  - BUFFER_PORT_INGRESS_PROFILE_LIST
+ *  - BUFFER_PORT_EGRESS_PROFILE_LIST   
+ * One thing we need to handle is to transform the separator from | to :
+ * The following items contain separator:
+ *  - keys of each item
+ *  - pool in BUFFER_PROFILE
+ *  - profile in BUFFER_PG
+ *  - profile_list in BUFFER_PORT_INGRESS_PROFILE_LIST and BUFFER_PORT_EGRESS_PROFILE_LIST
+ */
+void BufferMgr::doBufferTableTask(Consumer &consumer, ProducerStateTable &applTable)
+{
+    SWSS_LOG_ENTER();
+
+    auto it = consumer.m_toSync.begin();
+    while (it != consumer.m_toSync.end())
+    {
+        KeyOpFieldsValuesTuple t = it->second;
+        string key = kfvKey(t);
+
+        //transform the separator in key from "|" to ":"
+        transformSeperator(key);
+
+        string op = kfvOp(t);
+        if (op == SET_COMMAND)
+        {
+            vector<FieldValueTuple> fvVector;
+
+            SWSS_LOG_INFO("Inserting entry %s from CONFIG_DB to APPL_DB", key.c_str());
+
+            for (auto i : kfvFieldsValues(t))
+            {
+                SWSS_LOG_INFO("Inserting field %s value %s", fvField(i).c_str(), fvValue(i).c_str());
+                //transform the separator in values from "|" to ":"
+                if (fvField(i) == "pool")
+                    transformReference(fvValue(i));
+                if (fvField(i) == "profile")
+                    transformReference(fvValue(i));
+                if (fvField(i) == "profile_list")
+                    transformReference(fvValue(i));
+                fvVector.emplace_back(FieldValueTuple(fvField(i), fvValue(i)));
+                SWSS_LOG_INFO("Inserting field %s value %s", fvField(i).c_str(), fvValue(i).c_str());
+            }
+            applTable.set(key, fvVector);
+        }
+        else if (op == DEL_COMMAND)
+        {
+            SWSS_LOG_INFO("Removing entry %s from APPL_DB", key.c_str());
+            applTable.del(key);
+        }
+        it = consumer.m_toSync.erase(it);
+    }
 }
 
 void BufferMgr::doTask(Consumer &consumer)
@@ -207,6 +311,42 @@ void BufferMgr::doTask(Consumer &consumer)
     SWSS_LOG_ENTER();
 
     string table_name = consumer.getTableName();
+
+    if (table_name == CFG_BUFFER_POOL_TABLE_NAME)
+    {
+        doBufferTableTask(consumer, m_applBufferPoolTable);
+        return;
+    }
+
+    if (table_name == CFG_BUFFER_PROFILE_TABLE_NAME)
+    {
+        doBufferTableTask(consumer, m_applBufferProfileTable);
+        return;
+    }
+
+    if (table_name == CFG_BUFFER_PG_TABLE_NAME)
+    {
+        doBufferTableTask(consumer, m_applBufferPgTable);
+        return;
+    }
+
+    if (table_name == CFG_BUFFER_QUEUE_TABLE_NAME)
+    {
+        doBufferTableTask(consumer, m_applBufferQueueTable);
+        return;
+    }
+
+    if (table_name == CFG_BUFFER_PORT_INGRESS_PROFILE_LIST_NAME)
+    {
+        doBufferTableTask(consumer, m_applBufferIngressProfileListTable);
+        return;
+    }
+
+    if (table_name == CFG_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME)
+    {
+        doBufferTableTask(consumer, m_applBufferEgressProfileListTable);
+        return;
+    }
 
     auto it = consumer.m_toSync.begin();
     while (it != consumer.m_toSync.end())
--- a/cfgmgr/buffermgr.h
+++ b/cfgmgr/buffermgr.h
@@ -13,6 +13,8 @@ namespace swss {
 #define INGRESS_LOSSLESS_PG_POOL_NAME "ingress_lossless_pool"
 #define LOSSLESS_PGS "3-4"
 
+#define BUFFERMGR_TIMER_PERIOD 10
+
 typedef struct{
     std::string size;
     std::string xon;
@@ -29,7 +31,7 @@ typedef std::map<std::string, std::string> port_cable_length_t;
 class BufferMgr : public Orch
 {
 public:
-    BufferMgr(DBConnector *cfgDb, DBConnector *stateDb, std::string pg_lookup_file, const std::vector<std::string> &tableNames);
+    BufferMgr(DBConnector *cfgDb, DBConnector *applDb, std::string pg_lookup_file, const std::vector<std::string> &tableNames);
     using Orch::doTask;
 
 private:
@@ -38,6 +40,14 @@ private:
     Table m_cfgBufferProfileTable;
     Table m_cfgBufferPgTable;
     Table m_cfgLosslessPgPoolTable;
+
+    ProducerStateTable m_applBufferProfileTable;
+    ProducerStateTable m_applBufferPgTable;
+    ProducerStateTable m_applBufferPoolTable;
+    ProducerStateTable m_applBufferQueueTable;
+    ProducerStateTable m_applBufferIngressProfileListTable;
+    ProducerStateTable m_applBufferEgressProfileListTable;
+
     bool m_pgfile_processed;
 
     pg_profile_lookup_t m_pgProfileLookup;
@@ -46,6 +56,10 @@ private:
     void readPgProfileLookupFile(std::string);
     task_process_status doCableTask(std::string port, std::string cable_length);
     task_process_status doSpeedUpdateTask(std::string port, std::string speed);
+    void doBufferTableTask(Consumer &consumer, ProducerStateTable &applTable);
+
+    void transformSeperator(std::string &name);
+    void transformReference(std::string &name);
 
     void doTask(Consumer &consumer);
 };
--- a/cfgmgr/buffermgrd.cpp
+++ b/cfgmgr/buffermgrd.cpp
@@ -7,13 +7,17 @@
 #include "exec.h"
 #include "schema.h"
 #include "buffermgr.h"
+#include "buffermgrdyn.h"
 #include <fstream>
 #include <iostream>
+#include "json.h"
+#include "json.hpp"
 
 using namespace std;
 using namespace swss;
+using json = nlohmann::json;
 
-/* select() function timeout retry time, in millisecond */
+/* SELECT() function timeout retry time, in millisecond */
 #define SELECT_TIMEOUT 1000
 
 /*
@@ -34,22 +38,84 @@ mutex gDbMutex;
 
 void usage()
 {
-    cout << "Usage: buffermgrd -l pg_lookup.ini" << endl;
-    cout << "       -l pg_lookup.ini: PG profile look up table file (mandatory)" << endl;
-    cout << "       format: csv" << endl;
-    cout << "       values: 'speed, cable, size, xon,  xoff, dynamic_threshold, xon_offset'" << endl;
+    cout << "Usage: buffermgrd <-l pg_lookup.ini|-a asic_table.json [-p peripheral_table.json]>" << endl;
+    cout << "       -l pg_lookup.ini: PG profile look up table file (mandatory for static mode)" << endl;
+    cout << "           format: csv" << endl;
+    cout << "           values: 'speed, cable, size, xon,  xoff, dynamic_threshold, xon_offset'" << endl;
+    cout << "       -a asic_table.json: ASIC-specific parameters definition (mandatory for dynamic mode)" << endl;
+    cout << "       -p peripheral_table.json: Peripheral (eg. gearbox) parameters definition (mandatory for dynamic mode)" << endl;
+}
+
+void dump_db_item(KeyOpFieldsValuesTuple &db_item)
+{
+    SWSS_LOG_DEBUG("db_item: [");
+    SWSS_LOG_DEBUG("\toperation: %s", kfvOp(db_item).c_str());
+    SWSS_LOG_DEBUG("\thash: %s", kfvKey(db_item).c_str());
+    SWSS_LOG_DEBUG("\tfields: [");
+    for (auto fv: kfvFieldsValues(db_item))
+        SWSS_LOG_DEBUG("\t\tfield: %s value: %s", fvField(fv).c_str(), fvValue(fv).c_str());
+    SWSS_LOG_DEBUG("\t]");
+    SWSS_LOG_DEBUG("]");
+}
+
+void write_to_state_db(shared_ptr<vector<KeyOpFieldsValuesTuple>> db_items_ptr)
+{
+    DBConnector db("STATE_DB", 0, true);
+    auto &db_items = *db_items_ptr;
+    for (auto &db_item : db_items)
+    {
+        dump_db_item(db_item);
+
+        string key = kfvKey(db_item);
+        size_t pos = key.find(":");
+        if ((string::npos == pos) || ((key.size() - 1) == pos))
+        {
+            SWSS_LOG_ERROR("Invalid formatted hash:%s\n", key.c_str());
+            return;
+        }
+        string table_name = key.substr(0, pos);
+        string key_name = key.substr(pos + 1);
+        Table stateTable(&db, table_name);
+
+        stateTable.set(key_name, kfvFieldsValues(db_item), SET_COMMAND);
+    }
+}
+
+shared_ptr<vector<KeyOpFieldsValuesTuple>> load_json(string file)
+{
+    try
+    {
+        ifstream json(file);
+        auto db_items_ptr = make_shared<vector<KeyOpFieldsValuesTuple>>();
+
+        if (!JSon::loadJsonFromFile(json, *db_items_ptr))
+        {
+            db_items_ptr.reset();
+            return nullptr;
+        }
+
+        return db_items_ptr;
+    }
+    catch (...)
+    {
+        SWSS_LOG_WARN("Loading file %s failed", file.c_str());
+        return nullptr;
+    }
 }
 
 int main(int argc, char **argv)
 {
     int opt;
     string pg_lookup_file = "";
+    string asic_table_file = "";
+    string peripherial_table_file = "";
+    string json_file = "";
     Logger::linkToDbNative("buffermgrd");
     SWSS_LOG_ENTER();
 
     SWSS_LOG_NOTICE("--- Starting buffermgrd ---");
 
-    while ((opt = getopt(argc, argv, "l:h")) != -1 )
+    while ((opt = getopt(argc, argv, "l:a:p:h")) != -1 )
     {
         switch (opt)
         {
@@ -59,32 +125,86 @@ int main(int argc, char **argv)
         case 'h':
             usage();
             return 1;
+        case 'a':
+            asic_table_file = optarg;
+            break;
+        case 'p':
+            peripherial_table_file = optarg;
+            break;
         default: /* '?' */
             usage();
             return EXIT_FAILURE;
         }
     }
 
-    if (pg_lookup_file.empty())
-    {
-        usage();
-        return EXIT_FAILURE;
-    }
-
     try
     {
-        vector<string> cfg_buffer_tables = {
-            CFG_PORT_TABLE_NAME,
-            CFG_PORT_CABLE_LEN_TABLE_NAME,
-        };
+        std::vector<Orch *> cfgOrchList;
+        bool dynamicMode = false;
+        shared_ptr<vector<KeyOpFieldsValuesTuple>> db_items_ptr;
 
         DBConnector cfgDb("CONFIG_DB", 0);
         DBConnector stateDb("STATE_DB", 0);
+        DBConnector applDb("APPL_DB", 0);
 
-        BufferMgr buffmgr(&cfgDb, &stateDb, pg_lookup_file, cfg_buffer_tables);
+        if (!asic_table_file.empty())
+        {
+            // Load the json file containing the SWITCH_TABLE
+            db_items_ptr = load_json(asic_table_file);
+            if (nullptr != db_items_ptr)
+            {
+                write_to_state_db(db_items_ptr);
+                db_items_ptr.reset();
 
-        // TODO: add tables in stateDB which interface depends on to monitor list
-        std::vector<Orch *> cfgOrchList = {&buffmgr};
+                if (!peripherial_table_file.empty())
+                {
+                    //Load the json file containing the PERIPHERIAL_TABLE
+                    db_items_ptr = load_json(peripherial_table_file);
+                    if (nullptr != db_items_ptr)
+                        write_to_state_db(db_items_ptr);
+                }
+
+                dynamicMode = true;
+            }
+        }
+
+        if (dynamicMode)
+        {
+            vector<TableConnector> buffer_table_connectors = {
+                TableConnector(&cfgDb, CFG_PORT_TABLE_NAME),
+                TableConnector(&cfgDb, CFG_PORT_CABLE_LEN_TABLE_NAME),
+                TableConnector(&cfgDb, CFG_BUFFER_POOL_TABLE_NAME),
+                TableConnector(&cfgDb, CFG_BUFFER_PROFILE_TABLE_NAME),
+                TableConnector(&cfgDb, CFG_BUFFER_PG_TABLE_NAME),
+                TableConnector(&cfgDb, CFG_BUFFER_QUEUE_TABLE_NAME),
+                TableConnector(&cfgDb, CFG_BUFFER_PORT_INGRESS_PROFILE_LIST_NAME),
+                TableConnector(&cfgDb, CFG_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME),
+                TableConnector(&cfgDb, CFG_DEFAULT_LOSSLESS_BUFFER_PARAMETER),
+                TableConnector(&stateDb, STATE_BUFFER_MAXIMUM_VALUE_TABLE)
+            };
+            cfgOrchList.emplace_back(new BufferMgrDynamic(&cfgDb, &stateDb, &applDb, buffer_table_connectors, db_items_ptr));
+        }
+        else if (!pg_lookup_file.empty())
+        {
+            vector<string> cfg_buffer_tables = {
+                CFG_PORT_TABLE_NAME,
+                CFG_PORT_CABLE_LEN_TABLE_NAME,
+                CFG_BUFFER_POOL_TABLE_NAME,
+                CFG_BUFFER_PROFILE_TABLE_NAME,
+                CFG_BUFFER_PG_TABLE_NAME,
+                CFG_BUFFER_QUEUE_TABLE_NAME,
+                CFG_BUFFER_PORT_INGRESS_PROFILE_LIST_NAME,
+                CFG_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME
+            };
+            cfgOrchList.emplace_back(new BufferMgr(&cfgDb, &applDb, pg_lookup_file, cfg_buffer_tables));
+        }
+        else
+        {
+            usage();
+            return EXIT_FAILURE;
+        }
+
+        auto buffmgr = cfgOrchList[0];
 
         swss::Select s;
         for (Orch *o : cfgOrchList)
@@ -106,7 +226,7 @@ int main(int argc, char **argv)
             }
             if (ret == Select::TIMEOUT)
             {
-                buffmgr.doTask();
+                buffmgr->doTask();
                 continue;
             }
 
--- a/cfgmgr/buffermgrdyn.cpp
+++ b/cfgmgr/buffermgrdyn.cpp
@@ -910,10 +910,8 @@ task_process_status BufferMgrDynamic::doUpdateStaticProfileTask(buffer_profile_t
     return task_process_status::task_success;
 }
 
-task_process_status BufferMgrDynamic::handleBufferMaxParam(Consumer &consumer)
+task_process_status BufferMgrDynamic::handleBufferMaxParam(KeyOpFieldsValuesTuple &t)
 {
-    auto it = consumer.m_toSync.begin();
-    KeyOpFieldsValuesTuple t = it->second;
     string op = kfvOp(t);
 
     if (op == SET_COMMAND)
@@ -940,10 +938,8 @@ task_process_status BufferMgrDynamic::handleBufferMaxParam(Consumer &consumer)
     return task_process_status::task_success;
 }
 
-task_process_status BufferMgrDynamic::handleDefaultLossLessBufferParam(Consumer &consumer)
+task_process_status BufferMgrDynamic::handleDefaultLossLessBufferParam(KeyOpFieldsValuesTuple &t)
 {
-    auto it = consumer.m_toSync.begin();
-    KeyOpFieldsValuesTuple t = it->second;
     string op = kfvOp(t);
 
     if (op == SET_COMMAND)
@@ -961,10 +957,8 @@ task_process_status BufferMgrDynamic::handleDefaultLossLessBufferParam(Consumer 
     return task_process_status::task_success;
 }
 
-task_process_status BufferMgrDynamic::handleCableLenTable(Consumer &consumer)
+task_process_status BufferMgrDynamic::handleCableLenTable(KeyOpFieldsValuesTuple &t)
 {
-    auto it = consumer.m_toSync.begin();
-    KeyOpFieldsValuesTuple t = it->second;
     string op = kfvOp(t);
 
     task_process_status task_status = task_process_status::task_success;
@@ -1065,10 +1059,8 @@ task_process_status BufferMgrDynamic::handleCableLenTable(Consumer &consumer)
 // 3. if mtu isn't configured, take the default value
 // 4. if speed_updated or mtu_updated, update headroom size
 //    elif admin_status_updated, update buffer pool size
-task_process_status BufferMgrDynamic::handlePortTable(Consumer &consumer)
+task_process_status BufferMgrDynamic::handlePortTable(KeyOpFieldsValuesTuple &t)
 {
-    auto it = consumer.m_toSync.begin();
-    KeyOpFieldsValuesTuple t = it->second;
     auto &port = kfvKey(t);
     string op = kfvOp(t);
     bool speed_updated = false, mtu_updated = false, admin_status_updated = false;
@@ -1150,11 +1142,9 @@ task_process_status BufferMgrDynamic::handlePortTable(Consumer &consumer)
     return task_status;
 }
 
-task_process_status BufferMgrDynamic::handleBufferPoolTable(Consumer &consumer)
+task_process_status BufferMgrDynamic::handleBufferPoolTable(KeyOpFieldsValuesTuple &tuple)
 {
     SWSS_LOG_ENTER();
-    auto it = consumer.m_toSync.begin();
-    KeyOpFieldsValuesTuple tuple = it->second;
     string &pool = kfvKey(tuple);
     string op = kfvOp(tuple);
     vector<FieldValueTuple> fvVector;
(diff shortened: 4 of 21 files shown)


cfgmgr/buffermgr.cpp
```diff
@@ -194,20 +201,152 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, string speed)
...
+            {
+                SWSS_LOG_INFO("Inserting field %s value %s", fvField(i).c_str(), fvValue(i).c_str());
+                //transform the separator in values from "|" to ":"
+                if (fvField(i) == "pool")
+                    transformReference(fvValue(i));
+                if (fvField(i) == "profile")
+                    transformReference(fvValue(i));
+                if (fvField(i) == "profile_list")
+                    transformReference(fvValue(i));
+                fvVector.emplace_back(FieldValueTuple(fvField(i), fvValue(i)));
```
neethajohn, line 294: we need to take into account the change in table names as well

orchagent/bufferorch.cpp
```diff
@@ -26,41 +26,62 @@ static const vector<sai_buffer_pool_stat_t> bufferPoolWatermarkStatIds =
...
 {
     SWSS_LOG_ENTER();
     initTableHandlers();
-    initBufferReadyLists(db);
+    initBufferReadyLists(confDb);
     initFlexCounterGroupTable();
+    initBufferConstants();
+};
+
+BufferOrch::BufferOrch(DBConnector *applDb, DBConnector *confDb, DBConnector *stateDb, vector<TableConnector> &tableConnectors) :
```
neethajohn, line 53: remove this one. orchdaemon calls the other constructor

orchagent/bufferorch.cpp
```diff
@@ -80,24 +101,49 @@ void BufferOrch::initBufferReadyList(Table& table)
...
-
         auto tokens = tokenize(key, config_db_key_delimiter);
         if (tokens.size() != 2)
         {
             SWSS_LOG_ERROR("Wrong format of a table '%s' key '%s'. Skip it", table.getTableName().c_str(), key.c_str());
             continue;
         }
 
+        // We need transform the key from config db format to appl db format
+        auto appldb_key = tokens[0] + delimiter + tokens[1];
```
neethajohn, line 112: account for the table name change

orchagent/bufferorch.h
```diff
@@ -27,11 +27,13 @@ const string buffer_profile_field_name      = "profile";
...
 const string buffer_value_egress            = "egress";
 const string buffer_profile_list_field_name = "profile_list";
+const string buffer_headroom_type_field_name= "headroom_type";
 
 class BufferOrch : public Orch
 {
 public:
-    BufferOrch(DBConnector *db, vector<string> &tableNames);
+    BufferOrch(DBConnector *applDb, DBConnector *confDb, DBConnector *stateDb, vector<string> &tableNames);
+    BufferOrch(DBConnector *applDb, DBConnector *confDb, DBConnector *stateDb, vector<TableConnector> &tableConnectors);
```
neethajohn, line 36: remove

cfgmgr/buffermgr.cpp
```diff
@@ -194,20 +201,152 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, string speed)
...
+        doBufferTableTask(consumer, m_applBufferIngressProfileListTable);
+        return;
+    }
+
+    if (table_name == CFG_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME)
+    {
+        doBufferTableTask(consumer, m_applBufferEgressProfileListTable);
+        return;
+    }
+
```
neethajohn, line 349: better to move this after the lossless pg profiles are generated to avoid the second copy during init phase

cfgmgr/buffermgr.cpp
```diff
@@ -194,20 +201,124 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, string speed)
...
+ * The following tables have to be moved to APPL_DB because they reference
+ * some entries that have been moved to APPL_DB
+ *  - BUFFER_QUEUE
+ *  - BUFFER_PORT_INGRESS_PROFILE_LIST
+ *  - BUFFER_PORT_EGRESS_PROFILE_LIST   
+ * One thing we need to handle is to transform the separator from | to :
+ * The following items contain separator:
+ *  - keys of each item
+ *  - pool in BUFFER_POOL
+ *  - profile in BUFFER_PG
```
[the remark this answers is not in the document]
stephenxs: fixed

orchagent/bufferorch.cpp
stephenxs: Fixed.

orchagent/bufferorch.cpp
stephenxs: Actually, we should not change this because this consists of keys only without the table name part.

orchagent/bufferorch.h
stephenxs: fixed

orchagent/bufferorch.cpp
```diff
@@ -289,6 +342,13 @@ task_process_status BufferOrch::processBufferPool(Consumer &consumer)
             else if (field == buffer_pool_mode_field_name)
             {
                 string mode = value;
+                if (SAI_NULL_OBJECT_ID != sai_object)
+                {
+                    // We should skip the pool type because it's create only
```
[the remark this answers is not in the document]
stephenxs: this one, along with the following comments error, will be addressed in [#1430](https://github.com/Azure/sonic-swss/pull/1430). This snippet of code will be adjusted after that has been merged

orchagent/bufferorch.cpp
```diff
@@ -80,24 +101,49 @@ void BufferOrch::initBufferReadyList(Table& table)
...
+{
+    sai_status_t status;
+    sai_attribute_t attr;
+
+    attr.id = SAI_SWITCH_ATTR_TOTAL_BUFFER_SIZE;
+
+    status = sai_switch_api->get_switch_attribute(gSwitchId, 1, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to get Maxiuum memory size, rv:%d", status);
```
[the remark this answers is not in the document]
stephenxs: fixed

cfgmgr/buffermgr.cpp
```diff
@@ -194,20 +201,124 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, string speed)
...
+ *  - BUFFER_PG
+ * The following tables have to be moved to APPL_DB because they reference
+ * some entries that have been moved to APPL_DB
+ *  - BUFFER_QUEUE
+ *  - BUFFER_PORT_INGRESS_PROFILE_LIST
+ *  - BUFFER_PORT_EGRESS_PROFILE_LIST   
+ * One thing we need to handle is to transform the separator from | to :
+ * The following items contain separator:
+ *  - keys of each item
+ *  - pool in BUFFER_POOL
```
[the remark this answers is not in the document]
stephenxs: fixed

cfgmgr/buffermgr.cpp
stephenxs: I'm afraid that won't yield too much benefit because
- `doBufferTableTask` is called by `doTask` which is called for each **entry** in a table it listens to but not for the entire table.
- after a pg profile is generated, there will always be a call to `doTask` for the entry regardless of the order between pg profile generating and those calls to `doBufferTableTask`.

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1620 @@
...
+    m_bufferTableHandlerMap.insert(buffer_handler_pair(CFG_BUFFER_PORT_INGRESS_PROFILE_LIST_NAME, &BufferMgrDynamic::handleBufferPortIngressProfileListTable));
+    m_bufferTableHandlerMap.insert(buffer_handler_pair(CFG_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME, &BufferMgrDynamic::handleBufferPortEgressProfileListTable));
+    m_bufferTableHandlerMap.insert(buffer_handler_pair(CFG_PORT_TABLE_NAME, &BufferMgrDynamic::handlePortTable));
+    m_bufferTableHandlerMap.insert(buffer_handler_pair(CFG_PORT_CABLE_LEN_TABLE_NAME, &BufferMgrDynamic::handleCableLenTable));
+}
+
+// APIs to handle variant kinds of keys
+
+// Transform key from CONFIG_DB format to APPL_DB format
+void BufferMgrDynamic::transformSeperator(string &name)
```
kperumalbfn, line 179: these transform APIs could be common for both static and dynamic buffer files.

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1620 @@
...
+    if (!m_mmuSize.empty())
+        recalculateSharedBufferPool();
+}
+
+// For buffer pool, only size can be updated on-the-fly
+void BufferMgrDynamic::updateBufferPoolToDb(const string &name, const buffer_pool_t &pool)
+{
+    vector<FieldValueTuple> fvVector;
+
+    if (pool.ingress)
```
kperumalbfn, line 396: Can we keep a knob to not update buffer pool for lossy and lossless and just update the profile based on speed/cable length??? Some asic can fix the upper max lossy or lossless pool size, but don't want to change while packets are flowing.

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1620 @@
...
+        fvVector.emplace_back(make_pair("type", "egress"));
+
+    fvVector.emplace_back(make_pair("mode", pool.mode));
+
+    SWSS_LOG_INFO("Buffer pool %s is initialized", name.c_str());
+
+    fvVector.emplace_back(make_pair("size", pool.total_size));
+
+    m_applBufferPoolTable.set(name, fvVector);
+
```
kperumalbfn, line 408: Can we do error check for the max buffer pool sizes for all the configured pools and that shouldn't exceed max mmu size?

cfgmgr/buffermgrdyn.cpp
stephenxs: Yes. It can be done by providing ‘size’ for those pools.

cfgmgr/buffermgrd.cpp
```diff
@@ -34,22 +38,84 @@ mutex gDbMutex;
...
+    string asic_table_file = "";
+    string peripherial_table_file = "";
+    string json_file = "";
     Logger::linkToDbNative("buffermgrd");
     SWSS_LOG_ENTER();
 
     SWSS_LOG_NOTICE("--- Starting buffermgrd ---");
 
-    while ((opt = getopt(argc, argv, "l:h")) != -1 )
+    while ((opt = getopt(argc, argv, "l:h:a:p:j")) != -1 )
```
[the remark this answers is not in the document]
stephenxs: Originally, it was -j which represents the Jason for asic table.
Redundant argument. Will remove.

cfgmgr/buffermgr.h
```diff
@@ -13,6 +13,8 @@ namespace swss {
 #define INGRESS_LOSSLESS_PG_POOL_NAME "ingress_lossless_pool"
 #define LOSSLESS_PGS "3-4"
 
+#define BUFFERMGR_TIMER_PERIOD 10
```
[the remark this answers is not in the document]
stephenxs: 1. Warm reboot, wait for all buffer stuff to be calculated. Something like reconciling timer. This is for both dynamic and traditional mode.
2. Check the buffer pool calculations periodically. Dynamic mode only
3. After system started, check whether max my size is ready periodically. After it’s ready buffer pools can be calculated dynamically

cfgmgr/buffermgr.cpp
```diff
@@ -194,20 +201,153 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, string speed)
...
+        {
+            vector<FieldValueTuple> fvVector;
+
+            SWSS_LOG_INFO("Inserting entry %s from CONFIG_DB to APPL_DB", key.c_str());
+
+            for (auto i : kfvFieldsValues(t))
+            {
+                SWSS_LOG_INFO("Inserting field %s value %s", fvField(i).c_str(), fvValue(i).c_str());
+                //transform the separator in values from "|" to ":"
+                if (fvField(i) == "pool")
```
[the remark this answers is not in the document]
stephenxs: Yes. All the buffer tables from CONFIG DB are copied to APPL DB with references modified properly.

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1620 @@
...
+        m_portInitDone(false),
+        m_firstTimeCalculateBufferPool(true)
+{
+    SWSS_LOG_ENTER();
+
+    // Initialize the handler map
+    initTableHandlerMap();
+    parseGearboxInfo(gearboxInfo);
+
+    string platform = getenv("ASIC_VENDOR") ? getenv("ASIC_VENDOR") : "";
```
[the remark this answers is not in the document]
stephenxs: These Json files are provided by vendors. They can do that at a per SKU/platform basis.
The Kia plugin are supposed to be provided on a per asic vendor basis, which is why the asic vendor is loaded here

cfgmgr/buffermgrdyn.cpp
stephenxs: It’s a bit difficult to achieve that because all asic vendors don’t share the same logic of testing the thresholds.
I suggest doing this in these ways:
- pools with static size: make sure the configuration is correct
- pools dynamically calculated: check that in the lua plug in

cfgmgr/buffermgrdyn.cpp
stephenxs: Theoretically, yes.
To share them needs a new common class to be introduced since there is no inheriting relation between the original `BufferMgr` and `BufferMgrDyn`. It looks like an overdesign to me. So I suggest keeping as it is.

cfgmgr/buffermgrdyn.cpp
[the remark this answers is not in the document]
kperumalbfn: Vendors can provide different lua scripts, but plugin names(headroomPluginName,  bufferpoolPluginName  and checkHeadroomPluginName ) are per asic_vendor and not per SKU. How can this be different for different SKUs? Could you pls clarify?

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1620 @@
...
+    }
+}
+
+// For string "TABLE_NAME|objectname", returns "objectname"
+string BufferMgrDynamic::parseObjectNameFromKey(const string &key, size_t pos = 0)
+{
+    auto keys = tokenize(key, delimiter);
+    if (pos >= keys.size())
+    {
+        SWSS_LOG_ERROR("Failed to fetch %lu-th sector of key %s", pos, key.c_str());
```
kperumalbfn, line 218: return some null string and the caller to check for these errors and handle the errors??

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1620 @@
...
+        bufferPool.dynamic_size = true;
+        for (auto i = kfvFieldsValues(tuple).begin(); i != kfvFieldsValues(tuple).end(); i++)
+        {
+            string &field = fvField(*i);
+            string &value = fvValue(*i);
+
+            SWSS_LOG_DEBUG("field:%s, value:%s", field.c_str(), value.c_str());
+            if (field == buffer_size_field_name)
+            {
+                bufferPool.dynamic_size = false;
```
kperumalbfn, line 1154: Should we check the buffer pool size for any error if this is more than max mmu size??

cfgmgr/buffermgrdyn.h
```diff
@@ -0,0 +1,245 @@
...
+#include "orch.h"
+
+#include <map>
+#include <set>
+#include <string>
+
+namespace swss {
+
+#define INGRESS_LOSSLESS_PG_POOL_NAME "ingress_lossless_pool"
+#define DEFAULT_MTU_STR             "9100"
```
kperumalbfn, line 15: Currently, SAI default port MTU is 1514. We could maintain the same default MTU instead of jumbo mtu size.

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1620 @@
...
+                    // The cost is that if the mtu is provided in the following iteration
+                    // the current calculation is of no value and will be replaced.
+                    mtu = DEFAULT_MTU_STR;
+                }
+                task_status = refreshPriorityGroupsForPort(port, speed, cable_length, mtu);
+                break;
+
+            case PORT_READY:
+                task_status = refreshPriorityGroupsForPort(port, speed, cable_length, mtu);
+                break;
```
kperumalbfn, line 1110: For both port_table and cable_len table,we could invoke refreshPriorityGroupsForPort only when the port is in PORT_READY state. We could remove refresh in port_init state. Seems to be redundant.

cfgmgr/buffermgrdyn.cpp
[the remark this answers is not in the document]
stephenxs: The lua plugin is provided on a per vendor  basis. Different platform/SKUs can have different parameters, like cell size. This can be reflected in two ways:
- parameter passed by buffermgrd when calling lua plugins
- the lua plugin can also fetch these parameters via reading the tables in database, like ASIC_TABLE in STATE DB for cell size.
Lua plug-ins are not provided for each platform/SKU.

cfgmgr/buffermgrdyn.h
stephenxs: But the SONiC MTU is 9100. So we need to align it with SONiC

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1620 @@
...
+    SWSS_LOG_ENTER();
+
+    // Initialize the handler map
+    initTableHandlerMap();
+    parseGearboxInfo(gearboxInfo);
+
+    string platform = getenv("ASIC_VENDOR") ? getenv("ASIC_VENDOR") : "";
+    if (platform == "")
+    {
+        SWSS_LOG_ERROR("Platform environment variable is not defined");
```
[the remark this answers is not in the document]
stephenxs: This is for the VS image in which no platform defined. Probably the message should not be in ERROR level.

cfgmgr/buffermgrdyn.cpp
stephenxs: Will fix it

cfgmgr/buffermgrdyn.cpp
stephenxs: Take `handlePortTable` as an example:
Reaching `case PORT_INITIALIZING` means:
- the current state of the port is `PORT_INITIALIZING` because it lacks some info
- an event that supplies the lacking info, `speed` in this case, is being handled and the port's state is about to be transmitted to `PORT_READY`.

Hence, the `refreshPriorityGroupsForPort` should be called for the "first" time once its state is `PORT_READY`.

keboliu:
retest this please

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1674 @@
...
+        // The format of the result:
+        // a list of strings containing key, value pairs with colon as separator
+        // each is the size of a buffer pool
+
+        for ( auto i : ret)
+        {
+            auto pairs = tokenize(i, ':');
+            if ("result" == pairs[0])
+            {
+                if ("true" != pairs[1])
```
kperumalbfn, line 626: which method checks for the valid and available headroom and update the appDb??

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1674 @@
...
+
+    SWSS_LOG_NOTICE("Remove BUFFER_PG %s (profile %s, %s)", pg_key.c_str(), bufferPg.running_profile_name.c_str(), bufferPg.configured_profile_name.c_str());
+
+    // recalculate pool size
+    checkSharedBufferPoolSize();
+
+    if (!portInfo.speed.empty() && !portInfo.cable_length.empty())
+        portInfo.state = PORT_READY;
+    else
+        portInfo.state = PORT_INITIALIZING;
```
kperumalbfn, line 859: In this case, when does the port goes to port_Ready state after lossless PG is removed?

cfgmgr/buffermgrdyn.cpp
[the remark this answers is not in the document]
stephenxs: Fixed

cfgmgr/buffermgrdyn.cpp
stephenxs: Checking added

cfgmgr/buffermgrdyn.cpp
stephenxs: Done.

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1674 @@
...
+void BufferMgrDynamic::updateBufferProfileToDb(const string &name, const buffer_profile_t &profile)
+{
+    vector<FieldValueTuple> fvVector;
+    string mode = getPgPoolMode();
+
+    if (mode.empty())
+    {
+        // this should never happen if switch initialized properly
+        SWSS_LOG_ERROR("PG lossless pool is not yet created, creating profile %s failed", name.c_str());
+        return;
```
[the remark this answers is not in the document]
stephenxs: Actually, this check can be removed because it is called from two functions:
- `allocateProfile`, in which there is already a check ahead of calling
- `doUpdateStaticProfileTask`, which is triggered by CLI (or other management plane interfaces) and the ingress_lossless_pool has been created before the CLI ready.

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1674 @@
...
+                threshold = profile.threshold;
+            }
+            else
+            {
+                threshold = m_defaultThreshold;
+            }
+            newProfile = getDynamicProfileName(speed, cable_length, mtu, threshold, gearbox_model);
+            auto rc = allocateProfile(speed, cable_length, mtu, threshold, gearbox_model, newProfile);
+            if (task_process_status::task_success != rc)
+                return rc;
```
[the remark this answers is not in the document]
stephenxs: There is a piece of message in [allocateProfile](https://github.com/Azure/sonic-swss/pull/1338/files/80b5b1f1b4dbf7b16415882911cf7253f86c35f5#diff-908ebd83566b58a4ad7eb1133cac45f1fa4f328ada69180a47e496b25a34cee7R512) in case it doesn't return task_success

cfgmgr/buffermgrdyn.cpp
stephenxs: The lua plugin `buffer_check_headroom_<vendor>` will check it.
After the check passed, the PG will be updated by `updateBufferPgToDb`.
I believe the big picture of the logic can be found in [refreshPriorityGroupsForPort](https://github.com/Azure/sonic-swss/pull/1338/files/80b5b1f1b4dbf7b16415882911cf7253f86c35f5#diff-908ebd83566b58a4ad7eb1133cac45f1fa4f328ada69180a47e496b25a34cee7R670)

cfgmgr/buffermgrdyn.cpp
stephenxs: For headroom override (statically configured profile on a port), it will go to `PORT_READY` once both `speed` and `cable length` are configured.
For dynamic calculated profile, this sentence won't be reached because only if both `speed` and `cable length` are provided will the headroom (profile) be calculated.

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1655 @@
...
+        string &speed = portInfo.speed;
+
+        if (cable_length.empty() || speed.empty())
+        {
+            SWSS_LOG_WARN("Cable length for %s hasn't been configured yet, unable to calculate headroom", port.c_str());
+            // We don't retry here because it doesn't make sense until the cable length is configured.
+            return task_process_status::task_success;
+        }
+
+        if (speed_updated || mtu_updated)
```
kperumalbfn, line 1110: I believe we should call refreshPGs for cable length update even if speed or MTU is not updated.

stephenxs: Not really.
This function is dedicated to handling the speed or MTU update.
Cable length update is handled in another function, handleCableLenTable.

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1655 @@
...
+            auto portName = parseObjectNameFromKey(key);
+            auto &port = m_portInfoLookup[portName];
+
+            if (portsChecked.find(portName) != portsChecked.end())
+                continue;
+
+            SWSS_LOG_DEBUG("Checking PG %s for dynamic profile %s", key.c_str(), profileName.c_str());
+            portsChecked.insert(portName);
+
+            refreshPriorityGroupsForPort(portName, port.speed, port.cable_length, port.mtu);
```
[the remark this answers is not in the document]
stephenxs: Fixed

cfgmgr/buffermgrdyn.cpp
```diff
@@ -0,0 +1,1655 @@
...
+            if (fvField(i) == "mmu_size")
+            {
+                m_mmuSize = fvValue(i);
+                if (!m_mmuSize.empty())
+                {
+                    m_mmuSizeNumber = atol(m_mmuSize.c_str());
+                }
+                if (0 == m_mmuSizeNumber)
+                {
+                    SWSS_LOG_ERROR("BUFFER_MAX_PARAM: Got invalid mmu size %s", m_mmuSize.c_str());
```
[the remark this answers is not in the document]
stephenxs: Fixed

tests/test_buffer.py
```diff
@@ -0,0 +1,215 @@
...
+
+        # remove lossless PG 3-4 on interface
+        dvs.runcmd("config interface buffer priority-group lossless remove Ethernet0 3-4")
+
+    @pytest.mark.skip(reason="Traditional buffer model runs on VS image currently.")
+    def test_MultipleLosslessPg(self, dvs, testlog):
+        self.setup_db(dvs)
+
+        # configure lossless PG 3-4 on interface
+        dvs.runcmd("config interface buffer priority-group lossless add Ethernet0 3-4")
```
kperumalbfn, line 133: Could you add a negative test with multiple lossless pools with size more than MMU size?

stephenxs: By default, only one lossless pool is supported. dynamic buffer calculation assumes only `ingress_lossless_pool` is used.
For the negative test, in general, I think we can verify it in the following steps:
- Get the `mmusize` from `STATE_DB`
- Hack the plugin which is responsible for calculating the pool size, making it return a super large value, which requires
  - prepare a hacked version of the plugin
  - copy it to the DUT and replace the original one with the new one in the swss docker
  - reload config to make the hacked plugin loaded
- Check whether the large pool size is deployed
- Clean up: remove the hacked plugin

As it isn't a simple test, I suggest adding this test in the future.

kperumalbfn (approved):

tests/test_buffer.py
```diff
@@ -0,0 +1,215 @@
...
+        self.app_db.wait_for_field_match("BUFFER_PG_TABLE", "Ethernet0:3-4", {"profile": "[BUFFER_PROFILE_TABLE:" + expectedProfile + "]"})
+
+        dvs.runcmd("config interface buffer priority-group lossless set Ethernet0 3-4 test")
+        self.app_db.wait_for_field_match("BUFFER_PG_TABLE", "Ethernet0:3-4", {"profile": "[BUFFER_PROFILE_TABLE:test]"})
+        dvs.runcmd("config interface buffer priority-group lossless add Ethernet0 6 test")
+        self.app_db.wait_for_field_match("BUFFER_PG_TABLE", "Ethernet0:6", {"profile": "[BUFFER_PROFILE_TABLE:test]"})
+
+        dvs.runcmd("config interface buffer priority-group lossless remove Ethernet0")
+        self.app_db.wait_for_deleted_entry("BUFFER_PG_TABLE", "Ethernet0:3-4")
+        self.app_db.wait_for_deleted_entry("BUFFER_PG_TABLE", "Ethernet0:6")
```
neethajohn, line 197: can you also remove the static profile and check whether it gets removed?

tests/test_buffer.py
```diff
@@ -0,0 +1,215 @@
...
+
+        dvs.runcmd("config interface cable-length Ethernet0 " + self.originalCableLen)
+        self.app_db.wait_for_deleted_entry("BUFFER_PROFILE_TABLE", expectedProfile)
+        expectedProfile = self.make_lossless_profile_name(self.originalSpeed, self.originalCableLen)
+        self.app_db.wait_for_entry("BUFFER_PROFILE_TABLE", expectedProfile)
+        self.app_db.wait_for_field_match("BUFFER_PG_TABLE", "Ethernet0:3-4", {"profile": "[BUFFER_PROFILE_TABLE:" + expectedProfile + "]"})
+
+        # remove lossless PG 3-4 on interface
+        dvs.runcmd("config interface buffer priority-group lossless remove Ethernet0 3-4")
+
```
neethajohn, line 210: Please add testcase for mtu change, port admin status change and non default alpha as well?

tests/test_buffer.py
```diff
@@ -0,0 +1,215 @@
...
+        fvs = {"mmu_size": "12766208"}
+        self.state_db.create_entry("BUFFER_MAX_PARAM_TABLE", "global", fvs)
+
+        self.initialized = True
+
+    def make_lossless_profile_name(self, speed, cable_length):
+        return "pg_lossless_" + speed + "_" + cable_length + "_profile"
+
+    @pytest.mark.skip(reason="Traditional buffer model runs on VS image currently.")
+    def test_changeSpeed(self, dvs, testlog):
```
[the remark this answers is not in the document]
stephenxs: So far we only verify whether there is a new profile generated according to the new tuple. The numbers in the profile won't be verified.

tests/test_buffer.py
```diff
@@ -0,0 +1,215 @@
...
+        dvs.runcmd("config interface buffer priority-group lossless add Ethernet0 3-4")
+
+        # Change speed to speed1 and verify whether the profile has been updated
+        dvs.runcmd("config interface speed Ethernet0 " + self.speedToTest1)
+
+        expectedProfile = self.make_lossless_profile_name(self.speedToTest1, self.originalCableLen)
+        self.app_db.wait_for_entry("BUFFER_PROFILE_TABLE", expectedProfile)
+
+        # Check whether bufer pg align
+        bufferPg = self.app_db.wait_for_field_match("BUFFER_PG_TABLE", "Ethernet0:3-4", {"profile": "[BUFFER_PROFILE_TABLE:" + expectedProfile + "]"})
```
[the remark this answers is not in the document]
stephenxs: Hi @neethajohn 
The only difficulty is that the mapping between buffer profile name and the SAI oid is stored in orchagent and isn't seen from outside. Probably this can be achieved in the following steps:
1. fetch all the set of all the buffer profiles from `ASIC_STATE:BUFFER_PROFILE` table at the beginning of the test. notate it as `S0`
2. if a new buffer profile should be created, fetch the new set of all the buffer profiles as `S1`. The new profile should be `S1 - S0`
3. at the beginning of the test, we must make sure that a new profile has been created otherwise we can not tell which entry represents the new created profile
4. during the test it's possible that an old profile is removed and a new one is created after a certain step. In that case we can treat the sai object fetched in the previous step as the one for the new profile. for example, in the following scenario:
    - only one port has 1500 as MTU, thus only one profile has the name convention `*_mtu1500_*`.
    - after that, the MTU has been updated to 3000. so new profile `*_mtu3000_*` is created and the old one `*_mtu1500_*` has been removed. It is very likely that the SAI OID for `*_mtu1500_*` is reused for `*_mtu3000_*`. Hence, we are not able to distinguish which SAI OID stands for the `*_mtu3000_*` profile by comparing two sets and have to assume the previous SAI OID is reused.
5. there is still one scenario in which we aren't able to identify the expected SAI OID in the new step. For example, in case the new profile which is expected to be created during one step of test has already existed. In this case, we have to skip the verification of the sai part.

Any suggestions?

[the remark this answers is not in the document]
neethajohn: We should check for the number of buffer profiles in asic db and use that to determine if a new profile has been added or not wherever it is possible for us to do it.

tests/test_buffer.py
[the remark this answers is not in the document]
neethajohn: We should at least do it in the sonic-mgmt tests. Currently we don't have any tests which check whether we are updating the headroom and pool size correctly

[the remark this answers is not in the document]
stephenxs: Ok. Let’s do it in this way

tests/test_buffer.py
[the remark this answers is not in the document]
stephenxs: Ok. Let’s do it in this approach.

tests/test_buffer.py
stephenxs: Done

neethajohn:
@stephenxs , Dynamic buffer tests are failing. can you check?

stephenxs:
1. rebase and smash all the commits before the one @kperumalbfn approved (inclusive) into one in order to make it clear and resolve the conflict. No change to the code.
2. Adjust test cases according to @neethajohn 's comments. (see the last two comments)

stephenxs:
> @stephenxs , Dynamic buffer tests are failing. can you check?

Yes, @neethajohn , it’s expected because the sonic build image and sonic utilities haven’t been merged yet. We need the vsimage supporting dynamic buffer to pass the test.

tests/test_buffer_dynamic.py
```diff
@@ -39,12 +51,65 @@ def setup_db(self, dvs):
...
+        extra = ""
+        if mtu:
+            extra += "_mtu" + mtu
+        if dynamic_th:
+            extra += "_th" + dynamic_th
+
+        return "pg_lossless_" + speed + "_" + cable_length + extra + "_profile"
 
-    @pytest.mark.skip(reason="Traditional buffer model runs on VS image currently.")
     def test_changeSpeed(self, dvs, testlog):
```
neethajohn, line 113: check in asic db for this test case is missing

tests/test_buffer_dynamic.py
```diff
@@ -97,6 +161,7 @@ def test_changeCableLen(self, dvs, testlog):
         dvs.runcmd("config interface cable-length Ethernet0 " + self.cableLenTest1)
         expectedProfile = self.make_lossless_profile_name(self.originalSpeed, self.cableLenTest1)
         self.app_db.wait_for_entry("BUFFER_PROFILE_TABLE", expectedProfile)
+        self.check_new_profile_in_asic_db(dvs, expectedProfile)
         self.app_db.wait_for_field_match("BUFFER_PG_TABLE", "Ethernet0:3-4", {"profile": "[BUFFER_PROFILE_TABLE:" + expectedProfile + "]"})
 
         # Remove the lossless PGs
```
neethajohn, line 167: after the cable length is changed to the new one, can't we check asic db to ensure that profile is removed after L174?

tests/test_buffer_dynamic.py
```diff
@@ -208,8 +293,59 @@ def test_headroomOverride(self, dvs, testlog):
...
+        dvs.runcmd("config interface buffer priority-group lossless add Ethernet0 3-4")
+
+        self.app_db.wait_for_entry("BUFFER_PG_TABLE", "Ethernet0:3-4")
+        self.app_db.wait_for_entry("BUFFER_PROFILE_TABLE", expectedProfileMtu)
+        self.check_new_profile_in_asic_db(dvs, expectedProfileMtu)
+        self.app_db.wait_for_field_match("BUFFER_PG_TABLE", "Ethernet0:3-4", {"profile": "[BUFFER_PROFILE_TABLE:{}]".format(expectedProfileMtu)})
+
+        dvs.runcmd("config interface mtu Ethernet0 {}".format(default_mtu))
+
+        self.app_db.wait_for_deleted_entry("BUFFER_PROFILE_TABLE", expectedProfileMtu)
```
neethajohn, line 317: check in asic db for profile deleted?

tests/test_buffer_dynamic.py
```diff
@@ -208,8 +293,59 @@ def test_headroomOverride(self, dvs, testlog):
...
+
+    def test_nonDefaultAlpha(self, dvs, testlog):
+        self.setup_db(dvs)
+
+        test_dynamic_th_1 = '1'
+        expectedProfile_th1 = self.make_lossless_profile_name(self.originalSpeed, self.originalCableLen, dynamic_th = test_dynamic_th_1)
+        test_dynamic_th_2 = '2'
+        expectedProfile_th2 = self.make_lossless_profile_name(self.originalSpeed, self.originalCableLen, dynamic_th = test_dynamic_th_2)
+
+        dvs.runcmd("config buffer profile add non-default-dynamic --dynamic_th {}".format(test_dynamic_th_1))
```
neethajohn, line 331: remove. same as L333

tests/test_buffer_dynamic.py
```diff
@@ -208,8 +293,59 @@ def test_headroomOverride(self, dvs, testlog):
...
+        dvs.runcmd("config interface buffer priority-group lossless add Ethernet0 3-4 non-default-dynamic")
+
+        self.app_db.wait_for_entry("BUFFER_PROFILE_TABLE", expectedProfile_th1)
+        self.check_new_profile_in_asic_db(dvs, expectedProfile_th1)
+        self.app_db.wait_for_field_match("BUFFER_PG_TABLE", "Ethernet0:3-4", {"profile": "[BUFFER_PROFILE_TABLE:" + expectedProfile_th1 + "]"})
+
+        # modify the profile to another dynamic_th
+        dvs.runcmd("config buffer profile set non-default-dynamic --dynamic_th {}".format(test_dynamic_th_2))
+
+        self.app_db.wait_for_deleted_entry("BUFFER_PROFILE_TABLE", expectedProfile_th1)
```
neethajohn, line 345: check in asic db for profile deleted?

cfgmgr/buffermgrdyn.cpp
```diff
@@ -910,10 +910,8 @@ task_process_status BufferMgrDynamic::doUpdateStaticProfileTask(buffer_profile_t
     return task_process_status::task_success;
 }
 
-task_process_status BufferMgrDynamic::handleBufferMaxParam(Consumer &consumer)
+task_process_status BufferMgrDynamic::handleBufferMaxParam(KeyOpFieldsValuesTuple &t)
```
neethajohn, line 913: Please rename all occurrences of KeyOpFieldsValuesTuple 't' to 'tuple' so that it is more meaningful

tests/test_buffer_dynamic.py
```diff
@@ -39,12 +51,65 @@ def setup_db(self, dvs):
...
+            self.newProfileInAsicDb = diff.pop()
+        # in case diff is empty, we just treat the newProfileInAsicDb cached the latest one
+        fvs = self.app_db.get_entry("BUFFER_PROFILE_TABLE", profile)
+        if fvs.get('dynamic_th'):
+            sai_threshold_value = fvs['dynamic_th']
+            sai_threshold_mode = 'SAI_BUFFER_PROFILE_THRESHOLD_MODE_DYNAMIC'
+        else:
+            sai_threshold_value = fvs['static_th']
+            sai_threshold_mode = 'SAI_BUFFER_PROFILE_THRESHOLD_MODE_STATIC'
+        self.asic_db.wait_for_field_match("ASIC_STATE:SAI_OBJECT_TYPE_BUFFER_PROFILE", self.newProfileInAsicDb,
```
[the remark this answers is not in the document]
stephenxs: No. We can’t tell which if the new items is the one related to the new profile in that case. As only one profile should be created but two are observed, the test should fail.

tests/test_buffer_dynamic.py
stephenxs: No, we can’t. The old profile has been removed but a new one will be created almost at the same time. It’s very likely that the sairedis will reuse the OID of the removed one. In that case, we can not find a chance to check whether it is removed in asic db

tests/test_buffer_dynamic.py
stephenxs: No, we can’t. Similar as the previous one.

tests/test_buffer_mode.py
```diff
@@ -0,0 +1,14 @@
...
+import re
+
+from pprint import pprint
+from swsscommon import swsscommon
+
+class TestBufferModel(object):
+    def test_bufferModel(self, dvs, testlog):
+        config_db = dvs.get_config_db()
+        metadata = config_db.get_entry("DEVICE_METADATA", "localhost")
+        assert metadata["buffer_model"] == "traditional"
```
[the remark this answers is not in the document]
stephenxs: Not really. It’s always initiated either starting from scratch or upgraded from an old version.

tests/test_buffer_dynamic.py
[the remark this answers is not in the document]
neethajohn: How will the test fail when we see more new items? as per the check here, we are exactly looking for a diff of 1 and updating the new profile if that is the case else we are assuming that there is no diff so we are reusing the last profile.

tests/test_buffer_dynamic.py
stephenxs: The similar case as the previous one.

tests/test_buffer_mode.py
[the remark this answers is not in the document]
neethajohn: saw the changes in buildimage now

tests/test_buffer_dynamic.py
stephenxs: We can’t do that because it very unlikely the old profile is removed after speed updated because the old speed is still being referenced by other ports, which means we can not know which oid is the new one

tests/buffer_model.py
```diff
@@ -0,0 +1,70 @@
...
+    # check whether it's already running dynamic mode
+    device_meta = config_db.get_entry('DEVICE_METADATA', 'localhost')
+    if device_meta['buffer_model'] == 'dynamic':
+        return
+
+    device_meta['buffer_model'] = 'dynamic'
+    config_db.update_entry('DEVICE_METADATA', 'localhost', device_meta)
+
+    # stop the buffermgrd in order to avoid it copying tables to appl db
+    cmd_runner("supervisorctl stop buffermgrd")
```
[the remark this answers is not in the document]
stephenxs: Not really. We need dynamic buffermgrd to remove table entries from APPL DB after CONFIG DB entries have been removed.

tests/test_buffer_dynamic.py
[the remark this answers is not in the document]
stephenxs: Two scenarios:
1. This is the first time the new profile is created. It results in the newProfileInAsicDb being None, which should fail the test.
2. This is not the first time. In that case the newProfileInAsicDb should contain a valid value and the test will continue.

We can add an assert here to fail the test in the 1st case

neethajohn (dismissed):

liat-grozovik:
retest vs please

liat-grozovik:
@lguohan  and @neethajohn can you please merge this PR so we can cont with the merge flow as suggested above? i dont have the ability to do so if vs is failing and in this case it should fail until we finish the merge flow.

kperumalbfn (approved):

liat-grozovik:
retest vs please

stephenxs:
retest vs please

stephenxs:
retest this please

stephenxs:
The vs test failed only on the dynamic buffer test cases, which is expected. I think we are safe to merge it. 
@neethajohn @lguohan Could you help with this?
Thanks.

stephenxs:
retest this please

neethajohn:
@stephenxs, can you look at the buffer test failures and fix them?

stephenxs:
> @stephenxs, can you look at the buffer test failures and fix them?

Hi @neethajohn done. All tests passed

neethajohn (approved):

liat-grozovik (approved):

abdosi:
@stepanblyschak this commit is causing sonic-util-build-pr failure. Can you please check ?

https://sonic-jenkins.westus2.cloudapp.azure.com/job/common/job/sonic-utilities-build-pr/3478/console

stephenxs:
@abdosi I'm checking it.

daall:
@abdosi @stephenxs it looks like it is fixed based on the latest few PRs that are still running. I think what happened is that those PRs that failed were based on swss build #2369, which didn't have the code changes in yet, and the later ones are using https://sonic-jenkins.westus2.cloudapp.azure.com/job/vs/job/sonic-swss-build/2370/ or later, which does have the code change. 

I will look into updating the utilities test job to pull the bins and the tests from the same build rather than cloning the tests from the swss repo directly.

stephenxs:
Thanks @daall . Just checked it. You are correct.

PR #1343, opened 2020-07-06
[aclorch] Use IPv6 Next Header internally for protocol number on MLNX platform

daall:

**What I did**
I updated the qualifiers we use internally for IP_PROTOCOL to IPV6_NEXT_HEADER for Mellanox SKUs.

**Why I did it**
IPV6_NEXT_HEADER *should* be used for MIRRORV6 and L3V6 across the board, but some SAI libraries do not implement this qualifier yet. We will push the other SAI vendors to implement this qualifier so we can transition to it moving forward.

The reason this is important to do now for Mellanox is that certain ACL rules will crash if they include the IP_PROTOCOL field.

**How I verified it**
1. Install new SWSS with this change
2. Confirm that ASIC DB shows IPV6_NEXT_HEADER as a MIRRORV6 table qualifier
3. Confirm that ASIC DB shows IPV6_NEXT_HEADER as a qualifier for rules with IP_PROTOCOL
4. Confirm that rules with IP Protocol and other fields (e.g. src ipv6, src port, dscp, etc.) do not crash the ASIC

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -389,7 +389,8 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
 
     // NOTE: Temporary workaround to support matching protocol numbers on MLNX platform.
     // In a later SAI version we will transition to using NEXT_HEADER for IPv6 on all platforms.
-    string platform = getenv("platform") ? getenv("platform") : "";
+    auto platform_env_var = getenv("platform");
+    string platform = platform_env_var ? platform_env_var: "";
     if ((m_tableType == ACL_TABLE_MIRRORV6 || m_tableType == ACL_TABLE_L3V6)
             && platform == MLNX_PLATFORM_SUBSTRING
             && attr_name == MATCH_IP_PROTOCOL)
@@ -1367,7 +1368,8 @@ bool AclTable::create()
 
     // NOTE: Temporary workaround to support matching protocol numbers on MLNX platform.
     // In a later SAI version we will transition to using NEXT_HEADER for IPv6 on all platforms.
-    string platform = getenv("platform") ? getenv("platform") : "";
+    auto platform_env_var = getenv("platform");
+    string platform = platform_env_var ? platform_env_var: "";
     if ((type == ACL_TABLE_MIRRORV6 || type == ACL_TABLE_L3V6)
             && platform == MLNX_PLATFORM_SUBSTRING)
     {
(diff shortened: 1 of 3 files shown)


orchagent/aclorch.cpp
```diff
@@ -386,6 +387,16 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
         return false;
     }
 
+    // NOTE: Temporary workaround to support matching protocol numbers on MLNX platform.
+    // In a later SAI version we will transition to using NEXT_HEADER for IPv6 on all platforms.
+    string platform = getenv("platform") ? getenv("platform") : "";
```
qiluo-msft, line 392: >getenv [](start = 43, length = 6)

You called `getenv` twice? Could you do once?

qiluo-msft: I see one more in AclOrch::init()

---
In reply to: [450405491](https://github.com/Azure/sonic-swss/pull/1343#discussion_r450405491) [](ancestors = 450405491)

orchagent/aclorch.cpp
```diff
@@ -1354,9 +1365,22 @@ bool AclTable::create()
     attr.value.booldata = true;
     table_attrs.push_back(attr);
 
-    attr.id = SAI_ACL_TABLE_ATTR_FIELD_IP_PROTOCOL;
-    attr.value.booldata = true;
-    table_attrs.push_back(attr);
+    // NOTE: Temporary workaround to support matching protocol numbers on MLNX platform.
+    // In a later SAI version we will transition to using NEXT_HEADER for IPv6 on all platforms.
+    string platform = getenv("platform") ? getenv("platform") : "";
```
qiluo-msft, line 1370: >string platform = getenv("platform") ? getenv("platform") : ""; [](start = 4, length = 63)

The same

orchagent/aclorch.cpp
```diff
@@ -1354,9 +1365,22 @@ bool AclTable::create()
...
+    if ((type == ACL_TABLE_MIRRORV6 || type == ACL_TABLE_L3V6)
+            && platform == MLNX_PLATFORM_SUBSTRING)
```
qiluo-msft, line 1372: >MLNX_PLATFORM_SUBSTRING [](start = 27, length = 23)

SUBSTRING means to test if it is substring of a platform string.

orchagent/aclorch.cpp
daall: Yeah that's where I got it from! I will fix the additions for this PR and then go back and change it for the existing code in a follow-up (because I think it's done in a few more places as well ).

qiluo-msft:
@itaibaz Could you help review?

orchagent/aclorch.cpp
daall: I see it used in this `==` manner throughout aclorch and orchdaemon. Maybe the macro names should be updated?

qiluo-msft: I see it already mixed used. Let's keep as-is, and add an issue to completely fix for all platform string comparison.

---
In reply to: [450408833](https://github.com/Azure/sonic-swss/pull/1343#discussion_r450408833) [](ancestors = 450408833)

daall: Opened #1344. Thanks, Qi!

orchagent/aclorch.cpp
daall: Fixed both!

daall:
retest this please

daall:
retest this please

lguohan:
retest this please

PR #1368, opened 2020-07-28
[portsorch] add buffer drop FC group

mykolaf:

**Do not merge yet, as this also needs a SAI impl update**

**What I did**
Introduce separate FC group for port level buffer drop counters as discussed in Azure#1308
**Why I did it**

**How I verified it**

**Details if related**


daall:
retest vs please

daall:
@mykolaf I also missed until I took a closer look at the 201911 PR: I think "PORT_BUFFER_DROP" in FlexCounterOrch should refer to PORT_BUFFER_DROP_STAT_FLEX_COUNTER_GROUP instead of PORT_STAT_COUNTER_FLEX_COUNTER_GROUP. It doesn't look like it effects the behavior, but we should probably fix it to be consistent.

Issue #1312, opened 2020-06-04
[NAT] nat_zone config.db option makes orchagent core dump on config load with NAT configuration.   


monipko:
**Problem description:**
nat_zone config.db option makes orchagent core dump on config load with NAT configuration.   

Steps to reproduce:

1. Add NAT configuration:
`sudo config nat feature enable` 
`sudo config nat add interface Loopback0 -nat_zone 1`
`sudo config nat add interface PortChannel0001  -nat_zone 1`
`sudo config nat add static basic 10.1.0.32 192.168.0.101`
2. Save config:
`sudo config save -y`
3. Reload config:
`sudo config reload -y`

**Expected result:** configuration will be loaded and services restarted successfully .

**Actual result:** orchagent makes core dump:
`May 13 17:37:26.647826 dut ERR swss#orchagent: :- meta_sai_validate_route_entry: virtual router is set to null object id                                            `
`May 13 17:37:26.647909 dut ERR swss#orchagent: :- addIp2MeRoute: Failed to create IP2me route ip:10.0.0.56, rv:-5                                                              `
`May 13 17:37:26.648023 dut INFO swss#supervisord: orchagent terminate called after throwing an instance of 'std::runtime_error'`
`May 13 17:37:26.648131 dut INFO swss#supervisord: orchagent   what():  Failed to create IP2me route.       `
`May 13 17:37:31.883474 dut INFO swss#supervisord 2020-05-13 17:37:26,794 INFO exited: orchagent (terminated by SIGABRT (core dumped); not expected)`

**Environment:**
SONiC Software Version: SONiC.HEAD.18-dirty-20200413.124039
Distribution: Debian 9.12
Kernel: 4.9.0-11-2-amd64
Build commit: d09fba37
Build date: Mon Apr 13 12:48:37 UTC 2020

Platform: x86_64-arista_7170_64c
HwSKU: Arista-7170-64C

AkhileshSamineni:
@monipko  You can close it issue, as it got addressed with PR https://github.com/Azure/sonic-swss/pull/1372

monipko:
@AkhileshSamineni looks like for 201911 these changes have not been commit and issue is still reproducing for SONiC 201911 #114

arlakshm:
Issue is fixed.

PR #1371, opened 2020-07-31
[intfsorch] Retrieve Port object before setting NAT zone on router interfaces.

kperumalbfn:
Method(setRouterIntfsNatZoneId) is called using local Port object that causes orchagent to crash on config reload with non-zero NAT zone configuration.

Orchagent:
Retrieve Port object using getPort method before setting NAT zone on Router interfaces.

Addressing the issue : https://github.com/Azure/sonic-swss/issues/1312
Signed-off-by: Kumaresh Perumal kumaresh.perumal@intel.com

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -681,21 +681,24 @@ void IntfsOrch::doTask(Consumer &consumer)
                     continue;
                 }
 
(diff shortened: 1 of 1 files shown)


kperumalbfn:
@rlhui LGTM failures are known in 201911 branch. Please help review and merge the changes.

PR #1372, opened 2020-07-31
[intfsorch] Retrieve Port object before setting NAT zone on router in…

kperumalbfn:
Method(setRouterIntfsNatZoneId) is called using local Port object that causes orchagent to crash on config reload with non-zero NAT zone configuration.

Orchagent:
Retrieve Port object using getPort method before setting NAT zone on Router interfaces.

Addressing the issue : #1312
Signed-off-by: Kumaresh Perumal kumaresh.perumal@intel.com


kperumalbfn:
retest this please

arlakshm:
restest this please

orchagent/intfsorch.cpp
```diff
@@ -712,21 +712,24 @@ void IntfsOrch::doTask(Consumer &consumer)
                     continue;
                 }
 
-                /* Set nat zone id */
-                if ((!nat_zone.empty()) and (port.m_nat_zone_id != nat_zone_id))
+                if (gPortsOrch->getPort(alias, port))
```
arlakshm, line 715: this `gPortsOrch->getPort(alias, port))`check is already done above( line 665). Why do we need to do the check here again ?

kperumalbfn: @arlakshm RIF objects are created in setIntf() before calling set_nat_zone API and the RIF object handle is set in port.m_rif_id. To get the Port object with rif handle, getPort() is invoked before setting the NAT zone.

kperumalbfn:
Retest vs please

kperumalbfn:
retest this please

kperumalbfn:
retest this please

kperumalbfn:
retest this please

prsunny:
looks like this is required for 201911 as well, correct?

arlakshm:
> looks like this is required for 201911 as well, correct?

@prsunny, yes this is required for 201911. I have added the label.

kperumalbfn:
retest this please

kperumalbfn:
retest this please

kperumalbfn:
@arlakshm Could you merge this.

abdosi:
@arlakshm Will wait for PR#1411 for 201911 first and then cherry-pick this.

arlakshm:
> @arlakshm Will wait for PR#1411 for 201911 first and then cherry-pick this.

@abdosi, I think #1411 this PR are independent. We can cherry-pick this.

PR #1373, opened 2020-07-31
Fix the issue when persistent DVS is used to run pytest which has number of front-panel ports < 32

abdosi:
**Why/What I did**
 Fix the issue where persistent DVS is used to run pytest
    and ports are created dynamically (ref pr#
    https://github.com/Azure/sonic-buildimage/pull/4499). There were two
    issues:

    a) since number of dynamic front port can be < 32 test case fails
       as it expect always 32. Make sure to update persistent DVS to always have
       32 ports/server link as part of test run if forcedvs option provided else abort test case 
       and save the current config db

    b) after test is done persistent DVS need to be moved to original state.
       Make sure to remove extra port/server link and restore back config db

**How I verified it**

a) Verified Persistent DVS < 32 ports without forcedvs option then abort else DVS restarted with 32 ports.
b) Manually Verified VS Docker has correct interface config/port.ini/ip link after test is completed and restored original config db
c) Verified on host ip netns list has expected server link only after is completed


daall:
>since number of dynamic front port can be < 32 test case fails as it expect always 32

Which tests are always expecting 32 ports? I am OK with this change as a short-term fix, but I think the real issue here is that the tests are hard-coded for 32 ports instead of checking in the Config DB.

tests/conftest.py
```diff
@@ -271,7 +278,13 @@ def __init__(self, name=None, imgname=None, keeptb=False, fakeplatform=None):
     def destroy(self):
         if self.appldb:
             del self.appldb
-        if self.cleanup:
+        # In case persisten dvs was used removed all the extra server link
```
daall, line 281:
```suggestion
        # In case persistent dvs was used removed all the extra server link
```
small typo

tests/conftest.py
```diff
@@ -271,7 +278,13 @@ def __init__(self, name=None, imgname=None, keeptb=False, fakeplatform=None):
...
+        # that were created
+        if self.persistent:
+            for s in self.servers:
+                s.destroy()
```
daall, line 285: I'm confused by this part. 

It makes sense if we have <32 servers that we need to create more, but I don't see that done anywhere?

Also, we destroy all the servers, but do we need to recreate the originals before exiting?

abdosi: > I'm confused by this part.
> 
> It makes sense if we have <32 servers that we need to create more, but I don't see that done anywhere?
@daall 
Caller is this code. Ref: https://github.com/Azure/sonic-swss/blob/541bffbaa9f35236e6925087a7c16b6c68d6f324/tests/conftest.py#L227

Extra port are created in else condition.
Ref: https://github.com/Azure/sonic-swss/blob/541bffbaa9f35236e6925087a7c16b6c68d6f324/tests/conftest.py#L115

> Also, we destroy all the servers, but do we need to recreate the originals before exiting?
@daall if we server port created as part of persistent dvs we don't remove them. cleanup flag is False for those
Ref: https://github.com/Azure/sonic-swss/blob/541bffbaa9f35236e6925087a7c16b6c68d6f324/tests/conftest.py#L113

tests/conftest.py
abdosi: @daall Done

abdosi:
> > since number of dynamic front port can be < 32 test case fails as it expect always 32
> 
> Which tests are always expecting 32 ports? I am OK with this change as a short-term fix, but I think the real issue here is that the tests are hard-coded for 32 ports instead of checking in the Config DB. We should open an issue and track this.

@daall . Many test case are using hard-coded server indexing

For eg: test_fdb_update.py we use  dvs.servers[17].runcmd
test_route.py  we use dvs.servers[3].runcmd

tests/conftest.py
daall: cool, thanks for the thorough explanation! :)

daall:
> > > since number of dynamic front port can be < 32 test case fails as it expect always 32
> > 
> > 
> > Which tests are always expecting 32 ports? I am OK with this change as a short-term fix, but I think the real issue here is that the tests are hard-coded for 32 ports instead of checking in the Config DB. We should open an issue and track this.
> 
> @daall . Many test case are using hard-coded server indexing
> 
> For eg: test_fdb_update.py we use dvs.servers[17].runcmd
> test_route.py we use dvs.servers[3].runcmd

ack, I will open an issue to track this. Thanks for pointing this out!

abdosi:
retest this please

abdosi:
retest this please

abdosi:
retest this please

abdosi:
> lgtm, please wait for @lguohan's review

@lguohan Can you please review this

abdosi:
retest this please

abdosi:
retest this please

abdosi:
@lguohan can you review this.

abdosi:
retest this please

abdosi:
retest this please

abdosi:
> > lgtm, please wait for @lguohan's review
> 
> @lguohan Can you please review this

@lguohan is it ok to merge ?

lguohan:
maybe we can just check if persistent dvs has 32 ports, if not, then we just abort and say persistent dvs has to be 32 ports instead of reconfiguration here?

abdosi:
> maybe we can just check if persistent dvs has 32 ports, if not, then we just abort and say persistent dvs has to be 32 ports instead of reconfiguration here?

@lguohan i think it is better to keep using existing persistent dvs and we can run pytest on top of that
instead to abort it especially for local development/debugging which can have local swss debian package.

lguohan:
@abdosi , can you provide that as an option?

tests/conftest.py
```diff
@@ -41,6 +41,10 @@ def ensure_system(cmd):
 def pytest_addoption(parser):
     parser.addoption("--dvsname", action="store", default=None,
                       help="dvs name")
+    
+    parser.addoption("--forcedvs", action="store_true", default=False,
+                      help="force persistent dvs when ports < 32 ")
+
```
daall, line 47:
```suggestion
    parser.addoption("--forcedvs", action="store_true", default=False,
                      help="force persistent dvs when ports < 32")
```
I'd remove the extra space before and after this option + in the help string to keep it consistent.

daall:
retest vs please

PR #1376, opened 2020-08-04
swss: gearsyncd should return 0 on exit

sydlogan:
* changes made to supervisord conf require gearsyncd to return 0 otherwise
  restarts are performed, fail, and swss docker exits
* originally, gearsyncd did not exit, and if it did, 1 exit code was returned
  to indicate failure. We converted this to do its job and exit, but did not
  correct the exit code.
* prior state was not a problem until changes that are yet to merge for
  sonic-buildimage supervisord conf made returning a 1 problematic.

**What I did**

change exit code for gearsyncd to 0 from 1

**Why I did it**

for compatibility with supervisord.conf. Not to mention, processes should return 0 on success.

**How I verified it**

Local build.

HLD is located at https://github.com/Azure/SONiC/blob/b817a12fd89520d3fd26bbc5897487928e7f6de7/doc/gearbox/gearbox_mgr_design.md

  Signed-off-by: syd.logan@broadcom.com

### DIFF
--- a/gearsyncd/gearsyncd.cpp
+++ b/gearsyncd/gearsyncd.cpp
@@ -105,7 +105,7 @@ int main(int argc, char **argv)
         cerr << "Exception \"" << e.what() << "\" had been thrown in gearsyncd daemon" << endl;
         return EXIT_FAILURE;
     }
-    return 1;
+    return 0;
 }
 
 bool handleGearboxConfigFromConfigDB(ProducerStateTable &p, DBConnector &cfgDb, bool warm)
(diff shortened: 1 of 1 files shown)


sydlogan:
@jleveque  this is going to be required for supervisord conf changes in pull 4851 to work.

gearsyncd/gearsyncd.cpp
```diff
@@ -105,7 +105,7 @@ int main(int argc, char **argv)
         cerr << "Exception \"" << e.what() << "\" had been thrown in gearsyncd daemon" << endl;
         return EXIT_FAILURE;
     }
-    return 1;
+    return 0;
```
jleveque, line 108: You return `EXIT_FRAILURE` above. Suggest returning `EXIT_SUCCESS` (which is defined as 0) here for consistency.

sydlogan:
Retest vs please

sydlogan:
Retest vs please

PR #1378, opened 2020-08-05
[dvs_acl] Refactor and document dvs_acl library

### LINKED ISSUE #1154: [testing] vs tests in TestAcl in test_acl.py have dependency on each other
Currently all the test cases in the TestAcl class in test_acl.py rely on test_AclTableCreation or test_V6AclTableCreation for the tables that they need. This means that individual test cases in TestAcl cannot be run on their own.

ACL table creation/deletion should be added to the TestAcl class as a fixture so that the creation/deletion is still automatic, but individual test cases can be run during feature development.

daall:
- Add docstrings to dvs_acl
- Revert dvs_acl to a vanilla fixture
- Combine redundant dvs_acl methods
- Refactor test_acl, test_port_dpb_acl, and test_mirror_port_span to use new interface

**What I did/why I did it**
1. I changed `dvs_acl` back into an explicit fixture that is passed in to the test methods. The reason for this is because I figured out that tools like `pylint` and `flake8` do not handle the `request.cls...` update very well, which is going to be problematic very soon. In addition, this change makes it a little easier for new contributers/new `dvslib` adopters to see what is going on by making the inclusion of the `dvs_acl` module a little more explicit.

2. I merged a few methods together in `dvs_acl` that were checking the same things (primarily the ACL group and port binding checks). This is for maintainability and to keep the public interface cleaner.

3. I added a bunch of type hints and docstrings to `dvs_acl` to make the usage more clear.

4. I created some fixtures in `test_acl` so that each individual test case can be run standalone, if the user asks for it. This should also make the PR tests more stable as it will be easier for `flaky` to re-run test cases in a controlled manner. Fixes #1154

5. I deleted some dead ACL code from `conftest` as it wasn't being used by `test_acl` or `test_port_dpb_acl` anymore with the `dvslib` additions.

**How I verified it**
Ran the test suite locally and it passes w/ the refactor.

**Details if related**
Once this refactor is merged we can begin converting the rest of `test_acl_*` and `test_mirror_*` to use `dvslib`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,7 +16,7 @@ if sys.version_info < (3, 0):
 from datetime import datetime
 from swsscommon import swsscommon
 from dvslib import dvs_database as dvs_db
-from dvslib import dvs_acl
+from dvslib.dvs_acl import DVSAcl
 from dvslib import dvs_vlan
 from dvslib import dvs_lag
 from dvslib import dvs_mirror
@@ -265,8 +265,6 @@ class DockerVirtualSwitch(object):
         self.config_db = None
         self.flex_db = None
         self.state_db = None
-
-        self.dvs_acl = None
 
     def destroy(self):
         if self.appldb:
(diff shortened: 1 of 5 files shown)


daall:
retest vs please

daall:
retest vs please

daall:
test failures are not related to this change

PR #1382, opened 2020-08-06
[dvs] Convert sflow and speed tests to use dvslib

daall:

**What I did**
I converted the sflow and speed tests to use `dvslib` to poll redis rather than relying on static `time.sleep` calls.

**Why I did it**
Both tests have been failing a lot recently, so I updated them to help stabilize the PR tests.

**How I verified it**
Ran both locally x20, no failures.

**Details if related**


daall:
retest this please

PR #1383, opened 2020-08-06
Added Max Nexthopgroup/ECMP Count supported by device into State DB.

abdosi:
**What I did**
Added Max Nexthopgroup/ECMP Count supported by device into State DB
as part of Switch Capability Table.

Since we are using SWITCH_CAPABILITY Table moved the setting of table
field from aclorch to switchorch via public function
so that in future this can used by any agent.

**Why I did it**
Motivation for adding this to StateDB was to make sure
we can query this value (n) from external test case and make sure
we add (n) number of ECMP Group successfully and (n+1) should fail
without termination of orchagent.

**How I verified it**
admin@str-s6000-on-2:~$ redis-cli -n 6
127.0.0.1:6379[6]> hgetall "SWITCH_CAPABILITY|switch"
 1) "MIRROR"
 2) "true"
 3) "MIRRORV6"
 4) "true"
 5) "ACL_ACTIONS|INGRESS"
 6) "PACKET_ACTION,MIRROR_INGRESS_ACTION"
 7) "ACL_ACTIONS|EGRESS"
 8) "PACKET_ACTION"
 9) "ACL_ACTION|PACKET_ACTION"
10) "DROP,FORWARD"
11) "MAX_NEXTHOP_GROUP_COUNT"
12) "1024"

### DIFF
--- a/orchagent/switchorch.h
+++ b/orchagent/switchorch.h
@@ -12,20 +12,21 @@ struct WarmRestartCheck
 class SwitchOrch : public Orch
 {
 public:
-    SwitchOrch(swss::DBConnector *db, std::string tableName);
-
+    SwitchOrch(swss::DBConnector *db, std::string tableName, TableConnector switchTable);
     bool checkRestartReady() { return m_warmRestartCheck.checkRestartReadyState; }
     bool checkRestartNoFreeze() { return m_warmRestartCheck.noFreeze; }
     bool skipPendingTaskCheck() { return m_warmRestartCheck.skipPendingTaskCheck; }
     void checkRestartReadyDone() { m_warmRestartCheck.checkRestartReadyState = false; }
     void restartCheckReply(const std::string &op, const std::string &data, std::vector<swss::FieldValueTuple> &values);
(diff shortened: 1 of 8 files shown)


orchagent/switchorch.h
```diff
@@ -12,20 +12,21 @@ struct WarmRestartCheck
 class SwitchOrch : public Orch
 {
 public:
-    SwitchOrch(swss::DBConnector *db, std::string tableName);
-
+    SwitchOrch(swss::DBConnector *db, std::string tableName, TableConnector switchTable);
```
prsunny, line 15: Looks like the change is missed in constructor updating `m_switchTable`

abdosi: Thanks @prsunny .  Missed the update. Fixed now

abdosi:
> lgtm. One suggestion is to keep the commits while pushing instead of force-push. While merging, we are doing "squash commits".

sure @prsunny , this were commit with minor fixes so i did squash manually.

abdosi:
@prsunny can you approve this again.

PR #1385, opened 2020-08-06
[dvs] Validate that SWSS is ready to receive input before starting tests

daall:
- Enhance AsicDbValidator to be stricter
- Add a new SWSS ready-check to validate that orchagent is done processing ports

**What I did**
I added some additional ready-checks during the DVS initialization to ensure that the virtual switch is fully initialized before starting the tests.

**Why I did it**
We've noticed recently that a lot of tests are failing from the very first test case, even ones that we have gone through and stabilized in the past. We figured out by checking the logs that, even though orchagent was running, it was still busy processing the initial port config when the tests started running. Since most things in orchagent depend on portsorch initializing everything before they can proceed, nothing was ever written to ASIC DB, causing the tests to fail.

**How I verified it**
Ran tests locally.

**Details if related**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,7 +16,7 @@ if sys.version_info < (3, 0):
 from datetime import datetime
 from swsscommon import swsscommon
 from dvslib.dvs_database import DVSDatabase
-from dvslib.dvs_common import PollingConfig
+from dvslib.dvs_common import PollingConfig, wait_for_result
 from dvslib.dvs_acl import DVSAcl
 from dvslib import dvs_vlan
 from dvslib import dvs_lag
@@ -332,7 +332,7 @@ class DockerVirtualSwitch(object):
 
             time.sleep(1)
 
-    def check_swss_ready(self, timeout=60):
+    def check_swss_ready(self, timeout=300):
         """Verify that SWSS is ready to receive inputs.
 
         Almost every part of orchagent depends on ports being created and initialized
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -38,54 +39,37 @@ def pytest_addoption(parser):
...
-        # get default acl table and acl rules
-        atbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_ACL_TABLE")
-        keys = atbl.getKeys()
+        # Get default ACL table and ACL rules
+        self.default_acl_tables = self.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_ACL_TABLE")
+        assert len(self.default_acl_tables) >= 0
 
-        assert len(keys) >= 0
-        self.default_acl_tables = keys
+        self.default_acl_entries = self.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_ACL_ENTRY", 0)
```
jleveque, line 71: Why wait for 0 keys?

daall: The old behavior was to assert that there are 0 keys. The `wait_for` construct is basically that, but it will try polling the database a few more times before declaring that it is in an unexpected state. Is it necessary here? Probably /not/, but it doesn't hurt to include it just in case. :P

tests/conftest.py
```diff
@@ -333,6 +325,29 @@ def check_ready(self, timeout=30):
...
+    # number of ports. GitHub Issue: Azure/sonic-swss#1384.
+    def check_swss_ready(self, timeout=60):
+        """Verify that SWSS is ready to receive inputs.
+
+        Almost every part of orchagent depends on ports being created and initialized
+        before they can proceed with their processing. If we start the tests after orchagent
+        has started running but before it has had time to initialize all the ports, then the
+        first several tests will fail.
+        """
+        num_ports = 32
```
abdosi, line 339: @daall We can define macro for 32 (few places we are using this)

daall: True, added!

abdosi: I think this will break persistent DVS dynamic port < 32 scenario

abdosi: ok read your FIXME comment now

tests/conftest.py
```diff
@@ -38,54 +46,37 @@ def pytest_addoption(parser):
...
-
-        assert len(keys) >= 0
-        self.default_acl_tables = keys
+        # Get default ACL table and ACL rules
+        self.default_acl_tables = self.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_ACL_TABLE")
+        assert len(self.default_acl_tables) >= 0
 
-        atbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_ACL_ENTRY")
-        keys = atbl.getKeys()
+        self.default_acl_entries = self.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_ACL_ENTRY", 0)
```
abdosi, line 78: Missing assert

daall: `wait_for_n_keys` implicitly asserts that the count is 0. :)

lguohan:
nice to see green here!

Issue #1292, opened 2020-05-11
NAT mgr clears the iptables NAT rules added by other Apps during bootup


venkatmahalingam:
Observation:
   The iptables SNAT rules added from hostcfgd during system bootup, cleaned up by some module.

Root cause:
   NAT mgr clears the iptables NAT rules added by hostcfgd during bootup.

Potential solution:
   Clear the NAT rules added only by natmgr not all rules in kernel.

NAT feature pull request:
https://github.com/Azure/sonic-swss/pull/1059

@AkhileshSamineni As discussed, please use this issue for tracking.

AkhileshSamineni:
Raised a PR https://github.com/Azure/sonic-swss/pull/1386 to address this issue.

arlakshm:
#1386 is merged. Closing this issue

PR #1386, opened 2020-08-07
Clear the iptables NAT rules only which were added by NAT Mgr

AkhileshSamineni:
This PR is to address the issue https://github.com/Azure/sonic-swss/issues/1292

Natmgrd does flushing the iptables nat rules at 2 incidents
1. At time of swss container start (natmgr init) - https://github.com/Azure/sonic-swss/blob/master/cfgmgr/natmgr.cpp#L66 
2. At time of sigterm handler (when user issues “config reload”) - https://github.com/Azure/sonic-swss/blob/master/cfgmgr/natmgrd.cpp#L73

Natmgr is the only component adding the iptables so we went for flush, now I see other components (like hostcfgd) are adding the nat rules.  So added changes to clear the iptables NAT rules only which were added by NAT Mgr.


AkhileshSamineni:
@arlakshm @rlhui Could you please review these changes.

arlakshm (approved):

PR #1387, opened 2020-08-07
[rates] fix issue with rates init

mykolaf:

Fix 2 issues with the lua scripts:
 - introduce separate INIT flag for PORT and RIF. Previously could cause problems on init of one of the groups.
 - skip if there is no initial config (alpha in DB). When reboot/config reload, it happens that alpha is not set until the `enable_counters.py` is executed

**What I did**

**Why I did it**

**How I verified it**
reboot, config reload, 
check rates initialized correctly, w/o errors in log
**Details if related**


lguohan:
retest this please

PR #1389, opened 2020-08-07
[dvs] Upload logs even if failure occurs during startup

daall:

**What I did**
I updated `dvs` to use the module log path even if the failure occurs during virtual switch setup.

**Why I did it**
Previously the `get_logs()` call in `check_ready_status` passed None as the `modpath`, which means any failure during startup would reset the logs. This makes it hard to debug startup issues for particular test cases, and it deletes a lot of helpful data from other non-startup related test failures.

**How I verified it**
Inserted `assert 2 == 3` into the DockerVirtualSwitch constructor, verified that logs were output to `test_speed` (no dvs specified) and `my_virtual_switch` (persistent container specified) respectively.

**Details if related**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -158,7 +158,7 @@ class DockerVirtualSwitch(object):
     FLEX_COUNTER_DB_ID = 5
     STATE_DB_ID = 6
 
-    def __init__(self, name=None, imgname=None, keeptb=False, fakeplatform=None):
+    def __init__(self, name=None, imgname=None, keeptb=False, fakeplatform=None, log_path=None):
         self.basicd = ['redis-server',
                        'rsyslogd']
         self.swssd = ['orchagent',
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -970,12 +974,14 @@ def dvs(request) -> DockerVirtualSwitch:
...
+
+    dvs = DockerVirtualSwitch(name, imgname, keeptb, fakeplatform, log_path)
+
     yield dvs
-    if name == None:
-        dvs.get_logs(request.module.__name__)
-    else:
-        dvs.get_logs()
+
+    dvs.get_logs()
```
lguohan, line 984: I did not realize we only capture the syslog, we should probably dump the db and capture that as well (not in this pr though)

daall: We do also get sairedis.rec and supervisor logs as well, but I think dumping the DB should also help

lguohan: does show tech work in the vs docker? if it works, then we can use that, that captures every thing.

daall: It looks like it fails because `sudo` is not found. I'll dig a little deeper and see if we can make it work

daall:
retest vs please

PR #1392, opened 2020-08-10
[dvs] Add performance test for adding and deleting routes

shi-su:
**What I did**
Refactor route tests to use dvslib.
Add a performance test for adding and deleting routes.

**Why I did it**
The performance test allows quantifying the efficiency of interactions between orchagent and syncd. The performance test also makes it possible to check if changes can be pushed into hardware within the required time limits.

**How I verified it**
Run the tests locally.

**Details if related**

**Sample output of performance test**
Time to add 10000 routes is 6.55 seconds.
Time to remove 10000 routes is 5.86 seconds.

### DIFF
--- a/tests/test_route.py
+++ b/tests/test_route.py
@@ -9,70 +9,55 @@ from swsscommon import swsscommon
 
 class TestRoute(object):
     def setup_db(self, dvs):
-        self.pdb = swsscommon.DBConnector(0, dvs.redis_sock, 0)
-        self.adb = swsscommon.DBConnector(1, dvs.redis_sock, 0)
-        self.cdb = swsscommon.DBConnector(4, dvs.redis_sock, 0)
+        self.pdb = dvs.get_app_db()
+        self.adb = dvs.get_asic_db()
+        self.cdb = dvs.get_config_db()
 
     def set_admin_status(self, interface, status):
-        tbl = swsscommon.Table(self.cdb, "PORT")
-        fvs = swsscommon.FieldValuePairs([("admin_status", status)])
-        tbl.set(interface, fvs)
+        self.cdb.update_entry("PORT", interface, {"admin_status": status})
         time.sleep(1)
 
     def create_vrf(self, vrf_name):
-        tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_VIRTUAL_ROUTER")
-        initial_entries = set(tbl.getKeys())
+        initial_entries = set(self.adb.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_VIRTUAL_ROUTER"))
 
-        tbl = swsscommon.Table(self.cdb, "VRF")
-        fvs = swsscommon.FieldValuePairs([('empty', 'empty')])
-        tbl.set(vrf_name, fvs)
+        self.cdb.create_entry("VRF", vrf_name, {'empty': 'empty'})
         time.sleep(1)
 
-        tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_VIRTUAL_ROUTER")
-        current_entries = set(tbl.getKeys())
+        current_entries = set(self.adb.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_VIRTUAL_ROUTER"))
         assert len(current_entries - initial_entries) == 1
         return list(current_entries - initial_entries)[0]
 
     def remove_vrf(self, vrf_name):
-        tbl = swsscommon.Table(self.cdb, "VRF")
-        tbl._del(vrf_name)
+        self.cdb.delete_entry("VRF", vrf_name)
         time.sleep(1)
 
     def create_l3_intf(self, interface, vrf_name):
-        tbl = swsscommon.Table(self.cdb, "INTERFACE")
         if len(vrf_name) == 0:
-            fvs = swsscommon.FieldValuePairs([("NULL", "NULL")])
+            self.cdb.create_entry("INTERFACE", interface, {"NULL": "NULL"})
         else:
-            fvs = swsscommon.FieldValuePairs([("vrf_name", vrf_name)])
-        tbl.set(interface, fvs)
+            self.cdb.create_entry("INTERFACE", interface, {"vrf_name": vrf_name})
         time.sleep(1)
 
     def remove_l3_intf(self, interface):
-        tbl = swsscommon.Table(self.cdb, "INTERFACE")
-        tbl._del(interface)
+        self.cdb.delete_entry("INTERFACE", interface)
         time.sleep(1)
 
     def add_ip_address(self, interface, ip):
-        tbl = swsscommon.Table(self.cdb, "INTERFACE")
-        fvs = swsscommon.FieldValuePairs([("NULL", "NULL")])
-        tbl.set(interface + "|" + ip, fvs)
+        self.cdb.create_entry("INTERFACE", interface + "|" + ip, {"NULL": "NULL"})
         time.sleep(1)
 
     def remove_ip_address(self, interface, ip):
-        tbl = swsscommon.Table(self.cdb, "INTERFACE")
-        tbl._del(interface + "|" + ip)
+        self.cdb.delete_entry("INTERFACE", interface + "|" + ip)
         time.sleep(1)
 
     def create_route_entry(self, key, pairs):
-        tbl = swsscommon.ProducerStateTable(self.pdb, "ROUTE_TABLE")
-        fvs = swsscommon.FieldValuePairs(pairs)
+        tbl = swsscommon.ProducerStateTable(self.pdb.db_connection, "ROUTE_TABLE")
+        fvs = swsscommon.FieldValuePairs(list(pairs.items()))
         tbl.set(key, fvs)
-        time.sleep(1)
 
     def remove_route_entry(self, key):
-        tbl = swsscommon.ProducerStateTable(self.pdb, "ROUTE_TABLE")
+        tbl = swsscommon.ProducerStateTable(self.pdb.db_connection, "ROUTE_TABLE")
         tbl._del(key)
-        time.sleep(1)
 
     def clear_srv_config(self, dvs):
         dvs.servers[0].runcmd("ip address flush dev eth0")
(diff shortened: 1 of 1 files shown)


tests/test_route.py
```diff
@@ -524,21 +490,19 @@ def test_RouteAndNexthopInDifferentVrf(self, dvs, testlog):
...
-        tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_NEXT_HOP")
-        nexthop_entries = tbl.getKeys()
+        nexthop_entries = self.adb.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_NEXT_HOP")
         for key in nexthop_entries:
-            (status, fvs) = tbl.get(key)
+            fvs = self.adb.get_entry("ASIC_STATE:SAI_OBJECT_TYPE_NEXT_HOP", key)
+            status = bool(fvs)
             assert status == True
-            for fv in fvs:
+            for fv in list(fvs.items()):
```
qiluo-msft, line 505: >fvs [](start = 27, length = 3)

Could you use fvs as a dict, instead of converting back to list?

tests/test_route.py
```diff
@@ -623,3 +585,94 @@ def test_RouteAndNexthopInDifferentVrf(self, dvs, testlog):
...
+    """ Performance tests for route """
+    def test_PerfAddRemoveRoute(self, dvs, testlog):
+        self.setup_db(dvs)
+        self.clear_srv_config(dvs)
+        numRoutes = 10000   # number of routes to add/remove
+
+        # generate addresses of routes
+        addrs = []
+        for i in range(numRoutes):
+            addrs.append("%d.%d.%d.%d/%d" % (100 + int(i / 256 ** 2), int(i / 256), i % 256, 0, 24))
```
qiluo-msft, line 599: >addrs [](start = 12, length = 5)

they are not ip addresses, but ip prefixes.

tests/test_route.py
```diff
@@ -623,3 +585,94 @@ def test_RouteAndNexthopInDifferentVrf(self, dvs, testlog):
...
+
+        # confirm all routes are added
+        asicAddrs = set()
+        for key in self.adb.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY"):
+            route = json.loads(key)
+            asicAddrs.add(route["dest"])
+        for addr in addrs:
+            assert addr in asicAddrs
+
+        #remove route entries
```
qiluo-msft, line 645: >#remove [](start = 8, length = 7)

Add a blank in between.

tests/test_route.py
```diff
@@ -623,3 +585,94 @@ def test_RouteAndNexthopInDifferentVrf(self, dvs, testlog):
...
+
+        self.set_admin_status("Ethernet0", "down")
+        self.set_admin_status("Ethernet4", "down")
+
+        # remove ip address and default route
+        dvs.servers[0].runcmd("ip route del default dev eth0")
+        dvs.servers[0].runcmd("ip address del 10.0.0.1/31 dev eth0")
+
+        dvs.servers[1].runcmd("ip route del default dev eth0")
+        dvs.servers[1].runcmd("ip address del 10.0.0.3/31 dev eth0")
```
qiluo-msft, line 678: >) [](start = 67, length = 1)

Could you provide some sample output in PR description?

tests/test_route.py
shi-su: Fixed this and two later ones.

tests/test_route.py
shi-su: Fixed.

tests/test_route.py
shi-su: Fixed.

tests/test_route.py
shi-su: Added a sample out of the performance test.

tests/test_route.py
```diff
@@ -6,80 +6,66 @@
...
 
     def create_vrf(self, vrf_name):
-        tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_VIRTUAL_ROUTER")
-        initial_entries = set(tbl.getKeys())
+        initial_entries = set(self.adb.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_VIRTUAL_ROUTER"))
 
-        tbl = swsscommon.Table(self.cdb, "VRF")
-        fvs = swsscommon.FieldValuePairs([('empty', 'empty')])
-        tbl.set(vrf_name, fvs)
+        self.cdb.create_entry("VRF", vrf_name, {'empty': 'empty'})
```
daall, line 22:
```suggestion
        self.cdb.create_entry("VRF", vrf_name, {"empty": "empty"})
```

Would recommend sticking to double-quotes throughout for consistency.

tests/test_route.py
```diff
@@ -112,14 +98,12 @@ def test_RouteAddRemoveIpv4Route(self, dvs, testlog):
         time.sleep(1)
 
         # check application database
-        tbl = swsscommon.Table(self.pdb, "ROUTE_TABLE")
-        route_entries = tbl.getKeys()
+        route_entries = self.pdb.get_keys("ROUTE_TABLE")
         assert "2.2.2.0/24" in route_entries
```
daall, line 102: There's a method here that you can use to guarantee that this key exists in the route table before proceeding: https://github.com/Azure/sonic-swss/blob/3052fb5f25fea0ba2f60ad17c294e6da8b9eb70a/tests/dvslib/dvs_database.py#L326

tests/test_route.py
```diff
@@ -130,13 +114,11 @@ def test_RouteAddRemoveIpv4Route(self, dvs, testlog):
         time.sleep(1)
 
         # check application database
-        tbl = swsscommon.Table(self.pdb, "ROUTE_TABLE")
-        route_entries = tbl.getKeys()
+        route_entries = self.pdb.get_keys("ROUTE_TABLE")
```
daall, line 117: Similarly there is this method for deleted keys: https://github.com/Azure/sonic-swss/blob/3052fb5f25fea0ba2f60ad17c294e6da8b9eb70a/tests/dvslib/dvs_database.py#L357

tests/test_route.py
```diff
@@ -524,49 +490,43 @@ def test_RouteAndNexthopInDifferentVrf(self, dvs, testlog):
...
-                    nexthop1_found = True
-                    nexthop1_oid = key
+            if fvs["SAI_NEXT_HOP_ATTR_IP"] == "20.0.1.2":
+                nexthop2_found = True
+                nexthop2_oid = key
+            if fvs["SAI_NEXT_HOP_ATTR_IP"] == "10.0.0.2":
+                nexthop1_found = True
+                nexthop1_oid = key
 
         assert nexthop1_found == True and nexthop2_found == True
```
daall, line 512: This should be wrapped in a `wait_until`, see example here: https://github.com/Azure/sonic-swss/blob/3052fb5f25fea0ba2f60ad17c294e6da8b9eb70a/tests/conftest.py#L351

tests/test_route.py
```diff
@@ -112,14 +98,12 @@ def test_RouteAddRemoveIpv4Route(self, dvs, testlog):
...
         assert "2.2.2.0/24" in route_entries
 
         # check ASIC route database
-        tbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY")
         route_found = False
-        for key in tbl.getKeys():
+        for key in self.adb.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY"):
             route = json.loads(key)
             if route["dest"] == "2.2.2.0/24":
                 route_found = True
```
daall, line 109: I think this is probably OK since you are waiting for the key in ROUTE_TABLE above, but for some extra insurance you may consider doing something like this: https://github.com/Azure/sonic-swss/blob/3052fb5f25fea0ba2f60ad17c294e6da8b9eb70a/tests/test_sub_port_intf.py#L104

tests/test_route.py
```diff
@@ -623,3 +580,94 @@ def test_RouteAndNexthopInDifferentVrf(self, dvs, testlog):
...
+        # get number of routes before adding new routes
+        startNumRoutes = len(self.adb.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY"))
+
+        # add route entries
+        timeStart = time.time()
+        for i in range(numRoutes):
+            self.create_route_entry(prefixes[i], fieldValues[i % 2])
+
+        # wait until all routes are added into ASIC database
+        self.adb.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY", startNumRoutes + numRoutes) # default timeout is 5 seconds
```
daall, line 629: We might have to wait and see how this goes in the PR tests. The PR runners can be a little slow so it is possible that we need to bump up the timeout a little bit for this test.

daall: Just kidding, it has run a few times successfully already it looks like. This default timeout is probably OK. :)

tests/test_route.py
```diff
@@ -623,3 +580,94 @@ def test_RouteAndNexthopInDifferentVrf(self, dvs, testlog):
...
+        dvs.servers[0].runcmd("ip address add 10.0.0.1/31 dev eth0")
+        dvs.servers[0].runcmd("ip route add default via 10.0.0.0")
+
+        dvs.servers[1].runcmd("ip address add 10.0.0.3/31 dev eth0")
+        dvs.servers[1].runcmd("ip route add default via 10.0.0.2")
+
+        fieldValues = [{"nexthop": "10.0.0.1", "ifname": "Ethernet0"}, {"nexthop": "10.0.0.3", "ifname": "Ethernet4"}]
+
+        # get neighbor and arp entry
+        dvs.servers[0].runcmd("ping -c 1 10.0.0.3")
```
prsunny, line 618: what about 10.0.0.1?

tests/test_route.py
shi-su: Fixed.

tests/test_route.py
shi-su: Since this line only checks for one entry, I am using wait_for_entry() instead.

tests/test_route.py
shi-su: Using wait_for_deleted_entry() to wait until the key is deleted.

tests/test_route.py
shi-su: Wrapped these into check_route_entries_nexthop() function.

tests/test_route.py
shi-su: Updated the check for ASIC_DB into another wait_until function.

tests/test_route.py
shi-su: Added another ping for 10.0.0.1.

tests/test_route.py
```diff
@@ -5,81 +5,115 @@
...
+        self.pdb = dvs.get_app_db()
+        self.adb = dvs.get_asic_db()
+        self.cdb = dvs.get_config_db()
 
     def set_admin_status(self, interface, status):
-        tbl = swsscommon.Table(self.cdb, "PORT")
-        fvs = swsscommon.FieldValuePairs([("admin_status", status)])
-        tbl.set(interface, fvs)
+        self.cdb.update_entry("PORT", interface, {"admin_status": status})
         time.sleep(1)
```
daall, line 19: Do you know if these `sleep` calls are still needed with all of the new polling functions you added?

daall:
retest vs please

PR #1394, opened 2020-08-10
[dvs] Add options to limit CPU usage

daall:

**What I did**
I added a flag to limit the CPU usage of a DVS container.

**Why I did it**
I did it to try to set an expectation of how much CPU the DVS container /should/ be using, and to try to make the performance more deterministic b/w the PR runners and different peoples' local setups (without a limit Docker just uses as much as it can from the host).

**How I verified it**
Ran tests locally, still passing. Typical CPU usage was <1 core, but need to experiment/monitor more with the PR runner to see if that is sufficient.

**Details if related**

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -79,6 +79,18 @@ For those developing new features for SWSS or the DVS framework, you might find 
     ```
 
 ## Other useful test parameters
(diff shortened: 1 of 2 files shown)


daall:
retest this please

PR #1395, opened 2020-08-11
Fix log/syslog not being correct when last test fails for given module

abdosi:
What/Why I did:-
As reported by this issue PR#https://github.com/box/flaky/issues/128
and PR#https://github.com/pytest-dev/pytest-rerunfailures/issues/51
currently pytest will call session/module tear-down before flaky
can retry the test when last test case fails. And Session/Module will be setup again to run last test 
retry.
Because of above behaviour when last test fail for given module
all the logs are lost for all test-case upto that point since when
tear down is called dvs container is destroyed and to run retry instance
DVS container is setup up again and logs only belonging to this instance
of run are captured and overwrite all previous logs.

Workaround to have default dummy pass (as suggested in above Issue also)  test case at end always so we
avoid module tear-down and setup again and logs are not lost.

### DIFF
--- a/tests/test_acl.py
+++ b/tests/test_acl.py
@@ -492,3 +492,7 @@ class TestAclRuleValidation():
                 dvs.runcmd("supervisorctl restart syncd")
                 dvs.stop_swss()
                 dvs.start_swss()
+# Add Dummy always-pass test at end as workaroud
+# for issue when Flaky fail on final test it invokes module tear-down before retrying
+def test_nonflaky_dummy():
+    pass
--- a/tests/test_acl.py
+++ b/tests/test_acl.py
@@ -492,6 +492,8 @@ class TestAclRuleValidation():
                 dvs.runcmd("supervisorctl restart syncd")
                 dvs.stop_swss()
                 dvs.start_swss()
+
+
 # Add Dummy always-pass test at end as workaroud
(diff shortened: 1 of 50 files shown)


tests/test_acl.py
```diff
@@ -492,3 +492,7 @@ def test_AclActionValidation(self, dvs, dvs_acl):
                 dvs.runcmd("supervisorctl restart syncd")
                 dvs.stop_swss()
                 dvs.start_swss()
+# Add Dummy always-pass test at end as workaroud
+# for issue when Flaky fail on final test it invokes module tear-down before retrying
+def test_nonflaky_dummy():
```
daall, line 497: Just double-checking, this works when added outside of the classes, right? We won't have this syslog issue if, say, the last test case in `TestAcl` fails?

abdosi: @daall : Outside of class works fine. Have verified it tes_acl.py also. 
Module teardown happens after executing dummy test case

abdosi:
retest this please

abdosi:
> LGTM, only suggestion is to add 2 spaces b/w the test cases and the dummy test case throughout.
> 
> It might also be worth adding a note in the "known issues" part of the README explaining why this case is needed and linking to the `flaky` issue. :)

@daall Updated.

PR #1399, opened 2020-08-12
[fec] added logic that put port down before applying fec onfiguration

vadymhlushko-mlnx:
Signed-off-by: Vadym Hlushko <vadymh@mellanox.com>

**What I did**

Added logic that put port down before applying FEC configuration, like in the case with speed change

**Why I did it**

Because if we don't put port down FEC changes will not apply

**How I verified it**

_On the fanaout side:_
1. Put port down
2. Change FEC
3. Put port up

_On the switch side:_
1. Change FEC
2. Verify FEC state by using SDK script (in syncd docker)


volodymyrsamotiy:
retest this please

PR #1401, opened 2020-08-13
Ignore IPv6 link-local entries as Vnet routes

prsunny:
**What I did**
Ignore IPv6 link-local and multicast entries as Vnet routes

**Why I did it**
Otherwise creating stale VNET route entries and displayed in "show vnet routes" command

**How I verified it**
Run in DUT

**Details if related**


prsunny:
retest vs please

PR #1402, opened 2020-08-17
[dvs] Refactor AsicDbValidator

daall:
- Relax timeout for startup sequence
- Break constructor into logical segments

**What I did**
I updated the ASIC DB Validator to 1) wait 30 seconds for the DB to initialize and 2) break it up into a few logical segments.

**Why I did it**
A few tests failed over the weekend because ASIC DB wasn't finished creating all of the host interfaces, so I made the timeout longer than the normal 5s timeout. Adding this made `__init__` really long, so I broke the code up into private methods.

**How I verified it**
Re-run tests locally with the change, they all still pass.

**Details if related**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -53,34 +53,55 @@ def pytest_addoption(parser):
 
 
 class AsicDbValidator(DVSDatabase):
-    def __init__(self, db_id: str, connector: str):
+    def __init__(self, db_id: int, connector: str):
         DVSDatabase.__init__(self, db_id, connector)
+        self._wait_for_asic_db_to_initialize()
+        self._populate_default_asic_db_values()
+        self._generate_oid_to_interface_mapping()
 
-        # Get default .1Q Vlan ID
-        self.default_vlan_id = self.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_VLAN", 1)[0]
+    def _wait_for_asic_db_to_initialize(self) -> None:
+        """Wait up to 30 seconds for the default fields to appear in ASIC DB."""
+        def _verify_db_contents():
+            # We expect only the default VLAN
+            if len(self.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_VLAN")) != 1:
+                return (False, None)
 
-        # Build port OID to front port name mapping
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -53,34 +53,55 @@ def pytest_addoption(parser):
...
+        """Wait up to 30 seconds for the default fields to appear in ASIC DB."""
+        def _verify_db_contents():
+            # We expect only the default VLAN
+            if len(self.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_VLAN")) != 1:
+                return (False, None)
+
+            if len(self.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_HOSTIF")) != NUM_PORTS:
+                return (False, None)
+
+            if len(self.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_ACL_ENTRY")) != 0:
```
lguohan, line 72: what acl entry are we expecting?

tests/conftest.py
```diff
@@ -53,34 +53,55 @@ def pytest_addoption(parser):
...
+        self._populate_default_asic_db_values()
+        self._generate_oid_to_interface_mapping()
 
-        # Get default .1Q Vlan ID
-        self.default_vlan_id = self.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_VLAN", 1)[0]
+    def _wait_for_asic_db_to_initialize(self) -> None:
+        """Wait up to 30 seconds for the default fields to appear in ASIC DB."""
+        def _verify_db_contents():
+            # We expect only the default VLAN
+            if len(self.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_VLAN")) != 1:
```
lguohan, line 66: this might be tricky if we load the T1 topology, do we still expect vlan?

tests/conftest.py
daall: We aren't expecting any ACL entries, so if we find any keys then the check fails.

tests/conftest.py
daall: This check has been here for a while so I'm inclined to leave it as is for this PR, but I can open another one to make this more flexible w.r.t. the topology.

This check as a whole isn't very robust, it would be good if we could pass in an expected topology for the ASIC DB validator to refer to rather than hard-coding it in `conftest.py`.

PR #1403, opened 2020-08-17
[fec] Get FEC mode when port is already admin down

daall:

**What I did**
I added a line to fetch the FEC mode from the map before setting it.

**Why I did it**
This check was missed in an earlier PR, causing tests to fail in sairedis like so: https://sonic-jenkins.westus2.cloudapp.azure.com/job/vs/job/sonic-sairedis-build/1196/testReport/junit/test_port/TestPort/test_PortFec/

**How I verified it**
Re-run VS tests with this change.

**Details if related**
We're not totally sure why this was caught in the sairedis PR tests and not the swss PR tests. Investigation ongoing.

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2386,6 +2386,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                             else
                             {
                                 /* Port is already down, setting fec mode*/
+                                p.m_fec_mode = fec_mode_map[fec_mode]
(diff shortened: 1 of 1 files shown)


prsunny:
@vadymhlushko-mlnx

vadymhlushko-mlnx:
This PR is created due to my fault in [swss_PR_1399](https://github.com/Azure/sonic-swss/pull/1399), forgot to get fec_mode before applying it when a port in down state

daall:
retest this please

PR #1406, opened 2020-08-18
[dvs] Clean-up conftest.py

daall:
- Refactor check_services to use common polling method
- Convert remaining python 2 code to python 3 (f-strings, type hints, etc.)
- Make the formatting consistent
- Note remaining test dependencies in dvs

**What I did/Why I did it**
I cleaned-up/refactored `conftest.py`:

1. `check_services` implemented its own polling mechanism, so I refactored it to use `wait_for_result` instead.
2. There was some left-over python2 code from using `2to3` for the conversion. Because we are using python3 syntax in a lot of places now, and because python2 is EOL, it no longer makes sense to include the python2 support, so I refactored the rest of it out. This includes swapping `format` for `f-string`, `subprocess` for `command`, and adding type hints.
3. The formatting was a little all over the place so I refactored it to be consistent throughout (except for a chunk of methods in the middle that we are planning on replacing with `dvslib`).
4. I added a `dep` comment to a big chunk of the methods in DockerVirtualSwitch that are still being used by the test cases. The plan is to deprecate these and have the tests rely on `dvslib` for common code. The comments are there to keep track of which external dependencies are still needed. The comments, and their respective methods, should be removed as more functionality is added to `dvslib`.

**How I verified it**
Ran the tests in 3 different configurations:
- No persistent container
- Persistent container w/ 32 ports
- Persistent container w/ 2 ports (w/ and w/o the `--forcedvs` flag)

All work as expected.

**Details if related**
N/A

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,21 +1,18 @@
 import os
-import os.path
 import re
 import time
 import json
 import redis
 import docker
 import pytest
-import tarfile
-import io
 import random
 import string
 import subprocess
 import sys
-if sys.version_info < (3, 0):
-    import commands
 
+from typing import Dict, Tuple
 from datetime import datetime
+
 from swsscommon import swsscommon
 from dvslib.dvs_database import DVSDatabase
 from dvslib.dvs_common import PollingConfig, wait_for_result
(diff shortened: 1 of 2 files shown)


daall:
retest vs please

daall:
retest this please

lguohan (approved):

daall:
retest vs please

daall:
retest vs please

tests/conftest.py
```diff
@@ -114,166 +128,162 @@ def _populate_default_asic_db_values(self) -> None:
...
-                      'vlanmgrd',
-                      'vrfmgrd',
-                      'portmgrd']
-        self.syncd = ['syncd']
-        self.rtd   = ['fpmsyncd', 'zebra', 'staticd']
-        self.teamd = ['teamsyncd', 'teammgrd']
-        self.natd = ['natsyncd', 'natmgrd']
-        self.alld  = self.basicd + self.swssd + self.syncd + self.rtd + self.teamd + self.natd
-        self.client = docker.from_env()
-        self.appldb = None
```
vasant17, line 225: Deleted by accident?

PR #1408, opened 2020-08-19
[dvs] Fix unstable DPB and NAT test cases

### LINKED ISSUE #1222: [dvs] test_port_breakout_multiple test case unstable
`test_port_breakout_multiple` is failing ~50% of the time at the moment, need someone to take a look at it and see if it is a timing issue or if the virtual switch is crashing during one of the steps.

daall:
- Re-enable test_port_dpb
- Add polling to NAT

**What I did**
I re-enabled the DPB test because it is now passing consistently and I added a skip to the NAT test because it is broken.

**Why I did it**
To un-block two tests that have been removed from the test suite for an extended period of time.

**How I verified it**
Ran both locally x10.

**Details if related**
Fixes #1222

### DIFF
--- a/tests/test_nat.py
+++ b/tests/test_nat.py
@@ -1,6 +1,10 @@
 import time
 import pytest
 
+from dvslib.dvs_common import wait_for_result
+from dvslib.dvs_database import DVSDatabase
+
+
 class TestNat(object):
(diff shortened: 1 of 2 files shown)


daall:
retest vs please

PR #1410, opened 2020-08-19
[dvs/chassis] Bring up SONiC interfaces in virtual chassis

daall:

**What I did/why I did it**
While testing https://github.com/Azure/sonic-buildimage/pull/5192 we figured out that the port config that is generated from `port_config.ini` does not include any information about the admin status of the ports on the virtual switch.

However, with this PR, ports are set to `DOWN` by default in the DVS. This causes the connectivity checks in `test_virtual_chassis.py` to fail indefinitely.

This PR ensures that the necessary ports in the chassis are admin UP so that pings can be sent from the LC instances to the neighboring servers.

I kept mixing up which interface was which while working on this PR, so I also refactored some of the variable names to be more descriptive.

**How I verified it**
Ran the chassis tests against the current master `docker-sonic-vs` as well as the one from https://github.com/Azure/sonic-buildimage/pull/5192. Both are successful on all chassis topologies.

**Details if related**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1266,13 +1266,13 @@ class DockerVirtualChassisTopology(object):
 
     def create_vct_ctn(self, ctndir):
         cwd = os.getcwd()
-        cfgdir = cwd + "/virtual_chassis/" + ctndir
-        cfgfile =  cfgdir + "/default_config.json"
-        with open(cfgfile, "r") as cfg:
+        chassis_config_dir = cwd + "/virtual_chassis/" + ctndir
+        chassis_config_file =  chassis_config_dir + "/default_config.json"
+        with open(chassis_config_file, "r") as cfg:
             defcfg = json.load(cfg)["DEVICE_METADATA"]["localhost"]
             ctnname = defcfg["hostname"] + "." + self.ns
             vol = {}
-            vol[cfgdir] = {"bind": "/usr/share/sonic/virtual_chassis", "mode": "ro"}
+            vol[chassis_config_dir] = {"bind": "/usr/share/sonic/virtual_chassis", "mode": "ro"}
 
             # pass self.ns into the vs to be use for vs restarts by swss conftest.
             # connection to chassbr is setup by chassis_connect.py within the vs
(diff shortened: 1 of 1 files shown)


daall:
@saravanansv can you help check too?

tests/conftest.py
```diff
@@ -1300,57 +1300,69 @@ def get_inband(self, ctnname):
...
+            chassis_config_file = os.path.join(chassis_config_dir, "default_config.json")
+            instance_container_name = ""
+            with open(chassis_config_file, "r") as cfg:
+                device_info = json.load(cfg)["DEVICE_METADATA"]["localhost"]
+                instance_container_name = device_info["hostname"] + "." + self.ns
+
+            chassis_veth_intf = int(endpoints[chassis_instance].split("eth")[1])
+            chassis_host_intf = f"Ethernet{(chassis_veth_intf - 1) * 4}"
+
+            chassis_config_file = os.path.join(working_dir, "virtual_chassis", neighbor_instance, "default_config.json")
```
saravanansv, line 1322: the variable needs to be neighbor_config_file

tests/conftest.py
```diff
@@ -1300,57 +1300,69 @@ def get_inband(self, ctnname):
...
+
+            chassis_config_dir = os.path.join(working_dir, "virtual_chassis", chassis_instance)
+            chassis_config_file = os.path.join(chassis_config_dir, "default_config.json")
+            instance_container_name = ""
+            with open(chassis_config_file, "r") as cfg:
+                device_info = json.load(cfg)["DEVICE_METADATA"]["localhost"]
+                instance_container_name = device_info["hostname"] + "." + self.ns
+
+            chassis_veth_intf = int(endpoints[chassis_instance].split("eth")[1])
+            chassis_host_intf = f"Ethernet{(chassis_veth_intf - 1) * 4}"
```
saravanansv, line 1320: this needs to be neighbor_host_intf = f"Ethernet{(neighbor_veth_intf - 1) * 4}"

tests/conftest.py
```diff
@@ -1300,57 +1300,69 @@ def get_inband(self, ctnname):
...
                             continue
-                        if ctnname not in neighs_by_ctn:
-                            neighs_by_ctn[ctnname] = []
-                        neighs_by_ctn[ctnname].append((vethintf - 1, nbraddr))
-        return neighs_by_ctn
+
+                        if instance_container_name not in instance_to_neighbor_map:
+                            instance_to_neighbor_map[instance_container_name] = []
+
+                        instance_to_neighbor_map[instance_container_name].append((chassis_veth_intf - 1,
```
saravanansv, line 1339: instance_to_neighbor_map[instance_container_name].append((neighbor_veth_intf - 1, neighbor_address)

tests/conftest.py
daall: We're trying to shut down the link on the LC side, so I think it should be the `chassis_host_intf`.

tests/conftest.py
daall: Doesn't this config file reside in the LC instance, not the neighbor?

tests/conftest.py
daall: Makes sense, thank you!

tests/conftest.py
daall: Ah, I see now. Thanks!

tests/conftest.py
saravanansv: got it. thanks

tests/conftest.py
saravanansv: each instance has its own config file. so the chassis_instance and neighbor_instance configs are separate

tests/conftest.py
saravanansv: may be a comment can help on why we return some info about the neighbor-instance and some about the chassis-instance. 
Neighbor info is there so that verify command can run ping to the neighbor to check connectivity.

tests/conftest.py
daall: Fixed

tests/conftest.py
daall: Fixed

tests/conftest.py
```diff
@@ -1300,57 +1300,69 @@ def get_inband(self, ctnname):
...
+            neighbor_instance = conn.split('-')[1]
+
+            chassis_config_dir = os.path.join(working_dir, "virtual_chassis", chassis_instance)
+            chassis_config_file = os.path.join(chassis_config_dir, "default_config.json")
+            instance_container_name = ""
+            with open(chassis_config_file, "r") as cfg:
+                device_info = json.load(cfg)["DEVICE_METADATA"]["localhost"]
+                instance_container_name = device_info["hostname"] + "." + self.ns
+
+            chassis_veth_intf = int(endpoints[chassis_instance].split("eth")[1])
```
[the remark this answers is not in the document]
daall: Fixed

tests/conftest.py
daall: That makes sense! Rather than doing that, I separated the chassis port bring-up part from the neighbor info methods. 

`get_topo_neigh` and `handle_neighconn` should be the same as before now minus the variable names, all the chassis port admin status handling is in `get_chassis_instance_port_statuses` and `handle_chassis_connections` (it makes more sense to get that info from the chassis instance config files anyways :)).

tests/conftest.py
```diff
@@ -1300,57 +1301,110 @@ def get_inband(self, ctnname):
...
                             continue
-                        if ctnname not in neighs_by_ctn:
-                            neighs_by_ctn[ctnname] = []
-                        neighs_by_ctn[ctnname].append((vethintf - 1, nbraddr))
-        return neighs_by_ctn
+
+                        if chassis_container_name not in instance_to_neighbor_map:
+                            instance_to_neighbor_map[chassis_container_name] = []
+
+                        instance_to_neighbor_map[chassis_container_name].append((chassis_veth_intf - 1,
```
daall, line 1341: btw, I confirmed that this part is supposed to be `chassis_veth_intf`. `chassis_veth_intf` was `vethintf` before, which looks like the chassis half of the connection. This makes sense to me since we are mapping from the `dvs` instance to the neighbor. Let me know if you see any problem with this.

saravanansv: chassis_veth_intf seems correct.

PR #1414, opened 2020-08-24
[vnet] Maintain the reference count of the nexthop when creating a vn…

liron-ze:
**What I did**
 Maintain the reference count of the nexthop when creating  a vnet route

**Why I did it**
Due to sequence problems or other reasons, the vnet route in ASIC_DB has not been deleted, and if you directly delete the related nexthop or interface, syncd will report an error.

swss#orchagent: :- meta_generic_validation_remove: object 0x600000000099a reference count is 1, can't remove
swss#orchagent: :- removeRouterIntfs: Failed to remove router interface for port Ethernet0, rv:-5
swss#supervisord: orchagent terminate called after throwing an instance of 'std::runtime_error'
swss#supervisord: orchagent   what():  Failed to remove router interface.

**How I verified it**
Create a Vnet1
bind Ethernet0 to Vnet1
add ip for Ethernet0
add a route with the Ethernet0 as the next hop in Vnet1

delete ip and unbind Ethernet0
delete the route

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -163,6 +163,47 @@ bool VNetVrfObject::addRoute(IpPrefix& ipPrefix, tunnelEndpoint& endp)
     return true;
 }
 
+void VNetVrfObject::increaseNextHopRefCount(const nextHop& nh)
+{
+    /* Return when there is no next hop (dropped) */
+    if (nh.ips.getSize() == 0)
+    {
+        return;
+    }
+    else if (nh.ips.getSize() == 1)
+    {
+        NextHopKey nexthop(nh.ips.to_string(), nh.ifname);
+        if (nexthop.ip_address.isZero())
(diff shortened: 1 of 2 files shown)


orchagent/vnetorch.cpp
```diff
@@ -163,6 +163,47 @@ bool VNetVrfObject::addRoute(IpPrefix& ipPrefix, tunnelEndpoint& endp)
...
+{
+    /* Return when there is no next hop (dropped) */
+    if (nh.ips.getSize() == 0)
+    {
+        return;
+    }
+    else if (nh.ips.getSize() == 1)
+    {
+        NextHopKey nexthop(nh.ips.to_string(), nh.ifname);
+        if (nexthop.ip_address.isZero())
```
prsunny, line 176: Minor comment - can you provide open and closing braces here and the below "if" condition?

PR #1419, opened 2020-08-25
Update Configuration.md

vasant17:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/doc/Configuration.md
+++ b/doc/Configuration.md
@@ -1149,6 +1149,39 @@ The packet action could be:
 }
 ```
 
+### BREAKOUT_CFG
+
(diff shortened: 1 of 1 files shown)


doc/Configuration.md
```diff
@@ -1149,6 +1149,39 @@ The packet action could be:
 }
 ```
 
+### BREAKOUT_CFG
+
+This table is introduced as part of Dynamic Port Breakout(DPB) feature.
```
zhenggen-xu, line 1154: I think we should be more explicitly saying this table is auto-generated during the system init, and is also changed when we breakout the ports by CLI.

PR #1421, opened 2020-08-28
[bitmap_vnet] Fix VNET route priority issue

volodymyrsamotiy:

**What I did**
Provided logic to create "full mask" routes with highest priority.

**Why I did it**
To fix VNET route priority issue in order to support flow when some IPs from the subnet are defined by another route (nexthop tunnel) .

**How I verified it**
* Create VxLAN tunnel
* Create VNET
* Create VNET interface (```192.168.1.1/24```)
* Create neighbor for VNET interface (```192.168.1.2```)
* Create VNET tunnel route for the host from the local subnet (```192.168.1.10/32 via tunnel```)
* Send TCP packet to the switch with destination IP equal to ```192.168.1.2```
* Verify packet was routed correctly and received on the neighbor host
* Send TCP packet to the switch with destination IP equal to ```192.168.1.10```
* Verify that encapsulated packet was received on the tunnel host (instead of neighbor host)

**Details if related**
N/A

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -335,11 +335,20 @@ void VNetBitmapObject::recycleBitmapId(const string& vnet)
     }
 }
 
-uint32_t VNetBitmapObject::getFreeTunnelRouteTableOffset()
+uint32_t VNetBitmapObject::getFreeTunnelRouteTableOffset(IpPrefix ipPfx)
 {
     SWSS_LOG_ENTER();
 
-    for (uint32_t i = 0; i < tunnelOffsets_.size(); i++)
+    uint32_t offsetStart = VNET_ROUTE_FULL_MASK_OFFSET_MAX + 1;
+    uint32_t offsetEnd = tunnelOffsets_.size();
+
+    if (ipPfx.isFullMask())
+    {
+        offsetStart = 0;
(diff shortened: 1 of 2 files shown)


prsunny (approved):
@marian-pritsak , could you review

orchagent/vnetorch.cpp
```diff
@@ -335,11 +335,20 @@ void VNetBitmapObject::recycleBitmapId(const string& vnet)
     }
 }
 
-uint32_t VNetBitmapObject::getFreeTunnelRouteTableOffset()
+uint32_t VNetBitmapObject::getFreeTunnelRouteTableOffset(IpPrefix ipPfx)
 {
     SWSS_LOG_ENTER();
 
-    for (uint32_t i = 0; i < tunnelOffsets_.size(); i++)
+    uint32_t offsetStart = VNET_ROUTE_FULL_MASK_OFFSET_MAX + 1;
```
marian-pritsak, line 342: tunnelOffsets[VNET_ROUTE_FULL_MASK_OFFSET_MAX] will not be included in any range.
It should be `uint32_t offsetStart = VNET_ROUTE_FULL_MASK_OFFSET_MAX;`

marian-pritsak (approved):

PR #1429, opened 2020-09-05
new BGP Monitor Daemon to pull BGP peer state and store in State DB for MIB consumption

gechiang:
Added bgpmon (BGP Monitor Daemon) support to assist gathering BGP related information periodically based on BGP activity detection to store those information into State DB.  For components such as SNMP agent that uses this new BGP related DB in  state DB to support the MIB operation.
This PR has the implementation of the bgpmon processing code. There is another PR (https://github.com/Azure/sonic-buildimage/pull/5329) needed in the dockers/docker-fpm-frr to start up this bgpmon inside of the BGP docker.
Both of these PRs are needed to achieve the full functionality of bgpmon running inside of BGP docker.

The HLD doc PR can be found here:
- [Azure-SONiC] (https://github.com/Azure/SONiC/pull/667)

The following PR (SNMP Agent Changes to use State DB for BGP MIB )has a dependency on this PR:
- [Azure-sonic-snmpagent] (https://github.com/Azure/sonic-snmpagent/pull/158)

Signed-off-by: Gen-Hwa Chiang <gechiang@microsoft.com>

**- What I did**

Added a new daemon to be run under BGP docker periodically to gather the BGP peer state information and update them in 
the state DB.  In order not to continuously running even when there are no BGP peer activities, this new daemon will use the BGP frr.log file timestamp as an indicator whether there are any activities in BGP that warrants it to inspect the peer state info for possible updates.

**- How I did it**

- Start a periodic timer to get triggered every 15 second.
- Check if there is a change in the timestamp of the frr.log file against a cached timestamp from the previous update.
- Request the back-end handler (FRR/Zebra) to dump out the "show bgp summary json"
- For each peer state that it just received, check it against the previous state that was cached and if there is a change or this is new peer, update/add the corresponding state DB table entry.  
- At the end of processing all the peer info, check if there are any stale peer entry present and delete the corresponding state DB entry accordingly.

**- How to verify it**
For IPV4 You can perform "show ip bgp summary" and use that output to check against the state DB for each peer.
The state DB key would be something like "NEIGH_STATE_TABLE|10.0.0.33"  where 10.0.0.33 is the peer ip address.
you can then read out the state information by issuing the redis cmd hgetall "NEIGH_STATE_TABLE|10.0.0.33". Compare that with the corresponding output of "show ip bgp summary".
Here is a sample output from the State DB:
admin@str-s6000-acs-8:~$ redis-cli -n 6
127.0.0.1:6379[6]> keys "NEIGH*"
 1) "NEIGH_STATE_TABLE|fc00::1a"
 2) "NEIGH_STATE_TABLE|fc00::32"
 3) "NEIGH_STATE_TABLE|10.0.0.33"
 4) "NEIGH_STATE_TABLE|fc00::66"
 5) "NEIGH_STATE_TABLE|fc00::62"
 6) "NEIGH_STATE_TABLE|fc00::46"
 7) "NEIGH_STATE_TABLE|10.0.0.35"
 8) "NEIGH_STATE_TABLE|10.0.0.53"
 9) "NEIGH_STATE_TABLE|10.0.0.47"
10) "NEIGH_STATE_TABLE|10.0.0.9"
11) "NEIGH_STATE_TABLE|fc00::6e"
12) "NEIGH_STATE_TABLE|10.0.0.41"
13) "NEIGH_STATE_TABLE|10.0.0.63"
14) "NEIGH_STATE_TABLE|10.0.0.17"
15) "NEIGH_STATE_TABLE|fc00::4a"
16) "NEIGH_STATE_TABLE|fc00::52"
17) "NEIGH_STATE_TABLE|fc00::5e"
18) "NEIGH_STATE_TABLE|10.0.0.29"
19) "NEIGH_STATE_TABLE|10.0.0.45"
20) "NEIGH_STATE_TABLE|fc00::2a"
21) "NEIGH_STATE_TABLE|fc00::42"
22) "NEIGH_STATE_TABLE|10.0.0.1"
23) "NEIGH_STATE_TABLE|10.0.0.59"
127.0.0.1:6379[6]> hgetall "NEIGH_STATE_TABLE|10.0.0.1"
1) "state"
2) "Established"
127.0.0.1:6379[6]>

you can do the same for ipv6 by performing "show ipv6 bgp summary" and follow the same steps  above to validate with the corresponding peer states stored in state DB.

### DIFF
--- a/fpmsyncd/bgpmon.py
+++ b/fpmsyncd/bgpmon.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 """"
 Description: bgpmon.py -- populating bgp related information in stateDB.
@@ -34,22 +34,14 @@ import traceback
 
 class BgpStateGet():
     def __init__(self):
-        # list ipv4_n stores the IPV4 Neighbor peer Ip address
-        # dic ipv4_n_state stores the IPV4 Neighbor peer state entries
-        # list ipv6_n stores the IPV4 Neighbor peer Ip address
-        # dic ipv6_n_state stores the IPV4 Neighbor peer state entries
-        # list new_ipv4_n stores the new snapshot of IPV4 Neighbor ip address
-        # dic new_ipv4_n_state stores the new snapshot of IPV4 Neighbor states 
-        # list new_ipv6_n stores the new snapshot of IPV6 Neighbor ip address
-        # dic new_ipv6_n_state stores the new snapshot of IPV4 Neighbor states 
-        self.ipv4_n = []
-        self.ipv4_n_state = {}
-        self.ipv6_n = []
-        self.ipv6_n_state = {}
-        self.new_ipv4_n = []
-        self.new_ipv4_n_state = {}
-        self.new_ipv6_n = []
-        self.new_ipv6_n_state = {}
+        # list peer_l stores the Neighbor peer Ip address
+        # dic peer_state stores the Neighbor peer state entries
+        # list new_peer_l stores the new snapshot of Neighbor peer ip address
+        # dic new_peer_state stores the new snapshot of Neighbor peer states 
+        self.peer_l = []
+        self.peer_state = {}
+        self.new_peer_l = []
+        self.new_peer_state = {}
         self.cached_timestamp = 0
         self.db = swsssdk.SonicV2Connector()
         self.db.connect(self.db.STATE_DB, False)
@@ -73,87 +65,58 @@ class BgpStateGet():
 
     # Get a new snapshot of BGP neighbors and store them in the "new" location
     def get_all_neigh_states(self):
+        ipv6_peer_l = []
         try:
             cmd = "vtysh -c 'show bgp summary json'"
             output = commands.getoutput(cmd)
             peer_info = json.loads(output)
             # no exception, safe to Clean the "new" lists/dic for new sanpshot
-            del self.new_ipv4_n[:]
-            self.new_ipv4_n_state.clear()
-            del self.new_ipv6_n[:]
-            self.new_ipv6_n_state.clear()
+            del self.new_peer_l[:]
+            self.new_peer_state.clear()
             if "ipv4Unicast" in peer_info and "peers" in peer_info["ipv4Unicast"]:
-                self.new_ipv4_n = peer_info["ipv4Unicast"]["peers"].keys()
-                for i in range (0, len(self.new_ipv4_n)):
-                    self.new_ipv4_n_state[self.new_ipv4_n[i]] = \
-                    peer_info["ipv4Unicast"]["peers"][self.new_ipv4_n[i]]["state"]
+                self.new_peer_l = peer_info["ipv4Unicast"]["peers"].keys()
+                for i in range (0, len(self.new_peer_l)):
+                    self.new_peer_state[self.new_peer_l[i]] = \
+                    peer_info["ipv4Unicast"]["peers"][self.new_peer_l[i]]["state"]
 
             if "ipv6Unicast" in peer_info and "peers" in peer_info["ipv6Unicast"]:
-                self.new_ipv6_n = peer_info["ipv6Unicast"]["peers"].keys()
-                for i in range (0, len(self.new_ipv6_n)):
-                    self.new_ipv6_n_state[self.new_ipv6_n[i]] = \
-                    peer_info["ipv6Unicast"]["peers"][self.new_ipv6_n[i]]["state"]
+                ipv6_peer_l = peer_info["ipv6Unicast"]["peers"].keys()
+                self.new_peer_l.extend(ipv6_peer_l)
+                for i in range (0, len(ipv6_peer_l)):
+                    self.new_peer_state[ipv6_peer_l[i]] = \
+                    peer_info["ipv6Unicast"]["peers"][ipv6_peer_l[i]]["state"]
 
         except Exception:
             syslog.syslog(syslog.LOG_ERR, "*ERROR* get_all_neigh_states Exception: %s"
                     % (traceback.format_exc()))
 
     def update_neigh_states(self):
-        # handle IPV4 case
-        for i in range (0, len(self.new_ipv4_n)):
-            neighb = self.new_ipv4_n[i]
-            key = "NEIGH_STATE_TABLE|%s" % neighb
-            if neighb in self.ipv4_n:
+        for i in range (0, len(self.new_peer_l)):
+            peer = self.new_peer_l[i]
+            key = "NEIGH_STATE_TABLE|%s" % peer
+            if peer in self.peer_l:
                 # only update the entry if state changed
-                if self.ipv4_n_state[neighb] != self.new_ipv4_n_state[neighb]:
+                if self.peer_state[peer] != self.new_peer_state[peer]:
                     # state changed. Update state DB for this entry
-                    state = self.new_ipv4_n_state[neighb]
+                    state = self.new_peer_state[peer]
                     self.db.set(self.db.STATE_DB, key, 'state', state)
-                    self.ipv4_n_state[neighb] = state
+                    self.peer_state[peer] = state
                 # remove this neighbor from old list since it is accounted for
-                self.ipv4_n.remove(neighb)
+                self.peer_l.remove(peer)
             else:
                 # New neighbor found case. Add to dictionary and state DB
-                state = self.new_ipv4_n_state[neighb]
+                state = self.new_peer_state[peer]
                 self.db.set(self.db.STATE_DB, key, 'state', state)
-                self.ipv4_n_state[neighb] = state
+                self.peer_state[peer] = state
         # Check for stale state entries to be cleaned up
-        while len(self.ipv4_n) > 0:
+        while len(self.peer_l) > 0:
             # remove this from the stateDB and the current nighbor state entry
-            neighb = self.ipv4_n.pop(0)
-            del_key = "NEIGH_STATE_TABLE|%s" % neighb
+            peer = self.peer_l.pop(0)
+            del_key = "NEIGH_STATE_TABLE|%s" % peer
             self.db.delete(self.db.STATE_DB, del_key)
-            del self.ipv4_n_state[neighb]
+            del self.peer_state[peer]
         # Save the new List
-        self.ipv4_n = self.new_ipv4_n[:]
-
-        # handle IPV6 case
-        for i in range (0, len(self.new_ipv6_n)):
-            neighb = self.new_ipv6_n[i]
-            key = "NEIGH_STATE_TABLE|%s" % neighb
-            if neighb in self.ipv6_n:
-                # only update the entry if sate changed
-                if self.ipv6_n_state[neighb] != self.new_ipv6_n_state[neighb]:
-                    # state changed. Update state DB for this entry
-                    state = self.new_ipv6_n_state[neighb]
-                    self.db.set(self.db.STATE_DB, key, 'state', state)
-                    self.ipv6_n_state[neighb] = state
-                # remove this neighbor from old list since it is accounted for
-                self.ipv6_n.remove(neighb)
-            else:
-                # New neighbor found case. Add to dictionary and state DB
-                state = self.new_ipv6_n_state[neighb]
-                self.db.set(self.db.STATE_DB, key, 'state', state)
-                self.ipv6_n_state[neighb] = state
-        # Check for stale state entries to be cleaned up
-        while len(self.ipv6_n) > 0:
-            # remove this from the stateDB and the current nighbor state entry
-            neighb = self.ipv6_n.pop(0)
-            del_key = "NEIGH_STATE_TABLE|%s" % neighb
-            self.db.delete(self.db.STATE_DB, del_key)
-            del self.ipv6_n_state[neighb]
-        # Save the new List
-        self.ipv6_n = self.new_ipv6_n[:]
+        self.peer_l = self.new_peer_l[:]
 
 def main():
 
--- a/fpmsyncd/bgpmon.py
+++ b/fpmsyncd/bgpmon.py
@@ -29,7 +29,6 @@ import os
 import syslog
 import swsssdk
 import time
-import traceback
 
 PIPE_BATCH_MAX_COUNT = 50
 
--- a/fpmsyncd/bgpmon.py
+++ b/fpmsyncd/bgpmon.py
@@ -1,0 +1,176 @@
+#!/usr/bin/env python
+
+""""
+Description: bgpmon.py -- populating bgp related information in stateDB.
+    script is started by supervisord in bgp docker when the docker is started.
+
+    Initial creation of this daemon is to assist SNMP agent in obtaining the 
+    BGP related information for its MIB support. The MIB that this daemon is
+    assiting is for the CiscoBgp4MIB (Neighbor state only). If there are other
+    BGP related items that needs to be updated in a periodic manner in the 
+    future, then more can be added into this process.
+
+    The script check if there are any bgp activities by monitoring the bgp
+    frr.log file timestamp.  If activity is detected, then it will request bgp
+    neighbor state via vtysh cli interface. This bgp activity monitoring is 
+    done periodically (every 15 second). When triggered, it looks specifically
+    for the neighbor state in the json output of show ip bgp neighbors json
+    and update the state DB for each neighbor accordingly.
+    In order to not disturb and hold on to the State DB access too long and
+    removal of the stale neighbors (neighbors that was there previously on 
+    previous get request but no longer there in the current get request), a
+    "previous" neighbor dictionary will be kept and used to determine if there
+    is a need to perform update or the peer is stale to be removed from the
+    state DB
+"""
+
+import commands
+import copy
+import errno
+import json
+import os
+import sys
+import syslog
+import swsssdk
+import time
+import traceback
+from swsscommon import swsscommon
+from time import gmtime, strftime
+
+class BgpStateGet():
+    def __init__(self):
+        self.ipv4_n = []
+        self.ipv4_n_state = {}
+        self.ipv6_n = []
+        self.ipv6_n_state = {}
+        self.new_ipv4_n = []
+        self.new_ipv4_n_state = {}
+        self.new_ipv6_n = []
+        self.new_ipv6_n_state = {}
+        self.retrieve_fail_cnt = 0
+        self.cached_timestamp = 0
+        self.db = swsssdk.SonicV2Connector(host='127.0.0.1')
+        self.db.connect(self.db.STATE_DB, False)
+        self.db.delete_all_by_pattern(self.db.STATE_DB, "NEIGH_STATE_TABLE|*" )
+
+    # A quick way to check if there are anything happening within BGP is to
+    # check its log file has any activities. This is by checking its modified
+    # timestamp against the cached timestamp that we keep and if there is a
+    # difference, there is activity detected. In case the log file got wiped
+    # out, it will default back to constant pulling every 15 seconds
+    def bgp_activity_detected(self):
+        try:
+            timestamp = os.stat("/var/log/frr/frr.log").st_mtime
+            if timestamp != self.cached_timestamp:
+                #syslog.syslog("bgpmon timestamp diff detected: {}".format(timestamp))
+                self.cached_timestamp = timestamp
+                return True
+            else:
+                return False
+        except Exception, e:
+            return True
+
+    # Get a new snapshot of BGP neighbors and store them in the "new" location
+    def get_all_neigh_states(self):
+        # Clean up the "new" dictionaries
+        del self.new_ipv4_n[:]
+        self.new_ipv4_n_state.clear()
+        del self.new_ipv6_n[:]
+        self.new_ipv6_n_state.clear()
+        try:
+            cmd = "vtysh -c 'show bgp summary json'"
+            output = commands.getoutput(cmd)
+            peer_info = json.loads(output)
+            if "ipv4Unicast" in peer_info and "peers" in peer_info["ipv4Unicast"]:
+                self.new_ipv4_n = peer_info["ipv4Unicast"]["peers"].keys()
+                for i in range (0, len(self.new_ipv4_n)):
+                    self.new_ipv4_n_state[self.new_ipv4_n[i]] =
+                    peer_info["ipv4Unicast"]["peers"][self.new_ipv4_n[i]]["state"]
+
+            if "ipv6Unicast" in peer_info and "peers" in peer_info["ipv6Unicast"]:
+                self.new_ipv6_n = peer_info["ipv6Unicast"]["peers"].keys()
+                for i in range (0, len(self.new_ipv6_n)):
+                    self.new_ipv6_n_state[self.new_ipv6_n[i]] =
+                    peer_info["ipv6Unicast"]["peers"][self.new_ipv6_n[i]]["state"]
+
+        except Exception:
+            self.retrieve_fail_cnt += 1
+            syslog.syslog(syslog.LOG_ERR, "*ERROR* get_all_neigh_states Exception: %s"
+                    % (traceback.format_exc()))
+
+    def update_neigh_states(self):
+        # handle IPV4 case
+        for i in range (0, len(self.new_ipv4_n)):
+            neighb = self.new_ipv4_n[i]
+            key = "NEIGH_STATE_TABLE|%s" % neighb
+            if neighb in self.ipv4_n:
+                # only update the entry if sate changed
+                if self.ipv4_n_state[neighb] != self.new_ipv4_n_state[neighb]:
+                    # state changed. Update state DB for this entry
+                    state = self.new_ipv4_n_state[neighb]
+                    self.db.set(self.db.STATE_DB, key, 'state', state)
+                    self.ipv4_n_state[neighb] = state
+                # remove this neighbor from old list since it is accounted for
+                self.ipv4_n.remove(neighb)
+            else:
+                # New neighbor found case. Add to dictionary and state DB
+                state = self.new_ipv4_n_state[neighb]
+                self.db.set(self.db.STATE_DB, key, 'state', state)
+                self.ipv4_n_state[neighb] = state
+        # Check for stale state entries to be cleaned up
+        while len(self.ipv4_n) > 0:
+            # remove this from the stateDB and the current nighbor state entry
+            neighb = self.ipv4_n.pop(0)
+            del_key = "NEIGH_STATE_TABLE|%s" % neighb
+            self.db.delete(self.db.STATE_DB, del_key)
+            del self.ipv4_n_state[neighb]
+        # Save the new List
+        self.ipv4_n = self.new_ipv4_n[:]
+
+        # handle IPV6 case
+        for i in range (0, len(self.new_ipv6_n)):
+            neighb = self.new_ipv6_n[i]
+            key = "NEIGH_STATE_TABLE|%s" % neighb
+            if neighb in self.ipv6_n:
+                # only update the entry if sate changed
+                if self.ipv6_n_state[neighb] != self.new_ipv6_n_state[neighb]:
+                    # state changed. Update state DB for this entry
+                    state = self.new_ipv6_n_state[neighb]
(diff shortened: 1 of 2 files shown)


gechiang:
retest lgtm please

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env python
```
qiluo-msft, line 1: >python [](start = 15, length = 6)

Could you be specific on python2/3?

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+        self.ipv4_n_state = {}
+        self.ipv6_n = []
+        self.ipv6_n_state = {}
+        self.new_ipv4_n = []
+        self.new_ipv4_n_state = {}
+        self.new_ipv6_n = []
+        self.new_ipv6_n_state = {}
+        self.retrieve_fail_cnt = 0
+        self.cached_timestamp = 0
+        self.db = swsssdk.SonicV2Connector(host='127.0.0.1')
```
qiluo-msft, line 47: >host='127.0.0.1' [](start = 43, length = 16)

In multi-DB (not multi-ASIC) design, the connection info is from config_db.json. So please no hard code

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+        # Save the new List
+        self.ipv6_n = self.new_ipv6_n[:]
+
+def main():
+
+    print "bgpmon service started"
+
+    try:
+        bgp_state_get = BgpStateGet()
+    except Exception, e:
```
qiluo-msft, line 159:
```suggestion
    except Exception as e:
```

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+                return True
+            else:
+                return False
+        except Exception, e:
+            return True
+
+    # Get a new snapshot of BGP neighbors and store them in the "new" location
+    def get_all_neigh_states(self):
+        # Clean up the "new" dictionaries
+        del self.new_ipv4_n[:]
```
qiluo-msft, line 71: >del self.new_ipv4_n[:] [](start = 8, length = 22)

Just curios why not use `self.new_ipv4_n.clear()`

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+        self.new_ipv4_n_state.clear()
+        del self.new_ipv6_n[:]
+        self.new_ipv6_n_state.clear()
+        try:
+            cmd = "vtysh -c 'show bgp summary json'"
+            output = commands.getoutput(cmd)
```
qiluo-msft, line 77: >commands.getoutput [](start = 21, length = 18)

Suggest check return code

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+import json
+import os
+import syslog
+import swsssdk
+import time
+import traceback
+
+class BgpStateGet():
+    def __init__(self):
+        self.ipv4_n = []
```
qiluo-msft, line 37: >_n [](start = 17, length = 2)

What does `_n` mean?

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+                self.new_ipv4_n = peer_info["ipv4Unicast"]["peers"].keys()
+                for i in range (0, len(self.new_ipv4_n)):
+                    self.new_ipv4_n_state[self.new_ipv4_n[i]] = \
+                    peer_info["ipv4Unicast"]["peers"][self.new_ipv4_n[i]]["state"]
+
+            if "ipv6Unicast" in peer_info and "peers" in peer_info["ipv6Unicast"]:
+                self.new_ipv6_n = peer_info["ipv6Unicast"]["peers"].keys()
+                for i in range (0, len(self.new_ipv6_n)):
+                    self.new_ipv6_n_state[self.new_ipv6_n[i]] = \
+                    peer_info["ipv6Unicast"]["peers"][self.new_ipv6_n[i]]["state"]
```
qiluo-msft, line 89: ipv4 and ipv6 code are quite similar, could you reuse code?

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+
+        except Exception:
+            self.retrieve_fail_cnt += 1
```
qiluo-msft, line 92: >retrieve_fail_cnt [](start = 17, length = 17)

Not used?

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+            syslog.syslog(syslog.LOG_ERR, "*ERROR* get_all_neigh_states Exception: %s"
+                    % (traceback.format_exc()))
+
+    def update_neigh_states(self):
+        # handle IPV4 case
+        for i in range (0, len(self.new_ipv4_n)):
+            neighb = self.new_ipv4_n[i]
+            key = "NEIGH_STATE_TABLE|%s" % neighb
+            if neighb in self.ipv4_n:
+                # only update the entry if sate changed
```
qiluo-msft, line 102: >sate [](start = 43, length = 4)

typo?

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+        while len(self.ipv4_n) > 0:
+            # remove this from the stateDB and the current nighbor state entry
+            neighb = self.ipv4_n.pop(0)
+            del_key = "NEIGH_STATE_TABLE|%s" % neighb
+            self.db.delete(self.db.STATE_DB, del_key)
+            del self.ipv4_n_state[neighb]
+        # Save the new List
+        self.ipv4_n = self.new_ipv4_n[:]
+
+        # handle IPV6 case
```
qiluo-msft, line 125: ># handle IPV6 case [](start = 8, length = 18)

reuse ipv4 code?

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+    # A quick way to check if there are anything happening within BGP is to
+    # check its log file has any activities. This is by checking its modified
+    # timestamp against the cached timestamp that we keep and if there is a
+    # difference, there is activity detected. In case the log file got wiped
+    # out, it will default back to constant pulling every 15 seconds
+    def bgp_activity_detected(self):
+        try:
+            timestamp = os.stat("/var/log/frr/frr.log").st_mtime
+            if timestamp != self.cached_timestamp:
+                #syslog.syslog("bgpmon timestamp diff detected: {}".format(timestamp))
```
qiluo-msft, line 60: >#syslog.syslog [](start = 16, length = 14)

Remove dead code

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+                self.cached_timestamp = timestamp
+                return True
+            else:
+                return False
+        except Exception, e:
```
qiluo-msft, line 65:
```suggestion
        except Exception as e:
```

qiluo-msft, line 65: >Exception [](start = 15, length = 9)

Use more specific exception type

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+                    self.new_ipv4_n_state[self.new_ipv4_n[i]] = \
+                    peer_info["ipv4Unicast"]["peers"][self.new_ipv4_n[i]]["state"]
+
+            if "ipv6Unicast" in peer_info and "peers" in peer_info["ipv6Unicast"]:
+                self.new_ipv6_n = peer_info["ipv6Unicast"]["peers"].keys()
+                for i in range (0, len(self.new_ipv6_n)):
+                    self.new_ipv6_n_state[self.new_ipv6_n[i]] = \
+                    peer_info["ipv6Unicast"]["peers"][self.new_ipv6_n[i]]["state"]
+
+        except Exception:
```
qiluo-msft, line 91: >Exception [](start = 15, length = 9)

Use more specific exception type

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,171 @@
...
+        # handle IPV4 case
+        for i in range (0, len(self.new_ipv4_n)):
+            neighb = self.new_ipv4_n[i]
+            key = "NEIGH_STATE_TABLE|%s" % neighb
+            if neighb in self.ipv4_n:
+                # only update the entry if sate changed
+                if self.ipv4_n_state[neighb] != self.new_ipv4_n_state[neighb]:
+                    # state changed. Update state DB for this entry
+                    state = self.new_ipv4_n_state[neighb]
+                    self.db.set(self.db.STATE_DB, key, 'state', state)
```
qiluo-msft, line 106: >self.db.set [](start = 20, length = 11)

It will be more efficient to group all the DB operations in `update_neigh_states()` into pipeline.

qiluo-msft, line 106: >set [](start = 28, length = 3)

Suggest to use [https://redis.io/commands/expire](https://redis.io/commands/expire) in case this process crashes or always fails.

If this is the right direction, "only update the entry if sate changed" may be not that important since we use pipeline and there is only one transaction.

qiluo-msft (changes requested):
As comments

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft This script is not depended on specific version of python2 nor python3.
Do we have to specify with python2 or python3 and can not leave it as "python" if there are no specific dependency?
I did a quick grep from two repositories (sonic-swss, sonic-utilities) and the result seems to be using the same as what I did. only one script did specify python2.  Here is the result in case you are interested:
gechiang@gechiang-vm-01:~/sonic-utilities$ grep -R "/usr/bin/env python" .
./pcieutil/main.py:#!/usr/bin/env python
./sfputil/main.py:#!/usr/bin/env python
./utilities_common/util_base.py:#!/usr/bin/env python2
./scripts/decode-syseeprom:#!/usr/bin/env python
./scripts/dropconfig:#!/usr/bin/env python
./scripts/configlet:#!/usr/bin/env python
./scripts/pfcstat:#!/usr/bin/env python
./scripts/watermarkstat:#!/usr/bin/env python
./scripts/fast-reboot-dump.py:#!/usr/bin/env python
./scripts/queuestat:#!/usr/bin/env python
./scripts/dropstat:#!/usr/bin/env python
./scripts/lldpshow:#!/usr/bin/env python
./scripts/db_migrator.py:#!/usr/bin/env python
./scripts/route_check.py:#!/usr/bin/env python
./scripts/neighbor_advertiser:#!/usr/bin/env python
./scripts/boot_part:#!/usr/bin/env python
./scripts/update_json.py:#! /usr/bin/env python
./scripts/portstat:#!/usr/bin/env python
./scripts/pcmping:#!/usr/bin/env python
./scripts/watermarkcfg:#!/usr/bin/env python
./scripts/intfstat:#!/usr/bin/env python
./scripts/port2alias:#!/usr/bin/env python
./consutil/lib.py:#!/usr/bin/env python
./consutil/main.py:#!/usr/bin/env python
./crm/main.py:#!/usr/bin/env python
./ssdutil/main.py:#!/usr/bin/env python
./acl_loader/main.py:#!/usr/bin/env python
./pddf_psuutil/main.py:#!/usr/bin/env python
./pfc/main.py:#!/usr/bin/env python
./fwutil/lib.py:#!/usr/bin/env python
./fwutil/log.py:#!/usr/bin/env python
./fwutil/main.py:#!/usr/bin/env python
./pddf_fanutil/main.py:#!/usr/bin/env python
./show/mlnx.py:#!/usr/bin/env python
./pddf_ledutil/main.py:#!/usr/bin/env python
./config/nat.py:#!/usr/bin/env python
./config/aaa.py:#!/usr/bin/env python -u
./config/mlnx.py:#!/usr/bin/env python
./psuutil/main.py:#!/usr/bin/env python
./pddf_thermalutil/main.py:#!/usr/bin/env python
./watchdogutil/main.py:#!/usr/bin/env python
gechiang@gechiang-vm-01:~/sonic-utilities$ cd ~/sonic-swss/
gechiang@gechiang-vm-01:~/sonic-swss$ grep -R "/usr/bin/env python" .
./neighsyncd/restore_neighbors.py:#!/usr/bin/env python
./fpmsyncd/bgpmon.py:#!/usr/bin/env python
./fpmsyncd/bgp_eoiu_marker.py:#!/usr/bin/env python
gechiang@gechiang-vm-01:~/sonic-swss$

Let me know if "need to be specific about python2 or python 3" is the standard convention for SONiC?  If so, I can change it to python2.  If not,  will leave it the same as other scripts does.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Agreed.  Will update to fix this.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Agreed. Will change it per suggested.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft list clear() method is not available until python 3.3. that is why I am using "del my_list[:]" to do the list clearing operation.  It is interesting that dic.clear() is available but somehow was not included for list in the core python...

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft I think you are suggesting me to use rc, output = commands.getstatusoutput(cmd)
The reason that I am not using that is because this is a "fixed" cmd line that this script keeps running with and there are no user intervention or chance to modify this cmd.  So once it is working the cmd itself should not be the cause of the problem.  While there may be potentially other run time scenarios such as Zebra somehow stopped working or restarted and momentarily the command may fail.  But this code is in the "try"" clause already and the bgpmon daemon should continue to function in case zebra does restart and come back to operate again in which case this daemon will just continue when zebra becomes functional.  So I don't think the RC will provide any additional value thus elect to use commands.getoutput(cmd) instead.  Hope this clarify why it was coded this way.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Originally it was named as "self.ipv4_neighbor".  It is a list to keep all the ipv4 neighbor's ip address. But I chose to shorten the name because the lines are getting too long and require multiple lines to wrap in order to allow user to still read it with a small width terminal (trying to keep everything within 79 chars...).  So I made a decision to chop the name to shorten it.  I have added more comments in the code to explain each variable usage as following:
        # list ipv4_n stores the IPV4 Neighbor peer Ip address
        # dic ipv4_n_state stores the IPV4 Neighbor peer state entries
        # list ipv6_n stores the IPV4 Neighbor peer Ip address
        # dic ipv6_n_state stores the IPV4 Neighbor peer state entries
        # list new_ipv4_n stores the new snapshot of IPV4 Neighbor ip address
        # dic new_ipv4_n_state stores the new snapshot of IPV4 Neighbor states 
        # list new_ipv6_n stores the new snapshot of IPV6 Neighbor ip address
        # dic new_ipv6_n_state stores the new snapshot of IPV4 Neighbor state
Hope this helps...

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft I think readability is important and I do want to keep IPV4 and IPV6 in separate DS so that we have a choice to handle them independently/differently if needed in future.  The lines involved currently is very small already so Let's keep it this way for now until later when we add more feature to this daemon and find the separation becomes unnecessary and can then combine them at that time.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft It was part of my debug code where I wanted to see if there are any failure related to retrieving the output from Zebra. I was debating if I should expose this to syslog but later chose not to but did not remove it from my code either.  I will remove it.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Good catch! Typo indeed... will fix it.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Yes I can combine these but let's keep it separate for now as we will be adding more feature support in the near future and at that time if we don't have to handle IPV4 and IPV6 differently, we can simplify it by combining them together.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Agree. will fix it.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Agreed. Will change it.

gechiang: @qiluo-msft Under normal operation I do not expect any exception resulting from attempting to read the timestamp of the frr.log file.  The only exception that may occurred is that someone accidentally renamed the log file or deleted the log file so the file path no longer exists.  Another possible exception perhaps has to do with access permission such as for some reason this daemon does not have the read permission and resulted to an exception.  Since the purpose is to "try" to use the FRR log file timestamp as a way to detect any BGP activities and if this mechanism temporarily not working, this daemon should still operate but just won't be as efficient as we desire it to be (became periodically reading out the bgp summary whether there is changes or not).  So perhaps it is ok not to be specific in this case as we don't really care about the specific reason why getting the frr log file timestamp failed.  We will just go on executing the BGP info pulling periodically anyway...  Let me know if you are ok with the above reasoning.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft The purpose is to be able to handle Zebra not responding, crashed, restarting,... etc and move on with no state DB update performed.  But in any case the daemon shall continue to be waken up at the set periodic time to try again.  I am not sure which specific exception I can use here since potentially it could be multiple reasons but the daemon does not need to care as the issue may later heal on its own.  As long as it does not impact the integrity of what this daemon is performing...  Let me know if you are ok with my explanation.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft  Not sure if I understood your suggestion. Are you referring as to why not just overlay all the peer states in the stateDB without the need to check if there are any changes?  The reason is that in most of the time maybe a few may have changed/added/deleted in the new snapshot. If there are a lot of BGP peers in the DB, the unnecessary write that has no state changes can unnecessary busy the StateDB. let along the snmp proxy agent is trying to read this same table so if bgpmon locks it up for a longer time than necessary, it will have performance consequences for snmp proxy agent or any future agents that like to read these BGP peer stateDB entries. This is why it avoids unnecessary updates if the existing peer has no changes between the new snapshot and the previous cached state.  The desired design is to ensure that we favor efficiency for the "normal" steady state scenarios.  Please clarify If I misunderstood your suggestion.
Please clarify if I understood your

gechiang: @qiluo-msft Since this StateDB is required by SNMP proxy agent and potential future consumers, if this process always fails, it will not matter what the BGP StateDB entries gets properly deleted or not...  those agents will not operate correctly until bgpmon daemon is healthy again.  If my understanding is correct after reading through the "expire" documentation it will require bgpmon to periodically update the expire to ensure those BGP key states does not get expired/deleted when there is no changes needed.  Since we want to favor the design for the steady state condition (no more peer state changes in steady state), we should not use "expire" or else bgpmon in steady state will end up having to periodically update the state DB even there are no state changes.  In this desing, bgpmon when restart will force the state DB to clean up and provide the most updated state from the new snapshot it gathered from BGP/Zebra.  So StateDB clean up is taken cared of when necessary to prevent stale entries from staying in the state DB.  Also, using "expire" means someone (Redis) will need to keep track of the time for each key.  This operation is not free no matter how efficient Redis implements it...

fpmsyncd/bgpmon.py
gechiang: Done.

fpmsyncd/bgpmon.py
qiluo-msft: The reason is that python will be symlinked to different executable in different systems, or even does not exist. I assume you tested in python2, so no harm to be specific.

ref: [https://stackoverflow.com/a/19305076](https://stackoverflow.com/a/19305076)


---
In reply to: [484650010](https://github.com/Azure/sonic-swss/pull/1429#discussion_r484650010) [](ancestors = 484650010)

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,176 @@
...
+            del self.ipv4_n_state[neighb]
+        # Save the new List
+        self.ipv4_n = self.new_ipv4_n[:]
+
+        # handle IPV6 case
+        for i in range (0, len(self.new_ipv6_n)):
+            neighb = self.new_ipv6_n[i]
+            key = "NEIGH_STATE_TABLE|%s" % neighb
+            if neighb in self.ipv6_n:
+                # only update the entry if sate changed
```
qiluo-msft, line 135: >sate [](start = 43, length = 4)

Typo again. This happens to be a good argument for code reuse, so if you fix in one part, you fix every part.

fpmsyncd/bgpmon.py
qiluo-msft: Thanks for detailed analysis! I prefer conservative in order not to hide any issue. You extreme samples are the good ones for stop daemon and engage more investigation.

---
In reply to: [484672855](https://github.com/Azure/sonic-swss/pull/1429#discussion_r484672855) [](ancestors = 484672855)

fpmsyncd/bgpmon.py
qiluo-msft: No. I am talking about [https://github.com/andymccurdy/redis-py#pipelines](https://github.com/andymccurdy/redis-py#pipelines). And we have similar feature in swsscommon library.

---
In reply to: [484677456](https://github.com/Azure/sonic-swss/pull/1429#discussion_r484677456) [](ancestors = 484677456)

qiluo-msft: Let's *assume* there is a bug inside this daemon and it is stuck somewhere. It is better to expire and propagate this error to monitors. In that situation, it's dangerous to keep as false stable.

---
In reply to: [484678284](https://github.com/Azure/sonic-swss/pull/1429#discussion_r484678284) [](ancestors = 484678284)

fpmsyncd/bgpmon.py
gechiang: Ok. I will set it to python2.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft After giving some more thoughts, I agree with your comment.  Since there are no clear differences in the handling currently between IPV4 and IPV6 I will keep all the peers and peer state in the same list and dictionary thus removing the duplication of code handling.  In my next check in this should be reflected.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Yes. I changed my mind and agreed to eliminate the extra IPV6 Data structures.  If in future there is a need to handle it differently, then we will deal with it in the future. For now there is no difference so I will not take on the burden to keep them separate.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Yes I agree.  thanks for the reference.  with pipelines we can batch several StateDB operations into one single transaction and thus cuts down some overhead. I think there is one PR (https://github.com/Azure/sonic-py-swsssdk/pull/85) that started to use this pipeline model.  I would prefer making this performance enhancement with a separate PR.  One reason is to observe if there are any potential pitfall from using pipelines until we gain some more usage experiences.  The other reason is that we first make sure that the basic design works first and then take on more enhancement improvements.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Yes. I am removing the unnecessary duplication with my next check in.

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,139 @@
...
+            if timestamp != self.cached_timestamp:
+                self.cached_timestamp = timestamp
+                return True
+            else:
+                return False
+        except Exception, e:
+            return True
+
+    # Get a new snapshot of BGP neighbors and store them in the "new" location
+    def get_all_neigh_states(self):
```
qiluo-msft, line 67: >get_all_neigh_states [](start = 8, length = 20)

Could you add some unit test or vs test?

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft I changed my mind.  I have changed to use "rc, output = commands.getstatusoutput(cmd)" as it will give better error logging if it does fail.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Agreed. I have changed to use more specific exception so it will not mask out important exceptions.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Since I changed to use "rc, output = commands.getstatusoutput(cmd)" and check for the rc value to log the error in syslog, there is no longer a need to handle exception here any more. This change indirectly resolved your concern where the general exception may masked out un-anticipated exceptions that we really need to debug them if they arise.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft I have changed the code to use redis pipeline for the StateDB operations.

gechiang: @qiluo-msft Per our phone discussion I think we agreed that this is ok for now.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Code reuse completed.

fpmsyncd/bgpmon.py
gechiang: @qiluo-msft Will add unit test in a separate PR.

qiluo-msft: Prefer adding in the this PR if you can. It even benefit yourself in iterations.

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,174 @@
...
+                self.peer_l.remove(peer)
+            else:
+                # New neighbor found case. Add to dictionary and state DB
+                state = self.new_peer_state[peer]
+                data[key] = {'state':state}
+                pipe_count += 1
+                self.peer_state[peer] = state
+            if pipe_count > PIPE_BATCH_MAX_COUNT:
+                self.flush_pipe(data)
+                data = {}
```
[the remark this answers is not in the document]
gechiang: @qiluo-msft Agreed.  unline "c", python pass the object instead of copy the value so it make sense to clear the data{} inside this flush_pipe() method.

fpmsyncd/bgpmon.py
```diff
@@ -0,0 +1,174 @@
...
+            del_key = "NEIGH_STATE_TABLE|%s" % peer
+            data[del_key] = None
+            pipe_count += 1
+            del self.peer_state[peer]
+            if pipe_count > PIPE_BATCH_MAX_COUNT:
+                self.flush_pipe(data)
+                data = {}
+                pipe_count = 0
+        # If anything in the pipeline not yet flushed, flush them now
+        if pipe_count > 0:
```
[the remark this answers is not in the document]
gechiang: @qiluo-msft Agreed. There is no need to introduce the extra counter "pipe_count" since python dict implementation maintains an item count that can be accessed with O(1) time so it make sense to use the len(data) instead.

qiluo-msft (commented):
LGTM except missing test

gechiang:
Based on offline comment from Guohan, the bgpmon code should not reside in this repo.  Based on its functionality it should be moved to under sonic-buildimage/src/sonic-bgpcfgd .
I am closing this PR and have moved this same code to the following PR:
https://github.com/Azure/sonic-buildimage/pull/5329
Please use the above for further review.
There will be no more activities on this PR.

PR #1431, opened 2020-09-10
Implementation of System ports initialization, Interface & Neighbor Synchronization...

minionatwork:
This PR includes implementation of System ports initialization & orchestration, Interface & Neighbor Synchronization using GLOBAL_APP_DB, Inband Interface Configuration.

HLD: https://github.com/Azure/SONiC/pull/639
Dependency PR's
(sonic-swss-common): https://github.com/Azure/sonic-swss-common/pull/380
(sonic-sairedis): https://github.com/Azure/sonic-sairedis/pull/657


orchagent/intfsorch.cpp
```diff
@@ -1255,3 +1326,88 @@ void IntfsOrch::doTask(SelectableTimer &timer)
...
+    attr.id = SAI_ROUTER_INTERFACE_ATTR_PORT_ID;
+
+    status = sai_router_intfs_api->get_router_interface_attribute(rif_id, 1, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to get rif for %s:0x%lx, rv:%d", alias.c_str(), rif_id, status);
+        return false;
+    }
+
+    //Get the Real ID of the RIF ID
```
eswaranb, line 1373: Can you please explain why this is necessary?

vganesan-nokia: Since orchagent to syncd router interface create call is asynchronous, while trying to get the real id of router interface, we need to make sure that the router interface is actually created in SAI and real id is available in asic db. Attempt to retrieve any attribute of router interface object will suffice. Just need to make a GET call on the object to make sure the entry is available.

eswaranb: Ok, "attempt to retrieve any attribute of router interface object will suffice" makes sense to me. Can we avoid posting the RIF ID of this interface to the CHASSIS_APP_DB because that number is meaningless in the other instances? It can lead to confusion. Please also add a comment here about this.

orchagent/neighorch.cpp
```diff
@@ -39,6 +50,15 @@ bool NeighOrch::addNextHop(const IpAddress &ipAddress, const string &alias)
     }
 
     NextHopKey nexthop = { ipAddress, alias };
+    if(m_intfsOrch->isRemoteSystemPortIntf(alias))
+    {
+        //For remote system ports kernel nexthops are always on inband. Change the key
+        Port inbp;
+        if(gPortsOrch->getInbandPort(inbp))
```
ysmanman, line 57: Is it possible that getInbandPort may fail? Do we need to handle this case?

orchagent/neighorch.cpp
```diff
@@ -579,5 +650,438 @@ bool NeighOrch::removeNeighbor(const NeighborEntry &neighborEntry)
...
+{
+    SWSS_LOG_ENTER();
+
+    string cmd, res;
+    string ip_str = ip_addr.to_string();
+    string mac_str = mac_addr.to_string();
+
+    if(ip_addr.isV4())
+    {
+        cmd = "arp -s " + ip_str + " " + mac_str + " -i " + odev;
```
ysmanman, line 1035: Can you explain why using arp cmd for IPV4 neighbor, but ip neigh cmd for IPV6 neighbor? Can we use ip neigh cmd for both?

orchagent/neighorch.cpp
```diff
@@ -579,5 +650,438 @@ bool NeighOrch::removeNeighbor(const NeighborEntry &neighborEntry)
...
+
+    if(ip_addr.isV4())
+    {
+        cmd = "arp -d " + ip_str + " -i " + odev;
+        SWSS_LOG_NOTICE("IPv4 Nbr Del cmd: %s",cmd.c_str());
+    }
+    else
+    {
+        cmd = "ip -6 neigh del " + ip_str + " dev " + odev;
+        SWSS_LOG_NOTICE("IPv4 Nbr Del cmd: %s",cmd.c_str());
```
ysmanman, line 1073: typo: IPv6

vganesan-nokia: We'll fix the typo.

orchagent/neighorch.cpp
vganesan-nokia: Yes. we'll use "ip neigh" command for both IPv4 and IPv6

orchagent/neighorch.cpp
vganesan-nokia: This neighbor is on the remote system ports. Remote system port neighbors will never come here without Inband interface already up and active. So this getInbandPort() for remote system port neighbors will never fail.

orchagent/intfsorch.cpp
vganesan-nokia: The RIF ID is synced for future use when we have multiple vlan provisioning. O.k, for now, we'll avoid syncing RIF ID to CHASSIS_APP_DB. We'll also add comment about this.

orchagent/neighorch.cpp
ysmanman: If getInbandPort() will never fail, can we use assert here?

orchagent/neighorch.cpp
```diff
@@ -579,5 +650,438 @@ bool NeighOrch::removeNeighbor(const NeighborEntry &neighborEntry)
...
+    MacAddress inband_mac = gMacAddress;
+
+    auto it = consumer.m_toSync.begin();
+    while (it != consumer.m_toSync.end())
+    {
+        KeyOpFieldsValuesTuple t = it->second;
+        string key = kfvKey(t);
+        string op = kfvOp(t);
+
+        size_t found = key.find(':');
```
ysmanman, line 683: Instead of hardcoded ":", can we use getTableNameSeparator()?

orchagent/neighorch.cpp
vganesan-nokia: Yes. We'll use assert here.

orchagent/neighorch.cpp
vganesan-nokia: Yes. We'll use getTableNameSeparator() here.

orchagent/intfsorch.cpp
minionatwork: @eswaranb,

don't we require to have same rif_id (i.e vlan) when creating neighbors on the rif across all instances of it?.

cfgmgr/intfmgr.cpp
```diff
@@ -660,3 +667,51 @@ void IntfMgr::doTask(Consumer &consumer)
         it = consumer.m_toSync.erase(it);
     }
 }
+
+void IntfMgr::doCfgVoqInbandInterfaceTask(Consumer &consumer)
```
prsunny, line 671: Why is this a separate function? looks like its just duplicated?

orchagent/neighorch.h
```diff
@@ -67,6 +69,17 @@ class NeighOrch : public Orch, public Subject
...
 
     void doTask(Consumer &consumer);
+    void doVoqSystemNeighTask(Consumer &consumer);
+
+    unique_ptr<Table> m_tableVoqSystemNeighTable;
+    bool getSystemPortNeighEncapIndex(string alias, uint32_t &encap_index);
+    bool addVoqEncapIndex(string &alias, IpAddress &ip, vector<sai_attribute_t> &neighbor_attrs);
+    bool addKernelRoute(string odev, IpAddress ip_addr);
+    bool delKernelRoute(IpAddress ip_addr);
+    bool addKernelNeigh(string odev, IpAddress ip_addr, MacAddress mac_addr);
```
prsunny, line 79: So far, orchagent hasn't directly configured anything to kernel. This is a behavioral change.

minionatwork: Agree, in terms of ARP/Route, this is the exception and discussed in HLD, to keep VOQ specific changes to be simple enough.

orchagent/intfsorch.cpp
```diff
@@ -1255,3 +1326,88 @@ void IntfsOrch::doTask(SelectableTimer &timer)
...
+    Port port;
+    if(gPortsOrch->getPort(alias, port))
+    {
+        return(port.m_system_port_info.type == SAI_SYSTEM_PORT_TYPE_REMOTE);
+    }
+    //Given alias is system port alias of the local port
+    return false;
+}
+
+bool IntfsOrch::voqSyncAddIntf(string &alias, sai_object_id_t &rif_id)
```
vganesan-nokia, line 1341: The return value is not used anywhere. Make it void

orchagent/intfsorch.cpp
```diff
@@ -1255,3 +1326,88 @@ void IntfsOrch::doTask(SelectableTimer &timer)
...
+        attrs.push_back(rifIdFv);
+
+        m_tableVoqSystemInterfaceTable->set(alias, attrs);
+        return true;
+    }
+    SWSS_LOG_ERROR("Failed to get real ID from ASIC DB for RIF id 0x%lx of %s", rif_id, alias.c_str());
+    return false;
+}
+
+bool IntfsOrch::voqSyncDelIntf(string &alias)
```
vganesan-nokia, line 1390: Return value is not used anywhere. Make it void

orchagent/neighorch.h
prsunny: Can you explain why nbrmgr handle this scenario?

orchagent/main.cpp
```diff
@@ -140,6 +145,121 @@ void init_gearbox_phys(DBConnector *applDb)
...
+bool getSystemPortConfigList(vector<sai_system_port_config_t> &sysportcfglist)
+{
+    DBConnector cfgDb("CONFIG_DB", 0);
+    DBConnector appDb("APPL_DB", 0);
+
+    Table cfgDeviceMetaDataTable(&cfgDb, CFG_DEVICE_METADATA_TABLE_NAME);
+    Table cfgSystemPortTable(&cfgDb, CFG_SYSTEM_PORT_TABLE_NAME);
+    Table appSystemPortTable(&appDb, APP_SYSTEM_PORT_TABLE_NAME);
+
+    //At this point of time (i.e, at the time of orchagent start), the CONFIG_DB is completely fully
```
vganesan-nokia, line 157: Add checking confg_db_initialized before proceeding.

orchagent/neighorch.h
minionatwork: remote neighbors are created in CHASSIS_APP_DB ("app_db") and hence its processed inside OrchAgent. Now, OrchAgent could post another message to nbrmgr as if its config_db event, but its adding many indirection instead of syncing to ASIC_DB.

orchagent/neighorch.cpp
```diff
@@ -317,6 +354,14 @@ void NeighOrch::doTask(Consumer &consumer)
             continue;
         }
 
+        if(gPortsOrch->isInbandPort(alias))
```
vganesan-nokia, line 357: Check for the inband  port type. This is applicable only for the "port" type inband interface

orchagent/neighorch.cpp
```diff
@@ -579,5 +650,438 @@ bool NeighOrch::removeNeighbor(const NeighborEntry &neighborEntry)
 
     removeNextHop(ip_address, alias);
 
+    if(gMySwitchType == "voq")
+    {
+        //Sync the neighbor to delete from the GLOBAL_APP_DB
```
vganesan-nokia, line 655: Change GLOBAL_APP_DB name to CHASSIS_APP_DB for comments also.

orchagent/neighorch.h
prsunny: Another proposal is `CHASSIS_APP_DB -> Nbrmgr ->(Kernel, APP_DB) -> orchagent`. As we know that orchagent currently don't handle kernel programming, another issue with this model is related to warmboot reconciliation logic. We dont want to introduce orchagent to deal with reconciling kernel objects. Once orchagent gets the data from APP_DB, it can use `getSystemPortNeighEncapIndex `to get the EncapIndex.  Kindly revisit this design.

orchagent/main.cpp
```diff
@@ -140,6 +145,121 @@ void init_gearbox_phys(DBConnector *applDb)
     delete tmpGearboxTable;
 }
 
+bool getSystemPortConfigList(vector<sai_system_port_config_t> &sysportcfglist)
+{
+    DBConnector cfgDb("CONFIG_DB", 0);
+    DBConnector appDb("APPL_DB", 0);
```
abdosi, line 151: looks like we are not using it. can we remove it ?

prsunny (changes requested):
As comments for `NeighOrch`

orchagent/neighorch.cpp
```diff
@@ -579,5 +650,438 @@ bool NeighOrch::removeNeighbor(const NeighborEntry &neighborEntry)
...
+    else
+    {
+        SWSS_LOG_ERROR("Port does not exist for %s!", alias.c_str());
+        return false;
+    }
+
+    attr.id = SAI_NEIGHBOR_ENTRY_ATTR_ENCAP_INDEX;
+
+    status = sai_neighbor_api->get_neighbor_entry_attribute(&neighbor_entry, 1, &attr);
+    if (status != SAI_STATUS_SUCCESS)
```
abdosi, line 916: we we need to check encap_index > 0 here ? I see that assumption in some of below code.

orchagent/neighorch.cpp
```diff
@@ -579,5 +650,438 @@ bool NeighOrch::removeNeighbor(const NeighborEntry &neighborEntry)
...
+{
+    string cmd, res;
+
+    SWSS_LOG_ENTER();
+
+    string ip_str = ip_addr.to_string();
+
+    if(ip_addr.isV4())
+    {
+        cmd = "ip route del " + ip_str + "/32";
```
abdosi, line 1003: can we use macro IP_CMD

cfgmgr/intfmgr.cpp
vganesan-nokia: Agreed. We'll remove this function and inband config handling will be moved to existing doTask().

orchagent/main.cpp
vganesan-nokia: These are used in lines 153-155 below.

orchagent/neighorch.h
brcmnokia: Ok. We will go with nbrmgr for adding kernel ARP entry and static route. Adding static route from nbrmgr isn't clean either because nbrmgr doesnt deal with static routes as overall. Are we Ok to put static route addition in nbrmgr?.

minionatwork: Ok. We will go with nbrmgr for adding static ARP entry in kernel and also adding static route. Certainly nbrmgr isn't right place for adding static route but we dont want to create another daemon for adding static routes alone. is it ok put static routes adds from nbrmgr?. Any thoughts?

orchagent/intfsorch.cpp
eswaranb: Yes, when we do support vlan interfaces, some ID has to be exchanged to the remote ASICs. However, that will not be this RIF_ID, because this is just a local ID that SAI gives us here and doesn't mean anything in the hardware. We will probably have to add something like a SYSTEM_VLAN (similar to SYSTEM_PORT and SYSTEM_LAG) to carry that hw ID to reference the vlan at that point.

orchagent/neighorch.h
prsunny: IMO, it is ok to have nbrmgr to configure static route.. I think it is for the same neighbor entry that you are creating the static route as well, right?

minionatwork: That's right. its static route(/32 or /128) the neighbor entry being added.

orchagent/neighorch.cpp
vganesan-nokia: Yes. Fixed

orchagent/neighorch.cpp
vganesan-nokia: Yes. Fixed

orchagent/neighorch.h
minionatwork: With new design:

chassis_app_db (remote neighbors)----> nbrmgr ---> create kernel entries(route & arp) ---> nbrsyncd ---> app_db (local)
chassis_app_db (remote neighbors)----> orchagent ---> create sai neighbor for remote neighbor with encap-id.... all dependencies are resolved (system-port, system routed interface). 
app_db (local neighbor entry created for remote neighbor) ---> orchagent ---> mark it as already programmed. 

when orchagent reads the local app_db neighbor object, it will map it existing remote neighbor because its already programmed and this way its sync's remote neighbor entries faster.  In case of orchagent slow in reading chassis_app_db, it will mark local neighbor entry as unresolved and wait for remote neighbor entry with encap id to appear to program the neighbor. 

kernel programming is make to iBGP routes to be appear with actual nexthop. 

We wanted clarify the new approach here. Please let us know if your comments/views on this.

prsunny: Only one subscriber for chassis_app_db. Orchagent can read that for getting the encap_id. For chassis_app_db/app_db, there can be only one subscriber.

minionatwork: Do you mean, 

OrchAgent is a consumer of chassis_app_db
NbrMgr is a subscriber of chassis_app_db.

vganesan-nokia: Guohan's proposal: Let the orchagent program SAI for remote neighbor independently. Set state in state DB or local APP DB with mac address an additional attribute.. Nbrmgr programs kernel from the STATE DB/App DB entry only after making sure that SAI is programmed for the remote neighbors.  Nbrmgr will not use the chassis_app_db.

orchagent/portsorch.cpp
```diff
@@ -380,7 +384,7 @@ PortsOrch::PortsOrch(DBConnector *db, vector<table_name_with_pri_t> &tableNames)
 void PortsOrch::removeDefaultVlanMembers()
 {
     /* Get VLAN members in default VLAN */
-    vector<sai_object_id_t> vlan_member_list(m_portCount);
+    vector<sai_object_id_t> vlan_member_list(m_portCount + m_systemPortCount);
```
vganesan-nokia, line 387: Explore possibility of having m_portCount including the System port count.

orchagent/neighorch.h
vganesan-nokia: Fixed. Implemented the proposal with STATE_DB to have the entries to indicate SAI program completion. Orchagent will not program the kernel. Orchagent programs only SAI.

orchagent/main.cpp
```diff
@@ -140,6 +145,121 @@ void init_gearbox_phys(DBConnector *applDb)
...
+    Table cfgDeviceMetaDataTable(&cfgDb, CFG_DEVICE_METADATA_TABLE_NAME);
+    Table cfgSystemPortTable(&cfgDb, CFG_SYSTEM_PORT_TABLE_NAME);
+    Table appSystemPortTable(&appDb, APP_SYSTEM_PORT_TABLE_NAME);
+
+    //At this point of time (i.e, at the time of orchagent start), the CONFIG_DB is completely fully
+    //populated with the contents of config_db.json.
+
+    string value;
+    //Get the swith type.
+    if(!cfgDeviceMetaDataTable.hget("localhost", "switch_type", value))
```
ngoc-do, line 162: It seems not have consistent format with Sonic/other changes in the same diff --> space between "if" and "(!cfgDeviceMetaDataTable...)

if (!cfgDeviceMetaDataTable.hget("localhost", "switch_type", value))

orchagent/main.cpp
```diff
@@ -286,6 +406,35 @@ int main(int argc, char **argv)
         attrs.push_back(attr);
     }
 
+    //Get info required for VOQ system
+    vector<sai_system_port_config_t> sysportconfiglist;
+    if(getSystemPortConfigList(sysportconfiglist))
+    {
+        if (gMySwitchType == "voq")
```
ngoc-do, line 408: It's the right format.

orchagent/main.cpp
```diff
@@ -140,6 +145,121 @@ void init_gearbox_phys(DBConnector *applDb)
...
+        return false;
+    }
+
+    vector<string> spKeys;
+    cfgSystemPortTable.getKeys(spKeys);
+
+    //Retrieve system port configurations
+    vector<FieldValueTuple> spFv;
+    sai_system_port_config_t sysport;
+    for ( auto &k : spKeys )
```
ngoc-do, line 219: Fix the format:

for (auto &k : spKeys)
{
}

orchagent/main.cpp
```diff
@@ -140,6 +145,121 @@ void init_gearbox_phys(DBConnector *applDb)
...
+            if(fv.first == "speed")
+            {
+                sysport.speed = stoi(fv.second);
+                continue;
+            }
+            if(fv.first == "system_port_id")
+            {
+                sysport.port_id = stoi(fv.second);
+                continue;
+            }
```
ysmanman, line 249: Why don't we check sysport num_voq setting in SystemPortTable?

orchagent/main.cpp
ngoc-do, line 162: Could you move the code lines that get switch_type out of function getSystemPortConfigList()? Inside getSystemPortConfigList(), we could expect gMySwitchType must be "voq".

It is because if gMySwitchType is "fabric", we would do something else. 

Moving these code lines out of this function will help avoiding reading CONFIG_DB multiple times to identify asic type.

So, you could define DB connectors out of this function too, and have them as getSystemPortConfigList arguments.

orchagent/main.cpp
```diff
@@ -286,6 +406,35 @@ int main(int argc, char **argv)
         attrs.push_back(attr);
     }
 
+    //Get info required for VOQ system
+    vector<sai_system_port_config_t> sysportconfiglist;
+    if(getSystemPortConfigList(sysportconfiglist))
```
ngoc-do, line 406: I saw

    /* Must be last Attribute */
    attr.id = SAI_REDIS_SWITCH_ATTR_CONTEXT;
    attr.value.u64 = gSwitchId;
    attrs.push_back(attr);

Wonder if we should call this before setting SAI_REDIS_SWITCH_ATTR_CONTEXT?

orchagent/main.cpp
vganesan-nokia: Yes. We'll fix this

orchagent/main.cpp
vganesan-nokia: o.k. Got it.

orchagent/main.cpp
vganesan-nokia: I can check for non-zero.
We can not check for a specific value since this value is optional and will be ignored by some vendor SAIs. I'll check the value for  0 and pass it to SAI as is.

orchagent/main.cpp
vganesan-nokia: We'll fix this.

orchagent/main.cpp
vganesan-nokia: Yes. We'll fix as suggested. 
The DB connection to config db, app db and state db happens at later point of time in this function. We'll move the db connection before getting the switch type and system port config list and pass the db connectors as arguments to get switch type and get sytem port config list functions.

orchagent/main.cpp
vganesan-nokia: Yes.  Already fixed in previous review comments fix commit 2ca2a00

orchagent/main.cpp

orchagent/main.cpp
vganesan-nokia: Fixed

orchagent/main.cpp

orchagent/main.cpp

orchagent/main.cpp
```diff
@@ -401,33 +405,44 @@ int main(int argc, char **argv)
...
     }
 
-    //Get info required for VOQ system
+    // Instantiate database connectors
+    DBConnector appl_db("APPL_DB", 0);
+    DBConnector config_db("CONFIG_DB", 0);
+    DBConnector state_db("STATE_DB", 0);
+
+    // Get Switch type
+    getCfgSwitchType(&config_db, gMySwitchType);
```
ngoc-do, line 414: Wonderful change. Thanks for working on it.

Could you move these lines up a little bit higher? Should be before 

    if (gMacAddress)
    {
        attr.id = SAI_SWITCH_ATTR_SRC_MAC_ADDRESS;
        memcpy(attr.value.mac, gMacAddress.getMac(), 6);
        attrs.push_back(attr);
    }

It's because not all asics have mac addr, or need to set mac addr. Your gMySwitchType would help to avoid "getting unnecessary mac addr".

vganesan-nokia: Yes. Thanks. Fixed as suggested.

orchagent/main.cpp
```diff
@@ -255,18 +379,21 @@ int main(int argc, char **argv)
     attr.value.ptr = (void *)on_switch_shutdown_request;
     attrs.push_back(attr);
 
+    // Instantiate database connectors
+    DBConnector appl_db("APPL_DB", 0);
+    DBConnector config_db("CONFIG_DB", 0);
+    DBConnector state_db("STATE_DB", 0);
+
+    // Get switch_type
+    getCfgSwitchType(&config_db, gMySwitchType);
```
eswaranb, line 388: Return value not handled?

orchagent/main.cpp
```diff
@@ -286,6 +413,43 @@ int main(int argc, char **argv)
...
+        attr.id = SAI_SWITCH_ATTR_SWITCH_ID;
+        attr.value.u32 = gVoqMySwitchId;
+        attrs.push_back(attr);
+
+        attr.id = SAI_SWITCH_ATTR_MAX_SYSTEM_CORES;
+        attr.value.u32 = gVoqMaxCores;
+        attrs.push_back(attr);
+
+        gCfgSystemPorts = (uint32_t) sysportconfiglist.size();
+        if (gCfgSystemPorts)
```
eswaranb, line 435: assert that this is > 0  ?

orchagent/portsorch.cpp
```diff
@@ -4496,3 +4502,269 @@ bool PortsOrch::initGearboxPort(Port &port)
...
+
+            SWSS_LOG_NOTICE("Added system port %" PRIx64 " for %s", system_port_oid, alias.c_str());
+        }
+        else
+        {
+            //System port does not exist in the switch
+            //This can not happen since all the system ports are supposed to be created during switch creation itself
+
+            SWSS_LOG_NOTICE("System port %s does not exist in switch. Port not added!", alias.c_str());
+            continue;
```
eswaranb, line 4715: should this be a stronger error?

orchagent/neighorch.cpp
```diff
@@ -584,5 +666,289 @@ bool NeighOrch::removeNeighbor(const NeighborEntry &neighborEntry)
...
+                    it = consumer.m_toSync.erase(it);
+                }
+                else
+                {
+                    SWSS_LOG_ERROR("Failed to add voq neighbor %s to SAI", kfvKey(t).c_str());
+                    it++;
+                }
+            }
+            else
+                /* Duplicate entry */
```
eswaranb, line 786: LOG?

eswaranb (dismissed):

orchagent/main.cpp
vganesan-nokia: Yes. But instead of handling error outside, I'll make this function "void" and inside the function for any value other than "voq" or "fabric" gMySwitchType will be set "switch" (to indicate regular switch).

orchagent/main.cpp
vganesan-nokia: Yes. We'll fix this

orchagent/portsorch.cpp
vganesan-nokia: Yes. I'll change this to SWSS_LOG_ERROR

orchagent/neighorch.cpp
vganesan-nokia: Yes. I'll add SWSS_LOG_INFO here.

orchagent/main.cpp
vganesan-nokia: Fixed

orchagent/main.cpp
vganesan-nokia: Fixed. Instead of using assert, I log error and exit failure

orchagent/portsorch.cpp
vganesan-nokia: Fixed

orchagent/neighorch.cpp
vganesan-nokia: Fixed

lguohan:
retest this please

minionatwork:
retest vs please

minionatwork:
retest LGTM analysis: C/C++ please

lguohan:
@minionatwork , you need to do force push to trigger lgtm analysis

vganesan-nokia:
LGTM

minionatwork:
@lguohan : can you please take a look at this?.

eswaranb (approved):

lguohan:
i think this pr lacks for swss vs tests. this is a new feature needs to add swss vs test to cover the new code added.

lguohan (changes requested):
need to add swss vs tests

orchagent/main.cpp
```diff
@@ -158,10 +158,9 @@ bool getCfgSwitchType(DBConnector *cfgDb, string &switch_type)
     if (switch_type != "voq" && switch_type != "fabric" && switch_type != "switch")
     {
         cout << "Invalid switch type " << switch_type.c_str() << " configured";
```
ngoc-do, line 160: Where does the output of `cout ` go? To /var/log/syslog?

vganesan-nokia: This goes to /var/log/syslog since we call this after the logger system initialized. I'll change it  SWSS_LOG_ERROR() to make it clear.

vganesan-nokia:
> need to add swss vs tests

I"ll add swss vs tests

eswaranb:
> need to add swss vs tests

As we are absorbing this PR, we are adding vslib changes and some tests. @ysmanman is working on this on the Arista side. I wonder how we can avoid duplicating our work here. 

Is it possible to add this code and then follow-up with the test changes?

minionatwork:
> > need to add swss vs tests
> 
> As we are absorbing this PR, we are adding vslib changes and some tests. @ysmanman is working on this on the Arista side. I wonder how we can avoid duplicating our work here.
> 
> Is it possible to add this code and then follow-up with the test changes?

@eswaranb : Is it possible to share fork from this PR and start adding vs-tests?.

vganesan-nokia:
Added VS tests for VOQ system ports by commit 34e9f68. This commit is dependent on sonic-sairedis PR: https://github.com/Azure/sonic-sairedis/pull/696 and sonic-buildimage PR: VS test for VOQ System ports #5862.

saravanansv:
are these tests are still failing with latest diff
test_virtual_chassis.TestVirtualChassis.test_voq_switch
test_virtual_chassis.TestVirtualChassis.test_chassis_app_db_sync
test_virtual_chassis.TestVirtualChassis.test_chassis_system_interface

vganesan-nokia:
> are these tests are still failing with latest diff
> test_virtual_chassis.TestVirtualChassis.test_voq_switch
> test_virtual_chassis.TestVirtualChassis.test_chassis_app_db_sync
> test_virtual_chassis.TestVirtualChassis.test_chassis_system_interface

We need https://github.com/Azure/sonic-buildimage/pull/5862 to be merged first to have these tests pass.

tests/test_virtual_chassis.py
```diff
@@ -24,3 +26,96 @@ def test_connectivity(self, vct):
                 _, out = dv.runcmd(['sh', "-c", "ping -c 5 -W 0 -q %s" % ip])
                 print(out)
                 assert '5 received' in out
+
+    def test_voq_switch(self, vct):
+        """ Test VOQ switch objects configuration """
+        dvss = vct.dvss
+        for name in dvss.keys():
+            if not name.startswith("supervisor"):
```
ysmanman, line 34: Do we need to skip fabric switch as well?

tests/test_virtual_chassis.py
```diff
@@ -24,3 +26,96 @@ def test_connectivity(self, vct):
...
+                status, value = tbl.hget(switch_oid_key, "SAI_SWITCH_ATTR_MAX_SYSTEM_CORES")
+                assert status, "Got error while getting max system cores"
+                assert value == cfg_max_cores, "Max system cores is invalid"
+                
+                status, value = tbl.hget(switch_oid_key, "SAI_SWITCH_ATTR_SYSTEM_PORT_CONFIG_LIST")
+                assert status, "Got error while getting system port config list"
+                #Convert the spcfg string to dictionary
+                spcfg = ast.literal_eval(value)
+                assert spcfg['count'] == sp_count, "Number of systems ports configured is invalid"
+
```
ysmanman, line 73: Any reason we don't check the config of individual system port?

tests/test_virtual_chassis.py
```diff
@@ -24,3 +26,96 @@ def test_connectivity(self, vct):
...
+    def test_chassis_app_db_sync(self, vct):
+        """ Test chassis app db syncing """
+        dvss = vct.dvss
+        for name in dvss.keys():
+            if name.startswith("supervisor"):
+                dvs = dvss[name]
+                chassis_app_db = swsscommon.DBConnector(swsscommon.CHASSIS_APP_DB, dvs.redis_chassis_sock, 0)
+                tbl = swsscommon.Table(chassis_app_db, "SYSTEM_INTERFACE")
+                keys = list(tbl.getKeys())
+                assert len(keys), "No chassis app db syncing is done"
```
ysmanman, line 83: Should we validate that the number of system intfs is the same as system ports?

tests/test_virtual_chassis.py
vganesan-nokia: Yes. These configs are available only in line cards config db

tests/test_virtual_chassis.py
vganesan-nokia: No specific reason. To save testing time. I thought just reading and verifying a field in the system port config list will be enough to verify successful config and retrieval of system port config list. Since the VS SAI emulation does not do any real functionality we just want to verify config and retrieval. Since system port config list is one single attribute, I thought checking any field in the attribute structure will be good enough.
If it will be good to check all ports in the list, we can do that. Please let me know

tests/test_virtual_chassis.py
vganesan-nokia: Objective of this test is to verify successful syncing from line cards to supervisor card.  I just picked up one table to verity the syncing. It could be SYSTEM_NEIGH too. The next test "test_chassis_system_interface" does the actual verification of a sample system interface correctness.

We can validate the number of system interfaces, if the topology is finalized and we know exactly how many interfaces are expected to be synced.

tests/test_virtual_chassis.py
ysmanman: It seems the system port info is defined in default_config.json, instead of being generated from minigraph, in vs. Then I guess we don't really have too much to validate here. What you have should fine.

tests/test_virtual_chassis.py
ysmanman: Can we check switch_type, instead of name, to skip supervisor and fabric switch?

vganesan-nokia: Yes. I'll fix it.

vganesan-nokia: Fixed

orchagent/main.cpp

ysmanman (dismissed):
The latest revision of  VOQ test looks good to me.

lguohan:
can you point out the global neighbor table test added? I cannot find it somehow.

lguohan:
retest this please

vganesan-nokia:
> can you point out the global neighbor table test added? I cannot find it somehow.

I have not added neighbor test case yet since the existing virtual chassis test setup does not have front panel neighbor yet. I can create a static neighbor on a front panel port and test for syncing and records in ASIC DB.

lguohan:
> I can create a static neighbor on a front panel port and test for syncing and records in ASIC DB.

@vganesan-nokia that will be good.

vganesan-nokia:
retest vs please

vganesan-nokia:
> 
> 
> > I can create a static neighbor on a front panel port and test for syncing and records in ASIC DB.
> 
> @vganesan-nokia that will be good.

Added vs test case (commit https://github.com/Azure/sonic-swss/pull/1431/commits/233a1bedc8d6a5177cf4050d658551c47d52cda0) for testing neighbor programming with encap index and syncing to chassis app db. This requires sairedis PR https://github.com/Azure/sonic-sairedis/pull/725. Please note that testing for **remote** neigh programming is TODO since remote neigh processing requires inband interface setup which is not available in this PR #1431. This test will be enhanced in the  inband interface implementation PR.

vganesan-nokia:
> Can we use dvs.add_neighbor() ?

The thinking behind adding neighbor directly to kernel is to simulate a situation of a host attached to a front panel port. This is close to the setup of an eBGP neighbor attached to the front panel port. dvs.add_neighbor() injects the neighbor entry directly to the app db, which is couple of steps later in the host learning process. I think it is better have a neighbor statically configured directly to the kernel so that we can test local neighbor processing as close as to a real situation.

vganesan-nokia:
> > Can we use dvs.add_neighbor() ?
> 
> The thinking behind adding neighbor directly to kernel is to simulate a situation of a host attached to a front panel port. This is close to the setup of an eBGP neighbor attached to the front panel port. dvs.add_neighbor() injects the neighbor entry directly to the app db, which is couple of steps later in the neighbor add process. I think it is better have a neighbor statically configured directly to the kernel so that we can test local neighbor processing as close as to a real situation.

ysmanman (dismissed):

tests/test_virtual_chassis.py
```diff
@@ -24,3 +26,184 @@ def test_connectivity(self, vct):
...
                 assert '5 received' in out
+
+    def test_voq_switch(self, vct):
+        """ Test VOQ switch objects configuration """
+        dvss = vct.dvss
+        for name in dvss.keys():
+            dvs = dvss[name]
+            #Get the config info
+            config_db = swsscommon.DBConnector(swsscommon.CONFIG_DB, dvs.redis_sock, 0)
+            metatbl = swsscommon.Table(config_db, "DEVICE_METADATA")
```
daall, line 37: Would recommend using `dvslib/dvs_database.py` over the raw swsscommon connectors. There are some examples in `test_acl`, `test_vlan`, and `test_subport_intf`, although it doesn't look like there is any waiting/sleeping here so you could just use the raw `get_entry`/`get_keys` methods.

vganesan-nokia: Changed as recommended. Thanks

tests/test_virtual_chassis.py
```diff
@@ -24,3 +27,184 @@ def test_connectivity(self, vct):
...
                 print(out)
                 assert '5 received' in out
+
+    def test_voq_switch(self, vct):
+        """ Test VOQ switch objects configuration """
+        dvss = vct.dvss
+        for name in dvss.keys():
+            dvs = dvss[name]
+            #Get the config info
+            config_db = DVSDatabase(swsscommon.CONFIG_DB, dvs.redis_sock)
```
daall, line 37: You should be able to use the `dvs.get_config_db()` method instead here.

tests/test_virtual_chassis.py
```diff
@@ -24,3 +27,184 @@ def test_connectivity(self, vct):
...
+                cfg_switch_id = metatbl.get("switch_id")
+                assert cfg_switch_id != "", "Got error in getting switch_id from CONFIG_DB DEVICE_METADATA"
+
+                cfg_max_cores = metatbl.get("max_cores")
+                assert cfg_max_cores != "", "Got error in getting max_cores from CONFIG_DB DEVICE_METADATA"
+                
+                cfgspkeys = config_db.get_keys("SYSTEM_PORT")
+                sp_count = len(cfgspkeys)
+
+                asic_db = DVSDatabase(swsscommon.ASIC_DB, dvs.redis_sock)
```
daall, line 54: Same as above.

tests/test_virtual_chassis.py
```diff
@@ -24,3 +27,184 @@ def test_connectivity(self, vct):
...
+                spcfg = ast.literal_eval(value)
+                assert spcfg['count'] == sp_count, "Number of systems ports configured is invalid"
+    
+    def test_chassis_app_db_sync(self, vct):
+        """ Test chassis app db syncing """
+        dvss = vct.dvss
+        for name in dvss.keys():
+            if name.startswith("supervisor"):
+                dvs = dvss[name]
+                chassis_app_db = DVSDatabase(swsscommon.CHASSIS_APP_DB, dvs.redis_chassis_sock)
```
daall, line 81: I am not sure how best to handle this, but we could probably add a similar `get_chassis_app_db` handler to DVS if it is a supervisor module. In the interest of unblocking other chassis PRs, I think directly initializing the DB here is OK and we can refine the initialization/getter part later.

tests/test_virtual_chassis.py
```diff
@@ -24,3 +27,184 @@ def test_connectivity(self, vct):
                 _, out = dv.runcmd(['sh', "-c", "ping -c 5 -W 0 -q %s" % ip])
                 print(out)
                 assert '5 received' in out
+
+    def test_voq_switch(self, vct):
+        """ Test VOQ switch objects configuration """
+        dvss = vct.dvss
+        for name in dvss.keys():
+            dvs = dvss[name]
+            #Get the config info
```
daall, line 36: one space b/w `#` and the comment (here and throughout the rest of the python code too)

tests/test_virtual_chassis.py
```diff
@@ -24,3 +27,184 @@ def test_connectivity(self, vct):
                 _, out = dv.runcmd(['sh', "-c", "ping -c 5 -W 0 -q %s" % ip])
                 print(out)
                 assert '5 received' in out
+
+    def test_voq_switch(self, vct):
+        """ Test VOQ switch objects configuration """
```
daall, line 32: Two things:

1. I would remove the extra spaces around the comment, and
2. Could you give a brief summary of what object configuration we're checking in this method?

So, the doc comment should end up looking like this:
```suggestion
        """Test VOQ switch objects configuration.
        
        Insert brief summary here.
        """
```

tests/test_virtual_chassis.py
```diff
@@ -24,3 +27,184 @@ def test_connectivity(self, vct):
...
+        dvss = vct.dvss
+        for name in dvss.keys():
+            if name.startswith("supervisor"):
+                dvs = dvss[name]
+                chassis_app_db = DVSDatabase(swsscommon.CHASSIS_APP_DB, dvs.redis_chassis_sock)
+                keys = chassis_app_db.get_keys("SYSTEM_INTERFACE")
+                assert len(keys), "No chassis app db syncing is done"
+                
+    def test_chassis_system_interface(self, vct):
+        """ Test RIF record creation in ASIC_DB for remote interfaces """
```
daall, line 86: Same feedback as above.

tests/test_virtual_chassis.py
```diff
@@ -24,3 +27,184 @@ def test_connectivity(self, vct):
...
+
+    def test_chassis_system_neigh(self, vct):
+        """ Test neigh record creation and syncing to chassis app db """
+        
+        """
+        This test validates that:
+           (i)   Local neighbor entry is created with encap index
+           (ii)  Local neighbor is synced to chassis ap db with assigned encap index
+           TODO: (iii) Remote neighbor entry is created in ASIC_DB with received encap index
+        """
```
daall, line 138: Same feedback as above, but in this case you can just combine this two comments like so:
```suggestion
        """Test neigh record creation and syncing to chassis app db.
        
       This test validates that:
           (i)   Local neighbor entry is created with encap index
           (ii)  Local neighbor is synced to chassis ap db with assigned encap index
           TODO: (iii) Remote neighbor entry is created in ASIC_DB with received encap index
        """
```

daall (commented):
Had a few more minor formatting suggestions for the test

tests/test_virtual_chassis.py
vganesan-nokia: Yes. Changed as suggested

tests/test_virtual_chassis.py
vganesan-nokia: Fixed.

tests/test_virtual_chassis.py
vganesan-nokia: O.k. Agreed. Thanks.

tests/test_virtual_chassis.py
vganesan-nokia: Fixed here and throughout the file

tests/test_virtual_chassis.py
vganesan-nokia: Fixed as suggested.

tests/test_virtual_chassis.py
vganesan-nokia: Fixed

tests/test_virtual_chassis.py
vganesan-nokia: Fixed. Thanks.

daall (dismissed):
VS tests LGTM, thanks @vganesan-nokia! @abdosi and @gechiang to review the rest of the changes

tests/test_virtual_chassis.py
```diff
@@ -24,3 +26,96 @@ def test_connectivity(self, vct):
...
+                keys = list(tbl.getKeys())
+                switch_oid_key = keys[0]
+
+                status, value = tbl.hget(switch_oid_key, "SAI_SWITCH_ATTR_TYPE")
+                assert status, "Got error while getting switch type"
+                assert value == "SAI_SWITCH_TYPE_VOQ", "Switch type is not VOQ"
+                
+                status, value = tbl.hget(switch_oid_key, "SAI_SWITCH_ATTR_SWITCH_ID")
+                assert status, "Got error while getting switch id"
+                assert value == cfg_switch_id, "VOQ switch id is invalid"
```
ngoc-do: There is nothing wrong with this code. I just wanted to ask since I couldn't find the related code in this PR or in the repo. Probably it's in another PR.

Where do we specify `switch_type` which is `SAI_SWITCH_TYPE_VOQ` for the test? I assume it should be some where in

tests/virtual_chassis/1/default_config.json
{
    "DEVICE_METADATA": {
        "localhost": {
            "hostname" : "lc1",
            "instance_name": "Linecard1",
            "connect_to_chassis_db" : 1,
            "chassis_db_address" : "10.8.1.200",
            "inband_address" : "10.8.1.1/24",
            "switch_type" : "voq"
        }
    },

And where does the code read this file and parse `switch_type`?

Thanks.

ngoc-do: No, nevermind. I found the answers.

vganesan-nokia:
retest vs please

vganesan-nokia:
retest vs please

vganesan-nokia:
retest vs please

daall:
@abdosi @gechiang please help review so we can get this merged and get the dependent PRs unblocked

daall:
@lguohan pinging you too b/c changes were requested earlier

vganesan-nokia:
retest vs please

vganesan-nokia:
retest vs please

vganesan-nokia:
retest vs please

vganesan-nokia:
retest vs please

orchagent/portsorch.cpp
```diff
@@ -4675,3 +4681,269 @@ bool PortsOrch::initGearboxPort(Port &port)
...
+        if(system_port_id < 0 || switch_id < 0 || core_index < 0 || core_port_index < 0)
+        {
+            SWSS_LOG_ERROR("Invalid or Missing field values for %s! system_port id:%d, switch_id:%d, core_index:%d, core_port_index:%d",
+                    alias.c_str(), system_port_id, switch_id, core_index, core_port_index);
+            continue;
+        }
+
+        if(speed < 0)
+        {
+            speed = 400000;
```
abdosi, line 4812: what does this default value means here ?

orchagent/intfsorch.cpp
```diff
@@ -1009,6 +1067,12 @@ bool IntfsOrch::addRouterIntfs(sai_object_id_t vrf_id, Port &port)
 
     SWSS_LOG_NOTICE("Create router interface %s MTU %u", port.m_alias.c_str(), port.m_mtu);
 
+    if(gMySwitchType == "voq")
+    {
+        //Sync the interface to add to the SYSTEM_INTERFACE table of CHASSIS_APP_DB
```
abdosi, line 1072: is it true port.m_type == Port::SYSTEM is the only case this call is needed ?

cfgmgr/nbrmgr.cpp
```diff
@@ -64,6 +65,20 @@ NbrMgr::NbrMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, con
...
+    string swtype;
+    Table cfgDeviceMetaDataTable(cfgDb, CFG_DEVICE_METADATA_TABLE_NAME);
+    if(cfgDeviceMetaDataTable.hget("localhost", "switch_type", swtype))
+    {
+        //If this is voq system, let the neighbor manager subscribe to state of SYSTEM_NEIGH
+        //entries. This is used to program static neigh and static route in kernel for remote neighbors.
+        if(swtype == "voq")
+        {
+            string tableName = STATE_SYSTEM_NEIGH_TABLE_NAME;
+            Orch::addExecutor(new Consumer(new SubscriberStateTable(stateDb, tableName, TableConsumable::DEFAULT_POP_BATCH_SIZE, 0), this, tableName));
```
abdosi, line 78: why we are not calling Orch::addConsumer() here ? We don;t have to repeat this code

cfgmgr/nbrmgr.cpp
```diff
@@ -64,6 +65,20 @@ NbrMgr::NbrMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, con
...
+          
+    string swtype;
+    Table cfgDeviceMetaDataTable(cfgDb, CFG_DEVICE_METADATA_TABLE_NAME);
+    if(cfgDeviceMetaDataTable.hget("localhost", "switch_type", swtype))
+    {
+        //If this is voq system, let the neighbor manager subscribe to state of SYSTEM_NEIGH
+        //entries. This is used to program static neigh and static route in kernel for remote neighbors.
+        if(swtype == "voq")
+        {
+            string tableName = STATE_SYSTEM_NEIGH_TABLE_NAME;
```
abdosi, line 77: where we are populating STATE_SYSTEM_NEIGH_TABLE_NAME ? Is it in another PR ?

orchagent/portsorch.cpp
vganesan-nokia: If the speed attribute was not configured, 400 G was assumed as default. However, now, this is not needed anymore since SAI sets default value when speed is not supplied in system port config info (during initialization). In this function we just read and update the local data structure for information purpose. I'll remove this default handling

orchagent/intfsorch.cpp
vganesan-nokia: We need to sync the RIF of local port and local LAG to chassis app db. So syncing will be applicable for both PHY and LAG.

cfgmgr/nbrmgr.cpp
vganesan-nokia: Yes. We can call Orch::addConsumer() here. I'll fix this.

cfgmgr/nbrmgr.cpp
vganesan-nokia: We are populating this table in neighorch.cpp (doVoqSystemNeighTask()) in this PR #1431 only. For the remote neighbors, we need to add static neighbor and static route in the kernel. Since it was recommended not to add the kernel entries from orchagent, we program kernel entries from the nbrmgr (as per @prsunny review comment). In neighorch, after programming the remote neighbor in SAI, we add an entry in state db in STATE_SYSTEM_NEIGH_TABLE to trigger programming kernel entries here.

orchagent/intfsorch.cpp
abdosi: can we update the comment in that case ?

vganesan-nokia: Yes. I'll update the comment.

abdosi (changes requested):
I have reviewed the changes. There are some comments. Once those are addressed i will approve the PR

orchagent/portsorch.cpp
vganesan-nokia: Fixed

orchagent/intfsorch.cpp
vganesan-nokia: Updated the comment as suggested

vganesan-nokia:
retest vs please

abdosi (dismissed):
LGTM. Wait for others approval/comments

vganesan-nokia:
> 
> 
> LGTM. Wait for others approval/comments

o.k. Thanks

daall (approved):

daall:
@prsunny @lguohan please check on the changes you requested and approve if they've been addressed

orchagent/neighorch.cpp
```diff
@@ -865,3 +952,284 @@ bool NeighOrch::removeTunnelNextHop(const NextHopKey& nh)
...
+        {
+            SWSS_LOG_ERROR("Unknown operation type %s", op.c_str());
+            it = consumer.m_toSync.erase(it);
+        }
+    }
+}
+
+bool NeighOrch::addInbandNeighbor(string alias, IpAddress ip_address)
+{
+    //For "port" type inband, the inband reachability info syncing can be done through static
```
prsunny, line 1102: Can we remove these functions if not used?

cfgmgr/nbrmgr.cpp
```diff
@@ -294,8 +309,253 @@ void NbrMgr::doTask(Consumer &consumer)
...
+        /* Just log error and return */
+        SWSS_LOG_ERROR("Failed to delete route for %s, error: %d", ip_str.c_str(), ret);
+        return false;
+    }
+
+    SWSS_LOG_INFO("Deleted route for %s", ip_str.c_str());
+    return true;
+}
+
+bool NbrMgr::addKernelNeigh(string odev, IpAddress ip_addr, MacAddress mac_addr)
```
prsunny, line 499: Just curious, why don't we use `setNeighbor ` instead?

prsunny (dismissed):
nbrmgr changes lgtm..  minor comments.

orchagent/neighorch.cpp
vganesan-nokia: This function and delInbandNeighbor() are place holders for forthcoming Inband PRs (port type by Nokia and vlan type by Arista). These dummy functions are there with some brief comments as guidance/remainder for filling these functions, when Inband interface functionality is implemented.

cfgmgr/nbrmgr.cpp
vganesan-nokia: There is no specific reason why setNeighbor() is not used. While looking for kernel entry programming, I felt using shell commands to add/del kernel entries seemed simple. So I added this function and other functions for delete neigh, static route add and static route delete functions that use the shell commands for kernel entry programming.

vganesan-nokia:
retest vs please

vganesan-nokia:
@daall, There are two vs test failures in test_virtual_chassis.py due to  **swsscommon.swsscommon has not attribute CHASSIS_APP_DB** error. The previous build https://sonic-jenkins.westus2.cloudapp.azure.com/job/vs/job/sonic-swss-build-pr/2615/  of this PR did not show this error and passed all tests. It seems jenkins test is not using correct swsscommon.py. Is there anything that need to be done in vs test to address this?

daall:
> @daall, There are two vs test failures in test_virtual_chassis.py due to **swsscommon.swsscommon has no attribute CHASSIS_APP_DB** error. The previous build https://sonic-jenkins.westus2.cloudapp.azure.com/job/vs/job/sonic-swss-build-pr/2615/ of this PR did not show this error and passed all tests. It seems jenkins test is not using correct swsscommon.py. Is there anything that need to be done in vs test to address this?

There seems to be some inconsistency between the swsscommon version installed on the PR runners, let me look into it and I will trigger a retry once it's resolved

vganesan-nokia:
> 
> 
> > @daall, There are two vs test failures in test_virtual_chassis.py due to **swsscommon.swsscommon has no attribute CHASSIS_APP_DB** error. The previous build https://sonic-jenkins.westus2.cloudapp.azure.com/job/vs/job/sonic-swss-build-pr/2615/ of this PR did not show this error and passed all tests. It seems jenkins test is not using correct swsscommon.py. Is there anything that need to be done in vs test to address this?
> 
> There seems to be some inconsistency between the swsscommon version installed on the PR runners, let me look into it and I will trigger a retry once it's resolved

o.k. Thanks.

daall:
retest vs please

daall:
> > > @daall, There are two vs test failures in test_virtual_chassis.py due to **swsscommon.swsscommon has no attribute CHASSIS_APP_DB** error. The previous build https://sonic-jenkins.westus2.cloudapp.azure.com/job/vs/job/sonic-swss-build-pr/2615/ of this PR did not show this error and passed all tests. It seems jenkins test is not using correct swsscommon.py. Is there anything that need to be done in vs test to address this?
> > 
> > 
> > There seems to be some inconsistency between the swsscommon version installed on the PR runners, let me look into it and I will trigger a retry once it's resolved
> 
> o.k. Thanks.

It looks like `dpkg` got deadlocked on one of the nodes, it should be fixed now

vganesan-nokia:
> 
> 
> > > > @daall, There are two vs test failures in test_virtual_chassis.py due to **swsscommon.swsscommon has no attribute CHASSIS_APP_DB** error. The previous build https://sonic-jenkins.westus2.cloudapp.azure.com/job/vs/job/sonic-swss-build-pr/2615/ of this PR did not show this error and passed all tests. It seems jenkins test is not using correct swsscommon.py. Is there anything that need to be done in vs test to address this?
> > > 
> > > 
> > > There seems to be some inconsistency between the swsscommon version installed on the PR runners, let me look into it and I will trigger a retry once it's resolved
> > 
> > 
> > o.k. Thanks.
> 
> It looks like `dpkg` got deadlocked on one of the nodes, it should be fixed now

Great! Thanks for the help.

abdosi (dismissed):

abdosi:
@lguohan Can you review this again

minionatwork:
@eswaranb Can you please take a look ?

lguohan (approved):

gechiang (approved):

eswaranb (approved):

ysmanman (approved):

ngoc-do (dismissed):

lguohan:
image

lguohan:
@vganesan-nokia , it looks like there were a few checker pending, not sure why. can you do a force repush? i have captured the sign-off status. we do not need everybody sign-off for you repush.

vganesan-nokia:
> @vganesan-nokia , it looks like there were a few checker pending, not sure why. can you do a force repush? i have captured the sign-off status. we do not need everybody sign-off for you repush.

@lguohan, just finished issuing "git push --force". Thanks

rlhui:
retest this please

vganesan-nokia:
retest this please

vganesan-nokia:
vs test fails because of sonic-sairedis PR https://github.com/Azure/sonic-sairedis/pull/737, which introduced the fabric port creation in voq switch create. It seems we create fabric port for both NPU asic and fabric asic. But in this PR #1431 we do not have support to supply fabric port lane mapping in line cards in vs tests (test_virtual_chassis.py). Without fabriclanemap.ini info switch create fails and orchagent crashes. 

To fix this: 
    (1) We need the sonic-buildimage PR https://github.com/Azure/sonic-buildimage/pull/6185 and some changes in sonic-swss/tests to get fabric port lane mapping file or 
    (2) Temporarily comment out fabric port creation code for now to enable merging PR #1431 

@lguohan, @ngoc-do, @eswaranb any other suggestions to address this?

ngoc-do:
> vs test fails because of sonic-sairedis PR [Azure/sonic-sairedis#737](https://github.com/Azure/sonic-sairedis/pull/737), which introduced the fabric port creation in voq switch create. It seems we create fabric port for both NPU asic and fabric asic. But in this PR #1431 we do not have support to supply fabric port lane mapping in line cards in vs tests (test_virtual_chassis.py). Without fabriclanemap.ini info switch create fails and orchagent crashes.
> 
> To fix this:
> (1) We need the sonic-buildimage PR [Azure/sonic-buildimage#6185](https://github.com/Azure/sonic-buildimage/pull/6185) and some changes in sonic-swss/tests to get fabric port lane mapping file or
> (2) Temporarily comment out fabric port creation code for now to enable merging PR #1431
> 
> @lguohan, @ngoc-do, @eswaranb any other suggestions to address this?

Oh, that's right. Sorry, https://github.com/Azure/sonic-sairedis/pull/737 is blocking you now.

Yes, to fix it, seems option 2 is faster for you. I made this https://github.com/Azure/sonic-sairedis/pull/769. Hope it will remove the blocker. Of course, we won't merge https://github.com/Azure/sonic-buildimage/pull/6185 until this PR merges.

rlhui:
retest vs please

vganesan-nokia:
retest vs please

vganesan-nokia:
> 
> 
> > vs test fails because of sonic-sairedis PR [Azure/sonic-sairedis#737](https://github.com/Azure/sonic-sairedis/pull/737), which introduced the fabric port creation in voq switch create. It seems we create fabric port for both NPU asic and fabric asic. But in this PR #1431 we do not have support to supply fabric port lane mapping in line cards in vs tests (test_virtual_chassis.py). Without fabriclanemap.ini info switch create fails and orchagent crashes.
> > To fix this:
> > (1) We need the sonic-buildimage PR [Azure/sonic-buildimage#6185](https://github.com/Azure/sonic-buildimage/pull/6185) and some changes in sonic-swss/tests to get fabric port lane mapping file or
> > (2) Temporarily comment out fabric port creation code for now to enable merging PR #1431
> > @lguohan, @ngoc-do, @eswaranb any other suggestions to address this?
> 
> Oh, that's right. Sorry, [Azure/sonic-sairedis#737](https://github.com/Azure/sonic-sairedis/pull/737) is blocking you now.
> 
> Yes, to fix it, seems option 2 is faster for you. I made this [Azure/sonic-sairedis#769](https://github.com/Azure/sonic-sairedis/pull/769). Hope it will remove the blocker. Of course, we won't merge [Azure/sonic-buildimage#6185](https://github.com/Azure/sonic-buildimage/pull/6185) until this PR merges.

Thanks for fixing

rlhui (approved):

ngoc-do (approved):

PR #1433, opened 2020-09-11
[acl] Remove Ethertype from L3V6 qualifiers

daall:

**What I did**
I removed ethertype from the qset for L3V6 tables.

**Why I did it**
This change was already done for MIRRORV6 tables a while back. ETHER_TYPE is unnecessary for V6 tables as the ethertype should always be the same (86DD) for IPv6 traffic. Additionally, this qualifier breaks IPv6 support on some platforms (e.g. Mellanox).

**How I verified it**
1. Update C++ unit tests
2. Test table behavior on BRCM and MLNX devices

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1036,6 +1036,12 @@ bool AclRuleL3V6::validateAddMatch(string attr_name, string attr_value)
         return false;
     }
 
+    if (attr_name == MATCH_ETHER_TYPE)
+    {
+        SWSS_LOG_ERROR("Ethertype match is not supported for table type L3V6");
+        return false;
+    }
+
     return AclRule::validateAddMatch(attr_name, attr_value);
 }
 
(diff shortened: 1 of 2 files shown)


lguohan:
what platform is broken?

orchagent/aclorch.cpp
```diff
@@ -1140,6 +1146,12 @@ bool AclRuleMirror::validateAddMatch(string attr_name, string attr_value)
         return false;
     }
 
+    if (m_tableType == ACL_TABLE_MIRRORV6 && attr_name == MATCH_ETHER_TYPE)
```
zhenggen-xu, line 1149: Isn't this is done by above if() check.  original file Line 1136.

daall: Good catch, thanks!

zhenggen-xu (approved):

lguohan (approved):

PR #1438, opened 2020-09-15
[dvs] Update test_nhg.py to make it pass in synchronous mode

shi-su:
**What I did**
Add a wait for route entries in ASIC_DB in `test_nhg.py::TestNextHopGroup::test_route_nhg_exhaust` so that the test case could pass in synchronous mode.

**Why I did it**
The test case `test_nhg.py::TestNextHopGroup::test_route_nhg_exhaust` fails in synchronous mode due to a delay between the creation of next-hop groups and routes in ASIC_DB. Getting the count of routes entries immediately after the creation of next-hop groups results in an inaccurate result. This PR fixes this by waiting for the number of route entries in ASIC_DB to be the expected value.

**How I verified it**
Verified by running the test locally.

**Details if related**

### DIFF
--- a/tests/test_nhg.py
+++ b/tests/test_nhg.py
@@ -234,11 +234,13 @@ class TestNextHopGroup(object):
             assert dvs.servers[i].runcmd("ip link set up dev eth0") == 0
 
         app_db = dvs.get_app_db()
+        asic_db = dvs.get_asic_db()
         ps = swsscommon.ProducerStateTable(app_db.db_connection, "ROUTE_TABLE")
 
         # Add first batch of routes with unique nexthop groups in AppDB
         route_count = 0
         r = 0
+        asic_routes_count = len(asic_db.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY"))
         while route_count < MAX_ECMP_COUNT:
             r += 1
             fmt = '{{0:0{}b}}'.format(MAX_PORT_COUNT)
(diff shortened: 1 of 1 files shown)


tests/test_nhg.py
```diff
@@ -251,11 +253,13 @@ def asic_route_nhg_fvs(k):
...
             route_count += 1
 
-        asic_db = dvs.get_asic_db()
-
         # Wait and check ASIC DB the count of nexthop groups used
         asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_NEXT_HOP_GROUP", MAX_ECMP_COUNT)
-        asic_routes_count = len(asic_db.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY"))
+
+        # Wait and check ASIC DB the count of routes
+        # Expect asic_routes_count + MAX_ECMP_COUNT - 1 since 2.2.2.0/24 is already set in the previous test case and counted in asic_routes_count
```
qiluo-msft, line 260: >since 2.2.2.0/24 is already set in the previous test case and counted in asic_routes_count [](start = 56, length = 90)

We should prevent inter-testcase dependency. With gtest, it is easy to filter specific testcase to run, so never assume the selection or sequence on testcases.

qiluo-msft (changes requested):
As comment

tests/test_nhg.py
shi-su: Thanks for the suggestion! Updated the previous test case to remove the inter-testcase dependency. Also refactored `test_route_nhg` test case to use dvslib.

qiluo-msft (approved):
LGTM

daall (approved):

PR #29, opened 2016-05-23
fix switch mac logic.

lguohan:
if user set the switch mac via -m, then set the switch mac to it.
Otherwise, get switch mac from SAI.

PR #1440, opened 2020-09-17
Fix issue #5157 by identifying the dependency among objects and avoiding releasing an object still being referenced

stephenxs:
**What I did**
The issue is caused by the orchagent receiving notification in a different order in which they were sent.
orchagent doesn't have any dependency check and try notifying sai-redis to release an object which is still being referenced,
which causes sai-redis to complain and the object leaks.

The idea is to introduce a mechanism to identify the dependency thus preventing a referenced object from being released.

**Why I did it**

**Details if related**
1. Introduce a new type representing the dependency among the variant type of objects, including the following fields:
    - m_objsDependingOnMe, a set representing the objects that reference the current object.
    - m_objsReferencingByMe, a map from a field of the current object's to the object name it references.

    eg. given the following dependencies, 

    - `BUFFER_PROFILE.ingress_lossless_profile` references `BUFFER_POOL.ingress_lossless_pool`,
    - `BUFFER_PROFILE.pg_lossless_25000_5m_profile` references `BUFFER_POOL.ingress_lossless_pool`
    the objects should be like this:
```
obj(BUFFER_POOL.ingress_lossless_pool).m_objsDependingOnMe = [
        BUFFER_PROFILE.ingress_lossless_profile,
        BUFFER_PROFILE.pg_lossless_25000_5m_profile
]
obj(BUFFER_PROFILE.pg_lossless_25000_5m_profile).m_objsReferencingByMe[pool] = BUFFER_POOL.ingress_lossless_pool
obj(BUFFER_PROFILE.ingress_lossless_profile).m_objsReferencingByMe[pool] = BUFFER_POOL.ingress_lossless_pool
```
2. When a field of an object `A`
    - is referencing `B` now
    - is going to be updated with referencing another object `B'`,

    We will have the following steps:

    - obj[`A.m_objsReferencingByMe[field name]`].m_objsDependingOnMe.remove(`A`)
      where obj[`A.m_objsReferencingByMe[field name]`] should be `B`
    - `A.m_objsReferencingByMe[field name]` = `B'`
    - `B'.m_objsDependingOnMe.add(A)`
3. When an object `A` is about to be removed,
    - if `A.m_objsDependingOnMe` isn't empty set, return `task_need_retry`
    - else execute the normal remove flow, especially if the `A` is referencing someone else:
      - For each field in `A.m_objsReferencingByMe`: obj[A.m_objsReferencingByMe[field]].m_objsDependingOnMe.remove(`A`)

***Reference count vs. set***
When it comes to recording the references, the reference count approach is the solution a lot of developers come up with.
A frequently asked question is why the set is used to record dependency instead of the reference count.
So I'd like to compare these two approaches.
- for reference count
  - pro
    - small memory footprint
    - able to trace a large number of objects depending on the current object
  - con
    - need an extra flag for each referencing objects, indicating whether the reference count has been increased/decreased for the object
    - misoperate reference count leads to fatal error
- for the set of dependent objects
  - pro
    - able to provide more information to the user regarding the dependency, which is very helpful when debugging an error
    - no extra flag required. as the insert/erase operations are idempotent, misoperation on reference count is less likely to lead to fatal error
  - con
    - heavy memory footprint

As we don't have a big number of objects to trace and want to provide more detailed info, I prefer the set approach.

**How I verified it**
1. remove objects with non-zero reference.
  to create a buffer pool, a buffer profile and two buffer pgs in order and then remove the profile and then the pool and at last the two PGs. in this case, the profile and pool can't be removed because of non-zero referencing.
    - create the buffer pool by using the command `hset BUFFER_POOL|ingress_test_pool type ingress mode dynamic size 14542848`. the log is like following:
        ```
        NOTICE swss#orchagent: :- processBufferPool: Created buffer pool ingress_test_pool with type BUFFER_POOL
        ```
    - create the buffer profile referencing the pool by using the command `hset BUFFER_PROFILE|ingress_test_profile dynamic_th 1 pool [BUFFER_POOL|ingress_test_pool] size 0`.
        ```
        NOTICE swss#orchagent: :- processBufferProfile: Created buffer profile ingress_test_profile with type BUFFER_PROFILE
        INFO swss#orchagent: :- setObjectReference: Obj BUFFER_PROFILE.ingress_test_profile Field pool: Add reference to BUFFER_POOL ingress_test_pool (now 1)
        ```
    - create a buffer PG referencing the profile by using the command `hset BUFFER_PG|Ethernet8|1 profile [BUFFER_PROFILE|ingress_test_profile]`.
        ```
        NOTICE swss#orchagent: :- processPriorityGroup: Set buffer PG Ethernet8|1 to BUFFER_PROFILE:ingress_test_profile
        INFO swss#orchagent: :- setObjectReference: Obj BUFFER_PG.Ethernet8|1 Field profile: Add reference to BUFFER_PROFILE ingress_test_profile (now 1)
        ```
    - create the other buffer PG by using `hset BUFFER_PG|Ethernet16|2 profile [BUFFER_PROFILE|ingress_test_profile]`
        ```
        NOTICE swss#orchagent: :- processPriorityGroup: Set buffer PG Ethernet16|2 to BUFFER_PROFILE:ingress_test_profile
        INFO swss#orchagent: :- setObjectReference: Obj BUFFER_PG.Ethernet16|2 Field profile: Add reference to BUFFER_PROFILE ingress_test_profile (now 2)
        ```
    - remove the buffer profile by using the command `hdel BUFFER_PROFILE|ingress_test_profile dynamic_th pool size`. It can't be removed because of dependency.
        ```
        NOTICE swss#orchagent: :- processBufferProfile: Can't remove object ingress_test_profile due to being referenced (BUFFER_PROFILE ingress_test_profile first object: Ethernet16|2 reference count: 2)
        INFO swss#orchagent: :- doTask: Failed to process buffer task, retry it
        ```
    - remove the buffer pool by using the command `hdel BUFFER_POOL|ingress_test_pool type mode size`. It can't be removed because of dependency.
        ```
        NOTICE swss#orchagent: :- processBufferProfile: Can't remove object ingress_test_profile due to being referenced (BUFFER_PROFILE ingress_test_profile first object: Ethernet16|2 reference count: 2)
        INFO swss#orchagent: :- doTask: Failed to process buffer task, retry it
        ```
    - remove one buffer PG by using the command `hdel BUFFER_PG|Ethernet8|1 profile`. the reference count of buffer profile is reduced to 1.
        ```
        NOTICE swss#orchagent: :- processPriorityGroup: Remove buffer PG Ethernet8|1
        INFO swss#orchagent: :- removeMeFromObjsReferencedByMe: Obj BUFFER_PG.Ethernet8|1 Field profile: Remove reference to BUFFER_PROFILE ingress_test_profile (now 1)
        INFO swss#orchagent: :- removeObject: Obj BUFFER_PG:Ethernet8|1 is removed from store
        ```
    - remove the other buffer PG by using `hdel BUFFER_PG|Ethernet16|2 profile`. Now the reference of the profile is decreased to 0 and then the pool. Eventually, all objects are removed.
        ```
        NOTICE swss#orchagent: :- processPriorityGroup: Remove buffer PG Ethernet16|2
        INFO swss#orchagent: :- removeMeFromObjsReferencedByMe: Obj BUFFER_PG.Ethernet16|2 Field profile: Remove reference to BUFFER_PROFILE ingress_test_profile (now 0)
        INFO swss#orchagent: :- removeObject: Obj BUFFER_PG:Ethernet16|2 is removed from store
        ERR swss#orchagent: :- processPriorityGroup: PG profile 'Ethernet16|2' applied after port Ethernet16 is up
        NOTICE swss#orchagent: :- processBufferPool: Can't remove object ingress_test_pool due to being referenced (BUFFER_POOL ingress_test_pool first object: ingress_test_profile reference count: 1)
        INFO swss#orchagent: :- doTask: Failed to process buffer task, retry it
        NOTICE swss#orchagent: :- processBufferProfile: Remove buffer profile ingress_test_profile with type BUFFER_PROFILE
        INFO swss#orchagent: :- removeMeFromObjsReferencedByMe: Obj BUFFER_PROFILE.ingress_test_profile Field pool: Remove reference to BUFFER_POOL ingress_test_pool (now 0)
        INFO swss#orchagent: :- removeObject: Obj BUFFER_PROFILE:ingress_test_profile is removed from store
        NOTICE swss#orchagent: :- processBufferPool: Removed buffer pool ingress_test_pool with type BUFFER_POOL
        ```
2. remove the object with zero-reference.
  create the same objects and destroy buffer PGs first and then buffer profiles and at last buffer pools. In this case the objects are destroyed in order.
    - create the buffer pool: `hset BUFFER_POOL|ingress_test_pool type ingress mode dynamic size 14542848`
        ```
        NOTICE swss#orchagent: :- processBufferPool: Created buffer pool ingress_test_pool with type BUFFER_POOL
        ```
    - create the buffer profile: `hset BUFFER_PROFILE|ingress_test_profile dynamic_th 1 pool [BUFFER_POOL|ingress_test_pool] size 0`
        ```
        NOTICE swss#orchagent: :- processBufferProfile: Created buffer profile ingress_test_profile with type BUFFER_PROFILE
        INFO swss#orchagent: :- setObjectReference: Obj BUFFER_PROFILE.ingress_test_profile Field pool: Add reference to BUFFER_POOL ingress_test_pool (now 1)
        ```
    - create buffer PGs: `hset BUFFER_PG|Ethernet8|1 profile [BUFFER_PROFILE|ingress_test_profile]` and `hset BUFFER_PG|Ethernet16|2 profile [BUFFER_PROFILE|ingress_test_profile]`
        ```
        NOTICE swss#orchagent: :- processPriorityGroup: Set buffer PG Ethernet8|1 to BUFFER_PROFILE:ingress_test_profile
        INFO swss#orchagent: :- setObjectReference: Obj BUFFER_PG.Ethernet8|1 Field profile: Add reference to BUFFER_PROFILE ingress_test_profile (now 1)
        NOTICE swss#orchagent: :- processPriorityGroup: Set buffer PG Ethernet16|2 to BUFFER_PROFILE:ingress_test_profile
        INFO swss#orchagent: :- setObjectReference: Obj BUFFER_PG.Ethernet16|2 Field profile: Add reference to BUFFER_PROFILE ingress_test_profile (now 2)
        ```
    - destroy the buffer PGs: `hdel BUFFER_PG|Ethernet16|2 profile` and `hdel BUFFER_PG|Ethernet8|1 profile`
        ```
        NOTICE swss#orchagent: :- processPriorityGroup: Remove buffer PG Ethernet16|2
        INFO swss#orchagent: :- removeMeFromObjsReferencedByMe: Obj BUFFER_PG.Ethernet16|2 Field profile: Remove reference to BUFFER_PROFILE ingress_test_profile (now 1)
        INFO swss#orchagent: :- removeObject: Obj BUFFER_PG:Ethernet16|2 is removed from store
        NOTICE swss#orchagent: :- processPriorityGroup: Remove buffer PG Ethernet8|1
        INFO swss#orchagent: :- removeMeFromObjsReferencedByMe: Obj BUFFER_PG.Ethernet8|1 Field profile: Remove reference to BUFFER_PROFILE ingress_test_profile (now 0)
        INFO swss#orchagent: :- removeObject: Obj BUFFER_PG:Ethernet8|1 is removed from store
        ```
    - destroy the buffer profile: `hdel BUFFER_PROFILE|ingress_test_profile dynamic_th pool size`
        ```
        NOTICE swss#orchagent: :- processBufferProfile: Remove buffer profile ingress_test_profile with type BUFFER_PROFILE
        INFO swss#orchagent: :- removeMeFromObjsReferencedByMe: Obj BUFFER_PROFILE.ingress_test_profile Field pool: Remove reference to BUFFER_POOL ingress_test_pool (now 0)
        INFO swss#orchagent: :- removeObject: Obj BUFFER_PROFILE:ingress_test_profile is removed from store
        ```
    - destroy the buffer pool: `hdel BUFFER_POOL|ingress_test_pool type mode size`
        ```
        NOTICE swss#orchagent: :- processBufferPool: Removed buffer pool ingress_test_pool with type BUFFER_POOL
        ```

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -26,12 +26,12 @@ static const vector<sai_buffer_pool_stat_t> bufferPoolWatermarkStatIds =
 };
 
 type_map BufferOrch::m_buffer_type_maps = {
-    {CFG_BUFFER_POOL_TABLE_NAME, new object_map()},
-    {CFG_BUFFER_PROFILE_TABLE_NAME, new object_map()},
-    {CFG_BUFFER_QUEUE_TABLE_NAME, new object_map()},
-    {CFG_BUFFER_PG_TABLE_NAME, new object_map()},
-    {CFG_BUFFER_PORT_INGRESS_PROFILE_LIST_NAME, new object_map()},
-    {CFG_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME, new object_map()}
+    {CFG_BUFFER_POOL_TABLE_NAME, new object_reference_map()},
+    {CFG_BUFFER_PROFILE_TABLE_NAME, new object_reference_map()},
+    {CFG_BUFFER_QUEUE_TABLE_NAME, new object_reference_map()},
+    {CFG_BUFFER_PG_TABLE_NAME, new object_reference_map()},
+    {CFG_BUFFER_PORT_INGRESS_PROFILE_LIST_NAME, new object_reference_map()},
+    {CFG_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME, new object_reference_map()}
 };
 
 BufferOrch::BufferOrch(DBConnector *db, vector<string> &tableNames) :
@@ -175,7 +175,7 @@ void BufferOrch::generateBufferPoolWatermarkCounterIdList(void)
     for (const auto &it : *(m_buffer_type_maps[CFG_BUFFER_POOL_TABLE_NAME]))
     {
         sai_status_t status = sai_buffer_api->clear_buffer_pool_stats(
-                it.second,
+                it.second.m_saiObjectId,
                 static_cast<uint32_t>(bufferPoolWatermarkStatIds.size()),
                 reinterpret_cast<const sai_stat_id_t *>(bufferPoolWatermarkStatIds.data()));
         if (status ==  SAI_STATUS_NOT_SUPPORTED || status == SAI_STATUS_NOT_IMPLEMENTED)
@@ -201,7 +201,7 @@ void BufferOrch::generateBufferPoolWatermarkCounterIdList(void)
     bitMask = 1;
     for (const auto &it : *(m_buffer_type_maps[CFG_BUFFER_POOL_TABLE_NAME]))
     {
-        string key = BUFFER_POOL_WATERMARK_STAT_COUNTER_FLEX_COUNTER_GROUP ":" + sai_serialize_object_id(it.second);
+        string key = BUFFER_POOL_WATERMARK_STAT_COUNTER_FLEX_COUNTER_GROUP ":" + sai_serialize_object_id(it.second.m_saiObjectId);
 
         if (noWmClrCapability)
         {
@@ -225,7 +225,7 @@ void BufferOrch::generateBufferPoolWatermarkCounterIdList(void)
     m_isBufferPoolWatermarkCounterIdListGenerated = true;
 }
 
-const object_map &BufferOrch::getBufferPoolNameOidMap(void)
+const object_reference_map &BufferOrch::getBufferPoolNameOidMap(void)
 {
     // In the case different Orches are running in
     // different threads, caller may need to grab a read lock
@@ -247,7 +247,7 @@ task_process_status BufferOrch::processBufferPool(Consumer &consumer)
     SWSS_LOG_DEBUG("object name:%s", object_name.c_str());
     if (m_buffer_type_maps[map_type_name]->find(object_name) != m_buffer_type_maps[map_type_name]->end())
     {
-        sai_object = (*(m_buffer_type_maps[map_type_name]))[object_name];
+        sai_object = (*(m_buffer_type_maps[map_type_name]))[object_name].m_saiObjectId;
         SWSS_LOG_DEBUG("found existing object:%s of type:%s", object_name.c_str(), map_type_name.c_str());
     }
     SWSS_LOG_DEBUG("processing command:%s", op.c_str());
@@ -335,7 +335,7 @@ task_process_status BufferOrch::processBufferPool(Consumer &consumer)
                 SWSS_LOG_ERROR("Failed to create buffer pool %s with type %s, rv:%d", object_name.c_str(), map_type_name.c_str(), sai_status);
                 return task_process_status::task_failed;
             }
-            (*(m_buffer_type_maps[map_type_name]))[object_name] = sai_object;
+            (*(m_buffer_type_maps[map_type_name]))[object_name].m_saiObjectId = sai_object;
             SWSS_LOG_NOTICE("Created buffer pool %s with type %s", object_name.c_str(), map_type_name.c_str());
             // Here we take the PFC watchdog approach to update the COUNTERS_DB metadata (e.g., PFC_WD_DETECTION_TIME per queue)
             // at initialization (creation and registration phase)
@@ -347,6 +347,14 @@ task_process_status BufferOrch::processBufferPool(Consumer &consumer)
     }
     else if (op == DEL_COMMAND)
     {
+        if (isObjectBeingReferenced(m_buffer_type_maps, map_type_name, object_name))
+        {
+            auto hint = objectReferenceInfo(m_buffer_type_maps, map_type_name, object_name);
+            SWSS_LOG_NOTICE("Can't remove object %s due to being referenced (%s)", object_name.c_str(), hint.c_str());
+
+            return task_process_status::task_need_retry;
+        }
+
         sai_status = sai_buffer_api->remove_buffer_pool(sai_object);
         if (SAI_STATUS_SUCCESS != sai_status)
         {
@@ -376,11 +384,12 @@ task_process_status BufferOrch::processBufferProfile(Consumer &consumer)
     string map_type_name = consumer.getTableName();
     string object_name = kfvKey(tuple);
     string op = kfvOp(tuple);
+    string pool_name;
 
     SWSS_LOG_DEBUG("object name:%s", object_name.c_str());
     if (m_buffer_type_maps[map_type_name]->find(object_name) != m_buffer_type_maps[map_type_name]->end())
     {
-        sai_object = (*(m_buffer_type_maps[map_type_name]))[object_name];
+        sai_object = (*(m_buffer_type_maps[map_type_name]))[object_name].m_saiObjectId;
         SWSS_LOG_DEBUG("found existing object:%s of type:%s", object_name.c_str(), map_type_name.c_str());
     }
     SWSS_LOG_DEBUG("processing command:%s", op.c_str());
@@ -397,7 +406,7 @@ task_process_status BufferOrch::processBufferProfile(Consumer &consumer)
             if (field == buffer_pool_field_name)
             {
                 sai_object_id_t sai_pool;
-                ref_resolve_status resolve_result = resolveFieldRefValue(m_buffer_type_maps, buffer_pool_field_name, tuple, sai_pool);
+                ref_resolve_status resolve_result = resolveFieldRefValue(m_buffer_type_maps, buffer_pool_field_name, tuple, sai_pool, pool_name);
                 if (ref_resolve_status::success != resolve_result)
                 {
                     if(ref_resolve_status::not_resolved == resolve_result)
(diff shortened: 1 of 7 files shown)


stephenxs:
retest vs, please

orchagent/bufferorch.cpp
```diff
@@ -737,6 +798,9 @@ task_process_status BufferOrch::processIngressBufferProfileList(Consumer &consum
         SWSS_LOG_ERROR("Failed resolving ingress buffer profile reference specified for:%s", key.c_str());
         return task_process_status::task_failed;
     }
+
+    setObjectReference(m_buffer_type_maps, consumer.getTableName(), key, buffer_profile_list_field_name, profile_name_list);
```
neethajohn, line 802: move this just before exiting this function to ensure that the profile is actually applied to the port

orchagent/bufferorch.cpp
```diff
@@ -783,6 +850,9 @@ task_process_status BufferOrch::processEgressBufferProfileList(Consumer &consume
         SWSS_LOG_ERROR("Failed resolving egress buffer profile reference specified for:%s", key.c_str());
         return task_process_status::task_failed;
     }
+
+    setObjectReference(m_buffer_type_maps, consumer.getTableName(), key, buffer_profile_list_field_name, profile_name_list);
```
neethajohn, line 854: move this just before exiting this function to ensure that the profile is actually applied to the port

orchagent/bufferorch.cpp
```diff
@@ -530,17 +551,38 @@ task_process_status BufferOrch::processQueue(Consumer &consumer)
...
+
+            SWSS_LOG_ERROR("Resolving queue profile reference failed");
+            return task_process_status::task_failed;
         }
-        SWSS_LOG_ERROR("Resolving queue profile reference failed");
-        return task_process_status::task_failed;
+
+        SWSS_LOG_NOTICE("Set buffer queue %s to %s", key.c_str(), buffer_profile_name.c_str());
+
+        setObjectReference(m_buffer_type_maps, consumer.getTableName(), key, buffer_profile_field_name, buffer_profile_name);
```
neethajohn, line 572: update the ref only if the set api succeeds

orchagent/bufferorch.cpp
```diff
@@ -530,17 +551,38 @@ task_process_status BufferOrch::processQueue(Consumer &consumer)
...
+    }
+    else if (op == DEL_COMMAND)
+    {
+        sai_buffer_profile = SAI_NULL_OBJECT_ID;
+        SWSS_LOG_NOTICE("Remove buffer queue %s", key.c_str());
+        removeObject(m_buffer_type_maps, consumer.getTableName(), key);
```
neethajohn, line 578: remove object only if the set api succeeds

orchagent/bufferorch.cpp
```diff
@@ -634,17 +672,38 @@ task_process_status BufferOrch::processPriorityGroup(Consumer &consumer)
...
+
+            SWSS_LOG_ERROR("Resolving pg profile reference failed");
+            return task_process_status::task_failed;
         }
-        SWSS_LOG_ERROR("Resolving pg profile reference failed");
-        return task_process_status::task_failed;
+
+        SWSS_LOG_NOTICE("Set buffer PG %s to %s", key.c_str(), buffer_profile_name.c_str());
+
+        setObjectReference(m_buffer_type_maps, consumer.getTableName(), key, buffer_profile_field_name, buffer_profile_name);
```
neethajohn, line 693: update the ref only if the set api succeeds

orchagent/bufferorch.cpp
```diff
@@ -634,17 +672,38 @@ task_process_status BufferOrch::processPriorityGroup(Consumer &consumer)
...
+    }
+    else if (op == DEL_COMMAND)
+    {
+        sai_buffer_profile = SAI_NULL_OBJECT_ID;
+        SWSS_LOG_NOTICE("Remove buffer PG %s", key.c_str());
+        removeObject(m_buffer_type_maps, consumer.getTableName(), key);
```
neethajohn, line 699: remove obj only if set api succeeds

neethajohn (changes requested):

orchagent/orch.cpp
```diff
@@ -401,17 +402,113 @@ ref_resolve_status Orch::resolveFieldRefValue(
...
+
+string Orch::objectReferenceInfo(
+    type_map &type_maps,
+    const string &table,
+    const string &obj_name)
+{
+    auto &objsDependingSet = (*type_maps[table])[obj_name].m_objsDependingOnMe;
+    for (auto &depObjName : objsDependingSet)
+    {
+        string hint = table + " " + obj_name + " first object: " + depObjName;
```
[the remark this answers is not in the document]
stephenxs: Yes. Will fix it.

orchagent/orch.h
```diff
@@ -52,12 +53,23 @@ typedef enum
...
 } task_process_status;
 
+typedef struct
+{
+    // m_objsDependingOnMe stores names (without table name) of all objects depending on the current obj
+    std::set<std::string> m_objsDependingOnMe;
+    // m_objsReferencingByMe is a map from a field of the current object's to the object names it references
+    // the object names are with table name
+    // multiple objects being referenced are separated by ','
+    std::map<std::string, std::string> m_objsReferencingByMe;
```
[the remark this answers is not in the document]
stephenxs: Yes. But it doesn't matter. We don't need to keep order here.

orchagent/bufferorch.cpp
stephenxs: This is the way in which I implemented it originally.
Why did I change it to the current solution eventually? Because:
- the motivation of the PR is to prevent orchagent from releasing an object with non-zero reference.
- even if the SAI API failed, the dependencies among the orchagent are still there. in this sense, not to update the reference in case of SAI API failure will cause the dependency record be broken, which, I suspect, can cause a subsequent issue
- in case of SAI API failed, no steps can be taken in orchagent to recover it. It will depend on the application to recover it.

In this sense, to do it in the current way is better. How do you think? @neethajohn

neethajohn: I am in agreement. @prsunny , @lguohan , do you see any issues with this approach?

prsunny: Since the two returns below doesn't result in retry, lgtm. But just curious, what if there is a retry and `setObjectReference `is called multiple times?

stephenxs: Hi
It doesn’t matter how many times the setObjectReference is called because
- ‘set’ is used to record the reference
- operations like insert and remove are idempotent, which means calling it for many times will result in the same result as if calling it for only once.

You can also refer the ***Reference count vs. set*** section in the PR’s description

Meanwhile, if the reference count approach is chosen to record the reference count, a flag is required.

orchagent/orch.cpp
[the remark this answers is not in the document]
stephenxs: Fixed

stephenxs:
Hi @neethajohn ,
All the comments have been fixed/replied to. Could you please approve the PR?
Thanks.

neethajohn (approved):

kcudnik:
@stephenxs can you fix this:

```
orch.cpp: In member function 'ref_resolve_status Orch::resolveFieldRefArray(type_map&, const string&, swss::KeyOpFieldsValuesTuple&, std::vector<long unsigned int>&, std::string&)':
orch.cpp:609:41: error: 'strlen' argument missing terminating nul [-Werror=stringop-overflow=]
  609 |                     object_name_list += string(&list_item_delimiter);
      |                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from orch.cpp:6:
orch.h:25:12: note: referenced argument declared here
   25 | const char list_item_delimiter = ',';
      |            ^~~~~~~~~~~~~~~~~~~
```

i get this error on master with all enabled errors, im not sure why this don't pop up on jenkins build, my gcc: gcc version 9.1.0 (GCC)

stephenxs:
Hi @kcudnik ,
Looks like I'm not able to install gcc-9 on buster docker. Could you please provide steps in which you installed it?
I also searched this error and got this: [GCC 9: special_values_formatter.hpp:43:16: error: 'strlen' argument missing terminating nul #29](https://github.com/CauldronDevelopmentLLC/cbang/issues/29). According to this page, it's a bug of boost while compiling by gcc9 and can be suppressed by using `-Wno-stringop-overflow` which disables this warning.
Can you also check whether this works for you?
Thanks

kcudnik:
ok, can you add then this option co configure?

PR #1444, opened 2020-09-21
Use .clear() after std::move()

pavel-shirshov:
**What I did**
1. Use .clear() on a vector after using std::move() from it to make sure that the vector is in a correct state.
2. Remove the if condition which is not required here.

**Why I did it**
1. std::move() from an object leaves the object in correct state. The correct state could be not empty. So I clear the vector to make sure that the vector is empty.
2. We don't need to check the vector on the emptiness. The for loop will do nothing, if there is no elements in the vector.

**How I verified it**
Make a draft PR and wait to make sure that there is no regression.

**Details if related**

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -556,14 +556,12 @@ void FdbOrch::updateVlanMember(const VlanMemberUpdate& update)
 
     string port_name = update.member.m_alias;
     auto fdb_list = std::move(saved_fdb_entries[port_name]);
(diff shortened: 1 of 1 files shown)


wendani:
👍

pavel-shirshov:
retest all please

pavel-shirshov:
retest this please

pavel-shirshov:
retest vs please

prsunny (approved):

PR #1451, opened 2020-09-27
remove FDB entires of BridgePort before removing the BridgePort

madhanmellanox:
**What I did**
To fix an issue when removing bridgeport, I introduced code to remove FDB entries of the bridgeport before calling SAI API remove BridgePort.

**Why I did it**
I introduced this fix because by directly removing the bridgeport without flushing the bridge port's FDB entries, the bridge port removal was failing, since it had uncleaned ASIC DB FDB entries.

**How I verified it**
I verified it by sending traffic to the vlan port and creating a FDB entry for the vlan port. By, then when I remove the bridge port member from the VLAN port, the FDB entry was removed as well as bridge port is removed from VLAN membership.

**Details if related**
The fix is introduced in PortsOrch.cpp

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -444,6 +444,16 @@ void PortsOrch::removeDefaultBridgePorts()
         }
         if (attr.value.s32 == SAI_BRIDGE_PORT_TYPE_PORT)
         {
+            Port port;
+	    getPortByBridgePortId(bridge_port_list[i], port);
+
+            //Flush the FDB entires corresponding to the port
+            if (!flushFdbEntries(port))
+            {	    
+                SWSS_LOG_ERROR("Failed to flush FDB entries for port %s",
+                               port.m_alias.c_str());
+            }
+
             status = sai_bridge_api->remove_bridge_port(bridge_port_list[i]);
             if (status != SAI_STATUS_SUCCESS)
             {
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -3130,7 +3130,7 @@ void PortsOrch::doTask(Consumer &consumer)
     }
 }
 
-bool PortsOrch::flushFDBEntries(sai_object_id_t bridge_port_id)
+bool PortsOrch::flushFdbEntries(Port port)
 {
     vector<sai_attribute_t>    attrs;
     sai_attribute_t            attr;
@@ -3138,23 +3138,23 @@ bool PortsOrch::flushFDBEntries(sai_object_id_t bridge_port_id)
 
     SWSS_LOG_ENTER();
 
-    if (SAI_NULL_OBJECT_ID == bridge_port_id)
+    if (SAI_NULL_OBJECT_ID == port.m_bridge_port_id)
     {
-        SWSS_LOG_WARN("Couldn't flush FDB. Bridge port OID: 0x%" PRIx64,
-                      bridge_port_id);
+        SWSS_LOG_WARN("Couldn't flush FDB entries for port: %s",
+                      port.m_alias.c_str());
         return false;
     }
 
     attr.id = SAI_FDB_FLUSH_ATTR_BRIDGE_PORT_ID;
-    attr.value.oid = bridge_port_id;
+    attr.value.oid = port.m_bridge_port_id;
     attrs.push_back(attr);
 
-    SWSS_LOG_INFO("Flushing FDB bridge_port_oid: 0x%" PRIx64, bridge_port_id);
+    SWSS_LOG_INFO("Flushing FDB entries for port: %s", port.m_alias.c_str());
 
     rv = sai_fdb_api->flush_fdb_entries(gSwitchId, (uint32_t)attrs.size(), attrs.data());
     if (SAI_STATUS_SUCCESS != rv)
     {
-        SWSS_LOG_ERROR("Flushing FDB failed. rv:%d", rv);
+        SWSS_LOG_ERROR("Flushing FDB for port %s failed. rv:%d", port.m_alias.c_str(), rv);
 	return false;
     }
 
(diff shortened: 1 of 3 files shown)


orchagent/portsorch.cpp
```diff
@@ -3431,6 +3462,14 @@ bool PortsOrch::removeBridgePort(Port &port)
         return false;
     }
 
+    //Flush the FDB entires corresponding to the port
```
prsunny, line 3465: How about [removeDefaultBridgePorts](https://github.com/Azure/sonic-swss/blob/c165697df2124d971a1c80254e1100c760e99fde/orchagent/portsorch.cpp#L447)

madhanmellanox: addressed in the next commit.

orchagent/portsorch.cpp
```diff
@@ -444,6 +444,16 @@ void PortsOrch::removeDefaultBridgePorts()
         }
         if (attr.value.s32 == SAI_BRIDGE_PORT_TYPE_PORT)
         {
+            Port port;
+	    getPortByBridgePortId(bridge_port_list[i], port);
```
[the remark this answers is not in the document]
madhanmellanox: The flushFdbEntries() member function has a parameter of Port so that we can display the portname in the SWSS logs. Displaying port-oids in SWSS logs may not be useful, that is why we are passing Port as parameter to flushFdbEntries. That is the reason here we have to convert to Port and back again in the function, it will be using m_bridge_port_id. This was one of our Internal PR comments. Will align the code in the next commit.

orchagent/portsorch.cpp
```diff
@@ -444,6 +444,16 @@ void PortsOrch::removeDefaultBridgePorts()
...
+
+            //Flush the FDB entires corresponding to the port
+            if (!flushFdbEntries(port))
+            {	    
+                SWSS_LOG_ERROR("Failed to flush FDB entries for port %s",
```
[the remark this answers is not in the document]
madhanmellanox: Fixed in next commit

orchagent/portsorch.cpp
```diff
@@ -3130,6 +3140,37 @@ void PortsOrch::doTask(Consumer &consumer)
...
+    vector<sai_attribute_t>    attrs;
+    sai_attribute_t            attr;
+    sai_status_t               rv = SAI_STATUS_SUCCESS;
+
+    SWSS_LOG_ENTER();
+
+    if (SAI_NULL_OBJECT_ID == port.m_bridge_port_id)
+    {
+        SWSS_LOG_WARN("Couldn't flush FDB entries for port: %s",
+                      port.m_alias.c_str());
```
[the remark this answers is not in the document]
madhanmellanox: Fixed in next commit.

orchagent/portsorch.cpp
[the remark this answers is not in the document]
prsunny: for the log part, we can log from the caller.. but I feel it is better to avoid this port/bridge_port conversion..

orchagent/portsorch.cpp
```diff
@@ -3431,6 +3470,14 @@ bool PortsOrch::removeBridgePort(Port &port)
         return false;
     }
 
+    //Flush the FDB entires corresponding to the port
+    if (!flushFdbEntries(port))
+    {	    
+        SWSS_LOG_ERROR("Failed to flush FDB entries for port %s",
+                port.m_alias.c_str());
```
prsunny, line 3477: Please have it same line

orchagent/portsorch.cpp
```diff
@@ -3130,6 +3139,36 @@ void PortsOrch::doTask(Consumer &consumer)
...
+    vector<sai_attribute_t>    attrs;
+    sai_attribute_t            attr;
+    sai_status_t               rv = SAI_STATUS_SUCCESS;
+
+    SWSS_LOG_ENTER();
+
+    if (SAI_NULL_OBJECT_ID == port.m_bridge_port_id)
+    {
+        SWSS_LOG_WARN("Couldn't flush FDB entries for port: %s", port.m_alias.c_str());
+        return false;
```
prsunny, line 3153: IMO, this should return 'true'. We dont want retries etc to happen if there is no bridge_port

orchagent/portsorch.cpp
```diff
@@ -3130,6 +3139,36 @@ void PortsOrch::doTask(Consumer &consumer)
     }
 }
 
+bool PortsOrch::flushFdbEntries(Port port)
```
prsunny, line 3142: As we discussed, we can use `fdborch->flushFDBEntries` instead

madhanmellanox: addressed in next commit

orchagent/portsorch.cpp
```diff
@@ -26,6 +26,7 @@
 #include "countercheckorch.h"
 #include "notifier.h"
 #include "redisclient.h"
+#include "fdborch.h"
```
prsunny, line 29: do you need to include this?

madhanmellanox: yes, otherwise it is not able to get the definition of FdbOrch. I am getting compilation errors.

portsorch.cpp:44:8: error: 'FdbOrch' does not name a type; did you mean 'CrmOrch'?
 extern FdbOrch *gFdbOrch;
        ^~~~~~~
        CrmOrch
portsorch.cpp: In member function 'bool PortsOrch::removeBridgePort(swss::Port&)':
portsorch.cpp:3437:5: error: 'gFdbOrch' was not declared in this scope
     gFdbOrch->flushFDBEntries(port.m_bridge_port_id, SAI_NULL_OBJECT_ID);
     ^~~~~~~~
portsorch.cpp:3437:5: note: suggested alternative: 'gCrmOrch'
     gFdbOrch->flushFDBEntries(port.m_bridge_port_id, SAI_NULL_OBJECT_ID);
     ^~~~~~~~
     gCrmOrch

prsunny (approved):

abdosi:
@liat-grozovik cherry-pick has conflcit

madhanmellanox:
@abdosi this PR according to @prsunny should not be ported to 201911 because flushFDBEntries() function is not present in 201911.

lguohan:
included as PR #1516

yxieca:
This change, though cherry-picked cleanly to 201811, but is not able to build.

PR #1454, opened 2020-10-01
adding handler for sai_shell_enable

banagiri:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


orchagent/switchorch.cpp
```diff
@@ -107,6 +108,10 @@ void SwitchOrch::doTask(Consumer &consumer)
                         memcpy(attr.value.mac, mac_addr.getMac(), sizeof(sai_mac_t));
                         break;
 
+                   case SAI_SWITCH_ATTR_SWITCH_SHELL_ENABLE:
```
lguohan, line 111: in sonic-sairedis, we have startDiagShell function to enable debug shell, have you checked that?

PR #1462, opened 2020-10-09
[acl] Refactor port OID retrieval into aclorch

daall:

**What I did**
I moved getAclBindPortId from portsorch to aclorch.

**Why I did it**
This serves two purposes:
1. There was a bug wherein AclTable::update would fail to handle port deletion events because the getAclBindPortId method lived in portsorch. Because this method was using the provided alias to get the OID internally, this method would fail in the delete case because the port had already been deleted from portsorch. As a result, all delete events would fail and print a "Unable to retrieve OID" method. This was caught by the `test_po_update` pytest because it deletes the PortChannel at the end of the test case, which triggers a delete event, which triggers this bug.

2. The behavior implemented in this method is specific to the ACL implementation in SONiC, and it doesn't really make sense to include it in the more generic AclOrch. Furthermore, we have access to all the port info in AclOrch, so there is no need to take a dependency on PortsOrch to fetch the port OID info and perform the parsing logic. There was a `TODO` indicating the same.

**How I verified it**
`test_po_update` passes in the master image now. I also re-ran the `acl` and `everflow` tests and everything looks good to go.

**Details if related**
Fixes https://github.com/Azure/sonic-buildimage/issues/5273

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1555,10 +1555,10 @@ void AclTable::update(SubjectType type, void *cntx)
     }
 
     PortUpdate *update = static_cast<PortUpdate *>(cntx);
-
     Port &port = update->port;
+
     sai_object_id_t bind_port_id;
-    if (!gPortsOrch->getAclBindPortId(port.m_alias, bind_port_id))
+    if (!AclOrch::getAclBindPortId(port, bind_port_id))
     {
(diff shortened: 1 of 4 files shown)


daall:
retest vs please

daall:
retest vs please

daall:
retest vs please

lguohan (approved):

lguohan:
retest vs please

daall:
retest vs please

PR #1465, opened 2020-10-12
[mock tests] Update MockDBConnector to match new swsscommon interface

daall:

**What I did**
I updated the MockDBConnector class to respect the access-level of the m_conn member variable.

**Why I did it**
https://github.com/Azure/sonic-swss-common/pull/387 refactored the DBConnector class and changed the access level of a few member fields. So, to make the mock implementation correct, we need to use the proper setter method to update the redis connection.

**How I verified it**
Built locally w/ latest sairedis and swsscommon changes.

**Details if related**
This should fix the following jenkins issue:
https://sonic-jenkins.westus2.cloudapp.azure.com/job/vs/job/sonic-sairedis-build/1344/console

Which will in turn fix this issue:
https://sonic-jenkins.westus2.cloudapp.azure.com/job/vs/job/sonic-swss-build/1944/console

### DIFF
--- a/tests/mock_tests/mock_dbconnector.cpp
+++ b/tests/mock_tests/mock_dbconnector.cpp
@@ -9,33 +9,25 @@
 
 namespace swss
 {
-    DBConnector::~DBConnector()
-    {
-        close(m_conn->fd);
-        if (m_conn->connection_type == REDIS_CONN_TCP)
-            free(m_conn->tcp.host);
-        else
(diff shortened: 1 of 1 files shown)


lguohan (approved):

lguohan:
@qiluo-msft , the function does not seem like a mock. why do we need it?

qiluo-msft:
It is a mock. Without it, a Redis service is required to run gtest unit tests.

PR #1473, opened 2020-10-22
[intfsorch] Init proxy_arp variable while adding router interface.

Minkang-Tsai:
**What I did**
Init proxy_arp variable while adding router interface.

**Why I did it**
Proxy ARP will be considered as enabled while dealing with deletion action in void IntfsOrch::doTask(Consumer &consumer) if you never enable it.

**How I verified it**
Add an router interface and then delete it. 

**Details if related**


prsunny:
retest vs please

PR #1482, opened 2020-10-26
Create vnet tunnel map only if it doesn't exist

prsunny:
**What I did**
There was a bug that accessed nullpointer, if a vnet config is pushed and it already exist.  This is to check if vnet exists before creating tunnel map

**Why I did it**
Fix issue https://github.com/Azure/sonic-buildimage/issues/5706

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -1391,6 +1391,7 @@ bool VNetOrch::addOperation(const Request& request)
                 {
                     SWSS_LOG_ERROR("VNET '%s', tunnel '%s', map create failed",
                                     vnet_name.c_str(), tunnel.c_str());
+                    return false;
                 }
 
                 SWSS_LOG_NOTICE("VNET '%s' was added ", vnet_name.c_str());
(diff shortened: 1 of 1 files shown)


volodymyrsamotiy (approved):

orchagent/vnetorch.cpp
```diff
@@ -1384,17 +1384,21 @@ bool VNetOrch::addOperation(const Request& request)
...
-            }
 
-            VNetVrfObject *vrf_obj = dynamic_cast<VNetVrfObject*>(obj.get());
-            if (!vxlan_orch->createVxlanTunnelMap(tunnel, TUNNEL_MAP_T_VIRTUAL_ROUTER, vni,
-                                                  vrf_obj->getEncapMapId(), vrf_obj->getDecapMapId(), VXLAN_ENCAP_TTL))
+                VNetVrfObject *vrf_obj = dynamic_cast<VNetVrfObject*>(obj.get());
+                if (!vxlan_orch->createVxlanTunnelMap(tunnel, TUNNEL_MAP_T_VIRTUAL_ROUTER, vni,
+                                                      vrf_obj->getEncapMapId(), vrf_obj->getDecapMapId(), VXLAN_ENCAP_TTL))
+                {
+                    SWSS_LOG_ERROR("VNET '%s', tunnel '%s', map create failed",
```
theasianpianist, line 1392: Do we need to clean up the newly created VNET if this tunnel map creation fails? If the VNET gets created but tunnel map creation fails, if the add operation is performed again the tunnel map creation will not be attempted since the VNET already exists.

prsunny: Sure, added a retry handling if map create failed.

theasianpianist (approved):

prsunny:
retest vs please

Issue #1484, opened 2020-10-26
Orchagent cannot properly handle the scenario when the DEL operation fails in a DEL+SET sequence


shi-su:
**Description**
According to https://github.com/Azure/sonic-swss/pull/1184, m_toSync keeps a maximum of two values per key, DEL or SET, or DEL+SET. When the DEL operation of the DEL+SET sequence fails and the SET operation can still be executed and the DEL operation will remain in m_toSync. An example of neighbor operations is presented in https://github.com/Azure/sonic-swss/issues/1483, while we still need a general solution for other objects.

shi-su:
@lguohan @qiluo-msft @zhenggen-xu for visibility

Issue #1483, opened 2020-10-26
Orchagent receives neighbor DEL operation when the reference count of the neighbor is not zero after reboot


shi-su:
**Description**
The orchagent receives neighbor DEL operation when the reference count of the neighbor is not zero with ping traffic after reboot (cold/fast/warm).

**Steps to reproduce the issue**
Run warm/fast-reboot test (advanced-reboot). Confirmed the issue also happens with cold reboot with the same ping traffic of warm/fast-reboot test.

**Describe the results you received**
Orchagent receives neighbor DEL operation. The operation fails due to a non-zero reference count.
```
Oct 26 22:23:15.259716 vlab-01 INFO swss#orchagent: :- removeNeighbor: Failed to remove still referenced neighbor 1e:92:d0:9e:6d:64 on PortChannel0002
Oct 26 22:23:15.259851 vlab-01 INFO swss#orchagent: :- removeNeighbor: Failed to remove still referenced neighbor 26:74:37:ff:3b:fa on PortChannel0004
```

**Describe the results you received**
Expect not to receive neighbor deletion operations with reference counter.

shi-su:
@qiluo-msft @lguohan @zhenggen-xu for visibility

prsunny:
Looks like related to this  - https://github.com/Azure/sonic-buildimage/issues/4400#issuecomment-714814839

shi-su:
> Looks like related to this - [Azure/sonic-buildimage#4400 (comment)](https://github.com/Azure/sonic-buildimage/issues/4400#issuecomment-714814839)

Yes, I just submitted #1485  to resolve it for neighorch and fix https://github.com/Azure/sonic-buildimage/issues/4400. While the source of the DEL operation remains unknown, this issue aims to track the source of the unexpected operation

PR #1184, opened 2020-01-28
[orch] change Consumer class to support multiple values for the same key

zhenggen-xu:
**Description**

The Consumer class is used by the orch objects to deal with the redis consumer tables' popped items. It has m_toSync map to save the tasks.  During the operations (more items than the orch objects can handle), the tasks in the map is merged for optimization. However, since it is a map, we would only have one value for each key.  This potentially eliminate the necessary actions from Redis, e,g, we have a DEL action and SET action coming out from Redis for some key, we would overwrite the DEL action today in the code and thus not able to delete some objects as we intended. 

The PR changed the m_toSync  from map to multi-map to get multiple values for one key.  In this design, we keep maximun two values per key,  DEL or SET or DEL+SET.  We need strictly keep the order of DEL and SET.  It is possible to use map of vectors to fulfill this,  we chose multi-map because:
 1,  It will have less/no changes to different orch classes to iterate the m_toSync 
 2, The order can be guaranteed. The order of the key-value pairs whose keys compare equivalent is the order of insertion and does not change. (since C++11). See
  https://en.cppreference.com/w/cpp/container/multimap

The PR also refactors the consumer class so vlanmgr.cpp and routeorch.cpp will leverage the Consumer functions instead of operating on the members.  It also refactors the UT code (aclorch_ut.cpp) so it removes the redundant code and uses the same code.

Google UT tests were added for Consumer Class especially for different cases for addToSync() function.

**What I did**
Change the m_toSync in Consumer class to multimap so it could support both DEL and SET
Reload Consumer addToSync() and refactor vlanmgr/route-orch and ut code to use it
Add google ut for consumer class
    
**Why I did it**
See description.

**How I verified it**
Unit tests:
Running main() from gtest_main.cc
```
[==========] Running 19 tests from 5 test cases.
[----------] Global test environment set-up.
[----------] 1 test from AclTest
[ RUN      ] AclTest.Create_L3_Acl_Table
[       OK ] AclTest.Create_L3_Acl_Table (1 ms)
[----------] 1 test from AclTest (1 ms total)

[----------] 3 tests from AclOrchTest
[ RUN      ] AclOrchTest.ACL_Creation_and_Destorying
[       OK ] AclOrchTest.ACL_Creation_and_Destorying (1000 ms)
[ RUN      ] AclOrchTest.L3Acl_Matches_Actions
[       OK ] AclOrchTest.L3Acl_Matches_Actions (1001 ms)
[...]

PR #1485, opened 2020-10-26
[neighorch] Remove pending DEL operation after SET operation for the same key

shi-su:
**What I did**
Remove unfinished DEL operation after SET operation for the same key in neighbor orch.

**Why I did it**
Since DEL operation is supposed to be executed before SET for the same neighbor [ref: #1184 ],  a remaining DEL after the SET operation means the DEL operation failed previously and should not be executed anymore. The remaining DEL operation will be kept for retrying and block following warm-reboot.

**How I verified it**
Verified on kvm.

**Details if related**
Log messages in the presence of failed  DEL operation
```
Oct 26 22:23:15.259851 vlab-01 INFO swss#orchagent: :- removeNeighbor: Failed to remove still referenced neighbor 26:74:37:ff:3b:fa on PortChannel0004
Oct 26 22:23:15.260014 vlab-01 NOTICE swss#orchagent: :- doTask: Removed unfinished DEL operation for PortChannel0004:10.0.0.63 after SET operation
Oct 26 22:23:17.953626 vlab-01 INFO swss#orchagent: :- removeNeighbor: Failed to remove still referenced neighbor 1e:92:d0:9e:6d:64 on PortChannel0002
Oct 26 22:23:17.953626 vlab-01 NOTICE swss#orchagent: :- doTask: Removed unfinished DEL operation for PortChannel0002:10.0.0.59 after SET operation
```
This PR aims to fix the following issue.
https://github.com/Azure/sonic-buildimage/issues/4400
Two other related issues remain to be resolved.
https://github.com/Azure/sonic-swss/issues/1483
https://github.com/Azure/sonic-swss/issues/1484

### DIFF
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -377,7 +377,7 @@ void NeighOrch::doTask(Consumer &consumer)
             if (it_rem != consumer.m_toSync.end() && kfvOp(it_rem->second) == DEL_COMMAND)
             {
                 it = consumer.m_toSync.erase(it_rem);
-                SWSS_LOG_NOTICE("Removed unfinished DEL operation for %s after SET operation", key.c_str());
+                SWSS_LOG_NOTICE("Removed unfinished neighbor DEL operation for %s after SET operation", key.c_str());
             }
         }
         else if (op == DEL_COMMAND)
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -377,7 +377,7 @@ void NeighOrch::doTask(Consumer &consumer)
             if (it_rem != consumer.m_toSync.end() && kfvOp(it_rem->second) == DEL_COMMAND)
             {
                 it = consumer.m_toSync.erase(it_rem);
-                SWSS_LOG_NOTICE("Removed unfinished neighbor DEL operation for %s after SET operation", key.c_str());
+                SWSS_LOG_NOTICE("Removed pending neighbor DEL operation for %s after SET operation", key.c_str());
             }
         }
         else if (op == DEL_COMMAND)
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -373,11 +373,15 @@ void NeighOrch::doTask(Consumer &consumer)
              * Since DEL operation is supposed to be executed before SET for the same neighbor
              * A remaining DEL after the SET operation means the DEL operation failed previously and should not be executed anymore
              */
-            auto it_rem = consumer.m_toSync.find(key);
-            if (it_rem != consumer.m_toSync.end() && kfvOp(it_rem->second) == DEL_COMMAND)
+            if (it != consumer.m_toSync.begin())
             {
-                it = consumer.m_toSync.erase(it_rem);
-                SWSS_LOG_NOTICE("Removed pending neighbor DEL operation for %s after SET operation", key.c_str());
+                auto it_rem = it;
+                it_rem--;
+                if (it_rem->first == key && kfvOp(it_rem->second) == DEL_COMMAND)
+                {
+                    consumer.m_toSync.erase(it_rem);
+                    SWSS_LOG_NOTICE("Removed pending neighbor DEL operation for %s after SET operation", key.c_str());
+                }
             }
         }
         else if (op == DEL_COMMAND)
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -373,15 +373,11 @@ void NeighOrch::doTask(Consumer &consumer)
              * Since DEL operation is supposed to be executed before SET for the same neighbor
              * A remaining DEL after the SET operation means the DEL operation failed previously and should not be executed anymore
              */
-            if (it != consumer.m_toSync.begin())
+            auto rit = make_reverse_iterator(it);
+            while (rit != consumer.m_toSync.rend() && rit->first == key && kfvOp(rit->second) == DEL_COMMAND)
             {
-                auto it_rem = it;
-                it_rem--;
-                if (it_rem->first == key && kfvOp(it_rem->second) == DEL_COMMAND)
-                {
-                    consumer.m_toSync.erase(it_rem);
-                    SWSS_LOG_NOTICE("Removed pending neighbor DEL operation for %s after SET operation", key.c_str());
-                }
+                consumer.m_toSync.erase(next(rit).base());
+                SWSS_LOG_NOTICE("Removed pending neighbor DEL operation for %s after SET operation", key.c_str());
             }
         }
         else if (op == DEL_COMMAND)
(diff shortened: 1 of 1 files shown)


orchagent/neighorch.cpp
```diff
@@ -354,13 +354,31 @@ void NeighOrch::doTask(Consumer &consumer)
...
+                }
             }
             else
+            {
                 /* Duplicate entry */
                 it = consumer.m_toSync.erase(it);
+            }
+
+            /* Remove remaining DEL operation in m_toSync for the same neighbor.
+             * Since DEL operation is supposed to be executed before SET for the same neighbor [ref: https://github.com/Azure/sonic-swss/pull/1184]
```
prsunny, line 373: Just a suggestion, please remove reference to PRs/issues in the comment section.

orchagent/neighorch.cpp
```diff
@@ -354,13 +354,31 @@ void NeighOrch::doTask(Consumer &consumer)
...
+             * A remaining DEL after the SET operation means the DEL operation failed previously and should not be executed anymore
+             */
+            auto it_rem = consumer.m_toSync.find(key);
+            if (it_rem != consumer.m_toSync.end() && kfvOp(it_rem->second) == DEL_COMMAND)
+            {
+                it = consumer.m_toSync.erase(it_rem);
+                SWSS_LOG_NOTICE("Removed unfinished neighbor DEL operation for %s after SET operation", key.c_str());
```
prsunny, line 380: This can be at INFO level.

orchagent/neighorch.cpp
```diff
@@ -354,13 +354,31 @@ void NeighOrch::doTask(Consumer &consumer)
...
             }
             else
+            {
                 /* Duplicate entry */
                 it = consumer.m_toSync.erase(it);
+            }
+
+            /* Remove remaining DEL operation in m_toSync for the same neighbor.
+             * Since DEL operation is supposed to be executed before SET for the same neighbor [ref: https://github.com/Azure/sonic-swss/pull/1184]
+             * A remaining DEL after the SET operation means the DEL operation failed previously and should not be executed anymore
```
prsunny, line 374: What if a  DEL is indeed coming after SET but not the previous failed one?

orchagent/neighorch.cpp
shi-su: Removed

orchagent/neighorch.cpp
shi-su: I thought this is an unexpected scenario and it should be made easily noticeable by users. Also, this would only happen once for each key and should not log a lot of redundant messages.

orchagent/neighorch.cpp
shi-su: According to the design in #1184, when a DEL operation comes later than SET, orch will remove the SET and only keep the DEL. Therefore, DEL+SET happens if and only if DEL is expected to be executed earlier. The scenario where there are both DEL and SET operations in m_toSync and the DEL operation should be executed later shall never happen. This is also why I referred to the PR in the comment.

lguohan (dismissed):

orchagent/neighorch.cpp
prsunny: Agree, but this looks like an unactionable log. Anyways, if it is not expected to be high even with large number of neighbor entries, lgtm.

prsunny (approved):

orchagent/neighorch.cpp
```diff
@@ -354,13 +354,31 @@ void NeighOrch::doTask(Consumer &consumer)
...
+
+            /* Remove remaining DEL operation in m_toSync for the same neighbor.
+             * Since DEL operation is supposed to be executed before SET for the same neighbor
+             * A remaining DEL after the SET operation means the DEL operation failed previously and should not be executed anymore
+             */
+            auto it_rem = consumer.m_toSync.find(key);
+            if (it_rem != consumer.m_toSync.end() && kfvOp(it_rem->second) == DEL_COMMAND)
+            {
+                it = consumer.m_toSync.erase(it_rem);
+                SWSS_LOG_NOTICE("Removed unfinished neighbor DEL operation for %s after SET operation", key.c_str());
```
qiluo-msft, line 380: >unfinished [](start = 41, length = 10)

pending? I changed the PR title.

orchagent/neighorch.cpp
```diff
@@ -354,13 +354,31 @@ void NeighOrch::doTask(Consumer &consumer)
...
+            {
                 /* Duplicate entry */
                 it = consumer.m_toSync.erase(it);
+            }
+
+            /* Remove remaining DEL operation in m_toSync for the same neighbor.
+             * Since DEL operation is supposed to be executed before SET for the same neighbor
+             * A remaining DEL after the SET operation means the DEL operation failed previously and should not be executed anymore
+             */
+            auto it_rem = consumer.m_toSync.find(key);
```
qiluo-msft, line 376: >find [](start = 44, length = 4)

Is it more efficient to travel backward and erase everything with the same key?

qiluo-msft (commented):
As comments

orchagent/neighorch.cpp
```diff
@@ -354,13 +354,31 @@ void NeighOrch::doTask(Consumer &consumer)
...
+            if (it_rem != consumer.m_toSync.end() && kfvOp(it_rem->second) == DEL_COMMAND)
+            {
+                it = consumer.m_toSync.erase(it_rem);
```
zhenggen-xu, line 379: "it" was already in the right position, we don't need assign it again. 
I also think it is more efficient to travel backwards (it_rem = it--), and check if the key is the same as the current one, if so, just erase the key.

orchagent/neighorch.cpp
qiluo-msft: Try [https://en.cppreference.com/w/cpp/iterator/make_reverse_iterator](https://en.cppreference.com/w/cpp/iterator/make_reverse_iterator) to make code cleaner?

---
In reply to: [512406052](https://github.com/Azure/sonic-swss/pull/1485#discussion_r512406052) [](ancestors = 512406052)

zhenggen-xu: actually I had a second thought here, since it is a multi-map, lookup with key should be fast.  To make it simple, how about have simple code as below: 

rm_num = consumer.m_toSync.erase(key);
if (rm_num) 
     SWSS_LOG_NOTICE(...);

above will handle key exist or not exist case.  

@qiluo-msft @shi-su

shi-su: @zhenggen-xu @qiluo-msft Thanks for the suggestions. I think erasing by key has a logarithm complexity. I updated the code to use a reverse iterator to do the job. In this scenario, since there would be at most one pending DEL operation, using the reverse iterator should achieve a constant complexity. Please let me know what you think.

zhenggen-xu: yeah, understood it was logarithm but that extra time did not seem to be a concern in our case, but it gives us a better readability and maintainability. 
Anyway, I don't have strong opinion here, either way is not complex anyways.   So I leave it to @qiluo-msft

qiluo-msft (approved):

shi-su:
retest vs please

PR #1486, opened 2020-10-27
[everflow] Add retry mechanism for mirror sessions and policers

daall:

**What I did**
I added retry mechanisms to `AclOrch` for mirror sessions and `MirrorOrch` for policers.

**Why I did it**
Fixes https://github.com/Azure/sonic-buildimage/issues/5497

**How I verified it**
Tested on a DUT locally. ~**THIS STILL NEEDS VS TESTS AS WELL.**~

**Details if related**
This will not cherry-pick cleanly to 201911 b/c of the port mirroring feature, so I will open a separate PR for 201911 once this one is reviewed + merged.

~In draft while I finish the VS tests.~

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1083,12 +1083,6 @@ bool AclRuleMirror::validateAddAction(string attr_name, string attr_value)
 
     m_sessionName = attr_value;
 
-    if (!m_pMirrorOrch->sessionExists(m_sessionName))
-    {
-        SWSS_LOG_ERROR("Mirror rule reference mirror session that does not exists %s", m_sessionName.c_str());
-        return false;
-    }
-
     // insert placeholder value, we'll set the session oid in AclRuleMirror::create()
     m_actions[action] = sai_attribute_value_t{};
 
@@ -1177,6 +1171,12 @@ bool AclRuleMirror::create()
 
     sai_object_id_t oid = SAI_NULL_OBJECT_ID;
     bool state = false;
+
+    if (!m_pMirrorOrch->sessionExists(m_sessionName))
+    {
+        SWSS_LOG_ERROR("Mirror rule references mirror session \"%s\" that does not exist yet", m_sessionName.c_str());
+        return false;
+    }
 
     if (!m_pMirrorOrch->getSessionStatus(m_sessionName, state))
     {
@@ -3122,7 +3122,16 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
             }
 
 
-            newRule = AclRule::makeShared(type, this, m_mirrorOrch, m_dTelOrch, rule_id, table_id, t);
+            try
+            {
+                newRule = AclRule::makeShared(type, this, m_mirrorOrch, m_dTelOrch, rule_id, table_id, t);
+            }
+            catch (exception &e)
+            {
+                SWSS_LOG_ERROR("Error while creating ACL rule %s: %s", rule_id.c_str(), e.what());
+                it = consumer.m_toSync.erase(it);
+                return;
+            }
 
             for (const auto& itr : kfvFieldsValues(t))
             {
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -2564,31 +2564,6 @@ void AclOrch::update(SubjectType type, void *cntx)
     }
 }
 
-bool AclOrch::bake()
-{
-    SWSS_LOG_ENTER();
-
-    // Wait to install rules until after the orchagent restoration so that
-    // all mirror sessions are available
-    m_freezeRuleInstallation = true;
-
-    return Orch::bake();
-}
-
-bool AclOrch::postBake()
-{
-    SWSS_LOG_ENTER();
-
-    SWSS_LOG_NOTICE("Running AclOrch post-baking steps");
-
-    // Unfreeze the ACL rule updates
-    m_freezeRuleInstallation = false;
-
-    Orch::doTask();
-
-    return Orch::postBake();
-}
-
 void AclOrch::doTask(Consumer &consumer)
 {
     SWSS_LOG_ENTER();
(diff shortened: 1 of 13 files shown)


orchagent/orchdaemon.cpp
```diff
@@ -566,6 +566,17 @@ bool OrchDaemon::warmRestoreAndSyncUp()
...
+    /*
+     * MIRROR ACL rules depend on mirror sessions, which won't be up after warm
+     * reboot until after the postBake step has finished. So, we need to give
+     * AclOrch one more round to resolve any rules that are waiting for mirror
+     * sessions.
+     *
+     * TODO: Implement a more generic mechanism for handling these types of
+     * hard dependencies.
+     */
+    gAclOrch->Orch::doTask();
```
wendani, line 578: You can overwrite postBake in aclOrch

daall: We considered that approach too, but the main concern is that the ACL rule operations must be processed strictly after the mirror sessions are created, otherwise the warm reboot validation will fail and orchagent will crash.

We could enforce this through the ordering of the orchs in the orch list, but putting the code here after postBake is more explicit and it's less likely to be re-arranged in the future and cause the warm reboot behavior to regress.

We could also do some sort of retry for postBake, but that is a bigger structural change than this which is beyond the scope of this bug fix.

orchagent/orchdaemon.cpp
```diff
@@ -561,6 +561,17 @@ bool OrchDaemon::warmRestoreAndSyncUp()
         o->postBake();
```
qiluo-msft, line 561: >postBake [](start = 11, length = 8)

This is special to MirrorOrch. Although it is not your code, I am wondering if possible to make it similar to other Orch, and we could add one more iteration to above for loop to cover both MirrorOrch and gAclOrch.

yxieca:
retest vs please

orchagent/orchdaemon.cpp
daall: I took a closer look at the code and ended up with (I think) what you and @wendani were suggesting. It does seem a lot cleaner now, but let me know if you had something else in mind!

qiluo-msft: Offline discussed, we should explore the possibility to remove `postBake()` interface completely.
To postpone the MirrorOrch processing in warmboot, following below steps:

1. loop all other Orch(s) several rounds of `doTask()` in order to drain the consumer data
2. `MirrorOrch::doTask()`
3. `AclOrch doTask()`


---
In reply to: [519059988](https://github.com/Azure/sonic-swss/pull/1486#discussion_r519059988) [](ancestors = 519059988)

orchagent/orchdaemon.cpp
```diff
@@ -548,18 +549,24 @@ bool OrchDaemon::warmRestoreAndSyncUp()
...
     }
 
-    for (Orch *o : m_orchList)
-    {
-        o->postBake();
-    }
+    // MirrorOrch depends on everything else being settled before it can run,
+    // and mirror ACL rules depend on MirrorOrch, so run these two at the end
+    // after the rest of the data has been processed.
+    gMirrorOrch->Orch::doTask();
```
qiluo-msft, line 568: >doTask [](start = 23, length = 6)

It's a virtual function. We should call the `gMirrorOrch->doTask();`

qiluo-msft (commented):
One minor issue added.

orchagent/orchdaemon.cpp
daall: The compiler throws an error when I try to do this b/c MirrorOrch overloads doTask.

orchagent/aclorch.cpp
```diff
@@ -1178,6 +1172,12 @@ bool AclRuleMirror::create()
     sai_object_id_t oid = SAI_NULL_OBJECT_ID;
     bool state = false;
 
+    if (!m_pMirrorOrch->sessionExists(m_sessionName))
```
wendani, line 1175: Moving mirror session existence check here is to ensure acl rule will be retried if mirror session is absent instead of being erased (`bAllAttributesOk` == false logic)?

daall: That's right!

orchagent/mirrororch.cpp
```diff
@@ -364,7 +344,7 @@ void MirrorOrch::createEntry(const string& key, const vector<FieldValueTuple>& d
                 if (!entry.srcIp.isV4())
                 {
                     SWSS_LOG_ERROR("Unsupported version of sessions %s source IP address", key.c_str());
-                    return;
+                    return task_process_status::task_failed;
```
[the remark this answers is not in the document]
daall: I used `task_invalid_entry` for the case where the mirror session already exists b/c we're invalidating the entire create/set operation, and I used `task_failed` for everything else b/c we have started processing the mirror session and hit an invalid/unrecognized field.

orchagent/mirrororch.cpp
```diff
@@ -1429,26 +1406,31 @@ void MirrorOrch::doTask(Consumer& consumer)
...
         else
         {
             SWSS_LOG_ERROR("Unknown operation type %s", op.c_str());
         }
 
+        // Specifically retry the task when asked
+        if (task_status == task_process_status::task_need_retry)
+        {
+            it++;
+            continue;
```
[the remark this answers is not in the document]
daall: Good catch, thanks!

orchagent/mirrororch.cpp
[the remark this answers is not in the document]
qiluo-msft: Suggest names:
for the case where the mirror session already -> task_duplicated
hit an invalid/unrecognized field -> task_invalid_field
others->task_failed


---
In reply to: [539543206](https://github.com/Azure/sonic-swss/pull/1486#discussion_r539543206) [](ancestors = 539543206)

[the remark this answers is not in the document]
daall: Updated.

qiluo-msft (approved):

PR #1487, opened 2020-10-27
[acl] Update CRM to include LAG bindings for ACL tables

daall:

**What I did**
I updated the AclOrch to correctly report using LAG bindings.

**Why I did it**
Fixes the issue described here: https://github.com/Azure/sonic-utilities/pull/1178

**How I verified it**
Tested on a DUT locally. ~**THIS STILL NEEDS VS TESTS AS WELL.**~

**Details if related**
~In draft while I finish the VS tests.~

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1367,6 +1367,8 @@ bool AclTable::create()
         {
             gCrmOrch->incCrmAclUsedCounter(
                     CrmResourceType::CRM_ACL_TABLE, (sai_acl_stage_t)attr.value.s32, SAI_ACL_BIND_POINT_TYPE_PORT);
+            gCrmOrch->incCrmAclUsedCounter(
+                    CrmResourceType::CRM_ACL_TABLE, (sai_acl_stage_t)attr.value.s32, SAI_ACL_BIND_POINT_TYPE_LAG);
         }
 
         return status == SAI_STATUS_SUCCESS;
@@ -1551,6 +1553,7 @@ bool AclTable::create()
     if (status == SAI_STATUS_SUCCESS)
     {
         gCrmOrch->incCrmAclUsedCounter(CrmResourceType::CRM_ACL_TABLE, acl_stage, SAI_ACL_BIND_POINT_TYPE_PORT);
+        gCrmOrch->incCrmAclUsedCounter(CrmResourceType::CRM_ACL_TABLE, acl_stage, SAI_ACL_BIND_POINT_TYPE_LAG);
     }
 
     return status == SAI_STATUS_SUCCESS;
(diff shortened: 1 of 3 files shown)


orchagent/aclorch.cpp
```diff
@@ -2837,9 +2840,16 @@ bool AclOrch::removeAclTable(string table_id)
     if (deleteUnbindAclTable(table_oid) == SAI_STATUS_SUCCESS)
     {
         auto stage = m_AclTables[table_oid].stage;
+        auto type = m_AclTables[table_oid].type;
+
         sai_acl_stage_t sai_stage = (stage == ACL_STAGE_INGRESS) ? SAI_ACL_STAGE_INGRESS : SAI_ACL_STAGE_EGRESS;
         gCrmOrch->decCrmAclUsedCounter(CrmResourceType::CRM_ACL_TABLE, sai_stage, SAI_ACL_BIND_POINT_TYPE_PORT, table_oid);
 
+        if (type != ACL_TABLE_PFCWD)
```
prsunny, line 2848: If another ACL table is added similar to PFCWD (which doesn't bind to LAG), then this leaves a gap. But I don't think there is a better way other than having a logic to go over the bindpoint list of acl table and address it in a generic way. So may be for a future commit.

wangshengjun: @prsunny If there is not a generic way to address the issue, why we need this PR in aclorch?

prsunny: @wangshengjun , this PR is definitely addressing the issue. I was referring to a different approach to handle this so that crm can be incremented based on bind list rather than duplicating the code at multiple places. Like a slight re-design.

wangshengjun: @prsunny 
Scenario 1: only port bindings for ACL tables
With this PR
CRM: port + 1, lag + 1 (wrong)
Without this PR
CRM:port + 1 (right)
Scenario 2: only lag bindings for ACL tables
With this PR
CRM: port + 1, lag + 1 (wrong)
Without this PR
CRM:port + 1  (wrong)
Scenario 3: both port and lag bindings for ACL tables
With this PR
CRM: port + 1, lag + 1 (right)
Without this PR
CRM:port + 1  (wrong)

daall: @wangshengjun please take a look at: https://github.com/Azure/sonic-swss/blob/04e7341cb2f45141f8bb33cd444922f3dcfd1952/orchagent/aclorch.cpp#L1324

All ACL tables have port + LAG bindings regardless of what interfaces you add to them. Whether that is the correct behavior or not is beyond the scope of this PR.

@prsunny I agree with you that this is not the most robust way of handling this, but I think the overall create/delete flow for ACL tables is pretty fragmented and needs to be revisited. I think we should take this up in a future PR.

prsunny (dismissed):

rlhui:
@prsunny  - is this safe for 20191130?

prsunny:
@rlhui , yes it should be.. but not sure if there would be any conflicts.. @daall ?

daall:
> @rlhui , yes it should be.. but not sure if there would be any conflicts.. @daall ?

There may be a conflict w/ the VS test change, but not 100% sure. If it doesn't cherry-pick cleanly I can open a PR against 201911

PR #1488, opened 2020-10-30
[orchagent] Arm 32-bit arch compilation warning Fixes

Sabareesh-Kumar-Anandan:

**What I did**
Replaced %l variants with PRI format specifier to fix compilation warnings for arm 32-bit arch

**Why I did it**

**How I verified it**
Compiled and verified for arm 32-bit arch (marvell-armhf)

**Details if related**

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -748,7 +748,7 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
     if (set)
     {
         port_attr.value.objlist.count = 1;
-        port_attr.value.objlist.list = (long unsigned int *)calloc(port_attr.value.objlist.count, sizeof(sai_object_id_t));
+        port_attr.value.objlist.list = reinterpret_cast<sai_object_id_t *>((void *)calloc(port_attr.value.objlist.count, sizeof(sai_object_id_t)));
         port_attr.value.objlist.list[0] = sessionId;
     }
     else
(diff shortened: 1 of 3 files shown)


orchagent/mirrororch.cpp
```diff
@@ -748,7 +748,7 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
     if (set)
     {
         port_attr.value.objlist.count = 1;
-        port_attr.value.objlist.list = (long unsigned int *)calloc(port_attr.value.objlist.count, sizeof(sai_object_id_t));
+        port_attr.value.objlist.list = reinterpret_cast<sai_object_id_t *>((void *)calloc(port_attr.value.objlist.count, sizeof(sai_object_id_t)));
```
qiluo-msft, line 751: >calloc [](start = 83, length = 6)

`calloc` already returns a `void *`. Why you need to convert explicitly?

qiluo-msft (changes requested):
As comment

orchagent/mirrororch.cpp
Sabareesh-Kumar-Anandan: reinterpret cast reports error if we don't explicitly type cast

qiluo-msft: I mean why convert `calloc` to `void *`. It is already `void *`.

Sabareesh-Kumar-Anandan: removed explicit cast

qiluo-msft (approved):

PR #1491, opened 2020-10-31
[vstest]: stablize fgnhg test

lguohan:
**What I did**
use standard wait API to wait for asic db and state db updates

**Why I did it**
stabilize the tst

**How I verified it**
run fgnhg 10 times locally

**Details if related**


lguohan:
```
21:27:55  test_fgnhg.py::TestFineGrainedNextHopGroup::test_route_fgnhg PASSED      [ 32%]
```

PR #1493, opened 2020-11-01
[vstest]: fix swss test for synchronized api

lguohan:

**What I did**
fix swss test for synchronous mode.

**Why I did it**
see test pass

**How I verified it**

**Details if related**
this pr MUST NOT be merged.


kcudnik:
is this fixing number of "SET" operations ? if so then this seems not right approach

kcudnik:
fix in sairedis: https://github.com/Azure/sonic-sairedis/pull/689

PR #1126, opened 2019-11-11
Natsyncd changes in sonic-swss submodule to support NAT feature.

AkhileshSamineni:
Added natsyncd and warmboot related changes.

Link to NAT HLD : https://github.com/Azure/SONiC/blob/master/doc/nat/nat_design_spec.md

Depends on:
sonic-swss : https://github.com/Azure/sonic-swss/pull/1059/ and https://github.com/Azure/sonic-swss/pull/1125
sonic-swss-common : Azure/sonic-swss-common#304
sonic-linux-kernel : Azure/sonic-linux-kernel#100
sonic-sairedis : Azure/sonic-sairedis#519

PR #1498, opened 2020-11-10
[fast-reboot] Add a check for warmstart before cleaning up neigh table

bingwang-ms:
**What I did**
Fix https://github.com/Azure/sonic-buildimage/issues/5841 and https://github.com/Azure/sonic-buildimage/issues/5580

We found that neighbor table loaded by ```swssconfig``` from ```arp.json``` after ```fast-reboot``` is cleared by ```neighsyncd``` mistakenly at the initial stage. This PR adds a check for ```WarmStart``` before cleaning up, and only do that if ```WarmStart``` is enable.

**Why I did it**
This PR is to fix the issue that arp table is not recovered after fast-reboot.

**How I verified it**
Verified on Arista-7260, running 201911 image.
1. Run some test to populate ARP entries on DUT, such as ```test_fast_reboot```
2. Issue a fast-reboot
3. Verify the ```arp.json``` backed up by ```fast-reboot-dump.py``` is loaded and NEIGH_TABLE is restored.
```
admin@str-7260cx3-acs-2:~$ redis-cli -n 0 keys '*NEIGH_TABLE*'
  1) "NEIGH_TABLE:Vlan1000:192.168.0.231"
  2) "NEIGH_TABLE:Vlan1000:192.168.0.106"
  3) "NEIGH_TABLE:Vlan1000:192.168.1.102"
  ....
509) "NEIGH_TABLE:Vlan1000:192.168.0.47"
510) "NEIGH_TABLE:Vlan1000:192.168.1.92"
```

```
py.test --inventory ../ansible/str,../ansible/veos --host-pattern str-7260cx3-acs-2 --module-path ../ansible --testbed vms7-t0-7260-2 --testbed_file ../ansible/testbed.csv --junit-xml=tr.xml --log-cli-level info --collect_techsupport=False --topology=t0,any,util platform_tests/test_advanced_reboot.py::test_fast_reboot
========================================================================================= test session starts =========================================================================================                                                                                                                        
platform_tests/test_advanced_reboot.py::test_fast_reboot 
------------------------------------------------------------------------------------------- live log setup --------------------------------------------------------------------------------------------
PASSED                                                                                                                                                                                          [100%]
------------------------------------------------------------------------------------------ live log teardown ------------------------------------------------------------------------------------------
===================================================================================== 1 passed in 404.24 seconds ======================================================================================
```
**Details if related**

### DIFF
--- a/warmrestart/warmRestartAssist.cpp
+++ b/warmrestart/warmRestartAssist.cpp
@@ -79,7 +79,10 @@ void AppRestartAssist::registerAppTable(const std::string &tableName, ProducerSt
     m_psTables[tableName]  = psTable;
 
     // Clear the producerstate table to make sure no pending data for the AppTable
-    psTable->clear();
+    if (m_warmStartInProgress)
+    {
+        psTable->clear();
+    }
     m_appTables[tableName] = new Table(m_pipeLine, tableName, false);
 }
 
@@ -150,7 +153,7 @@ void AppRestartAssist::readTablesToMap()
     }
     return;
 }
- 
+
 /*
  * Check and insert to CacheMap Logic:
  * if delete_key:


warmrestart/warmRestartAssist.cpp
```diff
@@ -79,7 +79,10 @@ void AppRestartAssist::registerAppTable(const std::string &tableName, ProducerSt
     m_psTables[tableName]  = psTable;
 
     // Clear the producerstate table to make sure no pending data for the AppTable
-    psTable->clear();
+    if (m_warmStartInProgress)
+    {
+        psTable->clear();
```
tahmed-dev, line 84: Looks like this change will affect both `neighsyncd` and `natsyncd`. @lguohan is it OK to not clear the table for `natsyncd` when the dut is warm-rebooting? If not and to limit the change to limit `neighsyncd`, we can also check the table name to be `neighsyncd`

yxieca: Wouldn't nat hit same issue if this protection is not there?

tahmed-dev: It is not clear if it is required for NAT tables to be cleared unconditionally or not. Needs a bit of digging into how NAT is using this shared class. NAT is using this method for 4 table [here](https://github.com/Azure/sonic-swss/blob/master/natsyncd/natsync.cpp#L57)

zhenggen-xu: This is a bug introduced by https://github.com/Azure/sonic-swss/pull/1126 where the refactoring of the library to support multiple tables was calling this psTable->clear unconditionally. The way the fix doing here should be correct.  NAT tables or any client should not use the library to flush producer state table in non warm-reboot cases.

zhenggen-xu (approved):

warmrestart/warmRestartAssist.cpp
zhenggen-xu: btw: The reason we use psTable clear was to make sure the relevant table wouldn’t change in some corner cases after we dumped it to memory,  this is required only in warm-reboot/restart case before we dump the table.  Also, since the daemon using the library was the producer itself, it was safe to do so.   However, this assumption was broken if we use swssconfig to load the table at the same time,  which is the case for non warm-reboot cases for arp, nat tables etc.  In those cases, the library cleared the requests from swssconfig incorrectly and cause the issues reported.

qiluo-msft (approved):

PR #1501, opened 2020-11-11
[Routeorch] Fix next hop group reference count in bulk operation

shi-su:
**What I did**
Remove next-hop groups after updating the reference counter for a bulk of routes instead of removing next-hop groups in the loop of updating the reference counter.
Fix https://github.com/Azure/sonic-buildimage/issues/5813

**Why I did it**
The bulk route API has two loops of updating the reference counter: 
1. update the sairedis reference counter
2. update the orchagent reference counter

Currently, the removal of next-hop groups is triggered in the second loop of updating the orchagent reference counter when the reference counter decreases to zero. This may result in a reference counter mismatch between orchagent and sairedis since the sairedis reference counter has already included the operation of the whole bulk but the orchagent reference counter has not. Therefore, the removal of next-hop group may fail due to the mismatch in reference counter (e.g., there are some other routes point to the next-hop group but has not been counted in orchagent yet). 

To fix this problem, the next-hop group removal operation should be done after updating the reference counter of the whole bulk to make sure the reference counters sairedis and orchagent matches.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -1404,7 +1404,10 @@ bool RouteOrch::addRoutePost(const RouteBulkContext& ctx, const NextHopGroupKey 
         increaseNextHopRefCount(nextHops);
 
         decreaseNextHopRefCount(it_route->second);
-        bulkNhgReducedRefCnt.emplace(it_route->second);
+        if (it_route->second.getSize() > 1)
+        {
+            bulkNhgReducedRefCnt.emplace(it_route->second);
+        }
         SWSS_LOG_INFO("Post set route %s with next hop(s) %s",
                 ipPrefix.to_string().c_str(), nextHops.to_string().c_str());
     }
@@ -1543,7 +1546,10 @@ bool RouteOrch::removeRoutePost(const RouteBulkContext& ctx)
      * Decrease the reference count only when the route is pointing to a next hop.
      */
     decreaseNextHopRefCount(it_route->second);
-    bulkNhgReducedRefCnt.emplace(it_route->second);
+    if (it_route->second.getSize() > 1)
+    {
+        bulkNhgReducedRefCnt.emplace(it_route->second);
+    }
 
     SWSS_LOG_INFO("Remove route %s with next hop(s) %s",
             ipPrefix.to_string().c_str(), it_route->second.to_string().c_str());
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -22,8 +22,6 @@ extern CrmOrch *gCrmOrch;
 
 const int routeorch_pri = 5;
 
-std::set<NextHopGroupKey> bulkNhgReducedRefCnt;
-
 RouteOrch::RouteOrch(DBConnector *db, string tableName, SwitchOrch *switchOrch, NeighOrch *neighOrch, IntfsOrch *intfsOrch, VRFOrch *vrfOrch, FgNhgOrch *fgNhgOrch) :
         gRouteBulker(sai_route_api),
         gNextHopGroupMemberBulker(sai_next_hop_group_api, gSwitchId),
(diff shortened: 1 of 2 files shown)


orchagent/routeorch.cpp
```diff
@@ -1392,11 +1411,6 @@ bool RouteOrch::addRoutePost(const RouteBulkContext& ctx, const NextHopGroupKey
         increaseNextHopRefCount(nextHops);
 
         decreaseNextHopRefCount(it_route->second);
-        if (it_route->second.getSize() > 1
-            && m_syncdNextHopGroups[it_route->second].ref_count == 0)
```
qiluo-msft, line 1396: Is it easier to insert `bulkNhgReducedRefCnt` here?

orchagent/routeorch.cpp
```diff
@@ -1533,16 +1547,8 @@ bool RouteOrch::removeRoutePost(const RouteBulkContext& ctx)
 
     /*
      * Decrease the reference count only when the route is pointing to a next hop.
-     * Decrease the reference count when the route is pointing to a next hop group,
-     * and check whether the reference count decreases to zero. If yes, then we need
-     * to remove the next hop group.
      */
     decreaseNextHopRefCount(it_route->second);
-    if (it_route->second.getSize() > 1
```
qiluo-msft, line 1541: The same

orchagent/routeorch.cpp
```diff
@@ -708,6 +718,15 @@ void RouteOrch::doTask(Consumer& consumer)
                     it_prev++;
             }
         }
+
+        /* Remove next hop group if the reference count decreases to zero */
+        for (auto it_nhg = bulkNhgReducedRefCnt.begin(); it_nhg != bulkNhgReducedRefCnt.end(); it_nhg++)
+        {
+            if (m_syncdNextHopGroups[*it_nhg].ref_count == 0)
+            {
+                removeNextHopGroup(*it_nhg);
```
qiluo-msft, line 727: >removeNextHopGroup [](start = 16, length = 18)

Could you write a vs testcase to fail old code and pass new code?

lguohan: need to add a vs test case.

shi-su: A vs test case for this particular issue could be challenging form two aspects:
1. It requires a specific bulk pattern to happen, but we do not have control over how the operations are partition into bulks. Constantly reproducing the issue in vs test could be a problem.
2. The issue is transient, it will get resolved itself and the only trace would be the log messages of failed attempts of nexthop group removals. Basically, there is a limited sign of the issue ever happened.
I think it needs some thinking in developing this test case.

orchagent/routeorch.cpp
qiluo-msft: Keep old if-statement, I think it will keep the `bulkNhgReducedRefCnt` in small size.

---
In reply to: [521673770](https://github.com/Azure/sonic-swss/pull/1501#discussion_r521673770) [](ancestors = 521673770)

orchagent/routeorch.cpp
```diff
@@ -22,6 +22,8 @@ extern CrmOrch *gCrmOrch;
 
 const int routeorch_pri = 5;
 
+std::set<NextHopGroupKey> bulkNhgReducedRefCnt;
```
qiluo-msft, line 25: >bulkNhgReducedRefCnt [](start = 26, length = 20)

Change to a RouteOrch member variable?

lguohan:
is this affecting 201911 as well?

orchagent/routeorch.cpp
lguohan: i agree with you. I think maybe we need to look into the googletest framework for such sequencing testing. the tests under tests/mock_tests

shi-su:
> is this affecting 201911 as well?

This should not affect 201911 since the bulk mode is not in 201911

qiluo-msft (approved):
LGTM

PR #1514, opened 2020-11-20
[dvs] Use DB namespace to construct DVSDatabase objects

daall:

**What I did**
I updated the `DVSDatabase` to refer to DBs by name and depend on the database config as opposed to manually providing the IDs and redis socket path.

**Why I did it**
This is the preferred way of addressing databases now, and it's also a much cleaner abstraction than ID/redis socket.

**How I verified it**
Re-ran the tests w/ this change. Also verified that concurrent tests work as expected.

**Details if related**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -79,8 +79,8 @@ def random_string(size=4, chars=string.ascii_uppercase + string.digits):
 
 
 class AsicDbValidator(DVSDatabase):
-    def __init__(self, db_id: int, connector: str):
-        DVSDatabase.__init__(self, db_id, connector)
+    def __init__(self):
+        DVSDatabase.__init__(self, "ASIC_DB")
         self._wait_for_asic_db_to_initialize()
         self._populate_default_asic_db_values()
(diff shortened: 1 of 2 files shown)


daall:
retest this please

tests/conftest.py
```diff
@@ -361,6 +361,18 @@ def __init__(
...
 
+        # Initialize the database configuration
+        with open("/var/run/redis/sonic-db/database_config.json") as f:
+            db_conf = json.load(f)
+
+        db_conf["INSTANCES"]["redis"]["unix_socket_path"] = self.redis_sock
+        vs_db_conf_file = f"/tmp/database_config-{self.pid}.json"
+
+        with open(vs_db_conf_file, "w") as f:
+            json.dump(db_conf, f)
```
qiluo-msft, line 372: Why not create a vs_db_conf_file under `tests` directory, as source code, and initialize from it directly?

qiluo-msft (commented):
As comment

PR #1516, opened 2020-11-21
[201911 sonic-swss] Flushing FDB entries before removing BridgePort

madhanmellanox:
**What I did**
I added coded in sonic-swss orchagent to flush the FDB entries corresponding to the bridge port before removing the bridgeport.

**Why I did it**
I did it because if we don't flush the FDB entries before removing the bridge port and try to remove the bridge port, the bridge port removal results in a failure and the port operational status is set to down.

**How I verified it**
I created a VLAN interface, assigned IP address to it and added 2 Ethernet ports as tagged members to it. I waited for the switch to add its neighbors through these bridge ports. Then I removed VLAN membership of one of the bridge ports and see it is successful and the operational state of the port does not go down on removing the VLAN membership.

**Details if related**
flushFDBEntries() function is introduce in orchagent/fdborch.cpp and called from orchagent/portsorch.cpp

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -416,6 +416,46 @@ void FdbOrch::doTask(NotificationConsumer& consumer)
     }
 }
 
+void FdbOrch::flushFDBEntries(sai_object_id_t bridge_port_oid,
+                              sai_object_id_t vlan_oid)
+{
+    vector<sai_attribute_t>    attrs;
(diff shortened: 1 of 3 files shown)


madhanmellanox:
@prsunny can you please review this?

madhanmellanox:
retest this please

lguohan (approved):

PR #1519, opened 2020-11-25
[dvs/copp] Adjust policer checks in VS tests for default COPP policers

daall:

**What I did**
I updated the DVS tests to take into account the default COPP configuration when checking for policers.

**Why I did it**
https://github.com/Azure/sonic-buildimage/pull/4861 added a COPP config to the DVS and this is causing a few policer related tests to fail in PR builds across repos b/c there is a mismatch between the expected number of policers and the actual number.

**How I verified it**
Re-ran the failing tests locally w/ my changes.

**Details if related**


abdosi (approved):

daall:
known infra issue, not related to this PR

PR #1523, opened 2020-12-01
[pfcwd] Update PFC storm detection logic for Mellanox platforms

volodymyrsamotiy:
* Use "PFC duration" counters in micro seconds instead of quanta

**What I did**
Updated PFCWD storm detection logic for Mellanox platforms in order to use "PFC duration" counters in micro seconds instead of quanta.

Depends on below PRs so they should be merged first.
- https://github.com/Azure/sonic-sairedis/pull/731
- https://github.com/Azure/sonic-buildimage/pull/6156

**Why I did it**
SONiC PFCWD logic requires "pfc duration" value in micro seconds but in SAI it was provided as quanta of time. So it required additional conversion which used speed value to do such conversion and it could cause PFCWD to detect storm on operationally down port in case of link flapping.

Now there are new SAI attributes that provide "pfc duration" in micro seconds so PCWD storm detection logic is updated in order to use this new "pfc duration" counters. Such algorithm change helps to avoid false PFC storm detection in case of link flapping because conversion is not needed anymore.

**How I verified it**
1. Ran all PFCWD tests from https://github.com/Azure/sonic-mgmt/tree/master/tests/pfcwd and verified all passed.
2. Manual test: manually start storm on fanout and manually put port down on fanout few times. Verified that no storm was detected when link was down

**Details if related**
N/A

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,0 +1,41 @@
+path_classifiers:
+  test:
+    exclude: tests
+extraction:
+  python:
+    python_setup:
+      version: "2"
+  cpp:
+    prepare:
+      packages:
+      - libxml-simple-perl
+      - aspell
+      - aspell-en
+      - libhiredis-dev
+      - libnl-3-dev
+      - libnl-genl-3-dev
+      - libnl-route-3-dev
+      - libnl-nf-3-dev
+      - libzmq3-dev
+      - libzmq5
+      - swig3.0
+      - libpython2.7-dev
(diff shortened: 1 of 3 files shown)


neethajohn:
LGTM. Can you update the PR description and also specify what testing was done?

neethajohn (approved):

qiluo-msft:
LGTM analysis C/C++ is not working for 201811 branch right now. Ignore.

PR #1534, opened 2020-12-09
handled update() function of fdb orchagent for FDB FLUSH event

madhanmellanox:
**What I did**
I added code to handle update() function of FDB orchagent for the FDB FLUSH event.

**Why I did it**
I added code because earlier the FDB FLUSH event was not handled not fully, it was saying unsupported. So, I handled the event as like in master branch.

**How I verified it**
Adding a port as VLAN member, learning FDB entries for the port and then removing VLAN membership was causing ERROR logs in SWSS. With this fix, no more error logs were observed.

**Details if related**
Changeset is in orchagent/fdborch.cpp and orechagent/fdborch.h

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -159,8 +159,31 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
         break;
 
     case SAI_FDB_EVENT_FLUSHED:
-        if (bridge_port_id == SAI_NULL_OBJECT_ID && entry->bv_id == SAI_NULL_OBJECT_ID)
+
+        SWSS_LOG_INFO("FDB Flush event received: [ %s , 0x%" PRIx64 " ], \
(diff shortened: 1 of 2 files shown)


madhanmellanox:
retest this please

orchagent/fdborch.cpp
```diff
@@ -307,10 +355,11 @@ void FdbOrch::doTask(Consumer& consumer)
                 }
             }
 
+	    entry.port_name = port;
```
prsunny, line 358: Please align

prsunny (approved):
lgtm. This is directly a port from master, correct?

orchagent/fdborch.cpp
madhanmellanox: Fixed in next commit.

madhanmellanox:
> lgtm. This is directly a port from master, correct?

Yes, this is a direct port from master branch.

lguohan:
can you identify which master branch commit is it coming from?

PR #1537, opened 2020-12-11
[fgnhgorch] Fg Nhg link handling

anish-n:
**What I did**
Added fine grained ecmp link oper state handling changes

**Why I did it**
To adjust ECMP appropriately on link oper change

**How I verified it**

**Details if related**
HLD update associated with change: https://github.com/Azure/SONiC/pull/693

### DIFF
--- a/orchagent/fgnhgorch.cpp
+++ b/orchagent/fgnhgorch.cpp
@@ -388,6 +388,9 @@ bool FgNhgOrch::validNextHopInNextHopGroup(const NextHopKey& nexthop)
             }
 
             m_neighOrch->increaseNextHopRefCount(nexthop);
+
+            SWSS_LOG_INFO("FG nh %s for prefix %s is up",
+                    nexthop.to_string().c_str(), route_table.first.to_string().c_str());
         }
     }
 
@@ -455,6 +458,9 @@ bool FgNhgOrch::invalidNextHopInNextHopGroup(const NextHopKey& nexthop)
             }
 
             m_neighOrch->decreaseNextHopRefCount(nexthop);
+
+            SWSS_LOG_INFO("FG nh %s for prefix %s is down",
+                    nexthop.to_string().c_str(), route_table.first.to_string().c_str());
         }
     }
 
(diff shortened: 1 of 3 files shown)


anish-n:
retest this please

orchagent/fgnhgorch.cpp
```diff
@@ -1396,23 +1534,63 @@ bool FgNhgOrch::doTaskFgNhgMember(const KeyOpFieldsValuesTuple & t)
...
             /* query and check the next hop is valid in neighOrcch */
             if (!m_neighOrch->hasNextHop(nhk))
             {
                 SWSS_LOG_INFO("Nexthop %s is not resolved yet", nhk.to_string().c_str());
-                return true;
             }
-
-            /* add next-hop into SAI group */
-            if (!validNextHopInNextHopGroup(nhk))
+            else if (link_oper)
```
prsunny, line 1583: is the flow here handled for the case where port doesn't exists (get_port failed)?

orchagent/fgnhgorch.cpp
```diff
@@ -7,6 +7,9 @@
 #include "crmorch.h"
 #include <array>
 
+#define LINK_DOWN    false
```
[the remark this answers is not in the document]
anish-n: Makes sense, will change LINK_DOWN to 0 and LINK_UP to 1

orchagent/fgnhgorch.cpp
anish-n: If get_port fails then link_oper will have the default value of 1 per: https://github.com/Azure/sonic-swss/blob/9a586c6cdbc95e2b75e429964f19047d68c9b751/orchagent/fgnhgorch.cpp#L1498

So it will end up going into the else if and call into validNextHopInNextHopGroup(assuming hasNextHop is true).

anish-n:
retest this please

prsunny (approved):

anish-n:
retest this please

anish-n:
retest this please

anish-n:
retest this please

PR #1538, opened 2020-12-11
[fgnhgorch] add warm reboot support for fgnhg

weixchen1215:
**What I did**
1. Add bake() functions in FgnhgOrch
bake() function retrieves the state database information
so that when the new route is added, fgnhg orch is able to recover
the nexthop and bucket the same as before warm reboot.
2. Add priority to fgnhg tables so that fgnhg tables can be loaded prior to route table.
**Why I did it**
After warm reboot, it is expected that the fgnhg local structure and relationship between nexthop and bucket is retained.
**How I verified it**
Verified using VS swss test cases by stop and restart swss.

### DIFF
--- a/orchagent/fgnhgorch.cpp
+++ b/orchagent/fgnhgorch.cpp
@@ -6,6 +6,7 @@
 #include "swssnet.h"
 #include "crmorch.h"
 #include <array>
+#include <algorithm>
 
 #define LINK_DOWN    0
 #define LINK_UP      1
@@ -20,7 +21,7 @@ extern RouteOrch *gRouteOrch;
 extern CrmOrch *gCrmOrch;
 extern PortsOrch *gPortsOrch;
 
-FgNhgOrch::FgNhgOrch(DBConnector *db, DBConnector *appDb, DBConnector *stateDb, vector<string> &tableNames, NeighOrch *neighOrch, IntfsOrch *intfsOrch, VRFOrch *vrfOrch) :
+FgNhgOrch::FgNhgOrch(DBConnector *db, DBConnector *appDb, DBConnector *stateDb, vector<table_name_with_pri_t> &tableNames, NeighOrch *neighOrch, IntfsOrch *intfsOrch, VRFOrch *vrfOrch) :
         Orch(db, tableNames),
         m_neighOrch(neighOrch),
         m_intfsOrch(intfsOrch),
@@ -105,6 +106,39 @@ void FgNhgOrch::update(SubjectType type, void *cntx)
     }
 }
 
+bool FgNhgOrch::bake()
+{
+    SWSS_LOG_ENTER();
+
+    deque<KeyOpFieldsValuesTuple> entries;
+    vector<string> keys;
+    m_stateWarmRestartRouteTable.getKeys(keys);
+
+    SWSS_LOG_NOTICE("Warm reboot: recovering entry %lu from state", keys.size());
+
+    for (const auto &key : keys)
+    {
+        vector<FieldValueTuple> tuples;
+        m_stateWarmRestartRouteTable.get(key, tuples);
+
+        NextHopIndexMap nhop_index_map(tuples.size(), std::string());
+        for (const auto &tuple : tuples)
+        {
+            const auto index = stoi(fvField(tuple));
+            const auto nextHop = fvValue(tuple);
+
+            nhop_index_map[index] = nextHop;
+            SWSS_LOG_INFO("Storing next hop %s at index %d", nhop_index_map[index].c_str(), index);
+        }
+
+        // Recover nexthop with index relationship
+        m_recoveryMap[key] = nhop_index_map;
+
+        removeStateDbRouteEntry(key);
+    }
+
+    return Orch::bake();
+}
 
 /* calculateBankHashBucketStartIndices: generates the hash_bucket_indices for all banks
  * and stores it in fgNhgEntry for the group. 
@@ -189,6 +223,13 @@ void FgNhgOrch::setStateDbRouteEntry(const IpPrefix &ipPrefix, uint32_t index, N
         m_stateWarmRestartRouteTable.set(key, fvs);
     }
 
+}
+
+void FgNhgOrch::removeStateDbRouteEntry(const string& ipPrefix)
+{
+	SWSS_LOG_ENTER();
+
+	m_stateWarmRestartRouteTable.del(ipPrefix);
 }
 
 
@@ -881,6 +922,8 @@ bool FgNhgOrch::setNewNhgMembers(FGNextHopGroupEntry &syncd_fg_route_entry, FgNh
     SWSS_LOG_ENTER();
 
     sai_status_t status;
+    bool is_warm_reboot = false;
+    auto nexthopsMap = m_recoveryMap.find(ipPrefix.to_string());
     for (uint32_t i = 0; i < fgNhgEntry->hash_bucket_indices.size(); i++) 
     {
         uint32_t bank = i;
@@ -913,11 +956,33 @@ bool FgNhgOrch::setNewNhgMembers(FGNextHopGroupEntry &syncd_fg_route_entry, FgNh
             return false;
         }
 
+        // recover state before warm reboot
+        if (nexthopsMap != m_recoveryMap.end())
+        {
+            is_warm_reboot = true;
+        }
+
+        SWSS_LOG_INFO("Warm reboot is set to %d", is_warm_reboot);
+
         for (uint32_t j = fgNhgEntry->hash_bucket_indices[i].start_index;
                 j <= fgNhgEntry->hash_bucket_indices[i].end_index; j++)
         {
-            NextHopKey bank_nh_memb = bank_member_changes[bank].nhs_to_add[j % 
-                bank_member_changes[bank].nhs_to_add.size()];
+            NextHopKey bank_nh_memb;
+            if (is_warm_reboot)
+            {
+                bank_nh_memb = nexthopsMap->second[j];
+                SWSS_LOG_INFO("Recovering nexthop %s with bucket %d", bank_nh_memb.ip_address.to_string().c_str(), j);
+                // case nhps in bank are all down
+                if (fgNhgEntry->next_hops[bank_nh_memb.ip_address] != i)
+                {
+                    syncd_fg_route_entry.inactive_to_active_map[i] = fgNhgEntry->next_hops[bank_nh_memb.ip_address];
+                }
+            }
+            else
+            {
+                bank_nh_memb = bank_member_changes[bank].nhs_to_add[j %
+                    bank_member_changes[bank].nhs_to_add.size()];
+            }
 
             // Create a next hop group member
             sai_attribute_t nhgm_attr;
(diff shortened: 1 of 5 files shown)


prsunny:
Approach lgtm. Will wait for @qiluo-msft 's review

orchagent/fgnhgorch.h
```diff
@@ -95,6 +100,10 @@ class FgNhgOrch : public Orch
     FgPrefixOpCache m_fgPrefixAddCache;
     FgPrefixOpCache m_fgPrefixDelCache;
 
+    // warm reboot support for recovery
+    // < ip_prefix, < ip_next_hop, HashBuckets>>
+    map<string, NextHopIndexMap> m_recoveryMap;
```
anish-n, line 105: nit: can we define ```typedef map<string, NextHopIndexMap>WarmBootRecoveryMap; ``` above and have this written as ```WarmBootRecoveryMap m_recoveryMap;``` ?

tests/test_fgnhg.py
```diff
@@ -325,6 +326,45 @@ def test_route_fgnhg(self, dvs, testlog):
...
+        nh_memb_exp_count = {"10.0.0.1":10,"10.0.0.3":10,"10.0.0.5":10,"10.0.0.7":10,"10.0.0.9":10,"10.0.0.11":10}
+        program_route_and_validate_fine_grained_ecmp(app_db.db_connection, asic_db, state_db, ip_to_if_map,
+                            fg_nhg_prefix, nh_memb_exp_count, nh_oid_map, nhgid, bucket_size)
+
+        #############################################################################
+        #                                                                           #
+        #                        swss Warm-Restart Testing END                      #
+        #                            get to normal state                            #
+        #############################################################################
+
```
anish-n, line 366: Could we add a warm reboot test case which utilizes the inactive_to_active_map, to make sure that a case where all members in a bank are down will work fine after warm reboot in terms of in-memory state? 

We can do the following for the test case:
1. Create Fine Grained ECMP route with only members of 1 bank present(...7, ...9, ...11)
2. Run warm reboot
3. Check everything is same post warm reboot
4. Add a route which changes members in 1 bank, eg: (...9, ...11), check everything is configured fine
5. Add a route which adds a member in a previously down bank, eg(...1, ...9, ...11), check everything is configured fine

orchagent/fgnhgorch.h
weixchen1215: Make the change

tests/test_fgnhg.py
weixchen1215: test warm reboot with only one bank up and vs test works.

orchagent/orchdaemon.cpp
```diff
@@ -122,10 +122,12 @@ bool OrchDaemon::init()
     gIntfsOrch = new IntfsOrch(m_applDb, APP_INTF_TABLE_NAME, vrf_orch);
     gNeighOrch = new NeighOrch(m_applDb, APP_NEIGH_TABLE_NAME, gIntfsOrch, gFdbOrch, gPortsOrch);
 
-    vector<string> fgnhg_tables = {
-        CFG_FG_NHG,
-        CFG_FG_NHG_PREFIX,
-        CFG_FG_NHG_MEMBER
+    const int fgnhgorch_pri = 15;
```
qiluo-msft, line 125: >const [](start = 4, length = 5)

static const

orchagent/fgnhgorch.cpp
```diff
@@ -113,6 +147,13 @@ void FgNhgOrch::setStateDbRouteEntry(const IpPrefix &ipPrefix, uint32_t index, N
 
 }
 
+void FgNhgOrch::removeStateDbRouteEntry(const string& ipPrefix)
```
qiluo-msft, line 150: >removeStateDbRouteEntry [](start = 16, length = 23)

Better to remove the function and just use `del()` in the original caller.

orchagent/fgnhgorch.cpp
```diff
@@ -820,11 +863,33 @@ bool FgNhgOrch::setNewNhgMembers(FGNextHopGroupEntry &syncd_fg_route_entry, FgNh
             return false;
         }
 
+        // recover state before warm reboot
+        if (nexthopsMap != m_recoveryMap.end())
+        {
+            is_warm_reboot = true;
```
qiluo-msft, line 869: >is_warm_reboot [](start = 12, length = 14)

Is it better to check m_recoveryMap.empty()? If no, then it is warm-reboot.

orchagent/fgnhgorch.cpp
```diff
@@ -868,6 +933,11 @@ bool FgNhgOrch::setNewNhgMembers(FGNextHopGroupEntry &syncd_fg_route_entry, FgNh
         }
     }
 
+    if (is_warm_reboot)
```
qiluo-msft, line 936: >is_warm_reboot [](start = 8, length = 14)

If `nexthopsMap != m_recoveryMap.end()`, then you will not erase. Is it a bug?

tests/test_fgnhg.py
```diff
@@ -175,6 +175,21 @@ def startup_link(dvs, db, port):
     dvs.servers[port].runcmd("ip link set up dev eth0") == 0
     db.wait_for_field_match("PORT_TABLE", "Ethernet%d" % (port * 4), {"oper_status": "up"})
 
+def run_warm_reboot(dvs):
+    dvs.runcmd("config warm_restart enable swss")
+
+    # Stop swss before modifing the configDB
+    dvs.stop_swss()
+    time.sleep(1)
```
qiluo-msft, line 183: >time.sleep(1) [](start = 4, length = 13)

What if no sleep?

tests/test_fgnhg.py
```diff
@@ -175,6 +175,21 @@ def startup_link(dvs, db, port):
...
+
+    # start to apply new port_config.ini
+    dvs.start_swss()
+    dvs.runcmd(['sh', '-c', 'supervisorctl start neighsyncd'])
+    dvs.runcmd(['sh', '-c', 'supervisorctl start restore_neighbors'])
+
+    # Enabling some extra logging for validating the order of orchagent
+    dvs.runcmd("swssloglevel -l INFO -c orchagent")
+    time.sleep(5)
```
qiluo-msft, line 192: >time.sleep(5) [](start = 4, length = 13)

What if no sleep?

qiluo-msft (changes requested):
As commented

orchagent/fgnhgorch.cpp
weixchen1215: if nexthopsMap != m_recoveryMap.end(), it a warm restart case. after recover from the map, we will erase the entry from recovery. nexthopsMap == m_recoveryMap.end() meaning this route is a new entry, will handle in a normal way.

orchagent/fgnhgorch.cpp
weixchen1215: auto nexthopsMap = m_recoveryMap.find(ipPrefix.to_string()); this checks if the route-addition is required to be recover after warm reboot. so checking if the route prefix can be found in m_recoveryMap. 
I think we should check the prefix existing instead of only recoverymap is empty. what do you think?

orchagent/fgnhgorch.cpp
weixchen1215: removed the function removeStateDbRouteEntry

orchagent/orchdaemon.cpp
weixchen1215: changed

tests/test_fgnhg.py
weixchen1215: removed the sleep and test passed

tests/test_fgnhg.py
weixchen1215: removed sleep

orchagent/fgnhgorch.cpp
```diff
@@ -790,6 +823,8 @@ bool FgNhgOrch::setNewNhgMembers(FGNextHopGroupEntry &syncd_fg_route_entry, FgNh
     SWSS_LOG_ENTER();
 
     sai_status_t status;
+    bool is_warm_reboot = false;
```
qiluo-msft, line 826: >is_warm_reboot [](start = 9, length = 14)

Better to use camelCaseConvention for local var.

weixchen1215:
retest this please

qiluo-msft (approved):
LGTM. Please also check with other reviewers.

weixchen1215:
retest this please

tests/test_fgnhg.py
```diff
@@ -306,6 +320,20 @@ def test_route_fgnhg(self, dvs, testlog):
         asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_NEXT_HOP", asic_nh_count + 1)
         nh_oid_map = get_nh_oid_map(asic_db)
         # Now that ARP was resolved, 10.0.0.7 should be added as a valid fg nhg member
+        nh_memb_exp_count = {"10.0.0.7":20,"10.0.0.9":20,"10.0.0.11":20}
+        validate_fine_grained_asic_n_state_db_entries(asic_db, state_db, ip_to_if_map,
+                                nh_memb_exp_count, nh_oid_map, nhgid, bucket_size)
+        
+        run_warm_reboot(dvs)
```
lguohan, line 327: this test_route_fgnhg test is becoming too large now, please split warm reboot into a separate test.

weixchen1215: seperate into a new test for warm reboot

weixchen1215:
retest this please

weixchen1215:
retest this please

weixchen1215:
retest this please

weixchen1215:
retest this please

qiluo-msft (approved):
LGTM. Please also check with other reviewers.

prsunny (approved):
lgtm

anish-n (approved):

tests/test_fgnhg.py
```diff
@@ -626,12 +643,126 @@ def test_route_fgnhg(self, dvs, testlog):
...
+        # Wait for the software to receive the entries
+        time.sleep(1)
+
+        ps = swsscommon.ProducerStateTable(app_db.db_connection, ROUTE_TB)
+        fvs = swsscommon.FieldValuePairs([("nexthop","10.0.0.7,10.0.0.9,10.0.0.11"),
+            ("ifname", "Vlan16,Vlan20,Vlan24")])
+        ps.set(fg_nhg_prefix, fvs)
+        # No ASIC_DB entry we can wait for since ARP is not resolved yet,
+        # We just use sleep so that the sw receives this entry
+        time.sleep(1)
```
lguohan, line 706: can you get rid of such sleep and wait_till?

anish-n: This sleep should be harmless since there isn't any state/asic db obj to check for after the sleep. This logic is the same on master today, and have not seen flakiness with test because of sleep. I added the sleep so that the sw can receive the entries and trigger a retry logic for the route. Even if I remove the sleep the outcome of this test is the same. The advantage of sleep is that it gives more test coverage here by allowing route retry logic to kick in, but even if retry does not occur, there should not be flakiness occurring due to it.

lguohan: what about the sleep above?

PR #1543, opened 2020-12-14
Fix LGTM C++ build for other 201911 PR

qiluo-msft:
**What I did**
Add the missing lgtm.yml

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -5,3 +5,35 @@ extraction:
   python:
     python_setup:
       version: "2"
+  cpp:
+    prepare:
+      packages:
+      - libxml-simple-perl
+      - aspell
+      - aspell-en
+      - libhiredis-dev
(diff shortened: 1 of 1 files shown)


lgtm.yml
```diff
@@ -0,0 +1,41 @@
...
+      - libzmq3-dev
+      - libzmq5
+      - swig3.0
+      - libpython2.7-dev
+      - libgtest-dev
+      - dh-exec
+      - doxygen
+      - graphviz
+    after_prepare:
+      - git clone https://github.com/Azure/sonic-swss-common; pushd sonic-swss-common; ./autogen.sh; fakeroot dpkg-buildpackage -us -uc -b; popd
```
daall, line 28: Do we need to add a `git checkout 201911` step here as well?

PR #1547, opened 2020-12-15
[gearbox] Add support for "hwinfo" field

Nakrez:
**What I did**

This PR adds support for a new "hwinfo" attribute in gearbox_config.json. The field is propagated by syncd to the SAI by setting SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO.

**Why I did it**

This is necessary to propagate the MDIO interface & phy_id that should be used by the SAI driver for a particular gearbox.

**How I verified it**

Verified on an internal Arista build that the field from the configuration was propagated all the way to the SAI.

### DIFF
--- a/orchagent/saihelper.cpp
+++ b/orchagent/saihelper.cpp
@@ -15,6 +15,7 @@ extern "C" {
 #include <tuple>
 #include <vector>
 #include <linux/limits.h>
+#include <net/if.h>
 #include "timestamp.h"
 #include "sai_serialize.h"
 #include "saihelper.h"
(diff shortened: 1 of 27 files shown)


Nakrez:
retest vs please

orchagent/saihelper.cpp
```diff
@@ -286,9 +291,18 @@ sai_status_t initSaiPhyApi(swss::gearbox_phy_t *phy)
     attr.value.u32 = 0;
     attrs.push_back(attr);
 
+    if (phy->hwinfo.length() > HWINFO_MAX_SIZE)
```
lguohan, line 294: if i understand correctly, the comparison is really for interface_name part, it is not the combined string. for example, you can have interface_name > ifnamsiz, but phy_id less than 5 bytes, so in total it is still less than HWINFO_MAC_SIZE, but the phy driver will see error since the interface_name exceed the limit?

Nakrez: That's a good point. I only thought about memory constraints, not about sanity of the data. That's why the check is done this way. I'll post a new version with a proper sanity check for the sub components of hwinfo.

lguohan (approved):

PR #1550, opened 2020-12-16
[CRM] Typecast to unit64_t to avoid divide by 0 during overflow

prsunny:
**What I did**
Static cast to uint64_t

**Why I did it**
Fix orchagent crash in case of overflow resulting in divide by 0

**How I verified it**
Wrote a test program to simulate

**Details if related**


prsunny:
retest this please

lguohan (approved):

PR #1553, opened 2020-12-16
[FPMSYNCD/FDBSYNCD] EVPN Type-5 route removing prefix-len for host route and removing junk character present in the mac

kishorekunal01:
[FPMSYNCD/FDBSYNCD] 
To make EVPN prefix route inline with the existing code, removing the prefix-len for host routes(/32 and /128).
Some Junk character are present in the mac and port name while config MAC in the kernel.

What I did
If the prefix is a host route don't append the prefix-len while creating the prefix.
With the review comment fix one place memset was removed due to this MAC and port name is containing trailing junk character. Due to this MAC and port name has some Junk character and Kernel is returning error.

Why I did it
Remove the prefix-len for host-routes (/32 & /128)
Removed junk char while installing the FDB entry into Kernel

How I verified it
Verified using vsonic


1549.diff
```diff
@@ -0,0 +1,38 @@
+diff --git a/fdbsyncd/fdbsync.cpp b/fdbsyncd/fdbsync.cpp
```
prsunny, line 1: please remove this file

kishorekunal01: Done. My bad added by mistake.

prsunny (approved):

PR #1555, opened 2020-12-17
Add support for Inband VLan

ysmanman:
The PR adds Inband VLan support.

**What I did**

- intfsorch change to configure IP address for Inband Vlan in kernel.
- neighorch change to skip adding remote neighbors attached to Inband Vlan.
- portsorch change to add Inband Vlan.

**Why I did it**

Support inband VLan

**How I verified it**

**Details if related**

This PR has some dependency on PR #1431  and some minor change is needed once that PR is merged.


anshuv-mfst:
@vganesan-nokia  (Nokia) could you please review the PR, thanks.

lguohan (changes requested):
i do not see any swss vstest being added.

orchagent/portsorch.cpp
```diff
@@ -4631,3 +4651,97 @@ bool PortsOrch::initGearboxPort(Port &port)
     return true;
 }
 
+/*
+ * The following method is called from setVoqInbandIntf() introduced in the following PR:
+ * https://github.com/Azure/sonic-swss/pull/1431
+ */
+bool PortsOrch::addInbandVlan(const string &alias, const string &type)
```
vganesan-nokia, line 4658: Thanks for the comments. PR 1431 is merged. setVoqInbandIntf() can be modified to call this api if the inband type is "vlan". Please rebase and modify setVoqInbandIntf() to make this complete and the comments can be removed.

orchagent/portsorch.cpp
```diff
@@ -4631,3 +4651,97 @@ bool PortsOrch::initGearboxPort(Port &port)
...
+    {
+        SWSS_LOG_ERROR("Failed to add inband VLAN %s", alias.c_str());
+        return false;
+    }
+
+    if (!getPort(alias, inbandVlan))
+    {
+        SWSS_LOG_ERROR("Failed to get port object of inband VLAN %s", alias.c_str());
+        return false;
+    }
```
vganesan-nokia, line 4689: Can we not do this vlan addition processing via vlan configuration by having entry in VLAN table in config_db? So that this will be similar to inband port type. For port type, the VOQ_INBAND_INTERFACE configuration is just to designate a port or vlan for interface for cpu-to-cpu communication. We actually do not configure the port or vlan itself.

vganesan-nokia (changes requested):

ysmanman:
> i do not see any swss vstest being added.

Sure, I will add a test.

orchagent/neighorch.cpp
```diff
@@ -428,6 +428,35 @@ void NeighOrch::doTask(Consumer &consumer)
...
+         * neighsyncd adds the kernel neigh to neigh table in APPL_DB again.
+         * The following change is to skip adding these remote neighbor attached
+         * to inband Vlan.
+         *
+         * The collection m_remoteNeigh is added to track remote neighbor. It's
+         * updated in the following method introduced in the above PR:
+         *  
+         * doVoqSystemNeighTask()
+         */
+        if (gPortsOrch->isInbandPort(alias))
```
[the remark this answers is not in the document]
ysmanman: Thanks. I will rebase and remove comment.

orchagent/portsorch.cpp
ysmanman: Thanks. I will rebase and remove the comment.

orchagent/portsorch.cpp
```diff
@@ -3420,22 +3433,28 @@ bool PortsOrch::addBridgePort(Port &port)
...
-    /* And with hardware FDB learning mode set to HW (explicit default value) */
-    attr.id = SAI_BRIDGE_PORT_ATTR_FDB_LEARNING_MODE;
-    auto found = learn_mode_map.find(port.m_learn_mode);
-    if (found == learn_mode_map.end())
+    if (port.m_type != Port::SYSTEM)
     {
-        attr.value.s32 = SAI_BRIDGE_PORT_FDB_LEARNING_MODE_HW;
+        attr.id = SAI_BRIDGE_PORT_ATTR_ADMIN_STATE;
+        attr.value.booldata = true;
+        attrs.push_back(attr);
```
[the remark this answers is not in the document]
ysmanman: Sounds good. Will fix.

orchagent/neighorch.cpp
```diff
@@ -428,6 +428,35 @@ void NeighOrch::doTask(Consumer &consumer)
...
+        {
+            Port ibport;
+            gPortsOrch->getInbandPort(ibport);
+            if(ibport.m_type == Port::VLAN)
+            {
+                 /* Only create neighbor attached to inband Vlan */
+                if (m_remoteNeigh.find(ip_address) != m_remoteNeigh.end())
```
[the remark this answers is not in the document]
ysmanman: Yeah.  iBGP peer can be resolved over inband Vlan. This kind of inband vlan neighbor needs to be created. But remote neighbors can be skipped.

orchagent/portsorch.cpp
ysmanman: We thought about this too. But inband VLAN is not exactly the same as regular SONiC VLANs. For example, the inband VLAN members are cpu ports, which do not have host intfs being created today.  Therefore, inband VLAN processing is different from VLAN processing.

orchagent/neighorch.cpp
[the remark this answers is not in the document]
ysmanman: Fixed.

orchagent/portsorch.cpp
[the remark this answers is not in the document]
ysmanman: Fixed.

orchagent/neighorch.cpp
```diff
@@ -541,6 +563,26 @@ void NeighOrch::doTask(Consumer &consumer)
...
+                 * and the interface that the neigh is attached is inband Vlan.
+                 * neighsyncd adds the kernel neigh to neigh table in APPL_DB again.
+                 * The following change is to skip adding these remote neighbor attached
+                 * to inband Vlan.
+                 */
+                if (m_remoteNeigh.find(ip_address) != m_remoteNeigh.end())
+                {
+                    it = consumer.m_toSync.erase(it);
+                    continue;
+                }
```
vganesan-nokia, line 582: This logic can go as else part (after line 560). The getPortOrch->isInbandPort(alias) and getPortsOrch->getInbandPort(ibport) are redundantly called multiple times.

orchagent/neighorch.cpp
```diff
@@ -1203,6 +1249,19 @@ void NeighOrch::voqSyncAddNeigh(string &alias, IpAddress &ip_address, const MacA
...
+        SWSS_LOG_ERROR("Failed to get inband port/VLAN");
+        return;
+    }
+
+    if (inbandPort.m_type == Port::VLAN && inbandPort.m_alias == alias)
+    {
+        SWSS_LOG_NOTICE("Skip inband VLAN neigh: %s", ip_address.to_string().c_str());
+        return;
+    }
+
```
vganesan-nokia, line 1264: It will be great if comments are added for this block of code saying that "inband neighbors are not synced to CHASSIS_APP_DB for inband interface type VLAN since the inband neigibors are learned via arp learning within inband VLAN"

orchagent/intfsorch.cpp
```diff
@@ -1357,6 +1381,39 @@ void IntfsOrch::doTask(SelectableTimer &timer)
...
+         {
+             SWSS_LOG_ERROR("Command '%s' failed with rc %d", cmd.str().c_str(), ret);
+         }
+         else
+         {
+             m_stateVlanTable->hset(inbandVlan.m_alias, "state", "ok");
+             SWSS_LOG_INFO("Added inband VLAN %s to STATE_VLAN_TABLE", inbandVlan.m_alias.c_str());
+         }
+
+         m_inbandVlanTimer->stop();
```
vganesan-nokia, line 1413: Do we want to stop the timer when there is failure in bringing up inband vlan in the kernel? Will it be a good idea to leave the timer running for re-tries?

vganesan-nokia (changes requested):

orchagent/neighorch.cpp
ysmanman: Thanks. Fixed.

orchagent/neighorch.cpp
ysmanman: Added the comment as suggested.

orchagent/intfsorch.cpp
ysmanman: Yes, you are right. Fixed.

vganesan-nokia (approved):

vganesan-nokia:
@ysmanman, thanks for addressing the comments. Approved the changes.

vganesan-nokia (approved):

orchagent/intfsorch.cpp
```diff
@@ -99,8 +102,20 @@ IntfsOrch::IntfsOrch(DBConnector *db, string tableName, VRFOrch *vrf_orch, DBCon
...
+    if (gMySwitchType == "voq")
     {
+        //STATE DB connection for setting state of the VoQ inband VLAN interface
+        unique_ptr<DBConnector> stateDb;
+        stateDb = make_unique<DBConnector>("STATE_DB", 0);
+        m_stateVlanTable = unique_ptr<Table>(new Table(stateDb.get(), STATE_VLAN_TABLE_NAME));
+
+        //Timer to bring up inband Vlan netdev
+        auto intervT = timespec { .tv_sec = INBAND_INTF_TIMER_SEC , .tv_nsec = 0 };
+        m_inbandVlanTimer = new SelectableTimer(intervT);
```
abdosi, line 114: why do we need timer ? Can you point to design doc that cover this ?

orchagent/portsorch.cpp
```diff
@@ -4905,6 +4922,88 @@ bool PortsOrch::getSystemPorts()
...
+    {
+        SWSS_LOG_ERROR("Failed to set host intf oper status for inband VLAN");
+        return false;
+    }
+
+    for (auto &it: m_portList)
+    {
+        Port port = it.second;
+        if (port.m_type != Port::SYSTEM ||
+            port.m_system_port_info.type == SAI_SYSTEM_PORT_TYPE_LOCAL ||
```
abdosi, line 4983: can local cpu port be covered in this same loop ? basically this will make code common w.r.t API addBridgePort() and addVlanMemeber()

orchagent/portsorch.cpp
```diff
@@ -3513,23 +3526,26 @@ bool PortsOrch::addBridgePort(Port &port)
...
-    {
-        attr.value.s32 = SAI_BRIDGE_PORT_FDB_LEARNING_MODE_HW;
-    }
-    else
+    if (port.m_type != Port::SYSTEM)
     {
-        attr.value.s32 = found->second;
+        /* Create a bridge port with admin status set to UP */
+        attr.id = SAI_BRIDGE_PORT_ATTR_ADMIN_STATE;
+        attr.value.booldata = true;
```
abdosi, line 3533: why we skip this for system port ?

orchagent/portsorch.cpp
```diff
@@ -3513,23 +3526,26 @@ bool PortsOrch::addBridgePort(Port &port)
...
+        attrs.push_back(attr);
+
+        /* And with hardware FDB learning mode set to HW (explicit default value) */
+        attr.id = SAI_BRIDGE_PORT_ATTR_FDB_LEARNING_MODE;
+        auto found = learn_mode_map.find(port.m_learn_mode);
+        if (found == learn_mode_map.end())
+        {
```
abdosi, line 3540: also any reason to skip this for system port ? code here is generic for add systemport as bridge port.
Yes for inband use case this not needed but i feel this code should be kept generic

orchagent/neighorch.cpp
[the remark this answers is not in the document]
abdosi: @vganesan-nokia why same cannot happen with Inband port ? iBGP neighbor can happen on inband port also ?

orchagent/neighorch.cpp
```diff
@@ -1203,6 +1242,21 @@ void NeighOrch::voqSyncAddNeigh(string &alias, IpAddress &ip_address, const MacA
...
+    if (!gPortsOrch->getInbandPort(inbandPort))
+    {
+        SWSS_LOG_ERROR("Failed to get inband port/VLAN");
+        return;
+    }
+
+    // Inband neighbors are not synced to CHASSIS_APP_DB for VLAN type inband interface
+    // since these neigibors are learned within inband VLAN.
+    if (inbandPort.m_type == Port::VLAN && inbandPort.m_alias == alias)
+    {
```
abdosi, line 1255: what is this case only for inband vlan ? Can you please explain which case it falls as per https://github.com/Azure/SONiC/blob/master/doc/voq/voq_hld.md ?

orchagent/neighorch.cpp
```diff
@@ -1240,7 +1294,15 @@ void NeighOrch::voqSyncAddNeigh(string &alias, IpAddress &ip_address, const MacA
     FieldValueTuple eiFv ("encap_index", to_string(attr.value.u32));
     attrs.push_back(eiFv);
 
-    FieldValueTuple macFv ("neigh", mac.to_string());
+    FieldValueTuple macFv;
+    if (inbandPort.m_type == Port::VLAN)
+    {
+        macFv = FieldValueTuple("neigh", gMacAddress.to_string());
```
abdosi, line 1300: why we are putting our own mac when sync to chassis app db ?  remote side will use this to program SAI . Why we need to send our own mac to remote side ? as per design doc also SAI neighbor tables uses actual mac for vlan inband case. can you please clarify.

abdosi (changes requested):
Added some changes and clarification

orchagent/neighorch.cpp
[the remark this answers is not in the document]
vganesan-nokia: The only neighbors that would be available on the inband port are the static neigbors we add for the remote neighbors (which should be skipped here since these are already added to SAI as part of doVoqSystemNeighTask()). Unlike inband vlan, inband port will not have any other neighbors. So no need to collect or verify whether the neihbor is remote or not. Just check the interface of the neighbor and if the interface is inband interface, skip the neighbor. 

For inband port except the ip address of inband port of my asic, all the iBGP neighbors (which are ip address of inband port of other asics) are remote neighbors. These are already programmed in SAI by doVoqSystemNeighTask() so we skip here. But inband vlan will have both the iBGP neighbors (resolved via ARP) which should be programmed  here and remote neighbors which are added statically to inband vlan and SAI prgrammed in doVoqSystemNeighTask(). Therefore for inband vlan case, we need to indentify the remote neighbors and skip.

orchagent/intfsorch.cpp
ysmanman: Since SAI create call is not synchronized, we cannot bring up inband vlan in kernel immediately after creating inband vlan with SAI call.  We use this timer to periodically check if it's ready to bring up  inband vlan in kernel or not. The timer is stopped once inband vlan is brought up in kernel successfully. I think this is implementation details, and thus is not covered in HLD.

orchagent/neighorch.cpp
ysmanman: In inband vlan case,  each asic (precisely its cpu port) participates in inband vlan. Each asic has its own inband IP address that belongs to the same inband vlan network. Each asic learns the inband IPs of other asics as neighbors. These neighbors are mainly used for the communication between SONiC instances like routing protocol peering.  Therefor, we added this change to not add inband neighbors into system neigh table.  This is described in inband vlan section (2.5.2) in voq HLD.

orchagent/neighorch.cpp
ysmanman: We need to install kernel neighbor entries for remote neighbors. In inband vlan case, the mac of the kernel neighbor entry must be the router mac of remote asic, where the remote neighbor was learnt. This was described in section 2.5.2.2 of voq HLD.  Use neighbor nbr-0-1 as an example shown in that section. nbr-0-1 is learnt by asic0. A kernel neighbor entry is installed for the neighbor in asic1 with asic0's mac. This is why local router mac is published in chassis app db here.  SAI  programming  for remote neighbor uses encap index,  which is an indirection of the actual mac of remote neighbor.

orchagent/portsorch.cpp
ysmanman: Both local and remote cpu ports need to be added into inband vlan. The port type of remote cpu ports is system port, and thus we have to add the check here to skip the processing that's not required for system port.

orchagent/portsorch.cpp
ysmanman: Both local and remote cpu ports need to be added into inband vlan. The port type of remote cpu ports is system port, and thus we have to add the check here to skip the processing that's not required for system port.

orchagent/portsorch.cpp
ysmanman: This is a good point. I will try to see if I can move local cpu port processing into the for loop.

ysmanman: Fixed as what you suggested.

ysmanman:
> Added some changes and clarification

@abdosi Thanks for review. I replied/resolved all your comments.

orchagent/neighorch.cpp
abdosi: thanks. looks like  i overlooked it

orchagent/neighorch.cpp
abdosi: thanks

orchagent/intfsorch.cpp
abdosi: @ysmanman how is this different w.r.t Regular VLAN IP Interface ? We don;t need timer there.

orchagent/intfsorch.cpp
```diff
@@ -1357,6 +1381,39 @@ void IntfsOrch::doTask(SelectableTimer &timer)
...
+     {
+         return;
+     }
+
+     string value;
+     const auto id = sai_serialize_object_id(inbandVlan.m_rif_id);
+     if (m_vidToRidTable->hget("", id, value))
+     {
+         stringstream cmd;
+         cmd << "ip link set " << inbandVlan.m_alias << " up";
```
prsunny, line 1399: We cannot do "ip link" command from orchagent. This approach has to be revisited.

orchagent/intfsorch.cpp
ysmanman: @abdosi There are few differences between regular Vlan and Inban Vlan:
0) The members of Inband Vlan are local and remote CPU ports, which don't have kernel devices created today.
1) Inband Vlan is a static configuration and imembership is unlikely user configurable.
These are the main reasons that Inband Vlan was implemented different from regular Vlans and therefore we added the timer.

orchagent/intfsorch.cpp
ysmanman: @prsunny can you elaborate why we cannot do this in orchagent?

prsunny: There is no kernel interactions from orchagent. In the current architecture, this is handled by the manager daemons.

lguohan:
> This PR has some dependency on PR #1431 and some minor change is needed once that PR is merged.
is the change adapted?

lguohan:
i still do not see any vs tests being added.

ysmanman:
> > This PR has some dependency on PR #1431 and some minor change is needed once that PR is merged.
> 
> is the change adapted?

@lguohan yes, this is already addressed.

ysmanman:
> i still do not see any vs tests being added.

@lguohan The plan was to re-use the same test added in inband port PR https://github.com/Azure/sonic-swss/pull/1602 (so we run the same test twice: one with inband port config and one with inband vlan config). inband port PR was merged recently. I am working on adding the test in this PR.

orchagent/intfsorch.cpp
ysmanman: @prsunny Thanks for the explanation. The kernel interaction was moved to intfmgr as suggested.

vganesan-nokia (dismissed):

ysmanman:
> > i still do not see any vs tests being added.
> 
> @lguohan The plan was to re-use the same test added in inband port PR #1602 (so we run the same test twice: one with inband port config and one with inband vlan config). inband port PR was merged recently. I am working on adding the test in this PR.

@lguohan The inband vlan test was added. We reused the inband port test. With the change in this PR,  the same test is run twice: one with inband port, and one with inband vlan.

orchagent/intfsorch.cpp
lguohan: sai call is synchronized now starting from 202012 release

anshuv-mfst:
@lguohan , @abdosi - could you please approve and merge (critical PR)

vganesan-nokia (dismissed):

orchagent/portsorch.cpp
abdosi: @michaelli10 Can you help. We can add system port as part of Bridge. Can we set SAI_BRIDGE_PORT_ATTR_ADMIN_STATE for Remote Systrem Port and what does it mean ?

abdosi:
where are we doing this ?

image

ysmanman:
> where are we doing this ?
> 
> image

@abdosi I checked just now. When programming inband vlan neighbor, orchagent passes the RIF of inband vlan to SAI and in HW, the neighbor entry is programmed to local CPU port.  It seems SAI/Sdk is picking up local CPU port when the rif is inband vlan.  So the HLD may need to be updated.

ysmanman:
> > where are we doing this ?
> > image
> 
> @abdosi I checked just now. When programming inband vlan neighbor, orchagent passes the RIF of inband vlan to SAI and in HW, the neighbor entry is programmed to local CPU port. It seems SAI/Sdk is picking up local CPU port when the rif is inband vlan. So the HLD may need to be updated.
> 
> I don't agree with this. SAI should not be aware of Special case How does it know given RIF is inband RIF ? How do we track this with SAI ?

@abdosi DNX SAI is not aware of inband RIF. The SAI behavior is for all Vlans. More details. When the RIF type of a neighbor is Vlan and the MAC is not learnt, SAI uses local CPU port as destination. MAC learning of inband Vlan is not enabled, therefore SAI picks up cpu port in this case.

abdosi:
> > > where are we doing this ?
> > > image
> > 
> > 
> > @abdosi I checked just now. When programming inband vlan neighbor, orchagent passes the RIF of inband vlan to SAI and in HW, the neighbor entry is programmed to local CPU port. It seems SAI/Sdk is picking up local CPU port when the rif is inband vlan. So the HLD may need to be updated.
> > I don't agree with this. SAI should not be aware of Special case How does it know given RIF is inband RIF ? How do we track this with SAI ?
> 
> @abdosi DNX SAI is not aware of inband RIF. The SAI behavior is for all Vlans. More details. When the RIF type of a neighbor is Vlan and the MAC is not learnt, SAI uses local CPU port as destination. MAC learning of inband Vlan is not enabled, therefore SAI picks up cpu port in this case.

Still does not look correct. This also is not standard SAI Implementation . Some SAI do Drop, some might Flood in VLAN. So we can not rely on SAI implementation.

@prsunny do we know if SAI has define standard model what should happen if VLAN RIF Mac is not learnt ?

prsunny:
> > > > where are we doing this ?
> > > > image
> > > 
> > > 
> > > @abdosi I checked just now. When programming inband vlan neighbor, orchagent passes the RIF of inband vlan to SAI and in HW, the neighbor entry is programmed to local CPU port. It seems SAI/Sdk is picking up local CPU port when the rif is inband vlan. So the HLD may need to be updated.
> > > I don't agree with this. SAI should not be aware of Special case How does it know given RIF is inband RIF ? How do we track this with SAI ?
> > 
> > 
> > @abdosi DNX SAI is not aware of inband RIF. The SAI behavior is for all Vlans. More details. When the RIF type of a neighbor is Vlan and the MAC is not learnt, SAI uses local CPU port as destination. MAC learning of inband Vlan is not enabled, therefore SAI picks up cpu port in this case.
> 
> Still does not look correct. This also is not standard SAI Implementation . Some SAI do Drop, some might Flood in VLAN. So we can not rely on SAI implementation.
> 
> @prsunny do we know if SAI has define standard model what should happen if for VLAN RIF Mac is not learnt ?

Agree with @abdosi , if mac is not learnt, some platforms may flood within VLAN.

orchagent/portsorch.cpp
michaelli10: @abdosi SAI_BRIDGE_PORT_ATTR_ADMIN_STATE is not currently supported for remote system port.

arlakshm:
@ysmanman, today we add addDirectedBroadcast neighbor entry for vlan. Do we need to skip that for inband vlan
https://github.com/Azure/sonic-swss/blob/6c4c9adda3621d1ae5adead1a91468d532ff792f/orchagent/intfsorch.cpp#L501-L504

abdosi:
Based on latest discussion with Arista this PR is put on hold as of now.

rlhui:
@ysmanman - if this is not needed, please close it. Thanks.

ysmanman:
> @ysmanman - if this is not needed, please close it. Thanks.

@rlhui yes, this is not needed any more. I will close it.

PR #1565, opened 2020-12-21
[fgnhgorch] Match mode changes for Fine Grained ECMP

anish-n:
**What I did**
Added Fine Grained ECMP match mode changes 

**Why I did it**
So that Fine Grained ECMP routes can be filtered via next-hop IPs(match_mode next-hop-based). The originally implemented solution filtered routes via the prefix, this continues to be supported via match_mode route-based

**How I verified it**

**Details if related**
https://github.com/Azure/SONiC/pull/727

### DIFF
--- a/orchagent/fgnhgorch.cpp
+++ b/orchagent/fgnhgorch.cpp
@@ -332,43 +332,6 @@ bool FgNhgOrch::removeFineGrainedNextHopGroup(FGNextHopGroupEntry *syncd_fg_rout
 }
 
 
-bool FgNhgOrch::createFineGrainedRouteEntry(FGNextHopGroupEntry &syncd_fg_route_entry, FgNhgEntry *fgNhgEntry,
-        sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const NextHopGroupKey &nextHops)
-{
-    SWSS_LOG_ENTER();
-    sai_route_entry_t route_entry;
-    sai_attribute_t route_attr;
-    route_entry.vr_id = vrf_id;
-    route_entry.switch_id = gSwitchId;
-    copy(route_entry.destination, ipPrefix);
-    route_attr.id = SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID;
-    route_attr.value.oid = syncd_fg_route_entry.next_hop_group_id;
-    sai_status_t status = sai_route_api->create_route_entry(&route_entry, 1, &route_attr);
-    if (status != SAI_STATUS_SUCCESS)
-    {
-        SWSS_LOG_ERROR("Failed to create route %s with next hop(s) %s",
-                ipPrefix.to_string().c_str(), nextHops.to_string().c_str());
-
-        /* Clean up the newly created next hop group entry */
-        if (!removeFineGrainedNextHopGroup(&syncd_fg_route_entry))
-        {
-            SWSS_LOG_ERROR("Failed to clean-up after route creation failure");
-        }
-        return false;
-    }
-
-    if (route_entry.destination.addr_family == SAI_IP_ADDR_FAMILY_IPV4)
-    {
-        gCrmOrch->incCrmResUsedCounter(CrmResourceType::CRM_IPV4_ROUTE);
-    }
-    else
-    {
-        gCrmOrch->incCrmResUsedCounter(CrmResourceType::CRM_IPV6_ROUTE);
-    }
-    return true;
-}
-
-
 bool FgNhgOrch::validNextHopInNextHopGroup(const NextHopKey& nexthop)
 {
     SWSS_LOG_ENTER();
@@ -387,8 +350,8 @@ bool FgNhgOrch::validNextHopInNextHopGroup(const NextHopKey& nexthop)
             auto prefix_entry = m_fgNhgPrefixes.find(route_table.first);
             if (prefix_entry == m_fgNhgPrefixes.end())
             {
-                auto member_entry = m_fgNhgMembers.find(nexthop.ip_address);
-                if (member_entry == m_fgNhgMembers.end())
+                auto member_entry = m_fgNhgNexthops.find(nexthop.ip_address);
+                if (member_entry == m_fgNhgNexthops.end())
                 {
                     SWSS_LOG_ERROR("fgNhgOrch got a validNextHopInNextHopGroup for non-configured FG ECMP entry");
                     return false;
@@ -457,8 +420,8 @@ bool FgNhgOrch::invalidNextHopInNextHopGroup(const NextHopKey& nexthop)
             auto prefix_entry = m_fgNhgPrefixes.find(route_table.first);
             if (prefix_entry == m_fgNhgPrefixes.end())
             {
-                auto member_entry = m_fgNhgMembers.find(nexthop.ip_address);
-                if (member_entry == m_fgNhgMembers.end())
+                auto member_entry = m_fgNhgNexthops.find(nexthop.ip_address);
+                if (member_entry == m_fgNhgNexthops.end())
                 {
                     SWSS_LOG_ERROR("fgNhgOrch got an invalidNextHopInNextHopGroup for non-configured FG ECMP entry");
                     return false;
@@ -1059,8 +1022,8 @@ bool FgNhgOrch::isRouteFineGrainedECMP(sai_object_id_t vrf_id, const IpPrefix &i
     {
         for (NextHopKey nhk : next_hop_set)
         {
-            auto member_entry = m_fgNhgMembers.find(nhk.ip_address);
-            if (member_entry == m_fgNhgMembers.end())
+            auto member_entry = m_fgNhgNexthops.find(nhk.ip_address);
+            if (member_entry == m_fgNhgNexthops.end())
             {
                 return false;
             }
@@ -1109,7 +1072,8 @@ bool FgNhgOrch::containsRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix)
 }
 
 
-bool FgNhgOrch::addRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const NextHopGroupKey &nextHops)
+bool FgNhgOrch::addModifyFgNhg(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const NextHopGroupKey &nextHops, 
+                                    sai_object_id_t &next_hop_id, bool &prevNhgWasFineGrained)
 {
     SWSS_LOG_ENTER();
 
@@ -1124,8 +1088,8 @@ bool FgNhgOrch::addRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
     {
         for (NextHopKey nhk : next_hop_set)
         {
-            auto member_entry = m_fgNhgMembers.find(nhk.ip_address);
-            if (member_entry == m_fgNhgMembers.end())
+            auto member_entry = m_fgNhgNexthops.find(nhk.ip_address);
+            if (member_entry == m_fgNhgNexthops.end())
             {
                 SWSS_LOG_ERROR("fgNhgOrch got a route addition %s:%s for non-configured FG ECMP entry",
                                     ipPrefix.to_string().c_str(), nextHops.to_string().c_str());
@@ -1224,6 +1188,7 @@ bool FgNhgOrch::addRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
     if (syncd_fg_route_entry_it != m_syncdFGRouteTables.at(vrf_id).end())
     {
         FGNextHopGroupEntry *syncd_fg_route_entry = &(syncd_fg_route_entry_it->second);
+        prevNhgWasFineGrained = true;
 
         /* Route exists, update FG ECMP group in SAI */
         for (auto nhk : syncd_fg_route_entry->active_nexthops)
@@ -1249,6 +1214,7 @@ bool FgNhgOrch::addRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
     else
     {
         /* New route + nhg addition */
+        prevNhgWasFineGrained = false;
         if (next_hop_to_add == false)
         {
             SWSS_LOG_INFO("There were no valid next-hops to add %s:%s", ipPrefix.to_string().c_str(),
@@ -1264,11 +1230,6 @@ bool FgNhgOrch::addRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
         }
 
         if (!setNewNhgMembers(syncd_fg_route_entry, fgNhgEntry, bank_member_changes, nhopgroup_members_set, ipPrefix))
-        {
-            return false;
-        }
-
-        if (!createFineGrainedRouteEntry(syncd_fg_route_entry, fgNhgEntry, vrf_id, ipPrefix, nextHops))
         {
             return false;
         }
@@ -1296,11 +1257,12 @@ bool FgNhgOrch::addRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
         }
     }
 
+    next_hop_id =  m_syncdFGRouteTables[vrf_id][ipPrefix].next_hop_group_id;
     return true;
 }
 
 
-bool FgNhgOrch::removeRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix)
+bool FgNhgOrch::removeFgNhg(sai_object_id_t vrf_id, const IpPrefix &ipPrefix)
 {
     SWSS_LOG_ENTER();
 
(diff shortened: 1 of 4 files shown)


orchagent/routeorch.cpp
```diff
@@ -1227,14 +1227,46 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
...
+         * so that it can be added for fine grained ecmp after that. */ 
+        auto route_table = m_syncdRoutes.find(vrf_id);
+        if (route_table != m_syncdRoutes.end())
+        {
+            auto route_entry = route_table->second.find(ipPrefix);
+            if (route_entry != route_table->second.end()) 
+            {
+                /* Case where we transition route from Regular ECMP to Fine Grained ECMP */
+                NextHopGroupKey nhgk = route_entry->second;
+                removeRoute(ctx);
```
qiluo-msft, line 1243: >removeRoute [](start = 16, length = 11)

I am worried about the complex design.
We should not call `removeRoute` inside `addRoute`
We should not manipulate `gRouteBulker` inside `addRoute`.

anish-n: This is to handle the case where the route needs Fine Grained ECMP based on the nexthops of the route. The only other way I see to handle this(not do removeRoute/bulker calls) is to add checks in fgnhgorch to see if a route exists in routeorch already, and if so use sai apis to modify the nexthop of the route(set_entry_attr) to fine grained ecmp, but this would mean that fgnhgorch will need to call into routeorch to delete the routes from routeorch internal data structure and remove the next-hop group if the ref count drops to 0. IMO the full routeorch removeRoute, as implemented just now, seems cleaner to me and better from a maintainability standpoint, though I am open to the mentioned way/ other suggestions.

orchagent/fgnhgorch.h
```diff
@@ -99,10 +108,16 @@ class FgNhgOrch : public Orch, public Observer
     NeighOrch *m_neighOrch;
     IntfsOrch *m_intfsOrch;
     VRFOrch *m_vrfOrch;
+
     FgNhgs m_FgNhgs;
     FGRouteTables m_syncdFGRouteTables;
+    FgNhgMembers m_fgNhgMembers;
```
prsunny, line 114: I think for clarity, is it better to call this `m_fgNhgNexthops`?. Members is generic that it can also include prefixes

orchagent/fgnhgorch.cpp
```diff
@@ -965,10 +981,99 @@ bool FgNhgOrch::setNewNhgMembers(FGNextHopGroupEntry &syncd_fg_route_entry, FgNh
...
+                return false;
+            }
+
+            if (!fgNhgEntry)
+            {
+                fgNhgEntry = member_entry->second;
+            }
+            else
+            {
+                /* Case where fgNhgEntry is alredy found via previous nexthop
```
prsunny, line 1012: So if some nexthop is in FGNHG but some are not, we are just returning false, correct? May be we should handle that case as well with a warning log?

orchagent/routeorch.cpp
```diff
@@ -1227,14 +1227,46 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
...
+        SWSS_LOG_INFO("Reroute %s:%s to fgNhgOrch", ipPrefix.to_string().c_str(),
+                        nextHops.to_string().c_str());
         return m_fgNhgOrch->addRoute(vrf_id, ipPrefix, nextHops);
     }
+    else
+    {
+        /* Route needs regular next-hop/ECMP. Logic here deals with case where we transition 
+         * route from Fine Grained ECMP to Regular ECMP: we first delete the route from fine grained ECMP 
+         * and then the rest of this function will create the route with regular ECMP/nexthops */
+        if (!m_fgNhgOrch->removeRoute(vrf_id, ipPrefix))
```
prsunny, line 1265: Why to call this if it is not a FGNHG route? This is the else case, correct?

orchagent/fgnhgorch.cpp
```diff
@@ -1269,6 +1370,7 @@ bool FgNhgOrch::doTaskFgNhg(const KeyOpFieldsValuesTuple & t)
     string key = kfvKey(t);
     string fg_nhg_name = key; 
     auto fgNhg_entry = m_FgNhgs.find(fg_nhg_name);
+    uint8_t match_mode = -1;
```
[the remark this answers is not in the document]
anish-n: Makes sense, will do so

orchagent/fgnhgorch.h
anish-n: Thanks, will do so.

orchagent/fgnhgorch.cpp
anish-n: That is correct: false is returned and regular ECMP gets created in that case due to false return. Makes sense, will add a warning log for that case.

orchagent/routeorch.cpp
anish-n: This is for the case where there were Fine Grained next-hops a, b, c associated with prefix 1.1.1.1 and a route was created, now the route changes such that 1.1.1.1 points to regular ecmp next-hops d, e. When this happens, we need to remove the route from fgnhgorch and create it via routeorch in order to create regular ECMP.

orchagent/routeorch.cpp
anish-n: Made changes for compatibility and to address the concerns

orchagent/routeorch.cpp
```diff
@@ -1260,17 +1253,31 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
...
+        /* The route is pointing to a next hop */
         NextHopKey nexthop;
         if (overlay_nh)
         {
             nexthop = NextHopKey(nextHops.to_string(), overlay_nh);
         }
         else
         {
-            nexthop = NextHopKey(nextHops.to_string());
+            nexthop =  NextHopKey(nextHops.to_string());
```
qiluo-msft, line 1280: >  [](start = 22, length = 1)

Remove one extra blank character.

orchagent/fgnhgorch.cpp
```diff
@@ -965,10 +944,100 @@ bool FgNhgOrch::setNewNhgMembers(FGNextHopGroupEntry &syncd_fg_route_entry, FgNh
...
+    else
+    {
+        for (NextHopKey nhk : next_hop_set)
+        {
+            auto member_entry = m_fgNhgNexthops.find(nhk.ip_address);
+            if (member_entry == m_fgNhgNexthops.end())
+            {
+                SWSS_LOG_ERROR("fgNhgOrch got a route addition %s:%s for non-configured FG ECMP entry",
+                                    ipPrefix.to_string().c_str(), nextHops.to_string().c_str());
+                return true;
```
qiluo-msft, line 1034: >return [](start = 16, length = 6)

For early return cases (there are more below), set prevNhgWasFineGrained to false?

orchagent/routeorch.cpp
```diff
@@ -1233,19 +1233,12 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
...
-                nextHops.to_string().c_str());
-        return m_fgNhgOrch->addRoute(vrf_id, ipPrefix, nextHops);
-    }
-
     /* next_hop_id indicates the next hop id or next hop group id of this route */
     sai_object_id_t next_hop_id = SAI_NULL_OBJECT_ID;
     bool overlay_nh = false;
     bool status = false;
+    bool newNhgIsFineGrained = false;
+    bool prevNhgWasFineGrained = false;
```
qiluo-msft, line 1241: >false [](start = 33, length = 5)

It is an output parameter so no need to initialize.

orchagent/routeorch.cpp
```diff
@@ -1435,12 +1442,22 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
...
         }
 
-        route_attr.id = SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID;
-        route_attr.value.oid = next_hop_id;
+        if (newNhgIsFineGrained && prevNhgWasFineGrained)
+        {
+            /* Don't change route entry if the route is previously fine grained and new nhg is also fine grained. 
+             * We already modifed sai nhg objs as part of addModifyFgNhg to account for nhg change. */
+            object_statuses.emplace_back();
+            object_statuses.back() = SAI_STATUS_SUCCESS;
```
qiluo-msft, line 1450: You can combine 2 statements into one. Just `emplace_back(SAI_STATUS_SUCCESS)`

orchagent/routeorch.cpp
anish-n: Resolved, thanks

orchagent/routeorch.cpp
anish-n: Makes sense, removed init.

orchagent/routeorch.cpp
anish-n: Combined the statements

orchagent/fgnhgorch.cpp
anish-n: Made prevNhgWasFineGrained default to true, because for early return cases where addModifyFgNhg returns success, prevNhgWasFineGrained should be true. For early return cases where addModifyFgNhg returns failure, routeorch addRoute will return false directly, so we don't care about prevNhgWasFineGrained on early failure return.

orchagent/routeorch.cpp
```diff
@@ -1233,19 +1233,12 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
...
-        SWSS_LOG_INFO("Reroute %s:%s to fgNhgOrch", ipPrefix.to_string().c_str(),
-                nextHops.to_string().c_str());
-        return m_fgNhgOrch->addRoute(vrf_id, ipPrefix, nextHops);
-    }
-
     /* next_hop_id indicates the next hop id or next hop group id of this route */
     sai_object_id_t next_hop_id = SAI_NULL_OBJECT_ID;
     bool overlay_nh = false;
     bool status = false;
+    bool newNhgIsFineGrained = false;
```
qiluo-msft, line 1240: >newNhgIsFineGrained [](start = 9, length = 19)

Better name `curNhgIsFineGrained`

orchagent/fgnhgorch.h
```diff
@@ -52,6 +52,11 @@ typedef std::map<IpPrefix, NextHopGroupKey> FgPrefixOpCache;
 /* Map from link name to next-hop IP */
 typedef std::unordered_map<string, std::vector<IpAddress>> Links;
 
+enum FGMatchModes
```
qiluo-msft, line 55: >FGMatchModes [](start = 5, length = 12)

Better name `FGMatchMode`

orchagent/fgnhgorch.h
```diff
@@ -68,10 +73,13 @@ typedef struct FgNhgEntry
     Links links;                                      // Link to IP map for oper changes
     std::vector<IpPrefix> prefixes;                   // Prefix which desires FG behavior
     std::vector<BankIndexRange> hash_bucket_indices;  // The hash bucket indices for a bank
+    uint8_t match_mode;                               // Stores a match_mode from FGMatchModes
```
qiluo-msft, line 76: >uint8_t [](start = 4, length = 7)

FGMatchMode

orchagent/fgnhgorch.h
```diff
@@ -86,23 +94,30 @@ typedef struct
...
     FgNhgOrch(DBConnector *db, DBConnector *appDb, DBConnector *stateDb, vector<string> &tableNames, NeighOrch *neighOrch, IntfsOrch *intfsOrch, VRFOrch *vrfOrch);
 
     void update(SubjectType type, void *cntx);
-    bool addRoute(sai_object_id_t, const IpPrefix&, const NextHopGroupKey&);
-    bool removeRoute(sai_object_id_t, const IpPrefix&);
+    bool isRouteFineGrainedECMP(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const NextHopGroupKey &nextHops);
+    bool containsRoute(sai_object_id_t vrf_id, const IpPrefix &ipPrefix);
     bool validNextHopInNextHopGroup(const NextHopKey&);
     bool invalidNextHopInNextHopGroup(const NextHopKey&);
+    bool addModifyFgNhg(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const NextHopGroupKey &nextHops, sai_object_id_t &next_hop_id, bool &prevNhgWasFineGrained);
```
qiluo-msft, line 104: >addModifyFgNhg [](start = 9, length = 14)

Do you mean add or modify?
If yes, maybe `setFgNhg` is better.

anish-n: Yes it is add or modify, the fn will either create a new Fine Grained ECMP nexthop group or modify an existing Fine Grained ECMP nexthop group in sai based on some conditions.

orchagent/fgnhgorch.h
```diff
@@ -86,23 +94,30 @@ typedef struct
 class FgNhgOrch : public Orch, public Observer
 {
 public:
-    FgNhgPrefixes fgNhgPrefixes;
     FgNhgOrch(DBConnector *db, DBConnector *appDb, DBConnector *stateDb, vector<string> &tableNames, NeighOrch *neighOrch, IntfsOrch *intfsOrch, VRFOrch *vrfOrch);
 
     void update(SubjectType type, void *cntx);
-    bool addRoute(sai_object_id_t, const IpPrefix&, const NextHopGroupKey&);
-    bool removeRoute(sai_object_id_t, const IpPrefix&);
+    bool isRouteFineGrainedECMP(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const NextHopGroupKey &nextHops);
```
qiluo-msft, line 100: >isRouteFineGrainedECMP [](start = 9, length = 22)

I am a little confused  about your naming conventions.

Are they the same thing?
- FineGrainedECMP
- FgNhg
- FineGrained
- m_fgNhgNexthops

Is it concise and understandable to call it `FineGroup` `isFineGroup`.

`isRouteFineGrainedECMP` -> `containsRoute` ?

anish-n: This is what each of them mean:
FineGrainedECMP: The term for the full feature -  Fine Grained ECMP
FgNhg: Fine Grained Next Hop Group. Similar in concept to ECMP next-hop group, but for Fine Grained ECMP behavior
FineGrained: Same as FineGrainedECMP, a concise name
m_fgNhgNexthops: The individual nexthops which comprise the FgNhg

Guess we can call it isFineGrained or isFgNhg

``containsRoute`` checks if a Fine Grained ECMP is associated with prefix is created in the ASIC
``isRouteFineGrainedECMP`` looks at current nhs of a route, and determines if it needs fine grained behavior. It is possible for isRouteFineGrainedECMP to be true while containsRoute is false, and all combinations of true/false. 

Can change to following, if you think its better ``isRouteFineGrainedECMP`` -> ``isNhgFineGrained`` , ``containsRoute`` -> ``containsSyncdFineGrained``.

orchagent/routeorch.cpp
anish-n: Makes sense, thanks

orchagent/fgnhgorch.h
anish-n: Done, thanks!

orchagent/fgnhgorch.h
anish-n: Corrected, thanks!

orchagent/fgnhgorch.h
anish-n: Changed code to isRouteFIneGrained and syncdContainsFgNhg. Let me know if this and above makes sense.

orchagent/routeorch.cpp
```diff
@@ -1233,19 +1233,12 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
...
-                nextHops.to_string().c_str());
-        return m_fgNhgOrch->addRoute(vrf_id, ipPrefix, nextHops);
-    }
-
     /* next_hop_id indicates the next hop id or next hop group id of this route */
     sai_object_id_t next_hop_id = SAI_NULL_OBJECT_ID;
     bool overlay_nh = false;
     bool status = false;
+    bool curNhgIsFineGrained = false;
+    bool prevNhgWasFineGrained;
```
prsunny, line 1241: Please initialize to false. Else this will be undefined if `isRouteFineGrained` returns false

orchagent/fgnhgorch.h
```diff
@@ -52,6 +52,11 @@ typedef std::map<IpPrefix, NextHopGroupKey> FgPrefixOpCache;
 /* Map from link name to next-hop IP */
 typedef std::unordered_map<string, std::vector<IpAddress>> Links;
 
+enum FGMatchMode
+{
+    ROUTE_BASED,
+    NEXTHOP_BASED
```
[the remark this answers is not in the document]
anish-n: Makes sense, have changed to nexthop-based, will change the schema in the hld too.

orchagent/routeorch.cpp
anish-n: Thanks, put the initialization code in.

orchagent/fgnhgorch.cpp
qiluo-msft: I know your code is working. I am arguing about the individual function design. Here you assume the caller takes the responsibility to initialize the output parameter `prevNhgWasFineGrained`. This is not necessary. You can make the decision on its boolean value internally, and output in any cases.

Then outside this function, you don't need to initialize the variable used as an output argument.

---
In reply to: [549525398](https://github.com/Azure/sonic-swss/pull/1565#discussion_r549525398) [](ancestors = 549525398)

prsunny (dismissed):
lgtm.. @qiluo-msft to signoff before merge

orchagent/fgnhgorch.cpp
anish-n: Yes, it makes sense to have the fn logic take care of the value setting internally. I have set the value in the fn: https://github.com/Azure/sonic-swss/pull/1565/files#diff-076ab0a620bf8e45930d13ac4b8a494eca6d6349b7238bcd52bfeafc1acb7116R1025

The caller is still setting the value to false to prevent an uninitalized var: the value of ```prevNhgWasFineGrained``` is only set when ```isRouteFineGrained``` is true and ```setFgNhg``` is called:
https://github.com/Azure/sonic-swss/pull/1565/files#diff-dd5418e58fe44ea78956f7404758ba1a9eb88d9f1900e29a60a75c79af387480R1256

This uninit var will then be checked in https://github.com/Azure/sonic-swss/pull/1565/files#diff-dd5418e58fe44ea78956f7404758ba1a9eb88d9f1900e29a60a75c79af387480R1445 . So it may be better for readability and to prevent an uninit var issue to have to explicitly set to false.

qiluo-msft:
Please resolve conflicts and lgtm new alerts.

anish-n:
retest this please

qiluo-msft (dismissed):

anish-n:
retest this please

anish-n:
retest this please

anish-n:
retest this please

liushilongbuaa:
retest vs please

anish-n:
retest this please

liushilongbuaa:
retest vs please

anish-n:
retest vs please

lguohan:
retest vs please

lguohan (approved):

PR #1568, opened 2020-12-24
[linksync] Netdev oper status determination using IFF_RUNNING

vganesan-nokia:
Signed-off-by: vedganes <vedavinayagam.ganesan@nokia.com>

Changes in linksync to use IFF_RUNNING to determine netdev operational
status instead of using IFF_LOWER_UP.

**What I did**

Changed linksync.cpp to use IFF_RUNNING to determine the netdev operational status instead of using IFF_LOWER_UP.

**Why I did it**

The objective of this change is to make orchagent/linksync align with FRR/Zebra in handling interface states. Zebra uses IFF_RUNNING flag to determine interfaces' oper state while linksync uses IFF_LOWER_UP. Zebra rightly depends on IFF_RUNNING flag. As a routing daemon, zebra wants to know whether the interface is capable of passing packets or not. The flag IFF_RUNNING indicates exactly that (based on comment in if.h, this flag reflects the interface oper up state as defined in RFC2863). Since orchagent uses IFF_LOWER_UP, which comes earlier than IFF_RUNNING, there is interface state mismatch between zebra and orchagent for a window of time. Since with voq implementation we write static neigh/routes when the netdev (inband port) becomes oper up and bgp depends on these kernel entries, there is a need for this change to have the interface state timing same in both orchagent and zebra.  

Refer issue https://github.com/Azure/sonic-buildimage/issues/6295 for detailed information.

**How I verified it**

- Currently the operational state of netdev of front panel ports are not used. It is used only for the management port. All vs tests passed.

**Details if related**

- The operational state will be used for voq based chassis systems with inband interface type port  (Ref: https://github.com/Azure/SONiC/blob/master/doc/voq/voq_hld.md, https://github.com/Azure/SONiC/blob/master/doc/voq/architecture.md). With inband interface implementation, there will be an additional attribute called "netdev_oper_status" in the **PORT_TABLE** of STATE_DB. This "netdev_oper_status" will be set based on IFF_RUNNING.

### DIFF
--- a/portsyncd/linksync.cpp
+++ b/portsyncd/linksync.cpp
@@ -174,7 +174,7 @@ void LinkSync::onMsg(int nlmsg_type, struct nl_object *obj)
 
     unsigned int flags = rtnl_link_get_flags(link);
     bool admin = flags & IFF_UP;
-    bool oper = flags & IFF_LOWER_UP;
+    bool oper = flags & IFF_RUNNING;
 
     char addrStr[MAX_ADDR_SIZE+1] = {0};
     nl_addr2str(rtnl_link_get_addr(link), addrStr, MAX_ADDR_SIZE);


vganesan-nokia:
retest this please

anshuv-mfst:
@eswaran (Arista), @abdosi  - could you please review the PR, Thanks.

portsyncd/linksync.cpp
```diff
@@ -174,7 +174,7 @@ void LinkSync::onMsg(int nlmsg_type, struct nl_object *obj)
 
     unsigned int flags = rtnl_link_get_flags(link);
     bool admin = flags & IFF_UP;
-    bool oper = flags & IFF_LOWER_UP;
+    bool oper = flags & IFF_RUNNING;
```
prsunny, line 177: For e.g, this interface has LOWER_UP but state is UNKNOWN.  Will it have IFF_RUNNING set?

58: Loopback0: <BROADCAST,NOARP,UP,`LOWER_UP`> mtu 65536 qdisc noqueue state `UNKNOWN` mode DEFAULT group default qlen 1000
    link/ether 96:a7:77:d1:73:34 brd ff:ff:ff:ff:ff:ff

vganesan-nokia: Yes. Linux command "ip link show" does not show it here. The bgp "show interface" will show this flag. Here is an example of interface info got by linux command and bgp command.

admin@Linecard2:\~$ 
admin@Linecard2:\~$ ```ip link show | grep -A1 Loopback0```
10: Loopback0: <BROADCAST,NOARP,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/ether f6:55:7a:eb:85:eb brd ff:ff:ff:ff:ff:ff
admin@Linecard2:\~$ 
admin@Linecard2:\~$ ```docker exec bgp vtysh -c "show interface" | grep -A12 Loopback0```
Interface Loopback0 is ```up```, line protocol is up
  Link ups:       1    last: 2019/03/02 21:32:48.16
  Link downs:     0    last: (never)
  vrf: default
  index 10 metric 0 mtu 65536 speed 0 
  ```flags: <UP,BROADCAST,RUNNING,NOARP>```
  Type: Unknown
  HWaddr: f6:55:7a:eb:85:eb
  inet 10.1.0.1/32 unnumbered
  inet6 fe80::f455:7aff:feeb:85eb/64
  inet6 fc00:10::1/128
  Interface Type Other
  Interface Slave Type None
admin@Linecard2:~$

prsunny (approved):

vganesan-nokia:
@prsunny, would please re-start the AzurePipelines? It seems they are stuck and not progressing. I tried to start it. It seems I do not have permission to start AzurePipelines.
Thanks

vganesan-nokia:
@prsunny, thanks for restarting tests.

vganesan-nokia:
@eswaran, @abdosi - A gentle remainder, would you please review and approve?

lguohan:
HERE IS LINUX MANUAL, wonder what the different is between iff_running and if_lower_up? why do we think iff_running is the right thing to do?

              IFF_RUNNING       Resources allocated.
              IFF_LOWER_UP      Driver signals L1 up (since Linux 2.6.17)

vganesan-nokia:
> HERE IS LINUX MANUAL, wonder what the different is between iff_running and if_lower_up? why do we think iff_running is the right thing to do?
> 
> ```
>           IFF_RUNNING       Resources allocated.
>           IFF_LOWER_UP      Driver signals L1 up (since Linux 2.6.17)
> ```

The objective is to make orchagent/linksync align with FRR/Zebra in handling interface states. Zebra uses IFF_RUNNING flag to determine interfaces' oper state while linksync uses IFF_LOWER_UP. This leads to interface state mismatch between zebra and orchagent for a window of time (IFF_LOWER_UP and IFF_RUNNING events come at different points of time). Since with voq implementation we write static neigh/routes when the netdev (inband port) becomes oper up and bgp depends on these kernel entries, there is a need for having the interface state timing same in both orchagent and zebra. I have described this in detail in issue https://github.com/Azure/sonic-buildimage/issues/6295.

lguohan:
is iff_running after iff_lower_up? is zebra doing the right thing?

vganesan-nokia:
> is iff_running after iff_lower_up? 

Yes. IFF_LOWER_UP comes up earlier than IFF_RUNNING. But zebra is not acting on IFF_LOWE_UP.

> is zebra doing the right thing?

In our SONiC linux,  in the file if.h, the IFF_RUNNING has comment "@IFF_RUNNING: interface RFC2863 OPER_UP. Volatile."

As a routing daemon, zebra want to know whether the interface is capable of passing packets or not. So it rightly depends on this IFF_RUNNING flag which indicates whether the interface is able to pass packets or not.

So I guess zebra is doing the right thing.

lguohan:
make sense, can you add above discussion in the pr description. we will add it as part of the pr commit message.

vganesan-nokia:
> make sense, can you add above discussion in the pr description. we will add it as part of the pr commit message.

Updated the PR comment as suggested.

PR #1579, opened 2021-01-05
[dvs] Update/disable DVS tests with new FRR 7.5 behavior

daall:
- Fix erspan test behavior to use valid alternate route for mirror session
- Skip route tests w/ FRR VRF bug

**What I did**
I updated/disabled DVS tests to account for new FRR 7.5 behavior.

**Why I did it**
To unblock the PR pipeline.

**How I verified it**
Re-ran both test files, all tests pass now.

**Details if related**

### DIFF
--- a/tests/test_mirror_port_erspan.py
+++ b/tests/test_mirror_port_erspan.py
@@ -443,7 +443,7 @@ class TestMirror(object):
         # mirror session move round 2
         # remove port channel member
         self.dvs_lag.remove_port_channel_member("080", "Ethernet32")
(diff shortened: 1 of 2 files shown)


prsunny (approved):

lguohan:
not sure why vstests are all failing?

daall:
> not sure why vstests are all failing?

this got buried in the email thread yesterday, but there is currently a mismatch b/w the swss-common version we are using and the sairedis version we are using. once this merges and the sairedis build succeed, swss should be unblocked

PR #1585, opened 2021-01-08
[FPMSYNCD] Change done to support FRR7.5

kishorekunal01:
What I did
1) While extracting the RT Attribute from netlink message, If the message is NESTED msg remove the NESTED flag and extract the msg.
2) Get nexthop ifindex from Netlink msg RTA_OIF attribute and use the name based on the ifindex program in the APP_DB.

Why I did it
1) With FRR7.5 upgrade Netlink Msg from FRR has NLA_F_NESTED flag set in
RTA_TYPE while sending VNI and RMAC.
2) In the new patch to keep it in line with FRR upstreaming, we are not
sending the VLAN ID from FRR as it can be extracted from ifindex inside
the nexthop information.

How I verified it
Verified using vsonic

### DIFF
--- a/fpmsyncd/fpmlink.cpp
+++ b/fpmsyncd/fpmlink.cpp
@@ -18,6 +18,18 @@ void netlink_parse_rtattr(struct rtattr **tb, int max, struct rtattr *rta,
         {
             tb[rta->rta_type] = rta;
         }
+        else
+        {
+            /* FRR 7.5 is sending RTA_ENCAP with NLA_F_NESTED bit set*/
+            if (rta->rta_type & NLA_F_NESTED)
+            {
+                int rta_type = rta->rta_type & ~NLA_F_NESTED;
+                if (rta_type <= max)
+                {
+                   tb[rta_type] = rta;
+                }
+            }
+        }
         rta = RTA_NEXT(rta, len);
     }
 }
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -31,7 +31,6 @@ using namespace swss;
 
 #define VXLAN_VNI             0
 #define VXLAN_RMAC            1
-#define VXLAN_VLAN            2
 #define NH_ENCAP_VXLAN      100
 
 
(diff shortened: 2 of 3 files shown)


kishorekunal01:
test vs please

lguohan:
does this only impact vxlan route, or normal route as well?

kishorekunal01:
> does this only impact vxlan route, or normal route as well?

Only Vxlan routes

fpmsyncd/fpmlink.cpp
```diff
@@ -18,6 +18,18 @@ void netlink_parse_rtattr(struct rtattr **tb, int max, struct rtattr *rta,
         {
             tb[rta->rta_type] = rta;
         }
+        else
+        {
+            /* Latest FRR is sending RTA_ENCAP with NLA_F_NESTED bit set*/
```
prsunny, line 23: Please rephrase the comment to mention FRR version or remove the comment itself. "Latest" may not be relevant sometime in future

fpmsyncd/routesync.cpp
```diff
@@ -102,10 +101,9 @@ void RouteSync::parseEncap(struct rtattr *tb, uint32_t &encap_value, string &rma
     parseRtAttrNested(tb_encap, 3, tb);
     encap_value = *(uint32_t *)RTA_DATA(tb_encap[VXLAN_VNI]);
     memcpy(&mac_buf, RTA_DATA(tb_encap[VXLAN_RMAC]), MAX_ADDR_SIZE);
-    vlan = *(uint32_t *)RTA_DATA(tb_encap[VXLAN_VLAN]);
```
prsunny, line 105: How do we get the vlan data if it is not nested? Do we need to handle that?

kishorekunal01: Vlan information was used to create the nexthop,  Similar to normal route(no evpn routes) nexthop information is extracted from tb[RTA_OIF] attribute.

fpmsyncd/fpmlink.cpp
kishorekunal01: Sure, I will update the comment.

prsunny:
@kishorekunal01 , does it require this patch (https://github.com/Azure/sonic-buildimage/pull/4806)  to be merged before we can merge this PR?

kishorekunal01:
> @kishorekunal01 , does it require this patch ([Azure/sonic-buildimage#4806](https://github.com/Azure/sonic-buildimage/pull/4806)) to be merged before we can merge this PR?

No, this change is independent changes and doesn't depend on the FRR patch.

prsunny (approved):

PR #1590, opened 2021-01-11
[ci]: only copy artifacts when build is successful

lguohan:
**What I did**
only copy artifacts when build is successful

**Why I did it**
fix false negative 

**How I verified it**

**Details if related**

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -63,8 +63,7 @@ jobs:
     submodules: true
   - script: |
       ./autogen.sh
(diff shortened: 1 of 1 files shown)


lguohan:
prevent false negative here.

https://dev.azure.com/sonicswitch/build/_build/results?buildId=1297&view=logs&j=196c4937-6990-58b5-7d18-234ac2f0888b&t=fa1940da-ab03-5e4d-4f28-fa90286ca8a8

qiluo-msft (approved):

PR #1593, opened 2021-01-12
[201911-SWSS] Change Error log to WARN log for FDB flush notification failure

madhanmellanox:
**What I did**
I modified Error log for FDB flush notification failure due to VLAN removal to WARN log.

**Why I did it**
I modified it because, the VLAN is already removed and we are receiving a FLUSH notification for the same, it should be treated as a warning and not an error.

**How I verified it**
I created a VLAN and added a member port to it and I learnt FDB entry on that port for the VLAN. Then removed the member port and removed the VLAN. Once, I remove the VLAN, there will be a log "Failed to locate vlan", it should be a warning log.

**Details if related**
Code changes are in Sonic-swss fdb orchagent.

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -171,7 +171,7 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
 
             if (!m_portsOrch->getPort(entry->bv_id, vlan))
             {
-                SWSS_LOG_WARN("FdbOrch notification: Failed to locate vlan\
+                SWSS_LOG_NOTICE("FdbOrch notification: Failed to locate vlan\
                                 port from bv_id 0x%" PRIx64, entry->bv_id);
                 return;
             }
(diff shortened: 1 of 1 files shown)


orchagent/fdborch.cpp
```diff
@@ -171,7 +171,7 @@ void FdbOrch::update(sai_fdb_event_t type, const sai_fdb_entry_t* entry, sai_obj
 
             if (!m_portsOrch->getPort(entry->bv_id, vlan))
             {
-                SWSS_LOG_ERROR("FdbOrch notification: Failed to locate vlan\
+                SWSS_LOG_WARN("FdbOrch notification: Failed to locate vlan\
```
prsunny, line 174: Please change to NOTICE as other places have it as NOTICE for the same case. Also could you make the change to master to be consistent - https://github.com/Azure/sonic-swss/blob/master/orchagent/fdborch.cpp#L360

madhanmellanox: @prsunny modified ERR log to NOTICE log
raised a PR for master https://github.com/Azure/sonic-swss/pull/1595

prsunny (approved):

PR #1595, opened 2021-01-12
[sonic-swss] modified ERR log to NOTICE log for FDB notification failure after VLA…

madhanmellanox:
…N delete

**What I did**
I modified Error log for FDB flush notification failure due to VLAN removal to NOTICE log.

**Why I did it**
I modified it because, the VLAN is already removed and we are receiving a FLUSH notification for the same, it should be treated as a warning and not an error.

**How I verified it**
I created a VLAN and added a member port to it and I learnt FDB entry on that port for the VLAN. Then removed the member port and removed the VLAN. Once, I remove the VLAN, there will be a log "Failed to locate vlan", it should be a warning log.

**Details if related**
Code changes are in Sonic-swss fdb orchagent.

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -184,7 +184,7 @@ void FdbOrch::update(sai_fdb_event_t        type,
 
         if (!m_portsOrch->getPort(entry->bv_id, vlan))
         {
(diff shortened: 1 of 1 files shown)


madhanmellanox:
retest please

prsunny (approved):

PR #1596, opened 2021-01-12
[synchronous mode] Add failure notification for SAI failures in synchronous mode

shi-su:
**What I did**
Add function to notify users in the presence of SAI failures by throwing exceptions and trigger swss restart in synchronous mode. And include this function in routeorch and neighorch. 

This is a part of the first step in the SAI failure handling in orchagent with synchronous mode:
1. Failure notification mechanism to ensure enough notifications in the presence of SAI failures and avoid running switches with unhandled failures. 
2. Add general failure handling logic by status.
3. Develop fine-grain failure handling mechanism for each orch to properly handle different SAI failures.

**Why I did it**
This function aims to ensure enough notifications in the presence of SAI failures and avoid running switches with unhandled failures (on-par with asynchronous mode). 

**How I verified it**
Generate SAI failure in vs SAI. Verify that the function gets the error status and throws an exception.
```
vlab-01 ERR syncd#syncd: :- create: create status: SAI_STATUS_FAILURE
vlab-01 ERR syncd#syncd: :- sendApiResponse: api SAI_COMMON_API_BULK_CREATE failed in syncd mode: SAI_STATUS_FAILURE
vlab-01 ERR swss#orchagent: :- addRoutePost: Failed to create route 201.0.0.128/25 with next hop(s) 10.0.0.57@PortChannel0001,10.0.0.59@PortChannel0002,10.0.0.61@PortChannel0003,10.0.0.63@PortChannel0004
Jan 29 02:09:48.865390 vlab-01 ERR swss#orchagent: :- handleSaiCreateStatus: Encountered failure in create operation, SAI API: SAI_API_ROUTE, status: SAI_STATUS_FAILURE
vlab-01 NOTICE swss#orchagent: :- uninitialize: begin
vlab-01 NOTICE swss#orchagent: :- uninitialize: begin
```

**Details if related**
From the orchagent’s perspective, failures in sairedis may caused by three types of failure: a) SAI failure; b) communication failure; c) metadata failure. The function in this PR is on par with asynchronous mode in the first two failure scenarios. For metadata failures, triggering swss restart might be more extreme than asynchronous mode. However, probability of such a scenario is low and the presence of such failure is likely to indicate bugs in orchagent, we treat it the same as the first two scenarios.

### DIFF
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -685,7 +685,7 @@ Executor *Orch::getExecutor(string executorName)
     return NULL;
 }
 
-bool Orch::handleSaiCreateFailure(sai_api_t api, sai_status_t status)
+bool Orch::handleSaiCreateStatus(sai_api_t api, sai_status_t status)
 {
     /*
      * This function aims to provide coarse handling of failures in sairedis create
@@ -699,7 +699,7 @@ bool Orch::handleSaiCreateFailure(sai_api_t api, sai_status_t status)
     return false;
 }
 
-bool Orch::handleSaiSetFailure(sai_api_t api, sai_status_t status)
+bool Orch::handleSaiSetStatus(sai_api_t api, sai_status_t status)
 {
     /*
      * This function aims to provide coarse handling of failures in sairedis set
@@ -713,7 +713,7 @@ bool Orch::handleSaiSetFailure(sai_api_t api, sai_status_t status)
     return false;
 }
 
-bool Orch::handleSaiRemoveFailure(sai_api_t api, sai_status_t status)
+bool Orch::handleSaiRemoveStatus(sai_api_t api, sai_status_t status)
 {
     /*
      * This function aims to provide coarse handling of failures in sairedis remove
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -685,6 +685,46 @@ Executor *Orch::getExecutor(string executorName)
     return NULL;
 }
 
+bool Orch::handleSaiCreateFailure(sai_status_t status)
+{
+    /*
+     * This function aims to provide coarse handling of failures in sairedis create
+     * operation (i.e., notify users by throwing excepions when failures happen).
+     * TODO: 1. Add general handling logic for specific statuses (e.g., SAI_STATUS_ITEM_ALREADY_EXISTS)
+     *       2. Develop fine-grain failure handling mechanisms specific and replace
+     *          this coarse handling in each orch.
+     */
+    SWSS_LOG_THROW("Encountered failure in Create operation, status: %d", status);
+    return false;
+}
+
+bool Orch::handleSaiSetFailure(sai_status_t status)
+{
+    /*
+     * This function aims to provide coarse handling of failures in sairedis set
+     * operation (i.e., notify users by throwing excepions when failures happen).
+     * TODO: 1. Add general handling logic for specific statuses
+     *       2. Develop fine-grain failure handling mechanisms specific and replace
+     *          this coarse handling in each orch.
+     */
+    SWSS_LOG_THROW("Encountered failure in set operation, status: %d", status); // need to update message, comment, and status 
+    return false;
+}
+
+bool Orch::handleSaiRemoveFailure(sai_status_t status)
+{
+    /*
+     * This function aims to provide coarse handling of failures in sairedis remove
+     * operation (i.e., notify users by throwing excepions when failures happen).
+     * TODO: 1. Add general handling logic for specific statuses (e.g., SAI_STATUS_OBJECT_IN_USE,
+     *          SAI_STATUS_ITEM_NOT_FOUND)
+     *       2. Develop fine-grain failure handling mechanisms specific and replace
+     *          this coarse handling in each orch.
+     */
+    SWSS_LOG_THROW("Encountered failure in DEL operation, status: %d", status); // need to update message, comment, and status 
+    return false;
+}
+
 void Orch2::doTask(Consumer &consumer)
 {
     SWSS_LOG_ENTER();
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -709,7 +709,7 @@ bool Orch::handleSaiSetStatus(sai_api_t api, sai_status_t status)
      *          this coarse handling in each orch.
      *       3. Take the type of sai api into consideration.
      */
-    SWSS_LOG_THROW("Encountered failure in set operation, status: %d", status); // need to update message, comment, and status 
+    SWSS_LOG_THROW("Encountered failure in set operation, status: %d", status);
     return false;
 }
 
@@ -724,7 +724,7 @@ bool Orch::handleSaiRemoveStatus(sai_api_t api, sai_status_t status)
      *          this coarse handling in each orch.
      *       3. Take the type of sai api into consideration.
      */
-    SWSS_LOG_THROW("Encountered failure in del operation, status: %d", status); // need to update message, comment, and status 
+    SWSS_LOG_THROW("Encountered failure in remove operation, status: %d", status);
     return false;
 }
 
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -685,7 +685,7 @@ Executor *Orch::getExecutor(string executorName)
     return NULL;
 }
 
-bool Orch::handleSaiCreateFailure(sai_status_t status)
+bool Orch::handleSaiCreateFailure(sai_api_t api, sai_status_t status)
 {
     /*
      * This function aims to provide coarse handling of failures in sairedis create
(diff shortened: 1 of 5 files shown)


qiluo-msft (changes requested):
Block this PR, for discuss purpose only

orchagent/orch.cpp
```diff
@@ -685,6 +685,46 @@ Executor *Orch::getExecutor(string executorName)
...
+{
+    /*
+     * This function aims to provide coarse handling of failures in sairedis create
+     * operation (i.e., notify users by throwing excepions when failures happen).
+     * TODO: 1. Add general handling logic for specific statuses (e.g., SAI_STATUS_ITEM_ALREADY_EXISTS)
+     *       2. Develop fine-grain failure handling mechanisms specific and replace
+     *          this coarse handling in each orch.
+     */
+    SWSS_LOG_THROW("Encountered failure in Create operation, status: %d", status);
+    return false;
```
qiluo-msft, line 698: >false [](start = 11, length = 5)

above line with throw anyway, so this line is dead code? can we remove it?

qiluo-msft: To make this function more general, only throw if `status != SAI_STATUS_SUCCESS`, and return true/false

---
In reply to: [560590840](https://github.com/Azure/sonic-swss/pull/1596#discussion_r560590840) [](ancestors = 560590840)

qiluo-msft: And rename function `handleSaiCreateStatus`

---
In reply to: [560591437](https://github.com/Azure/sonic-swss/pull/1596#discussion_r560591437) [](ancestors = 560591437,560590840)

orchagent/orch.h
```diff
@@ -218,6 +218,11 @@ class Orch
     static void recordTuple(Consumer &consumer, const swss::KeyOpFieldsValuesTuple &tuple);
 
     void dumpPendingTasks(std::vector<std::string> &ts);
+
+    /* Failure handling */
+    bool handleSaiCreateFailure(sai_status_t status);
```
qiluo-msft, line 223: >bool [](start = 4, length = 4)

protected functions?
virtual functions?

orchagent/orch.cpp
```diff
@@ -685,6 +685,46 @@ Executor *Orch::getExecutor(string executorName)
...
+bool Orch::handleSaiCreateFailure(sai_status_t status)
+{
+    /*
+     * This function aims to provide coarse handling of failures in sairedis create
+     * operation (i.e., notify users by throwing excepions when failures happen).
+     * TODO: 1. Add general handling logic for specific statuses (e.g., SAI_STATUS_ITEM_ALREADY_EXISTS)
+     *       2. Develop fine-grain failure handling mechanisms specific and replace
+     *          this coarse handling in each orch.
+     */
+    SWSS_LOG_THROW("Encountered failure in Create operation, status: %d", status);
```
qiluo-msft, line 697: >SWSS_LOG_THROW [](start = 4, length = 14)

In old async mode, [https://github.com/Azure/sonic-swss/blob/master/orchagent/notifications.cpp#L20](https://github.com/Azure/sonic-swss/blob/master/orchagent/notifications.cpp#L20) on_switch_shutdown_request will be called. Do you want to mimic exact message and exit?

orchagent/orch.cpp
shi-su: Renamed the function to `handleSaiCreateStatus`. Also added sai api type as an input. For the return boolean value, I was thinking to use it as an indicator for whether the status is handled or not. Currently, our aim is to mimic the behavior as async mode, so no status is handled in the function and the return value is not very meaningful. As the function evolves, the situation should change.

orchagent/orch.h
shi-su: Made the function as protected virtual.

orchagent/orch.cpp
shi-su: The message in `on_switch_shutdown_request` is only "syncd stopped". Personally, I do not like this message very much for two reasons: 1. it does not carry much information; 2. It does not precisely describe the situation in the sync mode scenario (In sync mode, syncd hasn't really stopped at this point).

qiluo-msft: Log message is good. Do you still want to mimic the exit behavior?

---
In reply to: [560594351](https://github.com/Azure/sonic-swss/pull/1596#discussion_r560594351) [](ancestors = 560594351)

orchagent/orch.cpp
```diff
@@ -685,6 +685,49 @@ Executor *Orch::getExecutor(string executorName)
...
+{
+    /*
+     * This function aims to provide coarse handling of failures in sairedis create
+     * operation (i.e., notify users by throwing excepions when failures happen).
+     * TODO: 1. Add general handling logic for specific statuses (e.g., SAI_STATUS_ITEM_ALREADY_EXISTS)
+     *       2. Develop fine-grain failure handling mechanisms specific and replace
+     *          this coarse handling in each orch.
+     *       3. Take the type of sai api into consideration.
+     */
+    SWSS_LOG_THROW("Encountered failure in create operation, status: %d", status);
```
qiluo-msft, line 698: >status [](start = 74, length = 6)

You may also log the api info.

orchagent/orch.cpp
qiluo-msft: Could you clarify "use it as an indicator for whether the status is handled or not"? Do you assume `status != SAI_STATUS_SUCCESS` ? What is 'handled' ?
It should be good function level comment.


---
In reply to: [563944011](https://github.com/Azure/sonic-swss/pull/1596#discussion_r563944011) [](ancestors = 563944011)

orchagent/orch.cpp
kcudnik: make %s and use sai_serialize_status(status).c_str() for string status

orchagent/orch.cpp
```diff
@@ -685,6 +685,49 @@ Executor *Orch::getExecutor(string executorName)
...
+{
+    /*
+     * This function aims to provide coarse handling of failures in sairedis set
+     * operation (i.e., notify users by throwing excepions when failures happen).
+     * TODO: 1. Add general handling logic for specific statuses
+     *       2. Develop fine-grain failure handling mechanisms specific and replace
+     *          this coarse handling in each orch.
+     *       3. Take the type of sai api into consideration.
+     */
+    SWSS_LOG_THROW("Encountered failure in set operation, status: %d", status);
```
kcudnik, line 712: make %s and use sai_serialize_status(status).c_str() for string status

orchagent/orch.cpp
```diff
@@ -685,6 +685,49 @@ Executor *Orch::getExecutor(string executorName)
...
+    /*
+     * This function aims to provide coarse handling of failures in sairedis remove
+     * operation (i.e., notify users by throwing excepions when failures happen).
+     * TODO: 1. Add general handling logic for specific statuses (e.g., SAI_STATUS_OBJECT_IN_USE,
+     *          SAI_STATUS_ITEM_NOT_FOUND)
+     *       2. Develop fine-grain failure handling mechanisms specific and replace
+     *          this coarse handling in each orch.
+     *       3. Take the type of sai api into consideration.
+     */
+    SWSS_LOG_THROW("Encountered failure in remove operation, status: %d", status);
```
kcudnik, line 727: make %s and use sai_serialize_status(status).c_str() for string status

kcudnik (changes requested):
Did you tested any of this change ? since now it will throw exceptions on any failure

orchagent/orch.cpp
shi-su: I think the criteria could be whether the SAI call needs to be retried after running this function. For example, similar to https://github.com/Azure/sonic-swss/blob/master/orchagent/neighorch.cpp#L639, it returns true if the create operation fails with the entry already exists. On the other hand, a false return value could mean retry is needed. Added a comment message to clarify. Also as it is assumed `status != SAI_STATUS_SUCCESS`, I added a check for `SAI_STATUS_SUCCESS` and log a warning when the function receives such a status.

orchagent/orch.cpp
shi-su: Using exit behavior instead.

orchagent/orch.cpp
shi-su: Added api info in the log and updated to use sai_serialize functions for api and status. It appears that cfgmgr also includes orch as a source, it is needed to include saimetadata library for cfgmgr.

orchagent/orch.cpp
shi-su: Updated as suggested.

orchagent/orch.cpp
shi-su: Updated as suggested.

shi-su:
> Did you tested any of this change ? since now it will throw exceptions on any failure

I tested this change in a few artificially generated SAI failures. The goal of this change is to make sure we do no worse than async mode, which triggers orchagent exit in any SAI failures. Throwing exceptions on any failure is basically trying to follow that behavior (assuming failures coming from sairedis is rare and most likely inferring some bugs in implementations).

qiluo-msft:
lgtm and okay to merge

---
In reply to: [571774867](https://github.com/Azure/sonic-swss/pull/1596#issuecomment-571774867) [](ancestors = 571774867)

qiluo-msft (approved):
Unblock

kcudnik (approved):

orchagent/orch.cpp
```diff
@@ -685,6 +686,82 @@ Executor *Orch::getExecutor(string executorName)
     return NULL;
 }
 
+bool Orch::handleSaiCreateStatus(sai_api_t api, sai_status_t status)
```
qiluo-msft, line 689: >status [](start = 61, length = 6)

Considering this is an interface design, and future user may have extended requirement. Is it reasonable to add another optional parameter like `void *context`, so user could pass any parameter they want and use it in their derived functions?

shi-su: Thanks for this comment. I added an optional parameter `void *context` to the functions.

qiluo-msft (approved):

PR #1597, opened 2021-01-12
[ci]: download artifacts from master branch build

### LINKED ISSUE #1589: Should we explicitly indicate the specific branch version?
https://github.com/Azure/sonic-swss/blob/693a02ceb669d6ecfb616183068225051aafe3e3/azure-pipelines.yml#L36-L42

Here is the Spec of AzurePipeline: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/download-pipeline-artifact?view=azure-devops

image

According to my understanding, if you don't explicitly indicate the artifact you want to download, it will fetch the latest build artifact that may be a PR artifact.
So , I think the following attributes that should be added to indicate a stable artifact in the master branch.
```yml
    runVersion: LatestFromBranch
    runBranch: refs/heads/master
```

lguohan:

**What I did**
download artifacts from master branch build

**Why I did it**
Fix #1589 

**How I verified it**

**Details if related**

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -39,8 +39,8 @@ jobs:
       project: build
       pipeline: 9
       artifacts: sonic-swss-common
-      runVersion: LatestFromBranch
-      runBranch: refs/heads/master
+      runVersion: 'latestFromBranch'
+      runBranch: 'refs/heads/master'
     displayName: "Download sonic swss common deb packages"
   - task: DownloadPipelineArtifact@2
     inputs:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -39,13 +39,17 @@ jobs:
       project: build
       pipeline: 9
       artifacts: sonic-swss-common
+      runVersion: 'latestFromBranch'
+      runBranch: 'refs/heads/master'
```
qiluo-msft, line 43: >master [](start = 29, length = 6)

Not a blocking issue: is it possible to figure out the branch automatically, so it will work on any branches like 202012, etc.

qiluo-msft (approved):

azure-pipelines.yml
lguohan: you are right.

PR #1602, opened 2021-01-18
[voq/inbandif] Voq inbandif port

vganesan-nokia:
**What I did**

Changes to support inband interface configuration

**Why I did it**

The VOQ chassis needs inband interface in the kernel for cpu to cpu communication across different asic instance and remote neighbor programming in the kernel. This PR adds support for configuring inband interface so that kernel interface can be created via sonic configuration.

**How I verified it**
swss vs test (test_virtual_chassis.py) to verify the remote neigh programming.

**Details if related**

Reference: VOQ HLD: https://github.com/Azure/SONiC/blob/master/doc/voq/voq_hld.md

### DIFF
--- a/cfgmgr/nbrmgr.cpp
+++ b/cfgmgr/nbrmgr.cpp
@@ -391,6 +391,9 @@ void NbrMgr::doStateSystemNeighTask(Consumer &consumer)
             if (!addKernelNeigh(nbr_odev, ip_address, mac_address))
             {
                 SWSS_LOG_ERROR("Neigh entry add on dev %s failed for '%s'", nbr_odev.c_str(), kfvKey(t).c_str());
+                // Delete neigh to take care of deletion of exiting nbr for mac change. This makes sure that
+                // re-try will be successful and route addtion (below) will be attempted and be successful
+                delKernelNeigh(nbr_odev, ip_address);
                 it++;
                 continue;
             }
--- a/cfgmgr/nbrmgr.cpp
+++ b/cfgmgr/nbrmgr.cpp
@@ -444,11 +444,11 @@ bool NbrMgr::isIntfOperUp(const string &alias)
 {
     string oper;
 
-    if (m_statePortTable.hget(alias, "oper_status", oper))
+    if (m_statePortTable.hget(alias, "netdev_oper_status", oper))
     {
         if (oper == "up")
         {
-            SWSS_LOG_DEBUG("Intf %s is oper up", alias.c_str());
+            SWSS_LOG_DEBUG("NetDev %s is oper up", alias.c_str());
             return true;
         }
     }
--- a/cfgmgr/nbrmgr.cpp
+++ b/cfgmgr/nbrmgr.cpp
@@ -348,7 +348,8 @@ void NbrMgr::doStateSystemNeighTask(Consumer &consumer)
     //Get the name of the device on which the neigh and route are
     //going to be programmed.
     string nbr_odev;
-    if(!getVoqInbandInterfaceName(nbr_odev))
+    string ibif_type;
+    if(!getVoqInbandInterfaceName(nbr_odev, ibif_type))
     {
         //The inband interface is not available yet
         return;
@@ -380,7 +381,7 @@ void NbrMgr::doStateSystemNeighTask(Consumer &consumer)
                     mac_address = MacAddress(fvValue(*i));
             }
 
-            if (!isIntfOperUp(nbr_odev))
+            if (ibif_type == "port" && !isIntfOperUp(nbr_odev))
             {
                 SWSS_LOG_DEBUG("Device %s is not oper up, skipping system neigh %s'", nbr_odev.c_str(), kfvKey(t).c_str());
                 it++;
@@ -455,9 +456,8 @@ bool NbrMgr::isIntfOperUp(const string &alias)
     return false;
 }
 
-bool NbrMgr::getVoqInbandInterfaceName(string &ibif)
+bool NbrMgr::getVoqInbandInterfaceName(string &ibif, string &type)
 {
-
     vector<string> keys;
     m_cfgVoqInbandInterfaceTable->getKeys(keys);
 
@@ -466,9 +466,21 @@ bool NbrMgr::getVoqInbandInterfaceName(string &ibif)
         SWSS_LOG_NOTICE("Voq Inband interface is not configured!");
         return false;
     }
-    //key:"alias" = inband interface name
+
+    // key:"alias" = inband interface name
+
     vector<string> if_keys = tokenize(keys[0], config_db_key_delimiter);
+
     ibif = if_keys[0];
+
+    // Get the type of the inband interface
+
+    if (!m_cfgVoqInbandInterfaceTable->hget(ibif, "inband_type", type))
+    {
+        SWSS_LOG_ERROR("Getting Voq Inband interface type failed for %s", ibif.c_str());
+        return false;
+    }
+
     return true;
 }
 
--- a/cfgmgr/nbrmgr.cpp
+++ b/cfgmgr/nbrmgr.cpp
@@ -380,9 +380,9 @@ void NbrMgr::doStateSystemNeighTask(Consumer &consumer)
                     mac_address = MacAddress(fvValue(*i));
             }
 
-            if (!isIntfStateOk(nbr_odev))
+            if (!isIntfOperUp(nbr_odev))
             {
-                SWSS_LOG_DEBUG("Interface %s is not ready, skipping system neigh %s'", nbr_odev.c_str(), kfvKey(t).c_str());
+                SWSS_LOG_DEBUG("Device %s is not oper up, skipping system neigh %s'", nbr_odev.c_str(), kfvKey(t).c_str());
                 it++;
                 continue;
             }
@@ -402,6 +402,8 @@ void NbrMgr::doStateSystemNeighTask(Consumer &consumer)
             {
                 SWSS_LOG_ERROR("Route entry add on dev %s failed for '%s'", nbr_odev.c_str(), kfvKey(t).c_str());
                 delKernelNeigh(nbr_odev, ip_address);
+                // Delete route to take care of deletion of exiting route of nbr for mac change.
+                delKernelRoute(ip_address);
                 it++;
                 continue;
             }
@@ -435,6 +437,22 @@ void NbrMgr::doStateSystemNeighTask(Consumer &consumer)
 
         it = consumer.m_toSync.erase(it);
     }
+}
+
+bool NbrMgr::isIntfOperUp(const string &alias)
+{
+    string oper;
+
+    if (m_statePortTable.hget(alias, "oper_status", oper))
+    {
+        if (oper == "up")
+        {
+            SWSS_LOG_DEBUG("Intf %s is oper up", alias.c_str());
+            return true;
+        }
+    }
+
+    return false;
 }
 
 bool NbrMgr::getVoqInbandInterfaceName(string &ibif)
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -1199,11 +1199,6 @@ bool NeighOrch::addInbandNeighbor(string alias, IpAddress ip_address)
     memcpy(attr.value.mac, inband_mac.getMac(), 6);
     neighbor_attrs.push_back(attr);
 
-    if(!addVoqEncapIndex(alias, ip_address, neighbor_attrs))
-    {
-        return false;
-    }
-
     //No host route for neighbor of the Inband IP address
     attr.id = SAI_NEIGHBOR_ENTRY_ATTR_NO_HOST_ROUTE;
     attr.value.booldata = true;
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -1025,6 +1025,13 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
         return;
     }
 
+    // Wait till the Inband interface is both admin up and oper up
+    if (ibif.m_admin_state_up != true || ibif.m_oper_status != SAI_PORT_OPER_STATUS_UP)
+    {
+        // Inband port is not operational yet
+        return;
+    }
+
     auto it = consumer.m_toSync.begin();
     while (it != consumer.m_toSync.end())
     {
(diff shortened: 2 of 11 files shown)


anshuv-mfst:
@eswaran (Arista), @abdosi - could you please review the PR, thanks.

orchagent/neighorch.cpp
```diff
@@ -1099,20 +1109,131 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
...
+        gCrmOrch->incCrmResUsedCounter(CrmResourceType::CRM_IPV6_NEIGHBOR);
+    }
+
+    //Sync the neighbor to add to the CHASSIS_APP_DB
+    voqSyncAddNeigh(alias, ip_address, inband_mac, neighbor_entry);
 
     return true;
 }
 
 bool NeighOrch::delInbandNeighbor(string alias, IpAddress ip_address)
```
ysmanman, line 1187: When will the local inband neigh be deleted? It should never happen, right?

vganesan-nokia: Yes. So far I do not have a compelling situation when we'll delete Inband neigh. Right now this API is not used anywhere since we do not delete the Inband interface once configured. We also do not support changing this once configured. May be in the future we may have a situation to dynamically change inband interface or ip address, when we'll use this.

cfgmgr/nbrmgr.cpp
```diff
@@ -380,6 +381,8 @@ void NbrMgr::doStateSystemNeighTask(Consumer &consumer)
             {
                 SWSS_LOG_ERROR("Route entry add on dev %s failed for '%s'", nbr_odev.c_str(), kfvKey(t).c_str());
                 delKernelNeigh(nbr_odev, ip_address);
+                // Delete route to take care of deletion of exiting route of nbr for mac change.
```
ysmanman, line 384: Can you elaborate why would this take care of mac change?

orchagent/portsorch.cpp
```diff
@@ -4925,21 +4927,20 @@ bool PortsOrch::setVoqInbandIntf(string &alias, string &type)
...
-            SWSS_LOG_NOTICE("Port configured for inband intf %s is not ready!", alias.c_str());
-            return false;
-        }
+        SWSS_LOG_ERROR("Port/Vlan configured for inband intf %s is not ready!", alias.c_str());
+        return false;
     }
 
-    // Check for existence of host interface. If does not exist, may create
-    // host if for the inband here
+    if(type == "port" && !port.m_hif_id)
```
ysmanman, line 4938: Why do we need this check? I remember setVoqInbandIntf() is called from Intfsorch::doTask, which will not process anything until all ports are ready (so hostIf must be created too). If inband port is processed just like front panel ports by portsorch, then hostIf of inban port must be created already before setVoqInbandIntf() is called.

orchagent/routeorch.cpp
```diff
@@ -1282,6 +1282,15 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
 
         if (nexthop.ip_address.isZero())
         {
+            if(gPortsOrch->isInbandPort(nexthop.alias))
```
ysmanman, line 1285: This is applied for both inband port and vlan, right?

tests/test_virtual_chassis.py
```diff
@@ -224,7 +273,79 @@ def test_chassis_system_neigh(self, vct):
                 assert encap_index == sys_neigh_encap_index, "Encap index not sync-ed correctly"
                 
                 break
+            
+        # Verify programming of remote neighbor in asic db and programming of static route and static
+        # neigh in the kernel for the remote neighbor. The neighbor created in linecard 1  will be a
```
ysmanman, line 278: Would it be easy to also cover the case of neighbor deletion too?

cfgmgr/nbrmgr.cpp
vganesan-nokia: When there is mac change for a neighbor, the system neighbor in the chassis app db is updated with new mac. There is no removal of system neighbor. The neighorch (doVoqSystemNeighTask()) just adds the neighbor in SAI with updated mac. But in the kernel, the static route still exits. After updating the SAI, the state db is updated to signal kernel programming. The kernel programming in nbrmgr first adds neigh and adds static route in the kernel. Since kernel already has the route, it returns an error. So we delete the route entry here so that in the next re-try the route will be programmed. 

Please also note that there will be error while trying to program the neighbor also since the kernel neighbor is tried to be programmed with same system mac (the mac of the inband port. For port type inband, in the kernel all neighbors will be programmed with the same mac - the system mac) even though there is mac change (only SAI is updated with chnaged mac). I'll fix this

orchagent/portsorch.cpp
vganesan-nokia: True. With the inband port is available in PORT table, by the time setVoqInbandIntf() is called, the host should have been created already. The check is here is kind of sanity check. Also, it gives the flexibility that you can have any port that is not part of the PORT table can be used for inband interface as long as we make the host if available, which can be done out side port config.

orchagent/routeorch.cpp
vganesan-nokia: Yes. since static routes of system neighbors are added for both types of inband, this is applicable for both.

tests/test_virtual_chassis.py
vganesan-nokia: Yes. I'll add a test case for delete neighbor.

vganesan-nokia: Fixed. Added test case for system neigh delete

cfgmgr/nbrmgr.cpp
vganesan-nokia: Fixed

vganesan-nokia:
retest this please

vganesan-nokia:
retest vs please

vganesan-nokia:
@lguohan, would you please help restarting the Azure-sonic-swss(Test vstes)? It seems there is some cleanup issue.  "retest this" is not responding. I do not have sufficient permission to start "AzurePipeline run"

ysmanman (dismissed):

vganesan-nokia:
Azurepipelines run

abdosi:
reviewing this.

orchagent/neighorch.cpp
```diff
@@ -1135,20 +1145,131 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
...
+    neighbor_entry.switch_id = gSwitchId;
+    copy(neighbor_entry.ip_address, ip_address);
+
+    vector<sai_attribute_t> neighbor_attrs;
+    sai_attribute_t attr;
+    attr.id = SAI_NEIGHBOR_ENTRY_ATTR_DST_MAC_ADDRESS;
+    memcpy(attr.value.mac, inband_mac.getMac(), 6);
+    neighbor_attrs.push_back(attr);
+
+    if(!addVoqEncapIndex(alias, ip_address, neighbor_attrs))
```
abdosi, line 1179: why we need this ?

vganesan-nokia: Usually for local neighbors we let the SAI allocate encap index and the neighbor along with the allocated encap index is synced to all the remote asics. Here we call the encap index addition for local neighbor to give the flexibility for user given encap index (user has to make sure that the encap index unique within asic). Simply configure the voq neighbor table with encap index attribute (similar to remote neighbor encap index) and that encap index will be sent to the SAI. So far we have not used this approach. Currently since we do not have encap index for local neighbors in voq neigh table, we always let the SAI allocate the encap index.
This api is similar to the addNeighbor() api that handles adding encap index for both local and remote neighbors except that this does not add next hop and the host route addition is masked (specifically required for inband port type neighbors)

abdosi: I don't understand 
if gIntfsOrch->isRemoteSystemPortIntf(alias) is True we return from this API. 
addVoqEncapIndex works only gIntfsOrch->isRemoteSystemPortIntf(alias) is True.
So why we are even calling it ?

orchagent/neighorch.cpp
```diff
@@ -1135,20 +1145,131 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
...
+    }
+
+    //No host route for neighbor of the Inband IP address
+    attr.id = SAI_NEIGHBOR_ENTRY_ATTR_NO_HOST_ROUTE;
+    attr.value.booldata = true;
+    neighbor_attrs.push_back(attr);
 
-    //For "vlan" type inband, the inband reachability info syncinng can be ARP learning of other
-    //asics inband or static configuration or through CHASSIS_APP_DB sync (this function)
+    status = sai_neighbor_api->create_neighbor_entry(&neighbor_entry, static_cast<uint32_t>(neighbor_attrs.size()), neighbor_attrs.data());
```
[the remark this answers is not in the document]
abdosi: my question what is the use of creating this information from SAI perspective. How this will be used later ?

orchagent/neighorch.cpp
vganesan-nokia: Good point. The addVoqEncapIndex() is supposed to get encap index from voq neighbor table only for remote system port neighbors. The addVoqEncapIndex() is called for all neihbors (local and remote neighbors) from addNeighbor() function. We need this remote system port interface check so that only for remote neighbors, encap index attribute will be sent to SAI. 
For addInbandNeighbor() since we are sure that this is a local neighbor, we can avoid calling addVoqEncapIndex() function. I'll fix this.

orchagent/neighorch.cpp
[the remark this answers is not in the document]
vganesan-nokia: Just a brief info about inband interface and recycle port: Inband interface is for cpu-to-cpu communication. In this PR we use recycle port facility for inband interface. Think of recycle port as any other network port but the egress side is looped back to the ingress. Since host is reached via inband interface for all the protocol communications we need to add the self neighbor on the inband interface. This is like any other neighbor on the front panel port. Since the neighbor resides within  the host of the asic itself, we use the recycle facility to loop back the packet and add neighbor to consume the packet. 
After adding the neighbor we inform the neighbor to other asics in the same we do for any neighbor on any other front panel port. Other asics program this neighbor on the system port of inband interface of this asic, When remote asics sends packet to this asic, they inject packet on their cpu port and everything else is similar to regular routing handled in the hardware.
For more information on this please refer to the section 2.5.1 of the voq HLD https://github.com/Azure/SONiC/blob/master/doc/voq/voq_hld.md.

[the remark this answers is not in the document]
abdosi: @vganesan-nokia thanks i had gone through this document before posting the question. I am not able to follow why there is need to create self neighbor for recycle port.  Let me know what i am missing in below :
LC1 CPU want to send packet destined to LC2 Inband IP and need to be consumed by LC2 CPU. 
For this my understanding is 

- Packet from LC1 CPU will read LC1 Recycle Port via NetDev
- Packet will do Host Lookup in LC1 Ingress ASIC and will send packet to LC2 Recylce Port
- Packet will reach LC2 Recycle Port and looped back into Ingress ASIC of LC2
- LC2 will have /32 route for its Inband IP and will go to LC2 CPU via Netdev of LC2 Recycle port.

So I am not sure where the self neighbor entry is coming into picture and how SAI will use it.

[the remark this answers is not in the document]
vganesan-nokia: > @vganesan-nokia thanks i had gone through this document before posting the question. I am not able to follow why there is need to create self neighbor for recycle port. Let me know what i am missing in below :
> LC1 CPU want to send packet destined to LC2 Inband IP and need to be consumed by LC2 CPU.
> For this my understanding is
> 
> * Packet from LC1 CPU will reach LC1 Recycle Port via NetDev
```[vganesan-nokia] >>> Here, the packet injection is on asic's cpu port with packet's destination information pointing to egress of recycle port.```
> * Packet will get looped back do Host Lookup in LC1 Ingress ASIC and will send packet to LC2 Recylce Port
```[vganesan-nokia] >>> For here, when the packet is looped back, i.e., when packet is received on ingress of the same LC1 asic it has to be directed to router block```. **For this consumption to happen we need the neighbor info of self on the recycle port**. ```Once packet is in routing operation block, the routing operation determines the destination information i.e., the system port of inband interface of LC2```
> * Packet will reach LC2 Recycle Port and looped back into Ingress ASIC of LC2
```[vganesan-nokia] >>> Here from fabric, the packet reaches the egress of the recycle port in LC2. This is looped back to ingress where the host IP trap rule sends packet to CPU.``` 
> * LC2 will have /32 route for its Inband IP and will go to LC2 CPU via Netdev of LC2 Recycle port.
> 
> So I am not sure where the self neighbor entry is coming into picture and how SAI will use it.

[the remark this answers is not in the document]
skeesara-nokia: The inband port IP address of the "my-asic" is the next-hop in the routing tables of the "other-asics" - for routes that point to "my-asic" host IP stack. So "my-asic" needs to create this neighbor and let the other asics know about this neighbor and its encap index. This will allow the other asics to host traffic destined to "my-asic"me via the inband port of "my-asic".

This is the only reason to create this neighbor. Please note that for this neighbor we do not create a host route in SAI. The neighbor attribute SAI_NEIGHBOR_ENTRY_ATTR_NO_HOST_ROUTE is used for this purpose. We do that becuase -
1. The IP2ME trap rules can already send the packet to the CPU once it comes around to the ingress of the recyle port.
2. If we have the host-route for this neighbor - it loop around the recyle port instead of being trapped to the CPU.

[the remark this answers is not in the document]
skeesara-nokia: > 
> 
> sorry i forgot. Can you please tells us the reason why we are creating Neighbor record on SAI for the local Inband/Recycle Port interface ? How this will get used ? Which section of design document cover the need for doing this ?

[skeesara-nokia] This is needed so that we get an encap index that the other asics can use it for the case where the next-hop is the inband port IP of my-asic

orchagent/neighorch.cpp
abdosi: Has this been fixed ?

vganesan-nokia: Yes. Fixed.

orchagent/neighorch.cpp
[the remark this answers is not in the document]
abdosi: As discussed please update HLD with the information of L2Header (Src/Dest Mac) being used for the above packat flow.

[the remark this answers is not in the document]
skeesara-nokia: Done. A PR has been raised with the requested update.

https://github.com/Azure/SONiC/pull/776

abdosi (approved):

PR #1603, opened 2021-01-18
[vog/systemlag] Voq lagid allocator

vganesan-nokia:
**What I did**
Defined class for lag id allocator and added lua script for allocating/freeing lag id in atomic fashion

**Why I did it**

For portchannels in VOQ based chassis systems we need unique lag id across
the system. The lag id (aka system port aggreggator id) is allocated during portchannel
creation. The changes are for a class for lag id allocation in atomic fashion. The LAG ID is allocated
from central chassis app db. A lua script loaded in the redis at the
time of lag id allocator instantiation ensures allocating unique lag id when
multiple clients requests for lag id simultaneously. 

Ref: VOQ LAG HLD PR: https://github.com/Azure/SONiC/pull/697

**How I verified it**

- In VOQ based chassis systems, make sure chassis app db is initialized with lag id start and lag id end data in supervisor card.
- Configure port channel in one asic instance.
- Check the ASIC DB for system port aggregator id (spa) attribute programmed and verify that this spa value is not duplicated in portchannels from different ASICs.

**Details if related**

Note: This PR is dependent on sonic-swss-common PR https://github.com/Azure/sonic-swss-common/pull/447

### DIFF
--- a/orchagent/lagid.h
+++ b/orchagent/lagid.h
@@ -1,0 +1,44 @@
+#ifndef SWSS_LAGID_H
+#define SWSS_LAGID_H
+
+#include "dbconnector.h"
+#include "sal.h"
+#include "schema.h"
+#include "redisapi.h"
+
+using namespace swss;
+using namespace std;
+
+#define LAG_ID_ALLOCATOR_ERROR_DELETE_ENTRY_NOT_FOUND  0
+#define LAG_ID_ALLOCATOR_ERROR_TABLE_FULL             -1
+#define LAG_ID_ALLOCATOR_ERROR_GET_ENTRY_NOT_FOUND    -2
+#define LAG_ID_ALLOCATOR_ERROR_INVALID_OP             -3
+#define LAG_ID_ALLOCATOR_ERROR_DB_ERROR               -4
+
+class LagIdAllocator
+{
+public:
+
+    LagIdAllocator(
+            _In_ DBConnector* chassis_app_db);
+
+public:
+
+    int32_t lagIdAdd(
+            _In_ const string &pcname,
+            _In_ int32_t lag_id);
+
+    int32_t lagIdDel(
+            _In_ const string &pcname);
+
+    int32_t lagIdGet(
+            _In_ const string &pcname);
+
+private:
+
+    DBConnector* m_dbConnector;
+
+    string m_shaLagId;
+};
+
+#endif // SWSS_LAGID_H
--- a/orchagent/lagids.lua
+++ b/orchagent/lagids.lua
@@ -4,9 +4,9 @@
 -- ARGV[3] - current lag id (for "add" operation only)
 
 -- return lagid if success for "add"/"del"
--- return 0 if lag not exists for "del"
+-- return 0 if lag does not exist for "del"
 -- return -1 if lag table full for "add"
--- return -2 if log does exist for "get"
+-- return -2 if lag does not exist for "get"
 -- return -3 if invalid operation
 
 local op = ARGV[1]
@@ -24,8 +24,7 @@ if op == "add" then
     if dblagid then
         dblagid = tonumber(dblagid)
         if plagid == 0 then
-            -- no lagid propsed. Return the existing lagid
-            redis.call("sadd", "SYSTEM_LAG_ID_SET", tostring(dblagid))
+            -- no lagid proposed. Return the existing lagid
             return dblagid
         end
     end
@@ -34,7 +33,6 @@ if op == "add" then
     if plagid >= lagid_start and plagid <= lagid_end then
         if plagid == dblagid then
             -- proposed lagid is same as the lagid in database
-            redis.call("sadd", "SYSTEM_LAG_ID_SET", tostring(plagid))
             return plagid
         end
         -- proposed lag id is different than that in database OR
--- a/orchagent/lagids.lua
+++ b/orchagent/lagids.lua
@@ -1,4 +1,4 @@
--- KEYS - lagid_start, lagid_end, lagids, lagid_set
+-- KEYS - None
 -- ARGV[1] - operation (add/del/get)
 -- ARGV[2] - lag name
 -- ARGV[3] - current lag id (for "add" operation only)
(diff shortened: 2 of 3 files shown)


anshuv-mfst:
@judyjoseph - could you please review the PR, thanks.

orchagent/lagid.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include "dbconnector.h"
+#include "sal.h"
+#include "schema.h"
+#include "redisapi.h"
+
+using namespace swss;
+using namespace std;
+
+#define LAG_ID_ALLOCATOR_ERROR_DELETE_ENTRY_NOT_FOUND  0
```
judyjoseph, line 12: Not finding these error codes being used, can be used in lagids.lua ?

orchagent/lagids.lua
```diff
@@ -0,0 +1,90 @@
+-- KEYS - None
+-- ARGV[1] - operation (add/del/get)
+-- ARGV[2] - lag name
+-- ARGV[3] - current lag id (for "add" operation only)
+
+-- return lagid if success for "add"/"del"
+-- return 0 if lag not exists for "del"
+-- return -1 if lag table full for "add"
+-- return -2 if log does exist for "get"
```
judyjoseph, line 9: Thanks for these comments, misspelt "lag" as "log"

orchagent/lagids.lua
```diff
@@ -0,0 +1,90 @@
...
+            redis.call("sadd", "SYSTEM_LAG_ID_SET", tostring(dblagid))
+            return dblagid
+        end
+    end
+
+    -- lagid allocation request with a lagid proposal
+    if plagid >= lagid_start and plagid <= lagid_end then
+        if plagid == dblagid then
+            -- proposed lagid is same as the lagid in database
+            redis.call("sadd", "SYSTEM_LAG_ID_SET", tostring(plagid))
```
judyjoseph, line 37: We could actually skip this redis.call, as the SET already has this id ?

orchagent/lagids.lua
```diff
@@ -0,0 +1,90 @@
...
+            return plagid
+        end
+        -- proposed lag id is different than that in database OR
+        -- the portchannel does not exist in the database
+        -- If proposed lagid is available, return the same proposed lag id
+        if redis.call("sismember", "SYSTEM_LAG_ID_SET", tostring(plagid)) == 0 then
+            redis.call("sadd", "SYSTEM_LAG_ID_SET", tostring(plagid))
+            redis.call("srem", "SYSTEM_LAG_ID_SET", tostring(dblagid))
```
judyjoseph, line 45: dblagid could be invalid, incase the portchannel doesn't exist in DB. Can you confirm this call don't error out.

orchagent/lagids.lua
```diff
@@ -0,0 +1,90 @@
...
+
+    return 0
+
+end
+
+if op == "get" then
+
+    if redis.call("hexists", "SYSTEM_LAG_ID_TABLE", pcname) == 1 then
+        local lagid = redis.call("hget", "SYSTEM_LAG_ID_TABLE", pcname)
+        return tonumber(lagid)
```
judyjoseph, line 83: Should we convert tonumber() ? as I think the caller expects string

orchagent/lagids.lua
```diff
@@ -0,0 +1,90 @@
...
+
+end
+
+if op == "del" then
+
+    if redis.call("hexists", "SYSTEM_LAG_ID_TABLE", pcname) == 1 then
+        local lagid = redis.call("hget", "SYSTEM_LAG_ID_TABLE", pcname)
+        redis.call("srem", "SYSTEM_LAG_ID_SET", lagid)
+        redis.call("hdel", "SYSTEM_LAG_ID_TABLE", pcname)
+        return tonumber(lagid)
```
judyjoseph, line 72: Should we convert tonumber() ? as I think the caller expects string

orchagent/lagids.lua
```diff
@@ -0,0 +1,90 @@
...
+        dblagid = tonumber(dblagid)
+        if plagid == 0 then
+            -- no lagid propsed. Return the existing lagid
+            redis.call("sadd", "SYSTEM_LAG_ID_SET", tostring(dblagid))
+            return dblagid
+        end
+    end
+
+    -- lagid allocation request with a lagid proposal
+    if plagid >= lagid_start and plagid <= lagid_end then
```
judyjoseph, line 34: Need to handle case when plagid is invalid, need to return back error.

orchagent/lagid.h
vganesan-nokia: Yes. These errors are used in lagids.lua as hardcode. Here I have just defined these in order to indicate that these error numbers are taken to avoid conflict with any new error definition in the future

orchagent/lagids.lua
vganesan-nokia: Sorry for the typo. I'll fix them.

orchagent/lagids.lua
vganesan-nokia: Yes. "sadd" can be skipped here. I'll fix this. 
The "sadd" in line#28 can also be skipped. I'll fix this also.

orchagent/lagids.lua
vganesan-nokia: This will not error out, instead a new lag id will be allocated.
If the portchannel does not exist, dblagid will be "".  In that case, if the proposed lagid is available, we allocate the proposed lag id. If the proposed lad is not available, we continue and allocate new lagid.

orchagent/lagids.lua
vganesan-nokia: Yes. We need to return a number since the caller expects a number. runRedisScript() is enhanced in PR https://github.com/Azure/sonic-swss-common/pull/447 for this

orchagent/lagids.lua
vganesan-nokia: Yes. The caller expects a number.

orchagent/lagids.lua
vganesan-nokia: The only possible value that will fail this condition is the non-proposal value of 0. In that case, we must allocate a new lag id.
The lag id proposal is never user-driven. Either it will be 0 (i.e, no lag id proposed) or a valid lagid value within boundary (during warmboot, if implemented in the future, when portchannel will exist and existing lagid will be returned).

orchagent/lagids.lua
vganesan-nokia: Fixed

orchagent/lagids.lua
vganesan-nokia: Fixed

vganesan-nokia:
retest vs please

vganesan-nokia:
retest vs please

judyjoseph:
retest this please

judyjoseph:
retest vs please

vganesan-nokia:
retest vs please

orchagent/lagids.lua
```diff
@@ -0,0 +1,88 @@
+-- KEYS - None
```
judyjoseph, line 1: Shouldn't we add this new lagids.lua  into Makefile ?

judyjoseph:
retest vs please

judyjoseph:
retest this please

orchagent/lagids.lua
vganesan-nokia: Yes. It is added in the make file when it is used. In this PR, we just implement the lag id allocator. We are not using it yet. In PR #1605 we update the make file and use the lag id allocator.

judyjoseph (approved):
LGTM

xumia:
retest vs please

vganesan-nokia:
retest vs please

judyjoseph:
vs tests it not required anymore, Azure Pipeline run is good enough. I will merge it,

vganesan-nokia:
> vs tests it not required anymore, Azure Pipeline run is good enough. I will merge it,

o.k. thanks @judyjoseph. The system lag functionality PR (https://github.com/Azure/sonic-swss/pull/1605) depends on this PR to build successfully.

PR #1605, opened 2021-01-18
[voq/systemlag] Voq system lag functionality

vganesan-nokia:
**What I did**

System LAG (PortChannel for chassis system) support

**Why I did it**

In VOQ based chassis systems, to forward across different ASICs through Portchannels configured in different asics, a new attribute called "system port aggregator id" (spa) is defined for the LAG objects. The system lag implementation in this PR allocates unique ID for Portchannel ans the id is sent to SAI to create the LAG with spa. This chassis wide unique spa is sync-ed with chassis app db so that non-owner asics will create corresponding remote LAGs with the same spa.

For more detailed information refer: Distributed VOQ LAG HLD https://github.com/Azure/SONiC/pull/697

**How I verified it**

- test_virtual_chassis.py with system lag test case passes

**Details if related**

This PR depends on the following PRs:
- https://github.com/Azure/sonic-swss-common/pull/447
- https://github.com/Azure/sonic-swss/pull/1603
- https://github.com/Azure/sonic-swss-common/pull/448
- https://github.com/Azure/sonic-buildimage/pull/6488

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -1363,9 +1363,10 @@ bool IntfsOrch::isRemoteSystemPortIntf(string alias)
     Port port;
     if(gPortsOrch->getPort(alias, port))
     {
-
         if (port.m_type == Port::LAG)
+        {
             return(port.m_system_lag_info.switch_id != gVoqMySwitchId);
+        }
 
         return(port.m_system_port_info.type == SAI_SYSTEM_PORT_TYPE_REMOTE);
     }
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -35,6 +35,7 @@ extern BufferOrch *gBufferOrch;
 extern bool gIsNatSupported;
 extern NeighOrch *gNeighOrch;
 extern string gMySwitchType;
+extern int32_t gVoqMySwitchId;
 
 const int intfsorch_pri = 35;
 
@@ -1362,9 +1363,13 @@ bool IntfsOrch::isRemoteSystemPortIntf(string alias)
     Port port;
     if(gPortsOrch->getPort(alias, port))
     {
+
+        if (port.m_type == Port::LAG)
+            return(port.m_system_lag_info.switch_id != gVoqMySwitchId);
+
         return(port.m_system_port_info.type == SAI_SYSTEM_PORT_TYPE_REMOTE);
     }
-    //Given alias is system port alias of the local port
+    //Given alias is system port alias of the local port/LAG
     return false;
 }
 
@@ -1375,11 +1380,22 @@ void IntfsOrch::voqSyncAddIntf(string &alias)
     Port port;
     if(gPortsOrch->getPort(alias, port))
     {
-        if(port.m_system_port_info.type == SAI_SYSTEM_PORT_TYPE_REMOTE)
+        if (port.m_type == Port::LAG)
         {
-            return;
+            if (port.m_system_lag_info.switch_id != gVoqMySwitchId)
+            {
+                return;
+            }
+            alias = port.m_system_lag_info.alias;
         }
-        alias = port.m_system_port_info.alias;
+        else
+        {
+            if(port.m_system_port_info.type == SAI_SYSTEM_PORT_TYPE_REMOTE)
+            {
+                return;
+            }
+            alias = port.m_system_port_info.alias;
+        }
     }
     else
     {
@@ -1401,11 +1417,22 @@ void IntfsOrch::voqSyncDelIntf(string &alias)
     Port port;
     if(gPortsOrch->getPort(alias, port))
     {
-        if(port.m_system_port_info.type == SAI_SYSTEM_PORT_TYPE_REMOTE)
+        if (port.m_type == Port::LAG)
         {
-            return;
+            if (port.m_system_lag_info.switch_id != gVoqMySwitchId)
+            {
+                return;
+            }
+            alias = port.m_system_lag_info.alias;
         }
-        alias = port.m_system_port_info.alias;
+        else
+        {
+            if(port.m_system_port_info.type == SAI_SYSTEM_PORT_TYPE_REMOTE)
+            {
+                return;
+            }
+            alias = port.m_system_port_info.alias;
+        }
     }
     else
     {
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -28,6 +28,7 @@
 #include "countercheckorch.h"
 #include "notifier.h"
 #include "fdborch.h"
+#include "subscriberstatetable.h"
 
 extern sai_switch_api_t *sai_switch_api;
 extern sai_bridge_api_t *sai_bridge_api;
@@ -46,6 +47,10 @@ extern BufferOrch *gBufferOrch;
 extern FdbOrch *gFdbOrch;
 extern Directory<Orch*> gDirectory;
 extern sai_system_port_api_t *sai_system_port_api;
+extern string gMySwitchType;
+extern int32_t gVoqMySwitchId;
+extern string gMyHostName;
+extern string gMyAsicName;
 
 #define VLAN_PREFIX         "Vlan"
 #define DEFAULT_VLAN_ID     1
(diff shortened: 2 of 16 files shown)


anshuv-mfst:
@ysmanman (Arista) - could you please review this PR, thanks.

orchagent/intfsorch.cpp
```diff
@@ -1362,9 +1363,13 @@ bool IntfsOrch::isRemoteSystemPortIntf(string alias)
     Port port;
     if(gPortsOrch->getPort(alias, port))
     {
+
+        if (port.m_type == Port::LAG)
+            return(port.m_system_lag_info.switch_id != gVoqMySwitchId);
```
ysmanman, line 1368: Can we put the statement in {} to be consistent with the rest of the change?

orchagent/intfsorch.cpp
```diff
@@ -1375,11 +1380,22 @@ void IntfsOrch::voqSyncAddIntf(string &alias)
...
+            alias = port.m_system_lag_info.alias;
+        }
+        else
         {
-            return;
+            if(port.m_system_port_info.type == SAI_SYSTEM_PORT_TYPE_REMOTE)
+            {
+                return;
+            }
+            alias = port.m_system_port_info.alias;
```
ysmanman, line 1397: This piece of code is duplicated in another function like voqSyncDelIntf too . Can we wrap this into a helper method? This can avoid duplicating code and also make it easy to extend it in the future.

orchagent/intfsorch.cpp
vganesan-nokia: o.k. I'll fix this.

orchagent/intfsorch.cpp
vganesan-nokia: This is done intentionally. This repeating will be there in other places also. The intention of this logic is to get whether the port or the lag is remote or not. Since we need system alias which in the Port structure,, we get the Port structure first. Since the port structure is already available and it also has the type info, I check the type in the port structure itself instead of calling another API.

orchagent/portsorch.cpp
```diff
@@ -2937,12 +2968,37 @@ void PortsOrch::doLagTask(Consumer &consumer)
...
+                }
+            }
+
+            if (table_name == CHASSIS_APP_LAG_TABLE_NAME)
+            {
+                if (switch_id == gVoqMySwitchId)
+                {
+                    //Already created, syncd local lag from CHASSIS_APP_DB. Skip
+                    it = consumer.m_toSync.erase(it);
+                    continue;
```
ysmanman, line 2987: Is it always safe to simply skip here? Is it possible this is to retry a lag that was failing to add?

orchagent/portsorch.cpp
```diff
@@ -3909,7 +4051,7 @@ bool PortsOrch::addLagMember(Port &lag, Port &port, bool enableForwarding)
     attr.value.oid = port.m_port_id;
     attrs.push_back(attr);
 
-    if (!enableForwarding)
+    if (!enableForwarding && port.m_type != Port::SYSTEM)
```
ysmanman, line 4054: Does this mean we cannot disable forwarding for sys lag?

orchagent/portsorch.cpp
vganesan-nokia: Yes. We should always skip this. This lag is the Portchannel locally created already and synced to CHASSIS_APP_DB. Since this api, which is used for both local lags and remote lags, this is invoked for entries added to CHASSIS_APP_DB also. Here we identify the local entry by the switch id. If it is local entry, since this local is already programmed, we should always skip this and erase it. So there will not be re-try at all.

orchagent/portsorch.cpp
vganesan-nokia: Setting (enable or disable) the ingress and egress is not applicable for remote LAG members. SAI returns errors when we set these attributes for system port lag members These are done at the owner ASIC (as local physical port members)

tests/test_virtual_chassis.py
```diff
@@ -224,7 +256,145 @@ def test_chassis_system_neigh(self, vct):
...
+        
+    def test_chassis_system_lag(self, vct):
+        """Test portchannel in VOQ based chassis systems.
+        
+        This test validates that
+           (i)   PortChannel is created in local asic with system port aggregator id (system lag id)
+                     - Unique lag id is allocated from chassis app db in supervisor card
+                     - The unique lag id is sent in system port aggregator id attribute
+           (ii)  PortChannel members are successfully added in the PortChannel created 
+           (iii) Local PortChannel is synced in chassis app db
```
judyjoseph, line 268: Can we add few negative tests to cover the failure conditions to cover the logic in lua id generation either here .. or in the other PR https://github.com/Azure/sonic-swss/pull/1603

vganesan-nokia: Yes. I'll add test cases for table full error. For delete and get error, it will be difficult to drive the tests via app db. The execution will not go that for to invoke the lagid allocator apis. I'll try to add test cases for these by invoking lua directly from cli using redis eval command, if possible

orchagent/intfsorch.cpp
vganesan-nokia: Fixed

tests/test_virtual_chassis.py
vganesan-nokia: Fixed. Added test cases to test lag id allocator table full error,  lag id delete error, re-use of released lag id.

ysmanman (dismissed):

orchagent/portsorch.cpp
```diff
@@ -4042,6 +4199,12 @@ bool PortsOrch::removeLagMember(Port &lag, Port &port)
     LagMemberUpdate update = { lag, port, false };
     notify(SUBJECT_TYPE_LAG_MEMBER_CHANGE, static_cast<void *>(&update));
 
+    if (gMySwitchType == "voq")
```
judyjoseph, line 4202: Similarly should we check for SwitchId here ? or are we handling it with the parent lag interface ?

orchagent/portsorch.cpp
```diff
@@ -4007,6 +4158,12 @@ bool PortsOrch::addLagMember(Port &lag, Port &port, bool enableForwarding)
     LagMemberUpdate update = { lag, port, true };
     notify(SUBJECT_TYPE_LAG_MEMBER_CHANGE, static_cast<void *>(&update));
 
+    if (gMySwitchType == "voq")
```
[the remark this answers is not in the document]
vganesan-nokia: The syncing function voqSyncAddLagMember() has check for this. If the lag's switch id is not equal to the local switch id, i.e., if the lag is remote, it will not be sync-ed.

orchagent/portsorch.cpp
vganesan-nokia: This lag's switch id is checked in syncing function voqSyncDelLagMember() for whether it is remote lag or not. If the lag is remote, no syncing will happen.

judyjoseph:
@vganesan-nokia There are some build errors can you take a look. Let me know if you are not able to check the <Details> link.

vganesan-nokia:
> @vganesan-nokia There are some build errors can you take a look. Let me know if you are not able to check the

This is expected. We need the lag id allocator PR https://github.com/Azure/sonic-swss/pull/1603 to be merged for this to build successfully

orchagent/portsorch.cpp
```diff
@@ -3851,12 +3930,43 @@ bool PortsOrch::isVlanMember(Port &vlan, Port &port)
...
+            // Local PortChannel. Allocate unique lag id from central CHASSIS_APP_DB
+            // Use the chassis wide unique system lag name.
+
+            // Get the local switch id and derive the system lag name.
+
+            switch_id = gVoqMySwitchId;
+            system_lag_alias = gMyHostName + "|" + gMyAsicName + "|" + lag_alias;
+
+            // Allocate unique lag id
+            spa_id = m_lagIdAllocator->lagIdAdd(system_lag_alias, 0);
```
judyjoseph, line 3953: Currently Sonic in the teammgrd we don't set a unique oper key in the LACP BPDU, but we are in the process of adding this support (https://github.com/Azure/sonic-swss/pull/1660)

We could use this unique id allocation schema there as well. The change there would be to allocate the lagId in teammgrd when user creates a new LAG .. and here in orchagent it should give back the same id as it should be already existing ( created before in teammgrd and updated in SYSTEM LAG TABLE.)

vganesan-nokia: The need for this unique lag id allocation is to have unique lag id for all port channels across all the asics. So we let the id allocation be done in centralized location. As you mentioned, it will be interesting use case to use uniquely allocated id as key for LACP. This will work well for multi-asic and distributed chassis. For single asic, we need to set the id boundaries and load the script in local redis server. Yes, the lag id allocator gives the same id for a given portchannel, if it is already allocated,

vganesan-nokia:
Capturing review approval from @ysmanman before rebasing

image

vganesan-nokia:
Hi @judyjoseph, 4 tests of fine grained next group test suite failed. I tried to debug this. But I am unable to simulate in my machine. For me all tests (372 tests) passed. I think there may be some timing issue. Would you please re-start the Azurepipelines again?. Thanks

orchagent/portsorch.cpp
judyjoseph: Yes in future we would need to extend this to teamd, so that we use this logic with the LUA script itself to generate a unique LACP key filled in the LACP BPDU.

judyjoseph (approved):

PR #1606, opened 2021-01-23
[fpmsyncd] Skip routes to eth0 or docker0

shi-su:
**What I did**
Skip routes to eth0 or docker0 in fpmsyncd.

Fix https://github.com/Azure/sonic-buildimage/issues/6483

**Why I did it**
An FRR behavior change from 7.2 to 7.5 makes FRR update the default route to eth0 in interface up/down events. Skipping routes to eth0 or docker0 to avoid such behavior.

**How I verified it**
Tested locally that the default route does not change in interface up/down events.

**Details if related**

### DIFF
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -679,7 +679,7 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
     {
         /*
          * An FRR behavior change from 7.2 to 7.5 makes FRR update default route to eth0 in interface
-         * up/down events. Skipping routes to eth0 or docker0 to aviod such behavior
+         * up/down events. Skipping routes to eth0 or docker0 to avoid such behavior
          */
         if (alias == "eth0" || alias == "docker0")
         {
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -674,6 +674,21 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
     string nexthops = getNextHopGw(route_obj);
     string ifnames = getNextHopIf(route_obj);
 
+    vector<string> alsv = tokenize(ifnames, ',');
+    for (auto alias : alsv)
+    {
+        /*
+         * An FRR behavior change from 7.2 to 7.5 makes FRR update default route to eth0 in interface
+         * up/down events. Skipping routes to eth0 or docker0 to aviod such behavior
+         */
+        if (alias == "eth0" || alias == "docker0")
+        {
+            SWSS_LOG_NOTICE("Skip routes to eth0 or docker0: %s %s %s",
+                    destipprefix, nexthops.c_str(), ifnames.c_str());
+            return;
+        }
+    }
+
     vector<FieldValueTuple> fvVector;
     FieldValueTuple nh("nexthop", nexthops);
     FieldValueTuple idx("ifname", ifnames);


lguohan (approved):

imzyxwvu:
Which FRR behaviour change is it that causes fpmsyncd to update default route? What did this behaviour break?

fpmsyncd/routesync.cpp
```diff
@@ -674,6 +674,21 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
...
+    {
+        /*
+         * An FRR behavior change from 7.2 to 7.5 makes FRR update default route to eth0 in interface
+         * up/down events. Skipping routes to eth0 or docker0 to avoid such behavior
+         */
+        if (alias == "eth0" || alias == "docker0")
+        {
+            SWSS_LOG_NOTICE("Skip routes to eth0 or docker0: %s %s %s",
+                    destipprefix, nexthops.c_str(), ifnames.c_str());
+            return;
```
imzyxwvu, line 688: Take this as a example: At first route 0.0.0.0/0  had a multipath nexthop "Ethernet0,Ethernet1", and some minutes later it became "Ethernet0,Ethernet1,eth0". Last it becomes "eth0". Since the change filters all routes including the eth0 nexthop, in the example this changes prevent "Ethernet0,Ethernet1" from exiting the next-hop list in ASIC. This is a rare situation but I suggest to take this into consideration.

lguohan: in our design, we would not never has such mixed scenario.

fpmsyncd/routesync.cpp
```diff
@@ -674,6 +674,21 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
...
+    vector<string> alsv = tokenize(ifnames, ',');
+    for (auto alias : alsv)
+    {
+        /*
+         * An FRR behavior change from 7.2 to 7.5 makes FRR update default route to eth0 in interface
+         * up/down events. Skipping routes to eth0 or docker0 to avoid such behavior
+         */
+        if (alias == "eth0" || alias == "docker0")
+        {
+            SWSS_LOG_NOTICE("Skip routes to eth0 or docker0: %s %s %s",
```
prsunny, line 686: Minor comment, route updates can be in INFO level log.

imzyxwvu:
After some investigation, I figured that the problem is actually introduced by [this change](https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-frr/patch/0003-Use-vrf_id-for-vrf-not-tabled_id.patch). SONiC uses `rtm_table` field (which is designed to pass table ID) of netlink_route protocol to pass VRF ID to FPM, and table ID info is lost. Generally there should be only table `main` in the default VRF, so this is harmless. But in fact zebra allows multiple route tables per VRF (or at least the main VRF). This leads to a mix of default VRF route updates across all tables and leaves a broken APPL_DB ROUTE_TABLE.

See `other_tables` field in `struct zebra_vrf`: https://github.com/Azure/sonic-frr/blob/frr/7.5/zebra/zebra_vrf.h#L85

shi-su:
> After some investigation, I figured that the problem is actually introduced by [this change](https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-frr/patch/0003-Use-vrf_id-for-vrf-not-tabled_id.patch). SONiC uses `rtm_table` field (which is designed to pass table ID) of netlink_route protocol to pass VRF ID to FPM, and table ID info is lost. Generally there should be only table `main` in the default VRF, so this is harmless. But in fact zebra allows multiple route tables per VRF (or at least the main VRF). This leads to a mix of default VRF route updates across all tables and leaves a broken APPL_DB ROUTE_TABLE.
> 
> See `other_tables` field in `struct zebra_vrf`: https://github.com/Azure/sonic-frr/blob/frr/7.5/zebra/zebra_vrf.h#L85

@imzyxwvu Thanks for your investigation. I still have some trouble understanding how `other_tables` leads to this behavior. And it seems that this patch is basically reverting the changes in https://github.com/FRRouting/frr/commit/6dfcd75461ff35ebf89e86f64fe6e982ed09b3e3

imzyxwvu:
> @imzyxwvu Thanks for your investigation. I still have some trouble understanding how `other_tables` leads to this behavior. And it seems that this patch is basically reverting the changes in [FRRouting/frr@6dfcd75](https://github.com/FRRouting/frr/commit/6dfcd75461ff35ebf89e86f64fe6e982ed09b3e3)

I'd suggest you to have a look at `zebra_vrf_lookup_table_with_table_id`.

https://github.com/Azure/sonic-frr/blob/frr/7.5/zebra/zebra_vrf.c#L361

This function includes a lookup on `other_tables`. The comment also contains some information on `table_id`.

PR #1611, opened 2021-01-26
Port serdes changes

dgsudharsan:
**What I did**
Migrating the port serdes programming from existing port attributes to port serdes object

**Why I did it**
This change is required for supporting 6 tap filters which cannot be supported through existing port attributes

**How I verified it**
Load the changes on existing platforms and verify the pre-emphasis settings are correctly programmed in the hardware
Load the changes on new platforms requiring 6tap filter and verified the settings are programmed correctly in the hardware

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2238,10 +2238,6 @@ void PortsOrch::doPortTask(Consumer &consumer)
                     getPortSerdesVal(fvValue(i), attr_val);
                     serdes_attr.insert(serdes_attr_pair(SAI_PORT_SERDES_ATTR_TX_FIR_ATTN, attr_val));
                 }
-                else
-                {
-                    SWSS_LOG_ERROR("Unknown port field: %s", fvField(i).c_str());
-                }
             }
 
             /* Collect information about all received ports */
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1944,6 +1944,8 @@ void PortsOrch::deInitPort(string alias, sai_object_id_t port_id)
     /* remove port name map from counter table */
     m_counter_db->hdel(COUNTERS_PORT_NAME_MAP, alias);
 
+    removePortSerdesAttribute(p.m_port_id);
+
     m_portList[alias].m_init = false;
     SWSS_LOG_NOTICE("De-Initialized port %s", alias.c_str());
 }
@@ -2102,9 +2104,9 @@ void PortsOrch::doPortTask(Consumer &consumer)
         if (op == SET_COMMAND)
         {
             set<int> lane_set;
-            vector<uint32_t> pre_emphasis;
-            vector<uint32_t> idriver;
-            vector<uint32_t> ipredriver;
+            vector<uint32_t> attr_val;
+            map<sai_port_serdes_attr_t, vector<uint32_t>> serdes_attr;
+            typedef pair<sai_port_serdes_attr_t, vector<uint32_t>> serdes_attr_pair;
             string admin_status;
             string fec_mode;
             string pfc_asym;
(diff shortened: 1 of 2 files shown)


dgsudharsan:
retest this please

dgsudharsan:
retest vs please

orchagent/portsorch.cpp
```diff
@@ -2177,22 +2179,92 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 }
 
                 /* Set port serdes Pre-emphasis */
+                attr_val.clear();
                 if (fvField(i) == "preemphasis")
                 {
-                    getPortSerdesVal(fvValue(i), pre_emphasis);
+                    getPortSerdesVal(fvValue(i), attr_val);
+                    serdes_attr.insert(serdes_attr_pair(SAI_PORT_SERDES_ATTR_PREEMPHASIS,attr_val));
```
jleveque, line 2186: Add space after comma

orchagent/portsorch.cpp
```diff
@@ -2177,22 +2179,92 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
                 }
 
+                attr_val.clear();
                 /* Set port serdes idriver */
                 if (fvField(i) == "idriver")
                 {
-                    getPortSerdesVal(fvValue(i), idriver);
+                    getPortSerdesVal(fvValue(i), attr_val);
+                    serdes_attr.insert(serdes_attr_pair(SAI_PORT_SERDES_ATTR_IDRIVER,attr_val));
```
jleveque, line 2194: Add space after comma

orchagent/portsorch.cpp
```diff
@@ -2177,22 +2179,92 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
                 }
 
+                attr_val.clear();
                 /* Set port serdes ipredriver */
                 if (fvField(i) == "ipredriver")
                 {
-                    getPortSerdesVal(fvValue(i), ipredriver);
+                    getPortSerdesVal(fvValue(i), attr_val);
+                    serdes_attr.insert(serdes_attr_pair(SAI_PORT_SERDES_ATTR_IPREDRIVER,attr_val));
```
jleveque, line 2202: Add space after comma

jleveque (changes requested):

orchagent/portsorch.cpp
```diff
@@ -4387,29 +4429,95 @@ bool PortsOrch::removeAclTableGroup(const Port &p)
...
+        SWSS_LOG_ERROR("Failed to get port attr serdes id %d to port pid:0x%" PRIx64,
+                       port_attr.id, port_id);
+        return false;
+    }
 
-    attr.value.u32list.count = (uint32_t)serdes_val.size();
-    attr.value.u32list.list = serdes_val.data();
+    if (port_attr.value.oid != SAI_NULL_OBJECT_ID)
+    {
+        status = sai_port_api->remove_port_serdes(port_attr.value.oid);
```
prsunny, line 4454: Can port_serdes be removed on the fly? Or do you require the port to be shutdown etc?

orchagent/portsorch.cpp
```diff
@@ -4387,29 +4429,95 @@ bool PortsOrch::removeAclTableGroup(const Port &p)
...
+                           port_attr.value.oid, port_id);
+            return false;
+        }
+    }
+
+
+    port_serdes_attr.id = SAI_PORT_SERDES_ATTR_PORT_ID;
+    port_serdes_attr.value.oid = port_id;
+    attr_list.emplace_back(port_serdes_attr);
+    SWSS_LOG_ERROR("Creating serdes for port 0x%" PRIx64, port_id);
```
prsunny, line 4467: This cannot be error -> please change to INFO as you already have one NOTICE below

orchagent/portsorch.cpp
```diff
@@ -4387,29 +4429,95 @@ bool PortsOrch::removeAclTableGroup(const Port &p)
...
 
-    sai_status_t status = sai_port_api->set_port_attribute(port_id, &attr);
     if (status != SAI_STATUS_SUCCESS)
     {
-        SWSS_LOG_ERROR("Failed to set serdes attribute %d to port pid:%" PRIx64,
-                       attr_id, port_id);
-        return false;
+        SWSS_LOG_ERROR("Failed to create port serdes for port 0x%" PRIx64,
+                       port_id);
+         return false;
```
prsunny, line 4484: Extra space

orchagent/portsorch.cpp
```diff
@@ -2177,22 +2179,92 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
                 if (fvField(i) == "ipredriver")
                 {
-                    getPortSerdesVal(fvValue(i), ipredriver);
+                    getPortSerdesVal(fvValue(i), attr_val);
+                    serdes_attr.insert(serdes_attr_pair(SAI_PORT_SERDES_ATTR_IPREDRIVER, attr_val));
+                }
+
+                /* Set port serdes pre1 */
+                attr_val.clear();
+                if (fvField(i) == "pre1")
```
daall, line 2207: Can we have these values like "pre1", "post1", "attn", etc. be macros?

dgsudharsan: Hi @daall, The existing fields are also compared as plain strings and hence to maintain uniformity I didn't define macros. Moreover this string is just used in one place. Please let me know if you want to change these to macros or the strings should be good

orchagent/portsorch.cpp
dgsudharsan: It can be removed on the fly. It just removes the software object. The logic is introduced because the port serdes object has all attributes as create only and since multiple attributes need to be set which the current set attribute doesn't support, we have this logic of deleting the existing object and recreating it.

orchagent/portsorch.cpp
daall: That seems reasonable to me, we can keep these as is.

daall:
It looks like the same set of tests (acl, CRM, mirror, etc.) are failing consistently between re-runs. @dgsudharsan can you take a look?

dgsudharsan:
> It looks like the same set of tests (acl, CRM, mirror, etc.) are failing consistently between re-runs. @dgsudharsan can you take a look?

Sure. Let me run the tests locally. Meanwhile let me also try to rebuild here again to see if it succeeds.

dgsudharsan:
retest vs please

dgsudharsan:
> > It looks like the same set of tests (acl, CRM, mirror, etc.) are failing consistently between re-runs. @dgsudharsan can you take a look?
> 
> Sure. Let me run the tests locally. Meanwhile let me also try to rebuild here again to see if it succeeds.

Currently one test failing. I don't think the failing tests are due to any of my changes as my changes come into play only when media_settings.json is defined which is not for a vs

dgsudharsan:
retest vs please

prsunny (approved):
lgtm.. wait for other reviewers.

dgsudharsan:
retest vs please

dgsudharsan:
retest vs please

dgsudharsan:
retest this please

prsunny (approved):

jleveque:
retest vs please

orchagent/portsorch.cpp
```diff
@@ -2133,65 +2136,111 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+                }
+                /* Set port serdes attn */
+                else if (fvField(i) == "attn")
+                {
+                    getPortSerdesVal(fvValue(i), attr_val);
+                    serdes_attr.insert(serdes_attr_pair(SAI_PORT_SERDES_ATTR_TX_FIR_ATTN, attr_val));
+                }
+                else
+                {
+                    SWSS_LOG_ERROR("Unknown port field: %s", fvField(i).c_str());
```
daall, line 2243: The CRM tests are failing b/c the syslog is getting throttled due to this message:
```
Feb  2 23:46:10.029687 9a649eaefbe1 ERR #orchagent: message repeated 4793 times: [ :- doPortTask: Unknown port field: alias]
```

Do we need to add/skip "alias" here?

dgsudharsan: I can remove this as there might be other fields that could be added in the future which may not be significant for the SAI conversion.

dgsudharsan:
retest this please

dgsudharsan:
retest this please

prsunny (approved):

jleveque (approved):

daall (approved):

aravindmani-1:
@daall 

Looks like this PR is not merged into Azure master image.
Can we merge this PR since we need it for DellEMC Z9332f platform?.

Thanks,
Aravind.

kcudnik:
this pr is explicitly removing port serdes if it exists on a given port, tis rises a question whether this port serdes should be treated like vlan is (that needs to be explicitly removed before removing port) or whether it can be treated like a queue (that will be auto removed when port is removed)

when implementing portserdes in virtual switch on sairedis lib, first approach will break backward compatibility with previous unittets that are actually removing port, they would need to be updated to remove port serdes as well or remove for both scenarios could be implemented

PR #1360, opened 2020-07-23
[PFCWD + Asym PFC]: Allow PFCWD to detect PFC storm on all priorities when Asym PFC is enabled

volodymyrsamotiy:
Signed-off-by: Volodymyr Samotiy <volodymyrs@mellanox.com>

**What I did**
Add possibility for PFCWD  to detect PFC storm on all priorities when asymmetric PFC is enabled.

According to requirements PFCWD should detect PFC storm on the lossy priorities for the port on which asymmetric PFC is enabled.
(when asymmetric PFC is enabled on some port that means such port is enabled for receiving PFC frames on all priorities)

**Why I did it**
To fix issue https://github.com/Azure/sonic-swss/issues/1128.

**How I verified it**
1. Enable asymmetric PFC on test port.
```
pfc config asymmetric on <port>
```
2. Enable PFCWD on the test port.
```
pfcwd start --action drop --restoration-time 3000 <port> 1000
pfcwd interval 500
```
3. Run ```pause``` frames to the test port for all priorities.
4. Verify that PFCWD detected storm for all priorities.
```
pfcwd show stats
```

**Details if related**
N/A

PR #1612, opened 2021-01-26
[pfc] Introduce pfc config bitmask to track user-configured pfc enabled TCs on physical port

wendani:
**What I did**
Introduce PFC config bitmask to track user-configured PFC enabled TCs on physical port

This allows separate tracking of user-configured PFC bitmask (`pfc_enable` field in `CONFIG_DB`) from PFC bitmask status in ASIC. As when pfc wd is enabled and PFC storm occurs on a lossless TC, PFC bit on that TC is cleared in ASIC during the storm by `PfcWdLossyHandler`, making PFC bitmask status in ASIC to be different from that configured by user.

Such a difference causes the following misleading log message in syslog when a lossless TC is in PFC storm and pfc wd drop action is in place:
```
Aug 20 10:28:43.590517 sonic WARNING swss#orchagent: :- pfcFrameCounterCheck: Got PFC 933416836 frame(s) on lossy queue 3 port Ethernet0
```

**Why I did it**
Fix https://github.com/Azure/sonic-buildimage/issues/5910

**How I verified it**
On brcm

**Details if related**

When user wants to change PFC enable bits on TCs, if pfc wd is enabled, user still needs to toggle pfc wd stop and start to ensure pfc wd state machine is stopped on a PFC-disabled TC and is running on a PFC-enabled TC. However, before this change, command sequence matters: pfcwd must be stopped before PFC bits change on TCs, followed by `pfcwd start`. With this change, restriction on command sequence is no longer needed, i.e., PFC bits change on TC can be issued either before or after `pfcwd stop`, followed by `pfcwd start`.

Removing manual intervention described above (pfc wd stop and start toggle in changing PFC bits on TCs) is out of the scope of this PR, and will be in follow-up ones.

Covers fix to allow disabling last lossless TC in pending PR https://github.com/Azure/sonic-swss/pull/1055

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -446,7 +446,7 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::enableBigRedSwitchMode()
         }
         // By removing action handler, we expect pfc bit mask status in asic to
         // be the same as that in config
-        assert(pfcMaskStatus == cfgMaskCfg);
+        assert(pfcMaskStatus == pfcMaskCfg);
 
         for (uint8_t i = 0; i < PFC_WD_TC_MAX; i++)
         {
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -325,7 +325,7 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::setBigRedSwitchMode(const string 
 
     if (value == "enable")
     {
-        // When BIG_RED_SWITCH mode is enabled, pfcwd is automatically disabled
+        // When BIG_RED_SWITCH mode is enabled, pfcwd state machine is automatically disabled
         enableBigRedSwitchMode();
     }
     else if (value == "disable")
@@ -382,7 +382,7 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::enableBigRedSwitchMode()
     for (auto &it: allPorts)
     {
         Port port = it.second;
-        uint8_t pfcMask = 0;
+        uint8_t pfcMaskStatus = 0;
 
         if (port.m_type != Port::PHY)
         {
@@ -390,7 +390,7 @@ void PfcWdSwOrch<DropHandler, ForwardHandler>::enableBigRedSwitchMode()
             continue;
         }
 
-        if (!gPortsOrch->getPortPfc(port.m_port_id, &pfcMask))
+        if (!gPortsOrch->getPortPfc(port.m_port_id, &pfcMaskStatus))
         {
             SWSS_LOG_ERROR("Failed to get PFC mask on port %s", port.m_alias.c_str());
             return;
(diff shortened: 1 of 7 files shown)


lguohan:
can you add a test case for this?

orchagent/pfcwdorch.cpp
```diff
@@ -258,7 +258,7 @@ task_process_status PfcWdOrch<DropHandler, ForwardHandler>::createEntry(const st
 
     if (!startWdOnPort(port, detectionTime, restorationTime, action))
     {
-        SWSS_LOG_ERROR("Failed to start PFC Watchdog on port %s", port.m_alias.c_str());
+        SWSS_LOG_INFO("Failed to start PFC Watchdog on port %s", port.m_alias.c_str());
```
lguohan, line 261: why change the log level here?

orchagent/portsorch.cpp
```diff
@@ -907,7 +907,7 @@ bool PortsOrch::setPortFec(Port &port, sai_port_fec_mode_t mode)
     return true;
 }
 
-bool PortsOrch::getPortPfc(sai_object_id_t portId, uint8_t *pfc_bitmask)
+bool PortsOrch::getPortPfc(sai_object_id_t portId, uint8_t *pfc_bitmask_status, uint8_t *pfc_bitmask_cfg)
```
lguohan, line 910: use reference is better than pointer here.

orchagent/portsorch.cpp
```diff
@@ -951,18 +952,74 @@ bool PortsOrch::setPortPfc(sai_object_id_t portId, uint8_t pfc_bitmask)
...
     {
-        SWSS_LOG_ERROR("Failed to set PFC 0x%x to port id 0x%" PRIx64 " (rc:%d)", attr.value.u8, portId, status);
+        SWSS_LOG_ERROR("Failed to set PFC status 0x%x to port id 0x%" PRIx64 ", rc: %d", attr.value.u8, p.m_port_id, status);
+        return false;
+    }
+
+    return true;
+}
+
+bool PortsOrch::setPortPfcStatus(sai_object_id_t portId, uint8_t pfc_bitmask_status)
```
lguohan, line 967: it is confusing to have two functions with the same name as above.

orchagent/portsorch.h
```diff
@@ -260,6 +261,7 @@ class PortsOrch : public Orch, public Subject
     bool setPortFec(Port &port, sai_port_fec_mode_t mode);
     bool setPortPfcAsym(Port &port, string pfc_asym);
     bool getDestPortId(sai_object_id_t src_port_id, dest_port_type_t port_type, sai_object_id_t &des_port_id);
+    bool setPortPfcStatus(const Port &p, uint8_t pfc_bitmask_status);
```
lguohan, line 264: can you mark this private function as setPortPfcStatus_

orchagent/port.h
```diff
@@ -119,7 +119,8 @@ class Port
     std::vector<sai_object_id_t> m_queue_ids;
     std::vector<sai_object_id_t> m_priority_group_ids;
     sai_port_priority_flow_control_mode_t m_pfc_asym = SAI_PORT_PRIORITY_FLOW_CONTROL_MODE_COMBINED;
-    uint8_t   m_pfc_bitmask = 0;
+    uint8_t   m_pfc_bitmask_cfg = 0;    // bitmask from config
+    uint8_t   m_pfc_bitmask_status = 0; // bitmask status in asic
```
lguohan, line 123: a better way to differentiate is to have two variables named as

m_pfc_bitmask_usercfg;
m_pfc_bitmask_wdcfg;


technicially, they are both config, one is user config, one is pfcwd config

lguohan:
others look good to me.

neethajohn:
@wendani , FYI, https://github.com/Azure/sonic-swss/pull/1360 changes the signature of setPortPfc and getPortPfc methods

wendani:
> can you add a test case for this?

Done

wendani:
> @wendani , FYI, #1360 changes the signature of setPortPfc and getPortPfc methods

What is your plan to move forwar?

neethajohn:
> > @wendani , FYI, #1360 changes the signature of setPortPfc and getPortPfc methods
> 
> What is your plan to move forward?

That PR needs some rework. so you should be good. will review these changes soon

PR #1614, opened 2021-01-28
[PortsOrch] Add reference counting to ports for ACL bindings

chaoskao:
What I did:
Use reference count for protect interface, If interface bind to ACL the reference will increase and vice versa.

Why I did it:
If interface bind to ACL and remove interface causes the ACL rule become global rule to match all interfaces.

How I verified it:
Run the pytest check ACL and Port test case passed.


daall (approved):

lguohan:
@daall , is this needed for 202012?

daall:
> @daall , is this needed for 202012?

We should take it, yes. Label added

PR #1617, opened 2021-01-28
Vlan remove comm bug 779

allas-nvidia:
**What I did**
Changed severity level of log notification from ERR to INFO, in the case vlan is deleted and there are still members on this VLAN.

**Why I did it**
When we remove VLAN and there are members on this VLAN, the events which coming to portsorch.cpp doTask(...) are not according to the order that they are executed in vlan.py - file which activated this command.

Lets see the next flow: we have 3 members on the VLAN and user type "sudo config vlan del <>"
In vlan.py the flow is next:

Remove entries from VLAN_MEMBER_CFG_TABLE with the key of the VLAN user want to remove.

Remove entry from VLAN_CFG_TABLE with the key of the VLAN user want to remove.
However, the order of requests getting to VlanMgr.cpp file is different.

Request DEL entry from VLAN_MEMBER_CFG_TABLE, Request DEL entry from VLAN_CFG_TABLE, Request DEL entry from VLAN_MEMBER_CFG_TABLE, Request DEL entry from VLAN_MEMBER_CFG_TABLE
(3 members - 3 calls of delete entry from VLAN_MEMBER_CFG_TABLE, but between these calls we have call fromVLAN_CFG_TABLE )

As VlanMgr.cpp update APP_TABLEs, STATE_TABLEs of the same tables the request is coming from (VLAN_CFG_TABLE update VLAN_APP_TABLE, VLAN_STATE_TABLE , VLAN_MEMBER_CFG_TABLE update VLAN_MEMBER_APP_TABLE ...) PortsOrch.cpp receives the notification events in the same order.
It means that request to delete VLAN will come before all vlan members are deleted.

However, finally, when all members will be deleted - the VLAN will be deleted too, as doVlanTask function will try it again and again ... and finally it should be remove VLAN

This BUG was opened by community:
Azure/sonic-utilities#779
**How I verified it**
sudo config vlan add 200
sudo config vlan member add 200 Ethernet0
sudo config vlan member add 200 Ethernet8
sudo config vlan member add 200 Ethernet16
sudo config vlan del 200

The log had info messages :

removeVlan: VLAN Vlan200 still has members assigned: < Ethernet8 >...

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1825,7 +1825,7 @@ sai_status_t PortsOrch::removePort(sai_object_id_t port_id)
 
     Port port;
 
-    /*
+    /* 
      * Make sure to bring down admin state.
      * SET would have replaced with DEL
      */
@@ -3600,12 +3600,7 @@ bool PortsOrch::removeVlan(Port vlan)
     /* Vlan removing is not allowed when the VLAN still has members */
     if (vlan.m_members.size() > 0)
     {
-        string vlan_members;
-        for (auto &name: vlan.m_members)
-        {
-            vlan_members += " " + name;
-        }
-        SWSS_LOG_INFO("VLAN %s still has members assigned: [%s ].", vlan.m_alias.c_str(), vlan_members.c_str());
+        SWSS_LOG_ERROR("Failed to remove non-empty VLAN %s", vlan.m_alias.c_str());
         return false;
     }
 
@@ -4944,7 +4939,7 @@ bool PortsOrch::setVoqInbandIntf(string &alias, string &type)
     // host if for the inband here
 
     // May do the processing for other inband type like type=vlan here
-
+    
     //Store the name of the local inband port
     m_inbandPortName = alias;
 
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1825,7 +1825,7 @@ sai_status_t PortsOrch::removePort(sai_object_id_t port_id)
 
     Port port;
 
-    /* 
+    /*
      * Make sure to bring down admin state.
      * SET would have replaced with DEL
      */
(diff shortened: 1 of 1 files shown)


prsunny:
Not fully convinced on this change. IMO, the orchagent behavior is expected and fix should in sonic-utilities where it must check for membership before accepting the vlan delete command

prsunny (changes requested):
As comments

allas-nvidia:
Now, sonic utilities (vlan.py) before removing VLAN remove all members of this VLAN. The first question - do we want to leave this behaviour? If yes - VlanMgr will process it in the next order : VLAN_MEMBER_CFG (doVlanMemberTask), VLAB_CFG(doVlanTask), VLAN_MEMBER_CFG(doVlanMemberTask), VLAN_MEMBER_CFG (doVlanMemberTask)(in the case 3 members). This is the order PortOrch will receive notification.
Do you suggest to move to PortsOrch deletion of all members in the case the notification of remove Vlan was received in PortsOrch? It means the logic in PortsOrch should be changed completely, as in VlanMgr we update APP_VLAN_MEMBER_TABLE_NAME and PortsOrch will get notification and come to the function doVlanMemberTask where we try to find the particular Port, notification was received on via this table. In this way, we will have ASSERT  in PortsOrch ....

allas-nvidia:
Maybe the idea not to allow to delete VLAN if we have member ....?

prsunny:
> Maybe the idea not to allow to delete VLAN if we have member ....?

Yes.. and it can be checked in CLI itself

allas-nvidia:
I am doing it - but it seems will break backward compatibility, as it was allowed to remove VLAN with members ....
Is it OK?

prsunny:
> I am doing it - but it seems will break backward compatibility, as it was allowed to remove VLAN with members ....
> Is it OK?

That should be fine, we can have the change in master

liat-grozovik:
@allas-nvidia  i think this PR should be closed as the other one will comes from the sonic-utilities.

PR #1620, opened 2021-02-01
Updated PFCWD to use single ACL table for PFCWD and MUX

vmittal-msft:
**What I did**
Allow PFCWD and MUX features to use single ACL table on ingress side.
**Why I did it**
It is done to save space and to allow new features to make use of ACL table.
**How I verified it**
Verified by running tests on TD3 platform.
**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -109,7 +109,8 @@ static acl_table_type_lookup_t aclTableTypeLookUp =
     { TABLE_TYPE_CTRLPLANE,             ACL_TABLE_CTRLPLANE },
     { TABLE_TYPE_DTEL_FLOW_WATCHLIST,   ACL_TABLE_DTEL_FLOW_WATCHLIST },
     { TABLE_TYPE_DTEL_DROP_WATCHLIST,   ACL_TABLE_DTEL_DROP_WATCHLIST },
-    { TABLE_TYPE_MCLAG,                 ACL_TABLE_MCLAG }
+    { TABLE_TYPE_MCLAG,                 ACL_TABLE_MCLAG },
+    { TABLE_TYPE_DROP,                  ACL_TABLE_DROP }
 };
 
 static acl_stage_type_lookup_t aclStageLookUp =
@@ -620,6 +621,26 @@ bool AclRule::remove()
     return res;
 }
 
+void AclRule::update_in_ports(void *data)
+{
+    SWSS_LOG_ENTER();
+    sai_attribute_t attr;
+    sai_status_t status;
+
+    if (data != NULL)
+    {
+        attr.id = *(sai_acl_entry_attr_t *)data;
+        attr.value = m_matches[*(sai_acl_entry_attr_t *)data];
+        attr.value.aclfield.enable = true;
+        
+        status = sai_acl_api->set_acl_entry_attribute(m_ruleOid, &attr);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("Failed to update ACL rule %s, rv:%d", m_id.c_str(), status);
+        }
+    }
+}
+
 AclRuleCounters AclRule::getCounters()
 {
     SWSS_LOG_ENTER();
@@ -675,7 +696,8 @@ shared_ptr<AclRule> AclRule::makeShared(acl_table_type_t type, AclOrch *acl, Mir
         type != ACL_TABLE_MIRROR_DSCP &&
         type != ACL_TABLE_DTEL_FLOW_WATCHLIST &&
         type != ACL_TABLE_DTEL_DROP_WATCHLIST &&
-        type != ACL_TABLE_MCLAG)
+        type != ACL_TABLE_MCLAG &&
+        type != ACL_TABLE_DROP)
     {
         throw runtime_error("Unknown table type");
     }
@@ -722,6 +744,10 @@ shared_ptr<AclRule> AclRule::makeShared(acl_table_type_t type, AclOrch *acl, Mir
     else if (type == ACL_TABLE_MCLAG)
     {
         return make_shared<AclRuleMclag>(acl, rule, table, type);
+    }
+    else if (type == ACL_TABLE_DROP)
+    {
+        return make_shared<AclRulePfcwd>(acl, rule, table, type);
     }
 
     throw runtime_error("Wrong combination of table type and action in rule " + rule);
@@ -993,24 +1019,9 @@ bool AclRuleL3::validate()
     return true;
 }
 
-void AclRuleL3::update(SubjectType type, void *data)
+void AclRuleL3::update(SubjectType, void *)
 {
-    SWSS_LOG_ENTER();
-    sai_attribute_t attr;
-    sai_status_t status;
-
-    if (data != NULL)
-    {
-        attr.id = *(sai_acl_entry_attr_t *)data;
-        attr.value = m_matches[*(sai_acl_entry_attr_t *)data];
-        attr.value.aclfield.enable = true;
-        
-        status = sai_acl_api->set_acl_entry_attribute(m_ruleOid, &attr);
-        if (status != SAI_STATUS_SUCCESS)
-        {
-            SWSS_LOG_ERROR("Failed to update ACL rule %s, rv:%d", m_id.c_str(), status);
-        }
-    }
+    // Do nothing
 }
 
 AclRulePfcwd::AclRulePfcwd(AclOrch *aclOrch, string rule, string table, acl_table_type_t type, bool createCounter) :
@@ -1020,12 +1031,6 @@ AclRulePfcwd::AclRulePfcwd(AclOrch *aclOrch, string rule, string table, acl_tabl
 
 bool AclRulePfcwd::validateAddMatch(string attr_name, string attr_value)
 {
-    if ((attr_name != MATCH_TC) && (attr_name != MATCH_IN_PORTS))
-    {
-        SWSS_LOG_ERROR("%s is not supported for the tables of type PFCWD", attr_name.c_str());
-        return false;
-    }
-
     return AclRule::validateAddMatch(attr_name, attr_value);
 }
 
@@ -1340,7 +1345,7 @@ bool AclTable::create()
     vector<int32_t> bpoint_list;
 
     // PFC watch dog ACLs are only applied to port
-    if (type == ACL_TABLE_PFCWD)
+    if ((type == ACL_TABLE_PFCWD) || (type == ACL_TABLE_DROP))
     {
         bpoint_list = { SAI_ACL_BIND_POINT_TYPE_PORT };
     }
@@ -1538,6 +1543,10 @@ bool AclTable::create()
         table_attrs.push_back(attr);
 
         attr.id = SAI_ACL_TABLE_ATTR_FIELD_IP_PROTOCOL;
+        attr.value.booldata = true;
+        table_attrs.push_back(attr);
+
+        attr.id = SAI_ACL_TABLE_ATTR_FIELD_IN_PORTS;
         attr.value.booldata = true;
         table_attrs.push_back(attr);
     }
@@ -2986,7 +2995,7 @@ bool AclOrch::updateAclRule(shared_ptr<AclRule> rule, string table_id, string at
             attr_value.pop_back();
 
             rule->validateAddMatch(MATCH_IN_PORTS, attr_value);  
-            m_AclTables[table_oid].rules[rule->getId()]->update(SUBJECT_TYPE_PORT_CHANGE, &aclMatchLookup[attr_name]);
+            m_AclTables[table_oid].rules[rule->getId()]->update_in_ports(&aclMatchLookup[attr_name]);
         }
         break;
 
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -999,7 +999,7 @@ void AclRuleL3::update(SubjectType type, void *data)
     sai_attribute_t attr;
     sai_status_t status;
 
-    if(data != NULL)
+    if (data != NULL)
     {
         attr.id = *(sai_acl_entry_attr_t *)data;
         attr.value = m_matches[*(sai_acl_entry_attr_t *)data];
@@ -1008,8 +1008,7 @@ void AclRuleL3::update(SubjectType type, void *data)
         status = sai_acl_api->set_acl_entry_attribute(m_ruleOid, &attr);
         if (status != SAI_STATUS_SUCCESS)
         {
-            SWSS_LOG_ERROR("Failed to update ACL rule %s, rv:%d",
-                    m_id.c_str(), status);
+            SWSS_LOG_ERROR("Failed to update ACL rule %s, rv:%d", m_id.c_str(), status);
         }
     }
 }
@@ -1021,6 +1020,12 @@ AclRulePfcwd::AclRulePfcwd(AclOrch *aclOrch, string rule, string table, acl_tabl
 
 bool AclRulePfcwd::validateAddMatch(string attr_name, string attr_value)
 {
+    if ((attr_name != MATCH_TC) && (attr_name != MATCH_IN_PORTS))
+    {
+        SWSS_LOG_ERROR("%s is not supported for the tables of type PFCWD", attr_name.c_str());
+        return false;
+    }
+
     return AclRule::validateAddMatch(attr_name, attr_value);
 }
 
@@ -1349,7 +1354,7 @@ bool AclTable::create()
     attr.value.s32list.list = bpoint_list.data();
     table_attrs.push_back(attr);
 
-    if (type == ACL_TABLE_PFCWD)
+    if ((type == ACL_TABLE_PFCWD) || (type == ACL_TABLE_DROP))
     {
         attr.id = SAI_ACL_TABLE_ATTR_FIELD_TC;
         attr.value.booldata = true;
@@ -1359,7 +1364,7 @@ bool AclTable::create()
         attr.value.s32 = (stage == ACL_STAGE_INGRESS) ? SAI_ACL_STAGE_INGRESS : SAI_ACL_STAGE_EGRESS;
         table_attrs.push_back(attr);
         
-        if(stage == ACL_STAGE_INGRESS) {
+        if (stage == ACL_STAGE_INGRESS) {
             attr.id = SAI_ACL_TABLE_ATTR_FIELD_IN_PORTS;
             attr.value.booldata = true;
             table_attrs.push_back(attr);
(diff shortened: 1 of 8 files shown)


orchagent/pfcactionhandler.cpp
```diff
@@ -239,12 +240,39 @@ PfcWdAclHandler::PfcWdAclHandler(sai_object_id_t port, sai_object_id_t queue,
...
+            SWSS_LOG_NOTICE("Found the rule hence updating an existing one...%s:%s" PRIx64, m_strIngressTable.c_str(), m_strRule.c_str());
+            vector<sai_object_id_t> in_ports;
+            in_ports = ruleIter->second->getInPorts();
+            in_ports.push_back(port);
+
+            for (const auto& portIter: in_ports)
+            {
+                Port port;
+                gPortsOrch->getPort(portIter, port);
+                SWSS_LOG_NOTICE("Port number added to the list ................. %s" PRIx64, port.m_alias.c_str());
```
lguohan, line 271: syntax error.

orchagent/aclorch.cpp
```diff
@@ -993,11 +993,26 @@ bool AclRuleL3::validate()
...
+    {
+        attr.id = *(sai_acl_entry_attr_t *)data;
+        attr.value = m_matches[*(sai_acl_entry_attr_t *)data];
+        attr.value.aclfield.enable = true;
+        
+        status = sai_acl_api->set_acl_entry_attribute(m_ruleOid, &attr);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("Failed to update ACL rule %s, rv:%d",
+                    m_id.c_str(), status);
```
prsunny, line 1012: This could be in the same line as above

orchagent/pfcactionhandler.cpp
```diff
@@ -316,13 +354,42 @@ void PfcWdAclHandler::createPfcAclRule(shared_ptr<AclRulePfcwd> rule, uint8_t qu
...
 
+void PfcWdAclHandler::updatePfcAclRule(shared_ptr<AclRule> rule, uint8_t queueId, string strTable, vector<sai_object_id_t> in_ports)
+{
+    SWSS_LOG_ENTER();
+    string attr_name, attr_value;
+
+    for (const auto& portIter: in_ports)
+    {
+        Port p;
+        gPortsOrch->getPort(portIter, p);
```
prsunny, line 382: Suggest check for return value of this function call

orchagent/pfcactionhandler.cpp
```diff
@@ -226,25 +226,42 @@ PfcWdAclHandler::PfcWdAclHandler(sai_object_id_t port, sai_object_id_t queue,
...
+    sai_object_id_t  table_oid;
 
     // There is one handler instance per queue ID
     string queuestr = to_string(queueId);
-    m_strIngressTable = "IngressTable_PfcWdAclHandler_" + queuestr;
+    m_strIngressTable = "IngressTable_PfcWdAclHandler";
     m_strEgressTable = "EgressTable_PfcWdAclHandler_" + queuestr;
     m_strRule = "Rule_PfcWdAclHandler_" + queuestr;
-
+
```
prsunny, line 236: tabspace?

prsunny:
@vmittal-msft , as we discussed, can we have the in-port bitmap handling APIs generic, so that it can be used by other modules? Also you probably would want to revisit the table creation as it could be used before the PFC WD may start adding rules.

tests/test_pfcwd.py
```diff
@@ -0,0 +1,221 @@
...
+def getQueueOid(dvs, queue):
+   cnt_r = redis.Redis(unix_socket_path=dvs.redis_sock, db=swsscommon.COUNTERS_DB,
+                       encoding="utf-8", decode_responses=True)
+   return  cnt_r.hget("COUNTERS_QUEUE_NAME_MAP", queue)
+
+def setPfcStormQueueOid(dvs, queue_oid):
+   cnt_r = redis.Redis(unix_socket_path=dvs.redis_sock, db=swsscommon.COUNTERS_DB,
+                       encoding="utf-8", decode_responses=True)
+   cnt_r.hset("COUNTERS:"+queue_oid, "DEBUG_STORM","enabled")
+   import pdb;pdb.set_trace()
```
daall, line 63: `counters_db`

daall, line 63: extra debug line

tests/test_pfcwd.py
```diff
@@ -0,0 +1,221 @@
...
+    atbl = swsscommon.Table(self.adb, "ASIC_STATE:SAI_OBJECT_TYPE_PORT")
+    port_groups = []
+    for p in [dvs.asicdb.portnamemap[portname] for portname in bind_ports]:
+        (status, fvs) = atbl.get(p)
+        for fv in fvs:
+            if fv[0] == "SAI_PORT_ATTR_INGRESS_ACL":
+                assert fv[1] in acl_table_groups
+                port_groups.append(fv[1])
+
+    assert len(port_groups) == len(bind_ports)
```
daall, line 132: Take a look at `dvs_acl.py`, I think these methods are already implemented for you.

tests/test_pfcwd.py
```diff
@@ -0,0 +1,221 @@
...
+    assert set(port_groups) == set(acl_table_groups)
+
+def Verify_AclTableCreation(self, dvs):
```
daall, line 135: I would take a look at https://github.com/Azure/sonic-swss/blob/0b0d24cfec42c0921687eb14cdb16f37adbfe34f/tests/test_acl.py#L54

tests/test_pfcwd.py
```diff
@@ -0,0 +1,221 @@
...
+    acl_group_ids = atbl.getKeys()
+    assert len(acl_group_ids) == 2
+
+    # check acl table group member
+    verify_acl_group_member(acl_group_ids, acl_table_id)
+
+    # check port binding
+    verify_acl_port_binding(dvs, "Ethernet0")
+
+def Verify_AclRuleInPorts(self, dvs):
```
daall, line 154: I would take a look at https://github.com/Azure/sonic-swss/blob/0b0d24cfec42c0921687eb14cdb16f37adbfe34f/tests/test_acl.py#L101

tests/test_pfcwd.py
```diff
@@ -0,0 +1,221 @@
...
+    @pytest.fixture(scope="module", autouse=True)
+    def enable_queue_counters(self, dvs):
+        dvs.runcmd("counterpoll queue enable")
+
+    def test_ingress_pfcwd_handler(self, dvs):
+        self.setup_db(dvs)
+        config_db = dvs.get_config_db()
+        
+        #Enable PFC at global level
+        enable_device_pfcwd(config_db)
```
daall, line 189: This can probably be made a fixture since it will be used throughout all of the tests.

daall (changes requested):
as comments

orchagent/pfcactionhandler.cpp
```diff
@@ -316,6 +354,22 @@ void PfcWdAclHandler::createPfcAclRule(shared_ptr<AclRulePfcwd> rule, uint8_t qu
     attr_value = to_string(queueId);
     rule->validateAddMatch(attr_name, attr_value);
 
+    // Add MATCH_IN_PORTS as match creteria for ingress table
+    if(strTable == INGRESS_TABLE_DROP)
```
prsunny, line 358: space after `if`

orchagent/pfcactionhandler.cpp
```diff
@@ -225,35 +225,51 @@ PfcWdAclHandler::PfcWdAclHandler(sai_object_id_t port, sai_object_id_t queue,
...
         // First time of handling PFC for this queue, create ACL table, and bind
         createPfcAclTable(port, m_strIngressTable, true);
         shared_ptr<AclRulePfcwd> newRule = make_shared<AclRulePfcwd>(gAclOrch, m_strRule, m_strIngressTable, table_type);
-        createPfcAclRule(newRule, queueId, m_strIngressTable);
+        createPfcAclRule(newRule, queueId, m_strIngressTable, port);
     }
     else
     {
-        // Otherwise just bind ACL table with the port
-        found->second.bind(port);
```
daall, line 247: I think we still need to bind the ports to the ACL table before we update the rule so that the ACL group + group members are created for ports that don't exist in any other ACL tables.

orchagent/pfcactionhandler.cpp
```diff
@@ -225,35 +225,51 @@ PfcWdAclHandler::PfcWdAclHandler(sai_object_id_t port, sai_object_id_t queue,
...
+        table_oid = gAclOrch->getTableById(m_strIngressTable);
+        shared_ptr<AclRulePfcwd> newRule = make_shared<AclRulePfcwd>(gAclOrch, m_strRule, m_strIngressTable, table_type);
```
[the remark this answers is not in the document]
vmittal-msft: Make sense. Done.

orchagent/aclorch.cpp
```diff
@@ -2919,6 +2939,60 @@ bool AclOrch::removeAclRule(string table_id, string rule_id)
...
+            {
+                Port p;
+                gPortsOrch->getPort(port_iter, p);
+                attr_value += p.m_alias;
+                attr_value += ',';
+            }
+            attr_value.pop_back();
+
+            rule->validateAddMatch(MATCH_IN_PORTS, attr_value);  
+            m_AclTables[table_oid].rules[rule->getId()]->update(SUBJECT_TYPE_PORT_CHANGE, &aclMatchLookup[attr_name]);
```
neethajohn, line 2985: I do not understand why we are using SUBJECT_TYPE_PORT_CHANGE here. What is the change event here for the port? all we need to do here is call a function to set acl rule attributes.

orchagent/pfcactionhandler.cpp
```diff
@@ -266,10 +282,23 @@ PfcWdAclHandler::~PfcWdAclHandler(void)
...
+    {
+        gAclOrch->removeAclRule(m_strIngressTable, m_strRule);
+    }
+    else 
+    {
+        gAclOrch->updateAclRule(rule_iter->second, m_strIngressTable, MATCH_IN_PORTS, &port, RULE_OPER_DELETE);
+    } 
+
+    auto found = m_aclTables.find(m_strEgressTable);
     found->second.unbind(getPort());
```
neethajohn, line 302: if Danny's comment on L247 holds good, then I think we need some extra handling for ingress table. what if the same port is bound to another rule for a diff queue?

neethajohn (commented):
Haven't reviewed the vs test. will look at it when it is complete

orchagent/aclorch.cpp
vmittal-msft: ACL rule update is new function added as part of this PR. Subject type is a param used during rule update, I kept it to indicate different kind of rule updates we may be doing. Idea of passing SUBJECT_TYPE_PORT_CHANGE to indicate port mask is changing for the rule. Please let me know if i am incorrectly using subject type for this purpose.

orchagent/pfcactionhandler.cpp
vmittal-msft: Sure. Will test out the L247 changes and make necessary changes here. Will probably unbind the port once there are no rules left in ingress table else we may nee do do walk of the table every time there is rule update.

orchagent/pfcactionhandler.cpp
```diff
@@ -266,11 +283,30 @@ PfcWdAclHandler::~PfcWdAclHandler(void)
...
+    vector<sai_object_id_t> port_set = rule_iter->second->getInPorts();
+    sai_object_id_t port = getPort();
+
+    if ((port_set.size() == 1) && (port_set[0] == port))
+    {
+        gAclOrch->removeAclRule(m_strIngressTable, m_strRule);
+        if (table_size == 1) 
+        {
+            auto found = m_aclTables.find(m_strIngressTable);
+            found->second.unbind(port);
```
neethajohn, line 300: I am not sure if this is the right approach to unbind. we will need to unbind a port if its not used by any  rules in the table. @daall , @prsunny , can you confirm if we really need to bind the port to the table since now we are including it in the rule.

tests/test_pfcwd.py
```diff
@@ -0,0 +1,297 @@
...
+            # Enable PFC Storm on queue 3
+            self.set_pfc_storm_queue_oid(dvs, queue_oid)
+            
+            # Verify ASIC_DB for table creation
+            #verify_acl_table_creation(self, dvs)
+
+            #Verify ASIC_DB for rule creation
+            #verify_acl_rule_in_ports(dvs)
+
+       finally:
```
neethajohn, line 289: please make sure to clean up all the objects that you created for this particular test

orchagent/pfcactionhandler.cpp
prsunny: Agree with @neethajohn, that this may cause issues to existing rules in that table.

orchagent/aclorch.cpp
```diff
@@ -2919,6 +2949,63 @@ bool AclOrch::removeAclRule(string table_id, string rule_id)
...
+                attr_value += ',';
+            }
+            attr_value.pop_back();
+
+            rule->validateAddMatch(MATCH_IN_PORTS, attr_value);  
+            m_AclTables[table_oid].rules[rule->getId()]->update_in_ports(&aclMatchLookup[attr_name]);
+        }
+        break;
+
+        default:
```
daall, line 3002: It might be helpful to include a DEBUG log here in case someone tries to update a field that is not IN_PORTS

orchagent/aclorch.cpp
```diff
@@ -1519,6 +1545,10 @@ bool AclTable::create()
         attr.id = SAI_ACL_TABLE_ATTR_FIELD_IP_PROTOCOL;
         attr.value.booldata = true;
         table_attrs.push_back(attr);
+
+        attr.id = SAI_ACL_TABLE_ATTR_FIELD_IN_PORTS;
+        attr.value.booldata = true;
+        table_attrs.push_back(attr);
```
daall, line 1551: How is this related to your PFCWD changes?

orchagent/pfcactionhandler.cpp
```diff
@@ -295,14 +323,23 @@ void PfcWdAclHandler::createPfcAclTable(sai_object_id_t port, string strTable, b
...
     aclTable.link(port);
     aclTable.id = strTable;
-    aclTable.stage = ingress ? ACL_STAGE_INGRESS : ACL_STAGE_EGRESS;
+
+    if (ingress) 
+    {
+        aclTable.type = ACL_TABLE_DROP;
+        aclTable.stage = ACL_STAGE_INGRESS;
+    } else 
+    {
```
daall, line 334:
```suggestion
    } 
    else 
    {
```

tests/test_pfcwd.py
```diff
@@ -0,0 +1,98 @@
...
+
+            expected_sai_qualifiers = {
+                "SAI_ACL_ENTRY_ATTR_FIELD_TC" : dvs_acl.get_simple_qualifier_comparator("4&mask:0xff"),
+                "SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS": dvs_acl.get_port_list_comparator(["Ethernet8","Ethernet24"]),
+            }
+
+            dvs_acl.update_acl_rule(PFCWD_TABLE_NAME, PFCWD_RULE_NAME_2, config_qualifiers, action="DROP")
+            time.sleep(5)
+            rule_oid = self.find_acl_rule_oid(dvs)
+            dvs_acl.verify_acl_rule(expected_sai_qualifiers, action="DROP",acl_rule_id=rule_oid)
```
daall, line 89: IIRC if you leave off the OID it will automatically run a similar method to what you wrote in `find_acl_rule_oid`, so I think you can remove that method and just run this:

```suggestion
            dvs_acl.verify_acl_rule(expected_sai_qualifiers, action="DROP")
```

tests/test_pfcwd.py
```diff
@@ -0,0 +1,98 @@
...
+
+            expected_sai_qualifiers = {
+                "SAI_ACL_ENTRY_ATTR_FIELD_TC" : dvs_acl.get_simple_qualifier_comparator("3&mask:0xff"),
+                "SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS": dvs_acl.get_port_list_comparator(["Ethernet0","Ethernet16"])
+            }
+
+            dvs_acl.update_acl_rule(PFCWD_TABLE_NAME, PFCWD_RULE_NAME_1, config_qualifiers, action="DROP")
+            time.sleep(5)
+            rule_oid = self.find_acl_rule_oid(dvs)
+            dvs_acl.verify_acl_rule(expected_sai_qualifiers, action="DROP",acl_rule_id=rule_oid)
```
daall, line 58: same suggestion as below

tests/test_pfcwd.py
```diff
@@ -0,0 +1,98 @@
...
+                "IN_PORTS": "Ethernet0,Ethernet16"
+            }
+
+            expected_sai_qualifiers = {
+                "SAI_ACL_ENTRY_ATTR_FIELD_TC" : dvs_acl.get_simple_qualifier_comparator("3&mask:0xff"),
+                "SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS": dvs_acl.get_port_list_comparator(["Ethernet0","Ethernet16"])
+            }
+
+            dvs_acl.update_acl_rule(PFCWD_TABLE_NAME, PFCWD_RULE_NAME_1, config_qualifiers, action="DROP")
+            time.sleep(5)
```
daall, line 56: This sleep shouldn't be necessary as `verify_acl_rule` will continue to poll for the fields you specified.

tests/test_pfcwd.py
```diff
@@ -0,0 +1,98 @@
...
+
+            expected_sai_qualifiers = {
+                "SAI_ACL_ENTRY_ATTR_FIELD_TC" : dvs_acl.get_simple_qualifier_comparator("3&mask:0xff"),
+                "SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS": dvs_acl.get_port_list_comparator(["Ethernet0"])
+            }
+        
+            dvs_acl.create_acl_rule(PFCWD_TABLE_NAME, PFCWD_RULE_NAME_1, config_qualifiers, action="DROP")
+            time.sleep(5)
+            rule_oid = self.find_acl_rule_oid(dvs)
+            dvs_acl.verify_acl_rule(expected_sai_qualifiers, action="DROP", acl_rule_id=rule_oid)
```
daall, line 43: same feedback as below w.r.t. sleep + verify_acl_rule

tests/test_pfcwd.py
```diff
@@ -0,0 +1,98 @@
...
+       keys = atbl.getKeys()
+       acl_entry = [k for k in keys if k not in dvs.asicdb.default_acl_entries]
+
+       return acl_entry[0]
+
+    def test_PfcWdAclCreationDeletion(self, dvs, dvs_acl, testlog):
+        try:
+            dvs_acl.create_acl_table(PFCWD_TABLE_NAME, PFCWD_TABLE_TYPE, ["Ethernet0","Ethernet8", "Ethernet16", "Ethernet24"], stage="ingress")
+            dvs.runcmd("counterpoll queue enable")
+            time.sleep(10)
```
daall, line 28: This is OK for this PR, but it would be better to poll for some event (like the QUEUE_NAME_MAP appearing in the DB) that indicates this command was successful rather than adding a static 10s sleep.

daall: I also don't know if this step is still necessary for this PR or not given that we're going through config DB directly instead of triggering the PFCWD.

tests/test_pfcwd.py
vmittal-msft: Without sleep, it seems to be failing.

tests/test_pfcwd.py
vmittal-msft: Fixed verify_acl_rule but without sleep it seems to be failing.

tests/test_pfcwd.py
daall: OK I don't want to block this PR any further so let's revisit it later. :)

daall (approved):

orchagent/aclorch.h
```diff
@@ -196,6 +201,7 @@ class AclRule
     virtual bool create();
     virtual bool remove();
     virtual void update(SubjectType, void *) = 0;
+    virtual void update_in_ports(void *);
```
prsunny, line 204: please use same naming convention for the function as the rest of class

orchagent/aclorch.cpp
```diff
@@ -620,6 +621,26 @@ bool AclRule::remove()
     return res;
 }
 
+void AclRule::update_in_ports(void *data)
```
prsunny, line 624: Here its not required to be a void * param as the API is specific for inports

orchagent/aclorch.cpp
```diff
@@ -2919,6 +2945,64 @@ bool AclOrch::removeAclRule(string table_id, string rule_id)
...
+            attr_value.pop_back();
+
+            rule->validateAddMatch(MATCH_IN_PORTS, attr_value);  
+            m_AclTables[table_oid].rules[rule->getId()]->update_in_ports(&aclMatchLookup[attr_name]);
+        }
+        break;
+
+        default:
+            SWSS_LOG_ERROR("Acl rule update not supported for attr name %s", attr_name.c_str());
+            break;
```
prsunny, line 3000: please align break statement similar to other case

vmittal-msft:
retest vs please

prsunny (approved):
lgtm.

PR #1621, opened 2021-02-02
Fix the compiling errors in gcc9

stephenxs:
**What I did**
Fix the following error.
```
    orch.cpp: In member function 'ref_resolve_status Orch::resolveFieldRefArray(type_map&, const string&, swss::KeyOpFieldsValuesTuple&, std::vector<long unsigned int>&, std::string&)':
    orch.cpp:609:41: error: 'strlen' argument missing terminating nul [-Werror=stringop-overflow=]
      609 |                     object_name_list += string(&list_item_delimiter);
          |                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
    In file included from orch.cpp:6:
    orch.h:25:12: note: referenced argument declared here
       25 | const char list_item_delimiter = ',';
          |            ^~~~~~~~~~~~~~~~~~~
```

**Why I did it**

**How I verified it**

**Details if related**

**- Which release branch to backport (provide reason below if selected)**

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -606,7 +606,7 @@ ref_resolve_status Orch::resolveFieldRefArray(
                 SWSS_LOG_DEBUG("Resolved to sai_object:0x%" PRIx64 ", type:%s, name:%s", sai_obj, ref_type_name.c_str(), object_name.c_str());
                 sai_object_arr.push_back(sai_obj);
                 if (!object_name_list.empty())
-                    object_name_list += string(&list_item_delimiter, 1);
+                    object_name_list += list_item_delimiter;
                 object_name_list += ref_type_name + delimiter + object_name;
             }
             count++;
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -606,7 +606,7 @@ ref_resolve_status Orch::resolveFieldRefArray(
                 SWSS_LOG_DEBUG("Resolved to sai_object:0x%" PRIx64 ", type:%s, name:%s", sai_obj, ref_type_name.c_str(), object_name.c_str());
                 sai_object_arr.push_back(sai_obj);
                 if (!object_name_list.empty())
-                    object_name_list += string(&list_item_delimiter);
+                    object_name_list += string(&list_item_delimiter, 1);
                 object_name_list += ref_type_name + delimiter + object_name;
             }
             count++;
--- a/configure.ac
+++ b/configure.ac
@@ -83,6 +83,7 @@ CFLAGS_COMMON+=" -Wvariadic-macros"
 CFLAGS_COMMON+=" -Wno-switch-default"
 CFLAGS_COMMON+=" -Wno-long-long"
 CFLAGS_COMMON+=" -Wno-redundant-decls"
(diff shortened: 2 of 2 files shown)


qiluo-msft:
It seems compiler report a true code bug.
```c++
const char list_item_delimiter = ',';
object_name_list += string(&list_item_delimiter)
```

Could you check why you feed a non zero-terminated char* to string ctor?

qiluo-msft (changes requested):
As comment

stephenxs:
> It seems compiler report a true code bug.
> 
> ```c++
> const char list_item_delimiter = ',';
> object_name_list += string(&list_item_delimiter)
> ```
> 
> Could you check why you feed a non zero-terminated char* to string ctor?

```c++
const char list_item_delimiter = ',';
object_name_list += string(&list_item_delimiter, 1);
```
Need to add a `count `
Meanwhile, I don’t quite understand why it was not catched by gcc8

qiluo-msft:
This bug should be detected by either compiler or linter. Could you help explore a general solution to detect other unfound or future occurrences, during build process?

kcudnik:
change PR name to fix gcc9 warning

stephenxs:
> change PR name to fix gcc9 warning

Done.

kcudnik:
> This bug should be detected by either compiler or linter. Could you help explore a general solution to detect other unfound or future occurrences, during build process?

i actually checked g++ without and with O2 optimization, and if you use & on char constant, it puts data as a single character string which is padded by zeros, and with O2 even aligned with 0x10, but it could be just good practice by gcc, it may not be guaranteed, so warning is helpful in that case

orchagent/orch.cpp
```diff
@@ -606,7 +606,7 @@ ref_resolve_status Orch::resolveFieldRefArray(
                 SWSS_LOG_DEBUG("Resolved to sai_object:0x%" PRIx64 ", type:%s, name:%s", sai_obj, ref_type_name.c_str(), object_name.c_str());
                 sai_object_arr.push_back(sai_obj);
                 if (!object_name_list.empty())
-                    object_name_list += string(&list_item_delimiter);
+                    object_name_list += string(&list_item_delimiter, 1);
```
qiluo-msft, line 609:
```suggestion
                    object_name_list += list_item_delimiter;
```

stephenxs: It works. Done.

stephenxs:
> > This bug should be detected by either compiler or linter. Could you help explore a general solution to detect other unfound or future occurrences, during build process?
> 
> i actually checked g++ without and with O2 optimization, and if you use & on char constant, it puts data as a single character string which is padded by zeros, and with O2 even aligned with 0x10, but it could be just good practice by gcc, it may not be guaranteed, so warning is helpful in that case

I'm not sure whether it's related to optimization. I think the default alignment of the types whose size is less than a word length will be the word length of the machine, unless preprecess instructions like `#progma pack(1)` is used. This means there will be padding bytes after a char variable. But there isn't guarantee that the padding bytes will be zero.

qiluo-msft (approved):

kcudnik (approved):

PR #1622, opened 2021-02-02
[orchagent] Init field of supported speeds while port initialization

maksymbelei95:
* Reading of supported speed list from the ASIC and saving it inside
  PORT_TABLE:EthernetXX table in APPL_DB to be able to check supported
  speeds while changing a current speed through CLI commands.

Signed-off-by: Maksym Belei <Maksym_Belei@jabil.com>

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


orchagent/portsorch.cpp
```diff
@@ -3367,6 +3374,27 @@ bool PortsOrch::initializePort(Port &port)
         return false;
     }
 
+    /* 
+     * initialize field with supported speeds of the port.
+     */
+    auto supp_speed_temp = getSupportedSpeed(port.m_alias, port.m_port_id);
+    /* Copy the vector to be able to modify it */
+    auto supp_speed(supp_speed_temp);
+    if (supp_speed.size() > 0) {
```
prsunny, line 3383: { -> new line

orchagent/portsorch.cpp
```diff
@@ -3367,6 +3374,27 @@ bool PortsOrch::initializePort(Port &port)
...
+        sort(supp_speed.begin(), supp_speed.end());
+
+        string tmp_supp_speed_str = "";
+        /* Add (cound - 1) elements to the string with separator*/
```
prsunny, line 3387: count - typo

orchagent/portsorch.cpp
maksymbelei95: Done.

orchagent/portsorch.cpp
maksymbelei95: Done.

prsunny:
@dgsudharsan , can you review this?

dgsudharsan:
We also get the list of allowed speeds from platform.json eg ("breakout_modes": "1x100G[50G,40G,25G,10G],2x50G[40G,25G,10G]") The hardware supported speeds would be generally more than the allowable speeds listed in platform.json (which are the platform validated ones). Shouldn't we be using this instead of fetching from the hardware and using it as valid speeds?

maksymbelei95:
@dgsudharsan, @prsunny,

I am not sure about using of configuration of Breakout Mode feature for supported speed determination purpose. Is it a good idea to establish dependencies between those features? What if the Breakout Mode configuration parameters will not be provided in `platform.json` or a syntax on the parameters will be changed?
Currently, the system checks available speed from ASIC, but this makes after setting the desirable port speed to Config DB, and as there is no rollback mechanism for the speed parameter in `Orchagent`, we are getting invalid speed parameter, shown by command `show interface status`, for example. The idea is to prevent such situations, so, I decided to use information from ASIC little bit earlier, in `Sonic-utilities`.

dgsudharsan:
> @dgsudharsan, @prsunny,
> 
> I am not sure about using of configuration of Breakout Mode feature for supported speed determination purpose. Is it a good idea to establish dependencies between those features? What if the Breakout Mode configuration parameters will not be provided in `platform.json` or a syntax on the parameters will be changed?
> Currently, the system checks available speed from ASIC, but this makes after setting the desirable port speed to Config DB, and as there is no rollback mechanism for the speed parameter in `Orchagent`, we are getting invalid speed parameter, shown by command `show interface status`, for example. The idea is to prevent such situations, so, I decided to use information from ASIC little bit earlier, in `Sonic-utilities`.

Hi,
Please take a look at this. The speeds must be part of platform.json.
https://github.com/Azure/SONiC/blob/master/doc/dynamic-port-breakout/sonic-dynamic-port-breakout-HLD.md

A capability file for a platform with port breakout capabilities is provided. It defines the initial configurations like lanes, speed, alias etc. This file will be used for CLI later to pick the parent port and breakout mode. It can be used for speed checks based on the current mode. It (in conjunction with hwsku.json talked later) will also replace the functionality of the current existing port_config.ini.

Speeds in the breakout modes will be used for speed validation when we change the individual ports speed later. Eg, In 4x25G[10G] mode, we can change the individual port in that port group to 25G or 10G, but not some other speeds.

dgsudharsan:
@zhenggen-xu Please decide if this PR is needed given that port breakout handles the speed checks

zhenggen-xu:
I think we could keep both speed checks to get more protection. 

orchagent check is based on ASIC SAI for ports, and breakout mode and platform.json should be more restrict since the speed was based on the platform (not just ASIC) capabilities and breakout mods. In general, breakout speed check is better. However, not all platforms have adopted the breakout design,  and meanwhile having backend SAI checks won't be conflicting with that and can provide protection to some extend before all platforms are onboarded with breakout feature.

maksymbelei95:
@zhenggen-xu, so, if my understanding is correct, we could use this solution as a temporary, till breakout feature will be completely integrated to all the platforms? @prsunny, could you check the comment from @zhenggen-xu?

prsunny:
> @zhenggen-xu, so, if my understanding is correct, we could use this solution as a temporary, till breakout feature will be completely integrated to all the platforms? @prsunny, could you check the comment from @zhenggen-xu?

Fine with me. @zhenggen-xu, can we take up breakout later?

zhenggen-xu:
> > @zhenggen-xu, so, if my understanding is correct, we could use this solution as a temporary, till breakout feature will be completely integrated to all the platforms? @prsunny, could you check the comment from @zhenggen-xu?
> 
> Fine with me. @zhenggen-xu, can we take up breakout later?

We will make the check in CLI if dynamic breakout (DPB) is utilized, it is up to box vendor to onboard the platform with DPB feature.

maksymbelei95:
@prsunny, is there any updates regarding the PR?

dprital:
@prsunny  - Do you have an update for this PR ?

prsunny:
> @prsunny - Do you have an update for this PR ?

can you please rebase? Also @zhenggen-xu to signoff

KonstiantynHalushka:
@zhenggen-xu we made some changes regarding the use of common code
https://github.com/Azure/sonic-swss/commit/cbf362173dfd7b1e1188f82870332f20817f1505
https://github.com/Azure/sonic-swss/commit/b15a29f036f01076d909a1c0de4513980d502afb

zhenggen-xu:
> @zhenggen-xu we made some changes regarding the use of common code [cbf3621](https://github.com/Azure/sonic-swss/commit/cbf362173dfd7b1e1188f82870332f20817f1505) [b15a29f](https://github.com/Azure/sonic-swss/commit/b15a29f036f01076d909a1c0de4513980d502afb)

Can you update the PR with the changes so they could be reviewed here?

pshulik:
@zhenggen-xu, Could you please review the latest changes?

zhenggen-xu:
> @zhenggen-xu, Could you please review the latest changes?

@pshulik @KonstiantynHalushka  As mentioned above, please update the PR itself instead of sharing commits from different places so we can review the finial code here.

KonstiantynHalushka:
@zhenggen-xu @prsunny Sorry for the wrong update of the PR. Please, review the finial code.

orchagent/portsorch.cpp
```diff
@@ -1857,6 +1840,24 @@ void PortsOrch::getPortSupportedSpeeds(const std::string& alias, sai_object_id_t
...
+    // This method will return vector of supported speeds for the port
+    // The method will return empty vector if there was something wrong during method execution.
+
+    // "Lazy" query of supported speeds for given port
+    // Once received the list will be stored in m_portSupportedSpeeds
+
+    if (!m_portSupportedSpeeds.count(port_id))
+    {
+    	PortSupportedSpeeds supported_speeds;
+    	getPortSupportedSpeeds(alias, port_id, &supported_speeds);
```
zhenggen-xu, line 1854: should be `getPortSupportedSpeeds(alias, port_id, supported_speeds);`?

orchagent/portsorch.cpp
```diff
@@ -1873,6 +1874,22 @@ void PortsOrch::initPortSupportedSpeeds(const std::string& alias, sai_object_id_
     m_portStateTable.set(alias, v);
 }
 
+bool PortsOrch::isSpeedSupported(const std::string& alias, sai_object_id_t port_id, sai_uint32_t speed)
+{
+    // This method will return false if we get a list of supported speeds and the requested speed
+    // is not supported
+    // Otherwise the method will return true (even if did not receive list of supported speed)
+
```
zhenggen-xu, line 1882: Since we removed `initPortSupportedSpeeds` in this function, just want to make sure `initPortSupportedSpeeds` was called always before calling this function.

zhenggen-xu (changes requested):

KonstiantynHalushka:
@zhenggen-xu I accepted your comments

zhenggen-xu:
Just checked the PR again, I failed to understand the reason we introduce the new functions and code to collect the speeds.  Isn't that done in the existing function `initPortSupportedSpeeds()` where the speeds were saved to stateDB.  see: https://github.com/Azure/sonic-swss/blob/master/orchagent/portsorch.cpp#L1864 . Can we just leverage that for CLI check?   Any reason we need move them to appDB?   Even if we need to do so, why don't we just put the same info to appDB instead of all the new code?

Sorry for being late for these findings,  or maybe I am missing something here?

pshulik:
Hi zhenggen-xu 
The initPortSupportedSpeeds() is called one time during initialization. Issue that optical transceiver can be replaced during operation time  and supported speeds might be different, so, information obtained via initPortSupportedSpeeds()  might be not actual.

zhenggen-xu:
> Hi zhenggen-xu The initPortSupportedSpeeds() is called one time during initialization. Issue that optical transceiver can be replaced during operation time and supported speeds might be different, so, information obtained via initPortSupportedSpeeds() might be not actual.

The port speeds were obtained by SAI API CALLs which should not care about optical transceivers. This is purely ASIC capabilities,  and we can just reuse the same support speeds as got during init time?

PR #1629, opened 2021-02-08
[tlm teamd] Add retry mechanism before logging the ERR in get_dumps.

judyjoseph:
**Why I did**
Fix https://github.com/Azure/sonic-buildimage/issues/6632

There has been cases when the get_dumps API in tlm_teamd process is not able to get the right data and logs an error message. 

The issue occurs very rarely and it is due to the race condition between teammgrd/teamsyncd/tlm_teamd when a Portchannel is removed. In the teamd telemetry module there are two places where the get_dumps() is called.
1. When the portchannel object is add/removed. [https://github.com/Azure/sonic-swss/blob/master/tlm_teamd/main.cpp#L101]
2. On timeout of 1 sec. [https://github.com/Azure/sonic-swss/blob/master/tlm_teamd/main.cpp#L108]

In case of timeout call for get_dumps(), there could be an inconsistent state where the portchannel/teamd process is getting removed by teammgrd but the STATE table update to remove the lag interface is still not received by the tlm_teamd module.  

Seen below on a bad case where the get_dumps() call from TIMEOUT handler throws an ERR message - as the remove_lag message is not yet received.

On a good case 
```
Feb  7 02:03:27.576078 vlab-01 NOTICE teamd#tlm_teamd: :- remove_lag: The LAG 'PortChannel999' has been removed.
Feb  7 02:03:28.453829 vlab-01 INFO teamd#supervisord 2021-02-07 02:03:28,451 INFO reaped unknown pid 4747 (exit status 0)
Feb  7 02:03:28.458616 vlab-01 NOTICE teamd#teammgrd: :- removeLag: Stop port channel PortChannel999
```
On a bad case 

```
Feb  7 02:03:33.037401 vlab-01 ERR teamd#tlm_teamd: :- get_dump: Can't get dump for LAG 'PortChannel999'. Skipping
Feb  7 02:03:33.046179 vlab-01 NOTICE teamd#tlm_teamd: :- remove_lag: The LAG 'PortChannel999' has been removed.
Feb  7 02:03:33.997639 vlab-01 INFO teamd#supervisord 2021-02-07 02:03:33,996 INFO reaped unknown pid 4775 (exit status 0)
Feb  7 02:03:34.040126 vlab-01 NOTICE teamd#teammgrd: :- removeLag: Stop port channel PortChannel999
```

**How I did it**

Add retry mechanism before logging the ERR in get_dumps API(). The number of retries is set as 3. So that if the same error repeats 3 times - it is logged, other wise it is considered a transient condition - not an error. 

Additionally added a **to_retry** flag to get_dumps() API so that the caller can decide whether to use the retry mechanism or not.

**How I verified it**
Verified that the error message is no more seen in the syslog.
Confirmed by running ~ 200 times portchannel creation (which had reproduced the issue earlier on VS testbed). 

The new NOTICE message added in remove_lag shows that we had indeed hit the original issue earlier and clearing flags here.

```
admin@vlab-01:/var/log$ sudo zgrep -i "get dump for LAG" syslog*; sudo zgrep -i "clearing it" syslog*
syslog.1:Feb  8 06:41:54.995716 vlab-01 NOTICE teamd#tlm_teamd: :- remove_lag: The LAG 'PortChannel999' had errored while getting dump, clearing it
syslog.2.gz:Feb  8 06:31:32.360135 vlab-01 NOTICE teamd#tlm_teamd: :- remove_lag: The LAG 'PortChannel999' had errored while getting dump, clearing it
syslog.2.gz:Feb  8 06:36:16.617283 vlab-01 NOTICE teamd#tlm_teamd: :- remove_lag: The LAG 'PortChannel999' had errored while getting dump, clearing it
syslog.2.gz:Feb  8 06:37:56.906306 vlab-01 NOTICE teamd#tlm_teamd: :- remove_lag: The LAG 'PortChannel999' had errored while getting dump, clearing it
syslog.3.gz:Feb  8 06:25:44.442474 vlab-01 NOTICE teamd#tlm_teamd: :- remove_lag: The LAG 'PortChannel999' had errored while getting dump, clearing it
syslog.3.gz:Feb  8 06:27:02.539413 vlab-01 NOTICE teamd#tlm_teamd: :- remove_lag: The LAG 'PortChannel999' had errored while getting dump, clearing it
syslog.3.gz:Feb  8 06:27:42.785533 vlab-01 NOTICE teamd#tlm_teamd: :- remove_lag: The LAG 'PortChannel999' had errored while getting dump, clearing it
syslog.3.gz:Feb  8 06:29:33.510933 vlab-01 NOTICE teamd#tlm_teamd: :- remove_lag: The LAG 'PortChannel999' had errored while getting dump, clearing it
syslog.5.gz:Feb  8 06:08:03.643106 vlab-01 NOTICE teamd#tlm_teamd: :- remove_lag: The LAG 'PortChannel999' had errored while getting dump, clearing it
```

**Details if related**

### DIFF
--- a/tlm_teamd/main.cpp
+++ b/tlm_teamd/main.cpp
@@ -98,7 +98,7 @@ int main()
             if (res == swss::Select::OBJECT)
             {
                 update_interfaces(sst_lag, teamdctl_mgr);
-                values_store.update(teamdctl_mgr.get_dumps());
+                values_store.update(teamdctl_mgr.get_dumps(false));
             }
             else if (res == swss::Select::ERROR)
             {
@@ -108,7 +108,10 @@ int main()
             else if (res == swss::Select::TIMEOUT)
             {
                 teamdctl_mgr.process_add_queue();
-                values_store.update(teamdctl_mgr.get_dumps());
+		// When get_dumps API is triggered per TIMEOUT interval as done here, 
+		// there is a possible race condition on lag resource removal and the error
+		// will be reproted incorrectly. So set the param to_retry = true.
+                values_store.update(teamdctl_mgr.get_dumps(true));
             }
             else
             {
--- a/tlm_teamd/main.cpp
+++ b/tlm_teamd/main.cpp
@@ -108,9 +108,9 @@ int main()
             else if (res == swss::Select::TIMEOUT)
             {
                 teamdctl_mgr.process_add_queue();
-                // When get_dumps API is triggered per TIMEOUT interval as done here, 
-                // there is a possible race condition on lag resource removal and the error
-                // will be reproted incorrectly. So set the param to_retry = true.
+                // In the case of lag removal, there is a scenario where the select::TIMEOUT
+		// occurs, it triggers get_dumps incorrectly for resource which was in process of 
+		// getting deleted. The fix here is to retry and check if this is a real failure.
                 values_store.update(teamdctl_mgr.get_dumps(true));
             }
             else
--- a/tlm_teamd/main.cpp
+++ b/tlm_teamd/main.cpp
@@ -108,9 +108,9 @@ int main()
             else if (res == swss::Select::TIMEOUT)
             {
                 teamdctl_mgr.process_add_queue();
-		// When get_dumps API is triggered per TIMEOUT interval as done here, 
-		// there is a possible race condition on lag resource removal and the error
-		// will be reproted incorrectly. So set the param to_retry = true.
+                // When get_dumps API is triggered per TIMEOUT interval as done here, 
+                // there is a possible race condition on lag resource removal and the error
+                // will be reproted incorrectly. So set the param to_retry = true.
                 values_store.update(teamdctl_mgr.get_dumps(true));
             }
             else
--- a/tlm_teamd/main.cpp
+++ b/tlm_teamd/main.cpp
@@ -109,8 +109,8 @@ int main()
             {
                 teamdctl_mgr.process_add_queue();
                 // In the case of lag removal, there is a scenario where the select::TIMEOUT
-		// occurs, it triggers get_dumps incorrectly for resource which was in process of 
-		// getting deleted. The fix here is to retry and check if this is a real failure.
+                // occurs, it triggers get_dumps incorrectly for resource which was in process of 
+                // getting deleted. The fix here is to retry and check if this is a real failure.
                 values_store.update(teamdctl_mgr.get_dumps(true));
             }
             else
--- a/tlm_teamd/teamdctl_mgr.cpp
+++ b/tlm_teamd/teamdctl_mgr.cpp
@@ -5,6 +5,12 @@
 
 #include "teamdctl_mgr.h"
 
+#define MAX_RETRY 3
+
+/// Store the last errored lag name and the retry count.
+static std::string last_errored_lag_name= std::string("");
+static int no_of_retry = 0;
+
 ///
 /// Custom function for libteamdctl logger. IT is empty to prevent libteamdctl to spam us with the error messages
 /// @param tdc teamdctl descriptor
@@ -136,6 +142,15 @@ bool TeamdCtlMgr::remove_lag(const std::string & lag_name)
     {
         SWSS_LOG_WARN("The LAG '%s' hasn't been added. Can't remove it", lag_name.c_str());
     }
+
+    // If this lag interface errored last time, clear it
+    if ((lag_name.compare(last_errored_lag_name) == 0) && (no_of_retry != 0))
+    {
+        SWSS_LOG_NOTICE("The LAG '%s' had errored while getting dump, clearing it", lag_name.c_str());
+        last_errored_lag_name.clear();
+        no_of_retry = 0;
+    }
+
     return true;
 }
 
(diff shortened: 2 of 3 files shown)


judyjoseph:
retest vs please

judyjoseph:
retest this please

judyjoseph:
retest vs please

tlm_teamd/teamdctl_mgr.cpp
```diff
@@ -181,7 +197,32 @@ TeamdCtlDump TeamdCtlMgr::get_dump(const std::string & lag_name)
         }
         else
         {
-            SWSS_LOG_ERROR("Can't get dump for LAG '%s'. Skipping", lag_name.c_str());
+            // In case of failure and retry flag is set, check if it fails for MAX_RETRY times.
```
lguohan, line 200: in line 196, if there is no issue, do we need to clear out no_of_retry to 0 there?

tlm_teamd/teamdctl_mgr.cpp
```diff
@@ -136,6 +142,15 @@ bool TeamdCtlMgr::remove_lag(const std::string & lag_name)
     {
         SWSS_LOG_WARN("The LAG '%s' hasn't been added. Can't remove it", lag_name.c_str());
     }
+
+    // If this lag interface errored last time, clear it
+    if ((lag_name.compare(last_errored_lag_name) == 0) && (no_of_retry != 0))
```
lguohan, line 147: should we move this to get_dump()?

tlm_teamd/teamdctl_mgr.cpp
```diff
@@ -5,6 +5,12 @@
 
 #include "teamdctl_mgr.h"
 
+#define MAX_RETRY 3
+
+/// Store the last errored lag name and the retry count.
+static std::string last_errored_lag_name= std::string("");
+static int no_of_retry = 0;
```
lguohan, line 12: wonder if we can move all these static variable into get_dump() function?

tlm_teamd/teamdctl_mgr.cpp
judyjoseph: I had the static variables here and this approach to clear here originally.

On further testing, I moved it out to remove API , as I found that the get_dump for that particular lag_name was not getting called again as the remove lag request ( which comes in between ) clears off that entry for the lag_name from the m_handlers. So the no_of_retry doesn't get cleared.

This particular race condition don't happen in the add_lag sequence as the way we create the teamd process + entry in APP DB is a good. This issue happens with **remove lag** case -- as we remove teamd process instance first .. and the APP DB update comes very delayed.

lguohan: can you publish new changes so that i can take a look at?

judyjoseph: I have added the check to clear the no_of_retry here, in the good case for the lag in get_dump. Thanks

tlm_teamd/teamdctl_mgr.cpp
judyjoseph: We would need this check here in remove_lag()... as I mentioned in earlier comment -- that the lag could get cleared from the m_handlers here in this call to remote_lag() from update_interfaces(). Hence without this here, the no_retry count won't be reset by the logic even if we add it at line 196.

tlm_teamd/teamdctl_mgr.cpp
judyjoseph: We would need to access this variables in remove_lag API also.

judyjoseph:
@lguohan Could you take a look at the changes again - thanks

judyjoseph:
@lguohan. I had made the changes you recommended in L#196 - but I am keeping the logic in removeLag for the reason I put earlier in comment. 
<<<
We would need this check here in remove_lag()... The lag resource could get cleared from the m_handlers data structure here (teamdctl_mgr.cpp#L133) due to call from in main.cpp#L98 --> update_interfaces() --> remove_lag(). Since it can get cleared as above when get_dumps() is called on timeout (main.cpp#L114), it won't reach teamdctl_mgr.cpp#L196 as m_handlers[lag_name] for that "lag" don't exist.
>>>
Could you take a look again ?

judyjoseph:
Closing this PR for now, as this error is not seen any more.

liat-grozovik:
@judyjoseph the issue of https://github.com/Azure/sonic-buildimage/issues/6632 can be reproduced without this patch easily but not with it. I strongly suggest to move on with the review and approval flow and get it to 20212 and above. 
We were able to reproduce it while with a simple script of a loop of add bulk of multiple port channels and then remote them.

judyjoseph:
@prsunny @lguohan could you review this PR again. This is an old PR which we closed earlier as it was not reproducible. 
The test failures are unrelated, will trigger a re-run.

tlm_teamd/teamdctl_mgr.cpp
judyjoseph: I have kept these variables global ( it is not static anymore, in updated code.). It is used in remove_lag() and get_dump() APIs

liat-grozovik:
@judyjoseph kindly reminder. what is the plan to merge and take to 202012 and 202106?

tlm_teamd/main.cpp
```diff
@@ -108,7 +108,10 @@ int main()
             else if (res == swss::Select::TIMEOUT)
             {
                 teamdctl_mgr.process_add_queue();
-                values_store.update(teamdctl_mgr.get_dumps());
+                // In the case of lag removal, there is a scenario where the select::TIMEOUT
+		// occurs, it triggers get_dumps incorrectly for resource which was in process of 
+		// getting deleted. The fix here is to retry and check if this is a real failure.
```
lguohan, line 113: please convert the tab to spaces

judyjoseph: @lguohan updated as per comments, could you review.

lguohan (approved):

PR #1631, opened 2021-02-08
[Mux] Neighbor handling based on FDB entry

prsunny:
**What I did**

1. Neighbor handling based on FDB entries. When a new neighbor is learnt, the code lookup FDB table to get the associated physical port and based on the mux status for the port, the neighbor is programmed as local or tunnel nexthop.
2. Log level/message revisited based on log observations from current deployments

**Why I did it**

**How I verified it**

_Ethernet96 mux status is - Standby_

```
show arp
Address       MacAddress         Iface            Vlan
------------  -----------------  ---------------  ------
192.168.0.101  00:00:11:22:33:45  Ethernet96       1000

Feb  6 02:48:21.931004 str2-7050cx3-acs-02 ERR swss#orchagent: :- getPort: Failed to get bridge port ID for FDB entry 00:00:11:22:33:55, rv:-5
Feb  6 02:48:21.931004 str2-7050cx3-acs-02 NOTICE swss#orchagent: :- getMuxPort: FDB entry not found: Vlan Vlan1000, mac 00:00:11:22:33:55
Feb  6 02:48:21.935757 str2-7050cx3-acs-02 NOTICE swss#orchagent: :- addNeighbor: Created neighbor ip 192.168.0.101, 00:00:11:22:33:55 on Vlan1000
Feb  6 02:48:21.937385 str2-7050cx3-acs-02 NOTICE swss#orchagent: :- addNextHop: Created next hop 192.168.0.101 on Vlan1000
Feb  6 02:48:21.938899 str2-7050cx3-acs-02 INFO swss#orchagent: :- update: Neigh 192.168.0.101 on Vlan1000, add 1, state 2
Feb  6 02:48:21.938899 str2-7050cx3-acs-02 NOTICE swss#orchagent: :- disableNeighbor: Neighbor disable request for 192.168.0.101 
Feb  6 02:48:21.940604 str2-7050cx3-acs-02 NOTICE swss#orchagent: :- removeNeighbor: Removed next hop 192.168.0.101 on Vlan1000
Feb  6 02:48:21.952963 str2-7050cx3-acs-02 NOTICE swss#orchagent: :- removeNeighbor: Removed neighbor 00:00:11:22:33:55 on Vlan1000
Feb  6 02:48:21.952989 str2-7050cx3-acs-02 INFO swss#orchagent: :- addTunnelRoute: Add tunnel route DB 'Vlan1000:192.168.0.101/32'
Feb  6 02:48:21.954947 str2-7050cx3-acs-02 NOTICE swss#orchagent: :- create_route: Created tunnel route to 192.168.0.101/32 
```

**Details if related**

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -19,6 +19,7 @@
 #include "portsorch.h"
 #include "aclorch.h"
 #include "routeorch.h"
+#include "fdborch.h"
 
 /* Global variables */
 extern Directory<Orch*> gDirectory;
@@ -27,6 +28,7 @@ extern NeighOrch *gNeighOrch;
 extern RouteOrch *gRouteOrch;
 extern AclOrch *gAclOrch;
 extern PortsOrch *gPortsOrch;
+extern FdbOrch *gFdbOrch;
 
 extern sai_object_id_t gVirtualRouterId;
 extern sai_object_id_t  gUnderlayIfId;
@@ -195,6 +197,9 @@ static sai_object_id_t create_tunnel(const IpAddress* p_dst_ip, const IpAddress*
     attr.id = SAI_TUNNEL_ATTR_PEER_MODE;
     attr.value.s32 = SAI_TUNNEL_PEER_MODE_P2P;
     tunnel_attrs.push_back(attr);
+
+    attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_MODE;
+    attr.value.s32 = SAI_TUNNEL_TTL_MODE_PIPE_MODEL;
 
     if (p_src_ip != nullptr)
     {
@@ -474,7 +479,7 @@ void MuxCable::updateNeighbor(NextHopKey nh, bool add)
     {
         mux_orch_->addNexthop(nh, mux_name_);
     }
-    else
+    else if (mux_name_ == mux_orch_->getNexthopMuxName(nh))
     {
         mux_orch_->removeNexthop(nh);
     }
@@ -507,9 +512,11 @@ void MuxNbrHandler::update(NextHopKey nh, sai_object_id_t tunnelId, bool add, Mu
             break;
         case MuxState::MUX_STATE_ACTIVE:
             neighbors_[nh.ip_address] = gNeighOrch->getLocalNextHopId(nh);
+            gNeighOrch->enableNeighbor(nh);
             break;
         case MuxState::MUX_STATE_STANDBY:
             neighbors_[nh.ip_address] = tunnelId;
+            gNeighOrch->disableNeighbor(nh);
             mux_cb_orch->addTunnelRoute(nh);
             create_route(pfx, tunnelId);
             break;
@@ -806,7 +813,6 @@ sai_object_id_t MuxOrch::getNextHopTunnelId(std::string tunnelKey, IpAddress& ip
     return it->second.nh_id;
 }
 
-
 MuxCable* MuxOrch::findMuxCableInSubnet(IpAddress ip)
 {
     for (auto it = mux_cable_tb_.begin(); it != mux_cable_tb_.end(); it++)
@@ -821,7 +827,7 @@ MuxCable* MuxOrch::findMuxCableInSubnet(IpAddress ip)
     return nullptr;
 }
 
-bool MuxOrch::isNeighborActive(IpAddress nbr, string alias)
+bool MuxOrch::isNeighborActive(const IpAddress& nbr, const MacAddress& mac, string& alias)
 {
     MuxCable* ptr = findMuxCableInSubnet(nbr);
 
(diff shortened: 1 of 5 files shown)


orchagent/muxorch.cpp
```diff
@@ -830,9 +836,90 @@ bool MuxOrch::isNeighborActive(IpAddress nbr, string alias)
         return ptr->isActive();
     }
 
+    string port;
+    if (!getMuxPort(mac, alias, port))
+    {
+        SWSS_LOG_INFO("Mux get port from FDB failed for '%s' mac '%s'",
```
gechiang, line 842: Q: Is this a true error case or it can happen "normally" during init time or due to in certain mux state?
If it is truly an error case, then this needs to be logged as error.
If it can happens at init time then it is ok as is.
But if this can happen quite regularly and not an error, perhaps we should set the log level to DEBUG so it does not keep filling up the syslog...

orchagent/muxorch.cpp
```diff
@@ -830,9 +836,90 @@ bool MuxOrch::isNeighborActive(IpAddress nbr, string alias)
...
+                       nbr.to_string().c_str(), mac.to_string().c_str());
+        return true;
+    }
+
+    if (!port.empty() && isMuxExists(port))
+    {
+        MuxCable* ptr = getMuxCable(port);
```
gechiang, line 849: Q: Any possibility that getMuxCable() itself may error back with ptr being NULL?

orchagent/muxorch.cpp
```diff
@@ -830,9 +836,90 @@ bool MuxOrch::isNeighborActive(IpAddress nbr, string alias)
...
+}
+
+bool MuxOrch::getMuxPort(const MacAddress& mac, const string& alias, string& portName)
+{
+    portName = std::string();
+    Port rif, port;
+
+    if (!gPortsOrch->getPort(alias, rif))
+    {
+        SWSS_LOG_INFO("Interface '%s' not found in port table", alias.c_str());
```
gechiang, line 863: Q: If this is a true error case, then log level should be error and not info.
If this is expected to occur during normal operation perhaps due to race condition, then we should lower it to DEBUG level instead so that frequent encounter of this will not cause syslog to be filled up.

orchagent/muxorch.cpp
```diff
@@ -830,9 +836,90 @@ bool MuxOrch::isNeighborActive(IpAddress nbr, string alias)
...
+        return false;
+    }
+
+    if (rif.m_type != Port::VLAN)
+    {
+        SWSS_LOG_INFO("Interface type for '%s' is not Vlan, type %d", alias.c_str(), rif.m_type);
```
gechiang, line 869: Same question as above...

orchagent/muxorch.cpp
```diff
@@ -830,9 +836,90 @@ bool MuxOrch::isNeighborActive(IpAddress nbr, string alias)
...
+        return false;
+    }
+
+    if (!gFdbOrch->getPort(mac, rif.m_vlan_info.vlan_id, port))
+    {
+        SWSS_LOG_NOTICE("FDB entry not found: Vlan %s, mac %s", alias.c_str(), mac.to_string().c_str());
```
gechiang, line 875: Instead of using "alias.c_str()" for VLAN, should it be "rif.m_vlan_info.vlan_id" instead?
Also, please check LOG severity setting here to reflect the true severity (see my previous question above)

orchagent/muxorch.cpp
```diff
@@ -841,8 +928,44 @@ void MuxOrch::updateNeighbor(const NeighborUpdate& update)
         if (ptr->isIpInSubnet(update.entry.ip_address))
         {
             ptr->updateNeighbor(update.entry, update.add);
+            return;
```
gechiang, line 931: Q: Just for my understanding here. So when this function is entered, and the update ip address is in Subnet and is already in the mux_cable_tb, we just need to updateNeighbor and nothing else needs to be handled?

orchagent/muxorch.cpp
```diff
@@ -841,8 +928,44 @@ void MuxOrch::updateNeighbor(const NeighborUpdate& update)
...
+    {
+        return;
+    }
+    else if (update.add)
+    {
+        /* Check if the neighbor already exists */
+        old_port = getNexthopMuxName(update.entry);
+        addNexthop(update.entry);
+    }
+    else
```
gechiang, line 946: Assuming there are no other update type and when reached here it must be the delete case?

orchagent/muxorch.cpp
prsunny: This can return false if neighbor is learnt over a portchannel interface (T1) for which there is no mux port associated. For such entries, it is not expected to happen frequently.

orchagent/muxorch.cpp
prsunny: no, in that case `isMuxExists `returns false

orchagent/muxorch.cpp
prsunny: This is not expected in our normal operation, so just added for a safety check. I'll change it to ERROR

orchagent/muxorch.cpp
prsunny: The can happen in normal scenario. Will change to debug level

orchagent/muxorch.cpp
prsunny: both would convey the same data - `Vlan1000 `vs `1000`. I think, this severity can be at NOTICE level if it cannot find FDB entry.

orchagent/muxorch.cpp
prsunny: thats right.. because we already know the mux cable and doesn't have to look fdb for it

orchagent/muxorch.cpp
prsunny: yes

orchagent/muxorch.cpp
```diff
@@ -196,6 +198,9 @@ static sai_object_id_t create_tunnel(const IpAddress* p_dst_ip, const IpAddress*
     attr.value.s32 = SAI_TUNNEL_PEER_MODE_P2P;
     tunnel_attrs.push_back(attr);
 
+    attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_MODE;
+    attr.value.s32 = SAI_TUNNEL_TTL_MODE_PIPE_MODEL;
+
```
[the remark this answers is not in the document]
prsunny: good catch, will fix

orchagent/muxorch.cpp
gechiang: In the previous OS that I worked on the IP interface name could be named different from the VLAN that is is mapped to.  Not sure if that is also applicable to SONiC...  If Not possible with SONiC it should be fine.

gechiang (approved):

orchagent/muxorch.cpp
prsunny: IMO, its not applicable in Sonic. We have this naming convention of interface prefixed as "Vlan" and typically its configured as Vlan+id.

gechiang: Sure.  Change is not required.

prsunny:
retest vs please

orchagent/muxorch.cpp
```diff
@@ -821,28 +828,159 @@ MuxCable* MuxOrch::findMuxCableInSubnet(IpAddress ip)
...
+    return true;
+}
+
+bool MuxOrch::getMuxPort(const MacAddress& mac, const string& alias, string& portName)
+{
+    portName = std::string();
+    Port rif, port;
+
+    if (!gPortsOrch->getPort(alias, rif))
```
lguohan, line 867: >rif [](start = 36, length = 3)

rif is usually router interface. is it router interface here? i though it is a normal port.

prsunny: This is in the rif context.. alias here is for e.g 'Vlan1000' or any rif where neighbor is learnt.  Actual port is obtained at line 879

gechiang (approved):

lguohan (approved):

PR #1636, opened 2021-02-11
[ci]: run vstest

lguohan:

**What I did**
add vstest in azure pipeline

**Why I did it**

**How I verified it**
check pr

image

**Details if related**

### DIFF
--- a/.azure-pipelines/build-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/build-docker-sonic-vs-template.yml
@@ -1,0 +1,85 @@
+parameters:
+- name: arch
+  type: string
+  values:
+  - amd64
+  - armhf
+  - arm64
+  default: amd64
+
+- name: timeout
+  type: number
+  default: 60
+
+- name: swss_artifact_name
+  type: string
+
+- name: sairedis_artifact_name
+  type: string
+
+- name: swss_common_artifact_name
+  type: string
+
+- name: artifact_name
+  type: string
+
+jobs:
+- job:
+  displayName: ${{ parameters.arch }}
+  timeoutInMinutes: ${{ parameters.timeout }}
+
+  pool:
+    vmImage: 'ubuntu-20.04'
+
+  steps:
+  - task: DownloadPipelineArtifact@2
+    inputs:
+      source: specific
+      project: build
+      pipeline: 9
+      artifact: ${{ parameters.swss_common_artifact_name }}
+      runVersion: 'latestFromBranch'
+      runBranch: 'refs/heads/master'
+    displayName: "Download sonic swss common deb packages"
+  - task: DownloadPipelineArtifact@2
+    inputs:
+      source: specific
+      project: build
+      pipeline: 12
+      artifact: ${{ parameters.sairedis_artifact_name }}
+      runVersion: 'latestFromBranch'
+      runBranch: 'refs/heads/master'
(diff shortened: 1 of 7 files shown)


.azure-pipelines/test-docker-sonic-vs-template.yml
```diff
@@ -0,0 +1,55 @@
...
+      sudo .azure-pipelines/build_and_install_team.sh
+
+      sudo apt-get install -y libhiredis0.14
+      sudo dpkg -i --force-confask,confnew ../libswsscommon_1.0.0_amd64.deb || apt-get install -f
+      sudo dpkg -i ../python3-swsscommon_1.0.0_amd64.deb
+
+      # install packages for vs test
+      sudo apt-get install -y net-tools bridge-utils vlan
+      sudo apt-get install -y python3-pip
+      sudo pip3 install pytest==4.6.2 attrs==19.1.0 exabgp==4.0.10 distro==1.5.0 docker==4.4.1 redis==3.3.4 flaky dataclasses fstring
```
lguohan, line 41: @daall , I could not find the where the fstring is used in the test code. also dataclasses seems to be available natively after python3.6

daall: That is right, those instructions are only in the README for 18.04 users. Since I assume the azure pipeline is using 20.04 or later we don't need to explicitly add `fstring` or `dataclasses`.

.azure-pipelines/build-docker-sonic-vs-template.yml
```diff
@@ -0,0 +1,85 @@
...
+      project: build
+      pipeline: 12
+      artifact: ${{ parameters.sairedis_artifact_name }}
+      runVersion: 'latestFromBranch'
+      runBranch: 'refs/heads/master'
+    displayName: "Download sonic sairedis deb packages"
+  - task: DownloadPipelineArtifact@2
+    inputs:
+      artifact: ${{ parameters.swss_artifact_name }}
+    displayName: "Download sonic swss artifact"
```
daall, line 56: Question - how does this know where to get the built `swss` from? Is that covered by the `swss_artifact_name` param?

daall (commented):
See comments. As a general question, is there a way for `sonic-sairedis` and `sonic-swss-common` to access the components here so we can set up these tests to run on those PRs?

.azure-pipelines/build-docker-sonic-vs-template.yml
lguohan: for this one, it download the artifact that is produced in previous stage.

.azure-pipelines/test-docker-sonic-vs-template.yml
lguohan: will remvoe

lguohan:
> See comments. As a general question, is there a way for sonic-sairedis and sonic-swss-common to access the components here so we can set up these tests to run on those PRs?

i assume your question is about how can use this for sairedis and common test? I guess we would like replicate this in those repo.  i prefer to have everything needs in those repo so that whenever we change the ci file, it will trigger a rebuild and we can validate ci changes in the pr.

daall (approved):

lguohan:
@daall , all test passed. sign-off?

daall (approved):

PR #1639, opened 2021-02-13
[bufferorch] Handle NOT IMPLEMENTED status returned during set attr operation

neethajohn:
Signed-off-by: Neetha John <nejo@microsoft.com>

**What I did**
Set operations for some attributes of buffer pool and buffer profile may not be implemented by some vendors and they return SAI_STATUS_ATTR_NOT_IMPLEMENTED_0. Handle that status in the code and ignore that task

**Why I did it**
Reduce the noise in the logs due to these operations

**How I verified it**
With the code changes in place triggered a warm reboot where these were logged as errors and see them logged as NOTICE


vaibhavhd (approved):

yxieca (approved):

lguohan:
test failure fixed in https://github.com/Azure/sonic-swss/pull/1640

PR #1640, opened 2021-02-13
[test_virtual_chassis]: use wait_for to make test more robust

lguohan:

**What I did**
https://github.com/lguohan/sonic-swss/pull/new/vcha

**Why I did it**

**How I verified it**
pr 
**Details if related**


lguohan:
merge to avoid future test failures.

PR #1641, opened 2021-02-14
fixed unsupported resource issue

PrabhuSreenivasan:
Signed-off-by: Prabhu Sreenivasan <prabhu.sreenivasan@broadcom.com>

Fix Azure/sonic-buildimage#6563

**What I did**
Instead of erasing unsupported resource from m_resourcesMap, mark it as unsupported.

**Why I did it**
Erasing an entry while iterating using range based iterator makes the iterator invalid.
Also removing a resource from the m_resourcesMap impacts cross access while iterating other maps (eg crmUsedCntsTableMap).

**How I verified it**
Added a new unsupported resource for testing and verified the logs.

**Details if related**

### DIFF
--- a/orchagent/crmorch.cpp
+++ b/orchagent/crmorch.cpp
@@ -439,6 +439,10 @@ void CrmOrch::getResAvailableCounters()
         sai_attribute_t attr;
         attr.id = crmResSaiAvailAttrMap.at(res.first);
 
+        // ignore unsupported resources
+        if (res.second.resStatus != CrmResourceStatus::CRM_RES_SUPPORTED)
+            continue;
+
         switch (attr.id)
         {
             case SAI_SWITCH_ATTR_AVAILABLE_IPV4_ROUTE_ENTRY:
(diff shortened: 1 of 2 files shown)


orchagent/crmorch.cpp
```diff
@@ -439,6 +439,10 @@ void CrmOrch::getResAvailableCounters()
         sai_attribute_t attr;
         attr.id = crmResSaiAvailAttrMap.at(res.first);
 
+        // ignore unsupported resources
+        if (res.second.resStatus != CrmResourceStatus::CRM_RES_SUPPORTED)
+            continue;
```
nazariig, line 444: @PrabhuSreenivasan please use `{ }` notation and move it at the beginning of the loop

PrabhuSreenivasan: @nazariig , Thanks. I have made the suggested changes.

nazariig (approved):

orchagent/crmorch.cpp
nazariig: @PrabhuSreenivasan NP

liat-grozovik (approved):

qiluo-msft (approved):

PR #1643, opened 2021-02-17
[acl] Add VLAN bindings to dataplane ACL tables

daall:

**What I did**
I added the VLAN bind option to ACL tables.

**Why I did it**
So that ACLs can be bound to VLAN ports.

**How I verified it**
Updated tests. Also validating locally.

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1332,7 +1332,7 @@ bool AclTable::create()
     }
     else
     {
-        bpoint_list = { SAI_ACL_BIND_POINT_TYPE_PORT, SAI_ACL_BIND_POINT_TYPE_LAG };
+        bpoint_list = { SAI_ACL_BIND_POINT_TYPE_PORT, SAI_ACL_BIND_POINT_TYPE_LAG, SAI_ACL_BIND_POINT_TYPE_VLAN };
     }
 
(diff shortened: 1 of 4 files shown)


qiluo-msft (approved):

lguohan:
can you add vstest for this? 

what does the config db entry looks like, can you share example?

orchagent/aclorch.cpp
```diff
@@ -1332,7 +1332,7 @@ bool AclTable::create()
     }
     else
     {
-        bpoint_list = { SAI_ACL_BIND_POINT_TYPE_PORT, SAI_ACL_BIND_POINT_TYPE_LAG };
+        bpoint_list = { SAI_ACL_BIND_POINT_TYPE_PORT, SAI_ACL_BIND_POINT_TYPE_LAG, SAI_ACL_BIND_POINT_TYPE_VLAN };
```
prsunny, line 1335: will this be supported on all platforms?

daall: I am still checking, that is why this is in draft :)

liat-grozovik:
@daall i think you are missing tests to cover the new configuration. if this is now allowed need to add specific case for adding and deleting this type of acl. i agree that it will be better to have a description of the schema of how to configure this type of ACL. we have few examples in the swss chema md file. 

there is also a sonic.mgmt test which covers acl. will you have a seperated PR for sonic.mgmt to extend the test configuration and check functionality?

PR #1647, opened 2021-02-19
Update vxlanmgrd.cpp

RajeshPukhrajJain:
**What I did**
Removed the dependency of vxlanmgr to wait for the swss to reconcile during the warmboot process.

**Why I did it**
The orchagent architecture already address this use-case scenario where it will ensure that the sequence of the new events from application are picked only after replaying all the existing config during the warmboot process.

**How I verified it**
Tested the warmboot and ensured that the replay/reconcile of the vxlanmgrd is not dependent on the swss reconcile state.

**Details if related**

### DIFF
--- a/cfgmgr/vxlanmgrd.cpp
+++ b/cfgmgr/vxlanmgrd.cpp
@@ -94,27 +94,9 @@ int main(int argc, char **argv)
 
         if (WarmStart::isWarmStart())
         {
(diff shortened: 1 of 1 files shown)


srj102:
retest this please

prsunny:
retest this please

prsunny (dismissed):

prsunny (approved):

PR #1648, opened 2021-02-20
[acl] Enable VLAN ID qualifier for ACL rules

daall:

**What I did**
I added support for the OUTER_VLAN_ID field to ACL rules.

**Why I did it**
To allow users to create rules that match on the VLAN ID of a packet.

**How I verified it**
See VS test updates. I also ran traffic tests against 201911 and master image.

**Details if related**


orchagent/aclorch.cpp
```diff
@@ -288,6 +292,17 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
             value.aclfield.data.u16 = to_uint<uint16_t>(attr_value);
             value.aclfield.mask.u16 = 0xFFFF;
         }
+        else if (attr_name == MATCH_VLAN_ID)
+        {
+            value.aclfield.data.u16 = to_uint<uint16_t>(attr_value);
+            value.aclfield.mask.u16 = 0xFFF;
```
qiluo-msft, line 298: >0xFFF [](start = 38, length = 5)

0xFF ?

qiluo-msft (approved):

orchagent/aclorch.cpp
```diff
@@ -1392,6 +1407,10 @@ bool AclTable::create()
         table_attrs.push_back(attr);
     }
 
+    attr.id = SAI_ACL_TABLE_ATTR_FIELD_OUTER_VLAN_ID;
```
prsunny, line 1410: So this is expected to be set for all tables except PFC_WD and ACL_TABLE_MIRROR_DSCP?

orchagent/aclorch.cpp
prsunny, line 298: Should this be 0xFFFF?

qiluo-msft: 4095 == 0xFFF

orchagent/aclorch.cpp
daall: L3, L3V6, MIRROR, and MIRRORV6. DSCP and PFCWD are skipped in this method, and MCLAG and DTEL use their own init methods.

orchagent/aclorch.cpp
daall: Exactly what @qiluo-msft said, VLAN ID is a 12 bit field so we need a 12 bit mask for it.

prsunny (approved):

qiluo-msft (approved):

abdosi:
@daall please create pr for 201911. VS tests have conflict.

daall:
> @daall please create pr for 201911. VS tests have conflict.
> 
> @qiluo-msft

ack, thx

PR #1650, opened 2021-02-24
[intfsorch] Create subport with the entry contains necessary attributes

peter-yu:
redis APPL_DB has two entries for subport
8) "INTF_TABLE:Ethernet4.20"
9) "INTF_TABLE:Ethernet4.20:192.169.1.1/24"

sometimes, if the entry with ip key are execute first,
the syslog is like
Feb 23 02:20:43.650184 as7726-32x-3 NOTICE swss#orchagent: :- addRouterIntfs: Create router interface Ethernet4.20 MTU 32761
Feb 23 02:20:43.651383 as7726-32x-3 ERR syncd#syncd:[none] brcm_sai_create_router_interface:284 Invalid MTU size

the mtu value is not init and is too large. sai will return SAI_STATUS_INVALID_PARAMETER

Thus, mtu should be initialized to 0

**What I did**
initialize mtu to 0
**Why I did it**
sometimes gPortsOrch->addSubPort(port, alias, adminUp, mtu) is called and 
the mtu value is not init and is too large. sai will return SAI_STATUS_INVALID_PARAMETER

Thus, mtu should be initialized to 0

**How I verified it**
create/remove subport interface many times

**Details if related**

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -588,8 +588,8 @@ void IntfsOrch::doTask(Consumer &consumer)
         string vrf_name = "", vnet_name = "", nat_zone = "";
         MacAddress mac;
 
-        uint32_t mtu;
-        bool adminUp;
+        uint32_t mtu = 0;
+        bool adminUp = false;
         uint32_t nat_zone_id = 0;
         string proxy_arp = "";
         string inband_type = "";
@@ -742,7 +742,7 @@ void IntfsOrch::doTask(Consumer &consumer)
             Port port;
             if (!gPortsOrch->getPort(alias, port))
             {
-                if (isSubIntf)
+                if (!ip_prefix_in_key && isSubIntf)
                 {
                     if (!gPortsOrch->addSubPort(port, alias, adminUp, mtu))
                     {


andriy-kokhan-plv:
I'd extend PR description to explicitly mentioned that in case MTU=0 the subport will inherit MTU value from the parent port.

andriy-kokhan-plv (approved):

orchagent/intfsorch.cpp
```diff
@@ -588,7 +588,7 @@ void IntfsOrch::doTask(Consumer &consumer)
         string vrf_name = "", vnet_name = "", nat_zone = "";
         MacAddress mac;
 
-        uint32_t mtu;
+        uint32_t mtu = 0;
```
wendani, line 591: Port object is not supposed to be created with "INTF_TABLE:Ethernet4.20:192.169.1.1/24"

I believe on line 745, we need to amend the condition to create Port object as `&& !ip_prefix_in_key`

wendani (changes requested):
as comments

orchagent/intfsorch.cpp
```diff
@@ -744,7 +744,7 @@ void IntfsOrch::doTask(Consumer &consumer)
             {
                 if (isSubIntf)
                 {
-                    if (!gPortsOrch->addSubPort(port, alias, adminUp, mtu))
+                    if (!ip_prefix_in_key && !gPortsOrch->addSubPort(port, alias, adminUp, mtu))
```
wendani, line 747: Should be with line 745

orchagent/intfsorch.cpp
peter-yu: Thanks. Done.

orchagent/intfsorch.cpp
peter-yu: Done.

wendani (approved):

prsunny (approved):

PR #1656, opened 2021-03-01
[mux] VS test for neigh, route and fdb

prsunny:
**What I did**

VS test to cover:

1. IPv4 and V6 neighbor add and simulate mux state change.  Verify: _Active mux shall have host entry, Standby mux shall have tunnel route_
2. Route add and simulate mux state change. Verify: _Route should have next hop programmed based on mux state (Active shall have neighbor NH, standby shall have tunnel NH)_
3.  ECMP route add and simulate mux state change. Verify: _ECMP group shall be updated based on mux state change, active mux shall have regular nexthop member, standby shall have tunnel nexthop member_
3. Neighbor handing based on FDB learnt. Verify: _Neighbor with FDB learnt on active mux shall have host entry whereas on standby mux shall have tunnel route_

**Why I did it**
unit-test

**How I verified it**

Run `pytest test_mux.py --dvsname=vs --forcedvs`

**Details if related**

### DIFF
--- a/tests/test_mux.py
+++ b/tests/test_mux.py
@@ -201,7 +201,7 @@ class TestMuxTunnelBase(object):
 
         # In standby mode, the entry must not be added to Neigh table but Route
         asicdb.wait_for_matching_keys(self.ASIC_NEIGH_TABLE, existing_keys)
-        rtkeys = dvs_route.check_asicdb_route_entries([self.SERV2_IPV4+self.IPV4_MASK, self.SERV2_IPV6+self.IPV6_MASK])
+        dvs_route.check_asicdb_route_entries([self.SERV2_IPV4+self.IPV4_MASK, self.SERV2_IPV6+self.IPV6_MASK])
 
         # The first standby route also creates as tunnel Nexthop
         self.check_tnl_nexthop_in_asic_db(asicdb, 3)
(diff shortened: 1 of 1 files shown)


tests/test_mux.py
```diff
@@ -40,45 +46,307 @@ class TestMuxTunnelBase(object):
...
+            fvs = asicdb.get_entry(self.ASIC_NEXTHOP_TABLE, key)
+            if fvs.get("SAI_NEXT_HOP_ATTR_TYPE") == "SAI_NEXT_HOP_TYPE_TUNNEL_ENCAP":
+                tunnel_nh_id = key
+
+        assert tunnel_nh_id
+
+
+    def check_nexthop_in_asic_db(self, asicdb, key, standby=False):
+
+        global tunnel_nh_id
```
daall, line 118: Is this needed if we're not setting `tunnel_nh_id`?

tests/test_mux.py
```diff
@@ -40,45 +46,307 @@ class TestMuxTunnelBase(object):
...
+
+        fvs = asicdb.get_entry("ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY", key)
+
+        nhg_id = fvs["SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID"]
+
+        fvs = asicdb.get_entry("ASIC_STATE:SAI_OBJECT_TYPE_NEXT_HOP_GROUP", nhg_id)
+
+        assert bool(fvs)
```
daall, line 141: If you use `wait_for_entry` these two lines can be combined

tests/test_mux.py
```diff
@@ -40,45 +46,307 @@ class TestMuxTunnelBase(object):
...
+
+        keys = asicdb.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_NEXT_HOP_GROUP_MEMBER")
+
+        # Two NH group members are expected to be added
+        assert len(keys) == 2
```
daall, line 146: Same with these two lines and `wait_for_n_keys`

prsunny: done

tests/test_mux.py
prsunny: this is need as its used for checking in this function. Only one tunnel_nh_id exists in the system, so using it as a global var.

daall: That makes sense, but is `global` still needed if we're just reading the variable?

prsunny: I see your point. Addressed.

daall (approved):

PR #1658, opened 2021-03-02
[portorch] parse on/off value from autoneg

dmytroxshevchuk:
As part of PR in sonic-buildimage

**What I did**
Changed parsing `autoneg` from 1/0 to on/off
 
**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2228,7 +2228,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 /* Set autoneg and ignore the port speed setting */
                 else if (fvField(i) == "autoneg")
                 {
-                    an = (int)stoul(fvValue(i));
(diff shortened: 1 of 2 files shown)


orchagent/portsorch.cpp
```diff
@@ -2228,7 +2225,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 /* Set autoneg and ignore the port speed setting */
                 else if (fvField(i) == "autoneg")
                 {
-                    an = (int)stoul(fvValue(i));
+                    an = fvValue(i);
```
lguohan, line 2228: we can change this as an = (fvValue(i) == "on")

then, we do not need to change an throughout the code.

dmytroxshevchuk: done, please check

lguohan (approved):

PR #1659, opened 2021-03-03
Keep attribute order in bulk mode

shi-su:
**What I did**
Keep attributes for the same entry in the order that the attributes are set in bulk mode.

Fix https://github.com/Azure/sonic-buildimage/issues/6881

**Why I did it**
Some attributes have a dependency on others in SAI operations (e.g., setting nexthop requires packet action to be forward in route set operation). Without keeping the order of the attributes, the configuration may not be successfully applied to hardware. Therefore, it is necessary to keep the order of attributes in bulk mode.

**How I verified it**
Verify that sairedis applys `SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION` before `SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID` when setting default routes.

```
2021-03-03.06:01:12.003330|S|SAI_OBJECT_TYPE_ROUTE_ENTRY||{"dest":"::/0","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000022"}|SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION=SAI_PACKET_ACTION_FORWARD||{"dest":"::/0","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000022"}|SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID=oid:0x5000000000687
2021-03-03.06:01:12.945653|S|SAI_OBJECT_TYPE_ROUTE_ENTRY||{"dest":"0.0.0.0/0","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000022"}|SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION=SAI_PACKET_ACTION_FORWARD||{"dest":"0.0.0.0/0","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000022"}|SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID=oid:0x5000000000683
```

**Details if related**

### DIFF
--- a/orchagent/bulker.h
+++ b/orchagent/bulker.h
@@ -208,10 +208,10 @@ public:
         if (found_setting != setting_entries.end())
         {
             // Mark old one as done
-            auto& attrmap = found_setting->second;
-            for (auto& attr: attrmap)
+            auto& attrs = found_setting->second;
+            for (auto& attr: attrs)
             {
-                *attr.second.second = SAI_STATUS_SUCCESS;
+                *attr.second = SAI_STATUS_SUCCESS;
             }
             // Erase old one
             setting_entries.erase(found_setting);
@@ -252,27 +252,15 @@ public:
         if (!attr) throw std::invalid_argument("attr is null");
 
         // Insert or find the key (entry)
-        auto& attrmap = setting_entries.emplace(std::piecewise_construct,
+        auto& attrs = setting_entries.emplace(std::piecewise_construct,
                 std::forward_as_tuple(*entry),
                 std::forward_as_tuple()
         ).first->second;
 
-        // Insert or find the key (attr)
-        auto rc = attrmap.emplace(std::piecewise_construct,
-                std::forward_as_tuple(attr->id),
-                std::forward_as_tuple());
-        bool inserted = rc.second;
-        auto it = rc.first;
-
-        // If inserted new key, assign the attr
-        // If found existing key, overwrite the old attr
-        it->second.first = *attr;
-        if (!inserted)
-        {
-            // If found existing key, mark old status as success
-            *it->second.second = SAI_STATUS_SUCCESS;
-        }
-        it->second.second = object_status;
+        // Insert attr
+        attrs.emplace_back(std::piecewise_construct,
+                std::forward_as_tuple(*attr),
+                std::forward_as_tuple(object_status));
         *object_status = SAI_STATUS_NOT_EXECUTED;
     }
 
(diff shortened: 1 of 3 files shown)


orchagent/bulker.h
```diff
@@ -252,27 +252,15 @@ class EntityBulker
...
 
-        // Insert or find the key (attr)
-        auto rc = attrmap.emplace(std::piecewise_construct,
-                std::forward_as_tuple(attr->id),
-                std::forward_as_tuple());
-        bool inserted = rc.second;
-        auto it = rc.first;
-
-        // If inserted new key, assign the attr
-        // If found existing key, overwrite the old attr
```
qiluo-msft, line 268: >If found existing key, overwrite the old attr [](start = 11, length = 45)

One advantage or map structure is possible to merge setting operation on the same entry and same attribute. Is it still required?

shi-su: I think it does not seem very necessary to merge the operations for the same attribute and the same entry because of the following reasons:
1. The same entry should not show up more than once in one bulk.
2. For each route entry, there are two attributes that might be set in bulk mode: `PACKET_ACTION` and `NEXT_HOP_ID`, each gets to set only once in the routeorch logic.
3. For other types of entries (not supported in bulk yet), the to_sync logic should already merge the same attribute for the same entry. It does not seem necessary to do it again in bulk.

kcudnik:
what ? order should not matter at all in any way to any component including SAI

shi-su:
> what ? order should not matter at all in any way to any component including SAI

It is reported in https://github.com/Azure/sonic-buildimage/issues/6881 that the hardware has a contract that route action must not be DROP when setting next hops (otherwise it gets ignored). The order matters if this is the case since the action must be set first in this scenario.

lguohan:
missing unit test?

lguohan (dismissed):

qiluo-msft (approved):

tests/mock_tests/bulker_ut.cpp
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        void TearDown() override
+        {
+            delete sai_route_api;
+            sai_route_api = nullptr;
+        }
+    };
+
+    TEST_F(BulkerTest, BulkerAttrOrder)
```
qiluo-msft, line 29: >BulkerTest [](start = 11, length = 10)

Did this unit test fail old code?

qiluo-msft (approved):

PR #1660, opened 2021-03-03
[teammgrd]: Create unique LACP key for port-channel

DavidZagury:
**What I did**
Fix https://github.com/Azure/sonic-buildimage/issues/4009

When a member port is added to port-channel, create a unique LACP key.
When adding a member port to port-channel set the LACP key to a unique number.
The number is extracted from the port-channel name and will be the number at the end of the port-channel name with an additional digit at the beginning in order to make sure that this number will be unique in the system.

**Why I did it**
If LACP key is not set, then the peer will not be able to distinguish the ports which are connected to different port-channels, as it will receive the LACP key as 0 for all the ports from different port-channels.

**How I verified it**
I configure a SONiC switch to have two port-channels and on a second switch, I created one port-channel for both links between the switches.
On the second switch only one of the ports comes up in the PO and the other one stayed down.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

### DIFF
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -504,6 +504,52 @@ bool TeamMgr::removeLag(const string &alias)
     return true;
 }
 
+// Port-channel names are in the pattern of "PortChannel####"
+// 
+// The LACP key could be generated in 3 ways based on the value in config DB:
+//      1. "auto" - LACP key is extracted from the port-channel name and is set to be the number at the end of the port-channel name
+//                  We are adding 1 at the beginning to avoid LACP key collisions between similar LACP keys e.g. PortChannel10 and PortChannel010.
+//      2. n -      LACP key will be n.
+//      3. "" -     LACP key will be 0 - exists for backward compatibility.
+uint16_t TeamMgr::generateLacpKey(const string& lag)
+{
+    vector <FieldValueTuple> fvs;
+    m_cfgLagTable.get(lag, fvs);
+
+    auto it = find_if(fvs.begin(), fvs.end(), [](const FieldValueTuple& fv)
+    {
+        return fv.first == "lacp_key";
+    });
+    string lacp_key;
+    if (it != fvs.end())
+    {
+        lacp_key = it->second;
+        if (!lacp_key.empty())
+        {
+            try
+            {
+                if (lacp_key == "auto")
+                {
+                    return static_cast<uint16_t>(std::stoul("1" + lag.substr(lag.find_first_of("0123456789"))));
+                }
+                else
+                {
+                    return static_cast<uint16_t>(std::stoul(lacp_key));
+                }
+            }
+            catch (const std::exception& e)
+            {
+                SWSS_LOG_THROW("Failed to parse LACP key %s for port channel %s", lacp_key.c_str(), lag.c_str());
+            }
+        }
+        else
+        {
+            return 0;
+        }
+    }
+    return 0;
+}
+
 // Once a port is enslaved into a port channel, the port's MTU will
 // be inherited from the master's MTU while the port's admin status
 // will still be controlled separately.
@@ -520,11 +566,17 @@ task_process_status TeamMgr::addLagMember(const string &lag, const string &membe
 
     stringstream cmd;
     string res;
+    uint16_t keyId = generateLacpKey(lag);
 
     // Set admin down LAG member (required by teamd) and enslave it
     // ip link set dev <member> down;
+    // teamdctl <port_channel_name> port config update <member> { "lacp_key": <lacp_key>, "link_watch": { "name": "ethtool" } };
     // teamdctl <port_channel_name> port add <member>;
     cmd << IP_CMD << " link set dev " << shellquote(member) << " down; ";
+    cmd << TEAMDCTL_CMD << " " << shellquote(lag) << " port config update " << shellquote(member)
+        << " '{\"lacp_key\":"
+        << keyId
+        << ",\"link_watch\": {\"name\": \"ethtool\"} }'; ";
     cmd << TEAMDCTL_CMD << " " << shellquote(lag) << " port add " << shellquote(member);
 
     if (exec(cmd.str(), res) != 0)
--- a/tests/test_portchannel.py
+++ b/tests/test_portchannel.py
@@ -2,6 +2,7 @@ import time
 import re
 import json
 import pytest
+import itertools
 
 from swsscommon import swsscommon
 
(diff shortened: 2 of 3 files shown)


cfgmgr/teammgr.cpp
```diff
@@ -504,6 +519,52 @@ bool TeamMgr::removeLag(const string &alias)
...
+//      2. n -      LACP key will be n.
+//      3. "" -     LACP key will be 0 - exists for backward compatibility.
+uint16_t TeamMgr::generateLacpKey(const string& lag)
+{
+    vector <FieldValueTuple> fvs;
+    m_cfgLagTable.get(lag, fvs);
+
+    auto it = find_if(fvs.begin(), fvs.end(), [](const FieldValueTuple& fv)
+    {
+        return fv.first == "lacp_key";
```
lguohan, line 536: looks like if there is no lacp_key, then it is the old behavior?

tests/test_portchannel.py
```diff
@@ -89,6 +89,51 @@ def test_Portchannel(self, dvs, testlog):
...
+                            ("PortChannel000", "Ethernet12", 1000)]
+
+        self.adb = swsscommon.DBConnector(1, dvs.redis_sock, 0)
+        self.cdb = swsscommon.DBConnector(4, dvs.redis_sock, 0)
+        self.pdb = swsscommon.DBConnector(0, dvs.redis_sock, 0)
+
+        # Create 4 PortChannels
+        tbl = swsscommon.Table(self.cdb, "PORTCHANNEL")
+        fvs = swsscommon.FieldValuePairs(
+            [("admin_status", "up"), ("mtu", "9100"), ("oper_status", "up"), ("lacp_key", "auto")])
```
lguohan, line 105: can you test without lacp_key, lacp_key = "", lacp_key=n, all four cases?

dprital:
@lguohan  - please keep in mind that the fix for issue #4009 arrive with this PR and additional PR:
https://github.com/Azure/sonic-utilities/pull/1473

lguohan:
thank you. commented on 1473 pr.

cfgmgr/teammgr.cpp
```diff
@@ -395,6 +395,21 @@ bool TeamMgr::setLagMtu(const string &alias, const string &mtu)
     return true;
 }
 
+bool TeamMgr::setLacpKey(const string& alias, const string& lacp_key)
```
judyjoseph, line 398: This new function is not called anywhere ?

cfgmgr/teammgr.cpp
```diff
@@ -520,11 +581,17 @@ task_process_status TeamMgr::addLagMember(const string &lag, const string &membe
...
     stringstream cmd;
     string res;
+    uint16_t keyId = generateLacpKey(lag);
 
     // Set admin down LAG member (required by teamd) and enslave it
     // ip link set dev <member> down;
+    // teamdctl <port_channel_name> port config update <member> { lacp_key: <lacp_key>, link_watch: { name: ethtool } };
     // teamdctl <port_channel_name> port add <member>;
     cmd << IP_CMD << " link set dev " << shellquote(member) << " down; ";
+    cmd << TEAMDCTL_CMD << " " << shellquote(lag) << " port config update " << shellquote(member)
```
judyjoseph, line 591: Is teamd ok ( no errors seen here ) as we are doing a "port config update" before doing actual member add below using "port add". Should we do this call after "port add" ?

cfgmgr/teammgr.cpp
DavidZagury: Yes, we kept this behavior to not break warm updates from previous versions, on warm-update the PortChannel will need to keep the LACP key as 0 to avoid link drop

cfgmgr/teammgr.cpp
DavidZagury: Fixed

cfgmgr/teammgr.cpp
DavidZagury: No, there are no errors printed from teamd when we are running the commands in this error.

Trying to run the port add before the config does results in errors:
> root@arc-switch1038:/home/admin# ip link set dev Ethernet64 down
> root@arc-switch1038:/home/admin# teamdctl PortChannel0005 port add Ethernet64
> root@arc-switch1038:/home/admin# teamdctl PortChannel0005 port config update Ethernet64 '{ "lacp_key": 10005, "link_watch": { "name": "ethtool" } }'
> libteamdctl: cli_usock_process_msg: usock: Error message received: "ConfigUpdateFail"
> libteamdctl: cli_usock_process_msg: usock: Error message content: "Failed to update config."
> command call failed (Invalid argument)

tests/test_portchannel.py
DavidZagury: I extended the tests to cover all the lacp_key cases, thank you.

liat-grozovik:
@ judyjoseph any comments on this PR? can you please explain why the recent mentioned PR is related and how you would like to move forward with this one?

tests/test_portchannel.py
```diff
@@ -89,6 +90,69 @@ def test_Portchannel(self, dvs, testlog):
...
+
+        for portchannel in portchannelNamesAuto:
+            tbl.set(portchannel[0], fvs)
+            time.sleep(1)
+
+        fvs_no_lacp_key = swsscommon.FieldValuePairs(
+            [("admin_status", "up"), ("mtu", "9100"), ("oper_status", "up")])
+        tbl.set(portchannelNames[0][0], fvs_no_lacp_key)
+        time.sleep(1)
```
lguohan, line 117: why do you have all these sleeps?

cfgmgr/teammgr.cpp
```diff
@@ -504,6 +504,52 @@ bool TeamMgr::removeLag(const string &alias)
     return true;
 }
 
+// Port-channel names are in the pattern of "PortChannel####"
+// 
+// The LACP key could be generated in 3 ways based on the value in config DB:
+//      1. "auto" - LACP key is extracted from the port-channel name and is set to be the number at the end of the port-channel name
```
lguohan, line 510: since you added new config db knob, please add this to the sonic yang model. https://github.com/Azure/sonic-buildimage/tree/master/src/sonic-yang-models/yang-models

lguohan:
i have concerns on the sleeps in the test, code looks good to me. please add yang model as well in sonic-buildimage repo.

tests/test_portchannel.py
DavidZagury: I had them just because I based my test on the second test in this file that had them, but I removed them and it works

judyjoseph (approved):
LGTM, will wait for all reviewers.

liat-grozovik:
@lguohan any further feedback?

lguohan:
this looks like a new feature, not sure why needs backport.

PR #1665, opened 2021-03-08
Add failure notification for orchagent

shi-su:
**What I did**
Add SAI failure handling functions in aclorch, bufferorch, copporch, dtelorch, fdborch, fgnhgorch, intfsorch, mirrororch, natorch, policerorch, macsecorch, portsorch, qosorch, sfloworch, switchorch, tunneldecaporch, vrforch.

**Why I did it**
Failure notification mechanism to ensure enough notifications in the presence of SAI failures and avoid running switches with unhandled failures.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -3626,6 +3626,7 @@ sai_status_t AclOrch::createDTelWatchListTables()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to create table %s", flowWLTable.description.c_str());
+        handleSaiCreateStatus(SAI_API_ACL, status);
         return status;
     }
 
@@ -3687,6 +3688,7 @@ sai_status_t AclOrch::createDTelWatchListTables()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to create table %s", dropWLTable.description.c_str());
+        handleSaiCreateStatus(SAI_API_ACL, status);
         return status;
     }
 
@@ -3719,6 +3721,7 @@ sai_status_t AclOrch::deleteDTelWatchListTables()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to delete table %s", table_id.c_str());
+        handleSaiRemoveStatus(SAI_API_ACL, status);
         return status;
     }
 
@@ -3739,6 +3742,7 @@ sai_status_t AclOrch::deleteDTelWatchListTables()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to delete table %s", table_id.c_str());
+        handleSaiRemoveStatus(SAI_API_ACL, status);
         return status;
     }
 
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -3626,8 +3626,10 @@ sai_status_t AclOrch::createDTelWatchListTables()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to create table %s", flowWLTable.description.c_str());
-        handleSaiCreateStatus(SAI_API_ACL, status);
-        return status;
+        if (handleSaiCreateStatus(SAI_API_ACL, status) != task_success)
+        {
+            return status;
+        }
     }
 
     gCrmOrch->incCrmAclUsedCounter(CrmResourceType::CRM_ACL_TABLE, SAI_ACL_STAGE_INGRESS, SAI_ACL_BIND_POINT_TYPE_SWITCH);
@@ -3688,15 +3690,17 @@ sai_status_t AclOrch::createDTelWatchListTables()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to create table %s", dropWLTable.description.c_str());
-        handleSaiCreateStatus(SAI_API_ACL, status);
-        return status;
+        if (handleSaiCreateStatus(SAI_API_ACL, status) != task_success)
+        {
+            return status;
+        }
     }
 
     gCrmOrch->incCrmAclUsedCounter(CrmResourceType::CRM_ACL_TABLE, SAI_ACL_STAGE_INGRESS, SAI_ACL_BIND_POINT_TYPE_SWITCH);
     m_AclTables[table_oid] = dropWLTable;
     SWSS_LOG_INFO("Successfully created ACL table %s, oid: %" PRIx64, dropWLTable.description.c_str(), table_oid);
 
-    return status;
+    return SAI_STATUS_SUCCESS;
 }
 
 sai_status_t AclOrch::deleteDTelWatchListTables()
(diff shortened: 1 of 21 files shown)


orchagent/aclorch.cpp
```diff
@@ -3596,6 +3596,7 @@ sai_status_t AclOrch::createDTelWatchListTables()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to create table %s", flowWLTable.description.c_str());
+        handleSaiCreateStatus(SAI_API_ACL, status);
```
qiluo-msft, line 3599: >handleSaiCreateStatus [](start = 8, length = 21)

Check the return value? If successfully handled, you treat it as success?

orchagent/copporch.cpp
```diff
@@ -125,7 +125,7 @@ void CoppOrch::initDefaultHostIntfTable()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to create default host interface table, rv:%d", status);
-        throw "CoppOrch initialization failure";
+        handleSaiCreateStatus(SAI_API_HOSTIF, status);
```
qiluo-msft, line 128: >handleSaiCreateStatus [](start = 8, length = 21)

Do you want to keep old behavior even with your new handle function?

orchagent/copporch.cpp
```diff
@@ -489,6 +489,7 @@ task_process_status CoppOrch::processCoppRule(Consumer& consumer)
                 if (sai_status != SAI_STATUS_SUCCESS)
                 {
                     SWSS_LOG_ERROR("Failed to apply attribute:%d to trap group:%" PRIx64 ", name:%s, error:%d\n", trap_gr_attr.id, m_trap_group_map[trap_group_name], trap_group_name.c_str(), sai_status);
+                    handleSaiSetStatus(SAI_API_HOSTIF, sai_status);
```
qiluo-msft, line 492: >handleSaiSetStatus [](start = 20, length = 18)

Check the return value?

orchagent/dtelorch.cpp
```diff
@@ -506,7 +511,15 @@ void DTelOrch::doDtelTableTask(Consumer &consumer)
...
                 {
                     SWSS_LOG_ERROR("DTEL ERROR: Failed to enable INT endpoint mode");
-                    goto dtel_table_continue;
+                    if (handleSaiSetStatus(SAI_API_DTEL, status))
+                    {
+                        goto dtel_table_continue;
+                    }
+                    else
+                    {
+                        it++;
```
qiluo-msft, line 520: >it++; [](start = 24, length = 5)

I think you change the behavior. Old code will always `erase(it)` no matter success or not.

orchagent/fdborch.cpp
```diff
@@ -1097,7 +1098,7 @@ bool FdbOrch::addFdbEntry(const FdbEntry& entry, const string& port_name,
             SWSS_LOG_ERROR("Failed to create %s FDB %s in %s on %s, rv:%d",
                     fdbData.type.c_str(), entry.mac.to_string().c_str(),
                     vlan.m_alias.c_str(), port_name.c_str(), status);
-            return false; //FIXME: it should be based on status. Some could be retried, some not
```
qiluo-msft, line 1100: >FIXME [](start = 28, length = 5)

Did you truly fixed the `FIXME`?

orchagent/fdborch.cpp
```diff
@@ -1212,7 +1213,7 @@ bool FdbOrch::removeFdbEntry(const FdbEntry& entry, FdbOrigin origin)
     {
         SWSS_LOG_ERROR("FdbOrch RemoveFDBEntry: Failed to remove FDB entry. mac=%s, bv_id=0x%" PRIx64,
                        entry.mac.to_string().c_str(), entry.bv_id);
-        return true; //FIXME: it should be based on status. Some could be retried. some not
```
qiluo-msft, line 1215: >FIXME [](start = 23, length = 5)

The same.

orchagent/intfsorch.cpp
```diff
@@ -1056,7 +1061,7 @@ bool IntfsOrch::addRouterIntfs(sai_object_id_t vrf_id, Port &port)
     {
         SWSS_LOG_ERROR("Failed to create router interface %s, rv:%d",
                 port.m_alias.c_str(), status);
-        throw runtime_error("Failed to create router interface.");
+        return handleSaiCreateStatus(SAI_API_ROUTER_INTERFACE, status);
```
qiluo-msft, line 1064: >handleSaiCreateStatus [](start = 15, length = 21)

To exactly mimic old behavior, you can derive the handleSaiCreateStatus function in this class, and `throw` inside the function.

orchagent/aclorch.cpp
shi-su: There are a few SAI calls that do not take any action in the case of SAI failures. For these ones, since our goal is to make it on-par with async mode at this stage, I am not checking the return value for now so that it has the same behavior as before.

orchagent/copporch.cpp
shi-su: I think in the old behavior (async mode), this throw should not even be called since syncd should already trigger an swss restart before it reaches this throw. To mimic the old behavior close enough, I think a method could be running this handling function before this throw and keep this throw regardless of the return value of the handling function. I updated the code as such.

orchagent/copporch.cpp
shi-su: There are a few SAI calls that do not take any action in the case of SAI failures. For these ones, since our goal is to make it on-par with async mode at this stage, I am not checking the return value for now so that it has the same behavior as before.

orchagent/dtelorch.cpp
shi-su: Currently, the handling function always exits orchagent to achieve on-par handling with async mode. So this change will not be activated at the moment and the behavior should be the same as async mode. Once we have a better logic for the failure handling functions, we can fully control whether it should retry or not.

orchagent/fdborch.cpp
shi-su: I did not, added the FIXME back.

orchagent/fdborch.cpp
shi-su: Added the FIXME back.

orchagent/intfsorch.cpp
shi-su: I think in the old behavior (async mode), this throw should not even be called since syncd should already trigger an swss restart before it reaches this throw. To mimic the old behavior close enough, I think a method could be running this handling function before this throw and keep this throw regardless of the return value of the handling function. I updated the code as such.

orchagent/aclorch.cpp
qiluo-msft: Let's assume this function magically retry and fix the issue, so the return value is `task_success`. You should return `SAI_STATUS_SUCCESS` after this function.

---
In reply to: [595609272](https://github.com/Azure/sonic-swss/pull/1665#discussion_r595609272) [](ancestors = 595609272)

orchagent/copporch.cpp
qiluo-msft: You also need to check return value and conditional `throw`.

---
In reply to: [595610233](https://github.com/Azure/sonic-swss/pull/1665#discussion_r595610233) [](ancestors = 595610233)

orchagent/aclorch.cpp
shi-su: Agree, updated the logic to check if the return value is `task_success`, and proceed when it is.

orchagent/copporch.cpp
shi-su: Updated the logic to check return value before throwing.

orchagent/copporch.cpp
shi-su: Added a check for the return value of the handling function and return only when it is not `task_success`

orchagent/intfsorch.cpp
shi-su: Added a check for the return value of the handling function and throw exception only when it did not successfully handle the failure.

qiluo-msft (approved):

PR #1666, opened 2021-03-09
Deactivate mirror session only when status is true in updateLagMember

shi-su:
**What I did**
Deactivate mirror session only when status is true in updateLagMember

**Why I did it**
Fix the issue when mirrororch tries to deactivate a session when it is not activated
```
Mar  7 18:22:03.125662 e07265d2546e NOTICE #orchagent: :- removeLagMember: Remove member Ethernet88 from LAG PortChannel008 lid:20000000005ff lmid:1b000000000600
Mar  7 18:22:03.125691 e07265d2546e ERR #orchagent: :- meta_sai_validate_oid: object key SAI_OBJECT_TYPE_MIRROR_SESSION:oid:0xe000000000603 doesn't exist
Mar  7 18:22:03.125695 e07265d2546e ERR #orchagent: :- deactivateSession: Failed to deactivate mirroring session TEST_SESSION
Mar  7 18:22:04.126757 e07265d2546e NOTICE #orchagent: :- removeLag: Remove LAG PortChannel008 lid:20000000005ff
```

**How I verified it**
Verified that the above error message does not show up in dvs test

**Details if related**

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -1342,7 +1342,10 @@ void MirrorOrch::updateLagMember(const LagMemberUpdate& update)
             // If LAG is empty, deactivate session
             if (update.lag.m_members.empty())
             {
-                deactivateSession(name, session);
+                if (session.status)
+                {
+                    deactivateSession(name, session);
+                }
                 session.neighborInfo.portId = SAI_OBJECT_TYPE_NULL;
             }
             // Switch to a new member of the LAG


orchagent/mirrororch.cpp
```diff
@@ -1342,7 +1342,10 @@ void MirrorOrch::updateLagMember(const LagMemberUpdate& update)
             // If LAG is empty, deactivate session
             if (update.lag.m_members.empty())
             {
-                deactivateSession(name, session);
```
qiluo-msft, line 1345: Is it requested by end user? If yes, the error message is expected.

shi-su: It seems that the session got deactivated twice, and we should avoid deactivating the same session once it is already deactivated. I think checking the session status should be a good way to do the job. The function `deactivateSession()` even has an assertion that session status is true at the beginning of the function.

lguohan (approved):

lguohan:
is there an issue associated with this, if yes, can you link the issue to this pr?

shi-su:
> is there an issue associated with this, if yes, can you link the issue to this pr?

I did not create an issue for this one. Noticed it when adding on-par SAI failure handling and created this PR for fix.

PR #1667, opened 2021-03-09
[201911][acl] Handle uncaught exception in aclorch

daall:

**What I did**
I added a `catch` statement to prevent orchagent from crashing if an invalid field combination is provided.

**Why I did it**
https://github.com/Azure/sonic-swss/pull/1486 added a `catch` to prevent orchagent from crashing if an invalid field combination is provided, however this was not brought into 201911.

**How I verified it**
Pass in something illegal like a rule for a `MIRROR` table with a `PACKET_ACTION` and verify that orchagent reports an error to the syslog without crashing.

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -2903,8 +2903,16 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
                 type = table_id == m_mirrorTableId[stage] ? ACL_TABLE_MIRROR : ACL_TABLE_MIRRORV6;
             }
 
-
-            newRule = AclRule::makeShared(type, this, m_mirrorOrch, m_dTelOrch, rule_id, table_id, t);
+            try
+            {
+                newRule = AclRule::makeShared(type, this, m_mirrorOrch, m_dTelOrch, rule_id, table_id, t);
+            }
+            catch (exception &e)
+            {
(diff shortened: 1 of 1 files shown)


orchagent/aclorch.cpp
```diff
@@ -2903,8 +2903,16 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
                 type = table_id == m_mirrorTableId[stage] ? ACL_TABLE_MIRROR : ACL_TABLE_MIRRORV6;
             }
 
-
-            newRule = AclRule::makeShared(type, this, m_mirrorOrch, m_dTelOrch, rule_id, table_id, t);
+            try
+            {
+                newRule = AclRule::makeShared(type, this, m_mirrorOrch, m_dTelOrch, rule_id, table_id, t);
+            }
+            catch (exception &e)
```
qiluo-msft, line 2910: >exception [](start = 19, length = 9)

`exception` type is too generic and possibly hide code bug. Could you list all possible excpetion types?
This comment is also applicable to master branch. So if you found better solution, let's start a PR on master first.

qiluo-msft, line 2910: >catch [](start = 12, length = 5)

The communication between orchagent and syncd on 201911 is async mode. Is it related to a syncd/SAI error code? In that case, will you catch anything?

PR #1669, opened 2021-03-10
reduce severity of log to info in case of flush on non-existing member

allas-nvidia:
Signed-off-by: allas <allas@nvidia.com>

**What I did**
Reduced severity of log to info in case of flush on non-existing member.

**Why I did it**
There is a race in the infra which can cause the scenario: FDB FLUSH notification is received on port which was already removed.

**How I verified it**
    >  sudo config vlan add 3
    >  sudo config vlan member add 3 Ethernet8
    >  sudo config vlan member del 3 Ethernet8

**Details if related**

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -168,11 +168,24 @@ void FdbOrch::update(sai_fdb_event_t        type,
                    type, update.entry.mac.to_string().c_str(),
                    entry->bv_id, bridge_port_id);
 
+
     if (bridge_port_id &&
         !m_portsOrch->getPortByBridgePortId(bridge_port_id, update.port))
     {
-        SWSS_LOG_ERROR("Failed to get port by bridge port ID 0x%" PRIx64 ".",
+        if (type == SAI_FDB_EVENT_FLUSHED)
+        {
+            /* In case of flush - can be ignored due to a race.
+               There are notifications about FDB FLUSH (syncd/sai_redis) on port,
+               which was already removed by orchagent as a result of
+               removeVlanMember action (removeBridgePort) */
+            SWSS_LOG_INFO("Flush event: Failed to get port by bridge port ID 0x%" PRIx64 ".",
                         bridge_port_id);
+
+        } else {
+            SWSS_LOG_ERROR("Failed to get port by bridge port ID 0x%" PRIx64 ".",
+                        bridge_port_id);
+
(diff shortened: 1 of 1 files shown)


orchagent/fdborch.cpp
```diff
@@ -168,11 +168,23 @@ void FdbOrch::update(sai_fdb_event_t        type,
...
 
+
     if (bridge_port_id &&
         !m_portsOrch->getPortByBridgePortId(bridge_port_id, update.port))
     {
-        SWSS_LOG_ERROR("Failed to get port by bridge port ID 0x%" PRIx64 ".",
+        if (type == SAI_FDB_EVENT_FLUSHED)
+        {
+            /* in case of flush - can be ignored due to a race */
+            /* there are notifications about FDB FLUSH (syncd/sai_redis) on port which was already removed
```
prsunny, line 178: Please combine the comments from previous line and have as a single block of comment.

orchagent/fdborch.cpp
```diff
@@ -168,11 +168,23 @@ void FdbOrch::update(sai_fdb_event_t        type,
...
+             * by orchagent as a result of removeVlanMember action (removeBridgePort) */
+            SWSS_LOG_INFO("Flush ignored. Received event on bridge port ID 0x%" PRIx64 " which is not exist.",
```
prsunny, line 180: Can you please rephrase to "Flush event: Failed to get port by bridge port ID 0x%" PRIx64 "."

prsunny:
Also can you please respond to https://github.com/Azure/sonic-buildimage/issues/6936, before the PR merge?

orchagent/fdborch.cpp
allas-nvidia: I can :) but, IMO its better to avoid to use word "Failed"in INFO notification.
What do you think?

prsunny: Agree, but this has been the model for some exceptions.. For e.g check [this](https://github.com/Azure/sonic-swss/blob/master/orchagent/fdborch.cpp#L832).  IMO, even though it is INFO, it is still a failure in fetch the data.

liat-grozovik:
@allas-nvidia  please align the code based on the above comments.

liat-grozovik (approved):

liat-grozovik (approved):

PR #1670, opened 2021-03-12
[MUX/PFCWD] Use in_ports for acls instead of seperate ACL table

prsunny:
**What I did**

1. Use ACL_TABLE_DROP for both PFC_WD and MUX
2. Use MATCH_IN_PORTS instead of binding port to ACL table and program single ACL rule
3. Updated PFC_WD code for generic handling and use class data within member function. 
3. Update ACL rule by modifying IN_PORTS bitmap during mux transitions
4. VS test to cover the new changes

**Why I did it**
Optimize ACL TCAM use

**How I verified it**
Run test on DUT and validated the PBMP

**Details if related**

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -40,8 +40,8 @@ extern sai_router_interface_api_t* sai_router_intfs_api;
 
 /* Constants */
 #define MUX_TUNNEL "MuxTunnel0"
-#define MUX_ACL_TABLE_NAME "mux_acl_table";
-#define MUX_ACL_RULE_NAME "mux_acl_rule";
+#define MUX_ACL_TABLE_NAME INGRESS_TABLE_DROP
+#define MUX_ACL_RULE_NAME "mux_acl_rule"
 #define MUX_HW_STATE_UNKNOWN "unknown"
 #define MUX_HW_STATE_ERROR "error"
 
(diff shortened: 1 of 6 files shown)


orchagent/muxorch.h
```diff
@@ -101,7 +102,7 @@ class MuxCable
     bool stateInitActive();
     bool stateStandby();
 
-    bool aclHandler(sai_object_id_t, bool add = true);
+    bool aclHandler(sai_object_id_t, string, bool add = true);
```
daall, line 105: can we give the rest of the parameters names?

orchagent/muxorch.cpp
```diff
@@ -709,20 +715,44 @@ MuxAclHandler::MuxAclHandler(sai_object_id_t port)
...
 
     SWSS_LOG_NOTICE("Un-Binding port %" PRIx64 "", port_);
 
-    auto found = acl_table_.find(table_name);
-    found->second.unbind(port_);
+    AclRule* rule = gAclOrch->getAclRule(table_name, rule_name);
+    if (rule == nullptr)
+    {
+        SWSS_LOG_THROW("ACL Rule does not exist for port %s, rule %s", alias_.c_str(), rule_name.c_str());
+    }
```
daall, line 745: Are we guaranteed there will always be a rule here? If there's no PFC storm and no switchover then won't we throw an unnecessary exception?

prsunny: Yes it is guaranteed; In the constructor, we always create an ACL rule and in this destructor, it would be an exception if we don't find the rule.  Same is done for pfwdacl as well, @vmittal-msft,

vmittal-msft (approved):

orchagent/muxorch.cpp
daall: AH this is the handler not the orch, I understand now. Thanks!

daall (approved):

PR #1671, opened 2021-03-12
[acl] Move ACL table constants to acltable.h

daall:

**What I did**
I moved all of the ACL table related constants over to acltable.h.

**Why I did it**
aclorch.h and aclorch.cpp are both huge files that have become hard to maintain/update to support new features. This is part of a bigger effort to try to refactor it to be more manageable.

**How I verified it**
Confirmed that the compiler doesn't throw any new errors, unit tests still work the same as before, and I ran `test_acl` and `test_everflow` from sonic-mgmt to confirm there is no traffic impact.

**Details if related**


prsunny (approved):
lgtm

PR #1672, opened 2021-03-16
[nbrmgrd] added function to parse IP address from APP_DB

vadymhlushko-mlnx:
Signed-off-by: Vadym Hlushko <vadymh@nvidia.com>

**What I did**
Added specific function to parse keys from `APP_DB` which looks like:
`"NEIGH_RESOLVE_TABLE:Vlan2:2000:1::1"`
`"NEIGH_RESOLVE_TABLE:Vlan2:192.168.0.10"`

**Why I did it**
The fix for the [issues/7045](https://github.com/Azure/sonic-buildimage/issues/7045).

**How I verified it**
Provided the VS tests.
Also, it was verified manually, error messages and crash from the [issues/7045](https://github.com/Azure/sonic-buildimage/issues/7045) wasn't observed.

### DIFF
--- a/tests/test_neighbor.py
+++ b/tests/test_neighbor.py
@@ -391,6 +391,28 @@ class TestNeighbor(object):
             dec_neigh_entries_cnt = (old_neigh_entries_cnt - current_neigh_entries_cnt)
             assert dec_neigh_entries_cnt == 1
 
+    def test_FlushResolveNeighborIpv6(self, dvs, testlog):
+        appl_db = swsscommon.DBConnector(swsscommon.APPL_DB, dvs.redis_sock, 0)
+        prod_state_tbl = swsscommon.ProducerStateTable(appl_db, swsscommon.APP_NEIGH_RESOLVE_TABLE_NAME)
+        fvs = swsscommon.FieldValuePairs([("mac", "52:54:00:25:06:E9")])
+
+        prod_state_tbl.set("Vlan2:2000:1::1", fvs)
+        time.sleep(2)
+
+        (exitcode, output) = dvs.runcmd(['sh', '-c', "supervisorctl status nbrmgrd | awk '{print $2}'"])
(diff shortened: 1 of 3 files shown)


prsunny (commented):
changes lgtm.. as we discussed, we can add a simple VS test to add to APP_DB and verify. Recommend to have the test covered as part of the same PR

tests/test_neighbor.py
```diff
@@ -389,10 +389,32 @@ def test_NeighborAddRemoveIpv4WithVrf(self, dvs, testlog):
...
+    def test_FlushResolveNeighborIpv6(self, dvs, testlog):
+        appl_db = swsscommon.DBConnector(swsscommon.APPL_DB, dvs.redis_sock, 0)
+        prod_state_tbl = swsscommon.ProducerStateTable(appl_db, swsscommon.APP_NEIGH_RESOLVE_TABLE_NAME)
+        fvs = swsscommon.FieldValuePairs([("mac", "52:54:00:25:06:E9")])
+
+        prod_state_tbl.set("Vlan2:2000:1::1", fvs)
+        time.sleep(2)
+
+        (exitcode, output) = dvs.runcmd(['sh', '-c', "supervisorctl status nbrmgrd | awk '{print $2}'"])
+        assert output == "RUNNING\n"
```
prsunny, line 403: Is it ok if you can also check whether the req has been sent out? Like parsing `ip neigh show` for 2000:1::1. This way we can verify the functionality. It might be in 'failed' state which is ok.

vadymhlushko-mlnx: The `ip neigh show` command produced the empty output, it seems like because in `test_FlushResolveNeighborIpv6` we don't test the whole flow of flushing the neighbor entry, because it is impossible to do on the VS.

prsunny (approved):

PR #1673, opened 2021-03-18
Increase tunnel nexthop reference count when nh_tunnels create new entry

iris00522:
**What I did**
I found  nh_tunnels ref_count without increasing when entry create, 
the nh_tunnels ref_count may decrease to negative number when remove nexthop entry.

Increase tunnel nexthop reference count when nh_tunnels create new entry.

**Why I did it**
the nh_tunnels ref_count may decrease to negative number when remove nexthop entry.

**How I verified it**
test on broadcom dut.

### DIFF
--- a/orchagent/vxlanorch.cpp
+++ b/orchagent/vxlanorch.cpp
@@ -574,6 +574,7 @@ void VxlanTunnel::updateNextHop(IpAddress& ipAddr, MacAddress macAddress,
     if (it == nh_tunnels_.end())
     {
         nh_tunnels_[key] = {nh_id, 1};
+        nh_tunnels_[key].ref_count++;
         return;
     } 
     else 


orchagent/vxlanorch.cpp
```diff
@@ -574,6 +574,7 @@ void VxlanTunnel::updateNextHop(IpAddress& ipAddr, MacAddress macAddress,
     if (it == nh_tunnels_.end())
     {
         nh_tunnels_[key] = {nh_id, 1};
+        nh_tunnels_[key].ref_count++;
```
prsunny, line 577: This section gets executed only if `key `added to `nh_tunnels` for the first time. So how will this get incremented in subsequent updates? May be i'm missing something.

iris00522: when call createNextHopTunnel will increase ref_count every times except create nh_tunnels first time.

PR #1674, opened 2021-03-19
[MuxOrch] FDB ageout safety check

prsunny:
**What I did**
Added safety check to handle one scenario of fdb ageout and mux state change. 
If no fdb exists, return the state of current mux port associated. 

**Why I did it**

**How I verified it**

**Details if related**


lguohan:
vs test?

gechiang (approved):

tahmed-dev (approved):

PR #1607, opened 2021-01-25
[buffermgr] Support maximum port headroom checking

stephenxs:
**What I did**
Check the accumulative port headroom against the port's maximum headroom

On Mellanox platform, this PR depends on PR [#6566](https://github.com/Azure/sonic-buildimage/pull/6566) to be merged. In that PR the required SAI attribute is supported.
On other platforms, there is no dependency.

Signed-off-by: Stephen Sun <stephens@nvidia.com>

**Why I did it**
On some platforms, the SAI will notify orchagent shut down if a headroom size that causes the accumulative port headroom to exceed its limit is programmed to SAI.
To avoid that, we need to check this before programming it to SAI.

**How I verified it**
Run the regression test.

**Details if related**
- Fetch the maximum port headroom via fetching the port attribute `SAI_PORT_ATTR_QOS_MAXIMUM_HEADROOM_SIZE` when the orchagent starts and push the data into STATE_DB
- Check the accumulative port headroom against the maximum headroom in buffer_check_headroom_<vendor>.lua

**- Which release branch to backport (provide reason below if selected)**

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

PR #1675, opened 2021-03-19
Revert "[buffermgr] Support maximum port headroom checking"

prsunny:
Reverts Azure/sonic-swss#1607

Suspecting this for a test failure. Reverting to verify. 

Refer: https://github.com/Azure/sonic-buildimage/pull/6989. Failing when the PR is updated to this comment.

### DIFF
--- a/cfgmgr/buffer_check_headroom_mellanox.lua
+++ b/cfgmgr/buffer_check_headroom_mellanox.lua
@@ -10,15 +10,36 @@ local new_pg = ARGV[3]
 local accumulative_size = 0
 
 local appl_db = "0"
+local config_db = "4"
 local state_db = "6"
 
 local ret_true = {}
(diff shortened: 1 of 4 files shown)


lguohan (approved):

lguohan:
@stephenxs for awareness

stephenxs:
From the log I see there are messages which indicate SAI returned SAI_STATUS_NOT_IMPLEMENTED the call SAI_PORT_ATTR_QOS_MAXIMUM_HEADROOM_SIZE=20 for all ports and after that SAI continued to response other calls, which means the SAI didn’t stop immediately after this call. 
As this is used for dynamic buffer calculation only and that feature doesn’t run on vs image, I’m not sure that’s the reason why SAI doesn’t update the routing entries.

PR #1677, opened 2021-03-20
[Dynamic buffer calc] Bug fix: Remove PGs from an administratively down port.

stephenxs:
**What I did**
Bug fixes: Remove PGs from an administratively down port.

**Why I did it**
To fix bugs

**How I verified it**
Run regression and vs test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012

**Details if related**
- Remove PGs from an administratively down port.
  - Introduce a new state: `PORT_ADMIN_DOWN` which represents the port is administratively down.
  - Remove all PGs when the port is shut down and re-add all configured PGs when port is started up
  - Only record the new value but don't touch `BUFFER_PG_TABLE` if the following events come when a port is administratively down
    - a port's MTU, speed, or cable length is updated
    - a new PG is added to a port or an existing PG is removed from a port
  - Optimize the port event handling flow since `refreshPriorityGroupsForPort` should be called only once in case more than one fields are updated
  - Adjust the Lua plugin which calculates the buffer pool size accordingly

### DIFF
--- a/cfgmgr/buffer_pool_mellanox.lua
+++ b/cfgmgr/buffer_pool_mellanox.lua
@@ -12,7 +12,7 @@ local lossypg_400g = 0
 local result = {}
 local profiles = {}
 
-local count_up_port = 0
+local total_port = 0
 
 local mgmt_pool_size = 256 * 1024
 local egress_mirror_headroom = 10 * 1024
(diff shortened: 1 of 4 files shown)


neethajohn:
@stephenxs , can you move out the lua plugin optimizations out of this PR and keep this PR  only for the port down bug fix?

stephenxs:
> @stephenxs , can you move out the lua plugin optimizations out of this PR and keep this PR only for the port down bug fix?

Hi @neethajohn 
Actually the lua plugin is also a part of the fix. Probably I need to adjust the wording for optimize to adjust?

neethajohn (approved):

PR #1678, opened 2021-03-20
[Dynamic Buffer Calc] Bug fix: A suffix keeps being added if a non-exist buffer pool is referenced.

stephenxs:
**What I did**
Bug:
The buffermgrd can keep adding suffix to the buffer pool reference if the buffer pool isn't found when it is being referenced. In most of the cases, it's caused by wrong configuration.
Cause:
In handleBufferProfileTable, the value of each field is designated by a (lvalue) reference to the object in the tuple, which means the object in the tuple will be changed if the value is changed in the function.
Fix:
Pass the value of each field by value instead of reference.

**Why I did it**
Fix bug.

**How I verified it**
Manually test.

**Details if related**

### DIFF
--- a/cfgmgr/buffermgrdyn.cpp
+++ b/cfgmgr/buffermgrdyn.cpp
@@ -1472,8 +1472,8 @@ task_process_status BufferMgrDynamic::handleBufferProfileTable(KeyOpFieldsValues
         }
         for (auto i = kfvFieldsValues(tuple).begin(); i != kfvFieldsValues(tuple).end(); i++)
         {
-            string &field = fvField(*i);
-            string &value = fvValue(*i);
+            const string &field = fvField(*i);
+            string value = fvValue(*i);
 
             SWSS_LOG_DEBUG("field:%s, value:%s", field.c_str(), value.c_str());
             if (field == buffer_pool_field_name)


neethajohn:
@stephenxs, does any of the existing vs tests cover this scenario?

stephenxs:
> @stephenxs, does any of the existing vs tests cover this scenario?

For the 

> @stephenxs, does any of the existing vs tests cover this scenario?

We do have test to cover the case for adding correct Profiles but we don’t have one for adding wrong profiles. Typically we do vstest by checking the dbs. But for this bug there is not anything wrong in db and only an message with an incorrect pool reference is observed. So I think it doesn’t make sense to add the case for it.

neethajohn:
> > @stephenxs, does any of the existing vs tests cover this scenario?
> 
> For the
> 
> > @stephenxs, does any of the existing vs tests cover this scenario?
> 
> We do have test to cover the case for adding correct Profiles but we don’t have one for adding wrong profiles. Typically we do vstest by checking the dbs. But for this bug there is not anything wrong in db and only an message with an incorrect pool reference is observed. So I think it doesn’t make sense to add the case for it.

As discussed offline, since this error is seen by buffermgr, we cannot add a test for this

neethajohn (approved):

stephenxs:
azp /run

stephenxs:
The vs test failed by `test_TunnelDecap_v6`, `test_TunnelSymmetric_v4`, and `test_TunnelSymmetric_v6` which are not related to dynamic buffer calculation.

liat-grozovik:
@lguohan is there anyone working on these failed tests?

dprital:
@neethajohn  - Can you please merge it ?

PR #1680, opened 2021-03-20
[Dynamic Buffer Calc] Enhance the field checking in table handling

stephenxs:
**What I did**
Enhance the field checking in table handling
- Fix a typo in vs test.
- Add field name-checking in `BUFFER_PG` handling.
- Use `else if` instead of `if` if the `field` matches one field of the table because in that case, it won't match others

Signed-off-by: Stephen Sun stephens@nvidia.com

**Why I did it**

**How I verified it**
Run vs test and regression test.

**Details if related**


stephenxs:
Originally it was a typo in vstest. During fixing the typo I realized it's lacking the field name checking. So I add them as well.

neethajohn (approved):

PR #1690, opened 2021-04-03
[ci]: archive swss pytests

lguohan:

**What I did**
archive swss pytests to be used for other pipeline vstests

**Why I did it**
in other pipelines, they will download swss and pytests. this is make sure the swss.deb and pytests are matched.

**How I verified it**

**Details if related**

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -30,6 +30,10 @@ parameters:
 - name: artifact_name
   type: string
 
+- name: archive_pytests
+  type: boolean
+  default: false
+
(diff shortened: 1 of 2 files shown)


yxieca:
@lguohan I think you need to create the change for 202012 branch directly. This change cannot be cherry-picked cleanly.

xumia:
@lguohan , can we simple copy all the yaml files to the branch 202012?
I want to enable the PR checker for the branch 202012 of sonic-utilities, it has dependency on this PR.
See https://github.com/Azure/sonic-utilities/pull/1649.

Issue #1687, opened 2021-04-02
[pytest][azp] test_chassis_system_lag_id_allocator_del_id is failing in PR commits


smaheshm:
This test is flaky and has been failing fairly consistently. 

```

2021-04-02T01:34:40.6713838Z test_virtual_chassis.py:461: AssertionError
2021-04-02T01:34:40.6714224Z ________ TestVirtualChassis.test_chassis_system_lag_id_allocator_del_id ________
2021-04-02T01:34:40.6714450Z 
2021-04-02T01:34:40.6714741Z self = <test_virtual_chassis.TestVirtualChassis object at 0x7f5ebd1ce4f0>
2021-04-02T01:34:40.6715146Z vct = <conftest.DockerVirtualChassisTopology object at 0x7f5ebd0cfd60>
2021-04-02T01:34:40.6715340Z 
2021-04-02T01:34:40.6715647Z     def test_chassis_system_lag_id_allocator_del_id(self, vct):
2021-04-02T01:34:40.6716254Z         """Test lag id allocator's release id and re-use id processing.
2021-04-02T01:34:40.6716549Z     
2021-04-02T01:34:40.6716926Z         Pre-requisite:
2021-04-02T01:34:40.6717247Z             (i)  Test case: test_chassis_system_lag
2021-04-02T01:34:40.6717660Z             (ii) Test case: test_chassis_system_lag_id_allocator_table_full
2021-04-02T01:34:40.6718012Z         This test validates that
2021-04-02T01:34:40.6718373Z             (i)   Portchannel is deleted and id allocator does not return error
2021-04-02T01:34:40.6719051Z             (ii)  Should be able to add PortChannel to re-use released id
2021-04-02T01:34:40.6719667Z             (iii) Deleted portchaneels are removed from chassis app db
2021-04-02T01:34:40.6720127Z             (iv)  Remote asics remove the system lag corresponding to the deleted PortChannels
2021-04-02T01:34:40.6720508Z         """
2021-04-02T01:34:40.6720728Z     
2021-04-02T01:34:40.6720996Z         if vct is None:
2021-04-02T01:34:40.6721271Z             return
2021-04-02T01:34:40.6721516Z     
2021-04-02T01:34:40.6721784Z         test_lag1_name = "PortChannel0001"
2021-04-02T01:34:40.6722123Z         test_lag1_member = "Ethernet4"
2021-04-02T01:34:40.6722445Z         test_lag2_name = "PortChannel0002"
[...]

PR #1692, opened 2021-04-05
[vstest] VOQ system lag vs test timing issue fix

vganesan-nokia:
**What I did**

In Virtual chassis test suite, time sleep for 1 second after creating and deleting system lag is replaced by
DVS's wait_for_n_keys() function.

**Why I did it**

Potential fix for issue https://github.com/Azure/sonic-swss/issues/1687.

**How I verified it**

Ran pytest on docker sonic vs image multiple times with this fix. 100% pass. 

**Details if related**


lguohan (approved):

lguohan:
thank you.

PR #1694, opened 2021-04-06
[acl] Add regression test for `config acl` CLI command

daall:
Add a new test case to catch the bug where calling "config acl add table" multiple
times causes the "ports" field to be deleted from config DB.

**What I did/why I did it**
I added a test case to guard against regressions like the one described in https://github.com/Azure/sonic-utilities/pull/1519.

**How I verified it**
Ran the test case w/ and w/o the change from 1519.

### DIFF
--- a/tests/test_acl_cli.py
+++ b/tests/test_acl_cli.py
@@ -1,0 +1,33 @@
+class TestAclCli:
+    def test_AddTableMultipleTimes(self, dvs, dvs_acl):
+        dvs.runcmd("config acl add table TEST L3 -p Ethernet0")
+
+        cdb = dvs.get_config_db()
+        cdb.wait_for_field_match(
+            "ACL_TABLE",
+            "TEST",
+            {"ports": "Ethernet0"}
+        )
+
+        # Verify that subsequent updates don't delete "ports" from config DB
+        dvs.runcmd("config acl add table TEST L3 -p Ethernet4")
+        cdb.wait_for_field_match(
+            "ACL_TABLE",
+            "TEST",
+            {"ports": "Ethernet4"}
+        )
+
+        # Verify that subsequent updates propagate to ASIC DB
+        L3_BIND_PORTS = ["Ethernet0", "Ethernet4", "Ethernet8", "Ethernet12"]
+        dvs.runcmd(f"config acl add table TEST L3 -p {','.join(L3_BIND_PORTS)}")
+        acl_table_id = dvs_acl.get_acl_table_ids(1)[0]
+        acl_table_group_ids = dvs_acl.get_acl_table_group_ids(len(L3_BIND_PORTS))
+
+        dvs_acl.verify_acl_table_group_members(acl_table_id, acl_table_group_ids, 1)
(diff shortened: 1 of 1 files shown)


qiluo-msft (approved):

lguohan:
since 1519 is not merge, isn't this new test supposed to be fail?

daall:
> since 1519 is not merge, isn't this new test supposed to be fail?

I am trying to figure out if the underlying issue has been resolved in the master branch or if there is some weird discrepancy between the physical device and the virtual switch. Will update shortly.

daall:
> > since 1519 is not merge, isn't this new test supposed to be fail?
> 
> I am trying to figure out if the underlying issue has been resolved in the master branch or if there is some weird discrepancy between the physical device and the virtual switch. Will update shortly.

The underlying issue seems to have been fixed in the master branch, so let me track down the change and see if this is a bugfix we can also grab for 201911.

daall:
> > > since 1519 is not merge, isn't this new test supposed to be fail?
> > 
> > 
> > I am trying to figure out if the underlying issue has been resolved in the master branch or if there is some weird discrepancy between the physical device and the virtual switch. Will update shortly.
> 
> The underlying issue seems to have been fixed in the master branch, so let me track down the change and see if this is a bugfix we can also grab for 201911.

This is the reason for the difference: https://github.com/Azure/sonic-utilities/pull/1392 It seems like the bug exists in swsssdk but not swss-common. I think we should still merge this test case just to prevent any similar regressions in the future.

daall:
> > > since 1519 is not merge, isn't this new test supposed to be fail?
> > 
> > 
> > I am trying to figure out if the underlying issue has been resolved in the master branch or if there is some weird discrepancy between the physical device and the virtual switch. Will update shortly.
> 
> The underlying issue seems to have been fixed in the master branch, so let me track down the change and see if this is a bugfix we can also grab for 201911.

This is the reason for the difference: Azure/sonic-utilities#1392 It seems like the bug exists in swsssdk but not swss-common. I think we should still merge this test case just to prevent any similar regressions in the future. @lguohan do you agree?

qiluo-msft:
This is interesting. How confident are you that https://github.com/Azure/sonic-utilities/pull/1392 fixed the any bug? It is expected as refactoring. If you know the detail, could you explain your API call sequence into swss-common or swsssdk? I can help check the sequence to make sure it is fixed.

daall:
> This is interesting. How confident are you that [Azure/sonic-utilities#1392](https://github.com/Azure/sonic-utilities/pull/1392) fixed the bug? It is expected as refactoring. If you know the detail, could you explain your API call sequence into swss-common or swsssdk? I can help check the sequence to make sure it is fixed.

They're both calling `set_entry` from ConfigDBConnector. When ConfigDBConnector is imported from `swsscommon`, everything works as expected and we can call `set_entry` with `ports@` without issue. When ConfigDBConnector is imported from `swsssdk`, then the first call to `set_entry` works as expected, but the second call deletes the `ports@` field from config DB.

lguohan (approved):

lguohan:
test looks good to me, please proceed to merge.

PR #1695, opened 2021-04-06
[intfmgrd] reach reconciled state at start when there are no interfaces configuration to process

volodymyrsamotiy:

**What I did**

If there is no pending configuration intfmgrd should reach reconciled state immediately. Same does for example vlanmgrd.

**Why I did it**

fdbsyncd waits till intfmgrd reaches reconciled state otherwise exits with exception about timeout. As a result sanity checkers complain about that.

**How I verified it**

I tested it on the switch by running warm-reboot and observed no fdbsyncd failure.
Also I wrote a simple test for intfmgrd warm restart when there are no interfaces in config_db.

**Details if related**

### DIFF
--- a/cfgmgr/intfmgr.cpp
+++ b/cfgmgr/intfmgr.cpp
@@ -195,10 +195,10 @@ void IntfMgr::buildIntfReplayList(void)
 
     m_cfgLoopbackIntfTable.getKeys(intfList);
     std::copy( intfList.begin(), intfList.end(), std::inserter( m_pendingReplayIntfList, m_pendingReplayIntfList.end() ) );
-        
+
     m_cfgVlanIntfTable.getKeys(intfList);
     std::copy( intfList.begin(), intfList.end(), std::inserter( m_pendingReplayIntfList, m_pendingReplayIntfList.end() ) );
-        
+
     m_cfgLagIntfTable.getKeys(intfList);
     std::copy( intfList.begin(), intfList.end(), std::inserter( m_pendingReplayIntfList, m_pendingReplayIntfList.end() ) );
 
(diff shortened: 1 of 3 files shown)


prsunny:
@srj102 , @tapashdas , @nkelapur to review

cfgmgr/intfmgr.cpp
```diff
@@ -191,16 +195,25 @@ void IntfMgr::buildIntfReplayList(void)
...
+
     m_cfgLagIntfTable.getKeys(intfList);
     std::copy( intfList.begin(), intfList.end(), std::inserter( m_pendingReplayIntfList, m_pendingReplayIntfList.end() ) );
 
     SWSS_LOG_INFO("Found %d Total Intfs to be replayed", (int)m_pendingReplayIntfList.size() );
 }
 
+void IntfMgr::setWarmReplayDoneState()
+{
+    replayDone = true;
```
nkelapur, line 210: please follows the class member naming convention now that "replayDone" is a member variable.

nkelapur (commented):
Overall changes look good

cfgmgr/intfmgr.cpp
stepanblyschak: Thanks, fixed

prsunny:
@stepanblyschak , @nkelapur , this is required for 202012 branch, correct?

nkelapur:
> @stepanblyschak , @nkelapur , this is required for 202012 branch, correct?

Yes, this should be required for passing the sanity tests as mentioned by Stepan. However as I understand, in practical usage there will not be a scenario where there are no interfaces in the system, and hence the reported issue should not occur.

prsunny (approved):

stepanblyschak:
@nkelapur @prsunny This is required for 202012.
There can be a L2 config only deployment where there are no L3 interfaces in INTERFACE table, so it  might be  a production issue, not just a sanity check issue.

dprital:
@daall - Can you please merge to 202012 ?

PR #1686, opened 2021-03-31
sonic-swss changes for MPLS

qbdwlr:
What I did
SONiC swss support for MPLS:
1) RouteOrch support for SAI MPLS inseg
2) NeighOrch support for MPLS NHs.
3) CrmOrch support for MPLS inseg and NH accounting.
4) New sonic-swss/tests/test_mpls.py for verification.

Why I did it
SONiC swss support for MPLS

How I verified it
Unit-tests in sonic-swss/tests/test_mpls.py and sonic-utilities/tests
System tests in sonic-mgmt

Details if related
Refer to HLD: Azure/SONiC#706

PR #1699, opened 2021-04-07
fix for CRM available bug

qbdwlr:
**What I did**
Changes update switch statement in CrmOrch::getResAvailableCounters() to use CrmResourceType enum which guarantees the uniqueness of labels in the switch statement.

**Why I did it**
Previously, the switch statement in CrmOrch::getResAvailableCounters() used values from two different enums:  sai_switch_attr_t and sai_acl_table_attr_t.  There is no guarantee that the values from sai_switch_attr_t will not conflict with values from sai_acl_table_attr_t, so it is not good coding practice.  Any future development in this area should not be using values from multiple enums with overlapping values in a single switch statement.  
Additionally, future CRM development can use the new sai_object_type_get_availability() with sai_object_type_t instead of defining new SAI attrs per CrmResourceType without issue of enum conflict.
Refer to crmorch.cpp changes in https://github.com/Azure/sonic-swss/pull/1686 for an example of sai_object_type_get_availability() in CrmOrch::getResAvailableCounters().

**How I verified it**
Re-test of route and acl plus mpls CRM support in https://github.com/Azure/sonic-swss/pull/1686

**Details if related**

### DIFF
--- a/orchagent/crmorch.cpp
+++ b/orchagent/crmorch.cpp
@@ -442,24 +442,24 @@ void CrmOrch::getResAvailableCounters()
             continue;
         }
 
-        sai_attribute_t attr;
-        attr.id = crmResSaiAvailAttrMap.at(res.first);
-
-        switch (attr.id)
+        switch (res.first)
         {
-            case SAI_SWITCH_ATTR_AVAILABLE_IPV4_ROUTE_ENTRY:
-            case SAI_SWITCH_ATTR_AVAILABLE_IPV6_ROUTE_ENTRY:
-            case SAI_SWITCH_ATTR_AVAILABLE_IPV4_NEXTHOP_ENTRY:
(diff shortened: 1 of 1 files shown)


lguohan (approved):

lguohan:
@prsunny to look

prsunny (approved):

abdosi (approved):

qbdwlr:
@prsunny: What is the ETA for moving sonic-swss pointer forward in sonic-buildimage to include this PR?

prsunny:
> @prsunny: What is the ETA for moving sonic-swss pointer forward in sonic-buildimage to include this PR?

@qbdwlr , you can raise a submodule update PR to buildimage similar to https://github.com/Azure/sonic-buildimage/pull/7263

PR #1701, opened 2021-04-09
Add an option to disable orchagent exit in SAI failures in synchronous mode

shi-su:
**What I did**
Add an option to disable orchagent exit in SAI failures in synchronous mode.

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/cfgmgr/buffermgrd.cpp
+++ b/cfgmgr/buffermgrd.cpp
@@ -31,6 +31,7 @@ using json = nlohmann::json;
 int gBatchSize = 0;
 bool gSwssRecord = false;
 bool gLogRotate = false;
+bool gExitInSaiFailure = true;
 ofstream gRecordOfs;
 string gRecordFile;
 /* Global database mutex */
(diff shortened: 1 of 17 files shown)


cfgmgr/buffermgrd.cpp
```diff
@@ -31,6 +31,7 @@ using json = nlohmann::json;
 int gBatchSize = 0;
 bool gSwssRecord = false;
 bool gLogRotate = false;
+bool gExitInSaiFailure = true;
```
prsunny, line 34: Why do we have this Sai failure in manager daemons?

shi-su: This is because the orch class is used in the manager daemons. Thus, the global variables need to be defined here to avoid compilation errors. Here is an existing comment in the manager daemons that talks about why it is needed here.
```
/*
 * Following global variables are defined here for the purpose of
 * using existing Orch class which is to be refactored soon to
 * eliminate the direct exposure of the global variables.
 *
 * Once Orch class refactoring is done, these global variables
 * should be removed from here.
 */
```

PR #1703, opened 2021-04-12
[202012][intfmgrd] reach reconciled state at start when there are no interfaces configuration to process

stepanblyschak:
Signed-off-by: Stepan Blyschak stepanb@nvidia.com

**What I did**

If there is no pending configuration intfmgrd should reach reconciled state immediately. Same does for example vlanmgrd.

**Why I did it**

fdbsyncd waits till intfmgrd reaches reconciled state otherwise exits with exception about timeout. As a result sanity checkers complain about that.

**How I verified it**

I tested it on the switch by running warm-reboot and observed no fdbsyncd failure.
Also I wrote a simple test for intfmgrd warm restart when there are no interfaces in config_db.

**Details if related**

### DIFF
--- a/cfgmgr/intfmgr.cpp
+++ b/cfgmgr/intfmgr.cpp
@@ -207,7 +207,7 @@ void IntfMgr::buildIntfReplayList(void)
 
 void IntfMgr::setWarmReplayDoneState()
 {
-    replayDone = true;
+    m_replayDone = true;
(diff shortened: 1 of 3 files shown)


liat-grozovik:
@yxieca please note that this PR is against 202012 as it cherry pick to 202012 was not possible: https://github.com/Azure/sonic-swss/pull/1695

yxieca (approved):

prsunny (approved):

PR #1705, opened 2021-04-13
Allowing the first time FEC and AN configuration to be pushed to SAI

dgsudharsan:
**What I did**
Allowing the first time configuration for FEC and Autoneg to be pushed to SAI even if they are default.

**Why I did it**
In orchagent, the port struct is pre-initialized with fec = none and autoneg = false (Default SAI values). However in different hardware the underlying implementation might be different. There can also be requirement where the user values need to be forced. In current logic due to the port struct being initialized with default, fec = none or an = false configurations are not pushed from orchagent to SAI. In order to push it even in default case the conditional checks are modified.

**How I verified it**
Configure FEC as none and autoneg as false and check if ASIC DB is programmed.

**Details if related**
This change will force the FEC and autoneg configuration only if the configuration is explicitly present. There is no impact when there are no external FEC or autoneg configuration.
If an external FEC or autoneg config is present below are how different boot scenarios will be handled.
1) Warm boot will not be affected even if an external configuration is present since it will be reconciled. I tested the case and found no link flaps
2) Fast boot and Cold boot - Only if external configuration is present, it will be programmed during port initialization. No additional link flaps is seen or additional time is taken.

### DIFF
--- a/orchagent/port.h
+++ b/orchagent/port.h
@@ -152,6 +152,8 @@ public:
     SystemPortInfo   m_system_port_info;
     SystemLagInfo    m_system_lag_info;
 
+    bool m_fec_cfg = false;
+    bool m_an_cfg = false;
 };
 
 }
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2564,12 +2564,13 @@ void PortsOrch::doPortTask(Consumer &consumer)
             }
             else
             {
-                if (an != -1 && an != p.m_autoneg)
+                if (an != -1 && (!p.m_an_cfg || an != p.m_autoneg))
                 {
                     if (setPortAutoNeg(p.m_port_id, an))
                     {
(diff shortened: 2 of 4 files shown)


dgsudharsan:
This needs to be back ported to 202012

lguohan:
can you check the test failure, it seems related to the auto neg feature.

jleveque:
> This needs to be back ported to 202012

@dgsudharsan: FYI, the autoneg values have changed to on/off in master, but they are still 0/1 in 202012. Therefore, you will need to open a separate PR against 202012 once this is reviewed/merged.

dgsudharsan:
> > This needs to be back ported to 202012
> 
> @dgsudharsan: FYI, the autoneg values have changed to on/off in master, but they are still 0/1 in 202012. Therefore, you will need to open a separate PR against 202012 once this is reviewed/merged.

Sure Joe. I will raise a separate PR after this.

dgsudharsan:
retest vs please

jleveque (approved):

PR #1714, opened 2021-04-19
[sonic-swss] Add port auto negotiation support to swss

Junchao-Mellanox:
**What I did**

1. Added port auto negotiation attributes handle
2. Get supported speeds for each port and save them to CONFIG_DB
3. Added new test cases in VS test to verify the change 

**Why I did it**

To support port auto negotiation feature.

**How I verified it**

Manual test and vs test

**Details if related**


Junchao-Mellanox:
Depends on https://github.com/Azure/sonic-buildimage/pull/7361

Junchao-Mellanox:
Since Azure/sonic-buildimage#7361 is merged, could you please help re-triger the test?

jleveque (changes requested):

jleveque (approved):

orchagent/port.h
```diff
@@ -100,7 +100,7 @@ class Port
     uint32_t            m_mtu = DEFAULT_MTU;
     uint32_t            m_speed = 0;    // Mbps
     std::string         m_learn_mode = "hardware";
-    bool                m_autoneg = false;
+    int                 m_autoneg = -1;  // -1 means not set
```
jleveque, line 103: Suggest adding meanings of other values (e.g., `// -1 means not set, 0 = disabled, 1 = enabled`) in the comment for posterity.

jleveque (changes requested):

jleveque (dismissed):

Junchao-Mellanox:
Fixed conflict

jleveque (approved):

balanokia:
Hi @prsunny @prgeor @lguohan, the following AN feature code violates design of Gearbox port management and it would not be the correct behavior. We have a parent->child relationship in handling ASIC -> PHY ports(system/line) and hence we need to call setPort* APIs. 
When AN is enabled and discovery done, forwarding ASIC faceplate index properties have been inherited as per Gearbox HLD design recommended API setPortSpeed. But the else condition always updates GearboxPort that would impact all the related port parameters like Equalization, Link training, FEC.  
if (pCfg.speed.is_set)
                {
                    if (p.m_speed != pCfg.speed.value)
                    {
                    }
                    else
                    {
                        /* Always update Gearbox speed on Gearbox ports */
                        setGearboxPortsAttr(p, SAI_PORT_ATTR_SPEED, &pCfg.speed.value);
                    }
                }
 Kindly review this code and share your feedback. Thanks

lguohan:
@prgeor ，can you review

prsunny:
@balanokia , do you've a fix proposal/PR?

balanokia:
hi @prsunny We should comment out the "else part of applying hardcoded speed to GB ports" now to make code clean. For the differential speed configuration as well as discovered speed, GB ports should follow setPortSpeed API. Then AN support for GB ports to be handled as another PR (it has to handle system and line side in sync with ASIC and remote end point). I will make a proposal in upcoming release. Pls share your note.

balanokia:
> @balanokia , do you've a fix proposal/PR?
Hi @prsunny Pls share if we can comment it out for cleaner code. Its becoming a precedent for isolated handling of gearbox ports/external PHY ports(line or sys side) without handling from ASIC index. I see, this PR is quoted as a precedent by another PR, https://github.com/sonic-net/sonic-swss/pull/3529. 
Thanks
Bala

PR #1718, opened 2021-04-22
[MACsec]: Change the SCI value for no explict SCI

byu343:
**What I did**
SCI value passed to create_macsec_sc is later used as a mask in the SC/SA matching for packets. When NO explicit SCI is included in the packet, the SCI value has to be set as a wildcard.

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/macsecorch.cpp
+++ b/orchagent/macsecorch.cpp
@@ -1457,7 +1457,7 @@ bool MACsecOrch::createMACsecSC(
     attr.value.oid = flow_id;
     attrs.push_back(attr);
     attr.id = SAI_MACSEC_SC_ATTR_MACSEC_SCI;
-    attr.value.u64 = sci;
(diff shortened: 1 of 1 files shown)


lguohan:
@Pterosaur , to review

lguohan:
can you enhance the vs test?

Pterosaur:
When will the SCI pass to MACsec SC?
I didn't find related APIs in SAI header
https://github.com/opencomputeproject/SAI/blob/master/inc/saimacsec.h#L592-L598

byu343:
From my understanding, the SCI is not always necessary to be passed to the macsec chip. For the macsec chip on our device, it is only needed on ingress: if no explicit sci is expected for the incoming packets, it is set to all-0xf; otherwise, it is set to some real device-specific value.

byu343:
An alternative solution is to fix it in the driver, which seems to be more appropriate, so this pull request is closed.

PR #1719, opened 2021-04-22
[MACsec]: Set macsec to bypass by default

byu343:
**What I did**
In adapting to the MACsec driver for Arista 7280Cr3, we found it is necessary to set SAI_MACSEC_ATTR_PHYSICAL_BYPASS_ENABLE to True, to successfully change the MACsec status from enable to disable (expecting that the traffic can pass through without encryption thereafter). The reason behind is that the default value of the attribute is False, if it is False, the driver understands it as the user is not allowing to bypass MACsec, yielding the result of disabling MACsec not as expected.

**Why I did it**

**How I verified it**

**Details if related**


lguohan (approved):

lguohan:
@Pterosaur , please review.

PR #1720, opened 2021-04-22
[MACsec]: Create data ACL entry before binding

byu343:
**What I did**
In adapting to the MACsec driver for Arista 7280CR3, it is found that the driver requires the ACL entry for data plane to be created before binding the ACL table to port. Therefore, we would like to move the step of creating data ACL entry to before binding table to port. Also, the deletion of ACL entry for data plane will be together with the deletion of ACL table, which is similar to how the ACL entry for control plane is handled.

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/macsecorch.cpp
+++ b/orchagent/macsecorch.cpp
@@ -1376,6 +1376,12 @@ bool MACsecOrch::createMACsecSC(
                 : macsec_port.m_ingress_flow_id;
     }
 
+    auto table =
+        (direction == SAI_MACSEC_DIRECTION_EGRESS)
+            ? &macsec_port.m_egress_acl_table
+            : &macsec_port.m_ingress_acl_table;
+    sc->m_entry_id = table->m_data_entry_id;
+
     if (!createMACsecSC(
             sc->m_sc_id,
             switch_id,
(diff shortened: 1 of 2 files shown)


prsunny:
@Pterosaur to review

lguohan:
can you enhance the vs test?

Pterosaur:
According to my understanding from https://github.com/opencomputeproject/SAI/blob/master/doc/macsec-gearbox/SAI_MACsec_API_Proposal-v1.4.docx, we have an opportunity  to control each MACsec SC in one MACsec port separately. It means the mapping of SC and data ACL entry is 1 to 1.

So I have two concern:
1. this PR will only create one data ACL entry for one MACsec port that may be includes multiple MACsec SC
2. this PR sets the `sci` to zero if the flag `sci_in_sectag` is `true`, is it correct?

byu343:
1. Yes, under this change, if there are multiple SCs, they have to be attached to the same ACL entry and flow. Actually, we only considered and tested the case of one SC per port. I am wondering whether it is important to consider the use case for multiple SCs on the same port, as the user interface through config_db (attaching one profile to the port) seems not supporting it, which is:
{
    "PORT": {
        "Ethernet8": {
            "macsec": "test_profile1"
        }
    }
} 
2. I agree that the SCI passed here should be some meaningful value. For our device, we can set it to 0, because it is an unsupported  and not used field for our MACsec chip. The solution in this change is not generic to other devices.

Considering the issues brought by the change,  I will investigate whether we can skip this change and fix the issue in other ways like on the driver.

byu343:
The issue can be fixed in the driver,  so this request is closed.

PR #1727, opened 2021-04-26
[subintf] Mirror via a local sub port router interface

wendani:
**What I did**
Deal with the possibility that the chosen next hop object goes through a local sub port router interface

**Why I did it**
Mirror destination lookup is done in default VRF. As a sub port router interface can be the local router interface of the next hop or a next hop group member of an ip prefix in default VRF, here we deal with such a likelihood.

**How I verified it**
Newly added vs tests:

- Mirror session add removal, together with the following update event MirrorOrch observes:
    - ip prefix removal and add
    - neighbor mac update
    - lag member removal and add
- Mirror session status update at parent port oper status change for the case of mirror destination in directly connected subnet ()
- Next hop group update on LPM ip prefix that causes monitor port (Mirror to port) to migrate from via a local sub port router interface to via other local RIFs, and vice versa. Validate a local RIF migrated to/back from being one of the following types:
    - EthernetX
    - PortChannelX
    - Vlan interface of the same vlan id
    - Vlan interface of a different vlan id
    - EthernetX.Y sub port interface
    - PortChannelX.Y sub port interface
- LPM ip prefix update that cause monitor port (Mirror to port) to migrate from via a local sub port router interface to via other local RIFs, and vice versa. Validate a local RIF migrated to/back from being one of the following types listed above.

**Details if related**
Contains (dependency on) https://github.com/Azure/sonic-swss/pull/1725


prsunny:
Test failures?

PR #1729, opened 2021-04-27
[neighorch] VOQ encap index change handling

vganesan-nokia:
**What I did**

For the remote neighbors in the VOQ systems when there is change in the
encap index (due to syncd restart), the previously programmed remote
neighs should be re-written with the new changed encap index. The
current voq system neigh handling api in neighorch only checks for the
change of mac address. Because of this the change in the encap index is
ignored. This causes mismatch of encap index in the neighbor records in
owner asic and the remote asics after syncd restart situations like
config reload or failure recovery. This results in packet forwarding
failures for the packets forwarded across fabric and egressing in
different asics than ingress asics. The encap index that was previously
allocated by SAI and synced to the CHASSIS_APP_DB may change for the
same neighbor when syncd is restarted after config reload or any failure
recovery situations. When syncd restarts, the SAI may undergo full
re-initialization and may loose the memory of previous encap index
allocation for the neighbors. During reprogramming of the neighbors they
may not get the same encap index that were allocated before restart.

This fix is to store the allocated encap index in the orchagent
(neighorch) and compare with received encap index if the neighbor entry
already exits. This comparison of encap index is new to voq chassis.
This is done in addition to checking for mac change. For any received
(synced) neigh (received from CHASSIS_APP_DB) if neigh entry already
exists and if there is change in encap  index, the current neigh entry
is removed and re-added. As of now, since current SAI does not support
change of encap index (i.e, set of operaton of encap index attribute for
neigh records) del and re-add is done.

**Why I did it**

To fix the traffic loss problem in voq chassis after config reload. The problem is described in issue https://github.com/Azure/sonic-buildimage/issues/7451

**How I verified it**

- vs test for test_virtual_chassis.py fully pass.
- With setup as mentioned in issue https://github.com/Azure/sonic-buildimage/issues/7451, performed config reload multiple times, and verified that the traffic recovers fully after the line card becomes stable and operational.

**Details if related**

### DIFF
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -783,16 +783,16 @@ bool NeighOrch::addNeighbor(const NeighborEntry &neighborEntry, const MacAddress
     MuxOrch* mux_orch = gDirectory.get<MuxOrch*>();
     bool hw_config = isHwConfigured(neighborEntry);
 
+    if (gMySwitchType == "voq")
+    {
+        if (!addVoqEncapIndex(alias, ip_address, neighbor_attrs))
+        {
+            return false;
+        }
+    }
+
     if (!hw_config && mux_orch->isNeighborActive(ip_address, macAddress, alias))
     {
-        if (gMySwitchType == "voq")
-        {
-            if (!addVoqEncapIndex(alias, ip_address, neighbor_attrs))
-            {
-                return false;
-            }
-        }
-
         status = sai_neighbor_api->create_neighbor_entry(&neighbor_entry,
                                    (uint32_t)neighbor_attrs.size(), neighbor_attrs.data());
         if (status != SAI_STATUS_SUCCESS)
@@ -1230,8 +1230,30 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
             }
 
             if (m_syncdNeighbors.find(neighbor_entry) == m_syncdNeighbors.end() ||
-                    m_syncdNeighbors[neighbor_entry].mac != mac_address)
+                    m_syncdNeighbors[neighbor_entry].mac != mac_address ||
+                    m_syncdNeighbors[neighbor_entry].voq_encap_index != encap_index)
             {
+                // Handle encap index change. SAI does not support change of encap index for
+                // existing neighbors. Remove the neighbor but do not errase from consumer sync
+                // buffer. The next iteration will add the neighbor back with new encap index
+
+                if (m_syncdNeighbors.find(neighbor_entry) != m_syncdNeighbors.end() &&
+                    m_syncdNeighbors[neighbor_entry].voq_encap_index != encap_index)
+                {
+                    //Remove neigh from SAI
+                    if (removeNeighbor(neighbor_entry))
+                    {
+                        //neigh successfully deleted from SAI. Set STATE DB to signal to remove entries from kernel
+                        m_stateSystemNeighTable->del(state_key);
+                    }
+                    else
+                    {
+                        SWSS_LOG_ERROR("Failed to remove voq neighbor %s from SAI during encap index change", kfvKey(t).c_str());
+                    }
+                    it++;
+                    continue;
+                }
+
                 //Add neigh to SAI
                 if (addNeighbor(neighbor_entry, mac_address))
                 {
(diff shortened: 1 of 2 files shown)


anshuv-mfst:
@abdosi - could you please take a look, thanks.

anshuv-mfst:
@ysmanman please take a look.

orchagent/neighorch.cpp
```diff
@@ -1133,8 +1133,30 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
             }
 
             if (m_syncdNeighbors.find(neighbor_entry) == m_syncdNeighbors.end() ||
-                    m_syncdNeighbors[neighbor_entry].mac != mac_address)
+                    m_syncdNeighbors[neighbor_entry].mac != mac_address ||
+                    m_syncdNeighbors[neighbor_entry].voq_encap_index != encap_index)
             {
+                // Handle encap index change. SAI does not support change of encap index for
+                // existing neighbors. Remove the neighbor but do not errase from consumer sync
+                // buffer. The next iteration will add the neighbor back with new encap index
```
ysmanman, line 1141: Deleing neigh and adding it back may churn routers resolving over that neigh. So routes get deleted and added back.  I guess encap index change is common. Probably should be seen in syncd restart, as you mentioned, or linecard reload? Ideally, encap index change should be handled by SAI just like how it handles MAC change today.

ysmanman: typo: i meant to say encap index change is NOT common.

vganesan-nokia: I agree. SAI should handle the encap index change similar to mac change. But current SAI does not support it. Till SAI supports it, we will have to live with this delete and re-add approach. Once SAI is enhanced for the encap index change handling, we can change this.

orchagent/neighorch.cpp
```diff
@@ -1133,8 +1133,30 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
             }
 
             if (m_syncdNeighbors.find(neighbor_entry) == m_syncdNeighbors.end() ||
-                    m_syncdNeighbors[neighbor_entry].mac != mac_address)
+                    m_syncdNeighbors[neighbor_entry].mac != mac_address ||
```
ysmanman, line 1136: Why do we check mac for remote neighbor? Isn't checking encap sufficient?

orchagent/neighorch.cpp
```diff
@@ -1133,8 +1133,30 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
...
+                // buffer. The next iteration will add the neighbor back with new encap index
+
+                if (m_syncdNeighbors.find(neighbor_entry) != m_syncdNeighbors.end() &&
+                    m_syncdNeighbors[neighbor_entry].voq_encap_index != encap_index)
+                {
+                    //Remove neigh from SAI
+                    if (removeNeighbor(neighbor_entry))
+                    {
+                        //neigh successfully deleted from SAI. Set STATE DB to signal to remove entries from kernel
+                        m_stateSystemNeighTable->del(state_key);
```
ysmanman, line 1150: Why we only do this when removeNeighbor succeeds?  If removeNeighbor fails for some reasons, kernel entries are not removed and then traffic may still be forwarded with wrong encap index.  Should we signal removing kernel entries always when encap index changes?

orchagent/neighorch.cpp
vganesan-nokia: No. For mac change, the encap index is not changed. If we only check encap index, we'll not be processing the mac change.

orchagent/neighorch.cpp
vganesan-nokia: We do this for neat re-trying. Removal of kernel neigh without successful removal of neigh from SAI is partial processing. This will create problems in retries without additional logic.

ysmanman (dismissed):

anshuv-mfst:
@neethajohn - please review and approve.

orchagent/neighorch.cpp
abdosi: @vganesan-nokia can you create issue in SONiC for tracking SAI not supporting set operation

vganesan-nokia: > @vganesan-nokia can you create issue in SONiC for tracking SAI not supporting set operation

Yes. Created issue https://github.com/Azure/sonic-buildimage/issues/7879. Thanks

abdosi: i was discussing with @prsunny and we think better option will be call SAI api with set and if it fails (which it will) then we should do remove/add neigh. That way once the attribute is supported we don't need any sonic change going forward.

abdosi (changes requested):
as added comment.

orchagent/neighorch.cpp
vganesan-nokia: There is a problem with this approach. In current SAI (w.r.t brcm SAI 5.0), when set is done for encap index attribute in the neighbor record, SAI's neighbor entry set attribute API returns SAI_STATUS_FAILURE. When there is not SUCCESS from SAI, syncd sends orchagent shut down.

vganesan-nokia: Fixed. Since syncd is not sending orchagent shutdown for SET failures anymore, changes are done as suggested.

abdosi (approved):

abdosi:
For the remote neighbors in the VOQ systems when there is change in the
encap index (due to syncd restart), the previously programmed remote
neighs should be re-written with the new changed encap index. The
current voq system neigh handling api in neighorch only checks for the
change of mac address. Because of this the change in the encap index is
ignored.

PR #1730, opened 2021-04-28
Vlan status notifier

dr412113:
**What I did**
    - Maintain a count in orchagent VLAN strcuture indicating the number
    of VLAN members in UP state.
    - Send VLAn UP/DOWN to Vlanmgr either when the first VLAN member
    comes up or the last VLAN member goes down.
    - Vlanmgr will update VLAN admin state in kernel based on notification
    received from orchagent.

**Why I did it**
Vlan routing interface is UP even though the corresponding port-channel on which that vlan routing interface configured is DOWN.

**How I verified it**
Manual tests

**Details if related**


prsunny:
Please explain what is the impact if the Vlan interface is staying up?. Also what are the config changes to SAI/ASIC during this.

PR #1735, opened 2021-04-30
[bufferpool] Support for buffer pool type both

vganesan-nokia:
**What I did**

Changes done to allow orchagent accepting "both" type for buffer pool
configuration.
 
**Why I did it**

Currently sonic supports only 'ingress' and 'egress' types of buffer pools.
Since some devices like DNX based switches only support buffer pool
type 'both', we are unable to create buffer pools for these devices.
This patch fixes this problem by enhancing buffer pool configuration to
accept "both" type. 

**How I verified it**

In a switch that uses DNX arch based chips (like VOQ chassis using line cards with Jerico family of switches), the following configuration is applied
```
"BUFFER_POOL": {
        "lossless_pool": {
            "size": "12766208",
            "type": "both",
            "mode": "dynamic"
        }
 }
```
when the switch became operational, it was observed, there are no error logs indicating invalid parameter or failure of buffer pool creation.

**Details if needed**
- This PR is fix for the issue https://github.com/Azure/sonic-buildimage/issues/7497

### DIFF
--- a/orchagent/qosorch.cpp
+++ b/orchagent/qosorch.cpp
@@ -232,7 +232,7 @@ bool Dot1pToTcMapHandler::convertFieldValuesToAttributes(KeyOpFieldsValuesTuple 
     sai_qos_map_list_t dot1p_map_list;
 
     // Allocated resources are freed in freeAttribResources() call
-    dot1p_map_list.list = new sai_qos_map_t[kfvFieldsValues(tuple).size()];
+    dot1p_map_list.list = new sai_qos_map_t[kfvFieldsValues(tuple).size()]();
     int i = 0;
     for (const auto &fv : kfvFieldsValues(tuple))
     {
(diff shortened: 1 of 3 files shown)


anshuv-mfst:
@ysmanman, @abdosi - could you please review, thanks.

anshuv-mfst:
@neethajohn , @lguohan - please take a look, thanks.

orchagent/qosorch.cpp
```diff
@@ -232,7 +232,7 @@ bool Dot1pToTcMapHandler::convertFieldValuesToAttributes(KeyOpFieldsValuesTuple
     sai_qos_map_list_t dot1p_map_list;
 
     // Allocated resources are freed in freeAttribResources() call
-    dot1p_map_list.list = new sai_qos_map_t[kfvFieldsValues(tuple).size()];
+    dot1p_map_list.list = new sai_qos_map_t[kfvFieldsValues(tuple).size()]();
```
neethajohn, line 235: can you move this change to a separate PR?

vganesan-nokia: > can you move this change to a separate PR?

Yes.

vganesan-nokia: Fixed. The change is removed from this PR and put in a separate PR https://github.com/Azure/sonic-swss/pull/1746

ysmanman (approved):

neethajohn (approved):

anshuv-mfst:
@neethajohn - can you please help with merge, thanks!

PR #1737, opened 2021-05-08
Gcov for swss daemon

pettershao-ragilenetworks:
**What I did**
Introduce gcov-related automation script to test coverage rate for swss module
relate HLD: https://github.com/Azure/SONiC/pull/642
**Why I did it**
Attempt to integrate coverage report for each module in SONiC to Azure pipeline 
**How I verified it**
Using Azure pipeline to check whether a coverage report can be checked after a pull request
**Details if related**


pettershao-ragilenetworks:
@xumia @lguohan this experiment blocked by no master branch sai build, help build a successful master sai, thanks!(yesterday's build seems faild)
https://dev.azure.com/mssonic/build/_build?definitionId=12&_a=summary

.azure-pipelines/build-docker-sonic-vs-template.yml
```diff
@@ -3,8 +3,6 @@ parameters:
   type: string
   values:
   - amd64
-  - armhf
-  - arm64
```
lguohan, line 7: please revert it back.

.azure-pipelines/build-docker-sonic-vs-template.yml
```diff
@@ -72,6 +61,16 @@ jobs:
...
 
+      curl -k -O -L "https://sonic-build.azurewebsites.net/api/sonic/artifacts?branchName=master&platform=vs&target=target/debs/buster/libsaivs_1.0.0_amd64.deb"
+      curl -k -O -L "https://sonic-build.azurewebsites.net/api/sonic/artifacts?branchName=master&platform=vs&target=target/debs/buster/libsaivs-dev_1.0.0_amd64.deb"
+      curl -k -O -L "https://sonic-build.azurewebsites.net/api/sonic/artifacts?branchName=master&platform=vs&target=target/debs/buster/libsairedis_1.0.0_amd64.deb"
+      curl -k -O -L "https://sonic-build.azurewebsites.net/api/sonic/artifacts?branchName=master&platform=vs&target=target/debs/buster/libsairedis-dev_1.0.0_amd64.deb"
+      curl -k -O -L "https://sonic-build.azurewebsites.net/api/sonic/artifacts?branchName=master&platform=vs&target=target/debs/buster/libsaimetadata_1.0.0_amd64.deb"
+      curl -k -O -L "https://sonic-build.azurewebsites.net/api/sonic/artifacts?branchName=master&platform=vs&target=target/debs/buster/libsaimetadata-dev_1.0.0_amd64.deb"
+      curl -k -O -L "https://sonic-build.azurewebsites.net/api/sonic/artifacts?branchName=master&platform=vs&target=target/debs/buster/syncd-vs_1.0.0_amd64.deb"
```
lguohan, line 70: please revert back.

fdbsyncd/Makefile.am
```diff
@@ -12,5 +12,5 @@ fdbsyncd_SOURCES = fdbsyncd.cpp fdbsync.cpp $(top_srcdir)/warmrestart/warmRestar
 
 fdbsyncd_CFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(COV_CFLAGS)
 fdbsyncd_CPPFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(COV_CFLAGS)
-fdbsyncd_LDADD = -lnl-3 -lnl-route-3 -lswsscommon $(COV_LDFLAGS)
+fdbsyncd_LDADD = -lnl-3 -lnl-route-3 -lswsscommon $(COV_LDFLAGS) -lgcov_preload
```
lguohan, line 15: make it optional

tests/test_nhg.py
```diff
@@ -44,9 +44,7 @@ def test_route_nhg(self, dvs, dvs_route, testlog):
 
         dvs_route.check_asicdb_deleted_route_entries([rtprefix])
 
-        fvs = swsscommon.FieldValuePairs([("nexthop","10.0.0.1,10.0.0.3,10.0.0.5"),
```
lguohan, line 47: please revert all tests changes.

configure.ac
```diff
@@ -84,7 +84,10 @@ CFLAGS_COMMON+=" -Wno-switch-default"
 CFLAGS_COMMON+=" -Wno-long-long"
 CFLAGS_COMMON+=" -Wno-redundant-decls"
 
+CFLAGS_COMMON+=" -fprofile-arcs -ftest-coverage"
```
lguohan, line 87: should be configurable

.azure-pipelines/build-template.yml
```diff
@@ -100,6 +105,8 @@ jobs:
       set -x
       tar czf pytest.tgz tests
       cp -r pytest.tgz $(Build.ArtifactStagingDirectory)/
+      sudo cp libgcov_preload.so /usr/lib/
```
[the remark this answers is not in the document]
qiluo-msft: I see this PR contains a binary file `libgcov_preload.so`. It is not a good idea to check-in binary into code repo.

pettershao-ragilenetworks:
@lguohan hi guohan, help forward the related hld together, thanks: https://github.com/Azure/SONiC/pull/642

kcudnik:
please address lgtm issues

pettershao-ragilenetworks:
> please address lgtm issues
@kcudnik done, help forward this, thanks!

kcudnik:
@pettershao-ragilenetworks 
* this seems like good PR, could we do something similar to sonic-sairedis?
* is code coverage calculated during only local sonic-swss unittests run or also virtual switch ptf/azure pipeline run ?
* i noticed that in posted artifacts where gcov report is created, multiple html pages are create for the same cpp file, is that intended? or those are from different runs ?

pettershao-ragilenetworks:
> @pettershao-ragilenetworks
> 
> * this seems like good PR, could we do something similar to sonic-sairedis? yea, will do this soon.
> * is code coverage calculated during only local sonic-swss unittests run or also virtual switch ptf/azure pipeline run ? calculated in virtual switch ptf/azure pipeline run, since gcc version(and some other environment) is not same between vs docker and virtual switch.(gcda file generate in vs docker)(calculation will fail in vs docker as experiment before)
> * i noticed that in posted artifacts where gcov report is created, multiple html pages are create for the same cpp file, is that intended? or those are from different runs ? yea, they are generated by different pytest case,  every test case generate one vs docker, so there exist many gcda files for the same cpp file. and we have to generate info file for all the gcda files and merge the info files in the end.

prsunny:
@pettershao-ragilenetworks , can you please review https://github.com/Azure/sonic-swss/pull/1816

pettershao-ragilenetworks:
@prsunny please help check https://dev.azure.com/mssonic/build/_build/results?buildId=24291&view=results, the time for gcov report has reduce to 10 mins, seems a huge improvement. but the coverage rate is less, please help check.

kcudnik:
can you fix lgtm errors?

pettershao-ragilenetworks:
@prsunny help forward this, thanks!

kcudnik:
seems like you have the same gcov_support.sh in 3 different locations, can you reuse the same file ?

.azure-pipelines/gcov.yml
```diff
@@ -0,0 +1,109 @@
...
+    workingDirectory: $(Pipeline.Workspace)
+    displayName: "generate gcov reports"
+    condition: eq('${{ parameters.archive_gcov }}', true)
+
+  - publish: $(Build.ArtifactStagingDirectory)/gcov_output
+    artifact: gcov_info
+    displayName: "Archive gcov info"
+    condition: eq('${{ parameters.archive_gcov }}', true)
+
+  - task: PublishCodeCoverageResults@1
```
xumia, line 102: Looks like the task PublishCodeCoverageResults supports to publish multiple coverage files in single task, @pettershao-ragilenetworks, have you tried to merged the coverage files by the task itself? I am not sure it works or not.

pettershao-ragilenetworks: we only publish total.info(which later be convert into coverage.xml)one time. 
the reason why we merge the info in gcov_support.sh is that pipeline could't recognize the info files and will merge fail.

kcudnik:
why do you need gcovpreload added to every makefile ?

pettershao-ragilenetworks:
1. gcovpreload is used for process capture signal to generate gcda file, so every process need it.
2. yes, we have already do like this, condition is always, pleae check
3. will fix soon for 3 copies of sh file.

pettershao-ragilenetworks:
@kcudnik hlep check newest result, gcov time has reduce to 13mins, help forward this, thanks!

kcudnik:
> 1. gcovpreload is used for process capture signal to generate gcda file, so every process need it.
> 2. yes, we have already do like this, condition is always, pleae check.(no relation with gcovpreload)
> 3. will fix soon for 3 copies of sh file.

Ad 1. Not sure about that, i run unittests in sairedis repo and only what i need is flags for compilation and -lgcov for libraries, no need for gcovpreload extra library
Ad 2. when process exits normally when compiled with -lgcov, then gcda files are automatically written no need for special gcovpreload
Ad 3. ok

pettershao-ragilenetworks:
Actually, we do not import this lib at begining, only with LDFLAGS+=" -fprofile-arcs -lgcov" in configure.ac, but no gcda file generate after test.

kcudnik:
> Actually, we do not import this lib at begining, only with LDFLAGS+=" -fprofile-arcs -lgcov" in configure.ac, but no gcda file generate after test no matter what acction we take to stop container or process.

hmm this is interesting, can you try flags "-O0 -g -fprofile-arcs -ftest-coverage" ? this should require lgcov for linking stage

theasianpianist:
@pettershao-ragilenetworks DVS re-use PR has passed, can you try with the latest changes https://github.com/Azure/sonic-swss/pull/1816

pettershao-ragilenetworks:
> > Actually, we do not import this lib at begining, only with LDFLAGS+=" -fprofile-arcs -lgcov" in configure.ac, but no gcda file generate after test no matter what acction we take to stop container or process.
> 
> hmm this is interesting
> 
> from gcc man 2nd paragraph:
> 
> ```
>        -fprofile-arcs
>            Add code so that program flow arcs are instrumented.  During execution the program records how many times each branch and call is executed and
>            how many times it is taken or returns.  On targets that support constructors with priority support, profiling properly handles constructors,
>            destructors and C++ constructors (and destructors) of classes which are used as a type of a global variable.
> 
>            When the compiled program exits it saves this data to a file called auxname.gcda for each source file.  The data may be used for profile-
>            directed optimizations (-fbranch-probabilities), or for test coverage analysis (-ftest-coverage).  Each object file's auxname is generated from
>            the name of the output file, if explicitly specified and it is not the final executable, otherwise it is the basename of the source file.  In
>            both cases any suffix is removed (e.g. foo.gcda for input file dir/foo.c, or dir/foo.gcda for output file specified as -o dir/foo.o).
> ```
> 
> maybe because test container don't contain source files? or .libs or source directories are not existing in test container ?

actually, we have tried souce code on local(with artifict vs docker) also, not work either. And the root cause we concluded is that swss project consists of multiple modules, which are generated by gcda, but some are not. The reason is that many part of the module is processed in a loop and never exits.
Use kill -9 to close, if the program does not have the sigaction function, the gcda file cannot be generated.

pettershao-ragilenetworks:
> @pettershao-ragilenetworks DVS re-use PR has passed, can you try with the latest changes #1816

yea, have done yesterday, please check the result.

kcudnik:
> From the implementation principle of gcov, it can be known that if the user process does not call exit() to exit, the coverage statistics will not be obtained, and there will be no way to generate reports. Once the background service program is started, it rarely exits actively. In order to solve this problem, we can register a signal handler for the program under test to handle common forced exit signals such as SIGINT, SIGQUIT, SIGTERM, and actively call the exit or __gcov_flush function in the signal processing function to output statistics result

i understand your approach and i know why you doing it this way, but it seems something is wrong to me if you say that if service program (which exactly) exits actively. then it ether calls abort, or throws, those seems to bad scenarios or when tests crashes, when test is successfully there should be no crash or abort called. exceptions should be properly handled then. even if you expecting exceptions testing wrong paths, im using GTEST for testing that and test exceptions using EXPECT_THROW macro from gests, and this executed code is still written to gcda file, please take a look for example in sairedis/unittest/lib/TestUtils.cpp and you can run those tests using "make check" command, you can also see the compilator flags in Makefile.am for unittets  and for lib/Makefile.am for coverage

ps. i see that your code is merging tests from docker test bed, could you also merge tests from package unittests ? the probably needs another step before "make check" to execute them, which can add additional coverage %. I use unittests in sairedis repo to cover scenarios which are not used in production or vstests yes, byt are covered during make check, and this is actually significant portion of coverage %

pettershao-ragilenetworks:
> > From the implementation principle of gcov, it can be known that if the user process does not call exit() to exit, the coverage statistics will not be obtained, and there will be no way to generate reports. Once the background service program is started, it rarely exits actively. In order to solve this problem, we can register a signal handler for the program under test to handle common forced exit signals such as SIGINT, SIGQUIT, SIGTERM, and actively call the exit or __gcov_flush function in the signal processing function to output statistics result
> 
> i understand your approach and i know why you doing it this way, but it seems something is wrong to me if you say that if service program (which exactly) exits actively. then it ether calls abort, or throws, those seems to bad scenarios or when tests crashes, when test is successfully there should be no crash or abort called. exceptions should be properly handled then. even if you expecting exceptions testing wrong paths, im using GTEST for testing that and test exceptions using EXPECT_THROW macro from gests, and this executed code is still written to gcda file, please take a look for example in sairedis/unittest/lib/TestUtils.cpp and you can run those tests using "make check" command, you can also see the compilator flags in Makefile.am for unittets and for lib/Makefile.am for coverage
> 
> ps. i see that your code is merging tests from docker test bed, could you also merge tests from package unittests ? the probably needs another step before "make check" to execute them, which can add additional coverage %. I use unittests in sairedis repo to cover scenarios which are not used in production or vstests yes, byt are covered during make check, and this is actually significant portion of coverage %

1. I think this is also fit to creash scenarios, as kernel will send signal to stop process(actually, i have encountered a similar problem before without lcov preloader, process exit unexpected but gcda file generated)
2. this merging is about test speed, before optimize, the time for gcov is about 1.5h, now up to 12minus. So may not merge any other from unitests. I think we can do that when this done and forcurs on sairedis repo.

theasianpianist:
> > @pettershao-ragilenetworks DVS re-use PR has passed, can you try with the latest changes #1816
> 
> yea, have done yesterday, please check the result.

where is the result? Not sure what to look for, just let me know if you need further changes on the DVS management side

pettershao-ragilenetworks:
> > * I think this is also fit to crash scenarios, as kernel will send signal to stop process(actually, i have encountered a similar problem before without lcov preloader, process exit unexpected but gcda file generated)(actually preloader is an enhancement, will not affect origin logic)
> > * this merging is about test speed, before optimize, the time for gcov is about 1.5h, now up to 12minus. So may not merge any other from unitests. I think we can do that when this done and forcurs on sairedis repo.
> 
> Ad 1. is that kernel kill process desired scenario at all?  i think yes
> Ad 2. what was taking 1.5h ? merging gcda files? or which process? The huge number vs containers  cause long time for info file generation. Now microsoft team has try to reuse container and the time has been reduced.

pettershao-ragilenetworks:
> > > @pettershao-ragilenetworks DVS re-use PR has passed, can you try with the latest changes #1816
> > 
> > 
> > yea, have done yesterday, please check the result.
> 
> where is the result? Not sure what to look for, just let me know if you need further changes on the DVS management side

I think it is fine now, the time is reduce to 12 mins.

kcudnik:
> Ad 1. is that kernel kill process desired scenario at all? i think yes

can you find out why is this? this doesn't seem normal to me for kernel to kill the proces in test scenarios, since this seems too random and unpredictable, maybe test case is designed in wrong way

pettershao-ragilenetworks:
> > Ad 1. is that kernel kill process desired scenario at all? i think yes
> 
> can you find out why is this? this doesn't seem normal to me for kernel to kill the proces in test scenarios, since this seems too random and unpredictable, maybe test case is designed in wrong way

Hi, kcudnik, currently preloader is only used for resolving gcda file can't generate inside vs docker for swss pytest repo. For sai repo unitest, i think should open a new PR to do some experiment to check. do you think is fine? wait for your response, thanks!

prsunny:
@pettershao-ragilenetworks , could you please rebase and rerun as now we have the PR https://github.com/Azure/sonic-swss/pull/1816 merged

kcudnik:
> Hi, kcudnik, currently preloader is only used for resolving gcda file can't generate inside vs docker for swss pytest repo. For sai repo unitest, i think should open a new PR to do some experiment to check. do you think is fine? wait for your response, thanks!

if you could test that on sairedis repo with existing unittests, that would be great, but i think it could be tricky, and it would require some extra step or pipeline change, since currently code coverage on sairedis BuildDocker amd64 pipeline, and as for this chance, CC would be generated by other pipeline, maybe we could have a call on that how to approach this problem

merging 2 coverage.xml one from pytests and one from unittests should possible, or if required we could pull gcda from sairedis build and coverage.xml, export this as artefacts, and put gcov pytest pipeline after that build, and consume that artefact, and merge all gcda files together since you already have a way to do that

fyl @lguohan 
i think this would be very good for entire coverage, since from unittests im not able to test any traffic flow code, fdb events, and macsec flow, since that is done from kernel and tap interfaces. and that a huge chunk of code, which from my point of view it make no sense to develop substitute for kernel pipeline just to artificially in software simulate generated packets, and testing on existing vs platform would definitely bump up code coverage over all, which at this point it will be challenging to achieve 80%

pettershao-ragilenetworks:
hi  @prsunny  @theasianpianist help forward this, thanks!

Makefile.am
```diff
@@ -1,4 +1,9 @@
-SUBDIRS = fpmsyncd neighsyncd fdbsyncd portsyncd mclagsyncd natsyncd orchagent swssconfig cfgmgr tests gearsyncd
+if GCOV_ENABLED
+SUBDIRS = gcovpreload fpmsyncd neighsyncd portsyncd mclagsyncd natsyncd fdbsyncd orchagent swssconfig cfgmgr tests
```
theasianpianist, line 2: @pettershao-ragilenetworks please add `gearsyncd` here, believe this may be the reason the PR tests are consistently failing.

prsunny (approved):

prsunny:
@xumia , could you please sign-off ?

xumia (approved):

kcudnik (approved):
after this will be merged, may we check how much coverage this will generate for sonic-sairedis repo ? currently sairedis repo only runs local unittests which cover about 60% of the code, maybe running docker tests will bump that number up

kcudnik:
@pettershao-ragilenetworks hey, can we try use this approach on sonic-sairedis repo ? to find out what coverage could be achieved, since current unittest cover on sairedis repo in 60%, i will write mail

pettershao-ragilenetworks:
> @pettershao-ragilenetworks hey, can we try use this approach on sonic-sairedis repo ? to find out what coverage could be achieved, since current unittest cover on sairedis repo in 60%, i will write mail

OK!

PR #1738, opened 2021-05-08
Swss gcov support

pettershao-ragilenetworks:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


lguohan:
can you check the here. your azp changes has syntax error. https://dev.azure.com/mssonic/build/_build?definitionId=15

PR #1646, opened 2021-02-18
[flex-counters] Delay flex counters stats init for faster boot time

shlomibitton:
Signed-off-by: Shlomi Bitton <shlomibi@nvidia.com>

**What I did**
Update flex counters DB with counters stats only when counters are enabled.
As long as the polling counters are not enabled, flex counters information will stored internally on PortsOrch.

**Why I did it**
Creating flex counters objects on the DB will trigger 'SYNCD' to access the HW for query statistics capabilities.
This HW access takes time and will be better to finish boot before doing this (mainly for fast-reboot but good to have in general).
The flex counters are not crucial at boot time, we can delay it to the end of the boot process.

**How I verified it**
Reboot a switch and observer the flex counters DB populated after counters are enabled.

**Details if related**

PR #1739, opened 2021-05-10
Fix for delay flex counters flow

shlomibitton:

**What I did**
Exclude installing flex counter for CPU interface (mgmt) for proper query counter capability.
Fixing this PR: https://github.com/Azure/sonic-swss/pull/1646

**Why I did it**
The CPU interface does not suppose to be inserted to flex counter DB.
In addition, if we query the counter statistics capability by this interface it will cause all port counters id lists to be empty and break lua port rates functionality.

**How I verified it**
Enable all counters group types and observe all ID lists for ports populated correctly.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -4714,6 +4714,10 @@ void PortsOrch::generatePortCounterMap()
     auto port_counter_stats = generateCounterStats(PORT_STAT_COUNTER_FLEX_COUNTER_GROUP);
     for (const auto& it: m_portList)
     {
+        if (it.first == "CPU")
+        {
+            continue;
+        }
         port_stat_manager.setCounterIdList(it.second.m_port_id, CounterType::PORT, port_counter_stats);
     }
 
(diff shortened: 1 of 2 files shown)


liat-grozovik (changes requested):
missing unit test to make sure that when CPU port is the first it is working as expected. so then you can add the original fix which is now expected to be reverted.

shlomibitton:
@qiluo-msft @lguohan Can you please review this?
This is the fix for the issue introduced from PR: https://github.com/Azure/sonic-swss/pull/1646 https://github.com/Azure/sonic-swss/pull/1736
I added another unit test to verify this case.
To merge this fix properly we need to merge again the previous PR's and then merge this fix to both branches.

qiluo-msft:
Could you included previous code into this PR so we can merge this PR atomically?

shlomibitton:
> Could you included previous code into this PR so we can merge this PR atomically?

@qiluo-msft I created 2 new PR's with the previous code, issue fix and new unit test case for both branches.
master: https://github.com/Azure/sonic-swss/pull/1749
202012: https://github.com/Azure/sonic-swss/pull/1750

I am closing this PR since it is not relevant

PR #1741, opened 2021-05-12
[doc] Update CONFIG_DB manual to include NTP global settings

robocoder99:
Updated the Configuration Database documentation to include settings supported by the NTP daemon to globally configure its source port, which controls the port from which NTP requests are made (needs to be overridden in some cases) and also controls whether it runs within the mgmt vrf (which also needs to be set in some common configurations)


prsunny:
Is it related to PR https://github.com/Azure/sonic-buildimage/pull/7586 ?

yxieca (approved):

robocoder99:
> Is it related to PR [Azure/sonic-buildimage#7586](https://github.com/Azure/sonic-buildimage/pull/7586) ?

Yes. But only tangentially. No features were added in that PR but there was also no documentation of this feature that already existed.

PR #1744, opened 2021-05-13
[Bulk mode] Limit the size of bulker

shi-su:
**What I did**
Limit the size of the bulker.

**Why I did it**
Without a constraint on the size of bulkers, the bulked operation may take longer than the SAI timeout to finish. Such behavior would result in errors and swss crash. This PR limits the bulk size and avoids SAI timeout in bulk mode.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/bulker.h
+++ b/orchagent/bulker.h
@@ -288,12 +288,8 @@ public:
                     }
                 }
             }
+            flush_removing_entries(rs);
 
-            if (!rs.empty())
-            {
-                flush_removing_entries(rs);
-            }
-            
             removing_entries.clear();
         }
 
@@ -321,12 +317,8 @@ public:
                     }
                 }
             }
+            flush_creating_entries(rs, tss, cs);
 
-            if (!rs.empty())
-            {
-                flush_creating_entries(rs, tss, cs);
-            }
-            
             creating_entries.clear();
         }
 
@@ -358,12 +350,8 @@ public:
                     }
                 }
             }
+            flush_setting_entries(rs, ts, status_vector);
 
-            if (!rs.empty())
-            {
-                flush_setting_entries(rs, ts, status_vector);
-            }
-            
             setting_entries.clear();
         }
     }
@@ -426,6 +414,10 @@ private:
     sai_status_t flush_removing_entries(
         _Inout_ std::vector<Te> &rs)
     {
+        if (rs.empty())
+        {
+            return SAI_STATUS_SUCCESS;
+        }
         size_t count = rs.size();
         std::vector<sai_status_t> statuses(count);
         sai_status_t status = (*remove_entries)((uint32_t)count, rs.data(), SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR, statuses.data());
@@ -459,6 +451,10 @@ private:
         _Inout_ std::vector<sai_attribute_t const*> &tss,
         _Inout_ std::vector<uint32_t> &cs)
     {
+        if (rs.empty())
+        {
+            return SAI_STATUS_SUCCESS;
+        }
         size_t count = rs.size();
         std::vector<sai_status_t> statuses(count);
         sai_status_t status = (*create_entries)((uint32_t)count, rs.data(), cs.data(), tss.data()
@@ -495,6 +491,10 @@ private:
         _Inout_ std::vector<sai_attribute_t> &ts,
         _Inout_ std::vector<sai_status_t*> &status_vector)
     {
+        if (rs.empty())
+        {
+            return SAI_STATUS_SUCCESS;
+        }
         size_t count = rs.size();
         std::vector<sai_status_t> statuses(count);
         sai_status_t status = (*set_entries_attribute)((uint32_t)count, rs.data(), ts.data()
@@ -641,11 +641,7 @@ public:
                     }
                 }
             }
-
-            if (!rs.empty())
-            {
-                flush_removing_entries(rs);
-            }
+            flush_removing_entries(rs);
 
             removing_entries.clear();
         }
@@ -673,11 +669,7 @@ public:
                     }
                 }
             }
-
-            if (!rs.empty())
-            {
-                flush_creating_entries(rs, tss, cs);
-            }
+            flush_creating_entries(rs, tss, cs);
 
             creating_entries.clear();
         }
@@ -705,11 +697,7 @@ public:
                     }
                 }
             }
-
-            if (!rs.empty())
-            {
-                flush_setting_entries(rs, ts);
-            }
+            flush_setting_entries(rs, ts);
 
             setting_entries.clear();
         }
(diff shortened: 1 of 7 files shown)


orchagent/bulker.h
```diff
@@ -280,29 +282,19 @@ class EntityBulker
...
-                SWSS_LOG_ERROR("EntityBulker.flush remove entries failed, number of entries to remove: %zu, status: %s",
-                                removing_entries.size(), sai_serialize_status(status).c_str());
-            }
 
-            for (size_t ir = 0; ir < count; ir++)
-            {
-                auto& entry = rs[ir];
-                sai_status_t *object_status = removing_entries[entry];
-                if (object_status)
+                if (rs.size() >= MAX_BULK_SIZE)
```
qiluo-msft, line 286: >if [](http://example.com/codeflow?start=16&length=2)

You can move this if-block into inside of above if-block

orchagent/bulker.h
```diff
@@ -455,6 +429,93 @@ class EntityBulker
...
+            sai_status_t *object_status = removing_entries[entry];
+            if (object_status)
+            {
+                *object_status = statuses[ir];
+            }
+        }
+    }
+
+    void flush_creating_entries(
+        _In_ std::vector<Te> rs,
```
qiluo-msft, line 461: >vector [](http://example.com/codeflow?start=18&length=6)

Copy a vector is too heavy. Could you use reference?

orchagent/bulker.h
```diff
@@ -372,31 +355,22 @@ class EntityBulker
                         ts.push_back(attr);
                         status_vector.push_back(object_status);
                     }
+
+                    if (rs.size() >= MAX_BULK_SIZE)
+                    {
+                        flush_setting_entries(rs, ts, status_vector);
+                        rs.clear();
+                        ts.clear();
+                        status_vector.clear();
```
qiluo-msft, line 364: If you use reference in above function, you can move the 3 clear() into the above function.

orchagent/bulker.h
```diff
@@ -455,6 +429,93 @@ class EntityBulker
...
     typename Ts::bulk_remove_entry_fn                       remove_entries;
     typename Ts::bulk_set_entry_attribute_fn                set_entries_attribute;
+
+    void flush_removing_entries(
+        _In_ std::vector<Te> rs)
+    {
+        size_t count = rs.size();
+        std::vector<sai_status_t> statuses(count);
+        sai_status_t status = (*remove_entries)((uint32_t)count, rs.data(), SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR, statuses.data());
+        if (status == SAI_STATUS_SUCCESS)
```
qiluo-msft, line 439: >status [](http://example.com/codeflow?start=12&length=6)

The status if useful, let's return it.

orchagent/bulker.h
```diff
@@ -280,29 +282,19 @@ class EntityBulker
...
-            for (size_t ir = 0; ir < count; ir++)
-            {
-                auto& entry = rs[ir];
-                sai_status_t *object_status = removing_entries[entry];
-                if (object_status)
+                if (rs.size() >= MAX_BULK_SIZE)
                 {
-                    *object_status = statuses[ir];
+                    flush_removing_entries(rs);
+                    rs.clear();
```
qiluo-msft, line 289: >rs.clear(); [](http://example.com/codeflow?start=20&length=11)

If you use reference in above function, you can move the clear() into the above function.

orchagent/bulker.h
```diff
@@ -280,29 +282,19 @@ class EntityBulker
...
                 }
             }
+
+            if (!rs.empty())
```
qiluo-msft, line 293: >if (!rs.empty()) [](http://example.com/codeflow?start=12&length=16)

you can move the check into flush function.

orchagent/bulker.h
```diff
@@ -11,6 +11,8 @@
 #include "logger.h"
 #include "sai_serialize.h"
 
+#define MAX_BULK_SIZE 1000
```
lguohan, line 14: can we make this configurable? maybe via cmd line option or configuration file?

orchagent/bulker.h
shi-su: Moved it to the if-block above.

orchagent/bulker.h
shi-su: Passing reference instead.

orchagent/bulker.h
shi-su: Moved the clear() into the function.

orchagent/bulker.h
shi-su: Moved the clear() into the function.

orchagent/bulker.h
shi-su: Updated the functions to return the status.

orchagent/bulker.h
shi-su: Moved the check for empty into the flush function.

orchagent/bulker.h
shi-su: Added a command-line option to configure the max bulk size.

orchagent/main.cpp
```diff
@@ -397,6 +401,9 @@ int main(int argc, char **argv)
                 sairedis_rec_filename = optarg;
             }
             break;
+        case 'k':
+            gMaxBulkSize = atoi(optarg);
```
qiluo-msft, line 405: >gMaxBulkSize [](http://example.com/codeflow?start=12&length=12)

Please handle error and corner cases.

orchagent/bulker.h
```diff
@@ -11,6 +11,8 @@
 #include "logger.h"
 #include "sai_serialize.h"
 
+extern size_t gMaxBulkSize;
```
qiluo-msft, line 14: >gMaxBulkSize [](http://example.com/codeflow?start=14&length=12)

Better to define in below classes, also its default value.

tests/mock_tests/mock_orchagent_main.h
```diff
@@ -25,6 +25,7 @@ extern bool gLogRotate;
 extern bool gSaiRedisLogRotate;
 extern ofstream gRecordOfs;
 extern string gRecordFile;
+extern size_t gMaxBulkSize;
```
qiluo-msft, line 28: >gMaxBulkSize [](http://example.com/codeflow?start=14&length=12)

Try remove global var from test.

shi-su: Removed from tests.

orchagent/main.cpp
shi-su: Added a logic to handle errors and corner cases.

orchagent/bulker.h
shi-su: Defined a member variable in the classes for the max bulk size and setting the value in the orch that creates the bulker. As the default is currently global, the default value is not added to the class.

qiluo-msft (approved):

yxieca:
@shi-su  please help open PR for 202012 branch. Thanks.

PR #1747, opened 2021-05-17
[sonic-swss:TPID CONFIG]Added Orchagent TPID config support for Port/LAG

gechiang:
Signed-off-by: Gen-Hwa Chiang <gechiang@microsoft.com>

**What I did**
This PR is part of the TPID CONFIG SONiC feature set to support port/LAG TPID configuration.  
TPID HLD PR: (https://github.com/Azure/SONiC/pull/681)

There are total of 3 PRs across different submodules to achieve this feature support.

- SONiC-Utilities (https://github.com/Azure/sonic-utilities/pull/1618) - To handle TPID CLI related changes.
- SONiC Yang Model and Minigraph default (https://github.com/Azure/sonic-buildimage/pull/7630) - for the new port/LAG TPID attribute and Minigraph to configDB changes.
- This PR -  To handle Orchagent changes triggered from TPID configuration changes to interact with SYNCd SAI.

Please NOTE this PR has a  functional dependency on (https://github.com/Azure/sonic-buildimage/pull/7630)

Please refer to the TPID HLD listed above for more details on the TPID configuration feature support.
Here is a quick summary for this PR changes:

1. By default TPID for port/LAG is 0x8100.  With HW SKU that has capability to handle TPID configuration user is allowed to configure the port/LAG to non default TPID values such as (0x9100, 0x9200, 0x88A8). 
2. Orchagent is responsible to query HW SKU TPID attribute capability and store this information onto the StateDB so that the CLI handler can first validate if HW is capable before allowing any TPID CLI configuration to take place.
3. Orchagent listens to CLI cionfiguration changes and performs the SAI API call to perform the TPID attribute setting on the Port/LAG accordingly.

**Why I did it**
Having the capability to set TPID configuration it allows the HW to ignore the VALN TAG field if the TPID of received packet does not match the configured TPID value and treat it as Untagged packet.  This allows many desired feature such as the fanout switch configuration in  SONiC DUT PTF testing topology as well as potential 802.1Q Tunneling feature.

**How I verified it**
See the HLD for details on all the testing requirement.
Added unit testing on existing unit test modules (port and portchannel) to include the TPID config validation testing.

**Details if related**

### DIFF
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -415,8 +415,7 @@ bool TeamMgr::setLagTpid(const string &alias, const string &tpid)
     fvs.push_back(fv);
     m_appLagTable.set(alias, fvs);
 
-    SWSS_LOG_NOTICE("Set port channel %s TPID to %s",
-            alias.c_str(), tpid.c_str());
+    SWSS_LOG_NOTICE("Set port channel %s TPID to %s", alias.c_str(), tpid.c_str());
 
     return true;
 }
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -145,6 +145,7 @@ void TeamMgr::doLagTask(Consumer &consumer)
             string admin_status = DEFAULT_ADMIN_STATUS_STR;
             string mtu = DEFAULT_MTU_STR;
             string learn_mode;
+            string tpid;
 
             for (auto i : kfvFieldsValues(t))
             {
@@ -178,6 +179,11 @@ void TeamMgr::doLagTask(Consumer &consumer)
                     SWSS_LOG_INFO("Get learn_mode %s",
                             learn_mode.c_str());
                 }
+                else if (fvField(i) == "tpid")
+                {
+                    tpid = fvValue(i);
+                    SWSS_LOG_INFO("Get TPID %s", tpid.c_str());
+                 }
             }
 
             if (m_lagList.find(alias) == m_lagList.end())
@@ -197,6 +203,11 @@ void TeamMgr::doLagTask(Consumer &consumer)
             {
                 setLagLearnMode(alias, learn_mode);
                 SWSS_LOG_NOTICE("Configure %s MAC learn mode to %s", alias.c_str(), learn_mode.c_str());
+            }
+            if (!tpid.empty())
+            {
+                setLagTpid(alias, tpid);
+                SWSS_LOG_NOTICE("Configure %s TPID to %s", alias.c_str(), tpid.c_str());
             }
         }
         else if (op == DEL_COMMAND)
@@ -394,6 +405,22 @@ bool TeamMgr::setLagMtu(const string &alias, const string &mtu)
 
     return true;
 }
+
+bool TeamMgr::setLagTpid(const string &alias, const string &tpid)
+{
+    SWSS_LOG_ENTER();
+
+    vector<FieldValueTuple> fvs;
+    FieldValueTuple fv("tpid", tpid);
+    fvs.push_back(fv);
+    m_appLagTable.set(alias, fvs);
+
+    SWSS_LOG_NOTICE("Set port channel %s TPID to %s",
+            alias.c_str(), tpid.c_str());
+
+    return true;
+}
+
 
 bool TeamMgr::setLagLearnMode(const string &alias, const string &learn_mode)
 {
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -43,6 +43,9 @@ MACsecOrch *gMacsecOrch;
 
 bool gIsNatSupported = false;
 
+#define DEFAULT_MAX_BULK_SIZE 1000
+size_t gMaxBulkSize = DEFAULT_MAX_BULK_SIZE;
+
 OrchDaemon::OrchDaemon(DBConnector *applDb, DBConnector *configDb, DBConnector *stateDb, DBConnector *chassisAppDb) :
         m_applDb(applDb),
         m_configDb(configDb),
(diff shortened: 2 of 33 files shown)


cfgmgr/teammgr.cpp
```diff
@@ -395,6 +406,22 @@ bool TeamMgr::setLagMtu(const string &alias, const string &mtu)
...
+bool TeamMgr::setLagTpid(const string &alias, const string &tpid)
+{
+    SWSS_LOG_ENTER();
+
+    vector<FieldValueTuple> fvs;
+    FieldValueTuple fv("tpid", tpid);
+    fvs.push_back(fv);
+    m_appLagTable.set(alias, fvs);
+
+    SWSS_LOG_NOTICE("Set port channel %s TPID to %s",
```
prsunny, line 418: Please have it in same line

orchagent/orchdaemon.cpp
```diff
@@ -275,6 +280,56 @@ bool OrchDaemon::init()
 
     gMacsecOrch = new MACsecOrch(m_applDb, m_stateDb, macsec_app_tables, gPortsOrch);
   
+    Table m_switchTable(stateDbSwitchTable.first, stateDbSwitchTable.second);
+    // Check if SAI is capable of handling TPID config and store result in StateDB switch capability table
+    {
```
prsunny, line 285: Please move this section to switch init and make this part of switchorch class.

orchagent/orchdaemon.cpp
```diff
@@ -275,6 +280,56 @@ bool OrchDaemon::init()
...
+            {
+                fvVector.emplace_back(SWITCH_CAPABILITY_TABLE_LAG_TPID_CAPABLE, "true");
+            }
+            else
+            {
+                fvVector.emplace_back(SWITCH_CAPABILITY_TABLE_LAG_TPID_CAPABLE, "false");
+            }
+            SWSS_LOG_NOTICE("LAG TPID capability %d", capability.set_implemented);
+        }
+        m_switchTable.set("switch", fvVector);
```
prsunny, line 330: You can use the API `SwitchOrch::set_switch_capability`

orchagent/portsorch.cpp
```diff
@@ -2699,6 +2740,22 @@ void PortsOrch::doPortTask(Consumer &consumer)
                     }
                 }
 
+                if (tpid != 0 && tpid != p.m_tpid)
```
prsunny, line 2743: How do we reset? I think the check should be just `tpid != p.m_tpid`, correct?

cfgmgr/teammgr.cpp
gechiang: Will change the logging to same line.

orchagent/portsorch.cpp
gechiang: In this case "tpid !=0" check serves the purpose of identifying whether it is handling non-TPID operations in which case variable tpid is initialized to "0" and there is no need to perform TPID operation.  This is inline with how MTU and speed setting is being handled.  Hope this explains why this check is required.

orchagent/orchdaemon.cpp
gechiang: Got it. Will change this per your suggestion.

orchagent/orchdaemon.cpp
gechiang: Changed per suggestion in my new commit.

orchagent/orchdaemon.cpp
gechiang: Changed per suggestion in my new commit.

orchagent/portsorch.cpp
```diff
@@ -2384,6 +2416,15 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 {
                     mtu = (uint32_t)stoul(fvValue(i));
                 }
+                /* Set port TPID */
+                if (fvField(i) == "tpid")
+                {
+                    tpid_string = fvValue(i);
+                    // Need to get rid of the leading 0x
+                    tpid_string.erase(0,2);
```
prsunny, line 2424: Is it better to use str.find and then str.erase(0, pos)? Do we always expect a `0x`?

orchagent/portsorch.cpp
```diff
@@ -917,6 +917,36 @@ bool PortsOrch::setPortMtu(sai_object_id_t id, sai_uint32_t mtu)
     return true;
 }
 
+
+bool PortsOrch::setPortTpid(sai_object_id_t id, sai_uint16_t tpid)
```
prsunny, line 921: Can we combine this with Lag to have single function and pass `is_lag `as boolean since only two lines are different and rest are the same? Just a suggestion.

tests/test_port.py
```diff
@@ -6,6 +6,51 @@
...
+        cdb = swsscommon.DBConnector(4, dvs.redis_sock, 0)
+
+        # check application database for default TPID before set operation
+        pdb_port_tbl = swsscommon.Table(pdb, "PORT_TABLE")
+        (status, fvs) = pdb_port_tbl.get("Ethernet8")
+        assert status == True
+        tpid = "0x0000"
+        for fv in fvs:
+            if fv[0] == "tpid":
+                tpid = fv[1]
```
prsunny, line 21: By default, there is no `tpid `field in the port table until you set, correct? IMO, this check doesn't serve a purpose

prsunny (approved):
lgtm, minor comments

orchagent/portsorch.cpp
gechiang: The CLI enforces that the TPID values must be one of the following: "0x8100", "0x9100", "0x9200", or "0x88A8".
So yes it will always have the "0x" which we can get rid of.

orchagent/portsorch.cpp
gechiang: There is a chance that we may end up having to implement the LAG handling by having Orchagent apply the LAG TPID config to all its LAG members.  This is because currently only one SAI vendor supports TPID setting on Port only and not yet for LAG.  For our fanout switch usage, this is already good enough.  But if we ever want to apply it to LAG while SAI vendor not supporting it, it will be much easier to have the port/LAG handling having its own function to do the handling.  So for now, let's keep the way it is just in case the enhancement is needed in the future.

prsunny: ok

tests/test_port.py
gechiang: sure. let me take it out.

prsunny (dismissed):

prsunny (approved):

PR #1749, opened 2021-05-19
[flex-counters] Delay flex counters stats init for faster boot time

shlomibitton:

**What I did**
Update flex counters DB with counters stats only when counters are enabled.
As long as the polling counters are not enabled, flex counters information will stored internally on PortsOrch.

**Why I did it**
Creating flex counters objects on the DB will trigger 'SYNCD' to access the HW for query statistics capabilities.
This HW access takes time and will be better to finish boot before doing this (mainly for fast-reboot but good to have in general).
The flex counters are not crucial at boot time, we can delay it to the end of the boot process.

**How I verified it**
Reboot a switch and observer the flex counters DB populated after counters are enabled.

**Details if related**

### DIFF
--- a/orchagent/flexcounterorch.cpp
+++ b/orchagent/flexcounterorch.cpp
@@ -18,6 +18,11 @@ extern IntfsOrch *gIntfsOrch;
 extern BufferOrch *gBufferOrch;
 
 #define BUFFER_POOL_WATERMARK_KEY   "BUFFER_POOL_WATERMARK"
+#define PORT_KEY                    "PORT"
+#define PORT_BUFFER_DROP_KEY        "PORT_BUFFER_DROP"
+#define QUEUE_KEY                   "QUEUE"
+#define PG_WATERMARK_KEY            "PG_WATERMARK"
+#define RIF_KEY                     "RIF"
 
 unordered_map<string, string> flexCounterGroupMap =
 {
@@ -94,6 +99,16 @@ void FlexCounterOrch::doTask(Consumer &consumer)
                 }
                 else if(field == FLEX_COUNTER_STATUS_FIELD)
                 {
+                    if(gPortsOrch && (key == PORT_KEY) && (value == "enable"))
+                    {
+                        gPortsOrch->generatePortCounterMap();
+                        m_port_counter_enabled = true;
+                    }
+                    if(gPortsOrch && (key == PORT_BUFFER_DROP_KEY) && (value == "enable"))
+                    {
+                        gPortsOrch->generatePortBufferDropCounterMap();
+                        m_port_buffer_drop_counter_enabled = true;
+                    }
                     // Currently, the counters are disabled for polling by default
                     // The queue maps will be generated as soon as counters are enabled for polling
                     // Counter polling is enabled by pushing the COUNTER_ID_LIST/ATTR_ID_LIST, which contains
@@ -108,16 +123,15 @@ void FlexCounterOrch::doTask(Consumer &consumer)
                     // This can be because generateQueueMap() installs a fundamental list of queue stats
                     // that need to be polled. So my doubt here is if queue watermark stats shall be piggybacked
                     // into the same function as they may not be counted as fundamental
-                    if(gPortsOrch)
+                    if(gPortsOrch && (key == QUEUE_KEY) && (value == "enable"))
                     {
                         gPortsOrch->generateQueueMap();
-                        gPortsOrch->generatePriorityGroupMap();
                     }
-                    if(gPortsOrch)
+                    if(gPortsOrch && (key == PG_WATERMARK_KEY) && (value == "enable"))
                     {
                         gPortsOrch->generatePriorityGroupMap();
                     }
-                    if(gIntfsOrch)
+                    if(gIntfsOrch && (key == RIF_KEY) && (value == "enable"))
                     {
                         gIntfsOrch->generateInterfaceMap();
                     }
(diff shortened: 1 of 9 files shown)


liat-grozovik:
@shlomibitton  please resolve conflicts for the review to start.

shlomibitton:
> @shlomibitton please resolve conflicts for the review to start.

@liat-grozovik done.

liat-grozovik:
@qiluo-msft kindly reminder. we need the fix merged in order to test fastboot and get some improvements.

qiluo-msft:
Compared with https://github.com/Azure/sonic-swss/pull/1646, what are changed?

shlomibitton:
> Compared with #1646, what are changed?

This is with the fix to the bug where PORT_STAT is empty on COUNTERS DB as a result of query port counter statistics from the CPU interface.
A unit test improvements added as well.
As you requested this PR include the previous changes, the fix and the unit tests.

orchagent/flexcounterorch.cpp
```diff
@@ -94,6 +99,16 @@ void FlexCounterOrch::doTask(Consumer &consumer)
                 }
                 else if(field == FLEX_COUNTER_STATUS_FIELD)
                 {
+                    if(gPortsOrch && (key == PORT_KEY) && (value == "enable"))
```
qiluo-msft, line 102: >if(gPortsOrch && (key == PORT_KEY) && (value == "enable")) [](http://example.com/codeflow?start=20&length=58)

You have many if-blocks checking `gPortsOrch && (value == "enable")`, can you check only once? and inside the block, you can use `switch`/`case` on `key`.

orchagent/flexcounterorch.cpp
```diff
@@ -144,3 +158,13 @@ void FlexCounterOrch::doTask(Consumer &consumer)
         consumer.m_toSync.erase(it++);
     }
 }
+
+bool FlexCounterOrch::getPortCountersState()
+    {
+        return m_port_counter_enabled;
+    }
```
qiluo-msft, line 165: Could you follow the indentation convention of nearby code?

orchagent/flexcounterorch.h
```diff
@@ -15,10 +15,14 @@ class FlexCounterOrch: public Orch
     void doTask(Consumer &consumer);
     FlexCounterOrch(swss::DBConnector *db, std::vector<std::string> &tableNames);
     virtual ~FlexCounterOrch(void);
+    bool getPortCountersState();
```
qiluo-msft, line 18: >; [](http://example.com/codeflow?start=31&length=1)

Add `const` if it is a `get` function.

tests/mock_tests/portsorch_ut.cpp
```diff
@@ -146,8 +146,13 @@ namespace portsorch_test
...
         ASSERT_EQ(gPortsOrch, nullptr);
 
-        gPortsOrch = new PortsOrch(m_app_db.get(), m_state_db.get(), ports_tables, m_chassis_app_db.get());
+        vector<string> flex_counter_tables = {
+            CFG_FLEX_COUNTER_TABLE_NAME
+        };
+        auto* flexCounterOrch = new FlexCounterOrch(m_config_db.get(), flex_counter_tables);
+        gDirectory.set(flexCounterOrch);
 
+	gPortsOrch = new PortsOrch(m_app_db.get(), m_state_db.get(), ports_tables, m_chassis_app_db.get());
```
qiluo-msft, line 155: >gPortsOrch [](http://example.com/codeflow?start=1&length=10)

wrong indentation level?

qiluo-msft (changes requested):
As comments

orchagent/flexcounterorch.cpp
shlomibitton: Fixed.

tests/mock_tests/portsorch_ut.cpp
shlomibitton: Fixed.

orchagent/flexcounterorch.h
shlomibitton: We can't declare this return value as const for this function since the bool we are returning is changing on run time.

orchagent/flexcounterorch.cpp
shlomibitton: Adding a switch/case here will require hashing to map string to int and I think it is making the code more complicated.
I modified it to have one check for "gPortsOrch && (value == "enable")" and inside if/else for the different cases.

orchagent/flexcounterorch.h
qiluo-msft:
```suggestion
    bool getPortCountersState() const;
```
Not the return value, but the whole function

shlomibitton: Got you, fixed

qiluo-msft (approved):

orchagent/portsorch.cpp
```diff
@@ -2221,19 +2221,21 @@ bool PortsOrch::initPort(const string &alias, const int index, const set<int> &l
                 vector<FieldValueTuple> fields;
                 fields.push_back(tuple);
                 m_counterTable->set("", fields);
+
                 // Install a flex counter for this port to track stats
-                std::unordered_set<std::string> counter_stats;
-                for (const auto& it: port_stat_ids)
+                auto flex_counters_orch = gDirectory.get<FlexCounterOrch*>();
+                /* Delay installing the counters if they are yet enabled
```
qiluo-msft, line 2227: >Delay [](http://example.com/codeflow?start=19&length=5)

How long do you delay? I cannot find the constant in code. How did you test the delay?

shlomibitton: @qiluo-msft please check this file:
https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-orchagent/enable_counters.py#L55
it will enable the counters on CONFIG DB, we handle this change here:
https://github.com/Azure/sonic-swss/pull/1749/files#diff-77a2299548b8c0b1cfa565b28cdc46051dbb6c325d91011fab5fe9df283a03c8R112
and then update m_port_counter_enabled = true
If the counter is already enabled, the normal flow will take place like it used to be before.

PR #1752, opened 2021-05-19
[nhg]: Add support for weight in nexthop group member.

caizhenghui-juniper:
**What I did**
1. In fpmsyncd, parse weight field in netlink message, set APP_DB if weight is set.
2. In routeorch, collect weight and pass weight attribute to next hop group member object.

**Why I did it**
In order to set weight for next hop group members. To support weighted ECMP.
**How I verified it**
1. With cRPD as routing stack, configure bgp weight policy to different next hops. 
2. Bring up routes on both next hops. And send traffic to the BGP routes.
3. Verify traffic is distributed with the portion set by weight policy.
**Details if related**

### DIFF
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -982,11 +982,11 @@ string RouteSync::getNextHopWt(struct rtnl_route *route_obj)
     for (int i = 0; i < rtnl_route_get_nnexthops(route_obj); i++)
     {
         struct rtnl_nexthop *nexthop = rtnl_route_nexthop_n(route_obj, i);
-        /* Get the ID of next hop interface */
-        uint8_t if_weight = rtnl_route_nh_get_weight(nexthop);
-        if (if_weight)
+        /* Get the weight of next hop */
+        uint8_t weight = rtnl_route_nh_get_weight(nexthop);
+        if (weight)
         {
-            result += to_string(if_weight + 1);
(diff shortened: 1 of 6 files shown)


fpmsyncd/routesync.cpp
```diff
@@ -962,3 +968,36 @@ string RouteSync::getNextHopIf(struct rtnl_route *route_obj)
...
+    string result = "";
+
+    for (int i = 0; i < rtnl_route_get_nnexthops(route_obj); i++)
+    {
+        struct rtnl_nexthop *nexthop = rtnl_route_nexthop_n(route_obj, i);
+        /* Get the ID of next hop interface */
+        uint8_t if_weight = rtnl_route_nh_get_weight(nexthop);
+        if (if_weight)
+        {
+            result += to_string(if_weight + 1);
```
prsunny, line 989: please rename to `weight`. `if_weight `is confusing. Also, why is it incremented by 1?

prsunny (commented):
Please add VS tests

fpmsyncd/routesync.cpp
caizhenghui-juniper: @prsunny Incrementing by 1 is to agree with kernel behavior(ip route show). I'll change if_weight to weight

prsunny (approved):

shi-su (approved):

PR #1756, opened 2021-05-25
[tunneldecaporch] Set default MTU for the overlay loopback interface

volodymyrsamotiy:

**What I did**
Set default MTU for the underlay loopback interface to 9100 in order to make it working with jumbo frames.

**Why I did it**
Overlay loopback is used for IP-in-IP tunnel and without setting MTU jumbo frames are trapped to CPU but they should be successfully routed to the appropriate destination.
With this change, all internal loopback interfaces are created with 9100 MTU size (both underlay and overlay) in continues to PR https://github.com/Azure/sonic-swss/pull/1299

**How I verified it**
* Configure IP-in-IP tunnel
* Send jumbo frames and verify that they are correctly routed

**Details if related**
N/A

### DIFF
--- a/orchagent/tunneldecaporch.cpp
+++ b/orchagent/tunneldecaporch.cpp
@@ -6,6 +6,8 @@
 #include "logger.h"
 #include "swssnet.h"
 
+#define OVERLAY_RIF_DEFAULT_MTU 9100
+
 extern sai_tunnel_api_t* sai_tunnel_api;
 extern sai_router_interface_api_t* sai_router_intfs_api;
 extern sai_next_hop_api_t* sai_next_hop_api;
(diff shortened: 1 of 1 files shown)


liat-grozovik:
@abdosi  and @qiluo-msft it is cleanly be cherry picked to the different requested branches. Appreciate if you can take over once merged.

abdosi:
`Set default MTU for the underlay loopback interface to 9100 in order to make it working with jumbo frames.` Looks typo ?

prsunny (approved):

abdosi:
why is this needed for 201911 ? @prsunny

liat-grozovik:
A bug fix which is needed for sonic end user use case

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

abdosi:
@prsunny this is in 201911 now

PR #1765, opened 2021-06-03
[fpmsyncd]: Linux netlink runtime option

qbdwlr:
**What I did**
Added runtime option for fpmsyncd "-l net" to allow fpmsyncd to listen directly to netlink messages from Linux kernel rather than listening to FRR FPM TCP socket.

**Why I did it**
FRR currently does not transmit any MPLS related information on its FPM TCP socket to fpmsyncd, not even MPLS attributes on IP routes.  MPLS functionality verification had to be using alternate routing stacks (Juniper cRPD and MetaSwitch) but these routing stacks are not open-source and are not available to the community in general.  
Some unit-testing during MPLS development was performed by switching fpmsyncd to listen directly to Linux kernel messages, so it was proposed to upstream this capability for general use.

**How I verified it**
Unit-tests in sonic-swss/tests test_route.py and test_mpls.py were run with "fpmsyncd -l net".

**Details if related**
Main difference between default fpmsyncd and "fpmsyncd -l net" behavior is related to VRF support.  SONiC modifies actual netlink message from FRR to send VRF ifindex value in RTA_RT_TABLE_ID attribute, while Linux kernel netlink message is unmodified and contains the route table ID in RTA_RT_TABLE_ID attribute.  This prevents "-l net" option being used in scenarios that test routes with VRF.


fpmsyncd/routesync.cpp
```diff
@@ -581,7 +581,8 @@ void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
     char master_name[IFNAMSIZ] = {0};
 
     /* if the table_id is not set in the route obj then route is for default vrf. */
-    if (master_index)
+    if (master_index &&
+        (master_index != RT_TABLE_MAIN) && (master_index != RT_TABLE_DEFAULT))
```
abdosi, line 585: what is trigger of this change ?

qbdwlr: Main difference between default fpmsyncd and "fpmsyncd -l net" behavior is related to VRF support. SONiC modifies actual netlink message from FRR to send VRF ifindex value in RTA_RT_TABLE_ID attribute, while Linux kernel netlink message is unmodified and contains the route table ID in RTA_RT_TABLE_ID attribute. This prevents "-l net" option being used in scenarios that test routes with VRF.

qbdwlr: @abdosi I can change this logic to just check a boolean to see if Linux Netlink connection is used and skip VRF if it is.

PR #1769, opened 2021-06-05
[orchagent] Put port configuration to APPL_DB according to autoneg mode

Junchao-Mellanox:
**What I did**

Before puting port related configuration to APPL_DB, check autoneg mode first. If autoneg mode is enabled, "speed" and "interface_type" will not be put into APPL_DB; if autoneg mode is disabled, "adv_speeds" and "adv_interface_types" will not be put into APPL_DB; else all configuration will be put to APPL_DB for backward compatible.

**Why I did it**

The field "speed" in PORT_TABLE of APPL_DB has two meanings:
1. The configured port speed
2. The actual port speed

This is all right when SONiC has no auto negotiation feature because the actual speed must equal to the configured speed or the operational status is down. However, with port auto negotiation feature, if autoneg mode is enabled, the actual speed could be a value of adv_speeds which might not equal to field "speed". Now port auto negotiation feature handle it this way: 
1. if autoneg mode  is disabled, everything goes to the old way, the actual speed is always equal to field "speed"; 
2. if autoneg mode is enabled, and if operational status is up, get actual speed from SAI and update the field "speed" in APPL_DB.

For case#2, there is an issue. Think about following flow:
1. Set Ethernet0 speed=400000
2. Set Ethernet0 autoneg=on, adv_speeds=100000, and Ethernet0 is operational up. Now the speed of Ethernet0 in "show interfaces status" is 100G because orchagent get the actual speed from SAI. This is right.
3. Issue command: "config interface type Ethernet0 CR".

Step#3 won't affect the auto negotiation status because the autoneg mode is enabled, force interface type will not be applied to SAI. But it triggers PORT_TABLE change and portsyncd will push all Ethernet0 related configuration to APPL_DB. So, speed of Ethernet0 in "show interface status" changed to 400G.

This PR is to fix the issue.

**How I verified it**

Manual test

**Details if related**

### DIFF
--- a/portsyncd/portsyncd.cpp
+++ b/portsyncd/portsyncd.cpp
@@ -235,6 +235,10 @@ bool handlePortConfigFromConfigDB(ProducerStateTable &p, DBConnector &cfgDb, boo
 
 void handlePortConfig(ProducerStateTable &p, map<string, KeyOpFieldsValuesTuple> &port_cfg_map)
 {
+    string autoneg;
+    vector<FieldValueTuple> attrs;
+    vector<FieldValueTuple> autoneg_attrs;
+    vector<FieldValueTuple> force_attrs;
 
     auto it = port_cfg_map.begin();
     while (it != port_cfg_map.end())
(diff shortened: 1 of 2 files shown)


prsunny:
> /azp run

@jleveque , there is a build failure while downloading swss-common. Tracking it with build team.

portsyncd/portsyncd.cpp
```diff
@@ -250,7 +254,54 @@ void handlePortConfig(ProducerStateTable &p, map<string, KeyOpFieldsValuesTuple>
...
+                else if (autoneg == "off") // autoneg is off, only put speed and interface_type to APPL_DB
+                {
+                    attrs.insert(attrs.end(), force_attrs.begin(), force_attrs.end());
+                }
+                else // autoneg is not configured, put all attributes to APPL_DB
+                {
+                    attrs.insert(attrs.end(), autoneg_attrs.begin(), autoneg_attrs.end());
+                    attrs.insert(attrs.end(), force_attrs.begin(), force_attrs.end());
+                }
+                p.set(key, attrs);
```
prsunny, line 300: How is this handled during warmboot?

Junchao-Mellanox: No special treatment for warmboot

Junchao-Mellanox:
Hi @prsunny , it seems 2 test cases related to flex counter failed which should not relate to my change. Could you please help re-trigger the test?

liat-grozovik:
@prsunny can you approve the PR so once the tests are passing we can merge?

liat-grozovik:
@prsunny kindly reminder, can you please approve and merge? now that is is passing the tests.

PR #1772, opened 2021-06-07
[orchagent] Remove default prefix in non default vrf

msosyak:
**What I did**

1. Remove default prefix in non-default VRF instead of setting it to blackhole
2. Initialize `m_overlay_nexthops` in overloaded constructor. 

**Why I did it**
To fix https://github.com/Azure/sonic-buildimage/issues/6504

1. To remove default routes in non default VRF
2. To avoid garbage in `m_overlay_nexthops`. 

**How I verified it**
1. Add t0 topology
2. Load config generated by this test vrf/test_vrf.py::TestVrfDeletion::test_vrf1_neigh_after_restore
3. `sudo config vrf del Vrf1`
4. `redis-cli -n 6 keys "*VRF*"`
5. check that there is entries only for Vrf2. e.g:
```
1) "VRF_OBJECT_TABLE|Vrf2"
2) "VRF_TABLE|Vrf2"
```

**Details if related**

### DIFF
--- a/orchagent/nexthopgroupkey.h
+++ b/orchagent/nexthopgroupkey.h
@@ -33,6 +33,7 @@ public:
 
     NextHopGroupKey(const std::string &nexthops, const std::string &weights)
     {
+        m_overlay_nexthops = false;
         std::vector<std::string> nhv = tokenize(nexthops, NHG_DELIMITER);
         std::vector<std::string> wtv = tokenize(weights, NHG_DELIMITER);
         for (uint32_t i = 0; i < nhv.size(); i++)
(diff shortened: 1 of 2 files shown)


msosyak:
These changes fix Azure/sonic-buildimage#6504  partially.

I still observe that some routes in VRF are not deleted.

msosyak:
@prsunny  Could you have a look, please?

@iris00522 I see you fixed a ref count issue previously. Could you have a look, please, too?

prsunny (approved):
lgtm

PR #1776, opened 2021-06-09
[Dynamic Buffer Calc] Avoid creating lossy PG for admin down ports during initialization

stephenxs:
**What I did**
- Avoid creating lossy PGs for admin down ports during system initialization
- Make ADMIN_DOWN the default state of a port in dynamic buffer manager.
- Explicitly remove configured PGs from an admin down port during initialization
  This is to make sure the items will be added to m_ready_list.
- Use APPL_DB as the source of initial PGs and queues for port ready deducting for warm reboot

**Why I did it**
Lossy PGs for admin down ports should not be applied to `APPL_DB`.
Originally, we had the logic to remove lossy/lossless PGs when shutting a port. The same logic should be applied when the system is starting. Allocating lossy PG for an admin down port doesn't affect the system negatively except for wasting buffer. So we would like to resolve it.

**How I verified it**
Manually test

**Details if related**
1. A port is admin down if `admin_status` in `CONFIG_DB|PORT|<port>` is `down` or doesn't exist. In case it doesn't exist, `buffermgrd` doesn't have a chance to read and initialize admin status in its internal data structure. To simplify the flow, we make `PORT_ADMIN_DOWN` the default state of the port.
2. There is a logic in orchagent to make sure all configured PGs have been handled before the port initialization can proceed. The idea is to add check whether all configured PGs have been applied to SAI. As no PG will be applied to SAI on admin down ports, removing the PGs will notify the orchagent that the PGs have been handled.

### DIFF
--- a/cfgmgr/buffermgrdyn.h
+++ b/cfgmgr/buffermgrdyn.h
@@ -79,12 +79,12 @@ typedef struct {
 } buffer_pg_t;
 
 typedef enum {
+    // Port is admin down. All PGs programmed to APPL_DB should be removed from the port
+    PORT_ADMIN_DOWN,
     // Port is under initializing, which means its info hasn't been comprehensive for calculating headroom
     PORT_INITIALIZING,
     // All necessary information for calculating headroom is ready
-    PORT_READY,
-    // Port is admin down. All PGs programmed to APPL_DB should be removed from the port
-    PORT_ADMIN_DOWN
+    PORT_READY
 } port_state_t;
 
 typedef struct {
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -2,6 +2,7 @@
 #include "bufferorch.h"
 #include "logger.h"
 #include "sai_serialize.h"
+#include "warm_restart.h"
 
 #include <inttypes.h>
 #include <sstream>
@@ -45,7 +46,7 @@ BufferOrch::BufferOrch(DBConnector *applDb, DBConnector *confDb, DBConnector *st
 {
     SWSS_LOG_ENTER();
     initTableHandlers();
-    initBufferReadyLists(confDb);
+    initBufferReadyLists(confDb, applDb);
     initFlexCounterGroupTable();
     initBufferConstants();
 };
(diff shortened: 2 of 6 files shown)


dprital:
This fix is required also for 202012. Need to add the tag for it,. Thanks.

stephenxs:
> This fix is required also for 202012. Need to add the tag for it,. Thanks.

Done

liat-grozovik:
@neethajohn please help to review

orchagent/portsorch.cpp
```diff
@@ -2732,9 +2732,11 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 continue;
             }
 
-            if (!gBufferOrch->isPortReady(alias))
+            if (!gBufferOrch->isPortReady(alias) && (admin_status == "up"))
```
neethajohn, line 2735: as discussed, why is this check needed?  BUFFER_PG and BUFFER_QUEUE tables only include active ports and isPortReady() will return true for admin down ports

stephenxs: I understood the confusion.
In `isPortReady`, a port can be ready if one of the following conditions is satisfied:
- it's not in the list `m_port_ready_list_ref`, meaning there is not any `BUFFER_PG` item configured on the port. The `m_port_ready_list_ref` contains all the keys configured in `BUFFER_PG`
- it's in `m_ready_list`, meaning there are some configurations in the above table and all items in `BUFFER_PG` have been deployed to SAI.

Now that no buffer PG will be deployed to SAI if the port is admin down, which means it won't be `ready` until the port is admin up, we need to handle the case by testing whether the admin_status is `up`.
During the discussion, the `m_ready_list` wasn't found in the code somehow.

neethajohn: I have some concerns regarding the lossless profiles. Currently they get applied to even admin down ports. With this check, we end up removing that port from pendingPortSet even if the profile hasn't been pushed down to SAI

stephenxs: Fixed.

stephenxs:
Vs test failed due to environmental issue: “installing dependence “

neethajohn (dismissed):

liat-grozovik:
@neethajohn should we wait for additional reviewer or can merge?

neethajohn:
@stephenxs , can you please update the PR description?  We do not have the 'isPortReady' change now

stephenxs:
> @stephenxs , can you please update the PR description? We do not have the 'isPortReady' change now

Thanks for reminding. Done.

stephenxs:
Please don't merge for now. The latest commit introduces a further issue during warm-reboot. I'm checking and turned it to draft for now.

orchagent/bufferorch.cpp
```diff
@@ -61,31 +62,61 @@ void BufferOrch::initTableHandlers()
     m_bufferHandlerMap.insert(buffer_handler_pair(APP_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME, &BufferOrch::processEgressBufferProfileList));
 }
 
-void BufferOrch::initBufferReadyLists(DBConnector *db)
+void BufferOrch::initBufferReadyLists(DBConnector *applDb, DBConnector *confDb)
 {
-    // The motivation of m_ready_list is to get the preconfigured buffer pg and buffer queue items
-    // from the database when system starts.
-    // When a buffer pg or queue item is updated, if the item isn't in the m_ready_list
+    // Map m_ready_list and m_port_ready_list_ref are designed to track whether the ports are "ready" from buffer's POV
```
liat-grozovik, line 67: can you change it to a block of comment using /* instead of multiple lines of //?

cfgmgr/buffermgrdyn.h
```diff
@@ -79,12 +79,12 @@ typedef struct {
 } buffer_pg_t;
 
 typedef enum {
+    // Port is admin down. All PGs programmed to APPL_DB should be removed from the port
```
liat-grozovik, line 82: Why do you need to change order of the enum? does the enum number has meaning?

liat-grozovik (commented):
@stephenxs can you please confirm warmboot sonic-mgmt test is passing with this change as well?

cfgmgr/buffermgrdyn.h
stephenxs: The first enum value (0) is the default value after initialization.
Originally, the default state was `PORT_INITIALIZING`.
As sometimes the `admin_status` doesn't exist in `CONFIG_DB` if the port is admin down, we need to change the default state to `PORT_ADMIN` and change a port's state to another state after it is admin state is learned.

orchagent/bufferorch.cpp
stephenxs: Fixed

stephenxs:
> @stephenxs can you please confirm warmboot sonic-mgmt test is passing with this change as well?

Passed on Spectrum and Spectrum 3 platforms. I didn't test it on Spectrum 2 platform as there isn't an available test bed.
This is not platform independent so I think it's safe for warm reboot.

stephenxs:
The vs test failures have been fixed by PR Azure/sonic-swss#1807

liat-grozovik (approved):

neethajohn (approved):

PR #1783, opened 2021-06-11
[macsec] Fix acl entry id in updateMACsecPort

byu343:
**What I did**
ACL entry Id should be used as first parameter of setMACsecFlowActive. 

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/macsecorch.cpp
+++ b/orchagent/macsecorch.cpp
@@ -1106,9 +1106,9 @@ bool MACsecOrch::updateMACsecPort(MACsecPort &macsec_port, const TaskArgs &port_
                     SWSS_LOG_WARN("Cannot change the ACL entry action from packet action to MACsec flow");
                     return false;
                 }
-                auto an = macsec_sc->m_encoding_an;
+                auto entry_id = macsec_sc->m_entry_id;
                 auto flow_id = macsec_sc->m_flow_id;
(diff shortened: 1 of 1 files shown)


Pterosaur (changes requested):
I found there are same issues in other places using ` setMACsecFlowActive`
Could you also help to fix them?
E.G. 
https://github.com/Azure/sonic-swss/blob/a87ca108242701cf6e85aafa00fe35fde4f59b84/orchagent/macsecorch.cpp#L1111
https://github.com/Azure/sonic-swss/blob/a87ca108242701cf6e85aafa00fe35fde4f59b84/orchagent/macsecorch.cpp#L1697-L1702

byu343:
> I found there are same issues in other places using ` setMACsecFlowActive`
> Could you also help to fix them?
> E.G.
> https://github.com/Azure/sonic-swss/blob/a87ca108242701cf6e85aafa00fe35fde4f59b84/orchagent/macsecorch.cpp#L1111
> 
> 
> https://github.com/Azure/sonic-swss/blob/a87ca108242701cf6e85aafa00fe35fde4f59b84/orchagent/macsecorch.cpp#L1697-L1702

That is true. I have submitted another commit to fix it.

Pterosaur (approved):

PR #1785, opened 2021-06-14
[gearbox] Set port speed to SAI_PORT_ATTR_SPEED for gearbox

byu343:
**What I did**
Set port speed to SAI_PORT_ATTR_SPEED for gearbox system-side/line-side ports.  Lane speed was set before this change.

**Why I did it**
According to the example of "Creating ASIC Ports with 100G" in the SAI design doc: https://github.com/opencomputeproject/SAI/blob/master/doc/macsec-gearbox/SAI_Gearbox_API_Proposal-v1.0.docx, port speed should be set to SAI_PORT_ATTR_SPEED. However, the current code applies the lane speed to it. The SAI-based driver for the gearbox chip on Arista devices is assuming port speed for SAI_PORT_ATTR_SPEED, so we hope to submit this change to fix the issue we are facing.

**How I verified it**
Tested that traffic passing through the ports works and no related errors in syslog.

**Details if related**
1) If there exists other platforms implementing the gearbox SAI APIs, please help to check whether they are using port speed or lane speed on SAI_PORT_ATTR_SPEED. I think we need to follow the same definition.
2) Note that system_speed and line_speed in phy_config.json are still lane speed, following the design doc: https://github.com/Azure/SONiC/blob/master/doc/gearbox/gearbox_mgr_design.md.

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1800,9 +1800,7 @@ bool PortsOrch::setGearboxPortAttr(Port &port, dest_port_type_t port_type, sai_p
     sai_object_id_t dest_port_id;
     sai_attribute_t attr;
     string speed_attr;
-    uint32_t lane_speed = 0;
     sai_uint32_t speed = 0;
-    uint32_t lanes = 0;
 
     SWSS_LOG_ENTER();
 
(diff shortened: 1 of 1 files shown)


lguohan:
is there a unit test that cover this part? can you add swss vs test?

orchagent/portsorch.cpp
```diff
@@ -1839,23 +1835,19 @@ bool PortsOrch::setGearboxPortAttr(Port &port, dest_port_type_t port_type, sai_p
 
                     // Gearbox expects speed per lane
```
jimmyzhai, line 1836: Please remove the line of comment "// Gearbox expects speed per lane".

PR #1786, opened 2021-06-16
[BufferOrch] Don't call SAI API for BUFFER_POOL/PROFILE handling in case the op is DEL and the SAI OID is NULL

stephenxs:
**What I did**
Don't call SAI API for BUFFER_POOL/PROFILE handling in case the op is DEL and the SAI OID is NULL in order to avoid orchagent from exiting.
We need it only in 202106 or above. In 202012 the orchagent won't exit in such case.

**Why I did it**
Handle rare cases which cause SAI error eventually makes orchagent to exit.

**How I verified it**
Manually test.

**Details if related**
There is an optimization on orch: the SET event will be replaced by a DEL event if it is still pending in m_toSync when the DEL event is coming.
This is reasonable but can cause the SAI OID to be NULL in rare case:
 - The application creates an object and then destroy it after a very short period
 - The create notification is eliminated and replaced by destroying the notification
 - The SAI object hasn't been created and is NULL when it is removed

This causes SAI error which eventually makes orchagent exit.
We need to avoid it in 202106 and above
(In 202012, the orchagent won't exit so it's ok to ignore the error)
The solution is not to call SAI removing interface in case SAI OID is NULL

It can happen if a user configures something which causes the accumulative headroom to exceed the limit.
In this case, the buffer profile was created and then removed in a short time.

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -416,7 +416,7 @@ task_process_status BufferOrch::processBufferPool(KeyOpFieldsValuesTuple &tuple)
             return task_process_status::task_need_retry;
         }
 
-        if (SAI_NULL_OBJECT_ID == sai_object)
+        if (SAI_NULL_OBJECT_ID != sai_object)
         {
             sai_status = sai_buffer_api->remove_buffer_pool(sai_object);
             if (SAI_STATUS_SUCCESS != sai_status)
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -416,14 +416,17 @@ task_process_status BufferOrch::processBufferPool(KeyOpFieldsValuesTuple &tuple)
             return task_process_status::task_need_retry;
         }
 
-        sai_status = sai_buffer_api->remove_buffer_pool(sai_object);
-        if (SAI_STATUS_SUCCESS != sai_status)
+        if (SAI_NULL_OBJECT_ID == sai_object)
         {
-            SWSS_LOG_ERROR("Failed to remove buffer pool %s with type %s, rv:%d", object_name.c_str(), map_type_name.c_str(), sai_status);
-            task_process_status handle_status = handleSaiRemoveStatus(SAI_API_BUFFER, sai_status);
-            if (handle_status != task_process_status::task_success)
+            sai_status = sai_buffer_api->remove_buffer_pool(sai_object);
+            if (SAI_STATUS_SUCCESS != sai_status)
             {
-                return handle_status;
+                SWSS_LOG_ERROR("Failed to remove buffer pool %s with type %s, rv:%d", object_name.c_str(), map_type_name.c_str(), sai_status);
+                task_process_status handle_status = handleSaiRemoveStatus(SAI_API_BUFFER, sai_status);
+                if (handle_status != task_process_status::task_success)
+                {
+                    return handle_status;
+                }
             }
         }
         SWSS_LOG_NOTICE("Removed buffer pool %s with type %s", object_name.c_str(), map_type_name.c_str());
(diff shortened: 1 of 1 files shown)


stephenxs:
Looks like the LGTM failed due to some temporary network issue
```
[2021-06-16 23:08:24] [build-stdout] Get:51 http://storage.googleapis.com/lgtm-apt-mirror/ubuntu eoan/main amd64 dh-exec amd64 0.23.2 [25.0 kB]
[2021-06-16 23:08:24] [build-stdout] Fetched 12.9 MB in 1min 10s (184 kB/s)
[2021-06-16 23:08:24] [build-stderr] E: Failed to fetch http://storage.googleapis.com/lgtm-apt-mirror/ubuntu/pool/main/libs/libsodium/libsodium-dev_1.0.18-1_amd64.deb  Connection failed [IP: 173.194.198.128 80]
[2021-06-16 23:08:24] [build-stderr] E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
[2021-06-16 23:08:24] [build-stderr] Error: 'lgtm-install-packages' terminated with exit code: 100
```

orchagent/bufferorch.cpp
```diff
@@ -416,14 +416,17 @@ task_process_status BufferOrch::processBufferPool(KeyOpFieldsValuesTuple &tuple)
             return task_process_status::task_need_retry;
         }
 
-        sai_status = sai_buffer_api->remove_buffer_pool(sai_object);
-        if (SAI_STATUS_SUCCESS != sai_status)
+        if (SAI_NULL_OBJECT_ID == sai_object)
```
neethajohn, line 419: The condition should be '!='

orchagent/bufferorch.cpp
```diff
@@ -416,14 +416,17 @@ task_process_status BufferOrch::processBufferPool(KeyOpFieldsValuesTuple &tuple)
...
-                return handle_status;
+                SWSS_LOG_ERROR("Failed to remove buffer pool %s with type %s, rv:%d", object_name.c_str(), map_type_name.c_str(), sai_status);
+                task_process_status handle_status = handleSaiRemoveStatus(SAI_API_BUFFER, sai_status);
+                if (handle_status != task_process_status::task_success)
+                {
+                    return handle_status;
+                }
             }
         }
         SWSS_LOG_NOTICE("Removed buffer pool %s with type %s", object_name.c_str(), map_type_name.c_str());
```
neethajohn, line 432: Shouldn't L432-435 be also executed only if the object exists?

orchagent/bufferorch.cpp
```diff
@@ -615,14 +618,17 @@ task_process_status BufferOrch::processBufferProfile(KeyOpFieldsValuesTuple &tup
...
+            sai_status = sai_buffer_api->remove_buffer_profile(sai_object);
+            if (SAI_STATUS_SUCCESS != sai_status)
             {
-                return handle_status;
+                SWSS_LOG_ERROR("Failed to remove buffer profile %s with type %s, rv:%d", object_name.c_str(), map_type_name.c_str(), sai_status);
+                task_process_status handle_status = handleSaiRemoveStatus(SAI_API_BUFFER, sai_status);
+                if (handle_status != task_process_status::task_success)
+                {
+                    return handle_status;
+                }
```
neethajohn, line 631: L635-636 should also be executed only if object exists

neethajohn (changes requested):
How can we test this?

prsunny:
@stephenxs , please do not close and re-open the PR multiple times for retriggering the checks. It can be done via /azp run. I think once the code changes/review is completed, this can be taken care of.

stephenxs:
> @stephenxs , please do not close and re-open the PR multiple times for retriggering the checks. It can be done via /azp run. I think once the code changes/review is completed, this can be taken care of.

Thanks Prince. Looks like we don’t have the privilege to trigger azp rerun. Anyway I’ll avoid closing/reopening the PRs

stephenxs:
> @stephenxs , please do not close and re-open the PR multiple times for retriggering the checks. It can be done via /azp run. I think once the code changes/review is completed, this can be taken care of.

Thanks Prince. Looks like we don’t have the privilege to trigger azp rerun. Anyway I’ll avoid closing/reopening the PRs

> How can we test this?

We have a regression test (test exceeding headroom) for buffer profile. We don’t have a chance to verify the buffer pool because we seldom remove a buffer pool. I updated it because they shared the same logic.

orchagent/bufferorch.cpp
stephenxs: Fixed

orchagent/bufferorch.cpp
stephenxs: I tend to keep it as it is.
- `SAI_NULL_OBJECT_ID != sai_object` guarantees the object has been created by SAI, thus unnecessary to call the remove function
- it doesn't guarantee the object hasn't been tracked/recorded in `m_buffer_type_maps`. eg. the SAI returns success with a NULL OBJ ID (even very unlikely to happen)
- we use a map to track reference, which is idempotent and harmless to remove an unexisting reference.
- on the contrary, if we don't remove it from the map when sai_object is NULL OBJ ID, there can be a reference leak in case the SAI returns success with a NULL OBJ ID (again, very unlikely to happen)

Meanwhile, I don't have a strong opinion on it.

neethajohn:
> > How can we test this?
> 
> We have a regression test ([test_exceeding_headroom](https://github.com/Azure/sonic-mgmt/blob/master/tests/qos/test_buffer.py#L1492)) for buffer profile. We don’t have a chance to verify the buffer pool because we seldom remove a buffer pool. I updated it because they shared the same logic.

@stephenxs , does this test already cover this scenario for buffer profile?

neethajohn:
> > > How can we test this?
> > 
> > 
> > We have a regression test ([test_exceeding_headroom](https://github.com/Azure/sonic-mgmt/blob/master/tests/qos/test_buffer.py#L1492)) for buffer profile. We don’t have a chance to verify the buffer pool because we seldom remove a buffer pool. I updated it because they shared the same logic.
> 
> @stephenxs , does this test already cover this scenario for buffer profile?

Discussed offline, it is covered

neethajohn (approved):

liat-grozovik (approved):

PR #1788, opened 2021-06-17
Fix for inband port

ysmanman:
**What I did**
Fix issues seen with inband port
- Setting admin and oper status of recirc/inband port to up.
- Setting system port info for inband port.

**Why I did it**
Observed the following issues with inband port:
- SAI does not support setting admin and oper status of recirc ports today.
  As a result, the admin and oper status of inband port are down. This prevents
  remote neighbor from being programmed. 
- For recirc ports, SAI call get_system_port_attribute() returns a pid that's
  different from the pid returned from get_switch_attribute(SAI_SWITCH_ATTR_PORT_LIST).
  As a result, the system port info of recirc ports are not set correctly when adding
  system port of recirc port. 

**How I verified it**
Without the above fix, remote neighbor was not programmed. Verified that remote neighbor was programmed with the fix.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -5973,8 +5973,6 @@ bool PortsOrch::doProcessRecircPort(string alias, string role, set<int> lane_set
         Port p(alias, Port::PHY);
         p.m_port_id = port_id;
         p.m_init = true;
-        m_recircPortRole[alias] = role;
-        setPort(alias, p);
 
         string lane_str = "";
         for (auto lane : lane_set)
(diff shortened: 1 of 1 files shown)


ysmanman:
@vganesan-nokia According to our testing, inband port was not working without this PR. We wanted to make sure we tested inband port correctly. Can you please take a look at this PR and see if this PR is required?

vganesan-nokia:
> @vganesan-nokia According to our testing, inband port was not working without this PR. We wanted to make sure we tested inband port correctly. Can you please take a look at this PR and see if this PR is required?

@ysmanman, we did not face this problem. We use only one recycle port. It is defined as Ethernet-IB0 in port_config.ini. The config_db.json also defines this Ethernet-IB0 in PORT table. We get the correct oid (the same oid we get in PortsOrch constructor and later saved in **port.m_port_id** in initPort()) for SAI_SYSTEM_PORT_ATTR_PORT attribute when we call the get_system_port_attribute() with system port id of the Ethernet-IB0

PR #1792, opened 2021-06-18
[portsorch] Adjust port-buffer-drop counter default interval from 60000ms to 10000ms

Junchao-Mellanox:
**What I did**

Adjust port-buffer-drop counter default interval from 60000ms to 10000ms

**Why I did it**

There is a performance optimization for port-buffer-drop counter, with it the CPU usage for this counter decreased dramatically (SDK CPU usage from 40%~60% to 1% ~3%).

**How I verified it**

Manual test

**Details if related**


neethajohn:
@Junchao-Mellanox , there was a discussion to add a testcase to check the cpu usage so that this change can be validated on other asics as well

Junchao-Mellanox:
Hi @neethajohn, is this test case ready now?

keboliu (approved):

PR #1794, opened 2021-06-18
[fpmsyncd] updates for MPLS

qbdwlr:
What I did
SONiC swss fpmsyncd support for MPLS:
1) fpmsyncd support for routes with AF_MPLS and NHs with MPLS attributes.

Why I did it
SONiC swss fpmsyncd support for MPLS

How I verified it
Unit-tests in sonic-swss/tests/test_mpls.py
System tests in sonic-mgmt

Details if related
Refer to HLD: Azure/SONiC#706


fpmsyncd/routesync.cpp
```diff
@@ -897,6 +1004,123 @@ bool RouteSync::getIfName(int if_index, char *if_name, size_t name_len)
     return true;
 }
 
+#ifndef LIBNL3_ENCAP_MPLS
```
smaheshm, line 1007: lets remove this and update azure pipeline to fetch libnl3 debian package from sonic-buildimaage

qbdwlr: @smaheshm I am not familiar with the azure pipeline infrastructure and will need input with exact changes required to fetch libnl3 debian  packages from sonic-buildimage.

qbdwlr: @smaheshm I have submitted a new PR https://github.com/Azure/sonic-buildimage/pull/8064 that is the only other change required to make this PR work in the Azure pipeline as it is currently.

.azure-pipelines/build-template.yml
```diff
@@ -21,6 +21,9 @@ parameters:
 - name: sonic_slave
   type: string
 
+- name: buildimage_artifact_name
```
smaheshm, line 24: The value for this variable must be passed from "azure-pipelines.yml", I don't see that change in the PR.

qbdwlr: @smaheshm  Did you look at the pipeline details for this?  Doesn't even get to the point where the .azure-pipelines stuff is needed.  lgtm.yml is run first and fails... so this issue you have pointed out is irrelevant for now.

lgtm.yml
```diff
@@ -12,20 +12,39 @@ extraction:
...
+      - dpkg-deb -x libnl-route-3-200_3.5.0-1_amd64.deb $LGTM_WORKSPACE
+      - dpkg-deb -x libnl-route-3-dev_3.5.0-1_amd64.deb $LGTM_WORKSPACE
+      - dpkg-deb -x libnl-nf-3-200_3.5.0-1_amd64.deb $LGTM_WORKSPACE
+      - dpkg-deb -x libnl-nf-3-dev_3.5.0-1_amd64.deb $LGTM_WORKSPACE
+      - sudo dpkg -i libnl-3-dev_3.5.0-1_amd64.deb
+      - sudo dpkg -i libnl-genl-3-dev_3.5.0-1_amd64.deb
+      - sudo dpkg -i libnl-route-3-dev_3.5.0-1_amd64.deb
+      - sudo dpkg -i libnl-nf-3-dev_3.5.0-1_amd64.deb
+      - popd
+      - git clone https://github.com/qbdwlr/sonic-swss-common; pushd sonic-swss-common; git checkout azp; ./autogen.sh; fakeroot dpkg-buildpackage -us -uc -b; popd
```
qiluo-msft, line 47: Why using a personal repo?

smaheshm: I guess @qbdwlr  was trying out if this change will fix the LGTM analysis. Looks like it didn't.

fpmsyncd/routesync.cpp
```diff
@@ -758,6 +773,98 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
...
+    else if (nlmsg_type != RTM_NEWROUTE)
+    {
+        SWSS_LOG_INFO("Unknown message-type: %d for LabelRoute %s", nlmsg_type, destaddr);
+        return;
+    }
+
+    /* Get the index of the master device */
+    uint32_t master_index = rtnl_route_get_table(route_obj);
+    /* if the table_id is not set in the route obj then route is for default vrf. */
+    if (master_index)
```
smaheshm, line 805: How did you find out non-zero value implies a VRF table. "254" is the main table that ip route installs, 255 and 253 and local and default route table. All these are non-VRF tables.

qbdwlr: @smahesh You do know that SONiC has modified the fpm netlink message content from frr to something non-standard, right?  The rtnl_route_get_table() function will not return the RT_TABLE value in SONiC fpmsyncd, it returns the VRF ifindex value instead because SONiC has patched this.  
https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-frr/patch/0003-Use-vrf_id-for-vrf-not-tabled_id.patch

smaheshm: that's for the info, no I wasn't aware.

qbdwlr: @smaheshm Yes, this is the reason I have to disable VRF support in the other PR (https://github.com/Azure/sonic-swss/pull/1765) where I provide the fpmsyncd runtime option to listen to kernel netlink instead of frr fpm.  kernel netlink messages send the standard RT_TABLE ID in the netlink message, whereas frr fpm substitutes the patched VRF ifindex value.

rlhui:
@prsunny - would you please help review? thanks.

smaheshm:
> @prsunny - would you please help review? thanks.

@rlhui We will use either https://github.com/Azure/sonic-swss/pull/1870 or https://github.com/Azure/sonic-swss/pull/1871 for the merge since it includes AZP pipeline changes too.

PR #1797, opened 2021-06-22
[portsorch] fix errors when moving port from one lag to another.

stepanblyschak:
In schenario that is executed in sonic-mgmt in test_po_update.py a
portchannel member is deleted from one portchannel and added to another
portchannel:

```
config portchannel member del PortChannel9999 Ethernet0; config
portchannel member add PortChannel0001 Ethernet0;
```

It is possible that requests from teamsynd will arrive to PortsOrch
different order:

```
2021-06-17.14:01:19.405245|LAG_MEMBER_TABLE:PortChannel0001:Ethernet0|SET|status:enabled
2021-06-17.14:01:19.405318|LAG_MEMBER_TABLE:PortChannel999:Ethernet0|DEL
```

Consider following python simulated flow between teamsyncd and orchangent:

```python
from swsscommon import swsscommon

db = swsscommon.DBConnector('APPL_DB', 0)
table = swsscommon.ProducerStateTable(db, 'LAG_MEMBER_TABLE')
consumer = swsscommon.ConsumerStateTable(db, 'LAG_MEMBER_TABLE')

fvs = swsscommon.FieldValuePairs([('field', 'value')])

# teamsyncd deletes member from one lag and adds to another lag.
# orchagent is blocked due to other tasks (updating routes for example)
table.delete('PortChannel999:Ethernet112')
table.set('PortChannel0001:Ethernet112', fvs)

# orchagent unblocks and pops data
print(consumer.pop())
print(consumer.pop())

# orchagent will receive this:
# ['PortChannel0001:Ethernet112', 'SET', (('field', 'value'),)]
# ['PortChannel999:Ethernet112', 'DEL', ()]
```

This is a fundamental issue of Producer/ConsumerStateTable, thus orchagent must be aware of this and treat it as normal situation and figure out the right order and not crash or print an errors.

**What I did**

Check if port is already a lag member beforehand.
Added an UT to cover this scenario, this UT verifies that SAI API is not called in this case.
Refactored portsorch_ut.cpp by moving out Orchs creation/deletion into SetUp()/TearDown()

**Why I did it**

To fix errors in log.

**How I verified it**

Ran test_po_update.py test.

**Details if related**


tests/mock_tests/portsorch_ut.cpp
```diff
@@ -455,6 +429,102 @@ namespace portsorch_test
         ts.clear();
     }
 
+    TEST_F(PortsOrchTest, LagMemberDoesNotCallSAIApiWhenPortIsAlreadyALagMember)
```
liat-grozovik, line 432: i know all is writing this was but i would like to recommend adding a description of the motivation and the steps of each test.

liat-grozovik (dismissed):
small but not must comment

prsunny:
So here the netlink events is coming correctly but teamsyncd events are in different order to orchagent? @qiluo-msft to take a look

orchagent/portsorch.cpp
```diff
@@ -3704,8 +3704,15 @@ void PortsOrch::doLagMemberTask(Consumer &consumer)
...
             if (lag.m_members.find(port_alias) == lag.m_members.end())
             {
-                /* Assert the port doesn't belong to any LAG already */
-                assert(!port.m_lag_id && !port.m_lag_member_id);
+                /* Assert the port is not a LAG */
+                assert(port.m_lag_id == SAI_NULL_OBJECT_ID);
+
+                if (port.m_lag_member_id != SAI_NULL_OBJECT_ID)
+                {
+                    SWSS_LOG_NOTICE("Port %s is already a LAG member", port.m_alias.c_str());
```
prsunny, line 3712: IMO, please move it to INFO as we dont want this possible retry code to continuously spew logs. Otherwise the change lgtm. @judyjoseph to sign-off.

liat-grozovik:
@judyjoseph can you please provide your input?

orchagent/portsorch.cpp
stepanblyschak: This will not continuously print in logs, when the situation described above happens, you'll get only 1 notice message. I did not want to make it lower level because if user configured lag member which is already a part of different lag he will not see any notification in the log.

prsunny: But you are not erasing the object, right? So there will be a retry in the next doTask? Can you please confirm?

judyjoseph: I think Prince is referring to a case where Ethernet0 is already part of a different Port Channel & user issues this command only  --> config portchannel member add PortChannel0001 Ethernet0;   (no delete command issued )

judyjoseph:
> So here the netlink events is coming correctly but teamsyncd events are in different order to orchagent? @qiluo-msft to take a look

Even I would like to find a bit more on this ..shouldn't the producer events be added in the order it is received -- otherwise there could be issues with other use cases eg: vlan member addition

orchagent/portsorch.cpp
stepanblyschak: Does this mean orchagent should not print anything on such invalid configuration?

prsunny: @judyjoseph , in the above case, IMO, the CLI itself should return error and should not accept the command. My concern is about a race condition, which we expect to get resolved in next few retries. @stepanblyschak , afaik, this is not an invalid configuration. If you say its invalid config, then error should be returned from upper layers and not orchagent.

orchagent/portsorch.cpp
```diff
@@ -3704,8 +3704,15 @@ void PortsOrch::doLagMemberTask(Consumer &consumer)
 
             if (lag.m_members.find(port_alias) == lag.m_members.end())
             {
-                /* Assert the port doesn't belong to any LAG already */
-                assert(!port.m_lag_id && !port.m_lag_member_id);
+                /* Assert the port is not a LAG */
+                assert(port.m_lag_id == SAI_NULL_OBJECT_ID);
```
qiluo-msft, line 3708: >assert [](http://example.com/codeflow?start=16&length=6)

`assert` will be nop in release build. This is a runtime eror, we need check and treat it as an error.

stepanblyschak: Right, are you saying this needs to be handled in this PR?

qiluo-msft: yes, please

orchagent/portsorch.cpp
stepanblyschak: @prsunny "error should be returned from upper layers" - this is correct for CLI only for >=202012. There is a check in CLI whether port is already a lag member. In 201911 there is no such check, also a "config reload" check does not exist in any release meaning we either have to:
1. Don't print anything to log by default (INFO or DEBUG) making it harder for end user to figure out there is something wrong with configuration.
2. Have a one time NOTICE message in case the scenario described above happens which is rare and in case of invalid configuration user will see a lot of failed retries in the log rather then nothing.

prsunny: so this fix is only intended for 202012 and master, right. I dont see it marked for 201911 (although i'm *not* suggesting for 201911)

qiluo-msft: > in case of invalid configuration user will see a lot of failed retries in the log rather then nothing

In this case, infinite retry and logging are bad, the best effort orchagent could do:
1. error should be returned from upper layers
2. discard the invalid configuration and do not retry infintely

tests/mock_tests/portsorch_ut.cpp
stepanblyschak: Done

orchagent/portsorch.cpp
stepanblyschak: @qiluo-msft Then how to deal with reorder? or it is a ConsumerStateTable bug? although ConsumerStateTable always worked without preserving order as far as I know.

stepanblyschak: @prsunny I suggest mark it for 201911 as well as I see a possibility for this to happen there as well.

orchagent/portsorch.cpp
stepanblyschak: Removed this assert, placed a check for correct port type instead at the begginning.

stepanblyschak: Actually there are lots of asserts, only this is addressed

stepanblyschak:
@judyjoseph There could be other similar bugs in orchs that do not expect reorder.
@qiluo-msft Could you please comment on this?

qiluo-msft:
The design of ProducerStateTable and ConsumerStateTable only respect the final state, not the order. Even m_toSync also design with this philosophy.

You could use ProducerTable and ConsumerTable to conquer to first shortage, but not the second.

orchagent/portsorch.cpp
qiluo-msft: I am only comment on "infinite retry". For reorder case, I think retry will pass finally.

qiluo-msft (dismissed):
LGTM. Please check with other reviewers.

orchagent/portsorch.cpp
prsunny: I dont think this is required for 201911 as it is relatively stable. Secondly, please fix the log level as discussed above. As mentioned, if it is a reorder case, it will pass.. if not it will infinitely spew log.

stepanblyschak: @prsunny changed. Regarding 201911 I just want to make it clear that it is possible althoug I didn't tested it (a reporduction may require to put some artificial load on CPU while doing member add/remove in the loop). In 201911 it is even worth then in 202012. In 201911 the call is done in sairedis async mode which means it will crash orchagent if the call is not successful. In 202012 it will call SAI but orchagent will retry later.

liat-grozovik:
@judyjoseph and @prsunny could you please review and approve? we need this for 202012 as well.

qiluo-msft (approved):
LGTM. Please check with other reviewers.

orchagent/portsorch.cpp
```diff
@@ -3676,6 +3682,15 @@ void PortsOrch::doLagMemberTask(Consumer &consumer)
             continue;
         }
 
+        /* Fast failure if a port type is not a valid type for beeing a LAG member port.
```
prsunny, line 3685: Could you please revisit the comment? What is fast failure? Also beeing -> typo

prsunny (commented):
lgtm, minor comment. @judyjoseph to approve

orchagent/portsorch.cpp
stepanblyschak: Fixed

prsunny (approved):

liat-grozovik:
@judyjoseph kindly reminder, can you please review following the changes integrated?

orchagent/portsorch.cpp
```diff
@@ -3666,6 +3672,15 @@ void PortsOrch::doLagMemberTask(Consumer &consumer)
             continue;
         }
 
+        /* Fail if a port type is not a valid type for being a LAG member port.
+         * Erase invalid entry, no need to retry in this case. */
+        if (!isValidPortTypeForLagMember(port))
```
judyjoseph, line 3677: This specific check might not be needed, as it should be covered in the above API getPort() which will return false on failure to find a match. Was there a specific reason it was explicitly added

judyjoseph:
> @judyjoseph kindly reminder, can you please review following the changes integrated?

Sure, have a minor comment - else looks good !

orchagent/portsorch.cpp
stepanblyschak: getPort() does not check if this port can be a valid lag member, this check is needed.

judyjoseph: Sure ok .. I see. I was referring to the other overloaded call, but here we pass the alias as the parameter https://github.com/Azure/sonic-swss/blob/736f739fc7dc6593a03feb21706eafde99ae0b78/orchagent/portsorch.cpp#L668.

judyjoseph (approved):

qiluo-msft:
This commit could not be cleanly cherry-pick to 202012. Please submit another PR.

PR #1803, opened 2021-06-24
[flex-counters] Delay flex counters stats init for faster boot time

shlomibitton:

**What I did**
Update flex counters DB with counters stats only when counters are enabled.
As long as the polling counters are not enabled, flex counters information will stored internally on PortsOrch.

**Why I did it**
Creating flex counters objects on the DB will trigger 'SYNCD' to access the HW for query statistics capabilities.
This HW access takes time and will be better to finish boot before doing this (mainly for fast-reboot but good to have in general).
The flex counters are not crucial at boot time, we can delay it to the end of the boot process.

**How I verified it**
Reboot a switch and observer the flex counters DB populated after counters are enabled.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -284,9 +284,9 @@ static char* hostif_vlan_tag[] = {
 PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_with_pri_t> &tableNames, DBConnector *chassisAppDb) :
         Orch(db, tableNames),
         m_portStateTable(stateDb, STATE_PORT_TABLE_NAME),
-        port_stat_manager(PORT_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, true),
-        port_buffer_drop_stat_manager(PORT_BUFFER_DROP_STAT_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_BUFFER_DROP_STAT_POLLING_INTERVAL_MS, true),
-        queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, true)
+        port_stat_manager(PORT_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
+        port_buffer_drop_stat_manager(PORT_BUFFER_DROP_STAT_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_BUFFER_DROP_STAT_POLLING_INTERVAL_MS, false),
+        queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false)
 {
     SWSS_LOG_ENTER();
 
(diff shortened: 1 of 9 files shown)


tests/mock_tests/portsorch_ut.cpp
```diff
@@ -146,8 +146,13 @@ namespace portsorch_test
...
         ASSERT_EQ(gPortsOrch, nullptr);
 
-        gPortsOrch = new PortsOrch(m_app_db.get(), m_state_db.get(), ports_tables, m_chassis_app_db.get());
+        vector<string> flex_counter_tables = {
+            CFG_FLEX_COUNTER_TABLE_NAME
+        };
+        auto* flexCounterOrch = new FlexCounterOrch(m_config_db.get(), flex_counter_tables);
+        gDirectory.set(flexCounterOrch);
 
+	    gPortsOrch = new PortsOrch(m_app_db.get(), m_state_db.get(), ports_tables, m_chassis_app_db.get());
```
yxieca, line 155: Are you  using tab here? If so, please change it to white spaces. Otherwise please fix the indentation.

shlomibitton: Fixed

yxieca: What unit test is done with this new flexCounterOrch?

orchagent/portsorch.cpp
```diff
@@ -5072,6 +5077,46 @@ void PortsOrch::generatePriorityGroupMapPerPort(const Port& port)
...
+{
+    if (m_isPortCounterMapGenerated)
+    {
+        return;
+    }
+
+    auto port_counter_stats = generateCounterStats(PORT_STAT_COUNTER_FLEX_COUNTER_GROUP);
+    for (const auto& it: m_portList)
+    {
+        if (it.second.m_type != Port::Type::PHY)
```
yxieca, line 5090: Please comment why only PHY port needs to be queried.

orchagent/portsorch.cpp
```diff
@@ -5072,6 +5077,46 @@ void PortsOrch::generatePriorityGroupMapPerPort(const Port& port)
...
+{
+    if (m_isPortBufferDropCounterMapGenerated)
+    {
+        return;
+    }
+
+    auto port_buffer_drop_stats = generateCounterStats(PORT_BUFFER_DROP_STAT_FLEX_COUNTER_GROUP);
+    for (const auto& it: m_portList)
+    {
+        if (it.second.m_type != Port::Type::PHY)
```
yxieca, line 5110: Please comment

yxieca:
Please add unit test to guarantee that only PHY port will be added to the query list.

orchagent/portsorch.cpp
shlomibitton: Generating counter stats will eventually trigger syncd to query the counter statistics from the HW and this is possible only for physical ports.

tests/mock_tests/portsorch_ut.cpp
shlomibitton: This is to adapt to the new change, there is no new UT on this file.
Since PortsOrch object is calling functions on the FlexOrch this change was needed.
The UT added on a different python file.

shlomibitton:
> Please add unit test to guarantee that only PHY port will be added to the query list.

@yxieca there is a check to verify the port ID on the DB is not 0x0, from my testing PortChannels and CPU port are with this ID.
please check
https://github.com/Azure/sonic-swss/pull/1803/files#diff-574b26154866415f530c54dd5bc0feb404cbd5b121fd12c043057cf7b6b42b4cR107

orchagent/portsorch.cpp
yxieca: Please add comment in code so that people will understand this logic in the future.

shlomibitton: Added.

yxieca (approved):

PR #1808, opened 2021-06-28
[swss]: Allow portsyncd to run on system without ports

liorghub:
**What I did**
Remove Portsyncd error exit if there are no ports in config DB.

**Why I did it**
Allow system with no ports in config db run without errors
This is needed for modular system which should boot properly without line cards.

**How I verified it**
Remove ports from config DB and verify proper init flow. 

**Details if related**

### DIFF
--- a/portsyncd/portsyncd.cpp
+++ b/portsyncd/portsyncd.cpp
@@ -88,10 +88,8 @@ int main(int argc, char **argv)
 
         if (!handlePortConfigFromConfigDB(p, cfgDb, warm))
         {
-            // if port config is missing in ConfigDB
-            // program will exit with failure
-            SWSS_LOG_THROW("ConfigDB does not have port information, exiting...");
-            return EXIT_FAILURE;
+            SWSS_LOG_NOTICE("ConfigDB does not have port information, "
+                            "however ports can be added later on, continuing...");
         }
 
         LinkSync sync(&appl_db, &state_db);


portsyncd/portsyncd.cpp
```diff
@@ -86,13 +86,7 @@ int main(int argc, char **argv)
...
         cout << "Listen to link messages..." << endl;
 
-        if (!handlePortConfigFromConfigDB(p, cfgDb, warm))
-        {
-            // if port config is missing in ConfigDB
-            // program will exit with failure
-            SWSS_LOG_THROW("ConfigDB does not have port information, exiting...");
-            return EXIT_FAILURE;
-        }
+        handlePortConfigFromConfigDB(p, cfgDb, warm);
```
prsunny, line 89: please have all the code as before and log error here instead of throw. Also give the comment why continuing the flow.

liorghub: @prsunny 
Sure will do. Since zero ports is becoming a valid flow, may I use log warning instead of error?

prsunny: ok

liorghub: @prsunny I gave it some more thought. 
Do we really want to add a warning here? 
Since there is no problem and we just want to notify the user that there are no ports, we might as well change it to notice. 
What do you think?

prsunny: fine with me.

akokhan:
I believe this can be also useful to be able to start with an empty configuration:
https://github.com/Azure/sonic-buildimage/blob/24c90cedc42fe8f133389248d5363e500fe04c0e/src/sonic-config-engine/config_samples.py#L154

prsunny (approved):

Issue #1809, opened 2021-06-28
[VS test stability] Chassis test


prsunny:
The following tests starting consistently failing. Please analyze and fix.

```
2021-06-25T18:28:41.0688328Z test_virtual_chassis.py::TestVirtualChassis::test_chassis_app_db_sync FAILED [ 86%]
2021-06-25T18:28:51.5768542Z test_virtual_chassis.py::TestVirtualChassis::test_chassis_system_neigh FAILED [ 86%]
2021-06-25T18:29:02.0485211Z test_virtual_chassis.py::TestVirtualChassis::test_chassis_system_lag FAILED [ 87%]
```

Failure is [here](https://dev.azure.com/mssonic/build/_build/results?buildId=21347&view=logs&j=da806f27-156e-5225-7a93-a039d1b8cbbe&t=f1bfaaef-1c6a-5abf-8650-c6346f2cd003)

More logs, artifacts for all three attempts could be found at:

https://dev.azure.com/mssonic/build/_build/results?buildId=21347&view=artifacts&pathAsName=false&type=publishedArtifacts

prsunny:
@vganesan-nokia , @minionatwork , kindly prioritize this issue. Please email me, if you've further questions.

prsunny:
It is temporarily skipped as part of https://github.com/Azure/sonic-swss/issues/1809

vganesan-nokia:
> @vganesan-nokia , @minionatwork , kindly prioritize this issue. Please email me, if you've further questions.

The vs test for virtual chassis is broken due to problems introduced by couple of sonic-buildimage PRs merged on 06/23

(1) The changes from PR https://github.com/Azure/sonic-buildimage/pull/7857 introduced problem in construction of  config_db.json from /usr/share/sonic/virtual_chassis/default_config.json that is needed for VS virtual chassls tests.

(2) The fabricportlane mapping introduced by PR https://github.com/Azure/sonic-buildimage/pull/7629. Along with this PR merging FABRIC_NUM_PORTS in conftest.py should be set to 16 (as stated in PR https://github.com/Azure/sonic-swss/pull/1459)

I'll raise two PRs (one in sonic-buildimage and one in sonic-swss) to fix these.

vganesan-nokia:
Fixed by the following two PRs:

sonic-swss: https://github.com/Azure/sonic-swss/pull/1811
sonic-buildimage: https://github.com/Azure/sonic-buildimage/pull/8008

[...]

PR #1459, opened 2020-10-07
[fabricportsorch] Add fabric support

ngoc-do:
This code is to add support for fabric asics and NPU with fabric ports enabled.

**What I did**

1. Create FabricOrchDaemon for fabric asics, instead of using OrchDaemon which is used for NPU.

2. Create FabricPortsOrch to manage fabric ports. It collects information about port state, peer switch id and peer lane (stored in STATE_DB), sets up port stats and queue states. In future, it will also be used to enable/disable erroneous fabric ports.

3. Fabric port and queue stats are setup to be collected via FlexCounters.

One thing to note here: The code is waiting for https://github.com/Azure/sonic-swss/pull/1431 to start fabric flow for fabric asic/NPU with enabled fabric ports. So right now, there is no code that creates objects for FabricOrchDaemon/FabricPortsOrch yet. We need to know asic type (and if asic needs fabric ports to be enabled) to do that. These information should come from orchagent/main.cpp, and will be available after https://github.com/Azure/sonic-swss/pull/1431 merges.

**Why I did it**

Initialize fabric asics and support traffic across multiple NPUs.

**How I verified it**

Tested in systems with fabric asics (BCM 88790) and NPU with fabric ports (BCM 88690).

**Details if related**

This code will require to update a few other things: 

(1) syncd: to ask FlexCounter to  collect fabric counters (https://github.com/Azure/sonic-sairedis/pull/669)

(2) docker-orchagent: disable orchagent-portsyncd dependency. In this pull request, after fabric asic is initialized (by SAI), all fabric ports will be enabled by default. So portsyncd is not used for fabric asic at all, and orchagent should not wait for portsyncd (https://github.com/Azure/sonic-buildimage/pull/5569)

This pull request doesn't include:

- Cli commands to show portstat and queuestat for fabric ports. Typically fabric is used in multi-asic environment, and portstat, queuestat seem not to be supported for multi asic systems yet. We will add them later in next requests.

The implementation is based on the design doc proposed in https://github.com/Azure/SONiC/pull/668/.

PR #1811, opened 2021-06-29
[vstest] VS test failure fix after fabric port orch PR merge

vganesan-nokia:
**What I did**

- In tests/conftest.py Set the FABRIC_NUM_PORTS to 16. This is required since addition of
fabric port orch adds 16 fabrics in asic db internally.
- in tests/test_virtual_chassis.py added back the skipped tests

**Why I did it**

Fix for virtual chassis tests failure issue https://github.com/Azure/sonic-swss/issues/1809

**How I verified it**

With sonic-buildimage PR https://github.com/Azure/sonic-buildimage/pull/8008, run the vs test. All tests from test_virtual_chassis.py should pass consistently

**Details if related**

This PR is dependent on sonic-buildimage PR https://github.com/Azure/sonic-buildimage/pull/8008

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -36,7 +36,7 @@ NUM_PORTS = 32
 # FIXME: Voq asics will have 16 fabric ports created (defined in Azure/sonic-buildimage#6185).
 # Right now, we set FABRIC_NUM_PORTS to 0, and change to 16 when PR#6185 merges. PR#6185 can't
 # be merged before this PR. Otherwise it will cause swss voq test failures.
-FABRIC_NUM_PORTS = 0
+FABRIC_NUM_PORTS = 16
 
 def ensure_system(cmd):
     rc, output = subprocess.getstatusoutput(cmd)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,9 +33,7 @@ from buffer_model import enable_dynamic_buffer
 # a dynamic number of ports. GitHub Issue: Azure/sonic-swss#1384.
 NUM_PORTS = 32
 
-# FIXME: Voq asics will have 16 fabric ports created (defined in Azure/sonic-buildimage#6185).
-# Right now, we set FABRIC_NUM_PORTS to 0, and change to 16 when PR#6185 merges. PR#6185 can't
-# be merged before this PR. Otherwise it will cause swss voq test failures.
+# Voq asics will have 16 fabric ports created (defined in Azure/sonic-buildimage#7629).
 FABRIC_NUM_PORTS = 16
 
 def ensure_system(cmd):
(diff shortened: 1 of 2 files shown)


vganesan-nokia:
@ngoc-do, please review.

ngoc-do:
> @ngoc-do, please review.

Thanks for updating the port number. I was going to do that after https://github.com/Azure/sonic-buildimage/pull/7629 merged, but you already made it. Thanks a lot.

tests/test_virtual_chassis.py
```diff
@@ -136,7 +136,6 @@ def test_voq_switch(self, vct):
                 spcfg = ast.literal_eval(value)
                 assert spcfg['count'] == sp_count, "Number of systems ports configured is invalid"
 
-    @pytest.mark.skip(reason="This test is not stable enough")
```
ngoc-do, line 139: Can I have the context for this? 

`@pytest.mark.skip(reason="This test is not stable enough")`

Why was it added? Was it added after https://github.com/Azure/sonic-buildimage/pull/7629 merged? Did https://github.com/Azure/sonic-buildimage/pull/7629 cause test instability?

tests/conftest.py
```diff
@@ -35,7 +35,7 @@
 # FIXME: Voq asics will have 16 fabric ports created (defined in Azure/sonic-buildimage#6185).
 # Right now, we set FABRIC_NUM_PORTS to 0, and change to 16 when PR#6185 merges. PR#6185 can't
 # be merged before this PR. Otherwise it will cause swss voq test failures.
```
ngoc-do, line 37: Remove the comment.

tests/test_virtual_chassis.py
ngoc-do: OK. I saw you mentioned 2 reasons.

(1) The changes from PR Azure/sonic-buildimage#7857 introduced problem in construction of config_db.json from /usr/share/sonic/virtual_chassis/default_config.json that is needed for VS virtual chassls tests.

(2) The fabricportlane mapping introduced by PR Azure/sonic-buildimage#7629. Along with this PR merging FABRIC_NUM_PORTS in conftest.py should be set to 16 (as stated in PR #1459)

I didn't look at carefully reason 1, but I'm curious why reason 2 could cause test failure?

It's because I think that if FABRIC_NUM_PORTS = 0, it still works fine. Of course, we have to change to 16 to test fabric code, but the value of 0 should not cause test failures.

It's because we are testing at least n ports, not exact n ports. Please find the following code in `tests/conftest.py`.

```
        # Verify that we have "at least" NUM_PORTS + FABRIC_NUM_PORTS, rather exact number.
        # Right now, FABRIC_NUM_PORTS = 0. So it essentially waits for at least NUM_PORTS.
        # This will allow us to merge Azure/sonic-buildimage#6185 that creates 16 fabric ports.
        # When PR#6185 merges, FABRIC_NUM_PORTS should be 16, and so this verification (at least
        # NUM_PORTS) still holds.
        # Will update FABRIC_NUM_PORTS to 16, and revert back to wait exact NUM_PORTS + FABRIC_NUM_PORTS
        # when PR#6185 merges.

        wait_at_least_n_keys = True
        asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_PORT", num_ports + 1, wait_at_least_n_keys)  # +1 CPU Port

        # Verify that fabric ports are monitored in STATE_DB
        if metadata.get('switch_type', 'npu') in ['voq', 'fabric']:
            self.get_state_db()
            self.state_db.wait_for_n_keys("FABRIC_PORT_TABLE", FABRIC_NUM_PORTS, wait_at_least_n_keys)
```

`wait_at_least_n_keys` means we are verifying at least `NUM_PORTS + FABRIC_NUM_PORTS`, not exact `NUM_PORTS + FABRIC_NUM_PORTS`. Since FABRIC_NUM_PORTS = 0, the verification still returns true when Azure/sonic-buildimage#7629 merged.

And could you please update the above code in `tests/conftest.py` by removing the comment and `wait_at_least_n_keys` in the call `wait_for_n_keys`? I think we are now safe to verify the exact number of ports.

vganesan-nokia: Those 3 tests consistently failed. So PR https://github.com/Azure/sonic-swss/pull/1807 was raised to skip these tests.
These tests failed not because of  Azure/sonic-buildimage#7629 but because of https://github.com/Azure/sonic-buildimage/pull/7857. I noticed that once the problem created by https://github.com/Azure/sonic-buildimage/pull/7857 is fixed, the port count mismatch problem occurred, which was due to Azure/sonic-buildimage#7629

tests/conftest.py
vganesan-nokia: Fixed

tests/test_virtual_chassis.py
ngoc-do: Cool, thanks.

ngoc-do (approved):

prsunny:
@vganesan-nokia , i still see the vs test failed. can you please check?

vganesan-nokia:
> @vganesan-nokia , i still see the vs test failed. can you please check?

This is expected. This PR id dependent on PR https://github.com/Azure/sonic-buildimage/pull/8008. We need to merge https://github.com/Azure/sonic-buildimage/pull/8008 first.

prsunny:
can you please rebase to latest?

vganesan-nokia:
@prsunny, I rebased to latest. There are 2 failures unrelated to virtual chasiss tests. Would you please restart the azp/vs tests? Thanks

prsunny (approved):

PR #1814, opened 2021-07-01
Cleanup code

kcudnik:
**What I did**
Update gitignore, remove unused variables, fix commandline help message, remove explicit move

**Why I did it**
Keep code clean

**How I verified it**
Local run

**Details if related**

### DIFF
--- a/orchagent/countercheckorch.cpp
+++ b/orchagent/countercheckorch.cpp
@@ -164,7 +164,7 @@ PfcFrameCounters CounterCheckOrch::getPfcFrameCounters(sai_object_id_t portId)
 
     if (!m_countersTable->get(sai_serialize_object_id(portId), fieldValues))
     {
-        return (counters);
+        return counters;
     }
 
     for (const auto& fv : fieldValues)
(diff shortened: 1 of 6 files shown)


orchagent/countercheckorch.cpp
```diff
@@ -164,7 +164,7 @@ PfcFrameCounters CounterCheckOrch::getPfcFrameCounters(sai_object_id_t portId)
 
     if (!m_countersTable->get(sai_serialize_object_id(portId), fieldValues))
     {
-        return move(counters);
+        return (counters);
```
qiluo-msft, line 167: >( [](http://example.com/codeflow?start=15&length=1)

Do we need the parentheses?

orchagent/countercheckorch.cpp
```diff
@@ -182,7 +182,7 @@ PfcFrameCounters CounterCheckOrch::getPfcFrameCounters(sai_object_id_t portId)
         }
     }
 
-    return move(counters);
+    return(counters);
```
qiluo-msft, line 185: >n [](http://example.com/codeflow?start=9&length=1)

Add a blank after `return`

qiluo-msft (approved):

PR #1815, opened 2021-07-02
Ignore ALREADY_EXIST error in FDB creation

shi-su:
**What I did**
Ignore ALREADY_EXIST error in FDB creation.
Fix: https://github.com/Azure/sonic-buildimage/issues/7798

**Why I did it**
In FDB creation, there are scenarios where the hardware learns an FDB entry before orchagent. In such cases, the FDB SAI creation would report the status of SAI_STATUS_ITEM_ALREADY_EXISTS, and orchagent should ignore the error and treat it as entry was explicitly created.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -709,10 +709,10 @@ task_process_status Orch::handleSaiCreateStatus(sai_api_t api, sai_status_t stat
                     return task_success;
                 case SAI_STATUS_ITEM_ALREADY_EXISTS:
                     /*
-                    *  In FDB creation, there are scenarios where the hardware learns an FDB entry before orchagent.
-                    *  In such cases, the FDB SAI creation would report the status of SAI_STATUS_ITEM_ALREADY_EXISTS,
-                    *  and orchagent should ignore the error and treat it as entry was explicitly created.
-                    */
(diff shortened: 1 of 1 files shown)


prsunny (approved):

kcudnik (approved):

orchagent/orch.cpp
```diff
@@ -699,15 +699,38 @@ task_process_status Orch::handleSaiCreateStatus(sai_api_t api, sai_status_t stat
...
+    switch (api)
     {
-        case SAI_STATUS_SUCCESS:
-            SWSS_LOG_WARN("SAI_STATUS_SUCCESS is not expected in handleSaiCreateStatus");
-            return task_success;
+        case SAI_API_FDB:
+            switch (status)
+            {
+                case SAI_STATUS_SUCCESS:
+                    SWSS_LOG_WARN("SAI_STATUS_SUCCESS is not expected in handleSaiCreateStatus");
```
kcudnik, line 708: interesting, success is not expected :D

shi-su: Yes, this function aims to handle the failed scenarios, so it is not expected to get called for success. This is just to put a warning when this unexpected scenario happens.

PR #1819, opened 2021-07-15
[202012][portsorch] fix errors when moving port from one lag to another.

stepanblyschak:
In scenario that is executed in sonic-mgmt in test_po_update.py a portchannel member is deleted from one portchannel and added to another portchannel.
It is possible that requests from teamsynd will arrive in different order
This reordering happens because teamsyncd has single event handler/selectable TeamSync::TeamPortSync::onChange() per team device so when two of them are ready it is swss::Select implementation detail in which order they are going to be returned.
This is a fundamental issue of Producer/ConsumerStateTable, thus orchagent must be aware of this and treat it as normal situation and figure out the right order and not crash or print an errors.

Original PR https://github.com/Azure/sonic-swss/pull/1797

Signed-off-by: Stepan Blyschak stepanb@nvidia.com

**What I did**
Check if port is already a lag member beforehand.
Added an UT to cover this scenario, this UT verifies that SAI API is not called in this case.
Refactored portsorch_ut.cpp by moving out Orchs creation/deletion into SetUp()/TearDown()

**Why I did it**
To fix errors in log.

**How I verified it**
Ran test_po_update.py test.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -212,7 +212,7 @@ static char* hostif_vlan_tag[] = {
 
 static bool isValidPortTypeForLagMember(const Port& port)
 {
-    return (port.m_type == Port::Type::PHY || port.m_type == Port::Type::SYSTEM);
+    return port.m_type == Port::Type::PHY;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


prsunny:
Can you provide the link to master PR?

dprital:
> Can you provide the link to master PR?

Link to master PR:
https://github.com/Azure/sonic-swss/pull/1797

qiluo-msft (approved):

PR #1822, opened 2021-07-16
[Dynamic Buffer Calc] Bug fix: Don't create lossless buffer profile for active ports without speed configured

stephenxs:
**What I did**
Bugfix: Don't create lossless buffer profiles for active ports without speed configured

Root cause:
- In `handlePortTableUpdate`, `refreshPgsFromPort` is called if admin status is updated even if the speed is not configured.
  This is reasonable because the port can be configured as headroom override and the profile should be applied in that case.
- However, as a side-effect, the port's state is set to `PORT_READY` in `refreshPgsForPort` regardless of whether the speed is configured, which is not correct.
  This should be avoided and `PORT_READY` should be set by the caller if necessary

Fix:
- Don't set the port's state to `PORT_READY` in `refreshPgsForPort` and check the port's state before calling it.
- Explicitly set port's state to `PORT_READY` in `handlePortStateTable`. This has been done by all other callers.

Note:
- The change is covered by the existing vs tests.
- The scenario where the bug was originally found can not be covered by vs test because:
  - The speed is always configured in vs image by default
  - Removing speed is not handled in buffer manager since it's not a valid flow.
- A regression test will be opened to cover this case.

**Why I did it**

**How I verified it**
Regression test and vs test.

**Details if related**

### DIFF
--- a/cfgmgr/buffermgrdyn.cpp
+++ b/cfgmgr/buffermgrdyn.cpp
@@ -987,8 +987,6 @@ task_process_status BufferMgrDynamic::refreshPgsForPort(const string &port, cons
         SWSS_LOG_DEBUG("Nothing to do for port %s since no PG configured on it", port.c_str());
     }
 
-    portInfo.state = PORT_READY;
-
     // Remove the old profile which is probably not referenced anymore.
     if (!profilesToBeReleased.empty())
     {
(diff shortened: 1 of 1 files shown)


neethajohn (approved):

liat-grozovik (approved):

judyjoseph:
@stephenxs, not able to cherry-pick this PR in 202106 cleanly. Is there other depedent PR's ? Or please please create a separate PR for 202106.

stephenxs:
> @stephenxs, not able to cherry-pick this PR in 202106 cleanly. Is there other depedent PR's ? Or please please create a separate PR for 202106.

I’ll check it. Thanks

stephenxs:
> > @stephenxs, not able to cherry-pick this PR in 202106 cleanly. Is there other depedent PR's ? Or please please create a separate PR for 202106.
> 
> I’ll check it. Thanks

Hi @judyjoseph 
Looks like I'm able to cherry-pick it to the 202106 branch.
I tested based on commit 97a108f8 which is the latest commit in 202106. Can you please check it?
Thanks

PR #1829, opened 2021-07-20
[crm] Fix for Issue Azure/sonic-buildimage#8036

qbdwlr:
**What I did**
Added more appropriate log message for CRM queries of MPLS_INSEG and MPLS_NEXTHOP resources when platform does not support these resource types.

**Why I did it**
Mellanox has raised the issue officially here: https://github.com/Azure/sonic-buildimage/issues/8036

**How I verified it**
System test with VS SAI modified to return NOT_SUPPORTED.

**Details if related**
N/A


qbdwlr:
@stepanblyschak @Junchao-Mellanox Please review this PR fix for https://github.com/Azure/sonic-buildimage/issues/8036

Junchao-Mellanox (approved):

prsunny (approved):

PR #1830, opened 2021-07-20
Td2: Reclaim buffer from unused ports

neethajohn:
By default, lossless profiles are attached to PGs 3, 4 of all ports which results in a lot of buffer wastage when most of those ports are unused (not associated with a DEVICE_NEIGHBOR). In TD2, ingress pool size comprises of the reserved space as well.  Hence making use of  a special cable length of '0m' to identify unused ports and skip reserving space for those ports

**What I did**
* In buffermgr, if port with cable len '0m' is identified, return immediately without creating pg lossless profile or attaching profile to the lossless pgs of that port
* Listen to 'admin_status' update as well from the PORT table and update port-speed mapping. This is to handle add-rack scenario where a port is added later
         - The port starts of with cable length 0m
         - Configlet to add a port is applied. The order of operations specifc to the PORT/CABLE_LENGTH table are - port is initially set to admin down, cable length is updated for that port, port table attributes are defined and port is set to admin up
         - speed update might not be seen when the port is set to admin up. Hence port-speed mapping will capture the speed update whenever its seen and once the cable length is updated while the port is brought back up, profiles can be attached to the lossless pgs of the port

**How I verified it**
* Manual tests done. Verified that no space is reserved for unused ports
* Verified that when a port is added using 'add-rack' scenario, profile is attached to pgs of that port
* New VS test added

### DIFF
--- a/cfgmgr/buffermgr.cpp
+++ b/cfgmgr/buffermgr.cpp
@@ -378,17 +378,17 @@ void BufferMgr::doTask(Consumer &consumer)
                 }
                 if (m_pgfile_processed && table_name == CFG_PORT_TABLE_NAME && (fvField(i) == "speed" || fvField(i) == "admin_status"))
                 {
-	            if (fvField(i) == "speed")
-		        {
-		            m_speedLookup[port] = fvValue(i);
+                    if (fvField(i) == "speed")
+                    {
+                        m_speedLookup[port] = fvValue(i);
                     }
-
-		        if (m_speedLookup.count(port) != 0)
+                    
+                    if (m_speedLookup.count(port) != 0)
                     {
                         // create/update profile for port
                         task_status = doSpeedUpdateTask(port);
-	            }
-		    }
+                    }
+                }
                 if (task_status != task_process_status::task_success)
                 {
                     break;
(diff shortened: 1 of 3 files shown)


cfgmgr/buffermgr.cpp
```diff
@@ -132,7 +134,13 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, string speed)
     }
 
     cable = m_cableLenLookup[port];
+    if (cable == "0m")
+    {
+        SWSS_LOG_INFO("Not creating/updating PG profile for port %s. Cable length is set to %s", port.c_str(), cable.c_str());
```
yxieca, line 139: Raise this log to NOTICE so that it will by default recorded in syslog?

neethajohn: Done!

cfgmgr/buffermgr.cpp
```diff
@@ -368,11 +376,18 @@ void BufferMgr::doTask(Consumer &consumer)
...
-                    task_status = doSpeedUpdateTask(port, fvValue(i));
+                    if (fvField(i) == "speed")
+                    {
+                        m_speedLookup[port] = fvValue(i);
+                    }
+                    
+                    if (m_speedLookup.count(port) != 0)
+                    {
+                        // create/update profile for port
+                        task_status = doSpeedUpdateTask(port);
```
prsunny, line 389: Now speed/profile is updated every time admin status changes for a port even if there is no change to port speed. Is it expected?

neethajohn: It will not update every time admin status changes. If there is no change in port speed, in doSpeedUpdateTask, we will hit L188 and return from L209 (which is skipping the profile attachment to the pg)

yxieca (approved):

prsunny (approved):

PR #1838, opened 2021-07-26
[Draft]: Portchannel macsec

Pterosaur:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


cfgmgr/teammgr.cpp
```diff
@@ -594,6 +609,20 @@ task_process_status TeamMgr::addLagMember(const string &lag, const string &membe
...
 
+    vector<FieldValueTuple> fvs;
+    m_cfgPortTable.get(member, fvs);
+
+    // Get the MACsec status
+    auto it = find_if(fvs.begin(), fvs.end(), [](const FieldValueTuple &fv) {
+            return fv.first == "macsec";
+            });
+    // Portchannel needs to be enabled after MACsec enabled
+    if (it != fvs.end()) {
```
prsunny, line 620: { -> new line

cfgmgr/teammgr.cpp
```diff
@@ -594,6 +609,20 @@ task_process_status TeamMgr::addLagMember(const string &lag, const string &membe
...
     uint16_t keyId = generateLacpKey(lag);
 
+    vector<FieldValueTuple> fvs;
+    m_cfgPortTable.get(member, fvs);
+
+    // Get the MACsec status
+    auto it = find_if(fvs.begin(), fvs.end(), [](const FieldValueTuple &fv) {
+            return fv.first == "macsec";
+            });
+    // Portchannel needs to be enabled after MACsec enabled
```
prsunny, line 619: Just curios, how is this handled if `macsec `is enabled after portchannel member is added?

Pterosaur: The MACsec will fail to enable if the port is a member of port channel. This solution isn't graceful so I'm investigating a better solution.

PR #1863, opened 2021-08-11
Fix fdb type as dynamic check

vaibhavhd:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


vaibhavhd:
This is covered under https://github.com/Azure/sonic-swss/pull/1866

PR #1868, opened 2021-08-12
Add autoconf-archive package to lgtm

kcudnik:
Needed for code coverage in configure.ac


kcudnik:
LGTM fails again since it didn't picked up this change (not sure why)  
./configure: line 11876: AX_CODE_COVERAGE: command not found

PR #1879, opened 2021-08-17
[dpb] Remove wait and validate LAG removal from end of test_port_dpb_lag test case

alexrallen:
**What I did**

Removed lines of code that wait for and verify removal of lag following the `test_port_dpb_lag` test case which verified that a port will not be removed if there is a lag referencing it. 

**Why I did it**

On Azure CI there seems to be a non-determinate period of time it takes for the LAG to be removed. This is not experienced locally when running this test in isolation. 

**How I verified it**

Run on CI and verify test passes.

### DIFF
--- a/tests/test_port_dpb_lag.py
+++ b/tests/test_port_dpb_lag.py
@@ -40,8 +40,8 @@ class TestPortDPBLag(object):
         self.dvs_lag.remove_port_channel_member(lag, p.get_name())
 
         # 8. Verify that port is removed from ASIC DB
-        assert(p.not_exists_in_asic_db() == True) 
-        
+        assert(p.not_exists_in_asic_db() == True)
+
         # 9. Re-create port Ethernet0 and verify that it is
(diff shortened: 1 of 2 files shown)


tests/test_port_dpb_lag.py
```diff
@@ -41,8 +40,8 @@ def test_dependency(self, dvs):
         self.dvs_lag.remove_port_channel_member(lag, p.get_name())
 
         # 8. Verify that port is removed from ASIC DB
-        assert(p.not_exists_in_asic_db() == True)
-
+        assert(p.not_exists_in_asic_db() == True)
```
dgsudharsan, line 43: Can you please review the white spaces?

dgsudharsan (changes requested):

dgsudharsan (dismissed):

prsunny (dismissed):

dgsudharsan:
@alexrallen  Please close this since we merged https://github.com/Azure/sonic-swss/pull/1919

PR #1881, opened 2021-08-18
Fix vs test failure in test_buffer_traditional

stephenxs:
**What I did**

The test expects buffer PGs created for all ports.
However, the vs image is deployed as cable length being 0m,
which prevents the buffer PGs from being created and fails the test.

To fix the failure:
- Config the cable length of the port under test to 300m if it was 0m
- Startup the port after cable length configured
  By doing that the buffer PG can be configured correctly
- Collect the orig_profiles after the port is started up
  This is required to check whether the new profile has been created after cable length/speed is changed.

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/tests/test_buffer_traditional.py
+++ b/tests/test_buffer_traditional.py
@@ -74,11 +74,15 @@ class TestBuffer(object):
         orig_cable_len = None
         orig_speed = None
         try:
-            dvs.runcmd("config interface startup {}".format(self.INTF))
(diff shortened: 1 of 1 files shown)


prsunny:
Ref: https://github.com/Azure/sonic-swss/pull/1875/files

stephenxs:
> Ref: #1875

Thanks. Will remove the "skip".

PR #1883, opened 2021-08-23
[bfdorch] Orchagent support hardware BFD

shi-su:
**What I did**
Implement bfdorch to program hardware BFD sessions via bfd SAI.
Add vs test for bfd sessions.

**Why I did it**
To support hardware BFD.

**How I verified it**
Configure hardware BFD sessions on virtual switches and physical devices and confirm the BFD session is programmed.

**Details if related**


shi-su:
Build failure is expected before https://github.com/Azure/sonic-swss-common/pull/524 is merged.

orchagent/bfdorch.h
```diff
@@ -0,0 +1,41 @@
+#ifndef SWSS_BFDORCH_H
+#define SWSS_BFDORCH_H
+
+#include "orch.h"
+#include "intfsorch.h"
```
prsunny, line 5: Please move these headers to .cpp if not referenced in this header file.

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,412 @@
...
+            sai_bfd_session_state_t state = bfdSessionState[i].session_state;
+
+            SWSS_LOG_INFO("Get bfd session state change notification id:%" PRIx64 " state:%d", id, state);
+
+            if (state != bfd_session_lookup[id].state)
+            {
+                auto key = bfd_session_lookup[id].peer;
+                m_stateBfdSessionTable->hset(key, "state", session_state_loopup.at(state));
+
+                bfd_session_lookup[id].state = state;
```
prsunny, line 136: Can you please add the code for notifying observers? similar to port oper change?

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,412 @@
...
+        sai_bfd_session_state_notification_t *bfdSessionState = nullptr;
+
+        sai_deserialize_bfd_session_state_ntf(data, count, &bfdSessionState);
+
+        for (uint32_t i = 0; i < count; i++)
+        {
+            sai_object_id_t id = bfdSessionState[i].bfd_session_id;
+            sai_bfd_session_state_t state = bfdSessionState[i].session_state;
+
+            SWSS_LOG_INFO("Get bfd session state change notification id:%" PRIx64 " state:%d", id, state);
```
prsunny, line 129: I also suggest having a NOTICE log for the BFD session state for logging purposes

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+    attr.id = SAI_BFD_SESSION_ATTR_BFD_ENCAPSULATION_TYPE;
+    attr.value.s32 = encapsulation_type;
+    attrs.emplace_back(attr);
+
+    attr.id = SAI_BFD_SESSION_ATTR_IPHDR_VERSION;
+    attr.value.u8 = src_ip.isV4() ? 4 : 6;
+    attrs.emplace_back(attr);
+
+    if (!src_ip_provided)
```
prsunny, line 245: Suggest moving this check before  attr handling.

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,412 @@
...
+    {
+        SWSS_LOG_ERROR("Failed to create BFD session %s because source IP is not provided", key.c_str());
+        return false;
```
prsunny, line 248: return `true` here as we don't want to retry in this case.

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,412 @@
...
+    {
+        SWSS_LOG_ERROR("BFD session for %s already exists", key.c_str());
+        return true;
+    }
+
+    size_t found_vrf = key.find(delimiter);
+    if (found_vrf == string::npos)
+    {
+        SWSS_LOG_ERROR("Failed to parse key %s, no vrf is given", key.c_str());
+        return false;
```
prsunny, line 156: return `true` as no need to retry.

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    size_t found_ifname = key.find(delimiter, found_vrf + 1);
+    if (found_ifname == string::npos)
+    {
+        SWSS_LOG_ERROR("Failed to parse key %s, no ifname is given", key.c_str());
+        return false;
```
prsunny, line 163: return `true` as no need to retry

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,412 @@
...
+    copy(attr.value.ipaddr, src_ip);
+    attrs.emplace_back(attr);
+    fvVector.emplace_back("local_addr", src_ip.to_string());
+
+    attr.id = SAI_BFD_SESSION_ATTR_DST_IP_ADDRESS;
+    copy(attr.value.ipaddr, peer_address);
+    attrs.emplace_back(attr);
+
+    attr.id = SAI_BFD_SESSION_ATTR_MIN_TX;
+    attr.value.u32 = tx_interval * 1000;
```
prsunny, line 260: define this. don't use magic number

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+        attr.id = SAI_BFD_SESSION_ATTR_SRC_MAC_ADDRESS;
+        memcpy(attr.value.mac, port.m_mac.getMac(), sizeof(sai_mac_t));
+        attrs.emplace_back(attr);
+
+        if (!dst_mac_provided)
+        {
+            SWSS_LOG_ERROR("Failed to create BFD session %s: destination MAC address required when hardware lookup not valid",
+                            key.c_str());
+            return false;
```
prsunny, line 311: return `true` as no need to retry. Suggest check all return codes in this function

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,412 @@
...
+        }
+        else
+        {
+            VRFOrch* vrf_orch = gDirectory.get<VRFOrch*>();
+            attr.value.oid = vrf_orch->getVRFid(vrf_name);
+        }
+
+        attrs.emplace_back(attr);
+
+        if (dst_mac_provided)
```
prsunny, line 340: I think this check can be relaxed as its ok to ignore. In future when modifying attributes are supported, we don't want to have some restriction as to remove certain attributes when some other attributes are modified.

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,412 @@
...
+    {
+        SWSS_LOG_ERROR("Failed to create bfd session %s, rv:%d", key.c_str(), status);
+        task_process_status handle_status = handleSaiCreateStatus(SAI_API_BFD, status);
+        if (!parseHandleSaiStatusFailure(handle_status))
+        {
+            return false;
+        }
+    }
+
+    const string state_db_key = vrf_name + state_db_key_delimiter + alias + state_db_key_delimiter + peer_address.to_string();
```
prsunny, line 362: Can you move this to a separate function so it can be used from both create and remove functions?

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,412 @@
...
+    {SAI_BFD_SESSION_STATE_UP,          "Up"}
+};
+
+BfdOrch::BfdOrch(DBConnector *db, string tableName):
+    Orch(db, tableName)
+{
+    SWSS_LOG_ENTER();
+
+    m_state_db = shared_ptr<DBConnector>(new DBConnector("STATE_DB", 0));
+    m_stateBfdSessionTable = unique_ptr<Table>(new Table(m_state_db.get(), STATE_BFD_SESSION_TABLE));
```
prsunny, line 51: As mentioned, please pass this attributes from orch init

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,386 @@
...
+    sai_attribute_t attr;
+    vector<sai_attribute_t> attrs;
+    vector<FieldValueTuple> fvVector;
+
+    for (auto i : data)
+    {
+        auto value = fvValue(i);
+
+        if (fvField(i) == "tx_interval")
+            tx_interval = to_uint<uint32_t>(value);
```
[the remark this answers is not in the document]
shi-su: Added braces.

orchagent/orchdaemon.cpp
```diff
@@ -288,6 +289,8 @@ bool OrchDaemon::init()
 
     gMacsecOrch = new MACsecOrch(m_applDb, m_stateDb, macsec_app_tables, gPortsOrch);
   
+    gBfdOrch = new BfdOrch(m_applDb, APP_BFD_SESSION_TABLE_NAME);
```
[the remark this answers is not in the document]
shi-su: Thanks for this suggestion, updated the code to pass `m_stateDB` and `STATE_BFD_SESSION_TABLE_NAME` to BfdOrch.

orchagent/main.cpp
```diff
@@ -451,6 +451,10 @@ int main(int argc, char **argv)
     attr.value.ptr = (void *)on_port_state_change;
     attrs.push_back(attr);
 
+    attr.id = SAI_SWITCH_ATTR_BFD_SESSION_STATE_CHANGE_NOTIFY;
```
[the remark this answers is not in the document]
shi-su: This is a good point. I am not sure how it would behave for vendor SAI without BFD support. Will do some research on that.

orchagent/bfdorch.h
shi-su: Moved the headers to .cpp.

orchagent/bfdorch.cpp
shi-su: Added code for notifying observer. For this PR, there will not be anywhere processing these notifications though.

orchagent/bfdorch.cpp
shi-su: Moved the check before attr handling

orchagent/bfdorch.cpp
shi-su: Updated return value to `true`

orchagent/bfdorch.cpp
shi-su: Updated return value to `true`

orchagent/bfdorch.cpp
shi-su: Updated return value to `true`

orchagent/bfdorch.cpp
shi-su: Updated return value to `true` and checked other return codes

orchagent/bfdorch.cpp
shi-su: Got this defined as `BFD_SESSION_MILLISECOND_TO_MICROSECOND`

orchagent/bfdorch.cpp
shi-su: This check is per SAI definition that mac address is required when and only when `SAI_BFD_SESSION_ATTR_HW_LOOKUP_VALID` is `false`. Therefore, I added the check here and the opposite one for `SAI_BFD_SESSION_ATTR_HW_LOOKUP_VALID` is `true`.

orchagent/bfdorch.cpp
shi-su: Defined a separate function to generate keys.

orchagent/bfdorch.cpp
shi-su: Thanks for the comment. Moved the attributes to orch init.

orchagent/bfdorch.cpp
shi-su: Added a NOTICE log in the `if` condition below. As such, a notice level log would be added when and only when bfd session state actually changes.

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+const map<string, sai_bfd_session_type_t> session_type_map =
+{
+    {"demand_active",       SAI_BFD_SESSION_TYPE_DEMAND_ACTIVE},
+    {"demand_passive",      SAI_BFD_SESSION_TYPE_DEMAND_PASSIVE},
+    {"async_active",        SAI_BFD_SESSION_TYPE_ASYNC_ACTIVE},
+    {"async_passive",       SAI_BFD_SESSION_TYPE_ASYNC_PASSIVE}
+};
+
+const map<sai_bfd_session_type_t, string> session_type_loopup =
```
prsunny, line 34: lookup -> typo?. I would suggest to name as `session_type_map`. Similarly for `session_state`

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,441 @@
...
+{
+    {SAI_BFD_SESSION_TYPE_DEMAND_ACTIVE,    "demand_active"},
+    {SAI_BFD_SESSION_TYPE_DEMAND_PASSIVE,   "demand_passive"},
+    {SAI_BFD_SESSION_TYPE_ASYNC_ACTIVE,     "async_active"},
+    {SAI_BFD_SESSION_TYPE_ASYNC_PASSIVE,    "async_passive"}
+};
+
+const map<sai_bfd_session_state_t, string> session_state_loopup =
```
prsunny, line 42: lookup -> typo?

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,441 @@
...
+        fvVector.emplace_back("multihop", "false");
+    }
+
+    if (alias != "default")
+    {
+        Port port;
+        if (!gPortsOrch->getPort(alias, port))
+        {
+            SWSS_LOG_ERROR("Failed to locate port %s", alias.c_str());
+            return true;
```
prsunny, line 311: this should return `false `as a retry is expected here. port may get created later

orchagent/bfdorch.cpp
```diff
@@ -0,0 +1,441 @@
...
+            sai_bfd_session_state_t state = bfdSessionState[i].session_state;
+
+            SWSS_LOG_INFO("Get BFD session state change notification id:%" PRIx64 " state: %s", id, session_state_loopup.at(state).c_str());
+
+            if (state != bfd_session_lookup[id].state)
+            {
+                auto key = bfd_session_lookup[id].peer;
+                m_stateBfdSessionTable.hset(key, "state", session_state_loopup.at(state));
+
+                SWSS_LOG_NOTICE("Updated BFD session state for %s to %s", key.c_str(), session_state_loopup.at(state).c_str());
```
prsunny, line 139: can you please rephrase this? `BFD session state for %s changed from %s to %s`. This shall be a critical log message

orchagent/bfdorch.cpp
shi-su: It is.. Thanks for catching it.

prsunny (dismissed):

orchagent/bfdorch.cpp
shi-su: Fixed for this and a few other places.

orchagent/bfdorch.cpp
shi-su: Updated as suggested.

orchagent/bfdorch.cpp
shi-su: Updated the log message.

prsunny (approved):

PR #1885, opened 2021-08-25
[portsorch] Add an extra check before setting oper speed to APPL_DB

Junchao-Mellanox:
**What I did**

If port operational status is up and port operational speed is 0, don't save it to APPL_DB

**Why I did it**

Avoid put speed 0 to APPL_DB.

Port operational status is up, but operational speed is 0. It could be a valid case because
port UP event is an aync event, the valid flow could be:
   1. port UP event comes
   2. SONiC gets the UP event and starts to process the event
   3. port goes down due to any reason
   4. SONiC reads the oper speed, and gets a 0 because oper_state is down.
And it could also be a bug. So, we log a warning here.

**How I verified it**

Run sonic-mgmt regression test cases

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -5252,22 +5252,8 @@ void PortsOrch::doTask(NotificationConsumer &consumer)
                 sai_uint32_t speed;
                 if (getPortOperSpeed(port, speed))
                 {
-                    if (speed != 0) 
-                    {
-                        SWSS_LOG_NOTICE("%s oper speed is %d", port.m_alias.c_str(), speed);
-                        updateDbPortOperSpeed(port, speed);
-                    }
-                    else 
-                    {
-                        // Port operational status is up, but operational speed is 0. It could be a valid case because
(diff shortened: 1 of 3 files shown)


orchagent/portsorch.cpp
```diff
@@ -5252,8 +5252,22 @@ void PortsOrch::doTask(NotificationConsumer &consumer)
                 sai_uint32_t speed;
                 if (getPortOperSpeed(port, speed))
                 {
-                    SWSS_LOG_NOTICE("%s oper speed is %d", port.m_alias.c_str(), speed);
-                    updateDbPortOperSpeed(port, speed);
+                    if (speed != 0)
```
prsunny, line 5255: Should we just return false from `getPortOperSpeed` if speed is 0. Just a thought?

prsunny (approved):

Junchao-Mellanox:
Hi @judyjoseph, could you please help cherry-pick to 202016.

PR #1890, opened 2021-08-27
[copp] Add ISIS, LDP and micro-BFD trap types to CoPP manager

abanu-ms:
* Add ISIS, LDP and micro-BFD traps to copp

* Removed unnecessary pytest markup which was skipping the copp UTs

* Add UTs for the new traps

**What I did**
I've added ISIS, LDP and micro-BFD trap types to copp and enabled the copp UTs.

**Why I did it**
To enable ISIS, LDP and micro-BFD traps. The latter two have been added to SAI in https://github.com/opencomputeproject/SAI/commit/516406550d2069be943c8d7fc19bf672b5ced763.

**How I verified it**
Updated the existing `test_copp.py` to include the new traps.

**Details if related**


prsunny (dismissed):

prsunny:
@abanu-ms , can you investigate the failure?

abanu-ms:
> @abanu-ms , can you investigate the failure?

It looks like it's trying to compile the code against a version of sonic-sairedis which uses an older version of SAI that didn't contain these new traps at that point. I'm not sure how to fix this.

PR #1877, opened 2021-08-17
[Flex Counters] Delay flex counters even if tables are present in the DB

shlomibitton:
**What I did**
Check if delay indicator flag is exist and 'true', if it does, skip the counter enablement.

**Why I did it**
Currently if flex counters tables are present in config DB the delay mechanism will not take place.
This change is to make sure the delay will take place even if the tables are present in the DB.

**How I verified it**
Observer counters are created after enable_counters script is called.

**Details if related**

PR #1899, opened 2021-09-09
[Flex-counters] Fix the delay of flex counters flow to prevent infinite loop

shlomibitton:

**What I did**
Add 'consumer.m_toSync.erase(it++)' before 'continue' statement on while loop.

**Why I did it**
Running fast-reboot with a delay indicator will result in an infinite loop, following PR: https://github.com/Azure/sonic-swss/pull/1877

**How I verified it**
Run fast-reboot with this change.

**Details if related**

### DIFF
--- a/orchagent/flexcounterorch.cpp
+++ b/orchagent/flexcounterorch.cpp
@@ -86,9 +86,9 @@ void FlexCounterOrch::doTask(Consumer &consumer)
 
         if (op == SET_COMMAND)
         {
-            auto it = std::find(std::begin(data), std::end(data), FieldValueTuple(FLEX_COUNTER_DELAY_STATUS_FIELD, "true"));
+            auto delay_it = std::find(std::begin(data), std::end(data), FieldValueTuple(FLEX_COUNTER_DELAY_STATUS_FIELD, "true"));
 
-            if (it != data.end())
+            if (delay_it != data.end())
             {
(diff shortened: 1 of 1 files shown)


stepanblyschak (dismissed):

orchagent/flexcounterorch.cpp
```diff
@@ -86,10 +86,11 @@ void FlexCounterOrch::doTask(Consumer &consumer)
 
         if (op == SET_COMMAND)
         {
-            auto it = std::find(std::begin(data), std::end(data), FieldValueTuple(FLEX_COUNTER_DELAY_STATUS_FIELD, "true"));
+            auto delay_it = std::find(std::begin(data), std::end(data), FieldValueTuple(FLEX_COUNTER_DELAY_STATUS_FIELD, "true"));
```
stepanblyschak, line 89: I think our code style is to use camelCaseStyle for local variables

shlomibitton: Fixed.

stepanblyschak (approved):

prsunny (approved):

prsunny:
Looks like a bug introduced as part of https://github.com/Azure/sonic-swss/pull/1877. @kcudnik for viz

kcudnik (approved):

PR #1902, opened 2021-09-10
[MuxOrch] Mux state order change

prsunny:
**What I did**
Orchagent sometimes took >50ms to finish the state change operations. This change is to update notification to xcvrd (via APPDB) after execution is complete.  This enables xcvrd to simulate the direction change *after* orchagent has completed the state transition. 

**Why I did it**
Reduce packet drop during state transition. 

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -394,15 +394,14 @@ void MuxCable::setState(string new_state)
     SWSS_LOG_NOTICE("[%s] Set MUX state from %s to %s", mux_name_.c_str(),
                      muxStateValToString.at(state_).c_str(), new_state.c_str());
 
-    // Update HW Mux cable state anyways
-    mux_cb_orch_->updateMuxState(mux_name_, new_state);
-
     MuxState ns = muxStateStringToVal.at(new_state);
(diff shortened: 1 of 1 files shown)


orchagent/muxorch.cpp
```diff
@@ -394,15 +394,14 @@ void MuxCable::setState(string new_state)
...
-    mux_cb_orch_->updateMuxState(mux_name_, new_state);
-
     MuxState ns = muxStateStringToVal.at(new_state);
 
     auto it = muxStateTransition.find(make_pair(state_, ns));
 
     if (it ==  muxStateTransition.end())
     {
+        // Update HW Mux cable state anyways
+        mux_cb_orch_->updateMuxState(mux_name_, new_state);
```
shi-su, line 404: I am assuming that it will never run into the if condition in line 417 in normal operation. But as this state is updated anyway, should it also be added there?

prsunny: Thats right. If we run into that state, it means a fatal error and that's why I din't add it there.

shi-su (approved):

PR #1906, opened 2021-09-15
[portsorch] Avoid orchagent crash when set invalid interface types to port

Junchao-Mellanox:
**What I did**

When setting invalid interface types to port, do not call handleSaiSetStatus to avoid orchagent exit.

**Why I did it**

Currently, when setting invalid interface types to port, orchagent would crash, this is not necessary since user still has chance to correct it. Error log see:

```
Jun  1 03:35:22.172518 r-leopard-41 ERR swss#orchagent: :- set: set status: SAI_STATUS_INVALID_ATTR_VALUE_0
Jun  1 03:35:22.172518 r-leopard-41 ERR swss#orchagent: :- handleSaiSetStatus: Encountered failure in set operation, exiting orchagent, SAI API: SAI_API_PORT, status: SAI_STATUS_INVALID_ATTR_VALUE_0

```

**How I verified it**

Manual test

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2013,16 +2013,8 @@ bool PortsOrch::setPortAdvInterfaceTypes(sai_object_id_t port_id, std::vector<ui
     attr.value.u32list.count = static_cast<uint32_t>(interface_types.size());
 
     status = sai_port_api->set_port_attribute(port_id, &attr);
-    if (status != SAI_STATUS_SUCCESS)
-    {
-        task_process_status handle_status = handleSaiSetStatus(SAI_API_PORT, status);
-        if (handle_status != task_success)
-        {
-            return parseHandleSaiStatusFailure(handle_status);
-        }
-    }
 
-    return true;
+    return status == SAI_STATUS_SUCCESS;
 }
 
 bool PortsOrch::getQueueTypeAndIndex(sai_object_id_t queue_id, string &type, uint8_t &index)
@@ -2836,7 +2828,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         else
                         {
                             SWSS_LOG_ERROR("Failed to set port %s AN from %d to %d", alias.c_str(), p.m_autoneg, an);
-                            it++;
+                            it = consumer.m_toSync.erase(it);
                             continue;
                         }
                     }
@@ -2872,7 +2864,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         if (!setPortSpeed(p, speed))
                         {
                             SWSS_LOG_ERROR("Failed to set port %s speed from %u to %u", alias.c_str(), p.m_speed, speed);
-                            it++;
+                            it = consumer.m_toSync.erase(it);
                             continue;
                         }
 
@@ -2920,7 +2912,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                             SWSS_LOG_ERROR("Failed to set port %s advertised speed from %s to %s", alias.c_str(),
                                                                                                    ori_adv_speeds.c_str(),
                                                                                                    adv_speeds_str.c_str());
-                            it++;
+                            it = consumer.m_toSync.erase(it);
                             continue;
                         }
                         SWSS_LOG_NOTICE("Set port %s advertised speed from %s to %s", alias.c_str(),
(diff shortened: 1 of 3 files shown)


orchagent/portsorch.cpp
```diff
@@ -2013,16 +2013,8 @@ bool PortsOrch::setPortAdvInterfaceTypes(sai_object_id_t port_id, std::vector<ui
     attr.value.u32list.count = static_cast<uint32_t>(interface_types.size());
 
     status = sai_port_api->set_port_attribute(port_id, &attr);
-    if (status != SAI_STATUS_SUCCESS)
```
shi-su, line 2016: This change removes the failure notification completely. I believe this is not the expected behavior. According to your issue description, it should only be skipped if the returned status is `SAI_STATUS_INVALID_ATTR_VALUE_0`. Therefore, I think the change here should be either checking for status before this part like https://github.com/Azure/sonic-swss/blob/master/orchagent/tunneldecaporch.cpp#L760, or add is as a specific case in the failure handling function like https://github.com/Azure/sonic-swss/pull/1815/files

shi-su (changes requested):

orchagent/portsorch.cpp
```diff
@@ -2836,7 +2844,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         else
                         {
                             SWSS_LOG_ERROR("Failed to set port %s AN from %d to %d", alias.c_str(), p.m_autoneg, an);
-                            it++;
```
shi-su, line 2839: Do we need to make this change?

orchagent/portsorch.cpp
```diff
@@ -2872,7 +2880,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         if (!setPortSpeed(p, speed))
                         {
                             SWSS_LOG_ERROR("Failed to set port %s speed from %u to %u", alias.c_str(), p.m_speed, speed);
-                            it++;
```
shi-su, line 2875: Do we need to make this change?

orchagent/portsorch.cpp
```diff
@@ -2920,7 +2928,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                             SWSS_LOG_ERROR("Failed to set port %s advertised speed from %s to %s", alias.c_str(),
                                                                                                    ori_adv_speeds.c_str(),
                                                                                                    adv_speeds_str.c_str());
-                            it++;
```
shi-su, line 2923: Do we need to make this change?

orchagent/portsorch.cpp
```diff
@@ -2960,7 +2968,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         if (!setPortInterfaceType(p.m_port_id, interface_type))
                         {
                             SWSS_LOG_ERROR("Failed to set port %s interface type to %s", alias.c_str(), interface_type_str.c_str());
-                            it++;
```
shi-su, line 2963: Do we need to make this change?

orchagent/portsorch.cpp
```diff
@@ -2999,7 +3007,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         if (!setPortAdvInterfaceTypes(p.m_port_id, adv_interface_types))
                         {
                             SWSS_LOG_ERROR("Failed to set port %s advertised interface type to %s", alias.c_str(), adv_interface_types_str.c_str());
-                            it++;
```
shi-su, line 3002: I guess the reason to make this change is to avoid retrying. However, I think the problem is we typically treat `false` return value as need retry and this behavior deviate from the expected one. I am not sure what would be the best solution, I think a reasonable one could be making the return value type of `setPortAdvInterfaceTypes` to `task_process_status`. And adding the status of `SAI_STATUS_INVALID_ATTR_VALUE_0` for `SAI_API_PORT` as a special case to return `task_failed` in the failure handling function like https://github.com/Azure/sonic-swss/pull/1815. Then we can make the return value of `task_process_status` type from `handleSaiSetStatus` as the return value of `setPortAdvInterfaceTypes` in case of SAI failure. It follows that we can choose whether to retry here.

Junchao-Mellanox:
@shi-su I have fixed all comments, could you please review again?

Junchao-Mellanox:
The test failure is not related to the PR.

shi-su (approved):
Could you please include the error message related to the changes in the PR description?

Junchao-Mellanox:
Updated description with error message

Junchao-Mellanox:
Hi @judyjoseph, could you help cherry-pick to 202106?

PR #1908, opened 2021-09-15
[pfcwd] Convert polling interval from ms to us in LUA scripts

volodymyrsamotiy:

**What I did**
Converted polling interval from milliseconds to microseconds in PFCWD LUA scripts.

**Why I did it**
PFCWD storm detection and restoration LUA scripts require values in microseconds.
Due to recent changes polling interval is now passed in milliseconds by "FlexCounter".
* Azure/sonic-sairedis#878

So need to convert values to microseconds (as it was before) in order to make PFCWD working,

**How I verified it**
Ran PFCWD tests from sonic-mgmt.

**Details if related**
N/A

### DIFF
--- a/orchagent/pfc_detect_mellanox.lua
+++ b/orchagent/pfc_detect_mellanox.lua
@@ -1,12 +1,12 @@
 -- KEYS - queue IDs
 -- ARGV[1] - counters db index
 -- ARGV[2] - counters table name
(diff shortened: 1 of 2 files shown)


neethajohn (approved):

dgsudharsan:
Required for 202106

PR #1813, opened 2021-06-30
Revert recirc port change

ysmanman:
**What I did**
Revert the special handling for recirc port.

**Why I did it**
Recirc port support has been added to recent versions of DNX SAI. The earlier special handling for recirc port is not required anymore and thus reverted.

**How I verified it**
I tested the PR with SAI 5.0.0.5 and verified inband port came up and functioned correctly.

**Details if related**

PR #1911, opened 2021-09-16
Completely remove "recirc port" code

qiluo-msft:
**What I did**
Completely remove "recirc port" code. The intention of 1813 is to revert 1530, but it is not complete.
Reverts below commits
```
7c7c4513 2021-07-06 | Revert recirc port change (#1813) [Song Yuan]
5ef4b382 2021-06-01 | Add support for recirc port and everflow (#1530) [Song Yuan]
```

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -44,7 +44,6 @@ extern sai_port_api_t *sai_port_api;
 
 extern sai_object_id_t  gSwitchId;
 extern PortsOrch*       gPortsOrch;
-extern string           gMySwitchType;
 
 using namespace std::rel_ops;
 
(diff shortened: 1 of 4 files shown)


qiluo-msft:
@ysmanman Could you also help review?

prsunny:
@kcudnik for viz

prsunny (approved):

kcudnik:
Address lgtm alerts ? also will this be back, since it seems more right to find root cause of this and fix it instead of remove it completely

qiluo-msft:
LGTM result is actually positive, fixing 2.
Not sure about the feature itself. I am following the PR owner's intention in #1813.

ysmanman:
Why do we want to revert this change? This will break everflow support.

abdosi:
@qiluo-msft why we are reverting ? Idea of PR https://github.com/Azure/sonic-swss/pull/1813 was to revert code that was needed when SAI did not have  correct support so when SAI has correct support this PR reverted it.

abdosi (changes requested):
Need clarification why PR needs revert ?

PR #1915, opened 2021-09-17
[PFC] Fixed the issue that PFC priority on port can't be disabled completely.

steven-guo-ec:
**What I did**
Fixed the issue that the pfc priority is still functioning when disabling the last pfc priority.

**Why I did it**
The status between SONiC and chip is not consistent when disabling the last pfc priority of a port. After disabling the last pfc priority of a port, `show pfc priority` command shows the priority has been disabled, but it still exist in chip. The root cause is that swss treat the configuration of disabling last priority as no change to pfc priority.

**How I verified it**
Verified the status between SONiC and chip became consistent after the modification.

**Details if related**

### DIFF
--- a/orchagent/qosorch.cpp
+++ b/orchagent/qosorch.cpp
@@ -1310,6 +1310,7 @@ task_process_status QosOrch::handlePortQosMapTable(Consumer& consumer)
     string op = kfvOp(tuple);
 
     sai_uint8_t pfc_enable = 0;
+    bool have_pfc = false;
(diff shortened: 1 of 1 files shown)


liat-grozovik:
@volodymyrsamotiy could you please help to review?
@steven-guo-ec could you please help to fix conflicts?

steven-guo-ec:
This change is same as #1962 , so I close this PR.

PR #1917, opened 2021-09-17
[neighsyncd] Add setting to globally enable/disable IPv6 link-local n…

msupuran:
…eighbors

* routes added by ISIS have IPv6 link local addresses as next hops - in order to have those next hops in APPL_DB, the corresponding neighbors are required to be present in APPL_DB as well
* avoid having to enable IPv6 link-local neighbors on each interface when all interfaces require this capability
* updated the isLinkLocalEnabled() function to first check for the presence of the new global_ipv6_link_local_neighbors setting in CONFIG_DB's DEVICE_METADATA table
Signed-off-by: Marius Supuran v-msupuran@microsoft.com

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


prsunny:
could you please sign the CLA?

msupuran:
@prsunny Done, sorry for the delay

prsunny:
@AkhileshSamineni , @venkatmahalingam , could you please review?

TACappleman:
@AkhileshSamineni @venkatmahalingam Will you be able to review this shortly? Please let us know if not so we can find alternative reviewers.

neighsyncd/neighsync.cpp
```diff
@@ -141,6 +142,20 @@ bool NeighSync::isLinkLocalEnabled(const string &port)
 {
```
venkatmahalingam, line 142: Update this function description as it checks global Ipv6 LLA status as well now.

neighsyncd/neighsync.cpp
```diff
@@ -141,6 +142,20 @@ bool NeighSync::isLinkLocalEnabled(const string &port)
...
     vector<FieldValueTuple> values;
 
+    if(m_cfgDeviceMetadataTable.get("localhost", values))
+    {
+        auto it = std::find_if(values.begin(), values.end(), [](const FieldValueTuple &t)
+                               { return t.first == "global_ipv6_link_local_neighbors"; });
+        if (it != values.end())
+        {
+            if (it->second == "enable")
+            {
```
venkatmahalingam, line 152: What if we have global IPv6 LLA status is enabled but port level IPv6 LLA status is disabled? do we handle this case?

venkatmahalingam:
> @AkhileshSamineni @venkatmahalingam Will you be able to review this shortly? Please let us know if not so we can find alternative reviewers.

Yes, the review is done, please reply to the comments.

neighsyncd/neighsync.cpp
msupuran: Solved

neighsyncd/neighsync.cpp
msupuran: We do not handle this case - if global IPv6 LLA status is enabled, we consider the IPv6 LLA status as enabled for each port, regardless of the actual setting

PR #1919, opened 2021-09-20
Fix failing DPB LAG tests

dgsudharsan:
**What I did**
Fixed the failed DPB LAG tests

**Why I did it**
The DPB LAG test was not passing. The test case was meant to cover the race condition scenario where on port removal, port delete arrives at orchagent before lag removal(The dependencies would be removed by breakout command automatically). So port delete should wait until dependencies are removed.
To mimic this, app DB LAG member table should be created first and removed after dpb command. The test was earlier mimicing with cfg DB LAG table (which is not the actual use case). Moreover mimicking config_Db lag table has a race condition where removeLAGMember is called by teamd during link down and thus the flow completes and when actual remove lag member from config_db is called, teamd crashes. This results in further lag remove call to be unhandled leaving residual LAG objects in asic DB failing the tests. 

**How I verified it**
Modifying the tests to operate on top of app DB instead of config DB.

**Details if related**

### DIFF
--- a/tests/test_port_dpb_lag.py
+++ b/tests/test_port_dpb_lag.py
@@ -2,6 +2,7 @@ import time
 import pytest
 from port_dpb import Port
 from port_dpb import DPB
+from swsscommon import swsscommon
 
 @pytest.mark.usefixtures('dpb_setup_fixture')
 @pytest.mark.usefixtures('dvs_lag_manager')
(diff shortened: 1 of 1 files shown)


tests/test_port_dpb_lag.py
```diff
@@ -10,17 +11,39 @@ def check_syslog(self, dvs, marker, log, expected_cnt):
...
         lag = "0001"
         p = Port(dvs, "Ethernet0")
         p.sync_from_config_db()
 
         # 1. Create PortChannel0001.
-        self.dvs_lag.create_port_channel(lag)
+        self.create_port_channel(dvs, lag)
 
         # 2. Add Ethernet0 to PortChannel0001.
-        self.dvs_lag.create_port_channel_member(lag, p.get_name())
```
prsunny, line 23: this one is a common fixture.. is there an issue with using common function?

prsunny:
@zhenggen-xu to review

prsunny (approved):

PR #1920, opened 2021-09-21
[gearbox] Add gearbox unit test

jimmyzhai:
**What I did**
* Add test_gearbox
* Add DVS_ENV for module specific dvs env variables

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -244,7 +244,7 @@ class DockerVirtualSwitch:
         name: str = None,
         imgname: str = None,
         keeptb: bool = False,
-        fakeplatform: str = None,
+        env: list = [],
         log_path: str = None,
         max_cpu: int = 2,
         forcedvs: bool = None,
@@ -356,8 +356,6 @@ class DockerVirtualSwitch:
             self.mount = f"/var/run/redis-vs/{self.ctn_sw.name}"
             ensure_system(f"mkdir -p {self.mount}")
 
-            self.environment = [f"fake_platform={fakeplatform}"] if fakeplatform else []
-
             kwargs = {}
             if newctnname:
                 kwargs["name"] = newctnname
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -236,7 +236,7 @@ def __init__(
         name: str = None,
         imgname: str = None,
         keeptb: bool = False,
-        fakeplatform: str = None,
+        env: list = [],
```
lguohan, line 239: what is the reason to change to a list?

jimmyzhai: List type, in case multiple environment variables

jimmyzhai: The parameter _environment_ of **self.client.containers.run** is a **list** type

tests/test_gearbox.py
```diff
@@ -0,0 +1,143 @@
...
+                (status, fvs) = port_table.get(j)
+                assert status == True
+                self.phys[i]["ports"][j] = dict(fvs)
+
+            lane_table = swsscommon.Table(db, sr.join([phy_table.getKeyName(i), "lanes"]))
+            lane_list = [x for x in lane_table.getKeys() if sr not in x]
+            self.phys[i]["lanes"] = {}
+            for j in lane_list:
+                (status, fvs) = lane_table.get(j)
+                assert status == True
```
lguohan, line 43: do we need to wait here?

jimmyzhai: Do not need. In dvs creation, it has verified all services ready. The table GEARBOX_TABLE in APP_DB has been created at one time by gearsyncd. In case some error happens, the assertion statements are enough.

tests/conftest.py
jimmyzhai: /azpw run

lguohan (approved):

PR #1923, opened 2021-09-22
Skip to create port if the lane set isn't available in ASIC

Pterosaur:

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


prsunny:
Can you provide details in description?

Pterosaur:
> Can you provide details in description?

Sure, I'm testing it. If this PR is available, I will add more description about it.

orchagent/portsorch.cpp
```diff
@@ -2737,6 +2738,29 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+                    if (!lane_set_is_available)
+                    {
+                        string tmp_lane_str = "";
+                        for (auto s : it->first)
+                        {
+                            tmp_lane_str += to_string(s) + " ";
+                        }
+                        tmp_lane_str = tmp_lane_str.substr(0, tmp_lane_str.size() - 1);
+
+                        SWSS_LOG_WARN("The lane set %s isn't available in ASIC", tmp_lane_str.c_str());
```
lguohan, line 2759: it looks like this conflicts with line 2764. in case the port is not available, the orchagent tends to create that port.

orchagent/portsorch.cpp
```diff
@@ -457,6 +457,7 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
         for (j = 0; j < attr.value.u32list.count; j++)
         {
             tmp_lane_set.insert(attr.value.u32list.list[j]);
+            m_availableLanes.insert(attr.value.u32list.list[j]);
```
lguohan, line 460: >m_availableLanes [](http://example.com/codeflow?start=12&length=16)

isn't m_availableLanes the same the same a tmp_lane_set, do we need to introduce this variable?

orchagent/portsorch.cpp
Pterosaur: I don't think so. This block is to check whether the expected lanes of ports are available. But line 2764 is to check whether a port is configured for creating.

orchagent/portsorch.cpp
Pterosaur: I think we need this variable. Because tmp_lane_set was stored in the m_portListLaneMap, but m_portListLaneMap may be changed in line 2731 which will make us lose the real available lane set.
https://github.com/Azure/sonic-swss/blob/b6c62b9a9cf4c4b4ccc5ab82ddb9732c899f8dea/orchagent/portsorch.cpp#L2723-L2737

PR #1935, opened 2021-09-28
[multi-asic] SAI_REDIS_SWITCH_ATTR_CONTEXT should be the last attribute. 

judyjoseph:
**What I did**
Moved the code so that SAI_REDIS_SWITCH_ATTR_CONTEXT  is the last attribute in create switch call.

**Why I did it**
To be in sync with new logic and master code. gAsicInstance is needed for multi-asic platforms which gets missed as sai_redis library expects SAI_REDIS_SWITCH_ATTR_CONTEXT  as the last attribute,

**How I verified it**
Get a multi-asic platform up.

**Details if related**

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -292,11 +292,6 @@ int main(int argc, char **argv)
         attrs.push_back(attr);
     }
 
-    /* Must be last Attribute */
-    attr.id = SAI_REDIS_SWITCH_ATTR_CONTEXT;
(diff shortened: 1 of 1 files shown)


qiluo-msft:
Why it should be the last attribute? Is it required by SAI definition?

kcudnik:
yes, is expected to be last attribute because of convenience, but we can change that since this is only for switch

PR #1936, opened 2021-09-29
[cfgmgr] Fix for STATE_DB Port check logic in cfgmgr daemons

qbdwlr:
**What I did**
Updated checks for Port entry in STATE_DB in portmgrd, teammgrd, and intfmgrd to additionally check for presence of "state" attribute.

**Why I did it**
Prior to recent commits for DPB, 3 daemons in cfgmgr (portmgrd, teammgrd, and intfmgrd) would not allow configuration to proceed for a specific PORT until portsyncd detected the presence of the kernel device (EthernetN) associated with the PORT and created the associated entry for the PORT in the STATE_DB with attribute "state" and value "ok".
With recent commits for DPB, this logic is now broken due to creation of PORT entry in the STATE_DB by PortsOrch with only "supported_speed" attribute.
The logic in the 3 cfgmgr daemons has been updated to additionally check for the "state" attribute when they detect the presence of the PORT entry in STATE_DB.

**How I verified it**
SONiC unit-tests and platform tests.

**Details if related**
N/A

### DIFF
--- a/cfgmgr/intfmgr.cpp
+++ b/cfgmgr/intfmgr.cpp
@@ -9,6 +9,7 @@
 #include "shellcmd.h"
 #include "macaddress.h"
 #include "warm_restart.h"
+#include <swss/redisutility.h>
 
 using namespace std;
 using namespace swss;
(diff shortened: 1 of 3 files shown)


prsunny:
Change lgtm, but this would be a working function broken. Do you know which PR introduced the change?

qbdwlr:
> Change lgtm, but this would be a working function broken. Do you know which PR introduced the change?

@prsunny :  This PR introduced PortsOrch::initPortSupportedSpeeds() which breaks the cfgmgr logic:  https://github.com/Azure/sonic-swss/pull/1714

prsunny:
@Junchao-Mellanox , could you please take a look at this? In #1714, it is not mentioned in description about any STATE_DB update and looks like there are impact on cfgmgrs.

prsunny (approved):

PR #1951, opened 2021-10-11
[orchagent] Add trap flow counter support

Junchao-Mellanox:
Added new VS test case in this PR, and the new VS test case requires https://github.com/Azure/sonic-swss/pull/1967 and https://github.com/Azure/sonic-sairedis/pull/954

**What I did**

Add trap flow counter support. See HLD: https://github.com/Azure/SONiC/pull/858

**Why I did it**

Flow counters are usually used for debugging, troubleshooting and performance enhancement processes. Host interface trap counter can get number of received traps per Trap ID.

**How I verified it**

Manual test/VS test/sonic mgmt test

**Details if related**


prsunny:
@dgsudharsan for review

dgsudharsan (approved):

dgsudharsan (dismissed):

keboliu (dismissed):

orchagent/flow_counter/flow_counter_handler.cpp
```diff
@@ -0,0 +1,49 @@
+#include <inttypes.h>
```
prsunny, line 1: As discussed, lets move this to flexcounter folder

orchagent/copporch.cpp
```diff
@@ -97,11 +121,23 @@ const string default_trap_group = "default";
 const vector<sai_hostif_trap_type_t> default_trap_ids = {
     SAI_HOSTIF_TRAP_TYPE_TTL_ERROR
 };
+const uint HOSTIF_TRAP_COUNTER_POLLING_INTERVAL_MS = 1000;
```
prsunny, line 124: Lets have the default value to 10sec

dgsudharsan (dismissed):

dgsudharsan (approved):

xumia:
@Junchao-Mellanox , if you want to trigger a check rerun, please try to use "/azpw run Azure.sonic-swss"

Junchao-Mellanox:
Hi @xumia, sure, thanks.

dgsudharsan (approved):

orchagent/copporch.cpp
```diff
@@ -18,8 +24,11 @@ extern sai_switch_api_t*    sai_switch_api;
 
 extern sai_object_id_t      gSwitchId;
 extern PortsOrch*           gPortsOrch;
+extern Directory<Orch*>     gDirectory;
 extern bool                 gIsNatSupported;
 
+#define FLEX_COUNTER_UPD_INTERVAL 1
```
prsunny, line 30: Please change this to 10.

prsunny:
@Junchao-Mellanox , I think we have discussed this before, please do not "close" and "reopen" the PR multiple times. It is not clear whether you really want to close the PR without merging. If the intention is to to retrigger tests, please do "/azpw run".

orchagent/copporch.cpp
Junchao-Mellanox: This interval here is the timer to add flex counter to flex DB, it is not the query interval.

prsunny: But this is the interval where the select gets triggered, right? So why select this every 1 sec?

Junchao-Mellanox: Here we use a timer to timely add flex counter to FLEX COUNTER DB due to comment: [#1859 (comment)](https://github.com/Azure/sonic-swss/pull/1859#discussion_r727988279) . We set the interval 1 here because we want flex counter to be created as soon as possible. It won't cost much CPU because:

- the timer will be started when the first item is added to queue "m_pendingAddToFlexCntr"
- when the queue "m_pendingAddToFlexCntr" is empty, the timer will be stopped

Please check copporch.cpp line 770:
```cpp
    auto was_empty = m_pendingAddToFlexCntr.empty();
    m_pendingAddToFlexCntr[counter_id] = trap_name;

    if (was_empty)
    {
        m_FlexCounterUpdTimer->start();
    }
```
and copporch.cpp line 1229

```cpp
    if (m_pendingAddToFlexCntr.empty())
    {
        m_FlexCounterUpdTimer->stop();
    }
```

prsunny (approved):

PR #1955, opened 2021-10-12
[vnetorch] ECMP for vnet tunnel routes with endpoint health monitor

shi-su:
**What I did**
Add functions to create/remove next hop groups for vnet tunnel routes.
Count the reference count of next hop groups to create and remove as needed.
Share the counter of next hop groups with routeorch.
Adapt route endpoint according to the BFD state of endpoints.

**Why I did it**
To add support for overlay ECMP.

**How I verified it**
Verify ECMP groups are properly created and removed with the functions.
**Details if related**

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -813,7 +813,7 @@ bool VNetRouteOrch::removeNextHopGroup(const string& vnet, const NextHopGroupKey
 template<>
 bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipPrefix,
                                                NextHopGroupKey& nexthops, string& op, 
-                                               bool temp_change, const map<NextHopKey, IpAddress>& monitors)
+                                               const map<NextHopKey, IpAddress>& monitors)
 {
     SWSS_LOG_ENTER();
 
@@ -851,7 +851,6 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
     if (op == SET_COMMAND)
     {
         sai_object_id_t nh_id;
-        /* The route is pointing to one single endpoint */
         if (!hasNextHopGroup(vnet, nexthops))
         {
             setEndpointMonitor(vnet, monitors, nexthops);
@@ -891,7 +890,7 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
             }
             else
             {
-                if (it_route == syncd_tunnel_routes_[vnet].end() || temp_change)
+                if (it_route == syncd_tunnel_routes_[vnet].end())
                 {
                     route_status = add_route(vr_id, pfx, nh_id);
                 }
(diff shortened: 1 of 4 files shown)


orchagent/vnetorch.h
```diff
@@ -283,10 +285,25 @@ struct VNetNextHopObserverEntry
...
+struct BfdSessionInfo
+{
+    sai_bfd_session_state_t bfd_state;
+    std::map<std::string, NextHopKey> endpoints;
+};
+
 typedef std::map<NextHopGroupKey, NextHopGroupInfo> VNetNextHopGroupInfoTable;
 typedef std::map<IpPrefix, NextHopGroupKey> VNetTunnelRouteTable;
+typedef std::map<IpAddress, BfdSessionInfo> BfdSessionTable;
+typedef std::map<IpAddress, VNetNextHopInfo> VNetEndpointInfoTable;
```
prsunny, line 304: Lets discuss this map? I'm not fully understanding the purpose

orchagent/vnetorch.cpp
```diff
@@ -863,13 +881,32 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
...
-                route_status = add_route(vr_id, pfx, nh_id);
+                if (it_route != syncd_tunnel_routes_[vnet].end())
+                {
+                    del_route(vr_id, pfx);
+                }
             }
             else
             {
-                route_status = update_route(vr_id, pfx, nh_id);
+                if (it_route == syncd_tunnel_routes_[vnet].end() || temp_change)
```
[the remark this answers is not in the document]
shi-su: The aim of this is for route adaptation according to BFD state. For these routes, we only want to change the routes in hardware, while the original configuration needs to be saved.

orchagent/vnetorch.h
shi-su: The purpose of this map is to track the BFD state and nexthop groups of each endpoint so that the nexthop group can be updated accordingly. I believe we do need some discussion about the implementation. Will get it scheduled.

orchagent/vnetorch.cpp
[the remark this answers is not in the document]
shi-su: Made the health monitoring related changes a new function.

orchagent/vnetorch.cpp
```diff
@@ -949,6 +1010,12 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
         }
 
         vrf_obj->removeRoute(ipPrefix);
+
+        removeRouteState(vnet, ipPrefix);
+    }
+    else
```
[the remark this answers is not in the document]
shi-su: Removed.

orchagent/vnetorch.h
```diff
@@ -326,6 +357,11 @@ class VNetRouteOrch : public Orch2, public Subject
     VNetNextHopObserverTable next_hop_observers_;
     std::map<std::string, VNetNextHopGroupInfoTable> syncd_nexthop_groups_;
     std::map<std::string, VNetTunnelRouteTable> syncd_tunnel_routes_;
+    BfdSessionTable bfd_sessions_;
+    std::map<std::string, VNetEndpointInfoTable> nexthop_info_;
+    ProducerStateTable m_bfdSessionProducer;
```
prsunny, line 362: could you please use the same naming convention as in the rest of class

shi-su: Updated naming to match naming convention.

prsunny (approved):

PR #1956, opened 2021-10-13
[Buffermgr]Graceful handling of buffer model change

dgsudharsan:
Signed-off-by: Sudharsan Dhamal Gopalarathnam <sudharsand@nvidia.com>

**What I did**
Handled buffer model changes gracefully.  The use case is when config_db.json is loaded initially with no buffer configurations and when user executes 'config qos reload', it would result in buffer model changing from traditional to dynamic. This transition requires swss restart which will be shown as message to user.

**Why I did it**
When config qos reload is given in platforms with dynamic buffer model, swss restart is required. However, if swss is not restarted the buffermgrd will stay in static model and will program the orchagent with 'size' field not set in buffer pool due to dynamic mode checks in jinja2 template. This will result in orchagent calling SAI without SAI_BUFFER_POOL_ATTR_SIZE which is mandatory attribute.
Since this results in a SAI create API failure, it will result in orchagent crash.
So when the mode is changed to dynamic, buffermgrd will not process any configurations when running in static mode.

**How I verified it**
Added UT. Manually verified that config qos reload doesn't crash.

**Details if related**

### DIFF
--- a/tests/test_buffer_mode.py
+++ b/tests/test_buffer_mode.py
@@ -12,9 +12,16 @@ class TestBufferModel(object):
         app_db = dvs.get_app_db()
         keys = app_db.get_keys("BUFFER_POOL_TABLE")
         num_keys =  len(keys)
-        fvs = {'buffer_model' : 'dynamic'}
-        config_db.update_entry("DEVICE_METADATA", "localhost", fvs)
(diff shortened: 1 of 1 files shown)


dgsudharsan:
@neethajohn @prsunny Can you please add required for 202106 label?

PR #1959, opened 2021-10-13
[switchorch] Implement VXLAN src port range feature 

ayurkiv-nvda:

**What I did**
Added ability to configure UDP VXLAN src port range with help of new added parameters (vxlan_sport and vxlan_mask)
Сorresponding SAI attributes: SAI_SWITCH_TUNNEL_ATTR_VXLAN_UDP_SPORT and SAI_SWITCH_TUNNEL_ATTR_VXLAN_UDP_SPORT_MASK

**Why I did it**
Implemented configuration of  VXLAN src port feature according to requirements

**How I verified it**
Use appropriate VS test

`sudo pytest --dvsname=vs tests/test_switch.py::TestSwitch::test_switch_attribute -v -s`

Create json file with params.
Use swssconfig to enable feature and write config params to APP_DB : swssconfig /etc/swss/config.d/switch.json
**Details if related**
```
switch.json:
[
    {
        "SWITCH_TABLE:switch": {
			"vxlan_sport": "0xFFA0",
			"vxlan_mask": "3"
        },
        "OP": "SET"
    }
]
```

### DIFF
--- a/orchagent/switchorch.cpp
+++ b/orchagent/switchorch.cpp
@@ -55,29 +55,6 @@ SwitchOrch::SwitchOrch(DBConnector *db, vector<TableConnector>& connectors, Tabl
     querySwitchTpidCapability();
     auto executorT = new ExecutableTimer(m_sensorsPollerTimer, this, "ASIC_SENSORS_POLL_TIMER");
     Orch::addExecutor(executorT);
-
-    vector<sai_attribute_t> attrs;
-    sai_attribute_t attr;
-    sai_status_t status;
-
-    attr.id = SAI_SWITCH_TUNNEL_ATTR_TUNNEL_TYPE;
-    attr.value.s32 = SAI_TUNNEL_TYPE_VXLAN;
-    attrs.push_back(attr);
-    attr.id = SAI_SWITCH_TUNNEL_ATTR_TUNNEL_VXLAN_UDP_SPORT_MODE;
-    attr.value.s32 = SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_EPHEMERAL;
-    attrs.push_back(attr);
-
-    status = sai_switch_api->create_switch_tunnel(&switch_tunnel_id, gSwitchId, static_cast<uint32_t>(attrs.size()), attrs.data());
-
-    if (status != SAI_STATUS_SUCCESS)
-    {
-        SWSS_LOG_ERROR("Failed to create switch_tunnel object");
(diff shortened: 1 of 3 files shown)


orchagent/switchorch.h
```diff
@@ -35,11 +35,13 @@ class SwitchOrch : public Orch
     void doAppSwitchTableTask(Consumer &consumer);
     void initSensorsTable();
     void querySwitchTpidCapability();
+    sai_status_t setSwitchTunnelVxlanParams(swss::FieldValueTuple &val);
 
     swss::NotificationConsumer* m_restartCheckNotificationConsumer;
     void doTask(swss::NotificationConsumer& consumer);
     swss::DBConnector *m_db;
     swss::Table m_switchTable;
+    sai_object_id_t switch_tunnel_id;
```
prsunny, line 44: Please rename to `m_switchTunnelId`

orchagent/switchorch.cpp
```diff
@@ -49,6 +55,29 @@ SwitchOrch::SwitchOrch(DBConnector *db, vector<TableConnector>& connectors, Tabl
...
+
+    vector<sai_attribute_t> attrs;
+    sai_attribute_t attr;
+    sai_status_t status;
+
+    attr.id = SAI_SWITCH_TUNNEL_ATTR_TUNNEL_TYPE;
+    attr.value.s32 = SAI_TUNNEL_TYPE_VXLAN;
+    attrs.push_back(attr);
+    attr.id = SAI_SWITCH_TUNNEL_ATTR_TUNNEL_VXLAN_UDP_SPORT_MODE;
+    attr.value.s32 = SAI_TUNNEL_VXLAN_UDP_SPORT_MODE_EPHEMERAL;
```
prsunny, line 67: This is not required as the default is "EPHEMERAL". Moreover, it might be possible for SAI to return a failure if this attribute is not supported. So I don't really find a reason for setting it in application.

prsunny (changes requested):

orchagent/switchorch.cpp
ayurkiv-nvda: If  "EPHEMERAL" is default attr so SAI must support it anyway isn't it? Please correct me if I wrong. 

Anyway you proposal is to move **create_switch_tunnel(&switch_tunnel_id,** from constructor to **setSwitchTunnelVxlanParams** function?

prsunny: Thats not true. Attribute set support may or may not be present for all SAI implementation. I don't think we need to explicitly set default value here.

prsunny (approved):

PR #1987, opened 2021-10-26
Don't handle buffer pool watermark during warm reboot reconciling

stephenxs:
**What I did**
Don't handle buffer pool watermark during warm reboot reconciling

**Why I did it**
This is to fix the community issue Azure/sonic-sairedis#862 and Azure/sonic-buildimage#8722

**How I verified it**
Perform a warm reboot. Check whether
- buffer pool watermark handling is skipped during reconciling and handled after it.
- other watermark handling is handled during reconciling as it was before.

**Details if related**
The warm reboot flow is like this:
1. System starts. Orchagent fetches the items from database stored before warm reboot and pushes them into `m_toSync` of all orchagents. This is done by `bake`, which can be overridden by sub orchagent.
2. All sub orchagents handle the items in `m_toSync`. At this point, any notification from redis-db is blocked.
3. Warm reboot converges.
4. Orchagent starts to handle notifications from redis-db.

The fix is like this: in FlexCounterOrch::bake. the buffer pool watermark handling is skipped.

### DIFF
--- a/orchagent/flexcounterorch.cpp
+++ b/orchagent/flexcounterorch.cpp
@@ -1,5 +1,4 @@
 #include <unordered_map>
-#include "flexcounterorch.h"
 #include "portsorch.h"
 #include "fabricportsorch.h"
 #include "select.h"
@@ -43,6 +42,7 @@ unordered_map<string, string> flexCounterGroupMap =
 
 FlexCounterOrch::FlexCounterOrch(DBConnector *db, vector<string> &tableNames):
     Orch(db, tableNames),
+    m_flexCounterConfigTable(db, CFG_FLEX_COUNTER_TABLE_NAME),
     m_flexCounterDb(new DBConnector("FLEX_COUNTER_DB", 0)),
     m_flexCounterGroupTable(new ProducerTable(m_flexCounterDb.get(), FLEX_COUNTER_GROUP_TABLE))
 {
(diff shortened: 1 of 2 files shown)


stepanblyschak:
@stephenxs So this issue exists because we call ```get_buffer_pool_stats``` and ```clear_buffer_pool_stats``` in ```generateBufferPoolWatermarkCounterIdList``` that you want to postpone in this change?
Why not using ```sai_query_stats_capability``` which tells whether object supports READ, READ_CLEAR capabilities?

stephenxs:
> @stephenxs So this issue exists because we call `get_buffer_pool_stats` and `clear_buffer_pool_stats` in `generateBufferPoolWatermarkCounterIdList` that you want to postpone in this change? Why not using `sai_query_stats_capability` which tells whether object supports READ, READ_CLEAR capabilities?

@stepanblyschak 
Thanks for the suggestion. Yes, that's the motivation but I'm afraid `sai_query_stats_capability` is not able to satisfy the requirement.
```
/**
 * @brief Query statistics capability for statistics bound at object level
 *
 * @param[in] switch_id SAI Switch object id
 * @param[in] object_type SAI object type
 * @param[inout] stats_capability List of implemented enum values, and the statistics modes (bit mask) supported per value
 *
 * @return #SAI_STATUS_SUCCESS on success, #SAI_STATUS_BUFFER_OVERFLOW if lists size insufficient, failure status code on error
 */
```
`sai_query_stats_capability` provides the ability to query capability on a per-object-type basis. 
However, it is required to do it on a per-object basis here. Eg. on some platforms, `SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES` can be fetched on `ingress_lossless_pool` only. (This code is not here for now. It had been there but was reverted because of the bug we try to fix by this PR. We need to re-add it after the bug is fixed.)

stepanblyschak (approved):

PR #1993, opened 2021-10-28
[aclorch] add generic AclOrch::updateAclRule() method

stepanblyschak:
**What I did**

I added generic AclOrch::updateAclRule() implementation. This implementation takes a newly constructed shared_ptr\<AclRule\> provided by the user, calculates a difference between priority, matches and actions and calls sai_acl_api->set_acl_entry_attribute() for those attributes that were changed. The derivatives of AclRule can customize the behaviour of AclRule::update(const AclRule& updatedRule) by polymorphic override, although in that case the derivative needs dynamic_cast the updatedRule to a concrete derivative type.
Added unit test to cover update scenario. Currently this new API is not used by any clients nor handling ACL_RULE table updates. This API will be used by PBH ACL rule update flow.

**Why I did it**

To support the scenario for updating PBH ACL rule in the future.

**How I verified it**

Temporary changed the ACL_RULE table update to leverage this new API and tested updates through CONFIG_DB and unit test coverage.

**Details if related**


bingwang-ms:
Reviewing

orchagent/aclorch.cpp
```diff
@@ -650,6 +631,262 @@ void AclRule::updateInPorts()
     }
 }
 
+
+bool AclRule::update(const AclRule& updatedRule)
```
bingwang-ms, line 635: Could you confirm if the update of priority/match rules/actions is supported? I remembered that we only support updating binding ports before.

stepanblyschak: Yes, with this API priority, match and actions of ACL rule can be updated via SAI set API.
We support updating bind ports on ACL table.

bingwang-ms:
Please ping me when the PR is ready for review. Thanks

stepanblyschak:
@bingwang-ms This PR is ready.

orchagent/aclorch.cpp
```diff
@@ -3324,6 +3631,26 @@ bool AclOrch::updateAclRule(string table_id, string rule_id, bool enableCounter)
     return true;
 }
 
+bool AclOrch::updateAclRule(shared_ptr<AclRule> updatedRule)
```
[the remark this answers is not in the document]
stepanblyschak: @bingwang-ms This function will be used by PbhOrch code. PbhOrch needs to implement an update functionality of PBH rules. The PbhOrch will construct a new PBH rule - an instance of AclRule and here we will perfrom the diff between matches, actions and priority and set what was changed.

orchagent/aclorch.cpp
```diff
@@ -415,42 +438,12 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
...
-                                                           it.second.aclaction.parameter))
-            {
-                SWSS_LOG_ERROR("Action %s:%s is not supported by ASIC",
-                               attr_name.c_str(), attr_value.c_str());
-                return false;
-            }
-        }
-    }
-    return true;
+    return false;
```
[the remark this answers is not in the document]
stepanblyschak: @bingwang-ms Made it pure virtual

orchagent/aclorch.cpp
stepanblyschak: @bingwang-ms @prsunny I did not find an API in SAI to query whether updating attributes is supported.

bingwang-ms: That my only concern for this change since I noticed below lines
https://github.com/Azure/sonic-swss/blob/8727ae5c323d99e53dddcf84102372abced94568/orchagent/aclorch.cpp#L3292-3294
```
        default:
            SWSS_LOG_ERROR("Acl rule update not supported for attr name %s", attr_name.c_str());
```

stepanblyschak:
There is some issue with LGTM:
```
[2021-11-18 13:03:05] [build-stdout] g++ -DHAVE_CONFIG_H -I. -I.. -I .. -g -DNDEBUG  -ansi -fPIC -std=c++11 -Wall -Wcast-align -Wcast-qual -Wconversion -Wdisabled-optimization -Werror -Wextra -Wfloat-equal -Wformat=2 -Wformat-nonliteral -Wformat-security -Wformat-y2k -Wimport -Winit-self -Winvalid-pch -Wlong-long -Wmissing-field-initializers -Wmissing-format-attribute -Wmissing-include-dirs -Wmissing-noreturn -Wno-aggregate-return -Wno-padded -Wno-switch-enum -Wno-unused-parameter -Wpacked -Wpointer-arith -Wredundant-decls -Wshadow -Wstack-protector -Wstrict-aliasing=3 -Wswitch -Wswitch-default -Wunreachable-code -Wunused -Wvariadic-macros -Wno-write-strings -Wno-missing-format-attribute -Wno-long-long  -I/usr/include/libnl3  -Wdate-time -D_FORTIFY_SOURCE=2  -g -O2 -fdebug-prefix-map=/opt/src/sonic-swss-common=. -fstack-protector-strong -Wformat -Werror=format-security -c -o tests-saiaclschema_ut.o `test -f 'saiaclschema_ut.cpp' || echo './'`saiaclschema_ut.cpp
[2021-11-18 13:03:05] [build-stderr] saiaclschema_ut.cpp:1:10: fatal error: gmock/gmock.h: No such file or directory
[2021-11-18 13:03:05] [build-stderr]     1 | #include <gmock/gmock.h>
[2021-11-18 13:03:05] [build-stderr]       |          ^~~~~~~~~~~~~~~
[2021-11-18 13:03:05] [build-stderr] compilation terminated.
```

orchagent/aclorch.cpp
stepanblyschak: Lines that you are refering to are related to the current approach of doing AclRule update. As you can see in the code it is implemented only for SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS (That's why there is an error message that you pasted in the comment, not because it is not supported by SAI or SAI vendor). Let's say there is an Orch which wants to use some actions, matches that AclOrch does not know (like PbhOrch for example uses SET_LAG_HASH/SET_ECMP_HASH). Then, every time we add some new action or match that is only relevant to that new Orch we would have to change AclOrch and add a special logic for handling update for that new action and match. Appearantly the AclOrch becomes too huge a special pieces of which are related to other orch's functionality. So, instead of current approach we propose a generic enough AclOrch::updateAclRule that can work with any SAI match, any SAI action. I haven't removed the current ```bool AclOrch::updateAclRule(string table_id, string rule_id, string attr_name, void *data, bool oper)``` method because it is in use by MuxOrch, PfcWdOrch and moving them to a new API ```bool AclOrch::updateAclRule(shared_ptr<AclRule> updatedRule)``` requires a bit of refactor which I plan to do in a different PR. Eventually I propose to remove the old ```updateAclRule``` with the new one.

bingwang-ms (approved):

PR #1996, opened 2021-11-01
[Reclaiming buffer] Common code update

stephenxs:
**What I did**
Common code update for reclaiming buffer.

1. Loading zero_profiles when dynamic buffer manager starting
    The buffer manager won't consume it for now. This is to pass Azure CI.
2. Support removing a buffer pool.
3. Support exposing maximum PGs and queues per port
4. Support transmit between bitmap and map string
5. Change the log severity from `ERROR` to `NOTICE` when parsing buffer profile from buffer profile list failed. Typically this can be resolved by retrying.
    The severity of similar log when parsing buffer PG and queue is already `NOTICE`.

**Why I did it**

To split large PR into smaller ones and help pass CI.

**How I verified it**

vs test and regression test.

**Details if related**


keboliu (dismissed):

orchagent/portsorch.cpp
```diff
@@ -4084,9 +4084,19 @@ void PortsOrch::initializePortMaximumHeadroom(Port &port)
         return;
     }
 
-    vector<FieldValueTuple> fvVector;
     port.m_maximum_headroom = attr.value.u32;
-    fvVector.emplace_back("max_headroom_size", to_string(port.m_maximum_headroom));
+}
+
+void PortsOrch::initializePortBufferMaximumParameters(Port &port)
```
neethajohn, line 4090: can this be combined with initializePortMaximumHeadroom?

orchagent/orch.cpp
```diff
@@ -635,6 +635,138 @@ bool Orch::parseIndexRange(const string &input, sai_uint32_t &range_low, sai_uin
...
+set<string> Orch::generateIdListFromMap(unsigned long idsMap, sai_uint32_t maxId)
+{
+    unsigned long currentIdMask = 1;
+    bool started = false, needGenerateMap = false;
+    sai_uint32_t lower, upper;
+    set<string> extraIdsToReclaim;
+    for (sai_uint32_t id = 0; id <= maxId; id ++)
+    {
+        // currentIdMask represents the bit mask corresponding to id: (1<<id)
+        if (idsMap & currentIdMask)
```
neethajohn, line 691: This logic does not seem to work if the bit map is 11100110 (basically if there is a range whose upper bound is maxId)

tests/test_buffer_dynamic.py
```diff
@@ -636,3 +636,40 @@ def test_autoNegPort(self, dvs, testlog):
 
         # Shutdown interface
         dvs.runcmd('config interface shutdown Ethernet0')
+
+    def test_removeBufferPool(self, dvs, testlog):
+        self.setup_db(dvs)
```
neethajohn, line 641: please put this in a try-finally block so that the cleanup always happens

orchagent/bufferorch.cpp
```diff
@@ -210,6 +210,12 @@ bool BufferOrch::isPortReady(const std::string& port_name) const
     return result;
 }
 
+void BufferOrch::clearBufferPoolWatermarkCounterCounterIdList(const sai_object_id_t object_id)
```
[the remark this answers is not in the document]
stephenxs: It is required for reclaiming buffer. 
The zero buffer pool will be removed once the last admin down port is started up.
Originally, a buffer pool can not be removed cleanly because the counter id list was not removed. This is to support it.

orchagent/portsorch.cpp
stephenxs: ok. let's change the name to `initializePortBufferMaximumParameters` and then merge them together.

orchagent/orch.cpp
```diff
@@ -635,6 +635,138 @@ bool Orch::parseIndexRange(const string &input, sai_uint32_t &range_low, sai_uin
...
+ * Generates the bit map representing the idsMap in string
+ * Args:
+ *      idsStr: The string representing the IDs.
+ *              Typically it's part of a key of BUFFER_QUEUE or BUFFER_PG, like "3-4" from "Ethernet0|3-4"
+ * Return:
+ *      idsMap: The bitmap of IDs. The LSB stands for ID 0.
+ *
+ * Example:
+ *      Input idsMap: 3-4
+ *      Return: 00001100b
```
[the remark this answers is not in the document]
stephenxs: Fixed.

orchagent/portsorch.cpp
stephenxs: Fixed

tests/test_buffer_dynamic.py
stephenxs: Fixed

orchagent/orch.cpp
stephenxs: Can you elaborate? I didn't find any issue and this is exactly how we use this function in reclaiming buffer.

orchagent/orch.cpp
```diff
@@ -635,6 +635,138 @@ bool Orch::parseIndexRange(const string &input, sai_uint32_t &range_low, sai_uin
...
+ * Example:
+ *     Input idsMap: 00100110b, maxId: 8
+ *     Return vector: ["1-2", "5"]
+ */
+set<string> Orch::generateIdListFromMap(unsigned long idsMap, sai_uint32_t maxId)
+{
+    unsigned long currentIdMask = 1;
+    bool started = false, needGenerateMap = false;
+    sai_uint32_t lower, upper;
+    set<string> extraIdsToReclaim;
```
stephenxs, line 687: Rename it.

orchagent/orch.cpp
neethajohn: how does 9-15 get inserted into extraIdsToReclaim since needGenerateMap is set to False after you have inserted 7 and for 9-15,  L691 will be true and L709 will always be False. I am assuming the maxId in this case is 15 and after this we are done with the for loop. Did I miss something?

stephenxs: In case the input is `111111010000000`, the maxId should be 16. (refer to the example: 0010011b, maxId: 8)
I believe that's the confusion, right?

stephenxs: Adjusted the comment to make it clear.

neethajohn (approved):

orchagent/bufferorch.cpp
```diff
@@ -460,6 +466,7 @@ task_process_status BufferOrch::processBufferPool(KeyOpFieldsValuesTuple &tuple)
 
         if (SAI_NULL_OBJECT_ID != sai_object)
         {
+            clearBufferPoolWatermarkCounterIdList(sai_object);
```
neethajohn, line 469: check if 'm_isBufferPoolWatermarkCounterIdListGenerated' is True before the clear?

stephenxs: Sounds reasonable. Let’s check it in the clear function

stephenxs: Added checking in `clearBufferPoolWatermarkCounterIdList`: only remove it if `m_isBufferPoolWatermarkCounterIdListGenerated` is true.

neethajohn (approved):

keboliu (approved):

PR #1998, opened 2021-11-01
[macsec] Fix setting network_id in macsec

byu343:
**What I did**
It looks like we should use network_id in set_network, although I didn't find a related issue in running macsec before this change.
**Why I did it**

**How I verified it**

**Details if related**


Pterosaur (approved):

kenneth-arista:
Tagging @arlakshm for awareness

PR #1999, opened 2021-11-01
[macsec] Remove setting SCI to data acl entry

byu343:
**What I did**
The reason for this change is SAI_ACL_ENTRY_ATTR_FIELD_MACSEC_SCI is not supported by the Credo driver. For Credo B52 chip, its classification engine does not support SCI; still the SCI field in be verified by the security engine through SAI_MACSEC_SC_ATTR_MACSEC_SCI.  
**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/macsecorch.cpp
+++ b/orchagent/macsecorch.cpp
@@ -2283,14 +2283,6 @@ bool MACsecOrch::createMACsecACLDataEntry(
     attr.value.aclaction.parameter.s32 = SAI_PACKET_ACTION_DROP;
     attr.value.aclaction.enable = true;
     attrs.push_back(attr);
-    if (sci_in_sectag)
-    {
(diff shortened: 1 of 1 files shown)


Pterosaur:
Due to discussion offline, there is another method to solve this issue.

SarathBug:
Hi @Pterosaur 

Could you please help with alternate method to bypass this error?

Pterosaur:
> Hi @Pterosaur
> 
> Could you please help with alternate method to bypass this error?

Hi @SarathBug ,
Here is our discussion from byu's e-mail:
"
Using SCI in the ACL or classification stage is not supported by B52 hardware. Another way to solve this problem is to pass the value to the driver, but ignore it inside the driver.  I would prefer this way now, then the PR can be discarded.
"

SarathBug:
Hi @Pterosaur,

Thanks for clarifying.

PR #2002, opened 2021-11-02
[voq][syncd][bcmsai] Adjusting syncd response timeout during switch create

vganesan-nokia:
**What I did**

Fix for syncd response time out for switch create request from orchagent

**Why I did it**

In VOQ based chassis where syncd uses BCM VOQ SAI, if there are large number of front panel ports, SAI takes more than 1 minutes to complete the switch create initialization. Because of this, the switch create request sent by orchagent is not getting response within the default response wait time of 1 minute. So the orchagent declares switch create failure and crashes.

The number of ports need to be initialized by SAI depends on number of ports per asic and total number of system ports configured in the system. The total number of system ports in the system in turn depends on number of line cards supported, number of asics per line card and number of ports supported by each asic. Therefore in a fully populated system, which is an often expected scenario, this crashing will happen.

To fix this, in orchagent, the syncd response time out is set to 35 minutes before sending request for switch create and is set back to default wait time after the switch create. For now, since the problem is observed for BCM platforms, this adjustment is done only for BCM platform.

**How I verified it**

- In a VOQ based chassis that uses BCM SAI, populate the chassis with a  line card that has asics with more than 62 ports.
- Configure 192+ system ports
- Reboot the line card
- Observe that the orchagent does not crash with swith create failure
- Observe that the syncd does not show reponse time out error
- Observe that swss and syncd dockers are up and running and all the interfaces come up.

**Details if related**

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -81,6 +81,7 @@ int32_t gVoqMaxCores = 0;
 uint32_t gCfgSystemPorts = 0;
 string gMyHostName = "";
 string gMyAsicName = "";
+uint64_t gSyncdResponseTimeout = ( 35 * SAI_REDIS_DEFAULT_SYNC_OPERATION_RESPONSE_TIMEOUT);
 
 void usage()
 {
@@ -574,6 +575,29 @@ int main(int argc, char **argv)
     attr.value.u64 = gSwitchId;
     attrs.push_back(attr);
 
+
+    char *platform = getenv("platform");
+    if (platform && strstr(platform, BRCM_PLATFORM_SUBSTRING))
+    {
+        /* We set this long timeout in order for orchagent to wait enough time for
+         * response from syncd. It is needed since switch create takes more time
+         * than default time to create switch if there are lots of front panel ports
+         * and systems ports to initialize
+         */
+        attr.id = SAI_REDIS_SWITCH_ATTR_SYNC_OPERATION_RESPONSE_TIMEOUT;
+        attr.value.u64 = gSyncdResponseTimeout;
+        status = sai_switch_api->set_switch_attribute(gSwitchId, &attr);
+
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_WARN("Failed to set SAI REDIS response timeout");
+        }
+        else
+        {
+            SWSS_LOG_NOTICE("SAI REDIS response timeout set successfully to %" PRIu64 " ", attr.value.u64);
+        }
+    }
+
     status = sai_switch_api->create_switch(&gSwitchId, (uint32_t)attrs.size(), attrs.data());
     if (status != SAI_STATUS_SUCCESS)
     {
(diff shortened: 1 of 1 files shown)


orchagent/main.cpp
```diff
@@ -543,6 +544,29 @@ int main(int argc, char **argv)
     attr.value.u64 = gSwitchId;
     attrs.push_back(attr);
 
+
+    char *platform = getenv("platform");
+    if (platform && strstr(platform, BRCM_PLATFORM_SUBSTRING))
```
prsunny, line 549: can this be done for switchtype fabric?

vganesan-nokia: This change/the code path is applicable for both fabric and voq in chassis. In both "fabric" (supervisor card) and in "voq" (asic in linecard) the switch create with bcm SAI takes more than the default time.

vganesan-nokia:
@prsunny , would you please approve this PR and check if this can be merged?

kcudnik (dismissed):

vganesan-nokia:
@prsunny, would you please check if this PR can be merged? Thanks

vganesan-nokia:
@rlhui, @prsunny Would you please check if this PR can be merged?

orchagent/main.cpp
```diff
@@ -81,6 +81,8 @@ int32_t gVoqMaxCores = 0;
 uint32_t gCfgSystemPorts = 0;
 string gMyHostName = "";
 string gMyAsicName = "";
+uint64_t gSyncdRespTimeoutLC = ( 5 * SAI_REDIS_DEFAULT_SYNC_OPERATION_RESPONSE_TIMEOUT);
```
minionatwork, line 84: do you need this variable? Can #define useful. these variables aren't used other than switchcreate.

orchagent/main.cpp
```diff
@@ -81,6 +81,8 @@ int32_t gVoqMaxCores = 0;
...
+uint64_t gSyncdRespTimeoutFab = ( 10 * SAI_REDIS_DEFAULT_SYNC_OPERATION_RESPONSE_TIMEOUT);
```
judyjoseph, line 85: Do we really need a cumulative timeout value here?  I find induvidually each asic/syncd switch_create takes less time in SUP and in LC. I will get and add few logs from device later.

orchagent/main.cpp
```diff
@@ -574,6 +576,39 @@ int main(int argc, char **argv)
     attr.value.u64 = gSwitchId;
     attrs.push_back(attr);
 
+
+    char *platform = getenv("platform");
+    if (platform && strstr(platform, BRCM_PLATFORM_SUBSTRING) &&
```
judyjoseph, line 581: We can just have the check for voq, fabric alone ? 
Can remove the platform Broadcom check, I see Prince too suggested the same earlier.

orchagent/main.cpp

orchagent/main.cpp
vganesan-nokia: Fixed. Removed platfrom check.

prsunny (approved):

minionatwork (dismissed):

judyjoseph (approved):

judyjoseph:
@vganesan-nokia, Can you create a new PR for this change ? and close this PR  
azp runs are failing, I wanted to rule out if this is because this PR is very old.

vganesan-nokia:
> @vganesan-nokia, Can you create a new PR for this change ? and close this PR azp runs are failing, I wanted to rule out if this is because this PR is very old.

@judyjoseph, a new PR #2159 has been raised. The contents are identical.

prsunny (approved):

judyjoseph:
closing this as new PR is raised https://github.com/Azure/sonic-swss/pull/2159

PR #2003, opened 2021-11-03
[orchagent] Flush pipeline every 1 second, not only when select will timeout

kcudnik:
**What I did**
Added flush for async communication pipeline, every 1 second, not only if select will timeout.
**Why I did it**
pfcwdorch has a executor timeout which triggers select every second, and somehow from some time select never timeouts (its timeout is also 1 second) and every time is returning valid object, not causing async pipeline to flush, which causes for some operations to wait for next synchronous operation to be processed.
**How I verified it**
Tested thanks to @shi-su 
**Details if related**

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -1,5 +1,6 @@
 #include <unistd.h>
 #include <unordered_map>
+#include <chrono>
 #include <limits.h>
 #include "orchdaemon.h"
 #include "logger.h"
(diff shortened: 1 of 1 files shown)


kcudnik:
i actually don't like much this solution, so if you have any other one in mind, let me know

shi-su (approved):

kcudnik:
@prsunny hey, are you OK to merge this despite of failing tests ? each time is different set

PR #2008, opened 2021-11-04
[CRM][MPLS NHOP] Fix the mpls nexthop CRM attribute

smaheshm:
**What I did**
 Fix the mpls nexthop CRM attribute

MPLS v4 nexthops will share the same resource as IPv4 NHOP, there is no separate resource partition for MPLS nexthop.

**Why I did it**
Getting the following error in logs:
Oct 29 22:40:03.679316 str2-7804-lc3-1 ERR swss#orchagent: :- getResAvailableCounters: Failed to get availability for object_type 63 , rv:-5
.
.

**How I verified it**
Verified the message is not produced anymore

**Details if related**

### DIFF
--- a/orchagent/crmorch.cpp
+++ b/orchagent/crmorch.cpp
@@ -64,7 +64,7 @@ const map<CrmResourceType, uint32_t> crmResSaiAvailAttrMap =
     { CrmResourceType::CRM_SNAT_ENTRY, SAI_SWITCH_ATTR_AVAILABLE_SNAT_ENTRY },
     { CrmResourceType::CRM_DNAT_ENTRY, SAI_SWITCH_ATTR_AVAILABLE_DNAT_ENTRY },
     { CrmResourceType::CRM_MPLS_INSEG, SAI_OBJECT_TYPE_INSEG_ENTRY },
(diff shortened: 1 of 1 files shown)


orchagent/crmorch.cpp
```diff
@@ -64,7 +64,7 @@ const map<CrmResourceType, uint32_t> crmResSaiAvailAttrMap =
     { CrmResourceType::CRM_SNAT_ENTRY, SAI_SWITCH_ATTR_AVAILABLE_SNAT_ENTRY },
     { CrmResourceType::CRM_DNAT_ENTRY, SAI_SWITCH_ATTR_AVAILABLE_DNAT_ENTRY },
     { CrmResourceType::CRM_MPLS_INSEG, SAI_OBJECT_TYPE_INSEG_ENTRY },
```
prsunny, line 66: Is the INSEG_ENTRY correct?

smaheshm: yes, label routes usually have a separate tables than IP routes.

smaheshm: but in code it doesn't look right, will check further.

smaheshm: There is no SAI attribute for AVAILABLE INSEG entries. How are such cases handled. Should the code for inseg label CRM be removed?

kcudnik (approved):

PR #2009, opened 2021-11-05
Update default route status to state DB

prsunny:
**What I did**
Orchagent update of IPv4 and IPv6 default route status. If there is a valid default route, it shall be updated as `state:ok`. If there is no valid default route (with packet action drop), state db shall be updated as `state:na`

**Why I did it**
Default route state in state_db can be used as a signal for application to determine if there is a valid default route learned via BGP is available in orchagent

**How I verified it**

**Details if related**
```
redis-cli -n 6
127.0.0.1:6379[6]> KEYS *ROUTE*
1) "ROUTE_TABLE|::/0"
2) "ROUTE_TABLE|0.0.0.0/0"
127.0.0.1:6379[6]> HGETALL "ROUTE_TABLE|0.0.0.0/0"
1) "state"
2) "na"
```

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -326,6 +326,9 @@ MuxCable::MuxCable(string name, IpPrefix& srv_ip4, IpPrefix& srv_ip6, IpAddress 
     state_machine_handlers_.insert(handler_pair(MUX_STATE_STANDBY_ACTIVE, &MuxCable::stateActive));
     state_machine_handlers_.insert(handler_pair(MUX_STATE_INIT_STANDBY, &MuxCable::stateStandby));
     state_machine_handlers_.insert(handler_pair(MUX_STATE_ACTIVE_STANDBY, &MuxCable::stateStandby));
+
+    /* Set initial state to "standby" */
+    stateStandby();
 }
 
 bool MuxCable::stateInitActive()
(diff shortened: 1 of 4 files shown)


orchagent/muxorch.cpp
```diff
@@ -326,6 +326,9 @@ MuxCable::MuxCable(string name, IpPrefix& srv_ip4, IpPrefix& srv_ip6, IpAddress
     state_machine_handlers_.insert(handler_pair(MUX_STATE_STANDBY_ACTIVE, &MuxCable::stateActive));
     state_machine_handlers_.insert(handler_pair(MUX_STATE_INIT_STANDBY, &MuxCable::stateStandby));
     state_machine_handlers_.insert(handler_pair(MUX_STATE_ACTIVE_STANDBY, &MuxCable::stateStandby));
+
+    /* Set initial state to "standby" */
+    stateStandby();
```
shi-su, line 331: Just wondering if this is directly related to the subject of this PR?

orchagent/routeorch.h
```diff
@@ -206,6 +206,8 @@ class RouteOrch : public Orch, public Subject
     unsigned int getNhgCount() { return m_nextHopGroupCount; }
     unsigned int getMaxNhgCount() { return m_maxNextHopGroupCount; }
 
+    void updateDefRouteState(string ip, bool add=false);
```
shi-su, line 209: Is this necessary to be a public function?

orchagent/muxorch.cpp
prsunny: Sure. move to another PR as I've few additional changes

orchagent/routeorch.h
prsunny: Moved to private

shi-su (approved):

shi-su (approved):

shi-su (approved):

qiluo-msft:

PR #2018, opened 2021-11-09
Return task_need_retry when insufficient resources to create

abanu-ms:
* Add SAI_STATUS_INSUFFICIENT_RESOURCES case in handleSaiCreateStatus
* Remove redundant return

**What I did**
I added a custom switch case in handleSaiCreateStatus() for insufficient resources case to keep retrying the operation for when enough resources will be available.
Removed redundant return from the method as it would never be reached.

**Why I did it**
Orchagent would crash if the ASIC runs out of resources when it should keep retrying.

**How I verified it**
Live tested the fix on real hardware.

**Details if related**

### DIFF
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -728,13 +728,16 @@ task_process_status Orch::handleSaiCreateStatus(sai_api_t api, sai_status_t stat
                 case SAI_STATUS_SUCCESS:
                     SWSS_LOG_WARN("SAI_STATUS_SUCCESS is not expected in handleSaiCreateStatus");
                     return task_success;
+                case SAI_STATUS_INSUFFICIENT_RESOURCES:
+                    SWSS_LOG_INFO("SAI API: %s returned status: %s",
+                                  sai_serialize_api(api).c_str(), sai_serialize_status(status).c_str());
+                    return task_need_retry;
                 default:
                     SWSS_LOG_ERROR("Encountered failure in create operation, exiting orchagent, SAI API: %s, status: %s",
                                 sai_serialize_api(api).c_str(), sai_serialize_status(status).c_str());
                     exit(EXIT_FAILURE);
             }
     }
-    return task_need_retry;
 }
 
 task_process_status Orch::handleSaiSetStatus(sai_api_t api, sai_status_t status, void *context)


orchagent/orch.cpp
```diff
@@ -728,13 +728,16 @@ task_process_status Orch::handleSaiCreateStatus(sai_api_t api, sai_status_t stat
                 case SAI_STATUS_SUCCESS:
                     SWSS_LOG_WARN("SAI_STATUS_SUCCESS is not expected in handleSaiCreateStatus");
                     return task_success;
+                case SAI_STATUS_INSUFFICIENT_RESOURCES:
+                    SWSS_LOG_INFO("SAI API: %s returned status: %s",
+                                  sai_serialize_api(api).c_str(), sai_serialize_status(status).c_str());
+                    return task_need_retry;
```
shi-su, line 734: I am not fully sure if this is the best solution since the same call might get retried forever, and there might be no one to be aware of the problem. The info level log message might just hide the problem. I was wondering if there is any particular scenario that needs this behavior?

abanu-ms: For instance you might run out of NH / neighbor resources in ASIC - not due to a bug, but because of the scale of the network. While indeed this does not solve the problem, at least it doesn't stop the switch and lets it do its best while an operator should take care of the underlying issue.

I've put that log at INFO level because, as you said, it might get retried forever and not to add too much spam to the syslog. Usually I've seen error logs being printed by syncd and sairedis when something goes wrong in SAI, so it shouldn't pass unnoticed.

Also, this behavior already exists in at least one orch I'm aware of - the RouteOrch (or the newer NhgOrch). If there are not enough resources for a new NHG, its creation is postponed and will be retried until it succeeds. It doesn't make sense for orchagent to stop when reaching a resource limitation.

abanu-ms:
@shi-su Could you please review my comment about this use case and let me know if you agree - and if not, what other solution would you see for stopping orchagent from exiting in such scenarios? Thanks!

orchagent/orch.cpp
shi-su: I am still concerned if we should use this change to address this problem since it changes the current behavior for all SAI APIs for the `SAI_STATUS_INSUFFICIENT_RESOURCES`. I doubt if it may cause other problems. I was wondering if there is any urgent issue directly related to this issue (e.g., known production issues)? If not, I would prefer to leave it as is at the moment. In the future, we can introduce the ERROR_DB and notify to upstream process to act accordingly to address the issue.

abanu-ms: I do not think it's that urgent. It's something that can occur and it occurred in my scale testing (but it can also be a bug in the SAI implementation I had), but I'm not sure if anyone has problems with this right now. @TACappleman you okay with closing this PR as @shi-su suggests and wait for that ERROR_DB?

TACappleman: Fine with me

abanu-ms: Okay. I will close this PR for now then.

abanu-ms:
Will be waiting for ERROR_DB to be added and a better solution to be implemented.

PR #2023, opened 2021-11-11
[orchagent]: Support for setting switch level DSCP to TC QoS map

vmittal-msft:
**What I did**
Added support for setting switch level DSCP to TC QoS map.

**Why I did it**
This is required to set correct DSCP -> TC QoS map for tunnels else they will point to some default QoS map (DSCP/8 = TC) which is not correct.

**How I verified it**
This needs SAI change to support this capability. Verified this to be working fine on T0 testbed with mock dual ToR topology.
 
**Details if related**


orchagent/qosorch.cpp
```diff
@@ -207,12 +210,41 @@ bool DscpToTcMapHandler::convertFieldValuesToAttributes(KeyOpFieldsValuesTuple &
...
     sai_object_id_t sai_object;
     vector<sai_attribute_t> qos_map_attrs;
+    bool rv;
+
+    /* Query DSCP_TO_TC QoS map at switch capability */
+    rv = gSwitchOrch->querySwitchDscpToTcCapability(SAI_OBJECT_TYPE_SWITCH, SAI_SWITCH_ATTR_QOS_DSCP_TO_TC_MAP);
+    if (rv == false)
+    {
+        SWSS_LOG_ERROR("Switch level DSCP to TC QoS map configuration is not supported");
+        return false;
```
prsunny, line 246: why is it returned from here? IMO, we should check query capability and apply the switch setting if capability is true. Another suggestion is to make it part of  `applyDscpToTcMapToSwitch`

orchagent/switchorch.cpp
```diff
@@ -613,3 +613,22 @@ void SwitchOrch::querySwitchTpidCapability()
...
+    sai_status_t status = SAI_STATUS_SUCCESS;
+    sai_attr_capability_t capability;
+
+    status = sai_query_attribute_capability(gSwitchId, sai_object, attr_id, &capability);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_WARN("Could not query switch level DSCP to TC map %d", status);
+        return false;
+    }
+
```
prsunny, line 632: You have to check `capability.set_implemented` is true as well, before returning from this function

prsunny (commented):
we may need some VS test to cover this

orchagent/qosorch.cpp
vmittal-msft: If it is true the following code is applying DSCP to TC QoS map at switch.

orchagent/switchorch.cpp
vmittal-msft: I didn't check for capability.set_implemented to keep changes in SAI code to minimum. Please let me know if this is still needed.

orchagent/qosorch.cpp
```diff
@@ -12,6 +12,7 @@
 
 using namespace std;
 
+extern sai_switch_api_t *sai_switch_api;
```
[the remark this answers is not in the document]
vmittal-msft: Done

orchagent/qosorch.cpp
```diff
@@ -20,7 +21,9 @@ extern sai_qos_map_api_t *sai_qos_map_api;
 extern sai_scheduler_group_api_t *sai_scheduler_group_api;
 extern sai_switch_api_t *sai_switch_api;
 extern sai_acl_api_t* sai_acl_api;
+extern bool querySwitchDscpToTcCapability(sai_object_type_t, sai_attr_id_t);
```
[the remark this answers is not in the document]
vmittal-msft: This function is defined in another file hence added extern.

orchagent/qosorch.cpp
```diff
@@ -12,15 +12,17 @@
...
+extern sai_switch_api_t *sai_switch_api;
 extern sai_port_api_t *sai_port_api;
 extern sai_queue_api_t *sai_queue_api;
 extern sai_scheduler_api_t *sai_scheduler_api;
 extern sai_wred_api_t *sai_wred_api;
 extern sai_qos_map_api_t *sai_qos_map_api;
 extern sai_scheduler_group_api_t *sai_scheduler_group_api;
-extern sai_switch_api_t *sai_switch_api;
 extern sai_acl_api_t* sai_acl_api;
+bool querySwitchDscpToTcCapability(sai_object_type_t, sai_attr_id_t);
```
prsunny, line 23: Please remove this. It is not required

orchagent/qosorch.cpp
```diff
@@ -207,12 +209,41 @@ bool DscpToTcMapHandler::convertFieldValuesToAttributes(KeyOpFieldsValuesTuple &
...
+    SWSS_LOG_NOTICE("Applied DSCP_TO_TC QoS map to switch successfully");
+}
+
 sai_object_id_t DscpToTcMapHandler::addQosItem(const vector<sai_attribute_t> &attributes)
 {
     SWSS_LOG_ENTER();
     sai_status_t sai_status;
     sai_object_id_t sai_object;
     vector<sai_attribute_t> qos_map_attrs;
+    bool rv;
```
prsunny, line 246: No longer required

orchagent/qosorch.cpp
```diff
@@ -231,6 +262,10 @@ sai_object_id_t DscpToTcMapHandler::addQosItem(const vector<sai_attribute_t> &at
         return SAI_NULL_OBJECT_ID;
     }
     SWSS_LOG_DEBUG("created QosMap object:%" PRIx64, sai_object);
+
+
```
prsunny, line 266: Please remove extra line

prsunny (approved):
lgtm, tests are planned to be added as future commit

qiluo-msft:
This commit could not be cleanly cherry-pick to 202012. Please submit another PR.

PR #2025, opened 2021-11-12
Fix next hop compilation

TACappleman:
**What I did**
Fixed compilation of SWSS following last 2 commits

**Why I did it**
Last 2 commits don't compile together due to change in nexthop(group)key definitions

**How I verified it**
Compiled with this fix

**Details if related**

### DIFF
--- a/orchagent/nexthopgroupkey.h
+++ b/orchagent/nexthopgroupkey.h
@@ -21,7 +21,7 @@ public:
     }
 
     /* ip_string|if_alias|vni|router_mac separated by ',' */
-    NextHopGroupKey(const std::string &nexthops, bool overlay_nh, bool srv6_nh)
+    NextHopGroupKey(const std::string &nexthops, bool overlay_nh, bool srv6_nh = false)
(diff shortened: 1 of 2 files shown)


prsunny:
why is lgtm failing? @kperumalbfn , please review

TACappleman:
LGTM fails because it tries to compile the previous commit for comparison

prsunny (approved):

kperumalbfn:
@TACappleman Not sure why it is failing after merge.

kperumalbfn (approved):

TACappleman:
@kperumalbfn the previous commit (which merged around the same time) referenced one of the functions you'd added a variable to. There was no merge conflict between them, but the combo then couldn't compile

kperumalbfn:
> @kperumalbfn the previous commit (which merged around the same time) referenced one of the functions you'd added a variable to. There was no merge conflict between them, but the combo then couldn't compile

Thanks @TACappleman

PR #2027, opened 2021-11-12
[muxorch] Bind all ports to drop ACL table

devpatha:
**What I did**
BInd all physical ports to drop ACL table

**Why I did it**
With current implementation, ACL table is link/bind to the very first port that triggers creation of drop ACL table followed by rule creation for the same.
For subsequent ports only new rule is created.
If a port is not linked/bind to the drop ACL table, the rule is not applied for that port.

**How I verified it**
1. Verified using sairedis log and mux logs that was added with new code that all ports are linked/bind to the drop ACL table.
2. Verified that after this fix all standby ports are dropping packets as expected.

**Details if related**

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -790,9 +790,9 @@ void MuxAclHandler::createMuxAclTable(sai_object_id_t port, string strTable)
 
     acl_table.type = ACL_TABLE_DROP;
     acl_table.id = strTable;
-    acl_table.link(port);
     acl_table.stage = ACL_STAGE_INGRESS;
     gAclOrch->addAclTable(acl_table);
+    bindAllPorts(acl_table);
 }
 
 void MuxAclHandler::createMuxAclRule(shared_ptr<AclRuleMux> rule, string strTable)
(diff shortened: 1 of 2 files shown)


orchagent/muxorch.cpp
```diff
@@ -790,9 +790,9 @@ void MuxAclHandler::createMuxAclTable(sai_object_id_t port, string strTable)
 
     acl_table.type = ACL_TABLE_DROP;
     acl_table.id = strTable;
-    acl_table.link(port);
     acl_table.stage = ACL_STAGE_INGRESS;
     gAclOrch->addAclTable(acl_table);
+    bindAllPorts(acl_table);
```
prsunny, line 795: As you can see from line 783, the same table shall be used if pfcwd has already created this table. You may want to apply the same fix [here ](https://github.com/Azure/sonic-swss/blob/master/orchagent/pfcactionhandler.cpp#L410) as well -

devpatha: sure. I can update the PR with the changes but won't be able to verify as Cisco platform does not use drop ACL table for PFC wd. Is it fine to change?

devpatha: Sure. I can update the PR with requested changes. But cisco platform does not use drop ACL for PFC wd so I won't be able to verify it. Is it fine to change?

prsunny: ok, then leave it for now and revisit later

prsunny (approved):

PR #2030, opened 2021-11-14
[vstest/acl table] Fix ports field in acl table creation

wendani:
Signed-off-by: Wenda Ni <wonda.ni@gmail.com>

**What I did**

**Why I did it**
Be consistent with `dvs_acl.create_acl_table`

```
        table_attrs = { 
            "policy_desc": table_name,
            "type": table_type,
            "ports@": ",".join(ports)
        }   
```

**How I verified it**
vs test

**Details if related**


prsunny:
@wendani , does that mean old config should be failing?

PR #2031, opened 2021-11-15
[flex counter] Flex counter threads consume too much CPU resources #9202

StormLiangMS:
**What I did**
Watermark flex counter will consume many CPU resources on some platforms.
By increasing the default interval for those counters from 10 seconds to 60 seconds to
workaround this issue. The performance issue of flex counter  need to be addressed
separately, this is just to reduce CPU consumption by adjusting the granularity of watermark related counters .
If smaller granularity is needed, the interval could be adjusted through CLI under SONIC shell as below.
counterpoll watermark interval <interval_val_in_ms>

**Why I did it**

**How I verified it**

**Details if related**


orchagent/portsorch.cpp
```diff
@@ -63,8 +63,8 @@ extern string gMyAsicName;
 #define PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS     1000
 #define PORT_BUFFER_DROP_STAT_POLLING_INTERVAL_MS     60000
 #define QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS   10000
-#define QUEUE_WATERMARK_FLEX_STAT_COUNTER_POLL_MSECS "10000"
-#define PG_WATERMARK_FLEX_STAT_COUNTER_POLL_MSECS    "10000"
+#define QUEUE_WATERMARK_FLEX_STAT_COUNTER_POLL_MSECS "60000"
+#define PG_WATERMARK_FLEX_STAT_COUNTER_POLL_MSECS    "60000"
```
neethajohn, line 67: There is also a buffer pool watermark. The default needs to be changed for that as well

StormLiangMS: There are only 3 counters for the buffer pool watermark related, I didn't change the default value since it won't make difference on the CPU consumption. 
But I think you are right, since it would employ same configuration of watermark interval, it should be same with other watermark intervals by default.

prsunny (approved):
is this required to be backported?

neethajohn (approved):

StormLiangMS:
> is this required to be backported?

Yes, this one should be back ported to 201911 and 202012.

prsunny:
@StormLiangMS , for future PRs,  please create a fork of this repo and raise PR from there

PR #2034, opened 2021-11-18
[CoPP] Add always_enabled field to coppmgr logic

noaOrMlnx:
**What I did**
Add the new design for copp manager - always_enabled field in copp_cfg.json file. 

**Why I did it**
A change was done for traps needs to be installed but doesn't have feature (arp, udld, lacp, ip2me),
in the new implementation, coppmgr will not automatically install traps which has no entry in features table, but will check first if the trap has "always_enabled":"true" field.

**How I verified it**
run tests for making sure traps are installed and not when expecting them to. 

**Details if related**
related to sonic-buildimage change - https://github.com/Azure/sonic-buildimage/pull/9302

SONiC mgmt test plan can be found in https://github.com/Azure/SONiC/pull/903.


dgsudharsan (dismissed):

prsunny (approved):

prsunny:
@noaOrMlnx , please check the test failure

prsunny (approved):

dgsudharsan (dismissed):

liat-grozovik:
@noaOrMlnx can you please review the failures of Azure.sonic-swss ?

cfgmgr/coppmgr.cpp
```diff
@@ -610,21 +610,18 @@ void CoppMgr::doCoppTrapTask(Consumer &consumer)
...
-                std::vector<FieldValueTuple> feature_fvs;
+                m_coppTrapConfMap[key].is_always_enabled = is_always_enabled;
                 if (is_always_enabled == "true")
                 {
                     /* if the value was changed from false to true,
                     if the trap is not installed, install it.
                     otherwise, do nothing. */
 
-                    if (m_coppTrapConfMap.find(key) == m_coppTrapConfMap.end())
+                    // if the feature was not enabled, install the trap
```
dgsudharsan, line 620: I think feature enabled should be checked here before the cache is updated since feature enabled looks at the cache and the point of the logic is to get the feature state before the always enabled update. Can you please revisit?

cfgmgr/coppmgr.cpp
```diff
@@ -687,7 +683,7 @@ void CoppMgr::doCoppTrapTask(Consumer &consumer)
         }
         else if (op == DEL_COMMAND)
         {
-            if (conf_present && !m_coppTrapConfMap[key].trap_group.empty() && !m_coppTrapConfMap[key].trap_ids.empty())
+            if (conf_present)
```
dgsudharsan, line 686: Can you please explain why this check is removed?

noaOrMlnx: it is here https://github.com/Azure/sonic-swss/pull/2034/files#diff-39e9e71a7788fdd27fda2f3bd818b47ce232b447bb3d1064b1f1433e51657459R704

dgsudharsan (approved):

liat-grozovik (approved):

qiluo-msft:

prsunny:
@noaOrMlnx , can you please raise a new PR for 202012?

Issue #1971, opened 2021-10-20
Enable support to drop Rx traffic on WD detection for cisco-8000


alpeshspatel:
https://github.com/Azure/sonic-swss/pull/1748 enabled support for PFC-WD detection and restoration for cisco-8000 using the drop action handler. 

In the above referenced PR, support was added for dropping the Tx traffic from the port/tc where PFC-WD is detected.

This issue is created to enable drop the Rx traffic on the port/tc where PFC-WD is detected.

alpeshspatel:
Will open a PR against sonic-swss/202012 branch once internal testing is complete

PR #1942, opened 2021-10-05
fix the type for SAI_BUFFER_PROFILE_ATTR_BUFFER_SIZE

alpeshspatel:
Fixing the type for SAI_BUFFER_PROFILE_ATTR_BUFFER_SIZE

It is incorrectly set to u32 and is not clearing the higher 32 bits in the union

Sample code:

    attr.id = SAI_BUFFER_PROFILE_ATTR_POOL_ID;
    attr.value.oid = getPool(ingress); <<<<-- the higher order 32 bits were retrieved for  BUFFER_PROFILE_ATTR_BUFFER_SIZE
    attribs.push_back(attr);

    attr.id = SAI_BUFFER_PROFILE_ATTR_THRESHOLD_MODE;
    attr.value.u32 = SAI_BUFFER_PROFILE_THRESHOLD_MODE_DYNAMIC;
    attribs.push_back(attr);

    attr.id = SAI_BUFFER_PROFILE_ATTR_BUFFER_SIZE;
    attr.value.u64 = 0;      <<<<-------------------------------- This was u32 earlier
    attribs.push_back(attr);

Signed-off-by: Alpesh S Patel <alpesh@cisco.com>

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

PR #2041, opened 2021-11-21
patch for issue #1971 - enable Rx Drop handling for cisco-8000

alpeshspatel:
Signed-off-by: Alpesh S Patel <alpesh@cisco.com>

**What I did**

1. Enables support for Rx traffic drop for a port/tc by applying the zero buffer profile
2. Changed class hierarchy so that default getHwCounters() function is used (since the Rx counter support is enabled)
3. Fixes a pfc-wd off by 1 detection in case of PFC-WD detection without traffic

**Why I did it**
1. enabling a missing functionality
2. leveraging sonic code as our sdk now implements the missing counter
3. fixes a bug 

**How I verified it**
on a cisco-8000 router:
1. detected pfc-wd detection and restore happens within the (detection/restore-time + 1 poll interval) duration and that the changeset passes MSFT tests
2. Verified that when PFC-WD is detected, both Rx and Tx traffic for a port/tc is dropped and no forwarding happens

**Details if related**

this patch will need to be double committed to the 202012 branch along with #1942 , #1748 and #1962

### DIFF
--- a/orchagent/pfc_restore_cisco-8000.lua
+++ b/orchagent/pfc_restore_cisco-8000.lua
@@ -44,7 +44,7 @@ for i = n, 1, -1 do
         and (debug_storm ~= "enabled")
         -- DEBUG CODE END.
         then
-            if time_left <= 0 then
+            if time_left <= poll_time then
                 redis.call('PUBLISH', 'PFC_WD_ACTION', '["' .. KEYS[i] .. '","restore"]')
                 time_left = restoration_time
             else
--- a/orchagent/pfcactionhandler.cpp
+++ b/orchagent/pfcactionhandler.cpp
@@ -221,7 +221,7 @@ void PfcWdActionHandler::updateWdCounters(const string& queueIdStr, const PfcWdQ
 
 PfcWdSaiDlrInitHandler::PfcWdSaiDlrInitHandler(sai_object_id_t port, sai_object_id_t queue,
                                                uint8_t queueId, shared_ptr<Table> countersTable):
-    PfcWdActionHandler(port, queue, queueId, countersTable)
+    PfcWdZeroBufferHandler(port, queue, queueId, countersTable)
 {
     SWSS_LOG_ENTER();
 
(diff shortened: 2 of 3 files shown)


alpeshspatel:
@prsunny Can you go ahead and put me as Assignee (I cannot edit it)

orchagent/pfc_restore_cisco-8000.lua
```diff
@@ -44,7 +44,7 @@ for i = n, 1, -1 do
         and (debug_storm ~= "enabled")
         -- DEBUG CODE END.
         then
-            if time_left <= 0 then
+            if time_left <= poll_time then
```
neethajohn, line 47: We had this discussion during #1748 review. https://github.com/Azure/sonic-swss/pull/1748#discussion_r695934838
What has changed now?

neethajohn: @alpeshspatel , I am not clear why this is changed to compare with poll_time. in 1748, the change was made to compare with 0 and you had provided a small writeup for the reason.

orchagent/pfcactionhandler.cpp
```diff
@@ -469,6 +436,15 @@ PfcWdLossyHandler::PfcWdLossyHandler(sai_object_id_t port, sai_object_id_t queue
 {
     SWSS_LOG_ENTER();
 
+    string platform(getenv("platform"));
```
neethajohn, line 439: assign to empty string if "platform" does not exist

orchagent/pfcactionhandler.cpp
```diff
@@ -488,6 +464,15 @@ PfcWdLossyHandler::~PfcWdLossyHandler(void)
...
```
neethajohn, line 467: assign to empty string if "platform" does not exist

neethajohn (changes requested):

orchagent/pfc_restore_cisco-8000.lua
neethajohn: resolved offline

alpeshspatel:
@neethajohn : just pushed a change and marked the two comments as resolved for the time being. Please re-review and advise on next steps.

orchagent/pfcactionhandler.cpp
```diff
@@ -436,13 +436,20 @@ PfcWdLossyHandler::PfcWdLossyHandler(sai_object_id_t port, sai_object_id_t queue
 {
     SWSS_LOG_ENTER();
 
-    string platform(getenv("platform"));
-
-    if (platform == CISCO_8000_PLATFORM_SUBSTRING)
+    const char *env_plat = getenv("platform");
+    if (env_plat == nullptr)
```
neethajohn, line 440: string platform = getenv("platform") ? getenv("platform") : ""  would be much cleaner

sachinv-msft:
@neethajohn could you please review this PR?

neethajohn (approved):

alpeshspatel:
@prsunny 
What would be the next steps on this?

PR #2022, opened 2021-11-10
Dynamic port configuration - add port buffer cfg to the port ref counter

tomer-israel:
**What I did**
I added increasing/decreasing to the port ref counter each time a port buffer configuration is added or removed

Implemented according to the - [Dynamic Port Cfg HLD](https://github.com/Azure/SONiC/pull/900)

**Why I did it**
In order to avoid cases where a port is removed before the buffer configuration on this port were removed also

**How I verified it**
I added a vs test that test it.
we remove a port with buffer configuration and the port is not removed. only after all buffer configurations on this port were removed - this port will be removed.

**Details if related**

PR #2047, opened 2021-11-24
Add VS test cases for dynamic port cfg

tomer-israel:
Signed-off-by: tomeri <tomeri@nvidia.com>

DEPENDS ON: [Dynamic port configuration - add port buffer cfg to the port ref counter #2022](https://github.com/Azure/sonic-swss/pull/2022)

**What I did**
added a vs test to check functionality of add/remove ports dynamically on run time

**Why I did it**
in order to test add/remove ports functionality

**How I verified it**
test passed

**Details if related**

### DIFF
--- a/tests/test_port_add_remove.py
+++ b/tests/test_port_add_remove.py
@@ -1,0 +1,136 @@
+import pytest
+import time
+from dvslib.dvs_common import PollingConfig
+
+# the port to be removed and add
+PORT = "Ethernet64"
+
+"""
+DELETE_CREATE_ITERATIONS defines the number of iteration of delete and create to ports,
+we add different timeouts between delete/create to catch potential race condition that can lead to system crush.
+"""
+DELETE_CREATE_ITERATIONS = 10
+
+@pytest.mark.usefixtures('dvs_port_manager')
+class TestPortAddRemove(object):
+
+    def test_remove_port_with_buffer_cfg(self, dvs, testlog):
+        config_db = dvs.get_config_db()
+        asic_db = dvs.get_asic_db()
+
+        # get port info
+        port_info = config_db.get_entry("PORT", PORT)
+
+        # get the number of ports before removal
+        num_of_ports = len(asic_db.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_PORT"))
+
+        # try to remove this port
+        config_db.delete_entry('PORT', PORT)
+        num = asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_PORT",
+                                      num_of_ports-1,
+                                      polling_config = PollingConfig(polling_interval = 1, timeout = 5.00, strict = False))
+
+        # verify that the port wasn't removed since we still have buffer cfg
+        assert len(num) == num_of_ports
+
+        # remove buffer pg cfg for the port
+        pgs = config_db.get_keys('BUFFER_PG')
+        for key in pgs:
+            if PORT in key:
+                config_db.delete_entry('BUFFER_PG', key)
+
+        num = asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_PORT",
+                              num_of_ports-1,
+                              polling_config = PollingConfig(polling_interval = 1, timeout = 5.00, strict = False))
+
+        # verify that the port wasn't removed since we still have buffer cfg
+        assert len(num) == num_of_ports
+
+        # modify buffer queue entry to egress_lossless_profile instead of egress_lossy_profile
+        config_db.update_entry("BUFFER_QUEUE", "%s|0-2"%PORT, {"profile": "egress_lossless_profile"})
+
+        # remove buffer queue cfg for the port
+        pgs = config_db.get_keys('BUFFER_QUEUE')
+        for key in pgs:
+            if PORT in key:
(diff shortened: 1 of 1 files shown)


zhenggen-xu:
Are you following the vs tests section in this doc? https://github.com/Azure/SONiC/pull/900/files , if so, it seems we should add more test cases.

tomer-israel:
> Are you following the vs tests section in this doc? https://github.com/Azure/SONiC/pull/900/files , if so, it seems we should add more test cases.

I will update this section in the HLD.
Not all the cases I mentioned on the hld are implemented, some cases were problematic to add such as reboot test or error cases (hard to detect error messages in the log).

dprital:
@zhenggen-xu  - The comments on this PR were addressed. Can you please approve this PR ?

tests/test_port_add_remove.py
```diff
@@ -0,0 +1,136 @@
...
+                # remove a port
+                self.dvs_port.remove_port(key)
+    
+            # verify remove port
+            num = asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_PORT",
+                                          num_of_ports-len(ports))
+            assert len(num) == num_of_ports-len(ports)
+            
+            # add port
+            time.sleep(i%3)
```
zhenggen-xu, line 100: what is this waiting for?

tests/test_port_add_remove.py
```diff
@@ -0,0 +1,136 @@
...
+            for key in ports:
+                config_db.create_entry("PORT", key, ports_info[key])
+
+            # verify add port            
+            num = asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_PORT",
+                                  num_of_ports)
+            assert len(num) == num_of_ports
+                
+            time.sleep((i%2)+1)
```
zhenggen-xu, line 109: what is the waiting here, and why it varies?

tests/test_port_add_remove.py
```diff
@@ -0,0 +1,136 @@
...
+        
+        time.sleep(2)
+        
+        rc = dvs.servers[16].runcmd("ping -c 1 6.6.6.7")
+        assert rc == 0
+
+        rc = dvs.servers[17].runcmd("ping -c 1 6.6.6.6")
+        assert rc == 0
+
+        dvs.remove_vlan_member("6", "Ethernet68")
```
zhenggen-xu, line 134: Change the PORT status (like admin etc) to whatever before the test case?

tests/test_port_add_remove.py
tomer-israel: I added different timeouts between delete/create to catch potential race conditions that can lead to system crush.

tests/test_port_add_remove.py
tomer-israel: same as above:
different timeouts between delete/create can catch potential race conditions that can lead to system crush.

tests/test_port_add_remove.py
tomer-israel: changing the ports bach back to admin down was added

dprital:
@zhenggen-xu  - Can you please review and approve ?

liat-grozovik:
@zhenggen-xu could you please help to review?

tests/test_port_add_remove.py
zhenggen-xu: Can you add this as the comments in the code ?

tests/test_port_add_remove.py
```diff
@@ -0,0 +1,138 @@
...
+                                  num_of_ports)
+            assert len(num) == num_of_ports
+                
+            time.sleep((i%2)+1)           
+            
+        # run ping
+        dvs.setup_db()        
+        dvs.create_vlan("6")
+        dvs.create_vlan_member("6", PORT)
+        dvs.create_vlan_member("6", "Ethernet68")
```
zhenggen-xu, line 115: Can we make `Ethernet68` as defined constant too at beginning?

tests/test_port_add_remove.py
zhenggen-xu: Actually, we don't know if the port was admin down or not, I would think we should save the initial admin status, and then change to whatever it was before the test case.

tests/test_port_add_remove.py
tomer-israel: it is already inside on line 9-10
but maybe it's better to put it next to the time.sleep

tests/test_port_add_remove.py
tomer-israel: I noticed that on default it is down
I agree that the best thing to do is to save the initial admin status and change it back to the initial admin status in the end

tests/test_port_add_remove.py
tomer-israel: done

tests/test_port_add_remove.py
tomer-israel: i added those comments before the sleep

liat-grozovik:
@zhenggen-xu kindly reminder, please review recent changes following your feedback.

zhenggen-xu:
> @zhenggen-xu kindly reminder, please review recent changes following your feedback.

This is not addressed:
https://github.com/Azure/sonic-swss/pull/2047#discussion_r772895844

PR #2049, opened 2021-11-25
[cbf] Fix max FC value

abanu-ms:
**What I did**
Fixed the maximum FC value allowed by a switch.

**Why I did it**
There was a bit of confusion when I first wrote this, thinking the attribute should return the maximum FC value allowed by the switch, not the maximum number of FC classes. Because of this, the actual maximum FC value allowed is one less than the current one because we're counting from 0. As such, if the switch reports the maximum number of FCs is 255, the actual FC value must be in the 0-254 range.

**How I verified it**
Updated the existing UTs

**Details if related**


prsunny:
@kperumalbfn to review

kperumalbfn (approved):

abanu-ms:
@prsunny can we get it merged?

prsunny:
@arlakshm to signoff/merge

arlakshm (approved):

PR #2056, opened 2021-12-02
[cbf] Fix cbf sync error

abanu-ms:
**What I did**
Changed the CBF group members sync code to be non-assertive when one of its members is not yet synced.

**Why I did it**
There can be situations when the member of a CBF group is not yet synced by the time the CBF group is being synced. Some examples would be:

- CBF group entry is created before its member in APPL_DB
- Member sync is being blocked by something, like missing a neighbor/NH entry

In this scenarios, we should wait for the member to be synced instead of asserting as the member might become available shortly.

**How I verified it**
Added a new UT

**Details if related**


TACappleman:
@arlakshm could you review this PR please?

TACappleman:
@arlakshm could we get a review of this PR please?

arlakshm (approved):

PR #2058, opened 2021-12-02
[vnetorch] Advertise vnet tunnel routes

shi-su:
**What I did**
Advertise active vnet tunnel routes.

**Why I did it**
The overlay routes programmed on the device need to be advertised to BGP peers. This PR aims to meet this requirement.

**How I verified it**
Add an entry in `ADVERTISE_NETWORK_TABLE` for active overlay routes

**Details if related**

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -396,7 +396,7 @@ bool VNetOrch::addOperation(const Request& request)
     sai_attribute_t attr;
     vector<sai_attribute_t> attrs;
     set<string> peer_list = {};
-    bool peer = false, create = false;
+    bool peer = false, create = false, advertise_prefix = false;
     uint32_t vni=0;
     string tunnel;
     string scope;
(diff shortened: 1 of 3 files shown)


orchagent/vnetorch.cpp
```diff
@@ -1563,12 +1564,36 @@ void VNetRouteOrch::postRouteState(const string& vnet, IpPrefix& ipPrefix, NextH
     fvVector.emplace_back("state", route_state);
 
     state_vnet_rt_tunnel_table_->set(state_db_key, fvVector);
+
+    if (route_state == "active")
```
prsunny, line 1568: Advertisement must be done only if the VNET attribute is set to [advertise ](https://github.com/Azure/SONiC/blob/a2c68f2b6c40d4b5da164713a988146612ce5cf4/doc/vxlan/Overlay%20ECMP%20with%20BFD.md#212-vnetinterface-table). It should not break the previous behavior of default as no-advertise.

shi-su: Added support for `advertise_prefix` attribute and updated vs tests.

liat-grozovik:
@shi-su  should this PR go to 202111 as well? if so can you please add a label?

prsunny (approved):

PR #2062, opened 2021-12-03
Fix for 2053, Fix IPv6 BGP multipath-relax peer-type.

skbarista:
**What I did**
By default the metric for ipv6 static route is 1024. This makes it the
routes learned via the internal bgp sessions inferior to the Ebgp
routes. So make the metric 256 which make the static route metric same
as connected route metric.

**Why I did it**
To fix IPv6 BGP multipath-relax peer-type.

**How I verified it**

Verified voq chassis can learn a route which ecmps routes advertised by internal ibgp neighbor and ebgp neighbor.

### DIFF
--- a/cfgmgr/nbrmgr.cpp
+++ b/cfgmgr/nbrmgr.cpp
@@ -509,7 +509,12 @@ bool NbrMgr::addKernelRoute(string odev, IpAddress ip_addr)
     }
     else
     {
-        cmd = string("") + IP_CMD + " -6 route add " + ip_str + "/128 dev " + odev;
+        // In voq system, We need the static route to the remote neighbor and connected
+        // route to have the same metric to enable BGP to choose paths from routes learned
+        // via eBGP and iBGP over the internal inband port be part of same ecmp group.
+        // For v4 both the metrics (connected and static) are default 0 so we do not need
+        // to set the metric explicitly.
+        cmd = string("") + IP_CMD + " -6 route add " + ip_str + "/128 dev " + odev + " metric 256";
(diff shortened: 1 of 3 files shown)


arlakshm (approved):
lgtm

arlakshm:
@prsunny, can you take a look at this PR ?

cfgmgr/nbrmgr.cpp
```diff
@@ -509,7 +509,12 @@ bool NbrMgr::addKernelRoute(string odev, IpAddress ip_addr)
     }
     else
     {
-        cmd = string("") + IP_CMD + " -6 route add " + ip_str + "/128 dev " + odev;
+        // In voq system, We need the static route to the remote neighbor and connected
+        // route to have the same metric to enable BGP to choose paths from routes learned
+        // via eBGP and iBGP over the internal inband port be part of same ecmp group.
+        // For v4 both the metrics (connected and static) are default 0 so we do not need
+        // to set the metric explicitly.
+        cmd = string("") + IP_CMD + " -6 route add " + ip_str + "/128 dev " + odev + " metric 256";
```
prsunny, line 517: SYSTEM_NEIGH_TABLE is only updated for voq systems, correct?

prsunny (approved):

cfgmgr/nbrmgr.cpp
skbarista: Yes that is correct.

arlakshm:
@skbarista, the change cannot be cherry-picked clean to 202106 branch. Please create another PR for 202106

PR #2063, opened 2021-12-06
[Reclaiming buffer][202012] Support reclaiming buffer in traditional buffer model

stephenxs:
**What I did**
It's to port #2011 to 202012
To reclaim reserved buffer.
As the way to do it differs among vendors, `buffermgrd` will:

- Handle port admin down on Mellanox platform.
  - Not apply lossless buffer PG to an admin-down port 
  - Remove lossless buffer PG (3-4) from a port when it is shut down.
- Readd lossless buffer PG (3-4) to a port when a port is started up.

**Why I did it**
To support reclaiming reserved buffer when a port is shut down on Mellanox platform in traditional buffer model.

**How I verified it**

Regression test and vs test.

**Details if related**


liat-grozovik:
@neethajohn  could you please review?

neethajohn (approved):

PR #2067, opened 2021-12-06
[202012] Update default route status to state DB (#2009)

prsunny:
**What I did**
* Orchagent update of IPv4 and IPv6 default route status. If there is a valid default route, it shall be updated as state:ok. If there is no valid default route (with packet action drop), state db shall be updated as state:na - Schema ("ROUTE_TABLE|0.0.0.0/0")

cherry-pick PR https://github.com/Azure/sonic-swss/pull/2009

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -75,7 +75,11 @@ RouteOrch::RouteOrch(DBConnector *db, string tableName, SwitchOrch *switchOrch, 
 
     SWSS_LOG_NOTICE("Maximum number of ECMP groups supported is %d", m_maxNextHopGroupCount);
 
+    m_stateDb = shared_ptr<DBConnector>(new DBConnector("STATE_DB", 0));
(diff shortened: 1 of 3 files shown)


orchagent/routeorch.cpp
```diff
@@ -1898,7 +1913,10 @@ bool RouteOrch::addRoutePost(const RouteBulkContext& ctx, const NextHopGroupKey
         }
     }
 
-    m_syncdRoutes[vrf_id][ipPrefix] = nextHops;
```
shi-su, line 1901: What is this change for? The original change in the master branch does not seem to change this.

prsunny: fixed the cherry-pick error. Thanks for finding

shi-su (approved):

qiluo-msft (approved):

PR #2069, opened 2021-12-07
[voq] Neighbor entry impose encap index attribute deprecated

vganesan-nokia:
**What I did**

Changes done to avoid sending SAI_NEIGHBOR_ENTRY_ATTR_ENCAP_IMPOSE_INDEX in neighbor entry while creating neighbor records in SAI

**Why I did it**

Since neighbor entry attribute SAI_NEIGHBOR_ENTRY_ATTR_ENCAP_IMPOSE_INDEX is deprecated, and since the new version of SAI (version > 6.0.0.11; OCP version 1.9.1) returns error (previous versions just ignores this deprecated attribute instead of returning error), this attribute is no longer sent in the neighbor entry creation to avoid orchagent crash when syncd returns error.

**How I verified it**

- Load VOQ chassis line card with image that uses SAI version > 6.0.0.11 (OCP version 1.9.1)
- Bring up an interface and configure IP address to the interface.
- Observe that syncd has not returned any error, orchagent process is up and running.

**Details if related**

### DIFF
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -1541,10 +1541,6 @@ bool NeighOrch::addVoqEncapIndex(string &alias, IpAddress &ip, vector<sai_attrib
             attr.value.u32 = encap_index;
             neighbor_attrs.push_back(attr);
 
-            attr.id = SAI_NEIGHBOR_ENTRY_ATTR_ENCAP_IMPOSE_INDEX;
-            attr.value.booldata = true;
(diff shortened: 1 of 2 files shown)


vganesan-nokia:
@rlhui, Please tag this PR for chassis

vganesan-nokia:
@prsunny, Would you please tag this PR for chassis? Thanks

judyjoseph (approved):
Needed with SAI 6.0.0.13

PR #2071, opened 2021-12-08
[bulk mode] Fix bulk conflict when in case there are both remove and set operations

shi-su:
**What I did**
Check if there are items pending removal in bulk before calling bulk set API.

Fixes https://github.com/Azure/sonic-buildimage/issues/9434

**Why I did it**
When there are items pending removal in bulk before calling set API, it means the item will be removed before the set and it should do create instead.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -1827,7 +1827,7 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
      * (group) id. The old next hop (group) is then not used and the reference
      * count will decrease by 1.
      */
-    if (it_route == m_syncdRoutes.at(vrf_id).end())
+    if (it_route == m_syncdRoutes.at(vrf_id).end() || gRouteBulker.bulk_entry_pending_removal(route_entry))
     {
         if (blackhole)
         {
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -1826,6 +1826,10 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
      * in m_syncdRoutes, then we need to update the route with a new next hop
      * (group) id. The old next hop (group) is then not used and the reference
      * count will decrease by 1.
+     *
+     * In case the entry is already pending removal in the bulk, it would be removed
+     * from m_syncdRoutes during the bulk call. Therefore, such entries need to be
+     * re-created rather than set attribute.
      */
     if (it_route == m_syncdRoutes.at(vrf_id).end() || gRouteBulker.bulk_entry_pending_removal(route_entry))
     {
--- a/orchagent/bulker.h
+++ b/orchagent/bulker.h
@@ -414,6 +414,11 @@ public:
         return creating_entries.count(entry);
     }
 
+    bool bulk_entry_pending_removal(const Te& entry) const
+    {
+        return removing_entries.find(entry) != removing_entries.end();
+    }
+
 private:
     std::unordered_map<                                     // A map of
             Te,                                             // entry ->
--- a/orchagent/mplsrouteorch.cpp
+++ b/orchagent/mplsrouteorch.cpp
@@ -599,7 +599,7 @@ bool RouteOrch::addLabelRoute(LabelRouteBulkContext& ctx, const NextHopGroupKey 
      * (group) id. The old next hop (group) is then not used and the reference
      * count will decrease by 1.
      */
-    if (it_route == m_syncdLabelRoutes.at(vrf_id).end())
+    if (it_route == m_syncdLabelRoutes.at(vrf_id).end() || gLabelRouteBulker.bulk_entry_pending_removal(inseg_entry))
     {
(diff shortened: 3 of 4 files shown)


qiluo-msft:
>     auto it = consumer.m_toSync.begin();

I have another idea. For route orch, if m_toSync gets a DEL/SET, we can safely merge it into a SET. Then it will simplify the bulking logic. his is possible because the new SET has covered all the fields compared the existing fields before DEL.
And this will lead to a general solution.

---
Refers to: orchagent/routeorch.cpp:468 in 2a72407. [](commit_id = 2a72407b7c18f84637ad3d11a2c1349c011a144e, deletion_comment = False)

orchagent/routeorch.cpp
```diff
@@ -1826,8 +1826,12 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
      * in m_syncdRoutes, then we need to update the route with a new next hop
      * (group) id. The old next hop (group) is then not used and the reference
      * count will decrease by 1.
+     *
+     * In case the entry is already pending removal in the bulk, it would be removed
+     * from m_syncdRoutes during the bulk call. Therefore, such entries need to be
+     * re-created rather than set attribute.
      */
-    if (it_route == m_syncdRoutes.at(vrf_id).end())
+    if (it_route == m_syncdRoutes.at(vrf_id).end() || gRouteBulker.bulk_entry_pending_removal(route_entry))
```
qiluo-msft, line 1834: Add a unit test?

stephenxs (commented):
This logic works only if removing a routing entry always succeeds.
If removing a routing entry can fail and the routing entry is still in SAI table in that case, the logic will be broken. Can that happen?

stephenxs:
> > ```
> > auto it = consumer.m_toSync.begin();
> > ```
> 
> I have another idea. For route orch, if m_toSync gets a DEL/SET, we can safely merge it into a SET. Then it will simplify the bulking logic. his is possible because the new SET has covered all the fields compared the existing fields before DEL. And this will lead to a general solution.
> 

Sounds good but I don't know what happens if the routing entry switches from a normal nexthop/nhg to some special thing like a tunnel/MPLS. In that case, is removing the old entry a must or it can just override the attributes?
In [addToSync](https://github.com/Azure/sonic-swss/blob/master/orchagent/orch.cpp#L94) there is similar logic to merge multiple operations on the same key, but it doesn't merge a removal to a setting (`S, D => D`). It only merges a setting to a removal.

shi-su:
> This logic works only if removing a routing entry always succeeds. If removing a routing entry can fail and the routing entry is still in SAI table in that case, the logic will be broken. Can that happen?

This is a valid concern. This logic is based on the assumption that the remove operation is successful. In the current stage, the removal failure would cause orchagent exit, thus the situation should not happen. Also, the original behavior does not seem to be working. Let's assume the remove operation fails, the set operation would become a set_entry_attribute. In case that the set is successful and the remove gets retried, it might end up with the route getting removed though it is supposed to be there.

shi-su:
> > > ```
> > > auto it = consumer.m_toSync.begin();
> > > ```
> > 
> > 
> > I have another idea. For route orch, if m_toSync gets a DEL/SET, we can safely merge it into a SET. Then it will simplify the bulking logic. his is possible because the new SET has covered all the fields compared the existing fields before DEL. And this will lead to a general solution.
> 
> Sounds good but I don't know what happens if the routing entry switches from a normal nexthop/nhg to some special thing like a tunnel/MPLS. In that case, is removing the old entry a must or it can just override the attributes? In [addToSync](https://github.com/Azure/sonic-swss/blob/master/orchagent/orch.cpp#L94) there is similar logic to merge multiple operations on the same key, but it doesn't merge a removal to a setting (`S, D => D`). It only merges a setting to a removal.

I have some concerns about making merging DEL and SET into one SET a general solution. For some SAI APIs, some attributes may only be configurable on creation. In such cases, we should keep both operations and strictly do a DEL and then CREATE. Yet I do agree that it is a valid solution specific to route orch. And it could be an additional performance improvement on top of this fix.

stephenxs:
> > This logic works only if removing a routing entry always succeeds. If removing a routing entry can fail and the routing entry is still in SAI table in that case, the logic will be broken. Can that happen?
> 
> This is a valid concern. This logic is based on the assumption that the remove operation is successful. In the current stage, the removal failure would cause orchagent exit, thus the situation should not happen. Also, the original behavior does not seem to be working. Let's assume the remove operation fails, the set operation would become a set_entry_attribute. In case that the set is successful and the remove gets retried, it might end up with the route getting removed though it is supposed to be there.

Ok. That means it doesn't introduce new risks or degradation.

stephenxs (approved):

dprital:
@shi-su, Can you please retrigger the checkers for this PR ?

qiluo-msft (approved):

stephenxs:
vs test failed due to `##[error]Artifact sonic-gcov already exists for build 58142.`

shi-su:
> vs test failed due to `##[error]Artifact sonic-gcov already exists for build 58142.`

I retriggered the tests. I believe the status of sonic-gcov will be refreshed after this run if it passes.

PR #2073, opened 2021-12-09
[orchagent, crm]: Reset crm threshold exceed count when threshold type changed

link19430:
**What I did**
Reset crm threshold exceed count when threshold type changed

**Why I did it**
The issue is that exceed msg not work when exceed count ran out by pervious threshold type.
Root cause is that doesn't reset the exceed count when changed the threshold type

**How I verified it**
Step1. Trigger exceed high threshold until the exceed count ran out.
Step2. Change the threshold type to a different one and trigger exceed high threshold
Step3. Check the exceed count reset to 0 and the exceed msg display ok.

Signed-off-by: link_chiang <link_chiang@edge-core.com>


gord1306:
@prsunny could you help to review this PR, thanks

volodymyrsamotiy:
Changes look good to me.
Could you please also add VS test for this scenario?
_(VS tests for CRM are here: https://github.com/Azure/sonic-swss/blob/master/tests/test_crm.py)_

link19430:
@prsunny @volodymyrsamotiy
I have added the test case in vstest, please help review, thanks.

volodymyrsamotiy (dismissed):

tests/test_crm.py
```diff
@@ -720,6 +720,60 @@ def test_CrmDnatEntry(self, dvs, testlog):
...
+    def test_CrmResetThresholdExceedCount(self, dvs, testlog):
+
+        config_db = swsscommon.DBConnector(swsscommon.CONFIG_DB, dvs.redis_sock, 0)
+        intf_tbl = swsscommon.Table(config_db, "INTERFACE")
+        fvs = swsscommon.FieldValuePairs([("NULL","NULL")])
+        intf_tbl.set("Ethernet0", fvs)
+        intf_tbl.set("Ethernet0|10.0.0.0/31", fvs)
+        dvs.runcmd("config interface startup Ethernet0")
+
+        dvs.runcmd("crm config polling interval 1")
```
prsunny, line 732: @link19430 , please use `crm_update(dvs, "polling_interval", "1")` and fix all the CLI command usage. VS tests are modified *not* use CLI commands. You can take a look at other tests for reference.

prsunny (approved):

tests/test_crm.py
link19430: all cli commands had modified to use crm_update()

prsunny (approved):

PR #2075, opened 2021-12-09
set remote vtep the netdev down before delete.

iris00522:
**What I did**
set vtep-xxx the netdev down before delete vtep-xxx .
**Why I did it**
the vxlan netdev will be only delete only when remove vxlan tunnel map,
but the FRR doesn't remove IMET route entirely after receiving the delete event.
Therefore, add to set netdev down before delete the netdev to trigger FRR to remove related IMET route.

**How I verified it**
using t0 topology config
and flow the cmmand

redis-cli -n 4 -c DEL "VLAN_MEMBER|Vlan1000|Ethernet4"
sudo config interface ip add Ethernet4 10.0.0.64/31
vtysh
configure
ip route 100.0.0.65/31 10.0.0.65
exit
exit

sudo config vxlan add vtep 10.1.0.32
sudo config vxlan evpn_nvo add evpnnvo1 vtep
sudo config vxlan map add vtep 1000 10000

vtysh
configure
router bgp 65100
neighbor 10.0.0.65 remote-as 65200
address-family l2vpn evpn
neighbor 10.0.0.65 activate
advertise-all-vni
exit
exit
exit
exit

neighbor:
setting bgp and type 3 and typ2 packet 

**Details if related**

### DIFF
--- a/cfgmgr/vxlanmgr.cpp
+++ b/cfgmgr/vxlanmgr.cpp
@@ -643,6 +643,7 @@ bool VxlanMgr::doVxlanTunnelMapDeleteTask(const KeyOpFieldsValuesTuple & t)
     vxlan_dev_name = map_entry.vxlan_dev_name;
     vlan = map_entry.vlan;
     vni_id = map_entry.vni_id;
+    downVxlanNetdevice(vxlan_dev_name);
     deleteVxlanNetdevice(vxlan_dev_name);
 
     m_vxlanTunnelMapCache.erase(vxlanTunnelMapName);
@@ -906,11 +907,11 @@ void VxlanMgr::createAppDBTunnelMapTable(const KeyOpFieldsValuesTuple & t)
     std::replace(vxlanTunnelMapName.begin(), vxlanTunnelMapName.end(), config_db_key_delimiter, delimiter);
 
     /* Case 1: Entry exist - Erase from cache & return
-     * Case 2: Entry does not exist - Write to AppDB 
+     * Case 2: Entry does not exist - Write to AppDB
      * Case 3: Entry exist but modified - Not taken care. Will address later
      */
     if (m_in_reconcile)
-    { 
+    {
         auto it = find(m_appVxlanTunnelMapKeysRecon.begin(), m_appVxlanTunnelMapKeysRecon.end(), vxlanTunnelMapName);
         if (it != m_appVxlanTunnelMapKeysRecon.end())
         {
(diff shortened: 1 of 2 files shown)


prsunny:
@srj102 , @dgsudharsan for review

cfgmgr/vxlanmgr.cpp
```diff
@@ -78,6 +78,16 @@ static int cmdUpVxlan(const swss::VxlanMgr::VxlanInfo & info, std::string & res)
     return swss::exec(cmd.str(), res);
 }
 
+static int cmdDownVxlan(const swss::VxlanMgr::VxlanInfo & info, std::string & res)
```
dgsudharsan, line 81: This comes in VNET flow and not in regular EVPN flow. Is it needed?

cfgmgr/vxlanmgr.cpp
```diff
@@ -130,6 +140,16 @@ static int cmdUpVxlanIf(const swss::VxlanMgr::VxlanInfo & info, std::string & re
     return swss::exec(cmd.str(), res);
 }
 
+static int cmdDownVxlanIf(const swss::VxlanMgr::VxlanInfo & info, std::string & res)
```
dgsudharsan, line 143: Same as above. These APIs are called only for VNET table. Are these changes needed?

dgsudharsan (changes requested):

dgsudharsan:
There are some APIs which are in VNET flow that are changed. These may not be related to EVPN flow the issue is reported. Should we change? @prsunny 
@srj102  Can you also please review?

prsunny:
> There are some APIs which are in VNET flow that are changed. These may not be related to EVPN flow the issue is reported. Should we change? @prsunny @srj102 Can you also please review?

Agree with @dgsudharsan , this doesn't fall into the EVPN flow. Can you please check what's intended?

dgsudharsan (approved):

dgsudharsan:
@iris00522 Can you please sync your changes to latest master?

iris00522:
@dgsudharsan @prsunny  help to review again, thank you.

gord1306:
@dgsudharsan @prsunny can this PR be merged? Thanks

prsunny:
@dgsudharsan , can you please signoff?

dgsudharsan (approved):

prsunny (approved):

PR #1857, opened 2021-08-06
[buffer orch] Bugfix: Don't query counter SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES on a pool where it is not supported

stephenxs:
**What I did**
Currently, SAI_BUFFER_POOL_STAT_WATERMARK_BYTES and SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES are queried for buffer pools.
However, the latter is not supported on all pools and all platforms, which will result in sairedis complaint like this:
```
Jun 17 23:31:34.949613 sonic ERR syncd#SDK: :- updateSupportedBufferPoolCounters: BUFFER_POOL_WATERMARK_STAT_COUNTER: counter SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES is not supported on buffer pool oid:0x1000000018, rv: SAI_STATUS_NOT_IMPLEMENTED
Jun 17 23:31:34.951915 sonic ERR syncd#SDK: :- updateSupportedBufferPoolCounters: BUFFER_POOL_WATERMARK_STAT_COUNTER: counter SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES is not supported on buffer pool oid:0x1200000018, rv: SAI_STATUS_NOT_IMPLEMENTED
Jun 17 23:31:34.954177 sonic ERR syncd#SDK: :- updateSupportedBufferPoolCounters: BUFFER_POOL_WATERMARK_STAT_COUNTER: counter SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES is not supported on buffer pool oid:0x400000018, rv: SAI_STATUS_NOT_IMPLEMENTED
```
To avoid that, we need to test whether it is supported before putting it to FLEX_COUNTER_DB

Signed-off-by: Stephen Sun <stephens@nvidia.com>

**Why I did it**

**How I verified it**
Run vstest and manually test.

**Details if related**

PR #2081, opened 2021-12-14
Revert "Revert "[buffer orch] Bugfix: Don't query counter SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES on a pool where it is not supported (#1857)"

stephenxs:
**What I did**

Reverts Azure/sonic-swss#1945 which reverted PR #1857 
PR #1857 was introduced to avoid some error messages on some platforms. However, it made issue Azure/sonic-sairedis#862 more likely to be reproduced and has been reverted.
Now that the issue has been fixed by #1987, we can merge the PR.

Signed-off-by: Stephen Sun stephens@nvidia.com

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -23,6 +23,16 @@ extern sai_object_id_t gSwitchId;
 
 static const vector<sai_buffer_pool_stat_t> bufferPoolWatermarkStatIds =
 {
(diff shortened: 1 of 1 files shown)


stephenxs:
Sairedis has been updated accordingly so we do not need this PR any more.

PR #2085, opened 2021-12-20
[Buffer] Fix the bug that buffer configuration will cause failed logs during warm restart.

steven-guo-ec:
- What I do:
  Let BufferMgr not execute the entries which are already in APPL_DB after a warm restart.

- Why I did it:
  After warm start, the BufferMgr will set entries into APPL_DB which already exist.
  It will cause bufferorch to do a modification event.
  But, currently, modifying the buffer configuration is not supported in SAI, and it will print error messages.
  The error messages will cause test cases to fail.

- How I verify it:
  With this enhancement, the syslog will not appear the error messages about Buffer during a warm restart.

### DIFF
--- a/cfgmgr/buffermgr.cpp
+++ b/cfgmgr/buffermgr.cpp
@@ -287,6 +287,13 @@ void BufferMgr::doBufferTableTask(Consumer &consumer, ProducerStateTable &applTa
         string op = kfvOp(t);
         if (op == SET_COMMAND)
         {
+            if (m_warmStartAppTableKeys[applTable.getTableName()].find(key) != m_warmStartAppTableKeys[applTable.getTableName()].end())
+            {
+                m_warmStartAppTableKeys[applTable.getTableName()].erase(key);
+                it = consumer.m_toSync.erase(it);
+                continue;
+            }
+
             vector<FieldValueTuple> fvVector;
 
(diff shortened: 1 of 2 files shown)


stephenxs:
@steven-guo-ec 
can you please elaborate on which buffer attributes are not supported in SAI?
IMO buffer orch will not set create-only attributes

prsunny:
Please add VS tests

steven-guo-ec:
SAI_BUFFER_POOL_ATTR_SIZE and SAI_BUFFER_PROFILE_ATTR_BUFFER_SIZE  is not supported to set after they were created, they can be configured only when creating them.

stephenxs:
> SAI_BUFFER_POOL_ATTR_SIZE and SAI_BUFFER_PROFILE_ATTR_BUFFER_SIZE is not supported to set after they were created, they can be configured only when creating them.

Thanks. But this does not comply with SAI definition. Check [SAI_BUFFER_POOL_ATTR_SIZE](https://github.com/opencomputeproject/SAI/blob/master/inc/saibuffer.h#L298) and [SAI_BUFFER_PROFILE_ATTR_RESERVED_BUFFER_SIZE](https://github.com/opencomputeproject/SAI/blob/master/inc/saibuffer.h#L577) for details.
So IMO this should be fixed by vendor SAI.

PR #2086, opened 2021-12-20
[Route bulk] Fix bugs in case a SET operation follows a DEL operation in the same bulk

stephenxs:
**What I did**

Fix 2 issues in route bulk operation in the scenario where a SET operation follows a DEL operation on the same prefix.
1. The route entry SET operation can be ignored if it follows a DEL operation with the same prefix and next-hop.
2. DEL and then SET default route results in orchagent aborted

Added test case to cover the scenarios

**Why I did it**

**How I verified it**

Manually test
Unit test

**Details if related**
***The route entry SET operation can be ignored if it follows a DEL operation with the same prefix and next-hop.***
For example, the following operations
```
2021-12-18.06:17:54.793234|ROUTE_TABLE:1.1.1.0/24|SET|nexthop:10.0.0.33|ifname:Ethernet64
2021-12-18.06:19:53.187294|ROUTE_TABLE:1.1.1.0/24|DEL
2021-12-18.06:20:10.279850|ROUTE_TABLE:1.1.1.0/24|SET|nexthop:10.0.0.33|ifname:Ethernet64
```
result in the next sai operations
```
2021-12-18.06:19:31.203092|C|SAI_OBJECT_TYPE_ROUTE_ENTRY||{"dest":"1.1.1.0/24","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000010"}|SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID=oid:0x4000000000aae
2021-12-18.06:20:23.813475|R|SAI_OBJECT_TYPE_ROUTE_ENTRY||{"dest":"1.1.1.0/24","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000010"}
```
and the last SET is lost.

This is because:
Both DEL and the 2nd SET operations were bundled into bulk operations.
There is a logic that is not to program ASIC if the next-hop information in the route entry update isn't changed,
which is done by comparing the next-hop information between route entry update and m_syncdRoutes.
However, in bulk mode, a route entry that will be removed will not be removed from m_syncdRoutes until the bulk is flushed,
The bulk has not been flushed when the 2nd SET is handled,
which means the comparison returns "not changed" and the 2nd SET operation is ignored.

The fix is to check whether the prefix is in gRouteBulker.removing_entries and treat it as non-exist if yes.

***DEL and then SET default route results in orchagent aborted***

This is because:
1. packet forward action is initialized when route entry is created
2. the default route will be set to DROP if the application asks for removing it
3. if the default route is readded later in the same bulk session,
   it doesn't update packet forward action but provides a valid next-hop
4. SAI complains if both DROP and a valid next hop are provided

Solution:
Always set packet action for the default route

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -1880,6 +1880,25 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
         }
         else
         {
+            if (!blackhole && vrf_id == gVirtualRouterId && ipPrefix.isDefaultRoute())
+            {
+                // Always set packet action for default route to avoid conflict settings
+                // in case a SET follows a DEL on the default route in the same bulk.
+                // - On DEL default route, the packet action will be set to DROP
+                // - On SET default route, as the default route has NOT been removed from m_syncdRoute
+                //   it calls SAI set_route_attributes instead of crate_route
+                //   However, packet action is called only when a route entry is created
+                //   This leads to conflict settings:
+                //   - packet action: DROP
+                //   - next hop: a valid next hop id
+                // To avoid this, we always set packet action for default route.
+                route_attr.id = SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION;
+                route_attr.value.s32 = SAI_PACKET_ACTION_FORWARD;
+
+                object_statuses.emplace_back();
+                gRouteBulker.set_entry_attribute(&object_statuses.back(), &route_entry, &route_attr);
+            }
+
             route_attr.id = SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID;
             route_attr.value.oid = next_hop_id;
 
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -790,6 +790,11 @@ void RouteOrch::doTask(Consumer& consumer)
                     }
                 }
 
+                sai_route_entry_t route_entry;
+                route_entry.vr_id = vrf_id;
+                route_entry.switch_id = gSwitchId;
+                copy(route_entry.destination, ip_prefix);
+
                 if (nhg.getSize() == 1 && nhg.hasIntfNextHop())
                 {
                     if (alsv[0] == "unknown")
@@ -833,6 +838,7 @@ void RouteOrch::doTask(Consumer& consumer)
                 else if (m_syncdRoutes.find(vrf_id) == m_syncdRoutes.end() ||
                     m_syncdRoutes.at(vrf_id).find(ip_prefix) == m_syncdRoutes.at(vrf_id).end() ||
                     m_syncdRoutes.at(vrf_id).at(ip_prefix) != RouteNhg(nhg, ctx.nhg_index) ||
+                    gRouteBulker.bulk_entry_pending_removal(route_entry) ||
                     ctx.using_temp_nhg)
                 {
                     if (addRoute(ctx, nhg))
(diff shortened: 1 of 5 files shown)


orchagent/routeorch.cpp
```diff
@@ -833,6 +838,7 @@ void RouteOrch::doTask(Consumer& consumer)
                 else if (m_syncdRoutes.find(vrf_id) == m_syncdRoutes.end() ||
                     m_syncdRoutes.at(vrf_id).find(ip_prefix) == m_syncdRoutes.at(vrf_id).end() ||
                     m_syncdRoutes.at(vrf_id).at(ip_prefix) != RouteNhg(nhg, ctx.nhg_index) ||
+                    gRouteBulker.bulk_entry_pending_removal(route_entry) ||
```
qiluo-msft, line 841: >bulk_entry_pending_removal [](http://example.com/codeflow?start=33&length=26)

This is a valid fix.

However I am wondering how you find this issue. By design, m_toSync will already merge the sequence SET1,DEL,SET2 to DEL,SET2, so I think your first issue is not existing in orchagent process.
ref: [https://github.com/Azure/sonic-swss/blob/a4c80c3d8713d32b73d9428683b8670b46d65fcd/orchagent/orch.cpp#L108](https://github.com/Azure/sonic-swss/blob/a4c80c3d8713d32b73d9428683b8670b46d65fcd/orchagent/orch.cpp#L108)

Is it true?

qiluo-msft:
@shi-su Could you help review?

orchagent/routeorch.cpp
shi-su: I felt this is not a `SET1,DEL,SET2` problem. It sound more like a route was previously configured to a nexthop group. Afterward, there is a DEL request followed by a SET request to configure to the same nexthop group. Following the logic of route orch, it would first put the DEL into the bulk. Then for the SET, it will be ignored since the nexthop group is the same as it was before removal. So this is not a second SET operation involved but the SET operation is getting ignored.

orchagent/routeorch.cpp
```diff
@@ -1874,6 +1880,25 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
         }
         else
         {
+            if (!blackhole && vrf_id == gVirtualRouterId && ipPrefix.isDefaultRoute())
```
shi-su, line 1883: Does this mean we set packet action for the default route every time even if it is not changed? Is there any possibility to set it only when required?

orchagent/routeorch.cpp
stephenxs: Hi @qiluo-msft
Thanks for reviewing this.
The issue does exist in orchagent.
I understood that logic but It takes effect only if all three operations are received together and then handled. 
When this issue occurred, the first SET is received and handled, and then the DEL and 2nd SET. In fact, it doesn’t matter how long between the first SET and the DEL. Only the fact that a routing entry with a same next hop existing matters.
Actually this is a similar scenario with issue #9434. The difference is that the next hops were different between the existing entry and the SET operation, but are same here.
In regression test, this resulted in router check failed.
Is that clear?

orchagent/routeorch.cpp
stephenxs: Setting it to DROP was done in removeRoute and setting it to FORWARD is in addRoute and there is no global state recording this.

stephenxs:
@shi-su @qiluo-msft can we merge this pr if no other concern? We need it to fix azure pipeline checker failures. Thanks

shi-su (approved):

qiluo-msft (approved):

qiluo-msft:
Do we need to backport this PR to old branches?

stephenxs:
> Do we need to backport this PR to old branches?

I suggest doing so as it is a production issue. We need #2071 first

PR #2090, opened 2021-12-22
[Mellanox] '_8lane' not added to Mellanox 5xxx models with 800G

raphaelt-nvidia:
**What I did**

Previously, 400G buffer profiles for Mellanox 4xxx switches did not add '_8lane' to the profile name, because there is no other possibility. Now, the condition also applies to 800G buffer profiles for 5xxx switches. The work includes code that other vendors could use in buffermgrdyn if they need to distinguish between models.

**Why I did it**

No need for '_8lane' in profile name when there is no other possibility.

**How I verified it**

buffershow -l

**Details if related**


liat-grozovik:
@stephenxs can you please review?

stephenxs (approved):

raphaelt-nvidia:
azpw run

stephenxs (approved):

liat-grozovik (approved):

PR #2092, opened 2021-12-28
OA changes to support Ordered ECMP and DVS test for same.

abdosi:
What I did:-

- Orchagent changes to support Ordered ECMP as per design https://github.com/Azure/SONiC/pull/896
- While testing Order ECMP for Overlay ECMP use case found that for single nexthop based tunnel we create route without checking BFD state for endpoint. This results in not needed transaction of `add_route` followed by `del_route`.

How I verify:-

- Unit test updated to run for both scenario order ecmp as `true` and `false`.
- DVS unit-test will need this PR https://github.com/Azure/sonic-sairedis/pull/989 for passing.


orchagent/switchorch.h
```diff
@@ -10,6 +10,7 @@
 
 #define SWITCH_CAPABILITY_TABLE_PORT_TPID_CAPABLE                      "PORT_TPID_CAPABLE"
 #define SWITCH_CAPABILITY_TABLE_LAG_TPID_CAPABLE                       "LAG_TPID_CAPABLE"
+#define SWITCH_CAPABILITY_ORDERED_ECMP_CAPABLE                         "ORDERED_ECMP_CAPABLE"
```
shi-su, line 13: Should this be named `SWITCH_CAPABILITY_TABLE_ORDERED_ECMP_CAPABLE` for better consistency?

abdosi: updated

shi-su (dismissed):

shi-su (dismissed):

shi-su (approved):

PR #2094, opened 2022-01-04
Add support for route flow counter

Junchao-Mellanox:
HLD: https://github.com/Azure/SONiC/pull/908

Build dependency: https://github.com/Azure/sonic-swss-common/pull/576

**What I did**

Add support for route flow counter

**Why I did it**

For feature route flow counter

**How I verified it**

1. Added VS test cases
2. Manual test
3. sonic-mgmt test cases

**Details if related**


orchagent/routeorch.cpp
```diff
@@ -2547,3 +2565,968 @@ void RouteOrch::decNhgRefCount(const std::string &nhg_index)
...
+        flushUnbindFlowCounters(to_unbind);
+
+        mBoundRouteCounters.clear();
+        mPendingAddToFlexCntr.clear();
+    }
+
+    mUnboundRoutes.clear();
+}
+
+void RouteOrch::updateRoutePattern(const std::string &pattern, size_t max_match_count)
```
dgsudharsan, line 2612: The updateRoutePattern seems to add new pattern to the list, rather than replacing existing pattern. The requirement of 1 pattern per IP type is handled at CLI and so the backend is not enforcing it leaving for easy migration to multiple patterns per type in the future. Am I correct here?
If that's the case can we rename the API to addRoutePattern?

orchagent/vnetorch.cpp
```diff
@@ -547,11 +549,45 @@ bool VNetOrch::delOperation(const Request& request)
     return true;
 }
 
+bool VNetOrch::getVrfIdByVnetName(const std::string& vnet_name, sai_object_id_t &vrf_id)
```
dgsudharsan, line 552: Should there be logic to handle vnet addition similar to onAddVR?

dgsudharsan (changes requested):

orchagent/routeflowcounterorch.cpp
```diff
@@ -0,0 +1,87 @@
+#include "dbconnector.h"
```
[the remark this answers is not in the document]
Junchao-Mellanox: This is following arch design: to have a separate orch handles route pattern table change.

orchagent/routeorch.cpp
```diff
@@ -26,8 +33,10 @@ extern CbfNhgOrch *gCbfNhgOrch;
 extern size_t gMaxBulkSize;
 
 /* Default maximum number of next hop groups */
-#define DEFAULT_NUMBER_OF_ECMP_GROUPS   128
-#define DEFAULT_MAX_ECMP_GROUP_SIZE     32
+#define DEFAULT_NUMBER_OF_ECMP_GROUPS               128
+#define DEFAULT_MAX_ECMP_GROUP_SIZE                 32
+#define FLEX_COUNTER_UPD_INTERVAL                   1
+const uint ROUTE_FLOW_COUNTER_POLLING_INTERVAL_MS = 10000;
```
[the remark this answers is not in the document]
Junchao-Mellanox: Will change it to macro.

orchagent/routeorch.cpp
```diff
@@ -2547,3 +2565,968 @@ void RouteOrch::decNhgRefCount(const std::string &nhg_index)
...
+        if (!bulkBindFlowCounters(route_pattern.vrf_id, *prefix_iter, to_bind, current_bound_count))
+        {
+            SWSS_LOG_WARN("Failed to bind route flow counter, vrf=%s, ip_prefix=%s", sai_serialize_object_id(route_pattern.vrf_id).c_str(), prefix_iter->to_string().c_str());
+            iter->second.erase(prefix_iter);
+            continue;
+        }
+
+        if (current_bound_count >= route_pattern.max_match_count)
+        {
+            flushBindFlowCounters(route_pattern, to_bind, current_bound_count);
```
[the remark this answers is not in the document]
Junchao-Mellanox: yes, it is needed because there could be failures when binding counter to a route entry. Once there is a failure, we do "--current_bound_count" inside flushBindFlowCounters. And that means we still have "room" for more route flow counters.

orchagent/vnetorch.cpp
Junchao-Mellanox: Could yo please elaborate? I don't quite understand the comment here.

orchagent/routeorch.cpp
```diff
@@ -2547,3 +2565,968 @@ void RouteOrch::decNhgRefCount(const std::string &nhg_index)
...
+    std::set<RoutePattern> pending_fill;
+
+    for (auto &entry : mPendingUnbindRoutes)
+    {
+        const auto &route_pattern = entry.first;
+        auto vrf_id = route_pattern.vrf_id;
+        for (auto ip_prefix : entry.second)
+        {
+            // Check if the entry is in mPendingBindRoutes, if yes, remove it
+            auto pending_bind_iter = mPendingBindRoutes.find(route_pattern);
```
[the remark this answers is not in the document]
Junchao-Mellanox: According to issue https://github.com/Azure/sonic-buildimage/issues/9434, here it should do unbind and bind. I will remove the check here, it is not necessary.

orchagent/routeorch.cpp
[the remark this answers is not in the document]
dgsudharsan: What is the purpose of calling it at the end of the function? Is it redundant?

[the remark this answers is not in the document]
Junchao-Mellanox: In case it did not reach max_match_count in the loop.

dgsudharsan (approved):

stephenxs (dismissed):

Junchao-Mellanox:
Did a rebase to try to fix the checker failure, no code change.

stephenxs (dismissed):

orchagent/flex_counter/flex_counter_manager.cpp
```diff
@@ -44,6 +44,7 @@ const unordered_map<CounterType, string> FlexCounterManager::counter_id_field_lo
     { CounterType::ACL_COUNTER,     ACL_COUNTER_ATTR_ID_LIST },
     { CounterType::TUNNEL,          TUNNEL_COUNTER_ID_LIST },
     { CounterType::HOSTIF_TRAP,     FLOW_COUNTER_ID_LIST },
+    { CounterType::ROUTE,     FLOW_COUNTER_ID_LIST },
```
prsunny, line 47: Please align

orchagent/vnetorch.cpp
```diff
@@ -551,11 +553,45 @@ bool VNetOrch::delOperation(const Request& request)
...
+
+    return false;
+}
+
 /*
  * Vnet Route Handling
  */
 
-static bool del_route(sai_object_id_t vr_id, sai_ip_prefix_t& ip_pfx)
+static bool del_route(sai_object_id_t vr_id, sai_ip_prefix_t& ip_pfx, const IpPrefix& ipPrefix)
```
prsunny, line 594: This is too many changes in API definitions. Cherry-picking a future fix will be causing conflicts. Can you derive ipPrefix from ip_pfx?

orchagent/routeorch.h
```diff
@@ -248,6 +364,32 @@ class RouteOrch : public Orch, public Subject
     EntityBulker<sai_mpls_api_t>            gLabelRouteBulker;
     ObjectBulker<sai_next_hop_group_api_t>  gNextHopGroupMemberBulker;
 
+    std::shared_ptr<DBConnector> mAsicDb;
```
prsunny, line 367: This is significant changes to routeorch and prefer to have this handled in routeflowcounter.cpp. Let routeorch make some simple calls during add delete operation and rest be handled outside of this file. Please refer to examples of fgnhg or cbf.

prsunny (changes requested):

Junchao-Mellanox:
Hi @prsunny, @dgsudharsan, @stephenxs , as there is a big change, could you please review again?

Junchao-Mellanox:
Build dependency: https://github.com/Azure/sonic-swss-common/pull/593

orchagent/routeorch.h
```diff
@@ -16,13 +16,17 @@
 #include "bulker.h"
 #include "fgnhgorch.h"
 #include <map>
+#include <memory>
```
dgsudharsan, line 19: Do we need changes in this file?

orchagent/swssnet.h
```diff
@@ -76,6 +77,53 @@ inline static sai_ip_prefix_t& copy(sai_ip_prefix_t& dst, const IpAddress& src)
...
+
+inline static IpPrefix copy(const sai_ip_prefix_t& src)
+{
+    ip_addr_t ip;
+    switch(src.addr_family)
+    {
+        case SAI_IP_ADDR_FAMILY_IPV4:
+            ip.family = AF_INET;
+            ip.ip_addr.ipv4_addr = src.addr.ip4;
+            return IpPrefix(ip, getPrefixLenFromAddr(reinterpret_cast<const uint8_t*>(&src.mask.ip4), 4));
```
dgsudharsan, line 117: Why do we need to calculate the getPrefixLenFromAddr? Can we just copy? Can we also make inline with other copy APIs? by passing IpPrefix as argument? 
https://github.com/Azure/sonic-swss/blob/e40c7367232eaea2374ea94fb99771326057c2a0/orchagent/swssnet.h#L44

dgsudharsan (changes requested):

orchagent/flex_counter/flowcounterrouteorch.cpp
```diff
@@ -0,0 +1,965 @@
...
+}
+
+void FlowCounterRouteOrch::removeRouteFlowCounterFromDB(sai_object_id_t vrf_id, const IpPrefix& ip_prefix, sai_object_id_t counter_oid)
+{
+    SWSS_LOG_ENTER();
+    std::string nameMapKey;
+    getRouteFlowCounterNameMapKey(vrf_id, ip_prefix, nameMapKey);
+    mPrefixToPatternTable->hdel("", nameMapKey);
+    mPrefixToCounterTable->hdel("", nameMapKey);
+    mRouteFlowCounterMgr.clearCounterIdList(counter_oid);
```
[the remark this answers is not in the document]
Junchao-Mellanox: Already did it at [10:30 AM] Junchao Chen
https://github.com/Azure/sonic-sairedis/blob/473c99067c8132f8e5b3f8f2abd672b474b42b5a/syncd/FlexCounter.cpp#L986

orchagent/flex_counter/flowcounterrouteorch.cpp
```diff
@@ -0,0 +1,965 @@
...
+    {
+        for (auto &entry : pending_iter->second)
+        {
+            unbindFlowCounter(route_pattern, route_pattern.vrf_id, entry.first, entry.second);
+        }
+        mPendingAddToFlexCntr.erase(pending_iter);
+    }
+}
+
+void FlowCounterRouteOrch::onAddMiscRouteEntry(sai_object_id_t vrf_id, const sai_ip_prefix_t& ip_pfx, bool add_to_cache)
```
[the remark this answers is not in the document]
Junchao-Mellanox: yes.

orchagent/routeorch.h
Junchao-Mellanox: Will remove.

orchagent/swssnet.h
Junchao-Mellanox: Will rename copy to convert

dgsudharsan (dismissed):

orchagent/routeorch.cpp
```diff
@@ -22,12 +28,13 @@ extern CrmOrch *gCrmOrch;
...
 extern NhgOrch *gNhgOrch;
 extern CbfNhgOrch *gCbfNhgOrch;
+extern FlowCounterRouteOrch *gFlowCounterRouteOrch;
 
 extern size_t gMaxBulkSize;
 
 /* Default maximum number of next hop groups */
-#define DEFAULT_NUMBER_OF_ECMP_GROUPS   128
-#define DEFAULT_MAX_ECMP_GROUP_SIZE     32
+#define DEFAULT_NUMBER_OF_ECMP_GROUPS               128
```
prsunny, line 36: Why this change?, Please revert

orchagent/routeorch.cpp
```diff
@@ -1,13 +1,19 @@
...
 #include "routeorch.h"
 #include "nhgorch.h"
 #include "cbf/cbfnhgorch.h"
+#include "flexcounterorch.h"
 #include "logger.h"
+#include "flowcounterrouteorch.h"
 #include "swssnet.h"
 #include "crmorch.h"
 #include "directory.h"
+#include "sai_serialize.h"
```
prsunny, line 15: Please revert all the extra includes that are not used

orchagent/vnetorch.cpp
```diff
@@ -97,6 +99,7 @@ bool VNetVrfObject::createObj(vector<sai_attribute_t>& attrs)
                            vnet_name_.c_str(), status);
             throw std::runtime_error("Failed to create VR object");
         }
+        gFlowCounterRouteOrch->onAddVR(router_id);
```
prsunny, line 102: Vnet is a special case. I don't think we should add vnet routes to flowcounters. May be it can be considered for later phase

orchagent/swssnet.h
```diff
@@ -76,6 +77,53 @@ inline static sai_ip_prefix_t& copy(sai_ip_prefix_t& dst, const IpAddress& src)
     return dst;
 }
 
+static int getPrefixLenFromAddr(const uint8_t *addr, int len)
```
prsunny, line 80: Not sure I understand this. What is the change for? If I assume correct, please use` family=Ipv4?32:128`

orchagent/swssnet.h
```diff
@@ -76,6 +77,53 @@ inline static sai_ip_prefix_t& copy(sai_ip_prefix_t& dst, const IpAddress& src)
...
+        while(((non_zero >> j) & 0x1) == 0)
+        {
+            ++j;
+        }
+        return (i + 1) * 8 - (j + 1);
+    }
+    
+}
+
+inline static IpPrefix convert(const sai_ip_prefix_t& src)
```
prsunny, line 109: I don't think we should be having such a utility to convert back. Please revisit the use-case.

orchagent/flex_counter/flowcounterrouteorch.cpp
```diff
@@ -0,0 +1,965 @@
...
+            // Should not happen, just in case
+            SWSS_LOG_ERROR("Failed to get VRF/VNET name for vrf id %s", sai_serialize_object_id(vrf_id).c_str());
+        }
+    }
+    oss << ip_prefix.to_string();
+    key = oss.str();
+}
+
+void FlowCounterRouteOrch::handleRouteAdd(sai_object_id_t vrf_id, const IpPrefix& ip_prefix)
+{
```
prsunny, line 846: Why is `mRouteFlowCounterSupported `not checked here as in other APIs. Please add a check

prsunny (changes requested):

orchagent/swssnet.h
Junchao-Mellanox: I suppose we need this, e.g mask 255.255.0.0 should be convert to 16.

orchagent/swssnet.h
Junchao-Mellanox: This change is for your previous comments: https://github.com/Azure/sonic-swss/pull/2094#discussion_r834790424 
In order to fix this, need to convert from sai_ip_prefix to IpPrefix.

orchagent/vnetorch.cpp
Junchao-Mellanox: I prefer to keep this as this is required in the HLD review. Thanks.

dgsudharsan (dismissed):

orchagent/flex_counter/flowcounterrouteorch.cpp
```diff
@@ -0,0 +1,985 @@
...
+    }
+
+    IpPrefix ip_prefix = convert(ip_pfx);
+    onAddMiscRouteEntry(vrf_id, ip_prefix, add_to_cache);
+}
+
+void FlowCounterRouteOrch::onAddMiscRouteEntry(sai_object_id_t vrf_id, const IpPrefix& ip_prefix, bool add_to_cache)
+{
+    SWSS_LOG_ENTER();
+    if (!mRouteFlowCounterSupported)
```
prsunny, line 314: I think there is a gap here. My point was this code flow is executed only if `m_route_flow_counter_enabled` is true. Not just platform support.  The idea was, there is no change to the existing flow, if the feature is disabled.  We should check for both cases here.  `!mRouteFlowCounterSupported || !m_route_flow_counter_enabled`

orchagent/flex_counter/flowcounterrouteorch.cpp
```diff
@@ -0,0 +1,985 @@
...
+mPrefixToCounterTable(std::unique_ptr<Table>(new Table(mCounterDb.get(), COUNTERS_ROUTE_NAME_MAP))),
+mPrefixToPatternTable(std::unique_ptr<Table>(new Table(mCounterDb.get(), COUNTERS_ROUTE_TO_PATTERN_MAP))),
+mRouteFlowCounterMgr(ROUTE_FLOW_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, ROUTE_FLOW_COUNTER_POLLING_INTERVAL_MS, false),
+gRouteBulker(sai_route_api, gMaxBulkSize)
+{
+    SWSS_LOG_ENTER();
+    initRouteFlowCounterCapability();
+    if (mRouteFlowCounterSupported)
+    {
+        auto intervT = timespec { .tv_sec = FLEX_COUNTER_UPD_INTERVAL , .tv_nsec = 0 };
```
prsunny, line 42: I think this should be done if user enables the feature first time.

orchagent/routeorch.cpp
```diff
@@ -1,13 +1,18 @@
...
 #include "routeorch.h"
 #include "nhgorch.h"
 #include "cbf/cbfnhgorch.h"
+#include "flexcounterorch.h"
 #include "logger.h"
+#include "flowcounterrouteorch.h"
 #include "swssnet.h"
 #include "crmorch.h"
 #include "directory.h"
+#include "timer.h"
```
prsunny, line 15: Please remove the unused header files from this. `sstream, list, timer, flexcounterorch`

orchagent/routeorch.h
```diff
@@ -174,6 +174,7 @@ struct LabelRouteBulkContext
     }
 };
 
+
```
prsunny, line 177: Revert this change

orchagent/routeorch.h
```diff
@@ -214,10 +215,10 @@ class RouteOrch : public Orch, public Subject
 
     unsigned int getNhgCount() { return m_nextHopGroupCount; }
     unsigned int getMaxNhgCount() { return m_maxNextHopGroupCount; }
-
```
prsunny, line 217: Revert the change

orchagent/swssnet.h
prsunny: ok. Then I suggest rename to `getPrefixLenFromAddrMask`

orchagent/swssnet.h
prsunny: ok, then suggest rename to `getIpPrefixFromSaiPrefix`

orchagent/vrforch.cpp
```diff
@@ -11,15 +11,20 @@
...
 using namespace swss;
 
 extern sai_virtual_router_api_t* sai_virtual_router_api;
 extern sai_object_id_t gSwitchId;
-extern Directory<Orch*> gDirectory;
-extern PortsOrch*       gPortsOrch;
+
+extern Directory<Orch*>      gDirectory;
+extern PortsOrch*            gPortsOrch;
+extern RouteOrch*            gRouteOrch;
```
prsunny, line 26: Please remove unused extern

orchagent/vrforch.cpp
```diff
@@ -11,15 +11,20 @@
 #include "request_parser.h"
 #include "vrforch.h"
 #include "vxlanorch.h"
+#include "routeorch.h"
```
prsunny, line 14: Please remove unused header add in this PR

orchagent/flex_counter/flowcounterrouteorch.cpp
Junchao-Mellanox: We have to save it here. For example:
1. Some misc route entries are created, and we don't save it here. In addition, those misc route entries are not in DB/memory.
2. User enables route flow counter and configures route pattern
3. flexcounterrouteorch starts to search marched route entries, but misc route entries were not saved anywhere, hence it can never be matched.

prsunny: Only case of misc route is IP2ME and Vnet routes. For this, I don't think we should cache the entire entries if the feature is never enabled. As mentioned, this is a debug feature.

orchagent/swssnet.h
prsunny: I don't see this addressed

Junchao-Mellanox: Fixed

orchagent/flex_counter/flowcounterrouteorch.cpp
Junchao-Mellanox: Hi Prince, we don't cache VNET routes in flowcounterrouteorch, we call it like this:

```cpp
gFlowCounterRouteOrch->onRemoveMiscRouteEntry(vr_id, ip_pfx, false);
```

The last argument "false" indicates that we don't cache the routes.

But we do cache IP2ME routes. Like I said before, we have to do it because IP2ME routes are not saved to any memory data structures. So route pattern can never match an IP2ME routes if those routes are created before enabling route flow counter (no place to search from). IMHO, we need to cache IP2ME routes or we have to claim that we don't support IP2ME route for this counter, however, it was defined in the HLD.

orchagent/flex_counter/flowcounterrouteorch.cpp
Junchao-Mellanox: We don't enable the timer here, we just create it, so it would not affect the performance if used does not enable route flow counter. We must create timer inside ctor or the timer will never run. Please check OrchDaemon::start(), it add all selectable at the begin of the function.
```cpp
void OrchDaemon::start()
{
    SWSS_LOG_ENTER();

    for (Orch *o : m_orchList)
    {
        m_select->addSelectables(o->getSelectables());
    }
```

prsunny (approved):

tests/test_flex_counters.py
```diff
@@ -373,7 +454,244 @@ def test_remove_trap_group(self, dvs):
...
+        self.set_admin_status("Ethernet0", "up")
+        self.set_admin_status("Ethernet4", "up")
+        # set ip address and default route
+        dvs.servers[0].runcmd("ip address add 10.0.0.1/31 dev eth0")
+        dvs.servers[0].runcmd("ip route add default via 10.0.0.0")
+        dvs.servers[1].runcmd("ip address add 10.0.0.3/31 dev eth0")
+        dvs.servers[1].runcmd("ip route add default via 10.0.0.2")
+        dvs.servers[0].runcmd("ping -c 1 10.0.0.3")
+
+        dvs.runcmd("vtysh -c \"configure terminal\" -c \"ip route 1.1.1.0/24 10.0.0.1\"")
```
prsunny, line 479: Why do we have to change to vtysh? Is the previous approach to write to APP_DB not working?

Junchao-Mellanox: Yes, it does not work after some merge/rebase...I guess there is some new changes which introduce more strict check.

Junchao-Mellanox:
@prsunny could you please review and sign-off?

liat-grozovik:
as this was prev approved and the changes are due to recent master changes, I assumes we can move forward. @prsunny  any objection to merge it?

prsunny (approved):

PR #2096, opened 2022-01-04
Add initial value for weight in overlay nexthops

shi-su:
**What I did**
Add initial value for weight in overlay nexthops.

**Why I did it**
Without initializing the weight variable, it may be initialized as non-zero values. This PR aims to prevent it from happening.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/nexthopkey.h
+++ b/orchagent/nexthopkey.h
@@ -102,7 +102,7 @@ struct NextHopKey
         }
     }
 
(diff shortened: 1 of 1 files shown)


abdosi (approved):

liat-grozovik:
@shi-su should this PR goes to 202111? if so, can you please add the proper label?

PR #2106, opened 2022-01-07
[buffer orch][202012] Bugfix: Don't query counter SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES on a pool where it is not supported (#1857)

stephenxs:
**What I did**
Currently, SAI_BUFFER_POOL_STAT_WATERMARK_BYTES and SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES are queried for buffer pools.
However, the latter is not supported on all pools and all platforms, which will results in sairedis complaint
To avoid that, we need to test whether it is supported before putting it to FLEX_COUNTER table

It depends on #1987 to be cherry-picked to 202012.

**Why I did it**
[Bugfix] Don't query SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES if it is not supported by a pool

**How I verified it**
Run vstest and manually test.

**Details if related**

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -23,6 +23,16 @@ extern sai_object_id_t gSwitchId;
 
 static const vector<sai_buffer_pool_stat_t> bufferPoolWatermarkStatIds =
 {
+    SAI_BUFFER_POOL_STAT_WATERMARK_BYTES
+};
+
+static const vector<sai_buffer_pool_stat_t> bufferSharedHeadroomPoolWatermarkStatIds =
+{
+    SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES
+};
+
+static const vector<sai_buffer_pool_stat_t> bufferPoolAllWatermarkStatIds =
+{
     SAI_BUFFER_POOL_STAT_WATERMARK_BYTES,
     SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES
 };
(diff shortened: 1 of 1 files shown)


stephenxs:
Originally, this PR had been opened for eliminating the error message like `updateSupportedBufferPoolCounters: BUFFER_POOL_WATERMARK_STAT_COUNTER: counter SAI_BUFFER_POOL_STAT_XOFF_ROOM_WATERMARK_BYTES is not supported on buffer pool oid:0xf00000018, rv: SAI_STATUS_NOT_IMPLEMENTED` and merged to master. Later on, it was reverted due to issues Azure/sonic-sairedis#862 and Azure/sonic-buildimage#8722
Now the warm reboot issue has been resolved by #1987. However, we no longer need this PR for master since the error message has been eliminated in other ways (which can not be backported to 202012).
So we need to add this PR to 202012 only.

stephenxs:
Both ARM build failed due to the following error.
```
Generating script.
========================== Starting Command Output ===========================
/bin/bash --noprofile --norc /__w/_temp/12befde7-36ee-4d1a-abb4-fdbf4926cf20.sh
sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
##[error]Bash exited with code '1'.
Finishing: Install dependencies
```

liat-grozovik:
@neethajohn please review so we can have it on 202012.

neethajohn (approved):

PR #2110, opened 2022-01-12
Fix the unsafe usage of strncpy in portsorch.cpp

stephenxs:
**What I did**
Fix unsafe usage of `strncpy` in portsorch.cpp.
The following errors occurred when sonic-swss is being build inside slave docker:
```
portsorch.cpp: In member function 'bool PortsOrch::createVlanHostIntf(swss::Port&, std::__cxx11::string)':
portsorch.cpp:2180:12: error: 'char* strncpy(char*, const char*, size_t)' specified bound 32 equals destination size [-Werror=stringop-truncation]
     strncpy(attr.value.chardata, hostif_name.c_str(), sizeof(attr.value.chardata));//SAI_HOSTIF_NAME_SIZE);
     ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
make: *** [Makefile:942: orchagent-portsorch.o] Error 1
```

**Why I did it**
Originally, `strncpy` is used in the following way:
```
strncpy(attr.value.chardata, src_string, sizeof(attr.value.chardata));
```
where `attr.value.chardata` is a char array.

However, this is not safe in case `strlen(src_string)` >= `sizeof(attr.value.chardata)` because there will no space in `attr.value.chardata` to store the terminating character.
It will leave the string `attr.value.chardata` open, the receiver of attr cannot determine the end of the string and suffer buffer overflow.

According to SAI API definition, the actually length of `SAI_HOSTIF_ATTR_NAME` should be `SAI_HOSTIF_NAME_SIZE - 1` which is less than sizeof(attr.value.chardata)`. So a safe way to do it should be:
```
strncpy(attr.value.chardata, src_string, SAI_HOSTIF_NAME_SIZE);
attr.value.chardata[SAI_HOSTIF_NAME_SIZE - 1] = '\0'
```

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2174,8 +2174,7 @@ bool PortsOrch::createVlanHostIntf(Port& vl, string hostif_name)
     attr.id = SAI_HOSTIF_ATTR_NAME;
     if (hostif_name.length() >= SAI_HOSTIF_NAME_SIZE)
     {
-        SWSS_LOG_ERROR("Host interface name is too long");
-        return false;
+        SWSS_LOG_WARN("Host interface name %s is too long and will be truncated to %d bytes", hostif_name.c_str(), SAI_HOSTIF_NAME_SIZE - 1);
     }
     strncpy(attr.value.chardata, hostif_name.c_str(), SAI_HOSTIF_NAME_SIZE);
     attr.value.chardata[SAI_HOSTIF_NAME_SIZE - 1] = '\0';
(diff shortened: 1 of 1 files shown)


orchagent/portsorch.cpp
```diff
@@ -2172,7 +2172,13 @@ bool PortsOrch::createVlanHostIntf(Port& vl, string hostif_name)
     attrs.push_back(attr);
 
     attr.id = SAI_HOSTIF_ATTR_NAME;
-    strncpy(attr.value.chardata, hostif_name.c_str(), sizeof(attr.value.chardata));
+    if (hostif_name.length() >= SAI_HOSTIF_NAME_SIZE)
```
liat-grozovik, line 2175: @stephenxs I wonder why not to raise a warning and still set it. it indeed will not be as the user requested but much closer that it was before.

stephenxs: Hi @liat-grozovik 
According to the SAI interface, the maximum bytes in `SAI_HOSTIF_ATTR_NAME` is `SAI_HOSTIF_NAME_SIZE - 1`. So I'm afraid setting it to a string longer than the size can incur SAI error which eventually aborts orchagent.
```
    /**
     * @brief Name [char[SAI_HOSTIF_NAME_SIZE]]
     *
     * The maximum number of characters for the name is SAI_HOSTIF_NAME_SIZE - 1 since
     * it needs the terminating null byte ('\0') at the end.
     *
     * If Hostif is a generic netlink, this indicates the generic netlink family name.
     *
     * @type char
     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
     * @condition SAI_HOSTIF_ATTR_TYPE == SAI_HOSTIF_TYPE_NETDEV or SAI_HOSTIF_ATTR_TYPE == SAI_HOSTIF_TYPE_GENETLINK
     */
    SAI_HOSTIF_ATTR_NAME,
```
Meanwhile, the maximum size of net_device ifname is also 15 bytes. So this is just a protection and to eliminate compiling warnings. In most cases, it won't introduce a different behavior as it did.

stephenxs: > @stephenxs I wonder why not to raise a warning and still set it. it indeed will not be as the user requested but much closer that it was before.

Updated accordingly.

liat-grozovik (approved):

PR #2112, opened 2022-01-13
[mclagsyncd] fix out of order initialization

novikauanton:
**What I did**

Reorder class member variables.

**Why I did it**

I've encountered an issue with m_bufSize, it's initialization is random, some times it may be initializied with 0, some times with random value.

`m_bufSize` use `MSG_BATCH_SIZE` for initialization:

```
MclagLink::MclagLink(Select *select, int port) :
    MSG_BATCH_SIZE(256),
    m_bufSize(MCLAG_MAX_MSG_LEN * MSG_BATCH_SIZE),
    ...
```

but `MSG_BATCH_SIZE` declared after `m_bufSize`

```
class MclagLink : public Selectable {
  private:
    unsigned int m_bufSize;
    ...
    const int MSG_BATCH_SIZE;
    ...
```

so MSG_BATCH_SIZE will be initialized after m_bufSize, and m_bufSize will use MSG_BATCH_SIZE with uninitialized value.

**How I verified it**

Setup mclag topology, some times mclagsyncd will not be able to receive data after startup.

**Details if related**

[More about this issue in cpp reference](https://en.cppreference.com/w/cpp/language/constructor#Initialization_order)

IMHO it must be treated as compilation error, so it will be better to use`-Werror=reorder` parameter for compiler to avoid such issues.

### DIFF
--- a/mclagsyncd/mclaglink.h
+++ b/mclagsyncd/mclaglink.h
@@ -187,9 +187,10 @@ namespace swss {
     typedef  std::tuple<std::string, std::string> vlan_mbr;
 
     class MclagLink : public Selectable {
-        
+
         private:
-            Select *m_select;
+            const int MSG_BATCH_SIZE;
+
             unsigned int m_bufSize;
             char *m_messageBuffer;
             char *m_messageBuffer_send;
@@ -200,11 +201,12 @@ namespace swss {
             int m_server_socket;
             int m_connection_socket;
 
+            Select *m_select;
+
             bool is_iccp_up = false;
             std::string m_system_mac;
             std::set<vlan_mbr> m_vlan_mbrship; //set of vlan,mbr tuples
 
-            const int MSG_BATCH_SIZE;
             std::map<std::string, std:: string> *p_learn;
 
             unique_ptr<DBConnector> p_state_db;


novikauanton:
@gechiang could you please review?

msosyak (approved):

msosyak:
@prsunny  Could we merge this?

gechiang:
Adding @Praveen-Brcm for visibility and the compilation enabling question.
The changes look good to me.  Once Praveen approves it will merge it.
Thanks!

gechiang (approved):
LGTM. Let's wait for @Praveen-Brcm approval before I merge this in.

msosyak:
@Praveen-Brcm Have a look, please.

akokhan (approved):

Praveen-Brcm:
@gechiang , @akokhan @msosyak : sorry to respond late on this. I missed on it.
The changes looks good. Would like to understand if the issue is seen consistently.? We have not see such issue in Broadcom testing.

novikauanton:
@Praveen-Brcm it is undefined behavior, so the issue is not consistent, it is random. The issue could be easily see only when `MSG_BATCH_SIZE = 0` before initialization (depends on previous data on stack). On my setup the issue seen only during first start of mclagsyncd on specific hardware.

PR #2113, opened 2022-01-14
[DualToR] Handle race condition between tunnel_decap and mux orchestr…

devpatha:
**What I did**
Change orch list to make sure mux_orch and mux_cb_orch are placed after tunnel_decap_orch

**Why I did it**
handlePeerSwitch returns false because mux tunnel not yet created, and that results in create_nh_tunnel failure.
This moves ports into error state.

**How I verified it**
Verified that after changing the order ports don't move to error state.

**Details if related**

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -1219,8 +1219,6 @@ bool MuxOrch::handlePeerSwitch(const Request& request)
 
     if (op == SET_COMMAND)
     {
-        mux_peer_switch_ = peer_ip;
-
         // Create P2P tunnel when peer_ip is available.
         IpAddresses dst_ips = decap_orch_->getDstIpAddresses(MUX_TUNNEL);
         if (!dst_ips.getSize())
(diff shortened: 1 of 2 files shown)


prsunny (approved):

prsunny:
@devpatha , it seems the test is consistently failing.. can you check if the failure is due to the code change by any chance?

devpatha:
> @devpatha , it seems the test is consistently failing.. can you check if the failure is due to the code change by any chance?

Looked at the failure. It is in TestCopp.test_disabled_feature_always_enabled_trap.
It is not able to find LLDP trap. I don't see any intersection with changes in this PR.

```
            trap_found = False
            trap_type = traps_to_trap_type["lldp"]
            for key in trap_keys:
                (status, fvs) = self.trap_atbl.get(key)
                assert status == True
                for fv in fvs:
                    if fv[0] == "SAI_HOSTIF_TRAP_ATTR_TRAP_TYPE":
                        if fv[1] == trap_type:
                            trap_found = True
                if trap_found:
                    self.validate_trap_group(key,trap_group)
                    break
            assert trap_found == True

2022-01-27T04:49:08.7992702Z >           assert trap_found == True
2022-01-27T04:49:08.7992947Z E           assert False == True
2022-01-27T04:49:08.7994098Z E             -False
2022-01-27T04:49:08.7994317Z E             +True
2022-01-27T04:49:08.7994434Z 
```

prsunny:
@devpatha , seems like this got missed from merge to master. Can you please create a new PR?

devpatha:
> @devpatha , seems like this got missed from merge to master. Can you please create a new PR?

@prsunny created PR #2397

PR #2121, opened 2022-01-25
[swss/cfgmgr] teammgr configure lacp rate

globaltrouble:
#### What I did

Make lacp_rate configurable for portchannel.

```
	Option specifying the rate in which we'll ask our link partner
	to transmit LACPDU packets in 802.3ad mode.  Possible values
	are:

	slow
		Request partner to transmit LACPDUs every 30 seconds

	fast
		Request partner to transmit LACPDUs every 1 second

	The default is slow.
```

#### Why I did it

In case of slow lacp_rate configuration link down will be detected in 60-90 seconds, it may be to long (for example for MCLAG high availability), in case of using ` --fast-rate=true` link down will be detected in 2-3 seconds.

#### How I did it
   
* add optional argument to `config portchannel` command, default=slow for backward compatibility. https://github.com/Azure/sonic-utilities/pull/2036 
* parse argument in `teammgr` and forward it to `teamd` (this PR)
* update docs https://github.com/Azure/SONiC/pull/937

#### How to verify it

Confgiure bond on other side, then configure portchannel and sniff the traffic from it.

```
config portchannel add PortChannel0001  --fast-rate=true
config portchannel member add PortChannel0001 Ethernet0
config interface ip add PortChannel0001  192.168.1.2/24
tcpdump -ne
```


doc/Configuration.md
```diff
@@ -1139,21 +1139,25 @@ name as object key and member list as attribute.
 ```
 {
 "PORTCHANNEL": {
-        "PortChannel0003": {
```
novikauanton, line 1142: fix formatting, `"PortChannel0003"` contains all mandatory attrs, `"PortChannel0004"` additionally contains options attrs

novikauanton:
@stcheng  could you please review it?

cfgmgr/teammgr.cpp
```diff
@@ -553,6 +559,11 @@ task_process_status TeamMgr::addLag(const string &alias, int min_links, bool fal
         conf << ",\"fallback\":true";
     }
 
+    if (lacp_rate == "fast")
```
ArthiSivanantham, line 562: In "teamd.conf", (LACP RUNNER SPECIFIC OPTIONS), fast_rate is defined to be a boolean,

`runner.fast_rate (bool)
	      Option specifies the rate at which our link partner is asked to transmit LACPDU packets. If this is true then packets will  be  sent
	      once per second. Otherwise they will be sent every 30 seconds.`

https://www.unix.com/man-page/centos/5/teamd.conf/
http://manpages.ubuntu.com/manpages/xenial/man5/teamd.conf.5.html

While spawning the teamd process with it's config file, "slow|fast" should be converted to "false|true". Instead why can't we define "fast_rate/lacp_rate" to be a boolean in consistent with "teamd" configuration ?

venkatmahalingam: I agree with Arthi, please use fast_rate in the config-DB as well.

novikauanton: Answering your question:
* It's easier to read concreate status than to read a boolean flag and convert it to concreate status.
* Boolean flags are harder to maintain (update code base), in case new type need to be added - all code need to be change (boolean flags and conditions using them).
* If it's possible: modules should not depend on low level details. In this example for persion unfarmiliar with teamd it will be easier to find info only about lacp_rate, without dig in teamd.

Thank you for the review, I'll change it to be consistent with links you pointed.

@venkatmahalingam @ArthiSivanantham  could you please share you opinion, do we need to change `lacp_rate` param to `fast_rate` in [config tool too](https://github.com/Azure/sonic-utilities/pull/2036) ?

msosyak:
@ArthiSivanantham @venkatmahalingam @prsunny  Are there any other comments? If not, please, help to merge this and related PR
- https://github.com/Azure/SONiC/pull/937
- https://github.com/Azure/sonic-utilities/pull/2036

msosyak:
@ArthiSivanantham @venkatmahalingam @prsunny Could we merge this and related?

msosyak:
@venkatmahalingam @ArthiSivanantham do you have other comments? If not, please approve.

msosyak:
@qiluo-msft @prsunny @nazariig Please help review and merge

nazariig (changes requested):

cfgmgr/teammgr.cpp
nazariig: @venkatmahalingam / @ArthiSivanantham IMHO, the arguments provided by @globaltrouble seems to be reasonable. Do we have a strong opinion here? Although, the original approach is compliant with the `teamd.conf` but it is not user friendly.

prsunny:
> @qiluo-msft @prsunny @nazariig Please help review and merge

@msosyak , changes looks good to me. Please address the outstanding comments and also get the dependent PRs in sonic-utilities, HLD merged.

@nazariig , @venkatmahalingam , @ArthiSivanantham  to signoff

cfgmgr/teammgr.cpp
venkatmahalingam: We don't have any strong opinion @nazariig it's just a suggestion as well known teamd uses fast_rate to control the LACP rate and also we can avoid conversion (lacp_rate to fast_rate) in the back-end.

bratashX:
will be good to include these changes into 202111 branch

msosyak:
@nazariig Do you have other comments?

nazariig (approved):

msosyak:
@prsunny Could we merge this?

prsunny:
@judyjoseph , can you please review/signoff?

judyjoseph (approved):
LGTM, Can you also add the test results (the o/p of teamdctl state dump) in PR description with local/peer device having lacp rate configured as fast/slow rate respectively.

PR #2127, opened 2022-01-30
[macsecorch]: Fix deleting MACsec bug

Pterosaur:

**What I did**
Fix a segment fault in MACsec orchagent to record the next iterator before deleting the corresponding MACsec object.

**Why I did it**
The segment fault will happen if to delete the MACsec Port or MACsec SC that have MACsec SC(s) or MACsec SA(s), Because the original implementation used the for loop to iterate all MACsec sub objects, but the MACsec sub object will be delete on each iteration so that the removed iterator cannot go to the next one.

**How I verified it**
Use `appl_db_del(appl_db, "MACSEC_PORT_TABLE", ifname)` to delete a MACsec port that has MACsec SC, The segment fault shouldn't happen.
**Details if related**

### DIFF
--- a/orchagent/macsecorch.cpp
+++ b/orchagent/macsecorch.cpp
@@ -1280,7 +1280,7 @@ bool MACsecOrch::updateMACsecSCs(MACsecPort &macsec_port, std::function<bool(MAC
     auto sc = macsec_port.m_egress_scs.begin();
     while (sc != macsec_port.m_egress_scs.end())
     {
-        if (!action((sc++).second))
+        if (!action((sc++)->second))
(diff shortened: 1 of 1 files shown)


lguohan:
@Pterosaur , is the vstest failure related to this pr or not related, if it is not, can you ask @prsunny  to check?

Pterosaur:
> @Pterosaur , is the vstest failure related to this pr or not related, if it is not, can you ask @prsunny to check?

Sure, I have connected Prince.

judyjoseph (approved):

PR #2128, opened 2022-01-31
[test_nhg.py] Cleanup test resources at the end

abanu-ms:
**What I did**
Removed some leftover state from a previous test case that was sometimes messing up with a following test case.

**Why I did it**
VS test pipelines fail from time to time because of this.

**How I verified it**
Tried to repro the crash running the test file.

**Details if related**


abanu-ms:
@prsunny Do you know anything about "Path does not exist: /home/vsts/work/1/a/gcov_output" error? Seems to be blocking all PRs.

prsunny (approved):

PR #2130, opened 2022-02-02
Validate LAG has members before mirror session create

rupesh-k:
**What I did**
Check LAG is not empty before creating mirror session.

**Why I did it**
Fix issue https://github.com/Azure/sonic-buildimage/issues/7988#issue-930946691
**How I verified it**
Verified config with empty LAG.
**Details if related**

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -327,6 +327,7 @@ bool MirrorOrch::validateSrcPortList(const string& srcPortList)
             if (port.m_type == Port::LAG)
             {
                 vector<Port> portv;
+                int portCount = 0;
                 m_portsOrch->getLagMember(port, portv);
                 for (const auto p : portv)
                 {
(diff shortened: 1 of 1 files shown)


raphaelt-nvidia:
LGTM

orchagent/mirrororch.cpp
```diff
@@ -336,6 +337,13 @@ bool MirrorOrch::validateSrcPortList(const string& srcPortList)
                                   p.m_alias.c_str(), port.m_alias.c_str(), srcPortList.c_str());
                         return false;
                     }
+                    portCount++;
+                }
+                if(!portCount)
```
prsunny, line 342: Please have a space after` if`. Where do we ensure that when a port is added to portchannel, the mirror session is activated?

rupesh-k: Pushed patchset with  space change.
session is activated updateLagMember (during SUBJECT_TYPE_LAG_MEMBER_CHANGE notification)

prsunny (approved):

prsunny:
@bingwang-ms , please sign-off.

hari-selvam:
@rupesh-k : With this fix, mirror session is not programmed in hardware if the source interface is empty port-channel. Hence STATE_DB doesn't have this session config, mirror session status is shown as 'error' in 'show mirror_session' output.

But when I add LAG members into the port-channel, the mirror session should be programmed in hardware and its status should be moved from 'error' to 'active'. But this is not the case right now. May I know if this issue is addressed or not?

PR #2131, opened 2022-02-04
[Vstest] Check kernel logs

prsunny:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/.azure-pipelines/test-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/test-docker-sonic-vs-template.yml
@@ -22,7 +22,7 @@ jobs:
   timeoutInMinutes: ${{ parameters.timeout }}
 
   pool:
-    vmImage: 'ubuntu-20.04'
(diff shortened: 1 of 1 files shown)


.azure-pipelines/build_and_install_module.sh
```diff
@@ -66,6 +66,8 @@ function build_and_install_kmodule()
         cp drivers/net/team/*.ko $TEAM_DIR/
         modinfo $TEAM_DIR/team.ko
         depmod
+        echo 'dmesg'
+        dmesg
         modprobe team
```
saiarcot895, line 71: `set -e` is present at the top of the file, so it won't execute anything after this command fails. I think `modprobe team || true` immediately followed by `dmesg` would help.

prsunny: sure.. let me add that as well, meantime you can check if the current one has any clue

PR #2134, opened 2022-02-05
Create tunnel to peer switch when a packet to unresolved neighbor is received

sumukhatv:
**What I did**

**Why I did it**
To handle the case when a packet to an unresolved neighbor is received, the ToR will create a tunnel to the peer and forward the packet to the peer.
**How I verified it**
Write virtual switch tests
**Details if related**

### DIFF
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -733,7 +733,7 @@ void NeighOrch::doTask(Consumer &consumer)
                 if (mac_address == MacAddress("00:00:00:00:00:00"))
                 {
                     MuxOrch* mux_orch = gDirectory.get<MuxOrch*>();
-                    sai_object_id_t nexthop = mux_orch->createNextHopTunnel(MUX_TUNNEL, neighbor_entry.ip_address);
+                    mux_orch->createNextHopTunnel(MUX_TUNNEL, neighbor_entry.ip_address);
                 }
                 if (addNeighbor(neighbor_entry, mac_address))
                 {
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -22,6 +22,8 @@ extern Directory<Orch*> gDirectory;
 extern string gMySwitchType;
 extern int32_t gVoqMySwitchId;
 
+#define MUX_TUNNEL "MuxTunnel0"
+
 const int neighorch_pri = 30;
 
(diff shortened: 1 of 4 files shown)


orchagent/neighorch.cpp
```diff
@@ -728,6 +730,11 @@ void NeighOrch::doTask(Consumer &consumer)
             if (m_syncdNeighbors.find(neighbor_entry) == m_syncdNeighbors.end()
                     || m_syncdNeighbors[neighbor_entry].mac != mac_address)
             {
+                if (mac_address == MacAddress("00:00:00:00:00:00"))
```
prsunny, line 733: Looks like there is more change here and the muxorch changes.

tests/test_mux.py
```diff
@@ -664,6 +665,45 @@ def remove_and_test_tunnel(self, db, asicdb, tunnel_name):
...
+        for r in routes:
+            t = json.loads(r)        
+            if t["dest"] == "192.168.0.99/32":
+                route = r
+        assert json.loads(route)["dest"] == "192.168.0.99/32"
+
+        fvs1 = asicdb.get_entry(self.ASIC_ROUTE_TABLE, route)
+        oid = fvs1["SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID"]
+        fvs2 = asicdb.get_entry(self.ASIC_NEXTHOP_TABLE, oid)
+        assert fvs2["SAI_NEXT_HOP_ATTR_IP"] == "10.1.0.32"
```
prsunny, line 692: please use function `dvs_route.check_asicdb_route_entries` for checking asic_db

tests/test_mux.py
```diff
@@ -664,6 +665,45 @@ def remove_and_test_tunnel(self, db, asicdb, tunnel_name):
...
 
 
+    def remove_link_and_test_tunnel_create(self, appdb, asicdb, confdb, dvs):
+        self.create_vlan_interface(confdb, asicdb, dvs)
+        self.create_mux_cable(confdb)
+        self.create_and_test_tunnel(appdb, asicdb, tunnel_name="MuxTunnel0", tunnel_type="IPINIP",
+                                   dst_ip="10.1.0.32", dscp_mode="uniform",
+                                   ecn_mode="standard", ttl_mode="pipe")
+        
+        peer_attrs = {
```
prsunny, line 675: Please add the following tests as well:
1. Add a FAILED neighbor before configuring peer_switch. There should not be a tunnel route
2. Do mux status changes after resolving the entry and verify neighbor, and tunnel route
3. Add a FAILED entry and delete the entry. Make sure the tunnel route is removed as well.

PR #2142, opened 2022-02-09
[ci]: Cleanup the workspace

xumia:
**What I did**
Cleanup the workspace.
**Why I did it**
In some of the test cases, it is run as root, left some of the file cannot be removed during the checkout step.

Example: https://dev.azure.com/mssonic/build/_build/results?buildId=71009&view=logs&j=3f6395b2-1619-5ebe-f305-2aedcf353cb5&t=aa8a4b4e-56c0-53e8-6540-01f05ca983a1
```
$ ls -l /agent/_work/1/s/tests/.pytest_cache/CACHEDIR.TAG
-rw-r--r-- 1 root root 194 Feb  8 13:28 /agent/_work/1/s/tests/.pytest_cache/CACHEDIR.TAG

$ ls -l /agent/_work/1/s/ -d
drwxr-xr-x 25 AzDevOps AzDevOps 4096 Feb  8 17:11 /agent/_work/1/s/
```

**How I verified it**

**Details if related**

### DIFF
--- a/.azure-pipelines/test-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/test-docker-sonic-vs-template.yml
@@ -25,7 +25,7 @@ jobs:
 
   steps:
   - script: |
-      sudo rm -rf $(ls -A1)
+      ls -A1 | xargs -I{} sudo rm -rf {}
     displayName: "Clean workspace"
   - checkout: self
   - task: DownloadPipelineArtifact@2
--- a/.azure-pipelines/test-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/test-docker-sonic-vs-template.yml
@@ -25,8 +25,8 @@ jobs:
 
   steps:
   - script: |
-      sudo rm -rf s
-    displayName: "Init source folder"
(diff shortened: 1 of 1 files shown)


prsunny (dismissed):

.azure-pipelines/test-docker-sonic-vs-template.yml
```diff
@@ -24,6 +24,10 @@ jobs:
   pool: sonic-common
 
   steps:
+  - script: |
+      sudo rm -rf $(ls -A1)
```
qiluo-msft, line 28: >sudo rm -rf $(ls -A1) [](http://example.com/codeflow?start=6&length=21)

This combination has risk. For example there is a filename as 'a b', and `rm` will try `rm a b` which is two files.

And `ls` could output colors in ANSI sequence.

xumia: @qiluo-msft , the command is running before any script start, so it is to make the whole workspace clean. It is expected, there are no files as the agent hosted by Microsoft.

qiluo-msft (approved):

PR #2145, opened 2022-02-11
[VS] Increase test timout to 360min

prsunny:
**What I did**
Changing the docker pool, VS test is taking slightly more than 4hrs. 

**Why I did it**

**How I verified it**

**Details if related**


theasianpianist (approved):

qiluo-msft:
Do we need it on old branches?

prsunny:
> Do we need it on old branches?

Not very sure, will monitor on the failure and if it happens on timeout, we may want to increase

PR #2147, opened 2022-02-16
[CoPP] Add always_enabled field to coppmgr logic

noaOrMlnx:
- What I did
Add the new design for copp manager - always_enabled field in copp_cfg.json file.

- Why I did it
A change was done for traps needs to be installed but doesn't have feature (arp, udld, lacp, ip2me),
in the new implementation, coppmgr will not automatically install traps which has no entry in features table, but will check first if the trap has "always_enabled":"true" field.

- How I verified it
run tests for making sure traps are installed and not when expecting them to.

- Details if related
Related to sonic-buildimage change - Azure/sonic-buildimage#9302
SONiC mgmt test plan can be found in Azure/SONiC#903.

This commit is a cherry-pick of https://github.com/Azure/sonic-swss/pull/2034

### DIFF
--- a/cfgmgr/coppmgr.cpp
+++ b/cfgmgr/coppmgr.cpp
@@ -78,31 +78,42 @@ bool CoppMgr::checkTrapGroupPending(string trap_group_name)
 /* Feature name and CoPP Trap table name must match */
 void CoppMgr::setFeatureTrapIdsStatus(string feature, bool enable)
 {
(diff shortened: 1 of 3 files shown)


prsunny:
Isn't this dependent on https://github.com/Azure/sonic-buildimage/pull/9999 ?

prsunny (approved):

prsunny:
@noaOrMlnx , could you please raise a submodule update?

PR #2149, opened 2022-02-18
portsorch: initial support for link-training and remote advertisement

ds952811:
Signed-off-by: Dante Su <dante.su@broadcom.com>

**What I did**
1. Add Link-Training support to portsorch, while Gearbox is not in the scope
2. Add support for remote speed advertisement
3. Add support for capability checker

**Why I did it**
1. In the case of DAC, static pre-calibrated pre-emphasis is rarely available on SONIC, as most of the ODM are expecting this to be  done dynamically at runtime via link-training, hence we'll need this feature to improve the link quality
2. Help users identify the cause of autoneg link failure
3. Avoid un-necessary SAI calls for autoneg and link-training controls if this feature is not supported

**How I verified it**

1. Manual test
2. Ran the Unit-tests to the corresponding changes

**Details if related**

HLD: Azure/SONiC#924,  Azure/SONiC#925

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -61,6 +61,8 @@ extern string gMyAsicName;
 #define DEFAULT_VLAN_ID     1
 #define MAX_VALID_VLAN_ID   4094
 
+#define PORT_SPEED_LIST_DEFAULT_SIZE                     16
+#define PORT_STATE_POLLING_SEC                            5
 #define PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS     1000
 #define PORT_BUFFER_DROP_STAT_POLLING_INTERVAL_MS     60000
 #define QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS   10000
@@ -110,6 +112,26 @@ static map<string, int> autoneg_mode_map =
 {
     { "on", 1 },
     { "off", 0 }
+};
+
+static map<string, int> link_training_mode_map =
+{
+    { "on", 1 },
+    { "off", 0 }
+};
+
+static map<sai_port_link_training_failure_status_t, string> link_training_failure_map =
+{
+    { SAI_PORT_LINK_TRAINING_FAILURE_STATUS_NO_ERROR, "none" },
+    { SAI_PORT_LINK_TRAINING_FAILURE_STATUS_FRAME_LOCK_ERROR, "frame_lock"},
+    { SAI_PORT_LINK_TRAINING_FAILURE_STATUS_SNR_LOWER_THRESHOLD, "snr_low"},
+    { SAI_PORT_LINK_TRAINING_FAILURE_STATUS_TIME_OUT, "timeout"}
+};
+
+static map<sai_port_link_training_rx_status_t, string> link_training_rx_status_map =
+{
+    { SAI_PORT_LINK_TRAINING_RX_STATUS_NOT_TRAINED, "not_trained" },
+    { SAI_PORT_LINK_TRAINING_RX_STATUS_TRAINED, "trained"}
 };
 
 // Interface type map used for gearbox
@@ -327,7 +349,8 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
                 GBPORT_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ,
                 PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
         port_buffer_drop_stat_manager(PORT_BUFFER_DROP_STAT_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_BUFFER_DROP_STAT_POLLING_INTERVAL_MS, false),
-        queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false)
+        queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
+        m_port_state_poller(new SelectableTimer(timespec { .tv_sec = PORT_STATE_POLLING_SEC, .tv_nsec = 0 }))
 {
     SWSS_LOG_ENTER();
 
@@ -592,6 +615,9 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
 
         m_lagIdAllocator = unique_ptr<LagIdAllocator> (new LagIdAllocator(chassisAppDb));
     }
+
+    auto executor = new ExecutableTimer(m_port_state_poller, this, "PORT_STATE_POLLER");
+    Orch::addExecutor(executor);
 }
 
 void PortsOrch::removeDefaultVlanMembers()
@@ -1931,6 +1957,35 @@ void PortsOrch::initPortSupportedSpeeds(const std::string& alias, sai_object_id_
     m_portStateTable.set(alias, v);
 }
 
+void PortsOrch::initPortCapAutoNeg(Port &port)
+{
+    sai_status_t status;
+    sai_attribute_t attr;
+
+    attr.id = SAI_PORT_ATTR_SUPPORTED_AUTO_NEG_MODE;
+    status = sai_port_api->get_port_attribute(port.m_port_id, 1, &attr);
+    if (status == SAI_STATUS_SUCCESS)
+    {
+        port.m_cap_an = attr.value.booldata ? 1 : 0;
+    }
+    else
+    {
+        // To avoid breakage on the existing platforms, AN should be 1 by default
+        port.m_cap_an = 1;
+        SWSS_LOG_WARN("Unable to get %s AN support capability",
+                      port.m_alias.c_str());
+    }
+}
+
+void PortsOrch::initPortCapLinkTraining(Port &port)
+{
+    // TODO:
+    // Add SAI_PORT_ATTR_SUPPORTED_LINK_TRAINING_MODE query when it is
+    // available in the saiport.h of SAI.
+    port.m_cap_lt = 1;
+    SWSS_LOG_WARN("Unable to get %s LT support capability", port.m_alias.c_str());
+}
+
 /*
  * If Gearbox is enabled and this is a Gearbox port then set the attributes accordingly.
  */
@@ -2085,6 +2140,45 @@ bool PortsOrch::getPortSpeed(sai_object_id_t id, sai_uint32_t &speed)
     return true;
 }
 
+bool PortsOrch::getPortAdvSpeeds(const Port& port, bool remote, std::vector<sai_uint32_t>& speed_list)
+{
+    sai_object_id_t port_id = port.m_port_id;
+    sai_object_id_t line_port_id;
+    sai_attribute_t attr;
+    sai_status_t status;
+    std::vector<sai_uint32_t> speeds(PORT_SPEED_LIST_DEFAULT_SIZE);
+
+    attr.id = remote ? SAI_PORT_ATTR_REMOTE_ADVERTISED_SPEED : SAI_PORT_ATTR_ADVERTISED_SPEED;
+    attr.value.u32list.count = static_cast<uint32_t>(speeds.size());
+    attr.value.u32list.list = speeds.data();
+
+    if (getDestPortId(port_id, LINE_PORT_TYPE, line_port_id))
+    {
+        status = sai_port_api->get_port_attribute(line_port_id, 1, &attr);
+    }
+    else
+    {
+        status = sai_port_api->get_port_attribute(port_id, 1, &attr);
+    }
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_WARN("Unable to get advertised speed for %s", port.m_alias.c_str());
+        return false;
+    }
+    speeds.resize(attr.value.u32list.count);
+    speed_list.swap(speeds);
+    return true;
+}
+
+bool PortsOrch::getPortAdvSpeeds(const Port& port, bool remote, string& adv_speeds)
+{
+    std::vector<sai_uint32_t> speed_list;
+    bool rc = getPortAdvSpeeds(port, remote, speed_list);
+
+    adv_speeds = rc ? swss::join(',', speed_list.begin(), speed_list.end()) : "";
+    return rc;
+}
+
 task_process_status PortsOrch::setPortAdvSpeeds(sai_object_id_t port_id, std::vector<sai_uint32_t>& speed_list)
 {
     SWSS_LOG_ENTER();
@@ -2196,6 +2290,32 @@ task_process_status PortsOrch::setPortAutoNeg(sai_object_id_t id, int an)
         return handleSaiSetStatus(SAI_API_PORT, status);
     }
     SWSS_LOG_INFO("Set AutoNeg %u to port pid:%" PRIx64, attr.value.booldata, id);
+    return task_success;
+}
+
+task_process_status PortsOrch::setPortLinkTraining(const Port &port, bool state)
+{
+    SWSS_LOG_ENTER();
+
+    if (port.m_type != Port::PHY)
+    {
+        return task_failed;
+    }
+
+    sai_attribute_t attr;
+    attr.id = SAI_PORT_ATTR_LINK_TRAINING_ENABLE;
+    attr.value.booldata = state;
+
+    string op = state ? "on" : "off";
+    sai_status_t status = sai_port_api->set_port_attribute(port.m_port_id, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to set LT %s to port %s", op.c_str(), port.m_alias.c_str());
+        return handleSaiSetStatus(SAI_API_PORT, status);
+    }
+
+    SWSS_LOG_INFO("Set LT %s to port %s", op.c_str(), port.m_alias.c_str());
+
     return task_success;
 }
 
@@ -2706,7 +2826,9 @@ void PortsOrch::doPortTask(Consumer &consumer)
             uint32_t speed = 0;
             string learn_mode;
             string an_str;
+            string lt_str;
             int an = -1;
+            int lt = -1;
             int index = -1;
             string role;
             string adv_speeds_str;
@@ -2796,6 +2918,11 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 else if (fvField(i) == "adv_interface_types")
                 {
                     adv_interface_types_str = fvValue(i);
+                }
+                /* Set link training */
+                else if (fvField(i) == "link_training")
+                {
+                    lt_str = fvValue(i);
                 }
                 /* Set port serdes Pre-emphasis */
                 else if (fvField(i) == "preemphasis")
@@ -2964,6 +3091,18 @@ void PortsOrch::doPortTask(Consumer &consumer)
             {
                 if (!an_str.empty())
                 {
+                    if (p.m_cap_an < 0)
+                    {
+                        initPortCapAutoNeg(p);
+                        m_portList[alias] = p;
+                    }
+                    if (p.m_cap_an < 1)
+                    {
+                        SWSS_LOG_ERROR("%s: autoneg is not supported", p.m_alias.c_str());
+                        // Invalid auto negotiation mode configured, don't retry
+                        it = consumer.m_toSync.erase(it);
+                        continue;
+                    }
                     if (autoneg_mode_map.find(an_str) == autoneg_mode_map.end())
                     {
                         SWSS_LOG_ERROR("Failed to parse autoneg value: %s", an_str.c_str());
@@ -3006,6 +3145,62 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         SWSS_LOG_NOTICE("Set port %s AutoNeg from %d to %d", alias.c_str(), p.m_autoneg, an);
                         p.m_autoneg = an;
                         m_portList[alias] = p;
+                        m_portStateTable.hdel(p.m_alias, "rmt_adv_speeds");
+                        updatePortStatePoll(p, PORT_STATE_POLL_AN, (an > 0));
+                    }
+                }
+
+                if (!lt_str.empty() && (p.m_type == Port::PHY))
+                {
+                    if (link_training_mode_map.find(lt_str) == link_training_mode_map.end())
+                    {
+                        SWSS_LOG_ERROR("Failed to parse LT value: %s", lt_str.c_str());
+                        // Invalid link training mode configured, don't retry
+                        it = consumer.m_toSync.erase(it);
+                        continue;
+                    }
+
+                    lt = link_training_mode_map[lt_str];
+                    if (lt != p.m_link_training)
+                    {
+                        if (p.m_cap_lt < 0)
+                        {
+                            initPortCapLinkTraining(p);
+                            m_portList[alias] = p;
+                        }
+                        if (p.m_cap_lt < 1)
+                        {
+                            SWSS_LOG_WARN("%s: LT is not supported by the ASIC", alias.c_str());
+                            // Don't retry
+                            it = consumer.m_toSync.erase(it);
+                            continue;
+                        }
+
+                        auto status = setPortLinkTraining(p, lt > 0 ? true : false);
+                        if (status != task_success)
+                        {
+                            SWSS_LOG_ERROR("Failed to set port %s LT from %d to %d", alias.c_str(), p.m_link_training, lt);
+                            if (status == task_need_retry)
+                            {
+                                it++;
+                            }
+                            else
+                            {
+                                it = consumer.m_toSync.erase(it);
+                            }
+                            continue;
+                        }
+                        m_portStateTable.hset(alias, "link_training_status", lt_str);
+                        SWSS_LOG_NOTICE("Set port %s LT from %d to %d", alias.c_str(), p.m_link_training, lt);
+                        p.m_link_training = lt;
+                        m_portList[alias] = p;
+                        updatePortStatePoll(p, PORT_STATE_POLL_LT, (lt > 0));
+
+                        // Restore pre-emphasis when LT is transitioned from ON to OFF
+                        if ((p.m_link_training < 1) && (serdes_attr.size() == 0))
+                        {
+                            serdes_attr = p.m_preemphasis;
+                        }
                     }
                 }
 
@@ -3346,9 +3541,17 @@ void PortsOrch::doPortTask(Consumer &consumer)
 
                 if (serdes_attr.size() != 0)
                 {
-                    if (setPortSerdesAttribute(p.m_port_id, serdes_attr))
+                    if (p.m_link_training > 0)
                     {
-                        SWSS_LOG_NOTICE("Set port %s  preemphasis is success", alias.c_str());
+                        SWSS_LOG_NOTICE("Save port %s preemphasis for LT", alias.c_str());
+                        p.m_preemphasis = serdes_attr;
+                        m_portList[alias] = p;
+                    }
+                    else if (setPortSerdesAttribute(p.m_port_id, serdes_attr))
+                    {
+                        SWSS_LOG_NOTICE("Set port %s preemphasis is success", alias.c_str());
+                        p.m_preemphasis = serdes_attr;
+                        m_portList[alias] = p;
                     }
                     else
                     {
@@ -3356,7 +3559,6 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         it++;
                         continue;
                     }
-
                 }
 
                 /* Last step set port admin status */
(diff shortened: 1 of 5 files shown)


orchagent/portsorch.cpp
```diff
@@ -1860,6 +1887,24 @@ void PortsOrch::initPortSupportedSpeeds(const std::string& alias, sai_object_id_
     m_portStateTable.set(alias, v);
 }
 
+void PortsOrch::initPortCapabilityLinkTraining(const Port &port)
+{
+    sai_status_t status;
+    sai_attribute_t attr;
+
+    attr.id = SAI_PORT_ATTR_SUPPORTED_LINK_TRAINING_MODE;
+    status = sai_port_api->set_port_attribute(port.m_port_id, &attr);
```
dgsudharsan, line 1896: I don't get this logic. This is not the way to query and understand capabilities. We have separate API sai_query_attribute_capability. Please use that API.

orchagent/portsorch.cpp
```diff
@@ -2128,6 +2173,44 @@ task_process_status PortsOrch::setPortAutoNeg(sai_object_id_t id, int an)
     return task_success;
 }
 
+bool PortsOrch::getPortCapabilityLinkTraining(const Port& port)
+{
+    string supp;
+
+    if (m_stateLinkTrainingTable.hget(port.m_alias, "supported", supp))
```
dgsudharsan, line 2180: Please use internal Port data structure to cache the capability rather than writing to DB and reading from it within the same process

orchagent/portsorch.cpp
```diff
@@ -112,6 +113,26 @@ static map<string, int> autoneg_mode_map =
     { "off", 0 }
 };
 
+static map<string, int> link_training_mode_map =
+{
+    { "on", 1 },
```
dgsudharsan, line 118: Can't we use a bool instead of int? This will make code much simpler

orchagent/portsorch.cpp
```diff
@@ -2903,6 +2996,48 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+                            {
+                                it++;
+                            }
+                            else
+                            {
+                                it = consumer.m_toSync.erase(it);
+                            }
+                            continue;
+                        }
+                        m_stateLinkTrainingTable.hset(p.m_alias, "status", lt > 0 ? "on" : "off");
```
dgsudharsan, line 3034: Can't we use lt_str as such in status?

tests/test_port_lt.py
```diff
@@ -0,0 +1,139 @@
...
+
+        tbl = swsscommon.ProducerStateTable(db, "PORT_TABLE")
+        fvs = swsscommon.FieldValuePairs([("link_training","off")])
+        tbl.set("Ethernet0", fvs)
+
+        tbl = swsscommon.ProducerStateTable(db, "PORT_TABLE")
+        fvs = swsscommon.FieldValuePairs([("link_training","on")])
+        tbl.set("Ethernet4", fvs)
+
+        # validate if autoneg false is pushed to asic db when set first time
```
dgsudharsan, line 22: Typo in comments, should be link-training instead of autoneg

tests/test_port_lt.py
```diff
@@ -0,0 +1,139 @@
...
+        port_oid = adb.port_name_map["Ethernet0"]
+        expected_fields = {"SAI_PORT_ATTR_LINK_TRAINING_ENABLE":"false"}
+        adb.wait_for_field_match("ASIC_STATE:SAI_OBJECT_TYPE_PORT", port_oid, expected_fields)
+
+        # validate if autoneg true is pushed to asic db when set first time
```
dgsudharsan, line 27: Typo in comments, should be link-training instead of autoneg

dgsudharsan (changes requested):

orchagent/portsorch.cpp
```diff
@@ -296,9 +317,11 @@ static bool isValidPortTypeForLagMember(const Port& port)
 PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_with_pri_t> &tableNames, DBConnector *chassisAppDb) :
         Orch(db, tableNames),
         m_portStateTable(stateDb, STATE_PORT_TABLE_NAME),
+        m_stateLinkTrainingTable(stateDb, STATE_LINK_TRAINING_TABLE_NAME),
```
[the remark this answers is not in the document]
ds952811: It's now merged into 'link_training_status' of PORT_TABLE in STATE_DB

orchagent/portsorch.cpp
```diff
@@ -563,6 +586,10 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
 
         m_lagIdAllocator = unique_ptr<LagIdAllocator> (new LagIdAllocator(chassisAppDb));
     }
+
+    auto executor = new ExecutableTimer(m_timer, this, "PORT_STATUS_POLL");
+    Orch::addExecutor(executor);
+    m_timer->start();
```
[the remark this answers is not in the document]
ds952811: This will be migrated to use the implementation in the following PR, which is for AutoNeg, and it will only activate the poll when the port is admin up, link_training feature is available and is enabled.
i.e.
https://github.com/Azure/sonic-swss/pull/2215

orchagent/portsorch.cpp
ds952811: As discussed in https://github.com/Azure/sonic-swss/pull/2215, sai_query_attribute_capability does not support per-port basis capability query, while the AN/LT may only be available on certain ports on the switch ASIC

orchagent/portsorch.cpp
ds952811: This is now cached as member field of the internal Port data structure

orchagent/portsorch.cpp
ds952811: In order to distinguish 'NOT CONFIURED' and 'OFF', we need -1 for this, and this is the same approach for the AutoNeg in the current implementation of SONiC

orchagent/portsorch.cpp
ds952811: Thanks, will do.

tests/test_port_lt.py
ds952811: Thanks, will fix it

tests/test_port_lt.py
ds952811: Thanks, will fix it

dgsudharsan (changes requested):

orchagent/portsorch.cpp
```diff
@@ -2351,6 +2426,9 @@ bool PortsOrch::initPort(const string &alias, const string &role, const int inde
                 /* Create associated Gearbox lane mapping */
                 initGearboxPort(p);
 
+                /* Initialize port capabilities */
+                initPortCapLinkTraining(p);
```
dgsudharsan, line 2430: This function loops through all the ports and get capability. This would increase the port initialization time which is a bottleneck in fastboot scenarios. If the user has no LT configs this operation is not needed.
Instead init the capability only when the user configures LT and capability not in cache.

orchagent/portsorch.cpp
```diff
@@ -1860,6 +1886,28 @@ void PortsOrch::initPortSupportedSpeeds(const std::string& alias, sai_object_id_
     m_portStateTable.set(alias, v);
 }
 
+void PortsOrch::initPortCapLinkTraining(Port &port)
+{
+    sai_status_t status;
+    sai_attribute_t attr;
+
+#ifdef SAI_PORT_ATTR_SUPPORTED_LINK_TRAINING_MODE
```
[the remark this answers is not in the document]
ds952811: This is for the build failure in the PR and also the backward compatibility of the legacy SAI implementation, where SAI_PORT_ATTR_SUPPORTED_LINK_TRAINING_MODE is not implemented, while the other control attributes are supported

orchagent/portsorch.cpp
ds952811: Thanks, will fix it

orchagent/portsorch.cpp
ds952811: Noted, and it's done

orchagent/portsorch.cpp
```diff
@@ -61,6 +61,7 @@ extern string gMyAsicName;
 #define DEFAULT_VLAN_ID     1
 #define MAX_VALID_VLAN_ID   4094
 
+#define PORT_STAT_POLLING_SEC                             3
```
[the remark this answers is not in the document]
ds952811: Sure, it's done

orchagent/portsorch.cpp
Junchao-Mellanox: Still not fixed

orchagent/portsorch.cpp
```diff
@@ -1860,6 +1885,33 @@ void PortsOrch::initPortSupportedSpeeds(const std::string& alias, sai_object_id_
...
+    attr.id = SAI_PORT_ATTR_SUPPORTED_AUTO_NEG_MODE;
+#endif
+    status = sai_port_api->set_port_attribute(port.m_port_id, &attr);
+    if (status == SAI_STATUS_SUCCESS)
+    {
+        port.m_cap_lt = attr.value.booldata ? 1 : 0;
+    }
+    else
+    {
+        port.m_cap_lt = 1; /* Default to 1 for vstest */
```
Junchao-Mellanox, line 1910: I am not sure about this. It means that m_cap_lt will be 1 even if SAI return error. I would suggest to implement unit test if VS test does not support link training.

orchagent/portsorch.cpp
```diff
@@ -2930,6 +3016,60 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+                    if (lt != p.m_link_training)
+                    {
+                        if (p.m_cap_lt < 0)
+                        {
+                            initPortCapLinkTraining(p);
+                            m_portList[alias] = p;
+                        }
+                        if (p.m_cap_lt < 1)
+                        {
+                            SWSS_LOG_ERROR("%s: LT is not supported by the ASIC", alias.c_str());
```
Junchao-Mellanox, line 3039: should it be ERROR or WARNING?

orchagent/portsorch.cpp
```diff
@@ -5912,6 +6072,54 @@ bool PortsOrch::getPortOperSpeed(const Port& port, sai_uint32_t& speed) const
...
+    SWSS_LOG_ENTER();
+
+    if (port.m_type != Port::PHY)
+    {
+        return false;
+    }
+
+    sai_attribute_t attr;
+    attr.id = SAI_PORT_ATTR_LINK_TRAINING_RX_STATUS;
+    attr.value.u32 = 0;
```
Junchao-Mellanox, line 6086: no need?

orchagent/portsorch.cpp
```diff
@@ -5912,6 +6072,54 @@ bool PortsOrch::getPortOperSpeed(const Port& port, sai_uint32_t& speed) const
...
+    SWSS_LOG_ENTER();
+
+    if (port.m_type != Port::PHY)
+    {
+        return false;
+    }
+
+    sai_attribute_t attr;
+    attr.id = SAI_PORT_ATTR_LINK_TRAINING_FAILURE_STATUS;
+    attr.value.u32 = 0;
```
Junchao-Mellanox, line 6110: no need?

orchagent/portsorch.cpp
```diff
@@ -6883,3 +7091,87 @@ std::unordered_set<std::string> PortsOrch::generateCounterStats(const string& ty
...
+
+void PortsOrch::doTask(swss::SelectableTimer &timer)
+{
+    Port port;
+
+    for (auto it = m_port_state_poll.begin(); it != m_port_state_poll.end(); ++it)
+    {
+        if ((it->second == 0) || !getPort(it->first, port))
+        {
+            m_port_state_poll.erase(it);
```
Junchao-Mellanox, line 7161: it will invalid the iterator and cause unpredictable behavior

orchagent/portsorch.h
```diff
@@ -325,12 +330,26 @@ class PortsOrch : public Orch, public Subject
...
 
     bool getPortOperSpeed(const Port& port, sai_uint32_t& speed) const;
     void updateDbPortOperSpeed(Port &port, sai_uint32_t speed);
 
+    bool getPortLinkTrainingRxStatus(const Port &port, sai_port_link_training_rx_status_t &rx_status);
+    bool getPortLinkTrainingFailure(const Port &port, sai_port_link_training_failure_status_t &failure);
+
+    typedef enum {
+        PORT_STATE_POLL_NONE = 0,
+        PORT_STATE_POLL_AN   = 0x00000001, /* Auto Negotiation */
```
Junchao-Mellanox, line 345: never used?

orchagent/portsorch.cpp
```diff
@@ -6883,3 +7091,87 @@ std::unordered_set<std::string> PortsOrch::generateCounterStats(const string& ty
...
+        return;
+    }
+    if (active)
+    {
+        m_port_state_poll[port.m_alias] |= type;
+        m_timer->start();
+    }
+    else
+    {
+        m_port_state_poll[port.m_alias] &= ~type;
```
Junchao-Mellanox, line 7149: check if it should be remove from m_port_state_poll and stop timer if need

orchagent/portsorch.cpp
```diff
@@ -6883,3 +7091,87 @@ std::unordered_set<std::string> PortsOrch::generateCounterStats(const string& ty
...
+void PortsOrch::updatePortStatePoll(const Port &port, port_state_poll_t type, bool active)
+{
+    if (type == PORT_STATE_POLL_NONE)
+    {
+        return;
+    }
+    if (active)
+    {
+        m_port_state_poll[port.m_alias] |= type;
+        m_timer->start();
```
Junchao-Mellanox, line 7145: what if the timer is already started?

Junchao-Mellanox (changes requested):

orchagent/portsorch.h
ds952811: It's part of the review for AutoNeg, we're having 2 PRs in review at the same time, and they're using the same port state poller.

orchagent/portsorch.cpp
ds952811: Done

orchagent/portsorch.cpp
ds952811: Yes, this is redundant, will have it removed

orchagent/portsorch.cpp
```diff
@@ -2128,6 +2180,33 @@ task_process_status PortsOrch::setPortAutoNeg(sai_object_id_t id, int an)
...
+    string op = state ? "on" : "off";
+    sai_status_t status = sai_port_api->set_port_attribute(port.m_port_id, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to set LT %s to port pid:%" PRIx64,
+                       op.c_str(), port.m_port_id);
+        return handleSaiSetStatus(SAI_API_PORT, status);
+    }
+
+    SWSS_LOG_INFO("Set LT %s to port pid:%" PRIx64, op.c_str(), port.m_port_id);
```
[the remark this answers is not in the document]
ds952811: Done

orchagent/portsorch.cpp
```diff
@@ -2128,6 +2180,33 @@ task_process_status PortsOrch::setPortAutoNeg(sai_object_id_t id, int an)
...
+
+    sai_attribute_t attr;
+    attr.id = SAI_PORT_ATTR_LINK_TRAINING_ENABLE;
+    attr.value.booldata = state;
+
+    string op = state ? "on" : "off";
+    sai_status_t status = sai_port_api->set_port_attribute(port.m_port_id, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to set LT %s to port pid:%" PRIx64,
```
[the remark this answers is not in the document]
ds952811: Done

orchagent/portsorch.cpp
ds952811: This is an issue in VS, do you know where is the source of this SAI model? SAI/bm/p4-sai?
Could we this VS enhancement deal in a separate PR? We'll need to do this for both AutoNeg and LinkTraining

orchagent/portsorch.cpp
```diff
@@ -298,7 +319,8 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
         m_portStateTable(stateDb, STATE_PORT_TABLE_NAME),
         port_stat_manager(PORT_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
         port_buffer_drop_stat_manager(PORT_BUFFER_DROP_STAT_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_BUFFER_DROP_STAT_POLLING_INTERVAL_MS, false),
-        queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false)
+        queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
+        m_timer(new SelectableTimer(timespec { .tv_sec = PORT_STATE_POLLING_SEC, .tv_nsec = 0 }))
```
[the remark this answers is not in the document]
ds952811: Done

orchagent/portsorch.cpp
ds952811: Updated from ERROR to WARN

orchagent/portsorch.cpp
```diff
@@ -2927,6 +3055,49 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
                         p.m_autoneg = an;
                         m_portList[alias] = p;
+                        m_portStateTable.hdel(p.m_alias, "rmt_adv_speeds");
+                        updatePortStatePoll(p, PORT_STATE_POLL_AN, (an > 0));
+                    }
+                }
+
+                if (!lt_str.empty() && (p.m_type == Port::PHY))
+                {
+                    if (link_training_mode_map.find(lt_str) == link_training_mode_map.end())
```
dgsudharsan, line 3065: Similar to AN, I don't see a lt capability query here. Please add

orchagent/portsorch.cpp
```diff
@@ -5757,6 +5935,33 @@ void PortsOrch::updatePortOperStatus(Port &port, sai_port_oper_status_t status)
...
+            else
+            {
+                /* Restart autoneg state polling upon link down event */
+                m_portStateTable.hdel(port.m_alias, "rmt_adv_speeds");
+                updatePortStatePoll(port, PORT_STATE_POLL_AN, true);
+            }
+        }
+        if (port.m_admin_state_up && port.m_link_training > 0)
+        {
+            updatePortStateLinkTraining(port);
```
dgsudharsan, line 5954: Should this move inside the below if condition?

orchagent/portsorch.cpp
```diff
@@ -6883,3 +7132,115 @@ std::unordered_set<std::string> PortsOrch::generateCounterStats(const string& ty
     }
     return counter_stats;
 }
+
+void PortsOrch::updatePortStateAutoNeg(const Port &port)
```
dgsudharsan, line 7136: The function name is misleading. Should it be updatePortRemoveAdvSpeeds?

tests/test_port_an.py
```diff
@@ -293,6 +293,23 @@ def test_PortAutoNegWarm(self, dvs, testlog):
...
+        sdb = swsscommon.DBConnector(6, dvs.redis_sock, 0)
+
+        ctbl = swsscommon.Table(cdb, "PORT")
+        stbl = swsscommon.Table(sdb, "PORT_TABLE")
+
+        # set autoneg = true and admin_status = up
+        fvs = swsscommon.FieldValuePairs([("autoneg","on"),("admin_status","up")])
+        ctbl.set("Ethernet0", fvs)
+
+        time.sleep(10)
```
dgsudharsan, line 308: Why do we need 10 second sleep?

dgsudharsan (changes requested):

orchagent/portsorch.cpp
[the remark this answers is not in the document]
ds952811: Done

orchagent/portsorch.cpp
ds952811: This is intentionally removed at the earlier commit
1. LT is similar to AN, it's a port capability available on a per-port basis, not system-wide on a single switch ASIC
2. For a port capability available on a per-port basis, SAI_PORT_ATTR_SUPPORTED_XXXX should be introduced into saiport.h
3. As per the latest comments from Rita, the SAI spec. was finalized around March time frame, it's too late for 202205 right now
4. Since this is a port capability available on a per-port basis, using SAI_PORT_ATTR_LINK_TRAINING_ENABLE in sai_query_attribute_capability  is not going to work, and during the SAI subgroup meeting a few months back, there was a conclusion that per-port capability query should be tracked in a separated PR, as all the existing attributes should all be migrated, not just the new attribute for LT.

orchagent/portsorch.cpp
ds952811: We already have the if-statement for the port admin state and LT enablement, may I have more details of the additional checker that's missing? Thank you.

orchagent/portsorch.cpp
ds952811: This is a placeholder for the operational AN state, not specific to remote speed only, there are more upcoming in the PRs in the furture (e.g. tx/rx pause, FEC ....)

tests/test_port_an.py
ds952811: Yep, you're right now, I'll remove this, thank you!

orchagent/portsorch.cpp
ds952811: For your reference, the function prototype of sai_query_attribute_capability is as follows, it does not work on a per-port basis
https://github.com/opencomputeproject/SAI/blob/master/inc/saiobject.h#L221
```
sai_status_t sai_query_attribute_capability(
        _In_ sai_object_type_t object_type,
        _In_ sai_attr_id_t attr_id,
        _Out_ sai_attr_capability_t *attr_capability);
```

dgsudharsan: Can you please add a comment to make sure once SAI is there we revisit and add the check? It is important that we ensure it so that devices and specific ports where it is not supported will have check in place to avoid SAI failure thereby prevent orchagent crash

dgsudharsan (changes requested):

orchagent/portsorch.cpp
```diff
@@ -1860,6 +1886,24 @@ void PortsOrch::initPortSupportedSpeeds(const std::string& alias, sai_object_id_
     m_portStateTable.set(alias, v);
 }
 
+void PortsOrch::initPortCapAutoNeg(Port &port)
+{
+    sai_status_t status;
+    sai_attribute_t attr;
+
+    attr.id = SAI_PORT_ATTR_SUPPORTED_AUTO_NEG_MODE;
```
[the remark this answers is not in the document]
ds952811: Yes, it will cause breakage, hence the AN capability will be updated to use 1(supported) by default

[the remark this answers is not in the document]
dgsudharsan: Sure. Please use this capability to a warning log in case if SAI returns not supported for the attribute. SAI vendors can then mitigate to implement this attribute without the functional impact.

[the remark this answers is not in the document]
ds952811: Done

orchagent/portsorch.cpp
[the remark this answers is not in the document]
ds952811: Done

orchagent/portsorch.cpp
```diff
@@ -6883,3 +7132,115 @@ std::unordered_set<std::string> PortsOrch::generateCounterStats(const string& ty
...
+        m_portStateTable.hset(port.m_alias, "rmt_adv_speeds", adv_speeds);
+    }
+    else
+    {
+        m_portStateTable.hset(port.m_alias, "rmt_adv_speeds", "N/A");
+        updatePortStatePoll(port, PORT_STATE_POLL_AN, false);
+    }
+}
+
+void PortsOrch::updatePortStateLinkTraining(const Port &port)
```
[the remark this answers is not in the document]
ds952811: Done

orchagent/portsorch.cpp
```diff
@@ -2927,6 +3055,49 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+                        SWSS_LOG_ERROR("Failed to parse LT value: %s", lt_str.c_str());
+                        // Invalid link training mode configured, don't retry
+                        it = consumer.m_toSync.erase(it);
+                        continue;
+                    }
+
+                    lt = link_training_mode_map[lt_str];
+                    if (lt != p.m_link_training)
+                    {
+                        auto status = setPortLinkTraining(p, lt > 0 ? true : false);
```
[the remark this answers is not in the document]
ds952811: The LT could be enabled/disable regardless of the port admin state, it's a set of control registers on the SerDes

orchagent/portsorch.cpp
```diff
@@ -2128,6 +2211,32 @@ task_process_status PortsOrch::setPortAutoNeg(sai_object_id_t id, int an)
     return task_success;
 }
 
+task_process_status PortsOrch::setPortLinkTraining(const Port &port, bool state)
```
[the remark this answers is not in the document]
ds952811: The LT capability checker is now restored, such a request will be blocked in doPortTask()

orchagent/portsorch.cpp
ds952811: In the case of AN
- Link Up: Refresh AN states
- Link Down: Clear the remote advertisement
In the case of LT
- Link Up: Refresh LT states
- Link Down: Refresh LT states

Hence the updatePortStateLinkTraining() is invoked before checking the operation link status

orchagent/portsorch.cpp
ds952811: Done

orchagent/portsorch.cpp
```diff
@@ -1944,8 +1944,37 @@ void PortsOrch::initPortCapAutoNeg(Port &port)
     }
     else
     {
-        port.m_cap_an = 0;
-        SWSS_LOG_NOTICE("Unable to get %s AN capability", port.m_alias.c_str());
+        // To avoid breakage on the existing platforms, AN should be 1 by default
+        port.m_cap_an = 1;
+        SWSS_LOG_WARN("Unable to get %s AN capability, assumming it's supported",
```
dgsudharsan, line 1949: I prefer you just add "Unable to get AN support capability" in the log. Please remove "assuming it's supported"

dgsudharsan (changes requested):

orchagent/portsorch.cpp
ds952811: Done

orchagent/portsorch.cpp
```diff
@@ -1944,8 +1944,37 @@ void PortsOrch::initPortCapAutoNeg(Port &port)
...
+    sai_status_t status;
+    sai_attribute_t attr;
+
+    // TODO: Use SAI_PORT_ATTR_SUPPORTED_LINK_TRAINING_MODE for the query
+    //
+    // While SAI_PORT_ATTR_SUPPORTED_LINK_TRAINING_MODE is available at SAI master,
+    // it's not available in SAI v1.10 paried with SONiC.202205.
+    // And given that LT is part of AN, it should be okay to use
+    // SAI_PORT_ATTR_SUPPORTED_AUTO_NEG_MODE as a fallback plan.
+    attr.id = SAI_PORT_ATTR_SUPPORTED_AUTO_NEG_MODE;
```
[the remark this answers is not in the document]
ds952811: Done

orchagent/portsorch.cpp
```diff
@@ -7231,17 +7212,18 @@ void PortsOrch::refreshPortStateAutoNeg(const Port &port)
...
 
-    string adv_speeds;
+    string adv_speeds = "N/A";
 
-    if (getPortAdvSpeeds(port, true, adv_speeds))
-    {
-        m_portStateTable.hset(port.m_alias, "rmt_adv_speeds", adv_speeds);
-    }
-    else
+    if (port.m_admin_state_up)
```
dgsudharsan, line 7217: Previously the get remote adv speeds is done even when admin state is down and this new logic will restrict it to only when admin up. 
This means polling becomes irrelevant. Can you explain the change for the reason?

ds952811: When the port is admin down, the SerDes is stopped, and it's impossible to get a correct remote advertisement, hence this redundant operation should be skipped if the port is disabled.

prgeor:
"/azp run"

orchagent/portsorch.cpp
dgsudharsan: Then why do we need the poller? Can't we just update the remote advertised speeds on link up? The entire polling logic was introduced to get remote advertisement when link was down. Should that be removed?

ds952811: Sorry for the confusion, but that's a different story

1. MAC disabled + link down
This is a case we have port.m_admin_state_up == false, while the link is surely down, and the remote advertisement is not available since the MAC/SerDes is disabled

2. MAC enabled + link down
This is a case we have port.m_admin_state_up == true, while the link is operationally down, and the remote advertisement could be available as the MAC/SerDes is enabled, frame exchanges are happening on the PHY layer (i.e. It's the internal PHY, in this case. Precisely speaking, it's the TSC drivers on the Broadcom chips)

3. MAC enabled + link up
This is a general case of port.m_admin_state_up == true and link is  operationally up.

Conclusion:
While the remote advertisement could be available in case 2 and 3, it's not available in case 1, hence the poller should only fetching the status when MAC/SerDes is enabled.

Junchao-Mellanox (dismissed):

dgsudharsan (approved):

PR #2150, opened 2022-02-18
[vxlan] Remove tunnel map objects on VNET tunnel removal

volodymyrsamotiy:

**What I did**

* VxLAN tunnel map objects were not removed on VNET tunnel removal.
* It caused config lefovers which is not expected.
* If tunnel is removed then all related objects should be removed as well.
* To do so just used deleteTunnelHw() function which does proper remove.

**Why I did it**

To cleanup config leftovers after VNET tunnel removal.

**How I verified it**
* Configure 2 VNETs and then remove them one by one.
* Verify that after 1st VNET removed tunnel and all related objects still exist.
* Verify that after 2nd (last) VNET removed tunnel and all related objects are removed.

**Details if related**
N/A

### DIFF
--- a/orchagent/vxlanorch.cpp
+++ b/orchagent/vxlanorch.cpp
@@ -1542,28 +1542,12 @@ bool VxlanTunnelOrch::removeVxlanTunnelMap(string tunnelName, uint32_t vni)
     tunnel_obj->vlan_vrf_vni_count--;
     if (tunnel_obj->vlan_vrf_vni_count == 0)
     {
-        auto tunnel_term_id = vxlan_tunnel_table_[tunnelName].get()->getTunnelTermId();
-        try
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@volodymyrsamotiy  please add VS test to automate the way you verified it.

prsunny:
@volodymyrsamotiy , is this relevant to 201911?

volodymyrsamotiy:
@prsunny 
* I Will add VS test
* For the 201911, I am not sure but I will check

liat-grozovik:
@prsunny VS tests added. Please review

prsunny (approved):

prsunny:
@srj102 , could you please review/sign-off?

PR #2155, opened 2022-02-22
[aclorch] Fix and simplify DTel watchlist tables and entries

mickeyspiegel:
**What I did**

- Fixed the logic that determines which subclass of AclRule to create,
  by referencing the correct set of enum mappings for DTel actions.
- Removed the DTel drop watchlist table and entries, leaving only the
  DTel flow watchlist table and entries, which support the superset of
  actions for both flow and drop.
- Renamed AclRuleDTelFlowWatchListEntry to AclRuleDTelWatchListEntry
  (without ...Flow...) to clarify that there is only one common watchlist
  for DTel, covering both flow and drop actions.
  Note: Left definition of TABLE_TYPE_DTEL_FLOW_WATCHLIST unchanged
        for backwards compatibility.
- Updated the list of actions in the DTel Flow watchlist table so that
  it matches the existing allowed actions in DTel AclRule subclasses.

**Why I did it**

The significant rewrite of aclorch when adding ACL_TABLE_TYPE
configuration caused two major issues that prevent the DTel
watchlist from operating correctly:

1. Introduced a bug that prevents configuration of any DTel rules.
   This is due to a reference to an incorrect set of enum mappings,
   while determining which subclass of AclRule to create.

2. Changed the way that the AclRule subclass is determined so that it
   is no longer tied to ACL_TABLE_TYPE. The new logic assumes
   that each subclass of AclRule has a disjoint set of actions that do not
   overlap any of the actions in other AclRule subclasses. This assumption
   does not hold for DTel which currently has two AclRule subclasses,
   for flow and drop.

   Since the 201911 release, the DTel drop actions have been
   supported in the DTel flow watchlist as well as the DTel drop
   watchlist. This allows for simplified operation where there is a
   single watchlist (DTel flow watchlist) that contains both flow and
   drop actions. The ACL_TABLE_TYPE changes effectively reverted
   this change. A fix to this logic would remove the distinction
   between AclRules of subclass DTelFlow and DTelDrop, allowing for the
   larger set of actions supported by DTelFlow in either table.

   With this change in behavior, this seems like a good time to remove
   the DTel drop watchlist, leaving only one DTel watchlist that supports
   the superset of actions for both flow and drop.

**How I verified it**

Manual testing, verifying that syncd sends down the corresponding
acl entries, and that these rules are programmed in the switch hardware.

We also added sonic-swss tests for DTel acls that verify the SAI calls
generated for DTel acl entries.

**Details if related**

Signed-off-by: mickeyspiegel <mickey.spiegel@intel.com>


prsunny:
Changes lgtm, @bingwang-ms to review/merge

sunesh:
@bingwang-ms could u please help merge this PR

TomFletcher0:
@bingwang-ms can this be merged, or are we waiting for something?

mickeyspiegel:
@bingwang-ms Test failures between successive runs are disjoint. It does not seem like any of these failures are related to this patch.

sunesh:
@prsunny , @bingwang-ms can this PR be merged? Is there any action required from @mickeyspiegel

mickeyspiegel:
@prsunny I updated to latest master, all checks are passing now.

prsunny (approved):

PR #2162, opened 2022-02-28
Add mandatory action list to mirror ACL table

ysmanman:
**What I did**
Add mandatory action list to mirror ACL table, which is required by SAI.

**Why I did it**

**How I verified it**

**Details if related**


ysmanman:
This PR is replaced by https://github.com/Azure/sonic-swss/pull/2205.

PR #2170, opened 2022-03-07
Use abort instead of exit in case calling SAI API failure

stephenxs:
**What I did**
Use `abort` instead of `exit` in case calling SAI API failure
Currently, `exit` is used in functions `handleSai{Create,Remove,Set,Get}Status` on SAI failure while `abort` is used in some other functions on SAI failure.
IMO using `abort` yields benefits:
- consistent behavior in orchagent on SAI failure
- in mocked test, in case SAI failure occurs the test process will just exit silently, which makes it very difficult for developers to find out what happened. by using `abort`, a coredump will be generated, which helps developers nail down the issue.

So we would like to use `abort` in all cases of receiving SAI failure.

**Why I did it**

**How I verified it**

Manually test.

**Details if related**


prsunny (approved):

stephenxs:
it can be cherry-picked to 202111 smoothly (tested on b3e5e33f0007ea1c8716b215ac50f78ac6cdb313)

PR #2171, opened 2022-03-07
Update orchagent to support new field `pfcwd_sw_enable`

bingwang-ms:
Signed-off-by: bingwang <bingwang@microsoft.com>

<!--
Please make sure you have read and understood the contribution guildlines:
https://github.com/Azure/SONiC/blob/gh-pages/CONTRIBUTING.md

1. Make sure your commit includes a signature generted with `git commit -s`
2. Make sure your commit title follows the correct format: [component]: description
3. Make sure your commit message contains enough details about the change and related tests
4. Make sure your pull request adds related reviewers, asignees, labels

Please also provide the following information in this pull request:
-->

**What I did**
Currently, the entry `pfc_enable` in table `PORT_QOS_MAP` is used to specify pfc and pfc_watchdog are enabled on which queues.
To avoid PFC deadlock in Dual-ToR scrnario, we are going to introduce two extra lossless queues to carry bounced back traffic.[HLD](https://github.com/Azure/SONiC/pull/950).
The extra lossless queues require another two pfc_watchdogs, and the new watchdogs will be implemented by hardware due to limited resources. The hardware pfc watchdog is not covered in this PR.
To specify on which queue to enable  pfc watchdog, we need to define new table `pfcwd_sw_enable`.


|Table | Description|
|-- | --|
|pfc_enable | Specify on which queues to enable PFC|
|pfcwd_sw_enable | Specify on which queues to enable software PFC watchdog|


This PR is to update orchagent to support new field `pfcwd_sw_enable` .

As two extra lossless PGs (2 and 6) are to be added, `buffermgrd` is also updated in this PR to generate lossless profile for the new PGs.

**Why I did it**
Update orchagent to support new field `pfcwd_sw_enable` .

**How I verified it**
Verified by UT. 
```
sudo pytest3 --dvsname=vs tests/test_pfcwd.py::TestPfcwdFunc -v --pdb
========================================================================================= test session starts =========================================================================================
platform linux -- Python 3.6.9, pytest-7.0.1, pluggy-1.0.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/bingwang/work/sonic/sonic-buildimage-master/src/sonic-swss
plugins: flaky-3.7.0
collected 2 items                                                                                                                                                                                     

tests/test_pfcwd.py::TestPfcwdFunc::test_pfcwd_software_single_queue PASSED                                                                                                                     [ 50%]
tests/test_pfcwd.py::TestPfcwdFunc::test_pfcwd_software_multi_queue PASSED

sudo pytest test_buffer_traditional.py                
===================================================================================================================== test session starts ======================================================================================================================
platform linux -- Python 3.7.5, pytest-7.1.1, pluggy-1.0.0
rootdir: /home/bingwang/work/sonic/sonic-buildimage-master/src/sonic-swss/tests
collected 1 item                                                                                                                                                                                                                                               

test_buffer_traditional.py .                                                                                                                                                                                                                             [100%]


```
Test cases in sonic-mgmt is working in progress.

**Details if related**


orchagent/pfcwdorch.cpp
```diff
@@ -489,7 +489,7 @@ bool PfcWdSwOrch<DropHandler, ForwardHandler>::registerInWdDb(const Port& port,
 
     uint8_t pfcMask = 0;
 
-    if (!gPortsOrch->getPortPfc(port.m_port_id, &pfcMask))
+    if (!gPortsOrch->getPortPfcWatchdogStatus(port.m_port_id, &pfcMask))
```
neethajohn, line 492: Please make this change in  enableBigRedSwitchMode() for consistency even though we don't use it

tests/test_pfcwd.py
```diff
@@ -147,10 +147,12 @@ def _get_bitmask(self, queues):
 
         return str(mask)
 
-    def set_ports_pfc(self, status='enable', pfc_queues=[3,4]):
+    def set_ports_pfc(self, status='enable', pfc_queues=[3,4], watchdog_type='software'):
```
neethajohn, line 150: remove 'watchdog_type'

tests/test_pfcwd.py
```diff
@@ -253,11 +255,11 @@ def test_pfcwd_single_queue(self, dvs, setup_teardown_test):
             self.reset_pfcwd_counters(storm_queue)
             self.stop_pfcwd_on_ports()
 
-    def test_pfcwd_multi_queue(self, dvs, setup_teardown_test):
+    def test_pfcwd_software_multi_queue(self, dvs, setup_teardown_test):
         try:
             # enable PFC on queues
             test_queues = [3, 4]
-            self.set_ports_pfc(pfc_queues=test_queues)
+            self.set_ports_pfc(pfc_queues=test_queues, watchdog_type='software')
```
neethajohn, line 262: remove 'watchdog_type'

bingwang-ms:
@neethajohn The last two comments are addressed. Please take a look again. Thanks

neethajohn:
@bingwang-ms , to support the additional queues, you will also need to make a change in buffermgr to not hardcode the lossless pgs - https://github.com/Azure/sonic-swss/blob/master/cfgmgr/buffermgr.h#L14

bingwang-ms:
> @bingwang-ms , to support the additional queues, you will also need to make a change in buffermgr to not hardcode the lossless pgs - https://github.com/Azure/sonic-swss/blob/master/cfgmgr/buffermgr.h#L14

Updated. Thanks

cfgmgr/buffermgr.h
```diff
@@ -11,7 +11,7 @@
 namespace swss {
 
 #define INGRESS_LOSSLESS_PG_POOL_NAME "ingress_lossless_pool"
-#define LOSSLESS_PGS "3-4"
+#define LOSSLESS_PGS "2,3-4,6"
```
stephenxs, line 14: I don't think it's a good way to have 4 lossless PGs reserved for all ports. Can we reserve 4 lossless PGs for t0 uplink and t1 downlink in dual-ToR topo only?
I think there are 2 ways to do it
- reserve lossless PGs according to the roles of the ports which can be retrieved from `DEVICE_NEIGHBOR` table
- introduce an extra table or leverage the existing table to indicate the lossless PGs on a port

stephenxs (changes requested):

cfgmgr/buffermgr.h
neethajohn: the 'pfc_enable' field in PORT_QOS_MAP already captures the correct lossless PGs based on device role. We can use that to attach the lossless pgs in the BUFFER_PG table

bingwang-ms: Updated. Thanks so much

cfgmgr/buffermgr.cpp
```diff
@@ -442,7 +458,8 @@ void BufferMgr::doTask(Consumer &consumer)
                 if (m_speedLookup.count(port) != 0)
                 {
                     // create/update profile for port
-                    task_status = doSpeedUpdateTask(port, admin_up);
+                    string lossless_pgs = getPfcEnableQueuesForPort(port);
```
stephenxs, line 461: This is still risky. It assumes when BUFFER_PG is created all `PORT`, `CABLE_LENGTH` and `PORT_QOS_MAP` have been received, which is not guaranteed.
Consider this scenario:
1. buffer manager receives the `PORT` table and the `CABLE_LENGTH` table. It will handle them and generate `BUFFER_PG` table. But it hasn't received `PORT_QOS_MAP` yet, so the lossless PGs will be `3-4`.
2. Later it receives `PORT_QOS_MAP` table which contains `2,3-4,6` for dual ToR ports. but corresponding `BUFFER_PG` will not be updated.

I suggest generating `BUFFER_PG` only after all tables are available and not using `LOSSLESS_PGS` in case `pfc_enable` is empty. (I have a vague impression that removing BUFFER_PG is not supported on some platform.)

bingwang-ms: @stephenxs Thanks for the good catch. I update the code to add a retry if `PORT_QOS_MAP` table is not available. Please help take a look once more.

cfgmgr/buffermgr.cpp
```diff
@@ -153,94 +152,99 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, bool admin_up)
...
+        string buffer_pg_key = port + m_cfgBufferPgTable.getTableNameSeparator() + lossless_pg;
 
-    m_cfgBufferPgTable.get(buffer_pg_key, fvVectorPg);
+        m_cfgBufferPgTable.get(buffer_pg_key, fvVectorPg);
 
-    if (!admin_up && m_platform == "mellanox")
-    {
-        // Remove the entry in BUFFER_PG table if any
-        if (!fvVectorPg.empty())
+        if (!admin_up && m_platform == "mellanox")
```
neethajohn, line 167: if this condition is true, as per the current logic, before exiting the if block, we return success. That is the right behavior. With the current logic, the code flow continues. Better to put the for loop within this if block for this case

cfgmgr/buffermgr.cpp
```diff
@@ -153,94 +152,99 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, bool admin_up)
...
-    {
-        SWSS_LOG_NOTICE("Reusing existing profile '%s'", buffer_profile_key.c_str());
-    }
 
-    /* Check if PG Mapping is already then log message and return. */
-    for (auto& prop : fvVectorPg)
-    {
-        if ((fvField(prop) == "profile") && (profile_ref == fvValue(prop)))
+        /* Check if PG Mapping is already then log message and return. */
+        for (auto& prop : fvVectorPg)
```
neethajohn, line 234: with the changes suggested above, loop through the pgs again and retrieve the buffer pg table values for each lossless pg

cfgmgr/buffermgr.cpp
```diff
@@ -346,6 +350,54 @@ void BufferMgr::doBufferMetaTask(Consumer &consumer)
...
+    }
+
+}
+
+/*
+Read pfc_enable entry from PORT_QOS_MAP to decide on which queue to
+apply lossless PG.
+Return the true if PORT_QOS_MAP is already available, otherwise return false
+*/
+bool BufferMgr::getPfcEnableQueuesForPort(std::string port, std::string &pfc_status)
```
neethajohn, line 389: can remove this function

cfgmgr/buffermgr.cpp
stephenxs: Looks like there is still an issue in case a port's `pfc_enable` changed. The lossless buffers for the priorities that were but aren't enabled are kept untouched. In fact, they should be removed.
Probably we need to
- save `pfc_enable` for each port
- when `pfc_enable` is updated, remove the `BUFFER_PG` items generated according to the previous ones before generating the new ones.

sorry for the late response

cfgmgr/buffermgr.cpp
```diff
@@ -438,7 +496,17 @@ void BufferMgr::doTask(Consumer &consumer)
...
                 {
                     // create/update profile for port
-                    task_status = doSpeedUpdateTask(port, admin_up);
+                    string pfc_enable_status;
+                    if (!getPfcEnableQueuesForPort(port, pfc_enable_status))
+                    {
+                        // PORT_QOS_MAP is not ready yet. retry is required
+                        SWSS_LOG_INFO("pfc_enable status is not available for port %s", port.c_str());
+                        task_status = task_process_status::task_need_retry;
+                    }
```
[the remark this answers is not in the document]
bingwang-ms: Updated. Thanks

cfgmgr/buffermgr.cpp
bingwang-ms: Updated. Now the for loop is moved to the if block. Thanks

cfgmgr/buffermgr.cpp
bingwang-ms: Updated. Thanks

cfgmgr/buffermgr.cpp
bingwang-ms: Removed. Thanks

cfgmgr/buffermgr.cpp
```diff
@@ -185,9 +195,26 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, bool admin_up, con
...
+        // Clear applied recorded
+        if (m_portLosslessPgs.count(port) != 0)
+        {
+            m_portLosslessPgs.erase(port);
+        };
+        // Clear recorded pg for admin_down port
+        if (m_portPgLookup.count(port) != 0)
+        {
+            m_portPgLookup.erase(port);
+        }
```
neethajohn, line 208: you don't need any of these

cfgmgr/buffermgr.cpp
```diff
@@ -244,10 +271,78 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, bool admin_up, con
...
         m_cfgBufferPgTable.set(buffer_pg_key, fvVectorPg);
+        // Save the applied lossless PG
+        m_portLosslessPgs[port] = pfc_enable;
+        // Save the profile_ref
+        m_portPgLookup[port] = profile_ref;
     }
     return task_process_status::task_success;
 }
 
+void BufferMgr::updatePortLosslessPg(const string &port)
```
neethajohn, line 282: remove. we don't need to update lossless pg's on the fly. Many asics do not support the buffer pg deletion. Hence you don't need to record them

cfgmgr/buffermgr.cpp
```diff
@@ -185,9 +195,26 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, bool admin_up, con
...
+        }
+        return task_process_status::task_success;
+    }
+
+    for (auto lossless_pg : lossless_pgs)
+    {
+        vector<FieldValueTuple> fvVectorPg, fvVectorProfile;
+        string buffer_pg_key = port + m_cfgBufferPgTable.getTableNameSeparator() + lossless_pg;
+
+        m_cfgBufferPgTable.get(buffer_pg_key, fvVectorPg);
```
neethajohn, line 217: the buffer profile creation/existence check is not dependent on the pgs. so that needn26't be in a loop.  This entire block should move to after L258 and the for loop should include everything from L261.

The logic is create the buffer profile if does not exist. Then go through each lossless pg and check the buffer profile it references. if it is the same, don't do anything otherwise apply the buffer profile to the pg

tests/test_buffer_traditional.py
```diff
@@ -138,6 +141,30 @@ def test_zero_cable_len_profile_update(self, dvs, setup_teardown_test):
...
+                    self.asic_db.wait_for_field_match("ASIC_STATE:SAI_OBJECT_TYPE_INGRESS_PRIORITY_GROUP", orig_pg_name_map[pg], fvs)
+            
+            # Update buf_pg_profile
+            self.get_asic_buf_pg_profiles()
+
+            for pg in self.lossless_pgs:
+                self.app_db.wait_for_field_match("BUFFER_PG_TABLE", self.INTF + ":" + pg, {"profile": orig_lossless_profile})
+            for pg in self.pg_name_map:
+                fvs["SAI_INGRESS_PRIORITY_GROUP_ATTR_BUFFER_PROFILE"] = self.buf_pg_profile[pg]
+                self.asic_db.wait_for_field_match("ASIC_STATE:SAI_OBJECT_TYPE_INGRESS_PRIORITY_GROUP", self.pg_name_map[pg], fvs)
```
neethajohn, line 166: remove. the zero cable len is one use case. Do not overload it. Add a separate test 
1. admin up a port. 
2. record existing profile [pg_lossless_<orig speed>_<cable len>]
3. change the port speed
4. verify new buffer profile is created - [pg_lossless_<new speed>_<cable len]
5. chk the buffer profiles attached to all the lossless pgs of that port - it should not be the orig profile. it should be the new profile

neethajohn (changes requested):

cfgmgr/buffermgr.cpp
bingwang-ms: As discussed offline, we will not support update `BUFFER_PG` without a fully reload. So the change is reverted.

cfgmgr/buffermgr.cpp
bingwang-ms: Removed.

cfgmgr/buffermgr.cpp
bingwang-ms: Removed.

cfgmgr/buffermgr.cpp
bingwang-ms: Updated. Thanks for the great suggestion.

tests/test_buffer_traditional.py
bingwang-ms: Updated. The testcase `test_zero_cable_len_profile_update` is untouched now. I added another case `test_buffer_pg_update`. Thanks

cfgmgr/buffermgr.cpp
```diff
@@ -135,9 +135,10 @@ Create/update two tables: profile (in m_cfgBufferProfileTable) and port buffer (
 */
 task_process_status BufferMgr::doSpeedUpdateTask(string port, bool admin_up)
 {
-    vector<FieldValueTuple> fvVectorPg, fvVectorProfile;
     string cable;
     string speed;
+    string pfc_enable;
+    string applied_lossless_pg_status;
```
stephenxs, line 141: `applied_lossless_pg_status` can be removed since it is no longer used.

cfgmgr/buffermgr.cpp
```diff
@@ -152,47 +153,61 @@ task_process_status BufferMgr::doSpeedUpdateTask(string port, bool admin_up)
         return task_process_status::task_success;
     }
 
-    speed = m_speedLookup[port];
+    if (m_portPfcStatus.count(port) == 0)
+    {
+        // PORT_QOS_MAP is not ready yet. retry is required
+        SWSS_LOG_INFO("pfc_enable status is not available for port %s", port.c_str());
```
stephenxs, line 159: It is possible that `pfc_enable` will not be configured but `speed` is configured. In that case, the notification will remain in `m_toSync` forever.
Maybe we can return `task_success` and call `doSpeedUpdateTask` from `doPortQosTableTask` each time `pfc_enable` is updated.

bingwang-ms: Yes, that's a possible path. But even `pfc_enable` is not available when `speed` is configured, we will return `retry` here, and the main loop will skip this notification. When `pfc_enable` is configured eventually, the `speed` update notification will be consumed, right?
If we never have `pfc_enable` configured, yes, then the `BUFFER_PG` has no chance to be configured. But I think it also makes sense as we don't know on which PG to apply lossless profile.

cfgmgr/buffermgr.cpp
bingwang-ms: Good catch, will update.

cfgmgr/buffermgr.cpp
stephenxs: In case there is no `pfc_enable` configured, the notification will remain in `m_toSync`, and each time `doTask` is called the notification has to be checked. 
In addition, `doTask` will return once `doSpeedUpdateTask` returns a non-success value, leaving all the following notifications unhandled. Check [L506](https://github.com/Azure/sonic-swss/pull/2171/files#diff-ed9e5ec23caf0de71754703dac87fcf2be40f32e2443fa9e5617c1910665d7f2R505)
It's possible in your scenario if there are some inactive ports on the device because no `pfc_enable` configured on inactive ports.

bingwang-ms: Oh, I see. The main loop is break here
https://github.com/Azure/sonic-swss/blob/841f00389b338e91ddc4de460ace4ff96adfa796/cfgmgr/buffermgr.cpp#L444-L447
Thanks for pointing out. I'll post an update soon.

bingwang-ms: I removed the break if `task_status != task_process_status::task_success`. We should try to proceed the next port if one failed. 
With this change, we can consume other ports updates if one is lack of `pfc_enable`.

cfgmgr/buffermgr.cpp
```diff
@@ -399,6 +453,12 @@ void BufferMgr::doTask(Consumer &consumer)
         return;
     }
 
+    if (table_name == CFG_PORT_QOS_MAP_TABLE_NAME)
+    {
+        doPortQosTableTask(consumer);
+        return;
```
stephenxs, line 459: A rare scenario is that the `pfc_enable` of a port is received after the `speed` of any port. In that case, the flow returns here and the lossless PG will not be generated. Maybe call `doSpeedUpdateTask` from `doPortQosTableTask`?

cfgmgr/buffermgr.cpp
stephenxs: Thanks.
A concern is that the notification will remain in `m_toSync` forever if there is no `pfc_enable` configured. It's true for inactive ports that are not connected to ToR switches. As a result, those notifications will be checked and skipped every time a new notification is received.
A solution can be to remove such notification from `m_toSync` and to call `doSpeedUpdateTask` from `doPortQosTableTask`

bingwang-ms: Make sense. Fixed. Thanks

cfgmgr/buffermgr.cpp
bingwang-ms: Fixed. Could you help review again? Thanks

stephenxs (approved):

neethajohn (approved):

qiluo-msft:

PR #2177, opened 2022-03-14
[doc]Moving Configuration.md from swss to sonic-yang-models folder

dgsudharsan:
Signed-off-by: Sudharsan Dhamal Gopalarathnam <sudharsand@nvidia.com>

**What I did**
Moved Configuration.md from swss to yang folder. The configuration schema should be present along with yang models and when users add yang model they could update configuration document as well.

The PR to move to yang folder is https://github.com/Azure/sonic-buildimage/pull/10078
**Why I did it**

**How I verified it**

**Details if related**


qiluo-msft (approved):

qiluo-msft:
This PR only update document, will override checkers.

PR #2182, opened 2022-03-16
Fix for 2053, Fix IPv6 BGP multipath-relax peer-type.

skbarista:
By default the metric for ipv6 static route is 1024. This makes it the
routes learned via the internal bgp sessions inferior to the Ebgp
routes. So make the metric 256 which make the static route metric same
as connected route metric.

**What I did**
By default the metric for ipv6 static route is 1024. This makes it the
routes learned via the internal bgp sessions inferior to the Ebgp
routes. So make the metric 256 which make the static route metric same
as connected route metric.

**Why I did it**
To fix IPv6 BGP multipath-relax peer-type.

**How I verified it**
Verified voq chassis can learn a route which ecmps routes advertised by internal ibgp neighbor and ebgp neighbor.

### DIFF
--- a/cfgmgr/intfmgr.cpp
+++ b/cfgmgr/intfmgr.cpp
@@ -51,6 +51,13 @@ IntfMgr::IntfMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
             setWarmReplayDoneState();
         }
     }
+
+    string swtype;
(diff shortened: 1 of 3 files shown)


rlhui:
is fix needed in master and 202111 too?

arlakshm (approved):

skbarista:
@rlhui The fix merged to master. I will creat a pr for 202111. https://github.com/Azure/sonic-swss/pull/2062

PR #2183, opened 2022-03-16
[aclorch] Do not fail ACL rule remove flow if rule already deleted

liorghub:
**What I did**
Do not fail ACL rule remove flow if rule already deleted.

**Why I did it**
When ACL table that contains rules is being deleted, its rules are being deleted automatically.
In the case when ACL rule handler is called for a rule that was already deleted, handler should do nothing and pass.

**How I verified it**
config acl add table -p Ethernet72 -s ingress DATAACL L3
config acl update full /tmp/56521_acl_file.json
config acl add table -p Ethernet72 -s egress DATAACL L3
config acl update full 56521_acl_file.json
docker exec -it syncd sx_api_flex_acl_dump.py => verify ACL rules exist in "ACL Rules" table.

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -3617,7 +3617,9 @@ bool AclOrch::removeAclRule(string table_id, string rule_id)
     auto rule = getAclRule(table_id, rule_id);
     if (!rule)
     {
-        return false;
+        SWSS_LOG_NOTICE("ACL rule [%s] in table [%s] already deleted",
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@liorghub is this by design or race?

liorghub:
> @liorghub is this by design or race?

This is by design.

stepanblyschak (approved):

prsunny (approved):

bingwang-ms (approved):

PR #2123, opened 2022-01-26
 Add possibility to compile applications with address sanitizer. 

oleksandrivantsiv:
Add SIGTERM handling as a trigger to exit to all applications.
Fix "memory usage after free" error reported by ASAN.
Fix "SEGV on unknown address" error reported by ASAN.
Fix orchagent memory leaks reported by ASAN.

**What I did**
To add a possibility to compile SONiC applications with address sanitizer (ASAN).

> ASAN is a memory error detector for C/C++. It finds:
> - Use after free (dangling pointer dereference)
> - Heap buffer overflow
> - Stack buffer overflow
> - Global buffer overflow
> - Use after return
> - Use after the scope
> - Initialization order bugs
> - Memory leaks

Add SIGTERM handling as a trigger to exit to all applications. The correct shutdown of applications is needed to allow ASAN to generate a report. 

**Why I did it**
To add a possibility to detect memory errors in applications.

**How I verified it**
Compile SWSS with **--enable-asan** option. Install applications to DUT. Run tests that trigger different kinds of memory usage inside of applications. Stop applications with **SIGTERM** signal to trigger ASAN report generation. Reports will be placed in **/var/log/asan/** directory.

**Details if related**
Depends on:
https://github.com/Azure/sonic-buildimage/pull/9857
https://github.com/Azure/sonic-swss-common/pull/579

PR #2186, opened 2022-03-17
Add possibility to compile applications with address sanitizer.

oleksandrivantsiv:
**What I did**
To add a possibility to compile SONiC applications with address sanitizer (ASAN).

> ASAN is a memory error detector for C/C++. It finds:
> - Use after free (dangling pointer dereference)
> - Heap buffer overflow
> - Stack buffer overflow
> - Global buffer overflow
> - Use after return
> - Use after the scope
> - Initialization order bugs
> - Memory leaks

Add SIGTERM handling as a trigger to run a memory leak checker.

**Why I did it**
To add a possibility to detect memory errors in applications.

**How I verified it**
Compile SWSS with **--enable-asan** option. Install applications to DUT. Run tests that trigger different kinds of memory usage inside of applications. Stop applications with **SIGTERM** signal to trigger ASAN report generation. Reports will be placed in **/var/log/asan/** directory.

**Details if related**


liat-grozovik (approved):

liat-grozovik:
@prsunny , @qiluo-msft  kindly reminder to review and signoff.

oleksandrivantsiv:
@prsunny @qiluo-msft kindly remind to review and signoff

saiarcot895:
Rather than modifying all of the `Makefile.am` files, in `debian/rules`, you should be able to set `DEB_BUILD_MAINT_OPTIONS=sanitize=+address` and have it apply to everything. Debian build profiles can then also be used to conditionally enable ASAN builds.

oleksandrivantsiv:
> Rather than modifying all of the `Makefile.am` files, in `debian/rules`, you should be able to set `DEB_BUILD_MAINT_OPTIONS=sanitize=+address` and have it apply to everything. Debian build profiles can then also be used to conditionally enable ASAN builds.

Enabling ASAN in makefile gives us the flexibility to affect only applications we want.
Enabling it globally will affect unit tests as well. Unit tests have dozens of memory leaks, use after free errors, and other memory errors. They are running during the compilation time and it will break the compilation because of memory errors. Fixing memory errors in unit tests won't give and benefit and will produce hundreds of lines changes.
So I suggest going with the proposed solution.

saiarcot895:
For that purpose, it might be better to use an [ignore list](https://clang.llvm.org/docs/SanitizerSpecialCaseList.html) (GCC should work the same way, since the code is mostly copied from Clang to GCC).

cfgmgr/natmgrd.cpp
```diff
@@ -98,6 +102,12 @@ void sigterm_handler(int signo)
         natmgr->cleanupMangleIpTables();
         natmgr->cleanupPoolIpTable();
     }
+
+#if defined(ASAN_ENABLED)
+    __lsan_do_leak_check();
+#endif
+    signal(signo, SIG_DFL);
+    raise(signo);
```
qiluo-msft, line 110: Do you intend to move code into above #if block?

cfgmgr/buffermgrd.cpp
```diff
@@ -109,6 +114,15 @@ shared_ptr<vector<KeyOpFieldsValuesTuple>> load_json(string file)
     }
 }
 
+#if defined(ASAN_ENABLED)
+void sigterm_handler(int signo)
```
qiluo-msft, line 118: >sigterm_handler [](http://example.com/codeflow?start=5&length=15)

This function is implemented everywhere. Could you extract it into one single cpp file, and add it into each target in Makefile.

cfgmgr/buffermgrd.cpp
```diff
@@ -121,6 +135,14 @@ int main(int argc, char **argv)
 
     SWSS_LOG_NOTICE("--- Starting buffermgrd ---");
 
+#if defined(ASAN_ENABLED)
+    if (signal(SIGTERM, sigterm_handler) == SIG_ERR)
```
qiluo-msft, line 139: >signal [](http://example.com/codeflow?start=8&length=6)

This code lines are repeated in each main(). Could you define a common main function, and calling into old main(), which renamed to like buffermgrd_main().

saiarcot895: Could you also elaborate on adding a signal handler for calling `__lsan_do_leak_check` explicitly? Leak checks are atuomatically done when the process is exiting gracefully, so this shouldn't be needed at all.

saiarcot895:
> For that purpose, it might be better to use an [ignore list](https://clang.llvm.org/docs/SanitizerSpecialCaseList.html) (GCC should work the same way, since the code is mostly copied from Clang to GCC).

Looks like this (and `-fsanitize-blacklist=`) isn't supported on GCC yet, so ignore this.

cfgmgr/buffermgrd.cpp
oleksandrivantsiv: > Could you also elaborate on adding a signal handler for calling `__lsan_do_leak_check` explicitly? Leak checks are atuomatically done when the process is exiting gracefully, so this shouldn't be needed at all.

There are a couple of problems here why it is not working as expected. All applications are running as daemons and handing requests in the loop. To exit gracefully they need to receive a request to shut down, exit from the loop, and deallocate all the memory previously allocated. Then exit from the main function and after that address sanitizer will trigger the leak checker. The main problem here is that most if not all applications are running an infinite loop. They are not exiting correctly from the loop and are simply killed with the SIGTERM signal. Because of that when address sanitizer starts leaks checking it reports all the memory that was allocated during the initialization of the application as a leaked. So reports had dozens of false-positive memory leaks.
Even if we had decided to change the applications and exit from the loop correctly on a signal this wouldn't solve the problem. Because most of the objects and raw memory allocated on the initialization are not deallocated. And we again would have dozens of leaks in the reports.
Some time ago I created a pull request that does described changes: https://github.com/Azure/sonic-swss/pull/2123. And then decided to close it and change implementation to call __lsan_do_leak_check from the signal handler directly. This allows triggering leaks checker without the need to exit the loop and change of application. It reports only memory that is definitely lost and doesn't care about the memory that still has valid pointers.

cfgmgr/buffermgrd.cpp
oleksandrivantsiv: @qiluo-msft done

oleksandrivantsiv:
> > For that purpose, it might be better to use an [ignore list](https://clang.llvm.org/docs/SanitizerSpecialCaseList.html) (GCC should work the same way, since the code is mostly copied from Clang to GCC).
> 
> Looks like this (and `-fsanitize-blacklist=`) isn't supported on GCC yet, so ignore this.

@saiarcot895 -fsanitize-blacklist is available only for clang:
https://github.com/google/sanitizers/wiki/AddressSanitizer
```
You may also ignore certain functions using a blacklist: create a file my_ignores.txt and pass it to [AddressSanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer) at compile time using -fsanitize-blacklist=my_ignores.txt (This flag is new and is only supported by Clang now):
```

cfgmgr/natmgrd.cpp
oleksandrivantsiv: @qiluo-msft  yes, there is an issue here with closing an application after clean-up. 
Supervisord sends SIGTERM to all applications that it controls. Application after receiving the signal may optionally run a clean-up routine and should exit. natmngrd runs a cleanup from the signal handler but doesn't exit after that. After supervisord sends a signal it waits for 10 seconds (default timeout) to allow the application to exit correctly. If the application is not exiting in the given timeout supervisord kills the application. natmngrd is always killed by the supervisord. This affects swss container restart time. To exit from the user-defined signal_handler application must restore the default signal handler and raise the signal to itself again.
In the latest commits to this PR I removed this change to not mix it with ASAN changes and will create a separated PR for a bug fix.

oleksandrivantsiv:
@saiarcot895 @qiluo-msft I resolved all the comments. Please take a look.

cfgmgr/natmgrd.cpp
```diff
@@ -98,6 +100,10 @@ void sigterm_handler(int signo)
         natmgr->cleanupMangleIpTables();
         natmgr->cleanupPoolIpTable();
     }
+
+    if (old_sigaction.sa_handler != SIG_IGN && old_sigaction.sa_handler != SIG_DFL) {
```
qiluo-msft, line 104: >!= SIG_IGN [](http://example.com/codeflow?start=33&length=10)

Why compare a function pointer to integer macro?

cfgmgr/teammgrd.cpp
```diff
@@ -38,7 +45,15 @@ int main(int argc, char **argv)
...
     /* Register the signal handler for SIGTERM */
-    signal(SIGTERM, sig_handler);
+    struct sigaction sigact;
+    sigemptyset(&sigact.sa_mask);
+    sigact.sa_flags = 0;
+    sigact.sa_handler = sig_handler;
+    if (sigaction(SIGTERM, &sigact, &old_sigaction))
+    {
+        SWSS_LOG_ERROR("failed to setup SIGTERM action handler");
+        exit(1);
```
qiluo-msft, line 55: >1 [](http://example.com/codeflow?start=13&length=1)

Use `EXIT_FAILURE`

cfgmgr/natmgrd.cpp
qiluo-msft: Thanks for reporting this bug! Could you separate the bug fix into a standalone PR to make backport easier?

Suggest follow the practice of using `sig_atomic_t` and gracefully handle it in the main loop. This is not a trivial fix but worthy.

cfgmgr/natmgrd.cpp
```diff
@@ -129,7 +135,11 @@ int main(int argc, char **argv)
 
         cleanupNotifier = std::make_shared<swss::NotificationProducer>(&appDb, "NAT_DB_CLEANUP_NOTIFICATION");
 
-        if (signal(SIGTERM, sigterm_handler) == SIG_ERR)
+        struct sigaction sigact;
+        sigemptyset(&sigact.sa_mask);
+        sigact.sa_flags = 0;
+        sigact.sa_handler = sigterm_handler;
```
qiluo-msft, line 141: Duplicated code, better to extract a function.

cfgmgr/natmgrd.cpp
oleksandrivantsiv: SIG_IGN and SIG_DFL are valid values of sa_handler:
https://man7.org/linux/man-pages/man2/sigaction.2.html
```
       sa_handler specifies the action to be associated with signum and
       is be one of the following:

       * SIG_DFL for the default action.

       * SIG_IGN to ignore this signal.

       * A pointer to a signal handling function.  This function
         receives the signal number as its only argument.
```
This comparison ensures that sa_handler value is a valid function pointer and it is safe to call it.

cfgmgr/teammgrd.cpp
oleksandrivantsiv: done

cfgmgr/natmgrd.cpp
oleksandrivantsiv: I simplified initialization to remove code duplication.

cfgmgr/natmgrd.cpp
oleksandrivantsiv: https://github.com/Azure/sonic-swss/pull/2232

liat-grozovik:
@qiluo-msft kindly reminder to review and signoff

qiluo-msft (approved):

PR #2194, opened 2022-03-23
Dynamic port configuration - add port buffer cfg to the port ref counter

dprital:
**What I did**

This PR replace PR [#2022](https://github.com/Azure/sonic-swss/pull/2022)

Added increasing/decreasing to the port ref counter each time a port buffer configuration is added or removed
Implemented according to the - https://github.com/Azure/SONiC/pull/900

**Why I did it**
In order to avoid cases where a port is removed before the buffer configuration on this port were removed also

**How I verified it**
VS Test was added in order to test it. 
we remove a port with buffer configuration and the port is not removed. only after all buffer configurations on this port were 
removed - this port will be removed.

**Details if related**

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -42,6 +42,9 @@ map<string, string> buffer_to_ref_table_map = {
     {buffer_profile_list_field_name, APP_BUFFER_PROFILE_TABLE_NAME}
 };
 
+std::map<string, std::map<size_t, string>> pg_port_flags;
+std::map<string, std::map<size_t, string>> queue_port_flags;
+
 BufferOrch::BufferOrch(DBConnector *applDb, DBConnector *confDb, DBConnector *stateDb, vector<string> &tableNames) :
     Orch(applDb, tableNames),
     m_flexCounterDb(new DBConnector("FLEX_COUNTER_DB", 0)),
@@ -949,6 +952,38 @@ task_process_status BufferOrch::processQueue(KeyOpFieldsValuesTuple &tuple)
                     }
                 }
             }
+
+            /* when we apply buffer configuration we need to increase the ref counter of this port
+             * or decrease the ref counter for this port when we remove buffer cfg
+             * so for each priority cfg in each port we will increase/decrease the ref counter
+             * also we need to know when the set command is for creating a buffer cfg or modifying buffer cfg -
+             * we need to increase ref counter only on create flow.
+             * so we added a map that will help us to know what was the last command for this port and priority -
+             * if the last command was set command then it is a modify command and we dont need to increase the buffer counter
+             * all other cases (no last command exist or del command was the last command) it means that we need to increase the ref counter */
+            if (op == SET_COMMAND) 
+            {
+                if (queue_port_flags[port_name][ind] != SET_COMMAND) 
+                {
+                    /* if the last operation was not "set" then it's create and not modify - need to increase ref counter */
+                    gPortsOrch->increasePortRefCount(port_name);
+                }
+            } 
+            else if (op == DEL_COMMAND)
+            {
+                if (queue_port_flags[port_name][ind] == SET_COMMAND) {
+                    /* we need to decrease ref counter only if the last operation was "SET_COMMAND" */
+                    gPortsOrch->decreasePortRefCount(port_name);
+                }
+            } 
+            else 
+            {
+                SWSS_LOG_ERROR("operation value is not SET or DEL (op = %s)", op.c_str());
+                return task_process_status::task_invalid_entry;
+            }
+            /* save the last command (set or delete) */
+            queue_port_flags[port_name][ind] = op;
+
         }
     }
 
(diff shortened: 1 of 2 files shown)


tests/test_port_add_remove.py
```diff
@@ -0,0 +1,150 @@
...
+"""
+DELETE_CREATE_ITERATIONS defines the number of iteration of delete and create to  ports,
+we add different timeouts between delete/create to catch potential race condition that can lead to system crush.
+"""
+DELETE_CREATE_ITERATIONS = 10
+
+@pytest.mark.usefixtures('dvs_port_manager')
+class TestPortAddRemove(object):
+
+    def test_remove_port_with_buffer_cfg(self, dvs, testlog):
```
zhenggen-xu, line 18: we should remove_port, add_port, then remove_port  with buffer_cfg to see if things are as expected.  This is to explicitly  excise the ref counters during those operations.  maybe we can have a test case like "test_remove_add_remove_port_with_buffer_cfg" ?

orchagent/bufferorch.cpp
```diff
@@ -942,6 +945,38 @@ task_process_status BufferOrch::processQueue(KeyOpFieldsValuesTuple &tuple)
...
+            {
+                if (queue_port_flags[port_name][ind] != SET_COMMAND) 
+                {
+                    /* if the last operation was not "set" then it's create and not modify - need to increase ref counter */
+                    gPortsOrch->increasePortRefCount(port_name);
+                }
+            } 
+            else if (op == DEL_COMMAND)
+            {
+                if (queue_port_flags[port_name][ind] == SET_COMMAND) {
```
zhenggen-xu, line 967: minor comment,  "{" to separate line to be consistent.

dprital:
azpw

tests/test_port_add_remove.py
stephenxs: @zhenggen-xu 
to me, this is a negative test that verifies a port will not be removed if there is a buffer configuration on it. We should keep this test.
So i would like to have the test `test_remove_add_remove_port_with_buffer_cfg` like this:
- when removing the port for the first time, remove port first and then the buffer configuration. verify the port will not be removed without buffer configuration cleared.
- when adding the port, add the port first and then the buffer configuration, as you suggested.
- when removing the port for the second time, remove the buffer configuration and then the port, as you suggested.
By doing so, we are able to verify both orders.
How do you think?

zhenggen-xu: @stephenxs  sounds good to me.

zhenggen-xu: Sure, looks good to me.

zhenggen-xu (dismissed):

dprital:
@prsunny - Can you please review ?

tests/test_port_add_remove.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # Remove buffer queue
+        for key in buffer_queues.keys():
+            config_db.delete_entry('BUFFER_QUEUE', key)
+
+        num = asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_PORT",
+                                      num_of_ports-1,
+                                      polling_config = PollingConfig(polling_interval = 1, timeout = 5.00, strict = True))
+
+        # verify that the port wasn't removed since we still have buffer cfg
```
neethajohn, line 99: incorrect comment. at this point the port is expected to be removed

neethajohn (approved):

dprital:
@prsunny - As discussed, can you please check why the Warm Reboot VS tests fail?. Thanks.

liat-grozovik:
@dprital  what is the status of this PR? do you have an ETA to fix the tests

dprital:
@neethajohn  , @zhenggen-xu  - This is an old PR which was approved by you, as the checkers failed for some time, i fixed the test and now it pass. Would l ike to ask you to look at it (the code was not changed (but the test itself) and I would ilke you to approve so it will be able to be merged.  Thanks in advance!

tests/test_port_add_remove.py
```diff
@@ -8,30 +9,50 @@
...
     def test_remove_add_remove_port_with_buffer_cfg(self, dvs, testlog):
         config_db = dvs.get_config_db()
         asic_db = dvs.get_asic_db()
+        state_db = dvs.get_state_db()
+        app_db = dvs.get_app_db()
+
+        # set mmu size
+        fvs = {"mmu_size": "12766208"}
+        state_db.create_entry("BUFFER_MAX_PARAM_TABLE", "global", fvs)
+        bufferMaxParameter = state_db.wait_for_entry("BUFFER_MAX_PARAM_TABLE", PORT_A)
```
stephenxs, line 37: We should not assume the test `test_remove_add_remove_port_with_buffer_cfg` will alway run at first.
So let's move L35~36 to a commonplace, something like `setup_db`, and remove L37 which is unnecessary.
Otherwise, it will fail if we run other test cases without this one.

stephenxs (approved):
LGTM

liat-grozovik:
Note: approved and merged following prev approval and only tests updates.
@dprital is it needed for 202205? if so please ensure it can be cherry picked and passing. if not creale new PR against 202205

dprital:
> Note: approved and merged following prev approval and only tests updates. @dprital is it needed for 202205? if so please ensure it can be cherry picked and passing. if not creale new PR against 202205

Needed for 202205 and can be taken.

PR #2195, opened 2022-03-24
[VS test] Increase VS test time, skip dpb flaky test

prsunny:
**What I did**
DPB tests are flaky, need investigation before enabling. 
Increase VS test duration

**Why I did it**
Unblock some PRs

**How I verified it**

**Details if related**


StormLiangMS (approved):
LGTM

PR #2202, opened 2022-03-26
[BFD]Registering BFD state change callback during session creation

dgsudharsan:
**What I did**
Register BFD callback during the BFD session creation flow

**Why I did it**
BFD session callback is required to get state changes from SAI and there by remove or add endpoints. However not all vendors support it. So moved it from switch initialization and set it during first session creation with query for support from SAI.

**How I verified it**
Verified it manually with BFD show command which also relies on BFD session change notifications

**Details if related**

### DIFF
--- a/orchagent/bfdorch.cpp
+++ b/orchagent/bfdorch.cpp
@@ -6,6 +6,7 @@
 #include "notifier.h"
 #include "sai_serialize.h"
 #include "directory.h"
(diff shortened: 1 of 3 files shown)


liat-grozovik:
see comments as in https://github.com/Azure/sonic-swss/pull/2203

prsunny (approved):

PR #2203, opened 2022-03-26
[202012][BFD]Registering BFD state change callback during session creation

dgsudharsan:
**What I did**
Register BFD callback during the BFD session creation flow

**Why I did it**
BFD session callback is required to get state changes from SAI and there by remove or add endpoints.

**How I verified it**
Verified it manually with BFD show command which also relies on BFD session change notifications

**Details if related**

### DIFF
--- a/orchagent/bfdorch.cpp
+++ b/orchagent/bfdorch.cpp
@@ -6,6 +6,7 @@
 #include "notifier.h"
 #include "sai_serialize.h"
 #include "directory.h"
+#include "notifications.h"
 
 using namespace std;
 using namespace swss;
@@ -21,6 +22,7 @@ extern sai_bfd_api_t*       sai_bfd_api;
 extern sai_object_id_t      gSwitchId;
 extern sai_object_id_t      gVirtualRouterId;
 extern PortsOrch*           gPortsOrch;
+extern sai_switch_api_t*    sai_switch_api;
 extern Directory<Orch*>     gDirectory;
 
 const map<string, sai_bfd_session_type_t> session_type_map =
@@ -57,6 +59,7 @@ BfdOrch::BfdOrch(DBConnector *db, string tableName, TableConnector stateDbBfdSes
     m_bfdStateNotificationConsumer = new swss::NotificationConsumer(notificationsDb, "NOTIFICATIONS");
     auto bfdStateNotificatier = new Notifier(m_bfdStateNotificationConsumer, this, "BFD_STATE_NOTIFICATIONS");
     Orch::addExecutor(bfdStateNotificatier);
+    register_state_change_notif = false;
 }
 
 BfdOrch::~BfdOrch(void)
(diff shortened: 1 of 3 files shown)


orchagent/bfdorch.cpp
```diff
@@ -152,8 +155,52 @@ void BfdOrch::doTask(NotificationConsumer &consumer)
...
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Unable to query the BFD change notification capability");
+        return false;
+    }
+
+    if (!capability.set_implemented)
+    {
+        SWSS_LOG_ERROR("BFD register change notification not supported");
```
liat-grozovik, line 176: why it is an error and not just warning? it is OK that some vendors does not support this option and BFD can work

orchagent/bfdorch.cpp
```diff
@@ -152,8 +155,52 @@ void BfdOrch::doTask(NotificationConsumer &consumer)
...
+    attr.value.ptr = (void *)on_bfd_session_state_change;
+
+    status = sai_switch_api->set_switch_attribute(gSwitchId, &attr);
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to register BFD notification handler");
+        return false;
+    }
+    return true;
```
liat-grozovik, line 190: should we have here? 'register_state_change_notif = true' why only on create session and not just when registration is done?

orchagent/bfdorch.cpp
dgsudharsan: This error would be thrown when user explicitly configures BFD. In that case we need error LOG. It will not appear if BFD is not configured by the user

orchagent/bfdorch.cpp
dgsudharsan: Registration is done as part of first session creation just before we create BFD. 
When the user tries to create BFD session, registration support is checked and if not implemented a syslog error would be thrown.
If it is supported registration would be done and the first session would be created. Subsequent session creations will not perform registrations.

orchagent/bfdorch.cpp
dgsudharsan: I don't think deregistration is required. The only problem we might end up is registering callback when it is not supported and this PR handles it gracefully.
We have multiple other callbacks like port,fdb which are registered and there is no deregistering sequence.

orchagent/vnetorch.cpp
```diff
@@ -939,7 +939,7 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
             NextHopGroupKey nhg = it_route->second;
             if(--syncd_nexthop_groups_[vnet][nhg].ref_count == 0)
             {
-                if (nexthops.getSize() > 1)
+                if (nhg.getSize() > 1)
```
prsunny, line 942: Can you please remove this change?

orchagent/bfdorch.cpp
prsunny: Agree with @dgsudharsan

orchagent/vnetorch.cpp
dgsudharsan: Done

prsunny (approved):

PR #2205, opened 2022-03-28
Add mandatory ACL actions when creating mirror ACL table

ysmanman:
**What I did**
Add mandatory ACL actions when creating mirror ACL table

**Why I did it**
BRCM SAI requires to pass ACL actions when creating mirror ACL table. 

**How I verified it**
Loaded image on switch and checked SAI redis record to make sure counter & mirror_ingress actions are passed to SAI in mirror ACL creatation.

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -301,6 +301,12 @@ const set<sai_acl_action_type_t>& AclTableType::getActions() const
     return m_aclAcitons;
 }
 
+bool AclTableType::addAction(sai_acl_action_type_t action)
+{
+    m_aclAcitons.insert(action);
+    returne true;
+}
+
 AclTableTypeBuilder& AclTableTypeBuilder::withName(string name)
 {
     m_tableType.m_name = name;
@@ -1799,6 +1805,50 @@ AclTable::AclTable(AclOrch *pAclOrch, string id) noexcept : m_pAclOrch(pAclOrch)
 AclTable::AclTable(AclOrch *pAclOrch) noexcept : m_pAclOrch(pAclOrch)
 {
 
+}
+
+bool AclTable::addMandatoryActions()
+{
+    SWSS_LOG_ENTER();
+
+    if (stage == ACL_STAGE_UNKNOWN)
+    {
+        return false;
+    }
+
+    if (!m_pAclOrch->isAclActionListMandatoryOnTableCreation(stage))
+    {
+        // No op if action list is not mandatory on table creation.
+        return true;
+    }
+
+    if (type.getName() == TABLE_TYPE_MIRROR || type.getName() == TABLE_TYPE_MIRRORV6)
+    {
+
+        sai_acl_action_type_t acl_action = SAI_ACL_ACTION_TYPE_COUNTER;
+        if (m_pAclOrch->isAclActionSupported(stage, acl_action))
+        {
+            SWSS_LOG_INFO("Add counter acl action");
+            type.addAction(acl_action);
+        }
+
+        if (stage == ACL_STAGE_INGRESS)
+        {
+            acl_action = SAI_ACL_ACTION_TYPE_MIRROR_INGRESS;
+        }
+        else
+        {
+            acl_action = SAI_ACL_ACTION_TYPE_MIRROR_EGRESS;
+        }
+
+        if (m_pAclOrch->isAclActionSupported(stage, acl_action))
+        {
+            SWSS_LOG_INFO("Add ingress/egress acl action");
+            type.addAction(acl_action);
+        }
+    }
+
+    return true;
 }
 
 bool AclTable::validateAddType(const AclTableType &tableType)
@@ -3938,6 +3988,8 @@ void AclOrch::doAclTableTask(Consumer &consumer)
             }
 
             newTable.validateAddType(*tableType);
+
+            newTable.addMandatoryActions();
 
             // validate and create/update ACL Table
             if (bAllAttributesOk && newTable.validate())
--- a/orchagent/aclorch.h
+++ b/orchagent/aclorch.h
@@ -156,6 +156,8 @@ public:
     const set<sai_acl_range_type_t>& getRangeTypes() const;
     const set<sai_acl_action_type_t>& getActions() const;
 
+    bool addAction(sai_acl_action_type_t action);
+
 private:
     friend class AclTableTypeBuilder;
 
@@ -386,6 +388,9 @@ public:
     bool validateAddPorts(const unordered_set<string> &value);
     bool validate();
     bool create();
+
+    // Add actions to ACL table if mandatory action list is required on table creation.
+    bool addMandatoryActions();
 
     // validate AclRule match attribute against rule and table configuration
     bool validateAclRuleMatch(sai_acl_entry_attr_t matchId, const AclRule& rule) const;


orchagent/aclorch.cpp
```diff
@@ -1801,6 +1807,50 @@ AclTable::AclTable(AclOrch *pAclOrch) noexcept : m_pAclOrch(pAclOrch)
...
+        return false;
+    }
+
+    if (!m_pAclOrch->isAclActionListMandatoryOnTableCreation(stage))
+    {
+        // No op if action list is not mandatory on table creation.
+        return true;
+    }
+
+    if (type.getName() == TABLE_TYPE_MIRROR || type.getName() == TABLE_TYPE_MIRRORV6)
```
bingwang-ms, line 1825: Please also include `TABLE_TYPE_MIRROR_DSCP`

bingwang-ms:
@stepanblyschak Could you help take a look?

arlakshm:
@ysmanman, can you resolve the open comment

orchagent/aclorch.cpp
```diff
@@ -301,6 +301,12 @@ const set<sai_acl_action_type_t>& AclTableType::getActions() const
     return m_aclAcitons;
 }
 
+bool AclTableType::addAction(sai_acl_action_type_t action)
+{
+    m_aclAcitons.insert(action);
+    returne true;
```
arlakshm, line 307: typo ?

orchagent/aclorch.cpp
ysmanman: Thanks. Done.

orchagent/aclorch.cpp
ysmanman: @arlakshm Thanks for catching this. Fixed the typo.

ysmanman:
> @ysmanman, can you resolve the open comment

@arlakshm Sorry for the delay. Resolved the comment.

bingwang-ms (approved):

ysmanman:
@arlakshm @bingwang-ms I added test coverage as requested. Please take a look at it.

orchagent/aclorch.cpp
```diff
@@ -301,6 +301,12 @@ const set<sai_acl_action_type_t>& AclTableType::getActions() const
     return m_aclAcitons;
 }
 
+bool AclTableType::addAction(sai_acl_action_type_t action)
```
stepanblyschak, line 304: ```AclTableType``` was designed to be an immutable data structure, so you cannot change it after it has been created. This is due to the fact that matches, actions, bind point types are CREATE_ONLY SAI attributes. Adding mutable methods breaks this invariant. An object of type ```AclTableType``` may have been used to create tables already and changing the definition of a table type may cause divergence between software and hardware state.
Instead, a builder ```AclTableTypeBuilder``` is used to create ```AclTableType```s.
Why not adding neccessary actions in ```initDefaultTableTypes``` - https://github.com/Azure/sonic-swss/blob/master/orchagent/aclorch.cpp#L2895 ?

bingwang-ms: > `AclTableType` was designed to be an immutable data structure, so you cannot change it after it has been created. This is due to the fact that matches, actions, bind point types are CREATE_ONLY SAI attributes. Adding mutable methods breaks this invariant. An object of type `AclTableType` may have been used to create tables already and changing the definition of a table type may cause divergence between software and hardware state. Instead, a builder `AclTableTypeBuilder` is used to create `AclTableType`s. Why not adding neccessary actions in `initDefaultTableTypes` - https://github.com/Azure/sonic-swss/blob/master/orchagent/aclorch.cpp#L2895 ?

We also saw similar issue on broadcom paltform. Please see https://github.com/Azure/sonic-buildimage/issues/10425.
We may have two options to workaround this issue
1. Hardcode `is_mandatory` to `False` for the general ACL type, such as `L3`, `L3V6`
2. Add default action list for these ACL tables
What do you think?

stepanblyschak: Ok, I see, so the issue may be due to SAI telling that action list is mandatory even if it is not (because it worked without actions before ACL table types changes). In this case, hardcoding is_mandatory to be False for the types we know should work without explicitly passing action list should be a simpler workaround. Put a comment that this hardcode needs to be removed once SAI is fixed.

ysmanman: Based on our communication with broadcom,  action list is mandatory when creating mirror acl table. Otherwise,  SAI acl table creation will fail. Can you elaborate how would hardcoding is_mandatory to be False work around the SAI limitation?

ysmanman: If we only bypass the above check, aclorch attempts to create mirror acl table without passing mandatory actions and the corresponding SAI call will fail on platforms like broadcom that requires actions in acl table creation.

stepanblyschak: > Based on our communication with broadcom, action list is mandatory when creating mirror acl table. Otherwise, SAI acl table creation will fail. Can you elaborate how would hardcoding is_mandatory to be False work around the SAI limitation?

@ysmanman This means mirroring never worked on some platforms or wasn't tested? In 201911 for all table types there is no SAI_ACL_TABLE_ATTR_ACL_ACTION_TYPE_LIST https://github.com/Azure/sonic-swss/blob/201911/orchagent/aclorch.cpp#L1516.

ysmanman: @stepanblyschak Mirroring probably worked in fixed box before. My PR is to get mirroring work in broadcom VOQ chassis.

bingwang-ms: > 

Thanks for the update. Could you clarify why SAI call will fail if we bypass this check? Before this change, there is no check for action list existance. Is it a new feature of SAI?

ysmanman: @bingwang-ms @stepanblyschak Let me provide some context of this PR. We were testing mirroring in broadcom DNX or VOQ chassis with broadcom SAI 5.2 and observed orchagent failed to add ACL rule in mirroring ACL table.  The failure is because the ACL actions is not provided at the time of mirror ACL table creation.  Broadcom DNX  devices need ACL action list at the time of ACL table creation. To address this issue,  this PR adds required ACL actions to mirroring type ACL table if the action list is mandatory on table creation.

bingwang-ms: Thanks @ysmanman .
I saw similar issues when testing on Broadcom DNX devices. Please see https://github.com/Azure/sonic-buildimage/issues/10425.
Per your clarification, the action_list for ACL table is a must when we creating ACL table now? If that's the case, my by-pass solution will not work. The solution in this PR makes sense. We have to do similar change for other ACL table types (L3, L3V6, and etc.)

bingwang-ms (approved):

bingwang-ms:
@ysmanman Can you add some vstest case (in python) to cover the change? As far as I know, the C++ UT is not caculated as coverage. Thanks

bingwang-ms:
The change is verified on 7050cx3 device
```
collected 10 items                                                                                                                                                                                                                                             

everflow/test_everflow_testbed.py::TestEverflowV4IngressAclIngressMirror::test_everflow_basic_forwarding[cli-downstream] PASSED                                                                                                                          [ 10%] ^H
everflow/test_everflow_testbed.py::TestEverflowV4IngressAclIngressMirror::test_everflow_basic_forwarding[cli-upstream] PASSED                                                                                                                            [ 20%] ^H
everflow/test_everflow_testbed.py::TestEverflowV4IngressAclIngressMirror::test_everflow_neighbor_mac_change[cli-downstream] PASSED                                                                                                                       [ 30%]
everflow/test_everflow_testbed.py::TestEverflowV4IngressAclIngressMirror::test_everflow_neighbor_mac_change[cli-upstream] PASSED                                                                                                                         [ 40%] ^H
everflow/test_everflow_testbed.py::TestEverflowV4IngressAclIngressMirror::test_everflow_remove_unused_ecmp_next_hop[cli-downstream] PASSED                                                                                                               [ 50%] ^H
everflow/test_everflow_testbed.py::TestEverflowV4IngressAclIngressMirror::test_everflow_remove_unused_ecmp_next_hop[cli-upstream] PASSED                                                                                                                 [ 60%]
everflow/test_everflow_testbed.py::TestEverflowV4IngressAclIngressMirror::test_everflow_remove_used_ecmp_next_hop[cli-downstream]  ^HPASSED                                                                                                                 [ 70%]
everflow/test_everflow_testbed.py::TestEverflowV4IngressAclIngressMirror::test_everflow_remove_used_ecmp_next_hop[cli-upstream] PASSED                                                                                                                   [ 80%] ^H
everflow/test_everflow_testbed.py::TestEverflowV4IngressAclIngressMirror::test_everflow_dscp_with_policer[cli-downstream] PASSED                                                                                                                         [ 90%]
everflow/test_everflow_testbed.py::TestEverflowV4IngressAclIngressMirror::test_everflow_dscp_with_policer[cli-upstream]  ^HPASSED                                                                                                                           [100%]

```

bingwang-ms:
@ysmanman The change has been verified. I have raised another PR https://github.com/Azure/sonic-swss/pull/2298 in the same way as you did to fix other ACL table types. Changes in this PR is included. Could you help take a look? Thanks

ysmanman:
> @ysmanman The change has been verified. I have raised another PR #2298 in the same way as you did to fix other ACL table types. Changes in this PR is included. Could you help take a look? Thanks

@bingwang-ms Sure, I will take a look. Since your PR includes the change in my PR,  I will close my PR once your PR is merged. Thanks.

arlakshm:
Hi @ysmanman, #2298, is merged now. Can you please close this PR if your changes are all present in #2298

ysmanman:
> Hi @ysmanman, #2298, is merged now. Can you please close this PR if your changes are all present in #2298

Hi @arlakshm, yes, my changes is included. I am closing the PR.

PR #2206, opened 2022-03-28
[Bugfix][QosOrch] The notifications cannot be drained in QosOrch in case the first one needs to retry

stephenxs:
**What I did**
Bugfix: The notifications cannot be drained in QosOrch in case the first one needs to retry

**Why I did it**
Bugfix

**How I verified it**

Mock test and manual test.

**Details if related**
This is because both `doTask` and the table handlers take `consumer` as the argument.
In `doTask`, each notification is iterated in the main loop and passed to table handlers. However, `consumer` is passed to table handlers, which makes it unable to access the rest notifications following the first one. In case the first one needs to retry and remains in `m_toSync`, the rest notifications do not have opportunity to run.


orchagent/qosorch.cpp
```diff
@@ -1898,7 +1892,7 @@ void QosOrch::doTask(Consumer &consumer)
             continue;
         }
 
-        auto task_status = (this->*(m_qos_handler_map[qos_map_type_name]))(consumer);
+        auto task_status = (this->*(m_qos_handler_map[qos_map_type_name]))(consumer, it->second);
```
neethajohn, line 1895: Do we need to pass consumer any more?

stephenxs: @neethajohn  
Yes, we do.
`doTask` => `handleDscpToTcTable` => `processWorkItem`
`processWorkItem` is shared for many QoS map handlers. In this function, `consumer` is required for fetching table name which is required `m_qos_maps`.

stephenxs:
rebased on top of the latest master in order to resolve vs test failures

neethajohn (approved):

PR #2208, opened 2022-03-29
[202012] [vxlan] Remove tunnel map objects on VNET tunnel removal

volodymyrsamotiy:

### Porting changes in PR https://github.com/Azure/sonic-swss/pull/2150 to ```202012``` branch.

**What I did**

* VxLAN tunnel map objects were not removed on VNET tunnel removal.
* It caused config lefovers which is not expected.
* If tunnel is removed then all related objects should be removed as well.
* To do so just used deleteTunnelHw() function which does proper remove.

**Why I did it**

To cleanup config leftovers after VNET tunnel removal.

**How I verified it**
* Configure 2 VNETs and then remove them one by one.
* Verify that after 1st VNET removed tunnel and all related objects still exist.
* Verify that after 2nd (last) VNET removed tunnel and all related objects are removed.

**Details if related**
N/A


prsunny (approved):

prsunny:
Build and Tests has passed

PR #2215, opened 2022-04-01
[autoneg] add support for capability checker and remote speed advertisement

ds952811:
…sement

HLD: Azure/SONiC#924

**What I did**
1. Add support for remote speed advertisement
2. Add support for capability checker

**Why I did it**
1. Help users identify the cause of autoneg link failure
2. Avoid un-necessary SAI calls for autoneg controls if this feature is not supported

**How I verified it**
1. Built-in sanity test suite (i.e. sonic-swss/tests)
2. System tests

Signed-off-by: Dante Su <dante.su@broadcom.com>

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1879,7 +1879,7 @@ void PortsOrch::initPortCapAutoNeg(Port &port)
     }
     else
     {
-        port.m_port_cap_an = false;
+        port.m_port_cap_an = true; /* This is for vstest */
         SWSS_LOG_NOTICE("Unable to get %s AN capability", port.m_alias.c_str());
     }
 }
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -61,6 +61,7 @@ extern string gMyAsicName;
 #define DEFAULT_VLAN_ID     1
 #define MAX_VALID_VLAN_ID   4094
 
+#define PORT_STAT_POLLING_SEC                             1
 #define PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS     1000
 #define PORT_BUFFER_DROP_STAT_POLLING_INTERVAL_MS     60000
 #define QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS   10000
@@ -298,7 +299,8 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
         m_portStateTable(stateDb, STATE_PORT_TABLE_NAME),
         port_stat_manager(PORT_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
         port_buffer_drop_stat_manager(PORT_BUFFER_DROP_STAT_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_BUFFER_DROP_STAT_POLLING_INTERVAL_MS, false),
-        queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false)
+        queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
+        m_timer(new SelectableTimer(timespec { .tv_sec = PORT_STAT_POLLING_SEC, .tv_nsec = 0 }))
 {
     SWSS_LOG_ENTER();
 
@@ -563,6 +565,10 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
 
         m_lagIdAllocator = unique_ptr<LagIdAllocator> (new LagIdAllocator(chassisAppDb));
     }
+
+    auto executor = new ExecutableTimer(m_timer, this, "PORT_STATE_POLL");
+    Orch::addExecutor(executor);
+    m_timer->start();
 }
 
 void PortsOrch::removeDefaultVlanMembers()
@@ -1860,6 +1866,24 @@ void PortsOrch::initPortSupportedSpeeds(const std::string& alias, sai_object_id_
     m_portStateTable.set(alias, v);
 }
 
+void PortsOrch::initPortCapAutoNeg(Port &port)
+{
+    sai_status_t status;
+    sai_attribute_t attr;
+
+    attr.id = SAI_PORT_ATTR_SUPPORTED_AUTO_NEG_MODE;
+    status = sai_port_api->set_port_attribute(port.m_port_id, &attr);
+    if (status == SAI_STATUS_SUCCESS)
+    {
+        port.m_port_cap_an = attr.value.booldata;
+    }
+    else
+    {
+        port.m_port_cap_an = false;
+        SWSS_LOG_NOTICE("Unable to get %s AN capability", port.m_alias.c_str());
+    }
+}
+
 /*
  * If Gearbox is enabled and this is a Gearbox port then set the attributes accordingly.
  */
@@ -2012,6 +2036,61 @@ bool PortsOrch::getPortSpeed(sai_object_id_t id, sai_uint32_t &speed)
     }
 
     return true;
+}
+
+bool PortsOrch::getPortAdvSpeeds(const Port& port, bool remote, std::vector<sai_uint32_t>& speed_list)
+{
+    sai_object_id_t port_id = port.m_port_id;
+    sai_object_id_t line_port_id;
+    sai_attribute_t attr;
+    sai_status_t status;
+    const auto size_guess = 16; // Guess the size which could be enough
+    std::vector<sai_uint32_t> speeds(size_guess);
+
+    attr.id = remote ? SAI_PORT_ATTR_REMOTE_ADVERTISED_SPEED : SAI_PORT_ATTR_ADVERTISED_SPEED;
+    attr.value.u32list.count = static_cast<uint32_t>(speeds.size());
+    attr.value.u32list.list = speeds.data();
+
+    if (getDestPortId(port_id, LINE_PORT_TYPE, line_port_id))
+    {
+        status = sai_port_api->get_port_attribute(line_port_id, 1, &attr);
+    }
+    else
+    {
+        status = sai_port_api->get_port_attribute(port_id, 1, &attr);
+    }
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        return false;
+    }
+    speeds.resize(attr.value.u32list.count);
+
+    speed_list.clear();
+    for (auto it = speeds.begin(); it != speeds.end(); ++it)
+    {
+        speed_list.push_back(*it);
+    }
+    return true;
+}
+
+bool PortsOrch::getPortAdvSpeeds(const Port& port, bool remote, string& adv_speeds)
+{
+    std::vector<sai_uint32_t> speed_list;
+    bool rc = getPortAdvSpeeds(port, remote, speed_list);
+
+    adv_speeds = "";
+    if (rc)
+    {
+        for (auto it = speed_list.begin(); it != speed_list.end(); ++it)
+        {
+            adv_speeds += std::to_string(*it);
+            if (it + 1 != speed_list.end())
+            {
+              adv_speeds += ",";
+            }
+        }
+    }
+    return rc;
 }
 
 task_process_status PortsOrch::setPortAdvSpeeds(sai_object_id_t port_id, std::vector<sai_uint32_t>& speed_list)
@@ -2350,6 +2429,9 @@ bool PortsOrch::initPort(const string &alias, const string &role, const int inde
             {
                 /* Create associated Gearbox lane mapping */
                 initGearboxPort(p);
+
+                /* Initialize port capabilities */
+                initPortCapAutoNeg(p);
 
                 /* Add port to port list */
                 m_portList[alias] = p;
@@ -2885,6 +2967,13 @@ void PortsOrch::doPortTask(Consumer &consumer)
             {
                 if (!an_str.empty())
                 {
+                    if (!p.m_port_cap_an)
+                    {
+                        SWSS_LOG_ERROR("%s: autoneg is not supported", p.m_alias.c_str());
+                        // Invalid auto negotiation mode configured, don't retry
+                        it = consumer.m_toSync.erase(it);
+                        continue;
+                    }
                     if (autoneg_mode_map.find(an_str) == autoneg_mode_map.end())
                     {
                         SWSS_LOG_ERROR("Failed to parse autoneg value: %s", an_str.c_str());
@@ -2927,6 +3016,15 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         SWSS_LOG_NOTICE("Set port %s AutoNeg from %d to %d", alias.c_str(), p.m_autoneg, an);
                         p.m_autoneg = an;
                         m_portList[alias] = p;
+                        m_portStateTable.hdel(p.m_alias, "rmt_adv_speeds");
+                        if (an > 0)
+                        {
+                            m_port_state_poll[p.m_alias] |= PORT_STATE_POLL_AN;
+                        }
+                        else
+                        {
+                            m_port_state_poll[p.m_alias] &= ~PORT_STATE_POLL_AN;
+                        }
                     }
                 }
 
@@ -5757,6 +5855,20 @@ void PortsOrch::updatePortOperStatus(Port &port, sai_port_oper_status_t status)
     if (port.m_type == Port::PHY)
     {
         updateDbPortOperStatus(port, status);
+        if (port.m_autoneg > 0)
+        {
+            if (status == SAI_PORT_OPER_STATUS_UP)
+            {
+                updatePortStateAutoNeg(port);
+                m_port_state_poll[port.m_alias] &= ~PORT_STATE_POLL_AN;
+            }
+            else
+            {
+                /* Restart autoneg state polling upon link down event */
+                m_portStateTable.hdel(port.m_alias, "rmt_adv_speeds");
+                m_port_state_poll[port.m_alias] |= PORT_STATE_POLL_AN;
+            }
+        }
     }
     port.m_oper_status = status;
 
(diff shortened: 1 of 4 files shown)


orchagent/portsorch.cpp
```diff
@@ -5666,6 +5764,20 @@ void PortsOrch::updatePortOperStatus(Port &port, sai_port_oper_status_t status)
...
     {
         updateDbPortOperStatus(port, status);
+        if (port.m_autoneg > 0)
+        {
+            if (status == SAI_PORT_OPER_STATUS_UP)
+            {
+                updatePortStateAutoNeg(port);
+                m_port_state_poll[port.m_alias] &= ~PORT_STATE_POLL_AN;
+            }
+            else
```
dgsudharsan, line 5774: Should we make sure that port is not admin_down before enabling it for polling?

orchagent/portsorch.h
```diff
@@ -329,6 +342,9 @@ class PortsOrch : public Orch, public Subject
     bool getPortOperSpeed(const Port& port, sai_uint32_t& speed) const;
     void updateDbPortOperSpeed(Port &port, sai_uint32_t speed);
 
+    map<string, uint32_t> m_port_state_poll;
```
dgsudharsan, line 345: Can this be part of Port class instead of a map here?

orchagent/portsorch.cpp
```diff
@@ -563,6 +565,10 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
 
         m_lagIdAllocator = unique_ptr<LagIdAllocator> (new LagIdAllocator(chassisAppDb));
     }
+
+    auto executor = new ExecutableTimer(m_timer, this, "PORT_STATE_POLL");
+    Orch::addExecutor(executor);
+    m_timer->start();
```
dgsudharsan, line 571: Please ensure the timer is started only on the first poll config request and stopped when there are no poll requests. Else the orchagent will get unnecessary notifications every second

orchagent/portsorch.cpp
```diff
@@ -61,6 +61,7 @@ extern string gMyAsicName;
 #define DEFAULT_VLAN_ID     1
 #define MAX_VALID_VLAN_ID   4094
 
+#define PORT_STAT_POLLING_SEC                             1
```
dgsudharsan, line 64: Why do we need an agressive 1 second timer? Can this be something like 5 or 10 seconds if it is for show command purpose alone?

dgsudharsan (changes requested):
Is there plan to add UT coverage for these code changes?

orchagent/portsorch.cpp
```diff
@@ -5666,6 +5764,20 @@ void PortsOrch::updatePortOperStatus(Port &port, sai_port_oper_status_t status)
     if (port.m_type == Port::PHY)
     {
         updateDbPortOperStatus(port, status);
+        if (port.m_autoneg > 0)
+        {
+            if (status == SAI_PORT_OPER_STATUS_UP)
+            {
+                updatePortStateAutoNeg(port);
+                m_port_state_poll[port.m_alias] &= ~PORT_STATE_POLL_AN;
```
Junchao-Mellanox, line 5772: this is an async notification from low level, how could it make sure that the adv speeds has been retrieved already?

orchagent/portsorch.cpp
```diff
@@ -6792,3 +6904,37 @@ std::unordered_set<std::string> PortsOrch::generateCounterStats(const string& ty
...
+    {
+        m_port_state_poll[port.m_alias] &= ~PORT_STATE_POLL_AN;
+    }
+}
+
+void PortsOrch::doTask(swss::SelectableTimer &timer)
+{
+    SWSS_LOG_ENTER();
+
+    for (auto it = m_portList.begin(); it != m_portList.end(); ++it)
```
Junchao-Mellanox, line 6933: why we need loop m_portList? I suppose you can directly loop m_port_state_poll

orchagent/portsorch.cpp
```diff
@@ -2900,6 +2989,15 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
                         p.m_autoneg = an;
                         m_portList[alias] = p;
+                        m_portStateTable.hdel(p.m_alias, "rmt_adv_speeds");
+                        if (an > 0)
+                        {
+                            m_port_state_poll[p.m_alias] |= PORT_STATE_POLL_AN;
+                        }
+                        else
+                        {
+                            m_port_state_poll[p.m_alias] &= ~PORT_STATE_POLL_AN;
```
Junchao-Mellanox, line 2999: How about remove the port from m_port_state_poll if it no longer requires polling? The benefit is that we can stop the timer if  m_port_state_poll  is empty.

Junchao-Mellanox (changes requested):

orchagent/portsorch.cpp
```diff
@@ -1860,6 +1866,24 @@ void PortsOrch::initPortSupportedSpeeds(const std::string& alias, sai_object_id_
     m_portStateTable.set(alias, v);
 }
 
+void PortsOrch::initPortCapAutoNeg(Port &port)
+{
+    sai_status_t status;
+    sai_attribute_t attr;
+
+    attr.id = SAI_PORT_ATTR_SUPPORTED_AUTO_NEG_MODE;
```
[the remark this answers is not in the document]
ds952811: Here is the prototype of sai_query_attribute_capability(), it's obviously that it does not support per-port basis capability query, while the autoneg feature could be unavailable for some ports. (e.g. A switch ASIC without AN/LT support on the SFP/QSFP/QSFPDD, while there are native RJ45 ports available on the platform)

i.e. The prototype of sai_query_attribute_capability()
```
sai_status_t sai_query_attribute_capability(
        _In_ sai_object_type_t object_type,
        _In_ sai_attr_id_t attr_id,
        _Out_ sai_attr_capability_t *attr_capability);
```

orchagent/portsorch.cpp
```diff
@@ -2014,6 +2038,61 @@ bool PortsOrch::getPortSpeed(sai_object_id_t id, sai_uint32_t &speed)
     return true;
 }
 
+bool PortsOrch::getPortAdvSpeeds(const Port& port, bool remote, std::vector<sai_uint32_t>& speed_list)
+{
+    sai_object_id_t port_id = port.m_port_id;
+    sai_object_id_t line_port_id;
+    sai_attribute_t attr;
+    sai_status_t status;
+    const auto size_guess = 16; // Guess the size which could be enough
```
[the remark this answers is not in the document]
ds952811: Sure, will do

orchagent/portsorch.cpp
ds952811: Yes, will do.

orchagent/portsorch.cpp
ds952811: Sure, will do

orchagent/portsorch.cpp
ds952811: Sure, no problem

orchagent/portsorch.cpp
```diff
@@ -2014,6 +2038,61 @@ bool PortsOrch::getPortSpeed(sai_object_id_t id, sai_uint32_t &speed)
...
+        status = sai_port_api->get_port_attribute(port_id, 1, &attr);
+    }
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        return false;
+    }
+    speeds.resize(attr.value.u32list.count);
+
+    speed_list.clear();
+    for (auto it = speeds.begin(); it != speeds.end(); ++it)
```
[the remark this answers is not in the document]
ds952811: Thanks, it looks great

orchagent/portsorch.cpp
```diff
@@ -6792,3 +6904,37 @@ std::unordered_set<std::string> PortsOrch::generateCounterStats(const string& ty
...
+    }
+}
+
+void PortsOrch::doTask(swss::SelectableTimer &timer)
+{
+    SWSS_LOG_ENTER();
+
+    for (auto it = m_portList.begin(); it != m_portList.end(); ++it)
+    {
+        if (m_port_state_poll[it->second.m_alias] & PORT_STATE_POLL_AN)
```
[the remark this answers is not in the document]
ds952811: Sure, will do

orchagent/portsorch.cpp
ds952811: We'll check the port admin state and autoneg config to check if this query should be skipped

orchagent/portsorch.cpp
ds952811: Sure, will do

orchagent/portsorch.cpp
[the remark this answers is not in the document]
dgsudharsan: I believe get_attribute should have been used instead of set_attribute

dgsudharsan (changes requested):

orchagent/portsorch.cpp
```diff
@@ -61,6 +61,9 @@ extern string gMyAsicName;
 #define DEFAULT_VLAN_ID     1
 #define MAX_VALID_VLAN_ID   4094
 
+#define DEFAULT_VECTOR_SIZE 16
+
+#define PORT_STAT_POLLING_SEC                             5
```
dgsudharsan, line 66: Please rename this variable as requested in the other PR

orchagent/portsorch.cpp
```diff
@@ -298,7 +301,8 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
         m_portStateTable(stateDb, STATE_PORT_TABLE_NAME),
         port_stat_manager(PORT_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
         port_buffer_drop_stat_manager(PORT_BUFFER_DROP_STAT_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_BUFFER_DROP_STAT_POLLING_INTERVAL_MS, false),
-        queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false)
+        queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
+        m_timer(new SelectableTimer(timespec { .tv_sec = PORT_STAT_POLLING_SEC, .tv_nsec = 0 }))
```
dgsudharsan, line 305: Please rename PORT_STAT to PORT_STATE

orchagent/portsorch.cpp
```diff
@@ -2900,6 +2974,8 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         SWSS_LOG_NOTICE("Set port %s AutoNeg from %d to %d", alias.c_str(), p.m_autoneg, an);
                         p.m_autoneg = an;
                         m_portList[alias] = p;
+                        m_portStateTable.hdel(p.m_alias, "rmt_adv_speeds");
+                        updatePortStatePoll(p, PORT_STATE_POLL_AN, (an > 0));
```
dgsudharsan, line 2978: Should there be an admin status check here?

orchagent/portsorch.cpp
ds952811: This is a case of a link-up event, hence the remote advertisement is already received and processed by the ASIC, and the remote advertisement is surely available at this moment.

orchagent/portsorch.cpp
```diff
@@ -2014,6 +2038,61 @@ bool PortsOrch::getPortSpeed(sai_object_id_t id, sai_uint32_t &speed)
...
+
+bool PortsOrch::getPortAdvSpeeds(const Port& port, bool remote, string& adv_speeds)
+{
+    std::vector<sai_uint32_t> speed_list;
+    bool rc = getPortAdvSpeeds(port, remote, speed_list);
+
+    adv_speeds = "";
+    if (rc)
+    {
+        for (auto it = speed_list.begin(); it != speed_list.end(); ++it)
```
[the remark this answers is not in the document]
ds952811: Done

dgsudharsan (changes requested):

orchagent/portsorch.cpp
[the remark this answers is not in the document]
ds952811: Thanks!

orchagent/portsorch.cpp
ds952811: Done

orchagent/portsorch.cpp
ds952811: Done

orchagent/portsorch.cpp
ds952811: No, the AutoNeg could be enabled prior to having the port admin up, and the port could also be admin up/down during the 5 seconds polling interval, hence it's not necessary to check the admin status here.
In the poller thread, there is always a check for the admin status.

orchagent/portsorch.cpp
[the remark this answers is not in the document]
ds952811: Done

orchagent/portsorch.cpp
```diff
@@ -2014,6 +2038,61 @@ bool PortsOrch::getPortSpeed(sai_object_id_t id, sai_uint32_t &speed)
...
+    {
+        status = sai_port_api->get_port_attribute(line_port_id, 1, &attr);
+    }
+    else
+    {
+        status = sai_port_api->get_port_attribute(port_id, 1, &attr);
+    }
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        return false;
```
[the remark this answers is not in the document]
ds952811: Done

orchagent/portsorch.cpp
ds952811: Done, the poller will skip the operation if the port is admin down

orchagent/portsorch.cpp
```diff
@@ -1860,6 +1867,24 @@ void PortsOrch::initPortSupportedSpeeds(const std::string& alias, sai_object_id_
...
+
+    attr.id = SAI_PORT_ATTR_SUPPORTED_AUTO_NEG_MODE;
+    status = sai_port_api->get_port_attribute(port.m_port_id, 1, &attr);
+    if (status == SAI_STATUS_SUCCESS)
+    {
+        port.m_cap_an = attr.value.booldata ? 1 : 0;
+    }
+    else
+    {
+        port.m_cap_an = 1; /* Default to 1 for vstest */
```
Junchao-Mellanox, line 1883: Not sure about this, it means that m_cap_an will be 1 even if get_port_attribute returns error.

orchagent/portsorch.cpp
```diff
@@ -2014,6 +2039,45 @@ bool PortsOrch::getPortSpeed(sai_object_id_t id, sai_uint32_t &speed)
...
+    {
+        SWSS_LOG_WARN("Unable to get advertised speed for %s", port.m_alias.c_str());
+        return false;
+    }
+    speeds.resize(attr.value.u32list.count);
+    speed_list.swap(speeds);
+    return true;
+}
+
+bool PortsOrch::getPortAdvSpeeds(const Port& port, bool remote, string& adv_speeds)
```
Junchao-Mellanox, line 2072: getPortRemoteAdvSpeeds?

orchagent/portsorch.cpp
```diff
@@ -6883,3 +6975,70 @@ std::unordered_set<std::string> PortsOrch::generateCounterStats(const string& ty
...
+{
+    SWSS_LOG_ENTER();
+
+    Port port;
+
+    for (auto it = m_port_state_poll.begin(); it != m_port_state_poll.end(); ++it)
+    {
+        if ((it->second == 0) || !getPort(it->first, port))
+        {
+            m_port_state_poll.erase(it);
```
Junchao-Mellanox, line 7028: might lead to undefined behavior

Junchao-Mellanox (changes requested):

orchagent/portsorch.cpp
ds952811: Yes, this is intended to avoid test failure in vstest

orchagent/portsorch.cpp
```diff
@@ -61,6 +61,9 @@ extern string gMyAsicName;
 #define DEFAULT_VLAN_ID     1
 #define MAX_VALID_VLAN_ID   4094
 
+#define DEFAULT_VECTOR_SIZE 16
```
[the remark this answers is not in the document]
ds952811: Done

orchagent/portsorch.cpp
```diff
@@ -298,7 +301,8 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
         m_portStateTable(stateDb, STATE_PORT_TABLE_NAME),
         port_stat_manager(PORT_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
         port_buffer_drop_stat_manager(PORT_BUFFER_DROP_STAT_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_BUFFER_DROP_STAT_POLLING_INTERVAL_MS, false),
-        queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false)
+        queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
+        m_timer(new SelectableTimer(timespec { .tv_sec = PORT_STATE_POLLING_SEC, .tv_nsec = 0 }))
```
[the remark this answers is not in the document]
ds952811: Done

orchagent/portsorch.cpp
ds952811: The only difference between local and remote queries is the attribute id, and this function is intentionally designed to handle both cases in the same code logic to reduce the maintenance efforts.

orchagent/portsorch.cpp
ds952811: Done

orchagent/portsorch.cpp
Junchao-Mellanox: So, what if get_port_attribute actually return some error in production environment?

ds952811: It ends up with the AN test failures, and hence the code coverage will be reduced.
An alternative solution is to extend the SAI model in the docker-sonic-vs

ds952811:
Replaced by PR #2149

PR #2216, opened 2022-04-01
Make changes to support compiling on Bullseye with GCC 10

saiarcot895:
Signed-off-by: Saikrishna Arcot <sarcot@microsoft.com>

**What I did**

Types of changes done:
* Add missing includes in header files and .cpp files
* Don't use parentheses when doing list initialization in constructors
* Make sure variables are initialized before first use

**Why I did it**

Support building swss on Bullseye, so that containers can upgrade to Bullseye.

**How I verified it**

Not tested at this time, since there's no containers being built for Bullseye yet.

**Details if related**


lguohan (approved):

orchagent/mplsrouteorch.cpp
```diff
@@ -465,7 +465,7 @@ bool RouteOrch::addLabelRoute(LabelRouteBulkContext& ctx, const NextHopGroupKey
     Label& label = ctx.label;
 
     /* next_hop_id indicates the next hop id or next hop group id of this route */
-    sai_object_id_t next_hop_id;
+    sai_object_id_t next_hop_id = SAI_NULL_OBJECT_ID;
```
qiluo-msft, line 468: Seems a valid bug fix. If true, let's backport to other branches. And better to have a separate PR just for it.

PR #2225, opened 2022-04-06
[BFD]Retry create BFD with different source UDP port on failure

dgsudharsan:
**What I did**
The create_bfd_session API can fail when a source port already used by a different application is passed to SAI create. To avoid failure, a retry logic is added to try with a different port number three times before failing

**Why I did it**
To avoid BFD failure and orchagent crash due to port conflict.

**How I verified it**
Verified it manually by having an application binding the port and checking if retry is done when BFD tries the same port

**Details if related**

### DIFF
--- a/orchagent/bfdorch.cpp
+++ b/orchagent/bfdorch.cpp
@@ -422,6 +422,7 @@ bool BfdOrch::create_bfd_session(const string& key, const vector<FieldValueTuple
     {
         status = retry_create_bfd_session(bfd_session_id, attrs);
     }
+
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to create bfd session %s, rv:%d", key.c_str(), status);
@@ -453,8 +454,8 @@ void BfdOrch::update_port_number(vector<sai_attribute_t> &attrs)
        {
            auto old_num = attrs[attr_idx].value.u32;
            attrs[attr_idx].value.u32 = bfd_src_port();
-           SWSS_LOG_INFO ("BFD create using port number %d failed. Retrying with port number %d",
-                          old_num, attrs[attr_idx].value.u32);
+           SWSS_LOG_WARN("BFD create using port number %d failed. Retrying with port number %d",
+                         old_num, attrs[attr_idx].value.u32);
            return;
        }
     }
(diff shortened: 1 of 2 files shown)


orchagent/bfdorch.cpp
```diff
@@ -410,12 +430,24 @@ bool BfdOrch::create_bfd_session(const string& key, const vector<FieldValueTuple
...
+        attr_idx++;
     }
 
     fvVector.emplace_back("state", session_state_lookup.at(SAI_BFD_SESSION_STATE_DOWN));
 
     sai_object_id_t bfd_session_id = SAI_NULL_OBJECT_ID;
-    sai_status_t status = sai_bfd_api->create_bfd_session(&bfd_session_id, gSwitchId, (uint32_t)attrs.size(), attrs.data());
+    sai_status_t status = sai_bfd_api->create_bfd_session(&bfd_session_id, gSwitchId, attr_idx, attrs.data());
+
+    for (int retry = 0; (status != SAI_STATUS_SUCCESS) && (retry < NUM_BFD_SRCPORT_RETRIES); retry++)
```
prsunny, line 441: This is a corner case scenario. I think we should handle it in a separate function and invoke if status is Failure.  You can loop the attributes to find the src port attr index.  Managing index by attr_id++ is also error prone when someone adds a new attribute.

dgsudharsan: Updated

orchagent/bfdorch.cpp
```diff
@@ -439,6 +446,38 @@ bool BfdOrch::create_bfd_session(const string& key, const vector<FieldValueTuple
...
+sai_status_t BfdOrch::retry_create_bfd_session(sai_object_id_t &bfd_session_id, vector<sai_attribute_t> attrs)
+{
+    sai_status_t status = SAI_STATUS_FAILURE;
+
+    for (int retry = 0; retry < NUM_BFD_SRCPORT_RETRIES; retry++)
+    {
+        update_port_number(attrs);
+        status = sai_bfd_api->create_bfd_session(&bfd_session_id, gSwitchId,
+                                                 (uint32_t)attrs.size(), attrs.data());
+        if(status == SAI_STATUS_SUCCESS)
```
prsunny, line 473: space after if

prsunny (commented):
are you planning to add any VS test. code coverage _may_ fail if the additional code is not covered by test

orchagent/bfdorch.cpp
dgsudharsan: Updated

liat-grozovik:
LGTM. small comment
@prsunny  adding a code coverage for that will requires mock and not regular functional test. we need to make sure the port is used by other application and then try to create bfd. Do you still think adding a mock in the VS test is the right way? it is not unit test.

prsunny (approved):

prsunny:
@judyjoseph , can you please cherry-pick to 202111?

PR #2235, opened 2022-04-19
Fix issues in dynamic buffer in rare scenario after "config qos clear"

stephenxs:
1. Do not generate dynamic lossless buffer profiles in case default_dynamic_th is not ready.
2. For non-dynamic lossless buffer profiles
  - record threshold mode from CONFIG_DB
  - do not apply if the threshold mode doesn't aligh with buffer pool
3. Remove an item from m_toSync on failure

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


stephenxs:
Close it as it's the duplicate of #2234

PR #2238, opened 2022-04-22
[ACL]Avoid incrementing crm count when ACL rule create fails

dgsudharsan:
**What I did**
Avoid increment CRM counter when ACL rule creation fails. In existing code there is no return statement after failure during ACL rule creation which results in CRM getting incremented. Since the ACL rule recreation is retried this counter increments in a loop.
 
**Why I did it**
Added return statement in case of a failure.

**How I verified it**
Existing UT should verify the changes

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -846,10 +846,12 @@ bool AclRule::createRule()
                 m_id.c_str(), status);
         AclRange::remove(range_objects, range_object_list.count);
         decreaseNextHopRefCount();
-        return false;
     }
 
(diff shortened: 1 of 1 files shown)


liat-grozovik (approved):

liat-grozovik:
@prsunny could you please help to signoff?

liat-grozovik:
@dgsudharsan unrelated to this PR can you please check if we have such problem with other CRM counters? @volodymyrsamotiy  FYI

prsunny (approved):

PR #2239, opened 2022-04-25
Removing Vnet with scope default

sumukhatv:
**What I did**
If VirtualRouter ID is the default value, then don' remove it.  
**Why I did it**
When removing Vnet with scope `default`, we need not remove the Vrf object if it is of default value.
**How I verified it**
Added tests to verify it
**Details if related**

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -309,7 +309,7 @@ VNetVrfObject::~VNetVrfObject()
     set<sai_object_id_t> vr_ent = getVRids();
     for (auto it : vr_ent)
     {
-        if it != gVirtualRouterId {
+        if (it != gVirtualRouterId) {
             sai_status_t status = sai_virtual_router_api->remove_virtual_router(it);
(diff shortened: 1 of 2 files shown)


orchagent/vnetorch.cpp
```diff
@@ -309,11 +309,13 @@ VNetVrfObject::~VNetVrfObject()
     set<sai_object_id_t> vr_ent = getVRids();
     for (auto it : vr_ent)
     {
-        sai_status_t status = sai_virtual_router_api->remove_virtual_router(it);
-        if (status != SAI_STATUS_SUCCESS)
-        {
-            SWSS_LOG_ERROR("Failed to remove virtual router name: %s, rv:%d",
-                            vnet_name_.c_str(), status);
+        if (it != gVirtualRouterId) {
```
prsunny, line 312: {-> new line.

prsunny (commented):
Please add VS tests.

dgsudharsan:
@sumukhatv @prsunny This is also needed for 202111 branch. Can you please cherrypick?

PR #2240, opened 2022-04-26
Change DBConnector("CONFIG_DB") to const type

isabelmsft:
**What I did**
Change DBConnector("CONFIG_DB") to const type
**Why I did it**
Ensure that C++ only has read only operations to CONFIG_DB.
Ensuring C++ has no write operations to CONFIG_DB helps define the project scope for YANG config db validation, so we can ensure that all relevant write operations to CONFIG_DB are in python
**How I verified it**

**Details if related**


prsunny:
@isabelmsft , please provide some more details on "why I did". What was breaking, what would be the impact etc

PR #2244, opened 2022-04-28
[Vnet] Set BFD multihop to true for Vnet routes

prsunny:
**What I did**
Set multihop to true for BFD sessions created by Vnetorch

**Why I did it**
As per HLD

**How I verified it**

**Details if related**


theasianpianist (approved):

prsunny:
The code coverage is manually analyzed and failing likely due to compiler optimization. The code is indeed covered by the VS test.

PR #2249, opened 2022-05-02
[muxorch] Adding stateCurrent API to mux state machine

Ndancejic:
**What I did**

Added stateCurrent API to mux state machine, which allows active - active and standby - standby state change.

**How I verified it**

Tested in test_mux.py by switching from active to active and standby to standby and checking logs.

**Why I did it**

active-active and standby-standby state changes were throwing errors

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -1539,6 +1539,8 @@ void VNetRouteOrch::createBfdSession(const string& vnet, const NextHopKey& endpo
         FieldValueTuple fvTuple("local_addr", src_ip.to_string());
         data.push_back(fvTuple);
 
+	data.emplace_back("multihop", "true");
+
         bfd_session_producer_.set(key, data);
 
         bfd_sessions_[monitor_addr].bfd_state = SAI_BFD_SESSION_STATE_DOWN;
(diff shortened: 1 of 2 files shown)


orchagent/muxorch.cpp
```diff
@@ -431,6 +438,12 @@ bool MuxCable::stateStandby()
     return true;
 }
 
+bool MuxCable::stateCurrent()
```
prsunny, line 441: We don't have to go through metric update steps etc from line 468 for this case. I'm not thinking may be a simple check to see if existing and new state is the same, log as NOTICE and return. You've the test to cover that.

neighsyncd/neighsync.cpp
```diff
@@ -79,6 +79,16 @@ void NeighSync::onMsg(int nlmsg_type, struct nl_object *obj)
     key+= ":";
 
     nl_addr2str(rtnl_neigh_get_dst(neigh), ipStr, MAX_ADDR_SIZE);
+
+    /* Ignore IPv4 link-local addresses as neighbors */
```
prsunny, line 83: Can you please create a new PR for this? The title and description doesn't match with changes.

Ndancejic:
created a new pull request here: https://github.com/Azure/sonic-swss/pull/2280

PR #2257, opened 2022-05-06
Don't initialize P4Orch code when running on a Broadcom platform

saiarcot895:
Signed-off-by: Saikrishna Arcot <sarcot@microsoft.com>

**What I did**

The P4Orch initializer assumes that UDF is available on the platform,
which isn't available on Broadcom platforms. Therefore, for Broadcom,
don't initialize P4Orch.

**Why I did it**

On some platforms (such as Broadcom), UDF is not available, and trying to program it results in an error. This doesn't break anything during coldboot, but fails warmboot.

**How I verified it**

Tested on virtual switch and Broadcom.

**Details if related**

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -621,9 +621,11 @@ bool OrchDaemon::init()
 
     m_orchList.push_back(&CounterCheckOrch::getInstance(m_configDb));
 
-    vector<string> p4rt_tables = {APP_P4RT_TABLE_NAME};
-    gP4Orch = new P4Orch(m_applDb, p4rt_tables, vrf_orch, gCoppOrch);
-    m_orchList.push_back(gP4Orch);
+    if (platform == P4_PLATFORM_SUBSTRING) {
+        vector<string> p4rt_tables = {APP_P4RT_TABLE_NAME};
+        gP4Orch = new P4Orch(m_applDb, p4rt_tables, vrf_orch, gCoppOrch);
+        m_orchList.push_back(gP4Orch);
+    }
 
     if (WarmStart::isWarmStart())
     {
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -621,7 +621,7 @@ bool OrchDaemon::init()
 
     m_orchList.push_back(&CounterCheckOrch::getInstance(m_configDb));
 
-    if (platform == P4_PLATFORM_SUBSTRING)
+    if (platform != BRCM_PLATFORM_SUBSTRING)
     {
         vector<string> p4rt_tables = {APP_P4RT_TABLE_NAME};
         gP4Orch = new P4Orch(m_applDb, p4rt_tables, vrf_orch, gCoppOrch);
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -621,7 +621,8 @@ bool OrchDaemon::init()
 
     m_orchList.push_back(&CounterCheckOrch::getInstance(m_configDb));
 
-    if (platform == P4_PLATFORM_SUBSTRING) {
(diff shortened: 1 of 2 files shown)


saiarcot895:
@donNewtonAlpha FYI

orchagent/orchdaemon.cpp
```diff
@@ -621,9 +621,11 @@ bool OrchDaemon::init()
 
     m_orchList.push_back(&CounterCheckOrch::getInstance(m_configDb));
 
-    vector<string> p4rt_tables = {APP_P4RT_TABLE_NAME};
-    gP4Orch = new P4Orch(m_applDb, p4rt_tables, vrf_orch, gCoppOrch);
-    m_orchList.push_back(gP4Orch);
+    if (platform == P4_PLATFORM_SUBSTRING) {
```
prsunny, line 624: Please move `{` to new line.

saiarcot895: Done.

orchagent/orchdaemon.cpp
```diff
@@ -621,9 +621,12 @@ bool OrchDaemon::init()
 
     m_orchList.push_back(&CounterCheckOrch::getInstance(m_configDb));
 
-    vector<string> p4rt_tables = {APP_P4RT_TABLE_NAME};
-    gP4Orch = new P4Orch(m_applDb, p4rt_tables, vrf_orch, gCoppOrch);
-    m_orchList.push_back(gP4Orch);
+    if (platform == P4_PLATFORM_SUBSTRING)
```
baxia-lan, line 624: The p4orch should be able to initialized successfully on platform=broadcom switches. And the default UDF match created in p4orch during init is calling standard SAI API(https://github.com/Azure/sonic-swss/blob/eff725ff1803f7e4948a25e53abc68886e43b367/orchagent/p4orch/acl_table_manager.cpp#L40)

baxia-lan:
Hi @saiarcot895, could you please attach the details about the issues that you are seeing when init p4orch during warmreboot on broadcom?

saiarcot895:
Azure/sonic-buildimage#10410 has more details on this, but it appears that the Broadcom SAI library doesn't support UDF. During coldboot, there's an error message, but it's not treated as a fatal error. During warmboot, however, failing an ASIC operation is considered a fatal error.

bocon13 (changes requested):
Thanks for the PR, @saiarcot895

I understand you are having trouble with warm-boot and we should work with you on this.

The `p4` platform isn't related to the P4Orch; that target is an older software switch. The P4Orch is designed for many different platforms, including Broadcom.

cc: @mint570

saiarcot895:
Hi @bocon13, I've changed the conditional check to not initialize if running on a Broadcom platform. This will keep it available on other platforms.

Would this change be fine?

bocon13:
Hi @saiarcot895, the Broadcom platform is one of our main platforms that we aim to support with the P4Orch.

Can you provide some error logs (or excerpts) that we can take a look at? It would be good to work together to address the underlying issue.

saiarcot895:
Azure/sonic-buildimage#10410 has some output from syslog. Is there something more specific that you need?

mint570:
I believe we can do the UDF match initialization in the first entry instead of in the constructor.
This should address this particular issue.

(However, p4orch does not support warmboot yet. So warmboot will not work if we have programmed some P4RT tables. But it should work with the fix if there is no P4RT table.)

@baxia-lan, what do you think?

saiarcot895:
@baxia-lan meaning it would still do the UDF match initialization on warm reboot, even though nothing may be using it?

baxia-lan:
@saiarcot895  no, only if there is a ACL table created with UDF fields via p4orch, the UDF match creation will be called. I am preparing https://github.com/Azure/sonic-swss/pull/2282 to address this issue. Working on test coverage now

saiarcot895:
Closing in favor of #2282.

PR #2266, opened 2022-05-11
[unit test] Add unit test for FlowCounterRouteOrch

keboliu:
Signed-off-by: kebol@nvidia.com

**What I did**
Add unit test framework for FlowCounterRouteOrch
Add unit test case to cover the route pattern add/delete flow

**Why I did it**
The unit test framework is not available for FlowCounterRouteOrch which will block the developers to add UT to cover the code change in the future. So add the unit test framework for it. Add also add a test to cover the route pattern add/delete flow. 

**How I verified it**
run unit test and all of the new test can pass.

**Details if related**

### DIFF
--- a/tests/mock_tests/flowcounterrouteorch_ut.cpp
+++ b/tests/mock_tests/flowcounterrouteorch_ut.cpp
@@ -1,0 +1,361 @@
+#define private public
+#include "directory.h"
+#undef private
+#define protected public
+#include "orch.h"
+#undef protected
+#include "ut_helper.h"
+#include "mock_orchagent_main.h"
+#include "mock_table.h"
+
+extern string gMySwitchType;
+
+namespace flowcounterrouteorch_test
+{
+    using namespace std;
+    shared_ptr<swss::DBConnector> m_app_db;
+    shared_ptr<swss::DBConnector> m_config_db;
+    shared_ptr<swss::DBConnector> m_state_db;
+    shared_ptr<swss::DBConnector> m_chassis_app_db;
+
+    int num_created_counter;
+    sai_counter_api_t ut_sai_counter_api;
+    sai_counter_api_t *pold_sai_counter_api;
+    sai_create_counter_fn old_create_counter;
+    sai_remove_counter_fn old_remove_counter;
+
+    sai_status_t _ut_stub_create_counter(
+        _Out_ sai_object_id_t *counter_id, 
+	    _In_ sai_object_id_t switch_id, 
+	    _In_ uint32_t attr_count, 
+	    _In_ const sai_attribute_t *attr_list)
+    {
+        num_created_counter ++;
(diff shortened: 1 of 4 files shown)


Junchao-Mellanox (approved):

liat-grozovik:
@keboliu  well done on adding more tests to swss.

tests/mock_tests/flowcounterrouteorch_ut.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+            ASSERT_EQ(gNeighOrch, nullptr);
+            gNeighOrch = new NeighOrch(m_app_db.get(), APP_NEIGH_TABLE_NAME, gIntfsOrch, gFdbOrch, gPortsOrch, m_chassis_app_db.get());
+
+            auto* tunnel_decap_orch = new TunnelDecapOrch(m_app_db.get(), APP_TUNNEL_DECAP_TABLE_NAME);
+            vector<string> mux_tables = {
+                CFG_MUX_CABLE_TABLE_NAME,
+                CFG_PEER_SWITCH_TABLE_NAME
+            };
+            auto* mux_orch = new MuxOrch(m_config_db.get(), mux_tables, tunnel_decap_orch, gNeighOrch, gFdbOrch);
```
prsunny, line 183: In unit test, do we need to initialize those orch that are not being used? Suggest remove all other orchs that are not really required.

keboliu: @prsunny actually NeighOrch has a dependency on MuxOrch, see below code. W/O MuxOrch we will face segment fault when running the test. 
https://github.com/Azure/sonic-swss/blob/18c73a196704d7feb740fc13b15242a8aa30c5dc/orchagent/neighorch.cpp#L809
```
bool NeighOrch::addNeighbor(const NeighborEntry &neighborEntry, const MacAddress &macAddress)
{
    SWSS_LOG_ENTER();

    ...

    MuxOrch* mux_orch = gDirectory.get<MuxOrch*>();
    bool hw_config = isHwConfigured(neighborEntry);
   ...
}
```
The orch which are initialized here, if not directly needed, they are required by some other orch which are directly needed.

prsunny: so do you need neighorch in this?

keboliu: Yes, in order to successfully add a route entry, we need to add a neighbor entry beforehand.

prsunny: Thanks for clarifying. I assume all orchs are relevant here.

prsunny (approved):

Junchao-Mellanox (approved):

PR #2280, opened 2022-05-17
[muxorch] Adding case for maintaining current state

Ndancejic:
What I did:
Added a case where if current and previous Mux states
are the same, post a log entry that acknowleges this
change and returns without doing anything.

Why I did it:
Valid state changes such as active-active and
standby-standby were causing error logs

How I tested it:
added a case to check active-active state change
and standby-standby state change.

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -443,13 +443,19 @@ void MuxCable::setState(string new_state)
     new_state = muxStateValToString.at(ns);
 
     auto it = muxStateTransition.find(make_pair(state_, ns));
-
     if (it ==  muxStateTransition.end())
     {
         // Update HW Mux cable state anyways
         mux_cb_orch_->updateMuxState(mux_name_, new_state);
-        SWSS_LOG_ERROR("State transition from %s to %s is not-handled ",
-                        muxStateValToString.at(state_).c_str(), new_state.c_str());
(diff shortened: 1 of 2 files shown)


orchagent/muxorch.cpp
```diff
@@ -433,6 +433,12 @@ bool MuxCable::stateStandby()
 
 void MuxCable::setState(string new_state)
 {
+    if (strcmp(new_state.c_str(), muxStateValToString.at(state_).c_str()))
```
prsunny, line 436: For this, we still need to update the HW MuX cable state at line 455. So can you move this to 456 and else part have the error log

prsunny (dismissed):

prsunny:
@Ndancejic , seems like there are some VS failures due to the change:
```
test_mux.py::TestMuxTunnel::test_Tunnel PASSED                           [ 50%]
test_mux.py::TestMuxTunnel::test_Peer PASSED                             [ 50%]
test_mux.py::TestMuxTunnel::test_Neighbor FAILED                         [ 51%]
test_mux.py::TestMuxTunnel::test_Fdb FAILED                              [ 51%]
test_mux.py::TestMuxTunnel::test_Route FAILED                            [ 51%]
test_mux.py::TestMuxTunnel::test_acl FAILED                              [ 51%]
```

prsunny (approved):

PR #2286, opened 2022-05-23
[teammgr]: Waiting MACsec ready before doLagMemberTask

Pterosaur:
Signed-off-by: Ze Gan <ganze718@gmail.com>, Judy Joseph <jujoseph@microsoft.com>

**What I did**
If a member of portchannel has macsec profile attached in config, enable MACsec on the port before it's been added as a member of portchannel.

**Why I did it**
Due to some hardware limitation, cannot enable MACsec on a member of portchannel.  
So we enable the macsec on interface first and then add it as part of portchannel.

**Note: This is a work around which will be removed when h/w supports it future releases.**

The approach taken in this PR is 

> In the teamdMgr when an interface is added as part of the LAG, we wait for the macsecPort creation done in SAI and Ingress SA creation complete (if macsec is enabled on the interface)
> The above takes care of config reload, reboot scenario's where we cannot guarantee the sequence of macsec attach to interface, add interface as part of portchannel. 

If we do a manual removal of port from portchannel, or remove macsec config from the interface, Please follow this steps 
> First remove the portchannel member out of portchannel 
> Remove the macsec profile attached to interface.

**How I verified it**

Verified with config reload, reboot with the macsec profile attached to portchannel member interfaces.
Verified case when SAK rekey is enabled on macsec on portchannel members
Verified case when member interface link flaps 

**Details if related**

### DIFF
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -33,7 +33,8 @@ TeamMgr::TeamMgr(DBConnector *confDb, DBConnector *applDb, DBConnector *statDb,
     m_appPortTable(applDb, APP_PORT_TABLE_NAME),
     m_appLagTable(applDb, APP_LAG_TABLE_NAME),
     m_statePortTable(statDb, STATE_PORT_TABLE_NAME),
-    m_stateLagTable(statDb, STATE_LAG_TABLE_NAME)
+    m_stateLagTable(statDb, STATE_LAG_TABLE_NAME),
+    m_stateMACsecPortTable(statDb, STATE_MACSEC_PORT_TABLE_NAME)
 {
     SWSS_LOG_ENTER();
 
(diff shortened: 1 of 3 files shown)


cfgmgr/teammgr.cpp
```diff
@@ -98,6 +99,43 @@ bool TeamMgr::isLagStateOk(const string &alias)
     return true;
 }
 
+bool TeamMgr::isMACsecSetted(const std::string &port)
```
judyjoseph, line 102: May be change name to isMacsecEnabled

cfgmgr/teammgr.cpp
```diff
@@ -310,6 +348,12 @@ void TeamMgr::doLagMemberTask(Consumer &consumer)
                 continue;
             }
 
+            if (isMACsecSetted(member) && !isMACsecStateOk(member))
```
judyjoseph, line 351: Here we need to check if the macsec feature is enabled also. 

As there could be a possibility of "macsec profile" is attached to  PORT in config db, but the macsec feature is not enabled and the macsec dockers are not started. 

This would result in us waiting in isMACsecStateOk(member).

cfgmgr/teammgr.cpp
Pterosaur: This name was in my mind before. But I think `isMACsecEnabled` is more like that MACsec feature has been enabled on the port. But I just want to check whether the MACsec would like to be set on this port. So, I check the config DB instead of state DB in this function.

cfgmgr/teammgr.cpp
Pterosaur: Make sense, will add it.

Pterosaur: Done

cfgmgr/teammgr.cpp
```diff
@@ -400,6 +450,14 @@ void TeamMgr::doPortUpdateTask(Consumer &consumer)
             string lag;
             if (findPortMaster(lag, alias))
             {
+                if (isMACsecAttached(alias) && !isMACsecIngressSAOk(alias))
+                {
+                    it++;
+                    SWSS_LOG_INFO("MACsec is NOT ready on the port %s", alias.c_str());
+                    continue;
+                }
+                SWSS_LOG_NOTICE("MACsec is ready on the port %s", alias.c_str());
```
prsunny, line 459: Please remove this log as you already have the same in `isMACsecIngressSAOk`

Pterosaur:
image
image

The coverage tool has a false negative to this PR. Because I truly got the "missing(declared by coverage)" line from syslog. I think The most function have been covered by our test case.

@prsunny  Could you please help to merge this PR?

Pterosaur:
image
This is another evidence that I find in the log from log/test_inband_intf_mgmt_vrf/log/syslog (https://dev.azure.com/mssonic/build/_build/results?buildId=113863&view=artifacts&pathAsName=false&type=publishedArtifacts)

judyjoseph (approved):
Validated this logic is applied only on ports with macsec configured..

prsunny:
> Validated this logic is applied only on ports with macsec configured.. @prsunny to comment - the coverage details added by Ze looks not in sync with logs

Agree with Ze and Judy that this is a false negative. @theasianpianist for viz

Issue #2231, opened 2022-04-16
Don't query buffer profile attributes before APPLY_VIEW


kcudnik:
Change required to not query SAI_QUEUE_ATTR_BUFFER_PROFILE_ID  and SAI_INGRESS_PRIORITY_GROUP_ATTR_BUFFER_PROFILE  before APPLY_VIEW, because in warm boot scenario is bringing previous buffer profile into OA scope, and this is causing extra buffer profile to be created in syncd.

Issue in depth is described here: https://github.com/Azure/sonic-sairedis/issues/1030#issuecomment-1094077178

kcudnik:
@prsunny ping

prsunny:
I see the original issue refers to zero-buffer pools.  A related change is here - https://github.com/Azure/sonic-swss/pull/2304. @neethajohn , @stephenxs , can you check if its related?

stephenxs:
> I see the original issue refers to zero-buffer pools. A related change is here - #2304. @neethajohn , @stephenxs , can you check if its related?

I’ll check it. Originally the purpose was to share the zero buffer profile by PFCWD with the existing one. Now that we will not apply zero buffer profile on ingress side, this change is not required

svsivm:
> #2304

Hi, I doubt if this problem has anything to do with #2304. #2231 is seen specifically when warmboot is done while PFC storm is ongoing. After warmboot we see 4 zero buffer profiles instead of the expected and that throws things into disarray.

svsivm:
> > I see the original issue refers to zero-buffer pools. A related change is here - #2304. @neethajohn , @stephenxs , can you check if its related?
> 
> I’ll check it. Originally the purpose was to share the zero buffer profile by PFCWD with the existing one. Now that we will not apply zero buffer profile on ingress side, this change is not required

Hi Stephen, can you please look at #2319. Why were the zero buffer profile attribute values changed as part of #2185? Thanks.

stephenxs:
> > > I see the original issue refers to zero-buffer pools. A related change is here - #2304. @neethajohn , @stephenxs , can you check if its related?
> > 
> > 
> > I’ll check it. Originally the purpose was to share the zero buffer profile by PFCWD with the existing one. Now that we will not apply zero buffer profile on ingress side, this change is not required
> 
> Hi Stephen, can you please look at #2319. Why were the zero buffer profile attribute values changed as part of #2185? Thanks.

Hi I'm working on analyzing the issue. Will let you know once I have an answer

PR #2304, opened 2022-06-01
[PFC_WD] Avoid applying ZeroBuffer Profiles to ingress PG when a PFC storm is detected

vivekreddynv:
**What I did**

Avoid dropping traffic that is ingressing the port/pg that is in storm. The code changes in this PR avoid creating the ingress zero pool and profile and does not attach any zero profile to the ingress pg when pfcwd is triggered

Revert changes related to https://github.com/Azure/sonic-swss/pull/1480 where the retry mechanism was added to BufferOrch which caches the task retries and while the PG is locked by PfcWdZeroBufferHandler. 

Revert changes related to https://github.com/Azure/sonic-swss/pull/2164 in PfcWdZeroBufferHandler & ZeroBufferProfile & BufferOrch. 

Updated UT's accordingly

**How I verified it**

UT's.

Ran the sonic-mgmt test with these changes https://github.com/Azure/sonic-mgmt/pull/5665 and verified if they've passed. 

**Details if related**


stephenxs (dismissed):

orchagent/pfcactionhandler.cpp
```diff
@@ -696,37 +676,28 @@ void PfcWdZeroBufferHandler::ZeroBufferProfile::createZeroBufferProfile()
...
-            gSwitchId,
-            static_cast<uint32_t>(attribs.size()),
-            attribs.data());
-        if (status != SAI_STATUS_SUCCESS)
-        {
-            SWSS_LOG_ERROR("Failed to create dynamic zero buffer pool for PFC WD: %d", status);
-            return;
-        }
+    attr.id = SAI_BUFFER_POOL_ATTR_THRESHOLD_MODE;
+    attr.value.u32 = SAI_BUFFER_POOL_THRESHOLD_MODE_STATIC;
```
stephenxs, line 689: Let's use dynamic mode and dynamic_th -8 for creating profile. this is what was used before
L 689
```
    attr.value.u32 = SAI_BUFFER_POOL_THRESHOLD_MODE_DYNAMIC;
```
L 711
```
    attr.value.u32 = SAI_BUFFER_PROFILE_THRESHOLD_MODE_DYNAMIC;
```
L 718~719
```
    attr.id = SAI_BUFFER_PROFILE_ATTR_SHARED_DYNAMIC_TH;
    attr.value.s8 = 0;
```

orchagent/pfcactionhandler.cpp
```diff
@@ -708,14 +708,14 @@ void PfcWdZeroBufferHandler::ZeroBufferProfile::createZeroBufferProfile()
...
+    attr.value.u32 = SAI_BUFFER_PROFILE_THRESHOLD_MODE_DYNAMIC;
     attribs.push_back(attr);
 
     attr.id = SAI_BUFFER_PROFILE_ATTR_BUFFER_SIZE;
     attr.value.u64 = 0;
     attribs.push_back(attr);
 
-    attr.id = SAI_BUFFER_PROFILE_ATTR_SHARED_STATIC_TH;
+    attr.id = SAI_BUFFER_PROFILE_ATTR_SHARED_DYNAMIC_TH;
     attr.value.s8 = 0;
```
stephenxs, line 719: dynamic th should `-8` which stands for ALPHA_0

stephenxs (approved):

liat-grozovik:
@neethajohn  kindly reminder to review and signoff

orchagent/bufferorch.cpp
```diff
@@ -48,11 +48,7 @@ BufferOrch::BufferOrch(DBConnector *applDb, DBConnector *confDb, DBConnector *st
     m_flexCounterTable(new ProducerTable(m_flexCounterDb.get(), FLEX_COUNTER_TABLE)),
     m_flexCounterGroupTable(new ProducerTable(m_flexCounterDb.get(), FLEX_COUNTER_GROUP_TABLE)),
     m_countersDb(new DBConnector("COUNTERS_DB", 0)),
-    m_stateBufferMaximumValueTable(stateDb, STATE_BUFFER_MAXIMUM_VALUE_TABLE),
-    m_ingressZeroBufferPool(SAI_NULL_OBJECT_ID),
-    m_egressZeroBufferPool(SAI_NULL_OBJECT_ID),
-    m_ingressZeroPoolRefCount(0),
-    m_egressZeroPoolRefCount(0)
```
neethajohn, line 55: Is there any reason why we don't want use the same egress zero pool for pfcwd and reclaim buffer? @stephenxs

stephenxs: yes @neethajohn 
1. Currently, Only [ingress zero buffer pool](https://github.com/Azure/sonic-buildimage/blob/master/platform/mellanox/zero_profiles.j2#L3) is required for reclaiming PG. For the queues, we do not need an egress zero buffer pool because the zero profiles are applied based on normal egress buffer pools. Now that we will no longer apply zero buffer profile on the ingress side, the mechanism to share the zero buffer between reclaiming buffer and PFC WD isn't really required.
2. There is a logic in PFC WD to query the currently applied buffer profile on PG/queue from SAI before applying zero buffer profile. The motivation is to restore the original buffer profile once PFC storm has gone. However, this mechanism introduces an [issue#2231](https://github.com/Azure/sonic-swss/issues/2231) on another vendor's platform while upgrading the switch from an old version to a new one via warm reboot. This is because  PFC storm can happen before APPLY_VIEW during warm reboot when `GET` operation is prohibitted.  There has been a WA in sairedis to make `GET` works in this scenario. However, the zero buffer pool/profile that was applied by PFC WD differs from that created for reclaiming buffer, which fails the sairedis WA. (It is still very strange that the mechanism had existed for years but the issue was reported only last year). 

So we decide to revert it for now.
If we still want this mechanism to share zero buffer between reclaiming buffer and PFC WD, we may want to enhance PFC WD and buffer orch a bit more:
- buffer orch to store the buffer profiles currently applied on the queue/PG
- PFC WD not to query SAI for the currently applied buffer profiles

Let me know in case it is not clear to you

neethajohn: Thanks Stephen for the detailed explanation!

neethajohn (approved):

PR #2305, opened 2022-06-02
Port configuration incremental update support

Junchao-Mellanox:
HLD: https://github.com/Azure/SONiC/pull/985

**What I did**

1. portsyncd no longer handle port configuration change and put it to APP DB
2. Implement incremental configuration change in portmgr
3. Adjust portsorch to meet incremental configuration change requirement 

**Why I did it**

1. portsyncd and portmgrd both handles port configuration change, this causes unnecessary handling in ports orch
```
I set the mtu to 1500 and search the log:
 
admin@sonic:~$ sudo config interface mtu Ethernet0 1500                                                                                                                                                                            
admin@sonic:~$ sudo cat /var/log/syslog | grep portsyncd
 
 
Log shows:
 
May 31 07:41:25.275792 sonic NOTICE swss#portsyncd: :- handlePortConfig: Configure port Ethernet0 attribute admin_status value to up                                                                                               
May 31 07:41:25.275792 sonic NOTICE swss#portsyncd: :- handlePortConfig: Configure port Ethernet0 attribute alias value to etp1                                                                                                    
May 31 07:41:25.275792 sonic NOTICE swss#portsyncd: :- handlePortConfig: Configure port Ethernet0 attribute index value to 1                                                                                                       
May 31 07:41:25.275792 sonic NOTICE swss#portsyncd: :- handlePortConfig: Configure port Ethernet0 attribute lanes value to 0                                                                                                       
May 31 07:41:25.275792 sonic NOTICE swss#portsyncd: :- handlePortConfig: Configure port Ethernet0 attribute mtu value to 1500                                                                                                      
May 31 07:41:25.275792 sonic NOTICE swss#portsyncd: :- handlePortConfig: Configure port Ethernet0 attribute speed value to 25000
```

2. portmgrd shall do increment configuration update

**How I verified it**

1. manual test
2. regression test
3. added new unit test case

**Details if related**


cfgmgr/portmgr.cpp
```diff
@@ -117,14 +81,14 @@ void PortMgr::doTask(Consumer &consumer)
...
-            {
-                SWSS_LOG_INFO("Port %s is not ready, pending...", alias.c_str());
-                it++;
-                continue;
-            }
+            /* portOk=true indicates that the port has been created in kernel.
+             * We should not call any ip command if portOk=false. However, it is
+             * valid to put port configuration to APP DB which will trigger port creation in kernel.
+             */
+            bool portOk = isPortStateOk(alias);
```
prsunny, line 88: I dont think we need to change this? portsyncd does the `handlePortConfigFromConfigDB` . So portmgrd can still wait and keep the current flow.

Junchao-Mellanox: Hi @prsunny , `handlePortConfigFromConfigDB` is only called once at init stage. It cannot handle case like dynamic port breakout. So, I suppose we need keep this change. What do you think?

prsunny: ok. so the logic here retries for some part of the configuration (kernel), but write to app_db in the first place.

Junchao-Mellanox: yes

cfgmgr/portmgr.cpp
```diff
@@ -157,31 +121,57 @@ void PortMgr::doTask(Consumer &consumer)
...
             {
-                setPortTpid(alias, tpid);
+                writeConfigToAppDb(alias, "tpid", tpid);
                 SWSS_LOG_NOTICE("Configure %s TPID to %s", alias.c_str(), tpid.c_str());
             }
+
+            for (auto &entry : field_values)
+            {
+                writeConfigToAppDb(alias, fvField(entry), fvValue(entry));
+                SWSS_LOG_NOTICE("Configure %s %s to %s", alias.c_str(), fvField(entry).c_str(), fvValue(entry).c_str());
```
prsunny, line 173: i'm confused by this. It seems like we are anyways writing all the fields to APP DB, so why individual writes above like mtu, tpid?

cfgmgr/portmgr.cpp
prsunny: My concern is, developer has to make sure kernel writes cannot be attempted if portOk is false which is different from the previous logic.  Secondly, it keeps re-writing app_db during the wait period. We need to revisit this logic

Junchao-Mellanox: Thanks for your comment, any suggestion for this?

cfgmgr/portmgr.cpp
Junchao-Mellanox: I will fix this

cfgmgr/portmgr.cpp
prsunny: how about doing this in the `if case ` and keep the original logic. 

```
for (auto &entry : field_values)
            {
                writeConfigToAppDb(alias, fvField(entry), fvValue(entry));
```

Junchao-Mellanox: First let me confirm your solution:
```cpp
if (!isPortStateOk(alias))
{
    SWSS_LOG_INFO("Port %s is not ready, pending...", alias.c_str());
    for (auto& i : kfvFieldsValues(t))
    {
        writeConfigToAppDb(alias, fvField(i), fvValue(i));
    }
    it++;
    continue;
}

// rest of the code keeps as original one
```

This solution makes code much clear and simple. But It will repeatedly put the configuration to APP DB until the port state turns "ok".

prsunny: But its the same with your current implementation as well, as you are looping through all retry attributes and writing to app_db, correct?

Junchao-Mellanox: no, for the retry process there are 3 cases:
1. port turns ok, mtu / admin_status will be configured into kernel, other fields will not be put into APP DB again
2. port is still not ok and mtu/admin_status were changed by user, it will only put the new mtu / admin_status to APP DB
3. port is still not ok and mtu / admin_status were not changed, nothing will be put into APP DB

Junchao-Mellanox:
Hi @prsunny , could you please review and sign-off? The checker failure is a known issue which is not related to my change.

cfgmgr/portmgr.cpp
```diff
@@ -149,38 +118,42 @@ void PortMgr::doTask(Consumer &consumer)
...
-                setPortAdminStatus(alias, admin_status == "up");
-                SWSS_LOG_NOTICE("Configure %s admin status to %s", alias.c_str(), admin_status.c_str());
+                SWSS_LOG_INFO("Port %s is not ready, pending...", alias.c_str());
+
+                writeConfigToAppDb(alias, "mtu", mtu);
+                writeConfigToAppDb(alias, "admin_status", admin_status);
+                field_values.clear();
+                field_values.emplace_back("mtu", mtu);
+                field_values.emplace_back("admin_status", admin_status);
+                it->second = KeyOpFieldsValuesTuple{alias, SET_COMMAND, field_values};
```
prsunny, line 142: Can you confirm if in the retry process, any other fields (learnmode or mtu) are modified by the user, it gets reflected?

prsunny (approved):
lgtm

prsunny:
@Junchao-Mellanox , the VS test failure is tracke by https://github.com/Azure/sonic-swss/issues/2365. Please retry once this is fixed.

cfgmgr/portmgr.cpp
Junchao-Mellanox: Yes, this is covered by unit test case "ConfigureDuringRetry".

PR #2307, opened 2022-06-02
Add support for IP interface loopback action

liorghub:
**What I did**
Add support for IP interface loopback action.

**Why I did it**
New feature in SONiC.

**How I verified it**
New VS tests were added.

**Details if related**

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -504,13 +504,14 @@ set<IpPrefix> IntfsOrch:: getSubnetRoutes()
     return subnet_routes;
 }
 
-bool IntfsOrch::setIntf(const string& alias, sai_object_id_t vrf_id, const IpPrefix *ip_prefix, const bool adminUp, const uint32_t mtu, loopback_action_e loopbackAction)
+bool IntfsOrch::setIntf(const string& alias, sai_object_id_t vrf_id, const IpPrefix *ip_prefix,
+                        const bool adminUp, const uint32_t mtu, loopback_action_e loopbackAction)
 
 {
     SWSS_LOG_ENTER();
+
     Port port;
     gPortsOrch->getPort(alias, port);
-
     port.m_loopback_action = loopbackAction;
 
     auto it_intfs = m_syncdIntfses.find(alias);
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -57,18 +57,11 @@ static const vector<sai_router_interface_stat_t> rifStatIds =
     SAI_ROUTER_INTERFACE_STAT_OUT_ERROR_OCTETS,
 };
 
-// Translation of loopback action from string to sai type
-const unordered_map<string, loopback_action_e> IntfsOrch::m_loopback_action_map =
-{
-    {"drop",    LOOPBACK_ACTION_DROP},
-    {"forward", LOOPBACK_ACTION_FORWARD},
-};
-
 // Translation of loopback action from sonic to sai type
-const unordered_map<loopback_action_e, sai_packet_action_t> IntfsOrch::m_sai_loopback_action_map =
+const unordered_map<string, sai_packet_action_t> IntfsOrch::m_sai_loopback_action_map =
 {
-    {LOOPBACK_ACTION_DROP, SAI_PACKET_ACTION_DROP},
-    {LOOPBACK_ACTION_FORWARD, SAI_PACKET_ACTION_FORWARD},
+    {"drop", SAI_PACKET_ACTION_DROP},
+    {"forward", SAI_PACKET_ACTION_FORWARD},
 };
 
 IntfsOrch::IntfsOrch(DBConnector *db, string tableName, VRFOrch *vrf_orch, DBConnector *chassisAppDb) :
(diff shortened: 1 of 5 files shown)


orchagent/intfsorch.cpp
```diff
@@ -416,6 +430,62 @@ bool IntfsOrch::setIntfProxyArp(const string &alias, const string &proxy_arp)
     return true;
 }
 
+bool IntfsOrch::setIntfLoopbackAction(const Port &port)
```
prsunny, line 433: if you pass loopback action as a string to this function, no need for `getIntfLoopbackActionStr`

tests/test_interface.py
```diff
@@ -2193,6 +2196,81 @@ def test_VLanInterfaceIpv6LinkLocalOnly(self, dvs, testlog):
...
+        tbl.set(interface, fvs)
+        time.sleep(1)
+
+    def loopback_action_test(self, iface):
+        # create interface
+        self.create_l3_intf(iface, "")
+
+        # set interface loopback action in config database
+        action_map = {"drop": "SAI_PACKET_ACTION_DROP", "forward": "SAI_PACKET_ACTION_FORWARD"}
+        action = random.choice(list(action_map.keys()))
```
prsunny, line 2222: we dont prefer to have randomness in test. Please have coverage for both actions.

prsunny (commented):
IMO, the code can be simplified. Please let me know if you prefer an offline discussion.

orchagent/port.h
```diff
@@ -107,6 +114,7 @@ class Port
     }
 
     std::string         m_alias;
+    loopback_action_e   m_loopback_action;
```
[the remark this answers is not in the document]
liorghub: @prsunny I tried to follow your approach to remove from cache but I found out that we need it in cache, I will explain why. The cache is used not only for skipping multiple user configuration but also for skipping set flow after create flow. There are 2 flows for setting loopback action, set and create. Set flow is done when we set loopback action on an existing IP interface (func setIntfLoopbackAction). Create flow is done when loopback action set is part of the IP interface creation, this happens in init (func setIntf).  Both functions I mentioned are being called one after the other and the way to skip set flow after create flow took place, is using the cache. This is the approach taken for all IP interface attributes. I did however made the code less complicated by using string instead of enum for the action type ("drop" or "forward").

orchagent/intfsorch.cpp
```diff
@@ -1067,6 +1159,13 @@ bool IntfsOrch::addRouterIntfs(sai_object_id_t vrf_id, Port &port)
     attr.value.oid = vrf_id;
     attrs.push_back(attr);
 
+    if(port.m_loopback_action != LOOPBACK_ACTION_NONE)
```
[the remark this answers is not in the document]
liorghub: Done.

orchagent/intfsorch.cpp
```diff
@@ -906,6 +988,16 @@ void IntfsOrch::doTask(Consumer &consumer)
                         setRouterIntfsMpls(port);
                         gPortsOrch->setPort(alias, port);
                     }
+
+                    /* Set loopback action */
+                    if ((loopbackAction != LOOPBACK_ACTION_NONE) and (port.m_loopback_action != loopbackAction))
+                    {
+                        port.m_loopback_action = loopbackAction;
+                        if(setIntfLoopbackAction(port))
```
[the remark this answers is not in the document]
liorghub: Done.

orchagent/intfsorch.cpp
liorghub: Done.

tests/test_interface.py
liorghub: Done.

liorghub:
The fowling test failed:
/usr/local/lib/python3.9/dist-packages/health_checker/service_checker.py

Trying to rerun.

liorghub:
Test p4rt/test_p4rt_acl.py::TestP4RTAcl::test_AclRuleAddWithoutTableDefinitionFails ERROR failed, trying to rerun.

liorghub:
Test p4rt/test_p4rt_acl.py::TestP4RTAcl::test_AclRuleAddWithoutTableDefinitionFails ERROR failed again, trying to rerun.

orchagent/intfsorch.cpp
```diff
@@ -57,6 +57,13 @@ static const vector<sai_router_interface_stat_t> rifStatIds =
     SAI_ROUTER_INTERFACE_STAT_OUT_ERROR_OCTETS,
 };
 
+// Translation of loopback action from sonic to sai type
+const unordered_map<string, sai_packet_action_t> IntfsOrch::m_sai_loopback_action_map =
```
prsunny, line 61: You can have it a local map variable.. Also please follow the existing naming convention in the file. No need to have it as part of `IntfsOrch `class

orchagent/port.h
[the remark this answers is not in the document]
prsunny: I tend to disagree. Understand your point on create and set. But what is the concern if the same value is set again. For mtu and admin state, it is done currently for sub-port and those values are anyways part of  port class. Loopback setting is rare and I read its mostly as setting proxy_arp etc, for which there is no cache. This is a single event and lets get this simplified.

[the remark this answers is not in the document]
liorghub: Done.

orchagent/intfsorch.cpp
liorghub: Done.

prsunny (approved):

PR #2308, opened 2022-06-02
Purge package sonic-db-cli which depends on libswsscommon

qiluo-msft:
**What I did**
Purge package sonic-db-cli which depends on libswsscommon

**Why I did it**
Since sonic-db-cli depends on libswsscommon, we could not simply only purge libswsscommon, so we purge both together.

The build error is as below
```
Step 8/22 : RUN dpkg --purge libswsscommon
 ---> Running in 0297ee02869f
dpkg: dependency problems prevent removal of libswsscommon:
 sonic-db-cli depends on libswsscommon.
```

**How I verified it**

**Details if related**

### DIFF
--- a/.azure-pipelines/docker-sonic-vs/Dockerfile
+++ b/.azure-pipelines/docker-sonic-vs/Dockerfile
@@ -4,14 +4,7 @@ ARG docker_container_name
 
 ADD ["debs", "/debs"]
 
-RUN dpkg --purge python-swsscommon
-RUN dpkg --purge python3-swsscommon
-RUN dpkg --purge swss
-RUN dpkg --purge libsairedis
-RUN dpkg --purge libswsscommon
-RUN dpkg --purge libsaimetadata
-RUN dpkg --purge libsaivs
-RUN dpkg --purge syncd-vs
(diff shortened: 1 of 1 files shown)


yxieca (approved):

lguohan (approved):

theasianpianist:
Do we also need to re-install sonic-db-cli with the other packages so that it's available in the DVS?

lguohan:
i guess we should. but now sure if vstest is using sonic-db-cli?

theasianpianist:
> i guess we should. but now sure if vstest is using sonic-db-cli?

Don't think the tests are using it, but it may be useful for manual debugging if someone needs to download a DVS image from AZP

prsunny:
Following VS test failures on WB is being analyzed and not caused by this change. 

```
test_warm_reboot.py::TestWarmReboot::test_swss_neighbor_syncup FAILED    [ 96%]
test_warm_reboot.py::TestWarmReboot::test_OrchagentWarmRestartReadyCheck FAILED [ 96%]
test_warm_reboot.py::TestWarmReboot::test_swss_port_state_syncup FAILED  [ 96%]
test_warm_reboot.py::TestWarmReboot::test_routing_WarmRestart FAILED     [ 96%]
```
Inorder to unblock the rest of the PRs, need to force-merge this.

PR #2312, opened 2022-06-07
[sonic-swss] : PFCWD recovery changes using DLR_INIT

vmittal-msft:
**What I did**
This is changing PFCWD recovery from SONIC based to SAI/SDK DLR_INIT way.

**Why I did it**
Existing SONIC scheme will not work for upcoming BRCM devices starting with TH3.

**How I verified it**
Verified this to be working fine on HW platform where the issue was seen. Added sonic-mgmt test case to handle this scenario.

**Details if related**


prsunny:
You may want to raise a PR for master.

PR #2314, opened 2022-06-07
Apply `DSCP_TO_TC_MAP` from `PORT_QOS_MAP|global` to switch level

bingwang-ms:
Signed-off-by: bingwang <wang.bing@microsoft.com>

**What I did**
This PR is to update the code for applying switch level `DSCP_TO_TC_MAP`.
After PR https://github.com/Azure/sonic-buildimage/pull/10565, there will be two DSCP_TO_TC_MAP

- DSCP_TO_TC_MAP|AZURE is the default map, which is used at port level and switch level
- DSCP_TO_TC_MAP|AZURE_TUNNEL is used to remap the priority of tunnel traffic in dualtor deployment

To address the issue, an entry `PORT_QOS_MAP|global` will be added into `config_db`
```
"PORT_QOS_MAP": {
        "global": {
            "dscp_to_tc_map": "AZURE"
        }
}
```
The entry will be consumed by `qosorch`, and the specified map will be applied to switch.

**Why I did it**
This change is to ensure the correct `DSCP_TO_TC_MAP` is applied to switch level.

**How I verified it**
Verified by a new test case `test_dscp_to_tc_map_applied_to_switch`
```
collected 9 items                                                                                                                                                                                     

test_qos_map.py::TestDot1p::test_dot1p_cfg PASSED                                                                                                                                               [ 11%]
test_qos_map.py::TestDot1p::test_port_dot1p PASSED                                                                                                                                              [ 22%]
test_qos_map.py::TestCbf::test_dscp_to_fc PASSED                                                                                                                                                [ 33%]
test_qos_map.py::TestCbf::test_exp_to_fc PASSED                                                                                                                                                 [ 44%]
test_qos_map.py::TestCbf::test_per_port_cbf_binding PASSED                                                                                                                                      [ 55%]
test_qos_map.py::TestMplsTc::test_mpls_tc_cfg PASSED                                                                                                                                            [ 66%]
test_qos_map.py::TestMplsTc::test_port_mpls_tc PASSED                                                                                                                                           [ 77%]
test_qos_map.py::TestDscpToTcMap::test_dscp_to_tc_map_applied_to_switch PASSED                                                                                                                  [ 88%]
test_qos_map.py::test_nonflaky_dummy PASSED                                                                                                                                                     [100%]
```
**Details if related**

### DIFF
--- a/cfgmgr/buffermgr.cpp
+++ b/cfgmgr/buffermgr.cpp
@@ -15,6 +15,8 @@
 using namespace std;
 using namespace swss;
 
+#define PORT_NAME_GLOBAL "global"
+
 BufferMgr::BufferMgr(DBConnector *cfgDb, DBConnector *applDb, string pg_lookup_file, const vector<string> &tableNames) :
         Orch(cfgDb, tableNames),
         m_cfgPortTable(cfgDb, CFG_PORT_TABLE_NAME),
@@ -391,6 +393,11 @@ void BufferMgr::doPortQosTableTask(Consumer &consumer)
     {
         KeyOpFieldsValuesTuple tuple = it->second;
         string port_name = kfvKey(tuple);
+        if (port_name == PORT_NAME_GLOBAL)
+        {
+            // Ignore the entry for global level 
+            continue;
+        }
         string op = kfvOp(tuple);
         if (op == SET_COMMAND)
         {
--- a/cfgmgr/buffermgr.cpp
+++ b/cfgmgr/buffermgr.cpp
@@ -395,7 +395,8 @@ void BufferMgr::doPortQosTableTask(Consumer &consumer)
         string port_name = kfvKey(tuple);
         if (port_name == PORT_NAME_GLOBAL)
         {
-            // Ignore the entry for global level 
+            // Ignore the entry for global level
+            it = consumer.m_toSync.erase(it);
             continue;
         }
         string op = kfvOp(tuple);
--- a/orchagent/qosorch.cpp
+++ b/orchagent/qosorch.cpp
@@ -104,7 +104,7 @@ map<string, string> qos_to_ref_table_map = {
 #define DSCP_MAX_VAL 63
 #define EXP_MAX_VAL 7
 
-#define DEFAULT_DSCP_TO_TC_MAP_NAME "AZURE"
+#define PORT_NAME_GLOBAL "global"
 
 task_process_status QosMapHandler::processWorkItem(Consumer& consumer, KeyOpFieldsValuesTuple &tuple)
 {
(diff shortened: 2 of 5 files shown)


bingwang-ms:
> Hi It definitely works and resolves the issue to hardcode it for now but I don’t tend to think it is the best way to hardcode the name of the global DSCP_TO_TC_MAP. We'd better not assume that all SONiC customers will have AZURE as their default item name. Currently, `AZURE` does not exist in swss except for the test codes. I think maybe a formal way is to introduce a table in CONFIG_DB and designate the global DSCP_TO_TC_MAP there. If the info does not exist, we can choose
> 
> * to apply any DSCP_TO_TC_MAP once it is configured to the system, for the purpose of backward compatibility
> * not to apply any DSCP_TO_TC_MAP globally
> 
> I would prefer the second option because by doing so the global DSCP_TO_TC_MAP won't be applied to vendors who doesn't need it. Db_migrator can be leveraged for backward caompatibility. Regarding the table, we can
> 
> * either introduce a new table to indicate which DSCP_TO_TC_MAP to apply globally
> * or introduce a new key in `PORT_QOS_MAP`, like `PORT_QOS_MAP|global`.
> 
> Stephen

Thanks @stephenxs  for the great suggestions.
I think option 2 **introduce a new key in `PORT_QOS_MAP`, like `PORT_QOS_MAP|global`.** would be better.
I checked the code in `qosorch`, this special case can be handled by adding some logic in `handlePortQosMapTable` to handle the `DSCP_TO_TC_MAP` for `global`. 
And also, we need to update YANG to support entry whose name is not a port name.

orchagent/qosorch.cpp
```diff
@@ -148,6 +150,16 @@ task_process_status QosMapHandler::processWorkItem(Consumer& consumer, KeyOpFiel
                 freeAttribResources(attributes);
                 return task_process_status::task_failed;
             }
+            if ((qos_map_type_name == CFG_DSCP_TO_TC_MAP_TABLE_NAME) && (qos_object_name == DEFAULT_DSCP_TO_TC_MAP_NAME))
```
stephenxs, line 153: Can you please remove it from the switch attribute when `DSCP_TO_TC_MAP|AZURE` is being removed?

cfgmgr/buffermgr.cpp
```diff
@@ -391,6 +393,11 @@ void BufferMgr::doPortQosTableTask(Consumer &consumer)
     {
         KeyOpFieldsValuesTuple tuple = it->second;
         string port_name = kfvKey(tuple);
+        if (port_name == PORT_NAME_GLOBAL)
+        {
+            // Ignore the entry for global level 
+            continue;
```
stephenxs, line 399: Better to erase `it` from `m_toSync` otherwise this sentence will be hit every time.

orchagent/qosorch.cpp
```diff
@@ -303,23 +272,9 @@ bool DscpToTcMapHandler::removeQosItem(sai_object_id_t sai_object)
 
     if (sai_object == gQosOrch->m_globalDscpToTcMap)
```
stephenxs, line 273: The entire `if` statement can be removed because it is guaranteed that no reference to the map being removed once the flow reaches here.
`removeQosItem` is called from `QosMapHandler::processWorkItem` where the reference is checked before calling.
Originally, we did not have a CONFIG_DB item to track the reference of the global map so we had to check it here.
Now we do have it in CONFIG_DB and the reference is tracked so no need to check it anymore.

cfgmgr/buffermgr.cpp
bingwang-ms: Good catch. Thanks

orchagent/qosorch.cpp
bingwang-ms: Good suggestion. I remove the variable now. Thanks

orchagent/qosorch.cpp
```diff
@@ -1717,12 +1672,90 @@ task_process_status QosOrch::handleQueueTable(Consumer& consumer, KeyOpFieldsVal
...
+    SWSS_LOG_NOTICE("Applied DSCP_TO_TC QoS map to switch successfully");
+    return true;
+}
+
+task_process_status QosOrch::handleGlobalQosMap(const string &OP, KeyOpFieldsValuesTuple &tuple)
+{
+    task_process_status task_status = task_process_status::task_success;
+    if (OP == DEL_COMMAND)
+    {
+        // Ignore the DEL operation for switch level mapping
```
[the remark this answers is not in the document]
bingwang-ms: Updated. Thanks

neethajohn (approved):

PR #2317, opened 2022-06-08
[QoS] Support dynamic headroom calculation for Barefoot platforms

mmysle:
Signed-off-by: Michał Mysłek <michal.myslek@intel.com>

**What I did**
Adding the dynamic headroom calculation support for Barefoot platforms.

**Why I did it**
Enabling dynamic mode for barefoot case.

**How I verified it**
The community tests are adjusted and pass.

**Details if related**

### DIFF
--- a/cfgmgr/buffer_check_headroom_barefoot.lua
+++ b/cfgmgr/buffer_check_headroom_barefoot.lua
@@ -1,0 +1,17 @@
+-- KEYS - port name
+
+local port = KEYS[1]
+local state_db = "6"
+local ret = {}
+
+-- Connect to STATE_DB
+redis.call('SELECT', state_db)
+local max_headroom_size = tonumber(redis.call('HGET', 'BUFFER_MAX_PARAM_TABLE|' .. port, 'max_headroom_size'))
+if max_headroom_size == 0 or max_headroom_size == nil then
+    table.insert(ret, "result:true")
+    table.insert(ret, "debug:No need to check port headroom limit as shared headroom pool model is supported.")
+else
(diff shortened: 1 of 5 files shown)


cfgmgr/buffer_check_headroom_barefoot.lua
```diff
@@ -0,0 +1,17 @@
+-- KEYS - port name
+
+local port = KEYS[1]
+local state_db = "6"
+local ret = {}
+
+-- Connect to STATE_DB
+redis.call('SELECT', state_db)
+local max_headroom_size = tonumber(redis.call('HGET', 'BUFFER_MAX_PARAM_TABLE|' .. port, 'max_headroom_size'))
+if max_headroom_size == 0 or max_headroom_size == nil then
```
stephenxs, line 10: FYI.
`BUFFER_MAX_PARAM_TABLE|<port>|max_headroom_size` is the maximum accumulative headroom that can be applied to a port. It will always be exposed to `STATE_DB` during orchagent initialization as long as being supported by the vendor SAI. It won't be changed after initialization. It is not relevant to whether the shared headroom pool is enabled.
So In case the shared headroom pool is a must on Barefoot's platforms, I would like to suggest enabling it in `buffers_config.j2` in `DEFAULT_LOSSLESS_BUFFER_PARAMETER` and always returning true here.

mmysle:
@stephenxs, thank you for the review! The comments will be addressed here: https://github.com/Azure/sonic-swss/pull/2323
Sorry for the confusion.

PR #2328, opened 2022-06-14
[202012] [cherry-pick] Apply `DSCP_TO_TC_MAP` from `PORT_QOS_MAP|global` to switch level

bingwang-ms:
**What I did**
This PR is to cherry-pick https://github.com/Azure/sonic-swss/pull/2314 to `202012` branch after resolving conflicts.

This PR is to update the code for applying switch level `DSCP_TO_TC_MAP`.
After PR https://github.com/Azure/sonic-buildimage/pull/10565, there will be two DSCP_TO_TC_MAP

- DSCP_TO_TC_MAP|AZURE is the default map, which is used at port level and switch level
- DSCP_TO_TC_MAP|AZURE_TUNNEL is used to remap the priority of tunnel traffic in dualtor deployment

To address the issue, an entry `PORT_QOS_MAP|global` will be added into `config_db`
```
"PORT_QOS_MAP": {
        "global": {
            "dscp_to_tc_map": "[DSCP_TO_TC_MAP|AZURE]"
        }
}
```
The entry will be consumed by `qosorch`, and the specified map will be applied to switch.

**Why I did it**
This change is to ensure the correct `DSCP_TO_TC_MAP` is applied to switch level.

**How I verified it**
Verified by a new test case `test_dscp_to_tc_map_applied_to_switch`
```
collected 2 items                                                                                                                                                                                     

test_qos_map.py::TestDscpToTcMap::test_dscp_to_tc_map_applied_to_switch PASSED                                                                                                                  [ 50%]
test_qos_map.py::test_nonflaky_dummy PASSED                                                                                                                                                     [100%]
```
**Details if related**

### DIFF
--- a/orchagent/qosorch.cpp
+++ b/orchagent/qosorch.cpp
@@ -77,6 +77,8 @@ type_map QosOrch::m_qos_maps = {
 #define DSCP_MAX_VAL 63
 #define EXP_MAX_VAL 7
 
+#define PORT_NAME_GLOBAL "global"
+
 task_process_status QosMapHandler::processWorkItem(Consumer& consumer)
 {
     SWSS_LOG_ENTER();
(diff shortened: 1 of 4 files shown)


orchagent/qosorch.cpp
```diff
@@ -1436,6 +1422,93 @@ task_process_status QosOrch::ResolveMapAndApplyToPort(
...
+    {
+        SWSS_LOG_ERROR("Failed to apply DSCP_TO_TC QoS map to switch rv:%d", status);
+        return false;
+    }
+
+    SWSS_LOG_NOTICE("Applied DSCP_TO_TC QoS map to switch successfully");
+    return true;
+}
+
+task_process_status QosOrch::handleGlobalQosMap(const string &OP, KeyOpFieldsValuesTuple &tuple)
```
neethajohn, line 1453: Why are we not setting/removing obj reference in this PR?

tests/test_qos_map.py
```diff
@@ -105,6 +105,75 @@ def test_port_dot1p(self, dvs):
...
+            else:
+                for id in self.asic_qos_map_ids:
+                    fvs = self.asic_db.get_entry(self.ASIC_QOS_MAP_STR, id)
+                    if fvs.get("SAI_QOS_MAP_ATTR_TYPE") == "SAI_QOS_MAP_TYPE_DSCP_TO_TC":
+                        dscp_to_tc_map_id = id
+                        break
+            switch_oid = dvs.getSwitchOid()
+            # Check switch level DSCP_TO_TC_MAP doesn't before PORT_QOS_MAP|global is created
+            fvs = self.asic_db.get_entry(self.ASIC_SWITCH_STR, switch_oid)
+            assert("SAI_SWITCH_ATTR_QOS_DSCP_TO_TC_MAP" not in fvs)
```
neethajohn, line 157: remove this

orchagent/qosorch.cpp
bingwang-ms: It's because we are not setting/removing obj reference when handling the QOS map for port. Even I added obj reference for global map, the count is incorrect. That's why I didn't include that logic.

tests/test_qos_map.py
```diff
@@ -23,7 +23,7 @@
 CFG_PORT_TABLE_NAME = "PORT"
 
 
-class TestDot1p(object):
+class aaTestDot1p(object):
```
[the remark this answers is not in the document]
bingwang-ms: Fixed. Thanks

tests/test_qos_map.py
bingwang-ms: Removed. Thanks

neethajohn (approved):

PR #2332, opened 2022-06-15
Fix for "orchagent crashed when trying to delete fdb static entry with swssconfig #11046"

anilkpan:
Root cause:
When fdb flush was issued after adding a static mac, the mac was deleted due to the flush. Then when we try to remove the static mac, it thrown error and crashed due to the mac not found.

Fix:
Updated fdb flush to flush only dynamic mac.


orchagent/fdborch.cpp
```diff
@@ -1056,6 +1061,11 @@ void FdbOrch::flushFDBEntries(sai_object_id_t bridge_port_oid,
         attr.value.oid = vlan_oid;
         attrs.push_back(attr);
     }
+    
+    /* do not flush static mac */
+    attr.id = SAI_FDB_FLUSH_ATTR_ENTRY_TYPE;
```
gechiang, line 1066: With this change added here it means for any "FDB flush" operations whether it is based on port, or VLAN it will never flush out Static MACs.  Not sure if this may impact MLAG operations where they may want to flush out remotely learnt MAC that were programmed as Static MAC?  Adding MLAG owner @Praveen-Brcm to the review to ensure this is ok.

gechiang (approved):
Looks good to me.

anilkpan:
@qiluo-msft, will create the PR on 202012 this week.

anilkpan:
@qiluo-msft, created PR https://github.com/Azure/sonic-swss/pull/2374

liuh-80:
202012 branch PR merged: https://github.com/sonic-net/sonic-swss/pull/2374

PR #2335, opened 2022-06-16
[orchagent]: srv6orch support for uSID

svshah-intel:
uSID (microsid) support in swss/orchagent
-	uSID “action” mapping to end-point behavior and end-point behavior-flavor
-	uSID actions: uN, uA, uDX and uDT

Current srv6orch already has appropriate infrastructure in place. Only missing piece, to support uSID, is actions mapping to end-point behavior and behavior-flavor. This commit is to add that. Related SAI support also already is in place.

swss/tests
-	Add a test-case for uSID "action" in test_srv6.py

Signed-off-by: shitanshu.shah@intel.com

### DIFF
--- a/orchagent/srv6orch.cpp
+++ b/orchagent/srv6orch.cpp
@@ -20,6 +20,38 @@ extern sai_next_hop_api_t* sai_next_hop_api;
 extern RouteOrch *gRouteOrch;
 extern CrmOrch *gCrmOrch;
 
+const map<string, sai_my_sid_entry_endpoint_behavior_t> end_behavior_map =
+{
+    {"end",                SAI_MY_SID_ENTRY_ENDPOINT_BEHAVIOR_E},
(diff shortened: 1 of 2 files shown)


orchagent/srv6orch.cpp
```diff
@@ -423,9 +423,39 @@ bool Srv6Orch::sidEntryEndpointBehavior(string action, sai_my_sid_entry_endpoint
     {
         end_behavior = SAI_MY_SID_ENTRY_ENDPOINT_BEHAVIOR_B6_INSERT_RED;
     }
+    else if (action == "udx6")
```
prsunny, line 426: This would be better aligned in a 'switch' case construct.

svshah-intel: please review if new restructured code is better

prsunny (approved):
lgtm

reshmaintel:
@prsunny Hi Prince, thank you for the quick review.
Including @shishao7sxm 
Thanks

PR #2336, opened 2022-06-17
[swssconfig] Optimize performance of swssconfig

Junchao-Mellanox:
**What I did**

Optimize swssconfig:

- Use unix socket
- Cache producer table to avoid create it for same table name

**Why I did it**

We found that generating large scale static routes via swssconfig is very slow. 

**How I verified it**

After the optimization, generating 100K routes via swssconfig take 2 seconds, however, before the optimization it takes > 60 seconds.

**Details if related**


swssconfig/swssconfig.cpp
```diff
@@ -39,9 +43,28 @@ void dump_db_item(KeyOpFieldsValuesTuple &db_item)
...
 
+ProducerStateTable &get_producer_table(const std::string &table_name)
+{
+    auto iter = table_map.find(table_name);
+    if (iter != table_map.end())
+    {
+        return iter->second;
+    }
+
+    auto ret = table_map.emplace(table_name, ProducerStateTable(&pipeline, table_name, true));
```
qiluo-msft, line 54: >ProducerStateTable [](http://example.com/codeflow?start=45&length=18)

You may check this example: [https://github.com/Azure/sonic-swss-common/blob/92f54111172138a3c1b2ed2dfbdda23e84d9ae9d/common/dbinterface.cpp#L326](https://github.com/Azure/sonic-swss-common/blob/92f54111172138a3c1b2ed2dfbdda23e84d9ae9d/common/dbinterface.cpp#L326)

Junchao-Mellanox: Hi, not sure I understand the comment. What do you want me to check?

swssconfig/swssconfig.cpp
```diff
@@ -39,9 +43,28 @@ void dump_db_item(KeyOpFieldsValuesTuple &db_item)
     SWSS_LOG_DEBUG("]");
 }
 
+ProducerStateTable &get_producer_table(const std::string &table_name)
```
qiluo-msft, line 46: >get_producer_table [](http://example.com/codeflow?start=20&length=18)

This is a common pattern "find and insert". You may achieve it by one lookup. [https://stackoverflow.com/a/1409465/2514803](https://stackoverflow.com/a/1409465/2514803)

Junchao-Mellanox: sure, I will fix

swssconfig/swssconfig.cpp
```diff
@@ -50,23 +73,27 @@ bool write_db_data(vector<KeyOpFieldsValuesTuple> &db_items)
...
             producer.del(key_name, DEL_COMMAND);
         else
         {
+            flush_db_data(); // flush existing valid data to DB to keep logic the same as before
             SWSS_LOG_ERROR("Invalid operation: %s\n", kfvOp(db_item).c_str());
             return false;
         }
     }
+
+    flush_db_data();
```
qiluo-msft, line 96: >flush_db_data [](http://example.com/codeflow?start=4&length=13)

I see 3 `flush_db_data()`. Is it the same just calling once outside?

Junchao-Mellanox: Here, I want to keep the logic as same as before. Before the change, all data will be put into redis DB until a failure occurs. So, here I call `flush_db_data()` at 3 places (2 return false, 1 return true).

qiluo-msft:
Could you fix LGTM alert?

swssconfig/swssconfig.cpp
qiluo-msft: I see code improved s you must already check it :)

swssconfig/swssconfig.cpp
qiluo-msft: Understand your point. The pattern here is a python `finally` like syntax to achieve cleanup if out of scope. C++ has some best practice like RAII to achieve this.

Actually in this case, if you limit `table_map` scope to this function, not as a global function. The dtor will automatically flush all pipelines.

Junchao-Mellanox: I suppose I understand your point. RedisPipleLine will flush all data in its dtor, so if pipeline is defined in this function, it will automatically do flush. I will make this change.

Junchao-Mellanox:
It seems the checker is running a test case which is newer than my base code, I need to do a merge now.

qiluo-msft (approved):

PR #2340, opened 2022-06-21
[POC] Read buffer config from both static config DB and config DB

liuh-80:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/cfgmgr/buffer_headroom_mellanox.lua
+++ b/cfgmgr/buffer_headroom_mellanox.lua
@@ -32,6 +32,7 @@ local is_8lane = (ARGV[5] == "8")
 local appl_db = "0"
 local config_db = "4"
 local state_db = "6"
+local static_config_db = "15"
 
 local ret = {}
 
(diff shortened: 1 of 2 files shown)


stephenxs:
hi
can you explain a bit more about the motivation of this PR?
thanks

liuh-80:
> hi can you explain a bit more about the motivation of this PR? thanks

This still a POC PR, not ready for review, description updated.

cfgmgr/buffer_headroom_mellanox.lua
```diff
@@ -98,7 +99,14 @@ local default_lossless_param_keys = redis.call('KEYS', 'DEFAULT_LOSSLESS_BUFFER_
 local over_subscribe_ratio = tonumber(redis.call('HGET', default_lossless_param_keys[1], 'over_subscribe_ratio'))
 
 -- Fetch the shared headroom pool size
-local shp_size = tonumber(redis.call('HGET', 'BUFFER_POOL|ingress_lossless_pool', 'xoff'))
+-- POC, HUA: Fetch from both config DB and static config DB.
+local shp_size_val = redis.call('HGET', 'BUFFER_POOL|ingress_lossless_pool', 'xoff')
+if not shp_size_val then
```
stephenxs, line 104: There can be two cases which fails hget:
1. The entry doesn’t exist, we can fallback to profile db in this case 
2. The entry exists but no xoff field. In this case, we do not need to check profile db, right?

Or each entry must have all the fields so the 2nd case can’t happen?
Maybe we discussed it before, just to confirm

stephenxs (commented):
Just FYI. I updated this part for better handling lossy pools and profiles

Issue #2330, opened 2022-06-14
Pytest runs out of memory when coverage is enabled


mint570:
This is found in https://github.com/Azure/sonic-swss/pull/2237. The last few pytest failed.
We suspect that the cause is the coverage enabling: https://github.com/Azure/sonic-swss/blob/bf4d89082ad413ef71eb40a8ff2f279b87f1b002/.azure-pipelines/test-docker-sonic-vs-template.yml#L72

The "--keeptb" flag in pytest will keep the docker containers alive after the test finishes. In local experiments, each test will open two containers. After all the tests finish, there will be hundreds of containers open. If we add more pytest tests, the workflow will run out of memory.

We should try to run one test at a time. Collect the coverage artifacts, kill the containers. And then begin the next test.

mint570:
@pettershao-ragilenetworks 
@donNewtonAlpha

PR #2350, opened 2022-06-23
Run individule pytests instead of running all pytests at once.

mint570:
**What I did**
Run pytests one test at a time.

**Why I did it**
Fix https://github.com/Azure/sonic-swss/issues/2330

**How I verified it**

**Details if related**


mint570:
@lguohan

mint570:
@prsunny @pettershao-ragilenetworks 
Can you please take a look at this PR?

mint570:
@theasianpianist can you please take a look?

prsunny:
@theasianpianist , please review

prsunny:
lgtm, @liushilongbuaa , @xumia , could you please review?

liushilongbuaa (approved):
LGTM

prsunny:
@mint570 , as discussed offline with @theasianpianist , we see a significant increase in time for the test completion.

image

Can you please work on an alternative solution for #2373 and #2330?

mint570:
I don't have an obvious solution now. The increase in time is mostly due to the overhead of running individual test.
There might be some improvement we can do. Such as grouping the tests together (instead of running 100 tests, we can run 10 test groups, each includes 10 tests). But the overall time will still be increased compare to running all tests at once.

mint570:
Some random ideas. May or may not help:
1. Integrate the coverage file collection into the pytest. So that we can run all tests in one pytest call.
2. Share a single vs container with all tests.

Not sure if the above is possible or not. I tried to look at how the pytest dvs works. But unfortunately, I don't have enough expertise to understand that.

PR #2364, opened 2022-06-30
Fix for System interface for remote asic LAG not getting created if received prior to SYSTEM_LAG creation from Chassis DB

skeesara-nokia:
**What I did**
Changed interface and neighbor orch code in orchagent which handle updates from Chassis DB for system interface and remote neighbors. Fixes https://github.com/Azure/sonic-buildimage/issues/11290

**Why I did it**
If an update from the Chassis DB for a system interface or a remote neighbor is received prior to receiving an update for the creation of the SYSTEM LAG or SYSTEM_PORT - the current code incorrectly treats this as an update from the local asic and ignores it. As a result the SYSTEM_INTERFACE or the Remote neighbor is not creating when this specific sequence of updates are received.

This change correctly verifies that an update is from the local asic before ignoring it.

**How I verified it**
Prior to the fix - adding a the system_interface to the Chassis DB before adding the corresponding System LAG would result in the ROUTER_INTERFACE never getting created by SAI. After the fix the ROUTER_INTERFACE is correctly created for this scenario.

The sonic-mgmt supervisor reset tests that were failing prior to this fix - passed after the fix was applied.

**Details if related**

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -645,7 +645,7 @@ void IntfsOrch::doTask(Consumer &consumer)
 
         if(table_name == CHASSIS_APP_SYSTEM_INTERFACE_TABLE_NAME)
         {
-            if(!isRemoteSystemPortIntf(alias))
+            if(isLocalSystemPortIntf(alias))
             {
                 //Synced local interface. Skip
                 it = consumer.m_toSync.erase(it);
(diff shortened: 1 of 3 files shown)


judyjoseph:
Azp run

judyjoseph (approved):
LGTM

PR #2367, opened 2022-07-04
[orchagent]: Enhance initSaiPhyApi

andywongarista:
**What I did**
- Added support for PHYs that do not have firmware support and would return SAI_STATUS_NOT SUPPORTED when attempting to retrieve SAI_SWITCH_ATTR_FIRMWARE_MAJOR_VERSION
- Removed format check on gearbox_config.json hwinfo string

**Why I did it**
- Some PHYs (e.g. BCM54182) do not have firmware support and will return SAI_STATUS_NOT SUPPORTED when attempting to retrieve SAI_SWITCH_ATTR_FIRMWARE_MAJOR_VERSION. This should not fail initSaiPhyApi.
- SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO is defined to be a string but the exact format is vendor specific. Thus, the current format check in initSaiPhyApi should not be present.

**How I verified it**
Tested on Arista platform with BCM54182 and appropriate PAI driver

**Details if related**

### DIFF
--- a/orchagent/saihelper.cpp
+++ b/orchagent/saihelper.cpp
@@ -359,9 +359,6 @@ sai_status_t initSaiPhyApi(swss::gearbox_phy_t *phy)
     sai_status_t status;
     char fwPath[PATH_MAX];
     char hwinfo[HWINFO_MAX_SIZE + 1];
-    char hwinfoIntf[IFNAMSIZ + 1];
-    unsigned int hwinfoPhyid;
-    int ret;
 
     SWSS_LOG_ENTER();
 
@@ -377,19 +374,11 @@ sai_status_t initSaiPhyApi(swss::gearbox_phy_t *phy)
     attr.value.u32 = 0;
     attrs.push_back(attr);
 
-    ret = sscanf(phy->hwinfo.c_str(), "%" STR(IFNAMSIZ) "[^/]/%u", hwinfoIntf, &hwinfoPhyid);
-    if (ret != 2) {
-        SWSS_LOG_ERROR("BOX: hardware info doesn't match the 'interface_name/phyid' "
-                       "format");
-        return SAI_STATUS_FAILURE;
+    if( phy->hwinfo.length() > HWINFO_MAX_SIZE ) {
+       SWSS_LOG_ERROR( "hwinfo string attribute is too long." );
+       return SAI_STATUS_FAILURE;
     }
-
-    if (hwinfoPhyid > std::numeric_limits<uint16_t>::max()) {
-        SWSS_LOG_ERROR("BOX: phyid is bigger than maximum limit");
-        return SAI_STATUS_FAILURE;
-    }
-
-    strcpy(hwinfo, phy->hwinfo.c_str());
+    strncpy(hwinfo, phy->hwinfo.c_str(), HWINFO_MAX_SIZE);
 
     attr.id = SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO;
     attr.value.s8list.count = (uint32_t) phy->hwinfo.length();
--- a/orchagent/saihelper.cpp
+++ b/orchagent/saihelper.cpp
@@ -443,12 +443,17 @@ sai_status_t initSaiPhyApi(swss::gearbox_phy_t *phy)
 
     attr.id = SAI_SWITCH_ATTR_FIRMWARE_MAJOR_VERSION;
     status = sai_switch_api->get_switch_attribute(phyOid, 1, &attr);
-    if (status == SAI_STATUS_SUCCESS) {
+    if (status == SAI_STATUS_SUCCESS)
+    {
        phy->firmware_major_version = string(attr.value.chardata);
-    } else if( status == SAI_STATUS_NOT_SUPPORTED ) {
-       phy->firmware_major_version = "";
+    }
+    else if ( status == SAI_STATUS_NOT_SUPPORTED )
+    {
+       phy->firmware_major_version = "N/A";
        status = SAI_STATUS_SUCCESS;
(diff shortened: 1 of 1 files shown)


orchagent/saihelper.cpp
```diff
@@ -454,15 +443,14 @@ sai_status_t initSaiPhyApi(swss::gearbox_phy_t *phy)
 
     attr.id = SAI_SWITCH_ATTR_FIRMWARE_MAJOR_VERSION;
     status = sai_switch_api->get_switch_attribute(phyOid, 1, &attr);
-    if (status != SAI_STATUS_SUCCESS)
-    {
+    if (status == SAI_STATUS_SUCCESS) {
```
prsunny, line 446: {-> new line. Thats the general convention in swss repo

orchagent/saihelper.cpp
```diff
@@ -454,15 +443,14 @@ sai_status_t initSaiPhyApi(swss::gearbox_phy_t *phy)
...
+       phy->firmware_major_version = string(attr.value.chardata);
+    } else if( status == SAI_STATUS_NOT_SUPPORTED ) {
+       phy->firmware_major_version = "";
```
Junchao-Mellanox, line 449: I am not sure the empty string is good enough here. Should it be "N/A" or "Not Supported"?

orchagent/saihelper.cpp
```diff
@@ -454,15 +443,14 @@ sai_status_t initSaiPhyApi(swss::gearbox_phy_t *phy)
 
     attr.id = SAI_SWITCH_ATTR_FIRMWARE_MAJOR_VERSION;
     status = sai_switch_api->get_switch_attribute(phyOid, 1, &attr);
-    if (status != SAI_STATUS_SUCCESS)
-    {
+    if (status == SAI_STATUS_SUCCESS) {
+       phy->firmware_major_version = string(attr.value.chardata);
+    } else if( status == SAI_STATUS_NOT_SUPPORTED ) {
```
Junchao-Mellanox, line 448: Should this be handled in vendor SAI instead of swss?

orchagent/saihelper.cpp
andywongarista: I'm fine with changing it to N/A. I don't think there are any strict requirements here.

orchagent/saihelper.cpp
andywongarista: Vendor SAI will handle returning the appropriate status and setting the attr variable appropriately, and swss will update the local phy object appropriately. I'm not sure I understand why the responsibilities would need to change.

Junchao-Mellanox: If we do this change in swss, it would change the behavior for all platforms. However, BCM54182(or a few limited platforms) is the only platform that requires this change. That is why I thought it should be vendor SAI's responsibility to do so.
Anyway, if this change has no harm to other platforms, I can live with it. I don't have strong opinion here.

Junchao-Mellanox (dismissed):

orchagent/saihelper.cpp
andywongarista: Yes, this should not affect other platforms if they are still returning SAI_STATUS_FAILURE. But I think the change here is appropriate to allow initSaiPhyApi to return with success on platforms that don't support firmware.

prsunny:
Try `/azpw run`

orchagent/saihelper.cpp
```diff
@@ -454,15 +443,19 @@ sai_status_t initSaiPhyApi(swss::gearbox_phy_t *phy)
...
+    }
+    else if ( status == SAI_STATUS_NOT_SUPPORTED )
+    {
+       phy->firmware_major_version = "N/A";
+       status = SAI_STATUS_SUCCESS;
     }
     else
     {
-        phy->firmware_major_version = string(attr.value.chardata);
+        SWSS_LOG_ERROR("BOX: Failed to get firmware major version:%d rtn:%d", phy->phy_id, status);
```
prgeor, line 457: can we add PHY hwinfo string in the error output that may be more easily recognized than phy id?

orchagent/saihelper.cpp
```diff
@@ -454,15 +443,19 @@ sai_status_t initSaiPhyApi(swss::gearbox_phy_t *phy)
...
     attr.id = SAI_SWITCH_ATTR_FIRMWARE_MAJOR_VERSION;
     status = sai_switch_api->get_switch_attribute(phyOid, 1, &attr);
-    if (status != SAI_STATUS_SUCCESS)
+    if (status == SAI_STATUS_SUCCESS)
     {
-        SWSS_LOG_ERROR("BOX: Failed to get firmware major version:%d rtn:%d", phy->phy_id, status);
-        return status;
+       phy->firmware_major_version = string(attr.value.chardata);
+    }
+    else if ( status == SAI_STATUS_NOT_SUPPORTED )
```
jimmyzhai, line 450: If `phy->firmware` is not empty, we should ensure attr `SAI_SWITCH_ATTR_FIRMWARE_MAJOR_VERSION` returns `SAI_STATUS_SUCCESS`, otherwise should not try to get this attr.

orchagent/saihelper.cpp
andywongarista: Done

orchagent/saihelper.cpp
andywongarista: Done

Junchao-Mellanox (approved):

jimmyzhai (approved):

andywongarista:
I opened https://github.com/Azure/sonic-swss/pull/2383 to revert the change here that increases the hwinfo count

PR #2369, opened 2022-07-05
[muxorch] Always use direct link for SoC IPs

lolyu:
**What I did**
For SoC IPs of ports in `active-active` cable type, if mux is toggled to standby, still use direct link instead of changing next-hop to the tunnel.

**Why I did it**
In an `active-active` dualtor setup, changing the nexthop of route to SoC IP to the tunnel will   have the following problem:
If lower ToR is already standby, and somehow the upper ToR decides to toggle itself to standby, the toggle will fail:
linkmgrd decide to toggle to standby --> muxorch disables the SoC IP neighbor and change the route next-hop to the tunnel --> ycabled could not setup gRPC connection.

**How I verified it**
Add unittest and verify the change on `active-active` testbeds.

**Details if related**

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -534,9 +534,9 @@ bool MuxCable::nbrHandler(bool enable, bool update_rt)
 void MuxCable::updateNeighbor(NextHopKey nh, bool add)
 {
     sai_object_id_t tnh = mux_orch_->getNextHopTunnelId(MUX_TUNNEL, peer_ip4_);
-    if (add && skip_neighbors_.count(nh.ip_address) != 0)
+    if (add && skip_neighbors_.find(nh.ip_address) != skip_neighbors_.end())
     {
-        SWSS_LOG_NOTICE("Skip update neighbor %s on %s", nh.ip_address.to_string().c_str(), nh.alias.c_str());
+        SWSS_LOG_INFO("Skip update neighbor %s on %s", nh.ip_address.to_string().c_str(), nh.alias.c_str());
         return;
     }
     nbr_handler_->update(nh, tnh, add, state_);
@@ -1214,7 +1214,6 @@ bool MuxOrch::handleMuxCfg(const Request& request)
     auto srv_ip6 = request.getAttrIpPrefix("server_ipv6");
 
     std::set<IpAddress> skip_neighbors;
-    std::string cable_type = "active-standby";
 
     const auto& port_name = request.getKeyString(0);
     auto op = request.getOperation();
@@ -1233,15 +1232,6 @@ bool MuxOrch::handleMuxCfg(const Request& request)
             SWSS_LOG_NOTICE("%s: %s was added to ignored neighbor list", port_name.c_str(), soc_ip6.getIp().to_string().c_str());
             skip_neighbors.insert(soc_ip6.getIp());
         }
-        else if (name == "cable_type")
-        {
-            cable_type = request.getAttrString("cable_type");
-        }
-    }
-
-    if (cable_type != "active-active")
-    {
-        skip_neighbors.clear();
     }
 
     if (op == SET_COMMAND)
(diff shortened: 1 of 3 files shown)


prsunny (changes requested):
I prefer to take a different approach for this and not make it part of the nbr_handler

lolyu:
> I prefer to take a different approach for this and not make it part of the nbr_handler. We can just return from the 'update' function.

OK, I will change it to simply skip the neighbor handler

prsunny:
Propose to have it as below:

1. Extend Muxcable class to have a list of soc_ip4/6 addresses. Name it say, 'skip_neighbor_list'
2. In MuxOrch::updateNeighbor, check for this list and return if new neighbor is present in the 'add' case.
3. No changes to MuxNeighbor class

orchagent/muxorch.cpp
```diff
@@ -534,6 +534,11 @@ bool MuxCable::nbrHandler(bool enable, bool update_rt)
 void MuxCable::updateNeighbor(NextHopKey nh, bool add)
 {
     sai_object_id_t tnh = mux_orch_->getNextHopTunnelId(MUX_TUNNEL, peer_ip4_);
+    if (add && skip_neighbors_.count(nh.ip_address) != 0)
```
prsunny, line 537: Please use find instead of count as the intend is to check if the key exists. This is the current pattern in the file.

orchagent/muxorch.cpp
```diff
@@ -534,6 +534,11 @@ bool MuxCable::nbrHandler(bool enable, bool update_rt)
...
+    {
+        SWSS_LOG_NOTICE("Skip update neighbor %s on %s", nh.ip_address.to_string().c_str(), nh.alias.c_str());
```
prsunny, line 539: Please move to INFO level. We don't want this to be printed for each time a neighbor update happens for the skip neighbors

orchagent/muxorch.cpp
```diff
@@ -1208,9 +1213,37 @@ bool MuxOrch::handleMuxCfg(const Request& request)
     auto srv_ip = request.getAttrIpPrefix("server_ipv4");
     auto srv_ip6 = request.getAttrIpPrefix("server_ipv6");
 
+    std::set<IpAddress> skip_neighbors;
+    std::string cable_type = "active-standby";
```
prsunny, line 1217: We don't have to check for cable_type in orchagent. If 'soc_ipv4' is present, it must be skipped from tunnel neighbors. Please remove.

prsunny:
@lolyu , please dont force-push. I would like to review incremental changes.

lolyu:
> @lolyu , please dont force-push. I would like to review incremental changes.

@prsunny Sure, updated, thanks!

orchagent/muxorch.cpp
```diff
@@ -534,6 +534,11 @@ bool MuxCable::nbrHandler(bool enable, bool update_rt)
 void MuxCable::updateNeighbor(NextHopKey nh, bool add)
 {
     sai_object_id_t tnh = mux_orch_->getNextHopTunnelId(MUX_TUNNEL, peer_ip4_);
+    if (add && skip_neighbors_.find(nh.ip_address) != skip_neighbors_.end())
```
prsunny, line 537: Can you check if a 'del' operation does not impact the code flow?

prsunny (commented):
lgtm, minor comments

tests/test_mux.py
```diff
@@ -212,21 +221,23 @@ def create_and_test_neighbor(self, confdb, appdb, asicdb, dvs, dvs_route):
         dvs_route.check_asicdb_route_entries([self.SERV2_IPV4+self.IPV4_MASK, self.SERV2_IPV6+self.IPV6_MASK])
 
         # The first standby route also creates as tunnel Nexthop
-        self.check_tnl_nexthop_in_asic_db(asicdb, 3)
+        self.check_tnl_nexthop_in_asic_db(asicdb, 4)
```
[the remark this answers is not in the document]
prsunny: Discussed offline with @lolyu , and it is as per the current test expectation.

prsunny (approved):

PR #2372, opened 2022-07-08
Vlan stacking implementation

chaoskao:
**What I did**
Implemented the VLAN stacking.

**Why I did it**
N/A

**How I verified it**
Verified using vs tests.

**Details if related**
No.

### DIFF
--- a/cfgmgr/vlanmgr.cpp
+++ b/cfgmgr/vlanmgr.cpp
@@ -7,6 +7,7 @@
 #include "tokenize.h"
 #include "shellcmd.h"
 #include "warm_restart.h"
+#include "converter.h"
 #include <swss/redisutility.h>
 
 using namespace std;
@@ -31,6 +32,8 @@ VlanMgr::VlanMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
         m_stateVlanMemberTable(stateDb, STATE_VLAN_MEMBER_TABLE_NAME),
         m_appVlanTableProducer(appDb, APP_VLAN_TABLE_NAME),
         m_appVlanMemberTableProducer(appDb, APP_VLAN_MEMBER_TABLE_NAME),
+        m_appVlanStackingTableProducer(appDb, APP_VLAN_STACKING_TABLE_NAME),
+        m_appVlanXlateTableProducer(appDb, APP_VLAN_TRANSLATION_TABLE_NAME),
         replayDone(false)
 {
     SWSS_LOG_ENTER();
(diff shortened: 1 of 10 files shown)


prsunny:
@venkatmahalingam , please review

tests/test_vlan_stacking.py
```diff
@@ -0,0 +1,199 @@
+import pytest
```
prsunny, line 1: @prabhataravind , please review the test/cleanup

cfgmgr/vlanmgr.cpp
```diff
@@ -675,6 +678,167 @@ void VlanMgr::doVlanMemberTask(Consumer &consumer)
...
+        auto& t = it->second;
+
+        string key = kfvKey(t);
+
+        vector<string> keys = tokenize(key, config_db_key_delimiter);
+
+        if (keys.size() != 2)
+        {
+            SWSS_LOG_ERROR("Unknown VLAN Stacking config %s.", key.c_str());
+            it = consumer.m_toSync.erase(it);
```
venkatmahalingam, line 695: can we call goto vlan_stacking_parse_failed;?

cfgmgr/vlanmgr.cpp
```diff
@@ -675,6 +678,167 @@ void VlanMgr::doVlanMemberTask(Consumer &consumer)
...
+            continue;
+        }
+
+        if (!isValidVlanId(keys.at(1), 1, 4094))
+        {
+            SWSS_LOG_ERROR("Unknown VLAN Stacking config %s.", key.c_str());
+            it = consumer.m_toSync.erase(it);
```
venkatmahalingam, line 702: can we call goto vlan_stacking_parse_failed;?

tests/test_vlan_stacking.py
prabhataravind: Test/cleanup Looks good to me!

cfgmgr/vlanmgr.cpp
```diff
@@ -675,6 +678,167 @@ void VlanMgr::doVlanMemberTask(Consumer &consumer)
...
+                    try
+                    {
+                        s_vlan_priority = (uint8_t)stoi(attr_value);
+                    }
+                    catch (std::invalid_argument& e)
+                    {
+                        goto vlan_stacking_parse_failed;
+                    }
+
+                    if (s_vlan_priority < 0 || s_vlan_priority > 7)
```
prvattem, line 739: s_vlan_priority is unsigned int. It cant be "< 0". Need to relook at this check?

orchagent/portsorch.cpp
```diff
@@ -3999,6 +4003,490 @@ void PortsOrch::doLagMemberTask(Consumer &consumer)
...
+                {
+                    return task_failed;
+                }
+            }
+        }
+
+        sai_object_id_t port_oid = port.m_type == Port::PHY ? port.m_port_id : port.m_lag_id;
+        VlanStackKey stack_key = { port_oid, (uint16_t)to_uint<uint16_t>(s_vlanid) };
+
+        task_status = deleteVlanStackEntry(stack_key);
```
prvattem, line 4323: Why a delete is issued before Adding the Stacking Entry?

cfgmgr/vlanmgr.cpp
prvattem: BTW 0xFF is a valid value right?

chaoskao: no, 0xFF is not valid value

orchagent/portsorch.cpp
chaoskao: To make modify action easy, so use remove and add to support update

PR #2374, opened 2022-07-08
[202012] Fix for "orchagent crashed when trying to delete fdb static entry with swssconfig #11046"

anilkpan:
Root cause:
When fdb flush was issued after adding a static mac, the mac was deleted due to the flush. Then when we try to remove the static mac, it thrown error and crashed due to the mac not found.

Fix:
Updated fdb flush to flush only dynamic mac.


gechiang (approved):
LGTM!

prsunny (approved):

PR #2379, opened 2022-07-15
[QoS] Fix issue: the WRED profile can not be set if current min > new max or current max < new min

stephenxs:
**What I did**

Fix issue:
Setting a WRED profile can fail in case
 - the current min threshold is greater than the new max threshold
 - or the current max threshold is less than the new min threshold

for any color and at any time.
Eg.
- Current min=1M, max=2M, new min=3M, new max=4M
- The min threshold is set first, so current max (2M) < new min (3M), which violates the condition

This is because there can be only one attribute in each SAI SET operation,
which means the vendor SAI does not understand the whole information of the new attributes and can only perform the sanity check against each SET operation.

**Why I did it**

Fix the issue

**How I verified it**

Manual test and mock test.

**Details if related**
The fix
The thresholds that have been applied to SAI will be stored in orchagent.

The original logic is to handle each attribute to be set and append it to an attribute list.
To resolve the issue, a deferred attribute list is introduced and will be appended to the original attribute list after all the attributes have been handled.

In the new logic, each threshold to be set will be checked against the stored data.
In case it violates the condition, the violating attribute will be deferred, done via putting it into the deferred attributes list.

For any color, there can be only 1 threshold violating the condition.
Otherwise, it means both new min > old max and new max > old min, which means either old max < old min or new max < new min, which means either old or new data is illegal.
This can not happen because illegal data will not be applied and stored.

By doing so, the other threshold will be applied first, which extends the threshold range and breaks the violating condition.
A logic is also introduced to guarantee the min threshold is always less than the max threshold in the new profile to be set.

For the above example,
In the new logic, the min threshold will be deferred so the new max threshold will be applied first and then the new min is applied. In this flow, there is no violation at any time.
- min = 1M, max = 2M
- => min = 1M, max = 4M
- => min = 3M, max = 4M

### DIFF
--- a/orchagent/qosorch.cpp
+++ b/orchagent/qosorch.cpp
@@ -502,10 +502,10 @@ bool WredMapHandler::convertFieldValuesToAttributes(KeyOpFieldsValuesTuple &tupl
     sai_uint32_t threshold;
 
     /*
-     * Setting WRED profile can fail in case for any color
+     * Setting WRED profile can fail in case
      * - the current min threshold is greater than the new max threshold 
      * - or the current max threshold is less than the new min threshold
-     * on some vendor's platforms.
+     * for any color at any time, on some vendor's platforms.
      *
      * The root cause
      * There can be only one attribute in each SAI SET operation, which means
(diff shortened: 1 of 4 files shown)


liat-grozovik:
@neethajohn , @prsunny could you please help to review?

liat-grozovik:
@neethajohn  kindly reminder to help review this pr. this is a fix needed for 202205

neethajohn (approved):

PR #2383, opened 2022-07-20
Revert hwinfo count change

andywongarista:
**What I did**
Revert change from https://github.com/Azure/sonic-swss/pull/2367 which increases count associated with SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO by 1, as well as the memset.

**Why I did it**
Original intention of this change was to accommodate sairedis behaviour when copying null-terminated string; original behaviour is that the null-terminator would not be copied and so receiver of the hwinfo (PAI) would see non-null terminated string.

Reverting this change so that old behaviour is maintained and PAI driver is responsible for not relying on string to be null terminated.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/saihelper.cpp
+++ b/orchagent/saihelper.cpp
@@ -378,11 +378,10 @@ sai_status_t initSaiPhyApi(swss::gearbox_phy_t *phy)
        SWSS_LOG_ERROR( "hwinfo string attribute is too long." );
        return SAI_STATUS_FAILURE;
     }
-    memset(hwinfo, 0, HWINFO_MAX_SIZE + 1);
     strncpy(hwinfo, phy->hwinfo.c_str(), phy->hwinfo.length());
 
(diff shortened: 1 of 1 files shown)


jimmyzhai (approved):

jimmyzhai:
What I did
Revert change from https://github.com/Azure/sonic-swss/pull/2367 which increases count associated with SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO by 1, as well as the memset.

Why I did it
Original intention of this change was to accommodate sairedis behaviour when copying null-terminated string; original behaviour is that the null-terminator would not be copied and so receiver of the hwinfo (PAI) would see non-null terminated string.

Reverting this change so that old behaviour is maintained and PAI driver is responsible for not relying on string to be null terminated.

Blueve:
@andywongarista since this PR(https://github.com/sonic-net/sonic-swss/pull/2367 ) has been backported to 202012, i think this one need back port as well?

PR #2384, opened 2022-07-20
Revert "Add support for IP interface loopback action (#2307)"

prabhataravind:
This reverts commit 5043701cc87efc3cb642762b13217729b576cf59.

Revert "[VS Test] Skip failing subport tests (#2370)"

This reverts commit 71752450e5d2c3c2f126aabddfe65b1d14344490.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/cfgmgr/intfmgr.cpp
+++ b/cfgmgr/intfmgr.cpp
@@ -728,7 +728,6 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
     string grat_arp = "";
     string mpls = "";
     string ipv6_link_local_mode = "";
-    string loopback_action = "";
 
     for (auto idx : data)
     {
@@ -771,10 +770,6 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
         {
             vlanId = value;
         }
-        else if (field == "loopback_action")
-        {
-            loopback_action = value;
-        }
     }
 
     if (op == SET_COMMAND)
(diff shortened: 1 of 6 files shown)


prsunny:
@liorghub , please take a look

prsunny:
@liorghub , We just re-added the subport test in this PR - https://github.com/Azure/sonic-swss/pull/2389 and it failed. So somehow it is linked to the changes. Can you please investigate this with priority?

liorghub:
> @liorghub , We just re-added the subport test in this PR - #2389 and it failed. So somehow it is linked to the changes. Can you please investigate this with priority?

@prsunny I am on it

prsunny:
> > @liorghub , We just re-added the subport test in this PR - #2389 and it failed. So somehow it is linked to the changes. Can you please investigate this with priority?
> 
> @prsunny @dprital I tried to reproduce the issue that occured in vstest of PR #2389 but unfortunately I couldn't. In my run there were other failures. vstest in PR #2389 failed since there were 2 keys in table ASIC_STATE:SAI_OBJECT_TYPE_VIRTUAL_ROUTER. I searched the code to understand what is the flow that creates a new key and found that it happens once on switch init and each time we create VRF or EVPN tunnel. My tests are not creating VRF or tunnel, therefore they could not cause the failure. If you still think otherwise, appreciate if we can work together to resolve the issue before reverting my commit ([5043701](https://github.com/Azure/sonic-swss/commit/5043701cc87efc3cb642762b13217729b576cf59))

We are also not clear on the rootcause, but reverting your PR makes it pass, but with the changes, its failing. If you cannot repro it manually, can you try by reverting the changes one-by-one. Like remove intfmgrd but keep intforch and raise a PR. Lets try to narrow down.

PR #2387, opened 2022-07-21
[EVPN]Modified tunnel creation logic when creating tunnel in VRF-VNI map creation flow

dgsudharsan:
**What I did**
To fix issue https://github.com/Azure/sonic-buildimage/issues/11428
Modified the logic of tunnel map creation to create tunnel with tunnel map for vlan-vni map in addition to vrf-vni map when tunnel is created first time in the VRF-VNI map processing flow.

**Why I did it**
During the configuration phase when VRF-VNI map arrives before VLAN-VNI map, the tunnel is created without a tunnel map for vlan-vni membership. This is problematic when VLAN to VNI map arrives later, tunnel map entry cannot be created since the tunnel map doesn't exist and its a create only attribute in SAI.

**How I verified it**
Modified UT to add VRF-VNI map first and VLAN-VLAN map later

**Details if related**


volodymyrsamotiy (approved):

prsunny:
@srj102 , please review

tapashdas:
Please add a new test case instead of altering existing one.

dgsudharsan:
> Please add a new test case instead of altering existing one.

There are 4 tests that create vlan- vni map followed by vrf - vni map. All these test follow the same sequence for initialization. So I modified only one test to reverse the order.
This ensures all scenarios are covered in existing tests.
Adding new tests adds additional time for vs tests which already takes lots of time. So I reused the existing test.

Please also note that the test modified is in mellanox specific and so for the same test broadcom specific test is not modified which could also cover the scenario.

dgsudharsan:
@tapashdas  @srj102  @prsunny  Please let me know if there are any more comments.

dgsudharsan:
> @tapashdas @srj102 @prsunny Please let me know if there are any more comments.

Hi @tapashdas , @prsunny , @srj102  Can you please review and signoff? This is a critical bugfix required for us.

tapashdas (dismissed):

orchagent/vxlanorch.cpp
```diff
@@ -2013,19 +1952,21 @@ bool VxlanTunnelMapOrch::addOperation(const Request& request)
...
-            gPortsOrch->getPort(port_tunnel_name,tunPort);
-            gPortsOrch->addBridgePort(tunPort);
+            Port tunPort;
+            auto src_vtep = tunnel_obj->getSrcIP().to_string();
+            if (!tunnel_orch->getTunnelPort(src_vtep, tunPort, true))
+            {
+                auto port_tunnel_name = tunnel_orch->getTunnelPortName(src_vtep, true);
+                gPortsOrch->addTunnel(port_tunnel_name, tunnel_obj->getTunnelId(), false);
+                gPortsOrch->getPort(port_tunnel_name,tunPort);
+                gPortsOrch->addBridgePort(tunPort);
```
prsunny, line 1968: As discussed, lets move the bridgeport creation based on capability.

dgsudharsan: Hi Prince. I have added the check. I will look into the test failures and update.

PR #2392, opened 2022-07-22
[Everflow/ERSPAN] Set correct destination port and mac address when the nexthop is updated for ERSPAN mirror destination

saksarav-nokia:
Signed-off-by: Sakthivadivu Saravanaraj <sakthivadivu.saravanaraj@nokia.com>

**What I did**
Fixed the issues reported in https://github.com/Azure/sonic-buildimage/issues/11457
**Why I did it**
getNeighborEntry:  This function compares Neighbor alias with nexthop alias. Since the alias for nexthop was modified from systemport name to Inband port name in addNextHop, the alias comparison was failing and  mirror session was deactivated. Modified the getNeighborEntry to  compare the correct alias.
updateSessionDstMac: Modified the code to set the dest mac of the mirror to gMacAddress if it is voq switch
updateSessionDstPort: Set the dest port of the mirror to Rec port if the mirror is ERSPAN mirror
removeRanges: Delete the acl range only if it was created . If the acl entry with mirror action is created when the mirror is not activated, only counter is created. Acl range and acl rule are not created. So removeRanges returns false and hence "Failed to remove acl rule" errors printed many times in syslog.
AclRuleMirror::activate :  If Acl rule with mirror action was created before the mirror was activated and deleted, the counter was deleted. So activating the mirror calls AclRule:::createRule without creating counter and orchagent calls sairedis with counter Object as NULL and sai_meta layer returns error, so orchagent crashes. Added code to create counter if not valid before creating the rule.
**How I verified it**
Verified ingress and egress mirror oc tests passes
**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -1477,6 +1477,11 @@ const vector<AclRangeConfig>& AclRule::getRangeConfig() const
     return m_rangeConfig;
 }
 
+bool AclRule::getCreateCounter() const
+{
+    return m_createCounter;
+}
+
 shared_ptr<AclRule> AclRule::makeShared(AclOrch *acl, MirrorOrch *mirror, DTelOrch *dtel, const string& rule, const string& table, const KeyOpFieldsValuesTuple& data)
 {
     shared_ptr<AclRule> aclRule;
@@ -1624,6 +1629,13 @@ bool AclRule::createCounter()
 bool AclRule::removeRanges()
 {
     SWSS_LOG_ENTER();
+    if (!m_ranges.size())
+    {
+       //The Acl Rules which have mirror action will not have ranges created till the mirror becomes active
+       SWSS_LOG_INFO("No Acl Range created for ACL Rule %s in table %s", m_id.c_str(), m_pTable->getId().c_str());
+       return true;
+    }
+
     for (const auto& rangeConfig: m_rangeConfig)
     {
         if (!AclRange::remove(rangeConfig.rangeType, rangeConfig.min, rangeConfig.max))
@@ -1922,6 +1934,16 @@ bool AclRuleMirror::activate()
         attr.value.aclaction.parameter.objlist.list = &oid;
         attr.value.aclaction.parameter.objlist.count = 1;
         setAction(it.first, attr.value.aclaction);
+    }
+
+    // If the rule with mirror action is removed and then mirror is activated, create the counter before rule is created
+    if (!hasCounter())
+    {
+        if (getCreateCounter() && !createCounter())
+        {
+            SWSS_LOG_ERROR("createCounter failed for Rule %s session %s", m_id.c_str(), m_sessionName.c_str());
+            return false;
+       }
     }
 
     if (!AclRule::createRule())
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -1115,7 +1115,13 @@ bool MirrorOrch::updateSessionDstMac(const string& name, MirrorEntry& session)
 
     sai_attribute_t attr;
     attr.id = SAI_MIRROR_SESSION_ATTR_DST_MAC_ADDRESS;
-    memcpy(attr.value.mac, session.neighborInfo.mac.getMac(), sizeof(sai_mac_t));
+    if (gMySwitchType == "voq")
+    {
+        memcpy(attr.value.mac, gMacAddress.getMac(), sizeof(sai_mac_t));
+    } else
+    {
+        memcpy(attr.value.mac, session.neighborInfo.mac.getMac(), sizeof(sai_mac_t));
+    }
 
     sai_status_t status = sai_mirror_api->set_mirror_session_attribute(session.sessionId, &attr);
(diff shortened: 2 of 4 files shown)


saksarav-nokia:
@judyjoseph @prsunny , could you please review this PR since this is blocking  everflow testing with github image

orchagent/mirrororch.cpp
```diff
@@ -1115,13 +1134,19 @@ bool MirrorOrch::updateSessionDstMac(const string& name, MirrorEntry& session)
...
+    {
+         memcpy(attr.value.mac, session.neighborInfo.mac.getMac(), sizeof(sai_mac_t));
+    }
 
     sai_status_t status = sai_mirror_api->set_mirror_session_attribute(session.sessionId, &attr);
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to update mirror session %s destination MAC to %s, rv:%d",
-                name.c_str(), session.neighborInfo.mac.to_string().c_str(), status);
+                name.c_str(), sai_serialize_mac(attr.value.mac).c_str(), status);
```
judyjoseph, line 1149: Checking if there is any particular reason for using  sai_serialize_mac() here

orchagent/aclorch.cpp
```diff
@@ -1924,6 +1936,16 @@ bool AclRuleMirror::activate()
         setAction(it.first, attr.value.aclaction);
     }
 
+    // If the rule with mirror action is removed and then mirror is activated, create the counter before rule is created
+    if (!hasCounter())
```
judyjoseph, line 1940: @bingwang-ms to check this change as these are generic changes, not specific to VOQ. wondering why this is not affecting non-chassis platforms.

orchagent/mirrororch.cpp
saksarav-nokia: session.neighborInfo.mac is swss:MacAddress  class which has a function to_string. But attr.value.mac is sai_mac_t which is "typedef UINT8   sai_mac_t[6];". So i had to call sai_serialize_mac to convert the mac to string.

orchagent/mirrororch.cpp
```diff
@@ -583,13 +584,31 @@ void MirrorOrch::setSessionState(const string& name, const MirrorEntry& session,
     if (attr.empty() || attr == MIRROR_SESSION_MONITOR_PORT)
     {
         Port port;
-        m_portsOrch->getPort(session.neighborInfo.portId, port);
+        if ((gMySwitchType == "voq") && (session.type == MIRROR_SESSION_ERSPAN) &&
+            (gIntfsOrch->isRemoteSystemPortIntf(session.neighborInfo.neighbor.alias)))
```
judyjoseph, line 588: Do we need to check if this is RemoteSystemPort here, irrespective of local/remote port we need to use Recirc port ?

saksarav-nokia: I think we can remove gIntfsOrch->isRemoteSystemPortIntf(session.neighborInfo.neighbor.alias) check

judyjoseph:
@arlakshm, @ysmanman - could you take a look at the changes as well ? thanks !

orchagent/aclorch.cpp
bingwang-ms: Normally, we will not see this issue because the counter for the rule has been created at `AclRule::create()`. It can happen if the rule is deleted before the mirror session is activated. It's a corner case, I think.

orchagent/mirrororch.cpp
```diff
@@ -50,6 +50,7 @@ extern sai_port_api_t *sai_port_api;
 extern sai_object_id_t  gSwitchId;
 extern PortsOrch*       gPortsOrch;
 extern string           gMySwitchType;
+extern IntfsOrch        *gIntfsOrch;
```
bingwang-ms, line 53: `gIntfsOrch` is never referenced.

orchagent/aclorch.cpp
saksarav-nokia: This happens when the acl rule is created with MIRROR action and  deleted before the nexthop to the mirror destination is not resolved and activated. The counter is created only when the mirror session is activated.

orchagent/mirrororch.cpp
saksarav-nokia: Fixed it

orchagent/mirrororch.cpp
saksarav-nokia: fixed it

abdosi (commented):
LGTM.. Can we have UT that cam simulate this change ?

abdosi (approved):

bingwang-ms (approved):

judyjoseph (approved):

gechiang:
@saksarav-nokia , can you raise a new PR in 202205 branch to cherry pick this same change and I will help approve it and merge it so that we don't need to wait for 202205 branch owner to perform the cherry-pick.

gechiang:
Removed label "Request for 202205 branch" because we have already merged in by the following PR:
https://github.com/sonic-net/sonic-swss/pull/2455.

PR #2397, opened 2022-07-26
[DualToR] Handle race condition between tunnel_decap and mux orchestrator

devpatha:
**What I did**
Change orch list to make sure mux_orch and mux_cb_orch are placed after tunnel_decap_orch

**Why I did it**
handlePeerSwitch returns false because mux tunnel not yet created, and that results in create_nh_tunnel failure.
This moves ports into error state.

**How I verified it**
Verified that after changing the order ports don't move to error state.

**Details if related**
Creating a new PR for #2113 

PR #2114 has already been merged into 202012 branch


prsunny:
thanks @devpatha for the quick turnaround.

PR #2398, opened 2022-07-26
[macsec]: Set MTU for MACsec

Pterosaur:

**What I did**
Set extra MTU for MACsec enabled port.

**Why I did it**
MACsec frame will expend the packet with MACsec SecTAG, Otherwise if a packet length equals the MTU which will be dropped by SAI port.

**How I verified it**
Assume 9100 is the MTU, Check by the command `ping -s 9100 10.0.0.57`

**Details if related**

### DIFF
--- a/orchagent/macsecorch.cpp
+++ b/orchagent/macsecorch.cpp
@@ -1275,6 +1275,18 @@ bool MACsecOrch::createMACsecPort(
             phy);
     });
 
+    if (!m_port_orch->getPortMtu(port.m_port_id, macsec_port.m_original_mtu))
+    {
+        SWSS_LOG_WARN("Cannot get Port MTU at the port %s", port_name.c_str());
+        return false;
+    }
+    m_port_orch->setMACsecEnabledState(port.m_port_id, true);
+    if (!m_port_orch->setPortMtu(port.m_port_id, macsec_port.m_original_mtu))
+    {
+        SWSS_LOG_WARN("Cannot set MTU to %u at the MACsec enabled port %s", macsec_port.m_original_mtu, port_name.c_str());
+        return false;
+    }
+
     if (phy)
     {
         if (!setPFCForward(port_id, true))
@@ -1540,6 +1552,13 @@ bool MACsecOrch::deleteMACsecPort(
     {
         SWSS_LOG_WARN("Cannot delete MACsec ingress port at the port %s", port_name.c_str());
         result &= false;
+    }
+
+    m_port_orch->setMACsecEnabledState(port.m_port_id, false);
+    if (!m_port_orch->setPortMtu(port.m_port_id, macsec_port.m_original_mtu))
+    {
+        SWSS_LOG_WARN("Cannot set MTU to %u at the port %s", macsec_port.m_original_mtu, port_name.c_str());
+        return false;
     }
 
     if (phy)
--- a/orchagent/macsecorch.cpp
+++ b/orchagent/macsecorch.cpp
@@ -1275,17 +1275,7 @@ bool MACsecOrch::createMACsecPort(
             phy);
     });
 
-    if (!m_port_orch->getPortMtu(port.m_port_id, macsec_port.m_original_mtu))
-    {
-        SWSS_LOG_WARN("Cannot get Port MTU at the port %s", port_name.c_str());
-        return false;
-    }
-    m_port_orch->setMACsecEnabledState(port.m_port_id, true);
-    if (!m_port_orch->setPortMtu(port.m_port_id, macsec_port.m_original_mtu))
-    {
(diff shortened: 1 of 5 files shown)


orchagent/portsorch.cpp
```diff
@@ -1144,6 +1172,11 @@ bool PortsOrch::setPortMtu(sai_object_id_t id, sai_uint32_t mtu)
     /* mtu + 14 + 4 + 4 = 22 bytes */
     attr.value.u32 = (uint32_t)(mtu + sizeof(struct ether_header) + FCS_LEN + VLAN_TAG_LEN);
 
+    if (isMACsecPort(id))
```
judyjoseph, line 1175: This is definitely needed for non-gearbox based devices. 
As discussed, do we need this additional 32 bytes to be added for the gbsyncd based devices as well ?

Pterosaur: Yes, we also needed it for line_side_port for gearbox. @jimmyzhai  will add more on this PR for adapting gearbox mode.

jimmyzhai: Have updated the fix to update port (w/o gearbox) MTU for MACsec.

Pterosaur (commented):
lgtm

orchagent/portsorch.cpp
```diff
@@ -1161,27 +1161,35 @@ bool PortsOrch::getPortAdminStatus(sai_object_id_t id, bool &up)
...
-                attr.value.u32, id, status);
+                attr.value.u32, port.m_port_id, status);
         task_process_status handle_status = handleSaiSetStatus(SAI_API_PORT, status);
         if (handle_status != task_success)
         {
             return parseHandleSaiStatusFailure(handle_status);
         }
     }
-    SWSS_LOG_INFO("Set MTU %u to port pid:%" PRIx64, attr.value.u32, id);
+
```
judyjoseph, line 1190: We can add the check to see if Gearbox is enabled before setting the gearbox attribute
PortsOrch::isGearboxEnabled

orchagent/macsecorch.cpp
```diff
@@ -1275,6 +1275,8 @@ bool MACsecOrch::createMACsecPort(
             phy);
     });
 
+    m_port_orch->setMACsecEnabledState(port_id, true);
+
```
judyjoseph, line 1279: Since we cannot guarantee that the macsec on a Port will be enabled before the MTU is set, I feel we need to set the MTU explicitly here. 

Or another way will be to check for macsec profile present in the port in config table and update the MTU accordingly in the PortsOrch::setPortMtu API.

Pterosaur: We will always update the MTU value in this function. Meanwhile, in the setPortMTU function, we will also check whether the MACsec enabled. So, I don't think there is any issue if enabling MACsec comes before the MTU setting.

orchagent/portsorch.cpp
Pterosaur: Done

judyjoseph (approved):
LGTM

judyjoseph:
@prsunny could you take a quick look too. thanks.

orchagent/portsorch.cpp
```diff
@@ -1161,27 +1161,62 @@ bool PortsOrch::getPortAdminStatus(sai_object_id_t id, bool &up)
...
+    if (isMACsecPort(port.m_port_id))
+    {
+        mtu -= MAX_MACSEC_SECTAG_SIZE;
+    }
+
+    return true;
+}
+
+bool PortsOrch::setPortMtu(const Port& port, sai_uint32_t mtu)
```
prsunny, line 1188: Why are we changing this API? Please have it as the original one and have the caller pass the extra value.

Pterosaur: Due to the Gearbox architecture support, the original function cannot semantically support the functionality of `set MTU`. We just follow other existing functions, like set Speed, set Fec and so on, that needed by Gearbox platform.
https://github.com/sonic-net/sonic-swss/blob/9ec471adc95b9cea1ffcdc6686a875f5ed70f9fe/orchagent/portsorch.cpp#L2191
https://github.com/sonic-net/sonic-swss/blob/9ec471adc95b9cea1ffcdc6686a875f5ed70f9fe/orchagent/portsorch.cpp#L1253

prsunny: ok, from the changes I still think you can reuse the existing function. But I see your point as making it consistent across other `set` functions. Approving from my side.

prsunny (approved):

yxieca:
This change causes 202205 PR test to fail.

Pterosaur:
> This change causes 202205 PR test to fail.

This PR: https://github.com/sonic-net/sonic-sairedis/pull/1107 is for fixing the failure.

PR #2407, opened 2022-08-04
[muxorch] return true if the nbr IP is in the skip_neighbors list

Ndancejic:
Orchagent was not programming neighbor IP to h/w L3 host table during
initialization because they were set as Standby. When the links became
active, SoC IPs were added to the skip_neighbors_ list and never
programmed. fix:

- Added public method MuxCable::getSkipNeighborsSet()
- check if nbr is in skip_neighbors_ when checking if cable is Active

related to: https://github.com/sonic-net/sonic-swss/pull/2369

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -962,6 +962,10 @@ bool MuxOrch::isNeighborActive(const IpAddress& nbr, const MacAddress& mac, stri
 
     if (ptr)
     {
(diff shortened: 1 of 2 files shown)


prsunny (commented):
Changes lgtm. Can you add a test as well?

lolyu (approved):

prsunny:
@Ndancejic , are the test failures due to the change?

PR #2410, opened 2022-08-10
[MuxOrch] Set internal class state to reflect the actual state

prsunny:
**What I did**
Set internal class variable to set to STANDBY.
Default value was set to INIT and not modified after setting to standby. 

**Why I did it**
Fix for https://github.com/sonic-net/sonic-buildimage/issues/11654

**How I verified it**

**Details if related**


bingwang-ms (approved):

prsunny:
@Ndancejic

qiluo-msft:

prsunny:
This is already part of the 202012 with another PR - https://github.com/sonic-net/sonic-swss/pull/2151

PR #2411, opened 2022-08-10
Delete invalid if condition in intfsorch.cpp

1002574903:
**What I did**
delete invalid if condition
**Why I did it**
Line 778 expression if always false because "else if" condition matches previous condition at line 731
image

**How I verified it**
Run test
**Details if related**

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -775,10 +775,6 @@ void IntfsOrch::doTask(Consumer &consumer)
                 }
                 adminStateChanged = true;
             }
-            else if (field == "nat_zone")
-            {
-                nat_zone = value;
-            }
             else if (field == "proxy_arp")
             {
                 proxy_arp = value;


orchagent/intfsorch.cpp
```diff
@@ -775,10 +775,6 @@ void IntfsOrch::doTask(Consumer &consumer)
                 }
                 adminStateChanged = true;
             }
-            else if (field == "nat_zone")
```
prsunny, line 778: Not sure I understand what is the issue/fix here. Please explain in PR description.

prsunny (changes requested):

orchagent/intfsorch.cpp
1002574903: duplicate else if (field == "nat_zone") （line731 and line778）

prsunny (approved):
thanks for clarifying

PR #2414, opened 2022-08-15
Enhance orchagent and buffer manager in error handling

stephenxs:
**What I did**
Enhance orchagent and buffer manager
- Buffer manager: do not insert buffer queue into cache if the profile is illegal, which prevents an empty string from being inserted into `APPL_DB` during initialization.
- orchagent: handle the case that a field referencing other objects is an empty string.
  There had been such logic that was broken by a PR last year.

**Why I did it**
Enhance the error handling logic.
In most cases, a user will not encounter such scenarios in a production environment because it's the front-ends' (eg. CLI) responsibility to identify the wrong configuration and prevent them from being inserted to `CONFIG_DB`.
However, in some cases, like a wrong `config_db.json` composed and copied to the switch, front-ends can not prevent that.

**How I verified it**
Manual and mock tests.

**Details if related**
For the improvement in buffer manager:
- previously, the logic was:
  - declare a reference `portQueue` to `m_portQueueLookup[port][queues]` and then assign `fvValue(i)` to `portQueue.running_profile_name`
  - But `[]` operation on C++ map has a side-effect -- it will insert a new element into the map if there wasn't one. In case the validation check in `checkBufferProfileDirection` failed and there was not one in the map, the `portQueue.running_profile_name` will keep empty. This is not what we want.
  - In case there was an item configured in the map, we should not remove it on failure because we want to prevent the user from being affected by misconfiguration and alert user to correct the error. There is log in `checkBufferProfileDirection`
- Now it is improved in this way:
  - Avoid using reference and initialize `m_portQueueLookup[port][queues]` only if there is a valid egress profile configured

### DIFF
--- a/cfgmgr/buffermgrdyn.cpp
+++ b/cfgmgr/buffermgrdyn.cpp
@@ -3110,8 +3110,7 @@ task_process_status BufferMgrDynamic::handleSingleBufferQueueEntry(const string 
 
     if (op == SET_COMMAND)
     {
-        auto &portQueue = m_portQueueLookup[port][queues];
-
+        bool successful = false;
         SWSS_LOG_INFO("Inserting entry BUFFER_QUEUE_TABLE:%s to APPL_DB", key.c_str());
 
         for (auto i : kfvFieldsValues(tuple))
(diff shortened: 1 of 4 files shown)


liat-grozovik:
@neethajohn , @prsunny  kindly reminder as this is a bug fixc for 202205 appreciate if you can prioritize it.

prsunny (approved):

neethajohn (approved):

PR #2415, opened 2022-08-16
[muxorch] Returning true if nbr in skip_neighbor_ in isNeighborActive()

Ndancejic:
PR #2407 added a check which returned true in isNeighborActive if it is
in the skip_neighbors_ list, but only checked for Mux in Subnet. This
extends that check to Mux Ports and Nexthop Muxes.

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -963,7 +963,7 @@ bool MuxOrch::isNeighborActive(const IpAddress& nbr, const MacAddress& mac, stri
 
     if (ptr)
     {
-        return ptr->isActive() || ptr->isSkipNeighbor(nbr);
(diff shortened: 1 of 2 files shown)


orchagent/muxorch.cpp
```diff
@@ -963,11 +963,7 @@ bool MuxOrch::isNeighborActive(const IpAddress& nbr, const MacAddress& mac, stri
 
     if (ptr)
     {
-        if (ptr->getSkipNeighborsSet().find(nbr) != ptr->getSkipNeighborsSet().end())
-        {
-            return true;
-        }
-        return ptr->isActive();
+        return ptr->isActive() | ptr->isSkipNeighbor(nbr);
```
prsunny, line 966: Can you please make to an || condition?

prsunny (approved):

yxieca (approved):

PR #2420, opened 2022-08-18
[Not for merge] Checking failures in  subinterface tests

liorghub:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


liorghub:
@dprital @prsunny 
Prince, I removed all the skips from test_sub_port_intf.py but in test log the name of the test cases do not appear. Am I missing something?

PR #2422, opened 2022-08-22
[QoS] Enforce drop probability only for colors whose WRED are enabled

stephenxs:
**What I did**

Do not enforce drop probability for a color whose WRED is disabled.

**Why I did it**

Currently, there is a logic to enforce the drop probability if it is not explicitly designated for a color. However, the drop probability is not a mandatory attribute. It can incur vendor SAI complaints to set it when the color is disabled.
The logic was introduced from the very beginning (by PR #571) because no drop probability was defined in the QoS template at the time, which is no longer true.
So we will enforce drop probability only if it is not configured and the color is enabled.

**How I verified it**

Unit test and manual test

**Details if related**

### DIFF
--- a/orchagent/qosorch.cpp
+++ b/orchagent/qosorch.cpp
@@ -46,6 +46,12 @@ enum {
     RED_DROP_PROBABILITY_SET    = (1U << 2)
 };
 
+enum {
+    GREEN_WRED_ENABLED  = (1U << 0),
+    YELLOW_WRED_ENABLED = (1U << 1),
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@stephenxs it is for master only or a bug fix for 202205?

liat-grozovik:
@neethajohn , @prsunny  could you please help to review and signoff?

stephenxs:
> @stephenxs it is for master only or a bug fix for 202205?

Master only. No need to check pick it.

neethajohn (approved):

PR #2427, opened 2022-08-24
[neighsyncd] Enabling ipv4 link local entries for non-dualtor

Ndancejic:
Allow ipv4 link local entries to be programmed to the hardware unless on
a dual-tor setup.

fixing sonic-net/sonic-buildimage#11830

### DIFF
--- a/tests/test_neighbor.py
+++ b/tests/test_neighbor.py
@@ -431,12 +431,12 @@ class TestNeighbor(object):
         # check application database
         tbl = swsscommon.Table(self.pdb, "NEIGH_TABLE:Ethernet8")
         intf_entries = tbl.getKeys()
-        assert len(intf_entries) == 0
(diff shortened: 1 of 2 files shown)


tests/test_neighbor.py
```diff
@@ -431,12 +431,12 @@ def test_Ipv4LinkLocalNeighbor(self, dvs, testlog):
         # check application database
         tbl = swsscommon.Table(self.pdb, "NEIGH_TABLE:Ethernet8")
         intf_entries = tbl.getKeys()
-        assert len(intf_entries) == 0
+        assert len(intf_entries) == 1
```
prsunny, line 434: Can we add a test to cover dualtor case?

dgsudharsan (approved):

prsunny (approved):

PR #2428, opened 2022-08-24
Change warning to notice for missing configuration of cable length

dprital:
**What I did**
Change warning log message to notice

**Why I did it**
When system is initialized with zero ports configures and port is created without cable length, we receive this warning message, (we see it when buffer configuration is disabled)
This message is seen on each time we are adding a port.
It is Non functional issue - the port is created normally and traffic on this port works normally.

**How I verified it**

**Details if related**


stephenxs (approved):

liat-grozovik:
@dprital i suggest you add the explaination to the code as from first review it is not clear why this flow now becomes a flow which is in some cases legit.

PR #2440, opened 2022-09-05
[DPU] Simplify SWSS initialization to meet DPU requirements.

oleksandrivantsiv:
Signed-off-by: Oleksandr Ivantsiv <oivantsiv@nvidia.com>

**What I did**
DPU supports features subset from Switch SONiC. To allow SWSS to run on DPU the initialization flow was simplified. During the start of the SWSS applications unsupported attributes are not configured. The list of SAI attributes supported by DPU is documented in DASH HLD.
- The precondition is to have the switch type as 'dpu'
- SAI Redis timeouts were extended such as for other switch types
- Only SAI calls are masked, the rest is the same using defaults. It is assumed that the user should not invoke flows/CLI commands which DPU does not support. No extra validation has been added.

**Why I did it**
To allow the SWSS application to run on DPU.

**How I verified it**
Run SWSS on DPU platform. Verify that unsupported attributes are not called.
Run SONiC as VS set as switch type DPU

**Details if related**

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -21,6 +21,7 @@ using namespace swss;
 extern sai_switch_api_t*           sai_switch_api;
 extern sai_object_id_t             gSwitchId;
 extern bool                        gSaiRedisLogRotate;
+extern string                      gMySwitchType;
 
 extern void syncd_apply_view();
 /*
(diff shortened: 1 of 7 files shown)


orchagent/orchdaemon.cpp
```diff
@@ -629,9 +630,12 @@ bool OrchDaemon::init()
 
     m_orchList.push_back(&CounterCheckOrch::getInstance(m_configDb));
 
-    vector<string> p4rt_tables = {APP_P4RT_TABLE_NAME};
-    gP4Orch = new P4Orch(m_applDb, p4rt_tables, vrf_orch, gCoppOrch);
-    m_orchList.push_back(gP4Orch);
+    if (gMySwitchType != "dpu")
```
prsunny, line 633: Is this applicable? This should not be impacting even if initialized, correct?

portsyncd/linksync.cpp
```diff
@@ -114,33 +115,39 @@ LinkSync::LinkSync(DBConnector *appl_db, DBConnector *state_db) :
             }
         }
 
-        for (idx_p = if_ni.get();
-                idx_p != NULL && idx_p->if_index != 0 && idx_p->if_name != NULL;
-                idx_p++)
+        /* In DPU SONiC this step is not needed as netdevs in Kernel are created in the early stage of the syncd service start, 
+         * when driver is loading. And exist while the driver is not unloaded.
+        */
+        if (g_switchType != "dpu")
```
prsunny, line 121: Can we just add the check and return if `dpu`. If other platform require this in future, lets address it accordingly.

oleksandrivantsiv:
@prsunny I resolved all the comments. Can you please take a look?

orchagent/orchdaemon.cpp
oleksandrivantsiv: After reworking of P4Orch this is not needed anymore.

prsunny (approved):

PR #2441, opened 2022-09-06
[ci][asan] add DVS tests run with ASAN

Yakiv-Huryk:
This adds 3 new stages:

* BuildAsan: same as 'Build' but with ENABLE_ASAN=y
* BuildDockerAsan: same as 'BuildDocker' but uses asan-enabled swss and
adds swss-dbg package. It also uses asan-enabled docker-sonic-vs from
the sonic-buildimage ci.
* TestAsan: same as 'Test' but uses asan-enabled docker and publishes
the asan reports if any. It also shows a warning if asan report dir is not empty
(which means that there is some memory usage issue in swss)

**What I did**
Added new CI stages to run the DVS tests with ASAN and check the ASAN reports.

**Why I did it**
To utilize ASAN for catching memory usage issues on PR.

**How I verified it**
Run the CI on my fork.

**Details if related**

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -49,6 +49,19 @@ stages:
       archive_pytests: true
       archive_gcov: true
 
+- stage: BuildAsan
+
+  jobs:
+  - template: .azure-pipelines/build-template.yml
+    parameters:
+      arch: amd64
+      sonic_slave: sonic-slave-buster
+      common_lib_artifact_name: common-lib
+      swss_common_artifact_name: sonic-swss-common
+      sairedis_artifact_name: sonic-sairedis
+      artifact_name: sonic-swss-asan
+      asan: true
+
 - stage: BuildArm
   dependsOn: Build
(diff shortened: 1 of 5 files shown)


liat-grozovik:
@xumia could you please help to review?

liat-grozovik:
@prsunny , @xumia  could you please help to review

azure-pipelines.yml
```diff
@@ -49,6 +49,19 @@ stages:
       archive_pytests: true
       archive_gcov: true
 
+- stage: BuildAsan
+
```
xumia, line 53: Does the BuildAsan stage have dependency on Build stage, if not, you can add the dependsOn:[] to skip it, it will run in parallel with Build stage.
```
- stage: BuildAsan
  dependsOn: []
```

Yakiv-Huryk: Good idea, thanks. I've updated the PR.

xumia (approved):

PR #2453, opened 2022-09-15
[nbrmgr] Interface cache for Neighbor refresh

LaveenBrcm:
**What I did**

Added Interface cache in nbrmgrd to handle neighbor refresh 

Nbrmgrd subscribes to Interface Table events and adss/deletes interface cache ==> Interface to IPAddr list info, whenever ip address is added or removed from an L3 physical interface.

**Why I did it**

Please refer https://github.com/sonic-net/SONiC/pull/1043

**How I verified it**

There are no new processing in OA based on this change, hence no update to Syncd/SAI, only the cache information is updated in nbrmgrd.

**Details if related**

This is one part of the change to support neighbor refresh functionality. For the complete list of related Code PRs, please refer to https://github.com/sonic-net/SONiC/pull/1043.

### DIFF
--- a/cfgmgr/nbrmgr.cpp
+++ b/cfgmgr/nbrmgr.cpp
@@ -41,8 +41,8 @@ static bool send_message(struct nl_sock *sk, struct nl_msg *msg)
     return rc;
 }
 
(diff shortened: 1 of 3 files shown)


prsunny:
This seems to be overloading the nbrmgrd. Need to understand the motivation and limitations with existing arp_update mechanism before this can be merged

PR #2459, opened 2022-09-20
orchagent: DASH changes

prabhataravind:
* Introduce DASH specific orchs - DashOrch, DashVnetOrch and DashRouteOrch

* Introduce changes to handle various DASH table schemas and operations

* Changes in bulker to introduce new constructors for DASH-related Entity/Object bulkers

* Changes to support new DASH tables and orchs in orchdaemon

* Changes to support new DASH APIs in saihelper

* Makefile changes to support compilation

* References: https://github.com/Azure/DASH/blob/main/documentation/general/design/dash-sonic-hld.md

* Verifications done: Manual testing with configurations referenced in the HLD above.

* Issues: Intermittent failures with bulk operations on DASH_VNET_MAPPING_TABLE, dependency related failures when deleting DASH_VNET and DASH_ENI objects when there are objects with references to these objects that are yet to be deleted.

Signed-off-by: Prabhat Aravind <paravind@microsoft.com>

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


orchagent/dash/dashvnetorch.h
```diff
@@ -0,0 +1,110 @@
...
+    DashVnetTable vnet_table_;
+    DashVnetMapTable vnet_map_table_;
+    ObjectBulker<sai_dash_vnet_api_t> vnet_bulker_;
+    EntityBulker<sai_dash_outbound_ca_to_pa_api_t> outbound_ca_to_pa_bulker_;
+    EntityBulker<sai_dash_pa_validation_api_t> pa_validation_bulker_;
+
+    void doTask(Consumer &consumer);
+    void doTaskVnetTable(Consumer &consumer);
+    void doTaskVnetMapTable(Consumer &consumer);
+    bool addVnet(string key, DashVnetBulkContext& ctxt);
```
Pterosaur, line 94: I feel it's better to add the qualifier of `const` and use reference, `const std::string &` for the arguments like `std::string` and `std::vector<xxx>`.

prabhataravind: I agree. I will make those changes.

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,914 @@
...
+    swss::copy(outbound_routing_entry.destination, ctxt.destination);
+    sai_attribute_t outbound_routing_attr;
+    vector<sai_attribute_t> outbound_routing_attrs;
+    auto& object_statuses = ctxt.object_statuses;
+
+    outbound_routing_attr.id = SAI_OUTBOUND_ROUTING_ENTRY_ATTR_DST_VNET_ID;
+    outbound_routing_attr.value.oid = gVnetNameToId[ctxt.vnet];
+    outbound_routing_attrs.push_back(outbound_routing_attr);
+
+    outbound_routing_attr.id = SAI_OUTBOUND_ROUTING_ENTRY_ATTR_OVERLAY_IP;
```
marian-pritsak, line 406: This attribute should be valid only for VNET_DIRECT_ROUTE

BTW, how does it even pass SAI metadata validation?

marian-pritsak (changes requested):

marian-pritsak:
Do we intend to set admin state of ENI?

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,914 @@
...
+            {
+                entry.admin_state = (fvValue(i) == "enabled" ? true : false);
+            }
+            if (fvField(i) == "vnet")
+            {
+                entry.vnet = fvValue(i);
+            }
+            if (fvField(i) == "qos")
+            {
+                entry.qos_name = fvValue(i);
```
Pterosaur, line 268: `else if` may be more efficient.

orchagent/dash/dashrouteorch.cpp
prabhataravind: I have addressed this.

orchagent/dash/dashrouteorch.cpp
prabhataravind: I have address this

prabhataravind:
> 

Yes, I have fixed this.

orchagent/dash/dashvnetorch.h
```diff
@@ -0,0 +1,110 @@
...
+    std::deque<sai_status_t> outbound_ca_to_pa_object_statuses;
+    std::deque<sai_status_t> pa_validation_object_statuses;
+    VnetMapBulkContext() {}
+
+    VnetMapBulkContext(const VnetMapBulkContext&) = delete;
+    VnetMapBulkContext(VnetMapBulkContext&&) = delete;
+
+    void clear()
+    {
+        outbound_ca_to_pa_object_statuses.clear();
```
oleksandrivantsiv, line 75: Do we need to clear "pa_validation_object_statuses" also?

orchagent/dash/dashorch.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    if (routing_type_entries_.find(routing_type) != routing_type_entries_.end())
+    {
+        SWSS_LOG_INFO("Routing type entry already exists for %s", routing_type.c_str());
+        return true;
+    }
+
+    routing_type_entries_[routing_type] = entry;
+
+    SWSS_LOG_ERROR("Routing type entry added %s", routing_type.c_str());
```
prsunny, line 197: this must be NOTICE. please fix

orchagent/dash/dashorch.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    if (routing_type_entries_.find(routing_type) == routing_type_entries_.end())
+    {
+        SWSS_LOG_INFO("Routing type entry does not exist for %s", routing_type.c_str());
+        return true;
+    }
+
+    routing_type_entries_.erase(routing_type);
+
+    SWSS_LOG_ERROR("Routing type entry removed for %s", routing_type.c_str());
```
prsunny, line 213: Please change to NOTICE

orchagent/dash/dashorch.cpp
```diff
@@ -0,0 +1,299 @@
...
+    SWSS_LOG_ENTER();
+
+    if (routing_type_entries_.find(routing_type) == routing_type_entries_.end())
+    {
+        SWSS_LOG_INFO("Routing type entry does not exist for %s", routing_type.c_str());
+        return true;
+    }
+
+    routing_type_entries_.erase(routing_type);
+
```
prsunny, line 212: Please remove extra line

orchagent/dash/dashorch.cpp
```diff
@@ -0,0 +1,299 @@
...
+    auto it = consumer.m_toSync.begin();
+    while (it != consumer.m_toSync.end())
+    {
+        KeyOpFieldsValuesTuple t = it->second;
+        string routing_type = kfvKey(t);
+        string op = kfvOp(t);
+
+        if (op == SET_COMMAND)
+        {
+            RoutingTypeEntry entry;
```
prsunny, line 230: As captured in schema, this can come as a list of actions. Please handle

orchagent/dash/dashrouteorch.h
```diff
@@ -0,0 +1,140 @@
...
+
+private:
+    EniTable eni_entries_;
+    QosTable qos_entries_;
+    RoutingTable routing_entries_;
+    RoutingRuleTable routing_rule_entries_;
+    EntityBulker<sai_dash_outbound_routing_api_t> outbound_routing_bulker_;
+    EntityBulker<sai_dash_inbound_routing_api_t> inbound_routing_bulker_;
+
+    void doTaskEniTable(Consumer &consumer);
```
prsunny, line 119: May be ENI and QoS can be moved to dashorch as those are device specific

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,994 @@
...
+    SWSS_LOG_ENTER();
+}
+
+bool DashRouteOrch::addEniEntry(const string& eni)
+{
+    SWSS_LOG_ENTER();
+
+    if (eni_entries_[eni].eniEntryAdded)
+    {
+        SWSS_LOG_ERROR("ENI entry already exists for %s", eni.c_str());
```
prsunny, line 51: Lets change such 'already exists' or 'object not found' cases to WARNING as northbound APIs are expected to be idempotent

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,994 @@
...
+        return true;
+    }
+
+    EniEntry &entry = eni_entries_[eni].eni_entry;
+    const string &vnet = entry.vnet;
+
+    if (!vnet.empty() && gVnetNameToId.find(vnet) == gVnetNameToId.end())
+    {
+        SWSS_LOG_ERROR("Retry as vnet %s not found", vnet.c_str());
```
prsunny, line 60: Please move this to INFO as we don't want retry logic to spew continuous logs

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,994 @@
...
+    return true;
+}
+
+bool DashRouteOrch::addEniAddrMapEntry(const string& eni)
+{
+    SWSS_LOG_ENTER();
+
+    if (eni_entries_[eni].eniAddrMapEntryAdded)
+    {
+        SWSS_LOG_ERROR("ENI ether address map entry already exists for %s", eni.c_str());
```
prsunny, line 115: Same as above comment for log levels

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,994 @@
...
+    {
+        SWSS_LOG_ERROR("Failed to create ENI entry for %s", eni.c_str());
+        task_process_status handle_status = handleSaiCreateStatus((sai_api_t) SAI_API_DASH_ENI, status);
+        if (handle_status != task_success)
+        {
+            return parseHandleSaiStatusFailure(handle_status);
+        }
+    }
+
+    eni_entries_[eni].eniEntryAdded = true;
```
prsunny, line 102: This is an overhead. I suggest checking eni by doing a 'find' of eni.

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,994 @@
...
+    SWSS_LOG_INFO("Created ENI entry for %s", eni.c_str());
+
+    return true;
+}
+
+bool DashRouteOrch::addEniAddrMapEntry(const string& eni)
+{
+    SWSS_LOG_ENTER();
+
+    if (eni_entries_[eni].eniAddrMapEntryAdded)
```
prsunny, line 113: This might add empty entries to map for `eni` if we are not careful about [] operations. Please use `find `operation for safer checks.

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,994 @@
...
+    {
+        SWSS_LOG_ERROR("ENI ether address map entry already exists for %s", eni.c_str());
+        return true;
+    }
+    if (!eni_entries_[eni].eniEntryAdded)
```
prsunny, line 118: Please use `find `to check

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,994 @@
...
+        string op = kfvOp(t);
+
+        if (op == SET_COMMAND)
+        {
+            QosEntry entry;
+
+            for (auto i : kfvFieldsValues(t))
+            {
+                if (fvField(i) == "qos_id")
+                    entry.qos_id = fvValue(i);
```
prsunny, line 367: Please use `{ } `for all if checks

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,994 @@
...
+            IpAddress& overlay_ip = ctxt.overlay_ip;
+            string& eni = ctxt.eni;
+            IpPrefix& destination = ctxt.destination;
+
+            vector<string> keys = tokenize(key, ':');
+            eni = keys[0];
+            destination = IpPrefix(keys[1]);
+
+            if (op == SET_COMMAND)
+
```
prsunny, line 585: extra line

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,994 @@
...
+    bool exists = (routing_rule_entries_.find(key) != routing_rule_entries_.end());
+    if (exists)
+    {
+        SWSS_LOG_ERROR("Inbound routing entry already exists for %s", key.c_str());
+        return true;
+    }
+    if (eni_entries_.find(ctxt.eni) == eni_entries_.end())
+    {
+        SWSS_LOG_ERROR("Retry as ENI entry %s not found", ctxt.eni.c_str());
```
prsunny, line 690: Please ensure retry logic are not spewing continuous logs, in all places

prsunny (changes requested):

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,994 @@
...
+    vector<sai_attribute_t> eni_attrs;
+
+    eni_attr.id = SAI_ENI_ATTR_VNET_ID;
+    eni_attr.value.oid = gVnetNameToId[entry.vnet];
+    eni_attrs.push_back(eni_attr);
+
+    bool has_qos = qos_entries_.find(entry.qos_name) != qos_entries_.end();
+
+    eni_attr.id = SAI_ENI_ATTR_PPS;
+    eni_attr.value.u32 = has_qos ? qos_entries_[entry.qos_name].bw : 0;
```
oleksandrivantsiv, line 75: The configuration should be set only if it is specified in the config file.
SONiC should not override the vendor's default configuration.
This is applicable to all attributes (SAI_ENI_ATTR_PPS, SAI_ENI_ATTR_CPS, SAI_ENI_ATTR_FLOWS).

oleksandrivantsiv (changes requested):

orchagent/dash/dashrouteorch.cpp
Pterosaur: IMHO, the default value of an SAI attribution should be unified to different vendors if the default value was declared in SAI. 
In the other words, if the default value wasn't specified by SWSS(upper layer of SONiC), the implementation of vendor should follow the SAI definition. e.g. the default value of SAI_ENI_ATTR_PPS is `0` in SAI.(https://github.com/opencomputeproject/SAI/blob/dash/experimental/saiexperimentaldasheni.h#L113)
So, I think to explicitly specified the default value in the upper layer is also acceptable if there is a default value in SAI.

prabhataravind: @oleksandrivantsiv, I tend to agree with @Pterosaur . The default value of 0 was considered as that was the value specified in SAI headers. Please let us know. We can discuss offline also on this if needed.

orchagent/bulker.h
```diff
@@ -11,6 +11,35 @@
...
+        _In_ sai_bulk_op_error_mode_t mode,
+        _Out_ sai_status_t *object_statuses);
+
+typedef sai_status_t (*sai_bulk_set_outbound_ca_to_pa_entry_attribute_fn) (
+        _In_ uint32_t object_count,
+        _In_ const sai_outbound_ca_to_pa_entry_t *entry,
+        _In_ const sai_attribute_t *attr_list,
+        _In_ sai_bulk_op_error_mode_t mode,
+        _Out_ sai_status_t *object_statuses);
+
```
theasianpianist, line 27: nit: extra newline

orchagent/bulker.h
```diff
@@ -104,6 +188,54 @@ namespace std
...
         }
     };
+
+    template <>
+    struct hash<sai_inbound_routing_entry_t>
+    {
+        size_t operator()(const sai_inbound_routing_entry_t& a) const noexcept
+        {
+            size_t seed = 0;
+            boost::hash_combine(seed, a.switch_id);
```
theasianpianist, line 198: Do we need to include the switch ID here? @prsunny I was under the impression that the switch ID is set at init time and never changes

orchagent/bulker.h
```diff
@@ -104,6 +188,54 @@ namespace std
...
+        }
+    };
+
+    template <>
+    struct hash<sai_outbound_ca_to_pa_entry_t>
+    {
+        size_t operator()(const sai_outbound_ca_to_pa_entry_t& a) const noexcept
+        {
+            size_t seed = 0;
+            boost::hash_combine(seed, a.switch_id);
```
theasianpianist, line 210: Same as above

orchagent/bulker.h
```diff
@@ -104,6 +188,54 @@ namespace std
...
+
+    template <>
+    struct hash<sai_inbound_routing_entry_t>
+    {
+        size_t operator()(const sai_inbound_routing_entry_t& a) const noexcept
+        {
+            size_t seed = 0;
+            boost::hash_combine(seed, a.switch_id);
+            boost::hash_combine(seed, a.eni_id);
+            return seed;
```
theasianpianist, line 200: Given that the behavioral model uses ENI, VNI, and SIP as the lookup key for the inbound_routing table, do we need to add VNI and SIP here?

orchagent/bulker.h
```diff
@@ -104,6 +188,54 @@ namespace std
...
+
+    template <>
+    struct hash<sai_outbound_ca_to_pa_entry_t>
+    {
+        size_t operator()(const sai_outbound_ca_to_pa_entry_t& a) const noexcept
+        {
+            size_t seed = 0;
+            boost::hash_combine(seed, a.switch_id);
+            boost::hash_combine(seed, a.dst_vnet_id);
+            return seed;
```
theasianpianist, line 212: Similar to above - behavioral model uses VNET ID and DIP as table keys, is it necessary to add DIP here?

orchagent/bulker.h
```diff
@@ -104,6 +188,54 @@ namespace std
...
+        }
+    };
+
+    template <>
+    struct hash<sai_pa_validation_entry_t>
+    {
+        size_t operator()(const sai_pa_validation_entry_t& a) const noexcept
+        {
+            size_t seed = 0;
+            boost::hash_combine(seed, a.switch_id);
```
theasianpianist, line 222: Same as above

orchagent/bulker.h
```diff
@@ -104,6 +188,54 @@ namespace std
...
+            boost::hash_combine(seed, a.vnet_id);
+            return seed;
```
theasianpianist, line 224: Similar to above - BM uses VNET ID and SIP

orchagent/bulker.h
```diff
@@ -104,6 +188,54 @@ namespace std
...
+        }
+    };
+
+    template <>
+    struct hash<sai_outbound_routing_entry_t>
+    {
+        size_t operator()(const sai_outbound_routing_entry_t& a) const noexcept
+        {
+            size_t seed = 0;
+            boost::hash_combine(seed, a.switch_id);
```
theasianpianist, line 234: Same as above

orchagent/bulker.h
```diff
@@ -104,6 +188,54 @@ namespace std
...
+            boost::hash_combine(seed, a.eni_id);
+            return seed;
```
theasianpianist, line 236: Similar to above - BM uses ENI and DIP

orchagent/dash/dashorch.cpp
theasianpianist: Should we wait until we've finalized how multiple actions will be represented in the schema to change this?

orchagent/dash/dashrouteorch.cpp
```diff
@@ -0,0 +1,994 @@
...
+    MacAddress mac_addr = MacAddress(entry.mac_address);
+    sai_eni_ether_address_map_entry_t eni_ether_address_map_entry;
+    eni_ether_address_map_entry.switch_id = gSwitchId;
+    memcpy(eni_ether_address_map_entry.address, mac_addr.getMac(), sizeof(sai_mac_t));
+
+    sai_attribute_t eni_ether_address_map_entry_attr;
+    eni_ether_address_map_entry_attr.id = SAI_ENI_ETHER_ADDRESS_MAP_ENTRY_ATTR_ENI_ID;
+    eni_ether_address_map_entry_attr.value.oid = entry.eni_id;
+
+    sai_status_t status = sai_dash_eni_api->create_eni_ether_address_map_entry(&eni_ether_address_map_entry, 1, &eni_ether_address_map_entry_attr);
```
theasianpianist, line 134: Please avoid magic constants, prefer `(uint32_t)eni_ether_address_map_entry_attr.size()` here to stick with convention

orchagent/dash/dashrouteorch.cpp
theasianpianist: I'm a little confused here - since the default value is declared in the SAI headers, is it still necessary for orchagent to pass a default value to the SAI API? My understanding is that if we don't provide this attribute the API should automatically fill it with the default value.

@prsunny for other SAI attributes with default values, is it preferred to still program those defaults from orchagent, or let the implementation take care of it?

prsunny: Agree with @theasianpianist , we don't have to provide default from orchagent

theasianpianist: Looks like at least for neighorch, we are allowing the SAI implementation to handle default values as specified in the SAI headers. For `SAI_NEIGHBOR_ENTRY_ATTR_PACKET_ACTION`, [default is defined as `SAI_PACKET_ACTION_FORWARD`](https://github.com/opencomputeproject/SAI/blob/master/inc/saineighbor.h#L65) and it doesn't look like we set a value for it when [adding a new neighbor](https://github.com/sonic-net/sonic-swss/blob/ab299201be9d37654e3ba88f9e7e180023896722/orchagent/neighorch.cpp#L798). Similarly for `SAI_NEIGHBOR_ENTRY_ATTR_META_DATA` [with a default of 0](https://github.com/opencomputeproject/SAI/blob/master/inc/saineighbor.h#L102), we allow the SAI implementation to set the default and do not use it when creating a new neighbor entry.

orchagent/dash/dashvnetorch.cpp
```diff
@@ -0,0 +1,723 @@
...
+        SWSS_LOG_ERROR("Vnet already exists for %s", vnet_name.c_str());
+        return true;
+    }
+
+    auto& object_ids = ctxt.object_ids;
+    sai_attribute_t dash_vnet_attr;
+    dash_vnet_attr.id = SAI_VNET_ATTR_VNI;
+    dash_vnet_attr.value.u32 = ctxt.vni;
+    object_ids.emplace_back();
+    vnet_bulker_.create_entry(&object_ids.back(), 1, &dash_vnet_attr);
```
theasianpianist, line 56: Another magic constant, please avoid

orchagent/dash/dashvnetorch.cpp
```diff
@@ -0,0 +1,723 @@
...
+    swss::copy(pa_validation_entry.sip, ctxt.underlay_ip);
+    auto& object_statuses = ctxt.pa_validation_object_statuses;
+    sai_attribute_t pa_validation_attr;
+
+    pa_validation_attr.id = SAI_PA_VALIDATION_ENTRY_ATTR_ACTION;
+    pa_validation_attr.value.u32 = SAI_PA_VALIDATION_ENTRY_ACTION_PERMIT;
+
+    object_statuses.emplace_back();
+    pa_validation_bulker_.create_entry(&object_statuses.back(), &pa_validation_entry,
+            1, &pa_validation_attr);
```
theasianpianist, line 321: Magic constant

orchagent/dash/dashorch.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    sai_vip_entry_t vip_entry;
+    vip_entry.switch_id = gSwitchId;
+    swss::copy(vip_entry.vip, entry.sip);
+    sai_attribute_t appliance_attr;
+    vector<sai_attribute_t> appliance_attrs;
+    sai_status_t status;
+    appliance_attr.id = SAI_VIP_ENTRY_ATTR_ACTION;
+    appliance_attr.value.u32 = SAI_VIP_ENTRY_ACTION_ACCEPT;
+    status = sai_dash_vip_api->create_vip_entry(&vip_entry, 1, &appliance_attr);
```
theasianpianist, line 51: Magic constant, please avoid

orchagent/dash/dashorch.cpp
```diff
@@ -0,0 +1,299 @@
...
+            return parseHandleSaiStatusFailure(handle_status);
+        }
+    }
+
+    sai_direction_lookup_entry_t direction_lookup_entry;
+    direction_lookup_entry.switch_id = gSwitchId;
+    direction_lookup_entry.vni = entry.vm_vni;
+    appliance_attr.id = SAI_DIRECTION_LOOKUP_ENTRY_ATTR_ACTION;
+    appliance_attr.value.u32 = SAI_DIRECTION_LOOKUP_ENTRY_ACTION_SET_OUTBOUND_DIRECTION;
+    status = sai_dash_direction_lookup_api->create_direction_lookup_entry(&direction_lookup_entry, 1, &appliance_attr);
```
theasianpianist, line 67: magic constant

prabhataravind: I followed the convention in routeorch.cpp and a few other orchs where the constant "1" is used when there is only a single attribute to be added. A "vector of sai_attribute_t" is not used in such cases to get the size from and the constant "1" is passed as the attribute count. 
I will create a variable attr_count and assign that to 1 instead for better clarity.

orchagent/dash/dashorch.cpp
prabhataravind: I will defer this change for now and handle it as a subsequent commit.

orchagent/dash/dashrouteorch.h
prabhataravind: Agreed.

orchagent/dash/dashrouteorch.cpp
prabhataravind: Will do.

orchagent/dash/dashrouteorch.cpp
prabhataravind: Will do.

orchagent/dash/dashrouteorch.cpp
prabhataravind: I will make the change to not set these values explicitly unless there is a specific qos configuration.

orchagent/dash/dashrouteorch.cpp
prabhataravind: As agreed, I have added a variable attr_count set to 1.

orchagent/bulker.h
prabhataravind: Switch ID does not change, but I kept the existing convention used for sai_route_entry, sai_fdb_entry etc for consistency.

orchagent/bulker.h
prabhataravind: I have made the change

orchagent/bulker.h
prabhataravind: Fixed this.

orchagent/dash/dashorch.h
```diff
@@ -0,0 +1,48 @@
...
+
+private:
+    ApplianceTable appliance_entries_;
+    RoutingTypeTable routing_type_entries_;
+    void doTask(Consumer &consumer);
+    void doTaskApplianceTable(Consumer &consumer);
+    void doTaskRoutingTypeTable(Consumer &consumer);
+    bool addApplianceEntry(const std::string& appliance_id, const ApplianceEntry &entry);
+    bool removeApplianceEntry(const std::string& appliance_id);
+    bool addRoutingTypeEntry(const std::string& routing_type, const RoutingTypeEntry &entry);
```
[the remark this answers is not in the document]
prabhataravind: Will fix this in a subsequent commit along with handling list of actions for DASH_ROUTING_TYPE_TABLE

orchagent/bulker.h
```diff
@@ -40,6 +32,13 @@ typedef sai_status_t (*sai_bulk_set_outbound_routing_entry_attribute_fn) (
         _In_ sai_bulk_op_error_mode_t mode,
         _Out_ sai_status_t *object_statuses);
 
+typedef sai_status_t (*sai_bulk_set_inbound_routing_entry_attribute_fn) (
```
theasianpianist, line 35: Curious about this change, does order in which these are defined matter?

prabhataravind: No, the order doesn't matter. I just thought of keeping pa_validation and outbound_ca_to_pa entries together as they correspond to APP_DB DASH_VNET_MAPPING_TABLE.

prsunny (approved):

PR #2461, opened 2022-09-20
[orchagent] Fix issue: ip prefix shall be inited even if VRF/VNET is not ready

Junchao-Mellanox:
**What I did**

Fix issue: ip prefix of a route flow pattern shall be initialized even if VRF/VNET is not ready

**Why I did it**

The issue cause a crash:

```
Sep 15 13:44:39.986661 sonic INFO swss#supervisord: orchagent terminate called after throwing an instance of 'std::logic_error'
Sep 15 13:44:39.986661 sonic INFO swss#supervisord: orchagent   what():  basic_string::_M_construct null not valid
```

**How I verified it**

Manual test and added new unit test.

**Details if related**

The flow in which IpPrefix had an uninitialized member, consists of:

1. FlowCounterRouteOrch::addRoutePattern() with 'Vnet1|100.1.1.1/32'

2. Vnet1 does not exist yet, but the logic of addRoutePattern ignores that, assuming default vrf was meant and due to the way parseRouteKeyForRoutePattern works IpPrefix is left unitialized

3. Vnet1 is created

4. Flow counters are enabled, generateRouteFlowStats() -> createRouteFlowCounterByPattern() -> createRouteFlowCounterFromVnetRoutes() is called

5. Because we inserted a route pattern with defalt vrf id and unititliazed IpPrefix it leads to a crash

### DIFF
--- a/orchagent/flex_counter/flowcounterrouteorch.cpp
+++ b/orchagent/flex_counter/flowcounterrouteorch.cpp
@@ -636,7 +636,7 @@ void FlowCounterRouteOrch::createRouteFlowCounterByPattern(const RoutePattern &r
             {
                 return;
             }
-            
+
             if (route_pattern.is_match(route_pattern.vrf_id, entry.first))
             {
                 if (isRouteAlreadyBound(route_pattern, entry.first))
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@Junchao-Mellanox is it required for 202205 as well? if so, need to set label

Junchao-Mellanox:
Hi @prsunny , could you please help review?

prsunny (approved):

Junchao-Mellanox:
Hi @yxieca , could you please help cherry-pick to 202205? Thanks

PR #2468, opened 2022-09-27
Buffermgr adjusted according to the new fast reboot implementation

stephenxs:
**What I did**
1. Adjust fast reboot logic because WarmStart::isWarmStart() will be true for both warm/fast start in the new infra
2. Support mock functions used by warm relevant classes: DBConnector::get and Table::hset operations. 

Depends on sonic-net/sonic-swss-common#691

**Why I did it**
There is an optimization for fast reboot in dynamic buffer manager but now we are not able to distinguish between fast and warm reboot.
So we need to adjust it accordingly.

**How I verified it**

Manual test and mock test.

**Details if related**

### DIFF
--- a/cfgmgr/buffermgrdyn.cpp
+++ b/cfgmgr/buffermgrdyn.cpp
@@ -13,6 +13,7 @@
 #include "shellcmd.h"
 #include "schema.h"
 #include "warm_restart.h"
+#include "restart_waiter.h"
 
 /*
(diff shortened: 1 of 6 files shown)


stephenxs:
Test is failing because the depending PR has not been in.

stephenxs:
Closing it as it doesn't work as expected.

PR #2470, opened 2022-09-28
[dash]: ACL orchagent

Pterosaur:
**What I did**
Add DASH ACL orchagent implementation.

**Why I did it**
DASH needs ACL function.

**How I verified it**
```
$ sudo pytest -sv --num-ports=2 test_dash_acl.py --pdb
================================ test session starts ================================
platform linux -- Python 3.8.10, pytest-4.6.2, py-1.10.0, pluggy-0.13.1 -- /usr/bin/python3
cachedir: .pytest_cache
plugins: flaky-3.7.0
collected 4 items

test_dash_acl.py::TestAcl::test_acl_flow PASSED
test_dash_acl.py::TestAcl::test_acl_group PASSED
test_dash_acl.py::TestAcl::test_acl_rule PASSED
test_dash_acl.py::test_nonflaky_dummy PASSED

============================= 4 passed in 54.83 seconds =============================

```

**Details if related**


.azure-pipelines/build-template.yml
```diff
@@ -92,29 +93,29 @@ jobs:
   - task: DownloadPipelineArtifact@2
     inputs:
       source: specific
-      project: build
-      pipeline: Azure.sonic-sairedis
-      artifact: ${{ parameters.sairedis_artifact_name }}
+      project: nvidia
```
theasianpianist, line 96: Looks like this project is not accessible from `mssonic`: https://dev.azure.com/mssonic/build/_build/results?buildId=165280&view=logs&jobId=83516c17-6666-5250-abde-63983ce72a49&j=18fdc12d-bdfb-52d7-1391-aaa772f0f664

theasianpianist: Maybe it's better to add new steps for downloading the the DASH-specific DVS/libraries? This way when we eventually merge to master it won't interfere with the regular DVS image build

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,889 @@
...
+        attrs.back().value.u16rangelist.list = acl_rule.m_dst_ports.value().data();
+    }
+
+    if (!is_existing)
+    {
+        // Mandatory on create attributes should be setted, otherwise assign the default value.
+        // If the attributes don't have default value, just skip and wait for the user to set the value at the next message
+        if (!acl_rule.m_action || !acl_rule.m_terminating)
+        {
+            acl_rule.m_action == DashAclRuleEntry::Action::ALLOW;
```
theasianpianist, line 630: Are these default values defined somewhere/how did we decied on these?

Pterosaur: To the default value of protocol, just refer: https://github.com/Azure/DASH/blob/main/documentation/general/dash-sonic-hld.md#324-acl
```
if not provided, match on all protocols
```
To the action, just explicit assigning according to the SAI definition(https://github.com/opencomputeproject/SAI/blob/dash/experimental/saiexperimentaldashacl.h#L98).
But per our previous discussion, I can remove these explicit assigning.

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,881 @@
...
+    }
+
+    opt = value;
+
+    return true;
+}
+
+sai_attr_id_t getSaiStage(DashAclDirection d, sai_ip_addr_family_t f, DashAclStage s)
+{
+    const static map<tuple<DashAclDirection, sai_ip_addr_family_t, DashAclStage>, sai_attr_id_t> StageMaps =
```
theasianpianist, line 260: Would it be better to declare this outside this function just so the code is a bit cleaner?

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,881 @@
...
+
+void DashAclOrch::doTask(Consumer &consumer)
+{
+    SWSS_LOG_ENTER();
+
+    using TaskType = tuple<const string, const string>;
+    using TaskFunc = task_process_status (DashAclOrch::*)(
+        const string &,
+        const TaskArgs &);
+    const static map<TaskType, TaskFunc> TaskMap = {
```
theasianpianist, line 313: Same as above map, is it better to declare outside the function/at the top of the file?

orchagent/dash/dashaclorch.cpp
Pterosaur: IMHO, it's better to set the scope of a variable as small as possible.
Meanwhile, the initialization point of a static variable in a function will be determined at the first time that the function is called.  But if it's a global variable, the initialization sequence will be hard to determined. I believe a certain initialization sequence will be easier for maintaining if we want to expose these variables to other modules in the future.

orchagent/dash/dashaclorch.cpp
Pterosaur: Same as above reply.

.azure-pipelines/build-template.yml
Pterosaur: Fixed in another PR: https://github.com/sonic-net/sonic-swss/pull/2501

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+        stage = DashAclStage::STAGE4;
+    }
+    else if (buffer == "5")
+    {
+        stage = DashAclStage::STAGE5;
+    }
+    else
+    {
+        SWSS_LOG_ERROR("Invalid stage : %s", buffer.c_str());
+        throw std::invalid_argument("Invalid stage");
```
theasianpianist, line 91: nit: remove 'std::' namespace for consistency with other methods

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+        terminating = true;
+    }
+    else if (buffer == "false" || buffer == "False" || buffer == "0")
+    {
+        terminating = false;
+    }
+    else
+    {
+        SWSS_LOG_ERROR("Invalid terminating : %s", buffer.c_str());
+        throw std::invalid_argument("Invalid terminating");
```
theasianpianist, line 152: same as above

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+        {
+            // Range
+            SWSS_LOG_ERROR("Invalid port range : %s", token.c_str());
+            throw invalid_argument("Invalid port range");
+        }
+        ports.push_back(port);
+    }
+}
+
+}
```
theasianpianist, line 216: is this brace needed? seems to be extra

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+            attrs.back().value.s32 = *(acl_group.m_ip_version);
+        }
+    }
+
+    if (!attrs.empty())
+    {
+        if (!is_existing)
+        {
+            // Create a new ACL group
+            acl_group.m_status = m_dash_acl_group_bulker.create_entry(&acl_group.m_dash_acl_group_id, static_cast<uint32_t>(attrs.size()), attrs.data());
```
theasianpianist, line 447: Typically when we use bulk operations there is an associated 'post' method that is called after the bulk operation is finished, e.g. `addRoute` and `addRoutePost` in route orch. Do we need the same pattern here?

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+            acl_group.m_rule_count = 0;
+            m_defer_funcs.push_back([this, key]() {
+                if (m_dash_acl_group_table[key].m_dash_acl_group_id != SAI_NULL_OBJECT_ID)
```
theasianpianist, line 450: Do we need any error handling here in case the create fails?

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+            attrs.emplace_back();
+            attrs.back().id = SAI_DASH_ACL_RULE_ATTR_PROTOCOL;
+            attrs.back().value.u8list.count = static_cast<uint32_t>(acl_rule.m_protocols.value().size());
+            attrs.back().value.u8list.list = acl_rule.m_protocols.value().data();
+        }
+
+        if (!acl_rule.m_priority || !acl_rule.m_dst_ips || !acl_rule.m_src_ips || !acl_rule.m_dst_ports || !acl_rule.m_src_ports)
+        {
+            SWSS_LOG_WARN("ACL rule %s doesn't have all mandatory attributes, waiting for user to set the value", key.c_str());
+            return task_success;
```
theasianpianist, line 642: I think it makes more sense to return `task_failed` here since the rule was not actually created

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+    {
+        if (!is_existing)
+        {
+            // Create a new ACL rule
+
+            attrs.emplace_back();
+            attrs.back().id = SAI_DASH_ACL_RULE_ATTR_DASH_ACL_GROUP_ID;
+            attrs.back().value.oid = acl_group->m_dash_acl_group_id;
+
+            acl_rule.m_status = m_dash_acl_rule_bulker.create_entry(&acl_rule.m_dash_acl_rule_id, static_cast<uint32_t>(attrs.size()), attrs.data());
```
theasianpianist, line 656: same question as above re: bulk operations

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+            acl_group->m_rule_count++;
+
+            m_defer_funcs.push_back([this, group_id, key]() {
+                if (m_dash_acl_rule_table[key].m_dash_acl_rule_id != SAI_NULL_OBJECT_ID)
+                {
+                    SWSS_LOG_NOTICE("Created ACL rule %s", key.c_str());
+                }
+                else
+                {
+                    getAclGroup(group_id)->m_rule_count--;
```
theasianpianist, line 666: Might be a good idea to add an error log here?

orchagent/dash/dashaclorch.cpp
Pterosaur: It's needed for line: https://github.com/sonic-net/sonic-swss/blob/957e9288c98a4d026edb65e3c33c8ca1a2bd5582/orchagent/dash/dashaclorch.cpp#L61

prsunny (commented):
You would need to maintain a ref count b/w eni and acl group to prevent deleting group while it is already bind to an ENI.

orchagent/dash/dashaclorch.cpp
Pterosaur: Done, I changed the bulk operation to normal SAI operation.

orchagent/dash/dashaclorch.cpp
Pterosaur: Done

orchagent/dash/dashaclorch.cpp
Pterosaur: IMHO, we can allow to create an object with multiple APPL_DB messages. It means if the upper layer only provide partial metadata of the object that's acceptable instead of a failure. So, I return task_success instead of task_failed if the metadata isn't complete.

orchagent/dash/dashaclorch.cpp
Pterosaur: Done

orchagent/dash/dashaclorch.cpp
Pterosaur: Done

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+    if (acl_group == nullptr)
+    {
+        SWSS_LOG_WARN("ACL group %s doesn't exist", key.c_str());
+        return task_success;
+    }
+
+    // The member rules of group should be removed first
+    if (acl_group->m_rule_count != 0)
+    {
+        SWSS_LOG_WARN("ACL group %s still has %d rules", key.c_str(), acl_group->m_rule_count);
```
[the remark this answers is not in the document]
Pterosaur: Done

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+                if (m_dash_acl_group_table[key].m_dash_acl_group_id != SAI_NULL_OBJECT_ID)
+                {
+                    SWSS_LOG_NOTICE("Created ACL group %s", key.c_str());
+                }
+            });
+        }
+        else
+        {
+            // Update the ACL group's attributes
+            for (const auto &attr : attrs)
```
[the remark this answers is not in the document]
Pterosaur: OK, I removed the update block.

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+    auto acl_group = getAclGroup(key);
+
+    if (acl_group == nullptr)
+    {
+        SWSS_LOG_WARN("ACL group %s doesn't exist", key.c_str());
+        return task_success;
+    }
+
+    // The member rules of group should be removed first
+    if (acl_group->m_rule_count != 0)
```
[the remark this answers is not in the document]
Pterosaur: Done. Add a ref count to check that.

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+    if (!extractVariables(key, ':', group_id, rule_id))
+    {
+        SWSS_LOG_ERROR("Failed to parse key %s", key.c_str());
+        return task_failed;
+    }
+
+    auto acl_group = getAclGroup(group_id);
+    if (acl_group == nullptr)
+    {
+        SWSS_LOG_WARN("ACL group %s doesn't exist, waiting for group creating before creating rule %s", group_id.c_str(), rule_id.c_str());
```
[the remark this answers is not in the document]
Pterosaur: Done

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+        attrs.back().value.u16rangelist.list = acl_rule.m_dst_ports.value().data();
+    }
+
+    if (!is_existing)
+    {
+        // Mandatory on create attributes should be setted, otherwise assign the default value.
+        // If the attributes don't have default value, just skip and wait for the user to set the value at the next message
+        if (!acl_rule.m_protocols)
+        {
+            const static vector<uint8_t> all_protocols = {
```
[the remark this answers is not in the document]
Pterosaur: Done

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+                }
+                else
+                {
+                    getAclGroup(group_id)->m_rule_count--;
+                }
+            });
+        }
+        else
+        {
+            // Update the ACL group's attributes
```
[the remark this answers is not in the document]
Pterosaur: Done

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+    if (!extractVariables(key, ':', eni, stage))
+    {
+        SWSS_LOG_ERROR("Invalid key : %s", key.c_str());
+        return task_failed;
+    }
+
+    auto eni_entry = m_dash_orch->getEni(eni);
+    if (eni_entry == nullptr)
+    {
+        SWSS_LOG_WARN("eni %s cannot be found", eni.c_str());
```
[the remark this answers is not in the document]
Pterosaur: Done.

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+        {
+            SWSS_LOG_WARN("acl group %s cannot be found, wait for create", acl.m_acl_group_id->c_str());
+            acl.m_acl_group_id.reset();
+            return task_need_retry;
+        }
+
+        attr.id = getSaiStage(direction, *(acl_group->m_ip_version), stage);
+        attr.value.oid = acl_group->m_dash_acl_group_id;
+    }
+    else
```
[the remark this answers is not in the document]
Pterosaur: If the input string isn't valid or the new value equals than the old one.

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+    {
+        SWSS_LOG_WARN("ACL rule %s doesn't exist", key.c_str());
+        return task_success;
+    }
+
+    auto &acl_rule = itr->second;
+
+    bool is_existing = acl_rule.m_dash_acl_rule_id != SAI_NULL_OBJECT_ID;
+
+    if (!is_existing)
```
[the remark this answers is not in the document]
Pterosaur: In my opinion, a ACL rule can only be added to only one ACL group. So that the ref count is always one. If yes, why we need a count to record it?

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,873 @@
...
+    {
+        attrs.emplace_back();
+        attrs.back().id = SAI_DASH_ACL_RULE_ATTR_PRIORITY;
+        attrs.back().value.u32 = *(acl_rule.m_priority);
+    }
+
+    bool update_action = false;
+    update_action |= updateValue(data, "action", acl_rule.m_action);
+    update_action |= updateValue(data, "terminating", acl_rule.m_terminating);
+    if (update_action || (acl_rule.m_action && acl_rule.m_terminating && !is_existing))
```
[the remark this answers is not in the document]
Pterosaur: Here is an optimization I want.
I would always only pass these updated attributes to the SAI. But how can I know which attributes are updated if I move this to another section? I have no idea only if I introduce another repeated code to check and add them one by one to the `attrs`.

prsunny (approved):

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,844 @@
...
+        if (task != TaskMap.end())
+        {
+            task_status = (this->*task->second)(
+                kfvKey(message),
+                kfvFieldsValues(message));
+        }
+        else
+        {
+            SWSS_LOG_ERROR(
+                "Unknown task : %s - %s",
```
theasianpianist, line 331: Might be a good idea to include the actual task/key in this message as well to help with debugging

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,844 @@
...
+                table_name.c_str(),
+                op.c_str());
+        }
+
+        if (task_status == task_need_retry)
+        {
+            SWSS_LOG_DEBUG(
+                "Task %s - %s need retry",
```
theasianpianist, line 339: Same as above

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,844 @@
...
+                table_name.c_str(),
+                op.c_str());
+            ++itr;
+        }
+        else
+        {
+            if (task_status != task_success)
+            {
+                SWSS_LOG_WARN("Task %s - %s fail",
```
theasianpianist, line 348: Same as above

orchagent/dash/dashaclorch.cpp
```diff
@@ -0,0 +1,844 @@
...
+                              table_name.c_str(),
+                              op.c_str());
+            }
+            else
+            {
+                SWSS_LOG_DEBUG(
+                    "Task %s - %s success",
```
theasianpianist, line 355: Same as above

orchagent/dash/dashaclorch.cpp
Pterosaur: I believe all actual tasks/key have been logged in /var/log/swss/swss.rec by lib swsscommon so that we don't need repeated recording.

PR #2481, opened 2022-10-04
[orchdaemon]: Fixed sairedis record file rotation

bacrossland:
**What I did**
Fix Azure/sonic-buildimage#8162

Moved sairedis record file rotation logic out of flush() to fix issue.

**Why I did it**
Sairedis record file was not releasing the file handle on rotation. This is because the file handle release was inside the flush() which was only being called if a select timeout was triggered. Moved the logic to its own function which is called in the start() loop.

**How I verified it**
Ran a script to fill log and verified that rotation was happening correctly.

Signed-off-by: Bryan Crossland bryan.crossland@target.com

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -55,7 +55,6 @@ int gBatchSize = DEFAULT_BATCH_SIZE;
 bool gSairedisRecord = true;
 bool gSwssRecord = true;
 bool gLogRotate = false;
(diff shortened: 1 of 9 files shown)


bacrossland:
@qiluo-msft This is the PR for back porting https://github.com/sonic-net/sonic-swss/pull/2299 to 202012

PR #2484, opened 2022-10-06
[portinit] Do not call GET on SAI_PORT_ATTR_SPEED when AUTONEG is enabled

vaibhavhd:
**What I did**
During port init if AutoNeg is enabled do not do GET oper on `SAI_PORT_ATTR_SPEED` (do not call `getPortSpeed`)

**Why I did it**
This PR fixes an issue where in some platforms syncd crashes when AutoNeg is enabled and port is oper down.
The crash happens in the warmboot recovery path:
1. In the target image as part of portinit, GET operation on `SAI_PORT_ATTR_SPEED` returns a random value.
4. This random value does not match the speed in the base image.
5. Diff in speed causes syncd comparison logic to attempt to set newly detected speed.
6. Comparison logic crashes in APPLY_VIEW when doing a SET speed operation on port with the new speed.

Why SAI returns random value in GET oper on `SAI_PORT_ATTR_SPEED`:
SAI returns random value as when autoneg is enabled attribute `SAI_PORT_ATTR_SPEED` is not set in the first place.
When autoneg is enabled a list of speeds is set using attribute `SAI_PORT_ATTR_ADVERTISED_SPEED` (instead of `SAI_PORT_ATTR_SPEED` when autoneg is not enabled.

In autoneg enabled case, get oper on `SAI_PORT_ATTR_SPEED` is not allowed as the speed on the port is not certain, and instead depends on advertised/negotiated speed.

The autoneg speed handling is done separately after portinit logic.
https://github.com/sonic-net/sonic-swss/blob/e04bb436eabe135eca2482852dc492599432d44b/orchagent/portsorch.cpp#L3294

https://github.com/sonic-net/sonic-swss/blob/e04bb436eabe135eca2482852dc492599432d44b/orchagent/portsorch.cpp#L3508

**How I verified it**

Built and installed a new swss image on a physical device.

Without this change:
```
2022-12-02.21:39:53.553522|g|SAI_OBJECT_TYPE_PORT:oid:0x100000000005b|SAI_PORT_ATTR_ADMIN_STATE=false
2022-12-02.21:39:53.554071|G|SAI_STATUS_SUCCESS|SAI_PORT_ATTR_ADMIN_STATE=true
2022-12-02.21:39:53.554090|g|SAI_OBJECT_TYPE_PORT:oid:0x100000000005b|SAI_PORT_ATTR_SPEED=0
2022-12-02.21:39:53.556303|G|SAI_STATUS_SUCCESS|SAI_PORT_ATTR_SPEED=25000
```

With this change, a GET call on AUTO_NEG_MODE is done, if FALSE, GET on PORT_ATTR_SPEED is done:
```
2022-12-02.22:13:07.852038|g|SAI_OBJECT_TYPE_PORT:oid:0x100000000005b|SAI_PORT_ATTR_ADMIN_STATE=true
2022-12-02.22:13:07.852562|G|SAI_STATUS_SUCCESS|SAI_PORT_ATTR_ADMIN_STATE=true
2022-12-02.22:13:07.852581|g|SAI_OBJECT_TYPE_PORT:oid:0x100000000005b|SAI_PORT_ATTR_AUTO_NEG_MODE=true
2022-12-02.22:13:07.852942|G|SAI_STATUS_SUCCESS|SAI_PORT_ATTR_AUTO_NEG_MODE=false
2022-12-02.22:13:07.852961|g|SAI_OBJECT_TYPE_PORT:oid:0x100000000005b|SAI_PORT_ATTR_SPEED=0
2022-12-02.22:13:07.855123|G|SAI_STATUS_SUCCESS|SAI_PORT_ATTR_SPEED=25000
```

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -4681,9 +4681,15 @@ bool PortsOrch::initializePort(Port &port)
         return false;
     }
 
-    /* initialize port admin speed */
-    if (!getPortSpeed(port.m_port_id, port.m_speed))
+    if (port.m_autoneg == 1 && !getPortAdvSpeeds(port, true, port.m_adv_speeds))
     {
+        /* initialize port advertised speed */
+        SWSS_LOG_ERROR("Failed to get initial port advertised speed %d", port.m_adv_speeds);
+        return false;
+    }
+    else if (!getPortSpeed(port.m_port_id, port.m_speed))
+    {
+        /* initialize port admin speed */
         SWSS_LOG_ERROR("Failed to get initial port admin speed %d", port.m_speed);
         return false;
     }
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2495,11 +2495,7 @@ bool PortsOrch::isAutoNegEnabled(sai_object_id_t id)
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to get port AutoNeg status for port pid:%" PRIx64, id);
-        task_process_status handle_status = handleSaiGetStatus(SAI_API_PORT, status);
-        if (handle_status != task_process_status::task_success)
-        {
-            return false;
-        }
+        return false;
     }
 
     return attr.value.booldata;
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -4681,15 +4681,14 @@ bool PortsOrch::initializePort(Port &port)
         return false;
     }
 
+    /* initialize port advertised speed */
     if (port.m_autoneg == 1 && !getPortAdvSpeeds(port, true, port.m_adv_speeds))
     {
-        /* initialize port advertised speed */
-        SWSS_LOG_ERROR("Failed to get initial port advertised speed %d", port.m_adv_speeds);
+        SWSS_LOG_ERROR("Failed to get initial port advertised speed");
         return false;
     }
-    else if (!getPortSpeed(port.m_port_id, port.m_speed))
+    else if (!getPortSpeed(port.m_port_id, port.m_speed)) /* initialize port admin speed */
     {
-        /* initialize port admin speed */
         SWSS_LOG_ERROR("Failed to get initial port admin speed %d", port.m_speed);
         return false;
     }
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2481,6 +2481,27 @@ bool PortsOrch::getQueueTypeAndIndex(sai_object_id_t queue_id, string &type, uin
     return true;
 }
 
+bool PortsOrch::isAutoNegEnabled(sai_object_id_t id)
+{
+    SWSS_LOG_ENTER();
+
+    sai_attribute_t attr;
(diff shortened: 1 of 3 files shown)


prsunny (approved):
lgtm

orchagent/portsorch.cpp
```diff
@@ -2484,6 +2484,27 @@ bool PortsOrch::getQueueTypeAndIndex(sai_object_id_t queue_id, string &type, uin
     return true;
 }
 
+bool PortsOrch::isAutoNegEnabled(sai_object_id_t id)
```
dgsudharsan, line 2487: Why do we  use m_autoneg in port to identify if autoneg is enabled or not?

orchagent/portsorch.cpp
```diff
@@ -4758,7 +4779,7 @@ bool PortsOrch::initializePort(Port &port)
     }
 
     /* initialize port admin speed */
-    if (!getPortSpeed(port.m_port_id, port.m_speed))
+    if (!isAutoNegEnabled(port.m_port_id) && !getPortSpeed(port.m_port_id, port.m_speed))
```
dgsudharsan, line 4782: I don't understand the intention here. Are we getting speed only inf autoneg is not enabled? The title says get adv speed but I don't see get adv speed not called. Can you please explain more?

dgsudharsan (changes requested):

orchagent/portsorch.cpp
vaibhavhd: This is in the context of `initializePort`. I believe at that time, `m_autoneg` is not initialized.

orchagent/portsorch.cpp
vaibhavhd: We are polling for `SAI_PORT_ATTR_SPEED` speed only when auto neg is not enabled. 

https://github.com/opencomputeproject/SAI/blob/566d4a8ef2e90bee986133042c4f8c2d7478a107/inc/saiport.h#L772

As when autoneg is enabled, `SAI_PORT_ATTR_SPEED` is not set. We set and poll `SAI_PORT_ATTR_ADVERTISED_SPEED`

https://github.com/opencomputeproject/SAI/blob/566d4a8ef2e90bee986133042c4f8c2d7478a107/inc/saiport.h#L819

As for the GET call on adv speed, that is done here:
https://github.com/sonic-net/sonic-swss/blob/e04bb436eabe135eca2482852dc492599432d44b/orchagent/portsorch.cpp#L3464

dgsudharsan (approved):

prgeor:
@vaibhavhd can you populate the section `How I verified it`. Also please test this change in 202012 and 202205 separately

prgeor:
@vaibhavhd as discussed, if you can 
1. update the title of the PR
2. provide more context in the description why SAI_PORT_ATTR_SPEED is not supposed to be called

Also, lets take up with community on what should be returned by SAI_PORT_ATTR_SPEED if speed is NOT configured if AN is configured.

prgeor (approved):

qiluo-msft:

yxieca:
201811 cherry-picking is a precautionary action. There is no need for this change to go into any branch between 201811 and 202012.

PR #2490, opened 2022-10-18
Vxlan ecmp tunnel with priority Endpoints.[202012]

siqbal1986:
Added support for Priority endpoints in Vxlan ECMP tunnel.

**What I did**
This commit does not amend the existing behavior of the system. This change adds support for creating two NHG's (primary and secondary) so they can be switched when primary endpoints go down.
When a secondary endpoint comes back up, the next hop group with secondary endpoints is replaced with the primary one for the route.
If the primary endpoints list is provided,  it must be the same as the **"endpoint"** list.
**Why I did it**
To add support for Primary /Backup endpoint routing behavior.
**How I verified it**
Added 3  test cases to verify the functionality and existing tests are passing.
**Details if related**

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -686,7 +686,7 @@ bool VNetRouteOrch::hasNextHopGroup(const string& vnet, const NextHopGroupKey& n
 
 bool VNetRouteOrch::isActiveNextHopGroup(const string& vnet, const NextHopGroupKey& nexthops)
 {
-    if(syncd_nexthop_groups_[vnet].find(nexthops) != syncd_nexthop_groups_[vnet].end())
+    if (syncd_nexthop_groups_[vnet].find(nexthops) != syncd_nexthop_groups_[vnet].end())
     {
         return (!syncd_nexthop_groups_[vnet][nexthops].active_members.empty());
     }
@@ -896,7 +896,7 @@ bool VNetRouteOrch::selectNextHopGroup(const string& vnet,
     }
     if (!hasNextHopGroup(vnet, nexthops_secondary))
     {
-        SWSS_LOG_INFO("Creating secondary next hop group  %s", nexthops_primary.to_string().c_str());
+        SWSS_LOG_INFO("Creating secondary next hop group  %s", nexthops_secondary.to_string().c_str());
 
         if (!createNextHopGroup(vnet, nexthops_secondary, vrf_obj, monitors))
         {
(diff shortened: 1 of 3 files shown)


orchagent/vnetorch.cpp
```diff
@@ -1045,6 +1053,7 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
         {
             syncd_tunnel_routes_.erase(vnet);
         }
+        removePriorityEndpoints(vnet, ipPrefix);
```
prsunny, line 1056: You are invoking this after erasing `ipPrefix `from `syncd_tunnel_routes_ `but this is being accessed in your function. The could become an invalid access. Please move this to line 1050.

orchagent/vnetorch.cpp
```diff
@@ -1055,6 +1064,97 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
...
+    {
+        SWSS_LOG_INFO("Failed to find priority tunnel route entry, prefix %s\n",
+        ipPrefix.to_string().c_str());
+        return;
+    }
+    auto old_primary_nhgk = it_route->second.primary;
+    auto old_all_nhgk = it_route->second.all;
+
+    std::set<NextHopKey> nhks_primary = old_primary_nhgk.getNextHops();
+    for (auto nhk: nhks_primary)
```
prsunny, line 1094: Not sure I understand this logic. You want to erase `ipPrefix `from `priority_endpoints_`, right? Why not just loop and erase?

orchagent/vnetorch.cpp
```diff
@@ -1055,6 +1064,97 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
     return true;
 }
 
+void VNetRouteOrch::addPriorityEndpoints(const string& vnet, IpPrefix& ipPrefix, NextHopGroupKey& primary,  NextHopGroupKey& all)
+{
+    std::set<NextHopKey> nhks_primary = primary.getNextHops();
+    std::set<NextHopKey> nhks_all = all.getNextHops();
+    for (auto nkh : nhks_all)
```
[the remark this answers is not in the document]
siqbal1986: we want to keep track of all the endpoints, both primary and secondary. When a secondary EP will come up, we will check if all the primaries are down and then use that secondary to create NHG.

orchagent/vnetorch.cpp
siqbal1986: that correct. Ill fix that.

orchagent/vnetorch.cpp
siqbal1986: yes, I could use the iterator to erase it. ill change it.

siqbal1986: done

orchagent/vnetorch.cpp
```diff
@@ -986,7 +987,14 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
...
         syncd_nexthop_groups_[vnet][nexthops].tunnel_routes.insert(ipPrefix);
 
-        syncd_tunnel_routes_[vnet][ipPrefix] = nexthops;
+        VNetTunnelRouteEntry tunnel_route_entry;
+        tunnel_route_entry.ngh_key = nexthops;
+        tunnel_route_entry.all = nexthops_all;
+        tunnel_route_entry.primary = nexthops;
+        syncd_tunnel_routes_[vnet][ipPrefix] = tunnel_route_entry;
+
+	    addPriorityEndpoints(vnet, ipPrefix, nexthops, nexthops_all);
```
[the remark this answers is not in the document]
siqbal1986: done.

orchagent/vnetorch.cpp
```diff
@@ -1893,14 +1998,25 @@ bool VNetRouteOrch::handleTunnel(const Request& request)
         SWSS_LOG_ERROR("Peer monitor size of %zu does not match endpoint size of %zu", monitor_list.size(), ip_list.size());
         return false;
     }
-    
+    if (!primary_list.empty() && monitor_list.empty())
+    {
+        SWSS_LOG_ERROR("Primary/backup behaviour cannot function without endpoint monitoring.");
+        return false;
```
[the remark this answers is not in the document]
siqbal1986: done.

PR #2500, opened 2022-10-27
[202012][muxorch] Adding case for maintaining current state

Ndancejic:
What I did:
Added a case where if current and previous Mux states are the same, post a log entry that acknowleges this change and returns without doing anything.

Why I did it:
Valid state changes such as active-active and
standby-standby were causing error logs

How I tested it:
added a case to check active-active state change
and standby-standby state change.


prsunny (approved):

Ndancejic:
https://github.com/sonic-net/sonic-swss/pull/2280

tests/test_mux.py
```diff
@@ -1098,6 +1102,22 @@ def test_neighbor_miss_no_peer(
         for ip in test_ips:
             self.check_neighbor_state(dvs, dvs_route, ip, expect_route=False)
 
+    def test_mux_states(
```
prsunny, line 1105: Please rebase the test same as master branch PR

PR #2501, opened 2022-10-28
[azp]: Add Azp for DASH

Pterosaur:

**What I did**

Add Azp job for DASH.

**Why I did it**

DASH needs CLI.

**How I verified it**

Check Azp.

**Details if related**


.azure-pipelines/build-template.yml
```diff
@@ -84,38 +116,39 @@ jobs:
       artifact: ${{ parameters.swss_common_artifact_name }}
       runVersion: 'latestFromBranch'
       runBranch: 'refs/heads/$(BUILD_BRANCH)'
-      path: $(Build.ArtifactStagingDirectory)/download
+      allowFailedBuilds: ${{ parameters.allow_failed_artifact }}
```
theasianpianist, line 119: Why do we need to allow failed builds here?

azure-pipelines-dash.yml
```diff
@@ -0,0 +1,86 @@
+# C/C++ with GCC
+# Build your C/C++ project with GCC using make.
+# Add steps that publish test results, save build artifacts, deploy, and more:
+# https://docs.microsoft.com/azure/devops/pipelines/apps/c-cpp/gcc
+
+pr:
```
theasianpianist, line 6: Does this file need to be merged with the existing `azure-pipelines.yml`? Looks like the automatic PR build will only trigger jobs in `azure-pipelines.yml`

.azure-pipelines/build-template.yml
```diff
@@ -84,38 +116,39 @@ jobs:
       artifact: ${{ parameters.swss_common_artifact_name }}
       runVersion: 'latestFromBranch'
       runBranch: 'refs/heads/$(BUILD_BRANCH)'
```
theasianpianist, line 118: Looks like the swsscommon dash branch doesn't differ from the master branch - will this change in the future? Should we continue to use master branch here since the swsscommon job keeps failing due to the sairedis repo not having a dash branch?

prabhataravind: My understanding is that swsscommon for dash will always be on master branch. @prsunny to confirm.

.azure-pipelines/build-template.yml
prabhataravind: The default value for allow_failed_artifact is false. So perhaps that is ok?

theasianpianist: Looks like it's set to True in `azure-pipeline-dash.yml` below

.azure-pipelines/build-template.yml
theasianpianist: Discussed with @prsunny and @prabhataravind, we should always be using the master branch of swsscommon - can you add a new parameter like `swss_common_branch_name` so we can specify this for DASh jobs?

Pterosaur: Done

.azure-pipelines/build-template.yml
Pterosaur: Because we are using the branch, dash, of sonic-swss-common which also leverage the sairedis code in Azp: https://dev.azure.com/mssonic/build/_build/results?buildId=166787&view=logs&jobId=61b7eb0d-78fa-5af5-6df0-b6f82d3891a8&j=61b7eb0d-78fa-5af5-6df0-b6f82d3891a8&t=70f86778-6fb4-5097-2f4f-43a5d7a20e70 . But in the repo, sonic-sairedis, there is no related branch named 'dash'.
I don't think we can directly use the master branch of sonic-swss-common, because other artifacts from DASH like 'syncd' and so on were using  a specified commit of sonic-swss-common what was listed in sonic-buildimage.

azure-pipelines-dash.yml
Pterosaur: Done, please check it.

azure-pipelines-dash.yml
```diff
@@ -0,0 +1,89 @@
...
+
+stages:
+- stage: Build
+  jobs:
+  - template: .azure-pipelines/build-template.yml
+    parameters:
+      arch: amd64
+      sonic_slave: sonic-slave-buster
+      swss_common_artifact_name: sonic-swss-common
+      swss_common_branch: dash
```
theasianpianist, line 39: Please change to master branch as per https://github.com/sonic-net/sonic-swss/pull/2501#discussion_r1012389884

azure-pipelines-dash.yml
```diff
@@ -0,0 +1,89 @@
...
+
+- stage: BuildDocker
+  dependsOn: Build
+  condition: succeeded('Build')
+  jobs:
+  - template: .azure-pipelines/build-docker-sonic-vs-template.yml
+    parameters:
+      swss_common_artifact_name: sonic-swss-common
+      swss_common_branch: dash
+      swss_artifact_name: sonic-swss
```
theasianpianist, line 63: Please change to master branch as per https://github.com/sonic-net/sonic-swss/pull/2501#discussion_r1012389884

azure-pipelines-dash.yml
```diff
@@ -0,0 +1,89 @@
...
+- stage: Test
+  dependsOn: BuildDocker
+  condition: succeeded('BuildDocker')
+  jobs:
+  - template: .azure-pipelines/test-docker-sonic-vs-template.yml
+    parameters:
+      log_artifact_name: log
+      gcov_artifact_name: sonic-gcov
+      sonic_slave: sonic-slave-buster
+      swss_common_branch: dash
```
theasianpianist, line 85: Please change to master branch as per https://github.com/sonic-net/sonic-swss/pull/2501#discussion_r1012389884

.azure-pipelines/build-template.yml
theasianpianist: As per https://github.com/sonic-net/sonic-swss/pull/2501#discussion_r1012389884, we should be using the master branch of swss common. In this case, I don't think we still need to allow failed builds.

.azure-pipelines/build-docker-sonic-vs-template.yml
```diff
@@ -43,18 +83,27 @@ jobs:
       pipeline: Azure.sonic-swss-common
       artifact: ${{ parameters.swss_common_artifact_name }}
       runVersion: 'latestFromBranch'
-      runBranch: 'refs/heads/$(BUILD_BRANCH)'
+      runBranch: 'refs/heads/${{ parameters.swss_common_branch }}'
+      allowFailedBuilds: ${{ parameters.allow_failed_artifact }}
```
theasianpianist, line 87: Since we're using the master branch of swss common, please remove this

theasianpianist (changes requested):

theasianpianist (approved):
thanks!

PR #2503, opened 2022-10-31
[dual-tor] add missing SAI attribte in order to create IPNIP tunnel

ayurkiv-nvda:

**What I did**
add SAI_TUNNEL_ATTR_DECAP_TTL_MODE and SAI_TUNNEL_ATTR_DECAP_DSCP_MODE

**Why I did it**
Need to pass additional SAI attrib. Align encap end decap params

1 **SAI_TUNNEL_ATTR_DECAP_DSCP_MODE** 
Problem:
 
Long time ago this attribute was defined as “mandatory on create” in SAI API, but at some point it was changed to “optional”.
But MLNX SAI implementation didn’t change it and still expects this attribute to be always provided on tunnel creation.
SONiC “dual ToR” implementation, when it creates IP-in-IP tunnel, does not provide this attribute because it is “optional”.
So we get SAI error for creating tunnel because as mentioned above MLNX SAI implementation still expects this attribute as “mandatory on create”.
Need to always set this attribute in “Dual ToR orchagent” on IP-in-IP tunnel creation

2 **SAI_TUNNEL_ATTR_DECAP_TTL_MODE**
Problem:

Long time ago this attribute was defined as “mandatory on create” in SAI API, but at some point it was changed to “optional”.
SONiC “dual ToR” implementation, when it creates IP-in-IP tunnel, does not provide this attribute because now it is “optional”.

SAI team reported:
_"We only support pipe, which is not the default API value of uniform, so user must provide explicit pipe
This has always been the ip in ip behavior and limitation, which is also documented in our RN"_

So it is MLNX SAI implementation limitation that we support only “pipe” and since default value for this attribute is “uniform” we need to set it explicitly.
And in our case it is becomes “mandatory on create” attribute which must be set to “pipe” value.

As a result of this limitation we should update SONiC “Dual ToR” code in order to pass this attribute on IP-in-IP tunnel creation.

**How I verified it**
tunnel successfully created

**Details if related**

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -212,6 +212,10 @@ static sai_object_id_t create_tunnel(
     attr.value.s32 = SAI_TUNNEL_TTL_MODE_PIPE_MODEL;
     tunnel_attrs.push_back(attr);
 
+    attr.id = SAI_TUNNEL_ATTR_DECAP_TTL_MODE;
+    attr.value.s32 = SAI_TUNNEL_TTL_MODE_PIPE_MODEL;
+    tunnel_attrs.push_back(attr);
+
     if (dscp_mode_name == "uniform" || dscp_mode_name == "pipe")
     {
         sai_tunnel_dscp_mode_t dscp_mode;
@@ -224,6 +228,10 @@ static sai_object_id_t create_tunnel(
             dscp_mode = SAI_TUNNEL_DSCP_MODE_PIPE_MODEL;
         }
         attr.id = SAI_TUNNEL_ATTR_ENCAP_DSCP_MODE;
+        attr.value.s32 = dscp_mode;
+        tunnel_attrs.push_back(attr);
+
+        attr.id = SAI_TUNNEL_ATTR_DECAP_DSCP_MODE;
         attr.value.s32 = dscp_mode;
         tunnel_attrs.push_back(attr);
     }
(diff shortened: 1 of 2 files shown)


orchagent/muxorch.cpp
```diff
@@ -212,6 +212,10 @@ static sai_object_id_t create_tunnel(
     attr.value.s32 = SAI_TUNNEL_TTL_MODE_PIPE_MODEL;
     tunnel_attrs.push_back(attr);
 
+    attr.id = SAI_TUNNEL_ATTR_DECAP_TTL_MODE;
```
prsunny, line 215: This tunnel is used only as nexthop tunnel for encap purposes. Why is this required here?

ayurkiv-nvda: Regarding **SAI_TUNNEL_ATTR_DECAP_TTL_MODE**
 
Problem:
Long time ago this attribute was defined as “mandatory on create” in SAI API, but at some point it was changed to “optional”.
SONiC “dual ToR” implementation, when it creates IP-in-IP tunnel, does not provide this attribute because now it is “optional”.

SAI team reported:
_"We only support pipe, which is not the default API value of uniform, so user must provide explicit pipe
This has always been the ip in ip behavior and limitation, which is also documented in our RN"_
 
So it is MLNX SAI implementation limitation that we support only “pipe” and since default value for this attribute is “uniform” we need to set it explicitly.
And in our case it is becomes “mandatory on create” attribute which must be set to “pipe” value.
 
 As a result of this limitation we should update SONiC “Dual ToR” code in order to pass this attribute on IP-in-IP tunnel creation.

prsunny (approved):
lgtm. @bingwang-ms , do you've any comments?

bingwang-ms (approved):

bingwang-ms:
LGTM. I need to watch the test result of master image to confirm the extra attribute does't cause issue.

ayurkiv-nvda:
@bingwang-ms  @prsunny it looks like after branch alignment with latest master all checks passed successfully.
Can you please approve PR again?

bingwang-ms (approved):

ayurkiv-nvda:
can someone merge this PR to upstream?

PR #2504, opened 2022-11-01
Fix vs test issue: failed to remove vlan due to referenced by vlan interface

stephenxs:
**What I did**

Fix: remove the VLAN interface before removing the VLAN itself

**Why I did it**

There is a logic in VLAN vs test cases `test_VlanGratArp` and `test_VlanProxyArp` to create a VLAN, a VLAN interface and then remove the VLAN and then check whether the VLAN is removed from SAI.
However, the step to remove the VLAN interface is missed, which causes the VLAN to be referenced and prevents it from being deleted.

**How I verified it**

Run vs test.

**Details if related**

### DIFF
--- a/tests/dvslib/dvs_vlan.py
+++ b/tests/dvslib/dvs_vlan.py
@@ -46,6 +46,10 @@ class DVSVlan(object):
         member = "Vlan{}|{}".format(vlanID, interface)
         self.config_db.delete_entry("VLAN_MEMBER", member)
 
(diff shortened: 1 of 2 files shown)


stephenxs:
Hi @prsunny 
Could you please review this PR? It fixes VLAN vs test issue which blocks other PRs
Thanks

prsunny (approved):

lolyu:
Added `request for 202205` as the `test_vlan.py` suffers from the same issue.

PR #2508, opened 2022-11-10
[Bugfix]Vxlan ecmp configuration idempotent behavior fix.[202012]

siqbal1986:
The vxlan tunnel feature goes into an unpredictable state if the same tunnel configuration is applied twice.
Any further configuration update worsens the problem.
**What I did**
Added a fix to check if the configuration updates has no impact and avoid internal state update.
**Why I did it**
Bug causes feature failure.
**How I verified it**
Added 5 tests to verify idempotent behavior
These tests check with a single endpoint with/without BFD, and multiple endpoints with/without BFD scenarios.
**Details if related**

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -958,11 +958,11 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
             }
         }
 
-        if (it_route != syncd_tunnel_routes_[vnet].end())
+        if (it_route != syncd_tunnel_routes_[vnet].end() && it_route->second != nexthops)
         {
-            // In case of updating an existing route, decrease the reference count for the previous nexthop group
+            // In case of updating an existing route, decrease the reference count for the previous nexthop group if not same as new nhg
             NextHopGroupKey nhg = it_route->second;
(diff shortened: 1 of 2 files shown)


orchagent/vnetorch.cpp
```diff
@@ -983,13 +983,14 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
...
 
-        syncd_nexthop_groups_[vnet][nexthops].tunnel_routes.insert(ipPrefix);
-
-        syncd_tunnel_routes_[vnet][ipPrefix] = nexthops;
-        syncd_nexthop_groups_[vnet][nexthops].ref_count++;
-        vrf_obj->addRoute(ipPrefix, nexthops);
-
+            syncd_tunnel_routes_[vnet][ipPrefix] = nexthops;
+            syncd_nexthop_groups_[vnet][nexthops].ref_count++;
+            vrf_obj->addRoute(ipPrefix, nexthops);
```
prsunny, line 992: Can you check if its ok to call `postRouteState` if `second != nexthops`?

prsunny (approved):

prsunny:
Please raise a PR to master as well.

PR #2511, opened 2022-11-11
[ResponsePublisher] add pipeline support 

stepanblyschak:

**DEPENDS ON** https://github.com/sonic-net/sonic-swss-common/pull/708

**What I did**

I added ResponsePublisher support for redis pipeline

**Why I did it**

I did it to improve performance when sending many responses. Responses are buffered in redis client before beeing sent out to redis server, while orchagent has no more pending tasks to do, responses are flushed to redis.

**How I verified it**

I did manual tests and verified togather with bgp suppress feature.

**Details if related**


liat-grozovik:
@marian-pritsak @prsunny could you please help to review?

orchagent/response_publisher.cpp
```diff
@@ -90,7 +90,10 @@ void RecordResponse(const std::string &response_channel, const std::string &key,
 
 } // namespace
 
-ResponsePublisher::ResponsePublisher() : m_db("APPL_STATE_DB", 0)
+ResponsePublisher::ResponsePublisher(bool buffered) :
+    m_db(std::make_unique<swss::DBConnector>("APPL_STATE_DB", 0)),
+    m_pipe(std::make_unique<swss::RedisPipeline>(m_db.get())),
```
qiluo-msft, line 95: >m_db [](http://example.com/codeflow?start=49&length=4)

You are assuming m_pipe initialized after m_db. The sequence is tricky. Suggest move m_pipe to function body.

orchagent/response_publisher.cpp
```diff
@@ -107,17 +110,14 @@ void ResponsePublisher::publish(const std::string &table, const std::string &key
     }
 
     std::string response_channel = "APPL_DB_" + table + "_RESPONSE_CHANNEL";
-    if (m_notifiers.find(table) == m_notifiers.end())
-    {
-        m_notifiers[table] = std::make_unique<swss::NotificationProducer>(&m_db, response_channel);
-    }
+    swss::NotificationProducer notificationProducer{m_pipe.get(), response_channel, m_buffered};
```
qiluo-msft, line 113: >NotificationProducer [](http://example.com/codeflow?start=10&length=20)

Is it too expensive to create a new NotificationProducer for each function call?

orchagent/response_publisher.cpp
```diff
@@ -140,17 +140,14 @@ void ResponsePublisher::publish(const std::string &table, const std::string &key
 void ResponsePublisher::writeToDB(const std::string &table, const std::string &key,
                                   const std::vector<swss::FieldValueTuple> &values, const std::string &op, bool replace)
 {
-    if (m_tables.find(table) == m_tables.end())
-    {
-        m_tables[table] = std::make_unique<swss::Table>(&m_db, table);
-    }
+    swss::Table applStateTable{m_pipe.get(), table, m_buffered};
```
qiluo-msft, line 143: >applStateTable [](http://example.com/codeflow?start=16&length=14)

Is it too expensive to create a new Table for each function call?

orchagent/response_publisher.cpp
stepanblyschak: @qiluo-msft Same reply https://github.com/sonic-net/sonic-swss-common/pull/708#discussion_r1070343699

orchagent/response_publisher.cpp
stepanblyschak: I find it cheap, just arguments copy. It might be cheaper then a hash table lookup, depending on hash function and a load factor, so without it it takes a bit more deterministic time.

In my testing old and new implementation show the same result (measured published responses/sec), probably bound to redis IO anyway.

orchagent/response_publisher.cpp
stepanblyschak: @qiluo-msft  Same as above https://github.com/sonic-net/sonic-swss/pull/2511#discussion_r1070350279

qiluo-msft (approved):

orchagent/response_publisher.cpp
qiluo-msft: I assume the argument is "arguments copy". Seem the constructor is not just arguments copy. For example: `TableBase::TableBase()` includes a map find.

qiluo-msft (changes requested):
Add new comment

mint570 (commented):
How frequent is the OrchDaemon::flush being called?

Also, I saw a PR on doing pipeline on the notification API. Won't that covers this one already since the publisher just uses the notification producer.

orchagent/response_publisher.cpp
stepanblyschak: @qiluo-msft So it's one find with an integer key + arguments copy vs 2 finds with string keys.
I did a benchmark to prove my point and got these results:

```c++
static void BM_newResponsePublisher(benchmark::State& state) {
    ResponsePublisher publisher{};

    for (auto _ : state) {
        publisher.publish("SOME_TABLE", "SOME_KEY", {}, ReturnCode(SAI_STATUS_SUCCESS));
    }
}

static void BM_oldResponsePublisher(benchmark::State& state) {
    old::ResponsePublisher publisher{};

    for (auto _ : state) {
        publisher.publish("SOME_TABLE", "SOME_KEY", {}, ReturnCode(SAI_STATUS_SUCCESS));
    }
}
```

Compiled with:

```
g++ -std=c++14 -O2 -I ../orchagent/ -I /usr/include/swss -I /usr/include/sai/ ../orchagent/response_publisher.cpp ../orchagent/response_publisher_old.cpp main.cpp -o main -lswsscommon -lhiredis -lbenchmark
```

Full code is here - https://github.com/stepanblyschak/sonic-swss/blob/response-publisher-benchmarks/benchmarks/main.cpp

```
stepanb@333f1eb97224:/sonic/src/sonic-swss/benchmarks$ ./main 
2023-01-29T17:42:27+00:00
Running ./main
Run on (12 X 1600 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x12)
  L1 Instruction 32 KiB (x12)
  L2 Unified 256 KiB (x12)
  L3 Unified 15360 KiB (x2)
Load Average: 0.55, 0.32, 0.30
------------------------------------------------------------------
Benchmark                        Time             CPU   Iterations
------------------------------------------------------------------
BM_newResponsePublisher      61831 ns        46291 ns        14984
BM_oldResponsePublisher      63855 ns        47892 ns        15570
```

Most of the time it is IO bound of course, but I constantly get a little bit better result with new implementation every time a run the benchmark.  So, the caching makes it even a little bit worse and the new implementation is 3.16 % faster.  Please note, that both don't use redis pipeline.

stepanblyschak:
> How frequent is the OrchDaemon::flush being called?
> 
> Also, I saw a PR on doing pipeline on the notification API. Won't that covers this one already since the publisher just uses the notification producer.

@mint570 It is at least every second. It is not just about adding pipeline support to NotificationProducer. Someone must flush the pipeline when it is not full, otherwise events stay in the buffer until the buffer itself is not full.
It completelly follows the design of orchagent's async sairedis mode.

qiluo-msft (dismissed):

mint570:
> > How frequent is the OrchDaemon::flush being called?
> > Also, I saw a PR on doing pipeline on the notification API. Won't that covers this one already since the publisher just uses the notification producer.
> 
> @mint570 It is at least every second. It is not just about adding pipeline support to NotificationProducer. Someone must flush the pipeline when it is not full, otherwise events stay in the buffer until the buffer itself is not full. It completelly follows the design of orchagent's async sairedis mode.

This change will help in the batch cases, but it will take at most 1 second in a single request case.

stepanblyschak:
@mint570 Orch can call Orch::flushResponses() on demand

mint570 (approved):

qiluo-msft:
>                 copp_ut.cpp \

Why remove existing test file?

---
Refers to: tests/mock_tests/Makefile.am:33 in 235bbde. [](commit_id = 235bbde4aa2b0ed3629d26b26a7a3d11ee0932dc, deletion_comment = True)

qiluo-msft:
>                 warmrestartassist_ut.cpp \

Why remove existing test file?

---
Refers to: tests/mock_tests/Makefile.am:52 in 235bbde. [](commit_id = 235bbde4aa2b0ed3629d26b26a7a3d11ee0932dc, deletion_comment = True)

stepanblyschak:
@qiluo-msft I didn't touch these files

prsunny:
@stepanblyschak , can you please check code coverage?

qiluo-msft (approved):

PR #2512, opened 2022-11-11
[RouteOrch] Record ROUTE_TABLE entry programming status to APPL_STATE_DB

stepanblyschak:

**DEPENDS ON** https://github.com/sonic-net/sonic-swss/pull/2511

**What I did**

I added ROUTE_TABLE entry programming status recording to APPL_STATE_DB. I made a ResponsePublisher work in buffered mode to be more optimized for route bulk processing.

I always publish a route entry with return code SAI_STATUS_SUCCESS because orchagent immidiatelly exits on any error.

**Why I did it**

I did it to support route supression in BGP. Orchagent has to sends route programming feedback back to fpmsyncd which communicates with zebra.

**How I verified it**

I wrote UT, also on the switch, you can run:

```
admin@sonic:~$ redis-cli -n 14 keys '*ROUTE*'
```

**Details if related**


orchagent/routeorch.cpp
```diff
@@ -630,6 +632,11 @@ void RouteOrch::doTask(Consumer& consumer)
 
                     if (fvField(i) == "seg_src")
                         srv6_source = fvValue(i);
+
+                    if (fvField(i) == "protocol")
```
prsunny, line 636: Why do we have to introduce this field?

orchagent/routeorch.cpp
```diff
@@ -2574,3 +2591,22 @@ void RouteOrch::decNhgRefCount(const std::string &nhg_index)
...
+void RouteOrch::publishRouteState(const RouteBulkContext& ctx, const ReturnCode& status)
+{
+    SWSS_LOG_ENTER();
+
+    std::vector<FieldValueTuple> fvs;
+
+    /* If the operation type is "DEL" then ctx.protocol is empty and fvs is left empty.
+     * An empty fvs makes ResponsePublisher::publish() remove the state entry from APPL_STATE_DB
+     */
+    if (!ctx.protocol.empty())
```
prsunny, line 2604: Why introduce this field for identifying operation? I'm not sure if i understand the logic here. Can you please explaing this?

stepanblyschak: It is not the motivation for introducing this field. The context does not carry operation type, however, if the protocol is left empty that means it is a delete operation, since the delete update does not carry FVS (including protocol)

stepanblyschak: @prsunny Replaced with a flag saved in the context object, please check.

prsunny (dismissed):
lgtm,

tests/mock_tests/Makefile.am
```diff
@@ -165,7 +165,7 @@ tests_intfmgrd_INCLUDES = $(tests_INCLUDES) -I$(top_srcdir)/cfgmgr -I$(top_srcdi
 tests_intfmgrd_CFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_GTEST) $(CFLAGS_SAI)
 tests_intfmgrd_CPPFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_GTEST) $(CFLAGS_SAI) $(tests_intfmgrd_INCLUDES)
 tests_intfmgrd_LDADD = $(LDADD_GTEST) $(LDADD_SAI) -lnl-genl-3 -lhiredis -lhiredis \
-        -lswsscommon -lswsscommon -lgtest -lgtest_main -lzmq -lnl-3 -lnl-route-3 -lpthread
+        -lswsscommon -lswsscommon -lgtest -lgtest_main -lzmq -lnl-3 -lnl-route-3 -lpthread -lgmock -lgmock_main
```
prsunny, line 168: why do we need this change in intfmgrd?

stepanblyschak: @prsunny tests/mock_tests/fake_response_publisher.cpp tests_intfmgrd depends on is using google mock

prsunny (approved):

PR #2515, opened 2022-11-12
[fpmsyncd] Add support for SRv6

cscarpitta:
The `dplane_fpm_nl` module in FRR allows SONiC to learn the forwarding information computed by FRR. FRR already supports many SRv6 functionalities. There is an open PR (FRRouting/frr/pull/12301) to extend the `dplane_fpm_nl` module of FRR to export the required SRv6 information to SONiC.

OrchAgent (specifically, RouteOrch and SRv6Orch) in SONiC SWSS can already extract the SRv6 information from the APP_DB and push it towards the ASIC_DB ([sonic-net/sonic-swss/pull/1964](https://github.com/sonic-net/sonic-swss/pull/1964)).

However, the `fpmsyncd` in SONiC SWSS does not process the SRv6 information received from FRR. This PR extends `fpmsyncd` to parse the SRv6 information received from FRR and push it toward the APP_DB.

**What I did**

I added support for SRv6 to `fpmsyncd`:
* I added a handler `onSrv6RouteMsg()` which is able to parse Netlink messages containing an SRv6 Steering Route and create an entry in the `ROUTE_TABLE` of `APP_DB`.
* I added a handler `onSrv6LocalSidMsg()` which is able to parse Netlink messages containing an SRv6 Local SID and create an entry in the `SRV6_MY_SID_TABLE` of `APP_DB`.
* I changed `onMsgRaw()` to pass the Netlink messages to the appropriate handler depending on the encapsulation type.
* I added the unit tests to verify that `fpmsyncd` processes the Netlink messages and updates the `APP_DB` correctly.

**Why I did it**

FRR supports many SRv6 functionalities. There is an open PR ([FRRouting/frr/pull/12301](https://github.com/FRRouting/frr/pull/12301)) to extend the `dplane_fpm_nl` module of FRR to export the required SRv6 information to SONiC.

However, the `fpmsyncd` in SONiC SWSS did not process the SRv6 information received from FRR.

**How I verified it**

I added new unit tests in `tests/mock_tests/fpmsyncd/receive_srv6_localsids_ut.cpp` and `tests/mock_tests/fpmsyncd/receive_srv6_steer_routes_ut.cpp`.

**Details if related**
SRv6 HLD: https://github.com/sonic-net/SONiC/blob/master/doc/srv6/srv6_hld.md
SRv6 uSID HLD: https://github.com/sonic-net/SONiC/blob/master/doc/srv6/SRv6_uSID.md

### DIFF
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -37,6 +37,7 @@ using namespace swss;
 #define VXLAN_RMAC            1
 #define NH_ENCAP_VXLAN      100
 
+#define NH_ENCAP_SRV6_ROUTE         101
 
 #define IPV4_MAX_BYTE       4
 #define IPV6_MAX_BYTE      16
@@ -44,6 +45,12 @@ using namespace swss;
 #define IPV6_MAX_BITLEN    128
 
 #define ETHER_ADDR_STRLEN (3*ETH_ALEN)
+
+enum {
+    SRV6_ROUTE_UNSPEC            = 0,
+    SRV6_ROUTE_VPN_SID           = 1,
+    SRV6_ROUTE_ENCAP_SRC_ADDR    = 2,
+};
 
 /* Returns name of the protocol passed number represents */
 static string getProtocolString(int proto)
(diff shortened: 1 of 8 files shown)


fpmsyncd/routesync.cpp
```diff
@@ -547,6 +810,606 @@ void RouteSync::onEvpnRouteMsg(struct nlmsghdr *h, int len)
...
+    }
+
+    if (vpn_sid_str.empty())
+    {
+        SWSS_LOG_NOTICE("SRv6 IP Prefix: %s vpn_sid is empty", destipprefix);
+        return;
+    }
+
+    vector<FieldValueTuple> fvVectorRoute;
+    if (!vpn_sid_str.empty())
```
svshah-intel, line 1024: this is not required since this check is already performed few lines above

fpmsyncd/routesync.cpp
```diff
@@ -547,6 +810,606 @@ void RouteSync::onEvpnRouteMsg(struct nlmsghdr *h, int len)
...
+        FieldValueTuple vrf("vrf", vrf_str);
+        fvVector.push_back(vrf);
+    }
+    if (!adj_str.empty())
+    {
+        FieldValueTuple adj("adj", adj_str);
+        fvVector.push_back(adj);
+    }
+
+    m_srv6LocalSidTable.set(my_sid_table_key, fvVector);
```
svshah-intel, line 1335: srv6orch.cpp says
    /* Key for mySid : block_len:node_len:function_len:args_len:sid-ip */

here looks like only sid-ip is used as my_sid_table_key?

fpmsyncd/routesync.cpp
```diff
@@ -563,7 +1426,20 @@ void RouteSync::onMsgRaw(struct nlmsghdr *h)
...
+    switch (getEncapType(h))
+    {
+        case NH_ENCAP_SRV6_ROUTE:
+            onSrv6SteerRouteMsg(h, len);
+            break;
+        case NH_ENCAP_SRV6_LOCAL_SID:
+            onSrv6LocalSidMsg(h, len);
+            break;
+        default:
+            onEvpnRouteMsg(h, len);
```
svshah-intel, line 1439: it may be good to have a comment here to mention why default case handles onEvpnRouteMsg. I understand you are mainting functionality as before switch case, but for future reading it may be confusing without supporting comment

fpmsyncd/routesync.cpp
```diff
@@ -661,6 +1537,13 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
 
     if (nlmsg_type == RTM_DELROUTE)
     {
+        char destipaddress[MAX_ADDR_SIZE + 1] = {0};
+        nl_addr2str(dip, destipaddress, MAX_ADDR_SIZE);
+
+        /* Duplicated delete as we do not know if it is a seg6 route, seg6local
+         * route or regular route */
+        m_srv6LocalSidTable.del(destipaddress);
```
svshah-intel, line 1545: 1) don't you need to check ip type to be v6 first?
2) How about handling warmRestartInProgress check just like is handled for normal route delete?
3) Same as earlier comments - srv6orch.cpp seems to be expecting more than just ip address as a key value

fpmsyncd/routesync.cpp
cscarpitta: I've removed it, thanks!

fpmsyncd/routesync.cpp
cscarpitta: I added a comment, thanks for the suggestion!

fpmsyncd/routesync.cpp
cscarpitta: Exactly! I fixed it.

fpmsyncd/routesync.cpp
cscarpitta: I moved the delete code to `onSrv6LocalSidMsg()` and used the correct key value.

cscarpitta:
@svshah-intel Many thanks for your review! I addressed all your comments. Could you please take another look?

cscarpitta:
This PR has been replaced by https://github.com/sonic-net/sonic-swss/pull/3123. Closing it

PR #2469, opened 2022-09-27
[vlanmgr] Disable `arp_evict_nocarrier` for vlan host intf

lolyu:
**What I did**
To fix: https://github.com/sonic-net/sonic-buildimage/issues/11924

This PR depends on: https://github.com/sonic-net/sonic-linux-kernel/pull/293

Signed-off-by: Longxiang Lyu <lolv@microsoft.com>

**Why I did it**
Disable the sysctl parameter `arp_evict_nocarrier` for the host vlan interfaces.
So if its member ports are all down and host vlan interfaces become `NO-CARRIER`, the neighbor learned will not be flushed.

**How I verified it**
Build image together with the kernel change:
1. verify the sysctl param `arp_evict_no_carrier` for `Vlan1000`
```
# cat /proc/sys/net/ipv4/conf/Vlan1000/arp_evict_nocarrier
0
```
2. check the neighbor learned from `Vlan1000`
```
# ip neighbor | grep Vlan1000
192.168.0.13 dev Vlan1000 lladdr 82:7b:a1:ac:3d:0c STALE
192.168.0.15 dev Vlan1000 lladdr ea:5e:4b:f2:84:0e STALE
192.168.0.9 dev Vlan1000 lladdr 72:ce:37:c0:39:08 STALE
192.168.0.24 dev Vlan1000 lladdr ea:47:b9:01:4c:17 STALE
192.168.0.4 dev Vlan1000 lladdr 02:e1:1a:90:eb:03 STALE
192.168.0.11 dev Vlan1000 lladdr 56:a0:28:b5:b7:0a STALE
192.168.0.6 dev Vlan1000 lladdr 52:52:cb:52:7b:05 STALE
192.168.0.21 dev Vlan1000 lladdr 8e:a7:11:a3:1c:14 STALE
192.168.0.23 dev Vlan1000 lladdr 8e:83:cc:23:3b:16 STALE
192.168.0.2 dev Vlan1000 lladdr e2:48:09:46:76:01 STALE
192.168.0.17 dev Vlan1000 lladdr 16:0a:b6:0e:4d:10 STALE
192.168.0.12 dev Vlan1000 lladdr be:26:ff:0f:7e:0b STALE
192.168.0.19 dev Vlan1000 lladdr da:1b:c4:ca:10:12 STALE
192.168.0.14 dev Vlan1000 lladdr 66:e0:b9:f1:09:0d STALE
192.168.0.8 dev Vlan1000 lladdr c2:4d:53:03:bc:07 STALE
[...]

PR #2517, opened 2022-11-14
Revert "[vlanmgr] Disable `arp_evict_nocarrier` for vlan host intf"

liuh-80:
Need revert PR #2469 because new UT 'test_VlanMemberLinkDown' does not work on 202205 or older version.
And currently some 202205 branch P0 PR https://github.com/sonic-net/sonic-swss-common/pull/706 been block by this UT.


liuh-80:
The linux-kernel patch 0001-net-arp-introduce-arp_evict_nocarrier-sysctl-paramet.patch does not add to 202205 branch. I think that's why the UT failed.

liuh-80:
Offline sync with longxiang, he will create a revert on 202205

PR #2520, opened 2022-11-14
orchagent/portsorch: Keeping sched_group_ids in port

arvbb:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/qosorch.cpp
+++ b/orchagent/qosorch.cpp
@@ -1459,44 +1459,14 @@ sai_object_id_t QosOrch::getSchedulerGroup(const Port &port, const sai_object_id
     const auto it = m_scheduler_group_port_info.find(port.m_port_id);
     if (it == m_scheduler_group_port_info.end())
     {
-        /* Get max sched groups count */
-        attr.id = SAI_PORT_ATTR_QOS_NUMBER_OF_SCHEDULER_GROUPS;
-        sai_status = sai_port_api->get_port_attribute(port.m_port_id, 1, &attr);
-        if (SAI_STATUS_SUCCESS != sai_status)
-        {
-            SWSS_LOG_ERROR("Failed to get number of scheduler groups for port:%s", port.m_alias.c_str());
-            task_process_status handle_status = handleSaiGetStatus(SAI_API_PORT, sai_status);
-            if (handle_status != task_process_status::task_success)
-            {
-                return SAI_NULL_OBJECT_ID;
-            }
-        }
-
-        /* Get total groups list on the port */
-        uint32_t groups_count = attr.value.u32;
-        std::vector<sai_object_id_t> groups(groups_count);
-
-        attr.id = SAI_PORT_ATTR_QOS_SCHEDULER_GROUP_LIST;
-        attr.value.objlist.list = groups.data();
-        attr.value.objlist.count = groups_count;
-        sai_status = sai_port_api->get_port_attribute(port.m_port_id, 1, &attr);
-        if (SAI_STATUS_SUCCESS != sai_status)
-        {
-            SWSS_LOG_ERROR("Failed to get scheduler group list for port:%s", port.m_alias.c_str());
-            task_process_status handle_status = handleSaiGetStatus(SAI_API_PORT, sai_status);
-            if (handle_status != task_process_status::task_success)
-            {
-                return SAI_NULL_OBJECT_ID;
-            }
-        }
-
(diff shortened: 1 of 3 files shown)


neethajohn (approved):

orchagent/qosorch.cpp
```diff
@@ -1459,44 +1459,14 @@ sai_object_id_t QosOrch::getSchedulerGroup(const Port &port, const sai_object_id
...
-            if (handle_status != task_process_status::task_success)
-            {
-                return SAI_NULL_OBJECT_ID;
-            }
-        }
-
+        size_t groups_count = port.m_scheduler_group_ids.size();
         m_scheduler_group_port_info[port.m_port_id] = {
-            .groups = std::move(groups),
+            .groups = std::move(port.m_scheduler_group_ids),
```
stephenxs, line 1464: `move` generates a right value reference. Typically, it is used only if the object designated by the input argument is about to expire its lifespan or if we know for sure it will not be referenced afterward.
Here, a move constructor will be used to instantiate the object `m_scheduler_group_port_info`, which effectively passes the `port.m_scheduler_group_ids` memory to `m_scheduler_group_port_info` but `port.m_scheduler_group_ids` is still not expired, which means both objects point to the same area of memory.
Once the port is removed (eg by DPB), the memory is freed but is still referenced by `m_scheduler_group_port_info`, which can lead to memory errors.
I would like to suggest using `port.m_scheduler_group_ids` directly.
I also noticed `m_scheduler_group_port_info` doesn't handle port removal in DPB. By using `port.m_scheduler_group_ids` it can also be resolved.

stephenxs (changes requested):
As comments.
Also suggest adding a mock test to simulate port removal and make sure no error.

orchagent/qosorch.cpp
```diff
@@ -1459,44 +1459,14 @@ sai_object_id_t QosOrch::getSchedulerGroup(const Port &port, const sai_object_id
...
-            if (handle_status != task_process_status::task_success)
-            {
-                return SAI_NULL_OBJECT_ID;
-            }
-        }
-
+        size_t groups_count = port.m_scheduler_group_ids.size();
         m_scheduler_group_port_info[port.m_port_id] = {
-            .groups = std::move(groups),
+            .groups = port.m_scheduler_group_ids,
```
stephenxs, line 1464: This avoids the memory issue but there is still a memory leak issue in DPB. eg. if the old port is removed then a set of new ports are created, the SAI OID of ports will be changed but `m_scheduler_group_port_info[<SAI OID of old port>]` won't be removed. Eventually, `m_scheduler_group_port_info` will contain a lot of info of ports that have been removed.
I would like to suggest
1. moving the whole m_scheduler_group_port_info to `Port` struct and initialize it during port initialization or in this function.
2. or keeping in `m_scheduler_group_port_info` and removing items from it on port removing.
@neethajohn  how do you think

neethajohn: I think the second solution will be a lot cleaner

arvbb: Hi Stephen/Neetha, I am not familiar with DPB. How is the port removal scenario handled currently for scheduler group? Is it already getting covered in existing code before this PR or it is new functionality that needs to be added? Please clarify, thanks.

orchagent/qosorch.cpp
arvbb: Fixed std::move

orchagent/qosorch.cpp
neethajohn: @stephenxs, can you comment on the DPB scenario?

stephenxs: When it comes to DPB, I think what we care about is the port removal flow.
In short, the QoS orchagent should listen to the `config_db.PORT` table and handle the port removal event.
On receiving a port removal event, it should map the port name to OID and then erase the item indexed by the OID from `m_scheduler_group_port_info`.
I think we can use the function `bool PortsOrch::getPort(string alias, Port &p)` to map the port name to OID.
But what I'm not sure about is whether the mapping is still available in ports orch in the port removal scenario because there is no guarantee in terms of the order in which the orchagents (ports ~, QoS ~) handle the port removal notification.
this is why I proposed option `#1`. but if we want to use `#2` maybe we also have a way. I will check with the DPB expert in my team (if there is one) and get back to you.
thanks.

PR #2523, opened 2022-11-15
[202012] [ACL] Specify action list for `PFCWD` and `DROP` ACL table

bingwang-ms:
**What I did**
We found that without specifying `action_list`, the `PFCWD` and `DROP` ACL tables takes more TCAM resources than expected.
Actually, only `SAI_ACL_ACTION_TYPE_PACKET_ACTION` is required for `PFCWD` and `DROP` table.
This PR is to specify the `action_list` for `PFCWD` and `DROP` ACL table. 

**Why I did it**
This PR is to specify the `action_list` for `PFCWD` and `DROP` ACL table. 

**How I verified it**
Verified by vstest and physical testbed.

**Details if related**


neethajohn (approved):

prsunny (approved):

prsunny:
Can you please provide a link to the master PR?

bingwang-ms:
> Can you please provide a link to the master PR?

We don't need it for master branch, because the feature is already included in master branch. 
https://github.com/sonic-net/sonic-swss/blob/28aa309ed04a5b442487c0404db23229cec9b4c9/orchagent/aclorch.cpp#L253-L269

PR #2529, opened 2022-11-16
[Centec]for support mclag of centec to configure port isolate-group

youshcentec:
**What I did**
Add centec platform to supported platform list for port isolation group

**Why I did it**
Centec platform support port isolation group for mclag, but it was not in the supported platforms list

**How I verified it**
Build image, run regression

**Details if related**


youshcentec:
@prsunny please help to merge this pr for supporting mlag on centec asic

neethajohn (approved):

PR #2531, opened 2022-11-17
[POC] routeorch change to use share memory based state tables. 

liuh-80:
Still POC, please ignore this PR.


liuh-80:
Close PR because now using ZMQ.

PR #2539, opened 2022-11-21
Handle race condition to update FDB entries on one-arm MCLAG interface down.

prasanna-cls:
**What I did**
Handled the below two cases.
a) When FDB entry is being removed, and SAI returns SAI_STATUS_ITEM_NOT_FOUND, software cache table have the entry, then the SAI error is ignored.
b) In case of MCLAG advertised mac, and FDB entry is modified to have a new destination port, and SAI returns SAI_STATUS_ITEM_NOT_FOUND, then the FDB entry is created in hardware instead of modification.

**Why I did it**
In MCLAG scenario, when one arm of the MCLAG is becoming DOWN, the SWSS orchagent detects it as oper down and handles it by bulk flushing all the FDB entries on that MCLAG interface. This means , orchagent invokes the SAI bulk flush routine (to delete by port/trunk) and when the SAI notifies back , it clears the software cache entries. At the same time MCLAG module deletes/modifies such FDB entries in one-by-one fashion. Hence its is possible that the bulk flush call had cleared the SAI/hardware and yet software entry is still not cleared. This commit gracefully handles this condition where the FDB entry is not available in hardware but present in orchagent software database. 

**How I verified it**
Without this fix, the SAI error SAI_STATUS_ITEM_NOT_FOUND causes swss reboot but with this fix the FDB entries are programmed properly. Hardware and software MAC tables are in sync, in the MCLAG interface down scenario

**Details if related**
Dependent pull request , to return SAI_STATUS_ITEM_NOT_FOUND when FDB entry is not in SAI/hardware.
https://github.com/sonic-net/sonic-sairedis/pull/1154 

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205

Handling the SAI failures to modify/remove FDB entries when MCLAG interface goes down operationally.

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -1418,6 +1418,17 @@ bool FdbOrch::addFdbEntry(const FdbEntry& entry, const string& port_name,
             {
                 SWSS_LOG_ERROR("macUpdate-Failed for attr.id=0x%x for FDB %s in %s on %s, rv:%d",
                             itr.id, entry.mac.to_string().c_str(), vlan.m_alias.c_str(), port_name.c_str(), status);
+                if ((SAI_STATUS_ITEM_NOT_FOUND  == status) && (fdbData.origin == FDB_ORIGIN_MCLAG_ADVERTIZED)) {
+                    status = sai_fdb_api->create_fdb_entry(&fdb_entry, (uint32_t)attrs.size(), attrs.data());
+                    if (SAI_STATUS_SUCCESS == status) {
+                        SWSS_LOG_INFO("MAC-Update re-create success for FDB %s in %s on %s",
+                             entry.mac.to_string().c_str(), vlan.m_alias.c_str(), port_name.c_str());
+                        break;
+                    } else {
+                        SWSS_LOG_ERROR("MAC-Update re-create failed for FDB %s in %s on %s",
+                           entry.mac.to_string().c_str(), vlan.m_alias.c_str(), port_name.c_str());
+                    }
+                }
                 task_process_status handle_status = handleSaiSetStatus(SAI_API_FDB, status);
                 if (handle_status != task_success)
                 {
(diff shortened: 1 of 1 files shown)


prasanna-cls:
request to retest please.

orchagent/fdborch.cpp
```diff
@@ -1418,6 +1418,17 @@ bool FdbOrch::addFdbEntry(const FdbEntry& entry, const string& port_name,
             {
                 SWSS_LOG_ERROR("macUpdate-Failed for attr.id=0x%x for FDB %s in %s on %s, rv:%d",
                             itr.id, entry.mac.to_string().c_str(), vlan.m_alias.c_str(), port_name.c_str(), status);
+                if ((SAI_STATUS_ITEM_NOT_FOUND  == status) && (fdbData.origin == FDB_ORIGIN_MCLAG_ADVERTIZED)) {
```
gechiang, line 1421: Can you add comments here as you have explained in the PR for the Add case here as well?  Reason is by the time this PR is merged, reader will have a hard time to understand why this code was like this.
when I review this part, I had to go back to your PR description again in order to understand why...
Please add appropriate comment.

orchagent/fdborch.cpp
```diff
@@ -1601,7 +1612,7 @@ bool FdbOrch::removeFdbEntry(const FdbEntry& entry, FdbOrigin origin)
     fdb_entry.bv_id = entry.bv_id;
 
     status = sai_fdb_api->remove_fdb_entry(&fdb_entry);
-    if (status != SAI_STATUS_SUCCESS)
+    if ((status != SAI_STATUS_SUCCESS) && (status != SAI_STATUS_ITEM_NOT_FOUND))
```
gechiang, line 1615: Can you also add a comment here as why this could be a case that we should allow to move on and not treated as an error case.

gechiang:
@Praveen-Brcm, Can you please help review this as well? Thanks!

orchagent/fdborch.cpp
prasanna-cls: Thanks @gechiang . I have added comments now.

orchagent/fdborch.cpp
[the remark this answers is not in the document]
prasanna-cls: thanks @prsunny , addressed this comment.

prasanna-cls: Thanks @gechiang . I have added comments now.

prasanna-cls:
> Can you try add test to have code coverage?

Hi @prsunny , 
This code will be exexcuted only at certain race conditions. i.e. execution of events in below order.
a) Bulk Flush of  FDBs is initiated by Orchagent, SAI has removed the entries in hardware
b) MCLAG modifies the destination port of the MACs from MCLAG interface to ICCP link interface.
c) SAI notifies Orchagent about the FDB deletion and , Orchagent deletes its internal cache table

I tried simulating this using the pytest infra, i am not able to achieve this.
in the pytest environment, event (a) and (c) always happen together i.e. they are always atomic. Hence i am unable to include the coverage test case for this scenario.
I tried introducing even a new thread, to trigger events (c) and (b) separately, but of no help.

PR #2541, opened 2022-11-22
Fix `test_vlan.py`

lolyu:
**What I did**
The signature of `wait_for_n_keys` is:
```
def wait_for_n_keys(
    self,
    table_name,
    num_keys,
    wait_at_least_n_keys,
    polling_config = PollingConfig(),
    failure_message = None,
):
```
In the following methods:
`DVSVlan.get_and_verify_vlan_ids`
`DVSVlan.get_and_verify_vlan_hostif_ids`

They pass `polling_config` as the third argument, stored in the parameter `wait_at_least_n_keys` and always evaluated to `True`. This will cause trouble in the vlan or vlan interface removal, `wait_for_n_keys` will always return early as there are always more than expected keys existing in the table. So when next testcase starts, those keys might still exist in the table and fail the next testcase.

**Why I did it**
Stabilize `test_vlan.py`

**How I verified it**

**Details if related**

### DIFF
--- a/tests/dvslib/dvs_vlan.py
+++ b/tests/dvslib/dvs_vlan.py
@@ -72,7 +72,7 @@ class DVSVlan(object):
                                 polling_config=PollingConfig()):
         vlan_entries = self.asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_VLAN",
                                                     expected_num + 1,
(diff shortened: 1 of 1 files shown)


yxieca:
@lolyu can you look into the vstest failure?

theasianpianist (approved):

PR #2547, opened 2022-11-24
[mirrororch] ERSPAN supports LAG as a destination port.

inspurSDN:
[portsorch] Add a function to return a linked-up member of an input LAG.

- What I did Support LAG as a destination port in a erspan session. In this case, detect the LAG member that is linked down and switch the destination port to a LAG member that is linked up.

- Why I did it Current ERSPAN does not support LAG as a destination port in a mirror session. The mirrororch is unaware of the LAG member port link up or down.

- How to verify it The community tests and manual tests pass.

Signed-off-by: sdn@inspur.com


prsunny:
@inspurSDN , do you've an HLD to link here?

PR #2548, opened 2022-11-24
Modify coppmgr mergeConfig to support preserving copp tables through reboot.

arfeigin:
This PR should be merged together with https://github.com/sonic-net/sonic-utilities/pull/2524 and is required to 202205 and 202211.
This PR implements [fastboot] Preserve CoPP table HLD to improve fastboot flow (https://github.com/sonic-net/SONiC/pull/1107).

**What I did**
Modified coppmgr mergeConfig logic to support preserving copp tables contents through reboot. Handle duplicates, overwrites, unsupported keys preserved and new keys. According to https://github.com/sonic-net/SONiC/pull/1107

**Why I did it**
To shorten dataplane downtime on fast-reboot.

**How I verified it**
Manual tests (community fast-reboot test)

**Details if related**

### DIFF
--- a/cfgmgr/coppmgr.cpp
+++ b/cfgmgr/coppmgr.cpp
@@ -280,16 +280,13 @@ bool CoppMgr::isDupEntry(const std::string &key, std::vector<FieldValueTuple> &f
         {
             string field = fvField(fv);
             string value = fvValue(fv);
-            unordered_map<string, string>::const_iterator preserved_copp_it = preserved_copp_entry.find(field);
+            auto preserved_copp_it = preserved_copp_entry.find(field);
             bool field_found = (preserved_copp_it != preserved_copp_entry.end());
-            if (field_found)
+            if ((!field_found) || (field_found && preserved_copp_it->second.compare(value)))
             {
-                if (preserved_copp_it->second.compare(value))
-                {
-                    // overwrite -> delete preserved entry from copp table and set a new entry instead
-                    m_coppTable.del(key);
-                    return false;
-                }
+                // overwrite -> delete preserved entry from copp table and set a new entry instead
+                m_coppTable.del(key);
+                return false;
             }
         }
     }
@@ -315,6 +312,7 @@ CoppMgr::CoppMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
 
     std::vector<string> group_cfg_keys;
     std::vector<string> trap_cfg_keys;
+    unordered_set<string> supported_copp_keys;
 
     CoppCfg group_cfg;
     CoppCfg trap_cfg;
@@ -395,6 +393,7 @@ CoppMgr::CoppMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
 
         if (!trap_group_fvs.empty())
         {
+            supported_copp_keys.emplace(i.first);
             if (isDupEntry(i.first, trap_group_fvs))
             {
                 continue;
@@ -411,15 +410,6 @@ CoppMgr::CoppMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
     }
 
     // Delete unsupported keys from preserved copp tables
-    unordered_set<string> supported_copp_keys;
-    for (auto i : m_coppGroupInitCfg)
-    {
-        supported_copp_keys.emplace(i.first);
-    }
-    for (auto i : group_cfg_keys)
-    {
-        supported_copp_keys.emplace(i);
-    }
     for (auto it : preserved_copp_keys)
     {
         auto copp_it = supported_copp_keys.find(it);
--- a/cfgmgr/coppmgr.cpp
+++ b/cfgmgr/coppmgr.cpp
@@ -262,10 +262,6 @@ void CoppMgr::mergeConfig(CoppCfg &init_cfg, CoppCfg &m_cfg, std::vector<std::st
      * user config) will overwrie the copp table entries, current entry will be deleted from copp
      * tables.
      */
-    unordered_map<string, string> preserved_copp_entry;
-    unordered_map<string, string>::const_iterator preserved_copp_it;
-    string field;
-    string value;
     for (auto i : helper)
     {
         string table_key = i.first;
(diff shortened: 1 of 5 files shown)


stepanblyschak (approved):

cfgmgr/coppmgr.cpp
```diff
@@ -250,7 +253,65 @@ void CoppMgr::mergeConfig(CoppCfg &init_cfg, CoppCfg &m_cfg, std::vector<std::st
         {
             std::vector<FieldValueTuple> cfg_fvs;
             cfgTable.get(i, cfg_fvs);
-            m_cfg[i] = cfg_fvs;
+            helper[i] = cfg_fvs;
```
dgsudharsan, line 256: mergeConfig is used to merge both copp_trap and copp_group cfg. Its a generic function. I don't recommend to have the helper config present inside this API. The helper needs to be done only for copp group and not for copp trap.

cfgmgr/coppmgr.cpp
```diff
@@ -250,7 +253,65 @@ void CoppMgr::mergeConfig(CoppCfg &init_cfg, CoppCfg &m_cfg, std::vector<std::st
...
+                    if (!preserved_copp_it->second.compare(value))
+                    {
+                        // duplicate -> ignore and continue to next key.
+                        continue;
+                    }
+                    else
+                    {
+                        // overwrite -> delete preserved FV tuple from copp table and add helper fvs to m_cfg
+                        overwrite = true;
+                        m_coppTable.hdel(table_key, field);
```
dgsudharsan, line 303: I don't think there is logic in copporch to process hdel. It should be del and recreating the key

dgsudharsan (changes requested):

cfgmgr/coppmgr.cpp
arfeigin: I manually checked it, this worked well in the following scenario:

1. redis-cli -n 0 hgetall "COPP_TABLE:queue4_group2" (to see the current values)
2. redis-cli -n 0 hdel "COPP_TABLE:queue4_group2" "cbs"
3. redis-cli -n 0 hset "COPP_TABLE:queue4_group2" "cbs" "700"
4. perform fast-reboot
5. copp tables are preserved through db migration
6. redis-cli -n 0 hgetall "COPP_TABLE:queue4_group2" (as soon as platform is up to check that cbs still has 700 as value)
7. redis-cli -n 0 hgetall "COPP_TABLE:queue4_group2" (sometime after, 40~ seconds to check that cbs value was changed to 600 according to the default initialization file /etc/sonic/copp_cfg.json)

cfgmgr/coppmgr.cpp
arfeigin: Moved to a dedicated function to merge with preserved copp tables.

cfgmgr/coppmgr.cpp
```diff
@@ -255,6 +257,45 @@ void CoppMgr::mergeConfig(CoppCfg &init_cfg, CoppCfg &m_cfg, std::vector<std::st
...
+        {
+            preserved_copp_entry[fvField(prev_fv)] = fvValue(prev_fv);
+        }
+        for (auto fv: fvs)
+        {
+            string field = fvField(fv);
+            string value = fvValue(fv);
+            unordered_map<string, string>::const_iterator preserved_copp_it = preserved_copp_entry.find(field);
+            bool field_found = (preserved_copp_it != preserved_copp_entry.end());
+            if (field_found)
```
dgsudharsan, line 285: If field is not found, it means some new field is present in new fvs which must be handled right? In that case too shouldn't we delete and return false?

dgsudharsan (changes requested):

cfgmgr/coppmgr.cpp
```diff
@@ -255,6 +257,45 @@ void CoppMgr::mergeConfig(CoppCfg &init_cfg, CoppCfg &m_cfg, std::vector<std::st
...
+        unordered_map<string, string> preserved_copp_entry;
+        for (auto prev_fv : preserved_fvs)
+        {
+            preserved_copp_entry[fvField(prev_fv)] = fvValue(prev_fv);
+        }
+        for (auto fv: fvs)
+        {
+            string field = fvField(fv);
+            string value = fvValue(fv);
+            unordered_map<string, string>::const_iterator preserved_copp_it = preserved_copp_entry.find(field);
```
dgsudharsan, line 283: You can use auto instead of   unordered_map<string, string>::const_iterator

dgsudharsan (changes requested):

cfgmgr/coppmgr.cpp
```diff
@@ -352,15 +395,39 @@ CoppMgr::CoppMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
...
         {
             g_copp_init_set.insert(i.first);
         }
     }
+
+    // Delete unsupported keys from preserved copp tables
+    unordered_set<string> supported_copp_keys;
+    for (auto i : m_coppGroupInitCfg)
+    {
+        supported_copp_keys.emplace(i.first);
```
dgsudharsan, line 417: Can't the supported_copp_keys be filled inside the if loop in 397 when setting the app_db table?

dgsudharsan (changes requested):

cfgmgr/coppmgr.cpp
arfeigin: Correct, fixed the condition.

cfgmgr/coppmgr.cpp
arfeigin: Done.

dgsudharsan (approved):

dgsudharsan (approved):

cfgmgr/coppmgr.cpp
```diff
@@ -255,6 +257,42 @@ void CoppMgr::mergeConfig(CoppCfg &init_cfg, CoppCfg &m_cfg, std::vector<std::st
...
+            auto preserved_copp_it = preserved_copp_entry.find(field);
+            bool field_found = (preserved_copp_it != preserved_copp_entry.end());
+            if ((!field_found) || (field_found && preserved_copp_it->second.compare(value)))
+            {
+                // overwrite -> delete preserved entry from copp table and set a new entry instead
+                m_coppTable.del(key);
+                return false;
+            }
+        }
+    }
```
svshah-intel, line 292: seems this code covers case where new copp entry has a field that preserved does not. isn't it missing a case where preserved copp entry had a field that new copp entry does not carry anymore?

arfeigin: Good catch, I missed it.
I guess this can be done easily by comparing the sizes of fvs and preserved_fvs. I will test and raise a new PR with this fix.

Comparing sizes will be sufficient since if sizes differ it says that there is at least one difference so it is not a duplicate and the preserved entry should be deleted and new entry set.
In case the size of both is equal but some of the fields differs it will be caught then iterating over fvs.

svshah-intel: yeah, size check should work.

PR #2550, opened 2022-11-29
[ACL] Support ACTION_COUNTER action in custom ACL table type

bingwang-ms:
**What I did**
This PR is to update aclorch to support `SAI_ACL_ENTRY_ATTR_ACTION_COUNTER` action in custom ACL table type.
Before this change, ACL counter is not supported in custom ACL type. 

**Why I did it**
Because we need the custom ACL type with counter supported.

**How I verified it**
Verified by building a new image, and check the counter.

**Details if related**


bingwang-ms:
@stepanblyschak Can you please help review?

stepanblyschak (approved):

prsunny (approved):

PR #2551, opened 2022-11-29
[fpmsyncd] Implement pending route suppression feature

stepanblyschak:
**DEPENDS**: https://github.com/sonic-net/sonic-swss/pull/2512

**What I did**

I implemented support to enable pending routes suppression feature. When this feature is enabled, fpmsyncd will wait for reply from orchagent before sending offload status message to zebra.

**Why I did it**

This is done to not announce routes which aren't yet offloaded in HW.

**How I verified it**

UT and manual tests.

**Details if related**


fpmsyncd/routesync.cpp
```diff
@@ -788,6 +831,8 @@ void RouteSync::onLabelRouteMsg(int nlmsg_type, struct nl_object *obj)
     struct nl_addr *daddr;
     char destaddr[MAX_ADDR_SIZE + 1] = {0};
 
+    sendOffloadReply(route_obj);
```
venkatmahalingam, line 834: Hope you have a test case to test this MPLS label route.

fpmsyncd/routesync.cpp
```diff
@@ -1222,3 +1269,183 @@ string RouteSync::getNextHopWt(struct rtnl_route *route_obj)
...
+        if (field == "err_str")
+        {
+            isSuccessReply = (value == "SWSS_RC_SUCCESS");
+        }
+        else if (field == "protocol")
+        {
+            // If field "protocol" is present in the field values then
+            // it is a SET operation. This field is absent only if we are
+            // processing DEL operation.
+            isSetOperation = true;
```
venkatmahalingam, line 1362: Dont we get Del operation in the event? why should we assume this?

fpmsyncd/routesync.cpp
```diff
@@ -1222,3 +1269,183 @@ string RouteSync::getNextHopWt(struct rtnl_route *route_obj)
...
+
+    if (isSuccessReply)
+    {
+        flags |= RTM_F_OFFLOAD;
+    }
+
+    // Mark route as OFFLOAD
+    rtnl_route_set_flags(routeObject.get(), flags);
+
+    if (!sendOffloadReply(routeObject.get()))
```
venkatmahalingam, line 1414: When there are multiple route updates for the given route, how do we make sure that the response to sent for the right route-update?

stepanblyschak: The constructed RTM_NEWROUTE message contains prefix + vrf + protocol of the route

fpmsyncd/routesync.cpp
stepanblyschak: If we get an response for a delete operation we don't get original FVs. We use this knowledge to filter out DEL events we are not interested in.

ResponsePublisher for reference: https://github.com/sonic-net/sonic-swss/blob/master/orchagent/response_publisher.cpp#L106

fpmsyncd/routesync.cpp
stepanblyschak: I don't add additional flows for MPLS route handling. So it should be covered by existing MPLS route tests.

prsunny:
@stepanblyschak , can you please test for backward compatibility?

stepanblyschak:
@prsunny It can be merged without https://github.com/sonic-net/sonic-buildimage/pull/12853. BGP facts and FIB test passed on T1-64-LAG with this change.

Please note, that I had to mark VS tests with ```xfail``` because they require VS changes done in https://github.com/sonic-net/sonic-buildimage/pull/12853. Otherwise, it is not possible to merge one without the other in order to not break tests or image.
Later this mark has to be removed once sonic-buildimage PR is merged.

fpmsyncd/fpmlink.cpp
```diff
@@ -277,3 +283,36 @@ void FpmLink::processFpmMessage(fpm_msg_hdr_t* hdr)
...
+
+bool FpmLink::send(nlmsghdr* nl_hdr)
+{
+    fpm_msg_hdr_t hdr{};
+
+    size_t len = fpm_msg_align(sizeof(hdr) + nl_hdr->nlmsg_len);
+
+    if (len > m_bufSize)
+    {
+        SWSS_LOG_THROW("Message length %zu is greater then the send buffer size %d", len, m_bufSize);
```
StormLiangMS, line 295: wording issue, then -> than

StormLiangMS:
@jcaiMR could you help to review?

fpmsyncd/fpmsyncd.cpp
```diff
@@ -47,21 +51,47 @@ static bool eoiuFlagsSet(Table &bgpStateTable)
...
     DBConnector stateDb("STATE_DB", 0);
     Table bgpStateTable(&stateDb, STATE_BGP_TABLE_NAME);
 
+    NetLink netlink;
+
+    netlink.registerGroup(RTNLGRP_LINK);
+
     NetDispatcher::getInstance().registerMessageHandler(RTM_NEWROUTE, &sync);
     NetDispatcher::getInstance().registerMessageHandler(RTM_DELROUTE, &sync);
+    NetDispatcher::getInstance().registerMessageHandler(RTM_NEWLINK, &sync);
```
StormLiangMS, line 76: @stepanblyschak what's the usage for RTM_NETLINK?

fpmsyncd/fpmsyncd.cpp
```diff
@@ -182,6 +216,67 @@ int main(int argc, char **argv)
...
+
+                            if (shouldEnable && !sync.isSuppressionEnabled())
+                            {
+                                routeResponseChannel = std::make_unique<NotificationConsumer>(&applStateDb, routeResponseChannelName);
+                                sync.setSuppressionEnabled(true);
+                                s.addSelectable(routeResponseChannel.get());
+                            }
+                            else if (!shouldEnable && sync.isSuppressionEnabled())
+                            {
+                                sync.setSuppressionEnabled(false);
```
StormLiangMS, line 260: @stepanblyschak when do the disable, is that possible some in flight messages are lost?

fpmsyncd/fpmlink.cpp
stepanblyschak: Thanks, fixed

fpmsyncd/fpmsyncd.cpp
stepanblyschak: Once disable fpmsyncd will reply imidiatelly on receiving route from FRR, no messages are lost

StormLiangMS: @stepanblyschak 
Assume the route A sent to the orchagent to offload, fpmsyncd was waiting for the response. At this time, the route A is in hold status, if we disable the suppress, fpmsyncd won't care the msg sent back to announce route A is offloaded to HW successfully, right? In that case, the route A is offloaded to HW, but the status in FRR is suppressed, right?

stepanblyschak: @StormLiangMS Oh, I get it. Right, we will have suppressed route in FRR in case user disables prior to processing update from orchagent. That can be one or many routes.
We could listen to response channel regardless of suppress status and send back offload flag again but that would generate twice as much traffic to zebra in case feature is disabled.

How about in case of transition from enabled -> disabled fetch all routes present in APPL_DB (it will include those routes that are not set offloaded yet) and send offload status?

StormLiangMS: @stepanblyschak 
The other way we may have a snapshot for the routes in hold in FRR when user disable it, and keep listen to response channel and only send back offload flag which matched with this snapshot list?

In order to avoid the race condition for when we take the snapshot, we may allow a little bit overlap time, during that time frame, send twice traffic to Zebra.

StormLiangMS: > @stepanblyschak The other way we may have a snapshot for the routes in hold in FRR when user disable it, and keep listen to response channel and only send back offload flag which matched with this snapshot list?
> 
> In order to avoid the race condition for when we take the snapshot, we may allow a little bit overlap time, during that time frame, send twice traffic to Zebra.

I rethink it, maybe to make it simple, your idea is good enough, our purpose here it to make sure when suppress enabled, to make sure HW offload happened before advertised out, but if this is disabled, just make sure everything is advertised out would be ok, no need to make sure offload happened before advertise at this time.

prsunny:
@StormLiangMS , are we good to merge this? Can you please sign-off if no further comments?

fpmsyncd/fpmsyncd.cpp
StormLiangMS: @stepanblyschak should we revert the sequence, first, setSuppressionEnabled to disable the offload process, then do the markRoutesOffloaded? Otherwise it still timing that some inflight routes after the markRoutesOffloaded.

StormLiangMS (approved):
LGTM

StormLiangMS:
@prsunny this looks good to me.

StormLiangMS:
@stepanblyschak could you also have the testcase in sonic-mgmt to cover this FIB suppress feature, and pls involve Jcai and me for the code review.

stepanblyschak:
@StormLiangMS @jcaiMR BGP suppress FIB pending test cases are already available in PRs - test plan + implementation:
- https://github.com/sonic-net/sonic-mgmt/pull/7475
- https://github.com/sonic-net/sonic-mgmt/pull/7430

PR #2553, opened 2022-11-29
[routesync] Fix for stale dynamic neighbor

vganesan-nokia:
**What I did**

Changes are done to delete those route entries from APPL_DB which have next hop on eth0 or docker0 interface.

**Why I did it**

This is done to fix the issue https://github.com/sonic-net/sonic-buildimage/issues/12442. 
For more information about root cause and the solution, please see details below

**How I verified it**

Tested using the following scenario 

- Establish eBGP sessions and advertise default route (IPv4 or IPv6).
- Make sure that default routes exists in kernel, APPL_DB and ASIC_DB as expected.
 -Shutdown interfaces such that all the eBGP neighbors that advertised the default route are down.
- Dump the routes in the kernel and make sure that the defautl route has only one next on the "eth0"
- Clear the dynamically learned neighbors using "sonic-clearp ndp" or "sonic-clear arp" command.
- Dump the neighbors in the linux kernel using "ip neigh show" command
- Dump neighbors from ASIC_DB
- All the dynamically learned neighbors are cleared from linux kernel, APPL_DB and ASIC_DB when the neighbors are flushed after shutting down the eBGP neighbor interfaces.

**Details if related**

The root cause of this problem is that some routes learned from eBGP neighbors are not cleared from APPL_DB when the eBGP neighbors go down. The routes which show this issue are those routes which include the next hop on interface "eth0". This consistently occurs for default routes for both IPv4 and IPv6 in any asic instance. Since the route is not deleted from APPL DB (though route is deleted from kernel), the neighbor ref_count is not 0 and hence the neighbor is not deleted from the ASIC_DB.

All asic instances include the default route with next hop on eth0 (with large metric to avoid colliding with default routes learned via eBGP). The APPL_DB has this route with only the eBGP next hops due to a filtering in routesync that filters the kernel route updates that has a next hop on eth0 or docker0. Consequently, the next hop on eth0 is not included in the route entry in APPL_DB and ASIC_DB. When all the eBGP neighbor interfaces are shut down, all eBGP neighbors go down. When all eBGP neighbors go down, the default route is withdrawn. The next hop on eth0 becomes the only next hop for the default route. When kernel sends this update, because of the the above mentioned routesync filter the route update is not sent to APPL_DB. Hence the APPL_DB is left with a stale route entry with whatever next hops it had before the eth0 next hop became only next hop. Since these next hops still have this route referenced, these are not cleared when neighbors are flushed.

The soultion to both of these issues viz., (1) stale route entry in APPL_DB and ASIC_DB and (2) stale neighbor entry in ASIC_DB is to delete the routes from APPL_DB when kernel updates are received with next hop on "eth0" as the only next hop.

### DIFF
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -740,9 +740,24 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
             // path will be left with stale route entry
             if(alsv.size() == 1)
             {
-                SWSS_LOG_NOTICE("RouteTable del msg for route with only one nh on eth0/docker0: %s %s %s %s",
-                        destipprefix, gw_list.c_str(), intf_list.c_str(), mpls_list.c_str());
-                m_routeTable.del(destipprefix);
+                if (!warmRestartInProgress)
+                {
+                    SWSS_LOG_NOTICE("RouteTable del msg for route with only one nh on eth0/docker0: %s %s %s %s",
+                            destipprefix, gw_list.c_str(), intf_list.c_str(), mpls_list.c_str());
+
+                    m_routeTable.del(destipprefix);
+                }
+                else
+                {
+                    SWSS_LOG_NOTICE("Warm-Restart mode: Receiving delete msg for route with only nh on eth0/docker0: %s %s %s %s",
+                            destipprefix, gw_list.c_str(), intf_list.c_str(), mpls_list.c_str());
+
+                    vector<FieldValueTuple> fvVector;
+                    const KeyOpFieldsValuesTuple kfv = std::make_tuple(destipprefix,
+                                                                       DEL_COMMAND,
+                                                                       fvVector);
+                    m_warmStartHelper.insertRefreshMap(kfv);
+                }
             }
             return;
         }
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -733,6 +733,17 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
         {
             SWSS_LOG_DEBUG("Skip routes to eth0 or docker0: %s %s %s",
                     destipprefix, gw_list.c_str(), intf_list.c_str());
+            // If intf_list has only this interface, that means all of the next hops of this route 
+            // have been removed and the next hop on the eth0/docker0 has become the only next hop. 
+            // In this case since we do not want the route with next hop on eth0/docker0, we return. 
+            // But still we need to clear the route from the APPL_DB. Otherwise the APPL_DB and data 
+            // path will be left with stale route entry
+            if(alsv.size() == 1)
+            {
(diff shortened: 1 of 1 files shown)


prsunny:
@vganesan-nokia , i think you may want to follow the CLA guidelines to sign.

abdosi (dismissed):

abdosi:
@prsunny please check on this.

arlakshm (approved):

prsunny:
For some reason, this PR is consistently failing. @vganesan-nokia , can you please rebase to latest and force-push?

vganesan-nokia:
@prsunny: Rebased and force pushed. I still see vstest failures (test_warmreboot.py).

prsunny:
ok, seems like this test is consistently failing on the warmboot, can you please run the test manually?

vganesan-nokia:
@prsunny: Fixed unit test failures. Please make arrangements to get this reviewed.

arlakshm (approved):

arlakshm:
@prsunny, can you check the change related to warm reboot.

dprital:
@vganesan-nokia , I notived that this PR is not included in 202211 branch so I added a label for it. @StormLiangMS , can you please cherry pick it to 202211 ?

fpmsyncd/routesync.cpp
```diff
@@ -733,6 +733,32 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
         {
             SWSS_LOG_DEBUG("Skip routes to eth0 or docker0: %s %s %s",
                     destipprefix, gw_list.c_str(), intf_list.c_str());
+            // If intf_list has only this interface, that means all of the next hops of this route 
+            // have been removed and the next hop on the eth0/docker0 has become the only next hop. 
+            // In this case since we do not want the route with next hop on eth0/docker0, we return. 
+            // But still we need to clear the route from the APPL_DB. Otherwise the APPL_DB and data 
+            // path will be left with stale route entry
+            if(alsv.size() == 1)
```
peter-yu, line 741: From the context, it seems for del msg, should we check it by
if((nlmsg_type == RTM_DELROUTE) && (alsv.size() == 1))
?

vganesan-nokia: Yes.

prsunny: @vganesan-nokia , do you want to address the comment?

peter-yu: Even it is a RTM_NEWROUTE on eth0/docker0, it may run into

if (!warmRestartInProgress) 
{
  ... 
  m_routeTable.del(destipprefix)
}

then the route is deleted from APPL_DB and ASIC_DB ?
It happened on my real device with  default route (0.0.0.0/0) is removed.

vganesan-nokia: > @vganesan-nokia , do you want to address the comment?

Yes. I'll make a PR to fix this.

PR #2558, opened 2022-12-05
Avoid printing to syslog in case port status is not changed

arfeigin:
**What I did**
Print port status change to syslog only after checking that new status differs from current status.

**Why I did it**
Printing of messages of port being changed from up to up (while actually port status is not being changed) cause log analyzer to fail on fast-reboot test.
According to sonic-mgmt tests/platform_tests/reboot_timing_constants.py

**How I verified it**
Fast-reboot community test.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -6618,13 +6618,13 @@ void PortsOrch::doTask(NotificationConsumer &consumer)
 
 void PortsOrch::updatePortOperStatus(Port &port, sai_port_oper_status_t status)
 {
-    SWSS_LOG_NOTICE("Port %s oper state set from %s to %s",
-            port.m_alias.c_str(), oper_status_strings.at(port.m_oper_status).c_str(),
-            oper_status_strings.at(status).c_str());
     if (status == port.m_oper_status)
     {
         return;
     }
+    SWSS_LOG_NOTICE("Port %s oper state set from %s to %s",
+            port.m_alias.c_str(), oper_status_strings.at(port.m_oper_status).c_str(),
+            oper_status_strings.at(status).c_str());
 
     if (port.m_type == Port::PHY)
     {


orchagent/portsorch.cpp
```diff
@@ -6618,13 +6618,13 @@ void PortsOrch::doTask(NotificationConsumer &consumer)
...
 void PortsOrch::updatePortOperStatus(Port &port, sai_port_oper_status_t status)
 {
-    SWSS_LOG_NOTICE("Port %s oper state set from %s to %s",
-            port.m_alias.c_str(), oper_status_strings.at(port.m_oper_status).c_str(),
-            oper_status_strings.at(status).c_str());
     if (status == port.m_oper_status)
     {
         return;
     }
+    SWSS_LOG_NOTICE("Port %s oper state set from %s to %s",
```
prsunny, line 6625: lgtm, but why would this be notified to orchagent if there is no change in status?

arfeigin: I am not sure I understand.
The current behavior was that it was being notified even when status is not being changed and followed with returning from the function. I suggested to have first the check if port status is actually being changed - in case it doesn't nothing will be notified as status wasn't changed.

prgeor: @arfeigin what about updating the DB unnecessarily even though there is no actual status change?

prgeor:
@arfeigin can we fix the test case? my concern is if there is some bug in your SAI which keeps sending update even though there is no change, we keep writing to the DB unnecessarily or even otherwise since we don't expect unwanted notification.

arfeigin:
> @arfeigin can we fix the test case? my concern is if there is some bug in your SAI which keeps sending update even though there is no change, we keep writing to the DB unnecessarily or even otherwise since we don't expect unwanted notification.

I went over the flow, the update is done twice and not ongoing. One time port status is updated as part of portsorch's doTask in startup after reset, and the second time is done due to switching to apply view. This happens since fast-reboot is using warm-reboot and called by OrchDeamon's warmRestoreAndSyncUp function.

orchagent/portsorch.cpp
arfeigin: You are correct, I think a fix for that will be that I will change the function to return a bool indicating if port status was actually changed and only if it was actually changed the DB will be updated.
Shall I add it to the PR?

prgeor: @arfeigin can you please check why there are OA notification from your platform or SDK even though the port operational status has not changed?

liat-grozovik:
@prsunny can we move forward and merge it?

prgeor:
> @prsunny can we move forward and merge it?

@liat-grozovik @arfeigin what is the harm in seeing this notification twice during fast-reboot? My concern is we don't want to miss barrage of such notification if there is any underlying issue in the platform/sdk/syncd

arfeigin:
As discussed, PR is close, issue was created on this topic: https://github.com/sonic-net/sonic-buildimage/issues/13254

PR #2565, opened 2022-12-09
[202205][dual-tor] use 'egress' port for Mellanox platform for ACL rules for standby ToR

ayurkiv-nvda:

**What I did**
For Dual-Tor scenario on standby ports drop packets on the egress stage for mellanox platform only.
Remain old behavior for other platform.

**NOTE**
Currently, it is short-term solution which will be replaced with more generic one

**Why I did it**
Current SAI ACL implementation drops both data plane and control-plane traffic if rules are installed on ingress stage.
But DualToR expects only data-plane traffic to be dropped, and control-plane should not be affected.
In order to handle this limitation, we will allow traffic to go into ingress port, then handle control plane traffic and drop that traffic on egress ports

**How I verified it**
Setup Dual-ToR, both Active and Standby should have "healthy" status

**Details if related**

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -49,6 +49,13 @@ extern sai_router_interface_api_t* sai_router_intfs_api;
 #define MUX_HW_STATE_UNKNOWN "unknown"
 #define MUX_HW_STATE_ERROR "error"
 
+
+static inline bool isIngressAcl()
+{
+    string platform = getenv("platform") ? getenv("platform") : "";
+    return platform != MLNX_PLATFORM_SUBSTRING;
+}
+
 const map<std::pair<MuxState, MuxState>, MuxStateChange> muxStateTransition =
 {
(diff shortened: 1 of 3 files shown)


orchagent/muxorch.cpp
```diff
@@ -753,7 +759,7 @@ MuxAclHandler::MuxAclHandler(sai_object_id_t port, string alias)
     SWSS_LOG_ENTER();
 
     // There is one handler instance per MUX port
-    string table_name = MUX_ACL_TABLE_NAME;
+    string table_name = getPlatform() == MLNX_PLATFORM_SUBSTRING ? EGRESS_TABLE_DROP : MUX_ACL_TABLE_NAME;
```
prsunny, line 762: Can we rename this function to say, `isIngressAcl`() and chose the right one? You can temporarily add MLNX to that function and later just need to modify that function.

bingwang-ms (approved):

prsunny (approved):

PR #2571, opened 2022-12-13
Fixed set mtu for deleted subintf due to late notification

EdenGri:
**What I did**

Ignores errors on the set MTU command for subinterface when the subinterface state is not OK.

**Why I did it**

A race condition between the portmgrd and the intfmgrd sometimes causes running a set MTU command on a deleted subinterface.
The logs and the error:
```
INFO swss#supervisord: intfmgrd Cannot find device "Ethernet32.58"
ERR swss#intfmgrd: :- main: Runtime error: /sbin/ip link set "Ethernet32.58" mtu "9100" :
INFO swss#supervisord 2022-11-08 05:53:33,057 INFO exited: intfmgrd (exit status 255; not expected)
```

**How I verified it**

Run the test_loopback_action_reload test and saw no errors in the logs.

**Details if related**

### DIFF
--- a/cfgmgr/intfmgr.cpp
+++ b/cfgmgr/intfmgr.cpp
@@ -447,9 +447,19 @@ std::string IntfMgr::setHostSubIntfMtu(const string &alias, const string &mtu, c
     }
     SWSS_LOG_INFO("subintf %s active mtu: %s", alias.c_str(), subifMtu.c_str());
     cmd << IP_CMD " link set " << shellquote(alias) << " mtu " << shellquote(subifMtu);
-    EXEC_WITH_ERROR_THROW(cmd.str(), res);
+    std::string cmd_str = cmd.str();
+    int ret = swss::exec(cmd_str, res);
 
-    return subifMtu;
+    if (ret && !isIntfStateOk(alias))
+    {
+        // Can happen when a DEL notification is sent by portmgrd immediately followed by a new SET notif
+        SWSS_LOG_WARN("Setting mtu to alias:%s netdev failed with cmd:%s, rc:%d, error:%s", alias.c_str(), cmd_str.c_str(), ret, res.c_str());
+    }
+    else if(ret)
+    {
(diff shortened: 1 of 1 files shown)


dprital (dismissed):

cfgmgr/intfmgr.cpp
```diff
@@ -447,9 +447,20 @@ std::string IntfMgr::setHostSubIntfMtu(const string &alias, const string &mtu, c
     }
     SWSS_LOG_INFO("subintf %s active mtu: %s", alias.c_str(), subifMtu.c_str());
     cmd << IP_CMD " link set " << shellquote(alias) << " mtu " << shellquote(subifMtu);
-    EXEC_WITH_ERROR_THROW(cmd.str(), res);
+    std::string cmd_str = cmd.str();
+    int ret = swss::exec(cmd_str, res);
 
-    return subifMtu;
+    if (ret && !isIntfStateOk(alias))
```
prsunny, line 453: This is already checked in the `doIntfGeneralTask `under SET operation. So why check again here? is there a race condition?

prsunny (changes requested):

cfgmgr/intfmgr.cpp
EdenGri: We are not visiting the `"doIntfGeneralTask"` function in this flow.
The flow is:
1. changing the "mtu" field in the PORT_TABLE in state DB
2. `doTask`
3. `doPortTableTask`
4. `updateSubIntfMtu`
5. `setHostSubIntfMtu`

dprital:
@prsunny - Can you please approve this PR ?

dprital (approved):

cfgmgr/intfmgr.cpp
prsunny: ok, does this happen in the `updateSubIntfAdminStatus `flow as well? Or why only `mtu `change is handled?

prsunny (approved):

PR #2593, opened 2022-12-30
[Neighsync] Handle Mac address 'none' during neigh add

prsunny:
**What I did**
In some rare cases, kernel neighbor add notifications are received with Mac string as 'none'. This is from the following libnl code:

https://www.infradead.org/~tgr/libnl/doc/api/lib_2addr_8c_source.html

image

This can cause a neighsyncd crash as below from [here](https://github.com/sonic-net/sonic-swss-common/blob/master/common/macaddress.cpp#L20)

```
Dec 12 07:36:37.149555 ST12-M0 INFO swss#/supervisord: neighsyncd terminate called after throwing an instance of 'std::invalid_argument'
Dec 12 07:36:37.149555 ST12-01M0 INFO swss#/supervisord: neighsyncd   what():  can't parse mac address 'none'
```

**Why I did it**

Prevent crash by checking if MacString is 'none' during neighbor 'add' operation. 

**How I verified it**

**Details if related**

### DIFF
--- a/neighsyncd/neighsync.cpp
+++ b/neighsyncd/neighsync.cpp
@@ -143,6 +143,12 @@ void NeighSync::onMsg(int nlmsg_type, struct nl_object *obj)
         nl_addr2str(rtnl_neigh_get_lladdr(neigh), macStr, MAX_ADDR_SIZE);
     }
 
+    if (!delete_key && !strncmp(macStr, "none", MAX_ADDR_SIZE))
+    {
+        SWSS_LOG_NOTICE("Mac address is 'none' for ADD op, ignoring for %s", ipStr);
+        return;
+    }
+
     /* Ignore neighbor entries with Broadcast Mac - Trigger for directed broadcast */
     if (!delete_key && (MacAddress(macStr) == MacAddress("ff:ff:ff:ff:ff:ff")))
     {


neighsyncd/neighsync.cpp
```diff
@@ -143,6 +143,12 @@ void NeighSync::onMsg(int nlmsg_type, struct nl_object *obj)
         nl_addr2str(rtnl_neigh_get_lladdr(neigh), macStr, MAX_ADDR_SIZE);
     }
 
+    if (!delete_key && !strncmp(macStr, "none", MAX_ADDR_SIZE))
```
theasianpianist, line 146: Is it possible for the MAC to be set to "none" on a dual ToR device? If this happens and also triggers the zero MAC logic directly above, we could have the "none" MAC be overwritten by an empty MAC which would cause unexpected tunnel routes to be created.

prsunny: I thought about dualtor case and it is unlikely we hit this issue in a stable system. In case it happens, i think treating it as zero mac should be fine.

theasianpianist (approved):

weiliu-ivy:
@prsunny @theasianpianist could you help add the reason that this PR can bypass coverage failure? Thank you!

prsunny:
> @prsunny @theasianpianist could you help add the reason that this PR can bypass coverage failure? Thank you!

hi @weiliu-ivy , this cannot be simulated by the test and hence bypassed. This can be treated as an exception.

weiliu-ivy:
> > @prsunny @theasianpianist could you help add the reason that this PR can bypass coverage failure? Thank you!
> 
> hi @weiliu-ivy , this cannot be simulated by the test and hence bypassed. This can be treated as an exception.

@prsunny Got it. Thank you!

PR #2595, opened 2023-01-01
[202205] Fixed set mtu for deleted subintf due to late notification

dprital:
PR against 202205 based on the following PR: https://github.com/sonic-net/sonic-swss/pull/2571

**What I did**

Ignores errors on the set MTU command for subinterface when the subinterface state is not OK.

**Why I did it**

A race condition between the portmgrd and the intfmgrd sometimes causes running a set MTU command on a deleted subinterface.
The logs and the error:

> INFO swss#supervisord: intfmgrd Cannot find device "Ethernet32.58"
> ERR swss#intfmgrd: :- main: Runtime error: /sbin/ip link set "Ethernet32.58" mtu "9100" :
> INFO swss#supervisord 2022-11-08 05:53:33,057 INFO exited: intfmgrd (exit status 255; not expected)

**How I verified it**

Run the test_loopback_action_reload test and saw no errors in the logs.

**Details if related**


liat-grozovik (approved):
Original PR on master approved by prsunny. moving a head with this PR approval as well

PR #2601, opened 2023-01-06
[MuxOrch] Enabling neighbor when adding in active state

Ndancejic:
What I did: called enable when adding a neighbor in active state

Why I did it: NH routes weren't being updated if they existed in standby state before neighbor is added

### DIFF
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -591,6 +591,12 @@ void NeighOrch::decreaseNextHopRefCount(const NextHopKey &nexthop, uint32_t coun
     assert(hasNextHop(nexthop));
     if (m_syncdNextHops.find(nexthop) != m_syncdNextHops.end())
     {
+        if ((m_syncdNextHops[nexthop].ref_count - (int)count) < 0)
+        {
+            SWSS_LOG_ERROR("Ref count cannot be negative for next_hop_id: 0x%" PRIx64 " with ip: %s and alias: %s",
+                   m_syncdNextHops[nexthop].next_hop_id, nexthop.ip_address.to_string().c_str(), nexthop.alias.c_str());
+            return;
+        }
         m_syncdNextHops[nexthop].ref_count -= count;
     }
 }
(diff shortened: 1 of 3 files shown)


prsunny (dismissed):

orchagent/neighorch.cpp
```diff
@@ -591,6 +591,12 @@ void NeighOrch::decreaseNextHopRefCount(const NextHopKey &nexthop, uint32_t coun
     assert(hasNextHop(nexthop));
     if (m_syncdNextHops.find(nexthop) != m_syncdNextHops.end())
     {
+        if ((m_syncdNextHops[nexthop].ref_count - (int)count) < 0)
+        {
+            SWSS_LOG_ERROR("Ref count cannot be negative for next_hop_id: 0x%" PRIx64 " with ip: %s and alias: %s",
+                   m_syncdNextHops[nexthop].next_hop_id, nexthop.ip_address.to_string().c_str(), nexthop.alias.c_str());
+            return;
```
theasianpianist, line 598: Before we return, should we still set the ref count to 0?

tests/test_mux.py
```diff
@@ -537,6 +537,66 @@ def create_and_test_route(self, appdb, asicdb, dvs, dvs_route):
...
+        # Set state to active, learn neighbor again
+        self.set_mux_state(appdb, "Ethernet0", "active")
+        self.check_nexthop_in_asic_db(asicdb, rtkeys[0], True)
+        self.check_nexthop_in_asic_db(asicdb, rtkeys_ipv6[0], True)
+
+        self.add_neighbor(dvs, neigh_ip, mac)
+        self.add_neighbor(dvs, neigh_ipv6, mac)
+        self.check_nexthop_in_asic_db(asicdb, rtkeys[0])
+        self.check_nexthop_in_asic_db(asicdb, rtkeys_ipv6[0])
+
```
theasianpianist, line 587: Might be worth it to try flushing the neighbor again after re-learning it to make sure nothing crashes?

orchagent/neighorch.cpp
Ndancejic: Hm yeah, I think that should be alright especially if the decrement is expecting to bring the refcount down to 0, it would reset that expectation. I'll update the PR

theasianpianist (approved):

PR #2606, opened 2023-01-10
swss: Add voq counters.

skbarista:
This pr contains changes from https://github.com/sonic-net/sonic-swss/pull/2467 and https://github.com/sonic-net/sonic-swss/pull/2552.

* Add m_voq_ids to SystemPortInfo to maintain the list of queue ids.
* Add two new tables COUNTERS_SYSTEM_PORT_NAME_MAP and COUNTERS_VOQ_NAME_MAP to enable queuestat to differentiate between Port Tx queues and Voqs.
* Add a new function initializeVoqs that retrieves the number of voqs for a system port and stores the voq object ids in m_voq_ids
* Add code to handle queue type SAI_QUEUE_TYPE_UNICAST_VOQ.
* Initialize voqs and populate COUNTERS_SYSTEM_PORT_NAME_MAP in addSystemPorts function.
* Update generateQueueMap to generate queue maps for both Txqs and Voq.
* For PHY ports in a voq system both Txqs and Voqs are instantiated. For Voqs of remote system port, only Voq counters are initialized.

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -368,6 +368,8 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
     /* Initialize counter table */
     m_counter_db = shared_ptr<DBConnector>(new DBConnector("COUNTERS_DB", 0));
     m_counterTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_PORT_NAME_MAP));
+    m_counterSysPortTable = unique_ptr<Table>(
+                    new Table(m_counter_db.get(), COUNTERS_SYSTEM_PORT_NAME_MAP));
     m_counterLagTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_LAG_NAME_MAP));
     FieldValueTuple tuple("", "");
     vector<FieldValueTuple> defaultLagFv;
@@ -382,6 +384,7 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
 
     /* Initialize queue tables */
     m_queueTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_QUEUE_NAME_MAP));
+    m_voqTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_VOQ_NAME_MAP));
     m_queuePortTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_QUEUE_PORT_MAP));
     m_queueIndexTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_QUEUE_INDEX_MAP));
     m_queueTypeTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_QUEUE_TYPE_MAP));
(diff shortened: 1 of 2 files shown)


kenneth-arista:
@arlakshm @rlhui

orchagent/portsorch.cpp
```diff
@@ -6017,15 +6104,27 @@ void PortsOrch::generateQueueMapPerPort(const Port& port, FlexCounterQueueStates
...
         queueVector.emplace_back(name.str(), id);
-        queuePortVector.emplace_back(id, sai_serialize_object_id(port.m_port_id));
+        if (voq)
+        {
+            queuePortVector.emplace_back(id, sai_serialize_object_id(port.m_system_port_oid));
+        }
+        else
+        {
+            queuePortVector.emplace_back(id, sai_serialize_object_id(port.m_port_id));
+        }
```
vmittal-msft, line 6115: image

this code seems to be present in master but not here. is this needed ?

skbarista: Did you mean the setting of queueTypeVector and queueIndexVector?

orchagent/portsorch.cpp
```diff
@@ -5981,34 +6033,69 @@ void PortsOrch::generateQueueMap(map<string, FlexCounterQueueStates> queuesState
...
+            if (gMySwitchType == "voq")
+            {
+               generateQueueMapPerPort(it.second, queuesStateVector.at(it.second.m_alias), true);
+            }
+        }
+        if (it.second.m_type == Port::SYSTEM)
+        {
+            if (!queuesStateVector.count(it.second.m_alias))
+            {
+                auto maxQueueNumber = getNumberOfPortSupportedQueueCounters(it.second.m_alias);
```
arlakshm, line 6046: the `getNumberOfPortSupportedQueueCounters` is getting the `m_queue_ids.size()` for the port. 
For system port what will be the size of `m_queue_ids`. I dont see `m_queue_ids` getting updated for system port

arlakshm (commented):
Hi @skbarista, can you confirm if voq counter polling works with latest SAI 7.1 Debian? I think with this change we should see the voq counters also in the counter_db no additional changes are needed. Is that correct?

orchagent/portsorch.cpp
skbarista: The size will be zero but since queuesStateVector is not used for system ports, we do not have any issue.

skbarista: yes we will see the voq counters in counters db. I have ported https://github.com/sonic-net/sonic-utilities/pull/1827 in our env. I will create a pr for that too. That way you can see the voq counters using queuestat

skbarista: Created https://github.com/sonic-net/sonic-utilities/pull/2591 for the queuestat enhancement.

orchagent/portsorch.cpp
skbarista: This is existing code and not added by the pr. If you expand the diff you will see the code in 202205 as well.

skbarista:
[voq.txt](https://github.com/sonic-net/sonic-swss/files/10388072/voq.txt)

orchagent/portsorch.cpp
skbarista: https://github.com/sonic-net/sonic-swss/files/10388072/voq.txt has the output with the sai version of my chassis and output from queuestat -V. This is using the image with voq changes and fabric counter changes removed.

rlhui:
@skbarista , please add the corresponding master branch PR for this?

skbarista:
@rlhui This pr contains changes from https://github.com/sonic-net/sonic-swss/pull/2467 and https://github.com/sonic-net/sonic-swss/pull/2552. I will update the description as well.

arlakshm (approved):

vmittal-msft (approved):

PR #2621, opened 2023-01-13
[refactor]Refactoring sai handle status

dgsudharsan:
**What I did**
Moved SAI handle status APIs from Orch class to saihelper files.

**Why I did it**
Orch class is used in manager daemons. When extending saihelper APIs to include SAI calls this creates linker errors. Logically these APIs shouldn't be part of Orch class. Hence moved them to saihelper.

**How I verified it**
Existing UT should verify

**Details if related**


prsunny (approved):

prsunny:
@qiluo-msft , please review this and this is a refactoring for collecting sdkdump during SAI failures. Current infra in orch class is unable to handle it and hence moved to saihelper module.

qiluo-msft (approved):

prsunny:
This is code refactoring (no new code introduced) and hence bypassing coverage failure.

PR #2626, opened 2023-01-20
fix EvpnRemoteVnip2pOrch warmboot check failure

jcaiMR:
**What I did**
Fix #12361

The trigger of this issue is warm-restart check failure, and EvpnRemoteVnip2pOrch::addOperation() caused it.
it always return false because of evpn_orch->getEVPNVtep() always return false cause warmboot check failure. 
The issue only happened on Broadcom platform since it support P2P vxlan tunnel, and will run into EvpnRemoteVnip2pOrch.
Nvidia SAI currently only support P2MP tunnel, so no issue on that platform. 
 
```
Jan 18 11:37:14.701069 str-dx010-acs-5 WARNING swss#orchagent: :- addOperation: Vxlan tunnelPort doesn't exist: 192.168.8.1
Jan 18 11:37:14.701177 str-dx010-acs-5 WARNING swss#orchagent: :- addTunnelUser: Unable to find EVPN VTEP. user=0 remote_vtep=19
```

Seems vtep instance only created with vxlan tunnel when tunnel creation source is EVPN.
```
VxlanTunnel::VxlanTunnel(string name, IpAddress srcIp, IpAddress dstIp, tunnel_creation_src_t src) 

                :tunnel_name_(name), src_ip_(srcIp), dst_ip_(dstIp), src_creation_(src) 

{ 
   VxlanTunnelOrch* tunnel_orch = gDirectory.get<VxlanTunnelOrch*>(); 
   if (dstIp.isZero()) 
   { 
       tunnel_orch->addVTEP(this, srcIp); 
       vtep_ptr = NULL; 
   } 
   else if (src_creation_ == TNL_CREATION_SRC_EVPN)  
   { 
       vtep_ptr = tunnel_orch->getVTEP(srcIp); 
       tunnel_orch->addRemoveStateTableEntry(name,srcIp, dstIp, src, true); 
   } 
} 
```

So here the fix is let EvpnRemoteVnip2pOrch::addOperation() return true when local VTEP is not exists, the logic same like EvpnRemoteVnip2mpOrch::addOperation().

**Why I did it**

**How I verified it**
run arp/test_wr_arp.py on Broadcom platform. 

**Details if related**

### DIFF
--- a/orchagent/vxlanorch.cpp
+++ b/orchagent/vxlanorch.cpp
@@ -2347,6 +2347,14 @@ bool EvpnRemoteVnip2pOrch::addOperation(const Request& request)
         return true;
     }
 
+    EvpnNvoOrch* evpn_orch = gDirectory.get<EvpnNvoOrch*>();
+    auto vtep_ptr = evpn_orch->getEVPNVtep();
+    if (!vtep_ptr) {
+        SWSS_LOG_WARN("Remote VNI add: Source VTEP not found. remote=%s vid=%d",
+                      remote_vtep.c_str(), vlan_id);
+        return true;
+    }
+
     VxlanTunnelOrch* tunnel_orch = gDirectory.get<VxlanTunnelOrch*>();
     Port tunnelPort, vlanPort;
 


prsunny:
@srj102 , please review this as it is a blocker and fixing a warmboot issue

dgsudharsan:
@jcaiMR @prsunny Do we know why EvpnRemoteVnip2pOrch::addOperation is called in non EVPN scenario. This will generally be called when VXLAN_REMOTE_VNI table is present. Is it populated in non evpn scenario?

jcaiMR:
> @jcaiMR @prsunny Do we know why EvpnRemoteVnip2pOrch::addOperation is called in non EVPN scenario. This will generally be called when VXLAN_REMOTE_VNI table is present. Is it populated in non evpn scenario?

Yes, the issue happened in non-evpn scenario, seems in function OrchDaemon::init(), we directly create evpn_remote_vni_orch and push into m_orchList. 
```
 if (vxlan_tunnel_orch->isDipTunnelsSupported())
    {
        EvpnRemoteVnip2pOrch* evpn_remote_vni_orch = new EvpnRemoteVnip2pOrch(m_applDb, APP_VXLAN_REMOTE_VNI_TABLE_NAME);
        gDirectory.set(evpn_remote_vni_orch);
        m_orchList.push_back(evpn_remote_vni_orch);
    }
    else
    {
        EvpnRemoteVnip2mpOrch* evpn_remote_vni_orch = new EvpnRemoteVnip2mpOrch(m_applDb, APP_VXLAN_REMOTE_VNI_TABLE_NAME);
        gDirectory.set(evpn_remote_vni_orch);
        m_orchList.push_back(evpn_remote_vni_orch);
    }
```

And warmboot check function OrchDaemon::warmRestartCheck( ) checked all orch agents in m_orchList, and finally call addOperation() for each orch agent. I didn't find any check logic for VXLAN_REMOTE_VNI  table before walking orch agent list.

prsunny (approved):

vaibhavhd (approved):

orchagent/vxlanorch.cpp
```diff
@@ -2347,6 +2347,14 @@ bool EvpnRemoteVnip2pOrch::addOperation(const Request& request)
         return true;
     }
 
+    EvpnNvoOrch* evpn_orch = gDirectory.get<EvpnNvoOrch*>();
```
srj102, line 2350: Line 2373 has this same check and hence it can be removed from there if this change needs to go in.

orchagent/vxlanorch.cpp
```diff
@@ -2347,6 +2347,14 @@ bool EvpnRemoteVnip2pOrch::addOperation(const Request& request)
...
+    auto vtep_ptr = evpn_orch->getEVPNVtep();
+    if (!vtep_ptr) {
+        SWSS_LOG_WARN("Remote VNI add: Source VTEP not found. remote=%s vid=%d",
```
srj102, line 2353: What does this log say during warm boot ? Does it point to anything valid ?
ie what is the remote ip and what is the vid ?
If it is some spurious data then we need to investigate that as this can be seen on any of the other tables in other OAs. 

Quite clearly the null check cannot be the fix especially in a non EVPN scenario.
We would need to see why is this being called when there is no explanation for the presence of an entry in the VXLAN_REMOTE_VNI table.

srj102:
> > @jcaiMR @prsunny Do we know why EvpnRemoteVnip2pOrch::addOperation is called in non EVPN scenario. This will generally be called when VXLAN_REMOTE_VNI table is present. Is it populated in non evpn scenario?
> 
> Yes, the issue happened in non-evpn scenario, seems in function OrchDaemon::init(), we directly create evpn_remote_vni_orch and push into m_orchList.
> 
> ```
>  if (vxlan_tunnel_orch->isDipTunnelsSupported())
>     {
>         EvpnRemoteVnip2pOrch* evpn_remote_vni_orch = new EvpnRemoteVnip2pOrch(m_applDb, APP_VXLAN_REMOTE_VNI_TABLE_NAME);
>         gDirectory.set(evpn_remote_vni_orch);
>         m_orchList.push_back(evpn_remote_vni_orch);
>     }
>     else
>     {
>         EvpnRemoteVnip2mpOrch* evpn_remote_vni_orch = new EvpnRemoteVnip2mpOrch(m_applDb, APP_VXLAN_REMOTE_VNI_TABLE_NAME);
>         gDirectory.set(evpn_remote_vni_orch);
>         m_orchList.push_back(evpn_remote_vni_orch);
>     }
> ```
> 
> And warmboot check function OrchDaemon::warmRestartCheck( ) checked all orch agents in m_orchList, and finally call addOperation() for each orch agent. I didn't find any check logic for VXLAN_REMOTE_VNI table before walking orch agent list, Do I missed something ?
> 
> attached some trace during warmboot:
> 
> ```
> Thread 1 "orchagent" hit Breakpoint 1, EvpnRemoteVnip2mpOrch::addOperation (this=0x55ecbd1e7af0, request=...) at vxlanorch.cpp:2476
> 2476    vxlanorch.cpp: No such file or directory.
> (gdb) bt
> #0  EvpnRemoteVnip2mpOrch::addOperation (this=0x55ecbd1e7af0, request=...) at vxlanorch.cpp:2476
> #1  0x000055ecbb8d1c60 in Orch2::doTask (this=0x55ecbd1e7af0, consumer=...) at orch.cpp:1063
> #2  0x000055ecbb8d5a9e in Consumer::drain (this=0x55ecbd1e85b0) at orch.cpp:241
> #3  Consumer::drain (this=0x55ecbd1e85b0) at orch.cpp:238
> #4  Consumer::execute (this=0x55ecbd1e85b0) at orch.cpp:235
> #5  0x000055ecbb8c54c9 in OrchDaemon::start (this=this@entry=0x55ecbd0bfd80) at orchdaemon.cpp:755
> #6  0x000055ecbb843fe0 in main (argc=<optimized out>, argv=<optimized out>) at main.cpp:735
> (gdb) c
> ```

Having the p2premotevniorch in the orchList does not explain why there was an event. 
Could you dump the contents of the VXLAN_REMOTE_VNI table before the WB command is issued ?
Does it have any entries ?

keys *VXLAN_REMOTE_VNI*

srj102 (changes requested):

orchagent/vxlanorch.cpp
jcaiMR: > > > @jcaiMR @prsunny Do we know why EvpnRemoteVnip2pOrch::addOperation is called in non EVPN scenario. This will generally be called when VXLAN_REMOTE_VNI table is present. Is it populated in non evpn scenario?
> > 
> > 
> > Yes, the issue happened in non-evpn scenario, seems in function OrchDaemon::init(), we directly create evpn_remote_vni_orch and push into m_orchList.
> > ```
> >  if (vxlan_tunnel_orch->isDipTunnelsSupported())
> >     {
> >         EvpnRemoteVnip2pOrch* evpn_remote_vni_orch = new EvpnRemoteVnip2pOrch(m_applDb, APP_VXLAN_REMOTE_VNI_TABLE_NAME);
> >         gDirectory.set(evpn_remote_vni_orch);
> >         m_orchList.push_back(evpn_remote_vni_orch);
> >     }
> >     else
> >     {
> >         EvpnRemoteVnip2mpOrch* evpn_remote_vni_orch = new EvpnRemoteVnip2mpOrch(m_applDb, APP_VXLAN_REMOTE_VNI_TABLE_NAME);
> >         gDirectory.set(evpn_remote_vni_orch);
> >         m_orchList.push_back(evpn_remote_vni_orch);
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And warmboot check function OrchDaemon::warmRestartCheck( ) checked all orch agents in m_orchList, and finally call addOperation() for each orch agent. I didn't find any check logic for VXLAN_REMOTE_VNI table before walking orch agent list, Do I missed something ?
> > attached some trace during warmboot:
> > ```
> > Thread 1 "orchagent" hit Breakpoint 1, EvpnRemoteVnip2mpOrch::addOperation (this=0x55ecbd1e7af0, request=...) at vxlanorch.cpp:2476
> > 2476    vxlanorch.cpp: No such file or directory.
> > (gdb) bt
> > #0  EvpnRemoteVnip2mpOrch::addOperation (this=0x55ecbd1e7af0, request=...) at vxlanorch.cpp:2476
> > #1  0x000055ecbb8d1c60 in Orch2::doTask (this=0x55ecbd1e7af0, consumer=...) at orch.cpp:1063
> > #2  0x000055ecbb8d5a9e in Consumer::drain (this=0x55ecbd1e85b0) at orch.cpp:241
> > #3  Consumer::drain (this=0x55ecbd1e85b0) at orch.cpp:238
> > #4  Consumer::execute (this=0x55ecbd1e85b0) at orch.cpp:235
> > #5  0x000055ecbb8c54c9 in OrchDaemon::start (this=this@entry=0x55ecbd0bfd80) at orchdaemon.cpp:755
> > #6  0x000055ecbb843fe0 in main (argc=<optimized out>, argv=<optimized out>) at main.cpp:735
> > (gdb) c
> > ```
> 
> Having the p2premotevniorch in the orchList does not explain why there was an event. Could you dump the contents of the VXLAN_REMOTE_VNI table before the WB command is issued ? Does it have any entries ?
> 
> keys _VXLAN_REMOTE_VNI_

Thanks for comments. we don't have EVPN configured and "sonic-db-cli APPL_DB keys "VXLAN_REMOTE_VNI*"" is empty.

srj102: yes addOperation cannot be called unless the VXLAN_REMOTE_VNI table is populated. 

My understanding is that warmRestartCheck will not call addOperation. 
It just checks the sizes of the m_toSync map and if it is non zero then returns a false. 

So one possible explanation is there was an earlier operation which resulted in the VXLAN_REMOTE_VNI being populated and resulting in a failure that you saw.  
Under valid EVPN VXLAN configuration one is not expected to see this failure. 

WB operation later on results in the readiness check failing.

jcaiMR: Thanks for suggestion, will have some further debug to find the event trigger.

yxieca: @srj102 I agree with your comments. However, I think we need to include this work-around for the short term. @jcaiMR please leave the issue open after this PR merged. (remove the fix PR comment).

jcaiMR: Got it, have modified the title and removed Fix #. Will use other PR for the root cause solution.

orchagent/vxlanorch.cpp
jcaiMR: Line 2373 will not hit in this case, as it early return at line 2364.

srj102: @jcaiMR Currently the check for the vtep ptr is present only in line 2373 under a couple of if conditions. 

Since the check is now being moved to line 2350 outside of the if checks why do we need the same check in line 2373 ?

    if (tunnel_orch->getTunnelPort(remote_vtep,tunnelPort))
    {
        SWSS_LOG_INFO("Vxlan tunnelPort exists: %s", remote_vtep.c_str());

        if (gPortsOrch->isVlanMember(vlanPort, tunnelPort))
        {

We can remove the following check as it is replicated in line 2350. 

            _EvpnNvoOrch* evpn_orch = gDirectory.get<EvpnNvoOrch*>();
            auto vtep_ptr = evpn_orch->getEVPNVtep();
            if (!vtep_ptr)
            {
                SWSS_LOG_WARN("Remote VNI add: VTEP not found. remote=%s vid=%d",
                              remote_vtep.c_str(),vlan_id);
                return true;
            }_

jcaiMR: > @jcaiMR Currently the check for the vtep ptr is present only in line 2373 under a couple of if conditions.
> 
> Since the check is now being moved to line 2350 outside of the if checks why do we need the same check in line 2373 ?
> 
> ```
> if (tunnel_orch->getTunnelPort(remote_vtep,tunnelPort))
> {
>     SWSS_LOG_INFO("Vxlan tunnelPort exists: %s", remote_vtep.c_str());
> 
>     if (gPortsOrch->isVlanMember(vlanPort, tunnelPort))
>     {
> ```
> 
> We can remove the following check as it is replicated in line 2350.
> 
> ```
>         _EvpnNvoOrch* evpn_orch = gDirectory.get<EvpnNvoOrch*>();
>         auto vtep_ptr = evpn_orch->getEVPNVtep();
>         if (!vtep_ptr)
>         {
>             SWSS_LOG_WARN("Remote VNI add: VTEP not found. remote=%s vid=%d",
>                           remote_vtep.c_str(),vlan_id);
>             return true;
>         }_
> ```

I see, will update in latest diff. And Rajesh, I plan to use this PR as a workaround to make warmboot function pass first. Will raise another PR for the final root cause fix.

orchagent/vxlanorch.cpp
```diff
@@ -2347,6 +2347,14 @@ bool EvpnRemoteVnip2pOrch::addOperation(const Request& request)
         return true;
     }
 
+    EvpnNvoOrch* evpn_orch = gDirectory.get<EvpnNvoOrch*>();
+    auto vtep_ptr = evpn_orch->getEVPNVtep();
+    if (!vtep_ptr) {
```
prsunny, line 2352: Please follow convention. { -> new line

jcaiMR: Sure, will update in latest diff.

srj102 (approved):

orchagent/vxlanorch.cpp
jcaiMR: updated.

orchagent/vxlanorch.cpp
jcaiMR: Diff updated.

prsunny (approved):

PR #2632, opened 2023-01-24
swss: Fix Invalid port oid messages generated because of voq counters.

skbarista:
**What I did**

* Do not add Voq oids to CounterCheckOrch since CounterCheckOrch is not relevant for voqs.

**Why I did it**

To fix the invalid oids messages generated by CounterCheckOrch for voq.

**How I verified it**

Verified that the messages are no longer generated and voq counters are still gathered.

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -6150,16 +6150,12 @@ void PortsOrch::generateQueueMapPerPort(const Port& port, FlexCounterQueueStates
     else
     {
         m_queueTable->set("", queueVector);
+        CounterCheckOrch::getInstance().addPort(port);
     }
 
     m_queuePortTable->set("", queuePortVector);
     m_queueIndexTable->set("", queueIndexVector);
     m_queueTypeTable->set("", queueTypeVector);
-
-    if (!voq)
-    {
-       CounterCheckOrch::getInstance().addPort(port);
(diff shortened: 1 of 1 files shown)


orchagent/portsorch.cpp
```diff
@@ -6156,7 +6156,10 @@ void PortsOrch::generateQueueMapPerPort(const Port& port, FlexCounterQueueStates
     m_queueIndexTable->set("", queueIndexVector);
     m_queueTypeTable->set("", queueTypeVector);
 
-    CounterCheckOrch::getInstance().addPort(port);
+    if (!voq)
+    {
+       CounterCheckOrch::getInstance().addPort(port);
```
prsunny, line 6161: can this be moved to line 6152 so as to avoid repeatin `"if (voq)`

skbarista:
@prsunny @judyjoseph looks like the failures I am facing in this pr is being addressed by https://github.com/sonic-net/sonic-swss/pull/2648

kenneth-arista:
This PR is also needed in master. However, for expediency to address the problem in 202205, this PR is created. Another PR will be created for master.

judyjoseph:
To merge the PR raised in master https://github.com/sonic-net/sonic-swss/pull/2653 first and cherry-pick to 202205.

PR #2633, opened 2023-01-24
[202211][bfdorch] add bfd local discriminator to state DB

baorliu:
Double commit of https://github.com/sonic-net/sonic-swss/pull/2629

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


baorliu:
wrong target branch

PR #2637, opened 2023-01-25
Issue # 2636 : Midstone-100x & Silverstonex: Increasing the netlink b…

praveenraja1:
…uffer size to prevent netlink message drops

Why I did it
We need a way increase netlink buffer size to prevent message drops in swss. This needs to be done for certain platforms alone.

How I did it
Exported new 'env' var to swss docker - "supports_increased_netlink_buffer"

This is currently set only for Midstone & silverstone platforms, in swss code we check this and increase the netlink buffer size.

How to verify it
Check the env in swss container.
Run test case mentioned in https://github.com/sonic-net/sonic-swss/issues/2636 and see that no netlink message drop happens.


praveenraja1:
Related reviews:

https://github.com/sonic-net/sonic-buildimage/pull/13506

https://github.com/sonic-net/sonic-swss-common/pull/743

PR #2638, opened 2023-01-25
[autoneg]Fixing adv interface types to be set when AN is disabled

dgsudharsan:
**What I did**
Allowing advertised interface types to be set even when AN is disabled. The SAI header doesn't restrict this field. In some cases when this field is changed during AN disable, it may not be reflected in SAI leading to errors as seen in the issue below. This will occur when interface type mismatches with advertise interface type and the vendor SAI library has checks.

**Why I did it**
To fix https://github.com/sonic-net/sonic-buildimage/issues/13414

**How I verified it**
Added UT to verify

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -3588,9 +3588,9 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         continue;
                     }
 
-                    if (adv_interface_types != p.m_adv_interface_types && p.m_autoneg == 1)
+                    if (adv_interface_types != p.m_adv_interface_types)
                     {
-                        if (p.m_admin_state_up)
+                        if (p.m_admin_state_up && p.m_autoneg == 1)
                         {
                             /* Bring port down before applying speed */
                             if (!setPortAdminStatus(p, false))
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@prgeor  could you please help to review?

Junchao-Mellanox (approved):

prgeor:
@dgsudharsan we need to update [HLD](https://github.com/sonic-net/SONiC/blame/master/doc/port_auto_neg/port-auto-negotiation-design.md#L224).

prgeor:
> @dgsudharsan we need to update [HLD](https://github.com/sonic-net/SONiC/blame/master/doc/port_auto_neg/port-auto-negotiation-design.md#L224). 

Now that we allow changing interface adv type even if AN=disabled

image

dgsudharsan:
> > @dgsudharsan we need to update [HLD](https://github.com/sonic-net/SONiC/blame/master/doc/port_auto_neg/port-auto-negotiation-design.md#L224).
> 
> Now that we allow changing interface adv type even if AN=disabled
> 
> image

Hi @prgeor I believe even previously we did the same. The HLD captures the behavior correctly. We accept the advertised-interface type even without AN enabled but that will be effective only when AN is turned on. We previously didn't block.

prgeor (approved):

PR #2643, opened 2023-01-26
Remove counters for Mux ACL

devpatha:
**What I did**
Do not use counters for mux ACL

**Why I did it**
Noticed that with PR #1982 counters are enabled by default. It was not the case earlier.
User won't be able to see it with `aclshow` and it will waste valuable hardware counter resource.

**How I verified it**

**Details if related**


bingwang-ms (approved):

devpatha:
@bingwang-ms could we please re-run checks here? This has already been merged into 202205 and would like to be there in master as well.

kevinskwang (approved):

yxieca:
#1982 was only committed to master branch. @StormLiangMS  please check if it was included in 202211 branch. From merge time of #1982, it might have been included in 202211 branch.

PR #2644, opened 2023-01-27
[sai_failure_dump]Invoking dump during SAI failure

dgsudharsan:
HLD: https://github.com/sonic-net/SONiC/pull/1212
Depends on https://github.com/sonic-net/sonic-sairedis/pull/1198
**What I did**
Added logic to invoke SAI failure dump during any SAI programming failure before invoking abort.

**Why I did it**
To collect necessary dumps in problem state in syncd before abort is called and all processes restarts

**How I verified it**
Manual verification. Added UT to cover abort scenario as well.

**Details if related**


dgsudharsan:
@prsunny  I have added tests to cover all lines  except main file (which cannot be included) but still the coverage throws error. I feel there is some issue with coverage here. Can you please help?

prsunny (approved):

prsunny:
Tests are added for coverage but for some reason, it is not showing in the tool. Skipping the check as the changes are in the failure path.

yxieca:
@dgsudharsan please help create a separate PR for 202205 branch.

dgsudharsan:
@yxieca I already created and its merged https://github.com/sonic-net/sonic-swss/pull/2661. Can you please update submodule?

PR #2650, opened 2023-02-02
Remove counters for Mux ACL

devpatha:
**What I did**
Do not use counters for mux ACL

**Why I did it**
Noticed that with PR https://github.com/sonic-net/sonic-swss/pull/1982 counters are enabled by default. It was not the case earlier.
User won't be able to see it with aclshow and it will waste valuable hardware counter resource.

**How I verified it**

**Details if related**
Master PR: https://github.com/sonic-net/sonic-swss/pull/2643


kevinskwang (approved):

yxieca:
@kevinskwang please use squash and merge going forward.

PR #2654, opened 2023-02-06
202205 enable fabric counters

kenneth-arista:
**What I did**
Reverted 34995f149a34125f9108473d092c95ed67482af0 to re-enable fabric counters support. Broadcom SAI 7.1.32 has been integrated into 202205 which fixes the crash problem.

**Why I did it**
To re-enable fabric counters.

**How I verified it**
Loaded image to confirm that syncd no longer crashes.


kenneth-arista:
Will create another PR appropriately against 202205 and not master

PR #2656, opened 2023-02-07
multi-nexthop routes draft

Ndancejic:

DRAFT of multi-nexthop routes PR


orchagent/nexthopgroupkey.h
```diff
@@ -23,27 +23,13 @@ class NextHopGroupKey
...
-            m_srv6_nexthops = false;
-            auto nhv = tokenize(nexthops, NHG_DELIMITER);
-            for (const auto &nh_str : nhv)
-            {
-                auto nh = NextHopKey(nh_str, overlay_nh, srv6_nh);
-                m_nexthops.insert(nh);
-            }
-        }
-        else if (srv6_nh)
+        m_overlay_nexthops = overlay_nh;
```
prsunny, line 26: Please revert this change.

prsunny:
Please provide the Description with all details as per the template

Ndancejic:
> Please provide the Description with all details as per the template

Ah my bad, that usually auto generates with my commit message but I had a draft PR open before and it didn't update. I'll fix that

orchagent/routeorch.cpp
```diff
@@ -1643,15 +1681,8 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
...
-    if (nextHops.is_overlay_nexthop())
-    {
-        overlay_nh = true;
-    }
-
-    if (nextHops.is_srv6_nexthop())
-    {
-        srv6_nh = true;
-    }
+    overlay_nh = nextHops.is_overlay_nexthop();
```
prsunny, line 1684: Please revert

orchagent/routeorch.cpp
```diff
@@ -2202,14 +2233,23 @@ bool RouteOrch::addRoutePost(const RouteBulkContext& ctx, const NextHopGroupKey
...
+        for (auto nh = nexthop_list.begin(); nh != nexthop_list.end(); nh++)
         {
-            addNextHopRoute(nexthop, r_key);
+            if (!nh->ip_address.isZero())
+            {
+                addNextHopRoute(*nh, r_key);
+            }
         }
+        // update routes to reflect mux state
+        mux_orch->updateRoute(ipPrefix);
```
prsunny, line 2252: This should be if `MuxNexthops`

orchagent/muxorch.cpp
```diff
@@ -973,6 +973,132 @@ sai_object_id_t MuxOrch::getNextHopTunnelId(std::string tunnelKey, IpAddress& ip
...
+
+        SWSS_LOG_INFO("Removing multi-nexthop route: %s", pfx.getIp().to_string().c_str());
+        status = sai_route_api->remove_route_entry(&route_entry);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("Failed to remove nexthop route %s rv:%d",
+                    pfx.getIp().to_string().c_str(), status);
+        }
+
+        if (gRouteOrch->hasNextHopGroup(nhg_key))
```
prsunny, line 1032: Dont remove RouteOrch cache

orchagent/muxorch.cpp
```diff
@@ -973,6 +973,132 @@ sai_object_id_t MuxOrch::getNextHopTunnelId(std::string tunnelKey, IpAddress& ip
...
+        {
+            if (findMuxCableInSubnet(nh->ip_address)->isActive())
+            {
+                // If we find an active nexthop neighbor, program it to hardware.
+                attr.value.oid = gNeighOrch->getNextHopId(*nh);
+                attrs.push_back(attr);
+                SWSS_LOG_NOTICE("Updating route %s with nexthop: %s",
+                    pfx.getIp().to_string().c_str(), nh->to_string().c_str());
+
+                status = sai_route_api->create_route_entry(&route_entry, (uint32_t)attrs.size(), attrs.data());
```
prsunny, line 1053: Can you do set/unset nexthop member

orchagent/routeorch.cpp
```diff
@@ -1580,6 +1581,43 @@ bool RouteOrch::updateNextHopRoutes(const NextHopKey& nextHop, uint32_t& numRout
...
+    return nh_list;
+}
+
+/**
+ * @brief returns a route prefix associated with nexthopkey
+ * @param routekey_set empty set of routekeys to populate
+ * @param nexthop_key nexthop key to lookup
+ * @return true if found, false if not found.
+ */
+bool RouteOrch::getRoutesForNexthop(std::set<RouteKey>& routekey_set, const NextHopKey& nexthop_key)
```
prsunny, line 1609: remove `_set `from the param name

prsunny, line 1609: use name as nexthopKey

orchagent/muxorch.cpp
```diff
@@ -983,6 +1075,91 @@ sai_object_id_t MuxOrch::getNextHopTunnelId(std::string tunnelKey, IpAddress& ip
...
+        bool active_found = false;
+
+        /* get nexthops from nexthop group */
+        nextHops = nhg_key.getNextHops();
+
+        auto it = mux_multi_nh_route_tb.find(pfx);
+        if (it != mux_multi_nh_route_tb.end())
+        {
+            MuxCable* cable = findMuxCableInSubnet(it->second.ip_address);
+            if (cable == nullptr || cable->isActive())
```
theasianpianist, line 1112: If I'm reading this correctly, if our route nexthop is not one of the configured mux cable IPs, we take no action. This seems a bit optimistic to me, unless we can 100% guarantee that the multi-mux-nexthop scenario only occurs with nexthop IPs that are configured as mux cable IPs?

orchagent/muxorch.cpp
```diff
@@ -983,6 +1075,91 @@ sai_object_id_t MuxOrch::getNextHopTunnelId(std::string tunnelKey, IpAddress& ip
...
+        }
+
+        SWSS_LOG_NOTICE("Updating route %s pointing to Mux nexthops %s",
+                    pfx.to_string().c_str(), nhg_key.to_string().c_str());
+
+        for (auto it = nextHops.begin(); it != nextHops.end(); it++)
+        {
+            NextHopKey nexthop = *it;
+            MuxCable* cable = findMuxCableInSubnet(nexthop.ip_address);
+            if (cable == nullptr || (cable->isActive() && !active_found))
```
theasianpianist, line 1127: same as above comment - if the nexthop IP isn't a mux cable IP, it seems safer to assume it's standby

tests/test_mux.py
```diff
@@ -450,92 +504,214 @@ def create_and_test_route(self, appdb, asicdb, dvs, dvs_route):
...
+
+                # move neighbor 2 back
+                self.add_neighbor(dvs, neighbors[0], macs[i])
+
+                self.del_route(dvs, route)
+
+
+            # toggle mux states to check setState actions
+            print("Testing toggling state")
+            starting_states = [(ACTIVE, ACTIVE), (ACTIVE, STANDBY), (STANDBY, ACTIVE), (STANDBY, STANDBY)]
```
theasianpianist, line 565: Does this need to be re-defined?

tests/test_mux.py
```diff
@@ -450,92 +504,214 @@ def create_and_test_route(self, appdb, asicdb, dvs, dvs_route):
...
+                    self.check_route_nexthop(dvs_route, asicdb, route, neighbors[0])
+                elif start[1] == "active":
+                    self.check_route_nexthop(dvs_route, asicdb, route, neighbors[1])
+                else:
+                    self.check_route_nexthop(dvs_route, asicdb, route, tunnel_nh_id, True)
+
+                # move neighbor 2 back
+                self.add_neighbor(dvs, neighbors[0], macs[i])
+
+                self.del_route(dvs, route)
```
theasianpianist, line 560: Would prefer if this test case were parameterized so that we could use a fixture to do setup/teardown in case something goes wrong in the middle of the test. Depending on urgency, we can merge as-is and discuss this change in the future.

Ndancejic: yeah I agree with this. currently I think the try,finally block will teardown resources that I'm using in this test, but I can definitely rework it in that way.

orchagent/muxorch.cpp
Ndancejic: From my understanding from discussions with Prince we currently handle neighbors on the Vlan that aren't tied to a muxcable as active. maybe we can have a discussion on this?

orchagent/muxorch.cpp
Ndancejic: same as above

orchagent/routeorch.cpp
```diff
@@ -2234,6 +2284,9 @@ bool RouteOrch::addRoutePost(const RouteBulkContext& ctx, const NextHopGroupKey
                 ipPrefix.to_string().c_str(), nextHops.to_string().c_str());
     }
 
+    m_syncdRoutes[vrf_id][ipPrefix] = RouteNhg(nextHops, ctx.nhg_index);
```
prsunny, line 2287: why this change?

orchagent/routeorch.h
```diff
@@ -186,7 +186,9 @@ class RouteOrch : public Orch, public Subject
     RouteOrch(DBConnector *db, vector<table_name_with_pri_t> &tableNames, SwitchOrch *switchOrch, NeighOrch *neighOrch, IntfsOrch *intfsOrch, VRFOrch *vrfOrch, FgNhgOrch *fgNhgOrch, Srv6Orch *srv6Orch);
 
     bool hasNextHopGroup(const NextHopGroupKey&) const;
+    bool inNextHopGroup(const NextHopKey&, NextHopGroupKey& nhgKey);
     sai_object_id_t getNextHopGroupId(const NextHopGroupKey&);
+    NextHopGroupEntry getNextHopGroupEntry(const NextHopGroupKey&);
```
prsunny, line 191: I dont see this function being used. Please remove if not applicable

orchagent/routeorch.cpp
```diff
@@ -1577,6 +1596,19 @@ bool RouteOrch::updateNextHopRoutes(const NextHopKey& nextHop, uint32_t& numRout
         return true;
     }
 
+    /* Check if nexthop is mux nexthop */
+    MuxOrch* mux_orch = gDirectory.get<MuxOrch*>();
+    NextHopGroupKey nhg_key;
+    if (inNextHopGroup(nextHop, nhg_key) && mux_orch->isMuxNexthops(nhg_key))
```
prsunny, line 1602: What happens if a nexthop is part of NHG but also another route points to this single nexthop?
For e.g like NH1 below:

R1 -> NHG1 (NH1, NH2)
R2 -> NH1

orchagent/routeorch.cpp
```diff
@@ -1577,6 +1596,19 @@ bool RouteOrch::updateNextHopRoutes(const NextHopKey& nextHop, uint32_t& numRout
         return true;
     }
 
+    /* Check if nexthop is mux nexthop */
```
prsunny, line 1599: Please add some more details on the check and assumption.

orchagent/muxorch.h
```diff
@@ -196,13 +202,16 @@ class MuxOrch : public Orch2, public Observer, public Subject
...
+    bool containsNextHop(const NextHopKey&);
+    bool isMuxNexthops(const NextHopGroupKey&);
     string getNexthopMuxName(NextHopKey);
     sai_object_id_t getNextHopId(const NextHopKey&);
 
     sai_object_id_t createNextHopTunnel(std::string tunnelKey, IpAddress& ipAddr);
     bool removeNextHopTunnel(std::string tunnelKey, IpAddress& ipAddr);
     sai_object_id_t getNextHopTunnelId(std::string tunnelKey, IpAddress& ipAddr);
 
+    void updateRoute(const IpPrefix &pfx, bool remove);
```
prsunny, line 214: general convention is `bool add=True`, Use 'false' as non-default case during remove

orchagent/muxorch.cpp
```diff
@@ -983,6 +1075,91 @@ sai_object_id_t MuxOrch::getNextHopTunnelId(std::string tunnelKey, IpAddress& ip
...
+ * @param remove bool only true when route is getting removed
+ */
+void MuxOrch::updateRoute(const IpPrefix &pfx, bool remove)
+{
+    NextHopGroupKey nhg_key;
+    NextHopGroupEntry nhg_entry;
+
+    if (remove)
+    {
+        mux_multi_nh_route_tb.erase(pfx);
```
prsunny, line 1090: Does the route gets removed from SAI via routeorch?

orchagent/muxorch.cpp
prsunny: i think `findMuxCableInSubnet` returns if its directly connected server IP but not all learned neighbors (and is in config_db). But i think we have to handle it if the nexthop belongs to any mux cable.

orchagent/muxorch.cpp
```diff
@@ -566,6 +597,29 @@ void MuxCable::updateNeighbor(NextHopKey nh, bool add)
     else if (mux_name_ == mux_orch_->getNexthopMuxName(nh))
     {
         mux_orch_->removeNexthop(nh);
+        nbr_handler_->removeNeighborRoutes(nh);
```
prsunny, line 600: This part is not clear. If a nh is removed for some reason, why remove the Route itself?

StormLiangMS:
@Ndancejic cherry pick conflict, to 202211, could you raise separate PR for 202211 branch?

@prsunny  for vis.

PR #2661, opened 2023-02-09
[202205][sai_failure_dump]Invoking dump during SAI failure 

dgsudharsan:
Cherry-pick of https://github.com/sonic-net/sonic-swss/pull/2644

**What I did**
Added logic to invoke SAI failure dump during any SAI programming failure before invoking abort.

**Why I did it**
To collect necessary dumps in problem state in syncd before abort is called and all processes restarts

**How I verified it**
Manual verification. Added UT to cover abort scenario as well.

**Details if related**


prsunny (approved):

prsunny:
@yxieca for viz as this is a new addition to orchagent behavior for 202205

PR #2664, opened 2023-02-10
[202211][voq][chassis] Remove created ports from the default vlan.

arista-nwolfe:
This is a cherry pick of https://github.com/sonic-net/sonic-swss/pull/2607 for 202211


arista-nwolfe:
@StormLiangMS would you be able to take a look at the build failure?
`No builds currently exist in the pipeline definition supplied.`
I'm not sure how to get past this failure, this is a straighforward cherry-pick of the original patch.

PR #2669, opened 2023-02-16
Do not allow to add port to .1Q bridge while router port deletion is not completed 

liorghub:
**What I did**
Do not allow to add port to .1Q bridge while router port deletion is not completed.

**Why I did it**
Adding router port to .1Q bridge is not allowed and will cause to errors in the driver.

**How I verified it**

Create PortChannel102 and set it as router port.
Make sure PortChannel102 has active neighbors.
Now run the following commands using a shell script (not one by one).
1. config interface ip remove PortChannel102 10.0.0.0/31
2. config vlan add 1000
3. config vlan member add 1000 PortChannel102

Follow the logs and verify addBridgePort is not perfomed until removeRouterIntfs is completed.
Unitest added as well.

This PR fixes issue https://github.com/sonic-net/sonic-buildimage/issues/13714

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -4881,7 +4881,7 @@ bool PortsOrch::addBridgePort(Port &port)
 
     if (port.m_rif_id != 0)
     {
-        SWSS_LOG_NOTICE("Interface %s is a router port", port.m_alias.c_str());
+        SWSS_LOG_NOTICE("Cannot create bridge port, interface %s is a router port", port.m_alias.c_str());
         return false;
     }
 
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -4879,6 +4879,12 @@ bool PortsOrch::addBridgePort(Port &port)
         return true;
     }
 
+    if (port.m_rif_id != 0)
+    {
+        SWSS_LOG_NOTICE("Interface %s is still a router port", port.m_alias.c_str());
+        return false;
+    }
+
     sai_attribute_t attr;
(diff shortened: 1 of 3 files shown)


stepanblyschak (dismissed):

orchagent/portsorch.cpp
```diff
@@ -4879,6 +4879,12 @@ bool PortsOrch::addBridgePort(Port &port)
         return true;
     }
 
+    if (port.m_rif_id != 0)
+    {
+        SWSS_LOG_NOTICE("Interface %s is a router port", port.m_alias.c_str());
```
dgsudharsan, line 4884: Can you please add more context in the log message and change it to warning since we are returning here in a failure scenario?
"Cannot create bridge port. Interface %s is a router port"

dgsudharsan (changes requested):

orchagent/portsorch.cpp
liorghub: Done.

stepanblyschak (approved):

orchagent/portsorch.cpp
dgsudharsan: @liorghub  I recommend changing it to "WARN" log instead of notice since here are returning here in a failure scenario.

liorghub: @dgsudharsan this is actually a retry mechanism (not failure).
When we print this message we also return false in the next line, DB change is not consumed and we will try again till success.

dgsudharsan (approved):

liat-grozovik:
@prsunny any objection to merge?

PR #2670, opened 2023-02-18
[EVPN]Handling error scenarios during route programming and IMR add

dgsudharsan:
**What I did**
Handling error scenarios for the following cases
1) When route is propagated through FRR but SONiC is yet to process VRF - VNI mapping. The route needs to be skipped
2) When VNI is configured  as L3 in SONiC but misconfigured as L2 in FRR leading to remote IMR add, avoid adding the IMR since the local VLAN VNI map doesn't exist according to the recent updated design.

**Why I did it**
These scenarios will lead to issues and result in SAI failures which needs to be avoided

**How I verified it**
Added UT to verify.

**Details if related**

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -665,6 +665,8 @@ void RouteOrch::doTask(Consumer& consumer)
                 NextHopGroupKey& nhg = ctx.nhg;
                 vector<string> srv6_segv;
                 vector<string> srv6_src;
+                bool l3Vni = true;
+                uint32_t vni = 0;
 
                 /* Check if the next hop group is owned by the NhgOrch. */
                 if (nhg_index.empty())
@@ -694,6 +696,23 @@ void RouteOrch::doTask(Consumer& consumer)
                     {
                         SWSS_LOG_NOTICE("Route %s: resize ipv to match alsv, %zd -> %zd.", key.c_str(), ipv.size(), alsv.size());
                         ipv.resize(alsv.size());
+                    }
+
+                    for (auto &vni_str: vni_labelv)
+                    {
+                        vni = static_cast<uint32_t>(std::stoul(vni_str));
+                        if (!m_vrfOrch->isL3VniVlan(vni))
+                        {
+                            SWSS_LOG_WARN("Route %s is received on non L3 VNI %s", key.c_str(), vni_str.c_str());
+                            l3Vni = false;
+                            break;
+                        }
+                    }
+
+                    if (!l3Vni)
+                    {
+                        it++;
+                        continue;
                     }
 
                     /* Set the empty ip(s) to zero
--- a/orchagent/vxlanorch.cpp
+++ b/orchagent/vxlanorch.cpp
@@ -2376,6 +2376,13 @@ bool EvpnRemoteVnip2pOrch::addOperation(const Request& request)
         return false;
     }
 
+    VRFOrch* vrf_orch = gDirectory.get<VRFOrch*>();
+    if (vrf_orch->isL3VniVlan(vni_id))
+    {
+        SWSS_LOG_WARN("Ignoring remote VNI add for L3 VNI:%d, remote:%s", vni_id, remote_vtep.c_str());
+        return false;
+    }
+
     if (tunnel_orch->getTunnelPort(remote_vtep,tunnelPort))
     {
         SWSS_LOG_INFO("Vxlan tunnelPort exists: %s", remote_vtep.c_str());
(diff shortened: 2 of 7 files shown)


volodymyrsamotiy (approved):

dgsudharsan:
@srj102 Can you please review?

liat-grozovik:
@prsunny can you please help to review/approve?

srj102:
@tapashdas

orchagent/vxlanorch.cpp
```diff
@@ -2376,6 +2376,13 @@ bool EvpnRemoteVnip2pOrch::addOperation(const Request& request)
         return false;
     }
 
+    VRFOrch* vrf_orch = gDirectory.get<VRFOrch*>();
+    if (vrf_orch->isL3VniVlan(vni_id))
```
srj102, line 2380: 1. wouldn't this check is applicable for fdborch also ? 
2. If the vni became an L3VNI after IMR routes were installed then OA will remove VLAN-VNI mappings and add the VRF-VNI mapping, leaving us with a scenario where the SAI has MAC/IMR routes with no corresponding vlan vni mappings. 
3. What happens without this fix ? Are there SAI errors we see ?

orchagent/vxlanorch.cpp
```diff
@@ -2376,6 +2376,13 @@ bool EvpnRemoteVnip2pOrch::addOperation(const Request& request)
...
+    {
+        SWSS_LOG_WARN("Ignoring remote VNI add for L3 VNI:%d, remote:%s", vni_id, remote_vtep.c_str());
+        return false;
```
srj102, line 2383: Should this be return true instead of return false ? 

The log says "Ignore" but in reality it is being deferred.
Also a large number of IMR routes being held in the m_tosync queue and being processed in a continous loop will load the OA unnecessarily.

orchagent/routeorch.cpp
```diff
@@ -696,6 +698,23 @@ void RouteOrch::doTask(Consumer& consumer)
...
+                        {
+                            SWSS_LOG_WARN("Route %s is received on non L3 VNI %s", key.c_str(), vni_str.c_str());
+                            l3Vni = false;
+                            break;
+                        }
+                    }
+
+                    if (!l3Vni)
+                    {
+                        it++;
```
srj102, line 714: Similar comment as for IMR handling. Not Removing this from the m_tosyncqueue in a scale scenario and in a misconfiguration case will unnecessarily load the OA.

orchagent/routeorch.cpp
```diff
@@ -696,6 +698,23 @@ void RouteOrch::doTask(Consumer& consumer)
                         ipv.resize(alsv.size());
                     }
 
+                    for (auto &vni_str: vni_labelv)
+                    {
+                        vni = static_cast<uint32_t>(std::stoul(vni_str));
+                        if (!m_vrfOrch->isL3VniVlan(vni))
```
srj102, line 704: If the vrf-vni configuration is removed from OA and not from FRR after routes are installed then the routes still remain in the HW. 
Is this scenario handled ?

srj102 (changes requested):

orchagent/vxlanorch.cpp
dgsudharsan: I can correct the log. I prefer not to ignore but defer. If we ignore we might miss those remote. Since this is a corner error scenario I don't think it will overwhelm orchagent during regular processing

orchagent/vxlanorch.cpp
dgsudharsan: 1. In case of FDB orch we have check to ensure Vlan membership which will prevent FDB getting added since the IMR is prevented
2. Yes that's true. But as per the current behavior we will fail at removing VLAN VNI mapping at SAI level due to references. Do you think we should add a check at orchagent. I am not sure if this check is valid for other platforms and so I didn't prefer to add
3. Without the fix SAI returns error and orchagent crashes.

orchagent/routeorch.cpp
dgsudharsan: In our case the VNI removal leads to error logs in SAI since it has references. This scenario again depends on references to maintain which I believe we don't do it. IMO I am fixing the scenarios which are easier to fix with the current implementation and then document the ones which are harder to fix.

dgsudharsan: Currently SAI error is not creating orchagent crash since we are not handling status for this call. So we will get syslog error. For now I consider this a lower priority and document this.

srj102 (approved):

prsunny (approved):

PR #2673, opened 2023-02-21
[FDB]Fixing FDB consolidated flush for Remote MACs

dgsudharsan:
**What I did**
Fixed FDB flush for remote MACS in EVPN scenario. Even though FDB MAC is received as dynamic, it would be installed as static in SAI and thus during flush it would not be removed. However SONiC's db when processing the FLUSH event removes all the dynamic macs and thus clearing VXLAN remote mac from cache.

Hence during flush modified logic to check by sai_fdb_type rather than SONiC's type

**Why I did it**
To avoid Remote VxLAN FDB getting cleared wrongly in flush.

**How I verified it**
Added UT to verify

**Details if related**

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -213,9 +213,6 @@ void FdbOrch::handleSyncdFlushNotif(const sai_object_id_t& bv_id,
     // Consolidated flush will have a zero mac
     MacAddress flush_mac("00:00:00:00:00:00");
 
-    /* TODO: Read the SAI_FDB_FLUSH_ATTR_ENTRY_TYPE attr from the flush notif
-    and clear the entries accordingly, currently only non-static entries are flushed
-    */
     if (bridge_port_id == SAI_NULL_OBJECT_ID && bv_id == SAI_NULL_OBJECT_ID)
     {
         for (auto itr = m_entries.begin(); itr != m_entries.end();)
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -109,6 +109,7 @@ bool FdbOrch::storeFdbEntryState(const FdbUpdate& update)
 
         fdbdata.bridge_port_id = update.port.m_bridge_port_id;
         fdbdata.type = update.type;
+        fdbdata.sai_fdb_type = update.sai_fdb_type;
         fdbdata.origin = FDB_ORIGIN_LEARN;
         fdbdata.remote_ip = "";
         fdbdata.esi = "";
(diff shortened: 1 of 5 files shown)


keboliu (approved):

liat-grozovik:
@marian-pritsak @prsunny could you please help to approve?

orchagent/fdborch.cpp
```diff
@@ -206,20 +207,19 @@ Handles the SAI_FDB_EVENT_FLUSHED notification recieved from syncd
...
-    /* TODO: Read the SAI_FDB_FLUSH_ATTR_ENTRY_TYPE attr from the flush notif
-    and clear the entries accordingly, currently only non-static entries are flushed
-    */
     if (bridge_port_id == SAI_NULL_OBJECT_ID && bv_id == SAI_NULL_OBJECT_ID)
     {
         for (auto itr = m_entries.begin(); itr != m_entries.end();)
         {
             auto curr = itr++;
-            if (curr->second.type != "static" && (curr->first.mac == mac || mac == flush_mac) && curr->second.is_flush_pending)
+            if (curr->second.sai_fdb_type == sai_fdb_type &&
```
prsunny, line 221: why dont we keep both the state in sync?

gechiang:
@Praveen-Brcm can you help review this?
In general I have seen this MAC Flushing handling code got changed so many times I am very concerned with the design itself as it seems to be very buggy in general...  Any chance this handling can be redesigned so it is less buggy?  Have we covered all possible scenarios yet??

orchagent/fdborch.cpp
dgsudharsan: @prsunny Currently EVPN and McLAG treat the remote dynamic entries differently than the local. To prohibit local ageing they install the entries as static with mac move enabled in the hardware although they are advertised as dynamic. I don't think we can change the application type to dynamic since it conflicts with the advertised type and the CLIs display the type as dynamic. Hence we need to maintain two different states

dgsudharsan:
> @Praveen-Brcm can you help review this? In general I have seen this MAC Flushing handling code got changed so many times I am very concerned with the design itself as it seems to be very buggy in general... Any chance this handling can be redesigned so it is less buggy? Have we covered all possible scenarios yet??

I believe McLAG and EVPN have different different scenarios where they change type while programming SAI from what is advertised. On top of that we also need to support consolidated flush notifications which leads to lot of cases to handle. For now I have added UT to ensure the cases are covered but I think its better to redesign the entire FDBOrch in the future. There are lot of checks not only in flush path but also in add and mac learn path.

vivekrnv (approved):

liat-grozovik:
@prsunny can you please help to review/approve?

gechiang (approved):

yxieca:
@dgsudharsan please create separate PR for 202205 branch

yxieca:
@dgsudharsan please create separate PR for 202205 branch.

dgsudharsan:
> @dgsudharsan please create separate PR for 202205 branch.

Hi @yxieca Here is the cherry-pick https://github.com/sonic-net/sonic-swss/pull/2725

PR #2618, opened 2023-01-12
[bufferorch] : Support for buffer profiles for VoQ on chassis (#2465)

vmittal-msft:
* QoS changes for VoQ on chassis

**What I did**
Porting https://github.com/sonic-net/sonic-swss/pull/2465 from master branch

**Why I did it**
Due to merge conflict

**How I verified it**

**Details if related**

PR #2685, opened 2023-03-02
[Ci] Support to build sonic-swss-common

xumia:
**What I did**
[Ci] Support to build sonic-swss-common

**Why I did it**
It is to improve the chance to build successful rate on sonic-swss.
It removes the dependency of the build artifacts of sonic-swss-common in sonic-swss, always uses the latest code of the sonic-swss-common to build it.

**How I verified it**

**Details if related**

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -82,7 +82,8 @@ jobs:
       ./autogen.sh
       dpkg-buildpackage -us -uc -b -j$(nproc)
       sudo dpkg -i ../libswsscommon*.deb
-    workingDirectory: $(Build.SourcesDirectory)/sonic-swss-common 
+      sudo dpkg -i **/libswsscommon_*.deb
+    workingDirectory: $(System.DefaultWorkingDirectory)/sonic-swss-common
     displayName: "Build and isntall sonic-swss-common"
   - task: DownloadPipelineArtifact@2
     inputs:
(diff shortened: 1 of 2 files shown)


vmittal-msft:
@theasianpianist we need this urgently for https://github.com/sonic-net/sonic-swss/pull/2618 to go in. #2618 is needed for pilot testing on chassis.

theasianpianist:
Why do we want to start building swss-common instead of downloading artifacts?

vmittal-msft:
Let me copy you on the email thread. sonic-swss compilation is failing due to sonic-swss-common package doesn't seem to be latest enough for this to compile. So why do we have such dependency ? and not directly compile to avoid such issues.

xumia:
> Why do we want to start building swss-common instead of downloading artifacts?

The build sonic-swss-common is failed in test stage, it blocks the sonic-swss to use the lastest code, and the sonic-swss cannot be built successfully.

xumia:
> > Why do we want to start building swss-common instead of downloading artifacts?
> 
> The build sonic-swss-common is failed in test stage, it blocks the sonic-swss to use the lastest code, and the sonic-swss cannot be built successfully. It reduces the dependency to wait for the result of the another pipeline.

vmittal-msft:
As per @prsunny, this need to be included for 202205.
https://github.com/sonic-net/sonic-swss/pull/2648

prsunny (approved):

prsunny:
@theasianpianist , merging to unblock PR builds. Lets address any concerns as future commits.

PR #2691, opened 2023-03-07
[flowcounterrouter] Fix the route remove flow for non-yet bound prefixes

vivekrnv:
**What I did**

Update the isRouteAlreadyBound to consider both the maps for verifying whether the prefix is deleted or not.

**Why I did it**

- When a route is deleted and the flow counter for that route is not yet bound i.e. still in mPendingAddToFlexCntr (in rare cases), the orchagent will not delete the flow counters and they will remain even after the corresponding route is deleted

- Happens because isRouteAlreadyBound method presumes all the route counters for a route pattern would either be in mPendingAddToFlexCntr or mBoundRouteCounters. But in some cases they might exist in both the maps

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/flex_counter/flowcounterrouteorch.cpp
+++ b/orchagent/flex_counter/flowcounterrouteorch.cpp
@@ -602,18 +602,25 @@ bool FlowCounterRouteOrch::isRouteAlreadyBound(const RoutePattern &route_pattern
 {
     SWSS_LOG_ENTER();
 
-    auto iter = mBoundRouteCounters.find(route_pattern);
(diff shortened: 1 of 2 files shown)


Junchao-Mellanox (approved):

liat-grozovik:
@vivekrnv  on which branches this fix should be backport?

vivekrnv:
> @vivekrnv on which branches this fix should be backport?

Request for 202211 and 202205

prsunny (approved):

PR #2697, opened 2023-03-08
[portsorch]: Set default hostif TX queue

prabhataravind:
* To avoid data and control packets sharing the same TX queue and possibly leading to control packet losses, pin control traffic to queue 7 by default.

**What I did**
Set default queue for control packets to 7.
**Why I did it**
To avoid control packets like BGP keepalives from being lost when link utilization becomes 100%
**How I verified it**
Tested on multiple platorms with ICMP packets to make sure they use TX queue 0
**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2582,6 +2582,7 @@ bool PortsOrch::createVlanHostIntf(Port& vl, string hostif_name)
 
     vector<sai_attribute_t> attrs;
     sai_attribute_t attr;
+    sai_attr_capability_t capability;
 
     attr.id = SAI_HOSTIF_ATTR_TYPE;
     attr.value.s32 = SAI_HOSTIF_TYPE_NETDEV;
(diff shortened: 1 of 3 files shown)


prsunny:
@prabhataravind , please create a testgap in sonic-mgmt repo and link here.

orchagent/portsorch.cpp
```diff
@@ -2599,6 +2600,10 @@ bool PortsOrch::createVlanHostIntf(Port& vl, string hostif_name)
     attr.value.chardata[SAI_HOSTIF_NAME_SIZE - 1] = '\0';
     attrs.push_back(attr);
 
+    attr.id = SAI_HOSTIF_ATTR_QUEUE;
+    attr.value.u32 = DEFAULT_HOSTIF_TX_QUEUE;
```
dgsudharsan, line 2604: Should we have a query to check if SAI supports this and only then set?

dgsudharsan (changes requested):
Please add UT to the changes

orchagent/portsorch.cpp
stephenxs: It's possible that the attribute `SAI_HOSTIF_ATTR_QUEUE` is not supported on all platforms.
So, I think it's a must to query SAI capability.

keboliu: I second to this comment. Not all vendors support his attributes, w/o query will fail on some platforms and cause orchagent exit.

orchagent/portsorch.cpp
```diff
@@ -2600,9 +2601,32 @@ bool PortsOrch::createVlanHostIntf(Port& vl, string hostif_name)
     attr.value.chardata[SAI_HOSTIF_NAME_SIZE - 1] = '\0';
     attrs.push_back(attr);
 
-    attr.id = SAI_HOSTIF_ATTR_QUEUE;
-    attr.value.u32 = DEFAULT_HOSTIF_TX_QUEUE;
-    attrs.push_back(attr);
+    bool set_hostif_tx_queue = false;
+    if (sai_query_attribute_capability(gSwitchId, SAI_OBJECT_TYPE_HOSTIF,
```
prsunny, line 2605: please use `querycapability `api from switchorch.

prabhataravind: SwitchOrch::querySwitchCapability is checking for capability.set_implemented, here we are interested in checking for capability.create_implemented as this is in the context of create operation for the hostif itself. Though it is likely that create is also supported when set is supported, it is probably better to err on the side of caution here. Do we want to enhance SwitchOrch::querySwitchCapability to check capability for all APIs?

prsunny: i think we can just use the existing query api. As you mentioned, if set returns true.. it can be considered as supported.

prabhataravind: okay

prsunny (approved):

prabhataravind:
@dgsudharsan , could you please review when you get a chance?

dgsudharsan (approved):

PR #2702, opened 2023-03-14
[202012][muxorch] handling multiple mux nexthops for route

Ndancejic:
What I did: added logic to handle when a route points to a nexthop group with mux neighbors. In this case, only one active neighbor, or the tunnel nexthop will be programmed to the ASIC.

Why I did it: having a route with multiple mux neighbors caused a data loop which lead to packet loss when different neighbors were in different states.

How I did it: added logic to update routes when a route is changed, a mux neighbor is changed, or there is a mux state change.

HLD: https://github.com/sonic-net/SONiC/pull/1256

cherry-pick of https://github.com/sonic-net/sonic-swss/pull/2656

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -163,6 +163,34 @@ static sai_status_t remove_route(IpPrefix &pfx)
     return status;
 }
 
+/**
+ * @brief sets the given route to point to the given nexthop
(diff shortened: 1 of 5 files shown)


prsunny (approved):

Aravind-Subbaroyan:
@prsunny  - could you please merge this to 202205?

PR #2703, opened 2023-03-14
[Dual-ToR] handle 'mux_tunnel_ingress_acl' attrib in order to change ACL configuration (drop on ingress/egress) on standby ToR (202205)

ayurkiv-nvda:
**What I did**
Use "mux_tunnel_ingress_acl" to set ACL rules on ingress/egress side depending on attribute value ("disabled/enabled").

**Why I did it**
We need to drop data-plane traffic and handle Control-plane traffic in the Dual-ToR scenario.
But we can't do it on Mellanox platform and process traffic on ingress.
To workaround it we can set ACL rules on egress ports, so will process control plane on ingress and drop Data-plane traffic that came from standby port on egress.

**How I verified it**
check "show mux status" on standby ToR - Mux status should be healthy.
check "show what-just-happened" on standby ToR - no ICMP drop expected on standby ports.

**Details if related**

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -49,13 +49,6 @@ extern sai_router_interface_api_t* sai_router_intfs_api;
 #define MUX_HW_STATE_UNKNOWN "unknown"
 #define MUX_HW_STATE_ERROR "error"
 
-
-static inline bool isIngressAcl()
-{
-    string platform = getenv("platform") ? getenv("platform") : "";
-    return platform != MLNX_PLATFORM_SUBSTRING;
-}
-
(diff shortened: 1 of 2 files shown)


prsunny:
Please have the branch name tagged first in the title

ayurkiv-nvda:
This PR requires YANG model for SYSTEM_DEFAULT.
Waiting for https://github.com/sonic-net/sonic-buildimage/pull/11117 to be merged

ayurkiv-nvda:
Related PRs
https://github.com/sonic-net/sonic-swss/pull/2646 - logic implementation in master
https://github.com/sonic-net/sonic-buildimage/pull/13547 - config_db implementation in master

PR #2710, opened 2023-03-22
[dash] Do not use an action drop with the inbound routing table.

oleksandrivantsiv:
**What I did**
Do not use an action drop with the inbound routing table.

**Why I did it**
Action "drop" is not supported by the inbound routing table. SAI_INBOUND_ROUTING_ENTRY_ACTION_DENY action is no longer available in SAI v1.12

**How I verified it**
Run sonic-mgmt DASH VNET to VNET test.

**Details if related**


prabhataravind (approved):

prsunny:
@oleksandrivantsiv , can you help check failure?

oleksandrivantsiv:
> @oleksandrivantsiv , can you help check failure?

@prsunny  it is an infrastructure issue. It fails on an attempt to install dependencies.

PR #2712, opened 2023-03-24
Fix orchagent missing request when logrotate happens.

mint570:
**What I did**
Fix orchagent request missing when logrotate happens.

**Why I did it**
The sairedis logrotate has been moved into the daemon main select loop. After it finishes, it should not continue, because that will drop the event that has been selected.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -754,7 +754,6 @@ void OrchDaemon::start()
             gSaiRedisLogRotate = false;
 
             logRotate();
-            continue;
         }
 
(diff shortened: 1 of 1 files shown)


mint570:
@prsunny can you take a look at this? A very small bug fix.

theasianpianist (approved):
good find, thanks!

prsunny:
@theasianpianist , can you check if it needs to be backported?

prsunny:
Introduced by https://github.com/sonic-net/sonic-swss/pull/2299. 

@Blueve for viz.

PR #2717, opened 2023-03-28
Update build pipeline to build for Bullseye

saiarcot895:
**What I did**

With the merge of sonic-net/sonic-buildimage#13294, docker-sonic-vs is now based on Bullseye. Switch to building for Bullseye to match that change.

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -71,7 +71,7 @@ jobs:
       sudo apt-get install -y libzmq5 libzmq3-dev
       sudo apt-get install -qq -y \
           libhiredis-dev \
-          swig3.0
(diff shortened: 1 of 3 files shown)


theasianpianist:
Depends on https://github.com/sonic-net/sonic-sairedis/pull/1225 being merged first

theasianpianist:
Looks like this URL needs to be changed to Debian 11 for gcov to work:
https://github.com/sonic-net/sonic-swss/blob/7a215d0fa5c42060695817da63c12276ce5ca4c1/.azure-pipelines/gcov.yml#L62

Pterosaur (approved):

PR #2725, opened 2023-04-06
[202205][FDB]Fixing FDB consolidated flush for Remote MACs

dgsudharsan:
**What I did**
Cherrypick of https://github.com/sonic-net/sonic-swss/pull/2673 to 202205
Fixed FDB flush for remote MACS in EVPN scenario. Even though FDB MAC is received as dynamic, it would be installed as static in SAI and thus during flush it would not be removed. However SONiC's db when processing the FLUSH event removes all the dynamic macs and thus clearing VXLAN remote mac from cache.

Hence during flush modified logic to check by sai_fdb_type rather than SONiC's type

**Why I did it**
To avoid Remote VxLAN FDB getting cleared wrongly in flush.

**How I verified it**
Added UT to verify

**Details if related**

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -109,6 +109,7 @@ bool FdbOrch::storeFdbEntryState(const FdbUpdate& update)
 
         fdbdata.bridge_port_id = update.port.m_bridge_port_id;
         fdbdata.type = update.type;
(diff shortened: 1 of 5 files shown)


dgsudharsan:
@yxieca @prsunny Can you please help merging this?

prsunny (approved):

PR #2735, opened 2023-04-14
Combine v4 and v6 L3 ACL rules on optimized platforms #1267

rck-innovium:
**What I did**
Support a new ACL table type called L3V4V6.
This table supports both v4 and v6 Match types.
Add unit tests for this new ACL table type.
HLD: https://github.com/sonic-net/SONiC/pull/1267

Signed-off-by: Ravi(Marvell) rck@innovium.com

**Why I did it**
To optimise ACL TCAM usage in capable platforms.

**How I verified it**
Add unit tests for this new ACL table type.

```
sudo pytest -s test_acl_l3v4v6.py
collected 2 items

test_acl_l3v4v6.py ..

======================= 2 passed, 1 warning in 60.88s (0:01:00) 

```

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -3109,7 +3109,7 @@ void AclOrch::init(vector<TableConnector>& connectors, PortsOrch *portOrch, Mirr
             m_mirrorTableCapabilities[TABLE_TYPE_MIRROR] ? "yes" : "no");
     SWSS_LOG_NOTICE("    TABLE_TYPE_MIRRORV6: %s",
             m_mirrorTableCapabilities[TABLE_TYPE_MIRRORV6] ? "yes" : "no");
-    SWSS_LOG_NOTICE("    TABLE_TYPE_L3v4V6: Ingress [%s], Egress [%s]",
+    SWSS_LOG_NOTICE("    TABLE_TYPE_L3V4V6: Ingress [%s], Egress [%s]",
             m_L3V4V6Capability[ACL_STAGE_INGRESS] ? "yes" : "no",
             m_L3V4V6Capability[ACL_STAGE_EGRESS] ? "yes" : "no");
 
@@ -3248,6 +3248,7 @@ void AclOrch::initDefaultTableTypes()
             .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_L4_SRC_PORT))
             .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_L4_DST_PORT))
             .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_TCP_FLAGS))
+            .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_IN_PORTS))
             .build()
     );
 
@@ -4576,6 +4577,8 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
             }
             bool bHasTCPFlag = false;
             bool bHasIPProtocol = false;
+            bool bHasIPV4 = false;
+            bool bHasIPV6 = false;
             for (const auto& itr : kfvFieldsValues(t))
             {
                 string attr_name = to_upper(fvField(itr));
@@ -4585,6 +4588,14 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
                 if (attr_name == MATCH_TCP_FLAGS)
                 {
                     bHasTCPFlag = true;
+                }
+                if (attr_name == MATCH_SRC_IP || attr_name == MATCH_DST_IP)
+                {
+                    bHasIPV4 = true;
+                }
+                if (attr_name == MATCH_SRC_IPV6 || attr_name == MATCH_DST_IPV6)
+                {
+                    bHasIPV6 = true;
                 }
                 if (attr_name == MATCH_IP_PROTOCOL || attr_name == MATCH_NEXT_HEADER)
                 {
@@ -4633,6 +4644,15 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
                     SWSS_LOG_ERROR("Failed to add attribute '%s : %s'", attr_name.c_str(), attr_value.c_str());
                 }
             }
+
+            if (bHasIPV4 && bHasIPV6)
+	    {
+		    if (type == TABLE_TYPE_L3V4V6)
+		    {
+			    SWSS_LOG_ERROR("Rule '%s' is invalid since it has both v4 and v6 matchfields.", rule_id.c_str());
+			    bAllAttributesOk = false;
+		    }
+	    }
 
             // validate and create ACL rule
             if (bAllAttributesOk && newRule->validate())
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -34,6 +34,7 @@ extern CrmOrch *gCrmOrch;
 
 #define STATE_DB_ACL_ACTION_FIELD_IS_ACTION_LIST_MANDATORY "is_action_list_mandatory"
 #define STATE_DB_ACL_ACTION_FIELD_ACTION_LIST              "action_list"
+#define STATE_DB_ACL_L3V4V6_SUPPORTED                      "supported_L3V4V6"
 #define COUNTERS_ACL_COUNTER_RULE_MAP "ACL_COUNTER_RULE_MAP"
 
 #define ACL_COUNTER_DEFAULT_POLLING_INTERVAL_MS 10000 // ms
@@ -185,6 +186,26 @@ static acl_table_action_list_lookup_t defaultAclActionList =
     {
         // L3V6
         TABLE_TYPE_L3V6,
+        {
+            {
+                ACL_STAGE_INGRESS,
+                {
+                    SAI_ACL_ACTION_TYPE_PACKET_ACTION,
+                    SAI_ACL_ACTION_TYPE_REDIRECT
+                }
+            },
+            {
+                ACL_STAGE_EGRESS,
+                {
+                    SAI_ACL_ACTION_TYPE_PACKET_ACTION,
+                    SAI_ACL_ACTION_TYPE_REDIRECT
+                }
+            }
+        }
+    },
+    {
+        // L3V4V6
+        TABLE_TYPE_L3V4V6,
         {
             {
                 ACL_STAGE_INGRESS,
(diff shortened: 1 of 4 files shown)


orchagent/aclorch.cpp
```diff
@@ -3168,6 +3227,30 @@ void AclOrch::initDefaultTableTypes()
...
+            .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_SRC_IPV6))
+            .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_DST_IPV6))
+            .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_ICMPV6_CODE))
+            .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_ICMPV6_TYPE))
+            .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_IPV6_NEXT_HEADER))
+            .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_L4_SRC_PORT))
+            .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_L4_DST_PORT))
+            .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_TCP_FLAGS))
+            .build()
+    );
```
bingwang-ms, line 3252: Two questions
1. Do we need to handle conflicts? For example, one ACL rule contains both "SAI_ACL_TABLE_ATTR_FIELD_SRC_IP` and `SAI_ACL_TABLE_ATTR_FIELD_SRC_IPV6`.
2. If a rule doesn't contain  `SAI_ACL_TABLE_ATTR_FIELD_ETHER_TYPE` or `SAI_ACL_TABLE_ATTR_FIELD_ACL_IP_TYPE`, should it be allowed ?

bingwang-ms: Thanks, we need to ensure the TCAM slice is wide enough to hold so many fields.

rck-innovium: > Two questions
> 
> 1.  Do we need to handle conflicts? For example, one ACL rule contains both `SAI_ACL_TABLE_ATTR_FIELD_SRC_IP` and `SAI_ACL_TABLE_ATTR_FIELD_SRC_IPV6`
> 2.  If a rule doesn't contain `SAI_ACL_TABLE_ATTR_FIELD_ETHER_TYPE` or `SAI_ACL_TABLE_ATTR_FIELD_ACL_IP_TYPE`, should it be allowed ?

1.  IMO, orchagent should not handle these conflicts-- similar to how orchagent does not handle one ACL rule containing both ICMP type and L4 source port.
2.  It should be allowed by orchagent for reasons:
    1.  Allow users to specify ACL rules that match only IN\_PORTS
    2.  Based on [feedback](https://github.com/sonic-net/SONiC/pull/1267#discussion_r1174595677) from other SAI vendors, we updated the HLD to specify that when IP\_type or Ether-type is not specified, the underlying SAI platform to handle based on their capability
        1.  install a single rule if the hardware supports it, or
        2.  make multiple rules (one rule per IP type), or
        3.  throw error as unsupported.

orchagent/aclorch.cpp
```diff
@@ -3049,11 +3083,36 @@ void AclOrch::init(vector<TableConnector>& connectors, PortsOrch *portOrch, Mirr
...
+
+    }
+
+
     SWSS_LOG_NOTICE("%s switch capability:", platform.c_str());
     SWSS_LOG_NOTICE("    TABLE_TYPE_MIRROR: %s",
             m_mirrorTableCapabilities[TABLE_TYPE_MIRROR] ? "yes" : "no");
     SWSS_LOG_NOTICE("    TABLE_TYPE_MIRRORV6: %s",
             m_mirrorTableCapabilities[TABLE_TYPE_MIRRORV6] ? "yes" : "no");
+    SWSS_LOG_NOTICE("    TABLE_TYPE_L3v4V6: Ingress [%s], Egress [%s]",
```
bingwang-ms, line 3112:
```suggestion
    SWSS_LOG_NOTICE("    TABLE_TYPE_L3V4V6: Ingress [%s], Egress [%s]",
```

orchagent/aclorch.cpp
bingwang-ms: Regarding #1, If orchagent doesn't handle this exception, then SAI has to handle it.  But I don't think we have clear agreement about how SAI handles this.

rck-innovium: Added the checks and unit tests.

test\_acl\_l3v4v6.py::TestAcl::test\_ValidAclRuleCreation\_sip\_dip PASSED                                                                          test\_acl\_l3v4v6.py::TestAcl::test\_InvalidAclRuleCreation\_sip\_sipv6 PASSED                                                                    test\_acl\_l3v4v6.py::TestAcl::test\_InvalidAclRuleCreation\_dip\_sipv6 PASSED

bingwang-ms (approved):

rck-innovium:
> LGTM. Are you going to update acl loader to support this new type as well?

Thank you, for the review. The acl-loader support is under review: sonic-net/sonic-utilities#2794

It is currently assigned to Prince and Qi Luo for review. But appreciate it if you can also review the acl-loader PR.

rck-innovium:
@prsunny 

The PR has been reviewed by Bing. Can you please merge before 202305 fork.

orchagent/aclorch.cpp
```diff
@@ -3049,11 +3083,36 @@ void AclOrch::init(vector<TableConnector>& connectors, PortsOrch *portOrch, Mirr
         };
     }
 
+    if ( platform == MRVL_PLATFORM_SUBSTRING ||
+            platform == INVM_PLATFORM_SUBSTRING ||
+            platform == VS_PLATFORM_SUBSTRING)
```
prsunny, line 3088: Why change VS platform to combined? Does it impact any current acl tests? @bingwang-ms

bingwang-ms: I think the change to VS makes no difference. AFAIK, We don't run acl test on VS. But even though, I think it doesn't make sense to change the behavior of VS. I suggest removing the change to VS_PLATFORM. @rck-innovium

rck-innovium: @bingwang-ms 

After removing the VS platform, I ran the tests that create ACL tables of the new L3V4V6 type and saw that the ACL table creation fails. 

```
marvell@cpss-testvm2:~/rck/sonic-dbg/sonic-buildimage/src/sonic-swss/tests$ sudo pytest -v  test_acl_l3v4v6.py
[sudo] password for marvell:
============================================================================================================================ test session starts =============================================================================================================================
:
:
test_acl_l3v4v6.py::TestAcl::test_L3V4V6AclTableCreationDeletion FAILED                                                                                                                                                                                                [ 20%]
:
:
=================================================================================================================================== ERRORS ===================================================================================================================================
________________________________________________________________________________________________________ ERROR at setup of TestAcl.test_ValidAclRuleCreation_sip_dip _________________________________________________________________________________________________________

self = <test_acl_l3v4v6.TestAcl object at 0x7f8141afba60>, dvs_acl = <dvslib.dvs_acl.DVSAcl object at 0x7f8141a564c0>

    @pytest.fixture
    def l3v4v6_acl_table(self, dvs_acl):
        try:
            dvs_acl.create_acl_table(L3V4V6_TABLE_NAME,
                                     L3V4V6_TABLE_TYPE,
                                     L3V4V6_BIND_PORTS)
>           yield dvs_acl.get_acl_table_ids(1)[0]

test_acl_l3v4v6.py:16:
```

The reason for the error is that the VS platform does NOT support the new ACL table type L3V4V6.

```
May 13 14:49:13.301367 e01c265b5ca5 ERR #orchagent: :- validate: Table L3_V4V6_TEST: table type L3V4V6 in stage 1 not supported on this platform.
May 13 14:49:13.301645 e01c265b5ca5 ERR #orchagent: :- doAclTableTask: Failed to create ACL table L3_V4V6_TEST, invalid configuration
```

@prsunny 

This change does not affect any other existing ACL  features since this capability check is used only for the newly introduced ACL table type L3V4V6. Also, I verified that all the current swss ACL tests pass as well as PTF test\_acl.py.

bingwang-ms: @rck-innovium Probably we can mock the platform by setting `DVS_FAKE_PLATFORM`. Please search in existing test code.
Thanks

rck-innovium: @bingwang-ms 

I was not able to find DVS\_FAKE\_PLATFORM in existing swss test code

marvell@cpss-testvm2:~/rck/sonic-dbg/sonic-buildimage/src/sonic-swss$ grep  -ir DVS\_FAKE\_PLATFORM \*  
marvell@cpss-testvm2:~/rck/sonic-dbg/sonic-buildimage/src/sonic-swss$ grep  -r FAKE\_PLATFORM \*  
marvell@cpss-testvm2:~/rck/sonic-dbg/sonic-buildimage/src/sonic-swss$ cd ..  
marvell@cpss-testvm2:~/rck/sonic-dbg/sonic-buildimage/src$ grep  -r DVS\_FAKE\_PLATFORM \*  
marvell@cpss-testvm2:~/rck/sonic-dbg/sonic-buildimage/src$

bingwang-ms: @rck-innovium Can you please try to set the `DVS_FAKE_PLATFORM` as testing script `src/sonic-swss/tests/test_mirror_ipv6_combined.py` does? I didn't try it by myself. But it probably works.

bingwang-ms: @rck-innovium Sorry for the misleading. I just realize that the `DVS_FAKE_PLATFORM` has been removed from the latest testing code. I don't know if we have other method to mock the platform string. @prsunny       Any suggestions?

prsunny: E.g: 
```
DVS_FAKE_PLATFORM = "marvell"
fakeplatform = getattr(request.module, "DVS_FAKE_PLATFORM", None)
dvs = DockerVirtualSwitch(name, keeptb, fakeplatform)
```

rck-innovium: @prsunny  @bingwang-ms 

DVS_FAKE_PLATFORM has been removed from the entire SONiC codebase and in SWSS it has been replaced with DVS_ENV.  

 I digged through the past versions of SONiC to find that Nvidia has changed DVS_FAKE_PLATFORM to DVS_ENV = ["HWSKU=Mellanox-SN2700"]. The reason I presume is to support mimicking at a deeper HwSKU level instead of just the platform.
 
However, I see in today’s swss/tests, **_**only Mellanox-SN2700 SKU is being faked**_**.  In fact, I ran into several issues when faking Marvell and Innovium platforms.
I have raised  PR https://github.com/sonic-net/sonic-swss/issues/2785 to track the issues.

rck-innovium: @prsunny 
Coming back to the [original question](https://github.com/sonic-net/sonic-swss/pull/2735/files/456bbfe21d2475b69b4b5580c6af8ae6f005c7fb#r1192793544), the new checks do NOT affect any other feature since this capability check is only used by the new L3V4V6 ACL table feature.
So adding VS platform does not have any impact on any of the existing feature.

PR #2736, opened 2023-04-14
[bugfix] vnet ping missing with secondary endpoints empty in priority routes.

siqbal1986:
When a configuration is added with custom monitoring such that there are no secondary endpoints in the configuration, this does not result in vnet ping session creation. This change fixes that and adds a test.

**What I did**
added a fix.
**Why I did it**

**How I verified it**
added a test which emulates this scenario.
**Details if related**

### DIFF
--- a/tests/test_vnet.py
+++ b/tests/test_vnet.py
@@ -3171,6 +3171,30 @@ class TestVnetOrch(object):
         check_remove_state_db_routes(dvs, 'Vnet12', "100.100.1.67/32")
         #adv should be gone.
         check_remove_routes_advertisement(dvs, "100.100.1.0/24")
+
+		#Add priority route with no secondary enpoints
+        create_vnet_routes(dvs, "100.100.1.71/32", vnet_name, '19.0.0.1,19.0.0.2', ep_monitor='19.1.0.1,19.1.0.2', profile = "test_prf", primary ='19.0.0.1,19.0.0.2',monitoring='custom', adv_prefix='100.100.1.0/24')
+        update_monitor_session_state(dvs, '100.100.1.71/32', '19.1.0.1', 'up')
+        update_monitor_session_state(dvs, '100.100.1.71/32', '19.1.0.2', 'up')
+
+        time.sleep(2)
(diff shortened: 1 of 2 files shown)


tests/test_vnet.py
```diff
@@ -3171,6 +3171,30 @@ def test_vnet_orch_22(self, dvs, testlog):
         check_remove_state_db_routes(dvs, 'Vnet12', "100.100.1.67/32")
         #adv should be gone.
         check_remove_routes_advertisement(dvs, "100.100.1.0/24")
+
+		#Add priority route with no secondary enpoints
```
prsunny, line 3175: please fix alignment

tests/test_vnet.py
```diff
@@ -3171,6 +3171,30 @@ def test_vnet_orch_22(self, dvs, testlog):
...
+        create_vnet_routes(dvs, "100.100.1.71/32", vnet_name, '19.0.0.1,19.0.0.2', ep_monitor='19.1.0.1,19.1.0.2', profile = "test_prf", primary ='19.0.0.1,19.0.0.2',monitoring='custom', adv_prefix='100.100.1.0/24')
```
prsunny, line 3176: Lets have this test with both endpoint and endpoint monitor the same ips.

tests/test_vnet.py
```diff
@@ -3171,6 +3171,30 @@ def test_vnet_orch_22(self, dvs, testlog):
...
+        update_monitor_session_state(dvs, '100.100.1.71/32', '19.1.0.1', 'up')
+        update_monitor_session_state(dvs, '100.100.1.71/32', '19.1.0.2', 'up')
```
prsunny, line 3178: please also add a check on "no bfd" sessions created

tests/test_vnet.py
siqbal1986: sure, ill change it.

tests/test_vnet.py
siqbal1986: done

tests/test_vnet.py
siqbal1986: good idea. ill add it.

prsunny (approved):

PR #2737, opened 2023-04-17
Add orchagent heart beat message for watchdog.

liuh-80:
**What I did**
Improve orch agent: output heartbeat message to systemd.

**Why I did it**
Currently SONiC monit system only monit orchagent process exist or not. If orchagent process stuck and stop processing, current monit can't find and report it.

**How I verified it**
Pass all UT.
Manually validate the heartbeat message works correctly.

**Details if related**
Another inprogress PR will add watchdog for this heartbeat message:
https://github.com/sonic-net/sonic-buildimage/pull/14686

sonic-mgmt UT PR: https://github.com/sonic-net/sonic-mgmt/pull/8306

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -771,7 +771,7 @@ void OrchDaemon::start()
         {
             heart_beat = tend;
             // output heart beat message to supervisord with 'PROCESS_COMMUNICATION' event: http://supervisord.org/events.html
-            cout << "<!--XSUPERVISOR:BEGIN-->" << "process:orchagent\norchagent heart beat message." << "<!--XSUPERVISOR:END-->" << endl;
+            cout << "<!--XSUPERVISOR:BEGIN-->\norchagent heart beat message.\n<!--XSUPERVISOR:END-->" << endl;
         }
 
         auto *c = (Executor *)s;
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -18,6 +18,9 @@ using namespace swss;
 #define SELECT_TIMEOUT 1000
 #define PFC_WD_POLL_MSECS 100
 
+/* orchagent heart beat message interval */
+#define HEART_BEAT_INTERVAL_MSECS 60 * 1000
+
 extern sai_switch_api_t*           sai_switch_api;
 extern sai_object_id_t             gSwitchId;
 extern bool                        gSaiRedisLogRotate;
@@ -713,6 +716,7 @@ void OrchDaemon::start()
     }
 
     auto tstart = std::chrono::high_resolution_clock::now();
+    auto heart_beat = tstart;
 
     while (true)
     {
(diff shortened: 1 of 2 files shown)


orchagent/orchdaemon.cpp
```diff
@@ -760,6 +765,15 @@ void OrchDaemon::start()
             continue;
         }
 
+        // output heart beat message to SYSLOG
```
qiluo-msft, line 768: >beat [](http://example.com/codeflow?start=24&length=4)

In the above part of while-loop block, there are several `continue` statements. So there is flow like early loop. Is it possible that early loop like timeout prevent heart beat?

orchagent/orchdaemon.cpp
```diff
@@ -760,6 +765,15 @@ void OrchDaemon::start()
...
+        diff = std::chrono::duration_cast<std::chrono::milliseconds>(tend - heart_beat);
+        if (diff.count() >= HEART_BEAT_INTERVAL_MSECS)
+        {
+            heart_beat = tend;
+            // output heart beat message to supervisord with 'PROCESS_COMMUNICATION_STDOUT' event: http://supervisord.org/events.html
+            cout << "<!--XSUPERVISOR:BEGIN-->heartbeat<!--XSUPERVISOR:END-->" << endl;
```
qiluo-msft, line 774: >heartbeat [](http://example.com/codeflow?start=45&length=9)

Is this message general enough for future? Assume multiple processes inside one container want to send heartbeats, but the heartbeats should be tracked per process.

liuh-80: Yes, when supervisord receive this message, supervisortd will add header information to the message, include process name.

In my drafter PR the watchdog already support new process by check the heartbeat message process name: https://github.com/sonic-net/sonic-buildimage/pull/14686

orchagent/orchdaemon.cpp
liuh-80: Fixed, move code to new method and send message at beginning of every loop.

orchagent/orchdaemon.cpp
```diff
@@ -958,6 +963,20 @@ void OrchDaemon::addOrchList(Orch *o)
     m_orchList.push_back(o);
 }
 
+void OrchDaemon::heartBeat(std::chrono::time_point<std::chrono::high_resolution_clock> tcurrent)
+{
+    static auto tlast = std::chrono::high_resolution_clock::now();
```
qiluo-msft, line 968: >static [](http://example.com/codeflow?start=4&length=6)

You are assuming `OrchDaemon` has only single instance in the process. To be super safe, you can use a static member variable instead of a static  function variable.

liuh-80: Fixed, change to a member variable.

orchagent/orchdaemon.cpp
```diff
@@ -64,6 +68,8 @@ event_handle_t g_events_handle;
 #define DEFAULT_MAX_BULK_SIZE 1000
 size_t gMaxBulkSize = DEFAULT_MAX_BULK_SIZE;
 
+std::chrono::time_point<std::chrono::high_resolution_clock> OrchDaemon::m_lastHeartBeat = std::chrono::high_resolution_clock::now();
```
qiluo-msft, line 71: >m_lastHeartBeat [](http://example.com/codeflow?start=72&length=15)

If not static, should it be initialized in ctor?

orchagent/orchdaemon.cpp
liuh-80: Fixed, change to none static member.

orchagent/orchdaemon.cpp
liuh-80: Fixed, initialized in ctor

PR #2754, opened 2023-04-27
Dual-ToR][ACL] bind LAG to ACL table in order to guarantee rule coverage if lag menber will be added to LAG after binding

ayurkiv-nvda:
…

**What I did**
Bind LAG to ACL table in order to guarantee ACL rule coverage if LAG member will be added to LAG after ACL binding

**Why I did it**
After config reload or reboot due to timing, we often see race condition when port are not yet added to LAG, but egress ACL table group already created and port is bound to this group.
It causes that ACL rules are not working on LAG's members that were added after binding.

**How I verified it**
Run test:
`py.test dualtor_io/test_normal_op.py::test_normal_op_upstream[active-standby`]

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -3199,6 +3199,7 @@ void AclOrch::initDefaultTableTypes()
     addAclTableType(
         builder.withName(TABLE_TYPE_DROP)
             .withBindPointType(SAI_ACL_BIND_POINT_TYPE_PORT)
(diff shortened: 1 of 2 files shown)


ayurkiv-nvda:
Fix is related to https://github.com/sonic-net/sonic-swss/pull/2646/files

bingwang-ms (approved):

PR #2775, opened 2023-05-12
Revert "[dpu] use default operation response timeout (#2742)"

oleksandrivantsiv:
This reverts commit 506b3608482393ad091049d818c5a0291a2e0d2d.

**What I did**
Revert changes introduced in https://github.com/sonic-net/sonic-swss/pull/2742.

**Why I did it**
In some corner cases, 60 seconds is insufficient for the "create switch" operation on BF-2.

**How I verified it**
Run sonic-mgmt tests on DPU

**Details if related**

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -589,7 +589,7 @@ int main(int argc, char **argv)
     attr.value.u64 = gSwitchId;
     attrs.push_back(attr);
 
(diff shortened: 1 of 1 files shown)


prabhataravind (approved):
Approving the changes. To be reverted when BF2 is deprecated.

PR #2776, opened 2023-05-15
[Mclagsyncd] Ignore select event if its return value is error.

Minkang-Tsai:
**What I did**
Ignore select event if its return value is error.

**Why I did it**
In order to prevent the mclagsyncd terminated.


gechiang:
@Minkang-Tsai , 
I have a few asks for you:
1, Is there an issue filed for this issue that you are trying to fix?  If so, can you link it to this PR? 
2. Can you add Unit test testcase for this change.  I don't see any test code added/changed to cover your changes.

@Praveen-Brcm can you please help review this?
Thanks!

mclagsyncd/mclagsyncd.cpp
```diff
@@ -66,10 +66,16 @@ int main(int argc, char **argv)
...
                 Selectable *temps;
+                int ret;
 
                 /* Reading MCLAG messages forever (and calling "readData" to read them) */
-                s.select(&temps);
+                ret = s.select(&temps);
 
+                if (ret == Select::ERROR)
+                {
+                    SWSS_LOG_NOTICE("Error: %s!", strerror(errno));
```
gechiang, line 76: Since this is error case, should the log level be elevated to WARNING or ERROR?

Minkang-Tsai:
> @Minkang-Tsai , I have a few asks for you: 1, Is there an issue filed for this issue that you are trying to fix? If so, can you link it to this PR? 2. Can you add Unit test testcase for this change. I don't see any test code added/changed to cover your changes.
> 
> @Praveen-Brcm can you please help review this? Thanks!

Hi @gechiang 
Our customer try to kill docker database directly. Under this situation, select return value in mclagsyncd is ERROR, but it can continue to do the following steps to possibly cause the segment fault. Do you think that we need to create a PR to record that issue?

gechiang:
> > @Minkang-Tsai , I have a few asks for you: 1, Is there an issue filed for this issue that you are trying to fix? If so, can you link it to this PR? 2. Can you add Unit test testcase for this change. I don't see any test code added/changed to cover your changes.
> > @Praveen-Brcm can you please help review this? Thanks!
> 
> Hi @gechiang Our customer try to kill docker database directly. Under this situation, select return value in mclagsyncd is ERROR, but it can continue to do the following steps to possibly cause the segment fault. Do you think that we need to create a PR to record that issue?

You have two choices:
1. Raise a GIT hub issue under sonic-net/sonic-buildimage for what your customer tried and describes the steps, and what was observed, and what is the desired outcome after the trigger was made. and share the issue link in this PR.
2. Fully describe the issue as I described above without filing a git hub issue in this PR description section.  

Reason for this ask is we need to clearly describe the scenario involved so it helps the reviewer to see what is happening and what your changes will help to resolve the issue.

PR #2779, opened 2023-05-17
[dash] Change dash orchagent from Redis consumer state table to ZMQ consumer state table.

liuh-80:
Change dash orchagent from Redis consumer state table to ZMQ consumer state table.

**What I did**
Change dash orchagent from Redis consumer state table to ZMQ consumer state table.

**Why I did it**
To improve dash entry create performance, change dash orchagent from Redis consumer state table to ZMQ consumer state table.

**How I verified it**
Pass all UT.

**Details if related**

### DIFF
--- a/orchagent/dash/dashaclorch.cpp
+++ b/orchagent/dash/dashaclorch.cpp
@@ -285,8 +285,8 @@ sai_attr_id_t getSaiStage(DashAclDirection d, sai_ip_addr_family_t f, DashAclSta
     return stage->second;
 }
 
-DashAclOrch::DashAclOrch(DBConnector *db, const vector<string> &tables, DashOrch *dash_orch) :
-    Orch(db, tables),
+DashAclOrch::DashAclOrch(DBConnector *db, const vector<string> &tables, DashOrch *dash_orch, ZmqServer *zmqServer) :
+    Orch(db, tables, zmqServer),
     m_dash_orch(dash_orch)
 {
     SWSS_LOG_ENTER();
@@ -294,7 +294,7 @@ DashAclOrch::DashAclOrch(DBConnector *db, const vector<string> &tables, DashOrch
     assert(m_dash_orch);
 }
 
-void DashAclOrch::doTask(Consumer &consumer)
+void DashAclOrch::doTask(ZmqConsumer &consumer)
 {
     SWSS_LOG_ENTER();
 
--- a/orchagent/dash/dashaclorch.cpp
+++ b/orchagent/dash/dashaclorch.cpp
@@ -286,7 +286,7 @@ sai_attr_id_t getSaiStage(DashAclDirection d, sai_ip_addr_family_t f, DashAclSta
 }
 
 DashAclOrch::DashAclOrch(DBConnector *db, const vector<string> &tables, DashOrch *dash_orch, ZmqServer *zmqServer) :
-    Orch(db, tables, zmqServer),
+    ZmqOrch(db, tables, zmqServer),
     m_dash_orch(dash_orch)
 {
     SWSS_LOG_ENTER();
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -83,7 +83,7 @@ string gMyAsicName = "";
 
 void usage()
 {
-    cout << "usage: orchagent [-h] [-r record_type] [-d record_location] [-f swss_rec_filename] [-j sairedis_rec_filename] [-b batch_size] [-m MAC] [-i INST_ID] [-s] [-z mode] [-k bulk_size]" << endl;
+    cout << "usage: orchagent [-h] [-r record_type] [-d record_location] [-f swss_rec_filename] [-j sairedis_rec_filename] [-b batch_size] [-m MAC] [-i INST_ID] [-s] [-z mode] [-k bulk_size] [-q zmq_server_address]" << endl;
     cout << "    -h: display this message" << endl;
     cout << "    -r record_type: record orchagent logs with type (default 3)" << endl;
     cout << "                    Bit 0: sairedis.rec, Bit 1: swss.rec, Bit 2: responsepublisher.rec. For example:" << endl;
@@ -101,6 +101,7 @@ void usage()
     cout << "    -f swss_rec_filename: swss record log filename(default 'swss.rec')" << endl;
     cout << "    -j sairedis_rec_filename: sairedis record log filename(default sairedis.rec)" << endl;
     cout << "    -k max bulk size in bulk mode (default 1000)" << endl;
+    cout << "    -q zmq_server_address: ZMQ server address (default tcp://*:1234)" << endl;
 }
 
 void sighup_handler(int signo)
@@ -349,6 +350,7 @@ int main(int argc, char **argv)
     string record_location = ".";
     string swss_rec_filename = "swss.rec";
     string sairedis_rec_filename = "sairedis.rec";
+    string zmq_server_address = "tcp://*:1234";
     string responsepublisher_rec_filename = "responsepublisher.rec";
     int record_type = 3; // Only swss and sairedis recordings enabled by default.
 
@@ -430,6 +432,12 @@ int main(int argc, char **argv)
                 }
             }
             break;
+        case 'q':
+            if (optarg)
+            {
+                zmq_server_address = optarg;
+            }
+            break;
         default: /* '?' */
             exit(EXIT_FAILURE);
         }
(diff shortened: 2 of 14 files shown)


liuh-80:
The code change in this PR already reviewed by this PR: https://github.com/nvidia-sonic/sonic-swss/pull/1

orchagent/dash/dashaclorch.cpp
```diff
@@ -285,16 +285,16 @@ sai_attr_id_t getSaiStage(DashAclDirection d, sai_ip_addr_family_t f, DashAclSta
     return stage->second;
 }
 
-DashAclOrch::DashAclOrch(DBConnector *db, const vector<string> &tables, DashOrch *dash_orch) :
-    Orch(db, tables),
+DashAclOrch::DashAclOrch(DBConnector *db, const vector<string> &tables, DashOrch *dash_orch, ZmqServer *zmqServer) :
+    Orch(db, tables, zmqServer),
```
Pterosaur, line 289: According to the function addConsumer in orch.cpp, original redis db and new zmqServer are mutually exclusive. If yes, why did this contractor receive both DBConnector and ZmqConnector?

orchagent/main.cpp
```diff
@@ -349,6 +350,7 @@ int main(int argc, char **argv)
     string record_location = ".";
     string swss_rec_filename = "swss.rec";
     string sairedis_rec_filename = "sairedis.rec";
+    string zmq_server_address = "tcp://*:1234";
```
Pterosaur, line 353: Should we define the port to schema or /etc/sonic/constants.yml?

orchagent/main.cpp
```diff
@@ -445,6 +453,10 @@ int main(int argc, char **argv)
     DBConnector config_db("CONFIG_DB", 0);
     DBConnector state_db("STATE_DB", 0);
 
+    // Instantiate ZMQ server
```
Pterosaur, line 456: If ZMQ architecture is DASH-specific, should we skip creating the zmq_server for the normal device? Otherwise is there any security concern for extra exported ports?

orchagent/orch.cpp
```diff
@@ -264,6 +266,28 @@ void Consumer::drain()
...
+
+    drain();
+}
+
+void ZmqConsumer::drain()
+{
+    if (!m_toSync.empty())
+        m_orch->doTask(*this);
+}
+
```
Pterosaur, line 290: These two functions are very similar with the class Consumer's(execute, drain). Is it possible to reuse the same code?

Pterosaur (changes requested):

orchagent/dash/dashaclorch.cpp
liuh-80: This is because ZmqConsumerStateTable need use db connector to write received data to redis:
So the ctor of ZmqConsumerStateTable need a DBConnector as parameter:

    ZmqConsumerStateTable(DBConnector *db, const std::string &tableName, ZmqServer &zmqServer, int popBatchSize = DEFAULT_POP_BATCH_SIZE, int pri = 0, bool dbPersistence = true);

orchagent/main.cpp
liuh-80: Yes, as my understand create ZMQ server code will only merge to dash branch.
In master branch, we don't use ZMQ.

orchagent/orch.cpp
liuh-80: For ZmqConsumer::execute(), currently it's similer, but we have plan to use protocolbuffer for Dash related entity. after protobuffer PR merged, ZMQ will update to not send/receive KeyOpFieldsValuesTuple. so this method will update soon.

For ZmqConsumer::drain(), there are following parametric polymorphism methods, so we need the class information use corresponding method for ZmqConsumer and orginal redis based Consumer:
    virtual void doTask(Consumer &consumer) { };
    virtual void doTask(ZmqConsumer &consumer) { };

orchagent/main.cpp
liuh-80: Will create another PR to put this to sonic-swss-common schema.h and update later.

orchagent/orch.cpp
liuh-80: TODO: Try improve with template.

liuh-80:
This PR now depends on this merge first: https://github.com/sonic-net/sonic-swss-common/pull/795

orchagent/dash/dashaclorch.cpp
liuh-80: Fixed, add ZmqOrch class. all dash orch will inherit from ZmqOrch.

orchagent/main.cpp
liuh-80: Will by default disable ZMQ.
Will send emaill notify this change.

orchagent/main.cpp
```diff
@@ -101,6 +101,7 @@ void usage()
     cout << "    -f swss_rec_filename: swss record log filename(default 'swss.rec')" << endl;
     cout << "    -j sairedis_rec_filename: sairedis record log filename(default sairedis.rec)" << endl;
     cout << "    -k max bulk size in bulk mode (default 1000)" << endl;
+    cout << "    -q zmq_server_address: ZMQ server address (default tcp://*:8020)" << endl;
```
liuh-80, line 104: Change to 127.0.0.1

orchagent/orch.cpp
liuh-80: WIll improve after UT issue fixed.

orchagent/main.cpp
liuh-80: Fixed.

Pterosaur (approved):

orchagent/main.cpp
```diff
@@ -101,6 +101,7 @@ void usage()
     cout << "    -f swss_rec_filename: swss record log filename(default 'swss.rec')" << endl;
     cout << "    -j sairedis_rec_filename: sairedis record log filename(default sairedis.rec)" << endl;
     cout << "    -k max bulk size in bulk mode (default 1000)" << endl;
+    cout << "    -q zmq_server_address: ZMQ server address (default tcp://127.0.0.1:8020)" << endl;
```
qiluo-msft, line 104: >tcp://127.0.0.1:8020 [](http://example.com/codeflow?start=68&length=20)

You can construct the default string by code.

orchagent/dash/dashaclorch.cpp
```diff
@@ -285,16 +285,16 @@ sai_attr_id_t getSaiStage(DashAclDirection d, sai_ip_addr_family_t f, DashAclSta
...
+    ZmqOrch(db, tables, zmqServer),
     m_dash_orch(dash_orch)
 {
     SWSS_LOG_ENTER();
 
     assert(m_dash_orch);
 }
 
-void DashAclOrch::doTask(Consumer &consumer)
+void DashAclOrch::doTask(ZmqConsumer &consumer)
```
qiluo-msft, line 297: >ZmqConsumer [](http://example.com/codeflow?start=25&length=11)

Possible to use ConsumerBase?

orchagent/dash/dashaclorch.cpp
```diff
@@ -285,16 +285,16 @@ sai_attr_id_t getSaiStage(DashAclDirection d, sai_ip_addr_family_t f, DashAclSta
     return stage->second;
 }
 
-DashAclOrch::DashAclOrch(DBConnector *db, const vector<string> &tables, DashOrch *dash_orch) :
-    Orch(db, tables),
+DashAclOrch::DashAclOrch(DBConnector *db, const vector<string> &tables, DashOrch *dash_orch, ZmqServer *zmqServer) :
+    ZmqOrch(db, tables, zmqServer),
```
qiluo-msft, line 289: >ZmqOrch [](http://example.com/codeflow?start=4&length=7)

Is it possible to working with a original Orch or ZmqOrch based on runtime argument?

orchagent/main.cpp
liuh-80: This string removed by https://github.com/sonic-net/sonic-swss/pull/2828

orchagent/dash/dashaclorch.cpp
liuh-80: No, because DashAclOrch inherite from ZmqOrch, this method override a virtual method from ZmqOrch::doTask(ZmqConsumer &consumer).

We can't define ZmqOrch::doTask(ConsumerBase &consumer). because the doTask code implementation for Consumer  and ZmqConsumer  are different.

orchagent/dash/dashaclorch.cpp
liuh-80: It's difficult to switch between Orch and ZmqOrch by parameter because they are base class
Maybe I can rename ZmqOrch to DashOrchBase.

liuh-80: Created new PR: https://github.com/sonic-net/sonic-swss/pull/2836

PR #2781, opened 2023-05-17
[fdborch] Only call flushFDBEntries() when there is a bridge port OID in port state update

mint570:
**What I did**
Only call flushFDBEntries() when there is a bridge port OID in port state update for fdborch.

**Why I did it**
The flushFDBEntries() has the check at the beginning. But it is wrong to call the function to begin with.
This will get rid of the false warning log message in flushFDBEntries().

**How I verified it**

**Details if related**


dgsudharsan (approved):

prsunny (approved):

prsunny:
Pending on PR - https://github.com/sonic-net/sonic-swss/pull/2900

PR #2782, opened 2023-05-17
Periodic flush record files instead of flushing them on std::endl.

mint570:
**What I did**
Periodic flush record files instead of flushing them on std::endl.

**Why I did it**
For performance purposes. In our testing, this provides 8% end-to-end performance improvement.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -589,7 +589,7 @@ void Orch::recordTuple(ConsumerBase &consumer, const KeyOpFieldsValuesTuple &tup
 {
     string s = consumer.dumpTuple(tuple);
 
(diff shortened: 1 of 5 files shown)


lguohan:
@mint570 , @prsunny , this seems like a big improvement. @mint570, can you explain a little more on why such change would have 8% performance improvement?

mint570:
This PR comes together with https://github.com/sonic-net/sonic-sairedis/pull/1241.
There are some discussion there.

Flushing a file is expensive (depending on the FS). The original code flush the file on every write. This PR changes to flush every second (or when the buffer is full). This will increase the performance but the downside is that when the program crashes, we might miss some logs.

PR #2790, opened 2023-05-23
[202205][VNET] consider empty an peer string in BfdUpdate entry as a Warning

ayurkiv-nvda:
**What I did**
changed log severity to WARN

**Why I did it**
When BfdOrch sends an update notification, it is handled in VNetRouteOrch::update()
Sometimes instead of a valid BFD peer string (like this one):
`NOTICE swss#orchagent: :- doTask: BFD session state for default|default|104.0.0.241 changed from Up to Down `
We receive an empty string (like this one):
`NOTICE swss#orchagent: :- doTask: BFD session state for  changed from Admin_Down to Down`

This will also cause passing empty string in the notification mentioned above, and in updateVnetTunnel(), we will receive an error message due to the if condition:
`ERR swss#orchagent: :- updateVnetTunnel: Failed to parse key , no vrf is given `
There could be a possibility that while SAI gives notification for a session, the BFD object itself could have been removed, which may result in the scenario described above .

This is expected behavior, but loglevel severity can't be ERR or INFO
It should be WARN

**How I verified it**

**Details if related**

### DIFF
--- a/.azure-pipelines/test-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/test-docker-sonic-vs-template.yml
@@ -55,7 +55,7 @@ jobs:
       # install packages for vs test
       sudo apt-get install -y net-tools bridge-utils vlan
       sudo apt-get install -y python3-pip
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@ayurkiv-nvda is this still relevant? if so, please handle conflicts and move from draft

PR #2791, opened 2023-05-24
[VNET] [202211] consider empty an peer string in BFD entry as a Warning

ayurkiv-nvda:
**What I did**
changed log severity to WARN

**Why I did it**
When BfdOrch sends an update notification, it is handled in VNetRouteOrch::update()
Sometimes instead of a valid BFD peer string (like this one):
`NOTICE swss#orchagent: :- doTask: BFD session state for default|default|104.0.0.241 changed from Up to Down `
We receive an empty string (like this one):
`NOTICE swss#orchagent: :- doTask: BFD session state for  changed from Admin_Down to Down`

This will also cause passing empty string in the notification mentioned above, and in updateVnetTunnel(), we will receive an error message due to the if condition:
`ERR swss#orchagent: :- updateVnetTunnel: Failed to parse key , no vrf is given `
There could be a possibility that while SAI gives notification for a session, the BFD object itself could have been removed, which may result in the scenario described above .

This is expected behavior, but loglevel severity can't be ERR or INFO
**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -1730,7 +1730,7 @@ void VNetRouteOrch::updateVnetTunnel(const BfdUpdate& update)
     size_t found_vrf = key.find(state_db_key_delimiter);
     if (found_vrf == string::npos)
     {
-        SWSS_LOG_ERROR("Failed to parse key %s, no vrf is given", key.c_str());
+        SWSS_LOG_WARN("Failed to parse key %s, no vrf is given", key.c_str());
         return;
     }
 


ayurkiv-nvda:
@prsunny can you please take a look?

ayurkiv-nvda:
> Cherry-pick from master, added label to master PR

There are might be merge problem, because in master  and 202211 loglevel of changed messages are different:
**master** : 
**INFO -> WARN** 

**202211:**
**ERR->WARN**

prsunny (approved):

prsunny:
Master PR  - https://github.com/sonic-net/sonic-swss/pull/2792

liat-grozovik:
@prsunny  can you merge this one as well?

keboliu:
@StormLiangMS  would you please merge this PR?

StormLiangMS (commented):
LGTM

PR #2798, opened 2023-05-26
[202205] [pfcwd] Enhance DLR_INIT based recovery and DLR_PACKET_ACTION for broadcom platforms

neethajohn:
**What I did**
This PR contains the following changes
* for brcm platforms that support DLR_INIT based recovery, use that method for both drop and forward action
* While using DLR_INIT based recovery, do not update the pfc bitmask for that port
* Update the logic to handle allowing drop/forward action on all ports without requiring a restart of swss
* Mock tests for DLR_INIT and DLR_PACKET action

**Why I did it**
To provide support for DLR_INIT based pfcwd recovery on broadcom platforms in certain roles

**How I verified it**
* Mock tests
* Tested on various brcm platforms (TD3 dual tor, Th2 T1, TD3 single tor) with a custom swss deb along with updated bcm config and verified that the pfcwd recovery mechanism is as expected

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -604,7 +604,7 @@ bool OrchDaemon::init()
 
         if(gSwitchOrch->checkPfcDlrInitEnable())
         {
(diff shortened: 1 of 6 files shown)


neethajohn:
Closing. included the changes in #2807

PR #2801, opened 2023-05-30
[QoS] Enhance QosOrch::doTask by handling QUEUE table after all the rest tables handled to avoid retry

stephenxs:
**What I did**

Enhance `QosOrch::doTask` by handling QUEUE table after all the rest tables handled to avoid retry

**Why I did it**

**How I verified it**

**Details if related**

During system initialization, QoS table items are received before `gPortsOrch->allPortsReady()` becomes true and will be handled all together after it becomes true.
In most cases, it will be handled in the for loop in the following snippet of code in `OrchDaemon::start()`
```
        auto *c = (Executor *)s;
        c->execute();

        /* After each iteration, periodically check all m_toSync map to
         * execute all the remaining tasks that need to be retried. */

        /* TODO: Abstract Orch class to have a specific todo list */
        for (Orch *o : m_orchList)
            o->doTask();
```
The `QUEUE` table items reference `WRED_PROFILE` and `SCHEDULER_PROFILE` table items. In case the latter tables are handled after the `QUEUE` table, the `QUEUE` table needs to be retried in the next for loop, which is not necessary.
So, we adjust the order in which the tables are handled to guarantee that all QoS table items to be handled in a single call to `QosOrch::doTask`.

### DIFF
--- a/orchagent/qosorch.cpp
+++ b/orchagent/qosorch.cpp
@@ -2137,12 +2137,13 @@ void QosOrch::doTask()
     SWSS_LOG_ENTER();
 
     auto *port_qos_map_cfg_exec = getExecutor(CFG_PORT_QOS_MAP_TABLE_NAME);
+    auto *queue_exec = getExecutor(CFG_QUEUE_TABLE_NAME);
 
     for (const auto &it : m_consumerMap)
     {
         auto *exec = it.second.get();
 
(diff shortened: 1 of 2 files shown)


Junchao-Mellanox (approved):

stephenxs:
Added `Included in 202211 branch` as the PR has been cherry-picked and the submodule has been merged.

PR #2807, opened 2023-05-31
[pfcwd] Enhance DLR_INIT based recovery and DLR_PACKET_ACTION for broadcom platforms

neethajohn:
What I did
This PR contains the following changes
* for brcm platforms that support DLR_INIT based recovery, use that method for both drop and forward action
* While using DLR_INIT based recovery, do not update the pfc bitmask for that port
* Update the logic to handle allowing drop/forward action on all ports without requiring a restart of swss
* Mock tests for DLR_INIT and DLR_PACKET action

Why I did it
To provide support for DLR_INIT based pfcwd recovery on broadcom platforms in certain roles

How I verified it
Mock tests
Tested on various brcm platforms (TD3 dual tor, Th2 T1, TD3 single tor) with a custom swss deb along with updated bcm config and verified that the pfcwd recovery mechanism is as expected on 202205
Tested on td3 dual tor with custom deb along with updated bcm config and verified functionality working fine on internal branch

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -238,11 +238,11 @@ task_process_status PfcWdOrch<DropHandler, ForwardHandler>::createEntry(const st
                 {
                     if(gSwitchOrch->checkPfcDlrInitEnable())
                     {
-                        if(getPfcDlrPacketAction() == PfcWdAction::PFC_WD_ACTION_UNKNOWN)
+                        if((getPfcDlrPacketAction() == PfcWdAction::PFC_WD_ACTION_UNKNOWN) || m_pfcwd_ports.empty())
                         {
                             sai_attribute_t attr;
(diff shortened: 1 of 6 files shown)


orchagent/pfcwdorch.cpp
```diff
@@ -238,11 +238,11 @@ task_process_status PfcWdOrch<DropHandler, ForwardHandler>::createEntry(const st
                 {
                     if(gSwitchOrch->checkPfcDlrInitEnable())
                     {
-                        if(getPfcDlrPacketAction() == PfcWdAction::PFC_WD_ACTION_UNKNOWN)
+                        if((getPfcDlrPacketAction() == PfcWdAction::PFC_WD_ACTION_UNKNOWN) || m_pfcwd_ports.empty())
```
prsunny, line 241: What is this check on line 228? `(action == PfcWdAction::PFC_WD_ACTION_UNKNOWN)`

prsunny (approved):

PR #2815, opened 2023-06-08
swss: Init vxlan configuration optimization

chidanta-shu:
* Added a data structure to store Vxlan configurations that have not yet created a basic configuration
* A thread pool is added to monitor the events of basic configuration update, and asynchronously process the Vxlan configuration with missing conditions through the event trigger mechanism

  Signed-off-by: wangbo3@cmss.chinamobile.com


prsunny:
@chidanta-shu , we are not taking any features to 201911 branch. If you propose this feature, please have an HLD and then plan to create PR to master. 
@zhangyanzhao for viz

PR #2827, opened 2023-06-21
[muxorch] set mux state to init upon warm reboot

Ndancejic:
What I did:
Initialize mux state to INIT during warm reboot

Why I did it:
initializing mux state to Standby during warm reboot was causing orchagent to crash if mux was put in manual mode

How I tested it:
vstests included in PR

### DIFF
--- a/tests/test_mux.py
+++ b/tests/test_mux.py
@@ -1478,6 +1478,27 @@ class TestMuxTunnel(TestMuxTunnelBase):
         for ip in test_ips:
             self.check_neighbor_state(dvs, dvs_route, ip, expect_route=False)
 
+    def test_warm_boot_mux_state(
+            self, dvs, appdb, dvs_route, setup_vlan, setup_mux_cable, setup_tunnel,
+            remove_peer_switch, neighbor_cleanup, testlog
+    ):
+        """
+        test mux initialization during warm boot.
+        """
+        self.set_mux_state(appdb, "Ethernet0", "active")
+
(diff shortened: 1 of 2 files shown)


tests/test_mux.py
```diff
@@ -1478,6 +1478,27 @@ def test_neighbor_miss_no_peer(
...
             self.check_neighbor_state(dvs, dvs_route, ip, expect_route=False)
 
+    def test_warm_boot_mux_state(
+            self, dvs, appdb, dvs_route, setup_vlan, setup_mux_cable, setup_tunnel,
+            remove_peer_switch, neighbor_cleanup, testlog
+    ):
+        """
+        test mux initialization during warm boot.
+        """
+        self.set_mux_state(appdb, "Ethernet0", "active")
```
vaibhavhd, line 1488: To be consistent with warmboot expectations, we should test a scenario where warmboot is applicable in dualtor systems - when there is a mix of active and standby states.

Hence, please add a mix (some active, some standby) before restarting swss, and verify 1) that respective states remain unchanged after restarting swss in warm_restart mode. 2) There are no flaps -- avoiding false negatives where double flap can result in same state as before.

Ndancejic: Thanks Vaibhav, I'll be closing the PR and reopening one for master branch first then requesting cherry-pick for 202205. I'll copy and paste your feedback so that it doesn't get lost in the transition.

PR #2828, opened 2023-06-21
Enable/disable Dash ZmqConsumer and switch back to Redis based consumer by parameter.

liuh-80:
Enable/disable Dash ZmqConsumer by parameter.

**What I did**
Change code to allow disable ZMQ and switch back to Redis Consumer.

**Why I did it**
Dash orch UT still need Redis based Consumer, so improve to disable ZMQ and switch back to Redis Consumer.

**How I verified it**
Pass all UT.

**Details if related**

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -101,7 +101,7 @@ void usage()
     cout << "    -f swss_rec_filename: swss record log filename(default 'swss.rec')" << endl;
     cout << "    -j sairedis_rec_filename: sairedis record log filename(default sairedis.rec)" << endl;
     cout << "    -k max bulk size in bulk mode (default 1000)" << endl;
-    cout << "    -q zmq_server_address: ZMQ server address (default tcp://127.0.0.1:8020)" << endl;
+    cout << "    -q zmq_server_address: ZMQ server address (default disable ZMQ)" << endl;
 }
 
 void sighup_handler(int signo)
@@ -351,6 +351,7 @@ int main(int argc, char **argv)
     string swss_rec_filename = "swss.rec";
     string sairedis_rec_filename = "sairedis.rec";
     string zmq_server_address = "tcp://127.0.0.1:" + to_string(ORCH_ZMQ_PORT);
+    bool   enable_zmq = false;
     string responsepublisher_rec_filename = "responsepublisher.rec";
     int record_type = 3; // Only swss and sairedis recordings enabled by default.
 
@@ -436,6 +437,7 @@ int main(int argc, char **argv)
             if (optarg)
             {
                 zmq_server_address = optarg;
+                enable_zmq = true;
             }
             break;
         default: /* '?' */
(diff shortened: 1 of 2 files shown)


Pterosaur (approved):

orchagent/main.cpp
```diff
@@ -453,9 +455,17 @@ int main(int argc, char **argv)
...
 
-    // Instantiate ZMQ server 
-    SWSS_LOG_NOTICE("Instantiate ZMQ server : %s", zmq_server_address.c_str());
-    ZmqServer zmq_server(zmq_server_address.c_str());
+    // Instantiate ZMQ server
+    shared_ptr<ZmqServer> zmq_server = nullptr;
+    if (enable_zmq)
+    {
+        SWSS_LOG_NOTICE("Instantiate ZMQ server : %s", zmq_server_address.c_str());
+        zmq_server = make_shared<ZmqServer>(zmq_server_address.c_str());
```
qiluo-msft, line 463: >zmq_server_address [](http://example.com/codeflow?start=44&length=18)

If `zmq_server_address` is an invalid argument, the constructor will not throw, but later "ZmqServer::mqPollThread" will throw. Will orchagent catch this exception and print error messages?

orchagent/orch.cpp
```diff
@@ -894,12 +894,20 @@ void ZmqOrch::addConsumer(DBConnector *db, string tableName, int pri, ZmqServer
 {
     if (db->getDbId() == APPL_DB)
     {
-        SWSS_LOG_DEBUG("[ZMQ] ZmqConsumer initialize for: %s", tableName.c_str());
-        addExecutor(new ZmqConsumer(new ZmqConsumerStateTable(db, tableName, *zmqServer, gBatchSize, pri), this, tableName));
+        if (zmqServer != nullptr)
```
prsunny, line 897: Does this mean, once enabled, it will be ZMQ for all APP_DB tables?

liuh-80: No, Once enabled, all orch inherite from ZmqOrch will use ZMQ for APP_DB tables.
Currently only Dash orch inherite from ZmqOrch, so once enabled, all Dash orch will use ZMQ for APP_DB tables.

liuh-80:
Enable/disable Dash ZmqConsumer by parameter.

**What I did**
Change code to allow disable ZMQ and switch back to Redis Consumer.

**Why I did it**
Dash orch UT still need Redis based Consumer, so improve to disable ZMQ and switch back to Redis Consumer.

**How I verified it**
Pass all UT.

**Details if related**

orchagent/main.cpp
liuh-80: In the "ZmqServer::mqPollThread" it's already log error to syslog before throw.
Because it's in the thread that exception will terminate the orchagent.

If it's necessary, I can add a global exception handler to catch and handle exception.

PR #2836, opened 2023-06-25
[dash] Improve dash orchagent ZMQ code.

liuh-80:
Improve dash orchagent code.

**What I did**
Move ZmqOrch related code to new files.
Remove unused code.
Change ZmqOrch::doTask() to use ConsumerBase as parameter.
Fix orchagent start parameter parse bug.
Fix dash orch can't receive data when ZMQ disabled issue.

**Why I did it**
Improve dash orchagent code.
HLD: https://github.com/sonic-net/SONiC/blob/master/doc/sonic-swss-common/ZMQ%20producer-consumer%20state%20table%20design.md?plain=1

**How I verified it**
Pass all UT.

**Details if related**

### DIFF
--- a/orchagent/dash/dashorchbase.h
+++ b/orchagent/dash/dashorchbase.h
@@ -1,0 +1,34 @@
+#pragma once
+
+#include <vector>
+#include <string>
+#include <orch.h>
+#include "zmqserver.h"
+
+class ZmqConsumer : public ConsumerBase {
+public:
+    ZmqConsumer(swss::ZmqConsumerStateTable *select, Orch *orch, const std::string &name)
+        : ConsumerBase(select, orch, name)
+    {
+    }
+
+    swss::TableBase *getConsumerTable() const override
+    {
+        // ZmqConsumerStateTable is a subclass of TableBase
+        return static_cast<swss::ZmqConsumerStateTable *>(getSelectable());
+    }
+
+    void execute() override;
+    void drain() override;
+};
+
+class DashOrchBase : public Orch
+{
+public:
+    DashOrchBase(swss::DBConnector *db, const std::vector<std::string> &tableNames, swss::ZmqServer *zmqServer);
+
+    virtual void doTask(ConsumerBase &consumer) { };
+
+private:
+    void addConsumer(swss::DBConnector *db, std::string tableName, int pri, swss::ZmqServer *zmqServer);
+};
--- a/orchagent/dash/dashvnetorch.h
+++ b/orchagent/dash/dashvnetorch.h
@@ -11,7 +11,7 @@
 #include "ipaddresses.h"
 #include "macaddress.h"
 #include "timer.h"
-#include <zmqorch.h>
+#include "zmqorch.h"
 #include "zmqserver.h"
 
 struct VnetEntry
--- a/orchagent/dash/dashvnetorch.h
+++ b/orchagent/dash/dashvnetorch.h
@@ -11,7 +11,7 @@
 #include "ipaddresses.h"
 #include "macaddress.h"
 #include "timer.h"
-#include "dashorchbase.h"
+#include <zmqorch.h>
 #include "zmqserver.h"
 
 struct VnetEntry
@@ -77,7 +77,7 @@ struct VnetMapBulkContext
     }
 };
 
-class DashVnetOrch : public DashOrchBase
+class DashVnetOrch : public ZmqOrch
 {
 public:
     DashVnetOrch(swss::DBConnector *db, std::vector<std::string> &tables, swss::ZmqServer *zmqServer);
--- a/orchagent/dash/dashvnetorch.h
+++ b/orchagent/dash/dashvnetorch.h
@@ -11,6 +11,7 @@
 #include "ipaddresses.h"
 #include "macaddress.h"
 #include "timer.h"
+#include "dashorchbase.h"
 #include "zmqserver.h"
 
 struct VnetEntry
@@ -76,7 +77,7 @@ struct VnetMapBulkContext
     }
 };
 
-class DashVnetOrch : public ZmqOrch
+class DashVnetOrch : public DashOrchBase
 {
 public:
     DashVnetOrch(swss::DBConnector *db, std::vector<std::string> &tables, swss::ZmqServer *zmqServer);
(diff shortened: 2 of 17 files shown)


orchagent/dash/dashorchbase.h
```diff
@@ -0,0 +1,34 @@
...
+    {
+        // ZmqConsumerStateTable is a subclass of TableBase
+        return static_cast<swss::ZmqConsumerStateTable *>(getSelectable());
+    }
+
+    void execute() override;
+    void drain() override;
+};
+
+class DashOrchBase : public Orch
```
mint570, line 25: Can this be named as "ZmqOrchBase" for other non-dash orchs to use it?

liuh-80: Currently this class is only designed for dash orch. because only dash orch using ZMQ to receive data from GNMI.
The reason I rename this from ZmqOrch to DashOrchBase because there are comments that create consumer/zmqconsumer based on parameter is wried for ZmqOrch, because the behavior is only required by dash so I rename this class.

mint570: We are looking at this to find potential performance improvement in our use case as well.
Wouldn't it be nice to make it usable to other orchs (and eventually migrate other orchs to use zmq)?

liuh-80: Yes, I think eventually we will support other orch.
Migrate ZMQ also need client side change, currently the Dash orch only have gnmi as client. so fir the first step we plan only for Dash orch.
And if other Orch want ZMQ support, we can do it in another PR with more improvement, for example create a consumer builder interface for ZmqConsumer and redis based Consumer, and change all Orch ctor code to accept a build as parameter.
However that's need change all orch code, so it's better to put in a seprate PR.

qiluo-msft, line 25: >DashOrchBase [](http://example.com/codeflow?start=6&length=12)

Can we move renaming into another separate PR? Seems like Dash is not an accurate name, because it is not bound  to dash project. Waiting community to provide more feedback on the name.

Other changes are easy to review and merge.

liuh-80: Fixed, rename back to ZmqOrch

prsunny:
@liuh-80 , can you please fix the description of PR to better align with new code? Also, "why" section. Link to any HLD if there is one.

liuh-80:
> @liuh-80 , can you please fix the description of PR to better align with new code? Also, "why" section. Link to any HLD if there is one.

Fixed, Description updated also add HLD link.

orchagent/dash/dashvnetorch.h
```diff
@@ -11,6 +11,7 @@
 #include "ipaddresses.h"
 #include "macaddress.h"
 #include "timer.h"
+#include <zmqorch.h>
```
qiluo-msft, line 14: >< [](http://example.com/codeflow?start=9&length=1)

we prefer `"` for internal header files.

liuh-80: Fixed

orchagent/zmqorch.cpp
```diff
@@ -0,0 +1,65 @@
...
+    else
+    {
+        SWSS_LOG_WARN("ZmqOrch does not support create consumer for db: %d, table: %s", db->getDbId(), tableName.c_str());
+    }
+}
+
+void ZmqOrch::doTask(Consumer &consumer)
+{
+    // When ZMQ disabled, forward data from Consumer
+    doTask((ConsumerBase &)consumer);
```
Pterosaur, line 64: I feel it's better to use `static_cast` instead of "C convention" cast.

liuh-80: Fixed

orchagent/zmqorch.cpp
```diff
@@ -0,0 +1,65 @@
...
+using namespace std;
+
+extern int gBatchSize;
+
+void ZmqConsumer::execute()
+{
+    SWSS_LOG_ENTER();
+
+    size_t update_size = 0;
+    auto table = static_cast<swss::ZmqConsumerStateTable*>(getSelectable());
```
qiluo-msft, line 13: >static_cast [](http://example.com/codeflow?start=17&length=11)

You should use `dynamic_cast` for downcast, and check return result.

orchagent/orch.cpp
```diff
@@ -295,29 +271,6 @@ void Consumer::drain()
...
-        update_size = addToSync(entries);
-    } while (update_size != 0);
-
-    drain();
-}
-
-void ZmqConsumer::drain()
-{
-    if (!m_toSync.empty())
-        ((ZmqOrch *)m_orch)->doTask((ZmqConsumer&)*this);
```
qiluo-msft, line 318: `*this` is already in this class. Is this casting really needed?

orchagent/zmqorch.cpp
```diff
@@ -0,0 +1,65 @@
...
+        update_size = addToSync(entries);
+    } while (update_size != 0);
+
+    drain();
+}
+
+void ZmqConsumer::drain()
+{
+    if (!m_toSync.empty())
+        (static_cast<ZmqOrch*>(m_orch))->doTask(static_cast<ZmqConsumer&>(*this));
```
qiluo-msft, line 27: `*this` is already in this class `ZmqConsumer`. Is this casting really needed?

orchagent/zmqorch.cpp
```diff
@@ -0,0 +1,65 @@
...
+    else
+    {
+        SWSS_LOG_WARN("ZmqOrch does not support create consumer for db: %d, table: %s", db->getDbId(), tableName.c_str());
+    }
+}
+
+void ZmqOrch::doTask(Consumer &consumer)
+{
+    // When ZMQ disabled, forward data from Consumer
+    doTask(static_cast<ConsumerBase &>(consumer));
```
qiluo-msft, line 64: >ConsumerBase [](http://example.com/codeflow?start=23&length=12)

Prefer traditional casting for upcasting for readability.

orchagent/zmqorch.cpp
liuh-80: Fixed

qiluo-msft (approved):

orchagent/zmqorch.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include <vector>
+#include <string>
+#include <orch.h>
+#include "zmqserver.h"
+
+class ZmqConsumer : public ConsumerBase {
```
mint570, line 8: This does not have all the consumer methods. How does warm boot work?
Looks like the current codes just cast the executor to be Consumer in the warmboot logic. It won't work for ZmqConsumer.

liuh-80: I confirmed that currently the Dash no plan for warm-reboot support.

mint570: I have two PRs:
https://github.com/sonic-net/sonic-swss/pull/2857
https://github.com/sonic-net/sonic-swss-common/pull/803

Could you take a look at them please?

PR #2837, opened 2023-06-27
Refactor Orch class to separate recorder implementation

vivekrnv:
**What I did**

1) Removed the recorder implementation from the Orch class and added a new Recorder Interface
2) Removed the exposure of recorder related global variables from the CfgMgrs and Orchagent
3) Writing the Kfv tuple to the Rec file was moved from the Orch class into the ConsumerBase
4) Moved the gBatchSize declaration into the Orch and thereby removing the need to explicitly declare the global variable in CfgMgrs

**Why I did it**

Recorder implementation should not be tied to Orchagent

**How I verified it**

Install the swss on the switch and verified

```
root@r-lionfish-16:/home/admin# show logging | grep Recorder
Jun 23 19:26:15.741815 sonic NOTICE swss#orchagent: :- startRec: SwSS Recorder: Recording started at /var/log/swss/swss.rec
Jun 23 19:29:09.514177 sonic INFO swss#orchagent: :- logfileReopen: SwSS Recorder: LogRotate request handled
```

**Verify Logrotate:**

```
root@r-lionfish-16:/home/admin# ls -l /var/log/swss/
total 2752
-rw-r--r-- 1 root root  999877 Jun 23 19:26 sairedis.rec
-rw-r--r-- 1 root root 1559947 Jun 23 19:11 sairedis.rec.1
-rw-r--r-- 1 root root  110563 Jun 23 19:27 swss.rec
-rw-r--r-- 1 root root  141713 Jun 23 19:16 swss.rec.1


root@r-lionfish-16:/home/admin# logrotate -f /etc/logrotate.conf  (Force Logrotate)

root@r-lionfish-16:/home/admin# ls -l /var/log/swss/
total 1232
-rw-r--r-- 1 root root    140 Jun 23 19:28 sairedis.rec
-rw-r--r-- 1 root root 999877 Jun 23 19:26 sairedis.rec.1
-rw-r--r-- 1 root root 126798 Jun 23 19:11 sairedis.rec.2.gz
-rw-r--r-- 1 root root      0 Jun 23 19:28 swss.rec
-rw-r--r-- 1 root root 110563 Jun 23 19:27 swss.rec.1
-rw-r--r-- 1 root root  13089 Jun 23 19:16 swss.rec.2.gz

<After some time>

root@r-lionfish-16:/home/admin# ls -l /var/log/swss/
total 1476
-rw-r--r-- 1 root root 242404 Jun 23 19:29 sairedis.rec
-rw-r--r-- 1 root root 999877 Jun 23 19:26 sairedis.rec.1
-rw-r--r-- 1 root root 126798 Jun 23 19:11 sairedis.rec.2.gz
-rw-r--r-- 1 root root   1764 Jun 23 19:29 swss.rec
-rw-r--r-- 1 root root 110645 Jun 23 19:29 swss.rec.1
-rw-r--r-- 1 root root  13089 Jun 23 19:16 swss.rec.2.gz
```

**Details if related**


dgsudharsan (dismissed):

dgsudharsan (dismissed):

prsunny:
tagging @qiluo-msft, @yxieca

orchagent/orch.cpp
```diff
@@ -16,12 +13,7 @@
 
 using namespace swss;
 
-extern int gBatchSize;
-
-extern bool gSwssRecord;
-extern ofstream gRecordOfs;
-extern bool gLogRotate;
-extern string gRecordFile;
+int gBatchSize = 0;
```
qiluo-msft, line 16: I see you move the definition of this global variable from main.cpp to orch.cpp. Is it really necessary?

lib/recorder.h
```diff
@@ -0,0 +1,81 @@
...
+    bool isRotate()  { return log_rotate; }
+    std::string getLoc() { return location; }
+    std::string getFile() { return filename; }
+    std::string getName() { return m_name; }
+
+private:
+    bool enable_rec;
+    bool log_rotate;
+    std::string location;
+    std::string filename;
```
qiluo-msft, line 32: >filename [](http://example.com/codeflow?start=16&length=8)

filename -> m_filename

lib/recorder.h
```diff
@@ -0,0 +1,81 @@
...
+
+    /* getters */
+    bool isRecord()  { return enable_rec; }
+    bool isRotate()  { return log_rotate; }
+    std::string getLoc() { return location; }
+    std::string getFile() { return filename; }
+    std::string getName() { return m_name; }
+
+private:
+    bool enable_rec;
```
qiluo-msft, line 29: >enable_rec [](http://example.com/codeflow?start=9&length=10)

Please add verbose documents to new code. For example, what is `m_name`, if it is a name, why you have another `filename`. Seems like `m_name` should be `m_somename`.

lib/recorder.h
```diff
@@ -0,0 +1,81 @@
...
+/* Record Handler for Response Publisher Class */
+class ResPubRec : public RecWriter {
+public:
+    ResPubRec();
+};
+
+class SaiRedisRec : public RecBase {
+public:
+    SaiRedisRec();
+};
```
qiluo-msft, line 65: All 3 classes share the same interfaces. They fit better into 3 objects of class RecWriter.

lib/recorder.h
```diff
@@ -0,0 +1,81 @@
...
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <memory>
+
+namespace swss {
+
+class RecBase {
+public:
+    RecBase() {}
```
kcudnik, line 13: use default keyword {}

lib/recorder.h
```diff
@@ -0,0 +1,81 @@
...
+public:
+    static const std::string DEFAULT_DIR;
+    static const std::string REC_START;
+    static const std::string SWSS_FNAME;
+    static const std::string SAIREDIS_FNAME;
+    static const std::string RESPPUB_FNAME;
+    
+    static std::unique_ptr<SwSSRec> swss;
+    static std::unique_ptr<SaiRedisRec> sairedis;
+    static std::unique_ptr<ResPubRec> respub;
```
kcudnik, line 78: i would prefer to not have those static objects

orchagent/orch.h
```diff
@@ -216,7 +220,7 @@ class Orch
     Orch(swss::DBConnector *db, const std::vector<std::string> &tableNames);
     Orch(swss::DBConnector *db, const std::vector<table_name_with_pri_t> &tableNameWithPri);
     Orch(const std::vector<TableConnector>& tables);
-    virtual ~Orch();
+    virtual ~Orch() {};
```
kcudnik, line 223: use default keyword

orchagent/orch.cpp
vivekrnv: Yes, if not every daemon that inherits from Orch should define this. 

None of the cfgmgrs require the use of gBatchSize var, so it is defined as zero. Only orchagent needs to updated this and thus it is linked with "extern" and update in main.cpp

lib/recorder.h
vivekrnv: And that's why we have a Recorder interface.  

```
/* Interface to access recorder classes */
class Recorder {
public:
    static std::unique_ptr<SwSSRec> swss;
    static std::unique_ptr<SaiRedisRec> sairedis;
    static std::unique_ptr<ResPubRec> respub;
};
```

lib/recorder.h
vivekrnv: What would you recommend? The only other option i could think of is to define a singleton of Recorder and have these as public member objects.

But to access the singleton everytime, we might need to define a method like getInstance() and use it to access the object and their methods, kinds like the logger class under swsscomon

Eg:` Recorder::getInstance().swss.startRec()`

lib/recorder.h
vivekrnv: m_name is just the user readable name of the type of recorder. "SwSS", "Response Publisher" & "SaiRedis". Used  while printing the logs

Eg: 

`SWSS_LOG_NOTICE("%s Recorder: Recording started at %s", getName().c_str(), fname.c_str());`

`Jun 23 19:26:15.741815 sonic NOTICE swss#orchagent: :- startRec: SwSS Recorder: Recording started at /var/log/swss/swss.rec`

lib/recorder.h
kcudnik: you dont need actually singleton, just a instance of recorder in Orch class, if you have more static members, its harder to test

vivekrnv: Defining recorder instance as a member of Orch class will not work. There are multiple orch classes and all write to a same file. 

So, logically the scope of recorder is the daemon/process and not just orch. Even if we can workaround with using an instance of the recorder class per orch, we would still need static members for handing ofstream, path etc. 

Best case scenario is one static Recorder instance and using it as singleton.

kcudnik: ok

vivekrnv:
@kcudnik , @qiluo-msft . I've handled your comments. Please check

liat-grozovik:
@qiluo-msft @kcudnik could you please help to review following comments?

dgsudharsan (approved):

dgsudharsan:
@qiluo-msft Can you please review and signoff?

lib/recorder.h
qiluo-msft: The proposal "3 objects of class RecWriter" allow you to the same thing.

vivekrnv: Having 3 separate objects is flexible since we can easily extend the functionality for any one of these separately, if required.

PR #2848, opened 2023-07-05
Allow NOT_IMPLEMENTED sai return status for availability monitoring API

tjchadaga:
**What I did**
Added SAI_STATUS_NOT_IMPLEMENTED to SAI return status check while fetching resource availability.

**Why I did it**
If resource availability lookup API is not supported for a specific object, the API returns SAI_STATUS_NOT_IMPLEMENTED or SAI_STATUS_NOT_SUPPORTED. These return values do not indicate unavailability of the resource itself and need to be ignored.

**How I verified it**
Everflow tests on 202205 with SAI which returns SAI_STATUS_NOT_IMPLEMENTED

**Details if related**

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -361,7 +361,10 @@ bool MirrorOrch::isHwResourcesAvailable()
     );
     if (status != SAI_STATUS_SUCCESS)
     {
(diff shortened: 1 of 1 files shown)


tjchadaga:
@saksarav-nokia - please help review as well

saksarav-nokia (approved):

prsunny (approved):

PR #2853, opened 2023-07-09
Remove system neigh DEL operation if SET operation succeeds

ysmanman:
**What I did**
Remove system neighbor DEL operation in m_toSync if SET operation for the same neighbor succeeds. This is to avoid mistakenly removing the system neighbor.

**Why I did it**
Fix https://github.com/sonic-net/sonic-buildimage/issues/15266

**How I verified it**
https://github.com/sonic-net/sonic-buildimage/issues/15266 was uncovered by https://github.com/sonic-net/sonic-mgmt/blob/master/tests/voq/test_voq_disrupts.py.  The issue was not seen in the test anymore with the fix.

**Details if related**

### DIFF
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -1420,6 +1420,17 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
                     {
                         SWSS_LOG_NOTICE("VOQ encap index updated for neighbor %s", kfvKey(t).c_str());
                         it = consumer.m_toSync.erase(it);
+
+                       /* Remove remaining DEL operation in m_toSync for the same neighbor.
+                        * Since DEL operation is supposed to be executed before SET for the same neighbor
+                        * A remaining DEL after the SET operation means the DEL operation failed previously and should not be executed anymore
+                        */
+                       auto rit = make_reverse_iterator(it);
+                       while (rit != consumer.m_toSync.rend() && rit->first == key && kfvOp(rit->second) == DEL_COMMAND)
+                       {
+                           consumer.m_toSync.erase(next(rit).base());
+                           SWSS_LOG_NOTICE("Removed pending system neighbor DEL operation for %s after SET operation", key.c_str());
+                       }
                     }
                     continue;
                 }
@@ -1481,6 +1492,7 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
                 else
                 {
                     it++;
+                    continue;
                 }
             }
             else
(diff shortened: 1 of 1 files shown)


ysmanman:
Add @arlakshm @kenneth-arista for visibility.

arlakshm (approved):
lgtm

kenneth-arista:
This fix will likely also address https://github.com/sonic-net/sonic-mgmt/issues/8588

arlakshm:
@ysmanman, can you rebase the branch?

ysmanman:
> @ysmanman, can you rebase the branch?

Hi @arlakshm rebased just now.

StormLiangMS:
@prsunny  could we have an ADO to track this?

prsunny:
> @prsunny could we have an ADO to track this?

@arlakshm, do you have one?

arlakshm:
> > @prsunny could we have an ADO to track this?
> 
> @arlakshm, do you have one?

@StormLiangMS @prsunny  this is the ADO# 24596307

orchagent/neighorch.cpp
```diff
@@ -1420,6 +1420,17 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
...
                         SWSS_LOG_NOTICE("VOQ encap index updated for neighbor %s", kfvKey(t).c_str());
                         it = consumer.m_toSync.erase(it);
+
+                       /* Remove remaining DEL operation in m_toSync for the same neighbor.
+                        * Since DEL operation is supposed to be executed before SET for the same neighbor
+                        * A remaining DEL after the SET operation means the DEL operation failed previously and should not be executed anymore
+                        */
+                       auto rit = make_reverse_iterator(it);
+                       while (rit != consumer.m_toSync.rend() && rit->first == key && kfvOp(rit->second) == DEL_COMMAND)
+                       {
```
lguohan, line 1430: i thought m_toSync will only have one entry for each neighbor, i didn't know if we have multiple entry for the same neighbor, just curious.

orchagent/neighorch.cpp
```diff
@@ -1420,6 +1420,17 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
                     {
                         SWSS_LOG_NOTICE("VOQ encap index updated for neighbor %s", kfvKey(t).c_str());
                         it = consumer.m_toSync.erase(it);
+
+                       /* Remove remaining DEL operation in m_toSync for the same neighbor.
+                        * Since DEL operation is supposed to be executed before SET for the same neighbor
+                        * A remaining DEL after the SET operation means the DEL operation failed previously and should not be executed anymore
+                        */
+                       auto rit = make_reverse_iterator(it);
```
lguohan, line 1428: sign, we should have tab space aligned. @arlakshm , @ysmanman . I feel being insulted. @prsunny

orchagent/neighorch.cpp
lguohan: why this problem is voq neighbor specificially? for normal neighbor, why we do not have such problem?

ysmanman: > i thought m_toSync will only have one entry for each neighbor, i didn't know if we have multiple entry for the same neighbor, just curious.

m_toSync is type of std::multimap, so it allows multiple actions (like DEL and SET) for the same neighbor.

ysmanman: > why this problem is voq neighbor specificially? for normal neighbor, why we do not have such problem?

This problem is not voq specific. The same change/fix exists for local neighbor too. The PR is to apply the same fix for voq.

orchagent/neighorch.cpp
ysmanman: > sigh, we should have tab space aligned. @arlakshm , @ysmanman . I feel being insulted. @prsunny

I can raise a PR to fix the indentation.

PR #2855, opened 2023-07-12
[Chassis][orchagent] : Support WRED profiles on system ports

vmittal-msft:
**What I did**
Added support to apply wred profile on system ports

**Why I did it**
remote system ports were missing wred profile

**How I verified it**
Tested on sonic chassis

**Details if related**

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -20,6 +20,8 @@ extern PortsOrch *gPortsOrch;
 extern Directory<Orch*> gDirectory;
 extern sai_object_id_t gSwitchId;
 extern string gMySwitchType;
+extern string gMyHostName;
+extern string gMyAsicName;
 
 #define BUFFER_POOL_WATERMARK_FLEX_STAT_COUNTER_POLL_MSECS  "60000"
 
@@ -788,6 +790,8 @@ task_process_status BufferOrch::processQueue(KeyOpFieldsValuesTuple &tuple)
     vector<string> tokens;
     sai_uint32_t range_low, range_high;
     bool need_update_sai = true;
+    bool local_port = false;
+    string local_port_name;
 
     SWSS_LOG_DEBUG("Processing:%s", key.c_str());
     tokens = tokenize(key, delimiter);
@@ -805,6 +809,13 @@ task_process_status BufferOrch::processQueue(KeyOpFieldsValuesTuple &tuple)
         if (!parseIndexRange(tokens[3], range_low, range_high))
         {
             return task_process_status::task_invalid_entry;
+        }
+
+        if(tokens[0] == gMyHostName)
+        {
+           local_port = true;
+           local_port_name = tokens[2];
+           SWSS_LOG_INFO("System port %s is local port %d local port name %s", port_names[0].c_str(), local_port, local_port_name.c_str());
         }
     }
     else 
@@ -883,7 +894,13 @@ task_process_status BufferOrch::processQueue(KeyOpFieldsValuesTuple &tuple)
     for (string port_name : port_names)
     {
         Port port;
-        SWSS_LOG_DEBUG("processing port:%s", port_name.c_str());
+        SWSS_LOG_ERROR("processing port:%s", port_name.c_str());
+
+        if(local_port == true)
+        {
+            port_name = local_port_name;
+        }
+
         if (!gPortsOrch->getPort(port_name, port))
         {
             SWSS_LOG_ERROR("Port with alias:%s not found", port_name.c_str());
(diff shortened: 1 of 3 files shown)


gechiang:
@vmittal-msft , please fix test failure

gechiang:
@vmittal-msft can you fix the test failure?

vmittal-msft:
@gechiang  these failures are not related to my changes. I am trying to re-run the build but it is still failing.

arlakshm (commented):
please check if we can add UTs

orchagent/bufferorch.cpp
```diff
@@ -1001,8 +1018,17 @@ task_process_status BufferOrch::processQueue(KeyOpFieldsValuesTuple &tuple)
         // set order is detected.
         for (const auto &port_name : port_names)
         {
-            if (gPortsOrch->isPortAdminUp(port_name)) {
-                SWSS_LOG_WARN("Queue profile '%s' applied after port %s is up", key.c_str(), port_name.c_str());
+            if(local_port == true)
+            {
+                if (gPortsOrch->isPortAdminUp(local_port_name)) {
+                    SWSS_LOG_WARN("Queue profile '%s' applied after port %s is up", key.c_str(), port_name.c_str());
```
gechiang, line 1024: should this be changed to use local_port_name.c_str() instead of port_name.c_str()?
Of for local_port is true case, port_name is always the same as local_port_name?

orchagent/qosorch.cpp
```diff
@@ -1652,7 +1691,7 @@ bool QosOrch::applySchedulerToQueueSchedulerGroup(Port &port, size_t queue_ind,
     sai_status = sai_scheduler_group_api->set_scheduler_group_attribute(group_id, &attr);
     if (SAI_STATUS_SUCCESS != sai_status)
     {
-        SWSS_LOG_ERROR("Failed applying scheduler profile:0x%" PRIx64 " to scheduler group:0x%" PRIx64 ", port:%s", scheduler_profile_id, group_id, port.m_alias.c_str());
+         SWSS_LOG_ERROR("Failed applying scheduler profile:0x%" PRIx64 " to scheduler group:0x%" PRIx64 ", port:%s", scheduler_profile_id, group_id, port.m_alias.c_str());
```
gechiang, line 1694: can you fix the extra indentation on this line since you are not chaging the logged message anyway?

orchagent/bufferorch.cpp
```diff
@@ -1001,8 +1018,17 @@ task_process_status BufferOrch::processQueue(KeyOpFieldsValuesTuple &tuple)
         // set order is detected.
         for (const auto &port_name : port_names)
         {
-            if (gPortsOrch->isPortAdminUp(port_name)) {
-                SWSS_LOG_WARN("Queue profile '%s' applied after port %s is up", key.c_str(), port_name.c_str());
+            if(local_port == true)
```
neethajohn, line 1021: why is this check needed? in L899, this check is already done and port_name is assigned as the local_port_name

orchagent/bufferorch.cpp
vmittal-msft: i would like to keep port_name in the debug message since local_port_name will be subset of port_name.

orchagent/qosorch.cpp
vmittal-msft: Sure. Will do.

orchagent/bufferorch.cpp
vmittal-msft: we are looping through port_names here hence port_name is getting assigned back to original name. Alos, during testing, i did see this behavior as well.

arlakshm (approved):
lgtm

orchagent/bufferorch.cpp
```diff
@@ -1001,8 +1018,17 @@ task_process_status BufferOrch::processQueue(KeyOpFieldsValuesTuple &tuple)
...
+            {
+                if (gPortsOrch->isPortAdminUp(local_port_name)) {
+                    SWSS_LOG_WARN("Queue profile '%s' applied after port %s is up", key.c_str(), port_name.c_str());
+                }
+            }
+            else
+            {
+                if (gPortsOrch->isPortAdminUp(port_name)) {
+                    SWSS_LOG_WARN("Queue profile '%s' applied after port %s is up", key.c_str(), port_name.c_str());
+                }
```
kenneth-arista, line 1031: if `local_port` is true, is it necessary to iterate over all port names since `gPortsOrch->isPortAdminUp()` is always passed `local_port_name`?

vmittal-msft: I have seen port_names carries only one port_name so it doesn't matter.

gechiang:
MS ADO: 24595422

lguohan:
@gechiang , @vmittal-msft , @prsunny , why lots of code changes, but no single line of unit test added?

vmittal-msft:
@lguohan i will work on the unit case for this change. Sorry for the miss in initial commit.

gechiang:
@lguohan My bad when I review but missed to check the missing UT check that is required and approved/merged it prematurely.
I have already communicated this with @vmittal-msft and @yxieca that another PR for the missing UT will be made and at that point this PR will then be approved together with the new UT PR to be cherry-picked to other public branch such as 202205.
Sorry for the miss.

PR #2857, opened 2023-07-12
[dash] Add refillToSync() into ConsumerBase to support warmboot.

mint570:
Add refillToSync() into ConsumerBase to support warmboot.

**What I did**
Add warmboot support for zmq consumer.

**Why I did it**
This will be useful when we migrate other orchs to use zmq.

**How I verified it**
UT

**Details if related**

### DIFF
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -176,7 +176,7 @@ size_t ConsumerBase::addToSync(const std::deque<KeyOpFieldsValuesTuple> &entries
 }
 
 // TODO: Table should be const
-size_t Consumer::refillToSync(Table* table)
+size_t ConsumerBase::refillToSync(Table* table)
 {
     std::deque<KeyOpFieldsValuesTuple> entries;
     vector<string> keys;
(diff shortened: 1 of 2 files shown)


prsunny:
Why is this PR raised to 'dash' branch?

mint570:
The dash branch has this commit: https://github.com/sonic-net/sonic-swss/pull/2836
If we cherry-pick that commit into the main branch, I can move this PR to main branch as well.

orchagent/orch.cpp
```diff
@@ -214,14 +215,22 @@ size_t Consumer::refillToSync()
...
-        auto db = getDbConnector();
+        auto db = consumertable->getDbConnector();
         string tableName = getTableName();
         auto table = Table(db, tableName);
         return refillToSync(&table);
     }
+    else
+    {
+        // The consumerTable should be ZmqConsumerStateTable.
+        DBConnector db("APPL_DB", 0);
```
liuh-80, line 229: The ZmqConsumer is designed for Dash scenario, we still under discussion about if we need put the APPL_DB on remote side. if the DB on host, then code here need connect to a remote Redis instance.

I confirmed that the Dash has no plan to support warm-reboot, so we may need wait Dash finalize design then decide how to implement the refill.

mint570:
I created a new PR in the master branch.
https://github.com/sonic-net/sonic-swss/pull/2866

PR #2859, opened 2023-07-14
[intfmgr]: add interface arp timeout task

sempSONiC1:
[intfmgr]: add interface arp timeout task

### DIFF
--- a/cfgmgr/intfmgr.cpp
+++ b/cfgmgr/intfmgr.cpp
@@ -632,6 +632,18 @@ bool IntfMgr::setIntfProxyArp(const string &alias, const string &proxy_arp)
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


prsunny:
Please provide the motivation of this PR. Also it seems like a new config_db attribute is introduced, do we have the yang defined and approved by Yang community? @dgsudharsan for viz.

sempSONiC1:
> Please provide the motivation of this PR. Also it seems like a new config_db attribute is introduced, do we have the yang defined and approved by Yang community? @dgsudharsan for viz.

We add this feature in recent project, maybe it's a general requirement. The yang definition does not pull yet.

dgsudharsan (changes requested):
Please add UT for the changes.
Please submit yang model. It is a must for any configuration schema change

PR #2860, opened 2023-07-17
[Mellanox] Fix error in peer response time when headroom is calculated for 800G

stephenxs:
**What I did**

Fix an error in peer response time when headroom is calculated for 800G

**Why I did it**

Bug fix.

**How I verified it**

Manually test

**Details if related**

### DIFF
--- a/cfgmgr/buffer_headroom_mellanox.lua
+++ b/cfgmgr/buffer_headroom_mellanox.lua
@@ -39,6 +39,7 @@ local ret = {}
 -- the key of table pause_quanta_per_speed is operating speed at Mb/s
 -- the value of table pause_quanta_per_speed is the number of pause_quanta
 local pause_quanta_per_speed = {}
+pause_quanta_per_speed[800000] = 905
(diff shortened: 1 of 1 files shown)


keboliu (approved):

cfgmgr/buffer_headroom_mellanox.lua
```diff
@@ -39,6 +39,7 @@ local ret = {}
 -- the key of table pause_quanta_per_speed is operating speed at Mb/s
 -- the value of table pause_quanta_per_speed is the number of pause_quanta
 local pause_quanta_per_speed = {}
+pause_quanta_per_speed[800000] = 905
```
neethajohn, line 42: how is this the same as for 400G?

stephenxs: Yes, this is defined by the IEEE. 400G and 800G share the same value.

neethajohn (approved):

StormLiangMS:
@neethajohn do we have ADO for cherry-pick?

PR #2866, opened 2023-07-24
Add refillToSync() into ConsumerBase to support warmboot.

mint570:
**What I did**
Add refillToSync() into ConsumerBase to support warmboot.

**Why I did it**
Add warmboot support for zmq consumer.

**How I verified it**
This will be useful when we migrate other orchs to use zmq.

**Details if related**
UT

### DIFF
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -156,7 +156,7 @@ size_t ConsumerBase::addToSync(const std::deque<KeyOpFieldsValuesTuple> &entries
 }
 
 // TODO: Table should be const
-size_t Consumer::refillToSync(Table* table)
+size_t ConsumerBase::refillToSync(Table* table)
 {
     std::deque<KeyOpFieldsValuesTuple> entries;
     vector<string> keys;
@@ -178,7 +178,7 @@ size_t Consumer::refillToSync(Table* table)
     return addToSync(entries);
 }
 
-size_t Consumer::refillToSync()
+size_t ConsumerBase::refillToSync()
 {
     auto subTable = dynamic_cast<SubscriberStateTable *>(getSelectable());
     if (subTable != NULL)
@@ -194,14 +194,23 @@ size_t Consumer::refillToSync()
         } while (update_size != 0);
         return total_size;
     }
-    else
+    string tableName = getTableName();
+    auto consumerTable = dynamic_cast<ConsumerTableBase *>(getSelectable());
+    if (consumerTable != NULL)
     {
         // consumerTable is either ConsumerStateTable or ConsumerTable
-        auto db = getDbConnector();
-        string tableName = getTableName();
+        auto db = consumerTable->getDbConnector();
         auto table = Table(db, tableName);
         return refillToSync(&table);
     }
+    auto zmqTable = dynamic_cast<ZmqConsumerStateTable *>(getSelectable());
+    if (zmqTable != NULL)
+    {
+        auto db = zmqTable->getDbConnector();
+        auto table = Table(db, tableName);
+        return refillToSync(&table);
+    }
+    return 0;
 }
 
 string ConsumerBase::dumpTuple(const KeyOpFieldsValuesTuple &tuple)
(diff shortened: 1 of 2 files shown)


lguohan:
@liuh-80 , i see you comments in #2857 , as @mint570 said, it would be useful to migrate other orchs to zmq. do you see any downside of this code? if it doesn't hurt existing scenario, we should make our code more generic so that people can build other scenario around our code?

liuh-80:
> @liuh-80 , i see you comments in #2857 , as @mint570 said, it would be useful to migrate other orchs to zmq. do you see any downside of this code? if it doesn't hurt existing scenario, we should make our code more generic so that people can build other scenario around our code?

There is no downside of add refillToSync(), I will review this PR and give comments.

orchagent/orch.cpp
```diff
@@ -172,9 +172,10 @@ size_t Consumer::refillToSync(Table* table)
     return addToSync(entries);
 }
 
-size_t Consumer::refillToSync()
+size_t ConsumerBase::refillToSync()
 {
     auto subTable = dynamic_cast<SubscriberStateTable *>(getSelectable());
+    auto consumertable = dynamic_cast<ConsumerTableBase *>(getSelectable());
```
liuh-80, line 178: Suggest move this to line 192 to avoid unnecessary convert, also because first if block returned, so line 192 can change from 'else if' to 'if'

orchagent/orch.cpp
```diff
@@ -188,14 +189,22 @@ size_t Consumer::refillToSync()
...
-        auto db = getDbConnector();
+        auto db = consumertable->getDbConnector();
         string tableName = getTableName();
         auto table = Table(db, tableName);
         return refillToSync(&table);
     }
+    else
+    {
+        // The consumerTable should be ZmqConsumerStateTable.
+        DBConnector db("APPL_DB", 0);
```
liuh-80, line 203: Please check if the consumerTable is ZmqConsumerStateTable and use getDbConnector() in ZmqConsumerStateTable:

https://github.com/sonic-net/sonic-swss-common/blob/5b6377cd260e5421470263ebd7e27ac8aca42c52/common/zmqconsumerstatetable.h

The reason of this is because the APPL_DB may not in local and may not be 'APPL_DB', the design still under discussion, so if use DBConnector from ZMQ, the code can keep no change when design finalized.

orchagent/orch.cpp
mint570: Done

orchagent/orch.cpp
mint570: Done

liuh-80 (approved):

orchagent/orch.cpp
```diff
@@ -194,14 +194,25 @@ size_t Consumer::refillToSync()
...
+    auto zmqTable = dynamic_cast<ZmqConsumerStateTable *>(getSelectable());
+    if (zmqTable != NULL)
+    {
+        auto db = zmqTable->getDbConnector();
         auto table = Table(db, tableName);
         return refillToSync(&table);
     }
+    DBConnector db("APPL_DB", 0);
+    auto table = Table(&db, tableName);
+    return refillToSync(&table);
```
qiluo-msft, line 215: Why need this block? What is the class of `getSelectable()`? APPL_DB seems a magic choice.

mint570: This is the default case if none of the above condition matches. Today, orchagent reads the entries from APPL DB after warm boot. I hear dash is different, but using APPL DB shouldn't be completely unreasonable.

Do you suggest to remove this default case handling? I can do that.

prsunny:
@qiluo-msft to approve

orchagent/orch.cpp
qiluo-msft: I do not get "the class of getSelectable()?" in this case. If you do not either, let's remove this block.

mint570: Done

qiluo-msft (approved):

mint570:
The vs test failed. Not sure if it relates to this PR or not.

prsunny:
> The vs test failed. Not sure if it relates to this PR or not.

no, there is a known issue with submodule PR for sai-redis. will merge this by next week.

PR #2870, opened 2023-07-30
Make sure new binaries replace existing binaries in docker-sonic-vs

saiarcot895:
**What I did**

PR #2717 removed a step where existing deb packages were removed before installing the new deb packages. The problem is that due to Docker's change detection code, where it only checks the file size and the modification timestamp, even if the file gets modified when installing a newer debian package, the file size may remain the same, and the modification timestamp will remain the same (since this is based on the debian/changelog timestamp). This caused changed binaries to not actually replace the existing binaries.

Fix this by re-adding the line that removed existing packages first.

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/.azure-pipelines/docker-sonic-vs/Dockerfile
+++ b/.azure-pipelines/docker-sonic-vs/Dockerfile
@@ -5,6 +5,11 @@ ARG need_dbg
 
 COPY ["debs", "/debs"]
 
+# Remove existing packages first before installing the new/current packages. This is to overcome limitations with
+# Docker's diff detection mechanism, where only the file size and the modification timestamp (which will remain the
+# same, even though contents have changed) are checked between the previous and current layer.
+RUN dpkg --purge libswsscommon python3-swsscommon sonic-db-cli libsaimetadata libsairedis libsaivs syncd-vs swss sonic-eventd
+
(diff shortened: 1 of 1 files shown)


.azure-pipelines/docker-sonic-vs/Dockerfile
```diff
@@ -5,6 +5,11 @@ ARG need_dbg
 
 COPY ["debs", "/debs"]
 
+# Remove existing packages first before installing the new/current packages. This is to overcome limitations with
+# Docker's diff detection mechanism, where only the file size and the modification timestamp (which will remain the
+# same, even though contents have changed) are checked between the previous and current layer.
+RUN dpkg --purge libswsscommon python3-swsscommon sonic-db-cli libsaimetadata libsairedis libsaivs syncd-vs swss sonic-eventd
```
marian-pritsak, line 11: What about swss-dbg from line 22?

saiarcot895: Debug packages aren't installed in the docker-sonic-vs container that's built as part of sonic-buildimage, so that doesn't need to be removed.

lguohan (approved):

PR #2890, opened 2023-08-29
[Buffer Orch] Retry one more time when it fails to set buffer profiles' attributes to SAI

stephenxs:
**What I did**

Retry one more time when it fails to set buffer profiles' attributes to SAI

**Why I did it**

Vendor SAI can perform a sanity check on all the attributes when an attribute is modified or a new object is created.
However, while the user or buffer manager daemon passes all new attributes in an (atomic) operation, there can be only one attribute in each SAI SET operation, which means the vendor SAI does not understand the whole information of the new attributes and can only perform the sanity check against each SET operation.
In some rare scenarios, the sanity check can fail due to one single attribute not being compatible with the old attributes set but being compatible with the new attributes set. In this case, all new attributes will be applied to SAI successfully by retrying one more time.

**How I verified it**

**Details if related**
An example:
A vendor SAI requires `xon` + `xoff` <= `size` when the shared headroom pool is disabled.
If a user wants to decrease the values on the fly, eg. smaller `new size` and `new xoff` will be applied where `new xoff` + `xon` <= `new size`. If `new size` is set to SAI first, the sanity check can fail because the `new size` can be less than the sum of `xon` and `old xoff` but `new xoff` will still be set to SAI successfully because `new xoff` + `old xon` < `old size`
In this case, `size` will be set to SAI successfully after retry because the `new xoff` has been set to SAI and `new xoff` + `xon` <= `new size`

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -704,6 +704,7 @@ task_process_status BufferOrch::processBufferProfile(KeyOpFieldsValuesTuple &tup
         }
         if (SAI_NULL_OBJECT_ID != sai_object)
         {
+            vector<sai_attribute_t> attribs_to_retry;
             SWSS_LOG_DEBUG("Modifying existing sai object:%" PRIx64, sai_object);
             for (auto &attribute : attribs)
             {
@@ -715,7 +716,18 @@ task_process_status BufferOrch::processBufferProfile(KeyOpFieldsValuesTuple &tup
                 }
                 else if (SAI_STATUS_SUCCESS != sai_status)
                 {
-                    SWSS_LOG_ERROR("Failed to modify buffer profile, name:%s, sai object:%" PRIx64 ", status:%d", object_name.c_str(), sai_object, sai_status);
+                    SWSS_LOG_NOTICE("Unable to modify buffer profile, name:%s, sai object:%" PRIx64 ", status:%d, will retry one more time", object_name.c_str(), sai_object, sai_status);
+                    attribs_to_retry.push_back(attribute);
+                }
+            }
+
+            for (auto &attribute : attribs)
+            {
+                sai_status = sai_buffer_api->set_buffer_profile_attribute(sai_object, &attribute);
+                if (SAI_STATUS_SUCCESS != sai_status)
+                {
+                    // A retried attribute can not be "not implemented"
+                    SWSS_LOG_ERROR("Failed to modify buffer profile, name:%s, sai object:%" PRIx64 ", status:%d, will retry once", object_name.c_str(), sai_object, sai_status);
                     task_process_status handle_status = handleSaiSetStatus(SAI_API_BUFFER, sai_status);
                     if (handle_status != task_process_status::task_success)
                     {
--- a/tests/mock_tests/bufferorch_ut.cpp
+++ b/tests/mock_tests/bufferorch_ut.cpp
@@ -64,6 +64,53 @@ namespace bufferorch_test
         return pold_sai_buffer_api->set_ingress_priority_group_attribute(ingress_priority_group_id, attr);
     }
 
+    sai_uint64_t _ut_stub_buffer_profile_size;
+    sai_uint64_t _ut_stub_buffer_profile_xon;
+    sai_uint64_t _ut_stub_buffer_profile_xoff;
+    bool _ut_stub_buffer_profile_sanity_check = false;
+    sai_status_t _ut_stub_sai_set_buffer_profile_attribute(
+        _In_ sai_object_id_t buffer_profile_id,
+        _In_ const sai_attribute_t *attr)
+    {
+        if (_ut_stub_buffer_profile_sanity_check)
+        {
+            if (SAI_BUFFER_PROFILE_ATTR_BUFFER_SIZE == attr[0].id)
+            {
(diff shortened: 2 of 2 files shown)


liat-grozovik:
@prsunny could you please help to review or assign someone?

prsunny:
I dont think this must be fixed in orchagent but SAI implementation should handle such cases. @neethajohn , @vmittal-msft , can you take a look and comment?

stephenxs:
> I dont think this must be fixed in orchagent but SAI implementation should handle such cases. @neethajohn , @vmittal-msft , can you take a look and comment?

Hi @prsunny 
Theoretically, the problem is that OA passes only one attribute when calling the SAI set API.
When vendor SAI handles the set API, it cannot look forward and assume what the next attributes are. It can just perform a sanity check with the existing attributes + the new attribute. In case there is an issue, it should not apply the attribute to SDK.
Neither can it store the attribute for further use because it incurs inconsistency between OA and SAI/SDK.
Furthermore, it can not distinguish this case and another case which is OA calls SAI set API with a really wrong attribute. In the second case, SAI should return an error.

orchagent/bufferorch.cpp
```diff
@@ -715,7 +716,18 @@ task_process_status BufferOrch::processBufferProfile(KeyOpFieldsValuesTuple &tup
                 }
                 else if (SAI_STATUS_SUCCESS != sai_status)
                 {
-                    SWSS_LOG_ERROR("Failed to modify buffer profile, name:%s, sai object:%" PRIx64 ", status:%d", object_name.c_str(), sai_object, sai_status);
+                    SWSS_LOG_NOTICE("Unable to modify buffer profile, name:%s, sai object:%" PRIx64 ", status:%d, will retry one more time", object_name.c_str(), sai_object, sai_status);
+                    attribs_to_retry.push_back(attribute);
+                }
+            }
+
+            for (auto &attribute : attribs)
```
neethajohn, line 724: This retry logic is needed for specific buffer profile attributes. Instead of just retrying blindly, why can't we handle this combination (xoff, xon and size) the same way as we did for  wred attributes - min and max (https://github.com/sonic-net/sonic-swss/blob/master/orchagent/qosorch.cpp#L595)

stephenxs: Hi @neethajohn 
Unlike the wred/ecn profile which has an explicit logic shared by all vendors, the buffer profile logic can differ among various vendors. In this sense, we can’t find a common logic but prefer to retry here and leave the sanity check to the vendor SAI.

neethajohn (approved):

neethajohn:
@prsunny , any comments?

stephenxs:
FYI.
Without the retry logic, the UT failed as expected
```
Thread 1 "tests" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff753d537 in __GI_abort () at abort.c:79
#2  0x00005555555d3fa6 in handleSaiFailure (abort_on_failure=<optimized out>) at ../../orchagent/saihelper.cpp:768
#3  0x00005555559ae5bc in handleSaiSetStatus (api=api@entry=SAI_API_BUFFER, status=-9456, status@entry=-5, context=context@entry=0x0) at ../../orchagent/saihelper.cpp:630
#4  0x000055555593d177 in BufferOrch::processBufferProfile (this=0x5555561a3e30, tuple=std::tuple containing = {...}) at ../../orchagent/bufferorch.cpp:715
#5  0x0000555555938044 in BufferOrch::doTask (this=0x5555561a3e30, consumer=...) at ../../orchagent/bufferorch.cpp:1453
#6  0x000055555593688f in BufferOrch::doTask (this=0x5555561a3e30) at ../../orchagent/bufferorch.cpp:1410
#7  0x000055555573eca3 in bufferorch_test::BufferOrchTest_BufferOrchTestSetBufferProfile_Test::TestBody (this=<optimized out>) at bufferorch_ut.cpp:769
#8  0x0000555555c3fce7 in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#9  0x0000555555c35f7e in testing::Test::Run() ()
#10 0x0000555555c360d5 in testing::TestInfo::Run() ()
#11 0x0000555555c36569 in testing::TestSuite::Run() ()
#12 0x0000555555c36bb2 in testing::internal::UnitTestImpl::RunAllTests() ()
#13 0x0000555555c40257 in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::Unit\
TestImpl::*)(), char const*) ()
#14 0x0000555555c36198 in testing::UnitTest::Run() ()
#15 0x00005555556b21d0 in main ()
```

StormLiangMS:
This is an enhancement, should we take this to older branches? @prsunny  @neethajohn

stephenxs:
> This is an enhancement, should we take this to older branches? @prsunny @neethajohn

Hi @StormLiangMS 
This is a bug fix instead of an enhancement. Warm reboot upgrade test can fail if the switch is upgraded from 202012 to 202211 with dynamic buffer model because lossless buffer profile is upgraded between branches.

PR #2893, opened 2023-09-02
[FEC]Auto FEC initial changes

dgsudharsan:
**What I did**
Initial Changes to support Auto FEC feature. Added support for mode "auto" in FEC

**Why I did it**
To implement auto FEC feature HLD: https://github.com/sonic-net/SONiC/pull/1416

**How I verified it**
Added UT to verify. Additionally performed Manual testing.

**Details if related**


Junchao-Mellanox (approved):

dgsudharsan:
/azpw.run Azure.sonic-swss

ridahanif96:
@dgsudharsan please add me as reviewer

ridahanif96 (approved):

prsunny:
@prgeor to signoff

prgeor:
> @prgeor to signoff

ack

prgeor:
@dgsudharsan can you update the PR description to point to the HLD?

orchagent/port/porthlpr.h
```diff
@@ -6,6 +6,22 @@
 #include <string>
 
 #include "portcnt.h"
+#include "portschema.h"
```
nazariig, line 9: @dgsudharsan the original intention of `portsorch` refactoring was to hide the properties parsing metadata from the mainline code. I do recommend to consider a different approach

orchagent/port/porthlpr.h
```diff
@@ -6,6 +6,22 @@
...
+
+static const std::unordered_map<std::string, sai_port_fec_mode_t> portFecMap =
+{
+    { PORT_FEC_NONE, SAI_PORT_FEC_MODE_NONE },
+    { PORT_FEC_RS,   SAI_PORT_FEC_MODE_RS   },
+    { PORT_FEC_FC,   SAI_PORT_FEC_MODE_FC   },
+    { PORT_FEC_AUTO, SAI_PORT_FEC_MODE_NONE }
```
nazariig, line 16: @dgsudharsan imho, this trick is used just to overcome properties parsing issue. In fact, this has nothing to do with a set of FEC configuration attributes. I understand that FEC override is tightly coupled with a FEC configuration itself, and i am ok to combine this one with the rest in terms of Config DB schema, but parsing/analysis logic should use a separate containers to store the data

orchagent/portsorch.cpp
```diff
@@ -2357,10 +2343,19 @@ void PortsOrch::initPortCapLinkTraining(Port &port)
     SWSS_LOG_WARN("Unable to get %s LT support capability", port.m_alias.c_str());
 }
 
-bool PortsOrch::isFecModeSupported(const Port &port, sai_port_fec_mode_t fec_mode)
+bool PortsOrch::isFecModeSupported(const Port &port, sai_port_fec_mode_t fec_mode, bool is_auto)
```
nazariig, line 2346: @dgsudharsan for FEC override you do have a separate capability. Why not to have a separate method to check whether this mode is supported or not? It can be done while processing port configuration attributes in the main loop, while having a distinction based on `autoneg` configuration

orchagent/portsorch.h
```diff
@@ -379,7 +380,7 @@ class PortsOrch : public Orch, public Subject
     bool setPortTpid(Port &port, sai_uint16_t tpid);
     bool setPortPvid (Port &port, sai_uint32_t pvid);
     bool getPortPvid(Port &port, sai_uint32_t &pvid);
-    bool setPortFec(Port &port, sai_port_fec_mode_t fec_mode);
+    bool setPortFec(Port &port, sai_port_fec_mode_t fec_mode, bool auto_fec);
```
nazariig, line 383: @dgsudharsan this is really confusing. Setting FEC auto is only possible while passing a FEC NONE configuration, which has nothing to do with the override action itself. Suggest considering a separate method to operate exclusively on the FEC override attribute

nazariig (changes requested):

orchagent/portsorch.h
dgsudharsan: @nazariig auto fec is a different mode compared to none. There are four options in the CLI - none, rs,fc, auto.
When auto is configured SAI needs to be programmed with SAI_PORT_FEC_ATTR_NONE and override is false. However since none is shared between two modes, we need a flag to differentiate whether it is auto or none mode from CLI.
Previously this API used to accept the modes as string but due to the recent refactoring the parameter is changed to sai_port_fec_mode_t . Either we undo the refactoring or introduce a new flag.

orchagent/port/porthlpr.cpp
```diff
@@ -471,6 +457,11 @@ bool PortHelper::parsePortFec(PortConfig &port, const std::string &field, const
     port.fec.value = cit->second;
     port.fec.is_set = true;
 
+    if (value == PORT_FEC_AUTO)
```
nazariig, line 460: @dgsudharsan more specific config values should to be processed first and return immediately leaving the rest of the logic behind

nazariig (changes requested):

orchagent/portsorch.h
nazariig: @dgsudharsan this is what we have in the spec:
```
    /**
     * @brief FEC mode auto-negotiation override status
     *
     * If set to true, any auto-negotiated FEC mode will be
     * overridden by the value configured in SAI_PORT_ATTR_FEC_MODE
     *
     * @type bool
     * @flags CREATE_AND_SET
     * @default false
     */
    SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE,
```

orchagent/portsorch.cpp
dgsudharsan: @nazariig  The FEC override configuration is not related to autoneg. We need to have a single place to check if a specific FEC mode is supported. Due to refactoring this method parameter has changed from string to sai_port_fec_mode_t. We need a flag to differentiate auto vs none.

orchagent/portsorch.h
dgsudharsan: @nazariig  Please refer to the HLD https://github.com/sonic-net/SONiC/pull/1416.

orchagent/portsorch.cpp
```diff
@@ -2357,10 +2343,19 @@ void PortsOrch::initPortCapLinkTraining(Port &port)
...
 
+    if (!override_fec && !fec_override_sup)
+    {
+        return false;
+    }
+
+    if (!override_fec && !port.m_autoneg)
+    {
+        SWSS_LOG_NOTICE("Autoneg must be enabled for port fec mode auto to work");
+    }
```
nazariig, line 2358: @dgsudharsan i would suggest the following:
```
// override_fec - means that user set fec=auto in Config DB
if (override_fec)
{
    if (!fec_override_sup)
    {
        return false;
    }

    if (!port.m_autoneg)
    {
        SWSS_LOG_NOTICE("Autoneg must be enabled for port fec mode auto to work");
    }
}
```

nazariig (approved):

orchagent/port/porthlpr.cpp
```diff
@@ -468,8 +501,16 @@ bool PortHelper::parsePortFec(PortConfig &port, const std::string &field, const
...
+    const auto &override_cit = portFecOverrideMap.find(value);
+    if (override_cit == portFecOverrideMap.cend())
+    {
+        SWSS_LOG_ERROR("Failed to parse field(%s): invalid value(%s) in override map", field.c_str(), value.c_str());
+        return false;
+    }
+
     port.fec.value = cit->second;
     port.fec.is_set = true;
+    port.fec.override_fec =override_cit->second;
```
prgeor, line 513: @dgsudharsan only for none,rc,fc this will be true?

orchagent/portsorch.cpp
```diff
@@ -2357,10 +2355,19 @@ void PortsOrch::initPortCapLinkTraining(Port &port)
...
 
-bool PortsOrch::isFecModeSupported(const Port &port, sai_port_fec_mode_t fec_mode)
+bool PortsOrch::isFecModeSupported(const Port &port, sai_port_fec_mode_t fec_mode, bool override_fec)
 {
     initPortSupportedFecModes(port.m_alias, port.m_port_id);
 
+    if (!override_fec && !fec_override_sup)
+    {
+        return false;
+    }
```
prgeor, line 2365: @dgsudharsan suggest to move this check outside at the caller and log a print "Skipping FEC override setting for fec = auto-fec" and not change the meaning of this function which is not related to fec supported or not.

orchagent/portsorch.cpp
```diff
@@ -3867,9 +3878,9 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 if (pCfg.fec.is_set)
                 {
                     /* reset fec mode upon mode change */
-                    if (!p.m_fec_cfg || p.m_fec_mode != pCfg.fec.value)
+                    if (!p.m_fec_cfg || p.m_fec_mode != pCfg.fec.value || p.m_override_fec != pCfg.fec.override_fec)
                     {
-                        if (!isFecModeSupported(p, pCfg.fec.value))
+                        if (!isFecModeSupported(p, pCfg.fec.value, pCfg.fec.override_fec))
```
prgeor, line 3883: @dgsudharsan can we keep the existing function not coupled with fec override and indepently taken care outised?

orchagent/portsorch.cpp
```diff
@@ -7691,9 +7765,21 @@ bool PortsOrch::initGearboxPort(Port &port)
...
-            attr.value.s32 = fec_mode_map[m_gearboxPortMap[port.m_index].line_fec];
+            if (!m_portHlpr.fecToSaiFecMode(m_gearboxPortMap[port.m_index].line_fec, sai_fec))
+            {
+                SWSS_LOG_ERROR("Invalid line FEC mode %s", m_gearboxPortMap[port.m_index].line_fec.c_str());
+                return false;
+            }
+            attr.value.s32 = sai_fec;
             attrs.push_back(attr);
 
+            if (fec_override_sup)
```
prgeor, line 7776: @dgsudharsan can you add comment, that this will take effect only with AN enabled

orchagent/port/porthlpr.cpp
```diff
@@ -74,7 +74,8 @@ static const std::unordered_map<std::string, sai_port_fec_mode_t> portFecMap =
 {
     { PORT_FEC_NONE, SAI_PORT_FEC_MODE_NONE },
     { PORT_FEC_RS,   SAI_PORT_FEC_MODE_RS   },
-    { PORT_FEC_FC,   SAI_PORT_FEC_MODE_FC   }
+    { PORT_FEC_FC,   SAI_PORT_FEC_MODE_FC   },
+    { PORT_FEC_AUTO, SAI_PORT_FEC_MODE_NONE }
```
prgeor, line 78: @dgsudharsan why fec is None for auto?

dgsudharsan: This is as per the design in the HLD https://github.com/sonic-net/SONiC/pull/1416

orchagent/port/porthlpr.cpp
dgsudharsan: Yes. Please refer to portFecOverrideMap

orchagent/portsorch.cpp
dgsudharsan: Done

orchagent/portsorch.cpp
dgsudharsan: Done

orchagent/portsorch.cpp
dgsudharsan: Done

prgeor (approved):

prgeor:
@prsunny LGTM. please merge.

prsunny (approved):

PR #2894, opened 2023-09-05
[orchagent] Add missing attribute for tunnel term table

kcudnik:
**What I did**
Add SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_SRC_IP when creating tunnel term table entry
**Why I did it**
Because of SAI breaking change See: https://github.com/sonic-net/sonic-sairedis/pull/1255
**How I verified it**
Local build and make check
**Details if related**
From SAI v1.13.0 new attribute is required.

### DIFF
--- a/orchagent/tunneldecaporch.cpp
+++ b/orchagent/tunneldecaporch.cpp
@@ -540,6 +540,17 @@ bool TunnelDecapOrch::addDecapTunnelTermEntries(string tunnelKey, swss::IpAddres
             copy(attr.value.ipaddr, ia);
             tunnel_table_entry_attrs.push_back(attr);
 
+            if (SAI_API_VERSION > SAI_VERSION(1,12,0))
+            {
+                // because of breaking change https://github.com/sonic-net/sonic-sairedis/pull/1255
+
+                attr.id = SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_SRC_IP; // required since v1.13.0
+                attr.value.ipaddr.addr_family = SAI_IP_ADDR_FAMILY_IPV4; // TODO may require revisit
+                attr.value.ipaddr.addr.ip4 = 0;
+
+                tunnel_table_entry_attrs.push_back(attr);
+            }
+
             // create the tunnel table entry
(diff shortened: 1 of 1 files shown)


orchagent/tunneldecaporch.cpp
```diff
@@ -540,6 +540,17 @@ bool TunnelDecapOrch::addDecapTunnelTermEntries(string tunnelKey, swss::IpAddres
             copy(attr.value.ipaddr, ia);
             tunnel_table_entry_attrs.push_back(attr);
 
+            if (SAI_API_VERSION > SAI_VERSION(1,12,0))
+            {
+                // because of breaking change https://github.com/sonic-net/sonic-sairedis/pull/1255
+
+                attr.id = SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_SRC_IP; // required since v1.13.0
```
prsunny, line 547: What if this is used with SAI 1.13 but using older SAI implementation? I think it can potentially return failure, right?

kcudnik: > What if this is used with SAI 1.13 but using older SAI implementation? I think it can potentially return failure, right?

@prsunny @lguohan 
actually this problem is unsavable, because os that headers breaking change. because metadata is compiled against SAI headers in SAI repo in sairedis repo, so it can be like v1.12 or v1.13

then you compile swss against SAI headers you installed from libsairedis-dev, or from headers from libsai-dev from vendor, if the headers match, there is no problem, if library is older, than v1.13, then metadata will require that extra parameter since it was compiled agains v1.13, but vendor library does not support that attribute ether, if the swss and sairedis are v1.12, and vendor i older, then vendor will expect that extra attribute which will be not passed

the solution to this scenario is to have both the vendor and the saimetdata to be at the save level or higher for example both v1.12 or both >= than v1.13, until next breaking change, unfortunately thats the problem of breaking changes :(

kcudnik:
lets hold on on this, we will discuss on todays SAI OCP meeting whether to revert that breaking change

kcudnik:
reverted https://github.com/opencomputeproject/SAI/pull/1890

PR #2901, opened 2023-09-08
Apply default TC for BFD packets

kperumalbfn:
**What I did**
Apply default TC value for BFD packets when BFD sessions are created. Use SAI attribute SAI_BFD_SESSION_ATTR_TC to set the TC value.

**Why I did it**
To transmit BFD packets using non-default Queue.

**How I verified it**
With default TC value of 7 and QoSmap TC_TO_QUEUE mapping TC(7)-->Queue(7), BFD packets are using Queue 7 on Egress ports.


siqbal1986 (approved):
looks good.

prsunny:
@kperumalbfn , do we need to check for capability or can this be merged?

kperumalbfn:
Closing this PR as theSAI attribute(SAI_BFD_SESSION_ATTR_TC) is not required for QoS lookup to select queue 7.

PR #2908, opened 2023-09-18
[orchagent] Add SAI_SWITCH_ATTR_SWITCH_ID for fabric switch

kcudnik:
This will be required since SAI v1.13

**What I did**
Add extra parameter when creating switch when switch type is fabric

**Why I did it**
It will be required attribute since SAI v1.13

**How I verified it**
Build locally

**Details if related**
Build on THIS will fail, since it will require sairedis to updates first, and sairedis will fail because swss don't contain this change, both repos needs to be update before test will start to pass

OA will require to pass "switch_id" param in config for fabric devices, not only for voq, devices, because of that test updates may be required

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -578,6 +578,10 @@ int main(int argc, char **argv)
         attr.id = SAI_SWITCH_ATTR_TYPE;
         attr.value.u32 = SAI_SWITCH_TYPE_FABRIC;
         attrs.push_back(attr);
+
+        attr.id = SAI_SWITCH_ATTR_SWITCH_ID;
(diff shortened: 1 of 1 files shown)


prsunny (approved):

abdosi (approved):

arlakshm (approved):

kcudnik:
Required sairedis commit https://github.com/sonic-net/sonic-sairedis/pull/1292 already merged

kcudnik:
i'm not sure which sairedis artifacts swss is downloading here
whether it gets latest successfull artifact or not

kcudnik:
let's merge this and proceed with sairedis, then we will get back here, i dont have permissions to merge

PR #2909, opened 2023-09-18
[codeql] Add missing json library

kcudnik:
**What I did**
Install new json library in codeql workflow
**Why I did it**
Because swss right now will require this json library to build

**How I verified it**
on this PR build
**Details if related**


kcudnik:
@prsunny seems like build system didn't picked up this change

kcudnik:
duplicate here: https://github.com/sonic-net/sonic-swss/pull/2900

PR #2910, opened 2023-09-18
[orchagent] Do not set internal port count to the PortConfigDone DB value.

mint570:
**What I did**
Do not set internal port count to the PortConfigDone DB value.

**Why I did it**
Fix https://github.com/sonic-net/sonic-swss/issues/2904

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -3294,14 +3294,6 @@ void PortsOrch::doPortTask(Consumer &consumer)
 
             setPortConfigState(PORT_CONFIG_RECEIVED);
 
-            for (const auto &cit : kfvFieldsValues(keyOpFieldsValues))
-            {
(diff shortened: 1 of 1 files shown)


nazariig:
@mint570 do we need to remove port `count` attribute also from `portsyncd`?

mint570:
@nazariig is that count in the DB being used by anything else?
If not, I can remove that from the DB.

mint570:
Looks like the schema is used in other places as well: https://github.com/sonic-net/sonic-utilities/blob/3df1f18b0a339ab9d92cd8c28ad703a491b30b85/scripts/db_migrator.py#L180

To be safe, let's not remove the count from DB now.

dgsudharsan (approved):

nazariig (approved):

prsunny (approved):

PR #2914, opened 2023-09-21
Change log level for NH down during ECMP create

kperumalbfn:
**What I did**
Change the log level for NH down event during ECMP create.

**Why I did it**
During ECMP create, down NHs are skipped and these events are not captured in syslog. Change the log level to NOTICE to capture these events in syslog for debugging.

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -1267,7 +1267,7 @@ bool RouteOrch::addNextHopGroup(const NextHopGroupKey &nexthops)
         // skip next hop group member create for neighbor from down port
         if (m_neighOrch->isNextHopFlagSet(it, NHFLAGS_IFDOWN))
         {
-            SWSS_LOG_INFO("Interface down for NH %s, skip this NH", it.to_string().c_str());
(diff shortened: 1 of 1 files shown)


orchagent/routeorch.cpp
```diff
@@ -1267,7 +1267,7 @@ bool RouteOrch::addNextHopGroup(const NextHopGroupKey &nexthops)
         // skip next hop group member create for neighbor from down port
         if (m_neighOrch->isNextHopFlagSet(it, NHFLAGS_IFDOWN))
         {
-            SWSS_LOG_INFO("Interface down for NH %s, skip this NH", it.to_string().c_str());
+            SWSS_LOG_NOTICE("Interface down for NH %s, skip this NH", it.to_string().c_str());
```
prsunny, line 1270: this may overload syslog if there are interface flaps

kperumalbfn: There will be extra logs if the NH egress interfaces are down during route add/update event, but this will have impact on forwarding functionality. So, it will be useful log to capture in syslog for debugging.

kperumalbfn:
Closing this PR as this change will flood syslog with lot of log messages when there is a link flap and multiple routes pointing to that interface.

PR #2917, opened 2023-09-27
[muxorch] Reorder the neighbor disable operations

lolyu:
**What I did**
When mux transits from active to standby, orchagent first disable the
neighbor[1], then program the tunnel route[2].
There is a small period of time between [1] and [2] that the ASIC
is unable to forward the traffic to the neighbor.

So let's reorder those two operations - program the tunnel route first,
then disable the neighbor.

**Why I did it**
Packet loss observed during `config mux standby`

**How I verified it**
Validate that, with the fix, no more packet loss is observed when `config mux standby`

**Details if related**


Ndancejic (approved):

prsunny (approved):

lolyu:
Hi @yxieca, could you please help cherry-pick this into 202205?

PR #2924, opened 2023-10-11
Reducing the severity of oper fec attribute get failure

dgsudharsan:
**What I did**
Reduced the severity of log when get attribute of oper fec failed.

**Why I did it**
The oper fec is not a mandatory attribute and may not be implemented by vendor SAI, in which case throwing an error is wrong. Reduced the severity of the log message to notice. This will also fix the checker issues in https://github.com/sonic-net/sonic-buildimage/pull/16785

**How I verified it**

**Details if related**


prsunny (approved):

prgeor (approved):

prgeor:
@dgsudharsan do we need in 202305? if so please add the branch tag

dgsudharsan:
> @dgsudharsan do we need in 202305? if so please add the branch tag

Not required as the feature is only present in master.

PR #2925, opened 2023-10-11
Fix infinite loop issue in buffermgr when deleing a  port.

baxia-lan:
**What I did**
Fix infinite loop issue in buffermgr when deleing a  port.

**Why I did it**

**How I verified it**

**Details if related**


baxia-lan:
@mint570 @prsunny PTAL at a quick fix on infinite loop issue in buffermgr

prsunny:
@baxia-lan , please provide more details on the PR? How it was found? Which PR caused? Is it a regression?

PR #2928, opened 2023-10-12
Independent Module - swss code to support port SI configuration per speed

tshalvi:
HLD: https://github.com/sonic-net/SONiC/pull/1455

What I did
I added six new SI parameters to the relevant locations in SWSS and introduced a log to display the SI values sent by SONiC to SAI.

Why I did it
I made these additions to support the new SI parameters recently added to APP_DB.

How I verified it
I verified it using the existing unit tests, and I will be adding new unit tests in the near future.

Details if related

HLD: https://github.com/sonic-net/SONiC/pull/1455


tshalvi:
Opened by accident. I'm closing it

PR #2931, opened 2023-10-12
Fix data race in on_switch_shutdown_request()

Yakiv-Huryk:
The data race in on_switch_shutdown_request() is the following:

* the on_switch_shutdown_request() calls exit() which calls the destructors for global static variables (e.g.BufferOrch::m_buffer_type_maps)
* in parallel to that, orchagent accesses the global static variables

A fix is to avoid calling the destructors by using quick_exit() instead of exit()

**What I did**
Changed exit() -> quick_exit() in on_switch_shutdown_request().
Added __lsan_do_leak_check() call for ASAN build.

**Why I did it**
To address a data race in the on_switch_shutdown_request()

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/notifications.cpp
+++ b/orchagent/notifications.cpp
@@ -5,6 +5,10 @@ extern "C" {
 #include "logger.h"
 #include "notifications.h"
 
+#ifdef ASAN_ENABLED
+#include <sanitizer/lsan_interface.h>
+#endif
+
 void on_fdb_event(uint32_t count, sai_fdb_event_notification_data_t *data)
 {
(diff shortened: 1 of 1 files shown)


dgsudharsan (approved):

qiluo-msft (approved):

prsunny (approved):

dgsudharsan:
@StormLiangMS  This is fix for the bug https://github.com/sonic-net/sonic-buildimage/issues/15966

This is not branch specific flow and we have verified this in general. In addition we ran the sanity testing on 202305.

StormLiangMS:
ADO:25592769

PR #2944, opened 2023-10-26
[Chassis] : Added test case to verify WRED profile on system ports

vmittal-msft:
**What I did**
Added test case for wred profile on system ports

**Why I did it**
Added test case for  https://github.com/sonic-net/sonic-swss/pull/2855

**How I verified it**
Verified on VS testbed.

**Details if related**

### DIFF
--- a/orchagent/pfc_detect_mellanox.lua
+++ b/orchagent/pfc_detect_mellanox.lua
@@ -12,6 +12,17 @@ local rets = {}
 
 redis.call('SELECT', counters_db)
 
+-- Record the polling time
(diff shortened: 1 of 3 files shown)


vmittal-msft:
Test case for : https://github.com/sonic-net/sonic-swss/pull/2855

vmittal-msft:
Since VS tests were failing, removed my commits to check if ok.

vmittal-msft:
Since VS tests were failing, removed my commits to check if ok.

vmittal-msft:
sw

PR #2946, opened 2023-10-27
[orchagent] Remove deprecated sync option

kcudnik:
Replaced by -z parameter

**What I did**
Remove depreciated parameter

**Why I did it**
To simplify the code and use new params

**How I verified it**
Locally

**Details if related**


prsunny (approved):

kcudnik:
some test env needs to be fixed, since 170 tests, fails, this option is still used also in prod, probably it will be hard to track where are all the places

PR #2950, opened 2023-11-01
[muxorch] Fixing updateRoute logic

Ndancejic:
What I did:
- Removed logic that prevented routes from updating if an active nexthop was already found in updateRoute cache.
- added test to check for scenario where ecmp route update triggers after updateRoute programs route to point to single nh
- Added logs to some key paths in updateRoute logic to add visibility

Why I did it:
Bug in logic was causing routes getting updated multiple times to skip updateRoute logic

How I did it:
removing logic that checked cache for active nexthop

https://github.com/sonic-net/sonic-swss/pull/2952

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -676,6 +676,7 @@ void MuxCable::updateNeighbor(NextHopKey nh, bool add)
  */
 void MuxCable::updateRoutes()
 {
+    SWSS_LOG_INFO("Updating routes pointing to multiple mux nexthops");
     MuxNeighbor neighbors = nbr_handler_->getNeighbors();
     string alias = nbr_handler_->getAlias();
     for (auto nh = neighbors.begin(); nh != neighbors.end(); nh ++)
@@ -1099,6 +1100,8 @@ void MuxOrch::updateRoute(const IpPrefix &pfx, bool add)
 
     if (!add)
     {
+        SWSS_LOG_INFO("Removing route %s from mux_multi_active_nh_table",
+                pfx.to_string().c_str());
         mux_multi_active_nh_table.erase(pfx);
         return;
     }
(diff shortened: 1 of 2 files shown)


orchagent/muxorch.cpp
```diff
@@ -1168,7 +1154,7 @@ void MuxOrch::updateRoute(const IpPrefix &pfx, bool add)
                         pfx.to_string().c_str(), nexthop.to_string().c_str());
                 continue;
             }
-            SWSS_LOG_INFO("setting route %s with nexthop %s %" PRIx64 "",
+            SWSS_LOG_NOTICE("setting route %s with nexthop %s %" PRIx64 "",
                 pfx.to_string().c_str(), nexthop.to_string().c_str(), next_hop_id);
             mux_multi_active_nh_table[pfx] = nexthop;
```
prsunny, line 1159: Is this still being used?

Ndancejic: It's not necessary, but it can help us tell when this update is being triggered. We can still skip if the route is pointing to a single nexthop

tests/test_mux.py
```diff
@@ -603,6 +603,25 @@ def multi_nexthop_test(self, dvs, dvs_route, asicdb, appdb, route, neighbors, ma
...
                 self.del_route(dvs, route)
 
+            # Check route_updates
+            print("Testing route updates")
+            self.add_route(dvs, route, neighbors)
+            print("setting states to active")
+            for port in mux_ports:
+                self.set_mux_state(appdb, port, "active")
+            print("triggering another route update")
+            self.add_route(dvs, route, neighbors)
```
prsunny, line 613: for route update, you may want to try with a different set of neighbors. applying the same route with same set of neighbors may skip processing by routeOrch

prsunny (approved):

prsunny:
@yxieca , the VS tests are passing but failed at publishing gcov stage. Merging it!

PR #2951, opened 2023-11-02
[orchagent]Modifying NPU_SI_SETTINGS_SYNC_STATUS based on NPU SI settings application status

mihirpat1:
**What I did**
Add change to update NPU_SI_SETTINGS_SYNC_STATUS based on NPU SI settings application status.

**Why I did it**
The change set is part of implementation of the HLD https://github.com/sonic-net/SONiC/pull/1432

**How I verified it**
Please refer to https://github.com/sonic-net/sonic-platform-daemons/pull/403 for the test results.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -4066,6 +4066,15 @@ void PortsOrch::doPortTask(Consumer &consumer)
                             SWSS_LOG_NOTICE("Set port %s SI settings is successful", p.m_alias.c_str());
                             p.m_preemphasis = serdes_attr;
                             m_portList[p.m_alias] = p;
+
+                            string value;
+                            bool foundNPUSiSettingsSyncStatus = m_portTable->hget(p.m_alias, "NPU_SI_SETTINGS_SYNC_STATUS", value);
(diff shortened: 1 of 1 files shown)


mihirpat1:
@prgeor @shyam77git @jaganbal-a - It will be great if you can help in reviewing this.

orchagent/portsorch.cpp
```diff
@@ -4066,6 +4066,15 @@ void PortsOrch::doPortTask(Consumer &consumer)
                             SWSS_LOG_NOTICE("Set port %s SI settings is successful", p.m_alias.c_str());
                             p.m_preemphasis = serdes_attr;
                             m_portList[p.m_alias] = p;
+
```
prgeor, line 4069: @mihirpat1 there are more than one place where setPortSerdesAttribute() is called. So, its better to notify inside setPortSerdesAttribute()

mihirpat1: As discussed, since the other place where setPortSerdesAttribute() is called (through initGearboxPort) is triggered only during boot-up and is not directly related to the NPU SI settings being set from xcvrd.
Hence, we will keep the current code as is.

PR #2958, opened 2023-11-14
[dash] fix DASH ACL Rule protocol use-after-free

Yakiv-Huryk:
**What I did**
Updated the code to avoid using the vector (protocols) that goes out of scope before its data is used

**Why I did it**
To fix a use-after-free bug

**How I verified it**

**Details if related**


oleksandrivantsiv (approved):

liat-grozovik:
@prsunny please help to review or assign someone

prsunny (approved):

PR #2960, opened 2023-11-16
Run test orchagent coredump

inspurSDN:
Cause:
    No check before removeVlan and No use upper_bound.
Solution:
    Check vlan existsing before removeVlan and use upper_bound.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


prsunny:
Seems like a test PR. Closing it!

PR #2961, opened 2023-11-16
orchagent would crash, when portsorch failed to remove bridge port

inspurSDN:
Root cause:
        Failed to remove bridge port becasue reference count is not 0

    Solution:
        1. put DEL task into m_toSync, when portsorch failed to remove bridge port
        2. if port was removed from vlan but m_bridge_port_id is not SAI_NULL_OBJECT_ID,
           portsorch will try to remove bridgeport again until reference count is 0

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -4065,13 +4065,31 @@ void PortsOrch::doVlanMemberTask(Consumer &consumer)
                 {
                     if (m_portVlanMember[port.m_alias].empty())
                     {
-                        removeBridgePort(port);
+                        if (!removeBridgePort(port))
+                        {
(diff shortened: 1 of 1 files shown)


orchagent/portsorch.cpp
```diff
@@ -5437,11 +5455,15 @@ bool PortsOrch::removeBridgePort(Port &port)
     {
         SWSS_LOG_ERROR("Failed to remove bridge port %s from default 1Q bridge, rv:%d",
             port.m_alias.c_str(), status);
+#if 0
```
prsunny, line 5458: We cannot have dead code.

prsunny (changes requested):

prsunny:
Please add unit tests

PR #2971, opened 2023-11-29
[coppmgrd] Use Producer del instead of del from Table

vivekrnv:
**What I did**

When handling the copp entries that are updated in new life, coppmgrd is deleting them directly from the table instead of using ProducerStateTable Del. Thus updated the value.

**Why i did it**

Without this, the copp traps are not created during reconciliation phase (because the key is deleted entirely from APP_DB) and are created at a later point. This would case a delay in LACP trap creation and would result in a 30 sec delay for the LACP sessions to be established and dplane downtime

**How I verified it**

Did a fastboot from 202205 -> 202305 (it had a difference in default copp fields and so the the del logic is triggered) and verified if the dplane downtime is < 30 sec

**Details if related**

### DIFF
--- a/cfgmgr/coppmgr.cpp
+++ b/cfgmgr/coppmgr.cpp
@@ -285,7 +285,7 @@ bool CoppMgr::isDupEntry(const std::string &key, std::vector<FieldValueTuple> &f
             if ((!field_found) || (field_found && preserved_copp_it->second.compare(value)))
             {
                 // overwrite -> delete preserved entry from copp table and set a new entry instead
-                m_coppTable.del(key);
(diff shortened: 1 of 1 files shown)


dgsudharsan (approved):

prsunny (approved):

dgsudharsan:
> Can you add a vs test?

As discussed offline, we do have vstest to cover the scenario, but vs test cannot replicate the real time usecase where orchagent has so many tasks in the queue during fastboot.

StormLiangMS:
ADO: 26008390

PR #2973, opened 2023-11-30
[crm] add support check for CRM_ACL_* resources

Yakiv-Huryk:
**What I did**
add support check for CRM_ACL_* resources

**Why I did it**
Some platforms don't support all the CRM attributes which leads to a stream of errors in the log.

**How I verified it**

**Details if related**


oleksandrivantsiv (approved):

Yakiv-Huryk:
The fix to pass test_CrmAcl/test_CrmAclGroup : https://github.com/sonic-net/sonic-sairedis/pull/1333

oleksandrivantsiv:
@prsunny kindly reminder. Please merge the PR

prsunny (approved):

orchagent/crmorch.cpp
```diff
@@ -928,6 +939,16 @@ void CrmOrch::getResAvailableCounters()
...
                 {
                     sai_status_t status = sai_acl_api->get_acl_table_attribute(cnt.second.id, 1, &attr);
+                    if ((status == SAI_STATUS_NOT_SUPPORTED) ||
+                        (status == SAI_STATUS_NOT_IMPLEMENTED) ||
+                        SAI_STATUS_IS_ATTR_NOT_SUPPORTED(status) ||
+                        SAI_STATUS_IS_ATTR_NOT_IMPLEMENTED(status))
+                    {
+                        // mark unsupported resources
+                        res.second.resStatus = CrmResourceStatus::CRM_RES_NOT_SUPPORTED;
+                        SWSS_LOG_NOTICE("CRM resource %s not supported", crmResTypeNameMap.at(res.first).c_str());
```
prsunny, line 949: Please confirm this log won't be spewed periodically. If so, please reduce to INFO level.

Yakiv-Huryk: It will be printed only once since if the resource is marked as not supported, it will be skipped in further checks:
https://github.com/sonic-net/sonic-swss/blob/master/orchagent/crmorch.cpp#L845

PR #2981, opened 2023-12-06
[fpmsyncd][WR] Relax the static schema constraint for ROUTE_TABLE

vivekrnv:
**What I did**

Remove the assert when the schema changes and in that case, imply a diff and pass it down to orchagent. Fixed https://github.com/sonic-net/sonic-buildimage/issues/17322

**Why I did it**

warmRestartHelper assumes a static schema during warm upgrade reconciliation. To overcome this, the db_migrator is used to add the missing fields in the ROUTE_TABLE. This is leading to control plane downtime exceeding 90 sec. 

**How I verified it**

UT:

```
vkarri@11ca2f5ec619:/sonic/src/sonic-swss/tests/mock_tests$ ./tests --gtest_filter="*WRHelper*"
Running main() from /build/googletest-YnT0O3/googletest-1.10.0.20201025/googletest/src/gtest_main.cc
Note: Google Test filter = *WRHelper*
[==========] Running 1 test from 1 test suite.
[----------] 1 test from WRHelperTest
[ RUN      ] WRHelperTest.testReconciliation
[       OK ] WRHelperTest.testReconciliation (1 ms)
[----------] 1 test from WRHelperTest (1 ms total)

[==========] 1 test from 1 test suite ran. (1 ms total)
[  PASSED  ] 1 test.
```

**Details if related**

### DIFF
--- a/warmrestart/warmRestartHelper.cpp
+++ b/warmrestart/warmRestartHelper.cpp
@@ -264,7 +264,7 @@ void WarmStartHelper::reconcile(void)
  * Compare all field-value-tuples within two vectors.
  *
  * Example: v1 {nexthop: 10.1.1.1, ifname: eth1}
- *          v2 {nexthop: 10.1.1.2, ifname: eth2}
+ *          v2 {nexthop: 10.1.1.2, ifname: eth2, protocol: kernel, weight: 1}
  *
  * Returns:
  *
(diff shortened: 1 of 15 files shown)


dgsudharsan (dismissed):

stepanblyschak (approved):

warmrestart/warmRestartHelper.cpp
```diff
@@ -274,25 +274,24 @@ void WarmStartHelper::reconcile(void)
...
-         * "ifname" fields; neighsyncd is expected to make use of "family" and
-         * "neigh" fields, etc. The existing reconciliation logic will rely on
-         * this assumption.
-         */
-        assert(v1Iter != v1Map.end());
+
+        /* Return true when v2 has a new field */
+        if (v1Iter == v1Map.end())
+        {
+            return true;
```
prsunny, line 293: should this be continue?

prsunny: how about the fields that are same?

vivekrnv: In that case, this logic is invoked. And if compareOneFV return false, this would continue with the next field.

```
if (compareOneFV(v1Map[fvField(*v1Iter)], fvValue(v2fv)))
{
    return true;
}
```

vivekrnv: Not required, we just have to check if there is a diff or not (if v2 has a field that's not present in v1 eg: protocol, it already implies a diff)

If diff is found we push the entire v2 down to orchagent

```
if (compareAllFV(restoredFV, refreshedFV))
{
    SWSS_LOG_NOTICE("Warm-Restart reconciliation: updating entry %s",
                    printKFV(refreshedKey, refreshedFV).c_str());

    m_syncTable->set(refreshedKey, refreshedFV);
}
else
{
    SWSS_LOG_INFO("Warm-Restart reconciliation: no changes needed for "
                  "existing entry %s",
  }
```

prsunny:
wait for @yxieca to signoff

dgsudharsan (dismissed):

dgsudharsan (dismissed):

yxieca:
@vivekrnv can you confirm that this warm reboot helper is only used for routing table? If so, can you make sure you name the helper accordingly so that it is only relaxing the ROUTE_TABLE schema check?

dgsudharsan:
> @vivekrnv can you confirm that this warm reboot helper is only used for routing table? If so, can you make sure you name the helper accordingly so that it is only relaxing the ROUTE_TABLE schema check?

Hi @yxieca . Yes this is used only by fpmsyncd. However I would prefer raising the cosmetic changes as part of separate PR.

yxieca (approved):

dgsudharsan (approved):

dgsudharsan:
@StormLiangMS  Tested on top of 202305 commit https://github.com/sonic-net/sonic-buildimage/commit/688245a724e7462765530fd8ec3a745d7372dd5e

PR #2982, opened 2023-12-06
[muxorch] Fixing cache bug in updateRoute logic

Ndancejic:
**What I did**
Fixed a bug where cache was not removing entry when an ecmp mux route gets updated to a single nexthop route. And added test case to cover the edge case.

**Why I did it**

**How I verified it**
Removing cache from updateRoute logic, and relying on other logic to skip over multiple mux nexthop case.

**Details if related**

ADO: 26099588

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -1099,6 +1099,8 @@ void MuxOrch::updateRoute(const IpPrefix &pfx, bool add)
 
     if (!add)
     {
+        SWSS_LOG_INFO("Removing route %s from mux_multi_active_nh_table",
+                      pfx.to_string().c_str());
         return;
     }
 
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -1099,9 +1099,6 @@ void MuxOrch::updateRoute(const IpPrefix &pfx, bool add)
 
     if (!add)
     {
-        SWSS_LOG_INFO("Removing route %s from mux_multi_active_nh_table",
-                pfx.to_string().c_str());
-        mux_multi_active_nh_table.erase(pfx);
         return;
     }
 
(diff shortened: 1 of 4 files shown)


orchagent/routeorch.cpp
```diff
@@ -1585,13 +1585,9 @@ bool RouteOrch::updateNextHopRoutes(const NextHopKey& nextHop, uint32_t& numRout
...
-        /* Check if route is mux multi-nexthop route
-         * we define this as a route present in
-         * mux_multi_active_nh_table
-         * These routes originally point to NHG and should be handled by updateRoute()
-         */
-        MuxOrch* mux_orch = gDirectory.get<MuxOrch*>();
-        if (mux_orch->isMultiNexthopRoute((*rt).prefix))
+        /* Check if route points to nexthop group and skip */
+        NextHopGroupKey nhg_key = gRouteOrch->getSyncdRouteNhgKey(gVirtualRouterId, (*rt).prefix);
+        if (nhg_key.getSize() > 1)
```
theasianpianist, line 1590: Would this also capture the case of a route with the portchannels as nexthops?

orchagent/muxorch.cpp
```diff
@@ -1099,9 +1099,6 @@ void MuxOrch::updateRoute(const IpPrefix &pfx, bool add)
 
     if (!add)
     {
-        SWSS_LOG_INFO("Removing route %s from mux_multi_active_nh_table",
```
prsunny, line 1102: Can you please keep the log to NOTICE and update log message with pfx details since its returning from here?

prsunny:
@Ndancejic , please re-edit the description to keep the template?

orchagent/routeorch.cpp
Ndancejic: we have a few test cases for that already, I'll add more with the test PR I plan to do later this month

StormLiangMS:
@Ndancejic  @prsunny  I didn't see nightly failure on this, could you help to file a test gap?

prsunny:
@StormLiangMS , this is now covered with VS test and we are also adding more unit tests

PR #2984, opened 2023-12-07
[dash] reduce the memory used by DASH ACL rules

Yakiv-Huryk:
**What I did**
* add a smaller struct for storing only the essential information about DASH ACL Rule (DashAclRuleInfo)
* in the event of rule update/re-creation, the rule configuration is read from a Redis

**Why I did it**
To reduce memory consumption

**How I verified it**
Run dash acl vs tests

**Details if related**

### DIFF
--- a/orchagent/dash/dashaclgroupmgr.cpp
+++ b/orchagent/dash/dashaclgroupmgr.cpp
@@ -9,6 +9,7 @@
 #include "dashaclorch.h"
 #include "saihelper.h"
 #include "pbutils.h"
+#include "taskworker.h"
 
 extern sai_dash_acl_api_t* sai_dash_acl_api;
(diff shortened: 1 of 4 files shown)


oleksandrivantsiv (approved):

prsunny:
@Yakiv-Huryk , could you please check coverage?

Yakiv-Huryk:
> @Yakiv-Huryk , could you please check coverage?

The coverage report doesn't seem correct. 
For example, the `sai_dash_acl_api->create_dash_acl_rule` is marked as not covered but the sairedis.rec contains `"|c|SAI_OBJECT_TYPE_DASH_ACL_RULE ..."`

prsunny (approved):

PR #2985, opened 2023-12-07
[Fdbsyncd] Adding extern_learn flag with fdb entry so Kernel doesn't age out

kishorekunal01:
- What I did
    extern_learn flag is added while programming the fdb entry into the Kernel. This will make sure that kernel doesn't age out the fdb entry. (#15004)

- How I did it
    A flag extern_learn will be passed while programing the fdb entry. ([#15004](https://github.com/sonic-net/sonic-buildimage/issues/15004))

- How to verify it
    Tested MAC add/del to the Kernel from the local FDB entry. ([#15004](https://github.com/sonic-net/sonic-buildimage/issues/15004))

Signed-off-by: kishore.kunal@broadcom.com

### DIFF
--- a/fdbsyncd/fdbsync.cpp
+++ b/fdbsyncd/fdbsync.cpp
@@ -324,7 +324,7 @@ void FdbSync::updateLocalMac (struct m_fdb_info *info)
 
     if (fdb_type == FDB_TYPE_DYNAMIC)
     {
-        type = "dynamic";
+        type = "dynamic extern_learn";
     }
     else
     {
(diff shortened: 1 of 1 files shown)


dgsudharsan (changes requested):
Can you please add UT or update existing UT?

kishorekunal01:
UT is already part of the fdbsyncd, and it is passing with this change

dgsudharsan (approved):

dgsudharsan:
Fix for https://github.com/sonic-net/sonic-buildimage/issues/15004

PR #2991, opened 2023-12-13
Use json.hpp from nlohmann-json-dev instead of from swss-common (#2900)

kcudnik:
* Use json.hpp from nlohmann-json-dev instead of from swss-common

This header file comes from an external package, and a very old version of the header file has been checked into swss-common. This will cause problems for the upcoming Bookworm upgrade.

Change references to the header file to use the Debian package nlohmann-json-dev, instead of from swss-common.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -68,6 +68,7 @@ jobs:
             autoconf-archive \
             uuid-dev \
             libjansson-dev \
+            nlohmann-json3-dev \
             python \
             stgit
 
--- a/cfgmgr/buffermgrd.cpp
+++ b/cfgmgr/buffermgrd.cpp
@@ -11,7 +11,7 @@
 #include <fstream>
 #include <iostream>
 #include "json.h"
-#include "json.hpp"
+#include <nlohmann/json.hpp>
 #include "warm_restart.h"
 
 using namespace std;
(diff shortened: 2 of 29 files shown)


saiarcot895:
I thought the plan was to change sairedis 202305 branch to not use json.hpp from nlohmann-json-dev? Trying to move everything to use nlohmann-json-dev will be a bigger effort that should not happen on a release branch.

kcudnik:
> I thought the plan was to change sairedis 202305 branch to not use json.hpp from nlohmann-json-dev? Trying to move everything to use nlohmann-json-dev will be a bigger effort that should not happen on a release branch.

this change is just cherrypick from master, so we want to move in that direction, why this will be a problem ?

saiarcot895:
Two reasons:

1. Cherry-picks/backports are meant for necessary bug fixes; this backport (of using json.hpp from the nlohmann-json-dev package instead of the local copy in swss-common) is not a required bug fix for 202305, besides the fact that a very old version is being used here.
2. There are multiple submodules that use json.hpp from swss-common. With this, there is now more of a mix of code with some components using json.hpp from swss-common and some components using json.hpp from nlohmann-json-dev. Following this through will mean a bigger effort that opens the door to possible bugs than just using the existing json.hpp from swss-common.

prsunny:
@kcudnik , please link the master PR if you are cherry-picking

kcudnik:
> @kcudnik , please link the master PR if you are cherry-picking

number is in this PR name :D https://github.com/sonic-net/sonic-swss/pull/2900

PR #2994, opened 2023-12-15
[muxorch][202205] Fixing cache bug in updateRoute logic

Ndancejic:
**What I did**
Fixed a bug where cache was not removing entry when an ecmp mux route gets updated to a single nexthop route. Removed cache from updateRoute logic, and relying on other logic to skip over multiple mux nexthop case.

**Why I did it**

**How I verified it**
Added vstest and ran test cases on virtual switch

**Details if related**
cherry-pick of https://github.com/sonic-net/sonic-swss/pull/2982

MSFT ADO - 26099588

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -2049,6 +2049,7 @@ bool RouteOrch::addRoutePost(const RouteBulkContext& ctx, const NextHopGroupKey 
 
     auto it_status = object_statuses.begin();
     auto it_route = m_syncdRoutes.at(vrf_id).find(ipPrefix);
+    MuxOrch* mux_orch = gDirectory.get<MuxOrch*>();
     if (isFineGrained)
     {
         if (it_route == m_syncdRoutes.at(vrf_id).end())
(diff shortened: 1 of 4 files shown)


orchagent/routeorch.cpp
```diff
@@ -2185,7 +2182,24 @@ bool RouteOrch::addRoutePost(const RouteBulkContext& ctx, const NextHopGroupKey
             if (ol_nextHops.getSize() > 1
                 && m_syncdNextHopGroups[ol_nextHops].ref_count == 0)
             {
-                m_bulkNhgReducedRefCnt.emplace(ol_nextHops, 0);
+                if (m_syncdNextHopGroups[ol_nextHops].ref_count == 0)
```
prsunny, line 2185: This seems to be unrelated to your PR. Can you please revert?

Ndancejic: Oh yes, this is what I was talking about earlier. I forgot to commit, let me fix that

prsunny (approved):

prsunny:
Failure on test-chassis, unrelated to PR changes. Merging to unblock build

PR #2997, opened 2023-12-20
[202305] Revert bgp suppress fib pending

stepanblyschak:
**What I did**

Revert the feature.

**Why I did it**

Revert bgp suppress FIB functionality due to found FRR memory consumption issues and bugs.

**How I verified it**

Basic sanity check on t1-lag, regression in progress.

**Details if related**


liat-grozovik:
@stepanblyschak please check checkers failure in case there is also a sonic-mgmt test that should be fixed.

stepanblyschak:
vstest with asan passed, so I assume vstest failed due to some random existing issues.

StormLiangMS (approved):
LGTM

PR #2999, opened 2023-12-26
[routeorch] Fixing bug with multiple routes pointing to nhg

Ndancejic:
**What I did**

**Why I did it**

**How I verified it**


prsunny (approved):

StormLiangMS:
@Ndancejic @prsunny could you have ADO and test result with 202305 image?

prsunny:
> @Ndancejic @prsunny could you have ADO and test result with 202305 image?

@StormLiangMS , ADO - 26099588

PR #3000, opened 2023-12-28
[202205][routeorch] Fixing bug with multiple routes pointing to nhg

Ndancejic:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**
cherry-pick of https://github.com/sonic-net/sonic-swss/pull/2999


prsunny (approved):

tests/test_virtual_chassis.py
```diff
@@ -846,7 +847,8 @@ def test_chassis_system_lag_id_allocator_del_id(self, vct):
                     assert len(lagmemberkeys) == 0, "Stale system lag member entries in asic db"
                     
                     break
-                    
+
+    @pytest.mark.skip(reason="Failing. Under investigation")
```
prsunny, line 851: You may want to import pytest

PR #3013, opened 2024-01-12
[RouteOrch] Publish route state for route to Loopback interface

stepanblyschak:
**What I did**

Publish exception route state to APPL_STATE_DB.

**Why I did it**

Fixes https://github.com/sonic-net/sonic-buildimage/issues/17638

Route to loopback interface is not advertised to peers when FIB suppression is enabled.

The proper way would be to program route to HW and then publish the route state, however to keep the existing behaviour this W/A is proposed until a final solution.

**How I verified it**

Route is queued:
```
admin@sonic:~$ show ipv6 route |grep q
       > - selected route, * - FIB route, q - queued route, r - rejected route
S>qfc00:1::/64 [1/0] is directly connected, Loopback0, 00:02:00
```

With this change:
```
admin@sonic:~$ show ipv6 route | grep q
       > - selected route, * - FIB route, q - queued route, r - rejected route
```

**Details if related**

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -751,6 +751,9 @@ void RouteOrch::doTask(Consumer& consumer)
                             it = consumer.m_toSync.erase(it);
                         else
                             it++;
+
+                        /* Publish route state to advertise routes to Loopback interface */
+                        publishRouteState(ctx);
                         continue;
                     }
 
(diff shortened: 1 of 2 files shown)


stepanblyschak:
@StormLiangMS @prsunny

stepanblyschak:
I need to add a test

prsunny (approved):
lgtm

PR #3017, opened 2024-01-18
Fix memory leak and object copying bugs in orchagent

saiarcot895:
**What I did**

When compiled with GCC 12, there are two types of compiler errors that get revealed. The first is GCC saying that in a for-loop, if the variable used for the objects is const, then a reference to the original object in the list can be used, instead of making a copy. The second is that two variables could potentially be used uninitialized. This appears to be a false positive to me, but I assigned it to 0 regardless.

There are, however, issues seen when running the tests. In a Bookworm-based container, the CoppMgrTest.CoppTest test case failed when run as part of the full test suite. However, when run individually, it passes. The error message is the following:

```
[ RUN      ] CoppMgrTest.CoppTest
mkdir: cannot create directory '/etc/sonic/': File exists
copp_ut.cpp:16: Failure
Expected equality of these values:
  status
    Which is: 256
  0
sudo: pam_open_session: Too many open files
sudo: policy plugin failed session initialization
copp_ut.cpp:28: Failure
Expected equality of these values:
  status
    Which is: 256
  0
[  FAILED  ] CoppMgrTest.CoppTest (15 ms)
```

Looking further, by the end of the test, there are a bunch of unconnected unix sockets opened, more than 1000 sockets. I suspect sudo (specifically, the `pam_open_session` module) is complaining that there are too many open files.

Looking into the test cases and the code, `FdbOrch` is creating a new `DBConnector` object, but it doesn't look like anything cleans this up if `FdbOrch` is destroyed.

Additionally, in the `AclOrch` test suite, nothing is deleting the `gFlowCounterRouteOrch` object. I've added a delete call for this.

There are still a bunch of file descriptors being created, suggesting there's still a leak somewhere, but now at least the full tests (when run in this order) are passing.

**Why I did it**

This is needed to create the docker-swss-layer-bookworm layer in sonic-buildimage, to allow for containers to upgrade to Bookworm.

This also fixes the potential for a runtime memory leak.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -42,8 +42,8 @@ FdbOrch::FdbOrch(DBConnector* applDbConnector, vector<table_name_with_pri_t> app
     Orch::addExecutor(flushNotifier);
 
     /* Add FDB notifications support from ASIC */
-    DBConnector *notificationsDb = new DBConnector("ASIC_DB", 0);
-    m_fdbNotificationConsumer = new swss::NotificationConsumer(notificationsDb, "NOTIFICATIONS");
+    m_notificationsDb = shared_ptr<DBConnector>(new DBConnector("ASIC_DB", 0));
+    m_fdbNotificationConsumer = new swss::NotificationConsumer(m_notificationsDb.get(), "NOTIFICATIONS");
     auto fdbNotifier = new Notifier(m_fdbNotificationConsumer, this, "FDB_NOTIFICATIONS");
     Orch::addExecutor(fdbNotifier);
 }
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -42,7 +42,7 @@ FdbOrch::FdbOrch(DBConnector* applDbConnector, vector<table_name_with_pri_t> app
     Orch::addExecutor(flushNotifier);
 
     /* Add FDB notifications support from ASIC */
-    m_notificationsDb = shared_ptr<DBConnector>(new DBConnector("ASIC_DB", 0));
+    m_notificationsDb = make_shared<DBConnector>("ASIC_DB", 0);
     m_fdbNotificationConsumer = new swss::NotificationConsumer(m_notificationsDb.get(), "NOTIFICATIONS");
     auto fdbNotifier = new Notifier(m_fdbNotificationConsumer, this, "FDB_NOTIFICATIONS");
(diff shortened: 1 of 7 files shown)


prsunny (dismissed):
lgtm. @qiluo-msft , could you please review?

orchagent/fdborch.cpp
```diff
@@ -42,8 +42,8 @@ FdbOrch::FdbOrch(DBConnector* applDbConnector, vector<table_name_with_pri_t> app
     Orch::addExecutor(flushNotifier);
 
     /* Add FDB notifications support from ASIC */
-    DBConnector *notificationsDb = new DBConnector("ASIC_DB", 0);
-    m_fdbNotificationConsumer = new swss::NotificationConsumer(notificationsDb, "NOTIFICATIONS");
```
qiluo-msft, line 46: Smartpoint is a valid bug fix, and we should backport it as a smaller standalone PR.

orchagent/fdborch.h
```diff
@@ -113,6 +113,7 @@ class FdbOrch: public Orch, public Subject, public Observer
     Table m_mclagFdbStateTable;
     NotificationConsumer* m_flushNotificationsConsumer;
     NotificationConsumer* m_fdbNotificationConsumer;
+    shared_ptr<DBConnector> m_notificationsDb;
```
qiluo-msft, line 116: Is it really necessary to introduce new member?

saiarcot895: Yes. The lifetime of `m_notificationsDb` needs to be at least as long as `m_fdbNotificationConsumer`, since `m_fdbNotificationConsumer` uses `m_notificationsDb`. Because currently it's just a local variable with a pointer, and because `swss::NotificationConsumer` doesn't accept `shared_ptr`, we need to make sure that `shared_ptr` doesn't free `DBConnector` too early.

orchagent/fdborch.cpp
saiarcot895: Created #3035 for this purpose.

prsunny (approved):

abdosi (approved):

yxieca:
@saiarcot895 can you add ADO number to this PR comment?

saiarcot895:
@yxieca Done

StormLiangMS:
hi @saiarcot895 could you run test with 202305 image for this PR?

PR #3022, opened 2024-01-24
[POC] Improve route performance with ZMQ

liuh-80:
Please ignore this POC PR.


mint570:
Just want to bring out the attention that there is a race condition in the zmq code.
The fix is in https://github.com/sonic-net/sonic-swss-common/pull/850

PR #3024, opened 2024-01-25
[buffermgr] Fix port breakout fail due to buffer configuration can't …

chiourung:
…be removed

Reproduce step:
1. Add default buffer configuration for all ports.
2. Do port breakout for port 1.
3. Do port breakout for port 2 and would be failed.
```
Running Breakout Mode : 1x100G[40G](4)
Target Breakout Mode : 1x400G

Ports to be deleted :
{
   "Ethernet120": "100000"
}
Ports to be added :
{
   "Ethernet120": "400000"
}
!!!  Critical Failure, Ports                     are not Deleted from ASIC DB, Bail Out  !!!
Ports are present in ASIC DB after 60 secs
Ports are present in ASIC DB after 60 secs
[ERROR] Port breakout Failed!!! Opting Out
Failed to break out Port. Error:
```
The first port breakout would send DEL port event and buffermgr would keep in while loop.
The second port breakout would remove buffer configuration first, but would be failed.

**What I did**
It should erase the iterator for non-important event.

**Why I did it**

**How I verified it**
Do port breakout for more than two ports.

**Details if related**


chiourung:
Fix in https://github.com/sonic-net/sonic-swss/commit/91bacca55a7d18860f442ca6b0562671b8dba9c8

PR #3025, opened 2024-01-25
[portsorch] process only updated APP_DB fields when port is already   created

stepanblyschak:
**What I did**

Fixing an issue when setting some port attribute in APPL_DB triggers serdes parameters to be re-programmed with port toggling. Made portsorch to handle only those attributes that were pushed to APPL_DB, so that serdes programming happens only by xcvrd's request to do so.

**Why I did it**

To fix an issue seen on warm-reboot. After orchagent reconciles, portmgrd sends another update to APPL_DB PORT_TABLE to sync. That trigger port flapping as portsorch does not handle just the fields that got pushed but all fields it has cached in ```m_portConfigMap```.

This also causes unconditionally re-set serdes attributes on any port attribute change, e.g.:

```
config interface mtu Ethernet136 9100
```

```
2024-01-18.09:02:30.712260|s|SAI_OBJECT_TYPE_PORT:oid:0x100000000001e|SAI_PORT_ATTR_ADMIN_STATE=false
2024-01-18.09:02:30.718675|n|port_host_tx_ready|[{"host_tx_ready_status":"SAI_PORT_HOST_TX_READY_STATUS_NOT_READY","port_id":"oid:0x100000000001e","switch_id":"oid:0x21000000000000"}]|
2024-01-18.09:02:30.718724|g|SAI_OBJECT_TYPE_PORT:oid:0x100000000001e|SAI_PORT_ATTR_PORT_SERDES_ID=oid:0x0
2024-01-18.09:02:30.719833|G|SAI_STATUS_SUCCESS|SAI_PORT_ATTR_PORT_SERDES_ID=oid:0x5700000000056c
2024-01-18.09:02:30.719908|r|SAI_OBJECT_TYPE_PORT_SERDES:oid:0x5700000000056c
2024-01-18.09:02:30.721476|c|SAI_OBJECT_TYPE_PORT_SERDES:oid:0x5700000000056d|SAI_PORT_SERDES_ATTR_PORT_ID=oid:0x100000000001e|SAI_PORT_SERDES_ATTR_IDRIVER=8:60,60,60,60,60,60,60,60|SAI_PORT_SERDES_ATTR_TX_FIR_PRE1=8:-2,-2,-2,-2,-2,-2,-2,-2|SAI_PORT_SERDES_ATTR_TX_FIR_PRE2=8:0,0,0,0,0,0,0,0|SAI_PORT_SERDES_ATTR_TX_FIR_MAIN=8:32,32,32,32,32,32,32,32|SAI_PORT_SERDES_ATTR_TX_FIR_POST1=8:6,6,6,6,6,6,6,6|SAI_PORT_SERDES_ATTR_TX_PAM4_RATIO=8:4,4,4,4,4,4,4,4|SAI_PORT_SERDES_ATTR_TX_OUT_COMMON_MODE=8:15,15,15,15,15,15,15,15|SAI_PORT_SERDES_ATTR_TX_PMOS_COMMON_MODE=8:105,105,105,105,105,105,105,105|SAI_PORT_SERDES_ATTR_TX_NMOS_COMMON_MODE=8:95,95,95,95,95,95,95,95|SAI_PORT_SERDES_ATTR_TX_PMOS_VLTG_REG=8:30,30,30,30,30,30,30,30|SAI_PORT_SERDES_ATTR_TX_NMOS_VLTG_REG=8:170,170,170,170,170,170,170,170
2024-01-18.09:02:30.724023|s|SAI_OBJECT_TYPE_PORT:oid:0x100000000001e|SAI_PORT_ATTR_ADMIN_STATE=true
2024-01-18.09:02:30.727609|S|SAI_OBJECT_TYPE_ROUTE_ENTRY||{"dest":"::/0","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000006"}|SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION=SAI_PACKET_ACTION_DROP||{"dest":"::/0","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000006"}|SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID=oid:0x0||{"dest":"0.0.0.0/0","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000006"}|SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION=SAI_PACKET_ACTION_DROP||{"dest":"0.0.0.0/0","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000006"}|SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID=oid:0x0
2024-01-18.09:02:30.731232|s|SAI_OBJECT_TYPE_PORT:oid:0x100000000001e|SAI_PORT_ATTR_MTU=9122
2024-01-18.09:02:30.732856|s|SAI_OBJECT_TYPE_PORT:oid:0x100000000001e|SAI_PORT_ATTR_ADMIN_STATE=false
2024-01-18.09:02:30.736624|g|SAI_OBJECT_TYPE_PORT:oid:0x100000000001e|SAI_PORT_ATTR_PORT_SERDES_ID=oid:0x0
2024-01-18.09:02:30.737289|G|SAI_STATUS_SUCCESS|SAI_PORT_ATTR_PORT_SERDES_ID=oid:0x5700000000056d
2024-01-18.09:02:30.737345|r|SAI_OBJECT_TYPE_PORT_SERDES:oid:0x5700000000056d
2024-01-18.09:02:30.738226|c|SAI_OBJECT_TYPE_PORT_SERDES:oid:0x5700000000056e|SAI_PORT_SERDES_ATTR_PORT_ID=oid:0x100000000001e|SAI_PORT_SERDES_ATTR_IDRIVER=8:60,60,60,60,60,60,60,60|SAI_PORT_SERDES_ATTR_TX_FIR_PRE1=8:-2,-2,-2,-2,-2,-2,-2,-2|SAI_PORT_SERDES_ATTR_TX_FIR_PRE2=8:0,0,0,0,0,0,0,0|SAI_PORT_SERDES_ATTR_TX_FIR_MAIN=8:32,32,32,32,32,32,32,32|SAI_PORT_SERDES_ATTR_TX_FIR_POST1=8:6,6,6,6,6,6,6,6|SAI_PORT_SERDES_ATTR_TX_PAM4_RATIO=8:4,4,4,4,4,4,4,4|SAI_PORT_SERDES_ATTR_TX_OUT_COMMON_MODE=8:15,15,15,15,15,15,15,15|SAI_PORT_SERDES_ATTR_TX_PMOS_COMMON_MODE=8:105,105,105,105,105,105,105,105|SAI_PORT_SERDES_ATTR_TX_NMOS_COMMON_MODE=8:95,95,95,95,95,95,95,95|SAI_PORT_SERDES_ATTR_TX_PMOS_VLTG_REG=8:30,30,30,30,30,30,30,30|SAI_PORT_SERDES_ATTR_TX_NMOS_VLTG_REG=8:170,170,170,170,170,170,170,170
2024-01-18.09:02:30.740160|s|SAI_OBJECT_TYPE_PORT:oid:0x100000000001e|SAI_PORT_ATTR_ADMIN_STATE=true
2024-01-18.09:02:30.805561|n|port_host_tx_ready|[{"host_tx_ready_status":"SAI_PORT_HOST_TX_READY_STATUS_READY","port_id":"oid:0x100000000001e","switch_id":"oid:0x21000000000000"}]|
```

**How I verified it**

By changing mtu of the interface I verified there was not port toggling and running warm-reboot test on T0 topology.

**Details if related**

**Request for 202311**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -3572,28 +3572,61 @@ void PortsOrch::doPortTask(Consumer &consumer)
 
         if (op == SET_COMMAND)
         {
-            auto &fvMap = m_portConfigMap[key];
-
-            for (const auto &cit : kfvFieldsValues(keyOpFieldsValues))
+            if (m_portList.find(key) == m_portList.end())
             {
-                auto fieldName = fvField(cit);
-                auto fieldValue = fvValue(cit);
+                // Aggregate configuration while the port is not created.
+                auto &fvMap = m_portConfigMap[key];
 
-                SWSS_LOG_INFO("FIELD: %s, VALUE: %s", fieldName.c_str(), fieldValue.c_str());
+                for (const auto &cit : kfvFieldsValues(keyOpFieldsValues))
+                {
+                    auto fieldName = fvField(cit);
+                    auto fieldValue = fvValue(cit);
 
-                fvMap[fieldName] = fieldValue;
+                    SWSS_LOG_INFO("FIELD: %s, VALUE: %s", fieldName.c_str(), fieldValue.c_str());
+
+                    fvMap[fieldName] = fieldValue;
+                }
+
+                pCfg.fieldValueMap = fvMap;
+
+                if (!m_portHlpr.parsePortConfig(pCfg))
+                {
+                    it = taskMap.erase(it);
+                    continue;
+                }
+
+                if (!m_portHlpr.validatePortConfig(pCfg))
+                {
+                    it = taskMap.erase(it);
+                    continue;
+                }
+
+                /* Collect information about all received ports */
+                m_lanesAliasSpeedMap[pCfg.lanes.value] = pCfg;
             }
-
-            pCfg.fieldValueMap = fvMap;
-
-            if (!m_portHlpr.parsePortConfig(pCfg))
+            else
             {
-                it = taskMap.erase(it);
-                continue;
-            }
+                // Port is already created, gather updated field-values.
+                std::unordered_map<std::string, std::string> fvMap;
 
-            /* Collect information about all received ports */
-            m_lanesAliasSpeedMap[pCfg.lanes.value] = pCfg;
+                for (const auto &cit : kfvFieldsValues(keyOpFieldsValues))
+                {
+                    auto fieldName = fvField(cit);
+                    auto fieldValue = fvValue(cit);
+
+                    SWSS_LOG_INFO("FIELD: %s, VALUE: %s", fieldName.c_str(), fieldValue.c_str());
+
+                    fvMap[fieldName] = fieldValue;
+                }
+
+                pCfg.fieldValueMap = fvMap;
+
+                if (!m_portHlpr.parsePortConfig(pCfg))
+                {
+                    it = taskMap.erase(it);
+                    continue;
+                }
+            }
 
             // TODO:
             // Fix the issue below
@@ -3708,6 +3741,9 @@ void PortsOrch::doPortTask(Consumer &consumer)
             {
                 PortSerdesAttrMap_t serdes_attr;
                 getPortSerdesAttr(serdes_attr, pCfg);
+
+                // Saved configured admin status
+                bool admin_status = p.m_admin_state_up;
 
                 if (pCfg.autoneg.is_set)
                 {
@@ -4268,6 +4304,13 @@ void PortsOrch::doPortTask(Consumer &consumer)
 
                 /* create host_tx_ready field in state-db */
                 initHostTxReadyState(p);
+
+                // Restore admin status if the port was brought down
+                if (admin_status != p.m_admin_state_up)
+                {
+                    pCfg.admin_status.is_set = true;
+                    pCfg.admin_status.value = admin_status;
+                }
 
                 /* Last step set port admin status */
                 if (pCfg.admin_status.is_set)
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -3572,11 +3572,8 @@ void PortsOrch::doPortTask(Consumer &consumer)
 
         if (op == SET_COMMAND)
         {
-            if (m_portList.find(key) == m_portList.end())
+            auto parsePortFvs = [&](auto& fvMap) -> bool
             {
-                // Aggregate configuration while the port is not created.
-                auto &fvMap = m_portConfigMap[key];
-
                 for (const auto &cit : kfvFieldsValues(keyOpFieldsValues))
                 {
                     auto fieldName = fvField(cit);
@@ -3590,6 +3587,19 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 pCfg.fieldValueMap = fvMap;
 
                 if (!m_portHlpr.parsePortConfig(pCfg))
+                {
+                    return false;
+                }
+
+                return true;
+            };
+
+            if (m_portList.find(key) == m_portList.end())
+            {
+                // Aggregate configuration while the port is not created.
+                auto &fvMap = m_portConfigMap[key];
+
+                if (!parsePortFvs(fvMap))
                 {
                     it = taskMap.erase(it);
                     continue;
@@ -3609,19 +3619,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 // Port is already created, gather updated field-values.
                 std::unordered_map<std::string, std::string> fvMap;
 
-                for (const auto &cit : kfvFieldsValues(keyOpFieldsValues))
-                {
-                    auto fieldName = fvField(cit);
-                    auto fieldValue = fvValue(cit);
-
-                    SWSS_LOG_INFO("FIELD: %s, VALUE: %s", fieldName.c_str(), fieldValue.c_str());
-
-                    fvMap[fieldName] = fieldValue;
-                }
-
-                pCfg.fieldValueMap = fvMap;
-
-                if (!m_portHlpr.parsePortConfig(pCfg))
+                if (!parsePortFvs(fvMap))
                 {
                     it = taskMap.erase(it);
                     continue;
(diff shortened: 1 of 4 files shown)


orchagent/portsorch.cpp
```diff
@@ -4170,15 +4164,9 @@ void PortsOrch::doPortTask(Consumer &consumer)
                     }
                 }
 
-                if (!serdes_attr.empty())
+                if (!serdes_attr.empty() && p.m_preemphasis != serdes_attr)
```
prgeor, line 4167: @stepanblyschak this will prevent setting of **SAME** serdes SI setting by OA if user updates any of the PORT table attributes in CONFIG_DB. This will _unintentionally_ prevent applying SAME SI setting when user plugs out the optics and insert it back. To avoid this we should clear/unset the p.m_preemphasis map when OA gets notified of optics removal [here](https://github.com/sonic-net/sonic-swss/blob/master/orchagent/portsorch.cpp#L4598)

prgeor: @stepanblyschak another use case where you would like to re-apply **same** SI settings is when user wants to shut the interface followed by no-shut. This should retrain the link with same SI settings. So, you should also clear the p.m_preemphasis map during admin down

stepanblyschak: @prgeor Do you mean SONiC **must** set same serdes SI attributes if the module is re-inserted? Is it per SAI spec? What libSAI is supposed to return when user queries for serdes attributes when the module is re-inserted?
For example, there is no such requirement on NVIDIA platform.
In case of shutdown/startup - why libSAI does not handle that? For example, we don't re-set the speed/fec/autoneg parameters when the interface is admin down, then up.

stepanblyschak: @prgeor I think clearing m_preemphasis to handle plug out event could be risky because there's no guarantee orchagent will first process STATE_TRANSCEIVER_INFO_TABLE_NAME and then APP_PORT_TABLE_NAME if both their corresponding consumers have pending data.

I am working on a different solution to make serdes config entirelly dependent on xcvrd's request.

prgeor: @stepanblyschak can you elaborate on the risk? Do you suspect OA may miss the insert/remove event of the transceiver?

stepanblyschak: @prgeor No, the risk is that there's no guarantee orchagent will first process STATE_TRANSCEIVER_INFO_TABLE_NAME and then APP_PORT_TABLE_NAME if both consumers have data.
Then, serdes config won't be applied.

stepanblyschak: @prgeor Please check the new appraoch.

I made the change so that while the port is not yet created, field-value pairs are aggregated in ```m_portConfigMap```, but when the port is already created only updated field-value pairs are processed.
This makes orchagent unconditionally set serdes attributes whenever xcvrd pushes the changes to APP_DB and also fixes the problem observed when mtu is configured. When mtu is configured, only mtu ```is_set``` with the new approach and serdes isn't re-configured. Serdes configuration becomes fully controlled by xcvrd and there's no special logic in orchagent for that.

Regarding admin status change I do think it is out of scope of this PR. It is not required on Nvidia platform. Even if required by SAI spec, original serdes support implementation (looking at 202205 for example) did not handle admin status change and there was no field-value pairs aggregation so it didn't re-configure serdes on every port field update including admin status change.

orchagent/portsorch.cpp
```diff
@@ -4228,6 +4264,13 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 /* create host_tx_ready field in state-db */
                 initHostTxReadyState(p);
 
+                // Restore admin status if the port was brought down
+                if (admin_status != p.m_admin_state_up)
```
prgeor, line 4268: @stepanblyschak why special handling for admin_status? Isn't it already taken care here https://github.com/sonic-net/sonic-swss/blob/9c995f0c83892ae9a4f6657c1c4673164e803169/orchagent/port/porthlpr.cpp#L736

orchagent/portsorch.cpp
```diff
@@ -3531,28 +3531,61 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+
+                for (const auto &cit : kfvFieldsValues(keyOpFieldsValues))
+                {
+                    auto fieldName = fvField(cit);
+                    auto fieldValue = fvValue(cit);
+
+                    SWSS_LOG_INFO("FIELD: %s, VALUE: %s", fieldName.c_str(), fieldValue.c_str());
+
+                    fvMap[fieldName] = fieldValue;
+                }
```
prgeor, line 3579: @stepanblyschak can we have a function() seems repeating in line 3539 also

orchagent/portsorch.cpp
stepanblyschak: @prgeor Setting some attributes require bringing port admin down - like speed. Hence, at the end of an update we need to bring port admin up. If we'd aggregate all fvs on update, then we don't need any special handling, admin_status is always set as per APPL_DB.
However, because this aggregation is removed we need to take a special care to restore port admin state.

orchagent/portsorch.cpp
stepanblyschak: @prgeor fixed

prgeor (approved):

prsunny:
@stepanblyschak , @prgeor , in general orchagent code should be idempotent. So even if same value is set again, it should pass down the value to SAI implementation. From what I understand, its causing flaps in some cases. PR is good to me, but just want to share the thought.

StormLiangMS:
hi @mihirpat1  for cherry pick to 202305, could you have it tested with 202305 image to avoid regression?

mihirpat1:
> hi @mihirpat1 for cherry pick to 202305, could you have it tested with 202305 image to avoid regression?

Hi @StormLiangMS,
I have tested this on 202305 image.
MSFT ADO - 27786245

PR #3029, opened 2024-01-31
[PortOrch] Add FEC codeword errors in port stats

vdahiya12:
**What I did**

   * This set of counters corresponds to number of symbol errors in each FEC codeword received on the port.
    
   * The number of symbol errors that may be detected is dependent on the type of FEC in use. For instance, RS-528 FEC supports detection of up to 7symbol errors, while RS-544 FEC supports detection of up to 15 symbol errors. 

**Why I did it**
This PR is adding the FEC symbols as per PR https://github.com/opencomputeproject/SAI/pull/1617

**How I verified it**

**Details if related**


prsunny (approved):

yxieca:
@vdahiya12 please add ADO number to this PR.

PR #3045, opened 2024-02-07
[intfsorch] Enable ipv6 proxy ndp along with proxy arp

Ndancejic:
setting SAI_VLAN_ATTR_UNKNOWN_MULTICAST_FLOOD_CONTROL_TYPE to SAI_VLAN_FLOOD_CONTROL_TYPE_NONE when proxy arp is enabled. This fixes a bug where ipv6 NS packets were flooding ports with duplicate packets. We now set multicast flood type to none.

**What I did**
Set SAI_VLAN_ATTR_UNKNOWN_MULTICAST_FLOOD_CONTROL_TYPE to vlan_flood_type. This will either be SAI_VLAN_FLOOD_CONTROL_TYPE_NONE or SAI_VLAN_FLOOD_CONTROL_TYPE_ALL depending on if proxy arp is enabled.

**Why I did it**
Fixes issue where ndp packets were flooding ports ADO: 26395578

**How I verified it**
Tested change on a lab device, confirmed that when proxy-arp was enabled multicast packets were being blocked in the hardware

**Details if related**
sonic-buildimage fix to block packets in kernel to come, will link here.

ADO: 26395578

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -368,6 +368,21 @@ bool IntfsOrch::setIntfVlanFloodType(const Port &port, sai_vlan_flood_control_ty
         }
     }
 
+    // Also set ipv6 multicast flood type
+    attr.id = SAI_VLAN_ATTR_UNKNOWN_MULTICAST_FLOOD_CONTROL_TYPE;
+    attr.value.s32 = vlan_flood_type;
+
+    sai_status_t status = sai_vlan_api->set_vlan_attribute(port.m_vlan_info.vlan_oid, &attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to set multicast flood type for VLAN %u, rv:%d", port.m_vlan_info.vlan_id, status);
+        task_process_status handle_status = handleSaiSetStatus(SAI_API_VLAN, status);
+        if (handle_status != task_success)
+        {
+            return parseHandleSaiStatusFailure(handle_status);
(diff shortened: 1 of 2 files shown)


orchagent/intfsorch.cpp
```diff
@@ -368,6 +368,21 @@ bool IntfsOrch::setIntfVlanFloodType(const Port &port, sai_vlan_flood_control_ty
         }
     }
 
+    // Also set ipv6 multicast flood type
+    attr.id = SAI_VLAN_ATTR_UNKNOWN_MULTICAST_FLOOD_CONTROL_TYPE;
```
prsunny, line 372: You may want to pass this as an attribute vector

Ndancejic: SAI API doesn't seem to support passing an attribute vector for set_vlan_attribute, and looking at other examples it doesn't seem like we're using it that way anywhere else:
```
/**
 * @brief Set VLAN Attribute
 *
 * @param[in] vlan_id VLAN ID
 * @param[in] attr Attribute structure containing ID and value
 *
 * @return #SAI_STATUS_SUCCESS on success, failure status code on error
 */
typedef sai_status_t (*sai_set_vlan_attribute_fn)(
        _In_ sai_object_id_t vlan_id,
        _In_ const sai_attribute_t *attr);
```

prsunny (approved):

StormLiangMS:
hi @Ndancejic  pls have a test result with 202305 branch, need to avoid regression at this stage.

PR #3052, opened 2024-02-17
Add port flap count and last flap timestamp to APPL_DB

prgeor:
**What I did**
Add port flap count and last flap timestamp to APPL_DB

**Why I did it**
To be able to get number of times the port flapped over the life of the port
Record the last time when the port was UP/DOWN

**How I verified it**
1. Verified the flap count increments when the port operational status changes
2. Time stamp is recorded when the port went UP/DOWN last time
3. Verified the flap count and timestamp are restored on warm-reboot

**Details if related**

image

image


orchagent/portsorch.cpp
```diff
@@ -5307,9 +5339,14 @@ bool PortsOrch::initializePort(Port &port)
             {
                 operStatus = fvValue(i);
             }
+
+            if (fvField(i) == "flap_count")
```
dgsudharsan, line 5343: Is this to handle reconciliation during warm/fast reboot?

dgsudharsan (changes requested):
Please add UT. You can add gtest to simulate port oper change event

orchagent/portsorch.cpp
prgeor: @dgsudharsan yes

prgeor:
> Please add UT. You can add gtest to simulate port oper change event

@dgsudharsan please check the test added

dgsudharsan:
> > Please add UT. You can add gtest to simulate port oper change event
> 
> @dgsudharsan please check the test added

Can you please check why the test is failing in the pipeline?

orchagent/portsorch.cpp
```diff
@@ -5314,6 +5351,20 @@ bool PortsOrch::initializePort(Port &port)
...
+    // initalize port flap count
+    if (!flapCount.empty())
+    {
+        try
+        {
+            port.m_flap_count = to_uint<std::uint32_t>(flapCount);
+        }
+        catch (const std::exception &e)
+        {
+            SWSS_LOG_ERROR("Failed to get port (%s) flap_count: %s", port.m_alias.c_str(), e.what());
```
prsunny, line 5363: This is going to throw an unwanted error during WB from old versions which doesn't have this field in DB. Please change to DEBUG as this is not a real error scenario

orchagent/portsorch.cpp
```diff
@@ -3074,6 +3074,38 @@ bool PortsOrch::removeVlanHostIntf(Port vl)
     return true;
 }
 
+void PortsOrch::updateDbPortFlapCount(Port& port, sai_port_oper_status_t pstatus)
+{
+    SWSS_LOG_ENTER();
+
+    ++port.m_flap_count;
```
prsunny, line 3081: looks like this will be incremented unnecessary if SAI has some flow where it notifies the same oper status multiple times?

prgeor: @prsunny no, there is a check where OA return if same port status is updated :- https://github.com/sonic-net/sonic-swss/blob/master/orchagent/portsorch.cpp#L7565

orchagent/portsorch.cpp
```diff
@@ -5291,9 +5323,14 @@ bool PortsOrch::initializePort(Port &port)
...
             }
+
+            if (fvField(i) == "flap_count")
+            {
+                flapCount = fvValue(i);
+            }
         }
     }
-    SWSS_LOG_DEBUG("initializePort %s with oper %s", port.m_alias.c_str(), operStatus.c_str());
+    SWSS_LOG_NOTICE("Port %s with oper %s flap_count=%s", port.m_alias.c_str(), operStatus.c_str(), flapCount.c_str());
```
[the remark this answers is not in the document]
prgeor: @prsunny if the field is not available in the DB, flapCount will be 0 because it is initialized with "0" by default (see line 5317)

orchagent/portsorch.cpp
prgeor: @prsunny if the value is not there in the DB, flapCount default value = "0" (see line 5317 above)

dgsudharsan (approved):

dgsudharsan (dismissed):

prgeor:
> @prgeor , seems its causing WB failures. Can you please check?
> 
> ```
>     # Only WARM_RESTART_TABLE|orchagent state=reconciled operation may exist after port oper status change.
> ```
> 
> > ```
> >   assert orchStateCount == 1
> > ```
> 
> E assert 2 == 1 E -2 E +1

@prgeor all tests now passed

nazariig:
@prgeor why have to use `APP DB`, why not `STATE DB`?

tests/mock_tests/portsorch_ut.cpp
```diff
@@ -2102,4 +2170,4 @@ namespace portsorch_test
         ASSERT_FALSE(bridgePortCalledBeforeLagMember); // bridge port created on lag before lag member was created
     }
 
-}
+}
```
nazariig, line 2173: @prgeor please add a new line

orchagent/portsorch.h
```diff
@@ -147,6 +147,8 @@ class PortsOrch : public Orch, public Subject
 
     bool setHostIntfsOperStatus(const Port& port, bool up) const;
     void updateDbPortOperStatus(const Port& port, sai_port_oper_status_t status) const;
+    void updateDbPortFlapCount(Port& port, sai_port_oper_status_t pstatus);
+
```
nazariig, line 151: @prgeor please remove extra line

orchagent/portsorch.cpp
```diff
@@ -3095,6 +3095,38 @@ bool PortsOrch::removeVlanHostIntf(Port vl)
...
 
+void PortsOrch::updateDbPortFlapCount(Port& port, sai_port_oper_status_t pstatus)
+{
+    SWSS_LOG_ENTER();
+
+    ++port.m_flap_count;
+    vector<FieldValueTuple> tuples;
+    FieldValueTuple tuple("flap_count", std::to_string(port.m_flap_count));
+    tuples.push_back(tuple);
+    m_portTable->set(port.m_alias, tuples);
```
nazariig, line 3106: @prgeor should be at the end of the function: gather field/value tuples and then set at once. No need to have duplicate calls to DB

nazariig (changes requested):

orchagent/portsorch.h
prgeor: @nazariig removed

prgeor:
> @prgeor why have to use `APP DB`, why not `STATE DB`?

@nazariig in order to reconcile the flap count after warm-reboot.

orchagent/port.h
```diff
@@ -164,6 +164,7 @@ class Port
     uint32_t  m_nat_zone_id = 0;
     uint32_t  m_vnid = VNID_NONE;
     uint32_t  m_fdb_count = 0;
+    uint32_t  m_flap_count = 0;
```
nazariig, line 167: @prgeor why not uint64? Do we expect an overflow?  Should be reflected in the log? Any reaction from the SWSS is needed?

orchagent/portsorch.cpp
```diff
@@ -5335,6 +5364,20 @@ bool PortsOrch::initializePort(Port &port)
...
+    // initalize port flap count
+    if (!flapCount.empty())
+    {
+        try
+        {
+            port.m_flap_count = to_uint<std::uint32_t>(flapCount);
+        }
+        catch (const std::exception &e)
+        {
+            SWSS_LOG_ERROR("Failed to get port (%s) flap_count: %s", port.m_alias.c_str(), e.what());
```
nazariig, line 5376: @prgeor if we hit the exception, is it ok to still put the value to the DB?

orchagent/portsorch.cpp
```diff
@@ -5335,6 +5364,20 @@ bool PortsOrch::initializePort(Port &port)
...
+        }
+        m_portTable->hset(port.m_alias, "flap_count", flapCount);
```
nazariig, line 5378: @prgeor is this a typo? `flapCount` ->`port.m_flap_count`

nazariig (changes requested):

orchagent/portsorch.cpp
prgeor: @nazariig the intention is to write uint64_t string value

orchagent/portsorch.cpp
prgeor: @nazariig fixed

orchagent/port.h
prgeor: @nazariig done

nazariig (approved):

dgsudharsan (approved):

justin-wong-ce:
The recent backport to 202311 caused a regression on 202311.

The backported commit had an issue that was previously fixed with https://github.com/sonic-net/sonic-swss/pull/3111 on master, which was not backported. Making a backport request to backport https://github.com/sonic-net/sonic-swss/pull/3111 to 202311.

Issue #2865, opened 2023-07-21
[swss] BUFFER_PG and BUFFER_QUEUE configs are not getting applied on new ports created from dynamic breakout


maulik-marvell:
## **Description**

In sonic 202211, after we perform a dynamic port breakout on one of the parent ports and create new ports out of it, we are not able to configure BUFFER_PG/BUFFER_QUEUE tables using sonic-cfggen. We do not see any updates in syslog/swss.rec/sairedis.rec for the new buffer configs loaded via sonic-cfggen.
 
However, when we check the running configs, we do see the BUFFER_PG/BUFFER_QUEUE configs present there but they are not applied to syncd/device. Because of this issue, not able to configure the buffer profiles on IPG/Queues for new ports. The same steps work well in sonic 202111 and such behavior/issue is not seen there.
 
## **Steps to reproduce the issue:**

1. Perform config reload in DUT to get to a fresh start.
   ```
    root@sonic-dut:~# config reload -y
   ```

2. List all the interfaces and pick any port for breakout(Ethernet0 is considered for this test)
    ```
    root@sonic-dut:~# show in status
    Interface            Lanes    Speed    MTU    FEC    Alias    Vlan    Oper    Admin             Type    Asym PFC
    -----------  ---------------  -------  -----  -----  -------  ------  ------  -------  ---------------  ----------
    Ethernet0      89,90,91,92     100G   9100     rs     Eth1  routed      up       up  QSFP28 or later         off
    Ethernet8      81,82,83,84     100G   9100     rs     Eth2  routed      up       up  QSFP28 or later         off
    Ethernet16      73,74,75,76     100G   9100     rs     Eth3  routed      up       up  QSFP28 or later         off
    Ethernet24      65,66,67,68     100G   9100     rs     Eth4  routed      up       up  QSFP28 or later         off
    Ethernet32      57,58,59,60     100G   9100     rs     Eth5  routed      up       up  QSFP28 or later         off
    ```

3. Perform dynamic port breakout on Ethernet0 and create 4 new ports Ethernet0,Ethernet1,Ethernet2,Ethernet3, make sure no error is reported
    ```
    root@sonic-dut:~# config interface breakout Ethernet0 4x25G[10G]\(4\) -yf
    
    Running Breakout Mode : 1x100G(4)
    Target Breakout Mode : 4x25G[10G](4)
     
    Ports to be deleted :
    {
        "Ethernet0": "100000"
    }
    Ports to be added :
    {
[...]

PR #3055, opened 2024-02-19
[buffermgrd] Move switch-statement outside of if-statement in BufferMgr::doTask

amazor:
Fix for: 
[swss] BUFFER_PG and BUFFER_QUEUE configs are not getting applied on new ports created from dynamic breakout #2865
and most l likely other bugs related to dynamic buffer management.

**What I did**
Fixed bug that would cause an infinite loop in buffermgr.cpp when sending a non-set command (eg. delete)

**Why I did it**
The Bug would cause buffermgrd to be stuck in infinite loop, blocking its function.

**How I verified it**
Buffer configuration after deletion of a port(eg. dynamic port breakout command) will now be correctly processed. This was checked through log outputs on buffermgrd

**Details if related**
The switch statement which would normally erase buffer events was moved to be inside the if-statement which would only enter if the event is a SET event. This was introduced in commit 5329c39.
This would cause an infinite loop, since non-set events would never be erased/consumed.
The switch statement has now been moved to occur outside the if, allowing for non-SET commands to be processed.

**Which release branch to backport** 
These branches should be updated as the bug exists on all of these releases:
- 202205
- 202211
- 202305
- 202311

### DIFF
--- a/cfgmgr/buffermgr.cpp
+++ b/cfgmgr/buffermgr.cpp
@@ -549,24 +549,23 @@ void BufferMgr::doTask(Consumer &consumer)
                     task_status = doSpeedUpdateTask(port);
                 }
             }
-
-            switch (task_status)
-            {
-                case task_process_status::task_failed:
-                    SWSS_LOG_ERROR("Failed to process table update");
-                    return;
-                case task_process_status::task_need_retry:
-                    SWSS_LOG_INFO("Unable to process table update. Will retry...");
-                    ++it;
-                    break;
-                case task_process_status::task_invalid_entry:
-                    SWSS_LOG_ERROR("Failed to process invalid entry, drop it");
-                    it = consumer.m_toSync.erase(it);
-                    break;
-                default:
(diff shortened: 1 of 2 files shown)


matiAlfaro (approved):

neethajohn (approved):

stephenxs (approved):

prsunny (approved):

amazor:
@prsunny Will this PR be backported to previous release branches as mentioned in the PR request?

> Which release branch to backport
> These branches should be updated as the bug exists on all of these releases:
> 
> - 202205
> - 202211
> - 202305
> - 202311

amazor:
@prsunny, Can I get an update on the backport request? Is this done in a separate PR, or by the community?

prsunny:
@amazor , i've marked for backporting to 2023 branch. 2022 is frozen for most backports.
Can you please confirm the test result for 202305 branch with this change before it can be approved for cherrypick?

PR #3066, opened 2024-02-29
New p4orch development changes

mint570:
New p4orch development changes.

This is the first PR for upstreaming the recent P4Orch changes (has been a while since the last time).
The main changes include:
1. Add new status code SWSS_RC_NOT_EXECUTED
2. Enable redis pipeline in response publisher for P4Orch (performance improvement)
3. Enable background thread of write to db in response publisher for P4Orch (performance improvement)
4. P4Orch writes to APPL DB instead of APPL STATE DB for responses (prepare for the zmq change, APPL DB will be used as APPL STATE DB in P4Orch for performance improvement)

More details on the P4Orch response path changes (changes are only done on P4Orch, other orchs has no behavior changes):
1. Enable P4Orch to use redis pipeline in response publisher.
The redis pipeline feature in response publisher is done by the SONiC communicate to improve performance. We enabled P4Orch to use the pipeline feature. This will increase performance in batched request processing. This change has no behavior changes.

2. Enable background thread in response publisher to write to DB for P4Orch.
This change will move the APPL STATE DB write into a background thread, which will improve performance. This change has no behavior changes. The original change was done for APPL STATE DB since the response path writes into the APPL STATE DB. With the next change, APPL STATE DB is deprecated and replaced with APPL DB. So the overall change is that the P4Orch response path will write into APPL DB in a background thread.

3. P4Orch writes to APPL DB instead of APPL STATE DB in response path.
The APPL STATE DB is supposed to have successful entries programmed in the ASIC, while the APPL DB has the intend. However, for P4 tables, P4RT needs to clean up APPL DB to match with APPL STATE DB to keep them in sync when a request fails. This is done for warmboot purpose as we don't want to program the failed intend during warmboot. So for P4RT, the APPL STATE DB is redundant information. This change will replace APPL STATE DB with APPL DB. In later PR, we will also upstream the change for P4Orch to enable zmq. This will reduce half of the DB write operations. This change will require the later P4RT change to be in sync.


mint570:
This PR requires https://github.com/sonic-net/sonic-swss-common/pull/828

orchagent/response_publisher.cpp
```diff
@@ -64,26 +64,52 @@ void RecordResponse(const std::string &response_channel, const std::string &key,
 
 } // namespace
 
-ResponsePublisher::ResponsePublisher(bool buffered)
```
prsunny, line 67: Can you please provide the details of response_publisher in the description? What was the behavior before and whats the new change? Is it breaking any previous implementation?

prsunny:
Why change the return path to APP_DB instead of APPL_STATE_DB?

mint570:
> Why change the return path to APP_DB instead of APPL_STATE_DB?

I updated the PR description.
This is to prepare for the zmq change that we are going to upstream. P4Orch will only update APPL DB (in a background thread) on response path. APPL STATE DB is redundant for P4Orch (as we need to keep them in sync for warmboot purpose).

orchagent/response_publisher.cpp
mint570: Updated the PR description.

prsunny:
@qiluo-msft , could you please review the response publish section? As per description there is no change to other orch.

mint570:
Got vs test failure on fabric port test.

mint570:
> Got vs test failure on fabric port test.

Looks like this is not related to my change. I tried a test PR with no change it still failed the vs test.

prsunny:
@mint570 , there has been a VS issue in buildimage and is fixed now. I'll trigger all PRs to rerun

qiluo-msft:
@bocon13, could you please review the response publish section?

orchagent/response_publisher.h
```diff
@@ -57,8 +61,29 @@ class ResponsePublisher : public ResponsePublisherInterface
...
+
     std::unique_ptr<swss::DBConnector> m_db;
+    std::unique_ptr<swss::RedisPipeline> m_ntf_pipe;
     std::unique_ptr<swss::RedisPipeline> m_pipe;
 
     bool m_buffered{false};
+    // Thread to write to DB.
+    std::unique_ptr<std::thread> m_update_thread;
+    std::queue<entry> m_queue;
+    mutable std::mutex m_lock;
```
qiluo-msft, line 87: >mutable [](http://example.com/codeflow?start=4&length=7)

why mutable?

orchagent/response_publisher.h
```diff
@@ -57,8 +61,29 @@ class ResponsePublisher : public ResponsePublisherInterface
...
+        bool shutdown;
+    };
+
+    void dbUpdateThread();
+    void writeToDBInternal(const std::string &table, const std::string &key,
+                           const std::vector<swss::FieldValueTuple> &values, const std::string &op, bool replace);
+
     std::unique_ptr<swss::DBConnector> m_db;
+    std::unique_ptr<swss::RedisPipeline> m_ntf_pipe;
     std::unique_ptr<swss::RedisPipeline> m_pipe;
```
qiluo-msft, line 81: >m_pipe [](http://example.com/codeflow?start=41&length=6)

will you rename m_pipe similar to m_ntf_pipe naming convention?

orchagent/response_publisher.h
```diff
@@ -57,8 +61,29 @@ class ResponsePublisher : public ResponsePublisherInterface
     void setBuffered(bool buffered);
 
   private:
+    struct entry
```
qiluo-msft, line 64: >entry [](http://example.com/codeflow?start=11&length=5)

Where can I find HLD?

orchagent/response_publisher.cpp
```diff
@@ -150,22 +210,70 @@ void ResponsePublisher::writeToDB(const std::string &table, const std::string &k
...
 
 void ResponsePublisher::flush()
 {
-    m_pipe->flush();
+    m_ntf_pipe->flush();
+    if (m_update_thread != nullptr)
+    {
+        {
+            std::lock_guard<std::mutex> lock(m_lock);
+            m_queue.push(entry{
```
qiluo-msft, line 228: >push [](http://example.com/codeflow?start=20&length=4)

push -> emplace ?

orchagent/response_publisher.cpp
```diff
@@ -150,22 +210,70 @@ void ResponsePublisher::writeToDB(const std::string &table, const std::string &k
...
+            e = m_queue.front();
+            m_queue.pop();
+        }
+        if (e.shutdown)
+        {
+            break;
+        }
+        if (e.flush)
+        {
+            m_pipe->flush();
```
qiluo-msft, line 272: >m_pipe [](http://example.com/codeflow?start=12&length=6)

You should not share DBConnector between threads. It is not thread-safe.

mint570: This pipe object is only used in the "DB update thread". It is not used in the main thread beside constructor.

The DB write operations are completed handled by the "DB update thread" if thread mode is enabled.
(If thread mode is disabled, the main thread will update the DB. There is no "DB update thread" in that case.)

orchagent/response_publisher.h
mint570: This is just internal implementation details. We don't have HLD for this.

The overall design is to put the DB update operation into a different thread for the response publisher.

In the detailed implementation here, we use a FIFO queue to store the DB update events. The main thread will queue up the event into the queue, and the "DB update thread" will read from the queue and process the DB update.
The "entry" struct here is the "DB update event".

Let me know if you need more details on this.

orchagent/response_publisher.h
mint570: That is a common practice for mutex object.

The key word "mutable" means that the variable can be changed in a const method.
If we have a method to read a variable that is protected by a mutex, it is nature to declare the method as const since it only does read operation. But the mutex object needs to be mutable for the read method to get the lock.

In this case, we might not really need it to be mutable. But we should follow the common practice.

tests/p4rt/test_p4rt_acl.py
```diff
@@ -241,26 +235,6 @@ def test_AclRulesAddUpdateDelPass(self, dvs, testlog):
         assert status == True
         util.verify_attr(fvs, attr_list)
 
-        # query application state database for ACL tables
```
prsunny, line 244: Since its deleted, where is the check for APPL_STATE_DB entries?

orchagent/response_publisher.h
qiluo-msft: we might not really need it to be mutable -> let's just remove it.

Thanks for the explanation!

qiluo-msft:
LGTM.

tests/p4rt/test_p4rt_acl.py
mint570: P4Orch will no longer write to APPL_STATE_DB, so there is no APPL_STATE_DB entries for P4RT table.
(P4Orch writes to APPL_DB instead, which is no change at the moment as APPL_DB is already written when orchagent pops the request entries.)

But P4Orch will still send the response to the Redis channel, which is checked and verified in util.verify_response().
And the test also checks the APPL_DB entries. For now it doesn't mean much. But in the future when we upstream the ZMQ changes, P4Orch will not write to APPL_DB in pop; it will write to APPL_DB in response path instead. At that time the P4RT table in APPL_DB will represent the state (not the intent).

prsunny (approved):

PR #3070, opened 2024-03-07
Set HOST_TX_READY_NOTIFY attribute only after query capabilities

noaOrMlnx:
**What I did**
Changed SAI_SWITCH_ATTR_PORT_HOST_TX_READY_NOTIFY callback attribute to be set only after query capabilities approved that CMIS mgmt. feature is enabled. 

**Why I did it**
Before that change, in orchagent/main.cpp, that attribute was set without checking if SAI version is not supporting this feature. 
This could cause an issue to unsupported SAI versions with supported SONiC versions. 

**How I verified it**
Made sure all functionality is the same as before, and this attribute set will not break not-supporting SAI versions.

**Details if related**

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -505,10 +505,6 @@ int main(int argc, char **argv)
     attr.value.ptr = (void *)on_switch_shutdown_request;
     attrs.push_back(attr);
 
-    attr.id = SAI_SWITCH_ATTR_PORT_HOST_TX_READY_NOTIFY;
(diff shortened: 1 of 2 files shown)


prsunny (dismissed):

keboliu (approved):

kperumalbfn (approved):

dprital (approved):

prgeor (approved):

prsunny (approved):

noaOrMlnx:
@yxieca can you please cherry-pick to 202311? 
Thanks

PR #3076, opened 2024-03-13
Fix flex counter out-of-order issue by notifying counter operations using SelectableChannel

stephenxs:
**What I did**

Fix flex counter out-of-order issue by notifying counter operations using SelectableChannel
Depends on https://github.com/sonic-net/sonic-sairedis/pull/1362

**Why I did it**

Currently, the operations of SAI objects and their counters (if any) are triggered by different channels, which introduces racing conditions:
- the creation and destruction of the objects are notified using the `SelectableChannel`,
- the operations of counters, including starting and stopping polling the counters, are notified by listening to the `FLEX_COUNTER` and `FLEX_COUNTER_GROUP` tables in the `FLEX_COUNTER_DB`
- The orchagent always respects the order when starting/stopping counter-polling (which means to start counter-polling after creating the object and to stop counter-polling before destroying the object) but `syncd` can receive events in a wrong order, eg. it receives destroying an object first and then stopping counter polling on the object, it can poll counter for a non-exist object, which causes errors in vendor SAI.

The new solution is to extend SAI redis attributes on the SAI_SWITCH_OBJECT to notify counter polling. As a result, all the objects and their counters are notified using a unified channel, the `SelectableChannel`.

**How I verified it**

Manual test
Regression test
Mock test

**Details if related**
1. Extend SAI redis attribute on the switch object to identify counter-polling operations.
2. Introduce a CLI option of orchagent to identify whether the new or the old approach is used for notifying counter-polling. It can not be changed on the fly.
3. Move the logic to notify counter polling from each orchagent class and flex counter manager class to a commonplace.    The logic is:
    - For the new approach, notify counter polling using SAI extended redis attribute
    - For the old approach, notify counter-polling using flex database tables. The corresponding flex database table objects (`ProducerTable`) are initialized during orchagent initialization, before any flex counter operations.
4. Remove the definition of and the references to flex counter and flex counter group table in each orchagent class (usually `m_flexCounterTable` and `m_flexCounterGroupTable`)
5. Improve the logic to load the counters' Lua plugin for mock test. The Lua scripts are not available in the mock test scenarios, which causes exceptions to load them in mock test. In case it fails to load the Lua plugin,
    - Originally, it skipped any flex counter group operations.
    - Now, it will leave the SHA code of the Lua plugin empty, and continue to notify the flex counter operations to SAI redis.
      - In the production scenario, where it can hardly happen, the plugin field will not be notified to SAI.
      - In the mock test scenario, it will notify SAI as it is. This is for mock test to verify whether the plugin field is provided as expected.
6. Gearbox counter handling.
    - Maintain a mapping from each OID created based on the gearbox, from the object ID itself to the gearbox's OID.
    - To notify counter operations for such OIDs in the new approach, we need to first fetch the corresponding gearbox OID from the mapping and the notify the gearbox OID as `switch OID`.
    - The old approach looks buggy.
      - There can be more than one gearbox syncd docker in the system, which is a P2MP scenario, the OA needs to communicate to multiple gearbox syncd dockers.
      - The old approach leverages `ConsumerTable`, `ProducerTable` mechanism to communicate between OA and sairedis. However, it works for P2P scenarios only. There is a logic for `ConsumerTable` to **consume** the update once a gearbox syncd sees it, leaving all rest gearbox syncd daemons to see nothing. As a result, for each update there is only one gearbox syncd that sees and handles it.
7. There is a logic in intfsorch, copporch and vxlanorch which is to add a counter for an object only after it occurs in the `VIRTORID` table, which is a WA of the counter out-of-order issue. Now that the issue has been fixed in the new approach, it does not need to check anymore.

***Performance analysis***
The counter operations are handled in the same thread in both the new and old solutions.
In swss, the counter operation was asynchronous in the old solution and is synchronous now, which can introduce a bit more latency. However, as the number of counter operations is small, no performance degradation is observed.

### DIFF
--- a/orchagent/flex_counter/flex_counter_manager.cpp
+++ b/orchagent/flex_counter/flex_counter_manager.cpp
@@ -18,6 +18,10 @@ using swss::DBConnector;
 using swss::FieldValueTuple;
 using swss::ProducerTable;
 
+extern sai_switch_api_t *sai_switch_api;
+
+extern sai_object_id_t gSwitchId;
+
 const string FLEX_COUNTER_ENABLE("enable");
 const string FLEX_COUNTER_DISABLE("disable");
 
@@ -89,13 +93,13 @@ FlexCounterManager::FlexCounterManager(
         const uint polling_interval,
         const bool enabled,
         FieldValueTuple fv_plugin) :
-    FlexCounterManager("FLEX_COUNTER_DB", group_name, stats_mode,
+    FlexCounterManager(false, group_name, stats_mode,
             polling_interval, enabled, fv_plugin)
 {
 }
 
 FlexCounterManager::FlexCounterManager(
-        const string& db_name,
+        const bool is_gearbox,
         const string& group_name,
         const StatsMode stats_mode,
         const uint polling_interval,
@@ -106,11 +110,7 @@ FlexCounterManager::FlexCounterManager(
     polling_interval(polling_interval),
     enabled(enabled),
     fv_plugin(fv_plugin),
-    flex_counter_db(new DBConnector(db_name, 0)),
-    flex_counter_group_table(new ProducerTable(flex_counter_db.get(),
-                FLEX_COUNTER_GROUP_TABLE)),
-    flex_counter_table(new ProducerTable(flex_counter_db.get(),
-                FLEX_COUNTER_TABLE))
+    is_gearbox(is_gearbox)
 {
     SWSS_LOG_ENTER();
 
@@ -125,13 +125,10 @@ FlexCounterManager::~FlexCounterManager()
 
     for (const auto& counter: installed_counters)
     {
-        flex_counter_table->del(getFlexCounterTableKey(group_name, counter));
+        stopFlexCounterPolling(counter.second, getFlexCounterTableKey(group_name, counter.first));
     }
 
-    if (flex_counter_group_table != nullptr)
-    {
-        flex_counter_group_table->del(group_name);
-    }
+    delFlexCounterGroup(group_name, is_gearbox);
 
     SWSS_LOG_DEBUG("Deleted flex counter group '%s'.", group_name.c_str());
 }
@@ -140,19 +137,13 @@ void FlexCounterManager::applyGroupConfiguration()
 {
     SWSS_LOG_ENTER();
 
-    vector<FieldValueTuple> field_values =
-    {
-        FieldValueTuple(STATS_MODE_FIELD, stats_mode_lookup.at(stats_mode)),
-        FieldValueTuple(POLL_INTERVAL_FIELD, std::to_string(polling_interval)),
-        FieldValueTuple(FLEX_COUNTER_STATUS_FIELD, status_lookup.at(enabled))
-    };
-
-    if (!fvField(fv_plugin).empty())
-    {
-        field_values.emplace_back(fv_plugin);
-    }
-
-    flex_counter_group_table->set(group_name, field_values);
+    setFlexCounterGroupParameter(group_name,
+                                 std::to_string(polling_interval),
+                                 stats_mode_lookup.at(stats_mode),
+                                 fvField(fv_plugin),
+                                 fvValue(fv_plugin),
+                                 status_lookup.at(enabled),
+                                 is_gearbox);
 }
 
 void FlexCounterManager::updateGroupPollingInterval(
@@ -160,11 +151,7 @@ void FlexCounterManager::updateGroupPollingInterval(
 {
     SWSS_LOG_ENTER();
 
-    vector<FieldValueTuple> field_values =
-    {
-        FieldValueTuple(POLL_INTERVAL_FIELD, std::to_string(polling_interval))
-    };
-    flex_counter_group_table->set(group_name, field_values);
+    setFlexCounterGroupPollInterval(group_name, std::to_string(polling_interval), is_gearbox);
 
     SWSS_LOG_DEBUG("Set polling interval for flex counter group '%s' to %d ms.",
             group_name.c_str(), polling_interval);
@@ -181,11 +168,7 @@ void FlexCounterManager::enableFlexCounterGroup()
         return;
     }
 
-    vector<FieldValueTuple> field_values =
-    {
-        FieldValueTuple(FLEX_COUNTER_STATUS_FIELD, FLEX_COUNTER_ENABLE)
-    };
-    flex_counter_group_table->set(group_name, field_values);
+    setFlexCounterGroupOperation(group_name, FLEX_COUNTER_ENABLE, is_gearbox);
     enabled = true;
 
     SWSS_LOG_DEBUG("Enabling flex counters for group '%s'.",
@@ -203,11 +186,7 @@ void FlexCounterManager::disableFlexCounterGroup()
         return;
     }
 
-    vector<FieldValueTuple> field_values =
-    {
-        FieldValueTuple(FLEX_COUNTER_STATUS_FIELD, FLEX_COUNTER_DISABLE)
-    };
-    flex_counter_group_table->set(group_name, field_values);
+    setFlexCounterGroupOperation(group_name, FLEX_COUNTER_DISABLE, is_gearbox);
     enabled = false;
 
     SWSS_LOG_DEBUG("Disabling flex counters for group '%s'.",
@@ -219,7 +198,8 @@ void FlexCounterManager::disableFlexCounterGroup()
 void FlexCounterManager::setCounterIdList(
         const sai_object_id_t object_id,
         const CounterType counter_type,
-        const unordered_set<string>& counter_stats)
+        const unordered_set<string>& counter_stats,
+        const sai_object_id_t switch_id)
 {
     SWSS_LOG_ENTER();
 
@@ -231,12 +211,12 @@ void FlexCounterManager::setCounterIdList(
         return;
     }
 
-    std::vector<swss::FieldValueTuple> field_values =
-    {
-        FieldValueTuple(counter_type_it->second, serializeCounterStats(counter_stats))
-    };
-    flex_counter_table->set(getFlexCounterTableKey(group_name, object_id), field_values);
-    installed_counters.insert(object_id);
+    auto key = getFlexCounterTableKey(group_name, object_id);
+    auto counter_ids = serializeCounterStats(counter_stats);
+    auto effective_switch_id = switch_id == SAI_NULL_OBJECT_ID ? gSwitchId : switch_id;
+
+    startFlexCounterPolling(effective_switch_id, key, counter_ids, counter_type_it->second);
+    installed_counters[object_id] = effective_switch_id;
 
     SWSS_LOG_DEBUG("Updated flex counter id list for object '%" PRIu64 "' in group '%s'.",
             object_id,
@@ -258,7 +238,8 @@ void FlexCounterManager::clearCounterIdList(const sai_object_id_t object_id)
         return;
     }
 
-    flex_counter_table->del(getFlexCounterTableKey(group_name, object_id));
+    auto key = getFlexCounterTableKey(group_name, object_id);
+    stopFlexCounterPolling(installed_counters[object_id], key);
     installed_counters.erase(counter_it);
 
     SWSS_LOG_DEBUG("Cleared flex counter id list for object '%" PRIu64 "' in group '%s'.",
@@ -272,7 +253,7 @@ string FlexCounterManager::getFlexCounterTableKey(
 {
     SWSS_LOG_ENTER();
 
-    return group_name + flex_counter_table->getTableNameSeparator() + sai_serialize_object_id(object_id);
+    return group_name + ":" + sai_serialize_object_id(object_id);
 }
 
 // serializeCounterStats turns a set of stats into a format suitable for FLEX_COUNTER_DB.
--- a/orchagent/flex_counter/flex_counter_manager.h
+++ b/orchagent/flex_counter/flex_counter_manager.h
@@ -12,7 +12,6 @@
 
 extern "C" {
 #include "sai.h"
-#include "sairedis.h"
 }
 
 enum class StatsMode
--- a/orchagent/flex_counter/flex_counter_manager.h
+++ b/orchagent/flex_counter/flex_counter_manager.h
@@ -12,6 +12,7 @@
 
 extern "C" {
 #include "sai.h"
+#include "sairedis.h"
 }
 
 enum class StatsMode
@@ -53,7 +54,7 @@ class FlexCounterManager
         {}
 
         FlexCounterManager(
-                const std::string& db_name,
+                const bool is_gearbox,
                 const std::string& group_name,
                 const StatsMode stats_mode,
                 const uint polling_interval,
@@ -71,7 +72,8 @@ class FlexCounterManager
         void setCounterIdList(
                 const sai_object_id_t object_id,
                 const CounterType counter_type,
-                const std::unordered_set<std::string>& counter_stats);
+                const std::unordered_set<std::string>& counter_stats,
+                const sai_object_id_t switch_id=SAI_NULL_OBJECT_ID);
         void clearCounterIdList(const sai_object_id_t object_id);
 
         const std::string& getGroupName() const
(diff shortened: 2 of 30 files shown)


kcudnik:
Please fix build issues

stephenxs:
> Please fix build issues
Hi @kcudnik 
It depends on the sairedis PR (https://github.com/sonic-net/sonic-sairedis/pull/1362) to define extended attributes.

kcudnik:
Will review

stephenxs:
As mentioned in the description, the PR depends on sairedis PR (https://github.com/sonic-net/sonic-sairedis/pull/1362) merged a week ago.
However, the sairedis Debian packages fetched in the build were TOO OLD to satisfy the dependency.
```
##[debug]ProjectId: be1b070f-be15-4154-aade-b1d3bfb17054
##[debug]PipelineVersionToDownload: latestFromBranch
##[debug]Agent environment resources - Disk: / Available 25389.00 MB out of 29593.00 MB, Memory: Used 701.00 MB out of 15968.00 MB, CPU: Usage 2.53%
##[debug]PipelineId from non-trigerringBuild: 495079
Download from the specified build: #495079
##[debug]Processed: ##vso[task.setvariable variable=BuildNumber;issecret=False;]495079
Download artifact to: /data/work/1/a/download/sairedis
Using default max parallelism.
```
It fetched `495079` in each build. However, it was built on 9th March, more than 2 weeks before my sairedis PR. I think this is because there has been no successful build since `495079`. Can anyone help to look into it?
```
511120	20240330.8	master	Azure.sonic-sairedis	failed	2024-03-30T09:00:59	2024-03-30T10:55:25	594944dbe4	[Build Link](https://dev.azure.com/mssonic/be1b070f-be15-4154-aade-b1d3bfb17054/_build/results?buildId=511120)	[Artifacts](https://sonic-build.azurewebsites.net/ui/sonic/pipelines/12/builds/511120/artifacts?branchName=master)
508223	20240326.3	master	Azure.sonic-sairedis	failed	2024-03-26T14:26:18	2024-03-26T15:20:54	594944dbe4	[Build Link](https://dev.azure.com/mssonic/be1b070f-be15-4154-aade-b1d3bfb17054/_build/results?buildId=508223)	[Artifacts](https://sonic-build.azurewebsites.net/ui/sonic/pipelines/12/builds/508223/artifacts?branchName=master)
505863	20240323.8	master	Azure.sonic-sairedis	failed	2024-03-23T09:01:19	2024-03-23T10:32:08	bb948f632b	[Build Link](https://dev.azure.com/mssonic/be1b070f-be15-4154-aade-b1d3bfb17054/_build/results?buildId=505863)	[Artifacts](https://sonic-build.azurewebsites.net/ui/sonic/pipelines/12/builds/505863/artifacts?branchName=master)
500443	20240316.8	master	Azure.sonic-sairedis	failed	2024-03-16T09:01:01	2024-03-16T10:48:54	bb948f632b	[Build Link](https://dev.azure.com/mssonic/be1b070f-be15-4154-aade-b1d3bfb17054/_build/results?buildId=500443)	[Artifacts](https://sonic-build.azurewebsites.net/ui/sonic/pipelines/12/builds/500443/artifacts?branchName=master)
```

orchagent/p4orch/tests/test_main.cpp
```diff
@@ -240,3 +241,56 @@ int main(int argc, char *argv[])
...
+
+void setFlexCounterGroupParameter(const std::string &group,
+                                  const std::string &poll_interval,
+                                  const std::string &stats_mode,
+                                  const std::string &plugin_name,
+                                  const std::string &plugins,
+                                  const std::string &operation,
+                                  bool is_gearbox)
+{
+    return;
```
kcudnik, line 253: why those functions are emptY ?

orchagent/saihelper.cpp
```diff
@@ -800,3 +827,270 @@ void handleSaiFailure(bool abort_on_failure)
...
+    initSaiRedisCounterEmptyParameter(flex_counter_group_param.stats_mode);
+    initSaiRedisCounterEmptyParameter(flex_counter_group_param.plugin_name);
+    initSaiRedisCounterEmptyParameter(flex_counter_group_param.plugins);
+}
+
+inline void initSaiRedisCounterParameterFromString(sai_s8_list_t &sai_s8_list, const std::string &str)
+{
+    if (str.length() > 0)
+    {
+        sai_s8_list.list = (int8_t*)const_cast<char *>(str.c_str());
```
kcudnik, line 851: i think this is bad practice, sine you need to make sure that passed str will not get out of scope when s8 list will be used

orchagent/p4orch/tests/test_main.cpp
stephenxs: The Flex counter handling logic is not invoked by p4orch. These functions are stubs for link.
We usually add mock test in mock/tests folder and add empty functions as stubs in p4orch/tests folder.

orchagent/saihelper.cpp
stephenxs: I understand it's not a best practice but it doesn't look like there is another way to initialize sai_s8_list from a std::string. we have to use passing by reference here.
maybe we can make all the functions to initialize parameters (starting with `initSaiRedis`) `static` to avoid being called outside the file.
all the functions calling them are like this

1. call `initSaiRedisXXX` to initialize sai_s8_list from std::string
2. call `notifySyncdXXXX` to notify SAI redis

both std::string and sai_s8_list variables are local to the function and are safe.

how do you think or any suggestions? @kcudnik

kcudnik: making those static should be better

stephenxs: Done.

orchagent/vxlanorch.cpp
```diff
@@ -1219,7 +1220,10 @@ VxlanTunnelOrch::VxlanTunnelOrch(DBConnector *statedb, DBConnector *db, const st
     m_tunnelNameTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_TUNNEL_NAME_MAP));
     m_tunnelTypeTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_TUNNEL_TYPE_MAP));
 
-    m_vidToRidTable = unique_ptr<Table>(new Table(m_asic_db.get(), "VIDTORID"));
+    if (gTraditionalFlexCounter)
+    {
+        m_vidToRidTable = unique_ptr<Table>(new Table(m_asic_db.get(), "VIDTORID"));
```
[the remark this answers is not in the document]
stephenxs: Done.

orchagent/flex_counter/flex_counter_manager.cpp
```diff
@@ -272,7 +253,7 @@ string FlexCounterManager::getFlexCounterTableKey(
 {
     SWSS_LOG_ENTER();
 
-    return group_name + flex_counter_table->getTableNameSeparator() + sai_serialize_object_id(object_id);
+    return group_name + ":" + sai_serialize_object_id(object_id);
```
kcudnik, line 256: why getTableNameSeparator was removed and replaced with explici ":'"? previous version is more readable and convinient

orchagent/flex_counter/flex_counter_manager.h
```diff
@@ -12,6 +12,7 @@
 
 extern "C" {
 #include "sai.h"
+#include "sairedis.h"
```
kcudnik, line 15: why this is needed here ?

orchagent/intfsorch.cpp
```diff
@@ -64,44 +64,40 @@ IntfsOrch::IntfsOrch(DBConnector *db, string tableName, VRFOrch *vrf_orch, DBCon
...
     /* Initialize COUNTER_DB tables */
     m_rifNameTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_RIF_NAME_MAP));
     m_rifTypeTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_RIF_TYPE_MAP));
 
-    m_vidToRidTable = unique_ptr<Table>(new Table(m_asic_db.get(), "VIDTORID"));
+    if (gTraditionalFlexCounter)
+    {
+        m_vidToRidTable = unique_ptr<Table>(new Table(m_asic_db.get(), "VIDTORID"));
+    }
+
```
kcudnik, line 76: this table should never be accessed via orchagent, as for OA knowledge this table don't exits

orchagent/saihelper.h
```diff
@@ -4,10 +4,13 @@
 
 #include <string>
 #include "orch.h"
+#include "producertable.h"
+#include <sairedis.h>
```
kcudnik, line 8: this should be externed C

orchagent/vxlanorch.cpp
```diff
@@ -1219,7 +1220,10 @@ VxlanTunnelOrch::VxlanTunnelOrch(DBConnector *statedb, DBConnector *db, const st
     m_tunnelNameTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_TUNNEL_NAME_MAP));
     m_tunnelTypeTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_TUNNEL_TYPE_MAP));
 
-    m_vidToRidTable = unique_ptr<Table>(new Table(m_asic_db.get(), "VIDTORID"));
+    if (gTraditionalFlexCounter)
+    {
+        m_vidToRidTable = make_unique<Table>(m_asic_db.get(), "VIDTORID");
+    }
```
kcudnik, line 1226: again as for OA this should not exists

orchagent/flex_counter/flex_counter_manager.cpp
stephenxs: In the new solution, we do not have `flex_counter_table` anymore.

orchagent/intfsorch.cpp
stephenxs: this is the WA of the out of order issue. it is accessed in intfsorch, copporch and vxlanorch for the same proposes.
now, we initialize it only for traditional flex counter model.

orchagent/flex_counter/flex_counter_manager.h
stephenxs: looks like a left over. will remove it.

orchagent/saihelper.h
stephenxs: I will move it to the `extern C` block in `saihelp.c`. For other cpp files, they do not need this header.

orchagent/vxlanorch.cpp
stephenxs: only in traditional model as a WA of the out of order issue.

orchagent/saihelper.h
stephenxs: fixed

orchagent/intfsorch.cpp
stephenxs: fixed

orchagent/flex_counter/flex_counter_manager.h
stephenxs: fixed

stephenxs:
Rebased to resolve the conflict.

orchagent/saihelper.cpp
```diff
@@ -85,6 +86,16 @@ sai_dash_direction_lookup_api_t*    sai_dash_direction_lookup_api;
...
+extern bool gTraditionalFlexCounter;
+
+vector<sai_object_id_t> gGearboxOids;
+
+unique_ptr<DBConnector> gFlexCounterDb;
+unique_ptr<ProducerTable> gFlexCounterGroupTable;
+unique_ptr<ProducerTable> gFlexCounterTable;
+unique_ptr<DBConnector> gGearBoxFlexCounterDb;
+unique_ptr<ProducerTable> gGearBoxFlexCounterGroupTable;
+unique_ptr<ProducerTable> gGearBoxFlexCounterTable;
```
kcudnik, line 98: seems like all this is related, so it could be moved to struct/class, but it seems for now that this could be like that, since OA needs ad ground up refactoring to move all global stuff to proper classes/namespaces

orchagent/saihelper.cpp
```diff
@@ -261,6 +272,20 @@ void initSaiApi()
     sai_log_set(SAI_API_TWAMP,                  SAI_LOG_LEVEL_NOTICE);
 }
 
+void initFlexCounterTables()
```
kcudnik, line 275: yea, looks like a constructor task for proposed class :D

kcudnik, line 275: should this be static inline ?

orchagent/saihelper.cpp
```diff
@@ -803,3 +831,270 @@ void handleSaiFailure(bool abort_on_failure)
...
+static inline void initSaiRedisCounterParameterFromString(sai_s8_list_t &sai_s8_list, const std::string &str)
+{
+    if (str.length() > 0)
+    {
+        sai_s8_list.list = (int8_t*)const_cast<char *>(str.c_str());
+        sai_s8_list.count = (uint32_t)str.length();
+    }
+    else
+    {
+        sai_s8_list.list = nullptr;
```
kcudnik, line 860: ehter use initSaiRedisCounterEmptyParameter or remove that function

orchagent/saihelper.cpp
```diff
@@ -803,3 +831,270 @@ void handleSaiFailure(bool abort_on_failure)
...
+        sai_s8_list.count = 0;
+    }
+}
+
+static inline void notifySyncdCounterOperation(bool is_gearbox, const sai_attribute_t &attr)
+{
+    if (sai_switch_api == nullptr)
+    {
+        // This can happen during destruction of the orchagent daemon.
```
kcudnik, line 869: why syncd operation counter would be notifying syncd when OA is destroyed ? sai_switch_api  should be destroyed at the end, only scenario i can imaged, is that SAI API init failed and sai_switch_api was NULL yet, but in that case no other operations should be performed and OA should quit, if switch API is NULL you can do nothing, that means sairedis lib initialization failed, so you will not send any notification anywaahre, and such initialization failre should be logged with critical prioirty not just error (not this log)

orchagent/saihelper.cpp
```diff
@@ -803,3 +831,270 @@ void handleSaiFailure(bool abort_on_failure)
...
+        {
+            sai_switch_api->set_switch_attribute(gearbox_oid, &attr);
+        }
+    }
+}
+
+static inline void operateFlexCounterDbSingleField(std::vector<FieldValueTuple> &fvTuples,
+                                            const string &field, const string &value)
+{
+    if (!field.empty() && !value.empty())
```
kcudnik, line 890: is this prevention here for a reason? ot at some point OA can actually populate empty counter? if any of those is empty this should be logged as ERROR

orchagent/saihelper.cpp
```diff
@@ -803,3 +831,270 @@ void handleSaiFailure(bool abort_on_failure)
...
+                                  const string &plugin_name,
+                                  const string &plugins,
+                                  const string &operation,
+                                  bool is_gearbox)
+{
+    if (gTraditionalFlexCounter)
+    {
+        operateFlexCounterGroupDatabase(group, poll_interval, stats_mode, plugin_name, plugins, operation, is_gearbox);
+        return;
+    }
```
kcudnik, line 926: so, normally in the OOP, such cases is resolved in proper way, by having base abstract class, and pointer in that class in OA, then base on command line in OA whether we want to use traditional counters or new one, just one of 2 objects is createed and populate the pointer, then all operations are the same, add field to counder, add counter remove counter etc, and in all entire code there is no need for gTraditionalFlexCounter) check, and i saw is checked in multiple places, and old and new counter realise the same functionality at the end, but im aware that at this stage this conversion without refactoring OA maybe not possible, it should be designed from the start

kcudnik: yea, look all the below functions, they all contain the same check, and adding new code could confuse unaware programmer, so moving this to 2 classes with one new and second ol would be great solution to that

kcudnik (approved):
i believe this will work, but some of changes i proposed would be desired

dprital:
@qiluo-msft @prsunny , Can you please approve and merge ?

liat-grozovik:
@prsunny  please help to merge now that PR checkers are passing and approval is granted.

orchagent/saihelper.cpp
stephenxs: This is for mock test scenario only. 
In production scenario, the test `if (sai_switch_api == nullptr)` won't be true.

orchagent/saihelper.cpp
stephenxs: This is also for mock test.
The common logic to install plugin is like this:
1. load lua plugin and generate SHA code
2. install the lua plugin into {`plugin name`: `plugin SHA code`} the flex counter group

In mock test scenario, it fails to load the Lua plugin, which leaves the SHA code empty string. But we still want the lua plugin to be installed into the flex counter group. By doing so, we can verify whether the plugin name is correct and meet the coverage requirement.

In production scenario, we will check whether either is empty, if so, we will not install the plugin. This is just a production and can not happen.

orchagent/saihelper.cpp
stephenxs: Fixed

kcudnik (approved):

prsunny (approved):

PR #3080, opened 2024-03-20
Clang format change.

mint570:
**What I did**
This PR has no real code change. It is purely clang formatting. It only applies to the P4Orch codes.
Commands that I run:
find orchagent/p4orch -name *.h -o -name *.cpp | xargs clang-format -i -style="{BasedOnStyle: Microsoft, DerivePointerAlignment: false}"
find orchagent -name response_publisher* -o -name return_code.h | xargs clang-format -i -style="{BasedOnStyle: Microsoft, DerivePointerAlignment: false}"

**Why I did it**
In Google, we follow the Google clang format style, but SONiC uses Microsoft style. In order to upstream our changes, we have to translate the SONiC code into Google style first, then we can cherry-pick our changes. After that, we translate back to Microsoft style. The existing codes do not fully follow the clang Microsoft style. And hence causes many unnecessary file changes in a PR: https://github.com/sonic-net/sonic-swss/pull/3066.

**How I verified it**
N/A

**Details if related**
This PR might not meet the code coverage requirement. But it is purely formatting changes.

### DIFF
--- a/orchagent/p4orch/acl_rule_manager.cpp
+++ b/orchagent/p4orch/acl_rule_manager.cpp
@@ -9,7 +9,6 @@
 #include "crmorch.h"
 #include "dbconnector.h"
 #include "intfsorch.h"
-#include <nlohmann/json.hpp>
 #include "logger.h"
 #include "orch.h"
 #include "p4orch.h"
@@ -18,6 +17,7 @@
 #include "sai_serialize.h"
 #include "table.h"
 #include "tokenize.h"
+#include <nlohmann/json.hpp>
 extern "C"
 {
 #include "sai.h"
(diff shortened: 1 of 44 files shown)


prsunny (approved):

prsunny:
@mint570 , just realized, we need to have a consistent style. Understand that a later PR is converting it back but I think we can consider this in review and close this. What do you say?

mint570:
@prsunny what do you mean by "consistent style"? SONiC code today follow the Microsoft clang style mostly, but not exactly. We are using Microsoft style for code upstream by running the clang command in the PR description. This will ensure the same style on our PRs. We are doing this for P4Orch codes only. We are not doing this for other SWSS codes.

Or do you mean we can review https://github.com/sonic-net/sonic-swss/pull/3066 directly without this PR? The format change in that PR makes it very hard to review the real change. And we don't know the code coverage error in that PR is due to format change or not. If that's what you mean, I will suggest to merge this PR first. I can handle the rebasing on the other PR. Once this PR is merged, our later PRs will be much cleaner.

prsunny (approved):

PR #3082, opened 2024-03-20
[202311][muxorch] Fixing cache bug in updateRoute logic and Fixing bug with multiple routes pointing to nhg

Ndancejic:
**What I did**
Manual cherry-pick of https://github.com/sonic-net/sonic-swss/pull/2982 and https://github.com/sonic-net/sonic-swss/pull/2999

- Fixed a bug where cache was not removing entry when an ecmp mux route gets updated to a single nexthop route. And added test case to cover the edge case.

**Why I did it**
MSFT ADO - 26099588

**How I verified it**

**Details if related**
will fix dualtor tests on 202311

ADO:#26458505
ADO:#26099588
ADO:#26099588

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -1099,6 +1099,8 @@ void MuxOrch::updateRoute(const IpPrefix &pfx, bool add)
 
     if (!add)
     {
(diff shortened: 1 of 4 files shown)


prsunny (approved):

prsunny:
Could you please fix title and description? Copy from the original PRs

PR #3087, opened 2024-03-25
[portsorch] Handle TRANSCEIVER_INFO table on warm boot

stepanblyschak:
**What I did**

Process TRANSCEIVER_INFO STATE_DB table on warm start to configure SAI_PORT_ATTR_HOST_TX_SIGNAL_ENABLE on warm boot.

**Why I did it**

Fix an issue that the port becomes operationally down after APPLY_VIEW.

**How I verified it**

Run warm-reboot.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -3391,6 +3391,7 @@ bool PortsOrch::bake()
     addExistingData(APP_LAG_MEMBER_TABLE_NAME);
     addExistingData(APP_VLAN_TABLE_NAME);
     addExistingData(APP_VLAN_MEMBER_TABLE_NAME);
+    addExistingData(STATE_TRANSCEIVER_INFO_TABLE_NAME);
 
     return true;
 }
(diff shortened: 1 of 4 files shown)


vaibhavhd:
Please modify existing UTs or add new UT to cover for this new table.

prgeor:
@stepanblyschak why did you choose `TRANSCEIVER_INFO STATE_DB` to be restored during warm reboot in OA instead of in Xcvrd? Isn't it natural for Xcvrd to restore the table during Xcvrd's warm reboot since this table is owned or populated by Xcvrd and not OA?

stepanblyschak:
@prgeor You are correct that xcvrd manages this table and it is xcvid's responsibility to keep it in DB when doing warm-reboot. This is done in https://github.com/sonic-net/sonic-platform-daemons/pull/454.
This change, just makes orchagent consume that information from DB in the new life.

prgeor (approved):

PR #3089, opened 2024-03-25
Add force option for unisolate command

jfeng-arista:
**What I did**
Add force option to the unisolate link command, so users can make the links not isolate if they want. 

**Why I did it**

**How I verified it**

**Details if related**


jfeng-arista:
@vmittal-msft @arlakshm

arlakshm (approved):
changes lgtm, can we merge now or we need to wait for the buildimage change to be merged

jfeng-arista:
we depend on https://github.com/sonic-net/sonic-buildimage/pull/18447, if we can merge 18447 up first, will be great

vmittal-msft (approved):

PR #3094, opened 2024-03-28
Apply port speed again after disabling port AN

PJHsieh:
**What I did**
Apply port speed again after disabling port AN.

**Why I did it**
In PortsOrch, the port speed configuration will not be set to SAI when port AN is enabled.
Changing the port speed while port AN is enabled will not take effect after port AN is disabled.

**How I verified it**

1. Enable port AN.
2. Set port speed from speed1 to speed2
3. Disable port AN.
4. Make sure port speed is speed2 instead of speed1.

**Details if related**


PJHsieh:
Hi, @prgeor , @prsunny , the test error is fixed.

Please review and approve this PR.

PR #3099, opened 2024-04-03
Fix setting host_tx_ready from initializePort()

noaOrMlnx:
**What I did**
During port initialization, we now set Host Tx Ready key-val to DB using "hset()" instead of using the new function "setHostTxReady()"

**Why I did it**
Port must be initialized before calling setHostTxReady(). if it is not initialized, it won't find the port object and function won't set host_tx_ready. 

**How I verified it**
boot switch with new changes.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -5464,7 +5464,7 @@ bool PortsOrch::initializePort(Port &port)
         string hostTxReadyStr = hostTxReadyVal ? "true" : "false";
 
         SWSS_LOG_DEBUG("Received host_tx_ready current status: port_id: 0x%" PRIx64 " status: %s", port.m_port_id, hostTxReadyStr.c_str());
-	m_portStateTable.hset(port.m_alias, "host_tx_ready", hostTxReadyStr);
+        m_portStateTable.hset(port.m_alias, "host_tx_ready", hostTxReadyStr);
     }
 
     /*
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -5423,7 +5423,7 @@ bool PortsOrch::initializePort(Port &port)
         string hostTxReadyStr = hostTxReadyVal ? "true" : "false";
 
         SWSS_LOG_DEBUG("Received host_tx_ready current status: port_id: 0x%" PRIx64 " status: %s", port.m_port_id, hostTxReadyStr.c_str());
-        setHostTxReady(port.m_port_id, hostTxReadyStr);
+	m_portStateTable.hset(port.m_alias, "host_tx_ready", hostTxReadyStr);
     }
 
     /*


orchagent/portsorch.cpp
```diff
@@ -5464,7 +5464,7 @@ bool PortsOrch::initializePort(Port &port)
         string hostTxReadyStr = hostTxReadyVal ? "true" : "false";
 
         SWSS_LOG_DEBUG("Received host_tx_ready current status: port_id: 0x%" PRIx64 " status: %s", port.m_port_id, hostTxReadyStr.c_str());
-        setHostTxReady(port.m_port_id, hostTxReadyStr);
+	m_portStateTable.hset(port.m_alias, "host_tx_ready", hostTxReadyStr);
```
stepanblyschak, line 5467: Please correct indent, should be using spaces

noaOrMlnx: Done

stepanblyschak (approved):

dprital:
@prsunny @prgeor , Can you please review ?

prsunny:
I feel there is a gap here. Why was it missed in VS test? How did we find this issue?

noaOrMlnx:
@prsunny 
We found out about this issue while seeing the following log error - 
ERR swss#orchagent: :- setHostTxReady: Failed to get port object for port id 0x1000000000015

the issue here is that setHostTxReady requires the port to be initialize and that is why it can’t be called from initialization.

prsunny:
@prgeor , please approve

noaOrMlnx:
Hi @prgeor, a kind reminder

prgeor:
> @prsunny We found out about this issue while seeing the following log error - ERR swss#orchagent: :- setHostTxReady: Failed to get port object for port id 0x1000000000015
> 
> the issue here is that setHostTxReady requires the port to be initialize and that is why it can’t be called from initialization.

@noaOrMlnx How Host Tx signal be ready without port fully initialized?

orchagent/portsorch.cpp
```diff
@@ -5465,7 +5465,7 @@ bool PortsOrch::initializePort(Port &port)
         string hostTxReadyStr = hostTxReadyVal ? "true" : "false";
 
         SWSS_LOG_DEBUG("Received host_tx_ready current status: port_id: 0x%" PRIx64 " status: %s", port.m_port_id, hostTxReadyStr.c_str());
-        setHostTxReady(port.m_port_id, hostTxReadyStr);
+        m_portStateTable.hset(port.m_alias, "host_tx_ready", hostTxReadyStr);
```
prgeor, line 5468: @noaOrMlnx please check above in this functin, port.m_port_id is being used ...so all those are also failing?

noaOrMlnx: @prgeor 
The issue here is not port.X, but the fact that inside setHostTxReady() function, we call getPort(). 
getPort cannot return the port object until port is initialize.

prgeor: @noaOrMlnx Can you please remove the getPort() from setHostTxReady()? This way all host_tx_ready change will be contained in one single function

prgeor: There is no need for portid inside setHostTxReady,

`PortsOrch::setHostTxReady(swss::Port, const std::string &status)` and modify the syslog to show only Alias (no need for port ID)

noaOrMlnx:
Hi @prgeor @prsunny 
I  opened a new PR that does what you suggested. please check https://github.com/sonic-net/sonic-swss/pull/3133

closing the current PR.
Thanks

PR #3101, opened 2024-04-04
[EVPN]Fix fpmsyncd crash when EVPN type5 is received with bgp fib suppression enabled

dgsudharsan:
**What I did**
Fix fpmsyncd crash when evpn is configured with fib suppression enabled. When protocol is not set, it's value is empty and fpmsyncd throws exception trying to covert to int and exits https://github.com/sonic-net/sonic-swss/blob/04912ad05aab8f4ebc40bcec70f00492a886e2ce/fpmsyncd/routesync.cpp#L1437. Protocol field was not extracted in the evpn flow. Decoding protocol from raw message fixed the issue.

**Why I did it**
To handle BGP suppress fib and evpn interop

**How I verified it**
Manually verified the fix. Added UT to cover.

**Details if related**

### DIFF
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -465,6 +465,7 @@ void RouteSync::onEvpnRouteMsg(struct nlmsghdr *h, int len)
                 inet_ntop(rtm->rtm_family, dstaddr, buf, MAX_ADDR_SIZE), dst_len);
     }
 
+    auto proto_str = getProtocolString(rtm->rtm_protocol);
     SWSS_LOG_INFO("Receive route message dest ip prefix: %s Op:%s", 
                     destipprefix,
                     nlmsg_type == RTM_NEWROUTE ? "add":"del");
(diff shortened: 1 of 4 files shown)


stepanblyschak (approved):

fpmsyncd/routesync.cpp
```diff
@@ -550,17 +551,20 @@ void RouteSync::onEvpnRouteMsg(struct nlmsghdr *h, int len)
     FieldValueTuple intf("ifname", intf_list);
     FieldValueTuple vni("vni_label", vni_list);
     FieldValueTuple mac("router_mac", mac_list);
+    FieldValueTuple proto("protocol", proto_str);
```
prsunny, line 554: as i see, this is only for evpn routes and not for regular routes, right? and no impact to warmboot etc

dgsudharsan: That's correct. The flow will happen only for EVPN. No impact to warmboot.

prsunny (approved):

PR #3105, opened 2024-04-07
[Orchagent] Recursive nexthop group enhancement

utpalkantpintoo:
**What I did**

Implemented recursive nexthop group enhancement in Orchagent (NhgOrch). They have been proposed in:
https://github.com/sonic-net/SONiC/pull/1636

**Why I did it**

These changes are required to handle a new field - "nexthop_group" in the App DB NEXT_HOP_GROUP_TABLE. Such nexthop groups are called recursive nexthop groups. This field contains the list of member (singleton) nexthop groups. Such recursive nexthop groups are represented by NEXT_HOP_GROUP objects.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/nhgorch.cpp
+++ b/orchagent/nhgorch.cpp
@@ -84,12 +84,18 @@ void NhgOrch::doTask(Consumer& consumer)
                         is_recursive = true;
                 }
             }
-
+            /* A NHG should not have both regular(ip/alias) and recursive fields */
+            if (is_recursive && (!ips.empty() || !aliases.empty()))
+            {
+                SWSS_LOG_ERROR("Nexthop group %s has both regular(ip/alias) and recursive fields", index.c_str());
+                it = consumer.m_toSync.erase(it);
+                continue;
+            }
             /* Split ips and aliases strings into vectors of tokens. */
             vector<string> ipv = tokenize(ips, ',');
             vector<string> alsv = tokenize(aliases, ',');
             vector<string> mpls_nhv = tokenize(mpls_nhs, ',');
-            vector<string> nhgv = tokenize(nhgs, ',');
+            vector<string> nhgv = tokenize(nhgs, NHG_DELIMITER);
 
             /* Create the next hop group key. */
             string nhg_str;
@@ -131,7 +137,7 @@ void NhgOrch::doTask(Consumer& consumer)
                     if (nhgs.empty())
                         nhgs = nhgm;
                     else
-                        nhgs += ',' + nhgm;
+                        nhgs += NHG_DELIMITER + nhgm;
                 }
                 if (invalid_member)
                 {
@@ -146,7 +152,7 @@ void NhgOrch::doTask(Consumer& consumer)
                 }
 
                 /* Form nexthopgroup key with the nexthopgroup keys of available members */
-                nhgv = tokenize(nhgs, ',');
+                nhgv = tokenize(nhgs, NHG_DELIMITER);
 
                 for (uint32_t i = 0; i < nhgv.size(); i++)
                 {
@@ -173,6 +179,8 @@ void NhgOrch::doTask(Consumer& consumer)
             /* If the group does not exist, create one. */
             if (nhg_it == m_syncdNextHopGroups.end())
             {
+                SWSS_LOG_INFO("Create nexthop group %s with %s", index.c_str(), nhg_str.c_str());
+
                 /*
                 * If we've reached the NHG limit, we're going to create a temporary
                 * group, represented by one of it's NH only until we have
@@ -230,6 +238,8 @@ void NhgOrch::doTask(Consumer& consumer)
             /* If the group exists, update it. */
             else
             {
+                SWSS_LOG_INFO("Update nexthop group %s with %s", index.c_str(), nhg_str.c_str());
+
                 const auto& nhg_ptr = nhg_it->second.nhg;
 
                 /*
@@ -462,14 +472,6 @@ sai_object_id_t NextHopGroupMember::getNhId() const
     if (m_key.isIntfNextHop())
     {
         nh_id = gIntfsOrch->getRouterIntfsId(m_key.alias);
-
-        if ((nh_id == SAI_NULL_OBJECT_ID) &&
-            !m_key.alias.compare(0, strlen("Loopback"), "Loopback"))
-        {
-            Port cpu_port;
-            gPortsOrch->getCpuPort(cpu_port);
-            nh_id = cpu_port.m_port_id;
-        }
     }
     else if (gNeighOrch->hasNextHop(m_key))
     {
--- a/orchagent/nhgorch.cpp
+++ b/orchagent/nhgorch.cpp
@@ -637,8 +637,8 @@ bool NextHopGroup::sync()
         return true;
     }
 
-    /* If the group is non-recursive, the group ID will be the only member's NH ID */
-    if (!isRecursive())
+    /* If the group is non-recursive with single member, the group ID will be the only member's NH ID */
+    if (!isRecursive() && (m_members.size() == 1))
     {
         const NextHopGroupMember& nhgm = m_members.begin()->second;
         sai_object_id_t nhid = nhgm.getNhId();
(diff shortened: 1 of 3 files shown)


hasan-brcm (approved):

orchagent/nhgorch.cpp
```diff
@@ -73,24 +76,96 @@ void NhgOrch::doTask(Consumer& consumer)
...
+                }
+                /* If no members are present */
+                if (nhgs.empty())
+                {
+                    it++;
+                    continue;
+                }
+
+                /* Form nexthopgroup key with the nexthopgroup keys of available members */
+                nhgv = tokenize(nhgs, ',');
```
Gokulnath-Raja, line 149: Can we use NHG_DELIMITER instead of ","

orchagent/nhgorch.cpp
```diff
@@ -73,24 +76,96 @@ void NhgOrch::doTask(Consumer& consumer)
...
+                                nhgm.c_str(), index.c_str());
+
+                        invalid_member = true;
+                        break;
+                    }
+                    /* Keep only the members which exist in the local cache */
+                    if (nhgs.empty())
+                        nhgs = nhgm;
+                    else
+                        nhgs += ',' + nhgm;
```
Gokulnath-Raja, line 134: Can we use NHG_DELIMITER instead of ","

orchagent/nhgorch.cpp
```diff
@@ -687,6 +807,9 @@ bool NextHopGroup::syncMembers(const std::set<NextHopKey>& nh_keys)
 {
     SWSS_LOG_ENTER();
 
+    /* This method should not be called for single-membered non-recursive nexthop groups */
+    assert(isRecursive() || (m_members.size() > 1));
```
Gokulnath-Raja, line 811: Instead of assert can we handle as exception?

orchagent/nhgorch.cpp
```diff
@@ -532,11 +637,8 @@ bool NextHopGroup::sync()
         return true;
     }
 
-    /*
-     * If the group is temporary, the group ID will be the only member's NH
-     * ID.
-     */
-    if (m_is_temp)
+    /* If the group is non-recursive with single member, the group ID will be the only member's NH ID */
+    if (!isRecursive() && (m_members.size() == 1))
```
Gokulnath-Raja, line 641: Do we need m_is_temp check?

orchagent/nhgorch.cpp
```diff
@@ -367,7 +459,19 @@ sai_object_id_t NextHopGroupMember::getNhId() const
 
     sai_object_id_t nh_id = SAI_NULL_OBJECT_ID;
 
-    if (gNeighOrch->hasNextHop(m_key))
+    if (m_key.isIntfNextHop())
+    {
+        nh_id = gIntfsOrch->getRouterIntfsId(m_key.alias);
+
+        if ((nh_id == SAI_NULL_OBJECT_ID) &&
```
Gokulnath-Raja, line 466: We don't consider loopback as nexthop and trap packet to CPU during route add https://github.com/sonic-net/sonic-swss/blob/master/orchagent/routeorch.cpp#L1789, so do we need to add this CPU port for NHG?

orchagent/nhgorch.cpp
utpalkantpintoo: That'll be redundant as the current check will be true for "temp" nhgs as well.

orchagent/nhgorch.cpp
utpalkantpintoo: Will make the NhgOrch behavior in-line with RouteOrch handling.
But I see an issue where if we receive such a msg, we'll end up holding the msgs in the corres. m_toSync queue(s) till the route is withdrawn.

orchagent/nhgorch.cpp
```diff
@@ -73,24 +76,96 @@ void NhgOrch::doTask(Consumer& consumer)
...
             vector<string> mpls_nhv = tokenize(mpls_nhs, ',');
+            vector<string> nhgv = tokenize(nhgs, ',');
 
             /* Create the next hop group key. */
             string nhg_str;
 
-            for (uint32_t i = 0; i < ipv.size(); i++)
+            /* Keeps track of any non-existing member of a recursive nexthop group */
+            bool non_existent_member = false;
+
```
venkatmahalingam, line 99: Do we need to throw a warning if ipv.size() is non-zero when is_recursive is set?

orchagent/nhgorch.cpp
utpalkantpintoo: This is in-line with similar error handling elsewhere in the code. We do not have a way to gracefully handle once we end up here for a non-recursive singleton nhg. There are necessary checks in caller fn.

orchagent/nhgorch.cpp
```diff
@@ -73,24 +76,96 @@ void NhgOrch::doTask(Consumer& consumer)
...
-                    nhg_str += mpls_nhv[i] + LABELSTACK_DELIMITER;
+                    if (i) nhg_str += NHG_DELIMITER;
+                    if (!mpls_nhv.empty() && mpls_nhv[i] != "na")
+                    {
+                        nhg_str += mpls_nhv[i] + LABELSTACK_DELIMITER;
+                    }
+                    nhg_str += ipv[i] + NH_DELIMITER + alsv[i];
                 }
-                nhg_str += ipv[i] + NH_DELIMITER + alsv[i];
             }
```
venkatmahalingam, line 170: Can we add a SWSS_LOG_DEBUG for nhg_str?

orchagent/nhgorch.cpp
utpalkantpintoo: Done

orchagent/nhgorch.cpp
utpalkantpintoo: Done

orchagent/nhgorch.cpp
utpalkantpintoo: Done

orchagent/nhgorch.cpp
utpalkantpintoo: Done

orchagent/nhgorch.cpp
venkatmahalingam: We install IP2ME for loopback IPs, right? I think, we can safely ignore the message with error log if we get a NHG with loopback interface.

orchagent/nhgorch.cpp
```diff
@@ -84,12 +84,18 @@ void NhgOrch::doTask(Consumer& consumer)
                         is_recursive = true;
                 }
             }
-
+            /* A NHG should not have both regular(ip/alias) and recursive fields */
+            if (is_recursive && (!ips.empty() || !aliases.empty()))
+            {
+                SWSS_LOG_ERROR("Nexthop group %s has both regular(ip/alias) and recursive fields", index.c_str());
```
Gokulnath-Raja, line 90: Can we move this to info log ?

utpalkantpintoo: Since the nexthop-group is not created due to this check, it's better to log this as error.

Gokulnath-Raja (approved):

venkatmahalingam (approved):

srj102 (approved):

goomadao:
Besides, there might be two issues that:
1. If a route points to a singleton NEXTHOP_GROUP and it becomes non-singleton later, the route will still refer to the SAI ID of the NEXTHOP instead of the newly created NEXTHOP_GROUP
2. For recursive situations, how to update NEXTHOP_GROUP recursively? For example, `NEXTHOP_GROUP_TABLE:ID1` points to `ID2` & `ID3`, `ID2` refers to NEXTHOP `NHa` & `NHb`. When `ID2` is updated to refer to `NHa`, `NHb` & `NHc`, how can we add member `NHc` to `NEXTHOP_GROUP_TABLE:ID1` simultaneously?

utpalkantpintoo:
Hi @goomadao

Thanks for the comments. Please find below my response:

1. For any nexthop-group update which changes its SAI-ID, it is expected that the routing app update the associated routes as well. This is expected behavior with FRR.
Handling it in Orchagent (NhgOrch/RouteOrch) will require maintaining a nexthop-group to route entries data-structure.

2. With recursive nexthop-group scheme, it is expected that the routing app resolve the recursive nexthop-groups and provide a recursive nexthop-group with ultimate nexthop member(s) which are not recursive themselves. This is expected behavior with FRR.

goomadao:
@utpalkantpintoo Thanks for your explanation. You're right, I forgot that there is an upper application where the update could be triggered.

Could you please take a look at the pending review I raised before? I still can't figure out any situation to get into those two conditions.

PR #3106, opened 2024-04-08
add vrrpmgrd to support FRR-VRRP configuration

micas-net:
hld:https://github.com/sonic-net/SONiC/pull/1446
depends-on:https://github.com/sonic-net/sonic-swss-common/pull/813


philo-micas:
@vvbrcm @madhupalu please help review

cfgmgr/Makefile.am
```diff
@@ -5,7 +5,7 @@ LIBNL_LIBS = -lnl-genl-3 -lnl-route-3 -lnl-3
 SAIMETA_LIBS = -lsaimeta -lsaimetadata -lzmq
 COMMON_LIBS = -lswsscommon -lpthread
 
-bin_PROGRAMS = vlanmgrd teammgrd portmgrd intfmgrd buffermgrd vrfmgrd nbrmgrd vxlanmgrd sflowmgrd natmgrd coppmgrd tunnelmgrd macsecmgrd fabricmgrd
+bin_PROGRAMS = vlanmgrd teammgrd portmgrd intfmgrd buffermgrd vrfmgrd nbrmgrd vxlanmgrd sflowmgrd natmgrd coppmgrd tunnelmgrd macsecmgrd fabricmgrd macvlanmgrd
```
venkatmahalingam, line 8: I would prefer vrrpmgrd, looks like we used macvlanmgrd in the HLD, please make the changes if possible.

philo-micas: https://github.com/sonic-net/SONiC/pull/1446#issuecomment-2131108752

As suggested by @nser77, renaming vrrpmgrd to macvlanmgrd would enhance its generality within the whole SONiC architecture, making it more convenient for future functional components to use the macvlan interface type.

venkatmahalingam: What future functional components you are referring to here? the whole code is written for VRRP, wondering how you would use this macvlanmgr for other features.

venkatmahalingam: I'm not convinced with your answer, I don't see any mention of future functional components you are referring to here. You are trying to make the code changes with the assumption that some feature will make use of this macvlanmgrd.
If you really want to have a generic macvlanmgrd, remove VRRP name usage in the files and use it as a generic library so that any future module can use it, today VRRP feature can make use of generic MACVLAN mgr.

philo-micas: Yes，you are right. We will use vrrpmgrd and have also submitted the PR for HLD https://github.com/sonic-net/SONiC/pull/1834. Could you please review it. Thanks.

cfgmgr/vrrpmgr.cpp
```diff
@@ -0,0 +1,506 @@
...
+    return true;
+}
+
+bool VrrpMgr::addVirtualInterface(const std::string &intf_alias, const std::string &vrrp_name, const MacAddress &vrrp_mac, bool is_ipv4)
+{
+    stringstream cmd;
+    string res;
+
+    // link add vrrp type macvlan
+    cmd << IP_CMD << " link add " << shellquote(vrrp_name) << " link " << shellquote(intf_alias) << " type macvlan mode bridge && ";
```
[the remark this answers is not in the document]
philo-micas: This is a great suggestion. For now, we haven't encountered any issues during our normal validation process. Other components, such as intfmgr and vlanmgr, currently also do not check whether the module is loaded before execution. We can start a new task to implement this code reinforcement for all components in the future.

cfgmgr/vrrpmgr.cpp
```diff
@@ -0,0 +1,506 @@
...
+                return false;
+            }
+        }
+    }
+    else
+    {
+        // del vrrp intf
+        if (vrrp.isValid())
+        {
+            if (!delVirtualInterface(intf_alias, vrrp.getVrrpName()))
```
nmoray, line 112: IMO, MAC Vlan interface removal needs to be avoided when the VIP list is empty as tit is getting removed as soon as the assigned VIPs are removed from the VRRP interface.

prsunny:
@venkatmahalingam , can you review and signoff?

prsunny:
@dgsudharsan for viz

cfgmgr/vrrpmgr.cpp
philo-micas: This part mainly corresponds to the 'add Virtual Interface' code mentioned earlier. If vips is not empty, we add vrrp; if vips is empty, we remove vrrp. Additionally, from the perspective of the VRRP protocol, it cannot handle forwarding without a VIP, so it seems unnecessary to retain the VRRP interface in such cases.

nmoray: Okay. then in that case, we should not allow the user to remove the VIP assigned on a VRRP interface as it the VRRP MASTER stays in Initialise state if we remove the VIP from VRRP interface and resign / update as the underlying MAC Vlan interface has already been removed.

nmoray: it works if we simply remove this condition and do not remove the interface.

philo-micas: When all VIPs are removed from a VRRP interface, it cannot respond to VRRP protocol packets and enters the Initialise state(similarly, when a new VRRP interface without any VIP is added, it also starts in the Initialise state). According to RFC 3768 Section 6.4, only the Master state allows data packet forwarding. If the VRRP interface is not removed from the chip, data packets may still match the DMAC of the VRRP interface in the Initialise state, resulting in unintended forwarding.
To ensure consistency between the kernel and hardware layers, we remove both the kernel VLAN MAC interface and the VRRP interface from the chip simultaneously.

It’s worth noting that this does not impact configuration management. Since VRRP configurations and display states are stored in the FRR container, commands like show vrrp will continue to display accurate information even after the kernel VLAN MAC is removed.

nmoray: Okay. So in the case where user wants to change the VIP, it is expected from the user to first delete VRRP interface via the CLI and later re-add that with new IP? If so, better to restrict the user (handle it in the configuration part) in assigning a new VIP if the underlying MAC Vlan interface is absent. 

When we remove the MAC VLAN interface directly from the kernel here, SONIC doesn't know that the underlying interface has already been removed while reassigning or updating the VIP as the respective config still present in the CONFIG_DB.

philo-micas: "I understand that two issues are being raised here:
1. When changing the VIP, there may be cases where a VIP is added, but the MAC VLAN interface is absent in the chip (or kernel).
    - In the kernel, this is expected. As mentioned earlier: If vips is not empty, we add vrrp; if vips is empty, we remove vrrp.
    - In the chip, the appl_db structure ensures a one-to-one mapping between the VIP and the interface with the key format: VRRP_TABLE:interface_name:vip:type.
2. Directly deleting the MAC VLAN from the kernel may lead to inconsistencies between layers.
    - This issue exists for all logical interfaces in SONiC, such as loopback, subports, portchannels, etc. For now, I have chosen to ignore this in the VRRP case. It might require a corresponding optimization PR in the future :)

balanokia:
@micas-net thanks for adding this feature. Please let me know the release that this feature is planned for. Pls help merging to the master. thank you

lib/vrrpintf.h
```diff
@@ -0,0 +1,48 @@
...
+#define VRRP_V6_PREFIX "Vrrp6"
+
+#define VRRP_V4_MAC_PREFIX "00:00:5e:00:01:"
+#define VRRP_V6_MAC_PREFIX "00:00:5e:00:02:"
+
+const char vrrp_name_delimiter = '-';
+
+namespace swss {
+
+class VrrpIntf
```
balanokia, line 21: macvlan device nomenclature needs to be adapted as it appends only VRID to the parent interface. It creates duplicate issue to reassign the same VRID to the other interfaces. I would like to propose the following nomenclature for macvlan device,
For VRRPv3 compliant,
1. Vrrp4-<vrid>-<t><hhhh>
2. Vrrp6-<vrid>-<t><hhhh>
Fields,
- `<t>` is parent type class (`e|p|v|s`),

1. Type `e` to denote L3 physical single interface
2. Type `p` to denote L3 physical bundle/LAG interface
3. Type `v` to denote L3 logical VLAN interface
4. Type `s` to denote L3 logical VLAN subinterface

- `<hhhh>` is per-type allocated 4-hex token. It will be a free-running counter to index max of 64k for each VRID.

PR #3118, opened 2024-04-19
[ACL] Remove flex counter when updating ACL rule

bingwang-ms:
**What I did**
This PR is to fix https://github.com/sonic-net/sonic-buildimage/issues/18719 

When ACL rule is created for the first time, a flex counter is created and registered https://github.com/sonic-net/sonic-swss/blob/9e183a650cf38e0d5af46adc1342dcde85ecb84d/orchagent/aclorch.cpp#L4076

When the same ACL rule is being updated, the FlexCounter created before is not removed, and another FlexCounter is created and registered.

**Why I did it**
Fix the issue that FlexCounter is duplicated when updating existing ACL rule.

**How I verified it**
1. Verified by VS test
2. Verified by building orchagent and run on a physical device

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -2586,6 +2586,12 @@ bool AclTable::add(shared_ptr<AclRule> newRule)
     if (ruleIter != rules.end())
     {
         // If ACL rule already exists, delete it first
+        if (ruleIter->second->hasCounter())
+        {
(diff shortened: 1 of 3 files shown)


bingwang-ms:
@stepanblyschak Can you please help review?

stepanblyschak (approved):

bingwang-ms:
MSFT ADO: 27676985

PR #3126, opened 2024-04-28
[System ready] Report ready state to Sysmonitor daemon

fastiuk:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/portsyncd/portsyncd.cpp
+++ b/portsyncd/portsyncd.cpp
@@ -69,6 +69,7 @@ int main(int argc, char **argv)
         DBConnector appl_db("APPL_DB", 0);
         DBConnector state_db("STATE_DB", 0);
         ProducerStateTable p(&appl_db, APP_PORT_TABLE_NAME);
+        Table feature(&state_db, "FEATURE");
 
         Table cfgDeviceMetaDataTable(&cfgDb, CFG_DEVICE_METADATA_TABLE_NAME);
         if (!cfgDeviceMetaDataTable.hget("localhost", "switch_type", g_switchType))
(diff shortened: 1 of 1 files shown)


liat-grozovik:
@fastiuk please refer to the comment provided. also please update the PR description of what are you are trying to fix and the scope. if this is a bug to whihc branches it should go?

portsyncd/portsyncd.cpp
```diff
@@ -134,6 +135,13 @@ int main(int argc, char **argv)
                     p.set("PortInitDone", attrs);
                     SWSS_LOG_NOTICE("PortInitDone");
 
+                    /*
+                     * Notify "System ready" feature that we are ok.
+                     */
+                    FieldValueTuple app_state("up_status", "true");
```
fastiuk, line 141: > Why is portinitdone alone considered for sysready? In case of warmboot/fastboot scenario should we consider reconciliation done as sys ready?

@dgsudharsan we don't have other requirements and portinitdone is good enough. The previous design used "boot-up" of the service as "system ready".
I am not sure about "reconciliation done". Is warmboot/fastbot not set "PortInitDone"?

portsyncd/portsyncd.cpp
```diff
@@ -134,6 +135,13 @@ int main(int argc, char **argv)
...
+                    vector<FieldValueTuple> feature_attrs = { app_state };
+                    feature.set("swss", feature_attrs);
```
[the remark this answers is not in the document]
fastiuk: We don't have another requirement, so portsyncd is good enough.

fastiuk:
@prsunny any more comments?

portsyncd/portsyncd.cpp
qiluo-msft, line 141: There should be some code to set it "false"?
1. when orchagent starts
2. when orchagent exited gracefully
3. when orchagent receive a signal and exited
4. etc.

PR #3131, opened 2024-05-03
Restore preemphasis when AN is transitioned from ON to OFF

PJHsieh:
In certain switches, link training is enabled along with auto-negotiation(AN) enabled. In this case, the preemphasis settings may be changed due to the presence of enabled link training. Therefore, it's essential to restore the preemphasis value when auto-negotiation transitions from ON to OFF.

**What I did**
Restore preemphasis when AN is transitioned from ON to OFF

**Why I did it**
In certain switches, link training is enabled along with auto-negotiation(AN) enabled. In this case, the preemphasis settings may be changed due to the presence of enabled link training. Therefore, it's essential to restore the preemphasis value when auto-negotiation transitions from ON to OFF.

**How I verified it**
1. Check the current pre-emphasis value P1 by bcmcmd.
2. Enable AN on a port, make sure the port is up and the pre-emphasis is changed to P2 by link-training.
3. Disable AN on the testing port, check the pre-emphasis value is restored to P1 by bcmcmd.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -3805,6 +3805,15 @@ void PortsOrch::doPortTask(Consumer &consumer)
                             "Set port %s autoneg to %s",
                             p.m_alias.c_str(), m_portHlpr.getAutonegStr(pCfg).c_str()
                         );
+
+                        /* If no explicit pre-emphasis settings exist when AN is changed from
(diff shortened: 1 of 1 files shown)


PJHsieh:
Hi, @prsunny ,

Please review this PR, thank you.

PJHsieh:
@dgsudharsan , please merge this PR, thanks.

PR #3133, opened 2024-05-06
Refactor Host Tx Ready setting

noaOrMlnx:
**What I did**
Update the call to setHostTxReady() to receive Port instead of port_id.

**Why I did it**
So we won't need to call getPort() inside the function, and we will be able to call setHostTxReady from within port initialization. 

**How I verified it**
Tested it on running switch.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1469,7 +1469,7 @@ void PortsOrch::initHostTxReadyState(Port &port)
 
     if (hostTxReady.empty())
     {
-        setHostTxReady(port.m_port_id, "false");
+        setHostTxReady(port, "false");
         SWSS_LOG_NOTICE("initialize host_tx_ready as false for port %s",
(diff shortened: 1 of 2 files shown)


prgeor (approved):

prsunny:
Can you please check coverage?

noaOrMlnx:
@prsunny 
As discussed in teams, Prince G approved to merge this PR without UT added, because it complains of missing coverage of one function call, which probably cannot be mocked (general SAI API which we cannot hook to any sai group as switch/port)
can you please check if this can be merged?

prsunny:
Skipping coverage as this is a code refactor

noaOrMlnx:
@yxieca can you please cherry-pick to 202311?

noaOrMlnx:
cherry-pick to 202311 PR: https://github.com/sonic-net/sonic-swss/pull/3157

PR #3135, opened 2024-05-07
Capability query for MACSEC ACL attribute

smvmarvell:
**What I did**
-Modified the MACsec orchestration logic to conditionally include or exclude the ability to match the SCI in ACL configurations based on the ASIC's capabilities.
-Implemented a capability check in SONiC to determine whether the SAI_ACL_TABLE_ATTR_FIELD_MACSEC_SCI attribute is supported by the ASIC, thus ensuring that neither SAI_ACL_TABLE_ATTR_FIELD_MACSEC_SCI nor SAI_ACL_ENTRY_ATTR_FIELD_MACSEC_SCI is used when unsupported.

**Why I did it**
- The current implementation attempts to use the `SAI_ACL_ENTRY_ATTR_FIELD_MACSEC_SCI` attribute even when it's not supported by the underlying ASIC or driver, causing failures in the vendor's SAI/SDK code.
- The capability check prevents these errors, ensuring compatibility with different hardware.

**How I verified it**
- Compiled the code successfully without errors.
- Conducted manual tests on Marvell platforms without support for the `SAI_ACL_TABLE_ATTR_FIELD_MACSEC_SCI` attribute to validate the conditional logic.

**Details if related**
Fixes the issue raised in https://github.com/sonic-net/sonic-swss/issues/3134

### DIFF
--- a/orchagent/macsecorch.cpp
+++ b/orchagent/macsecorch.cpp
@@ -38,6 +38,7 @@ extern sai_switch_api_t *sai_switch_api;
 constexpr bool DEFAULT_ENABLE_ENCRYPT = true;
 constexpr bool DEFAULT_SCI_IN_SECTAG = false;
 constexpr sai_macsec_cipher_suite_t DEFAULT_CIPHER_SUITE = SAI_MACSEC_CIPHER_SUITE_GCM_AES_128;
+bool saiAclFieldSciMatchSupported = true;
 
 static const std::vector<std::string> macsec_sa_attrs =
     {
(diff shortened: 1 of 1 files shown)


gurugmarvell:
This change is already merged as part of https://github.com/sonic-net/sonic-swss/pull/3385.

gurugmarvell:
@prsunny , Could you please close this PR , this change is merged as part of https://github.com/sonic-net/sonic-swss/pull/3385.

PR #3143, opened 2024-05-10
Vlanmgr and Fdborch changes for PAC

anilkpan:
Added Vlanmgr and Fdborch changes to support  PAC.
The following PAC operations are supported to be passed down from Vlanmgr to OA for updating the ASIC DB:

1. Port learning mode
2. FDB entry
3. VLAN membership

PAC HLD: https://github.com/sonic-net/SONiC/blob/master/doc/pac/Port%20Access%20Control.md

### DIFF
--- a/cfgmgr/vlanmgr.cpp
+++ b/cfgmgr/vlanmgr.cpp
@@ -21,8 +21,9 @@ using namespace swss;
 
 extern MacAddress gMacAddress;
 
-VlanMgr::VlanMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, const vector<string> &tableNames) :
-        Orch(cfgDb, tableNames),
+VlanMgr::VlanMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, const vector<string> &tableNames,
+        const vector<string> &stateTableNames) :
+        Orch(cfgDb, stateDb, tableNames, stateTableNames),
         m_cfgVlanTable(cfgDb, CFG_VLAN_TABLE_NAME),
         m_cfgVlanMemberTable(cfgDb, CFG_VLAN_MEMBER_TABLE_NAME),
         m_statePortTable(stateDb, STATE_PORT_TABLE_NAME),
@@ -31,6 +32,8 @@ VlanMgr::VlanMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
         m_stateVlanMemberTable(stateDb, STATE_VLAN_MEMBER_TABLE_NAME),
         m_appVlanTableProducer(appDb, APP_VLAN_TABLE_NAME),
         m_appVlanMemberTableProducer(appDb, APP_VLAN_MEMBER_TABLE_NAME),
+        m_appFdbTableProducer(appDb, APP_FDB_TABLE_NAME),
+        m_appPortTableProducer(appDb, APP_PORT_TABLE_NAME),
         replayDone(false)
 {
     SWSS_LOG_ENTER();
@@ -642,6 +645,7 @@ void VlanMgr::doVlanMemberTask(Consumer &consumer)
                 m_stateVlanMemberTable.set(kfvKey(t), fvVector);
 
                 m_vlanMemberReplay.erase(kfvKey(t));
+                m_PortVlanMember[port_alias][vlan_alias] = tagging_mode;
             }
         }
         else if (op == DEL_COMMAND)
@@ -654,6 +658,7 @@ void VlanMgr::doVlanMemberTask(Consumer &consumer)
                 key += port_alias;
                 m_appVlanMemberTableProducer.del(key);
                 m_stateVlanMemberTable.del(kfvKey(t));
+                m_PortVlanMember[port_alias].erase(vlan_alias);
             }
             else
             {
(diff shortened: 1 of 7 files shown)


anilkpan:
The following PR needs to be merged to resolve the compilation error:
https://github.com/sonic-net/sonic-swss-common/pull/871

orchagent/fdborch.cpp
```diff
@@ -1410,7 +1416,9 @@ bool FdbOrch::addFdbEntry(const FdbEntry& entry, const string& port_name,
             attrs.push_back(attr);
         }
     }
-
+    attr.id = SAI_FDB_ENTRY_ATTR_PACKET_ACTION;
```
sutharsansr, line 1419: Should there be changes in fdborch to push the packet to PACD socket? https://github.com/sonic-net/SONiC/blob/master/doc/pac/Port%20Access%20Control.md#313-mab-packet-receive-flow. How are the packets delivered for authentication when MAB is enabled on the interface please?

vijaya-ops: These packets that come to CPU are received by PACD as it registers socket for the same. If  MAB is enabled, then PACD informs , the information that is required for authentication,  such as mac address and the interface to MAB process. The communication between the PACD and MABD uses simple client-server socket communication.

adyeung:
@sutharsansr pls review and sign off if you have no further comments

sutharsansr (approved):

cfgmgr/vlanmgr.cpp
```diff
@@ -669,6 +674,307 @@ void VlanMgr::doVlanMemberTask(Consumer &consumer)
...
+                    learn_mode = fvValue(i);
+                }
+            }
+            if (!learn_mode.empty())
+            {
+                SWSS_LOG_NOTICE("set port learn mode port %s learn_mode %s\n", alias.c_str(), learn_mode.c_str());
+                vector<FieldValueTuple> fvVector;
+                FieldValueTuple portLearnMode("learn_mode", learn_mode);
+                fvVector.push_back(portLearnMode);
+                m_appPortTableProducer.set(alias, fvVector);
```
praveenraja1, line 707: What about the FDB learnt on the port, prior to PAC being enabled on it? Do we ensure those get flushed when PAC is enabled and port is set to s/w learning mode

cfgmgr/vlanmgr.cpp
```diff
@@ -669,6 +674,307 @@ void VlanMgr::doVlanMemberTask(Consumer &consumer)
...
+    int vlan_id = stoi(vlan_alias.substr(4));
+    if (addHostVlanMember(vlan_id, membername, tagging_mode))
+    {
+        std::string key = VLAN_PREFIX + to_string(vlan_id);
+        key += DEFAULT_KEY_SEPARATOR;
+        key += membername;
+        vector<FieldValueTuple> fvVector;
+        FieldValueTuple s("tagging_mode", tagging_mode);
+        fvVector.push_back(s);
+        FieldValueTuple s1("dynamic", "no");
```
praveenraja1, line 940: what does the "dynamic" field indicate here?

praveenraja1 (approved):

cfgmgr/vlanmgr.cpp
anilkpan: It should be done in the orchagent code while processing learn mode request, I will add that as well.

cfgmgr/vlanmgr.cpp
anilkpan: "dynamic" means that the vlan member was not added by config, but was added by an application. This field can be used to indicate non-configured vlan from show commands.

cfgmgr/vlanmgr.cpp
anilkpan: @praveenraja1 , the change related to flushing FDB on learning mode set is a common fix needed in OA, so its not really a part of PAC implementation. I will create a separate PR for that.

orchagent/orch.cpp
```diff
@@ -30,6 +30,20 @@ Orch::Orch(DBConnector *db, const vector<string> &tableNames)
     }
 }
 
+Orch::Orch(swss::DBConnector *db1, swss::DBConnector *db2,
```
prsunny, line 33: What is this change? Seems not related to feature. Please remove

prsunny (changes requested):
Please add tests and fix code coverage

orchagent/orch.cpp
anilkpan: This change is needed, as Vlanmgr needs to process updates from 2 db's now, config db and state db.

prsunny: Please explain which tables. and where is it added in Vlanmgr code?

anilkpan: Currently, Vlanmgr processes updates only from the config db. PAC sends updates to Vlanmgr through the following tables in the state db:
#define STATE_OPER_VLAN_TABLE_NAME "OPER_VLAN"
#define STATE_OPER_VLAN_MEMBER_TABLE_NAME "OPER_VLAN_MEMBER"
#define STATE_OPER_FDB_TABLE_NAME "OPER_FDB"
#define STATE_OPER_PORT_TABLE_NAME "OPER_PORT"

To process these updates, Vlanmgr needs to add the state db tables as well to the Orch list:

VlanMgr::VlanMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, const vector<string> &tableNames,
        const vector<string> &stateTableNames) :
        Orch(cfgDb, stateDb, tableNames, stateTableNames),

So, in addition to cfgDb, we also add stateDb in the Orch list. The PAC related tables are in stateTableNames

adyeung:
@prsunny please help review and signoff, this is planned for 202411

anilkpan:
vstest failing after rebasing with latest upstream/master

anilkpan:
@prsunny, can you please check and approve the PR for merge for 202411?

orchagent/orch.cpp
prsunny: Why is this not coming via APP_DB? I don't think we have a current flow where orchagent listens to STATE_DB.

anilkpan: Orchagent does not listen to STATE_DB. As per the PAC design in the HLD, Vlanmgr listens to STATE_DB for updates from PAC. Vlanmgr uses Orch infra for subscription.

adyeung: HLD 3.1.2 has the call flow, vlanmgr will subscribe and publish to state_db, it will also update to appl_db. No change to existing OA pub/sub model with PAC, OA will subscribe to appl_db for PAC update and publish to asic_db

prsunny (approved):
Thanks for clarifying @anilkpan

PR #3148, opened 2024-05-13
[muxorch] Using bulker to program routes/neighbors during switchover

Ndancejic:
Uses entity bulker to program routes and neighbors during mux
switchover. Mux switchover performance suffers when switching over with
a large number of neighbors on the mux port. This uses the optimization
of programming the neighbors and routes in bulk to avoid sequentially
programming each.

**What I did**
Changed mux switchover logic to use neighbor and bulk switchover instead of programming neighbors sequentially.

**Why I did it**
Testing shows this improves switchover time by an average of 30% at 128 neighbors.

**How I verified it**
added a test to vstests to test functionality of code, and tested performance on a dualtor lab testbed.

**Details if related**
ado:#25072867

[bulk_switchover_test_results.xlsx](https://github.com/sonic-net/sonic-swss/files/15300898/bulk_switchover_test_results.xlsx)


orchagent/neighorch.cpp
```diff
@@ -1199,6 +1201,434 @@ bool NeighOrch::removeNeighbor(const NeighborEntry &neighborEntry, bool disable)
     return true;
 }
 
+/**
+ * @brief Creates a neighbor add entry and adds it to bulker.
+ * @param ctx NeighborBulkContext contains neighbor information and list of object statuses.
+ */
+bool NeighOrch::addBulkNeighbor(NeighborBulkContext& ctx)
```
prsunny, line 1208: does it use bulk neighbor only for mux scenario or even for the existing path? I think we should enable it only for mux flow.

Ndancejic: it's only for mux flow for this PR

prsunny:
@dgsudharsan for viz

prsunny:
Dependant on PR https://github.com/sonic-net/sonic-sairedis/pull/1373

Ndancejic:
still pulling from wrong sairedis pipeline, running again:
`Download from the specified build: #558894`

Ndancejic:
Last run was still failing because the correct syncd was not installed for some reason. Ran mux tests locally with syncd package pulled from https://dev.azure.com/mssonic/build/_build/results?buildId=559999&view=results and only had one test_multi_nexthop test fail due to a small bug I was able to resolve. Tested with latest commit and all tests are passing.

=================================================================
conftest.py:1880
  /data/sonic-swss/tests/conftest.py:1880: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    @pytest.yield_fixture(scope="module")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=================================================== 31 passed, 1 warning in 1031.63s (0:17:11) =====

prsunny (approved):
lgtm, can you post the dualttor sonic-mgmt test results?

orchagent/muxorch.cpp
```diff
@@ -795,22 +804,48 @@ bool MuxNbrHandler::enable(bool update_rt)
...
+        {
+            gRouteBulker.clear();
+            return false;
+        }
+
+        it = neighbors_.begin();
+        while (it != neighbors_.end())
+        {
+            NextHopKey nh_key = NextHopKey(it->first, alias_);
+            if (update_rt)
```
prsunny, line 831: Why is `update_rt `checked here?  Its already done above.

orchagent/muxorch.cpp
```diff
@@ -795,22 +804,48 @@ bool MuxNbrHandler::enable(bool update_rt)
...
+        }
+        gRouteBulker.flush();
+        if (!processBulkRouteEntries(bulk_route_ctx_list))
+        {
+            gRouteBulker.clear();
+            return false;
+        }
+
+        it = neighbors_.begin();
+        while (it != neighbors_.end())
```
prsunny, line 828: We've introduced three loops now in the new code. Can we optimize it?

orchagent/muxorch.cpp
```diff
@@ -795,22 +804,48 @@ bool MuxNbrHandler::enable(bool update_rt)
...
+        {
+            NextHopKey nh_key = NextHopKey(it->first, alias_);
+            if (update_rt)
+            {
+                updateTunnelRoute(nh_key, false);
```
prsunny, line 833: Why don't we move this to original section and do this loop only if Bulk api fails. So we don't have to do it everytime

prsunny (changes requested):

orchagent/neighorch.cpp
```diff
@@ -1227,6 +1229,434 @@ bool NeighOrch::removeNeighbor(const NeighborEntry &neighborEntry, bool disable)
...
+            neighbor_attr.id = SAI_NEIGHBOR_ENTRY_ATTR_NO_HOST_ROUTE;
+            neighbor_attr.value.booldata = 1;
+            neighbor_attrs.push_back(neighbor_attr);
+        }
+    }
+
+    PortsOrch* ports_orch = gDirectory.get<PortsOrch*>();
+    auto vlan_ports = ports_orch->getAllVlans();
+
+    for (auto vlan_port: vlan_ports)
```
prsunny, line 1284: This is a lot of repeated code from existing `addNeighbor`. I think we should only add the bulk function as discussed offline.

orchagent/muxorch.cpp
Ndancejic: good catch, will fix that

orchagent/muxorch.cpp
Ndancejic: I see what you're saying, I think I just wanted to make sure the order was maintained. I'll fix that

Ndancejic: I think it's safe to take out this loop entirely then, if bulk api fails we will fall back, which will remove the tunnel routes anyways

orchagent/muxorch.cpp
Ndancejic: I took the loop out from the comment below, I believe the two left over are necessary, we need the neighbors enabled before we update routes to point to their next hops

orchagent/neighorch.cpp
Ndancejic: Reduced the amount of repeated code in latest commit

orchagent/muxorch.cpp
```diff
@@ -795,22 +805,28 @@ bool MuxNbrHandler::enable(bool update_rt)
...
+
         it++;
     }
 
+    if (update_rt && !removeRoutes(route_ctx_list))
+    {
+        return false;
+    }
+
+    gRouteBulker.clear();
```
prsunny, line 821: should we clear this even if its returned false above

orchagent/neighorch.cpp
```diff
@@ -1060,6 +1081,7 @@ bool NeighOrch::addNeighbor(const NeighborEntry &neighborEntry, const MacAddress
     }
     else if (isHwConfigured(neighborEntry))
     {
+        ctx.set_neigh_attr_count = (int)neighbor_attrs.size();
```
prsunny, line 1084: This can be removed

orchagent/muxorch.cpp
```diff
@@ -795,22 +805,28 @@ bool MuxNbrHandler::enable(bool update_rt)
...
         {
-            if (remove_route(pfx) != SAI_STATUS_SUCCESS)
-            {
-                return false;
-            }
-            updateTunnelRoute(nh_key, false);
+            route_ctx_list.push_back(MuxRouteBulkContext(pfx));
         }
 
+        updateTunnelRoute(nh_key, false);
```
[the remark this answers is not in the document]
Ndancejic: yes, I'll move that back

orchagent/muxorch.cpp
Ndancejic: Took a look at this, it's not needed since we call clear from removeRoutes

orchagent/neighorch.cpp
```diff
@@ -1227,6 +1260,185 @@ bool NeighOrch::removeNeighbor(const NeighborEntry &neighborEntry, bool disable)
...
+    IpAddress ip_address = neighborEntry.ip_address;
+
+    if (m_syncdNeighbors.find(neighborEntry) == m_syncdNeighbors.end())
+    {
+        return true;
+    }
+
+    SWSS_LOG_INFO("Checking neighbor remove entry status %s on %s.", ip_address.to_string().c_str(), m_syncdNeighbors[neighborEntry].mac.to_string().c_str());
+
+    if (isHwConfigured(neighborEntry))
```
Ndancejic, line 1391: Took a loot at this, we still need the HwConfigured check since this isn't checked for in the disableNeighbors patch

prsunny (approved):

PR #3156, opened 2024-05-21
[tests] mark suppress-fib-pending VS test as expected to fail

stepanblyschak:
**What I did**

Revert suppress FIB pending feature

**Why I did it**

Some unresolved FRR issues in current version

**How I verified it**

Build and run

**Details if related**

### DIFF
--- a/tests/test_route.py
+++ b/tests/test_route.py
@@ -1073,7 +1073,7 @@ class TestFpmSyncResponse(TestRouteBase):
         route_entry = json.loads(output)
         return bool(route_entry[route][0].get('offloaded'))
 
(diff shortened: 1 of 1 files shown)


prsunny:
I don't think we need to revert swss changes.

stepanblyschak:
@prsunny In theory yes, we could keep it. However, please note, as part of this PR I revert a VS test that will not work without FRR config. Reverting just a VS test means I remove the coverage for suppress FIB code.

prsunny:
may be we can mark the test skipped for now and create an issue to track.

stepanblyschak:
@prsunny Update PR to just mark the test as x(pected)fail. It was already marked that way to solve a cyclic dependency between sonic-buildimage and sonic-swss changes at that time and forgotten to be removed, now, I just updated the reason message.

prsunny (approved):

PR #3161, opened 2024-05-22
Parse subport value added to config_db

bobbymcgonigle:
**What I did**
https://github.com/sonic-net/sonic-buildimage/pull/18499 Added support for subport values to be automatically calculated and inserted on creation of config_db.json. This change succeeds in adding the subport value into config_db.json - but it results in a log 

`WARNING #orchagent: message repeated 18564 times: [ :- parsePortConfig: Unknown field(subport): skipping ...]`

being printed. This prevents the log being printed by handling the subport value.

**Why I did it**
To prevent the orchagent log

**How I verified it**
I manually tested on an Arista device that the log no longer appears

**Details if related**
Question to reviews - should we assert on an unhandled/unknown field ? Rather than logging this message, it is easy to miss and the value still ends up in config_db

### DIFF
--- a/orchagent/port/portcnt.h
+++ b/orchagent/port/portcnt.h
@@ -202,6 +202,11 @@ public:
         bool is_set = false;
     } description; // Port description
 
+    struct {
+        std::string value;
+        bool is_set = false;
+    } subport; // Port subport
+
     std::string key;
     std::string op;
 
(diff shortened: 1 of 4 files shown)


prsunny (approved):

prsunny:
Bypassing coverage to unblock other PRs

kcudnik:
thanks for quick fix, i was alreay writting PR XD

PR #3164, opened 2024-05-23
Test CRM tests

kcudnik:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


kcudnik:
resolved here: https://github.com/sonic-net/sonic-swss/pull/3161

PR #3167, opened 2024-05-24
[swss] add static anycast gateway support

yanjundeng:
**What I did**
Resync the original SAG support to latest master branch:
https://github.com/sonic-net/sonic-swss/pull/1974

- Add SAG support in intfmgr and intforch
- Add support for SAG mac-address related IPv6 link-local toMe route
- Add related testcases

Some improvements based on the original PR:

- only add SAG-related IPv6 linklocal route for the VRF which has SAG-enabled interface
- add SAG related FDB entry into linux bridge for the SAG-enabled vlan interface
- Recover the vlan interface's mac in linux kernel to system mac once this SAG is disabled internally, for example unbind VRF or delete VRF case
- fix some pytest issues in pipeline

**Why I did it**

- To support static anycast gateway feature

**How I verified it**

**Details if related**

### DIFF
--- a/tests/test_sag.py
+++ b/tests/test_sag.py
@@ -31,7 +31,7 @@ class TestSag(object):
     def create_vrf(self, vrf_name):
         initial_entries = set(self.asic_db.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_VIRTUAL_ROUTER"))
         self.cfg_db.create_entry(swsscommon.CFG_VRF_TABLE_NAME, vrf_name, {"empty": "empty"})
-        time.sleep(1)
+        time.sleep(2)
 
         current_entries = set(self.asic_db.get_keys("ASIC_STATE:SAI_OBJECT_TYPE_VIRTUAL_ROUTER"))
         assert len(current_entries - initial_entries) == 1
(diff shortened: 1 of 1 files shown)


orchagent/p4orch/tests/fake_routeorch.cpp
```diff
@@ -0,0 +1,232 @@
+extern "C"
```
prvattem, line 1: Please add some description about the purpose of fake_routeorch.

VladimirKuk (changes requested):
Shouldn't it be tested for unicast value only ?

orchagent/intfsorch.cpp
```diff
@@ -1129,6 +1204,121 @@ void IntfsOrch::doTask(Consumer &consumer)
...
+            {
+                sag_mac_str = value;
+            }
+
+            if (!sag_mac_str.empty())
+            {
+                MacAddress sag_mac;
+                try
+                {
+                    sag_mac = MacAddress(sag_mac_str);
```
VladimirKuk, line 1227: Shouldn't be tested for unicast value only ?

yanjundeng: Yes, we should validate the mac in CLI input.

VladimirKuk: What will happen if the configuration received from CONFIG_DB without CLI ?

pbrisset (approved):

aki-luxor:
Any plan to merge this?

PR #3168, opened 2024-05-25
[orchagent] Do not pollute logs when unknown field is found

kcudnik:
This will prevent future unittest failing since there is syslog rate limitting in place, causing message drop when large amount of messages is arriving

**What I did**
Reduce number of messages logged when unknown field is spotted in port config

**Why I did it**
To reduce number of logged messages

**How I verified it**
Locally

**Details if related**

### DIFF
--- a/orchagent/port/porthlpr.cpp
+++ b/orchagent/port/porthlpr.cpp
@@ -6,6 +6,7 @@
 #include <unordered_map>
 #include <string>
 #include <exception>
+#include <set>
 
 #include <boost/algorithm/string.hpp>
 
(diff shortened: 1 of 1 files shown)


prgeor (approved):

orchagent/port/porthlpr.cpp
```diff
@@ -1029,6 +1032,11 @@ bool PortHelper::parsePortConfig(PortConfig &port) const
         }
         else
         {
+            if (unknown.find(field) != unknown.end())
+                continue; // do not pollute logs with the same unknown field
```
nazariig, line 1036: @kcudnik this won't be working since we have two modes: 1. Aggregation mode before the port is created 2. Parameter(s) processing mode after port is created. Thus, you will have to maintain a permanent storage to filter the key-value pairs to avoid duplicated logs (relevant for the port create stage). It should be disposed once port is created/removed

nazariig (changes requested):

orchagent/port/porthlpr.cpp
kcudnik: so let's move set to class member ?

PR #3199, opened 2024-06-13
Remove buffer maximum parameter table when the port is removed

stephenxs:
**What I did**

Remove the buffer maximum parameter table entry when the port is removed.

**Why I did it**

To avoid the informational message `handleBufferMaxParam: BUFFER_MAX_PARAM: Port Ethernet190 is not configured, need retry` after the port is removed.

**How I verified it**

Unit test and manual test

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -3551,6 +3551,9 @@ void PortsOrch::deInitPort(string alias, sai_object_id_t port_id)
     /* Remove the associated port serdes attribute */
     removePortSerdesAttribute(p.m_port_id);
 
+    /* Remove the entry from buffer maximum parameter table*/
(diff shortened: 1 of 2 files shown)


stephenxs:
New UT fails without the fix as expected
```
[ RUN      ] PortsOrchTest.PortBulkCreateRemove
portsorch_ut.cpp:642: Failure
Value of: keys.empty()
  Actual: false
Expected: true
```

Junchao-Mellanox (approved):

stephenxs:
Failed due to `##[error]The agent did not connect within the alloted time of 45 minute(s).`

prsunny (approved):

dprital:
@bingwang-ms , Can you please cherry pick to 202405 ?

PR #3201, opened 2024-06-20
Support allocate buffer pool size based on percentage of available memory

stephenxs:
**What I did**

Support allocate buffer pool size based on percentage of available memory by `percentage` field in `BUFFER_POOL` table.
It is used in the dynamic buffer model only and represents the percentage of a buffer pool's size compared to the available memory size.
On Nvidia devices, if the percentage is defined, the buffer pool size is `available memory` * `percentage`.

**Why I did it**

**How I verified it**

Mock test

**Details if related**


prsunny (approved):

dprital:
@bingwang-ms , Can you please cherry pick to 202405 ?

PR #3210, opened 2024-07-01
Issue:SAI initializes break out ports as part of default vlan

prasanna-cls:
**What I did**

When port-break-out is created (new ports are created in SAI), Remove the vlan member for the [new-break-out-port, default-vlan] and default bridge port created for the new break-out ports that created by SAI.

**Why I did it**
When SAI port OID is created as part of trigger from break-out operations, SAI makes this new port part of default vlan (creates a vlan member ) as well as creates a default bridge port. per sonic Orchagent, default vlan association or default bridge port creation must be done only as a part of operator configuration, cannot be SAI default.

**How I verified it**
create port-break-out dynamically, ensure they are not
a) part of default default vlan membership
b) having any bridge port by default.

202311
**Details if related**

Issue:SAI initializes break out ports as part of default vlan
RCA:when a break out port is created in SAI, it make it as a vlan member for default vlan.
Fix: after SAI creates a port, Orchagent gets the default vlan port membership, if any of the vlan member or bridge port is part of newly created portlist, then its removed.

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1079,6 +1079,7 @@ bool PortsOrch::addPortBulk(const std::vector<PortConfig> &portList)
         removeDefaultVlanMembers();
         removeDefaultBridgePorts();
     }
+    removeNonDefaultConfig(oidList);
 
     SWSS_LOG_NOTICE("Created ports: %s", swss::join(',', oidList.begin(), oidList.end()).c_str());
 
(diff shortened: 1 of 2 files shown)


prsunny:
Please address coverage

dgsudharsan (changes requested):
This PR assumes that new port is added to default vlan and we create bridge port. But this behavior seems to be specific to the SAI library. SAI headers don't mandate adding a port to default vlan and creating a bridge port in the flow where new port is created through breakout.
Adding to a default vlan exist only in switch init flow.
This PR will break all vendor SAI today. Hence please handle this change in SAI library.

PR #2080, opened 2021-12-14
Set tunnel id to saiOidToAlias for support update tunnel status.

iris00522:
**What I did**
1.Set tunnel id to saiOidToAlias for support update tunnel status.
2. if create tunnel port successfully notify change tunnel port status from down to up 
**Why I did it**
update tunnel status 
admin@as7726-32x-2:~$ show vxlan remotevtep
+-----------+-----------+-------------------+--------------+
| SIP             | DIP           | Creation Source   | OperStatus   |
+===========+===========+===================+==============+
| 10.1.0.32    | 10.0.0.65  | EVPN                    | oper_up      |
+-----------+-----------+-------------------+--------------+
Total count : 1

**How I verified it**
1.create l2 vni environment
2.check tunnel status "show vxlan remotevtep"
3.check syslog
~~~
/var/log/syslog.1:Dec 14 10:25:43.929560 as7726-32x-2 NOTICE swss#orchagent: :- createDynamicDIPTunnel: Created P2P Tunnel remote IP 10.0.0.65
/var/log/syslog.1:Dec 14 10:25:43.929560 as7726-32x-2 NOTICE swss#orchagent: :- addTunnelUser: diprefcnt for remote 10.0.0.65 = 1
/var/log/syslog.1:Dec 14 10:25:43.930397 as7726-32x-2 INFO syncd#syncd: [none] brcm_sai_create_bridge_port:227 bridge_port_id: 0x  2a043a00000004 tunnel_id 4
/var/log/syslog.1:Dec 14 10:25:43.930895 as7726-32x-2 NOTICE swss#orchagent: :- addBridgePort: Add bridge port Port_EVPN_10.0.0.65 to default 1Q bridge
/var/log/syslog.1:Dec 14 10:25:43.931588 as7726-32x-2 INFO syncd#syncd: [none] _brcm_sai_bridge_port_valid:1297 idx 4
/var/log/syslog.1:Dec 14 10:25:43.932158 as7726-32x-2 NOTICE swss#orchagent: :- addVlanMember: Add member Port_EVPN_10.0.0.65 to VLAN Vlan1000 vid:1000 pid0
/var/log/syslog.1:Dec 14 10:25:43.932536 as7726-32x-2 NOTICE swss#orchagent: :- doTask: Get port state change notification id:2a00000000069f status:1
[...]

PR #3216, opened 2024-07-03
VxLAN fix tunnel oper_state issue

praveenraja1:
**What I did**

In portsorch layer ensure tunnel port_oper state is updated correctly after tunnel comes up. 

Additionally made changes so that 

1) non L3 packets which gets encapsulated have outer IP hdr TTL as 64.
2) For L3 packets the inner IpHdr TTL is preserved in outer header. 

**Why I did it**

After tunnel was up and traffic was flowing, tunnel oper_state was displayed as down.

```
root@sonic:~#  show vxlan remotevtep
+---------+---------+-------------------+--------------+
| SIP     | DIP     | Creation Source   | OperStatus   |
+=========+=========+===================+==============+
| 1.1.1.1 | 2.2.2.2 | EVPN              | oper_down    |
+---------+---------+-------------------+--------------+
Total count : 1
```

**How I verified it**

```
root@sonic:~# show vxlan tunnel
vxlan tunnel name    source ip    destination ip    tunnel map name    tunnel map mapping(vni -> vlan)
-------------------  -----------  ----------------  -----------------  ---------------------------------
vtep                 2.2.2.2                        map_2000_Vlan20    2000 -> Vlan20
                                                    map_3000_Vlan30    3000 -> Vlan30
root@sonic:~#
root@sonic:~#
root@sonic:~#  show vxlan remotevtep
+---------+---------+-------------------+--------------+
| SIP     | DIP     | Creation Source   | OperStatus   |
+=========+=========+===================+==============+
| 2.2.2.2 | 1.1.1.1 | EVPN              | oper_up      |
+---------+---------+-------------------+--------------+
Total count : 1
```

**Details if related**

### DIFF
--- a/orchagent/vxlanorch.cpp
+++ b/orchagent/vxlanorch.cpp
@@ -358,6 +358,15 @@ create_tunnel(
         attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_VAL;
         attr.value.u8 = encap_ttl;
         tunnel_attrs.push_back(attr);
+    } else {
+        attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_MODE;
+        attr.value.s32 = SAI_TUNNEL_TTL_MODE_PIPE_MODEL;
+        tunnel_attrs.push_back(attr);
+
+        attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_VAL;
+        attr.value.u8 = DEFAULT_TUNNEL_ENCAP_TTL;
+        tunnel_attrs.push_back(attr);
+
     }
 
     sai_object_id_t tunnel_id;
--- a/.azure-pipelines/build_and_install_module.sh
+++ b/.azure-pipelines/build_and_install_module.sh
@@ -26,15 +26,26 @@ function build_and_install_kmodule()
     SUBLEVEL=$(echo $KERNEL_MAINVERSION | cut -d. -f3)
 
     # Install the required debian packages to build the kernel modules
+    apt-get update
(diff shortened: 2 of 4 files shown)


orchagent/vxlanorch.cpp
VladimirKuk, line 1: I think it would better to use DEFAULT_TUNNEL_ENCAP_TTL as default value for the encap_ttl (instead of 0).  It will reuse the existing code and you could also remove the "if (encap_ttl != 0)".

bradh352:
Looks like this supersedes #2080 as it includes changes from that, and then also fixes https://github.com/sonic-net/sonic-buildimage/issues/10050

I'd think this *really* needs to get merged for the next release.  VXLAN just doesn't work (at least on my Dell S5248F switches) without the TTL change, and the oper_status fix is very useful for debugging.

bradh352:
@prsunny I see you merge a lot of PRs, I'd really like to see this in the next SONiC stable release branch... it would mean I wouldn't have to continue maintaining my own automated build system and patches. Thanks!

bradh352:
I'm replacing this PR with #3383

VladimirKuk:
> @VladimirKuk what do we need to do to get this merged? I see you had asked if `updateDbPortOperStatus(port, status)` could be moved outside the `if`, implying just always call it. I can test to see if that works today if you want.
> 
> I can say that this patch is critical to VXLAN support for me, as soon as I apply it, I can ping across the VXLAN tunnels to addresses learned via ARP by the VTEPs. Without it, just nothing...

Makes sense, since updateDbPortOperStatus wasn't called for tunnel case. Looks good to me.

bradh352:
@VladimirKuk  can you look at my modifications in #3383 to verify it looks right and comment there?

PR #3217, opened 2024-07-03
Validate interface name length in cfgmgr

arfeigin:
This PR is accompanied by https://github.com/sonic-net/sonic-swss-common/pull/889 and https://github.com/sonic-net/sonic-utilities/pull/3397
**What I did**
Validate interface name length does not exceed the limitation of IFNAMSIZ to align to kernel restrictions.
This PR covers the cases of adding interface directly to the DB not using the cli - similar validation on the CLI side is implemented in a sonic-utilities PR.

**Why I did it**
Kernel restricts interface name length to not exceed IFNAMSIZ and this was not checked which can lead to inconsistency when creating such interface as it will be created in config db but it will be blocked when trying to actually create such interface.

**How I verified it**
Manual testing.

**Details if related**

### DIFF
--- a/cfgmgr/intfmgr.cpp
+++ b/cfgmgr/intfmgr.cpp
@@ -5,6 +5,7 @@
 #include "tokenize.h"
 #include "ipprefix.h"
 #include "intfmgr.h"
+#include "interface.h"
 #include "exec.h"
 #include "shellcmd.h"
 #include "macaddress.h"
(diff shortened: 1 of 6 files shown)


arfeigin:
Checkers fail as https://github.com/sonic-net/sonic-swss-common/pull/889 should be merged first so it will compile.

nazariig (approved):

dprital:
@prsunny , Can you please approve this PR ?

PR #3220, opened 2024-07-04
[mirrororch] Fixed the issue that mirror references port without incr…

inspurSDN:
…easing the RefCount

Why I did it
Mirror references port without increasing the RefCount.

What I did
When the mirror refers to port, increase the number of port RefCount.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


prsunny:
@inspurSDN , you would need to add UT for the code fixes

PR #3222, opened 2024-07-05
Remove the API setBridgePortLearningFDB() and use setBridgePortLearnMode() only

DennisChiuEC:
**What I did**
1.Remove the API setBridgePortLearningFDB()
2.Use setBridgePortLearnMode() as instead

**Why I did it**
Two API setBridgePortLearningFDB() and setBridgePortLearnMode() do the same thing which configured port learning mode.
The code can be refined to have only one API.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -902,7 +902,13 @@ void OrchDaemon::start()
                         for (auto& pair: gPortsOrch->getAllPorts())
                         {
                             auto& port = pair.second;
-                            gPortsOrch->setBridgePortLearningFDB(port, SAI_BRIDGE_PORT_FDB_LEARNING_MODE_DISABLE);
+                            // TODO: how to support 1D bridge?
+                            if (port.m_type != Port::PHY && port.m_type != Port::LAG)
+                            {
+                                continue;
+                            }
+
+                            gPortsOrch->setBridgePortLearnMode(port, SAI_BRIDGE_PORT_FDB_LEARNING_MODE_DISABLE);
                         }
                     }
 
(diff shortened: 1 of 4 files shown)


prsunny:
@DennisChiuEC , understand the motivation to use one api. What is the other change during warmboot? Is it fixing any issue? @dgsudharsan , @vaibhavhd for viz

prsunny:
> The other change is aimed to make the L2 portchannel can forward packet during warmboot. Maybe I should separate the change into another PR? And this PR is focus on the change to use one api to set port learn mode.

that would be great

DennisChiuEC:
> > The other change is aimed to make the L2 portchannel can forward packet during warmboot. Maybe I should separate the change into another PR? And this PR is focus on the change to use one api to set port learn mode.
> 
> that would be great

Done, has removed the change which is not related to refine the API.

orchagent/orchdaemon.cpp
```diff
@@ -904,7 +904,13 @@ void OrchDaemon::start()
                         for (auto& pair: gPortsOrch->getAllPorts())
                         {
                             auto& port = pair.second;
-                            gPortsOrch->setBridgePortLearningFDB(port, SAI_BRIDGE_PORT_FDB_LEARNING_MODE_DISABLE);
+                            // TODO: how to support 1D bridge?
```
nazariig, line 907: @DennisChiuEC is this the single place where `setBridgePortLearningFDB` is being used? If not, that might cause some issues, since behavior has been changed

nazariig (changes requested):

orchagent/orchdaemon.cpp
DennisChiuEC: Yes, this is the single place where setBridgePortLearningFDB() is called.

PR #3223, opened 2024-07-05
[orchagent]: Skip installing ACL counter when ACL mirror rule is inactive

fountzou:
* Don't install counters when session rule is inactive.

Mirror session is not activated and thus, the ACL rule is not created in sairedis/SAI (there is no next hop for the mirror destination IP). However, orchagent creates and registers the ACL counter with the flexcounter (FC) and the counter is being polled every polling interval. Since the ACL counter is not attached to any ACL rule, the BCM SAI introduces print errors in syslog.

**What I did**
Removed the ACL counter when ACL mirror rule / session is inactive.

**Why I did it**
To solve [#18844](https://github.com/sonic-net/sonic-buildimage/issues/18844) issue.

**How I verified it**
- Running everflow and acl testcases.
- Checking syslog for the offending error logs.

**Details if related**


orchagent/aclorch.cpp
```diff
@@ -2055,6 +2055,11 @@ bool AclRuleMirror::activate()
 
     if (!state)
     {
+        // if the mirror session is inactive, do not create the counter yet. FlexCounter registration will be skipped as well.
```
DanielaMurin, line 2058: In this situation, where the mirror session is inactive, was the ACL rule created and then removed?

DanielaMurin: If the ACL rule was never created, is there a way to not create the counter at the beginning?
(Instead of it being created and then removed)

fountzou: Looking into it. Thanks for the suggestion.

deepak-singhal0408:
@bingwang-ms can you please help review this?

orchagent/aclorch.cpp
fountzou: PR has been updated based on your comment. Thank you!

orchagent/aclorch.cpp
```diff
@@ -2023,6 +2023,18 @@ bool AclRuleMirror::validate()
...
+bool AclRuleMirror::createCounter()
+{
+    SWSS_LOG_ENTER();
+
+    // If the mirror session is inactive, do not create the ACL counter from the first place
+    bool state = false;
+
+    m_pMirrorOrch->getSessionStatus(m_sessionName, state);
+
+    return !state ? true : AclRule::createCounter();
```
bingwang-ms, line 2035: This logic is a little confusing. Can you please write it in a better way?

bingwang-ms: If the `createCounter` function is overridden, do you also override the `removecounter` function for AclRuleMirror?

orchagent/aclorch.cpp
DanielaMurin: Wonderful! :)

orchagent/aclorch.cpp
fountzou: Thank you for the review. Code refactoring has been done (kindly see new commit).
`AclRuleMirror::createCounter` leverages its functionality to` AclRule::createCounter` iff the mirror session is active; otherwise, it performs NOP.
I haven't override `removeCounter` since it would behave exactly the same as `AclRule::removeCounter`.

mlok-nokia:
@prsunny @bingwang-ms @rlhui  Please review this PR.

bingwang-ms:
LGTM now. Thanks for the fix.

arlakshm:
@prsunny, please help merge this PR.

prsunny:
Restarting the VS tests

prsunny:
@fountzou , could you check the test failure?

fountzou:
Hi @prsunny. Title has been fixed -- thanks.

Looking into this but not sure yet if it is related to my code changes; Noticed in the pipelines that there are several PRs that fail in the same testcase.
https://dev.azure.com/mssonic/build/_build?definitionId=15&_a=summary
master: https://dev.azure.com/mssonic/build/_build?definitionId=15&_a=summary&repositoryFilter=154&branchFilter=11183%2C11183

mlok-nokia:
@fountzou Do we need this PR for 202405 branch?

fountzou:
> @fountzou Do we need this PR for 202405 branch?

@mlok-nokia Yes, we do -- thanks for bringing it up.  Appropriate label should be added.

mlok-nokia:
@prsunny Hi Prince, This PR is required by branch 202405.  Please help to merge to it.  Thanks.

mlok-nokia:
@prsunny  Please cherry-pick this PR 202405. Thanks

fountzou:
@prsunny Could you please add `chassis` label on this one so that gets the visibility and eventually cherry-picked? -- thanks

PR #3232, opened 2024-07-11
Add a check if OID exists before setting Host Tx Ready Signal Enable 

noaOrMlnx:
**What I did**
I added a query before setting host tx ready - must check that Oid exists before setting attribute to SAI.

**Why I did it**

When we do dynamic port breakout to a Fw-control port, portsorch receives a notification that something was deleted in TRANSCEIVER_INFO table from State DB, and tries to set HOST_TX_READY_SIGNAL_ENABLE attribute to false for the deleted OID. 
The issue is the Oid is already deleted and we receive an error from SAI. 

**How I verified it**
Run DPB on the switch
**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -958,7 +958,7 @@ bool PortsOrch::addPortBulk(const std::vector<PortConfig> &portList)
             attr.value.booldata = false;
             attrList.push_back(attr);
         }
-        
+
         if (cit.pt_intf_id.is_set)
         {
             if (!m_isPathTracingSupported)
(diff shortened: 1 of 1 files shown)


noaOrMlnx:
@prgeor 
can you please review this host_tx_ready fix?

Junchao-Mellanox (approved):

prgeor:
@noaOrMlnx Why OA is acting upon deleted ports?

noaOrMlnx:
@prgeor OA is acting upon port deletion, if a module was plugged out, and its entry was deleted from TRANSCEIVER_INFO table is State DB, so it could set HostTxReady enable to false.
In DPB case, the port OID is deleted, and we can't set any attribute to SAI.

orchagent/portsorch.cpp
```diff
@@ -5128,14 +5128,19 @@ bool PortsOrch::setSaiHostTxSignal(const Port &port, bool enable)
...
+        sai_status_t status = sai_port_api->set_port_attribute(port.m_port_id, &attr);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("Could not set SAI_PORT_ATTR_HOST_TX_SIGNAL_ENABLE to port 0x%" PRIx64, port.m_port_id);
+            return false;
+        }
+        return true;
     }
 
+    SWSS_LOG_NOTICE("Could not set SAI_PORT_ATTR_HOST_TX_SIGNAL_ENABLE - OID does not exist 0x%" PRIx64, port.m_port_id);
```
prgeor, line 5143: @noaOrMlnx can we print the port name in the notice message?

prgeor (approved):

bingwang-ms:
Please fix the conflict to 202311 branch.

yxieca:
@noaOrMlnx can you help creating separate PR for 202311 branch? Thanks!

noaOrMlnx:
@yxieca 
Hi, PR for 202311 was already created and merged - https://github.com/sonic-net/sonic-swss/pull/3233
Thanks :)

PR #3237, opened 2024-07-18
Negative counters

shiraez:
**What I did**
Fix for:
https://github.com/sonic-net/sonic-buildimage/issues/19594
https://github.com/sonic-net/sonic-swss/issues/1892

After changing the speed the counter is reset, 
We need to reset the <COUNTER_NAME>_last

**Why I did it**
Check if the counter creates an illegal number and then resets it

**How I verified it**
```
sudo config interface autoneg Ethernet0 disabled
sudo config interface speed Ethernet0 10
portstat 
sudo config interface speed Ethernet0 100
```


shiraez:
@Stas0k

shiraez:
@prsunny @prgeor , Can you please review?

PR #3241, opened 2024-07-22
[fpmsyncd] support pipeline to flush with a timer

a114j0y:
**What I did**
- change the pipeline size for `RouteSync`
- modify the fpmsyncd flush behavior so that it could delay the flush with an expiring timer
    - the timer is simulated by setting the timeout for `select` function

**Why I did it**
- to batch the data that fpmsyncd flush to `APP_DB`

**How I verified it**
measure the performance with [PerformanceTimer](https://github.com/sonic-net/sonic-swss-common/pull/893)  GitHub issue/pull request detail

### DIFF
--- a/fpmsyncd/fpmsyncd.cpp
+++ b/fpmsyncd/fpmsyncd.cpp
@@ -338,7 +338,10 @@ void flushPipeline(RedisPipeline& pipeline) {
 
     int idle = pipeline.getIdleTime();
 
-    // flush right away
+    // flush the pipeline if
+    // 1. traffic is not scaled (only prevent fpmsyncd from flushing ppl too frequently in the scaled case)
+    // 2. the idle time since last flush has exceeded gFlushTimeout
+    // 3. idle <= 0, due to system clock drift, should not happen since we already use steady_clock for timing
     if (remaining < SMALL_TRAFFIC || idle >= gFlushTimeout || idle <= 0) {
 
         pipeline.flush();
@@ -346,10 +349,12 @@ void flushPipeline(RedisPipeline& pipeline) {
         gSelectTimeout = INFINITE;
 
         SWSS_LOG_DEBUG("Pipeline flushed");
-    
-        return;
     }
-
-    // postpone the flush
-    gSelectTimeout = gFlushTimeout - idle;
+    else
+    {
+        // skip flushing ppl and set the timeout of fpmsyncd select function to be (gFlushTimeout - idle)
+        // so that fpmsyncd select function would block at most for (gFlushTimeout - idle)
+        // by doing this, we make sure every entry eventually gets flushed
+        gSelectTimeout = gFlushTimeout - idle;
+    }
 }
--- a/fpmsyncd/fpmsyncd.cpp
+++ b/fpmsyncd/fpmsyncd.cpp
@@ -17,8 +17,8 @@
 using namespace std;
 using namespace swss;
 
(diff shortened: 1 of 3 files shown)


siqbal1986 (commented):
Is there a test covering this change. how can we see if this is working as intended.
Also the build is failing. Can you please look into it.

a114j0y:
> Is there a test covering this change. how can we see if this is working as intended. Also the build is failing. Can you please look into it.

Its compilation needs this [pr] to be merged(https://github.com/sonic-net/sonic-swss-common/pull/895)

fpmsyncd/fpmsyncd.cpp
```diff
@@ -9,13 +9,35 @@
...
+#include "fpmsyncd/fpmsyncd.h"
 #include "fpmsyncd/routesync.h"
 
 #include <netlink/route/route.h>
 
 using namespace std;
 using namespace swss;
 
+// SELECT_TIMEOUT specifies the maximum wait time in milliseconds (-1 == infinite)
+static int SELECT_TIMEOUT;
```
liuh-80, line 21: rename to gSelectTimeout?

fpmsyncd/fpmsyncd.cpp
```diff
@@ -304,3 +327,30 @@ int main(int argc, char **argv)
...
+
+    if (remaining == 0) {
+        SELECT_TIMEOUT = INFINITE;
+        return;
+    }
+
+    int idle = pipeline.getIdleTime();
+
+    // flush right away
+    if (remaining < SMALL_TRAFFIC || idle >= gFlushTimeout || idle <= 0 || scheduled) {
```
liuh-80, line 343: Why flush with small traffic?
If there are lots of data, the pipeline will not flush until next timeout?

a114j0y: This optimization aims to reduce pipeline flushes in the case of scaled traffic.

Hence for SMALL_TRAFFIC, do immediate flush to maintain the logic before optimization. So that it would not affect the performance when you want to dispatch several critical routes.

But since we skip several flushes, we need to set a timeout to make sure the routes would eventually get flushed, that's when ``` idle >= gFlushTimeout``` 

```idle <=0``` happens when the system clock is not steady, but since we use steady_clock it shouldn't happen.

File: fpmsyncd/fpmsyncd.cpp (hunk 1 above)
Comment (liuh-80) [reply fpmsyncd/fpmsyncd.cpp]: #Closed

**Review by liuh-80 — APPROVED** [review]

File: fpmsyncd/fpmsyncd.cpp (hunk 2 above)
Comment (liuh-80) [reply fpmsyncd/fpmsyncd.cpp]: #Closed

Suggest add this info to code comments.

**Comment by a114j0y** [comment]
This PR also depends on merging [ PR#954](https://github.com/sonic-net/sonic-swss-common/pull/954)

**Review by prsunny — APPROVED** [review]

File: fpmsyncd/fpmsyncd.cpp [hunk 3]
```diff
@@ -304,3 +326,30 @@ int main(int argc, char **argv)
...
+
+    if (remaining == 0) {
+        gSelectTimeout = INFINITE;
+        return;
+    }
+
+    int idle = pipeline.getIdleTime();
+
+    // flush right away
+    if (remaining < SMALL_TRAFFIC || idle >= gFlushTimeout || idle <= 0) {
```
dgsudharsan, line 342: Please add comments in the code as suggested.

prsunny (approved):

PR #3242, opened 2024-07-22
[orchagent] implement ring buffer feature with a flag

a114j0y:
**What I did**
- add a ring thread for orchdaemon, which would be kicked off if gRingMode is turned on
- support ring buffer feature, currently only enabled for route table executor, which has a scaled use case
- fix the covariant return type issue of `swss::TableBase* Consumer::getConsumerTable() const override`
  - it should return `swss::ConsumerTableBase *`

**Why I did it**
- increase the speed for `APP_ROUTE_TABLE` consumers doing tasks

**How I verified it**
measure the performance with [PerformanceTimer](https://github.com/sonic-net/sonic-swss-common/pull/893)  GitHub issue/pull request detail


siqbal1986:
can you please add sone swss tests for this functionality. There are no tests mentioned in the PR. if you have created a separate PR , Please link it here.

siqbal1986 (commented):
what is hte need of rign buffer. Can you please add some detail regarding its need.

siqbal1986 (approved):

a114j0y:
rebased and conflict resolved

prsunny:
@a114j0y , added other folks to review. Also, this would need unit tests and coverage.

orchagent/orch.cpp
```diff
@@ -239,23 +247,67 @@ void ConsumerBase::dumpPendingTasks(vector<string> &ts)
...
 {
-    // ConsumerBase::execute_impl<swss::ConsumerTableBase>();
     SWSS_LOG_ENTER();
 
-    auto table = static_cast<swss::ConsumerTableBase *>(getSelectable());
-    std::deque<KeyOpFieldsValuesTuple> entries;
-    table->pops(entries);
+    size_t total_size = 0;
+
+    while (true)
```
liuh-80, line 254: This while loop cause high priority notification blocked by low priority notification bug, and removed by this PR:
https://github.com/sonic-net/sonic-swss/pull/3269

Why add it back?

orchagent/orch.cpp
```diff
@@ -239,23 +247,67 @@ void ConsumerBase::dumpPendingTasks(vector<string> &ts)
...
+    {
+        // execute the task right now in this thread if gRingBuffer is not initialized
+        // or the ring thread is not created, or this executor is not served by gRingBuffer
+        task();
+    }
+    else if (!gRingBuffer->Serves(getName())) // not served by ring thread
+    {
+        while (!gRingBuffer->IsEmpty() || !gRingBuffer->Idle) {
+            gRingBuffer->notify();
+            std::this_thread::sleep_for(std::chrono::milliseconds(SLEEP_MSECONDS));
```
liuh-80, line 291: Every consumer in orchagent work in same thread. which means the code here will sleep entire orchagent 0.5seconds.

I'm warry about 0.5seconds delay for critical event handling.

a114j0y: In this case, the consumer of APP_ROUTE_TABLE is actively working in the ring thread...so the consumer in the main thread waits until the ring thread pauses, since we don't want the two threads to work at the same time (because the two threads are currently dealing with different consumers). Actually not the entire orchagent sleeps for 0.5s since route table consumer is still running.

In the original single-thread case, all consumers work in the main thread and execute one by one, the later consumer executes after the previous consumer...so the 0.5 seconds is reasonable, i think

orchagent/orch.cpp
a114j0y: thanks for reminding...have added a commit to remove the loop : )

orchagent/orch.cpp
liuh-80: I noticed in the Orch::addExecutor(Executor* executor) the APP_ROUTE_TABLE_NAME been add to both m_consumerMap and gRingBuffer, is that by design?

And that make APP_ROUTE_TABLE consumer shared in 2 threads. but sonic-swss-common does not support multiple thread and code may crash.

a114j0y: yes, but  APP_ROUTE_TABLE is added to gRingBuffer just for housekeeping, so that gRingBuffer->Serves() could determine whether a table is served by ring.

we don't need sonic-swss-common to support multi-threading, because in our design, when a route table consumer is running, the main thread only performs pops, and addToSync() and drain() are done in ring thread, the two threads don't share common data, they can work in parallel. And only the route table consumer has multi-threading. For other consumers, since the main thread would wait for the ring to pause, and they never push tasks to ring thread, everything is only done in the main thread, it's still single-thread here.

liuh-80: I understand the async task for addToSync and drain are running in ring buffer thread in OrchDaemon::popRingBuffer().

However, the sleep 0.5seconds is running in main thread.

If main thread need wait ring buffer finish, then why not wait by signal or mutex?

If the task in ring buffer take < 0.5seconds, this PR actually make orchagent slow.

If the task in ring buffer take > 0.5seconds, then there some code may running in both ring buffer thread and main thread and cause race conduction issue, because the drain() method will call orch->dotask(), in some orch, dotask() method will call method defined in other orch.

orchagent/orch.cpp
liuh-80: # Closed

orchagent/orch.cpp
liuh-80: For more information, the consumer of APP_ROUTE_TABLE_NAME will call RouteOrch::DoTask()

    const int routeorch_pri = 5;
    vector<table_name_with_pri_t> route_tables = {
        { APP_ROUTE_TABLE_NAME,        routeorch_pri },
        { APP_LABEL_ROUTE_TABLE_NAME,  routeorch_pri }
    };
    gRouteOrch = new RouteOrch(m_applDb, route_tables, gSwitchOrch, gNeighOrch, gIntfsOrch, vrf_orch, gFgNhgOrch, gSrv6Orch);

However:
1. the consumer of APP_LABEL_ROUTE_TABLE_NAME, will run in main thread, it also will call RouteOrch::DoTask().
2. In the code of RouteOrch::DoTask(), it will call method of gSwitchOrch, gNeighOrch, gIntfsOrch, vrf_orch, gFgNhgOrch, gSrv6Orch, which means there are at least 6 other consumers running in main thread will have potential race condition issue with ring buffer.

a114j0y: > I understand the async task for addToSync and drain are running in ring buffer thread in OrchDaemon::popRingBuffer().
> 
> However, the sleep 0.5seconds is running in main thread.
> 
> If main thread need wait ring buffer finish, then why not wait by signal or mutex?
> 
> If the task in ring buffer take < 0.5seconds, this PR actually make orchagent slow.
> 
> If the task in ring buffer take > 0.5seconds, then there some code may running in both ring buffer thread and main thread and cause race conduction issue, because the drain() method will call orch->dotask(), in some orch, dotask() method will call method defined in other orch.

1. if < 0.5 sec, since main thread already finishes pops when the ring is running, it may not be slow
2. if > 0.5 sec, no race condition, since main thread would not call doTask when ring is active
3. we can use mutex, although no lock already achieves correctness

liuh-80: Let's user gSwitchOrch as example, gSwitchOrch own a consumer for APP_SWITCH_TABLE_NAME.

This table does not serve by ring buffer. my question is:

The drain() method of the APP_SWITCH_TABLE_NAME consumer is running in main thread or in ring buffer thread?

According to the code here, when this consumer execute(), the code here will:
1. notify ring buffer.
2. wait 0.5 seconds
3. run the drain() method at line 279. which is in main thread.

Then what will happen of the async task in ring buffer thread run longer than 0.5seconds?

a114j0y: currently, only route table consumer has the chance to run two threads at the same time (and the two threads don't have thread issue, because the main thread only performs pops, which only modifies redis table and the local variable), for other consumers, when they are chosen to execute(), they can only do pops, and proceed only if the ring thread is not active, otherwise  wait for the ring thread to pause (to avoid thread issue). so that we maintain the order among consumers, the multi-threading is only within the route table consumer itself.

how to make sure the ring thread has paused?
- it's empty && idle flag is true (since even if the ring is empty, maybe it's still running the last popped task, so need to check the idle flag)

to make sure we don't change the sequence order, we need to make sure the ring buffer is empty before allowing the next consumer to do addToSync and drain.

to avoid the ring thread being stuck, when the main thread determines that it needs to wait, it not only waits for 0.5s, but also sends out cv.notify to make sure there is no deadlock.

To answer your question,
1. all tables other than route table, runs in main thread
2. when it's selected to execute(), it immediately performs pops, without waiting. In original design, this pops happens after the previous consumer's drain, but currently they can be parallel
3. when async task in ring runs > 0.5 sec, main thread checks whether ring thread is paused, and found that idle is false, so it notifies the ring and wait another 0.5s
4. after the 0.5s, the main thread checks whether ring thread is paused again
    - you may question, maybe 0.5s is too much, but for the main thread, compared with the original design, even if it wastes the 0.5s here, it saves the time for pops, they can be balanced out  
5. it can continue to do addToSync and drain only after the ring buffer is empty and the ring thread is not running

liuh-80: I still don't understand:
1. for this: "when it's selected to execute(), it immediately performs pops, without waiting. In original design, this pops happens after the previous consumer's drain, but currently they can be parallel"

For other tables, the async task of drain still running in main thread, the main thread still needs wait the async task finish to continue, which means it's not parallel.


2. "Not 6 seconds, still 1 sec"

For every event, after pop, the main thread will sleep o.5 seconds, so if there are 10 event, the main thread sleep time will be 5 seconds. why still 1 sec?

a114j0y: the wait is in a while loop, the wait is not once, if ring thread is still active, APP_SWITCH_TABLE_NAME consumer  keeps waiting

liuh-80: Oh, sorry, my mistake.
But then do you mean when APP_SWITCH_TABLE_NAME consumer running in background, the main thread actually waiting in the while loop?

Also, the second point means, when an APP_SWITCH_TABLE_NAME events happen with any another table event, the second event will always block 0.5seconds.

a114j0y: the 0.5 sec delay would only be possible if APP_SWITCH_TABLE_NAME follows APP_ROUTE_TABLE_NAME, if it follows other kinds of table, the condition ```(!gRingBuffer->IsEmpty() || !gRingBuffer->Idle)``` is 100% false

liuh-80: Is there test result show how much performance improve? because the only improve is APP_ROUTE_TABLE_NAME table pops data parallel with drain data. and pops data actually very fast.

Not sure any critical event follows after an APP_ROUTE_TABLE_NAME event will have 0.5seconds delay is acceptable.

a114j0y: it shows benefits in the scaled case, when pops is really slow (redis i/o)
https://github.com/sonic-net/SONiC/pull/1521 shows it helps to accelerate BGP loading by 50%

liuh-80: #Closed

orchagent/orchdaemon.cpp
```diff
@@ -105,6 +111,43 @@ OrchDaemon::~OrchDaemon()
...
+    // make sure there is only one thread created to run popRingBuffer()
+    if (!gRingBuffer || gRingBuffer->threadCreated)
+        return;
+
+    gRingBuffer->threadCreated = true;
+    SWSS_LOG_NOTICE("OrchDaemon starts the popRingBuffer thread!");
+
+    while (!ring_thread_exited)
+    {
+        gRingBuffer->pause_thread();
```
liuh-80, line 127: RingBufffer is a singleton share by 2 threads, however there is no multithread protect.
So when RingBufffer instance delete by main thread, the gRingBuffer here will become a dangling pointer.

a114j0y: but the testcase is single-threaded

and in the orchdaemon destructor, we set ring_thread_exited as true to avoid delete gRingBuffer before detach the ring thread

liuh-80: The test case may be failed by other reason.

The variable name ring_thread_exited suggests that the ring thread has exited, but this is not actually the case. When ring_thread_exited is true, the while loop might still be executing, and you cannot predict the ring thread or the main thread exist first.

These two threads are scheduled by the system. The system might suspend the ring thread for a few milliseconds at any point in the while loop. During that time, the main thread might exit and destruct the ring buffer.

liuh-80: There is another possible reason for the UT issue:

The dtor of orchdaemon will access gRingBuffer, also dtor of RingBuffer class will delete the instance. which dtor will invoke first? if RingBuffer class destroy first, then gRingbuffer also will become a dangling pointer.

My suggestion is RingBuffer should not be a singleton and release in the dtor of RingBuffer class.

a114j0y: sure, 
1. a try catch block is added for the while loop in ring thread, to cover the case ring buffer is released before ring thread is detached.
2. ring buffer is released by its own dtor
in orchdaemon dtor , the local variable `gRingBuffer` is only used in this line to indicate whether we has enabled gRingMode
```
if (gRingBuffer) {
        ring_thread_exited = true;
        ring_thread.detach();}
```
if it's not nullptr, it indicates we enabled gRingMode, so need to detach the thread and set  ring_thread_exited as true, nothing else, so should be good here

liuh-80: When access a dangling pointer on linux system, it will trigger a SIGSEGV signal.
And as I know a signal is not a C++ exception. so, try-catch will not work.

My suggestion is:
1. continue use singleton, but in while loop, always use get() method to access singleton. and get() method need improve to handle multithread, also while loop need handle get() return null pointer case (when singleton destructued). which is really difficult and low performance.

2.Don't use singleton, create a global instance and manage by OrchDaemon.

a114j0y: singleton has API Get() to access the global and unique ring buffer

gRingBuffer is just a variable that records what Get() returns

we use it to compare with nullptr to check whether ring is enabled, we can always use OrchRing::Get() to access the unique ring buffer

liuh-80: What will happen when dtor of ring buffer called before popRingBuffer() thread exit? popRingBuffer() thread will cause SIGSEGV and make orchagent crash.

This kind crash is very difficult to find root cause.

a114j0y: No one would call the dtor of ringbuffer, its protected and can't be called by others

prsunny:
@a114j0y , can you check the coverage? can it be fixed?

a114j0y:
depending on [#pr3482](https://github.com/sonic-net/sonic-swss/pull/3482)

liuh-80 (approved):

orchagent/orch.cpp
```diff
@@ -253,23 +330,52 @@ void ConsumerBase::dumpPendingTasks(vector<string> &ts)
...
+        }
+        gRingBuffer->notify();
+    }
 }
 
 void Consumer::drain()
 {
-    if (!m_toSync.empty())
-        ((Orch *)m_orch)->doTask((Consumer&)*this);
+    m_orch->doTask(*this);
```
prsunny, line 378: Why is this change? seems like modifying existing flow. Can you confirm there is no change to current flow if RingBuffer is *NOT* enabled?

orchagent/orchdaemon.cpp
```diff
@@ -871,6 +925,21 @@ void OrchDaemon::start(long heartBeatInterval)
              * requests live in it. When the daemon has nothing to do, it
              * is a good chance to flush the pipeline  */
             flush();
+
+            if (gRingBuffer)
+            {
+
```
prsunny, line 931: extra line

orchagent/orchdaemon.cpp
```diff
@@ -888,9 +957,9 @@ void OrchDaemon::start(long heartBeatInterval)
         /* After each iteration, periodically check all m_toSync map to
          * execute all the remaining tasks that need to be retried. */
 
-        /* TODO: Abstract Orch class to have a specific todo list */
-        for (Orch *o : m_orchList)
-            o->doTask();
+        if (!gRingBuffer || (gRingBuffer->IsEmpty() && gRingBuffer->IsIdle()))
```
prsunny, line 960: can you add braces `{`  for the `if` check?

orchagent/orchdaemon.cpp
```diff
@@ -902,7 +971,18 @@ void OrchDaemon::start(long heartBeatInterval)
             bool ret = warmRestartCheck();
             if (ret)
             {
+
```
prsunny, line 974: please remove extra line

orchagent/orchdaemon.h
```diff
@@ -97,14 +115,15 @@ class OrchDaemon
...
-    
+
     std::chrono::time_point<std::chrono::high_resolution_clock> m_lastHeartBeat;
 
     void flush();
 
     void heartBeat(std::chrono::time_point<std::chrono::high_resolution_clock> tcurrent, long interval);
 
     void freezeAndHeartBeat(unsigned int duration, long interval);
+
```
prsunny, line 126: remove extra line thats not related to this PR

orchagent/orchdaemon.h
```diff
@@ -97,14 +115,15 @@ class OrchDaemon
 
     std::vector<Orch *> m_orchList;
     Select *m_select;
-    
+
```
prsunny, line 118: Same as other comment

orchagent/orch.cpp
```diff
@@ -253,23 +330,52 @@ void ConsumerBase::dumpPendingTasks(vector<string> &ts)
...
 
-    auto table = static_cast<swss::ConsumerTableBase *>(getSelectable());
-    std::deque<KeyOpFieldsValuesTuple> entries;
-    table->pops(entries);
+    auto entries = std::make_shared<std::deque<KeyOpFieldsValuesTuple>>();
+    getConsumerTable()->pops(*entries);
 
-    // add to sync
-    addToSync(entries);
+    pushRingBuffer([=](){
```
prsunny, line 338: IMO, this should be invoked only if gRingBuffer is true. Otherwise, follow the existing flow. @qiluo-msft , @liuh-80 , what do you think?

a114j0y: if gRingBuffer is not true, calling pushRingBuffer is equivalent to directly executing the functions inside pushRingBuffer

orchagent/orch.cpp
a114j0y: they seem to be equivalent, those type conversion seems to be unnecessary, can you confirm that @liuh-80 ?

orchagent/orch.cpp
a114j0y: do you think we can use a more general function name to replace `pushRingBuffer` with, for example, `processLambdaFunc`, where a bunch of tasks are packaged into a lambda function body for processing, we either directly execute this task package in the main thread or push it to the ring buffer for the ring thread to execute

eddieruan-alibaba:
@prsunny do you have any more comments?

orchagent/orch.cpp
prsunny: yes, prefer to have cleaner function names and separate out the ring buffer handling

a114j0y:
Modified.

orchagent/orch.cpp
a114j0y: https://github.com/sonic-net/sonic-swss/pull/3242/commits/a2fc65b37653de3f74f0ba185689f0a7becb8caf

changed to `processTasks`

there would be if else branches inside to decide whether to push into ringbuffer or directly execute the task

liuh-80 (approved):

prsunny (approved):

liuh-80:
@a114j0y , I recently testing improve route performance with ring buffer/zmq/multiple-db, and the ring buffer sometimes will cause stuck when there are massive bgp routes, I check this on a lab mellanox 4600 device which has 12000+ BGP routes:

~$ sonic-db-cli ASIC_DB eval "return #redis.call('keys', 'ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY*')" 0
12951

~$ show ip bgp summary

IPv4 Unicast Summary:
BGP router identifier 10.1.0.32, local AS number 65100 vrf-id 0
BGP table version 22923
RIB entries 12851, using 1644928 bytes of memory
Peers 24, using 494208 KiB of memory
Peer groups 2, using 128 bytes of memory


Neighbhor      V     AS    MsgRcvd    MsgSent    TblVer    InQ    OutQ  Up/Down      State/PfxRcd  NeighborName
-----------  ---  -----  ---------  ---------  --------  -----  ------  ---------  --------------  --------------
10.0.0.1       4  65200       3201       5537     22923      0       0  00:00:34             6370  ARISTA01T2
10.0.0.5       4  65200       3201       5537     22923      0       0  00:00:35             6370  ARISTA03T2
10.0.0.9       4  65200       3201       5537     22923      0       0  00:00:34             6370  ARISTA05T2
10.0.0.13      4  65200       3201       5537     22923      0       0  00:00:34             6370  ARISTA07T2
,,,

Then run "sudo config bgp shutdown all" or "sudo config bgp startup all", and check the syslog, some time I get following log:

2025 Mar 17 06:56:42.396204 DEVICE_NAME WARNING swss#orchagent: :- processAnyTask: ring is full...push again
2025 Mar 17 06:56:42.439894 DEVICE_NAME  INFO swss#rsyslogd: imuxsock[pid: 48, name: /usr/bin/orchagent] from <bjw2-can-4600c-3:orchagent>: begin to drop messages due to rate-limiting
2025 Mar 17 06:57:42.043606 DEVICE_NAME  WARNING swss#supervisor-proc-exit-listener: Process 'orchagent' is stuck in namespace 'host' (1.0 minutes).
2025 Mar 17 06:58:13.002017 DEVICE_NAME  INFO swss#rsyslogd: imuxsock[pid: 48, name: /usr/bin/orchagent]: 23441221 messages lost due to rate-limiting (20000 allowed within 300 seconds)
2025 Mar 17 06:58:13.082530 DEVICE_NAME  INFO swss#rsyslogd: imuxsock[pid: 48, name: /usr/bin/orchagent] from <DEVICE_NAME :orchagent>: begin to drop messages due to rate-limiting

Seems there are some deadlock issue in the ring buffer code.

PR #3258, opened 2024-08-16
[Orchagent] Add optional create_switch timeout parameter

pavannaregundi:
Change adds optional create_switch timeout as command line parameter to orchagent.

**What I did**
Change adds optional create_switch timeout as command line parameter to orchagent.

**Why I did it**
Older platforms are seeing increase in time required in bookworm based branches. 

**How I verified it**
Changes are verified to see passed timeout value is reflected only for create_switch api call from orchagent.

**Details if related**
Migrating some older armhf platform from SONiC 202311 to master and 202405 branch, resulted in syncd create_switch timeout.
It is analyzed to see that SAI SDK code remains the same and time consume by SAI SDK during init when some(snmp mgmt-framework lldp gnmi dhcp_relay radv teamd pmon bgp) of the docker in sonic are disabled is same as running on 202311(with same environment).
Some of the other observations are:

* Switch create time varies from 50sec to 85 sec randomly depending on CPU load.
* Running bullseye based 5.10 sonic kernel yielded same result on latest master.
* Running docker ps -a in 202405 shows all dockers attempting to start, while in 202311 it delays most dockers
* No single docker seems to have a significant impact


Blueve:
@prsunny can you help review?

@kcudnik this is the alternative way to allow some low performance switch run higher SONiC version
Previous PR has closed: https://github.com/sonic-net/sonic-sairedis/pull/1406

tests/test_zmq.py
```diff
@@ -57,8 +57,9 @@ class TestZmqDash(object):
     @pytest.fixture(scope="class")
     def enable_orchagent_zmq(self, dvs):
         # change orchagent to use ZMQ
+        # change orchagent to use custom create_switch_timeout
         dvs.runcmd("cp /usr/bin/orchagent.sh /usr/bin/orchagent.sh_zmq_ut_backup")
-        dvs.runcmd("sed -i.bak 's/\/usr\/bin\/orchagent /\/usr\/bin\/orchagent -q tcp:\/\/127.0.0.1:8100 /g' /usr/bin/orchagent.sh")
+        dvs.runcmd("sed -i.bak 's/\/usr\/bin\/orchagent /\/usr\/bin\/orchagent -q tcp:\/\/127.0.0.1:8100 -t 60 /g' /usr/bin/orchagent.sh")
```
prsunny, line 62: why this specific test is changed?

pavannaregundi: As the change is in main() passing a new argument to orchagent, found we that we are doing a similar argument change in this test case. This change does not affect the behavior of the test as timeout value we are sending is still same as default(60 sec). But helps to cover the UT for new code change.

orchagent/main.cpp
```diff
@@ -69,10 +69,11 @@ uint32_t gCfgSystemPorts = 0;
 string gMyHostName = "";
 string gMyAsicName = "";
 bool gTraditionalFlexCounter = false;
+uint32_t create_switch_timeout = 0;
 
 void usage()
 {
-    cout << "usage: orchagent [-h] [-r record_type] [-d record_location] [-f swss_rec_filename] [-j sairedis_rec_filename] [-b batch_size] [-m MAC] [-i INST_ID] [-s] [-z mode] [-k bulk_size] [-q zmq_server_address] [-c mode]" << endl;
+    cout << "usage: orchagent [-h] [-r record_type] [-d record_location] [-f swss_rec_filename] [-j sairedis_rec_filename] [-b batch_size] [-m MAC] [-i INST_ID] [-s] [-z mode] [-k bulk_size] [-q zmq_server_address] [-c mode] [-t create_switch_timeout]" << endl;
```
dgsudharsan, line 76: Are we planning to change the existing timeout? How will this be controlled from user perspective?

Blueve: This option provides the flexibility to override the existing timeout, it is optional because we most of platforms' performance are good enough to create switch within default timeout. 

The timeout setting can be controlled by platform designer, e.g. https://github.com/sonic-net/sonic-buildimage/pull/19928/files

Blueve:
@pavannaregundi can you create PR for 202311 as there is conflict?

Blueve (approved):

prsunny (approved):

Blueve:
ADO: 29208990

kcudnik:
switch_create must take less than 30 seconds on warm boot scenario to not cause dataplane interruption, higher values are not acceptable in production at all

PR #3260, opened 2024-08-21
[portsorch] query SAI_HOSTIF_ATTR_QUEUE switch capability once

stepanblyschak:
**What I did**

Support for SAI_HOSTIF_ATTR_QUEUE is a switch level capability, no need to query it per port.

**Why I did it**

As part of the goal to optimize boot time for systems with many ports. This one is the first step into that direction, extracting all non per-port operations.

**How I verified it**

Boot the system, verify no call to query SAI_HSOTIF_ATTR_QUEUE per port happens.

**Details if related**


prsunny (approved):

stepanblyschak:
@prsunny @liat-grozovik Could you please help to merge?

PR #3264, opened 2024-08-23
[MuxOrch] Init  mux cables in startup

lolyu:
**What I did**
let's init the mux cables from mux config table in the `MuxOrch` construction.

**Why I did it**
To fix: https://github.com/sonic-net/sonic-buildimage/issues/19996

**How I verified it**
UT

**Details if related**


lolyu:
This is not a viable solution to fix this.
During orchagent init stage, there is no mux tunnel, which should be created by tunnel decap orch. So the `tryInitMuxCable` cannot create the mux cables.

PR #3269, opened 2024-08-28
Fix port up/bfd sessions bringup notification delay issue.

liuh-80:
Fix port up/bfd sessions bringup notification delay issue.

#### Why I did it
Fix following issue:
https://github.com/sonic-net/sonic-buildimage/issues/19569

##### Work item tracking
- Microsoft ADO: 29192284

#### How I did it
Revert change in Consumer::execute(), which introduced by this commit:
https://github.com/sonic-net/sonic-swss/commit/92589789aa79bf1e70937a35cb06eff8a358ab6b#diff-96451cb89f907afccbd39ddadb6d30aa21fe6fbd01b1cbaf6362078b926f1f08

The change in this commit add a while loop:
do
{
    std::deque<KeyOpFieldsValuesTuple> entries;
    table->pops(entries);
    update_size = addToSync(entries);
} while (update_size != 0);

The addToSync sync method will return the size of entries
Which means, if there are massive routes notification, other high priority notification for example port up notification will blocked until all routes notification been handled.

#### How to verify it
Pass all UT.
Manually verify issue fixed.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Tested branch (Please provide the tested image version)

- [x]  SONiC.master-20030.629638-f370e2fa8

Fix port up/bfd sessions bringup notification delay issue.


mlok-nokia:
@liuh-80 With this change, we have tried multiple times to reboot single linecard, all ports come up in 7 mins (The existing takes more than 15 mins).  all the 34k routes are installed also.  It seems this change fixes ports up delay.   We still need to run more tests, include OC.

mlok-nokia:
@liuh-80 I built an image with this change and tested.  For the first time boot up after installation on a single linecard,  all ports come up in 8 minutes and all 34k routes are also installed.  For subsequent reboot a single linecard, it takes about 7 minutes for all linkup and 34k routes installed.  It seems this change addresses the issue.  We need to do more testing to verify that, includes the OC testing.

liuh-80:
@bocon13 , we found a performance issue which cause by your PR, can you review this fix?

Performance issue: https://github.com/sonic-net/sonic-buildimage/issues/19569
PR cause performance issue: https://github.com/sonic-net/sonic-swss/pull/1992

mlok-nokia (approved):

lguohan:
@liuh-80 , we must need some UT to prevent such regression.

wenyiz2021:
thanks @liuh-80 ! just curious what is the result of Comsumer pop notification once VS pop until size of entry is 0?

anamehra (approved):

qiluo-msft (approved):

liuh-80:
> Comsumer

Will add sonic-swss test case to prevent this issue happen again,

liuh-80:
> thanks @liuh-80 ! just curious what is the result of Comsumer pop notification once VS pop until size of entry is 0?

This will make Consumer pop all notifications belong to current consumer, so higher priority notification will be blocked.

liuh-80:
The PR validation failure seems not caused by code change, a PR change nothing also failed:
https://github.com/sonic-net/sonic-swss/pull/3276

liuh-80:
> @liuh-80 , we must need some UT to prevent such regression.

Fixed, added new unit test to prevent regression.

wenyiz2021:
> > thanks @liuh-80 ! just curious what is the result of Comsumer pop notification once VS pop until size of entry is 0?
> 
> This will make Consumer pop all notifications belong to current consumer, so higher priority notification will be blocked.

thanks! then does current design in this PR ensure the priority task notify, or it only notify the first come notification?

saksarav-nokia:
@liuh-80 , we ran OC with this fix and noticed orchagent crashes in all test beds. Analyzing the syslog files. I will update my findings soon.

prsunny:
@mint570 for viz

liuh-80:
> > > thanks @liuh-80 ! just curious what is the result of Comsumer pop notification once VS pop until size of entry is 0?
> > 
> > 
> > This will make Consumer pop all notifications belong to current consumer, so higher priority notification will be blocked.
> 
> thanks! then does current design in this PR ensure the priority task notify, or it only notify the first come notification?

With this PR, each Consumer will pop less than 128 notifications every time. which means orchagent will check and handle high priority notification more frequently.

prsunny:
> > > > thanks @liuh-80 ! just curious what is the result of Comsumer pop notification once VS pop until size of entry is 0?
> > > 
> > > 
> > > This will make Consumer pop all notifications belong to current consumer, so higher priority notification will be blocked.
> > 
> > 
> > thanks! then does current design in this PR ensure the priority task notify, or it only notify the first come notification?
> 
> With this PR, each Consumer will pop less than 128 notifications every time. which means orchagent will check and handle high priority notification more frequently.

does this mean, bulk-api's (SAI) that orchagent invoke for routes etc will now have a limit of 128 entries

tests/dvslib/dvs_common.py
```diff
@@ -16,7 +16,7 @@ class PollingConfig:
         strict: If the strict flag is set, reaching the timeout will cause tests to fail.
     """
 
-    polling_interval: float = 0.01
+    polling_interval: float = 0.1
```
liuh-80, line 19: Try fix test case failed
Code change in this PR will not pops all entries within one Consumer::exec() API call, which may cause data base update delay.

liuh-80:
> > > > > thanks @liuh-80 ! just curious what is the result of Comsumer pop notification once VS pop until size of entry is 0?
> > > > 
> > > > 
> > > > This will make Consumer pop all notifications belong to current consumer, so higher priority notification will be blocked.
> > > 
> > > 
> > > thanks! then does current design in this PR ensure the priority task notify, or it only notify the first come notification?
> > 
> > 
> > With this PR, each Consumer will pop less than 128 notifications every time. which means orchagent will check and handle high priority notification more frequently.
> 
> does this mean, bulk-api's (SAI) that orchagent invoke for routes etc will now have a limit of 128 entries

Yes, it will have a 128 entries limit.
This is an orchagent design issue.

orchagent/orch.cpp
```diff
@@ -242,14 +242,12 @@ void Consumer::execute()
...
     SWSS_LOG_ENTER();
 
-    size_t update_size = 0;
     auto table = static_cast<swss::ConsumerTableBase *>(getSelectable());
-    do
-    {
-        std::deque<KeyOpFieldsValuesTuple> entries;
-        table->pops(entries);
-        update_size = addToSync(entries);
-    } while (update_size != 0);
```
liuh-80, line 252: Multiple test case failed.
Will try different solution in another POC PR

liuh-80:
> The following tests in pc suite seems to be triggering the crashes. orchagent tries to remove the neighbor and nexthop, either meta layer or SAI is not in sync with orchagent and returns error which causes orchagent to exit. test_po_update_io_no_loss test_po_update

I think it's a timing issue, for example the validation of this PR has lots of test case failed, but after I increase the wait_for_n_keys timeout, many test case passed.

However this change do impact performance, because after this change every doTask() call can only handle 128 entries, so some scenario take longer time.

I'm trying to only change RouteOrch to improve performance.

siqbal1986 (approved):

orchagent/orch.cpp
```diff
@@ -789,7 +787,7 @@ void Orch::addConsumer(DBConnector *db, string tableName, int pri)
     }
     else
     {
-        addExecutor(new Consumer(new ConsumerStateTable(db, tableName, gBatchSize, pri), this, tableName));
+        addExecutor(new Consumer(new ConsumerStateTable(db, tableName, gBatchSize * CONSUMER_POP_MAX_BATCH_COUNT, pri), this, tableName));
```
liuh-80, line 790: This change will cause an issue:

https://github.com/sonic-net/sonic-buildimage/commit/286ec3edbf91ec1bd09b6eae3a4d69801bdd94a9


Background running lua script may cause redis-server quite busy if batch size is 8192.
If handling time exceeded default 5s, the redis-server will not response to other process and will cause syncd crash.

So that's also part of reason why add a loop in execute()

orchagent/orch.cpp
```diff
@@ -242,14 +244,16 @@ void Consumer::execute()
     // ConsumerBase::execute_impl<swss::ConsumerTableBase>();
     SWSS_LOG_ENTER();
 
-    size_t update_size = 0;
     auto table = static_cast<swss::ConsumerTableBase *>(getSelectable());
+    int batch_count = 0;
+    size_t update_size = 0;
```
liuh-80, line 249: Will move this loop to swss-common ConsumerStateTable::pops() method.

https://github.com/sonic-net/sonic-swss-common/pull/916

orchagent/orch.cpp
```diff
@@ -13,6 +13,8 @@
 #include "zmqconsumerstatetable.h"
 #include "sai_serialize.h"
 
+#define CONSUMER_POP_MAX_BATCH_COUNT 8
```
lguohan, line 16: i do not understand this POP MAX_BATCH, it looks like another batch. We should only have one batch size. hard to reasoning if we have another batch control.

liuh-80: I will revert this change to the original fix

qiluo-msft (approved):

saksarav-nokia:
We tried the fix in the PR in the Voq chassis and seeing the following issue
  The IMM has two asics and has 2 pot channels in each asic and 2 port members in each port channel.
The ip address is configured on each port channel and  bgp is eanbled. The neighbor and routes are learned  on these port channel.
In sonic-mgmt pc suite, the test case po-update removes the port members from one of the port channel, removes the ip address configured on that port channel, creates new port channel, adds the same port members to the new port channel, adds the same ip address to the new port channel.
In the remote asic, before all the routes learned on the old port channel are removed by routeOrch, the neighbor and nexthop for the old portchannel are being attempted to be removed. But since the routes are pending, the old nexthop and neighbor are not removed. Then the neighbor and nexthop for the new port channel are being added. If the neighbor is learned on remote system port in remote asic, the nexthop is added with alias as inband port's alias, so the key (ip,alias) is same for both old nexthop and new nexthop. When the new nexthop is added , it calls hasNextHop function to check if the nexthop with (ip-address, alias) as key and since the old nexthop is not removed yet, the hasNextHop returns true, however the assert(!hasNextHop) does n't trigger the crash. So addNextHop function replace the old nexthop with old rif-id with new nexthop with new old rif-id in the nexthop map. Then after all the routes learned on old port channel is removed, the old neighbor and old nexthop are being removed. Sine the old nexthop was replaced with new nexthop, when orchagent tries to delete the old nexthop, it   actually deletes the new nexthop from SAI. Then when it tries to remove the old neighbor, SAI returns error since orchagent removed the new nexthop from SAI instead of old nexthop and old neighbor is still referenced by the old nexthop in SAI. So orchagent crashes when SAI returns error.

liuh-80:
> We tried the fix in the PR in the Voq chassis and seeing the following issue The IMM has two asics and has 2 pot channels in each asic and 2 port members in each port channel. The ip address is configured on each port channel and bgp is eanbled. The neighbor and routes are learned on these port channel. In sonic-mgmt pc suite, the test case po-update removes the port members from one of the port channel, removes the ip address configured on that port channel, creates new port channel, adds the same port members to the new port channel, adds the same ip address to the new port channel. In the remote asic, before all the routes learned on the old port channel are removed by routeOrch, the neighbor and nexthop for the old portchannel are being attempted to be removed. But since the routes are pending, the old nexthop and neighbor are not removed. Then the neighbor and nexthop for the new port channel are being added. If the neighbor is learned on remote system port in remote asic, the nexthop is added with alias as inband port's alias, so the key (ip,alias) is same for both old nexthop and new nexthop. When the new nexthop is added , it calls hasNextHop function to check if the nexthop with (ip-address, alias) as key and since the old nexthop is not removed yet, the hasNextHop returns true, however the assert(!hasNextHop) does n't trigger the crash. So addNextHop function replace the old nexthop with old rif-id with new nexthop with new old rif-id in the nexthop map. Then after all the routes learned on old port channel is removed, the old neighbor and old nexthop are being removed. Sine the old nexthop was replaced with new nexthop, when orchagent tries to delete the old nexthop, it actually deletes the new nexthop from SAI. Then when it tries to remove the old neighbor, SAI returns error since orchagent removed the new nexthop from SAI instead of old nexthop and old neighbor is still referenced by the old nexthop in SAI. So orchagent crashes when SAI returns error.

Hi @saksarav-nokia , the code change in this PR does not change any orch logic, so the as my understand this issue is a bug in orchagent and it's already there. is there a plan to fix the issue?

orchagent/orch.cpp
liuh-80: Change reverted

liuh-80:
If this PR need cherry-pick, following PR also need cherry-pick first:
https://github.com/sonic-net/sonic-swss/pull/3304
https://github.com/sonic-net/sonic-swss/pull/3305

yejianquan:
@bingwang-ms  for viz

PR #3278, opened 2024-09-04
[POC] Improve routeorch to stop process routes when high priority notification coming.

liuh-80:
Please ignore this POC


liuh-80:
> @liuh-80 , Is this PR ready to try in our test beds?

@saksarav-nokia , no, there still many sonic-mgmt test case failed:
https://github.com/sonic-net/sonic-buildimage/pull/20143

I will investigation the reason and fix it this week

liuh-80:
This approach can pass all sonic-swss test case but can't pass sonic-mgmt test case because performance issue.
Will try another approach.

PR #1926, opened 2021-09-23
[swss] Install IPv6 link local route for new VRF

iMasaruOki:
Signed-off-by: Masaru OKI <masaru.oki@gmail.com>

**What I did**
Fix for issue: Azure/sonic-buildimage#8831

**Why I did it**
When created new VRF, IPv6 link local route is not installed for the VRF.
It causes BGP unnumbered (ipv4 route via ipv6 nexthop) is not working.

**How I verified it**

`Ethernet0` of DUT is connected to another switch speaking BGP (ASN 100).

Setting:
```
sudo config vrf add Vrf1
sudo config interface vrf bind Ethernet0 Vrf1
sudo config ipv6 enable link-local
vtysh \
-c "conf t" \
-c "router bgp 100 vrf Vrf1" \
-c "  router-id 1.1.1.1" \
-c "  neighbor Ethernet0 interface remote-as internal"
```

Wait for minutes, and run `vtysh -c 'show bgp vrf all summary'`

**Details if related**

PR #1862, opened 2021-08-09
Added VRF changes for IPv6 link local enhancements

AkhileshSamineni:
The VRF changes for IPv6 link local enhancements =>  Adding/Deleting IPv6 LL /10 and /128 routes for every VR.

These are missing changes from PR - https://github.com/Azure/sonic-swss/pull/1463

Signed-off-by: Akhilesh Samineni <akhilesh.samineni@broadcom.com>

PR #1463, opened 2020-10-10
Code changes to support IPv6 Link local enhancements

AkhileshSamineni:
Code changes :

1. Changes to handle the "ipv6_use_link_local_only" option on a interface which is added/deleted based on Global Config or Interface config.
2. Adds/Deletes the RIF for an interface in the absence of IPv6 global address on a interface based on "ipv6_use_link_local_only".
3. Allow the ipv6 link-local address as neighbors.
4. Allow adding the IPv6 routes with link-local nexthops.
5. Changes in fpm to handle ipv6 nexthop with ipv4 prefix (as BGP is going to send ipv6 link-local as the nexthop in case of 5549).
6. Skip Ipv4LL neighbor add in Orchagent for RFC5549.

As per HLD - https://github.com/Azure/SONiC/pull/625

Depends on :
https://github.com/Azure/sonic-utilities/pull/1159
https://github.com/Azure/sonic-buildimage/pull/5584

Signed-off-by: Akhilesh Samineni <akhilesh.samineni@broadcom.com>

PR #3280, opened 2024-09-06
Fix IPv6 link-local address can't work in VRF

link19430:
#### What I did
Add/Delete IPv6 link-local route to me in VRF when VRF add/delete to trap packet to CPU

#### Why I did it
The IPv6 link-local address can't work in VRF.
The packet which ip destination is IPv6 link-local address would be dropped.
The root cause is that there is not route to me for the IPv6 link-local address on VRF

#### How to verify it
Ping the IPv6 link-local address of DUT via the interface which is added in VRF

### DIFF
--- a/orchagent/p4orch/tests/fake_fdborch.cpp
+++ b/orchagent/p4orch/tests/fake_fdborch.cpp
@@ -1,0 +1,94 @@
+#include "fdborch.h"
+
+const int FdbOrch::fdborch_pri = 20;
+
+FdbOrch::FdbOrch(DBConnector* applDbConnector, vector<table_name_with_pri_t> appFdbTables,
+    TableConnector stateDbFdbConnector, TableConnector stateDbMclagFdbConnector, PortsOrch *port) :
+    Orch(applDbConnector, appFdbTables),
+    m_portsOrch(port),
+    m_fdbStateTable(stateDbFdbConnector.first, stateDbFdbConnector.second),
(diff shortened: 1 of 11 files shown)


prsunny:
There is a link-local subnet route to cpu. Why do you see link-local is not working?

link19430:
Current link-local subnet route to cpu works only for default VRF.

https://github.com/sonic-net/sonic-swss/blob/3c230d2b51ebf2ffc7163b2641ffab7ef358bfd4/orchagent/routeorch.cpp#L147

https://github.com/sonic-net/sonic-swss/blob/3c230d2b51ebf2ffc7163b2641ffab7ef358bfd4/orchagent/routeorch.cpp#L154

orchagent/p4orch/tests/fake_fdborch.cpp
```diff
@@ -0,0 +1,94 @@
+#include "fdborch.h"
```
prsunny, line 1: What are these test changes?

puffc: @prsunny This change is actually related to the following PRs: #1463, #1862, #1926, #3280 and issue https://github.com/sonic-net/sonic-buildimage/issues/8831.

puffc: I believe the purpose of these test changes is to fix the compilation error in the p4Orch/tests directory.

PR #3287, opened 2024-09-11
[POC] Fix high priority notification blocked by massive route notification issue.

liuh-80:
Please ignore this POC PR.

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -34,7 +34,7 @@ extern size_t gMaxBulkSize;
 #define DEFAULT_NUMBER_OF_ECMP_GROUPS   128
 #define DEFAULT_MAX_ECMP_GROUP_SIZE     32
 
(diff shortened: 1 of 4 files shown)


orchagent/routeorch.cpp
```diff
@@ -498,6 +500,13 @@ void RouteOrch::doTask(Consumer& consumer)
         // Add or remove routes with a route bulker
         while (it != consumer.m_toSync.end())
         {
+            exitentryCount++;
+            if (exitentryCount >= 5120)
```
liuh-80, line 504: A protentional issue: this will break warm reboot, because there will be items left in m_toSync when warmreboot message coming, and will cause 
2024 Sep 11 07:56:54.799275 vlab-01 NOTICE swss#orchagent: :- warmRestoreValidation: There are pending consumer tasks after restore orchangen restore failed after warm reboot: 

2024 Sep 11 08:01:19.708042 vlab-01 NOTICE swss#orchagent: :- setWarmStartState: orchagent warm start state changed to restored
2024 Sep 11 08:01:19.708042 vlab-01 ERR swss#orchagent: :- warmRestoreAndSyncUp: Orchagent state restore failed

PR #3288, opened 2024-09-13
[ACL] Add support for COPY action

devpatha:
**What I did**
Added COPY packet action, that uses SAI_PACKET_ACTION_COPY attribute.

**Why I did it**
This allows ACL to copy packets to CPU. It can be used to dump and debug transit packets.

**How I verified it**
Install the ACL with COPY action and verify that packets are punted towards the CPU with matching criteria.

**Details if related**


prsunny (approved):

prsunny:
@bingwang-ms for viz

devpatha:
@prsunny Could we please get it in 202405? Shall I raise a separate PR for it?

bingwang-ms:
@developfast , @prsunny Sorry I missed the review of this PR. The only concern for this change is different capabilities on different platforms. Not sure if we have checked that? 
And some YANG model change is also required to support this new action.

devpatha:
@bingwang-ms shall I update yang to add PACKET_ACTION enumeration of COPY? https://github.com/sonic-net/SONiC/blob/master/sonic-acl.yang#L138
AFAIK if a platform doesn't have support for SAI_PACKET_ACTION_COPY, the ACL rule will remain inactive?

bingwang-ms:
> @bingwang-ms shall I update yang to add PACKET_ACTION enumeration of COPY? https://github.com/sonic-net/SONiC/blob/master/sonic-acl.yang#L138 AFAIK if a platform doesn't have support for SAI_PACKET_ACTION_COPY, the ACL rule will remain inactive?

Yes, the YANG model needs to be updated. 
If the new action is not supported, we need to ensure orchagent/syncd won't crash.

PR #3289, opened 2024-09-14
[POC] test build VS image for pops with 1024

liuh-80:
please ignore draft PR.


orchagent/orch.cpp
```diff
@@ -13,6 +13,8 @@
 #include "zmqconsumerstatetable.h"
 #include "sai_serialize.h"
 
+#define CONSUMER_POP_MAX_BATCH_COUNT 2
```
liuh-80, line 16: Verify the test case failed because not pop twice.

PR #3297, opened 2024-09-24
[crm][dash] Do not probe DASH resources on devices other than the DPU.

oleksandrivantsiv:
**What I did**
Probe the DASH resource only when run on the DPU.  

**Why I did it**
The changes optimize the CRM Orch to not access the DASH resources when it is clear that the device running the SONiC doesn't support them.
Also this prevents potential SAI library incompatibility caused by SAI metadata changes introduced in https://github.com/opencomputeproject/SAI/pull/2028 PR

**How I verified it**
Run CRM UT.

**Details if related**


dgsudharsan (approved):

liat-grozovik:
@prsunny  this is urgent fix for master stability. can you please help to approve and merge?

dgsudharsan (approved):

PR #3301, opened 2024-09-25
add orchagent for debug generate dump

aviramd:
This update adds support for the sai_dbg_gen_dump API, enabling the generation of a debug dump file by the SAI when there is no SAI failure. For scenarios involving SAI failures, a separate flow called dump_on_sai_failure is available.

An HLD will be publish soon with all relevant PRs

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


prsunny:
Please link HLD and also associated tests. Suggest move this to draft until then

PR #3305, opened 2024-09-29
Fix the issue where the test case randomly fails due to not waiting after updating the config DB.

liuh-80:
Fix the issue where the test case randomly fails due to not waiting after updating the config DB.

#### Why I did it
Some test case randomly fails due to not waiting after updating the config DB.

#### How I did it
Add few seconds wait after update config DB.

#### How to verify it
Pass all UT.
Manually verify issue fixed.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

Fix the issue where the test case randomly fails due to not waiting after updating the config DB.

### DIFF
--- a/tests/dash/test_dash_acl.py
+++ b/tests/dash/test_dash_acl.py
@@ -87,10 +87,12 @@ class ProduceStateTable(object):
             pairs_str.append((to_string(k), to_string(v)))
         self.table.set(key, pairs_str)
         self.keys.add(key)
+        time.sleep(3)
 
     def __delitem__(self, key: str):
         self.table.delete(str(key))
         self.keys.discard(key)
+        time.sleep(3)
 
     def get_keys(self):
         return self.keys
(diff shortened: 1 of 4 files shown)


matiAlfaro (approved):

tests/dash/test_dash_acl.py
```diff
@@ -87,10 +87,12 @@ def __setitem__(self, key: str, pairs: Union[dict, list, tuple]):
             pairs_str.append((to_string(k), to_string(v)))
         self.table.set(key, pairs_str)
         self.keys.add(key)
+        time.sleep(3)
```
qiluo-msft, line 90: >sleep [](http://example.com/codeflow?start=13&length=5)

Since it is in the lowest level, it may dramatically slow down the testrun, assuming there will lots of set operations.

liuh-80: Improve by change sleep 1second.
I checked the code, there will be 5-10 set/delete operation for every test case.

qiluo-msft (approved):

liuh-80:
202405 cherry-pick PR created: https://github.com/sonic-net/sonic-swss/pull/3323

PR #3307, opened 2024-09-29
Added support for "UNDERLAY_SET_DSCP" and "UNDERLAY_SET_DSCPV6" tables

siqbal1986:
**What I did**
This PR adds the logical table UNDERLAY_SET_DSCP and UNDERLAY_SET_DSCPV6.
This feature allows SONIC to match an ingress L4 packet and change the DSCP field of the outer header when the packet is egressing.
These tables are only created on Cisco-8000, Mlnx and VS platforms. 
The PR also adds support to check if the platform supports the following SAI attributes.
SAI_SWITCH_ATTR_ACL_USER_META_DATA_RANGE
SAI_ACL_ENTRY_ATTR_ACTION_SET_ACL_META_DATA
SAI_ACL_ENTRY_ATTR_FIELD_ACL_USER_META
Verified on Mlnx 2700
image
**Why I did it**
The match set for these tables are following. The action for both of these tables is SET_DSCP
UNDERLAY_SET_DSCP 
- SRC_IP
- DST_IP
- IP_PROTOCOL
- L4_SRC_PORT
- L4_DST_PORT
- TCP_FLAGS
- DSCP

UNDERLAY_SET_DSCPV6
- SRC_IPV6
- DST_IPV6
- IPV6_NEXT_HEADER
- L4_SRC_PORT
- L4_DST_PORT
- DSCP
Note: These tables are not created but only their names are used.

Since we require matching based on L4 parameters, all the mentioned match attributes are necessary. Merging these into a single table would result in a larger TCAM footprint which may be impact existing ACL usage scenarios.

These table names translate into MARK_META and MARK_METAV6 and EGR_SET_DSCP tables.
The translation is such that an attempt to create UNDERLAY_SET_DSCP or UNDERLAY_SET_DSCPV6 results into creation of
MARK_META & EGR_SET_DSCP  or MARK_METAV6 & EGR_SET_DSCP .
The EGR_SET_DSCP table is shared and created only once. if both UNDERLAY_SET_DSCP and UNDERLAY_SET_DSCPV6 are created then only one instance of EGR_SET_DSCP is created to save tcam resource. The EGR_SET_DSCP is created in the Egress stage.

These internal MARK_META/V6 and EGR_SET_DSCP tables use SAI_ACL_ENTRY_ATTR_ACTION_SET_ACL_META_DATA and SAI_ACL_ENTRY_ATTR_FIELD_ACL_USER_META to set and match the metaData field.
For Example:
if the intent is to match an ingressing packet with fields 
"DST_IP": "20.0.0.1/32",
"SRC_IP": "10.0.0.0/32",
"DSCP": "1"
"SRC_PORT" : "10"
"DST_PORT" : "20"
and set its outer DSCP to 12 after encapsulation,
This would be done so by MARK_META entry in ingress tcam matching based on above mentioned critera and setting a metaData value e.g. "1"
Another EGR_SET_DSCP entry in egress stage would be created with match criteria of metaData=1 and action of SET_DSCP=12.

Each entry in the EGR_SET_DSCP table is refcounted and shared among all the rules interested in same DSCP value.
There are 7 metadata values available [1...7]. This is currently hardcoded but this can be enhanced based on capibility check.

A Metadata Manager class governs the allcoation and freeing of each metadata value and they are shared based on DSCP value.
**How I verified it**
Added 16 Tests in sonic-mgmt. All passing.
image

Manual Test:
Tests added for the tables and Metadata Manager. 9 Test cases have been added which verfiy the table creation, deletion, EGR_SET_DSCP referencing, Entry creation/deletion, metadata value allcation/ release and exhausion scenarios.

**Details if related**
tested on mlnx 2700 with following rule
{   "ACL_RULE": {
        "OVERLAY_MARK_META_TEST|RULE0": {
            "PRIORITY": "999",
            "DSCP_ACTION": "40",
            "SRC_IP": "1.1.1.1/32"
        },
        "OVERLAY_MARK_META_TEST|RULE1": {
            "PRIORITY": "998",
            "DSCP_ACTION": "42",
            "SRC_IP": "2.2.2.2/32"
        }
} and a vxlan tunnel
dscp value change

reviewed HLD.https://github.com/sonic-net/SONiC/pull/1743


orchagent/aclorch.cpp
```diff
@@ -3190,6 +3368,93 @@ void AclOrch::init(vector<TableConnector>& connectors, PortsOrch *portOrch, Mirr
...
+
+        sai_status_t status = SAI_STATUS_SUCCESS;
+        sai_attr_capability_t capability;
+        uint16_t metadataMin = 0;
+        uint16_t metadataMax = 0;
+        m_switchMetaDataCapabilities[TABLE_ACL_USER_META_DATA_MIN] = "0";
+        m_switchMetaDataCapabilities[TABLE_ACL_USER_META_DATA_MAX] = "0";
+        m_switchMetaDataCapabilities[TABLE_ACL_USER_META_DATA_RANGE_CAPABLE] = "false";
+        m_switchMetaDataCapabilities[TABLE_ACL_ENTRY_ATTR_META_CAPABLE] = "false";
+            m_switchMetaDataCapabilities[TABLE_ACL_ENTRY_ACTION_META_CAPABLE] = "false";
```
bingwang-ms, line 3396: Please fix the alignment

orchagent/aclorch.cpp
```diff
@@ -3190,6 +3368,93 @@ void AclOrch::init(vector<TableConnector>& connectors, PortsOrch *portOrch, Mirr
...
+        {
+            SWSS_LOG_WARN("Could not query ACL_USER_META_DATA_RANGE %d", status);
+        }
+        else
+        {
+            if (capability.get_implemented)
+            {
+                sai_attribute_t attrs[1];
+                attrs[0].id = SAI_SWITCH_ATTR_ACL_USER_META_DATA_RANGE;
+                sai_status_t status = sai_switch_api->get_switch_attribute(gSwitchId, 2, attrs);
```
bingwang-ms, line 3409: Should it be `get_switch_attribute(gSwitchId, 1, attrs)` ? The sz of attrs is 1

orchagent/aclorch.cpp
```diff
@@ -3903,10 +4228,194 @@ bool AclOrch::updateAclTable(AclTable &currentTable, AclTable &newTable)
...
+                SWSS_LOG_ERROR("Failed to create ACL table EgressSetDSCP");
+                // since we failed to create the table, there is no need for rollback.
+                return EgressSetDscpTableStatus::EGRESS_SET_DSCP_TABLE_FAILED;
+            }
+        }
+        else
+        {
+            if (updateAclTable(m_AclTables[egrSetDscp_oid], egrSetDscpTable))
+            {
+                SWSS_LOG_INFO("Successfully updated existing ACL table EgressSetDSCP");
```
bingwang-ms, line 4303: Suggest changing log level to NOTICE.

orchagent/aclorch.cpp
```diff
@@ -3903,10 +4228,194 @@ bool AclOrch::updateAclTable(AclTable &currentTable, AclTable &newTable)
...
+                SWSS_LOG_ERROR("Failed to update existing ACL table EgressSetDSCP");
+                //  there is no need for roollback as we have not made any changes to the MARK_META/V6 tables.
+                // We can simply return false.
+                return EgressSetDscpTableStatus::EGRESS_SET_DSCP_TABLE_FAILED;
+            }
+        }
+        // keep track of the fact that this table is now associated with the EGR_SET_DSCP table.
+
+        m_egrSetDscpRef.insert(table_id);
+        SWSS_LOG_INFO("Added ACL table %s to EgrSetDscpRef", table_id.c_str());
```
bingwang-ms, line 4318: Suggest changing log level to NOTICE.

orchagent/aclorch.cpp
```diff
@@ -3903,10 +4228,194 @@ bool AclOrch::updateAclTable(AclTable &currentTable, AclTable &newTable)
...
+{
+    m_egrSetDscpRef.erase(table_id);
+    SWSS_LOG_INFO("attempting to remove %s reference", table_id.c_str());
+    if (m_egrSetDscpRef.size() == 0)
+    {
+        if (!removeAclTable(EGR_SET_DSCP_TABLE_ID))
+        {
+            SWSS_LOG_INFO("Failed to remove the %s table so readding the reference of %s", EGR_SET_DSCP_TABLE_ID, table_id.c_str());
+            m_egrSetDscpRef.insert(table_id);
+            SWSS_LOG_ERROR("Failed to remove ACL table %s", EGR_SET_DSCP_TABLE_ID);
```
bingwang-ms, line 4336: Duplicated log

orchagent/aclorch.cpp
```diff
@@ -3903,10 +4228,194 @@ bool AclOrch::updateAclTable(AclTable &currentTable, AclTable &newTable)
...
+        AclTable dummyTable(this);
+        dummyTable.id = EGR_SET_DSCP_TABLE_ID;
+        dummyTable.stage = ACL_STAGE_EGRESS;
+        if (!updateAclTable(EGR_SET_DSCP_TABLE_ID, dummyTable, ""))
+        {
+            m_egrSetDscpRef.insert(table_id);
+            SWSS_LOG_ERROR("Failed to remove ACL table %s", EGR_SET_DSCP_TABLE_ID);
+            return false;
+        }
+        SWSS_LOG_INFO("Successfully removed the %s table from the reference of %s", table_id.c_str(), EGR_SET_DSCP_TABLE_ID);
```
bingwang-ms, line 4365: Please change log level to NOTICE

orchagent/aclorch.cpp
```diff
@@ -3903,10 +4228,194 @@ bool AclOrch::updateAclTable(AclTable &currentTable, AclTable &newTable)
...
+bool AclOrch::removeEgrSetDscpTable(string table_id)
+{
+    m_egrSetDscpRef.erase(table_id);
+    SWSS_LOG_INFO("attempting to remove %s reference", table_id.c_str());
+    if (m_egrSetDscpRef.size() == 0)
+    {
+        if (!removeAclTable(EGR_SET_DSCP_TABLE_ID))
+        {
+            SWSS_LOG_INFO("Failed to remove the %s table so readding the reference of %s", EGR_SET_DSCP_TABLE_ID, table_id.c_str());
+            m_egrSetDscpRef.insert(table_id);
```
bingwang-ms, line 4335: Suggest removing the table from `m_egrSetDscpRef` if the removal is successful so that you don't need to add it back in failing case.

orchagent/aclorch.cpp
```diff
@@ -4583,7 +5307,7 @@ void AclOrch::doAclTableTask(Consumer &consumer)
         }
         else if (op == DEL_COMMAND)
         {
-            if (removeAclTable(table_id))
+            if (removeAclTableWithEgrDscp(table_id))
```
bingwang-ms, line 5310: Suggest keeping the original function name and add the check into `removeAclTable`. Otherwise people may get confused when to use `removeAclTable` and when to use `removeAclTableWithEgrDscp`

orchagent/aclorch.cpp
```diff
@@ -4741,18 +5465,18 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
...
+			        SWSS_LOG_ERROR("Rule '%s' is invalid since it has both v4 and v6 matchfields.", rule_id.c_str());
+			        bAllAttributesOk = false;
+		        }
+	        }
 
             // validate and create ACL rule
             if (bAllAttributesOk && newRule->validate())
             {
-                if (addAclRule(newRule, table_id))
+                if (addAclRuleWithEgrSetDscp(newRule, table_id))
```
bingwang-ms, line 5479: Same here. Suggest keeping the original function name

bingwang-ms:
Creating the egress table implicitly when creating the ingress table still looks tricky to me. That also makes code logic more complicated. I still suggest separating them.

orchagent/aclorch.cpp
siqbal1986: done

orchagent/aclorch.cpp
siqbal1986: done

orchagent/aclorch.cpp
siqbal1986: done

orchagent/aclorch.cpp
siqbal1986: the m_egrSetDscpRef is an internal structure which keeps track of the active tables which are using EGR_SET_DSCP_TABLE. we also use it during the port set calcualtions. when removing a table it has to be derefed form this dictionary first bacause we use this to find out what other tables are referencing the EGR_SET_DSCP_TABLE The only other solution is to create a copy, remove the table from the copy and send it down. I think that would overcomplecate the solution.

orchagent/aclorch.cpp
siqbal1986: i can remove it.

orchagent/aclorch.cpp
siqbal1986: the removeAclTable is the existing function. the removeAclTableWithEgrDscp enhances the functionality without touching the orignal well tested code. I think its better to keep them segrigated. this way when someone is making changes to this code they can easily figure out that removeAclTableWithEgrDscp was added as a separate effort.

orchagent/aclorch.cpp
siqbal1986: done

orchagent/aclorch.cpp
siqbal1986: again, I would prefer that the reader can clearly distingush that addAclRuleWithEgrSetDscp handles a special case and addAclRule is a general rule handler.

bingwang-ms (dismissed):

prsunny:
Can you link the HLD to PR?

orchagent/aclorch.cpp
```diff
@@ -4207,6 +4819,19 @@ bool AclOrch::removeAclRule(string table_id, string rule_id)
     return m_AclTables[table_oid].remove(rule_id);
 }
 
+bool AclOrch::removeAclRuleWithEgrSetDscp(string table_id, string rule_id)
```
prsunny, line 4822: same here as add. Keep the logic in removeAclRule

orchagent/aclorch.cpp
```diff
@@ -2183,6 +2262,105 @@ void AclRuleMirror::onUpdate(SubjectType type, void *cntx)
...
+    actionData.parameter.u32 = 0;
+
+    SWSS_LOG_INFO("attr_name: %s, attr_value: %s int val %d", attr_name.c_str(), attr_value.c_str(), to_uint<uint32_t>(attr_value));
+    // we only handle DSCP and META_DATA actions for now.
+    if (attr_name == ACTION_DSCP || attr_name == ACTION_META_DATA)
+    {
+        actionData.parameter.u32 = to_uint<uint32_t>(attr_value);
+        if (attr_name == ACTION_META_DATA && (actionData.parameter.u32 < m_pAclOrch->getAclMetaDataMin() || actionData.parameter.u32 > m_pAclOrch->getAclMetaDataMax()))
+        {
+            return false;
```
kperumalbfn, line 2336: For any failures, should we decrement the Metadata Ref count??

orchagent/aclorch.cpp
```diff
@@ -4182,6 +4766,34 @@ bool AclOrch::addAclRule(shared_ptr<AclRule> newRule, string table_id)
...
+        string dscpAction = std::to_string(std::static_pointer_cast<AclRuleUnderlaySetDscp>(newRule)->getDscpValue());
+        if (!addEgrSetDscpRule(key, dscpAction))
+        {
+            SWSS_LOG_ERROR("Failed to add Egress Set Dscp rule for Rule %s in table %s.",
+                newRule->getId().c_str(), table_id.c_str());
+            return false;
+        }
+    }
+    // add the regular rule.
+    bool status = addAclRule(newRule, table_id);
```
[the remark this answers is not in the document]
siqbal1986: done

orchagent/aclorch.cpp
siqbal1986: and done

orchagent/aclorch.cpp
```diff
@@ -1052,6 +1115,18 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
                 return false;
             }
         }
+        else if (attr_name == MATCH_METADATA)
+        {
+            matchData.data.u16 = to_uint<uint16_t>(attr_value);
+            matchData.mask.u16 = 0xFF;
+
```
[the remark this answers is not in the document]
siqbal1986: i am aware this field comes in as u32 but one platform supports 4 bits and other supports 12 bits for the ACL META. thats why i chose to use a u16. but i get your point, i dont anticipate any platform to support a 32 bit field here given how TCAM entries are very congested in the first place, But the to be on the safer side, ill change it do u32. just dont like to waste space :)

orchagent/aclorch.cpp
```diff
@@ -3458,7 +3723,43 @@ void AclOrch::initDefaultTableTypes(const string& platform, const string& sub_pl
                 .build()
         );
     }
+    if (isAclMetaDataSupported())
+    {
+        addAclTableType(
+            builder.withName(TABLE_TYPE_MARK_META)
+                .withBindPointType(SAI_ACL_BIND_POINT_TYPE_PORT)
+                .withBindPointType(SAI_ACL_BIND_POINT_TYPE_LAG)
```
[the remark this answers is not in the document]
siqbal1986: both port and port channel. for our use case its mostly portchannels.

orchagent/aclorch.cpp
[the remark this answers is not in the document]
siqbal1986: excellent point. ill address this

orchagent/aclorch.cpp
```diff
@@ -3458,7 +3723,43 @@ void AclOrch::initDefaultTableTypes(const string& platform, const string& sub_pl
...
 
+        addAclTableType(
+            builder.withName(TABLE_TYPE_MARK_META_V6)
+                .withBindPointType(SAI_ACL_BIND_POINT_TYPE_PORT)
+                .withBindPointType(SAI_ACL_BIND_POINT_TYPE_LAG)
+                .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_SRC_IPV6))
+                .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_DST_IPV6))
+                .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_IPV6_NEXT_HEADER))
+                .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_L4_SRC_PORT))
+                .withMatch(make_shared<AclTableMatch>(SAI_ACL_TABLE_ATTR_FIELD_L4_DST_PORT))
```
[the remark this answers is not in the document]
siqbal1986: excellent catch. fixed

orchagent/aclorch.cpp
siqbal1986: I do the deref on removal.failure results in failure status, when the rule is removed I deref the metadata as well. This check serves as validator and safety net. We hit this when we have exhausted the metadata values or when we have an invalid metadata value. so we cannot proceed.

orchagent/aclorch.cpp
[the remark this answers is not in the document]
siqbal1986: done

orchagent/aclorch.cpp
bingwang-ms: Reopen this comment.
Suggest using the original function name and add the logic inside `removeAclTable` so that other maintainers can be clearer about which function to be used.

siqbal1986: done

orchagent/aclorch.cpp
```diff
@@ -1057,6 +1120,18 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
                 return false;
             }
         }
+        else if (attr_name == MATCH_METADATA)
+        {
+            matchData.data.u32 = to_uint<uint32_t>(attr_value);
+            matchData.mask.u32 = 0xFFFFFFFF;
+
+            if (matchData.data.u16 < m_pAclOrch->getAclMetaDataMin() || matchData.data.u16 > m_pAclOrch->getAclMetaDataMax())
```
kperumalbfn, line 1128: u32?

orchagent/aclorch.cpp
```diff
@@ -3195,6 +3373,93 @@ void AclOrch::init(vector<TableConnector>& connectors, PortsOrch *portOrch, Mirr
...
+                }
+                else
+                {
+                    SWSS_LOG_NOTICE("ACL_USER_META_DATA_RANGE, min: %u, max: %u", attrs[0].value.u32range.min, attrs[0].value.u32range.max);
+                    m_switchMetaDataCapabilities[TABLE_ACL_USER_META_DATA_RANGE_CAPABLE] = "true";
+                    m_switchMetaDataCapabilities[TABLE_ACL_USER_META_DATA_MIN] = std::to_string(attrs[0].value.u32range.min);
+                    m_switchMetaDataCapabilities[TABLE_ACL_USER_META_DATA_MAX] = std::to_string(attrs[0].value.u32range.max);
+                    metadataMin = uint16_t(attrs[0].value.u32range.min);
+                    metadataMax = uint16_t(attrs[0].value.u32range.max);
+                }
```
kperumalbfn, line 3427: Do we need typecast here?

orchagent/aclorch.cpp
```diff
@@ -3195,6 +3373,93 @@ void AclOrch::init(vector<TableConnector>& connectors, PortsOrch *portOrch, Mirr
...
+        sai_attr_capability_t capability;
+        uint16_t metadataMin = 0;
+        uint16_t metadataMax = 0;
+        m_switchMetaDataCapabilities[TABLE_ACL_USER_META_DATA_MIN] = "0";
+        m_switchMetaDataCapabilities[TABLE_ACL_USER_META_DATA_MAX] = "0";
+        m_switchMetaDataCapabilities[TABLE_ACL_USER_META_DATA_RANGE_CAPABLE] = "false";
+        m_switchMetaDataCapabilities[TABLE_ACL_ENTRY_ATTR_META_CAPABLE] = "false";
+        m_switchMetaDataCapabilities[TABLE_ACL_ENTRY_ACTION_META_CAPABLE] = "false";
+
+        status = sai_query_attribute_capability(gSwitchId, SAI_OBJECT_TYPE_SWITCH, SAI_SWITCH_ATTR_ACL_USER_META_DATA_RANGE, &capability);
```
kperumalbfn, line 3403: SImilar to validating USER_META_DATA capability, do we need to check ACL action set_dscp capability?

orchagent/aclorch.cpp
```diff
@@ -4746,13 +5447,13 @@ void AclOrch::doAclRuleTask(Consumer &consumer)
...
 
             if (bHasIPV4 && bHasIPV6)
-	    {
-		    if (type == TABLE_TYPE_L3V4V6)
-		    {
-			    SWSS_LOG_ERROR("Rule '%s' is invalid since it has both v4 and v6 matchfields.", rule_id.c_str());
-			    bAllAttributesOk = false;
-		    }
-	    }
+	        {
```
kperumalbfn, line 5450: pls check indentation here

orchagent/aclorch.h
```diff
@@ -168,6 +185,24 @@ class AclTableRangeMatch: public AclTableMatchInterface
...
+    uint16_t getFreeMetaData(uint8_t dscp);
+    void recycleMetaData(uint16_t metadata);
+    bool isValidMetaData(uint16_t metadata);
+
+private:
+    bool initComplete = false;
+    uint16_t metaMin = 0;
+    uint16_t metaMax = 0;
+    list<uint16_t> m_freeMetadata;
+    map<uint8_t, uint16_t> m_dscpMetadata;
```
kperumalbfn, line 202: pls check for uint32 for all references to metadata.

orchagent/aclorch.cpp
siqbal1986: the metadata values supproted by platforms are  1-7 , and 1-4095 bit. I dont think any platform would support a 32 bit for this field or any tcam would have 4 billion enties.

orchagent/aclorch.cpp
siqbal1986: that is already present on all platforms as we have tables using this capability.

orchagent/aclorch.cpp
siqbal1986: tabs

orchagent/aclorch.h
siqbal1986: i have limited the metadata to be between 0-4095. beyond this range, i double we ll have a platform which would allow 4k acl entries in a group.

orchagent/aclorch.cpp
siqbal1986: i have changed this code to restrict the range to 0-4095

orchagent/aclorch.cpp
siqbal1986: done

bingwang-ms (approved):
LGTM now.

kperumalbfn (approved):
LGTM

PR #3309, opened 2024-10-01
[subport] Allow to add Subport without vlan argument

gord1306:
**What I did**
Allow to add Subport without vlan argument

**Why I did it**
The alias of subport already contains the VLAN ID information. If no vlan be specified in the input argument, it would use the vlan in the alias as the vlan id

**How I verified it**
Use swssconfig to generate config to make sure the subport be created

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1449,13 +1449,21 @@ bool PortsOrch::addSubPort(Port &port, const string &alias, const string &vlan, 
     subIntf subIf(alias);
     string parentAlias = subIf.parentIntf();
     sai_vlan_id_t vlan_id;
+
     try
     {
-        vlan_id = static_cast<sai_vlan_id_t>(stoul(vlan));
+        if (vlan.empty())
+        {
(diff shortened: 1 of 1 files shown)


prsunny (commented):
Change lgtm, but please add tests.

dgsudharsan (changes requested):
@gord1306  There are two schemas to add subport. Short and long.
1) Long format. Name contains interface.vlan_id (Where vlan id is derived from the name) https://github.com/sonic-net/sonic-swss/blob/70ade3433200c7146b398cbe5908a4c5a3ef337b/cfgmgr/intfmgr.cpp#L753C13-L753C19
The vlan id is populated here and pushed to orchagent.
2) Name contains subinterface id. Vlan is derived from kv pair so by default pushed to orchagent.
https://github.com/sonic-net/SONiC/blob/master/doc/subport/sonic-sub-port-intf-hld.md

Then why do we need this change? Are we expecting to directly program app_db? If we are using config_db I don't think we need this change

PR #3314, opened 2024-10-03
Fix to enable/disable Lag member interface quickly when macsec session is up/down

judyjoseph:
**What I did**

Set the Lag member state to enable/disable when the macsec session on the lag member interface go up/down.
Fixes isue : https://github.com/sonic-net/sonic-buildimage/issues/19254

**Why I did it**

Usually when macsec session goes down, since macsec is point to point -- the interface also goes down, and it is removed from LAG.

But in case the interface don't go down, and it is just macsec session that went down -- it waits for 90secs which is LACP timeout to get the interface out of LAG.

With this fix -- macsec session timesout in 6sec .. and we immediately remove it out ot LAG, to minimize traffic impact.

**How I verified it**

     <In progress>

**Details if related**


abdosi:
@judyjoseph are we ready to review this ?

rlhui:
> @judyjoseph are we ready to review this ?

@judyjoseph  ping again? thx

PR #3326, opened 2024-10-14
[FC] process FC after apply view

stepanblyschak:
**What I did**

Simplify approach to delaying counters on warm boot and fast boot. Removed FLEX_COUNTER_DELAY_STATUS_FIELD and instead postpone all FC processing to happen after apply view to not delay data plane configuration.

The CONFIG_DB should not be updated in runtime anymore for counters to be delayed.

**Why I did it**

To address https://github.com/sonic-net/sonic-buildimage/issues/20302.

**How I verified it**

Run warm-boot - make sure FC orch runs only after APPLY_VIEW.

**Details if related**


Junchao-Mellanox (approved):

stepanblyschak:
Putting back delay for 60 sec as we found some cases where oper state update handling is delayed due to FC configuration after APPLY_VIEW

orchagent/flexcounterorch.cpp
```diff
@@ -66,12 +69,22 @@ unordered_map<string, string> flexCounterGroupMap =
...
 FlexCounterOrch::FlexCounterOrch(DBConnector *db, vector<string> &tableNames):
     Orch(db, tableNames),
-    m_flexCounterConfigTable(db, CFG_FLEX_COUNTER_TABLE_NAME),
     m_bufferQueueConfigTable(db, CFG_BUFFER_QUEUE_TABLE_NAME),
     m_bufferPgConfigTable(db, CFG_BUFFER_PG_TABLE_NAME),
     m_deviceMetadataConfigTable(db, CFG_DEVICE_METADATA_TABLE_NAME)
 {
     SWSS_LOG_ENTER();
+    m_delayTimer = new SelectableTimer(timespec{.tv_sec = FLEX_COUNTER_DELAY_SEC, .tv_nsec = 0});
+    if (WarmStart::isWarmStart())
```
Junchao-Mellanox, line 78: Just to confirm this will also handle fast-reboot.

orchagent/flexcounterorch.cpp
```diff
@@ -254,6 +258,15 @@ void FlexCounterOrch::doTask(Consumer &consumer)
     }
 }
 
+void FlexCounterOrch::doTask(SelectableTimer &timer)
+{
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_NOTICE("Processing counters");
+    m_delayTimer->stop();
```
Junchao-Mellanox, line 266: How about following?
```
if (!m_delayTimerExpired)
{
    m_delayTimer->stop();
    m_delayTimerExpired = true;
}
```

orchagent/flexcounterorch.cpp
```diff
@@ -254,6 +258,15 @@ void FlexCounterOrch::doTask(Consumer &consumer)
     }
 }
 
+void FlexCounterOrch::doTask(SelectableTimer &timer)
```
Junchao-Mellanox, line 261: should we delete the timer here?

orchagent/flexcounterorch.cpp
```diff
@@ -66,12 +69,22 @@ unordered_map<string, string> flexCounterGroupMap =
...
+    m_delayTimer = new SelectableTimer(timespec{.tv_sec = FLEX_COUNTER_DELAY_SEC, .tv_nsec = 0});
+    if (WarmStart::isWarmStart())
+    {
+        auto executor = new ExecutableTimer(m_delayTimer, this, "FLEX_COUNTER_DELAY");
+        Orch::addExecutor(executor);
+        m_delayTimer->start();
+    }
+    else
+    {
+        m_delayTimerExpired = true;
```
Junchao-Mellanox, line 86: We used to delay counter polling for all reboot types, and now it turns that counter polling will start right after APPLY_VIEW for cold reboot. Although we don't have a hard requirement for cold reboot time, there is still a question: could you please share some data about how this change would affect cold reboot time?

orchagent/flexcounterorch.cpp
stepanblyschak: Yes

orchagent/flexcounterorch.cpp
stepanblyschak: Mainly due to VS tests expecting FC counters to appear immediately, which would fail without putting some delays. Therefore I limited this change to fast/warm reboot only. I am looking for a way to not delay counters only for VS, then I can have this delay for all reboot types.

orchagent/flexcounterorch.cpp
stepanblyschak: Done

orchagent/flexcounterorch.cpp
stepanblyschak: Done

bingwang-ms:
@wen587 Can you help review?

orchagent/flexcounterorch.cpp
bingwang-ms: @stepanblyschak Since the change is limited to warm-reboot now, may I know how much time can be saved in warm-reboot after delaying flex counter?

stepanblyschak: @bingwang-ms No time savings, the delay was limited to fast and warm boot already. No delay if FC are present in CONFIG_DB already in cold boot.

orchagent/flexcounterorch.cpp
```diff
@@ -69,12 +72,22 @@ unordered_map<string, string> flexCounterGroupMap =
 
 FlexCounterOrch::FlexCounterOrch(DBConnector *db, vector<string> &tableNames):
     Orch(db, tableNames),
-    m_flexCounterConfigTable(db, CFG_FLEX_COUNTER_TABLE_NAME),
     m_bufferQueueConfigTable(db, CFG_BUFFER_QUEUE_TABLE_NAME),
     m_bufferPgConfigTable(db, CFG_BUFFER_PG_TABLE_NAME),
     m_deviceMetadataConfigTable(db, CFG_DEVICE_METADATA_TABLE_NAME)
 {
     SWSS_LOG_ENTER();
+    m_delayTimer = new SelectableTimer(timespec{.tv_sec = FLEX_COUNTER_DELAY_SEC, .tv_nsec = 0});
```
qiluo-msft, line 80: >new [](http://example.com/codeflow?start=19&length=3)

where do you delete it? in destructor?

qiluo-msft: will you use smart pointer?

stepanblyschak: @qiluo-msft I am ok using smart pointer. Did not choose to in this situation due to the lifetime of FlexCounterOrch and therefore m_delayTimer is the entire duration of the program, therefore no resource leak. Looking at other usages it seems to me that we aren't really calling destructor in other places as well, e.g. https://github.com/sonic-net/sonic-swss/blob/master/orchagent/portsorch.cpp#L538

bingwang-ms:
@qiluo-msft Can you please help review? Thanks

orchagent/flexcounterorch.cpp
```diff
@@ -69,12 +72,22 @@ unordered_map<string, string> flexCounterGroupMap =
...
+    if (WarmStart::isWarmStart())
+    {
+        auto executor = new ExecutableTimer(m_delayTimer, this, "FLEX_COUNTER_DELAY");
```
vaibhavhd, line 83: I am not convinced with this approach. I can use some to understand the approach.

Previously we had hardcoded delay in FC.
The hardcoded delay became a problem and we needed more deterministic delayed approach. So we added fast-reboot new infra which used warm-reboot kind of reconciliation and used fast-reboot done status to enable FC.

But now we again go back to hardcoded delay. I am afraid that to solve the new problem we (in a way) drift towards the day1 problem. Some scenarios that I can image that will now become problematic can be when 60s not being enough and overlaps w/ apply_view. Or similar race conditions that led to more deterministic delay.

Also, I think the concerns that Jingwen raised about new entries showing up in the table will still be present, right?
The config will still look different than what init config would want.

@wen587 , what do you think about this? Does this change really solve the concern that you have raised?

stepanblyschak: @vaibhavhd This is the problem that is being solved https://github.com/sonic-net/sonic-buildimage/issues/20302
The hardcoded delay is not a reported problem and this PR does not attempt to address that.
Not sure about race conditions, could you provide an example? The timer can't overlap with apply view as it is triggered in a different flow.

wen587: Hi @vaibhavhd , i think so. We want the config just stay as it is in init_cfg.json. We do not want any derived changes after service restart. Because all config db are static in Golden Config. We need Golden Config to be ground of truth.

orchagent/flexcounterorch.cpp
qiluo-msft: Just to clarify: we need to ensure no memory leak. If you want to improve "other usages" (in another PR), it is welcome. Smart pointer is just one easy solution to fix memory leak, but you can also choose other options.

qiluo-msft (approved):

stepanblyschak:
Infra issue with tests: failure to install dotnet package:

```
+ sudo apt-get install -y dotnet-sdk-8.0
Reading package lists...
Building dependency tree...
Reading state information...
E: The package keysinuse needs to be reinstalled, but I can't find an archive for it.
```

Restarting

orchagent/flexcounterorch.cpp
```diff
@@ -69,12 +72,22 @@ unordered_map<string, string> flexCounterGroupMap =
 
 FlexCounterOrch::FlexCounterOrch(DBConnector *db, vector<string> &tableNames):
     Orch(db, tableNames),
-    m_flexCounterConfigTable(db, CFG_FLEX_COUNTER_TABLE_NAME),
     m_bufferQueueConfigTable(db, CFG_BUFFER_QUEUE_TABLE_NAME),
     m_bufferPgConfigTable(db, CFG_BUFFER_PG_TABLE_NAME),
     m_deviceMetadataConfigTable(db, CFG_DEVICE_METADATA_TABLE_NAME)
 {
     SWSS_LOG_ENTER();
+    m_delayTimer = std::make_unique<SelectableTimer>(timespec{.tv_sec = FLEX_COUNTER_DELAY_SEC, .tv_nsec = 0});
```
wen587, line 80: @vaibhavhd , could you review the warm boot part?

liat-grozovik:
@wen587  @vaibhavhd  kindly reminder. this is needed to get in as other PRs are based on that. it is crucial for 202411. @bingwang-ms  please help to track it as well please.

vaibhavhd (approved):

bingwang-ms (approved):

PR #3335, opened 2024-10-20
BFD Tx, Rx interval support for Vnet Monitored routes.

siqbal1986:
**What I did**
This change add support for
 **rx_monitor_timer
tx_monitor_timer
check_directly_connected**
**Why I did it**
This change is beneficial because it adds support for BFD Tx, Rx interval for Vnet Monitored routes as required by DASH project.
**How I verified it**
Added a unit test to verify the functionality.
**Details if related**
The Tx and Rx values are added on per-route basis.  Therefore the same values are used for each nexthop. if the user needs to se different values for each next hop then they can add the values and nexthops incrementally.
e.g.
```
Add route with nexthop 10.0.0.1(1s,1s) -> BFD(10.0.0.1) with Tx =1s Rx= 1s
Update same route and add another nexthop 10.0.0.2(2s,2s) -> BFD(10.0.0.1) with Tx =1s Rx= 1s |  BFD(10.0.0.2) with Tx =2s Rx= 2s
```
The user can also update these values by selectively removing and readding a nexthop with different Tx and Rx values.

### DIFF
--- a/orchagent/vnetorch.cpp
+++ b/orchagent/vnetorch.cpp
@@ -1080,6 +1080,7 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
                                                NextHopGroupKey& nexthops, string& op, string& profile,
                                                const string& monitoring, NextHopGroupKey& nexthops_secondary,
                                                const IpPrefix& adv_prefix,
+                                               bool check_directly_connected,
                                                const map<NextHopKey, IpAddress>& monitors)
 {
     SWSS_LOG_ENTER();
@@ -1829,7 +1830,8 @@ void VNetRouteOrch::delRoute(const IpPrefix& ipPrefix)
     syncd_routes_.erase(route_itr);
 }
 
-void VNetRouteOrch::createBfdSession(const string& vnet, const NextHopKey& endpoint, const IpAddress& monitor_addr)
+void VNetRouteOrch::createBfdSession(const string& vnet, const NextHopKey& endpoint, const IpAddress& monitor_addr,
+                                     bool has_monitor_interval, u_int32_t tx_monitor_interval, u_int32_t rx_monitor_interval)
 {
     SWSS_LOG_ENTER();
 
@@ -1857,6 +1859,13 @@ void VNetRouteOrch::createBfdSession(const string& vnet, const NextHopKey& endpo
         // when the device goes into TSA.  The following parameter ensures that these session are
         // brought down while transitioning to TSA and brought back up when transitioning to TSB.
         data.emplace_back("shutdown_bfd_during_tsa", "true");
+        if (has_monitor_interval)
+        {
+            FieldValueTuple fvTuple1("tx_interval", to_string(tx_monitor_interval));
+            data.push_back(fvTuple1);
+            FieldValueTuple fvTuple2("rx_interval", to_string(rx_monitor_interval));
+            data.push_back(fvTuple2);
+        }
         bfd_session_producer_.set(key, data);
         bfd_sessions_[monitor_addr].bfd_state = SAI_BFD_SESSION_STATE_DOWN;
     }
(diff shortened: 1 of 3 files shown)


orchagent/vnetorch.cpp
```diff
@@ -1857,6 +1859,13 @@ void VNetRouteOrch::createBfdSession(const string& vnet, const NextHopKey& endpo
         // when the device goes into TSA.  The following parameter ensures that these session are
         // brought down while transitioning to TSA and brought back up when transitioning to TSB.
         data.emplace_back("shutdown_bfd_during_tsa", "true");
+        if (has_monitor_interval)
```
kperumalbfn, line 1862: Please add a note that without tx/rx interval in the DB, BFD session will use the default timers BFD_SESSION_DEFAULT_TX_INTERVAL and BFD_SESSION_DEFAULT_RX_INTERVAL

orchagent/vnetorch.cpp
```diff
@@ -1080,6 +1080,7 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
                                                NextHopGroupKey& nexthops, string& op, string& profile,
                                                const string& monitoring, NextHopGroupKey& nexthops_secondary,
                                                const IpPrefix& adv_prefix,
+                                               bool check_directly_connected,
```
kperumalbfn, line 1083: This check will be added later, right?

orchagent/vnetorch.cpp
```diff
@@ -1857,6 +1859,13 @@ void VNetRouteOrch::createBfdSession(const string& vnet, const NextHopKey& endpo
         // when the device goes into TSA.  The following parameter ensures that these session are
         // brought down while transitioning to TSA and brought back up when transitioning to TSB.
         data.emplace_back("shutdown_bfd_during_tsa", "true");
+        if (has_monitor_interval)
+        {
+            FieldValueTuple fvTuple1("tx_interval", to_string(tx_monitor_interval));
```
kperumalbfn, line 1864: Please include sonic-mgmt tests with Rx/Tx timer to check if these are supported by vendor SAI for BFD hardware offload.

orchagent/vnetorch.cpp
```diff
@@ -2746,9 +2785,32 @@ bool VNetRouteOrch::handleTunnel(const Request& request)
...
         adv_prefix = ip_pfx;
     }
+    if (has_monitor_intervals)
+    {
+        struct VnetRouteMonitorIntervals intervals;
+        intervals.rx_monitor_timer = rx_monitor_timer;
+        intervals.tx_monitor_timer = tx_monitor_timer;
+        prefix_to_monitor_intervals_[ip_pfx] = intervals;
+    }
+    if ( op == DEL_COMMAND)
```
kperumalbfn, line 2795: extra space

orchagent/vnetorch.cpp
siqbal1986: yes

orchagent/vnetorch.cpp
siqbal1986: ill add it.

orchagent/vnetorch.cpp
siqbal1986: ill add those tests after I have made the changes for directly connected nexthops.

orchagent/vnetorch.cpp
```diff
@@ -2620,6 +2641,10 @@ bool VNetRouteOrch::handleTunnel(const Request& request)
     swss::IpPrefix adv_prefix;
     bool has_priority_ep = false;
     bool has_adv_pfx = false;
+    bool has_monitor_intervals = false;
+    uint64_t rx_monitor_timer = 0;
```
prsunny, line 2645: these must be initialized to defaults instead of 0

siqbal1986: The default value is defined in the BFD orch so specifying a default here might result in discrepency if that value is changed. here we rely on has_monitor_interval to make the dicision to use the given values. I can add a comment explaining that.

orchagent/vnetorch.cpp
siqbal1986: fixed

PR #3336, opened 2024-10-21
Remove RIF from m_rifsToAdd before deleting it

tshalvi:
**What I did**
I extended the RIF removal functionality to also remove the port from the m_rifsToAdd list.

**Why I did it**
Typically, the counter and object handling logic follows a strict sequence:

Create an object, then start counter polling.
Stop counter polling, then remove the object.
However, there is deferred logic for RIF counters, where counter polling starts based on a timer rather than immediately.

This process generally works as follows:

1. Create an object and add it to a list upon receiving an APP_DB update.
2. Start counter polling for all objects in the list during the timer event.
3. Stop counter polling for an object.
4. Remove the object.

If RIF creation and removal occur frequently, removal can happen before the timer event. As a result, the timer may start counter polling for an object that has just been removed, causing the following error message:
```ERR syncd#SDK: :- processFlexCounterEvent: port VID oid:0x600000000099d, was not found (probably port was removed/splitted) and will remove from counters now```

**How I verified it**
Tested in regression.

**Details if related**

### CODE REVIEW

**Review by dgsudharsan — APPROVED** [review]

**Comment by liat-grozovik** [comment]
any further request or this can be merged?

File: orchagent/intfsorch.cpp [hunk 1]
```diff
@@ -1746,4 +1759,5 @@ void IntfsOrch::voqSyncIntfState(string &alias, bool isUp)
         m_tableVoqSystemInterfaceTable->hset(port_alias, "oper_status", isUp ? "up" : "down");
     }
 
-}
\ No newline at end of file
+}
+
```
prsunny, line 1763: Please remove unrelated changes from PR

PR #3337, opened 2024-10-22
Fix p4orch tests after SAI update

saiarcot895:
**What I did**

Make sure that the p4orch tests are always run (not just when code coverage is enabled), and fix the p4orch tests following the SAI update.

**Why I did it**

sonic-net/sonic-sairedis#1431 updated the SAI submodule to the latest version, and #3321 fixed some of the compatibility issues, but there were still some that got through, because p4orch tests are enabled only when code coverage is enabled.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/Makefile.am
+++ b/orchagent/Makefile.am
@@ -8,9 +8,7 @@ INCLUDES = -I $(top_srcdir)/lib \
            -I pbh \
            -I nhg
 
-if GCOV_ENABLED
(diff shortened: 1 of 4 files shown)


kcudnik (approved):

prsunny (dismissed):

prsunny (approved):
lgtm, @mint570 , let us know if you have any concerns

PR #3347, opened 2024-10-29
Fix State Db LAG_MEMBER_TABLE removal not happening.

abdosi:
What I did:
Original issue and and PR: https://github.com/sonic-net/sonic-swss/pull/3333

Why I did:
Fix the failure in `test_po_update.py` as seen in this PR checker: https://github.com/sonic-net/sonic-buildimage/pull/20610
Previous fix blocked State Db `LAG_MEMBER_TABLE` deletion which is not correct as this table is created by tlm_teamd (owner).
Intention was to prevent deletion of State Db `LAG_TABLE` owner of which is teamsyncd.

### DIFF
--- a/tlm_teamd/values_store.cpp
+++ b/tlm_teamd/values_store.cpp
@@ -281,6 +281,14 @@ void ValuesStore::remove_keys_db(const std::vector<std::string> & keys)
         const auto & p = split_key(key);
         const auto & table_name = p.first;
         const auto & table_key = p.second;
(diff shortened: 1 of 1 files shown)


abdosi:
@yejianquan for viz.

yejianquan:
Hi @prsunny , could you please help to review and merge if it got approved?

judyjoseph (approved):

PR #3353, opened 2024-11-05
Changes to read route tag value from Netlink Message and take appropriate action

abdosi:
What I did:

Added support to compile and load code to read yaml file in
routesync.cpp. In routesyncd we read tag from constants.yml and take action
of not to send route to appdb and also fallback to default route.

Why I did:

Corresponding PR on FRR : https://github.com/sonic-net/sonic-buildimage/pull/20692 to send route tag to fpmsyncd.
PR to specify constants.yml tag value: https://github.com/sonic-net/sonic-buildimage/pull/20224
PR for installing libyaml package: https://github.com/sonic-net/sonic-buildimage/pull/20698.

Reference to full context of this chnages: 
[Swss_route_enhancemnts.docx](https://github.com/user-attachments/files/18149089/Swss_route_enhancemnts.docx)


abdosi:
@prsunny : can you help review this.

fpmsyncd/routesync.cpp
```diff
@@ -86,6 +87,20 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
...
+    try
+    {
+        root = YAML::LoadFile("/etc/sonic/constants.yml");
+        route_tag_not_to_appdb = root["constants"]["bgp"]["route_do_not_send_appdb_tag"].as<int>();
+        route_tag_fallback_to_default_route = root["constants"]["bgp"]["route_eligible_for_fallback_to_default_tag"].as<int>();
+    }
+    catch (const exception &e)
+    {
+        cout << "Exception \"" << e.what() << "\" had been thrown in daemon in loading constants.yml" << endl;
+        route_tag_not_to_appdb = 0;
```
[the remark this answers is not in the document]
abdosi: only if fails to reads constants.yml which should not happen. this attribute are define in constants.yml (irrespective of the role) 
https://github.com/sonic-net/sonic-buildimage/pull/20224/files#diff-e6f2fe13a6c276dc2f3b27a5bef79886f9c103194be4fcb28ce57375edf2c23c

fpmsyncd/routesync.cpp
```diff
@@ -693,6 +710,13 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
         destipprefix[strlen(vrf)] = ':';
     }
 
+    tag = rtnl_route_get_priority(route_obj);
+
+    if (tag == route_tag_not_to_appdb)
```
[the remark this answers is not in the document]
abdosi: it should not be zero. It always reads form constants.yml.  Only time this will be zero in case of fail condition of loading/reading the file.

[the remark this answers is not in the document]
abdosi: > it should not be zero. It always reads from constants.yml. Updated the fail case to be non-zero as `0xffffffff`

fpmsyncd/routesync.cpp
```diff
@@ -693,6 +710,13 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
...
+        return;
+    else if (tag == route_tag_fallback_to_default_route)
+       route_eligible_for_fallback_to_default_route = true;
```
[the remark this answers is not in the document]
abdosi: updated.

.azure-pipelines/docker-sonic-vs/Dockerfile
```diff
@@ -11,21 +11,19 @@ COPY ["debs", "/debs"]
...
-RUN dpkg -i /debs/libdashapi_1.0.0_amd64.deb \
-            /debs/libswsscommon_1.0.0_amd64.deb \
-            /debs/python3-swsscommon_1.0.0_amd64.deb \
-            /debs/sonic-db-cli_1.0.0_amd64.deb \
-            /debs/libsaimetadata_1.0.0_amd64.deb \
-            /debs/libsairedis_1.0.0_amd64.deb \
-            /debs/libsaivs_1.0.0_amd64.deb \
-            /debs/syncd-vs_1.0.0_amd64.deb \
-            /debs/swss_1.0.0_amd64.deb
+RUN apt-get update
```
dgsudharsan, line 14: Can you please explain motivation for this change?

fpmsyncd/routesync.cpp
```diff
@@ -1714,6 +1738,12 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
         fvVector.push_back(wt);
     }
 
+    if (route_eligible_for_fallback_to_default_route)
+    {
+        FieldValueTuple tag("fallback_to_default_route", "true");
```
dgsudharsan, line 1743: Where is this field processed?

abdosi: @dgsudharsan : Here https://github.com/sonic-net/sonic-swss/pull/3389

.azure-pipelines/docker-sonic-vs/Dockerfile
abdosi: @dgsudharsan : To resolve `lyaml-cpp` dependency of swss debian package.  As `apt install` take care of resolve all dependency

prsunny:
@abdosi , it will be good if we have a document to capture the end-to-end behavior. It will be easier to understand on who sets what and when. We can discuss this in the swss infra subgroup meeting.

abdosi:
> > @abdosi , it will be good if we have a document to capture the end-to-end behavior. It will be easier to understand on who sets what and when. We can discuss this in the swss infra subgroup meeting.
> 
> sure @prsunny , I will prepare write-up on this.
@prsunny : Document attached

rlhui:
@prsunny  reminder, thx

judyjoseph:
@abdosi are we planning to merge this PR in master

abdosi:
@prsunny : can you help sign-off this

arlakshm (dismissed):

fpmsyncd/routesync.cpp
```diff
@@ -2288,6 +2316,11 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
         }
     }
 
+    if (route_eligible_for_fallback_to_default_route)
```
dgsudharsan, line 2319: How will this change work if there are two routes, one which is eligible for fallback to default route but other is not while they share the same ecmp? The core logic here involves updating the underlying ECMP with members from default route. In such scenarios, rather than updating the members with ECMP nexthop of default route, we should make this route point to ECMP group created from default route so scenarios like I mentioned won't be a problem.

dgsudharsan (changes requested):

fpmsyncd/routesync.cpp
abdosi: currently we don't have such use case.  this property (`tag`) is controlled via FRR route-maps which we are per peer-group so a given route learnt from user-specified peer only and is controlled by associating it to the tag

Goal of this change is fast convergence where we don't want to SET operation on Route to update it's nexthop group which will slow down the convergance

Also this change is not covering all possible case. It only cover specific scenarios.

dgsudharsan: I am afraid we are updating generic logic for a use case specific case. The design here is you obtained the tag via a route but then update nexthop which can be shared among other routes. While this may not be a problem in your use case, this may be against the generic design logic. @prsunny  what is your feedback on this?

abdosi: that is not scope of this change. If any application has such use case design need to be changed accordingly . This change does not break any existing behavior. This has been presented in swss sub group and was agreed upon by the community.

swss PR has  already merged https://github.com/sonic-net/sonic-swss/pull/3389.

PR #3356, opened 2024-11-06
[dualtor][warm reboot] Fix neighbor inconsistency for standby mux ports after warm reboot

lolyu:
**What I did**
Let's enable mux orch to cache the neighbor updates during warm-reboot restoration, and do the updates after everything else being settled (mux configs/peer switch config are applied), so the neighbors could be reprogrammed based on the mux state.

* Microsoft ADO (number only): 30148066

**Why I did it**
The neighbors on standby mux port, they are always installed directly into the ASIC after warm reboot.
They should be installed as the mux standby with tunnel to the active side.

**How I verified it**
1. verify on dualtor testbed, together with https://github.com/sonic-net/sonic-sairedis/pull/1432, that neighbors are reprogrammed consistently based on mux states after warm reboot, no orchagent/sycnd crash.
2. UT pass

**Details if related**


orchagent/muxorch.cpp
```diff
@@ -1862,6 +1869,30 @@ bool MuxOrch::isStandaloneTunnelRouteInstalled(const IpAddress& neighborIp)
     return standalone_tunnel_neighbors_.find(neighborIp) != standalone_tunnel_neighbors_.end();
 }
 
+void MuxOrch::updateCachedNeighbors()
+{
+    if (!enable_cache_neigh_updates_)
+    {
+        SWSS_LOG_NOTICE("Skip process cached neighbor updates");
```
prsunny, line 1876: lets change to INFO. this would be logged frequently otherwise

lolyu: `updateCachedNeighbors` is only called once in `OrchDaemon::warmRestoreAndSyncUp`, and it will only be called during warm reboot.

prsunny (approved):

prsunny:
@Ndancejic for viz

PR #2883, opened 2023-08-22
[buffers] Add handler for the 'create_only_config_db_buffers' configuration knob

vadymhlushko-mlnx:
**What I did**
Add support in the orchagent for the configuration knob:

```
{
    "DEVICE_METADATA": {
        "localhost": {
            "create_only_config_db_buffers": "true"
            ...
        }
    }
}
```

**Why I did it**
If the `"create_only_config_db_buffers"` is equal to `"true"` - buffers will be created according to the config_db configuration (for example `BUFFER_QUEUE|*` table).
If the `"create_only_config_db_buffers"` is equal to `"false"` or does not exist - the maximum available buffers (which are read from SAI) will be created, regardless of `the config_db` buffer config.

**How I verified it**
Add UT.

Manual verification:
1. Install the image with this PR included on the not `MSFT SKU` switch
2. Check the `show queue counters` output and verify that only configured in `CONFIG_DB` buffers are created
```
root@sonic:/home/admin# show queue counters
     Port    TxQ    Counter/pkts    Counter/bytes    Drop/pkts    Drop/bytes
---------  -----  --------------  ---------------  -----------  ------------
Ethernet0    UC0               0                0            0           N/A
Ethernet0    UC1               0                0            0           N/A
Ethernet0    UC2               0                0            0           N/A
Ethernet0    UC3               0                0            0           N/A
Ethernet0    UC4               0                0            0           N/A
Ethernet0    UC5               0                0            0           N/A
Ethernet0    UC6               0                0            0           N/A
```
3. Open the `/usr/share/sonic/device/$DEVICE/$SKU/create_only_config_db_buffers.json` and change it to:
```
"create_only_config_db_buffers": "false"
```
4. Do `config reload`
5. Check the `show queue counters` output and verify that all available buffers are created
```
root@sonic:/home/admin# show queue counters
     Port    TxQ    Counter/pkts    Counter/bytes    Drop/pkts    Drop/bytes
---------  -----  --------------  ---------------  -----------  ------------
Ethernet0    UC0               0                0            0           N/A
Ethernet0    UC1               0                0            0           N/A
Ethernet0    UC2               0                0            0           N/A
Ethernet0    UC3               0                0            0           N/A
Ethernet0    UC4               0                0            0           N/A
Ethernet0    UC5               0                0            0           N/A
Ethernet0    UC6               0                0            0           N/A
Ethernet0    UC7              60            15346            0           N/A
Ethernet0    MC8             N/A              N/A          N/A           N/A
[...]

PR #3360, opened 2024-11-07
Do not start counter polling for queues with zero buffer profiles if create_only_config_db_buffers is enabled

stephenxs:
**What I did**

Do not start counter-polling for queues with zero buffer profiles if create_only_config_db_buffers is enabled.

**Why I did it**

If create_only_config_db_buffers is enabled, counters will be polled only on queues with a buffer configured based on the hypothesis that a customer must configure a buffer profile before using the queue.
The system performance can be optimized by reducing the number of counter polled.

However, zero buffer profiles are configured on queues of inactive ports to reclaim unused buffers. Those queues are not used by customers. Hence, we do not need to start counter polling for queues with such buffer profiles configured.

**How I verified it**

Manual and regression test.

Flows verified:
1. System starts up with inactive ports
2. Shutdown a port
3. Start up a port

**Details if related**

Regarding `create_only_config_db_buffers`, it was introduced in PR #2883 for polling configured buffer counters only.
The idea is that a customer does not want to use certain buffer objects, like queues, and priority groups, if he/she does not configure a buffer profile on the BUFFER_PG or BUFFER_QUEUE table. In that case, it does not make sense to poll counters for unconfigured buffer objects.
However, zero buffer profile, which is a special type of buffer profile, can be configured on a BUFFER_PG or BUFFER_QUEUE table just to [reclaim unused buffers](https://github.com/sonic-net/SONiC/blob/master/doc/qos/reclaim-reserved-buffer.md) on inactive ports. In this case, we should not poll counters for those buffer objects as well.

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -360,6 +360,25 @@ const object_reference_map &BufferOrch::getBufferPoolNameOidMap(void)
     return *m_buffer_type_maps[APP_BUFFER_POOL_TABLE_NAME];
 }
 
+void BufferOrch::getBufferObjectsWithNonZeroProfile(vector<string> &nonZeroQueues, const string &table)
+{
+    for (auto &&queueRef: (*m_buffer_type_maps[table]))
+    {
+        for (auto &&profileRef: queueRef.second.m_objsReferencingByMe)
+        {
+            if (profileRef.second.find("_zero_") == std::string::npos)
+            {
+                SWSS_LOG_INFO("Selected key %s with profile %s", queueRef.first.c_str(), profileRef.second.c_str());
+                nonZeroQueues.push_back(queueRef.first);
+            }
+            else
+            {
+                SWSS_LOG_INFO("Skipped key %s with profile %s", queueRef.first.c_str(), profileRef.second.c_str());
+            }
+        }
+    }
+}
+
 task_process_status BufferOrch::processBufferPool(KeyOpFieldsValuesTuple &tuple)
 {
     SWSS_LOG_ENTER();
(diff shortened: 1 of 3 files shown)


stephenxs:
Hi @kperumalbfn 
Could you review the PR? I added details about `create_only_config_db_buffers` and `zero buffer profiles`.
Thanks.

orchagent/bufferorch.cpp
```diff
@@ -1161,14 +1199,16 @@ task_process_status BufferOrch::processPriorityGroup(KeyOpFieldsValuesTuple &tup
                     {
                         auto flexCounterOrch = gDirectory.get<FlexCounterOrch*>();
                         auto pgs = tokens[1];
-                        if (op == SET_COMMAND &&
+                        if (!counter_was_added && counter_needs_to_add &&
```
kperumalbfn, line 1202: @stephenxs As part of PFC detection, some platform applies zero buffer profile to queues. That profile is created only in ASIC_DB, does this counter poll bypass affect those queues as well? Can you confirm?

stephenxs: @kperumalbfn 
no it doesn't 
It only bypasses the zero profiles that are applied from APPL DB.
For PFC WD, zero profiles they are applied by calling SAI API directly. In APPL DB they are still normal profiles.
So counter polling remains.

liat-grozovik:
@stephenxs can you please follow up on the checker failure?
@kperumalbfn any additional feedback or you can signoff for this PR?

stephenxs:
> @stephenxs can you please follow up on the checker failure?
> 
> @kperumalbfn any additional feedback or you can signoff for this PR?

The check failure doesn't look relevant to the PR. Retrying.

stephenxs:
> > @stephenxs can you please follow up on the checker failure?
> > @kperumalbfn any additional feedback or you can signoff for this PR?
> 
> The check failure doesn't look relevant to the PR. Retrying.

codeQL failure has been fixed by the community.
Rebased PR to pass the test.

bingwang-ms:
@stephenxs What's the impact if this change is missing?

stephenxs:
> @stephenxs What's the impact if this change is missing?

This is to optimize boot up time on SKUs where there are admin down ports.

PR #3361, opened 2024-11-08
Fixing fpmsyncd to handle scaled nexthops

dgsudharsan:
**What I did**
Fixing fpmsyncd to handle scaled nexthops(e.g. 256 nexthops). In this scenario the max message len is increased to accommodate zebra message size. In addition initialized nlmsg_set_default_size to allow netlink to process the increased message size as well without which libnl would crash.

**Why I did it**
To support increased nexthops.

**How I verified it**
Running test with 256 nexthops

**Details if related**


liat-grozovik:
@prsunny kindly reminder for the review. this is needed for 202411

prsunny (approved):
lgtm

liat-grozovik:
@dgsudharsan could you please check the failure on the codeQL? could it be infra issue and rerun is needed?

dgsudharsan:
@liat-grozovik  This is the issue that @saiarcot895  is working on. This is due to libnl failure and affects all PRs

@saiarcot895 do you have ETA?

PR #3363, opened 2024-11-09
Add port FEC BER feature swss

vincentpcng:
**What I did**
This is to add the swss changes for the feature port FEC BER. 
It modify the port_rates.lua script to compute the BER and stored them in the DB .
Two additonal PR(s) will address the cli ( sonic-utilities) and the sonic-mgmt changes

**Why I did it**
The HLD for this feature is HLD#1829

**How I verified it**
We verify the counters internally
(1) verify when link has correctable FEC errors,  the BER were calcuated accordingly
(2) verify 400G with 4x100 and the BER , four serdes with 100G rate
(3) run a redis-script to polling the DB counters and verify the calculation
(4) stop polling manually  modify the uncorrectable counter and verify the cli display and calculation

**Details if related**


orchagent/port_rates.lua
```diff
@@ -83,6 +193,21 @@ local function compute_rate(port)
...
             redis.call('HSET', state_table, 'INIT_DONE', 'DONE')
         end
+
+        -- only do the calculation when all info present
+        
+        if fec_corr_bits and fec_uncorr_frames and lanes_count and serdes_speed then
+            local fec_corr_bits_last = redis.call('HGET', rates_table_name .. ':' .. port, 'SAI_PORT_STAT_IF_FEC_CORRECTED_BITS_last')
+            local fec_uncorr_frames_last = redis.call('HGET', rates_table_name .. ':' .. port, 'SAI_PORT_STAT_IF_FEC_NOT_CORRECTABLE_FARMES_last')
+
+            local serdes_rate_total = lanes_count * serdes_speed * delta / 1000
```
prgeor, line 203: @vincentpcng this is actually total bits during the delta period

orchagent/port_rates.lua
```diff
@@ -29,11 +32,117 @@ logit(alpha)
...
+    local fec_corr_bits, fec_uncorr_frames
+    local fec_corr_bits_ber_new, fec_uncorr_bits_ber_new = -1, -1
+    -- HLD review suggest to use the statistical average when calculate the post fec ber
+    local rs_average_frame_ber = 1e-8
+    local lanes_speed, serdes_speed, lanes_count = 0, 0, 0
+
+    -- lookup interface name from oid
+    local interface_name = find_interface_name_from_oid(port)
+    if interface_name then
+        lanes_count, lanes_speed, serdes_speed = find_lanes_and_serdes(interface_name)
```
prgeor, line 137: @vincentpcng i was wondering if serdes speed could be populated by OA in APPL_DB port table so that we don't have to compute every time.....may be an optimization for the future

prgeor (approved):

orchagent/port_rates.lua
vincentpcng: In the future if we can add the serdes speed to the appl_db then we can save this calculation.
Agree that we can do it as a future enhancement

orchagent/port_rates.lua
prgeor: @vincentpcng i was referring to the variable name `serdes_rate_total` which actually could be renamed `total_bits` since its not `rate`

prgeor (approved):

prgeor:
@prsunny can you merge?

PR #3367, opened 2024-11-14
Vlanmgrd

abdosi:
What I did:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/20684

Why I did:
It is seen when doing `config reload`  on multi-asic (multiple instances of swss/teamd) their might be race condition when teammrd might be getting SIGTERM and cleaning up the port-channel in kernel and vlanmgrd might be adding the port channel as member in kernel and this can cause the exception as this operation as being done in async.

How I did:
To handle the failure gracefully if we get an exception when portchannel as vlan member in kernel we check if Portchannel exists and if now we just retry to minimize the risk of failure due to above race-condition.


abdosi:
@yejianquan for viz.

cyw233:
Hey @abdosi, looks like the code coverage check did not pass

yejianquan:
Hi @abdosi , let's try to merge this PR early so that we can include it in this week's nightly test

abdosi:
> Hi @abdosi , let's try to merge this PR early so that we can include it in this week's nightly test

yes, will add UT for this tomorrow.

prsunny:
Bypassing code coverage since its handling exception scenario. @abdosi planning to add UT coverage.

yejianquan:
@bingwang-ms  could you help to cherry-pick this?

PR #3370, opened 2024-11-14
[vlanmgrd]: Preventing VLAN interfaces from going down when all of their member ports are down

mramezani95:
**What I did**
Started up the "dummy" interface that is already connected to the bridge named "Bridge" so that when all ports connected to the bridge are down, the bridge interface remains up. This ensures that when there is one VLAN interface connected to the bridge, it remains up even when all of the VLAN's member ports are down.

**Why I did it**
Before this change, when all member ports of a VLAN interface went down, the VLAN interface also went operationally down. This effectively disabled the advertisement of the VLAN interface's subnet IP to neighbors and prevented the IP-in-IP decapsulation feature for the VLAN interface from working properly.

**How I verified it**
I ran the tests submitted in PR #15244 for sonic-mgmt repo on a local T0 testbed and verified that when all member ports of a VLAN are brought down, the VLAN interface remains operationally up and BGP advertisement and IP-in-IP decapsulation features work as expected.

### DIFF
--- a/cfgmgr/vlanmgr.cpp
+++ b/cfgmgr/vlanmgr.cpp
@@ -79,7 +79,8 @@ VlanMgr::VlanMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
     //               /sbin/bridge vlan del vid 1 dev Bridge self;
     //               /sbin/ip link del dummy 2>/dev/null;
     //               /sbin/ip link add dummy type dummy &&
-    //               /sbin/ip link set dummy master Bridge"
+    //               /sbin/ip link set dummy master Bridge &&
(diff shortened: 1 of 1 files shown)


prsunny:
@dgsudharsan , fyi

dgsudharsan:
If I understand, this Vlan interface without neighbors is used only for decapsulation purpose? Why can't we use Loopback interface?

prsunny:
> If I understand, this Vlan interface without neighbors is used only for decapsulation purpose? Why can't we use Loopback interface?

no, we use Vlan GW IP as well for decap in addition to Loopback.

dgsudharsan (dismissed):

prsunny (approved):

PR #3383, opened 2024-11-20
[orchagent]: VXLAN: Fix oper_status and tunnel encapsulation TTL

bradh352:
**What I did**

This fixes 2 issues across a range of open tickets building upon patches created by others with modifications as requested by @VladimirKuk.

The first issue this resolves is the status shown for remote vteps which in the fact that it is wrong makes debugging nearly impossible:
```
# show vxlan remotevtep
+------------+------------+-------------------+--------------+
| SIP        | DIP        | Creation Source   | OperStatus   |
+============+============+===================+==============+
| 172.16.0.1 | 172.16.0.2 | EVPN              | oper_down    |
+------------+------------+-------------------+--------------+
Total count : 1
```

The remote VTEP is really up.

Original PR for that is #2080.

Also fixes https://github.com/sonic-net/sonic-buildimage/issues/10004 or at least the error message which hurts debugging.

The next issue is in reachabiity across VXLANs.  This fixes IP/MAC learning via ARP.  The original PR for that is #3216, however it appears it has its origins in
https://github.com/sonic-net/sonic-buildimage/issues/10050 which goes into greater detail about the issue itself.  Also there is talk about it here https://github.com/kamelnetworks/sonic/issues/9 as well as another similar patch here: https://github.com/kamelnetworks/sonic-swss/commit/02ee3e3

**Why I did it**

Fixes #3216
Fixes #2080
Fixes https://github.com/sonic-net/sonic-buildimage/issues/10050
Fixes https://github.com/sonic-net/sonic-buildimage/issues/10004

**How I verified it**

Pulled into my private sonic-swss fork:
https://github.com/bradh352/sonic-swss/commits/bradh352/master

Which is pulled in by my private sonic-buildimage fork:
https://github.com/bradh352/sonic-buildimage/tree/bradh352/master

Which is then automatically built when changes are made.  Then the uploaded asset of sonic-broadcom.bin is installed onto Dell S5248F switches and tested.

Update: also now tested by me on Mellanox/NVidia SN2201

**Details if related**

This should also be backported to 202411, 202405

### DIFF
--- a/orchagent/vxlanorch.cpp
+++ b/orchagent/vxlanorch.cpp
@@ -275,7 +275,7 @@ create_tunnel(
     sai_ip_address_t *dst_ip,
     sai_object_id_t underlay_rif,
     bool p2p,
-    sai_uint8_t encap_ttl=0)
+    sai_uint8_t encap_ttl)
 {
     sai_attribute_t attr;
     std::vector<sai_attribute_t> tunnel_attrs;
@@ -349,16 +349,13 @@ create_tunnel(
         tunnel_attrs.push_back(attr);
     }
 
-    if (encap_ttl != 0)
-    {
-        attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_MODE;
-        attr.value.s32 = SAI_TUNNEL_TTL_MODE_PIPE_MODEL;
-        tunnel_attrs.push_back(attr);
+    attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_MODE;
+    attr.value.s32 = SAI_TUNNEL_TTL_MODE_PIPE_MODEL;
+    tunnel_attrs.push_back(attr);
 
-        attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_VAL;
-        attr.value.u8 = encap_ttl;
-        tunnel_attrs.push_back(attr);
-    }
+    attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_VAL;
+    attr.value.u8 = encap_ttl;
+    tunnel_attrs.push_back(attr);
 
     sai_object_id_t tunnel_id;
     sai_status_t status = sai_tunnel_api->create_tunnel(
--- a/orchagent/vxlanorch.cpp
+++ b/orchagent/vxlanorch.cpp
@@ -275,7 +275,7 @@ create_tunnel(
     sai_ip_address_t *dst_ip,
     sai_object_id_t underlay_rif,
     bool p2p,
-    sai_uint8_t encap_ttl)
+    sai_uint8_t encap_ttl=0)
 {
     sai_attribute_t attr;
     std::vector<sai_attribute_t> tunnel_attrs;
@@ -349,13 +349,16 @@ create_tunnel(
         tunnel_attrs.push_back(attr);
     }
 
-    attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_MODE;
-    attr.value.s32 = SAI_TUNNEL_TTL_MODE_PIPE_MODEL;
-    tunnel_attrs.push_back(attr);
+    if (encap_ttl != 0)
+    {
+        attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_MODE;
+        attr.value.s32 = SAI_TUNNEL_TTL_MODE_PIPE_MODEL;
+        tunnel_attrs.push_back(attr);
 
-    attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_VAL;
-    attr.value.u8 = encap_ttl;
-    tunnel_attrs.push_back(attr);
+        attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_VAL;
+        attr.value.u8 = encap_ttl;
+        tunnel_attrs.push_back(attr);
+    }
 
     sai_object_id_t tunnel_id;
     sai_status_t status = sai_tunnel_api->create_tunnel(
--- a/tests/test_vxlan_tunnel.py
+++ b/tests/test_vxlan_tunnel.py
@@ -154,7 +154,7 @@ def check_vxlan_tunnel(dvs, src_ip, dst_ip, tunnel_map_ids, tunnel_map_entry_ids
                         'SAI_TUNNEL_ATTR_PEER_MODE': 'SAI_TUNNEL_PEER_MODE_P2MP',
                         'SAI_TUNNEL_ATTR_ENCAP_SRC_IP': src_ip,
                         'SAI_TUNNEL_ATTR_ENCAP_TTL_MODE': 'SAI_TUNNEL_TTL_MODE_PIPE_MODEL',
-                        'SAI_TUNNEL_ATTR_ENCAP_TTL_VAL': '64'
+                        'SAI_TUNNEL_ATTR_ENCAP_TTL_VAL': '255'
                     }
                 )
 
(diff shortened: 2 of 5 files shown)


VladimirKuk:
Tests do fail from time to time.
At least to me, these tests are unrelated to the change.

lukasstockner:
Thank you for pushing this forward!
For the record, we've been running these changes in production for ~2 years without issues, so I'd be quite confident that they work as expected.

bradh352:
@VladimirKuk I ended up having to sprinkle your suggestion in 2 places to get it fully working.

VladimirKuk (approved):

orchagent/vxlanorch.cpp
```diff
@@ -275,7 +275,7 @@ create_tunnel(
     sai_ip_address_t *dst_ip,
     sai_object_id_t underlay_rif,
     bool p2p,
-    sai_uint8_t encap_ttl=0)
```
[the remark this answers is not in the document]
bradh352: that is a static function that only exists in that file and is never called without the encap_ttl argument.  The public functions both set the default value and are optional.  I can of course set to a proper default value there if you'd prefer, its just never used so I didn't see a point.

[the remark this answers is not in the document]
bradh352: alternatively we can just go back to default of 0 everywhere to explicitly mean "choose something for me", then check for 0 in this and set it to 64.  Some old revisions did that but @VladimirKuk suggested to change it.  I personally don't care either way, both ways definitively fix the issue at hand.

bradh352:
@prsunny how would you like to proceed on this?  I think this is a critical issue since community SONiC doesn't support ARP/ND suppression so things just don't work at all without this.

bradh352:
@prsunny looks like the internal msconflict is resolved now since pvst was merged to main.  Please review and hopefully merge.  Thanks!

bradh352:
rebased on top of current master to get rid of merged stuff that is now upstream

bradh352:
I can't tell from the output if somehow the test failures are related to this PR.  I'm not familiar enough with the test case output to differentiate expected failures from real ones.

bradh352:
@prsunny figured out the test case failure and corrected it, it was due to specifically checking tunnel attributes and we added 2 new ones.

bradh352:
@abdosi, @judyjoseph, @srj102 any review comments?  This is a pretty large blocker for VXLAN EVPN

orchagent/vxlanorch.cpp
```diff
@@ -349,16 +349,13 @@ create_tunnel(
         tunnel_attrs.push_back(attr);
     }
 
-    if (encap_ttl != 0)
-    {
-        attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_MODE;
-        attr.value.s32 = SAI_TUNNEL_TTL_MODE_PIPE_MODEL;
-        tunnel_attrs.push_back(attr);
+    attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_MODE;
```
srj102, line 352: The current behaviour of this helper function allowed for uniform as well as pipe model with the default being uniform (encap_ttl=0) 
With this change sonic forces tunnels to be in pipe mode which is a step back. 

Hence I would say stick with the current implementation.

srj102 (changes requested):

orchagent/vxlanorch.cpp
bradh352: What's the solution here to the issue at hand then?  What we're seeing is ARP across the tunnel no longer works as it gets a TTL of 0 and gets dropped, therefore machines are unable to find eachother across the network.  At least on Broadcom Trident 3.

bradh352: or are you saying to try to just set a TTL > 0 in Uniform mode?  So just don't change the mode attribute and see what happens?

srj102: Is the ARP TTL issue on Trident3 still seen with the latest SAI versions ?

bradh352: as of 3 weeks ago on master anyhow, I haven't tried without this patch since then

bradh352: @srj102 how do you want me to proceed here?  I can test anything you want me to test within a few hours, such as setting the TTL be leaving in uniform mode.  That said, I must say with VXLAN EVPN I'm not sure uniform mode make sense anyhow unless I'm missing something.  MPLS it does of course.  I've also seen some vendors indicate they *only* support pipe mode on VXLAN EVPN (https://documentation.extremenetworks.com/slxos/sw/20xx/20.4.3/l2config/GUID-537EC20E-95A2-4CA7-A417-D45D596AD575.shtml)

srj102: The default of uniform mode seems to be coming from SAI defaults.

The foll is my suggestion. 
a. Not to change the create_tunnel helper function. 
b. Pass a non default encap_ttl from createTunnelHw()

I am fine with having the default encap ttl mode to be PIPE. 
@prsunny  @dgsudharsan  please comment. 

If this is not acceptable,
 then we might have to have this as PIPE only for BRCM platforms. 
Define a default_encap_ttl and set it to PIPE only for BRCM platforms (platform == BRCM_PLATFORM_SUBSTRING )

bradh352: @srj102 ping

bradh352: @srj102 I was really just asking for feedback on the latest changeset, if that's more inline with what you were wanting...

bradh352:
@prsunny @dgsudharsan @srj102 any further feedback here?  I'd really like to get this merged.  There are lots of people using this patch already.

srj102:
@bradh352 Thanks for the fixes. 

By keeping the create_tunnel helpers default values we are not impacting the other orchs like nvgre/srv6/tunneldecap.
The latest change you made is fine. 

vnetorch also keeps the VXLAN tunnel created in the PIPE mode and hence we should be able to go with PIPE mode for BGP EVPN vxlan tunnels.

bradh352:
> @bradh352 Thanks for the fixes.
> 
> By keeping the create_tunnel helpers default values we are not impacting the other orchs like nvgre/srv6/tunneldecap. The latest change you made is fine.
> 
> vnetorch also keeps the VXLAN tunnel created in the PIPE mode and hence we should be able to go with PIPE mode for BGP EVPN vxlan tunnels.

Can you mark it as approved?

Also, I recently acquired a Mellanox Spectrum1 switch (SN2201) for testing and can confirm this patch also works on that ASIC.

tests/test_vxlan_tunnel.py
```diff
@@ -152,7 +152,9 @@ def check_vxlan_tunnel(dvs, src_ip, dst_ip, tunnel_map_ids, tunnel_map_entry_ids
                         'SAI_TUNNEL_ATTR_DECAP_MAPPERS': decapstr,
                         'SAI_TUNNEL_ATTR_ENCAP_MAPPERS': encapstr,
                         'SAI_TUNNEL_ATTR_PEER_MODE': 'SAI_TUNNEL_PEER_MODE_P2MP',
-                        'SAI_TUNNEL_ATTR_ENCAP_SRC_IP': src_ip
+                        'SAI_TUNNEL_ATTR_ENCAP_SRC_IP': src_ip,
+                        'SAI_TUNNEL_ATTR_ENCAP_TTL_MODE': 'SAI_TUNNEL_TTL_MODE_PIPE_MODEL',
+                        'SAI_TUNNEL_ATTR_ENCAP_TTL_VAL': '255'
                     }
                 )
```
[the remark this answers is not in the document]
bradh352: done

bradh352:
looks like builds were hanging yesterday ... I rebased my commits on current master to force a rebuild ... hopefully its fixed today.

bradh352:
just forced a rebuild, tests failed in srv6 tests ... shouldn't be related in any way to my changes

bradh352:
only failures in the log are srv6 related, just like the minimal PR in #3482 also failing due to SRV6 tests failing

bradh352:
@prsunny looks like tests are passing now that the SRv6 fixes got merged yesterday.  Can this be merged now?

PR #3387, opened 2024-11-22
[buffer] indicate shared pool by putting a very small shared buffer pool size

Junchao-Mellanox:
**What I did**

Put a very small shared buffer pool size when shared pool is detected enabling.

**Why I did it**

Avoid a timing issue like following:

1. the buffer pool sizes, xoff have initialized to 0, which means the shared headroom pool is disabled
2. but the buffer profiles already indicate the shared headroom pool is enabled
3. later on the buffer pool sizes are updated with xoff being non-zero

In case the orchagent starts handling buffer configuration between 2 and 3, it is inconsistent between buffer pools and profiles, which fails Mellanox SAI sanity check and causes orchagent abort

**How I verified it**

Manual test
Run sonic-mgmt regression

**Details if related**


stephenxs (approved):

Junchao-Mellanox:
Hi @prsunny , could you please help review? Thanks.

bingwang-ms:
@kperumalbfn Could you please help review?

kperumalbfn (approved):
LGTM

PR #3390, opened 2024-11-24
[muxorch] Process nexthops using bulker during mux switchover

Ndancejic:
Using nexthop object bulker when processing neighbors. This is only enabled during mux switchover.

When processing neighbors during mux switchover:
- add nexthop object to add/remove using nexhop bulker
- flush nexthop bulker
- post-process bulker results

**What I did**
Added nexthop bulking operations to addNexthop function in neighOrch.
This is currently only used when processing neighbors during mux switchover

**Why I did it**
Neighbor bulking during mux switchover was still limited by sai calls made to create/remove nexthops. This uses the existing sai bulk nexthop api to also bulk program nexthops

**How I verified it**
Ran switchover testing on 7260cx3 dualtor testbed running 202405

### DIFF
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -28,6 +28,7 @@ const int neighorch_pri = 30;
 
 NeighOrch::NeighOrch(DBConnector *appDb, string tableName, IntfsOrch *intfsOrch, FdbOrch *fdbOrch, PortsOrch *portsOrch, DBConnector *chassisAppDb) :
         gNeighBulker(sai_neighbor_api, gMaxBulkSize),
+        gNextHopBulker(sai_next_hop_api, gSwitchId, gMaxBulkSize),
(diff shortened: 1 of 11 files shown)


orchagent/neighorch.cpp
```diff
@@ -267,6 +269,13 @@ bool NeighOrch::addNextHop(const NextHopKey &nh)
     next_hop_attrs.push_back(next_hop_attr);
 
     sai_object_id_t next_hop_id;
+
+    if (ctx.bulk_op)
```
prsunny, line 273: Just to clarify, the bulk operation is done only from mux context, right? or is it becoming the default flow with this change?

Ndancejic: Yes, currently the bulk_op flag is only set while processing mux neighbors during switchover

PR #3394, opened 2024-11-26
Added change not to create ECMP Group in SAI and program the route if none of ECMP members are active/link-up

abdosi:
What I did:
Added change not to create ECMP Group in SAI and program the route if none of the ECMP members are active/link-up.
Also do not program the Temp Route if Neigh is not active (Link Down)

Also as part of this change if Route is not programmed and if we remove that route than decrement VRF Reference count in `removeRoute` as `removeRoutePost` will not be called in this case.

Why I did:
In scale setup of T2 it's possible all links can go down simultaneously which case we can get Route messages with all nexthops being in down state. In such case we might create empty Nexthop Group in SAI for the given route which causes not needed SAI call for Nexthop Group creation and also create traffic blackhole for the route where that route can still forward traffic via default route if eligible/applicable.

Also in this case no point to add Temp Route if neighbor is link down.

How I verify:
1. Compile time test was failing because of this chnage so added oper_status as UP in the mock tests
2. Modified subport router interface UT for this scenario.
3. Ixia Convergence tests 

Reference to full context of this changes:
[Swss_route_enhancemnts.docx](https://github.com/user-attachments/files/18149172/Swss_route_enhancemnts.docx)


rlhui:
@prsunny  reminder to review, thanks.

prsunny:
@abdosi , i think the test failures are specific to this PR as I see it passing for all others

abdosi:
> @abdosi , i think the test failures are specific to this PR as I see it passing for all others

@prsunny : Please help review/sign-off. Fixed all the UT's.

prsunny (approved):

prsunny:
@Ndancejic , please review and check if there is any impact for mux route logic

bingwang-ms:
Readding label for cherry-pick

orchagent/routeorch.cpp
```diff
@@ -2423,8 +2433,14 @@ bool RouteOrch::removeRoute(RouteBulkContext& ctx)
     size_t creating = gRouteBulker.creating_entries_count(route_entry);
     if (it_route == it_route_table->second.end() && creating == 0)
     {
-        SWSS_LOG_INFO("Failed to find route entry, vrf_id 0x%" PRIx64 ", prefix %s\n", vrf_id,
-                ipPrefix.to_string().c_str());
+       if (it_route_table->second.size() == 0)
```
stephenxs, line 2436: I suspect it's unsafe to remove `m_syncdRoutes[vrf_id]` here.
`gRouteBulker.creating_entries_count(route_entry)` can give unexpected result. it indexes the creating entries using `sai_route_entry_t` as a key, which consists of union of IPv4 and IPv6. In case it is IPv4, the bytes for IPv6 only (bytes 4~15) are uninitialized in `copy`. This means the key `route_entry` here can have a different value in bytes 4~15 compared to value when the corresponding routing entry was added.
As a result, the `m_syncdRoutes[vrf_id]` can be removed even if there was ongoing entries in the VRF table.

We saw the following uncaught C++ exception
```
(gdb) bt
#0  0x00007f5791aedebc in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f5791a9efb2 in raise () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f5791a89472 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007f5791de0919 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007f5791debe1a in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007f5791debe85 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007f5791dec0d8 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007f5791de3240 in std::__throw_out_of_range(char const*) () from /lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x00005594e856d956 in std::map<unsigned long, std::map<swss::IpPrefix, RouteNhg, std::less<swss::IpPrefix>, std::allocator<std::pair<swss::IpPrefix const, RouteNhg> > >, std::less<unsigned long>, std::allocator<std::pair<unsigned long const, std::map<swss::IpPrefix, RouteNhg, std::less<swss::IpPrefix>, std::allocator<std::pair<swss::IpPrefix const, RouteNhg> > > > > >::at (this=<optimized out>, __k=<optimized out>) at /usr/include/c++/12/bits/stl_map.h:551
#9  0x00005594e8564beb in RouteOrch::addRoutePost (this=this@entry=0x5594ea13e080, ctx=..., nextHops=...) at ./orchagent/routeorch.cpp:2145
#10 0x00005594e856b0b2 in RouteOrch::doTask (this=0x5594ea13e080, consumer=...) at ./orchagent/routeorch.cpp:1021
#11 0x00005594e85282d2 in Orch::doTask (this=0x5594ea13e080) at ./orchagent/orch.cpp:553
#12 0x00005594e851909a in OrchDaemon::start (this=this@entry=0x5594ea0a0950) at ./orchagent/orchdaemon.cpp:895
#13 0x00005594e8485632 in main (argc=<optimized out>, argv=<optimized out>) at ./orchagent/main.cpp:818
(gdb) frame 9
#9  0x00005594e8564beb in RouteOrch::addRoutePost (this=this@entry=0x5594ea13e080, ctx=..., nextHops=...) at ./orchagent/routeorch.cpp:2145
```

stephenxs: https://github.com/sonic-net/sonic-swss/pull/3477 to fix it

PR #3399, opened 2024-11-27
[ACL] Add Tunnel Next Hop redirect support

vivekrnv:
**What I did**

- Add support for Tunnel Next Hop as the Redirect Action in ACL Rules
- Supported schema for REDIRECT_ACTION field : `<ip>@<tunnel_name> (Eg: 20.0.0.3@tunnel0)`
- Add a new Macro in the SAI mocking infra to clear extern mock_fns set
- Add UT's

**Why I did it**

**How I verified it**

```
[----------] 4 tests from AclRedirectActionTest
[ RUN      ] AclRedirectActionTest.TunnelNH
[       OK ] AclRedirectActionTest.TunnelNH (10 ms)
[ RUN      ] AclRedirectActionTest.TunnelNH_ExistingNhObject
[       OK ] AclRedirectActionTest.TunnelNH_ExistingNhObject (10 ms)
[ RUN      ] AclRedirectActionTest.TunnelNH_InvalidTunnel
[       OK ] AclRedirectActionTest.TunnelNH_InvalidTunnel (10 ms)
[ RUN      ] AclRedirectActionTest.TunnelNH_InvalidNextHop
[       OK ] AclRedirectActionTest.TunnelNH_InvalidNextHop (10 ms)
[----------] 4 tests from AclRedirectActionTest (40 ms total)
```

**Details if related**


vivekrnv:
```
[----------] 15 tests from MuxRollbackTest
[ RUN      ] MuxRollbackTest.StandbyToActiveNeighborAlreadyExists
[       OK ] MuxRollbackTest.StandbyToActiveNeighborAlreadyExists (86 ms)
[ RUN      ] MuxRollbackTest.ActiveToStandbyNeighborNotFound
pure virtual method called
terminate called without an active exception
FAIL tests (exit status: 134)
```

Unrelated UT failure. @lolyu Are you familiar on why this might fail?

oleksandrivantsiv (approved):

orchagent/aclorch.cpp
```diff
@@ -2001,21 +2042,38 @@ sai_object_id_t AclRulePacket::getRedirectObjectId(const string& redirect_value)
...
     {
         NextHopKey nh(target);
-        if (!m_pAclOrch->m_neighOrch->hasNextHop(nh))
+        if (m_pAclOrch->m_neighOrch->hasNextHop(nh))
         {
-            SWSS_LOG_ERROR("ACL Redirect action target next hop ip: '%s' doesn't exist on the switch", nh.to_string().c_str());
-            return SAI_NULL_OBJECT_ID;
+            m_redirect_target_next_hop = target;
+            m_pAclOrch->m_neighOrch->increaseNextHopRefCount(nh);
+            return m_pAclOrch->m_neighOrch->getNextHopId(nh);
```
prsunny, line 2049: Looks like logic is changed. Previously it returned NULL OBJECT if `hasnexthop `failed. Should we keep the old behaviour?

vivekrnv: This field can take many variants, it can be Port, Lag, NH IP, NHG and now Tunnel NH. 

Previous logic is incorrect as if get NH fails, it should try to check if it is one of other variants instead of returning  SAI_NULL_OBJECT_ID

orchagent/aclorch.cpp
```diff
@@ -617,6 +619,35 @@ bool AclTableRangeMatch::validateAclRuleMatch(const AclRule& rule) const
...
+
+    VxlanTunnelOrch* vxlan_orch = gDirectory.get<VxlanTunnelOrch*>();
+    /* Only the first call creates the SAI object, further calls just increment the ref count */
+    oid = vxlan_orch->createNextHopTunnel(tunnel_name, endpoint_ip, mac, vni);
+}
+
+void AclRule::TunnelNH::parse(const std::string& target)
+{
+    /* Supported Format: endpoint_ip@tunnel_name */
+    auto at_pos = target.find('@');
```
prsunny, line 634: Can you add this schema in the PR description and how it would look like with an example?

vivekrnv: updated

PR #3401, opened 2024-11-28
Improve log information when exception is caught in Orch2 class

DavidZagury:
**What I did**
Add the information regarding which class the exception is originated from when Orch2 catches an exception.

**Why I did it**
An Exception was thrown:
`ERR swss#orchagent: :- doTask: Logic error: map::at
`
This gives almost no information on where the exception happened.

**How I verified it**
Compiled SWSS which I forced an exception to be thrown from addOperation in VxlanTunnelOrch.
The log resulted is:

`ERR swss#orchagent: :- doTask: Logic error in 15VxlanTunnelOrch: map::at
`
Which focus you to a specific class.
Bigger improvement would need a larger refactor. 

**Details if related**

### DIFF
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -845,19 +845,19 @@ void Orch2::doTask(Consumer &consumer)
         }
         catch (const std::invalid_argument& e)
         {
-            SWSS_LOG_ERROR("Parse error: %s", e.what());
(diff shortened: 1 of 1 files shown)


prsunny (approved):

prsunny:
Bypassing coverage since its an exception path

bingwang-ms:
@DavidZagury, @prsunny I added the cherry-pick label for 202411 branch. It will not be cherry-picked to 202405 as it's considered as an enhancement.

PR #3405, opened 2024-12-01
[orchagent] fix "parsePortConfig: Unknown field(mode)"

bradh352:
**What I did**

When using "mode" as defined in
https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-port.yang this warning is emitted for each interface:
```
WARNING swss#orchagent: :- parsePortConfig: Unknown field(mode): skipping ...
```

It appears this field is not needed in orchagent at all, so we need to at least recognize it as valid then skip processing.

**Why I did it**

Warnings harm debugging and can cause wasted cycles when debugging real issues.

**How I verified it**

built via my private fork https://github.com/bradh352/sonic-buildimage and installed on physical switches

**Details if related**

### DIFF
--- a/orchagent/port/porthlpr.cpp
+++ b/orchagent/port/porthlpr.cpp
@@ -1229,6 +1229,12 @@ bool PortHelper::parsePortConfig(PortConfig &port) const
                 return false;
             }
         }
+        else if (field == PORT_MODE)
+        {
+            /* Placeholder to prevent warning. Not needed to be parsed here.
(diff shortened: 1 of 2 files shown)


bradh352:
test failures appear unrelated to this PR

bradh352:
@prsunny ping

prsunny (approved):

bradh352:
@prsunny please merge

PR #3411, opened 2024-12-05
Fix dash appliance delete error if SAI api is not implemented

mukeshmv:
RemoteRedisSaiInterface wrongly returns a stale SAI object ID even if the underlying SAI API object create returns NOT_IMPLEMENTED. Fix the error caused by saving and using this stale dash appliance object ID during subsequent object delete.


prsunny (approved):

prsunny:
Exception path logic changed. Bypassed coverage as the fix is specific failure case.

PR #3432, opened 2024-12-19
[portmgrd] regression: prevent runtime exception (crash) in configuring portchannel at boot

bradh352:
**What I did**
Prevent setting a default port MTU on PortChannel member ports as it will fail (at least on Dell S5248F) during boot and cause portmgrd to exit.  The current code in portmgr.cpp is setting a default value for an MTU (9100) even when its a PortChannel member, so this patch prevents that default value from being set. 

Also if a user were to incorrectly specify an MTU on a Port that is a member of the port channel via `config_db.json` this too would bring down portmgrd, so catch that and just emit a warning instead (NOTE: the YANG model does NOT support checking/preventing an MTU set on a PORT that is part of a PORTCHANNEL, so this secondary issue should be caught and handled gracefully).

In order to not add much overhead for large port count systems, we are also lazily caching portchannel members (using a local variable to `doTask()` so it is short-lived) and only using that cache on a new port being brought up or on failure to set an MTU.

This code is only called if the port is ***not*** in the `PORT_TABLE`. I'm not aware of any instances where ports are added to `PORT_TABLE` after startup.

During startup, it is expected that `doTask()` will not be invoked per port, but rather receive events for multiple (likely all) ports at once, so I optimized for that case by adding a hashtable cache so each port won't have to pull the list for the db, its pulled at most once per `doTask()` call.

Therefore, I expect the overhead of this patch to be exactly 1 db query per switch startup (***not per port*** and ***not after startup***). Then a very cheap hashtable lookup *once* per port, again, only during startup.

**Why I did it**
The current code *always* attempts to set an MTU on the PORT by setting a default here:
https://github.com/sonic-net/sonic-swss/blob/c20902f3195b5bf8a941045e131aa1b863b69fd0/cfgmgr/portmgr.cpp#L163-L172
Then applies it here:
https://github.com/sonic-net/sonic-swss/blob/c20902f3195b5bf8a941045e131aa1b863b69fd0/cfgmgr/portmgr.cpp#L222-L226

So it isn't crashing because the user configured the MTU in the PORT config, but rather because it is done by default (in portmgr.cpp) when the port is created.   (But it also would crash if a user set an MTU on a port which is bad since YANG doesn't do anything to prevent this).

**NOTE**: this only appears to crash on a freshly loaded config at boot, if you take an existing running configuration and modify it to add a portchannel it works since the port already exists in `PORT_TABLE` so the default MTU setting path isn't taken in the above referenced code.

This regression was caused by 8b99543fa274f083d81f6a07442b267741ceede9 (Oct 2024)... but just reverting that patch isn't the right solution.  The startup logic does not appear to be proper as it tries to set a default MTU regardless if its valid to do so for the port or not.

Logs show this issue which is a critical failure causing the entire switch to go down:
```
2024 Dec 17 19:26:20.964259 sw1 INFO swss#supervisord: portmgrd RTNETLINK answers: Operation not permitted
2024 Dec 17 19:26:20.965353 sw1 ERR swss#portmgrd: :- main: Runtime error: /sbin/ip link set dev "Ethernet0" mtu "9100" : 
2024 Dec 17 19:26:20.967020 sw1 INFO swss#supervisord 2024-12-17 19:26:20,966 WARN exited: portmgrd (exit status 255; not expected)
```

It is possible this won't happen on all switches as I'm guessing this is likely a race condition between PortChannel creation and Port creation since they both listen to the configdb separately.  So at least on my Dell S5248F switches, the PortChannel is triggered first by `teammgrd` before the physical Ports are added to PORT_TABLE by `portmgrd`.

**How I verified it**

Apply patch and verify this config no longer causes crash on Dell S5248F (Broadcom Trident3) during startup/boot.

You will observe the configuration below ***does not*** contain an *mtu* at all, because the primary issue is code internal to SONiC setting a default mtu.

Tested on 202411 and master.

```
{
    "PORT": {
        "Ethernet0": {
            "admin_status": "up",
            "alias": "twentyfiveGigE1/1/1",
            "autoneg": "off",
            "description": "PortChannel1 mgmt",
            "fec": "rs",
            "index": "1",
            "lanes": "49",
            "speed": "25000"
        },
        "Ethernet1": {
            "admin_status": "up",
            "alias": "twentyfiveGigE1/1/2",
            "autoneg": "off",
            "description": "PortChannel1 mgmt",
            "fec": "rs",
            "index": "2",
            "lanes": "50",
            "speed": "25000"
        }
    },
    "PORTCHANNEL": {
        "PortChannel0001": {
            "admin_status": "up",
            "description": "management interface",
            "lacp_key": "auto",
            "min_links": "1"
        }
    },
    "PORTCHANNEL_INTERFACE": {
        "PortChannel0001": {
            "ipv6_use_link_local_only": "enable",
            "mac_addr": "02:d3:ab:fe:fd:c4"
        },
        "PortChannel0001|10.0.0.11/24": {}
    },
    "PORTCHANNEL_MEMBER": {
        "PortChannel0001|Ethernet0": {},
        "PortChannel0001|Ethernet1": {}
    }
}
```

**Details if related**

### DIFF
--- a/cfgmgr/portmgr.cpp
+++ b/cfgmgr/portmgr.cpp
@@ -45,7 +45,14 @@ bool PortMgr::setPortMtu(const string &alias, const string &mtu)
     }
     else
     {
-        throw runtime_error(cmd_str + " : " + res);
+        // This failure can happen on PortChannels during system startup.  A PortChannel enslaves
+        // members before a default MTU is set on the port (set in this file, not via the config!).
+        // Therefore this error is always emitted on startup for portchannel members.
+        // In theory we shouldn't log in this case, the correct fix is to detect the
+        // port is part of a portchannel and not even try this but that is rejected for
+        // possible performance implications.
+        SWSS_LOG_WARN("Setting mtu to alias:%s netdev failed (isPortStateOk=true) with cmd:%s, rc:%d, error:%s", alias.c_str(), cmd_str.c_str(), ret, res.c_str());
+        return false;
     }
     return true;
 }
--- a/cfgmgr/portmgr.cpp
+++ b/cfgmgr/portmgr.cpp
@@ -7,7 +7,6 @@
 #include "exec.h"
 #include "shellcmd.h"
 #include <swss/redisutility.h>
-#include <unordered_map>
 
 using namespace std;
 using namespace swss;
@@ -23,40 +22,7 @@ PortMgr::PortMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
 {
 }
 
-bool PortMgr::isLagMember(const std::string &alias, std::unordered_map<std::string, int> &lagMembers)
-{
-    /* Lag members are lazily loaded on the first call to isLagMember and cached
-     * within a variable inside of doTask() for future calls.
-     */
-    if (lagMembers.empty())
-    {
-        vector<string> keys;
-        m_cfgLagMemberTable.getKeys(keys);
-        for (auto key: keys)
-        {
-            auto tokens = tokenize(key, config_db_key_delimiter);
-            std::string member = tokens[1];
-            if (!member.empty()) {
-                lagMembers[member] = 1;
-            }
-        }
-
-        /* placeholder to state we already read lagmembers even though there are
-         * none */
-        if (lagMembers.empty()) {
-            lagMembers["none"] = 1;
-        }
-    }
-
-    if (lagMembers.find(alias) != lagMembers.end())
-    {
-        return true;
-    }
-
-    return false;
-}
-
-bool PortMgr::setPortMtu(const string &alias, const string &mtu, std::unordered_map<std::string, int> &lagMembers)
+bool PortMgr::setPortMtu(const string &alias, const string &mtu)
 {
     stringstream cmd;
     string res, cmd_str;
@@ -75,12 +41,6 @@ bool PortMgr::setPortMtu(const string &alias, const string &mtu, std::unordered_
     {
         // Can happen when a DEL notification is sent by portmgrd immediately followed by a new SET notif
         SWSS_LOG_WARN("Setting mtu to alias:%s netdev failed with cmd:%s, rc:%d, error:%s", alias.c_str(), cmd_str.c_str(), ret, res.c_str());
-        return false;
-    }
-    else if (isLagMember(alias, lagMembers))
-    {
-        // Catch user improperly specified an MTU on the PortChannel
-        SWSS_LOG_WARN("Setting mtu to alias:%s which is a member of a PortChannel is invalid", alias.c_str());
         return false;
     }
     else
@@ -168,16 +128,9 @@ void PortMgr::doSendToIngressPortTask(Consumer &consumer)
 
 }
 
-
 void PortMgr::doTask(Consumer &consumer)
 {
     SWSS_LOG_ENTER();
-
-    /* Variable to lazily cache lag members upon first call into isLagMember(). We
-     * don't want to always query for lag members if not needed, and we also don't
-     * want to query it for each call to isLagMember() which may be on every port.
-     */
-    std::unordered_map<std::string, int> lagMembers;
 
     auto table = consumer.getTableName();
     if (table == CFG_SEND_TO_INGRESS_PORT_TABLE_NAME)
@@ -203,7 +156,6 @@ void PortMgr::doTask(Consumer &consumer)
             bool portOk = isPortStateOk(alias);
 
             string admin_status, mtu;
-            bool   isMtuSet = false;
             std::vector<FieldValueTuple> field_values;
 
             bool configured = (m_portList.find(alias) != m_portList.end());
@@ -215,6 +167,7 @@ void PortMgr::doTask(Consumer &consumer)
             {
                 admin_status = DEFAULT_ADMIN_STATUS_STR;
                 mtu = DEFAULT_MTU_STR;
+
                 m_portList.insert(alias);
             }
             else if (!portOk)
@@ -228,10 +181,6 @@ void PortMgr::doTask(Consumer &consumer)
                 if (fvField(i) == "mtu")
                 {
                     mtu = fvValue(i);
-                    /* mtu read might be "", so we can't just use .empty() to
-                     * know if its set.  There's test cases that depend on this
-                     * logic ... */
-                    isMtuSet = true;
                 }
                 else if (fvField(i) == "admin_status")
                 {
@@ -241,11 +190,6 @@ void PortMgr::doTask(Consumer &consumer)
                 {
                     field_values.emplace_back(i);
                 }
-            }
-
-            /* Clear default MTU if LAG member as this will otherwise fail. */
-            if (!isMtuSet && isLagMember(alias, lagMembers)) {
-                mtu = "";
             }
 
             if (!portOk)
@@ -277,14 +221,14 @@ void PortMgr::doTask(Consumer &consumer)
 
             if (!mtu.empty())
             {
+                setPortMtu(alias, mtu);
                 SWSS_LOG_NOTICE("Configure %s MTU to %s", alias.c_str(), mtu.c_str());
-                setPortMtu(alias, mtu, lagMembers);
             }
 
             if (!admin_status.empty())
             {
-                SWSS_LOG_NOTICE("Configure %s admin status to %s", alias.c_str(), admin_status.c_str());
                 setPortAdminStatus(alias, admin_status == "up");
+                SWSS_LOG_NOTICE("Configure %s admin status to %s", alias.c_str(), admin_status.c_str());
             }
         }
         else if (op == DEL_COMMAND)
--- a/cfgmgr/portmgr.cpp
+++ b/cfgmgr/portmgr.cpp
@@ -37,18 +37,17 @@ bool PortMgr::setPortMtu(const string &alias, const string &mtu)
         // the port MTU and possibly the port based router interface MTU
         return writeConfigToAppDb(alias, "mtu", mtu);
     }
-
-    // Failure happens on PortChannels during system startup.  PortChannel enslaves
-    // members before a default MTU on the port (set in this file, not via the config!).
-    // Therefore this error is always emitted on startup for portchannel members.
-    // In theory we shouldn't log in this case, the correct fix is to detect the
-    // port is part of a portchannel and not even try this but that is rejected for
-    // possible performance implications.
-    //
-    // This can also happen when a DEL notification is sent by portmgrd
-    // immediately followed by a new SET notif
-    SWSS_LOG_WARN("Setting mtu to alias:%s netdev failed with cmd:%s, rc:%d, error:%s", alias.c_str(), cmd_str.c_str(), ret, res.c_str());
-    return false;
+    else if (!isPortStateOk(alias))
+    {
+        // Can happen when a DEL notification is sent by portmgrd immediately followed by a new SET notif
+        SWSS_LOG_WARN("Setting mtu to alias:%s netdev failed with cmd:%s, rc:%d, error:%s", alias.c_str(), cmd_str.c_str(), ret, res.c_str());
+        return false;
+    }
+    else
+    {
+        throw runtime_error(cmd_str + " : " + res);
+    }
+    return true;
 }
 
 bool PortMgr::setPortAdminStatus(const string &alias, const bool up)
--- a/cfgmgr/portmgr.cpp
+++ b/cfgmgr/portmgr.cpp
@@ -7,6 +7,7 @@
 #include "exec.h"
 #include "shellcmd.h"
 #include <swss/redisutility.h>
+#include <unordered_map>
 
 using namespace std;
 using namespace swss;
@@ -22,7 +23,40 @@ PortMgr::PortMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
 {
 }
 
-bool PortMgr::setPortMtu(const string &alias, const string &mtu)
+bool PortMgr::isLagMember(const std::string &alias, std::unordered_map<std::string, int> &lagMembers)
+{
+    /* Lag members are lazily loaded on the first call to isLagMember and cached
+     * within a variable inside of doTask() for future calls.
+     */
+    if (lagMembers.empty())
+    {
+        vector<string> keys;
+        m_cfgLagMemberTable.getKeys(keys);
+        for (auto key: keys)
+        {
+            auto tokens = tokenize(key, config_db_key_delimiter);
+            std::string member = tokens[1];
+            if (!member.empty()) {
+                lagMembers[member] = 1;
+            }
+        }
+
+        /* placeholder to state we already read lagmembers even though there are
+         * none */
+        if (lagMembers.empty()) {
+            lagMembers["none"] = 1;
+        }
+    }
+
+    if (lagMembers.find(alias) != lagMembers.end())
+    {
+        return true;
+    }
+
+    return false;
+}
+
+bool PortMgr::setPortMtu(const string &alias, const string &mtu, std::unordered_map<std::string, int> &lagMembers)
 {
     stringstream cmd;
     string res, cmd_str;
@@ -41,6 +75,12 @@ bool PortMgr::setPortMtu(const string &alias, const string &mtu)
     {
         // Can happen when a DEL notification is sent by portmgrd immediately followed by a new SET notif
         SWSS_LOG_WARN("Setting mtu to alias:%s netdev failed with cmd:%s, rc:%d, error:%s", alias.c_str(), cmd_str.c_str(), ret, res.c_str());
+        return false;
+    }
+    else if (isLagMember(alias, lagMembers))
+    {
+        // Catch user improperly specified an MTU on the PortChannel
+        SWSS_LOG_WARN("Setting mtu to alias:%s which is a member of a PortChannel is invalid", alias.c_str());
         return false;
     }
     else
@@ -128,9 +168,16 @@ void PortMgr::doSendToIngressPortTask(Consumer &consumer)
 
 }
 
+
 void PortMgr::doTask(Consumer &consumer)
 {
     SWSS_LOG_ENTER();
+
+    /* Variable to lazily cache lag members upon first call into isLagMember(). We
+     * don't want to always query for lag members if not needed, and we also don't
+     * want to query it for each call to isLagMember() which may be on every port.
+     */
+    std::unordered_map<std::string, int> lagMembers;
 
     auto table = consumer.getTableName();
     if (table == CFG_SEND_TO_INGRESS_PORT_TABLE_NAME)
@@ -156,6 +203,7 @@ void PortMgr::doTask(Consumer &consumer)
             bool portOk = isPortStateOk(alias);
 
             string admin_status, mtu;
+            bool   isMtuSet = false;
             std::vector<FieldValueTuple> field_values;
 
             bool configured = (m_portList.find(alias) != m_portList.end());
@@ -167,7 +215,6 @@ void PortMgr::doTask(Consumer &consumer)
             {
                 admin_status = DEFAULT_ADMIN_STATUS_STR;
                 mtu = DEFAULT_MTU_STR;
-
                 m_portList.insert(alias);
             }
             else if (!portOk)
@@ -181,6 +228,10 @@ void PortMgr::doTask(Consumer &consumer)
                 if (fvField(i) == "mtu")
                 {
                     mtu = fvValue(i);
+                    /* mtu read might be "", so we can't just use .empty() to
+                     * know if its set.  There's test cases that depend on this
+                     * logic ... */
+                    isMtuSet = true;
                 }
                 else if (fvField(i) == "admin_status")
                 {
@@ -190,6 +241,11 @@ void PortMgr::doTask(Consumer &consumer)
                 {
                     field_values.emplace_back(i);
                 }
+            }
+
+            /* Clear default MTU if LAG member as this will otherwise fail. */
+            if (!isMtuSet && isLagMember(alias, lagMembers)) {
+                mtu = "";
             }
 
             if (!portOk)
@@ -221,14 +277,14 @@ void PortMgr::doTask(Consumer &consumer)
 
             if (!mtu.empty())
             {
-                setPortMtu(alias, mtu);
                 SWSS_LOG_NOTICE("Configure %s MTU to %s", alias.c_str(), mtu.c_str());
+                setPortMtu(alias, mtu, lagMembers);
             }
 
             if (!admin_status.empty())
             {
-                setPortAdminStatus(alias, admin_status == "up");
                 SWSS_LOG_NOTICE("Configure %s admin status to %s", alias.c_str(), admin_status.c_str());
+                setPortAdminStatus(alias, admin_status == "up");
             }
         }
         else if (op == DEL_COMMAND)
--- a/cfgmgr/portmgr.cpp
+++ b/cfgmgr/portmgr.cpp
@@ -37,17 +37,18 @@ bool PortMgr::setPortMtu(const string &alias, const string &mtu)
         // the port MTU and possibly the port based router interface MTU
         return writeConfigToAppDb(alias, "mtu", mtu);
     }
-    else if (!isPortStateOk(alias))
-    {
-        // Can happen when a DEL notification is sent by portmgrd immediately followed by a new SET notif
-        SWSS_LOG_WARN("Setting mtu to alias:%s netdev failed with cmd:%s, rc:%d, error:%s", alias.c_str(), cmd_str.c_str(), ret, res.c_str());
-        return false;
-    }
-    else
-    {
-        throw runtime_error(cmd_str + " : " + res);
-    }
-    return true;
+
+    // Failure happens on PortChannels during system startup.  PortChannel enslaves
+    // members before a default MTU on the port (set in this file, not via the config!).
+    // Therefore this error is always emitted on startup for portchannel members.
+    // In theory we shouldn't log in this case, the correct fix is to detect the
+    // port is part of a portchannel and not even try this but that is rejected for
+    // possible performance implications.
+    //
+    // This can also happen when a DEL notification is sent by portmgrd
+    // immediately followed by a new SET notif
+    SWSS_LOG_WARN("Setting mtu to alias:%s netdev failed with cmd:%s, rc:%d, error:%s", alias.c_str(), cmd_str.c_str(), ret, res.c_str());
+    return false;
 }
 
 bool PortMgr::setPortAdminStatus(const string &alias, const bool up)
--- a/.azure-pipelines/build-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/build-docker-sonic-vs-template.yml
@@ -21,6 +21,9 @@ parameters:
   type: string
   default: '$(BUILD_BRANCH)'
 
+- name: debian_version
+  type: string
+
 - name: artifact_name
   type: string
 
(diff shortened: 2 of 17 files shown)


bradh352:
@prsunny please review

cfgmgr/portmgr.cpp
```diff
@@ -22,6 +22,28 @@ PortMgr::PortMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
 {
 }
 
+bool PortMgr::isLag(const std::string &port)
```
dgsudharsan, line 25: Can we rename this as isLagMember?

dgsudharsan (changes requested):
Please add UT to cover this scenario.

dgsudharsan:
@bradh352 We do have this check at CLI level https://github.com/sonic-net/sonic-utilities/blob/80d469886f120bfe9bc60024f608c039dce06646/config/main.py#L4948

Why do we need such checks at multiple places? @prsunny  what are your thoughts on this?

bradh352:
> @bradh352 We do have this check at CLI level https://github.com/sonic-net/sonic-utilities/blob/80d469886f120bfe9bc60024f608c039dce06646/config/main.py#L4948
> 
> Why do we need such checks at multiple places? @prsunny what are your thoughts on this?

People using things like Ansible, don't use the CLI to set configuration.  They modify the `/etc/sonic/config_db.json` which does nothing to prevent this.  ALSO, in this case, as you can see from the `/etc/sonic/config_db.json` example I provided, no MTU is provided at all in the interface configuration.  Its being autopopulated somewhere as a default.  I didn't try to track that down.

bradh352:
> Please add UT to cover this scenario.

I committed one, no idea if its right.

bradh352:
coverage looks good, any other comments?

prsunny:
Agree with @dgsudharsan , this is checked in the CLI and Yang level. I dont think its really needed in swss.

bradh352:
@prsunny
> Agree with @dgsudharsan , this is checked in the CLI and Yang level. I dont think its really needed in swss.

Your evaluation is not correct.

First, its not occurring at the yang level as per:
https://github.com/sonic-net/sonic-buildimage/blob/d28b734ce545762cd7b49d110575eec8549458cf/src/sonic-yang-models/yang-models/sonic-port.yang#L162-L166
There are no restrictions on MTU being set for a port.

But that's besides the point.  My configuration that triggered this behavior, as can be seen in the PR summary, did not have an MTU set at all on the port:
```
"PORT": {
        "Ethernet0": {
            "admin_status": "up",
            "alias": "twentyfiveGigE1/1/1",
            "autoneg": "off",
            "description": "PortChannel1 mgmt",
            "fec": "rs",
            "index": "1",
            "lanes": "49",
            "speed": "25000"
        },
       ...
}
```

That means its getting inherited somewhere (I didn't look where), triggering it to call `/sbin/ip link set dev "Ethernet0" mtu "9100"`

bradh352:
Ok, found where its being set by default to 9100:

https://github.com/sonic-net/sonic-swss/blob/c20902f3195b5bf8a941045e131aa1b863b69fd0/cfgmgr/portmgr.cpp#L163-L172

Which is right above:

https://github.com/sonic-net/sonic-swss/blob/c20902f3195b5bf8a941045e131aa1b863b69fd0/cfgmgr/portmgr.cpp#L222-L226

Which is where it crashes due to the exception.

The macro for the default value is here:

https://github.com/sonic-net/sonic-swss/blob/c20902f3195b5bf8a941045e131aa1b863b69fd0/cfgmgr/portmgr.h#L15

There's no way that any CLI or Yang verification can prevent a hardcoded default in the code path.

prsunny:
> Ok, found where its being set by default to 9100, a hardcoded code path:
> 
> https://github.com/sonic-net/sonic-swss/blob/c20902f3195b5bf8a941045e131aa1b863b69fd0/cfgmgr/portmgr.cpp#L163-L172
> 
> Which is right above:
> 
> https://github.com/sonic-net/sonic-swss/blob/c20902f3195b5bf8a941045e131aa1b863b69fd0/cfgmgr/portmgr.cpp#L222-L226
> 
> Which is where it crashes due to the exception caused by `setPortMtu()`.
> 
> The macro for the default value is here:
> 
> https://github.com/sonic-net/sonic-swss/blob/c20902f3195b5bf8a941045e131aa1b863b69fd0/cfgmgr/portmgr.h#L15
> 
> There's no way that any CLI or Yang verification can prevent a hardcoded default in the code path.

Not clear I understand. This code path is legacy and we don't see an issue so far. So what exactly is triggering this?

bradh352:
> > Ok, found where its being set by default to 9100, a hardcoded code path:
> > https://github.com/sonic-net/sonic-swss/blob/c20902f3195b5bf8a941045e131aa1b863b69fd0/cfgmgr/portmgr.cpp#L163-L172
> > 
> > Which is right above:
> > https://github.com/sonic-net/sonic-swss/blob/c20902f3195b5bf8a941045e131aa1b863b69fd0/cfgmgr/portmgr.cpp#L222-L226
> > 
> > Which is where it crashes due to the exception caused by `setPortMtu()`.
> > The macro for the default value is here:
> > https://github.com/sonic-net/sonic-swss/blob/c20902f3195b5bf8a941045e131aa1b863b69fd0/cfgmgr/portmgr.h#L15
> > 
> > There's no way that any CLI or Yang verification can prevent a hardcoded default in the code path.
> 
> Not sure I understand. This code path is legacy and we don't see an issue so far. So what exactly is triggering this?

Well, I can 100% confirm without this patch, trying to bring up a portchannel interface with the config in the summary on a Dell S5248F (Broadcom Trident 3) fails due to the aforementioned error of:
```
2024 Dec 17 19:26:20.964259 sw1 INFO swss#supervisord: portmgrd RTNETLINK answers: Operation not permitted
2024 Dec 17 19:26:20.965353 sw1 ERR swss#portmgrd: :- main: Runtime error: /sbin/ip link set dev "Ethernet0" mtu "9100" : 
2024 Dec 17 19:26:20.967020 sw1 INFO swss#supervisord 2024-12-17 19:26:20,966 WARN exited: portmgrd (exit status 255; not expected)
```

Once portmgrd exits, its all over, the switch is down on both 202411 and master.  This is observed on a fresh boot with the configuration, I did observe that if adding the ports to a portchannel with a switch already running the error does not occur, presumable some order of operations aspect of bringing up a portchannel at boot vs configuring it at runtime.

I haven't tried portchannel on any other releases.  I'm assuming this is a recent issue with the 'ip link' command returning failure these days where as previously maybe it was accepted ... or maybe there was a change the order in the way portchannels vs ports are brought up during boot. I can't say.

Either way this PR corrects the issue.

bradh352:
@prsunny any other comments on this?  Did you want any changes to the PR?  Or can it be merged?

prsunny:
My concern is, each time you configure port, it goes through this Config_DB validation of lagmember causing atleast some delays. If this can be prevented during the config itself, prefer to handle it outside of swss. @dgsudharsan , let me know what you think.

dgsudharsan:
@prsunny Completely agree with your concern. We are introducing systems with 256 and 512 ports. The user configurations should come through click commands since they have validations. If the user introduces config through config_db bypassing CLI checks its user's responsibility to validate. Introducing this check will have impact on boot time and this is something we would like to avoid if not necessary.

bradh352:
@prsunny @dgsudharsan but the issue is not with ***user*** config, its with the ***default*** setting.  In this case the user ***did*** ***not*** set an MTU at all.

prsunny:
> By the way, I think the regression that caused this is [8b99543](https://github.com/sonic-net/sonic-swss/commit/8b99543fa274f083d81f6a07442b267741ceede9) committed in Oct 2024... I need to do some testing to confirm. Timeline matches and its saving an MTU that the user never specified.

thanks. We are not aware of a regression from the tests. It will be good if you can test and confirm.

bradh352:
I just reverted this patch + 8b99543 in my own branch and can confirm that commit is the cause of the regression.

I don't think just reverting 8b99543 is the proper solution though. 

@liuh-80 you wrote this patch, can you comment?

I'm not familiar enough with the startup sequence involved here, or why `mtu` and `admin_status` are excluded from being inserted into field_values in the first place from the fields passed in (in the for loop):
https://github.com/sonic-net/sonic-swss/blob/4eb74f0082f0f8c4537fe58621ac902c870d217c/cfgmgr/portmgr.cpp#L179-L202

bradh352:
I just updated this patch to try to address the expressed concerns and take into account the changes introduced in 8b99543fa274f083d81f6a07442b267741ceede9 to hopefully not regress that.  I'm in process of building and testing this so it may not be in a final state.

That said, I will take any feedback you have to offer if this is going in an acceptable direction or not.

bradh352:
I've installed this patch on a physical Dell S5248F and verified it works.  However, it appears some test cases are being run by Azure CI are failing.  I guess I need to look at them.

bradh352:
i've made as minimal a patch as possible while still passing all test cases now.  please review.  hopefully this is acceptable with the lazy caching of portchannel members.

bradh352:
@prsunny @dgsudharsan does the current change sufficiently address your performance concerns?

prsunny:
> @prsunny @dgsudharsan does the current change sufficiently address your performance concerns?

I still see this as a gap for dynamic lag members which are created after the config_db read. Can you confirm if you can modify the config_db (may be as a workaround) to have the correct values as expected and it works?

bradh352:
> > @prsunny @dgsudharsan does the current change sufficiently address your performance concerns?
> 
> I still see this as a gap for dynamic lag members which are created after the config_db read. Can you confirm if you can modify the config_db (may be as a workaround) to have the correct values as expected and it works?

I'm not sure I understand what you are saying.

The issue has nothing to do with config_db.json at all, it is a regression affecting all portchannels brought up during boot (at least on real hardware as of Oct 2024).  This is reproduced with *no* mtu specified at all in the config_db.json, because a default of `9100` is set in the c++ code of portmgr, this is not a value read from the config at all.  

This is not observed in the current upstream code after the initial port bring-up, such as if adding a portchannel after boot because the code to set the default mtu is NOT called in that case.  So yes, you have to reboot the switch with your saved configuration having a portchannel configured to observe the issue, but that is the only way in which it relates to config_db.json.

If you look at the minimal changeset in doTask(), I tried to make the change as comprehensible as possible there.  The main fix there is to unset the MTU specified earlier ONLY IF it was ***not*** specified in the port configuration *and* it is a LAG Member.

bradh352:
> > > @prsunny @dgsudharsan does the current change sufficiently address your performance concerns?
> > 
> > 
> > I still see this as a gap for dynamic lag members which are created after the config_db read. Can you confirm if you can modify the config_db (may be as a workaround) to have the correct values as expected and it works?
> 
> I'm not sure I understand what you are saying.
> 
> The issue has nothing to do with config_db.json at all, it is a regression affecting all portchannels brought up during boot (at least on real hardware as of Oct 2024 due to [8b99543](https://github.com/sonic-net/sonic-swss/commit/8b99543fa274f083d81f6a07442b267741ceede9)). This is reproduced with _no_ mtu specified at all in the config_db.json, because a default of `9100` is set in the c++ code of portmgr, this is not a value read from the config at all.
> 
> This is not observed in the current upstream code after the initial port bring-up, such as if adding a portchannel after boot because the code to set the default mtu is NOT called in that case. So yes, you have to reboot the switch with your saved configuration having a portchannel configured to observe the issue, but that is the only way in which it relates to config_db.json.
> 
> If you look at the minimal changeset in doTask(), I tried to make the change as comprehensible as possible there. The main fix there is to unset the MTU specified earlier ONLY IF it was _**not**_ specified in the port configuration _and_ it is a LAG Member.

@prsunny please reply or approve

bradh352:
> > > > @prsunny @dgsudharsan does the current change sufficiently address your performance concerns?
> > > 
> > > 
> > > I still see this as a gap for dynamic lag members which are created after the config_db read. Can you confirm if you can modify the config_db (may be as a workaround) to have the correct values as expected and it works?
> > 
> > 
> > I'm not sure I understand what you are saying.
> > The issue has nothing to do with config_db.json at all, it is a regression affecting all portchannels brought up during boot (at least on real hardware as of Oct 2024 due to [8b99543](https://github.com/sonic-net/sonic-swss/commit/8b99543fa274f083d81f6a07442b267741ceede9)). This is reproduced with _no_ mtu specified at all in the config_db.json, because a default of `9100` is set in the c++ code of portmgr, this is not a value read from the config at all.
> > This is not observed in the current upstream code after the initial port bring-up, such as if adding a portchannel after boot because the code to set the default mtu is NOT called in that case. So yes, you have to reboot the switch with your saved configuration having a portchannel configured to observe the issue, but that is the only way in which it relates to config_db.json.
> > If you look at the minimal changeset in doTask(), I tried to make the change as comprehensible as possible there. The main fix there is to unset the MTU specified earlier ONLY IF it was _**not**_ specified in the port configuration _and_ it is a LAG Member.
> 
> @prsunny please reply or approve

@dgsudharsan can you also reply or hopefully approve given my description?

bradh352:
> > > > > @prsunny @dgsudharsan does the current change sufficiently address your performance concerns?
> > > > 
> > > > 
> > > > I still see this as a gap for dynamic lag members which are created after the config_db read. Can you confirm if you can modify the config_db (may be as a workaround) to have the correct values as expected and it works?
> > > 
> > > 
> > > I'm not sure I understand what you are saying.
> > > The issue has nothing to do with config_db.json at all, it is a regression affecting all portchannels brought up during boot (at least on real hardware as of Oct 2024 due to [8b99543](https://github.com/sonic-net/sonic-swss/commit/8b99543fa274f083d81f6a07442b267741ceede9)). This is reproduced with _no_ mtu specified at all in the config_db.json, because a default of `9100` is set in the c++ code of portmgr, this is not a value read from the config at all.
> > > This is not observed in the current upstream code after the initial port bring-up, such as if adding a portchannel after boot because the code to set the default mtu is NOT called in that case. So yes, you have to reboot the switch with your saved configuration having a portchannel configured to observe the issue, but that is the only way in which it relates to config_db.json.
> > > If you look at the minimal changeset in doTask(), I tried to make the change as comprehensible as possible there. The main fix there is to unset the MTU specified earlier ONLY IF it was _**not**_ specified in the port configuration _and_ it is a LAG Member.
> > 
> > 
> > @prsunny please reply or approve
> 
> @dgsudharsan can you also reply or hopefully approve given my description?

any update here @prsunny @dgsudharsan ?

dgsudharsan:
@bradh352 Are you able to hit this issue 100% with the method you mentioned? We have lot of regression running on t1-lag topologies and we execute all forms of reboot but never hit this. The only theoretical way in which it could be hit is the port gets added to port-channel and lag interface is created in kernel even before its settings are set. @liuh-80 Can you please confirm if you see any issues after your PR was merged?

cfgmgr/portmgr.cpp
```diff
@@ -22,7 +23,40 @@ PortMgr::PortMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
...
-bool PortMgr::setPortMtu(const string &alias, const string &mtu)
+bool PortMgr::isLagMember(const std::string &alias, std::unordered_map<std::string, int> &lagMembers)
+{
+    /* Lag members are lazily loaded on the first call to isLagMember and cached
+     * within a variable inside of doTask() for future calls.
+     */
+    if (lagMembers.empty())
+    {
+        vector<string> keys;
+        m_cfgLagMemberTable.getKeys(keys);
```
dgsudharsan, line 34: This will not handle if lag member gets added after bootup.

bradh352:
@dgsudharsan

> @bradh352 Are you able to hit this issue 100% with the method you mentioned? We have lot of regression running on t1-lag topologies and we execute all forms of reboot but never hit this. The only theoretical way in which it could be hit is the port gets added to port-channel and lag interface is created in kernel even before its settings are set. @liuh-80 Can you please confirm if you see any issues after your PR was merged?

Yes, every single boot, 100% failure rate on Broadcom Trident3-X7 (Dell S5248F).  If I configure a lag post-boot, it does not cause a crash.  This fixes the crash at boot issue.

cfgmgr/portmgr.cpp
bradh352: doTask() isn't a long-running process and the cache is local to doTask() and not global, so it doesn't survive past a single call to doTask().  The only thing the cache really helps with is on boot where a large number of ports are being brought up so it will have more than 1 task to iterate across in doTask() ... I'd assume in most instance doTask() will only have a single task outside of bootup (I could be wrong here, I haven't tried to trace it).

dgsudharsan: doTask is called for every update from config_db. So the cache would be outdated if someone creates a portchannel using CLI.

bradh352: No, that's incorrect. 

The cache is a local variable to doTask().  Cache is destroyed at the end of each doTask call because the cache variable goes out of scope at the end of the function call.

So subsequent calls would create a new temporary cache. its not persistent from call to call.

Unless I'm seriously missing something here, but it looks like doTask() is a short-lived process.  It loops through an iterator processing events and exits ... just to be called again ... just looking at the main() loop for portmgrd.cpp its not expected to block for the life of the daemon or anything like that.

bradh352:
> @dgsudharsan
> 
> > @bradh352 Are you able to hit this issue 100% with the method you mentioned? We have lot of regression running on t1-lag topologies and we execute all forms of reboot but never hit this. The only theoretical way in which it could be hit is the port gets added to port-channel and lag interface is created in kernel even before its settings are set. @liuh-80 Can you please confirm if you see any issues after your PR was merged?
> 
> Yes, every single boot, 100% failure rate on Broadcom Trident3-X7 (Dell S5248F). If I configure a lag post-boot, it does not cause a crash. This fixes the crash at boot issue.

ping

cfgmgr/portmgr.cpp
prsunny: This could have performance issues with scaled up lag member and each time traversing through the list. I think for your specific issue, we could follow how this is done currently in sonic-mgmt tests as a work-around, to configure the lag

bradh352: what do you mean by "scaled up lag member"? 

This code is only called if the port is NOT in the PORT_TABLE.  I'm not aware of any instances where ports are added to PORT_TABLE after startup.  

During startup, it is expected that doTask() will not be invoked per port, but rather receive events for multiple (likely ***all***) ports at once, so I optimized for that case by adding a hashtable cache each port won't have to pull the list for the db, its pulled at most once per doTask() call.

Therefore, I expect the overhead of this patch to be exactly 1 db query per startup.  Then a very cheap hashtable lookup once per port.

bradh352:
@stepanblyschak you seem to do a bit on portsorch bring-up, can you look at this PR to fix a regression in portchannel bring-up that causes an exception?

stepanblyschak:
@bradh352 As I understand it, you enounter a crash that reveals design issues of kernel configuration of ports/lags - port and lag kernel configuration are inherently dependent but portmgrd and teammgrd are running completelly unaware of each other. The ```mtu``` field of the port becomes meaningless when port is enslaved as it inherits lag mtu, but is used later in teammgrd when removing a lag member to restore port mtu to its original value.
Given solution to check lag member configuration is still prone to data races post boot.
On the other hand, you seem to be more concerned about the crash behaviour, how about not crash when ```ip``` command fails, instead, report an error in syslog? A correct mtu value will be restored by teammgrd if user removes lag member configuration.

bradh352:
@stepanblyschak 
> @bradh352 As I understand it, you enounter a crash that reveals design issues of kernel configuration of ports/lags - port and lag kernel configuration are inherently dependent but portmgrd and teammgrd are running completelly unaware of each other. The `mtu` field of the port becomes meaningless when port is enslaved as it inherits lag mtu, but is used later in teammgrd when removing a lag member to restore port mtu to its original value. Given solution to check lag member configuration is still prone to data races post boot. On the other hand, you seem to be more concerned about the crash behaviour, how about not crash when `ip` command fails, instead, report an error in syslog? A correct mtu value will be restored by teammgrd if user removes lag member configuration.

First, there is no race condition here post-boot.  The default MTU value is only set if the port didn't previously exist in the PORT table, so that's a boot-time thing, not a post-boot thing.  That said, I could see during a quick review where that's not obvious due to the github default context it outputs when viewing the diff, you need to expand and you'll see the MTU is only set in the `if (!configured)` codepath.

This patch also includes the logic you suggest, where it won't cause portmgrd to exit if that command fails (which at this point is only if the user explicitly configured an MTU on a port belonging to a port channel which they should not be doing).

I'm a big advocate that logs should not contain warnings or errors that users should NOT be looking at.  Anything that could cause confusion while debugging a real issue should not be introduced.  So simply doing your suggestion is quite honestly not acceptable.

stepanblyschak:
> @stepanblyschak
> 
> > @bradh352 As I understand it, you enounter a crash that reveals design issues of kernel configuration of ports/lags - port and lag kernel configuration are inherently dependent but portmgrd and teammgrd are running completelly unaware of each other. The `mtu` field of the port becomes meaningless when port is enslaved as it inherits lag mtu, but is used later in teammgrd when removing a lag member to restore port mtu to its original value. Given solution to check lag member configuration is still prone to data races post boot. On the other hand, you seem to be more concerned about the crash behaviour, how about not crash when `ip` command fails, instead, report an error in syslog? A correct mtu value will be restored by teammgrd if user removes lag member configuration.
> 
> First, there is no race condition here post-boot. The default MTU value is only set if the port didn't previously exist in the PORT table, so that's a boot-time thing, not a post-boot thing. That said, I could see during a quick review where that's not obvious due to the github default context it outputs when viewing the diff, you need to expand and you'll see the MTU is only set in the `if (!configured)` codepath.
> 
> This patch also includes the logic you suggest, where it won't cause portmgrd to exit if that command fails (which at this point is only if the user explicitly configured an MTU on a port belonging to a port channel which they should not be doing).
> 
> I'm a big advocate that logs should not contain warnings or errors that users should NOT be looking at. Anything that could cause confusion while debugging a real issue should not be introduced. So simply doing your suggestion is quite honestly not acceptable.

Ports can be added at runtime in case of port breakout/split.

Post boot race I can think of is:
1. User writes to ```PORT``` table a new port
2. ```portmgrd``` reads ```LAG_MEMBER``` table and ```isLagMember``` returns false.
3. User writes to ```LAG_MEMBER``` table
4. ```teammgrd``` enslaves port in a lag
5.  ```portmgrd``` executes ```ip link set mtu``` and crash

And by the user here I mean a CLI or a script that does the update non atomically and fast enough to trigger a race.

Regarding a case when mtu is set by the user - I do not fully understand what the user should do if port is already configured with mtu and he wants to add port to a LAG? There's no such thing as "unset" mtu. You can enforce mtu on a port to match mtu on a LAG in CLI but you can't block user from adding a port to a lag if it has mtu configured nor crash in that case either.

I fully agree with your opinion about the log. My point is a bit more general - there's no reason for ```portmgrd``` to crash when ```ip``` command fails. A failure to set port mtu or admin state in the kernel does not invalidate ```portmgrd```'s state and so I think the daemon just needs to continue being operational and handle configs for other ports.

So, I mean there are actually two different issues:
1. ```portmgrd``` takes down the whole system (when autorestart is enabled) when failing to execute ```ip``` command
2. ```portmgrd``` assigns mtu on lag members when it makes no sense to do so

Does solving # 1 unblocks you from deploying your systems?

bradh352:
@stepanblyschak
> So, I mean there are actually two different issues:
> 
> 1. `portmgrd` takes down the whole system (when autorestart is enabled) when failing to execute `ip` command
> 2. `portmgrd` assigns mtu on lag members when it makes no sense to do so
> 
> Does solving # 1 unblocks you from deploying your systems?

Yes, #1 would unblock deployment, but will have unwanted log messages.

bradh352:
> @stepanblyschak
> 
> > So, I mean there are actually two different issues:
> > 
> > 1. `portmgrd` takes down the whole system (when autorestart is enabled) when failing to execute `ip` command
> > 2. `portmgrd` assigns mtu on lag members when it makes no sense to do so
> > 
> > Does solving # 1 unblocks you from deploying your systems?
> 
> Yes, # 1 would unblock deployment, but will have unwanted log messages.

What's the guidance here on this issue?

stepanblyschak:
tag @prsunny to take the final decision

prsunny:
@bradh352 , as discussed in the review, this adds delays in bootup and can potentially impact warmboot/fastboot. Also, this issue is not reported in sonic-mgmt tests or from community. Please check the config model in sonic-mgmt and verify if its aligned to your config generation model. Also, see if this can be handled outside of swss, like in CLI.

bradh352:
> @bradh352 , as discussed in the review, this adds delays in bootup and can potentially impact warmboot/fastboot. Also, this issue is not reported in sonic-mgmt tests. Please check the config model in sonic-mgmt and verify if its aligned to your config generation model. Also, see if this can be handled outside of swss, like in CLI.

@prsunny Your response is incorrect and shows you haven't taken the time to properly review the comments and scenarios in which this occurs.

I can provide you remote access to a real switch if you need it to prove this is an issue.  Again, this can't be handled by the CLI or yang because the mtu setting causing the issue is in C++ code as a ***default***.  I really don't know how else to explain it.

This is a ***REGRESSION*** introduced in 8b99543 (Oct 2024).

The fact that the sonic-mgmt test cases don't happen to trigger this isn't relevant in any form what-so-ever, this is a crash on switch **startup** with PortChannels on real hardware. Likely sonic-mgmt isn't testing this scenario as it is not restarting the vSonic image for this.  This will NOT happen when configured post-startup.  And the fact that you say this isn't reported by the community is odd considering **I AM REPORTING THIS FROM THE COMMUNITY**.  Plus how much of the community is running 202411 or later at this point in time with PortChannels?

I'm sure you can feel my frustration at this point.

I need to understand what would be an acceptable fix to this **CRASH**.

@stepanblyschak can you help?  @prgeor?  Anyone, please help.

bradh352:
> @prsunny , maintaniers, please help make a decision on this PR. Thanks!

@gechiang anyone else that can take the time to review this?  I can provide remote access to real hardware with a minimal config_db.json where you can reboot the switch and observe it crashes like I say it does on current master. I mean sure, everyone should be L3 these days ... but for those that aren't unless we're removing all L2 support like portchannels completely this shouldn't break.

prsunny:
I'm referring to this function during bootup which is incomplete for a dynamic LAG config - `isLagMember`. If you are saying its introduced by a commit, why not suggest reverting that commit whereas I see a different approach to handle the issue. Regarding "The fact that the sonic-mgmt test cases don't happen to trigger this isn't relevant in any form what-so-ever", I disagree that it is not relevant as sonic-mgmt is run on real hardware and should have reload tests with portchannels. If there is a gap, we should be addressing the tests as well.

bradh352:
> I'm referring to this function during bootup which is incomplete for a dynamic LAG config - `isLagMember`. If you are saying its introduced by a commit, why not suggest reverting that commit whereas I see a different approach to handle the issue. Regarding "The fact that the sonic-mgmt test cases don't happen to trigger this isn't relevant in any form what-so-ever", I disagree that it is not relevant as sonic-mgmt is run on real hardware and should have reload tests with portchannels. If there is a gap, we should be addressing the tests as well.

The commit the caused the regression fixed a different real issue, so reverting the commit also isn't the proper solution.

I'm not yet familiar enough with the deep internals of sonic-mgmt to write a test case to reproduce this, if this could be triggered by a simple config change to a running config it would be easy, but the fact that it depends on boot-time config reading to reproduce I am not sure how to write such a test case.

If I modify this to only check if the port is part of a portchannel ***on failure of setting the MTU***, would that be acceptable?  There would be zero performance hit EXCEPT during a failure that would otherwise throw an exception and ***crash***.

prsunny:
To unblock the PR, suggest catch the exception and log a `WARN `message as the crash scenario is not affecting MTU functionality.

bradh352:
@prsunny is the current commit what you want?

prsunny:
> @prsunny is the current commit what you want?

Not really, keep the old code as-is and log WARN instead of 'throw'

bradh352:
@prsunny ok, figured less code was better by removing another conditional that is irrelevant, but if you want it there, so be it.  Please see latest commit.

stepanblyschak (approved):

dgsudharsan (approved):

prsunny:
Suggest following Sonic github community guidelines for commit messages as reviewers can have better understanding of the code that's added/removed in each commit.

dgsudharsan (commented):
Please modify commit messages as suggested.

bhouse-nexthop:
> Suggest following Sonic github community guidelines for commit messages as reviewers can have better understanding of the code that's added/removed in each commit.

Do you just want them all squashed into 1 commit with the relevant message?  The first commit in the series has a proper commit message, but it is no longer relevant to the current overall patch set.  The rest of the commits obviously do not have properly formed messages since I was more concerned with coming up with something that was acceptable to you from a code standpoint.

prsunny:
> > Suggest following Sonic github community guidelines for commit messages as reviewers can have better understanding of the code that's added/removed in each commit.
> 
> Do you just want them all squashed into 1 commit with the relevant message? The first commit in the series has a proper commit message, but it is no longer relevant to the current overall patch set. The rest of the commits obviously do not have properly formed messages since I was more concerned with coming up with something that was acceptable to you from a code standpoint.

Leave it for this PR and address for future PRs to have meaningful commits. Will merge once the PR checkers pass.

prsunny:
Bypassed coverage temporarily as this is an exception path.

bradh352:
Please tag for backport to 202411 as well, or I can make a PR for that if needed.

PR #3435, opened 2024-12-20
Update buffermgrdyn.cpp

marwanelmessiry72:
**What I did**
i added if condetion that will ensure that the poolobj.direction will not size of the string m_bufferZeroProfileName so that the string will not get out of boundaries
**Why I did it**
as the checkmarx tool was telling this as high severity valunerability in the code 
**How I verified it**
added the if condition and regenerated a repoert and the valunerability disappeared
**Details if related**
so based on that this will also make it safe in case any one add any value to the enum of the poolobj.direction as it will make safe from being out of boundaries

### DIFF
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -16,6 +16,8 @@
 #include <net/if.h>
 #include <sys/ioctl.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
+#include <sys/types.h>
 #include <signal.h>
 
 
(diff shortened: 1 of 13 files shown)


cfgmgr/buffermgrdyn.cpp
```diff
@@ -325,10 +325,13 @@ void BufferMgrDynamic::loadZeroPoolAndProfiles()
                         auto &poolObj = poolSearchRef->second;
                         if (poolObj.zero_profile_name.empty())
                         {
+                        if(poolObj.direction>= 0 && poolObj.direction <= 2) // ensuring the poolpbj.direction will not get out of boundaries
```
prsunny, line 328: please fix alignment

Issue #3436, opened 2024-12-20
[RouteOrch] Orchagent crashed when add and delete route at same time.


yangyue8:
**Description**
The refcount of nexthop is incorrect when delete and add route at the same time, which causes orchagent crashed.

**Error Log**
2024-11-25.06:00:49.527708|R|SAI_OBJECT_TYPE_ROUTE_ENTRY||
{"dest":"100.1.40.0/24","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000477"}
2024-11-25.06:00:49.530093|C|SAI_OBJECT_TYPE_ROUTE_ENTRY||
{"dest":"100.1.40.0/24","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000478"}
|SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID=oid:0x4000000000491
2024-11-25.06:00:49.532539|a|SYNCD_INVOKE_DUMP

2024 Nov 25 06:00:49.532717 sonic ERR swss#orchagent: :- removeMplsNextHop: Failed to remove next hop push17/148+100.1.48.1@Ethernet1, rv:-17
2024 Nov 25 06:00:49.532717 sonic ERR swss#orchagent: :- handleSaiRemoveStatus: Encountered failure in remove operation, exiting orchagent, SAI API: SAI_API_NEXT_HOP, status: SAI_STATUS_OBJECT_IN_USE
2024 Nov 25 06:00:49.532753 sonic NOTICE swss#orchagent: :- notifySyncd: sending syncd: SYNCD_INVOKE_DUMP

PR #3438, opened 2024-12-20
Fixed Orchagent crashed when add and delete route at same time(#3436)

yangyue8:
**What I did**
reference： https://github.com/sonic-net/sonic-swss/pull/1501 
Remove mpls next-hop after updating the reference counter for a bulk of routes instead of removing it in the loop of updating the reference counter.
Fix https://github.com/sonic-net/sonic-swss/issues/3436

**Why I did it**
As described in issue #1501, I did this to fixed the same problem for mpls nexthop.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -975,6 +975,7 @@ void RouteOrch::doTask(Consumer& consumer)
         // Go through the bulker results
         auto it_prev = consumer.m_toSync.begin();
         m_bulkNhgReducedRefCnt.clear();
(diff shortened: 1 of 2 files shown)


yangyue8:
@prsunny  Please help to review, thank you.

prsunny:
@yangyue8 , can you add some tests to simulate this issue?

yangyue8:
Hi @prsunny ,I have added some tests to verify my changes, please help to review. 
My test case: Delete a route with **mpls nexthop** from vrf1, and add it to vrf2 at the same time.

PR #630, opened 2018-09-28
[portsyncd]: Store eth0 oper status into state DB

stcheng:
portsyncd listens to the netlink messages. In order for SNMP
to get the current oper status of the managemnt interface,
portsyncd will store eth0 oper status explicitly into the
state database.

Signed-off-by: Shu0T1an ChenG <shuche@microsoft.com>

PR #3439, opened 2024-12-20
Remove update of mgmt oper status in swss

SuvarnaMeenakshi:
**What I did**
Issue to be fix: Currently operational status of mgmt interface is not present or correct for multi-asic devices.

**Why I did it**
Initial PR that added mgmt oper status feature in swss: https://github.com/sonic-net/sonic-swss/pull/630
https://github.com/sonic-net/sonic-buildimage/pull/21245 adds a script to update oper status of management interface periodically. In doing so, we no longer need to update oper status of mgmt interface in swss.

**How I verified it**
Verified on single-asic platform
1. verified on single-asic platform and multi-asic Chassis platfrom.
Single ASIC Arista device verification along with https://github.com/sonic-net/sonic-buildimage/pull/21245 changes
Ran the below bash script to verify the state of STATE_DB: MGMT_OPER_STATUS table and also execute config_reload, verify if STATE_DB is flushed out and repopulated after monit starts periodic script.
```
#!/bin/bash
CUR_STATUS=`sonic-db-cli STATE_DB hgetall "MGMT_PORT_TABLE|eth0"`
echo "current status in STATE_DB is $CUR_STATUS :: expected to have up state"
snmp_result=`docker exec snmp snmpwalk -v2c -c <comm> <IP> 1.3.6.1.2.1.2.2.1.8.10000`
echo "snmp result  before config reload 5min $snmp_result :: expected to have 1 in snmp result"
sudo config reload -y
CUR_STATUS=`sonic-db-cli STATE_DB hgetall "MGMT_PORT_TABLE|eth0"`
echo "current status in STATE_DB after config_reload is $CUR_STATUS :: expected to have empty"
# sleep for snmp to start
sleep 60
snmp_result=`docker exec snmp snmpwalk -v2c -c <comm> <IP> 1.3.6.1.2.1.2.2.1.8.10000`
echo "snmp result  after config reload $snmp_result :: expected to return error since STATE_DB is not yet populated"
# monit will populate mgmt oper status after each 5min cycle
sleep 240
CUR_STATUS=`sonic-db-cli STATE_DB hgetall "MGMT_PORT_TABLE|eth0"`
echo "current status in STATE_DB after config_reload after 5min $CUR_STATUS :: expected to have up state"
snmp_result=`docker exec snmp snmpwalk -v2c -c <comm> <IP> 1.3.6.1.2.1.2.2.1.8.10000`
echo "snmp result  after config reload after 5min $snmp_result :: expected to have 1 in snmp result"
```
Result of above script:
```
current status in STATE_DB is {'oper_status': 'up'} :: expected to have up state
snmp result  before config reload 5min iso.3.6.1.2.1.2.2.1.8.10000 = INTEGER: 1 :: expected to have 1 in snmp result
Acquired lock on /etc/sonic/reload.lock
Disabling container and routeCheck monitoring ...
Stopping SONiC target ...
Running command: /usr/local/bin/sonic-cfggen -j /etc/sonic/init_cfg.json -j /etc/sonic/config_db.json --write-to-db
Running command: /usr/local/bin/db_migrator.py -o migrate
Running command: /usr/local/bin/sonic-cfggen -d -y /etc/sonic/sonic_version.yml -t /usr/share/sonic/templates/sonic-environment.j2,/etc/sonic/sonic-environment
Restarting SONiC target ...
Enabling container and routeCheck monitoring ...
Reloading Monit configuration ...
Reinitializing monit daemon
Released lock on /etc/sonic/reload.lock
current status in STATE_DB after config_reload is {} :: expected to have empty
snmp result  after config reload iso.3.6.1.2.1.2.2.1.8.10000 = No Such Object available on this agent at this OID :: expected to return error since STATE_DB is not yet populated
current status in STATE_DB after config_reload after 5min {'oper_status': 'up'} :: expected to have up state
snmp result  after config reload after 5min iso.3.6.1.2.1.2.2.1.8.10000 = INTEGER: 1 :: expected to have 1 in snmp result
```

**Details if related**


gechiang (approved):
LGTM.

prsunny:
@SuvarnaMeenakshi , could you link the original PR that added this feature? Is there any other dependency that we've to address?

SuvarnaMeenakshi:
> @SuvarnaMeenakshi , could you link the original PR that added this feature? Is there any other dependency that we've to address?

hi @prsunny this is the PR that added this feature: https://github.com/sonic-net/sonic-swss/pull/630

portsyncd/linksync.cpp
```diff
@@ -50,43 +49,6 @@ LinkSync::LinkSync(DBConnector *appl_db, DBConnector *state_db) :
     {
         string key = idx_p->if_name;
 
-        /* Explicitly store management ports oper status into the state database.
```
prsunny, line 53: Why do we need to keep the` for loop` as this was also added as part of #630

portsyncd/linksync.cpp
```diff
@@ -197,17 +159,6 @@ void LinkSync::onMsg(int nlmsg_type, struct nl_object *obj)
                        nlmsg_type, key.c_str(), admin, oper, addrStr, ifindex, master);
     }
 
-    if (!key.compare(0, MGMT_PREFIX.length(), MGMT_PREFIX))
```
prsunny, line 200: I think we should remove the check at line 134 as well

arlakshm:
@SuvarnaMeenakshi, can you please address the comments?

portsyncd/linksync.cpp
SuvarnaMeenakshi: Yes, we don't need this, removed to match initial PR changes

portsyncd/linksync.cpp
SuvarnaMeenakshi: Removed this : key.compare(0, MGMT_PREFIX.length(), MGMT_PREFIX)

prsunny (approved):

PR #3447, opened 2025-01-03
[orchagent] Do not restore port admin if port admin is configured

PJHsieh:
**What I did**
Only restore port admin if port admin is not configured.

**Why I did it**
The pCfg.admin_status might be overridden by prior configurations.

**How I verified it**

**Details if related**


prsunny:
@PJHsieh , do you've some logs and examples of the issue you are fixing in this PR?

PJHsieh:
> @PJHsieh , do you've some logs and examples of the issue you are fixing in this PR?

Hi, @prsunny , please check the following flow.

Example:
[Initialization](https://github.com/sonic-net/sonic-swss/blob/eba4d4bb95a95eed44cc49fde051112c30be132a/orchagent/portsorch.cpp#L4046)
The variable admin_status is initialized with the value of p.m_admin_state_up. Assume p.m_admin_state_up is true.

[User set port FEC](https://github.com/sonic-net/sonic-swss/blob/eba4d4bb95a95eed44cc49fde051112c30be132a/orchagent/portsorch.cpp#L4560):
When configuring FEC, the variable p.m_admin_state_up is set to false.


[User set port admin](https://github.com/sonic-net/sonic-swss/blob/eba4d4bb95a95eed44cc49fde051112c30be132a/orchagent/portsorch.cpp#L4694):
Additionally, the user configures the port admin to false in the same iteration. In this case, the value of pCfg.admin_status.value will be restored to true.

orchagent/portsorch.cpp
```diff
@@ -4692,7 +4692,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                 initializePortOperErrors(p);
 
                 // Restore admin status if the port was brought down
-                if (admin_status != p.m_admin_state_up)
+                if (admin_status != p.m_admin_state_up && pCfg.admin_status.is_set == false)
```
stepanblyschak, line 4695: suggest to add unit test

PJHsieh:
@prsunny @prgeor @stepanblyschak 
Hi, Pals,
the unit test is added. Please review this PR, thanks.

stepanblyschak (approved):

prsunny (approved):

PR #3449, opened 2025-01-04
Fix stporch init exception

divyachandralekha:
**What I did**
Fixed stporch init exception
**Why I did it**
SAI retrieval of STP attributes may return either SAI_STATUS_NOT_SUPPORTED or SAI_STATUS_NOT_IMPLEMENTED. An exception has been added to handle such failures, preventing Orchagent from starting on systems where STP is not supported


divyachandralekha:
@wajahatrazi  @srj102 @oleksandrivantsiv  
Please help to merge the changes

prsunny (approved):

PR #3450, opened 2025-01-06
solving vulnerabilities

marwanelmessiry:
solving vulnerabilities of out of boundaries risk of string
**What I did**
i added if condition to ensure that the size of IFNAMSIZ didn't change during the runtime  
**Why I did it**
insure that the string if name will still in boundaries
**How I verified it**
using checkmarx tool as after that the vulnerability is solved


prsunny:
Please remove space/alignment changes and keep only the required change

PR #3451, opened 2025-01-06
solving vulnerabilities in orch.cpp

marwanelmessiry:
**What I did**
i added the range of the field to ensure being into the boundaries of the string

**Why I did it**
because ithere is risk of leaving it ike that so the field changes in the runtime which makes it a high sverity vulnerability
**How I verified it**
using my checkmarx tool it ensured that it was valnurable and after the edit the vulnerability is gone


prsunny:
@marwanelmessiry , can you please update PR with ONLY the required change? Remove the alignment/space changes to help review easier

PR #3452, opened 2025-01-07
[BufferOrch] Use SAI bulk API to configure port, PG and queue

stepanblyschak:
**What I did**

Make use of SAI set bulk API to improve switch boot up performance, especially in warm-boot and fast-boot scenarios.

The general concept:
1. First, tasks are processed one by one by corresponding ```process*``` methods which add the SAI operation with a context to a bulk buffer. Bulk buffers are split by DB operation.
2. Bulk buffer is flushed to syncd using SAI bulk API, first DELETE operations are pushed in bulk then SET operations are pushed. Status code for each operation is updated in the task context structure.
3. Lastly, corresponding ```process*Post``` methods are invoked to handle SAI status code and perform post set operations like enabling FC counter for a PG/queue upon success.

This design allows re-use of all existing code that is written to handle one task at a time and a small change is needed to maintain task context persistence throughout steps 1-3.

**Why I did it**

To improve fast-boot/warm-boot convergence time.

**How I verified it**

Manual verification by booting up the T0 topology configuration and running basic tests with static and dynamic buffer models. Fast-boot tested and observed ~2 times improvement in buffer configuration time on 256 port system from 3 sec to 1.7 sec when syncd falls back to single SET API. A bigger improvement is expected with proper SAI driver bulk SET support.

**Details if related**

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -72,6 +72,11 @@ void BufferOrch::initTableHandlers()
     m_bufferHandlerMap.insert(buffer_handler_pair(APP_BUFFER_PG_TABLE_NAME, &BufferOrch::processPriorityGroup));
     m_bufferHandlerMap.insert(buffer_handler_pair(APP_BUFFER_PORT_INGRESS_PROFILE_LIST_NAME, &BufferOrch::processIngressBufferProfileList));
     m_bufferHandlerMap.insert(buffer_handler_pair(APP_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME, &BufferOrch::processEgressBufferProfileList));
+
+    m_bufferFlushHandlerMap.insert(buffer_flush_handler_pair(APP_BUFFER_PORT_INGRESS_PROFILE_LIST_NAME, &BufferOrch::processIngressBufferProfileListBulk));
+    m_bufferFlushHandlerMap.insert(buffer_flush_handler_pair(APP_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME, &BufferOrch::processEgressBufferProfileListBulk));
+    m_bufferFlushHandlerMap.insert(buffer_flush_handler_pair(APP_BUFFER_PG_TABLE_NAME, &BufferOrch::processPriorityGroupBulk));
+    m_bufferFlushHandlerMap.insert(buffer_flush_handler_pair(APP_BUFFER_QUEUE_TABLE_NAME, &BufferOrch::processQueueBulk));
 }
 
 void BufferOrch::initBufferReadyLists(DBConnector *applDb, DBConnector *confDb)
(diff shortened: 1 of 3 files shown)


developfast:
Will review shortly

orchagent/bufferorch.cpp
```diff
@@ -974,7 +988,52 @@ task_process_status BufferOrch::processQueue(KeyOpFieldsValuesTuple &tuple)
...
+task_process_status BufferOrch::processQueuePost(const QueueTask& task)
+{
+    SWSS_LOG_ENTER();
+
+    const auto& key = kfvKey(task.kofvs);
+    const auto& op = kfvOp(task.kofvs);
+    const auto tokens = tokenize(key, delimiter);
+    Port port;
+
+    for (const auto& portContext: task.ports)
```
kperumalbfn, line 1021: Any capability check to query vendor SAI's bulk API implementation?

stepanblyschak: Not required, please check https://github.com/jianyuewu/sonic-sairedis/blob/master/syncd/Syncd.cpp#L2186

stepanblyschak:
VS tests fail due to bulk set not supported for PGs in init view in syncd:
```
ERR #syncd: :- run: Runtime error: :- processBulkQuadEventInInitViewMode: SAI_OBJECT_TYPE_INGRESS_PRIORITY_GROUP is not supported in init view mode
```

Either need to add such support or disable bulk for systems that use init view

stepanblyschak:
@kperumalbfn @developfast @prsunny Kind reminder to review

kperumalbfn:
> VS tests fail due to bulk set not supported for PGs in init view in syncd:
> 
> ```
> ERR #syncd: :- run: Runtime error: :- processBulkQuadEventInInitViewMode: SAI_OBJECT_TYPE_INGRESS_PRIORITY_GROUP is not supported in init view mode
> ```
> 
> Either need to add such support or disable bulk for systems that use init view

@stepanblyschak Should we enable this for VS?

If one vendor SAI doesn't support bulk APIs for PG, queue, do we have fallback to regular API?

stepanblyschak:
> > VS tests fail due to bulk set not supported for PGs in init view in syncd:
> > ```
> > ERR #syncd: :- run: Runtime error: :- processBulkQuadEventInInitViewMode: SAI_OBJECT_TYPE_INGRESS_PRIORITY_GROUP is not supported in init view mode
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Either need to add such support or disable bulk for systems that use init view
> 
> @stepanblyschak Should we enable this for VS?
> 
> If one vendor SAI doesn't support bulk APIs for PG, queue, do we have fallback to regular API?

@kperumalbfn Bulk set support in init view added in this PR https://github.com/sonic-net/sonic-sairedis/pull/1496.
Yes, fallback implemented in syncd - https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/Syncd.cpp#L2234

kperumalbfn (approved):

stepanblyschak:
```
>           raise HTTPError(http_error_msg, response=self)
E           requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http+docker://localhost/v1.47/images/debian:jessie/json
```

stephenxs (approved):

bingwang-ms:
@prsunny Could you help review & merge? Thanks

PR #3453, opened 2025-01-07
solved string vulnerability 

marwanelmessiry:
**What I did**
add try and catch to ensure that the map can contain what in field
edits in line 460
**Why I did it**
as it wa s a vulnerability that there is a ris of the change of field which make vulnerability
**How I verified it**
through tool called checkmarx and it was solved through the tool that gets vulnerabilities


prsunny:
@marwanelmessiry , thanks for the fix. Can you please resubmit the PR with only the change needed for fix? Spacing changes can be removed

PR #3457, opened 2025-01-08
solving vulnerability of  string last element not null neighsync.cpp

marwanelmessiry:
**What I did**
added try and catch for the string  macstr and making last element = NULL 
**Why I did it**
as it gives vulnerability of being not sure that the last element of the string is not NULL
**How I verified it**
by using checkmarx  it gave vulnerability of that string last element is not assigned to zero by these edits it was solved

### DIFF
--- a/neighsyncd/neighsync.cpp
+++ b/neighsyncd/neighsync.cpp
@@ -135,12 +135,32 @@ void NeighSync::onMsg(int nlmsg_type, struct nl_object *obj)
 
     if (use_zero_mac)
     {
+       try
+    {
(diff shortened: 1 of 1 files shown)


neighsyncd/neighsync.cpp
```diff
@@ -134,13 +134,33 @@ void NeighSync::onMsg(int nlmsg_type, struct nl_object *obj)
     }
 
     if (use_zero_mac)
+    {
+       try
```
prsunny, line 138: Please fix alignment

PR #3458, opened 2025-01-09
Add heartbeat interval parameter.

liuh-80:
Add heartbeat interval parameter.

#### Why I did it
Make this feature can be disable, because log spam issue on small disk device:
https://github.com/sonic-net/sonic-buildimage/issues/21157

##### Work item tracking
- Microsoft ADO: 30594076

#### How I did it
Add heartbeat interval parameter. and disable heartbeat  when interval is 0.

#### How to verify it
Pass all UT.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

Add heartbeat interval parameter.

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -351,7 +351,7 @@ int main(int argc, char **argv)
     string responsepublisher_rec_filename = Recorder::RESPPUB_FNAME;
     int record_type = 3; // Only swss and sairedis recordings enabled by default.
 
-    while ((opt = getopt(argc, argv, "b:m:r:f:j:d:i:hsz:k:q:c:t:v:")) != -1)
+    while ((opt = getopt(argc, argv, "b:m:r:f:j:d:i:hsz:k:q:c:t:v:I:")) != -1)
     {
         switch (opt)
         {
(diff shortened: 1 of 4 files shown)


orchagent/main.cpp
```diff
@@ -450,6 +451,12 @@ int main(int argc, char **argv)
                 vrf = optarg;
             }
             break;
+        case 'I':
+            if (optarg)
+            {
+                g_heart_beat_interval = atoi(optarg);
```
qiluo-msft, line 457: >g_heart_beat_interval [](http://example.com/codeflow?start=16&length=21)

Could you add more error handling?

qiluo-msft, line 457: >g_heart_beat_interval [](http://example.com/codeflow?start=16&length=21)

you may init g_heart_beat_interval before the while loop, instead of init global static var which the sequence is not well defined.

liuh-80: Fixed

qiluo-msft: Should we treat 0 as error? an invalid string will also lead to 0.

liuh-80: 0 means disable heartbeat feature

qiluo-msft (approved):

PR #3461, opened 2025-01-10
Fix VRF update handling for loopback interfaces in IntfsOrch

ram25794:
**What I did**

Addressed the scenario where an interface exists in m_syncdIntfses, but the incoming VRF does not match the updated VRF for the loopback interface. The solution involves updating the loopback’s VRF ID, decrementing the reference count for the old VRF, and incrementing the reference count for the new VRF

**Why I did it**

Sometimes DEL is missing in the orchagent as mentioned in the issue (https://github.com/sonic-net/sonic-swss/issues/3339), leading to a scenario where, when a loopback interface is added to a non-default VRF, an ip2me route is incorrectly added to the default VRF. This issue occurs because the IntfsOrch is not handling the case where the interface exists in m_syncdIntfses, but the VRF does not match the updated VRF configuration.

**How I verified it**

Reproduced the issue with the fix applied and verified in the SAI Redis log that the ip2me route was correctly added to the non-default VRF

2025-01-07.20:48:12.797119|c|SAI_OBJECT_TYPE_ROUTE_ENTRY:{"dest":"8.8.8.8/32","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000427"}|SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION=SAI_PACKET_ACTION_FORWARD|SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID=oid:0x1000000000001

Added Unit Tests

**Details if related**

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -846,6 +846,19 @@ void IntfsOrch::doTask(Consumer &consumer)
                         m_syncdIntfses[alias] = intfs_entry;
                         m_vrfOrch->increaseVrfRefCount(vrf_id);
                     }
+                    else if (m_syncdIntfses[alias].vrf_id != vrf_id)
+                    {
+                        if (m_syncdIntfses[alias].ip_addresses.size() == 0)
+                        {
+                            m_vrfOrch->decreaseVrfRefCount(m_syncdIntfses[alias].vrf_id);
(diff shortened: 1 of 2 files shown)


dgsudharsan (approved):

dprital:
@kperumalbfn , Can you please cherry pick to 202411 ?

PR #3476, opened 2025-01-18
[vlanmgrd]: Fixing an issue causing mismatch between MAC and link-local IPv6 addresses of VLAN and Bridge interfaces

mramezani95:
**What I did**
1. Added code to bring down the Bridge interface before changing the MAC address of a VLAN interface and the Bridge, and then starting up the Bridge after the MAC is changed. This automatically brings down and starts up all VLAN interfaces added to the Bridge.
2. Added code to bring down and then immediately start up the Bridge after the dummy interface is started up. This is useful to ensure that MAC and link-local IPv6 addresses of the Bridge are consistent in case no VLANs are added to the Bridge later. 
3. Added a VS test to verify these behaviors.

**Why I did it**
After PR #3370, the Bridge becomes operationally UP after the dummy interface is started up. As a result, all VLAN interfaces created under the Bridge are immediately operationally UP after creation. This can cause an issue later if their MAC address is changed since the kernel does not update the link-local IPv6 address of an interface if it is operationally UP. This behavior caused the IPv6 version of the following test cases in sonic-mgmt to fail for the dualtor topology, which were temporarily skipped:
```
arp/test_arp_dualtor.py::test_arp_update_for_failed_standby_neighbor
arp/test_arp_dualtor.py::test_standby_unsolicited_neigh_learning
arp/test_arp_extended.py::test_proxy_arp
```

**How I verified it**
1. Verified that the above sonic-mgmt tests can be passed after this change.
2. Verified that the link-local IPv6 addresses are updated correctly after the Bridge is brought down and started up again.

### DIFF
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -16,6 +16,8 @@
 #include <net/if.h>
 #include <sys/ioctl.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
+#include <sys/types.h>
 #include <signal.h>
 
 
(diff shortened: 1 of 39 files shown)


prsunny:
Can you add a test to fix the coverage on `setHostVlanMac` ?

prsunny (approved):
lgtm

lolyu:
Hi @prsunny, is this issue on 202411? Should we backport this fix?

vivekverma-arista:
> Hi @prsunny, is this issue on 202411? Should we backport this fix?

yes the issue is seen on 202411 as well and we have verified that this PR fixes it

we should backport this

StormLiangMS:
MSFT ADO:30590329
@prsunny  @kperumalbfn could you help to take a look at the 202411 cherrypick? do we need this one on 202411?

bingwang-ms:
@prsunny Do we need this change for 202405 branch?

prsunny:
> @prsunny Do we need this change for 202405 branch?

Not required @bingwang-ms , the original PR dint make to 202405

PR #3479, opened 2025-01-22
Git ignore .gcda and .gcno in all folders

stephenxs:
**What I did**

Ignore `.gcda` and `.gcno` in all folders to avoid seeing a large number of untracked files in `git status`

**Why I did it**

**How I verified it**

**Details if related**


liat-grozovik:
@prsunny  please review and if no comments, please merge
@stephenxs  only needed for master?

prsunny (approved):

PR #3480, opened 2025-01-22
[vxlanorch]: p2mp tunnels not stored in VXLAN state table

bradh352:
**Why I did it**

Whether a VXLAN tunnel is brought up as p2p or p2mp depends on the silicon in use and isn't otherwise expressed to the end user.  The side effect of this is `show vxlan remotevtep` shows no destinations when the silicon only supports p2mp such as Mellanox Spectrum1.  This can cause a user to believe the vxlan tunnel wasn't properly formed when it really was.

p2mp tunnels do not get a Port created, so do not rely on an operstatus being set by portsorch to know if they are up or down.  This is presumably why the tunnel entries were not recorded. 

**What I did**

This introduces a new status of 'p2mp' in this case to denote this difference and records p2mp tunnels in the VXLAN state table.

Due to some past restructuring, calls into `updateRemoteEndPointIpRef()` were omitted as p2mp tunnels are not calling `addTunnelUser()`.  This has been resolved and this function now calls into `addRemoveStateTableEntry()` for tracking of those remote vteps.

**How I verified it**

Brought up vxlan tunnels on an Nvidia SN2201 switch and verified entries are now present in `show vxlan remotevtep`.

Result:
```
+------------+------------+-------------------+--------------+
| SIP        | DIP        | Creation Source   | OperStatus   |
+============+============+===================+==============+
| 172.16.0.3 | 172.16.0.1 | EVPN              | oper_p2mp    |
+------------+------------+-------------------+--------------+
| 172.16.0.3 | 172.16.0.2 | EVPN              | oper_p2mp    |
+------------+------------+-------------------+--------------+
Total count : 2
```

**Details if related**


bradh352:
test failures are in SRV6, not related to this

bradh352:
rebased to force rebuild after merge of srv6 test failure fixes

PR #3482, opened 2025-01-24
Handle duplicate SF in coverage.info to make sure coverage.xml gets correct statistics

a114j0y:
**What I did**
use `--add-tracefile` option in `debian/rules` and `tests/conftest.py` to sanitize `coverage.info` generated by `lcov`

**Why I did it**
`lcov` generates an initial `coverage.info` file based on collected `.gcno` and `.gcda` files, this `.info` file contains coverage information for different source files (marked as `SF`).  Sometimes you would observe that the same `SF` appears multiple times, it means `lcov` gets multiple copies of coverage information for this file, since this file may have appeared in multiple compilation units, and for each copy, the hit times of its lines are different. 

Then `lcov_cobertura` generates  `coverage.xml` based on `coverage.info`. However, it can't deal with duplicate SF in `coverage.info` properly. If it sees duplicate coverage information for a source file from `coverage.info`, it always overwrites the old copy with the new copy, hence only the last copy would be counted. However, if the last copy considers the functions as missing, the function is considered as missing in `coverage.xml`, which is used to determine whether the new PR passes the coverage threshold.

The proper way is to add the hit times of all the copies, which could be achieved by lcov `add-tracefile` option.

**How I verified it**
Before using `--add-tracefile`, RingBuffer related functions in this PR are considered missing, though they are covered in testcases. After adding `--add-tracefile`, it passes the coverage check.

[[orchagent] implement ring buffer feature with a flag #3242](https://github.com/sonic-net/sonic-swss/pull/3242)

**Details if related**
By downloading the artifact of PR#3242,  we could check coverage.xml, which has already had the correct statistics.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -44,6 +44,7 @@ override_dh_auto_install:
 ifeq ($(ENABLE_GCOV), y)
 	mkdir -p debian/swss/tmp/gcov
 	lcov -c --directory . --no-external --exclude "$(shell pwd)/tests/*" --exclude "$(shell pwd)/**/tests/*" --ignore-errors gcov --output-file coverage.info
+	lcov --add-tracefile coverage.info -o coverage.info
 	lcov_cobertura coverage.info -o coverage.xml
 	find ./ -type f -regex '.*\.\(h\|cpp\|gcno\|info\)' | tar -cf debian/swss/tmp/gcov/gcov-source.tar -T -
 endif
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -462,7 +462,7 @@ class DockerVirtualSwitch:
                 self.runcmd('supervisorctl stop all')
 
             # Generate the converage info by lcov and copy to the host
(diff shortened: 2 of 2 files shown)


a114j0y:
[coverage.zip](https://github.com/user-attachments/files/18609193/coverage.zip)
This .zip contains the coverage.info and coverage.xml without the change of this PR.
@saiarcot895 @theasianpianist

you can check search `SF:/__w/1/s/orchagent/orch.cpp` in this file and would find 4 copies there
then you can search `FN:36,_ZN10RingBuffer6notifyEv` in this file and would see 4 copies of `FNDA` there, only one copy is non-zero

saiarcot895:
The best fix for this is to rework the automake build files to not need to compile the same files repeatedly. This probably involves switching to non-recursive automake, and takes a bit of time to do so without breaking anything. This is a good enough short-term fix, at least.

saiarcot895 (approved):

prsunny (approved):

PR #3485, opened 2025-01-27
[ci]Use sonictest instead of sonic-common for test stage.

liushilongbuaa:
**What I did**
Something wrong in agent pool sonictest and sonic-common. No one maintain the agent initial script.
Now I did a quick fix on these two agent pool. sonictest agent pool works but sonic-common still has issues.

**Why I did it**
1. sonictest is by design to run swss related tests. swss-common and sairedis tests run on sonictest. Only swss run on sonic-common. It is not as expected.

**How I verified it**

**Details if related**


liushilongbuaa:
@prsunny , description updated.

prsunny (approved):

PR #3502, opened 2025-02-08
[copp]: Use non-zero trap priority for default trap group

prabhataravind:
**What I did**
Microsoft ADO **29981358**
Updated swss code to use a priority of 1 instead of 0 for default trap group.

**Why I did it**

Some vendor SAI implementations use zero as a default lowest priority internally which can impact TTL 1 packet trapping behavior in certain platforms.

**How I verified it**
 * By running copp tests which include rate-limiting tests for TTL 1 packets on on Cisco/Mellanox/Arista platforms.
 * By manual tests with TTL 1 packets generated by scapy
 
**Details if related**
Example packet:
def send_pkt(intf):
    pkt = Ether(src='00:06:07:08:09:0a',dst='84:7f:c2:88:57:a7') \
            /IP(dst='20.0.0.46',src='20.0.0.47',ttl=1) \
            /TCP(sport=1025,dport=179)
    sendp(pkt,count=50000,iface=intf)

### DIFF
--- a/orchagent/copporch.cpp
+++ b/orchagent/copporch.cpp
@@ -188,13 +188,17 @@ void CoppOrch::initDefaultTrapIds()
     attr.value.oid = m_trap_group_map[default_trap_group];
     trap_id_attrs.push_back(attr);
 
-    /* Mellanox platform doesn't support trap priority setting */
-    /* Marvell platform doesn't support trap priority. */
+    /*
+     * Use a default trap priority > 0 to avoid undesirable packet trapping
+     * behavior on some platforms that use 0 as default SAI-internal priority.
+     * Note: Mellanox and Marvell platforms don't support trap priority setting.
+     */
+
     char *platform = getenv("platform");
     if (!platform || (!strstr(platform, MLNX_PLATFORM_SUBSTRING) && (!strstr(platform, MRVL_PRST_PLATFORM_SUBSTRING))))
     {
         attr.id = SAI_HOSTIF_TRAP_ATTR_TRAP_PRIORITY;
-        attr.value.u32 = 0;
+        attr.value.u32 = 1;
         trap_id_attrs.push_back(attr);
     }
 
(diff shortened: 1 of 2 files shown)


orchagent/copporch.cpp
```diff
@@ -188,13 +188,17 @@ void CoppOrch::initDefaultTrapIds()
...
+     * behavior on some platforms that use 0 as default SAI-internal priority.
+     * Note: Mellanox and Marvell platforms don't support trap priority setting.
+     */
+
     char *platform = getenv("platform");
     if (!platform || (!strstr(platform, MLNX_PLATFORM_SUBSTRING) && (!strstr(platform, MRVL_PRST_PLATFORM_SUBSTRING))))
     {
         attr.id = SAI_HOSTIF_TRAP_ATTR_TRAP_PRIORITY;
-        attr.value.u32 = 0;
+        attr.value.u32 = 1;
```
kperumalbfn, line 201: In https://github.com/sonic-net/sonic-buildimage/blob/b66bfc0b25065273787b46ed20e0115f80ad6578/files/image_config/copp/copp_cfg.j2, trap_priority for queue1_group1, queue1_group2 and queue2_group1 are trap_priority 1. Should we change the priority for those trap groups?

prabhataravind: I tested with ip2me packets with TTL1 and TTL> 1. With TTL 1 ip2me packets, we now hit default trap first and packets go to queue 0, whereas with TTL> 1 packet, it hits the ip2me trap. So with the current logic, if any traps have the same priority as default trap (SAI_HOSTIF_TRAP_TYPE_TTL_ERROR), then corresponding packets will hit queue0 if they had TTL 1 and the configured queue for the trap if the packets had a TTL > 1. That is a reasonable behavior right?

kperumalbfn: Since the priority is same for default and ip2me, order of the trap_group could dictate the ACL match.

kperumalbfn (approved):

kperumalbfn:
Will add this to 202411

zhixzhu (approved):

PR #3504, opened 2025-02-09
Optimize counter initialization by reducing the number of bulk counter poll calls and communication between swss/sairedis

stephenxs:
**What I did**

Optimize counter initialization by reducing the number of bulk counter poll calls and communication between swss(orchagent)/sairedis(syncd) during initialization.

Originally, `orchagent` notifies `syncd` to initialize the counter using an extended sairedis call `SAI_REDIS_SWITCH_ATTR_FLEX_COUNTER` for each SAI object with the object ID as the key, which means the number of the extended sairedis calls is identical as the number of objects. It takes time to finish all the extended sairedis calls.

Now, for counter groups that have many objects (e.g., port, PG, queues, etc), `orchagent` notifies `syncd` to initialize the counter using a single extend sairedis call with many objects' ID as the key (format: `<key1>,<key2>,...<keyn>`). So, it takes much less time to initialize the counters because fewer extend sairedis calls are required.

HLD https://github.com/sonic-net/SONiC/pull/1862
Depends on https://github.com/sonic-net/sonic-sairedis/pull/1527

**Why I did it**

**How I verified it**

Unit test (mock test) and regression.

**Details if related**
- In sairedis, the bulk counter is supported for all counter groups except `Buffer Pool Counter` and `DASH ENI counter`.
- In swss, bulk counter for the following counter groups

  - priority group watermark
  - priority group drop
  - queue watermark
  - queue stat
  - PFC watchdog
  - WRED/ECN counter


stephenxs:
need https://github.com/sonic-net/sonic-sairedis/pull/1527 to pass vs test. dependency mentioned in the PR description.

stephenxs:
Did a test on local VS env. all counter-relevant cases passed with the corresponding sairedis package.

keboliu (approved):

orchagent/portsorch.cpp
```diff
@@ -8412,7 +8365,19 @@ void PortsOrch::addWredQueueFlexCountersPerPortPerQueueIndex(const Port& port, s
...
+void PortsOrch::flushCounters()
+{
+    port_stat_manager.flush();
+    port_buffer_drop_stat_manager.flush();
+    queue_stat_manager.flush();
+    queue_watermark_manager.flush();
+    pg_watermark_manager.flush();
+    pg_drop_stat_manager.flush();
+    wred_port_stat_manager.flush();
+    wred_queue_stat_manager.flush();
```
kcudnik, line 8380: should those be in some set or vector if they share the same base ? so we could use or loop to flush them ?

kcudnik (approved):

stephenxs:
@kperumalbfn could you please review this PR? Thanks

orchagent/p4orch/tests/fake_portorch.cpp
```diff
@@ -21,8 +26,12 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
                         PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, true),
       port_buffer_drop_stat_manager(PORT_BUFFER_DROP_STAT_FLEX_COUNTER_GROUP, StatsMode::READ,
                                     PORT_BUFFER_DROP_STAT_POLLING_INTERVAL_MS, true),
-      queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ,
-                         QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, true)
+      queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
+      queue_watermark_manager(QUEUE_WATERMARK_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ_AND_CLEAR, QUEUE_WATERMARK_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
```
kperumalbfn, line 30: For PG/Queue watermark, READ_AND_CLEAR mode is set. Is this for all watermark poll?

stephenxs: yes. for all.
we only have per-object counter mode for buffer pool counters.
for all other counter groups, like watermark, queue, etc, all objects of a group share a unified mode.
btw this is test stub code but in the production code it's the same. :)

orchagent/portsorch.cpp
```diff
@@ -553,8 +564,21 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
                 PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
         port_buffer_drop_stat_manager(PORT_BUFFER_DROP_STAT_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_BUFFER_DROP_STAT_POLLING_INTERVAL_MS, false),
         queue_stat_manager(QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, QUEUE_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
+        queue_watermark_manager(QUEUE_WATERMARK_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ_AND_CLEAR, QUEUE_WATERMARK_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
+        pg_watermark_manager(PG_WATERMARK_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ_AND_CLEAR, PG_WATERMARK_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
+        pg_drop_stat_manager(PG_DROP_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, PG_DROP_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
         wred_port_stat_manager(WRED_PORT_STAT_COUNTER_FLEX_COUNTER_GROUP, StatsMode::READ, PORT_STAT_FLEX_COUNTER_POLLING_INTERVAL_MS, false),
```
kperumalbfn, line 570: What would be the difference between PERIODIC_WATERMARK and USER_WATERMARK? Could you add note to this?

orchagent/portsorch.cpp
```diff
@@ -7607,7 +7620,7 @@ void PortsOrch::generateQueueMapPerPort(const Port& port, FlexCounterQueueStates
             {
                 continue;
             }
-            queueTypeVector.emplace_back(id, queueType);
+            queueTypeVector.emplace_back(id, sai_queue_type_string_map[queueType]);
```
kperumalbfn, line 7623: For the queue counters, could you check if this bulk counter poll includes all the types including WRED ECN that got added recently?

orchagent/p4orch/tests/fake_portorch.cpp
kperumalbfn: Agree, it is is in p4orch test code. But we have similar mode in portsorch.cpp :)

orchagent/portsorch.cpp
stephenxs: yes. my last commit adjust WRED ECN accordingly. https://github.com/sonic-net/sonic-swss/pull/3504/commits/e8fc577cfa66ad8058ca154a3b81c0c020a63745

orchagent/portsorch.cpp
stephenxs: Hi
They are not handled here and is not relevant. the periodic, user and system watermark is handled in the lua plugins, eg https://github.com/sonic-net/sonic-swss/blob/master/orchagent/watermark_queue.lua for queue watermark,  which are invoked after each counter-polling.
we do not change that logic in this PR.

orchagent/portsorch.cpp
kperumalbfn: Thanks @stephenxs

liat-grozovik:
@kperumalbfn can this be approved and merged?

kperumalbfn (approved):

PR #3505, opened 2025-02-10
portsorch: don't call updateDbPortOperStatus on all port types

bradh352:
**What I did**

Only call `updateDbPortOperStatus()` on PHY and TUNNEL ports.

**Why I did it**

PORT_TABLE contains PortChannel oper_status entries which are not expected by portsorch which leads to warm/fastreboot failures like:
```
2025 Feb 10 09:33:07.111055 sonic NOTICE swss#orchagent: :- bake: foundPortConfigDone = 1
2025 Feb 10 09:33:07.111080 sonic NOTICE swss#orchagent: :- bake: foundPortInitDone = 1
2025 Feb 10 09:33:07.111395 sonic NOTICE swss#orchagent: :- bake: m_portTable->getKeys 263
2025 Feb 10 09:33:07.111403 sonic NOTICE swss#orchagent: :- bake: portCount = 257, m_portCount = 0
2025 Feb 10 09:33:07.111403 sonic ERR swss#orchagent: :- bake: Invalid port table: portCount, expecting 257, got 261
```

Regression caused by #3383

**How I verified it**

Asked @stepanblyschak to verify who reported issue.

**Details if related**
Fixes https://github.com/sonic-net/sonic-buildimage/issues/21688

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -8268,7 +8268,10 @@ void PortsOrch::updatePortOperStatus(Port &port, sai_port_oper_status_t status)
         return;
     }
 
-    updateDbPortOperStatus(port, status);
+    if (port.m_type == Port::PHY || port.m_type == Port::TUNNEL)
+    {
+        updateDbPortOperStatus(port, status);
+    }
 
     if (port.m_type == Port::PHY)
     {
--- a/tests/test_portchannel.py
+++ b/tests/test_portchannel.py
@@ -454,7 +454,7 @@ class TestPortchannel(object):
 
         # verify a PORT_TABLE entry containing the PortChannel is NOT created
         # in APPDB (sonic-buildimage Issue #21688)
-        tbl = swsscommon.Table(app_db, "PORTCHANNEL_MEMBER")
+        tbl = swsscommon.Table(app_db, "PORT_TABLE")
         status, _ = tbl.get("PortChannel111")
         assert status is False
 
--- a/tests/test_portchannel.py
+++ b/tests/test_portchannel.py
@@ -452,6 +452,12 @@ class TestPortchannel(object):
         fvs = dict(fvs)
         assert fvs['netdev_oper_status'] == 'up'
 
+        # verify a PORT_TABLE entry containing the PortChannel is NOT created
+        # in APPDB (sonic-buildimage Issue #21688)
+        tbl = swsscommon.Table(app_db, "PORTCHANNEL_MEMBER")
+        status, _ = tbl.get("PortChannel111")
+        assert status is False
+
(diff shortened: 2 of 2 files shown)


stepanblyschak (commented):
Looks good to me, can you cover this with a VS test?

stepanblyschak:
> > Looks good to me, can you cover this with a VS test?
> 
> I'm not at all familiar with the vs testing framework. It would be a lot easier in test_portchannel.py to check to make sure there are no PortChannel entries in the port table. Would that be acceptable since that would have caught this regression?

Sure

bradh352:
> > > Looks good to me, can you cover this with a VS test?
> > 
> > 
> > I'm not at all familiar with the vs testing framework. It would be a lot easier in test_portchannel.py to check to make sure there are no PortChannel entries in the port table. Would that be acceptable since that would have caught this regression?
> 
> Sure

Test added, I haven't tested locally but CI here should validate I didn't make a typo or something.

bradh352:
@stepanblyschak 
> > > > Looks good to me, can you cover this with a VS test?
> > > 
> > > 
> > > I'm not at all familiar with the vs testing framework. It would be a lot easier in test_portchannel.py to check to make sure there are no PortChannel entries in the port table. Would that be acceptable since that would have caught this regression?
> > 
> > 
> > Sure
> 
> Test added, I haven't tested locally but CI here should validate I didn't make a typo or something.

Looks like all tests passed.  Can you approve?

bradh352:
@stepanblyschak maybe now approve it? :)

bingwang-ms:
@prsunny Can you help review?

prsunny (approved):

prsunny:
@prgeor, @vaibhavhd  for viz

vaibhavhd:
Has someone verified that the warm-reboot issue does not exist with the TUNNEL oper status still present in the DB? The failure was seen due to portchannel in the DB and that I believe is fixed. But, we may still have a problem with TUNNEL status in DB. That test_wr_arp validates that part by running warm-reboot with VXLAN tunnel programmed. Please check.

bradh352:
> Has someone verified that the warm-reboot issue does not exist with the TUNNEL oper status still present in the DB? The failure was seen due to portchannel in the DB and that I believe is fixed. But, we may still have a problem with TUNNEL status in DB. That test_wr_arp validates that part by running warm-reboot with VXLAN tunnel programmed. Please check.

The tunnel option gets routed to a different table so there's no way it could cause the same issue:
https://github.com/sonic-net/sonic-swss/blob/d3d95abec269b459f95ef7ad6086698e6a1133a6/orchagent/portsorch.cpp#L3427-L3442

PR #3506, opened 2025-02-10
[portsorch] Ports initialization speedup

stepanblyschak:
DEPENDS ON: https://github.com/sonic-net/sonic-sairedis/pull/1509

**What I did**

Optimized port initialization process using the following approach:
1. Attributes that are known at initialization (whether because their default values are defined by the SAI spec or because we created the port with a value we already know) like speed and MTU.
2. Heavy operations like querying list of PGs and queues are done in bulk to save on SAIRedis call overhead.
3. Port initialization that has no dependncy on subsequent data plane configuration are extracted into a seperate "post initialization" routine and executed after data plane configuration is done.
4. Save SAI set API call when possible to provide a value during creation.

**Why I did it**

To improve boot time for systems with many ports.

**How I verified it**

Ran fast-reboot and measure. Observed 45% reduction in time.

**Details if related**

### DIFF
--- a/orchagent/port.h
+++ b/orchagent/port.h
@@ -26,7 +26,7 @@ extern "C" {
  * Orchagent adds extra 22 bytes for Ethernet header and FCS,
  * hence setting to 1492 (1514 - 22)
  */
-#define DEFAULT_MTU             1492 - MAX_MACSEC_SECTAG_SIZE
+#define DEFAULT_MTU             (1492 - MAX_MACSEC_SECTAG_SIZE)
 
 /*
  * Default TPID is 8100
@@ -213,7 +213,6 @@ public:
     uint32_t  m_fdb_count = 0;
     uint64_t  m_flap_count = 0;
     uint32_t  m_up_member_count = 0;
-    uint32_t  m_maximum_headroom = 0;
     std::set<uint32_t> m_adv_speeds;
     sai_port_interface_type_t m_interface_type = SAI_PORT_INTERFACE_TYPE_NONE;
     std::set<sai_port_interface_type_t> m_adv_interface_types;
--- a/orchagent/port.h
+++ b/orchagent/port.h
@@ -18,12 +18,15 @@ extern "C" {
 #include <sairedis.h>
 
 #define DEFAULT_PORT_VLAN_ID    1
+
+#define MAX_MACSEC_SECTAG_SIZE 32
+
 /*
  * Default MTU is derived from SAI_PORT_ATTR_MTU (1514)
  * Orchagent adds extra 22 bytes for Ethernet header and FCS,
  * hence setting to 1492 (1514 - 22)
  */
-#define DEFAULT_MTU             1492
+#define DEFAULT_MTU             1492 - MAX_MACSEC_SECTAG_SIZE
 
 /*
  * Default TPID is 8100
@@ -268,6 +271,8 @@ public:
     uint32_t m_suppress_threshold = 0;
     uint32_t m_reuse_threshold = 0;
     uint32_t m_flap_penalty = 0;
+
+    Role m_role;
 };
 
 }
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -3560,6 +3560,11 @@ bool PortsOrch::initExistingPort(const PortConfig& port)
     p.m_index = index;
     p.m_port_id = id;
 
+    if (gMySwitchType != "dpu")
+    {
+        initializeSchedulerGroups(p);
+    }
+
     /* initialize port admin status */
     if (!getPortAdminStatus(p.m_port_id, p.m_admin_state_up))
     {
@@ -3580,10 +3585,11 @@ bool PortsOrch::initExistingPort(const PortConfig& port)
         SWSS_LOG_ERROR("Failed to get initial port mtu %d", p.m_mtu);
     }
 
-    return initPortsBulk({p});
+    std::vector<Port> ports = {p};
+    return initPortsBulk(ports);
 }
 
-bool PortsOrch::initPortsBulk(std::vector<Port> ports)
+bool PortsOrch::initPortsBulk(std::vector<Port>& ports)
 {
     SWSS_LOG_ENTER();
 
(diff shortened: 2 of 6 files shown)


nazariig (approved):

orchagent/port.h
```diff
@@ -18,12 +18,15 @@ extern "C" {
...
+
+#define MAX_MACSEC_SECTAG_SIZE 32
+
 /*
  * Default MTU is derived from SAI_PORT_ATTR_MTU (1514)
  * Orchagent adds extra 22 bytes for Ethernet header and FCS,
  * hence setting to 1492 (1514 - 22)
  */
-#define DEFAULT_MTU             1492
+#define DEFAULT_MTU             (1492 - MAX_MACSEC_SECTAG_SIZE)
```
prsunny, line 29: I dont think we should be changing this default value as there could be multple dependencies/test making this assumption.

prsunny:
IMO, the changes are significant in this one PR. Was wondering if you could split the PR to smaller self contained PRs.

stepanblyschak:
> IMO, the changes are significant in this one PR. Was wondering if you could split the PR to smaller self contained PRs.

@prsunny Sure, working on splitting into separate PRs. Will keep this PR around until I create all sub PRs.
Please review the first cut off - https://github.com/sonic-net/sonic-swss/pull/3562

orchagent/port.h
stepanblyschak: I see it is used to set default m_mtu value. Later the code in getPortMtu reads it on start from SAI and substracts MAX_MACSEC_SECTAG_SIZE. 
The default for SAI_PORT_ATTR_MTU is defined by headers, so it always sets m_mtu to the value 1492 - MAX_MACSEC_SECTAG_SIZE, thus theoretically we can save SAI call per port if we set it here to a constant.

Not sure if that was the intention to substract macsec tag size regardless of whether macsec is enabled or not but it is current behaviour.

bingwang-ms:
@stepanblyschak Do we still need this PR since we have separated PRs #3563 and #3562 ?
cc @prsunny

stepanblyschak:
@bingwang-ms Yes, after https://github.com/sonic-net/sonic-swss/pull/3562 is merged I will rebase this PR

prgeor:
@stepanblyschak What optimization are you talking about? How are you measuring the baseline? Link up time?

Ran fast-reboot and measure. Observed 45% reduction in time.

stepanblyschak:
@prgeor Measuring the time it takes to initialize all ports in fast boot. Specifically using these log messages:

Before:

```
2025 Jun  3 12:17:26.259927 sonic NOTICE swss#orchagent: :- initializePort: Initializing port alias:Ethernet0 pid:100000000000d
2025 Jun  3 12:17:33.519707 sonic NOTICE swss#orchagent: :- initPort: Initialized port Ethernet512
```

Takes 7.26 sec.

After:

```
2025 Jun  3 08:58:51.220497 sonic NOTICE swss#orchagent: :- initPortsBulk: initPortsBulk took 3.541895 sec
```

prgeor:
@stepanblyschak cherry pick labels does not look correct? 202205?

stepanblyschak:
@prsunny Thanks for pointing out, it's not correct, I changed the labels

orchagent/portsorch.cpp
```diff
@@ -92,6 +93,50 @@ struct PortAttrValue
...
+        }
+
+        std::vector<sai_attribute_t*> attrs(count);
+
+        for (size_t idx = 0; idx < count; idx++)
+        {
+            attrs[idx] = &attrList[idx];
+        }
+
+        sai_port_api->get_ports_attribute(count, oids.data(), attrCount.data(),
```
prgeor, line 138: @stepanblyschak where is this get used?

orchagent/port.h
prgeor: @stepanblyschak this change does not look to be relevant to port initialization speed up. can we take this separately?

bingwang-ms:
@prgeor Can you help review please?

orchagent/portsorch.cpp
stepanblyschak: @prgeor This is inside executeGet(), executeGet() called by initializePriorityGroupsBulk, initializeQueuesBulk, initializePortHostTxReadyBulk

orchagent/port.h
stepanblyschak: @prgeor It is related. We have two flows of port initialization - 
1. SAI starts with default ports and we get MTU in getPortMtu() which calls SAI get API.
2. We create ports not providing MTU value. In this case the MTU should be default as defined by SAI headers. So I avoid getting it from SAI.
This define is used to initialize m_mtu variable. Since getPortMtu() always substracts MAX_MACSEC_SECTAG_SIZE I updated this define to match the value that would have been returned by getPortMtu().

stepanblyschak:
> @stepanblyschak this change does not look to be relevant to port initialization speed up. can we take this separately?

@prgeor It is related. We have two flows of port initialization -

- SAI starts with default ports and we get MTU in getPortMtu() which calls SAI get API.
- We create ports not providing MTU value. In this case the MTU should be default as defined by SAI headers. So I avoid getting it from SAI.

This define is used to initialize m_mtu variable. Since getPortMtu() always substracts MAX_MACSEC_SECTAG_SIZE I updated this define to match the value that would have been returned by getPortMtu().

stepanblyschak:
This PR build break because dash API change: https://github.com/sonic-net/sonic-dash-api/pull/42

prsunny:
@zjswhhh , @theasianpianist , could you please check and unblock?

orchagent/port.h
stepanblyschak: Discussed online with @prgeor : keep the mtu change seperate from this PR.
Initial MTU may be diifferent from SAI spec on some platforms.

stepanblyschak:
@prgeor MTU change reverted as discussed

orchagent/portsorch.cpp
```diff
@@ -3734,109 +3744,139 @@ bool PortsOrch::initPort(const PortConfig &port)
...
         }
+
+        SWSS_LOG_NOTICE("Initialized port %s", alias.c_str());
     }
-    else
+
+    return status;
+}
+
+void PortsOrch::registerPort(Port &p)
```
prgeor, line 3897: @stepanblyschak what does registerPort() mean? can you add comments?

stepanblyschak: @prgeor I added a comment on the method. I extracted some code into a seperate function, this function does not do any more SAI calls to init a port. For a lack of better word I chose registerPort (addPort and initPort already used). The conceptual flow is like this - 1) addPortBulk - calls SAI bulk create 2) initPortsBulk - fetches all neccessary info from SAI and fills Port structure 3) registerPort - adds the port the internal data structures, generates counter maps, notifies other Orchs on added port.

stepanblyschak:
@prgeor @prsunny Checkers pass, comments addressed, could you please merge ?

prsunny:
@prgeor , would you please signoff? @vaibhavhd for viz

bingwang-ms:
@prgeor Kindly reminding on this.

orchagent/portsorch.cpp
```diff
@@ -97,6 +97,49 @@ struct PortAttrValue
...
+        }
+
+        std::vector<sai_attribute_t*> attrs(count);
+
+        for (size_t idx = 0; idx < count; idx++)
+        {
+            attrs[idx] = &attrList[idx];
+        }
+
+        sai_port_api->get_ports_attribute(count, oids.data(), attrCount.data(),
```
dgsudharsan, line 138: What if the bulk get is unsupported by a vendor? Is there a fallback mechanism required?

dgsudharsan, line 138: In addition to bulk ports get why can't we leverage bulk set_ports_attribute?

stepanblyschak: Yes, in syncd

stepanblyschak: we can, out of scope for this pr, no significant gains in fast boot

dgsudharsan (approved):

prsunny:
approved for merge as this is pending for few months.

PR #3508, opened 2025-02-10
[portsorch][portchannel] fix portchannel collection distribution

baorliu:
**What I did**
Checking portchannel and its member operational status when set portchannel collection/distribution, triggered either by enable/disable lag member or when portchannel status changes.

**Why I did it**
fix the issue https://github.com/sonic-net/sonic-swss/issues/3507

**How I verified it**
1, add a UT testcases
2, built a sonic image with this fix and tested it in a testbed.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2567,6 +2567,91 @@ bool PortsOrch::setHostIntfsStripTag(Port &port, sai_hostif_vlan_tag_t strip)
     return true;
 }
 
+// Set collection and distribution for all members of a portchannel,
+// or directly set one of the members
+bool PortsOrch::setLagCollectionDistribution(Port &port, bool oper_status)
+{
+    SWSS_LOG_ENTER();
+    vector<Port> portv;
+    bool isLagUp = oper_status;
+    std::string lag_alias = port.m_alias;
+
+    if(port.m_type == Port::TUNNEL)
+    {
+        return true;
+    }
+
+    if (port.m_type == Port::PHY)
+    {
+        Port lag;
+        bool enable;
+        if (port.m_lag_member_id == 0 || port.m_lag_id == 0) {
+            return true;
+        }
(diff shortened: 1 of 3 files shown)


baorliu:
convert to draft for now to solve the unit test issues.

orchagent/portsorch.cpp
```diff
@@ -8172,6 +8262,19 @@ void PortsOrch::updatePortOperStatus(Port &port, sai_port_oper_status_t status)
             SWSS_LOG_ERROR("Failed to set host interface %s operational status %s", port.m_alias.c_str(),
                     isUp ? "up" : "down");
         }
+        if (port.m_lag_member_id != 0 && !setLagCollectionDistribution(port, isUp))
```
prabhataravind, line 8265: I think this should be driven by LACP and should not be done directly here. This can potentially lead to some unwanted race conditions with LACP.

prabhataravind (commented):
I am not sure how this condition can arise. If the portchannel is down on both sides, how is the peer device able to send packets to the DUT?

baorliu:
> I am not sure how this condition can arise. If the portchannel is down on both sides, how is the peer device able to send packets to the DUT?

Hi @prabhataravind ,
This happens in the following topo,
sonic1 <--portchannel_1--> switch<--portchannel_2-->sonic2
 there is a switch in the middle, when portchannel_1 is down and portchannel_2 is up, sonic2 does not know the portchannel status in sonic1 and it could send packet to sonic1.

orchagent/portsorch.cpp
baorliu: As described here https://github.com/sonic-net/sonic-swss/issues/3507, LACP can make portchannel status DOWN but SWSS won't block traffic for the UP member(s).
The action here is to block/unblock traffic (not includes LACP packets) based on portchannel status.

PR #3515, opened 2025-02-13
[swss]add bfd session deletion handling in acceleration logic

baorliu:
This PR handles a corner case of BFD session deletion.
When BFD session is deleted at runtime (without restart swss), no BFD session deletion event is sent to neighorch. the acceleration logic is in neighorch, but neighorch is not aware of this BFD session deletion, so it cannot clear the flag in the future if the flag was set (setNextHopFlag(nhop->first, NHFLAGS_IFDOWN);) and no other event trigger the clear.

**What I did**
1, add a subject type for bfd session deletion.
2, notify consumer when bfd session is deleted in bfdorch
3, handle the new SUBJECT_TYPE_BFD_SESSION_DELETION subject type in neighorch, check port status before clear the flag.

**Why I did it**

**How I verified it**

**Details if related**


orchagent/bfdorch.cpp
```diff
@@ -581,6 +581,11 @@ bool BfdOrch::remove_bfd_session(const string& key)
         }
     }
 
+    BfdUpdate update;
+    update.peer = bfd_session_lookup[bfd_session_id].peer;
+    update.state = SAI_BFD_SESSION_STATE_UP;
```
baorliu, line 586: because the notification subject is SUBJECT_TYPE_BFD_SESSION_DELETION, so the bfd state does not matter here. The reason to use bfd state "SAI_BFD_SESSION_STATE_UP" here is to simplify consumer side design. 
It can also use real current BFD session state in bfdorch, and let consumer check SUBJECT_TYPE_BFD_SESSION_DELETION subject and decide how to handle the deletion subject.

baorliu: decide to use the second approach later

abdosi:
@baorliu do we still need this PR ?

baorliu:
> @baorliu do we still need this PR ?

It is better to fix it. we already hit this issue in a real project. and the fix was verified for that issue.
in the BFD offload for bgp case, some configuration change may remove BFD session but BFD session was DOWN at that time, traffic was totally blocked for ever after that (before restarting swss)

prsunny:
@baorliu , do you've an HLD update on this PR? Would like to review the usecase and design. @abdosi

PR #3517, opened 2025-02-14
Set Port UPDATE_DSCP attribute when TC_TO_DSCP map is attached

kperumalbfn:
**What I did**
- Set Port SAI attribute SAI_PORT_ATTR_UPDATE_DSCP when TC_TO_DSCP map is attached to the port.

**Why I did it**
- Some vendor SAI expects Sonic to set this attribute explicitly when TC_TO_DSCP map is attached to the port to modify DSCP value of the packet.

**How I verified it**
- Verifed after attaching TC_TO_DSCP map to the port, DSCP value of the packet is modified as expected.

**Details if related**


orchagent/qosorch.cpp
```diff
@@ -2088,6 +2088,24 @@ task_process_status QosOrch::handlePortQosMapTable(Consumer& consumer, KeyOpFiel
                         return task_process_status::task_invalid_entry;
                     }
                 }
+                if (attr.id == SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP) {
```
prsunny, line 2091: {-> new line

prsunny, line 2091: Since this attribute is set above, and if implemention doesn't support setting, looks like its failing and returning from above check itself. can you please confirm?

orchagent/qosorch.cpp
```diff
@@ -2195,6 +2213,24 @@ task_process_status QosOrch::handlePortQosMapTable(Consumer& consumer, KeyOpFiel
                     return task_process_status::task_invalid_entry;
                 }
             }
+            if (attr.id == SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP) {
```
prsunny, line 2216: Same comment as above

orchagent/qosorch.cpp
kperumalbfn: Yes @prsunny If the set_attribute API for SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP fails, below code of setting the new SAI attribute(UPDATE_DSCP) will not be invoked.

kperumalbfn: Updated

orchagent/qosorch.cpp
kperumalbfn: Updated

orchagent/qosorch.cpp
prsunny: then why do we need to check capability here?

kperumalbfn: Some vendor SAI/SDK may not support this additional attribute(SAI_PORT_ATTR_UPDATE_DSCP) to update the DSCP.

prsunny: in that case, it would be anyways failing in the previous` if` condition, right?

prsunny: ok

prsunny (approved):

PR #3520, opened 2025-02-18
Added Change to Skip Route Programming if NH is link/oper down

abdosi:
What I did:
Added Change to Skip Route Programming if NH is link/oper down. With Scale Route testing of 60K+ routes when we toggle all the interfaces[14+ interface back to back]  as done here: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/snappi_tests/multidut/bgp/test_bgp_outbound_uplink_multi_po_flap.py we see  because of slowness of FRR Route APP_DB processing compare to Link Notification Handling where we have updated the Nexthop Group as part of Link Notification handling to point to default route via https://github.com/sonic-net/sonic-swss/pull/3389 [if eligible] FRR slowness can reprogram the Route back to Nexthop which is link down.

This change is similar to https://github.com/sonic-net/sonic-swss/pull/3394 which was done for Nexthop Group.

How I verify:
Manual Verification and Ixia tetsing
Also UT has been updated to make sure it passes after this change.


arlakshm (approved):
lgtm

deepak-singhal0408 (approved):

orchagent/routeorch.cpp
```diff
@@ -1870,6 +1870,11 @@ bool RouteOrch::addRoute(RouteBulkContext& ctx, const NextHopGroupKey &nextHops)
                         return false;
                     }
                 }
+                else if (m_neighOrch->isNextHopFlagSet(nexthop, NHFLAGS_IFDOWN))
```
prsunny, line 1873: Just thinking if it impact tunnel route in dualtor case if NHFLAG is down? @Ndancejic , can you check this change?

Ndancejic: I think tunnel routes should be safe, I believe line 1839 would catch interface nexthops. also I think routes to tunnel nh get skipped in doTask:

```
/* skip direct routes to tun0 */
else if (alsv[0] == "tun0")
{
}
```

Ndancejic (approved):

prsunny (approved):

abdosi:
@arlakshm added label for msft/202405

arlakshm:
@abdosi, looks like there is a cherry-pick conflict

kperumalbfn:
@abdosi could you resolve the conflicts for 202411

r12f:
Hi @abdosi , I wonder if you could help with the cherry pick to 202411? Once it lands there, it will be automatically merged into 202412.

r12f:
Creating change directly to 202412 will create a merge conflict in later merges and break the automation. : D

PR #3529, opened 2025-02-22
[portsorch] add more attributes on port creation

stepanblyschak:
Depends on https://github.com/sonic-net/sonic-swss/pull/3506

**What I did**

Extend create_ports() supported attribute list with FEC, TPID, asymetric PFC configuration.

**Why I did it**

On platforms which start with no ports on fast-reboot ```sai_port_api->create_ports()``` is called as it is much more efficient to configure desired port attributes rather then have SAI driver to create default ports with default SAI attribute settings as required by SAI spec. Therefore, the most effective approach is to create ports and pass as much configuration at the time of port creation and not have to call ```sai_port_api->set_port_attribute``` for every attribute that needs to change its value.

**How I verified it**

Manual tests, unit tests, run fast-reboot test.

**Details if related**


StephenWangGoogle (dismissed):

StephenWangGoogle (commented):
approved by mistake

bingwang-ms:
@prgeor Can you help review?

bingwang-ms:
@prgeor Can you help review?

balanokia:
Hi, @stepanblyschak 
As per Gearbox design, all of the attributes have to be calculated and applied on ASIC faceplate port and setGearboxPortsAttr would invoke specific system/line side handling. We need to be aware that, Gearbox ports would have their own state machine in CRUD operations. All of FEC operations have to be encapsulated using setPortFec that would cascade the changes for System and Line as per gearbox port configuration (Tx FIR, Training, FEC). Its impacting port up/down and FEC config updates with external PHYs.  Pls refer, https://github.com/sonic-net/sonic-swss/pull/3660.
I have a pending CSP with Broadcom. I would be creating an issue to remove the following code.  

Can you pls share the logic for these changes?   I would be creating and issue to delete the following code during SET operation.      /*            else
                    {
                        setGearboxPortsAttr(p, SAI_PORT_ATTR_FEC_MODE, &pCfg.fec.value, pCfg.fec.override_fec);
                    }
*/

Also port speed should be following the Gearbox HLD. It needs to be removed as well

stepanblyschak:
@balanokia This PRs addresses an issue where SAI_PORT_ATTR_FEC_MODE is configured twice - when creating a port and then configured again in doPortTask. However, since gearbox port is not available at create the FEC attribute needs to be set at some point. This `else` block that you mentioned is exactly the same as in the code for speed handling. So, basically I applied the same approach as for speed.

balanokia:
Hi, @stepanblyschak. There are 2 types of ports in the system - ASIC hosting and PHY hosting. ASIC MAC port index is at the hierarchically higher level and it needs to be indexed by all SONiC operations. setPortFec would cascase the necessary changes from ASIC to the the PHY hosting. PHY/PAI has a host of configuration and it needs to be computed using gearbox configuration. Its the same design for Speed as well. Its not compliant with the current Gearbox design. Its not clear how you tested using any physical PHY devices. I would be creating an issue to delete the code that calls directly gearbox attributes modification during SET operation.  Thanks.

stepanblyschak:
@balanokia I understand your concern. I do not have any devices to test with PHY ports and rely on UT/VS for this part. I am ok with you removing this block for fec that I added and for speed that was already there as it does not do anything on my platform.

PR #3538, opened 2025-02-28
Initialize Port oper error status map only once

prgeor:
**What I did**
Initialize the port error status map only once 

**Why I did it**

Any change in CONFIG_DB PORT table was resulting in updating the port error status map resulting in change in error count = 0 

**How I verified it**

Verified by raising RF and LF error events on the port

**Details if related**


vdahiya12 (approved):

prsunny (approved):

prgeor:
@kperumalbfn could you cherry pick this bug fix to 202411

PR #3546, opened 2025-03-06
Initialize number of counters to 0 before querying the capabilities

stephenxs:
**What I did**

Initialize the argument `sai_stat_capability_list_t` to a zero size when calling `sai_query_stats_capability` for the first time in `initCounterCapabilities`.

**Why I did it**

Currently, it is initialized as the sum of all the counter IDs that orchagent requires. This works, but it is misleading because the purpose of this SAI API is to return all the counters that are supported by the vendor SAI, regardless of what the SAI client requires.
Usually, it should be either 0 or the number of counters that vendor SAI supports. It is strange to pass any other value for vendor SAI and might indicate an error. Some vendors can report an error in this case.
To avoid this kind of error message, we always pass 0 in this case.

**How I verified it**

Run unit test and manual test.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1535,12 +1535,6 @@ void PortsOrch::initCounterCapabilities(sai_object_id_t switchId)
 {
     sai_stat_capability_list_t queue_stats_capability, port_stats_capability;
 
-    uint32_t  queue_stat_count = (uint32_t) queue_stat_ids.size() +
-                                 (uint32_t) queueWatermarkStatIds.size() +
-                                 (uint32_t) wred_queue_stat_ids.size();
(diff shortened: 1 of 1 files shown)


dgsudharsan (approved):

prsunny:
@stephenxs , can you please fill in "Why I did It"?. What is currently the issue you encounter?

stephenxs:
> @stephenxs , can you please fill in "Why I did It"?. What is currently the issue you encounter?

@prsunny Sure. Update.
In sort, it is not common to pass a value that is neither 0 nor the number of counters that are supported by SAI to that SAI API, which might indicate an error. Some vendor SAIs report error on this.

PR #3547, opened 2025-03-07
Move setting of redis communication mode to initSaiRedis()

prabhataravind:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


prsunny (approved):

orchagent/notifications.cpp
```diff
@@ -18,10 +19,25 @@ void on_fdb_event(uint32_t count, sai_fdb_event_notification_data_t *data)
...
+ * libsairedis in a separate thread which causes concurrency issues.
+ * For platforms which use zmq between orchagent and syncd, it is an acceptable
+ * workaround to forward the notifications from the callback handler to the
+ * redis notifications channel processed by portsorch.
+ */
 void on_port_state_change(uint32_t count, sai_port_oper_status_notification_t *data)
 {
-    // don't use this event handler, because it runs by libsairedis in a separate thread
-    // which causes concurrency access to the DB
+    if (gRedisCommunicationMode == SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC)
```
vivekrnv, line 31: Wouldn't we need this for other handlers as well? I agree DPU's won't be using other handlers but i suppose this change is generic.

PR #3553, opened 2025-03-10
Add support for local ports with asic and dut name in getPort

arista-hpandya:
**What I did**
Modified `PortsOrch::getPort(string alias, Port &p)` method to search for system port name stripped from the device and namespaces. To implement this we introduced `isLocalSystemPortIntf(Port& port)` method to decouple getPort and `isLocalSystemPortIntf(string alias)`

**Why I did it**
Currently, we store local ports and remote ports using different convention in m_portList. Remote ports follow: <dut name>|<asic>|<interface> format and local ports are saved as <interface>. Hence when <dut name>|<asic>|<interface> format is used to 'get' a local port it returns false. This generates a number of "port not found" info messages in the syslog.

**How I verified it**
Testing:
- [x] Verified syntax by compiling the orchagent
- [x] Tested on a multi-asic LC by running T2 tests and identified no failures related to the change. 
- [x] Verified the logs no longer appear in syslog when loglevel is set to INFO

**Details if related**
Fixes https://github.com/sonic-net/sonic-buildimage/issues/20214

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -1458,6 +1458,22 @@ bool PortsOrch::getPort(string alias, Port &p)
 
     if (m_portList.find(alias) == m_portList.end())
     {
+        // For local ports strip the asic names and check for interface name
+        size_t pos = alias.find('|');
+        if (pos != std::string::npos) {
+            pos = alias.find_last_of('|');
+            string intf_name = alias.substr(pos + 1);
+
+            if (m_portList.find(intf_name) != m_portList.end())
+            {
+                Port port = m_portList[intf_name];
+                if (gIntfsOrch->isLocalSystemPortIntf(port))
+                {
+                    p = port;
+                    return true;
+                }
+            }
+        }
         return false;
     }
     else
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -1652,17 +1652,20 @@ bool IntfsOrch::isRemoteSystemPortIntf(string alias)
     return false;
 }
 
+bool IntfsOrch::isLocalSystemPortIntf(Port& port)
+{
+    if (port.m_type == Port::LAG)
+        return(port.m_system_lag_info.switch_id == gVoqMySwitchId);
+    else
+        return(port.m_system_port_info.type != SAI_SYSTEM_PORT_TYPE_REMOTE);
+}
+
 bool IntfsOrch::isLocalSystemPortIntf(string alias)
 {
     Port port;
     if(gPortsOrch->getPort(alias, port))
     {
-        if (port.m_type == Port::LAG)
-        {
-            return(port.m_system_lag_info.switch_id == gVoqMySwitchId);
-        }
-
-        return(port.m_system_port_info.type != SAI_SYSTEM_PORT_TYPE_REMOTE);
+        return isLocalSystemPortIntf(port);
     }
     //Given alias is system port alias of the local port/LAG
     return false;
--- a/orchagent/intfsorch.h
+++ b/orchagent/intfsorch.h
@@ -72,6 +72,7 @@ public:
 
     bool isRemoteSystemPortIntf(string alias);
     bool isLocalSystemPortIntf(string alias);
+    bool isLocalSystemPortIntf(Port& port);
     void voqSyncIntfState(string &alias, bool);
 
 private:


arista-hpandya:
The PR has been updated based on the feedback, can I get someone to review it? 
cc: @arlakshm

arista-hpandya:
Pipeline failure seem unrelated to the PR:

```===End Flaky Test Report===
=================== 12 passed, 1 xpassed in 1713.59 seconds ====================
+ break

##[debug]Process exited with code 123 and signal null for tool '/usr/bin/bash'
##[debug]STDIO streams have closed and received exit code 123 and signal null for tool '/usr/bin/bash'
##[error]Bash exited with code '123'.
##[debug]Processed: ##vso[task.issue type=error;source=TaskInternal;correlationId=4f608eee-f6ff-44e2-a353-0e0f86c4fef6;]Bash exited with code '123'.
##[debug]task result: Failed
##[debug]Processed: ##vso[task.complete result=Failed;done=true;]
Finishing: Run vs tests
```

vmittal-msft (approved):

vmittal-msft:
@arista-hpandya please rebase.

orchagent/portsorch.cpp
```diff
@@ -5652,8 +5652,16 @@ void PortsOrch::doLagMemberTask(Consumer &consumer)
         Port lag, port;
         if (!getPort(lag_alias, lag))
         {
-            SWSS_LOG_INFO("Failed to locate LAG %s", lag_alias.c_str());
-            it++;
+            size_t pos = lag_alias.find('|');
+            std::string port_hostname = (pos != std::string::npos) ? lag_alias.substr(0, pos) : lag_alias;
+
+            if (gMyHostName != port_hostname){
```
prsunny, line 5658: this seems to be not voq specific.

arista-hpandya: Yes gMyHostName is only defined for VOQ, will extend the conditional to check for gMySwitchType

arlakshm:
Please add UT for this change

orchagent/portsorch.cpp
```diff
@@ -5652,8 +5652,16 @@ void PortsOrch::doLagMemberTask(Consumer &consumer)
         Port lag, port;
         if (!getPort(lag_alias, lag))
         {
-            SWSS_LOG_INFO("Failed to locate LAG %s", lag_alias.c_str());
-            it++;
+            size_t pos = lag_alias.find('|');
+            std::string port_hostname = (pos != std::string::npos) ? lag_alias.substr(0, pos) : lag_alias;
+
+            if (gMySwitchType == "voq" && gMyHostName != port_hostname){
```
prsunny, line 5658: you must do the whole change if Switchtype is "voq".

orchagent/portsorch.cpp
```diff
@@ -5652,8 +5652,16 @@ void PortsOrch::doLagMemberTask(Consumer &consumer)
...
+                SWSS_LOG_INFO("Failed to locate LAG %s", lag_alias.c_str());
+                it++;
+            }
+            else {
```
prsunny, line 5662: `{ `-> next line here and the above if case.

arista-hpandya: Done

orchagent/portsorch.cpp
arista-hpandya: I modified the code to make sure we drop the consumer and do not trigger the syslog message if the system is VOQ and the hostname matches. Otherwise the default behaviour is retained. Can you confirm this is what you meant by `you must do the whole change if Switchtype is "voq"`?

orchagent/portsorch.cpp
```diff
@@ -5688,8 +5688,18 @@ void PortsOrch::doLagMemberTask(Consumer &consumer)
         Port lag, port;
         if (!getPort(lag_alias, lag))
         {
-            SWSS_LOG_INFO("Failed to locate LAG %s", lag_alias.c_str());
-            it++;
+            size_t pos = lag_alias.find('|');
```
prsunny, line 5691: I meant, this change can be place under `gMySwitchType == "voq"`

arista-hpandya: Sure, made the change!

judyjoseph:
@prsunny  could you review and merge pls

judyjoseph (approved):

abdosi:
@prsunny : can you help sign-off

orchagent/neighorch.cpp
```diff
@@ -1839,7 +1840,9 @@ void NeighOrch::doVoqSystemNeighTask(Consumer &consumer)
 
         string alias = key.substr(0, found);
 
-        if(gIntfsOrch->isLocalSystemPortIntf(alias))
+        size_t pos = alias.find('|');
+        std::string port_hostname = (pos != std::string::npos) ? alias.substr(0, pos) : alias;
+        if(gIntfsOrch->isLocalSystemPortIntf(alias) || (gMySwitchType == "voq" && gMyHostName == port_hostname))
```
BYGX-wcr, line 1845: @arista-hpandya, won't this change cause the orchagent to skip VoQ system neighbor task for neighbor ASICs on the LC?

@arlakshm for viz.

PR #3562, opened 2025-03-18
[portsorch] postpone non-critical port init part in warm/fast-reboot

stepanblyschak:
**What I did**

Port initialization that has no dependncy on subsequent data plane configuration are extracted into a seperate "post initialization" routine and executed after data plane configuration is done.

**Why I did it**

To save time in fast-boot.

**How I verified it**

Boot system with many ports, notice a time difference in port intialization time

**Details if related**

### DIFF
--- a/orchagent/orch.h
+++ b/orchagent/orch.h
@@ -295,6 +295,16 @@ public:
     virtual void doTask(swss::NotificationConsumer &consumer) { }
     virtual void doTask(swss::SelectableTimer &timer) { }
 
+    /*
+     * Called once after APPLY_VIEW in warm/fast boot scenario.
+     * Orch can override this method to perform orch specific operations after boot is finished.
+     * These operations are not meant to produce additional ASIC configuration,
+     * instead a capability fetch and STATE_DB update here is encouraged.
+     * Orch is not expected to call the base method implementation as it must remain
+     * empty for compatibility reasons.
+     */
+    virtual void onWarmBootEnd() { }
+
     void dumpPendingTasks(std::vector<std::string> &ts);
 
     /**
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -1094,8 +1094,10 @@ bool OrchDaemon::warmRestoreAndSyncUp()
 
     syncd_apply_view();
 
-    /* Start dynamic state sync up */
-    gPortsOrch->refreshPortStatus();
+    for (Orch *o : m_orchList)
+    {
+        o->onWarmBootEnd();
+    }
 
     /*
      * Note. Arp sync up is handled in neighsyncd.
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -34,6 +34,7 @@
 #include "switchorch.h"
 #include "stringutility.h"
 #include "subscriberstatetable.h"
+#include "warm_restart.h"
 
 #include "saitam.h"
 
@@ -579,7 +580,8 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
                 ref(wred_port_stat_manager),
                 ref(wred_queue_stat_manager)
             }),
-        m_port_state_poller(new SelectableTimer(timespec { .tv_sec = PORT_STATE_POLLING_SEC, .tv_nsec = 0 }))
+        m_port_state_poller(new SelectableTimer(timespec { .tv_sec = PORT_STATE_POLLING_SEC, .tv_nsec = 0 })),
+        m_isWarmRestoreStage(WarmStart::isWarmStart())
 {
     SWSS_LOG_ENTER();
 
@@ -3776,6 +3778,12 @@ bool PortsOrch::initPort(const PortConfig &port)
                     createPortBufferQueueCounters(p, to_string(p.m_host_tx_queue), false);
                 }
 
+                // In warm-reboot postPortInit is postponed.
+                if (!m_isWarmRestoreStage)
+                {
+                    postPortInit(m_portList[alias]);
+                }
+
                 SWSS_LOG_NOTICE("Initialized port %s", alias.c_str());
             }
             else
@@ -4186,9 +4194,6 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         continue;
                     }
 
-                    initPortSupportedSpeeds(it->second.key, m_portListLaneMap[it->first]);
-                    initPortSupportedFecModes(it->second.key, m_portListLaneMap[it->first]);
-
                     it++;
                 }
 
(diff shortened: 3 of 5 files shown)


nazariig (approved):

prgeor:
@stepanblyschak can you explain what do you mean by port here? Are these referring to front panel port?

stepanblyschak:
> @stepanblyschak . Please add more details. What is the target initialization time? How much was the previous vs now?
> 
> How I verified it
> 
> Boot system with many ports, notice a time difference in port intialization time

@prgeor Shared testing data in teams

orchagent/orchdaemon.cpp
```diff
@@ -1053,8 +1053,10 @@ bool OrchDaemon::warmRestoreAndSyncUp()
 
     syncd_apply_view();
 
-    /* Start dynamic state sync up */
-    gPortsOrch->refreshPortStatus();
+    for (Orch *o : m_orchList)
+    {
```
prgeor, line 1057: @stepanblyschak where does the refreshPortStatus() happen now?

orchagent/portsorch.cpp
```diff
@@ -5843,6 +5845,40 @@ void PortsOrch::doLagMemberTask(Consumer &consumer)
...
+{
+    SWSS_LOG_ENTER();
+
+    if (gMySwitchType != "dpu")
+    {
+        initializePortBufferMaximumParameters(p);
+    }
+
+    initPortSupportedSpeeds(p.m_alias, p.m_port_id);
+    initPortSupportedFecModes(p.m_alias, p.m_port_id);
```
prgeor, line 5879: @stepanblyschak these two are called in doPortTask(), so this is no longer needs to be called in doPortTask()?

orchagent/orchdaemon.cpp
stepanblyschak: @prgeor Moved to PortsOrch::onWarmBootEnd(), seems it was a good place for refreshPortStatus() to happen as well

orchagent/portsorch.cpp
stepanblyschak: In the original flow ```initPortSupportedSpeeds()``` and ```initPortSupportedFecModes()``` is called twice - when initializing a port and inside doPortTask() and I do not change that.
Both methods are checking if supported speeds and fec modes were queried before and in that case do nothing - https://github.com/sonic-net/sonic-swss/blob/master/orchagent/portsorch.cpp#L3012.

The subsequent calls in doPortTask() might be unnecccesary, but I do not want to change that in this PR.

prgeor (dismissed):

stepanblyschak:
@prgeor Could you please merge?

orchagent/orchdaemon.cpp
vaibhavhd: Why does this change only need to be made for fast/warm reboot? Do you see concerns with making this a standard process for all kinds of reboots (eg., cold boot)?

Asking so that we do not have too many difference in boot up sequence in for diff kinds of reboots. We already have too many, and should avoid adding more if we can.

orchagent/portsorch.cpp
```diff
@@ -6137,7 +6173,6 @@ bool PortsOrch::initializePort(Port &port)
         initializePriorityGroups(port);
         initializeQueues(port);
         initializeSchedulerGroups(port);
-        initializePortBufferMaximumParameters(port);
```
vaibhavhd, line 6142: Why is this change made to move init for buffer parms?

orchagent/orch.h
```diff
@@ -295,6 +295,16 @@ class Orch
     virtual void doTask(swss::NotificationConsumer &consumer) { }
     virtual void doTask(swss::SelectableTimer &timer) { }
 
+    /*
+     * Called once after APPLY_VIEW in warm/fast boot scenario.
+     * Orch can override this method to perform orch specific operations after boot is finished.
+     * These operations are not meant to produce additional ASIC configuration,
+     * instead a capability fetch and STATE_DB update here is encouraged.
+     * Orch is not expected to call the base method implementation as it must remain
```
vaibhavhd, line 303: > These operations are not meant to produce additional ASIC configuration. instead a capability fetch and STATE_DB update here is encouraged

Is this change in expectation? What happens (before and after this change) when these operations result in additional ASIC configuration?

Also, how do you propose to enforce this expectation and add tests to prevent this?
When there is true capability change (from an old SAI/SDK to new, for eg.) across warm/fast reboot then how to still take that into effect?

vaibhavhd:
I don't see UTs or SWSS VS tests for this. Can you add them?

orchagent/portsorch.cpp
stepanblyschak: @vaibhavhd This function sets max_headroom_size in STATE_DB for use by buffermgrd for buffer calculations. In fast boot, buffer settings already present in APPL_DB/CONFIG_DB. Also, no code uses max_headroom_size in orchagent.

orchagent/orch.h
stepanblyschak: @vaibhavhd In this change no configuration is done in PortsOrch::onWarmBootEnd.
If additional ASIC configuration is done, it is after APPLY_VIEW - depending on what exact configuration is done it could be either syncd crash, no impact at all or comparison logic removing configs from ASIC, resulting in traffic drops.
Can be enforced by flag in sairedis to disallow create/set/remove operations, but IMO it's unlikelly anyone will put some configuration under this method instead of usual doTask(), but would like to know your opinion

stepanblyschak:
> I don't see UTs or SWSS VS tests for this. Can you add them?

No new functionality has been added, diff coverage satisfied by existing tests

orchagent/orchdaemon.cpp
stepanblyschak: @vaibhavhd agree, would like to have same flow for all boot types. Unfortunatelly no indication of "_configuration end_" is available for cold boot - APPL_DB stats empty and is filled by mgrd/sync deamons incrementally with no indication of when initial configuration has been set. Unlike cold boot, in fast/warm we know when all configuration has been set to perform delayed post init routines.

prsunny:
@vaibhavhd , approve this?

prsunny:
> I don't see UTs or SWSS VS tests for this. Can you add them?

The modified code must have some scenario-based UT coverage. Is there any limitation in adding tests?

stepanblyschak:
@prgeor @prsunny @vaibhavhd Added UT, please check

prsunny:
> @prgeor @prsunny @vaibhavhd Added UT, please check

Can you please add a description of how the UT is done? From the code change, i dont fully understand whats covered by UT

stepanblyschak:
> > @prgeor @prsunny @vaibhavhd Added UT, please check
> 
> Can you please add a description of how the UT is done? From the code change, i dont fully understand whats covered by UT

@prsunny 
The test covers postPortInit() called after a call to onWarmBootEnd().
I don't find this UT was neccessary in the first place, as this code path is already covered by other test cases, but since requested I added one.

stepanblyschak:
@prsunny @prgeor @vaibhavhd Can be merged ?

liat-grozovik:
@prgeor @vaibhavhd the approval is lost after PR was updated with unit test. Can you please help to review and merge? this one is blocking others to get in as well.

PR #3563, opened 2025-03-18
[portsorch] pass oper status at hostif creation

stepanblyschak:
**What I did**

Save SAI set API call when possible to provide a value during creation.

**Why I did it**

Save time on fast boot.

**How I verified it**

Test fast-boot, make sure host-if brought up after boot.

**Details if related**

Extracted from https://github.com/sonic-net/sonic-swss/pull/3506


dgsudharsan (approved):

stepanblyschak:
@prsunny Can this be merged ?

PR #3565, opened 2025-03-20
Add Loopback interface admin status handling

anish-n:
**What I did**
Added admin_status field and associated handling to shut/unshut a Loopback interface. 

Table example with Lo admin shut:
```
LOOPBACK_INTERFACE|Loopback0:
{
    "admin_status": "down"
} 
```
**Why I did it**
To support shutting down a loopback interface using the standard SONiC way via config DB and CLIs. Upon shutting down the loopback interface, the associated Linux interface goes down, and if this Loopback is announced via BGP, the IP gets auto-withdrawn from neighbors

**How I verified it**
Using SWSS DVS based test, and manual test in VS docker.

**Details if related**


cfgmgr/intfmgr.cpp
```diff
@@ -843,6 +842,28 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
...
+            }
+            else if (adminStatus != "up" && adminStatus != "down")
+            {
+                SWSS_LOG_ERROR("Got incorrect value for admin_status as %s for intf %s, defaulting as up", adminStatus.c_str(), alias.c_str());
+                adminStatus = "up";
+            }
+
+            try
+            {
+                string intf_admin = setHostSubIntfAdminStatus(alias, adminStatus, "up");
```
prsunny, line 858: why do we call `subintf `api here?

anish-n: the ```subintf```` API functions like a regular set interface admin status API when the parent admin status is passed in as "up", I used this API in order to leverage common code. Changed the approach a bit: modified the code to introduce a new fn setIntfAdminStatus, and called setIntfAdminStatus from within setHostSubIntfAdminStatus to keep common code across both subintf and loopback interface.

cfgmgr/intfmgr.cpp
```diff
@@ -843,6 +857,30 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
...
+            {
+                SWSS_LOG_ERROR("Got incorrect value for admin_status as %s for intf %s, defaulting as up", adminStatus.c_str(), alias.c_str());
+                adminStatus = "up";
+            }
+
+            try
+            {
+                if(setIntfAdminStatus(alias, adminStatus))
+                {
+                    FieldValueTuple newAdminFvTuple("admin_status", adminStatus);
```
prsunny, line 875: is this new attribute for Loopback handled in orchagent? I believe, it should be ok. but can you please check?

cfgmgr/intfmgr.cpp
```diff
@@ -843,6 +857,30 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
...
+            }
+            else if (adminStatus != "up" && adminStatus != "down")
+            {
+                SWSS_LOG_ERROR("Got incorrect value for admin_status as %s for intf %s, defaulting as up", adminStatus.c_str(), alias.c_str());
+                adminStatus = "up";
+            }
+
+            try
+            {
+                if(setIntfAdminStatus(alias, adminStatus))
```
prsunny, line 873: space after` if`

cfgmgr/intfmgr.cpp
anish-n: admin_status handling is skipped in intforch for loopback:
https://github.com/sonic-net/sonic-swss/blob/master/orchagent/intfsorch.cpp#L877

The set operation upon admin_status update will basically return after ensuring that the interface itself exists in syncd, and not perform any change as such.

cfgmgr/intfmgr.cpp
anish-n: Fixed, thanks

cfgmgr/intfmgr.cpp
```diff
@@ -843,6 +857,30 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
                 m_loopbackIntfList.insert(alias);
                 SWSS_LOG_INFO("Added %s loopback interface", alias.c_str());
             }
+
+            if (adminStatus.empty())
+            {
+                adminStatus = "up";
+            }
+            else if (adminStatus != "up" && adminStatus != "down")
```
prsunny, line 865: Check seems incorrect. Shouldn't it be `||` ?

cfgmgr/intfmgr.cpp
```diff
@@ -843,6 +857,30 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
...
+            {
+                SWSS_LOG_ERROR("Got incorrect value for admin_status as %s for intf %s, defaulting as up", adminStatus.c_str(), alias.c_str());
```
prsunny, line 867: Since ERR is handled, please change to WARN level

cfgmgr/intfmgr.cpp
```diff
@@ -843,6 +857,30 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
...
+            {
+                if (setIntfAdminStatus(alias, adminStatus))
+                {
+                    FieldValueTuple newAdminFvTuple("admin_status", adminStatus);
+                    data.push_back(newAdminFvTuple);
+                }
+            }
+            catch (const std::runtime_error &e)
+            {
+                SWSS_LOG_NOTICE("Lo interface ip link set admin status %s failure. Runtime error: %s", adminStatus.c_str(), e.what());
```
prsunny, line 881: Please change to WARN instead of NOTICE

cfgmgr/intfmgr.cpp
anish-n: It should be ```&&```. If we make it ```||``` then we will end up reporting that we got an incorrect value even if the adminStatus gets passed in as ```down``` when it hits the 1st adminStatus check.

cfgmgr/intfmgr.cpp
```diff
@@ -843,6 +857,30 @@ bool IntfMgr::doIntfGeneralTask(const vector<string>& keys,
...
+                return false;
```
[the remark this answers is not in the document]
anish-n: I think that it won't be fully backward compatible cause the following piece of code is removed from addLoopbackIntf ```cmd << IP_CMD << " link set " << alias << " up";```, this logic is now moved to setIntfAdminStatus, so if the setting of admin status fail, we probably want it to retry? 

Alternatively, we can change it to not return here and let it move foward, that way it will be most similar to what happens today, but still won't be 100% backward compatible because the loopback may remain in down state due to the change in setIntfAdminStatus.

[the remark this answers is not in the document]
prsunny: agree.. lets remove the return.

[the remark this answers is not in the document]
anish-n: Done, thanks

cfgmgr/intfmgr.cpp
anish-n: Changed thanks

prsunny (approved):

PR #3566, opened 2025-03-21
[fgnhgorch] Fix incorrect state DB updates due to unordered field-val…

inspurSDN:
…ue tuples

### What I did
Fixed an issue where the order of field-value tuples retrieved from the state DB was not guaranteed, causing incorrect updates when using `fvs[index] = fv;`.

### How I did it
- Iterated through the field-value tuples to find the entry matching the given `index`.
- Updated the matching entry instead of assuming the order.
- If no matching entry was found, added a new one.
- Ensured the final `set()` call happens after all updates.

### How to verify it
1. Configure FG NHG and ensure routes are populated in the state DB.
2. Modify existing routes and check that the correct entries are updated.
3. Verify logs to confirm that updates occur only for matching indexes.
4. Ensure new entries are added when necessary.


prsunny:
@anish-n to review

inspurSDN:
@anish-n please help to review it

orchagent/fgnhgorch.cpp
```diff
@@ -207,23 +207,30 @@ void FgNhgOrch::setStateDbRouteEntry(const IpPrefix &ipPrefix, uint32_t index, N
     // check if profile already exists - if yes - skip creation
     m_stateWarmRestartRouteTable.get(key, fvs);
 
-    //bucket rewrite
-    if (fvs.size() > index)
+    bool isUpdated = false;
```
anish-n, line 210: Instead of this logic, why don't we just do an hset(as opposed to the get +set) on the table with key as prefix, and fv pair as <index, nh key>.

anish-n (changes requested):

anish-n:
> @anish-n please help to review it

@inspurSDN have reviewed the change, please take a look at the comments provided, thanks!

inspurSDN:
> > @anish-n please help to review it
> 
> @inspurSDN have reviewed the change, please take a look at the comments provided, thanks!

Hi @anish-n 
I have already finished the modification and verify by your suggestion, please help to review it again, thanks!

anish-n (approved):

prsunny (approved):
@vaibhavhd for viz

PR #3570, opened 2025-03-25
[wip] Normalize include header paths

lmeninato:
**What I did**
Ran the following script:

```
#!/usr/bin/python3

import subprocess
import os
import sys

# need to brew install gnu-sed if running on mac
SED_CMD = "gsed" if sys.platform == "darwin" else "sed"

DIRS = [
    "lib",
    "orchagent",
    "cfgmgr",
    "portsyncd",
    "fpmsyncd",
    "gearsyncd",
    "neighsyncd",
    "teamsyncd",
    "tlm_teamd",
    "fdbsyncd",
    "mclagsyncd",
    "natsyncd",
    "swssconfig",
    "warmrestart",
    "tests",
]

ALL_HEADERS = subprocess.run(
    ["find", *DIRS, "-type", "f", "-name", "*.h"],
    capture_output=True,
    text=True,
    check=True,
).stdout.splitlines()

def normalize_includes_find_sed(directory):
    """
    Normalizes include paths using find and sed commands.

    run:

    find . -name "*.cpp" -o -name "*.h" | awk -F/ '{print $NF}' | sort | uniq -c | awk '$1 > 1 {print $2}'

    to check if there are duplicates to be wary of (would require an additional patch to manually handle conflicts)

    Args:
        directory (str): The directory to search for files in to normalize headers.
    """
    header_files = set(ALL_HEADERS)
    for header_path in header_files:
        header_filename = os.path.basename(header_path)
        # Escape slashes for sed
        escaped_new_include = header_path.replace("/", "\\/")
        cmd = f'find {directory} -type f \( -name "*.cpp" -o -name "*.h" \) -exec {SED_CMD} -i \'s/#include ["<]{header_filename}["<]/#include "{escaped_new_include}"/g\' {{}} \;'
        os.system(cmd)

        parent_dirs = header_path.split("/")[:-1]
        for i in range(1, len(parent_dirs)):
            # given path a/b/c.h -> replace instances of c.h, b/c.h, with a/b/c.h
            possible_filename = "/".join(parent_dirs[1:i+1]) + "/" + header_filename
            escaped_possible_filename = possible_filename.replace("/", "\\/")
            cmd = f'find {directory} -type f \( -name "*.cpp" -o -name "*.h" \) -exec {SED_CMD} -i \'s/#include ["<]{escaped_possible_filename}["<]/#include "{escaped_new_include}"/g\' {{}} \;'
            os.system(cmd)

        print(f"Replaced includes for {header_filename}")

    print("All replacements completed.")

if __name__ == "__main__":
    for dir in DIRS:
        normalize_includes_find_sed(dir)
```

**Why I did it**

This normalizes all include paths in *.cpp and *.h files. Rather than needing to `-I dir` to find some header `a.h` in `orchagent/dir/a.h`,  by converting all references to `a.h` to the full path relative to the root of the project: `orchagent/dir/a.h`, we no longer need to pass in `-I dir` as a g++ flag.

This should both improve compilation times slightly (fewer locations to lookup header files) and it should improve readability (no need to guess where a header file is coming from).

**How I verified it**

No change to logic. All existing tests should pass. (todo: modify makefile.am files to not add unnecessary include paths to g++).

**Details if related**


mint570 (approved):

lmeninato:
@prsunny Do you mind taking a look? I've noticed that both the latest commit on master and this PR have had the same failing test issue at CI time. All this PR does is re-write include header paths to be more explicit (vs relying on include paths being added), so it shouldn't change code behavior. 

Eventually, I'm hoping to upstream a change where internally at Google we've migrated to Bazel, which should help with build times significantly. Since bazel does not allow including sibling directories as c++ compilation option (e.g. `-I ../a` is not allowed), this is a necessary first step. I also think it will make it a bit easier to know where a header is coming from if the path is normalized relative to the workspace root.

lmeninato:
@prsunny  @liushilongbuaa 

Same failures here in CI as in latest master commit:

https://dev.azure.com/mssonic/build/_build/results?buildId=822171&view=logs&jobId=3f6395b2-1619-5ebe-f305-2aedcf353cb5&j=3f6395b2-1619-5ebe-f305-2aedcf353cb5&t=56fb64f8-b013-579d-11b0-5843431d20c9

lmeninato:
@prsunny Do you mind taking a look again - CI is passing now!

PR #3580, opened 2025-03-31
AclOrch changes for Inner-Source-Mac-Rewrite

kavyavanibedida:
**What I did**
Added orchagent changes in aclorch for inner source mac rewrite feature- that rewrites INNER SRC MAC for VXLAN packet
- new MATCH INNER_SRC_IP
- new ACTION INNER_SRC_MAC_REWRITE_ACTION
- new aclrule AclRuleInnerSrcMacRewrite class
   -  3 new methods (validateAddAction, validate, onupdate)
- Mock tests UT (1 test case) covers
   -  Creation/Deletion of ACL table of the type inner src mac rewrite
   -  Creation/Deletion of ACL rule of inner src mac rewrite
   -  Verification of the action
   -  Verification of the rule update(TUNNEL VNI, INNER SRC IP with/without prefixes, ACTION)
   -  Verification of the ACL rule counter
- VS tests (3 testcases) covers
   -  Creation/Deletion of ACL table of the type inner src mac rewrite
   -  Creation/Deletion of ACL rules for the table type inner src mac rewrite
   -  Verification of the rule update(VNI, SRC IP with prefixes, ACTION) with ASIC entries, Verification of the ACL rule counter 

**Reference PRs**

- Associated SAI API PR https://github.com/opencomputeproject/SAI/pull/2079/files
- Associated HLD PR https://github.com/sonic-net/SONiC/pull/1961
- Associated Yang Model PR https://github.com/sonic-net/sonic-buildimage/pull/21906/files

**Why I did it**
Added orchagent changes for rewriting of inner src mac of VXLAN packets

**How I verified it**
- mock test UT
- VS test

**Details if related**


orchagent/aclorch.cpp
```diff
@@ -1060,7 +1073,7 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
             matchData.data.u8 = to_uint<uint8_t>(attr_value);
             matchData.mask.u8 = 0xFF;
         }
-        else if (attr_name == MATCH_SRC_IP || attr_name == MATCH_DST_IP)
+        else if (attr_name == MATCH_SRC_IP || attr_name == MATCH_DST_IP || attr_name == MATCH_INNER_SRC_IP)
```
SoumyaMishra18, line 1076: Do we need to explicitly check for action here?

anish-n:
Could you please link the SAI API PR in the description here?

kavyavanibedida:
> Could you please link the SAI changes PR in the description here?

Added in description

orchagent/aclorch.cpp
kavyavanibedida: yes, inner_src_ip is added as new match type, so its required to add here

tests/mock_tests/aclorch_ut.cpp
```diff
@@ -2056,4 +2056,228 @@ namespace aclorch_test
...
+        );
+
+        orch->doAclRuleTask(
+            deque<KeyOpFieldsValuesTuple>(
+                {
+                    {
+                        aclTableName + "|" + aclRuleName,
+                        SET_COMMAND,
+                        {
+                            { MATCH_INNER_SRC_IP, "1.1.1.1/32" },
```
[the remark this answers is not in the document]
kavyavanibedida: i have added the /24 route in the same ut
image

orchagent/aclorch.cpp
```diff
@@ -2133,6 +2150,65 @@ void AclRulePacket::onUpdate(SubjectType, void *)
...
+    {
+        return false;
+    }
+
+    return true;
+ }
+
+ void AclRuleInnerSrcMacRewrite::onUpdate(SubjectType type, void *cntx)
+ {
+    //do nothing
```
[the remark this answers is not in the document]
kavyavanibedida: from my understanding, AclRule::Update is updating the mac address, vni and src ip as below 
image

orchagent/aclorch.cpp
```diff
@@ -167,7 +173,8 @@ static const acl_capabilities_t defaultAclActionsSupported =
             {
                 SAI_ACL_ACTION_TYPE_PACKET_ACTION,
                 SAI_ACL_ACTION_TYPE_MIRROR_INGRESS,
-                SAI_ACL_ACTION_TYPE_NO_NAT
+                SAI_ACL_ACTION_TYPE_NO_NAT,
+                SAI_ACL_ACTION_TYPE_SET_INNER_SRC_MAC
```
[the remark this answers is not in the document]
kavyavanibedida: removed the action

orchagent/aclorch.cpp
```diff
@@ -2133,6 +2150,65 @@ void AclRulePacket::onUpdate(SubjectType, void *)
...
+
+    actionData.enable = true;
+    return setAction(aclL3InnerActionLookup[action_str], actionData);
+ }
+
+ bool AclRuleInnerSrcMacRewrite::validate()
+ {
+    SWSS_LOG_ENTER();
+
+    if ((m_rangeConfig.empty() && m_matches.size() != 2) || m_actions.size() != 1 )
```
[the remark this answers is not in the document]
kavyavanibedida: removed the match size limitation

orchagent/aclorch.cpp
```diff
@@ -2133,6 +2150,65 @@ void AclRulePacket::onUpdate(SubjectType, void *)
...
+            catch (invalid_argument &e)
+            {
+                SWSS_LOG_ERROR("Mac address in the wrong format");
+                return false;
+            }
+            memcpy(actionData.parameter.mac ,  MacAddress(_attr_value).getMac(), sizeof(sai_mac_t));
+            action_str = ACTION_INNER_SRC_MAC_REWRITE_ACTION;
+            SWSS_LOG_INFO("Converting the mac address to SAI acl action parameter");
+        }    
+        else{
```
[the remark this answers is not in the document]
kavyavanibedida: added the space

orchagent/aclorch.cpp
```diff
@@ -1769,6 +1782,10 @@ shared_ptr<AclRule> AclRule::makeShared(AclOrch *acl, MirrorOrch *mirror, DTelOr
         {
             return make_shared<AclRulePacket>(acl, rule, table);
         }
+        else if (aclL3InnerActionLookup.find(action) != aclL3InnerActionLookup.cend() || action == ACTION_INNER_SRC_MAC_REWRITE_ACTION)
```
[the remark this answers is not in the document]
kavyavanibedida: removed the check. thanks for pointing it out

orchagent/aclorch.cpp
```diff
@@ -110,6 +111,11 @@ static acl_rule_attr_lookup_t aclL3ActionLookup =
     { ACTION_DO_NOT_NAT_ACTION,                SAI_ACL_ENTRY_ATTR_ACTION_NO_NAT },
 };
 
+static acl_rule_attr_lookup_t aclL3InnerActionLookup =
```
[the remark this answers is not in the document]
kavyavanibedida: have set the name aclInnerActionLookup as per your suggestion

orchagent/aclorch.h
```diff
@@ -400,6 +402,16 @@ class AclRulePacket: public AclRule
     sai_object_id_t getRedirectObjectId(const string& redirect_param);
 };
 
+class AclRuleInnerSrcMacRewrite: public AclRule
+ {
+ public:
+     AclRuleInnerSrcMacRewrite(AclOrch *m_pAclOrch, string rule, string table, bool createCounter = true);
```
anish-n, line 408: createCounter = true can cause all counter resources on the ASIC to be fully consumed at high scale, can you please set this to false.

tests/test_acl_inner_src_,mac_rewrite.py
```diff
@@ -0,0 +1,191 @@
...
+        dvs_acl.verify_acl_rule_status(TABLE_NAME, RULE_NAME, None)
+
+        # Remove second rule and check status in ASIC_DB
+        self.remove_acl_rule(dvs, TABLE_NAME, RULE_NAME+"2")
+        dvs_acl.verify_acl_rule_status(TABLE_NAME, RULE_NAME+"2", None)
+
+        dvs_acl.verify_no_acl_rules()
+
+
+    def test_InnerSrcMacRewriteAclRuleUpdate(self, dvs, dvs_acl, innersrcmacrewrite_acl_table):
```
anish-n, line 129: can you add a test case where a counter gets added to the existing acl rule.

tests/test_acl_inner_src_,mac_rewrite.py
```diff
@@ -0,0 +1,191 @@
...
+            dvs_acl.remove_acl_table_type(TABLE_TYPE)
+            dvs_acl.verify_acl_table_count(0)
+
+    def test_InnerSrcMacRewriteAclRuleCreationDeletion(self, dvs, dvs_acl, innersrcmacrewrite_acl_table):
+
+        # This test checks for ACL rule creation(more than one) deletion for the table type
+
+        self.setup_db(dvs)
+        # Add the rule
+        config_qualifiers = {"INNER_SRC_IP": "10.10.10.10/32", "TUNNEL_VNI": "5000"}
```
anish-n, line 97: can you add a test case where inner src ip is a prefix thats non /32, and assert that asic object is setup correctly. Our use case has multiple different ip ranges to support.

anish-n (changes requested):

orchagent/aclorch.h
kavyavanibedida: Have set it to false

tests/test_acl_inner_src_,mac_rewrite.py
kavyavanibedida: Previously, it was added in Mock tests, now added a new testcase for counter update in DVS tests

tests/test_acl_inner_src_,mac_rewrite.py
kavyavanibedida: Previously, it was added in Mock tests, now added in DVS test as well

anish-n:
Please link the following HLD PR in description https://github.com/sonic-net/SONiC/pull/1961

anish-n:
@prsunny, @bingwang-ms I see that current aclorch doesn't allow dynamic addition/removal of a counter on an ACL rule(its more of a create time attribute directly put into the code and not configurable via a config db field value), I wonder if it might help to introduce a new field value called enable_counter to ACL_RULE, which aclorch consumes and we may enable/disable it on demand for issue investigations using config_db. Reason why its needed: for this use-case the ACL_RULE scale+ number of matched packets will be very high, hence we cannot have a per rule counter enabled by default, hence if counters are ever needed it would need to be an on-demand thing and turned on for some particular ACL_RULE only. If we choose to not have counters then packet capture would be the alternative way to investigate problems.

orchagent/aclorch.cpp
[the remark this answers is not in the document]
kavyavanibedida: Discussed on offline chat- OnUpdate method wont be required anymore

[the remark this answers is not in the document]
kavyavanibedida: Discussed offline with Anish, won't be adding new update method

prsunny:
> @prsunny, @bingwang-ms I see that current aclorch doesn't allow dynamic addition/removal of a counter on an ACL rule(its more of a create time attribute directly put into the code and not configurable via a config db field value), I wonder if it might help to introduce a new field value called enable_counter to ACL_RULE, which aclorch consumes and we may enable/disable it on demand for issue investigations using config_db. Reason why its needed: for this use-case the ACL_RULE scale+ number of matched packets will be very high, hence we cannot have a per rule counter enabled by default, hence if counters are ever needed it would need to be an on-demand thing and turned on for some particular ACL_RULE only. If we choose to not have counters then packet capture would be the alternative way to investigate problems.

@anish-n , agree to your point that we should have a flexibility but this would need some good amount of changes and need to ensure default behavior is not changed. For this usecase, could you also share in description, the scale of ACL rules we are looking at? For the most part, it is possible that ASICs have that many counters.

anish-n:
> > @prsunny, @bingwang-ms I see that current aclorch doesn't allow dynamic addition/removal of a counter on an ACL rule(its more of a create time attribute directly put into the code and not configurable via a config db field value), I wonder if it might help to introduce a new field value called enable_counter to ACL_RULE, which aclorch consumes and we may enable/disable it on demand for issue investigations using config_db. Reason why its needed: for this use-case the ACL_RULE scale+ number of matched packets will be very high, hence we cannot have a per rule counter enabled by default, hence if counters are ever needed it would need to be an on-demand thing and turned on for some particular ACL_RULE only. If we choose to not have counters then packet capture would be the alternative way to investigate problems.
> 
> @anish-n , agree to your point that we should have a flexibility but this would need some good amount of changes and need to ensure default behavior is not changed. For this usecase, could you also share in description, the scale of ACL rules we are looking at? For the most part, it is possible that ASICs have that many counters.

Makes sense, and agree that it will be some amount of change and requires code+validation for backwards compatibility. The scale is 10k ACL rules(included in the HLD as well), will check on whether the ASIC side can support it easily. @kavyavanibedida could you please help add 10k as rule scale to the PR description as suggested?

orchagent/aclorch.h
```diff
@@ -400,6 +402,16 @@ class AclRulePacket: public AclRule
     sai_object_id_t getRedirectObjectId(const string& redirect_param);
 };
 
+class AclRuleInnerSrcMacRewrite: public AclRule
+ {
+ public:
+     AclRuleInnerSrcMacRewrite(AclOrch *m_pAclOrch, string rule, string table, bool createCounter = false);
```
anish-n, line 408: After conversation with ASIC vendors, it seems that its safe to change this createCounter to true.

tests/test_acl_inner_src_mac_rewrite.py
```diff
@@ -0,0 +1,255 @@
...
+        self.remove_acl_rule(dvs, TABLE_NAME, RULE_NAME)
+        dvs_acl.verify_acl_rule_status(TABLE_NAME, RULE_NAME, None)
+
+        # Remove second rule and check status in ASIC_DB
+        self.remove_acl_rule(dvs, TABLE_NAME, RULE_NAME+"2")
+        dvs_acl.verify_acl_rule_status(TABLE_NAME, RULE_NAME+"2", None)
+
+        dvs_acl.verify_no_acl_rules()
+
+    def test_InnerSrcMacRewriteAclRuleCounterUpdate(self, dvs, dvs_acl, innersrcmacrewrite_acl_table):
```
anish-n, line 137: In the SWSS code there isn't any support for acl counter update, so this UT is not testing anything. If you add asserts for whether a counter actually exists I believe one won't exist.

anish-n (changes requested):

tests/test_acl_inner_src_mac_rewrite.py
```diff
@@ -0,0 +1,225 @@
...
+        fvs={
+            "PRIORITY": priority,
+            "INNER_SRC_MAC_REWRITE_ACTION": action
+        }
+        for k, v in qualifiers.items():
+            fvs[k] = v
+
+        formatted_entry = swsscommon.FieldValuePairs(list(fvs.items()))
+        tbl.set(table_name + "|" + rule_name, formatted_entry)
+        time.sleep(1)
```
anish-n, line 65: Can we remove this and all other sleeps, all your validation should check for ASIC DB entry presence and poll for some period of time.

anish-n (changes requested):

tests/test_acl_inner_src_mac_rewrite.py
```diff
@@ -0,0 +1,225 @@
...
+                assert fvs.get("SAI_ACL_ENTRY_ATTR_ACTION_SET_INNER_SRC_MAC") == "66:BB:AA:C3:3E:AB"
+
+            # Verify the previous counter is removed
+            if counter_id:
+                dvs_acl.check_acl_counter_not_in_counters_map(counter_id)
+            
+            acl_rule_id_new = dvs_acl.get_acl_rule_id()
+
+            # New counter should be created
+            dvs_acl._check_acl_entry_counters_map(acl_rule_id_new)
```
anish-n, line 181: Upon looking at the dvs_acl code https://github.com/sonic-net/sonic-swss/blob/master/tests/dvslib/dvs_acl.py, it looks like the function will return(not assert) if no counter exists(hence it won't test if the counter is successfully created). 

```
    def _check_acl_entry_counters_map(self, acl_entry_oid: str):
        entry = self.asic_db.wait_for_entry("ASIC_STATE:SAI_OBJECT_TYPE_ACL_ENTRY", acl_entry_oid)
        counter_oid = entry.get("SAI_ACL_ENTRY_ATTR_ACTION_COUNTER")
        if counter_oid is None:
            return
        rule_to_counter_map = self.counters_db.get_entry("ACL_COUNTER_RULE_MAP", "")
        counter_to_rule_map = {v: k for k, v in rule_to_counter_map.items()}
        assert counter_oid in counter_to_rule_map     
```

I think that if you call get_acl_counter_ids, that contains an assert within it for the counter existing. This will then provide the required coverage.

anish-n (changes requested):

orchagent/aclorch.cpp
```diff
@@ -2133,6 +2148,66 @@ void AclRulePacket::onUpdate(SubjectType, void *)
...
+        {
+            try {
+                MacAddress(_attr_value).getMac();
+            }
+            catch (invalid_argument &e)
+            {
+                SWSS_LOG_ERROR("Mac address in the wrong format");
+                return false;
+            }
+            memcpy(actionData.parameter.mac, MacAddress(_attr_value).getMac(), sizeof(sai_mac_t));
```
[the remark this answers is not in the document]
kavyavanibedida: updated

orchagent/aclorch.h
kavyavanibedida: updated to true

tests/test_acl_inner_src_mac_rewrite.py
```diff
@@ -0,0 +1,225 @@
...
+            dvs_acl.remove_acl_table(TABLE_NAME)
+            dvs_acl.remove_acl_table_type(TABLE_TYPE)
+            dvs_acl.verify_acl_table_count(0)
+
+    def create_entry(self, tbl, key, pairs):
+        fvs = swsscommon.FieldValuePairs(pairs)
+        tbl.set(key, fvs)
+        time.sleep(1)
+
+    def get_entry(self, tbl, key, pairs):
```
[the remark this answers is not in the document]
kavyavanibedida: removed

tests/test_acl_inner_src_mac_rewrite.py
```diff
@@ -0,0 +1,225 @@
...
+        try:
+            dvs_acl.create_acl_table_type(TABLE_TYPE, CUSTOM_TABLE_TYPE_MATCHES, CUSTOM_TABLE_TYPE_BPOINT_TYPES, CUSTOM_TABLE_TYPE_ACTIONS)
+            dvs_acl.create_acl_table(TABLE_NAME, TABLE_TYPE, BIND_PORTS, stage="egress")
+            yield dvs_acl.get_acl_table_ids(1)[0]
+        finally:
+            dvs_acl.remove_acl_table(TABLE_NAME)
+            dvs_acl.remove_acl_table_type(TABLE_TYPE)
+            dvs_acl.verify_acl_table_count(0)
+
+    def create_entry(self, tbl, key, pairs):
```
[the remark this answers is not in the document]
kavyavanibedida: removed

tests/test_acl_inner_src_mac_rewrite.py
kavyavanibedida: removed sleep and added one function for verifying SAI entries. Added the polling of 5 retries in the testcase

tests/test_acl_inner_src_mac_rewrite.py
```diff
@@ -0,0 +1,225 @@
...
+        # This test checks for ACL table and table type creation deletion for inner src mac rewrite
+        try:
+            dvs_acl.create_acl_table_type(TABLE_TYPE, CUSTOM_TABLE_TYPE_MATCHES, CUSTOM_TABLE_TYPE_BPOINT_TYPES, CUSTOM_TABLE_TYPE_ACTIONS)
+            dvs_acl.create_acl_table(TABLE_NAME, TABLE_TYPE, BIND_PORTS, stage="egress")
+            acl_table_id = dvs_acl.get_acl_table_ids(1)[0]
+            assert acl_table_id is not None
+            acl_table_group_ids = dvs_acl.get_acl_table_group_ids(len(BIND_PORTS))
+
+            dvs_acl.verify_acl_table_group_members(acl_table_id, acl_table_group_ids, 1)
+            dvs_acl.verify_acl_table_port_binding(acl_table_id, BIND_PORTS, 1, stage="egress")
```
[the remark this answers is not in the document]
kavyavanibedida: added

tests/test_acl_inner_src_mac_rewrite.py
kavyavanibedida: added the verification of counter after update of the rule. It checks the counter id existence in get_acl_counter_ids as per the change request

tests/test_acl_inner_src_mac_rewrite.py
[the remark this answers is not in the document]
kavyavanibedida: Also. checks for the expected action list

tests/test_acl_inner_src_mac_rewrite.py
```diff
@@ -0,0 +1,245 @@
...
+            self.update_acl_rule(dvs, TABLE_NAME, RULE_NAME, {"INNER_SRC_IP": "15.15.15.11/20"})
+
+            # Expected SAI entries for the rule update 1 
+            new_expected_sai_qualifiers={"SAI_ACL_ENTRY_ATTR_FIELD_INNER_SRC_IP": "15.15.15.11&mask:255.255.240.0",
+                                         "SAI_ACL_ENTRY_ATTR_FIELD_TUNNEL_VNI": "4000&mask:0xffffffff",
+                                         "SAI_ACL_ENTRY_ATTR_PRIORITY": "1001",
+                                         "SAI_ACL_ENTRY_ATTR_ACTION_SET_INNER_SRC_MAC": "66:BB:AA:C3:3E:AB"} 
+    
+            retry = 5
+            while dvs_acl.get_acl_rule_id() == acl_rule_id and retry >= 0:
```
anish-n, line 170: this is just waiting for 5s it seems. 

Since the ASIC object already exists you need to wait for specific entries on the asic db entry, for that you can follow this example to create an access function and wait for specific fields within the asic db to show up: https://github.com/sonic-net/sonic-swss/blob/master/tests/test_fgnhg.py#L50

kavyavanibedida: I added the access function for validation similar to the file, although I had to keep the sleep to one sec cause its taking some time for the entry to be updated

anish-n: The point of the access function is to poll until the value in the member changes to the specified value. We should not have sleep here, if you need to poll for longer then the access function should poll for longer

kavyavanibedida: increased the polling time

SoumyaMishra18 (approved):

tests/test_acl_inner_src_mac_rewrite.py
```diff
@@ -0,0 +1,224 @@
...
+
+            for qualifer in expected_qualifier:
+                if qualifer not in fvs:
+                    return false_ret
+
+                if fvs[qualifer] != expected_qualifier[qualifer]:
+                    return false_ret
+
+            return (True, key)
+        val, result = wait_for_result(_access_function, polling_config = PollingConfig(polling_interval=1, timeout=1.0, strict=True),
```
anish-n, line 86: why is the polling interval + timeout set to 1, this seems like it will make the test flaky, why do we specify a Polling config here instead of using the default, which is polling every 0.01 seconds with a timeout of 20 seconds, thats the default and will prevent the test from being flaky.

anish-n (changes requested):

tests/test_acl_inner_src_mac_rewrite.py
kavyavanibedida: using the default polling config caused the polling to fail. I have tried the following combinations, polling interval-0.01,1,0.1 sec, timeout: 20 secs, 50 secs etc. The current config worked. It takes 1 sec for the rule id to be updated so accordingly used the same.

kavyavanibedida: image

kavyavanibedida: had set to default

tests/test_acl_inner_src_mac_rewrite.py
```diff
@@ -0,0 +1,223 @@
...
+
+            # Expected SAI entries for the rule update #1 
+            new_expected_sai_qualifiers={"SAI_ACL_ENTRY_ATTR_FIELD_INNER_SRC_IP": "15.15.15.15&mask:255.255.240.0"}
+
+            # Verify the rule id and SAI entries are updated in ASIC DB
+            self.validate_asic_acl_entries(dvs_acl, dvs_acl.asic_db, rule_id, new_expected_sai_qualifiers, True)
+        
+            # Verify the rule with counter id to be present in ASIC DB
+            counter_id_2 = dvs_acl.get_acl_counter_oid()
+            rule_id_2 = dvs_acl.get_acl_rule_id()
```
anish-n, line 180: how does the logic ensure that you got rule_id_2 and not rule_id_1?

tests/test_acl_inner_src_mac_rewrite.py
```diff
@@ -0,0 +1,223 @@
...
+            # Expected SAI entries for the rule update #2 
+            new_expected_sai_qualifiers={"SAI_ACL_ENTRY_ATTR_FIELD_INNER_SRC_IP": "20.20.20.20&mask:255.255.240.0",
+                                         "SAI_ACL_ENTRY_ATTR_FIELD_TUNNEL_VNI": "111&mask:0xffffffff"} 
+
+            # Verify the rule id and SAI entries are updated in ASIC DB 
+            self.validate_asic_acl_entries(dvs_acl, dvs_acl.asic_db, rule_id_2, new_expected_sai_qualifiers, True)
+
+            # Verify the rule with counter id to be present in ASIC DB
+            counter_id_3 = dvs_acl.get_acl_counter_oid()
+            rule_id_3 = dvs_acl.get_acl_rule_id()
```
anish-n, line 195: same comment here

tests/test_acl_inner_src_mac_rewrite.py
kavyavanibedida: added assert to ensure rule_id_2 is not same as rule_id_1.

tests/test_acl_inner_src_mac_rewrite.py
kavyavanibedida: added assert to ensure rule_id_3 is not same as rule_id_2 and rule_id

kavyavanibedida: added assert also it checks in validate_asic_entries

tests/test_acl_inner_src_mac_rewrite.py
kavyavanibedida: Rule_id is updated for rule update. The updated rule_id is polled using self.validate_asic_acl_entries(dvs_acl, dvs_acl.asic_db, rule_id, new_expected_sai_qualifiers, True). since, we are looking for update we pass in the old rule_id in the function and if the old_rule id is no longer existent the polling stops. So, the updated rule_id_2 is got after the call of validate_asic_acl_entries. Furthermore, I added the assert to check for rule_id_2 and rule_id not being same

tests/test_acl_inner_src_mac_rewrite.py
kavyavanibedida: Had to revert to the old polling config since the pr checker failed for VS testcase

anish-n: Looking through the code what I see as root cause for why your test fails with default polling is the following piece of code in the UT:
```
            if update == False:
                key = rule_id     
            else:
                keys = asic_db.get_keys(ASIC_STATE_ACL)
                for k in keys:
                    if k == rule_id:
                        return false_ret
```

Please remove this code and see how it goes.

anish-n (approved):

prsunny (approved):
lgtm, @bingwang-ms please have a look to see if any concerns. Will plan to merge in sometime.

bingwang-ms (approved):
LGTM

PR #3582, opened 2025-04-01
[POC] Fix priority inversion issue because high priority notification depends on low priority events.

liuh-80:
Please this POC PR.

Fix priority inversion issue because high priority notification depends on low priority events. 

#### Why I did it
Fix issue https://github.com/sonic-net/sonic-buildimage/issues/20941
When fix above issue, because the sonic-swss-common selectable cmp algorithm changed to FIFO, in some scenarios, orchagent dead lock because following reason:
1. APP_PORT table has priority 45, The AllPortsReady flag will update according to event from this table.
2. port status notification has priority 100, which higher than APP_PORT table.
3. when port status notification event come, it will check AllPortsReady flag, if this flag is false, then the notification will not pops. which means priority inversion.
4. Because cmp algorithm changed, in some scenario, port status notification will income before AllPortsReady set to true. Then the APP_PORT table event will never been processed because port status notification block it.

#### How I did it
Pops notification to m_toSync, and process it in drain method.

##### Work item tracking
- Microsoft ADO: 30468564

#### How to verify it
Pass all test cases.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Fix priority inversion issue because high priority notification depends on low priority events.

### DIFF
--- a/cfgmgr/portmgr.cpp
+++ b/cfgmgr/portmgr.cpp
@@ -45,7 +45,14 @@ bool PortMgr::setPortMtu(const string &alias, const string &mtu)
     }
     else
     {
-        throw runtime_error(cmd_str + " : " + res);
+        // This failure can happen on PortChannels during system startup.  A PortChannel enslaves
+        // members before a default MTU is set on the port (set in this file, not via the config!).
+        // Therefore this error is always emitted on startup for portchannel members.
+        // In theory we shouldn't log in this case, the correct fix is to detect the
+        // port is part of a portchannel and not even try this but that is rejected for
+        // possible performance implications.
+        SWSS_LOG_WARN("Setting mtu to alias:%s netdev failed (isPortStateOk=true) with cmd:%s, rc:%d, error:%s", alias.c_str(), cmd_str.c_str(), ret, res.c_str());
+        return false;
     }
     return true;
 }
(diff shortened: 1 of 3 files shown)


vivekrnv:
Hi @liuh-80, Regarding the scenario you mentioned

> 1. APP_PORT table has priority 45, The AllPortsReady flag will update according to event from this table.
> 2. port status notification has priority 100, which higher than APP_PORT table.
> 3. when port status notification event come, it will check AllPortsReady flag, if this flag is false, then the notification will not pops. which means priority inversion.
> 4. Because cmp algorithm changed, in some scenario, port status notification will income before AllPortsReady set to true. Then the APP_PORT table event will never been processed because port status notification block it.

I don't think priority inversion is the reason here. 

After point 3, When there are no new notifications in orchagent, it'll call drain() in all individual orch's and portsorch calls drain() on each Executor.  It's the unimplemented drain() API in the Notifier that's. causing the port status notifications to never be handled.

Can you elaborate a different use case on why this fix is needed?

liuh-80:
> each Executor. It's the unimplemented drain() API in the Notifier that's causing the port status notifications to never be handled.
> 
> Can you elaborate a different use case on why this fix is needed?

Basically, we are on same point, this fix does 2 changes:
1. pops notification to m_toSync, so low priority event can be select and pops in orchagent main loop.
2. implement notification drain() API, so the notification in m_toSync can be handled in it.

PR #3584, opened 2025-04-01
[vrfmgr] Fix mgmt access loss after reboot when mgmt VRF is enabled

Xichen96:
**What I did**

- Normalize the canonical `from all lookup local` rule to preference `1001` when `VrfMgr` starts.
- Detect the rule's actual preference instead of only handling the kernel's cold-boot preference `0`.
- Apply the correction independently to IPv4 and IPv6.
- Add DVS regression coverage that moves the rule to a stale preference and verifies that `vrfmgrd` restores it.

**Why I did it**

When management VRF has been enabled, a reboot can restore the `local` lookup rule at preference `32765` instead of the normal cold-boot preference `0`:

```text
1000:  from all lookup [l3mdev-table]
32765: from all lookup local
32765: from <management IP> lookup <mgmt/default>
```

The existing startup logic only searches for a rule at preference `0`. It therefore skips normalization in this warm-state case. The `local` rule can then take precedence over the management source rule at the same preference, causing management traffic to be routed incorrectly and making `eth0` unreachable by SSH or ping.

This change finds the canonical `from all lookup local` rule directly and moves it to preference `1001` whenever necessary:

```text
1000:  from all lookup [l3mdev-table]   # present when management VRF is enabled
1001:  from all lookup local
32765: from <management IP> lookup <mgmt/default>
32766: from all lookup main
32767: from all lookup default
```

The behavior now covers cold boot (`0`), the observed stale reboot state (`32765`), an already-correct rule (`1001`), and any other detected preference. IPv4 and IPv6 are normalized separately so a correct rule in one family does not prevent recovery of the other.

**How I verified it**

- Reproduced the original failure on physical DUT `bjw2-can-7050-1`: after enabling management VRF and rebooting, the IPv4 and IPv6 `local` rules moved from preference `1001` to `32765`, and management connectivity was disrupted.
- Ran the complete management-VRF test module plus standalone cold-, warm-, and fast-reboot coverage against the master PR build: 12 test bodies in total. Confirmed after every run that the canonical IPv4 and IPv6 rules remained at preference `1001`, `vrfmgrd` remained healthy, and the DUT remained reachable.
- Added DVS regression coverage for both address families.
- Completed the full Azure pipeline for this master-targeted PR successfully, including normal and ASAN VSTest lanes.
- Validated the exact `202605` backport from https://github.com/sonic-net/sonic-swss/pull/4779 on physical hardware:
  - Azure build: https://dev.azure.com/mssonic/be1b070f-be15-4154-aade-b1d3bfb17054/_build/results?buildId=1171731
  - Artifact: `sonic-swss-trixie`
  - DUT image: `SONiC.20260510.06`, Debian 13/Trixie
  - Candidate `vrfmgrd` SHA-256: `b85628d5bda90c6bd75b9b6e4eef591b217c8a3391f0cc03edf4272d5864982d`
  - Test: `mvrf/test_mgmtvrf.py::TestReboot::test_reboot`
  - Result: `1 passed in 1051.68s`
  - Reboot proof: DUT boot time changed to `2026-07-22 17:07:23 UTC` / `2026-07-23 03:07:23 Sydney`
  - Post-reboot: management SSH reachable; IPv4 and IPv6 `from all lookup local` rules both at preference `1001`; `vrfmgrd` `RUNNING`; system state `running`; management VRF disabled
  - Restored stock `vrfmgrd` SHA-256 `a031378563fa4d2e22b11c3c11c4c34200cec4787200e717e9a52778a9617b2e` and re-verified daemon, system, and rule health

**Details if related**

| Boot-time `local` rule | Previous behavior | Behavior with this change |
|---|---|---|
| `0: from all lookup local` | Move to `1001` | Move to `1001` |
| `32765: from all lookup local` | No-op; management access can be lost | Move to `1001` |
| `1001: from all lookup local` | No-op | No-op |
| Any other preference | No-op | Move to `1001` |

The `202605` physical run used `--skip-yang` only to bypass an unrelated module-level empty-patch validation failure caused by pre-existing `EVERFLOW|RULE_8` data. The management-VRF reboot test body and its connectivity/rule assertions remained enabled.

**Backport request**

- [x] 202605
- Backport PR: https://github.com/sonic-net/sonic-swss/pull/4779
- 202605 branch-specific physical validation: passed

##### Work item tracking
- Microsoft ADO (number only): 39058191

### DIFF
--- a/cfgmgr/vrfmgr.cpp
+++ b/cfgmgr/vrfmgr.cpp
@@ -95,13 +95,18 @@ VrfMgr::VrfMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, con
 
     cmd.str("");
     cmd.clear();
-    cmd << IP_CMD << " rule | grep '^0:'";
-    if (swss::exec(cmd.str(), res) == 0)
+    // Get local table pref
+    cmd << IP_CMD << " rule | grep 'from all lookup local' | awk -F'[^0-9]+' '{ print $1 }'";
+    swss::exec(cmd.str(), res);
+    table = static_cast<uint32_t>(stoul(res));
+    // If local table pref is not expected, set it to be expected
+    if (table != TABLE_LOCAL_PREF)
     {
         cmd.str("");
         cmd.clear();
-        cmd << IP_CMD << " rule add pref " << TABLE_LOCAL_PREF << " table local && " << IP_CMD << " rule del pref 0 && "
-            << IP_CMD << " -6 rule add pref " << TABLE_LOCAL_PREF << " table local && " << IP_CMD << " -6 rule del pref 0";
+        cmd << IP_CMD << " rule add pref " << TABLE_LOCAL_PREF << " table local && " << IP_CMD << " rule del pref "
+            << table << " && " << IP_CMD << " -6 rule add pref " << TABLE_LOCAL_PREF << " table local && " << IP_CMD
+            << " -6 rule del pref " << table;
         EXEC_WITH_ERROR_THROW(cmd.str(), res);
     }
 
--- a/cfgmgr/vrfmgr.cpp
+++ b/cfgmgr/vrfmgr.cpp
@@ -93,21 +93,24 @@ VrfMgr::VrfMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, con
         }
     }
 
-    cmd.str("");
-    cmd.clear();
-    // Get local table pref
-    cmd << IP_CMD << " rule | grep 'from all lookup local' | awk -F'[^0-9]+' '{ print $1 }'";
-    swss::exec(cmd.str(), res);
-    table = static_cast<uint32_t>(stoul(res));
-    // If local table pref is not expected, set it to be expected
-    if (table != TABLE_LOCAL_PREF)
+    for (const char *family : {" -4", " -6"})
     {
         cmd.str("");
         cmd.clear();
-        cmd << IP_CMD << " rule add pref " << TABLE_LOCAL_PREF << " table local && " << IP_CMD << " rule del pref "
-            << table << " && " << IP_CMD << " -6 rule add pref " << TABLE_LOCAL_PREF << " table local && " << IP_CMD
-            << " -6 rule del pref " << table;
+        res.clear();
+        cmd << IP_CMD << family << " rule show table local | grep 'from all lookup local'";
         EXEC_WITH_ERROR_THROW(cmd.str(), res);
+        table = static_cast<uint32_t>(stoul(res));
+
+        if (table != TABLE_LOCAL_PREF)
+        {
+            cmd.str("");
+            cmd.clear();
+            res.clear();
+            cmd << IP_CMD << family << " rule add pref " << TABLE_LOCAL_PREF << " table local && "
+                << IP_CMD << family << " rule del pref " << table << " table local";
+            EXEC_WITH_ERROR_THROW(cmd.str(), res);
+        }
     }
 
     if (!WarmStart::isWarmStart())
@@ -557,4 +560,3 @@ uint32_t VrfMgr::getVRFmappedVNI(const std::string& vrf_name)
         return 0;
     }
 }
-
--- a/tests/test_vrf.py
+++ b/tests/test_vrf.py
@@ -163,6 +163,55 @@ class TestVrf(object):
         r = random.choice(values)
         return r[0], r[1]
 
+    @pytest.mark.parametrize("family", ["-4", "-6"])
+    def test_VRFMgr_LocalRulePriority(self, dvs, testlog, family):
+        other_family = "-6" if family == "-4" else "-4"
+
+        def get_local_rule_pref(rule_family):
+            status, output = dvs.runcmd(["ip", rule_family, "rule", "show", "table", "local"])
+            assert status == 0, output
+            for rule in output.splitlines():
+                preference, selector = rule.split(":", 1)
+                if selector.strip() == "from all lookup local":
+                    return int(preference.strip())
+            assert False, "Missing {} canonical local rule".format(rule_family)
+
+        original_pref = get_local_rule_pref(family)
+        assert original_pref == 1001
+        assert get_local_rule_pref(other_family) == 1001
+        try:
+            status, output = dvs.runcmd(
+                ["ip", family, "rule", "add", "pref", "32765", "table", "local"]
+            )
+            assert status == 0, output
+            status, output = dvs.runcmd(
+                ["ip", family, "rule", "del", "pref", str(original_pref), "table", "local"]
+            )
+            assert status == 0, output
+
+            status, output = dvs.runcmd(["supervisorctl", "restart", "vrfmgrd"])
+            assert status == 0, output
+
+            for _ in range(10):
+                if get_local_rule_pref(family) == 1001:
+                    break
+                time.sleep(1)
+            assert get_local_rule_pref(family) == 1001
+            assert get_local_rule_pref(other_family) == 1001
+        finally:
+            current_pref = get_local_rule_pref(family)
+            if current_pref != 1001:
+                status, output = dvs.runcmd(
+                    ["ip", family, "rule", "add", "pref", "1001", "table", "local"]
+                )
+                assert status == 0, output
+                status, output = dvs.runcmd(
+                    ["ip", family, "rule", "del", "pref", str(current_pref), "table", "local"]
+                )
+                assert status == 0, output
+            status, output = dvs.runcmd(["supervisorctl", "restart", "vrfmgrd"])
+            assert status == 0, output
+
     def test_VRFMgr_Comprehensive(self, dvs, testlog):
         self.setup_db(dvs)
 
--- a/tests/test_vrf.py
+++ b/tests/test_vrf.py
@@ -163,6 +163,67 @@ class TestVrf(object):
         r = random.choice(values)
         return r[0], r[1]
 
+    @pytest.mark.parametrize("family", ["-4", "-6"])
+    def test_VRFMgr_LocalRulePriority(self, dvs, testlog, family):
+        other_family = "-6" if family == "-4" else "-4"
+
+        def get_local_rule_prefs(rule_family):
+            status, output = dvs.runcmd(["ip", rule_family, "rule", "show", "table", "local"])
+            assert status == 0, output
+            preferences = []
+            for rule in output.splitlines():
+                preference, selector = rule.split(":", 1)
+                if selector.strip() == "from all lookup local":
+                    preferences.append(int(preference.strip()))
+            return preferences
+
+        def get_local_rule_pref(rule_family):
+            preferences = get_local_rule_prefs(rule_family)
+            assert preferences, "Missing {} canonical local rule".format(rule_family)
+            return preferences[0]
+
+        original_pref = get_local_rule_pref(family)
+        assert original_pref == 1001
+        assert get_local_rule_pref(other_family) == 1001
+        try:
+            status, output = dvs.runcmd(
+                ["ip", family, "rule", "add", "pref", "32765", "table", "local"]
+            )
+            assert status == 0, output
+            status, output = dvs.runcmd(
+                ["ip", family, "rule", "del", "pref", str(original_pref), "table", "local"]
+            )
+            assert status == 0, output
+
+            status, output = dvs.runcmd(["supervisorctl", "restart", "vrfmgrd"])
+            assert status == 0, output
+
+            for _ in range(10):
+                if get_local_rule_pref(family) == 1001:
+                    break
+                time.sleep(1)
+            assert get_local_rule_pref(family) == 1001
+            assert get_local_rule_pref(other_family) == 1001
+        finally:
+            current_prefs = get_local_rule_prefs(family)
+            if 1001 not in current_prefs:
+                status, output = dvs.runcmd(
+                    ["ip", family, "rule", "add", "pref", "1001", "table", "local"]
+                )
+                assert status == 0, output
+            for current_pref in current_prefs:
+                if current_pref == 1001:
+                    continue
+                status, output = dvs.runcmd(
+                    [
+                        "ip", family, "rule", "del", "pref", str(current_pref),
+                        "from", "all", "table", "local",
+                    ]
+                )
+                assert status == 0, output
+            status, output = dvs.runcmd(["supervisorctl", "restart", "vrfmgrd"])
+            assert status == 0, output
+
     def test_VRFMgr_Comprehensive(self, dvs, testlog):
         self.setup_db(dvs)
 


prsunny:
Can you provide an example of the expanded command in description with and without vrf?

Xichen96:
Vrfmgr will adjust ip rules to suit its current need whenever there is reboot
or config change. To summarize the current logic for dealing with ip rules.
If there is pref 0, remove pref 0 and add local table to 1001.

########################################################################

Ip rule changes for expected case. We expect the ip rules to be 
the same as the first one when mvrf is off, and to be the same 
as the second one when mvrf is on.

target ip rule when mvrf off
1001:   from all lookup local
32764:  from all to PTF lookup default
32764:  from all to GATEWAY lookup default
32764:  from all to DNS lookup default
32764:  from all to DNS lookup default
32764:  from all to NTP lookup default
32765:  from MGMT_IP lookup default
32766:  from all lookup main
32767:  from all lookup default

target ip rule when mvrf on
1000:   from all lookup [l3mdev-table]
1001:   from all lookup local
32764:  from all to PTF lookup mgmt
32764:  from all to GATEWAY lookup mgmt
32764:  from all to DNS lookup mgmt
32764:  from all to DNS lookup mgmt
32764:  from all to NTP lookup mgmt
32765:  from MGMT_IP lookup mgmt
32766:  from all lookup main
32767:  from all lookup default

#######################################################################

Default case, when mvrf was not relevant at all.
Lookup local was removed from pref 0 and added to 1001, making room for the
future l3mdev-table at pref 1000. Works fine. Finished ip rule is the same
as target ip rule with mvrf off.

config off mvrf off startup ip rule
0:	from all lookup local
32764:	from all to PTF lookup default
32764:	from all to GATEWAY lookup default
32764:	from all to DNS lookup default
32764:	from all to DNS lookup default
32764:	from all to NTP lookup default
32765:	from MGMT_IP lookup default
32766:	from all lookup main
232767:	from all lookup default

config off mvrf off finished ip rule
1001:   from all lookup local
32764:  from all to PTF lookup default
32764:  from all to GATEWAY lookup default
32764:  from all to DNS lookup default
32764:  from all to DNS lookup default
32764:  from all to NTP lookup default
32765:  from MGMT_IP lookup default
32766:  from all lookup main
32767:  from all lookup default

#####################################################

No mvrf configured in config db, but mvrf was on before reboot.
Finishing ip rule should match finishing ip rule from above.
To compare the 2, difference is local table at pref 1001 vs pref 32765,
and l3mdev-table at 1000 (l3mdev-table is empty, doesnt matter for the time being).
So changes to be made, even if no pref 0 exist, old local table look up at 32765
should still be removed and placed at 1001.

config off mvrf on startup ip rule
1000:	from all lookup [l3mdev-table]
32764:	from all to PTF lookup default
32764:	from all to GATEWAY lookup default
32764:	from all to DNS lookup default
32764:	from all to DNS lookup default
32764:	from all to NTP lookup default
32765:	from all lookup local
32765:	from MGMT_IP lookup default
32766:	from all lookup main
32767:	from all lookup default

config off mvrf on finished ip rule
1000:   from all lookup [l3mdev-table]
32764:  from all to PTF lookup default
32764:  from all to GATEWAY lookup default
32764:  from all to DNS lookup default
32764:  from all to DNS lookup default
32764:  from all to NTP lookup default
32765:  from all lookup local
32765:  from MGMT_IP lookup default
32766:  from all lookup main
32767:  from all lookup default


##########################################################

mvrf is configured, and mvrf is on before reboot.
We expect ip rules to match target ip rules when mvrf is off.
The only difference is the lookup local which should be at pref 1001
is at 32765. The changes proposed above would fix this one as well.

config on vrf on startup ip rule
1000:	from all lookup [l3mdev-table]
32764:	from all to PTF lookup 5000
32764:	from all to GATEWAY lookup 5000
32764:	from all to DNS lookup 5000
32764:	from all to DNS lookup 5000
32764:	from all to NTP lookup 5000
32765:	from all lookup local
32765:	from MGMT_IP lookup 5000
32766:	from all lookup main
32767:	from all lookup default

config on vrf on finished ip rule
1000:   from all lookup [l3mdev-table]
32764:  from all to PTF lookup mgmt
32764:  from all to GATEWAY lookup mgmt
32764:  from all to DNS lookup mgmt
32764:  from all to DNS lookup mgmt
32764:  from all to NTP lookup mgmt
32765:  from all lookup local
32765:  from MGMT_IP lookup mgmt
32766:  from all lookup main
32767:  from all lookup default

####################################################

mvrf is configured and mvrf was off before reboot. We also expect ip rules
to match target ip rules for mvrf on. The only difference is the lookup 
local which should be at pref 1001 is at 32765. The changes proposed above 
would fix this one as well.

config on vrf off startup ip rule
1000:	from all lookup [l3mdev-table]
32764:	from all to PTF lookup 5000
32764:	from all to GATEWAY lookup 5000
32764:	from all to DNS lookup 5000
32764:	from all to DNS lookup 5000
32764:	from all to NTP lookup 5000
32765:	from all lookup local
32765:	from MGMT_IP lookup 5000
32766:	from all lookup main
32767:	from all lookup default

config on vrf off finished ip rule
1000:   from all lookup [l3mdev-table]
32764:  from all to PTF lookup mgmt
32764:  from all to GATEWAY lookup mgmt
32764:  from all to DNS lookup mgmt
32764:  from all to DNS lookup mgmt
32764:  from all to NTP lookup mgmt
32765:  from all lookup local
32765:  from MGMT_IP lookup mgmt
32766:  from all lookup main
32767:  from all lookup default

Xichen96:
@prsunny hi Prince, could you help review again, I added analysis on changes of ip rules before and after vrfmgr and how it should be fixed.

Xichen96:
@yxieca @prsunny @venkatmahalingam Description is updated with full root-cause analysis and the four boot-case table. Change is +9/-4 in `cfgmgr/vrfmgr.cpp`: generalises the existing `local`-table-pref normalisation so it also fires on warm boots where mgmt VRF was previously enabled (current code only handles cold-boot pref 0). Looking for a review.

cc @bsun-sudo @nkelapur @ypcisco — past work on this file's reboot / mgmt-VRF paths.

cfgmgr/vrfmgr.cpp
```diff
@@ -93,16 +93,24 @@ VrfMgr::VrfMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, con
...
+
+        if (table != TABLE_LOCAL_PREF)
+        {
+            cmd.str("");
+            cmd.clear();
+            res.clear();
+            cmd << IP_CMD << family << " rule add pref " << TABLE_LOCAL_PREF << " table local && "
+                << IP_CMD << family << " rule del pref " << table << " table local";
+            EXEC_WITH_ERROR_THROW(cmd.str(), res);
+        }
```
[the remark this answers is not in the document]
Xichen96: Keeping this variable name intentionally. This block immediately consumes the parsed preference, and we reviewed the rename tradeoff while keeping the implementation compact.

tests/test_vrf.py
```diff
@@ -163,6 +163,58 @@ def packet_action_gen(self):
...
+            for _ in range(10):
+                preferences = [get_local_rule_prefs(rule_family) for rule_family in families]
+                if preferences == [[1001], [1001]]:
+                    break
+                time.sleep(1)
+            assert preferences == [[1001], [1001]]
+        finally:
+            for rule_family in families:
+                set_local_rule_pref(rule_family, 1001)
+            dvs.runcmd(["supervisorctl", "start", "vrfmgrd"])
```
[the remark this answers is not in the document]
Xichen96: Accepted and addressed in the current head: cleanup now uses `supervisorctl restart vrfmgrd` and asserts the returned status and output.

cfgmgr/vrfmgr.cpp
```diff
@@ -93,16 +93,24 @@ VrfMgr::VrfMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, con
...
```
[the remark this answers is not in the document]
Xichen96: The underlying failure-reporting concern is addressed in the current head without adding a C++ parser: the command now ends with `grep 'from all lookup local'`, so `EXEC_WITH_ERROR_THROW` surfaces a missing canonical rule before `stoul` is called.

tests/test_vrf.py
```diff
@@ -163,6 +163,45 @@ def packet_action_gen(self):
...
+                if get_local_rule_pref(family) == 1001:
+                    break
+                time.sleep(1)
+            assert get_local_rule_pref(family) == 1001
+            assert get_local_rule_pref(other_family) == 1001
+        finally:
+            current_pref = get_local_rule_pref(family)
+            if current_pref != 1001:
+                dvs.runcmd(["ip", family, "rule", "add", "pref", "1001", "table", "local"])
+                dvs.runcmd(["ip", family, "rule", "del", "pref", str(current_pref), "table", "local"])
```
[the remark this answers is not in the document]
Xichen96: Accepted and addressed in the current head: each cleanup `ip rule add`/`del` command now asserts its status and output.

tests/test_vrf.py
```diff
@@ -163,6 +163,51 @@ def packet_action_gen(self):
...
         return r[0], r[1]
 
+    @pytest.mark.parametrize("family", ["-4", "-6"])
+    def test_VRFMgrLocalRulePriority(self, dvs, testlog, family):
+        other_family = "-6" if family == "-4" else "-4"
+
+        def get_local_rule_pref(rule_family):
+            status, output = dvs.runcmd(["ip", rule_family, "rule", "show", "table", "local"])
+            assert status == 0, output
+            return int(output.split(":", 1)[0].strip())
```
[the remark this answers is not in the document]
Xichen96: Accepted and addressed in the current head: `get_local_rule_pref()` now scans all returned rules and selects the exact `from all lookup local` entry.

tests/test_vrf.py
```diff
@@ -163,6 +163,51 @@ def packet_action_gen(self):
         r = random.choice(values)
         return r[0], r[1]
 
+    @pytest.mark.parametrize("family", ["-4", "-6"])
+    def test_VRFMgrLocalRulePriority(self, dvs, testlog, family):
+        other_family = "-6" if family == "-4" else "-4"
```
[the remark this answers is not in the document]
Xichen96: Accepted and addressed in the current head: the test is now named `test_VRFMgr_LocalRulePriority` to match the existing convention.

StormLiangMS:
@Xichen96 — thanks for the detailed fix and the 202605 physical validation. The per-family normalization is directionally correct, and the current CI is green.

One concern before approval: `cfgmgr/vrfmgr.cpp` only parses the first `from all lookup local` rule. If adding preference `1001` succeeds but deleting the stale rule fails or is interrupted, the next restart will see `1001` first and can leave the duplicate stale rule indefinitely. The new test helper also returns only the first match, so it cannot detect this state, and its cleanup can leave the injected `32765` rule behind.

Could you please consider collecting all matching preferences, ensuring exactly one `1001` rule remains, and asserting that complete state in the test?

@prsunny, could you please help review the IP-rule normalization and reboot behavior?

Xichen96:
Thanks, Storm. I agree that ensuring exactly one `1001` rule remains would make the normalization more robust. However, I believe the duplicate-rule state is a pre-existing partial-operation failure mode rather than a regression introduced by this PR.

The existing preference-`0` implementation also uses a non-transactional add-then-delete sequence and does not enumerate or reconcile duplicate local-table rules. If that sequence is interrupted after the add succeeds, duplicate state can already be left behind.

This PR intentionally preserves the existing normalization model and stays focused on the reproduced reboot issue, where the local-table rule moves to preference `32765`. That specific fix has current CI coverage and has also been physically validated on the `202605` image.

I would prefer to address full reconciliation separately: collect all matching preferences, ensure exactly one `1001` rule remains, clean up duplicates, and add corresponding idempotence and fault-injection coverage. This keeps the current PR scoped to the demonstrated bug while leaving the broader hardening for a focused follow-up.

StormLiangMS:
@Xichen96 — re-review complete on current head `afeec730`. The IPv4/IPv6 rule normalization correctly fixes the reproduced management-VRF reboot failure, and current CI/DCO/CLA plus physical 202605 validation are green. I found no merge-blocking issue.

One follow-up remains: startup reconciles only the first canonical rule, so duplicate-rule repair is still tracked separately in #4826. Given that the failure removes device management access, I would treat this as **Medium priority / Medium urgency**, rather than Low/Low. Comfortable with merge after the required reviewer approval.

Xichen96:
@StormLiangMS Addressed the PR-local cleanup concern in head `afeec730a9b416c0ea0a1e0b78f9a63b57d35336`. The test cleanup now enumerates every exact `from all lookup local` rule, restores `1001` if absent, and removes each non-`1001` entry without touching source-specific local-table rules.

The broader production reconciliation is tracked separately in https://github.com/sonic-net/sonic-swss/issues/4826. Exact-head CI is running.

prabhataravind:
@Xichen96 @StormLiangMS  is this a day-1 issue or a regression? From the description, it seems like a day-1 issue but curious why we didn't see this until now.

cfgmgr/vrfmgr.cpp
```diff
@@ -93,16 +93,24 @@ VrfMgr::VrfMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, con
...
-    if (swss::exec(cmd.str(), res) == 0)
+    for (const char *family : {" -4", " -6"})
     {
         cmd.str("");
         cmd.clear();
-        cmd << IP_CMD << " rule add pref " << TABLE_LOCAL_PREF << " table local && " << IP_CMD << " rule del pref 0 && "
-            << IP_CMD << " -6 rule add pref " << TABLE_LOCAL_PREF << " table local && " << IP_CMD << " -6 rule del pref 0";
+        res.clear();
+        cmd << IP_CMD << family << " rule show table local | grep 'from all lookup local'";
         EXEC_WITH_ERROR_THROW(cmd.str(), res);
```
prabhataravind, line 102: This line and stoul in the next line can throw from the VrfMgr constructor. If the canonical from all lookup local rule is temporarily absent or the ip output is unexpected during startup, vrfmgrd exits instead of recovering or reporting a controlled error, potentially leaving management access unavailable. This might need more explicit output parsing and exception handling.

cfgmgr/vrfmgr.cpp
```diff
@@ -93,16 +93,24 @@ VrfMgr::VrfMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, con
...
-    cmd << IP_CMD << " rule | grep '^0:'";
-    if (swss::exec(cmd.str(), res) == 0)
+    for (const char *family : {" -4", " -6"})
     {
         cmd.str("");
         cmd.clear();
-        cmd << IP_CMD << " rule add pref " << TABLE_LOCAL_PREF << " table local && " << IP_CMD << " rule del pref 0 && "
-            << IP_CMD << " -6 rule add pref " << TABLE_LOCAL_PREF << " table local && " << IP_CMD << " -6 rule del pref 0";
+        res.clear();
+        cmd << IP_CMD << family << " rule show table local | grep 'from all lookup local'";
```
prabhataravind, line 101: grep can return multiple from all lookup local rules, but stoul(res) parses only the first preference and ignores the remaining output. If the first rule is already 1001, the code skips normalization entirely and leaves any additional stale local rules in place. This means startup does not actually enforce the intended single-rule invariant, and a stale rule with a higher precedence could still affect management traffic.

Xichen96:
This is a latent day-one issue, not a recent regression. The 2019 logic only normalizes the usual cold-boot preference `0`; it misses the `32765` state restored after a management-VRF reboot. That specific state also lacked regression coverage.

cfgmgr/vrfmgr.cpp
Xichen96: If the canonical `from all lookup local` rule cannot be identified, `vrfmgrd` reports the error and exits rather than guessing how to modify an unknown rule state. This PR repairs the known stale-preference case; it cannot safely infer the correct recovery when the fundamental rule invariant is missing.

cfgmgr/vrfmgr.cpp
Xichen96: The broader duplicate-rule reconciliation is intentionally tracked separately in https://github.com/sonic-net/sonic-swss/issues/4826. This PR remains scoped to the reproduced reboot state where the single canonical local rule is restored at stale preference `32765`.

Xichen96:
@StormLiangMS @prsunny This PR is ready for review on the current head. Xichen review, exact-head CI, and required validation are complete; please review when available.

cfgmgr/vrfmgr.cpp
prabhataravind: I agree that vrfmgrd should not guess how to reconstruct a missing canonical rule. However, not modifying the rule and terminating the entire daemon are different choices. This changes the previous missing-rule behavior into a fatal constructor failure and also leaves malformed stoul() input unhandled. Could we explicitly handle command and parsing failures, including std::invalid_argument and std::out_of_range, log the complete unexpected rule state, skip normalization for only the affected address family, and allow vrfmgrd to continue its other responsibilities?

cfgmgr/vrfmgr.cpp
```diff
@@ -93,16 +93,24 @@ VrfMgr::VrfMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, con
...
         EXEC_WITH_ERROR_THROW(cmd.str(), res);
+        table = static_cast<uint32_t>(stoul(res));
+
+        if (table != TABLE_LOCAL_PREF)
+        {
+            cmd.str("");
+            cmd.clear();
+            res.clear();
+            cmd << IP_CMD << family << " rule add pref " << TABLE_LOCAL_PREF << " table local && "
+                << IP_CMD << family << " rule del pref " << table << " table local";
```
prabhataravind, line 111: Line 101 finds the specific from all lookup local rule, but this line deletes a rule using only its preference and table. If another rule has the same preference and also looks up the local table, ip rule del could delete that rule instead of the one found on line 101. Could we include from all in the delete command, for example rule del pref <pref> from all table local, to ensure it deletes the intended canonical rule?

tests/test_vrf.py
```diff
@@ -163,6 +163,67 @@ def packet_action_gen(self):
...
+            assert status == 0, output
+
+            status, output = dvs.runcmd(["supervisorctl", "restart", "vrfmgrd"])
+            assert status == 0, output
+
+            for _ in range(10):
+                if get_local_rule_pref(family) == 1001:
+                    break
+                time.sleep(1)
+            assert get_local_rule_pref(family) == 1001
```
prabhataravind, line 205: This assertion checks only the first matching from all lookup local rule. The test could therefore pass if the rule at preference 1001 was added but the stale rule at 32765 was not removed. Could we assert that get_local_rule_prefs(family) == [1001] here so the test verifies that exactly one canonical rule remains?

PR #3585, opened 2025-04-01
Move timestamps out of counter table to avoid update too frequently

stephenxs:
**What I did**

The timestamp of PFC watchdog polling is moved out of the COUNTER table. The PFC watchdog Lua plugin should be adjusted accordingly.

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/pfc_detect_mellanox.lua
+++ b/orchagent/pfc_detect_mellanox.lua
@@ -25,9 +25,22 @@ if timestamp_last ~= false then
     redis.call('HSET', 'TIMESTAMP', 'effective_pfcwd_poll_time_last', global_effective_poll_time)
 end
 
+-- Get timestamp from TIME_STAMP table for PFC_WD counters
+-- Use a field name without spaces to avoid issues
+local port_timestamp_current = tonumber(redis.call('HGET', 'COUNTERS:TIME_STAMP', 'PFC_WD_Port_Counter_time_stamp'))
+local port_timestamp_last = tonumber(redis.call('HGET', 'COUNTERS:TIME_STAMP', 'PFC_WD_Port_Counter_time_stamp_last'))
+
+-- Update the last timestamp for all ports at once
+if port_timestamp_current ~= nil then
+    redis.call('HSET', 'COUNTERS:TIME_STAMP', 'PFC_WD_Port_Counter_time_stamp_last', port_timestamp_current)
+end
+
 local effective_poll_time
-local effective_poll_time_lasttime
-local port_timestamp_last_cache = {}
+if port_timestamp_current ~= nil and port_timestamp_last ~= nil then
+    effective_poll_time = (port_timestamp_current - port_timestamp_last) / 1000
+else
+    effective_poll_time = global_effective_poll_time
+end
 
 local debug_storm_global = redis.call('HGET', 'DEBUG_STORM', 'enabled') == 'true'
 local debug_storm_threshold = tonumber(redis.call('HGET', 'DEBUG_STORM', 'threshold'))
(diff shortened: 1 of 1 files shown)


orchagent/pfc_detect_mellanox.lua
```diff
@@ -132,8 +124,8 @@ for i = n, 1, -1 do
                                 local pfc_rx_packets_string = '"pfc_rx_packets","' .. tostring(pfc_rx_packets) .. '","pfc_rx_packets_last","' .. tostring(pfc_rx_packets_last) .. '",'
                                 local storm_condition_string = '"pfc_duration","' .. tostring(pfc_duration) .. '","pfc_duration_last","' .. tostring(pfc_duration_last) .. '",'
                                 local timestamps = '"timestamp","' .. timestamp_string .. '","timestamp_last","' .. timestamp_last .. '","effective_poll_time","' .. effective_poll_time .. '"'
-                                if effective_poll_time_lasttime ~= false then
-                                    timestamps = timestamps .. ',"effective_pfcwd_poll_time_last","' .. effective_poll_time_lasttime .. '"'
+                                if global_effective_poll_time_lasttime ~= false then
```
kperumalbfn, line 127: @stephenxs global_effective_pfcwd_poll_time_last is the variable that gets the timestamp, right? why are we comparing with boolean(false)? Pls check

orchagent/pfc_detect_mellanox.lua
```diff
@@ -132,8 +124,8 @@ for i = n, 1, -1 do
                                 local pfc_rx_packets_string = '"pfc_rx_packets","' .. tostring(pfc_rx_packets) .. '","pfc_rx_packets_last","' .. tostring(pfc_rx_packets_last) .. '",'
                                 local storm_condition_string = '"pfc_duration","' .. tostring(pfc_duration) .. '","pfc_duration_last","' .. tostring(pfc_duration_last) .. '",'
```
kperumalbfn, line 125: Also instead of HDEL redis operation, could we change to set to reduce the time?

orchagent/pfc_detect_mellanox.lua
stephenxs: Not really 
It's timestamp if it succeeds to fetch it from db
Or false if it fails to fetch it from db
Comparing it with false means to check whether it is in the db

orchagent/pfc_detect_mellanox.lua
stephenxs: Do you mean line 120-121?
I suggest keeping as it is
This happens very rarely. Only if we receive PFC storm covering 99% of polling interval.
In addition, this is old code.

stephenxs (commented):
@kperumalbfn could you please check my reply? Thanks

stephenxs:
> Thanks @stephenxs Could you link the sairedis PR as well?

Hi @kperumalbfn 

https://github.com/sonic-net/sonic-sairedis/pull/1567

orchagent/pfc_detect_mellanox.lua
```diff
@@ -63,27 +76,6 @@ for i = n, 1, -1 do
...
-                        effective_poll_time = global_effective_poll_time
-                    end
-                    effective_poll_time_lasttime = false
-                else
-                    effective_poll_time = global_effective_poll_time
-                    effective_poll_time_lasttime = global_effective_poll_time_lasttime
-                end
-
                 -- Get all counters
                 local occupancy_bytes = redis.call('HGET', counters_table_name .. ':' .. KEYS[i], 'SAI_QUEUE_STAT_CURR_OCCUPANCY_BYTES')
```
kperumalbfn, line 80: @stephenxs should we move this to HGETALL and get all counters for the queue and use that cache? this will also reduce queries inside the loop.

stephenxs: Thanks @kperumalbfn 
I tend to keep it as it is.
This indeed reduces a hash lookup. However, we can hardly say whether it can yield any performance improvements without testing it.
In addition we need extra logic to convert the resulting list into a dictionary in Lua. Such logic introduces overheads and makes it difficult to understand.

Previously we measured the performance of PFC watchdog on a 256 port SKU. It is very fast to execute the whole Lua plugin part for all queues, including the communication time.
 
We can optimize it if we see it takes more time in the future but for now we don't need to do that.
Thanks

kperumalbfn: Thanks. Can we update this as an improvement? For 512 port SKUs, this can have some impact.

kperumalbfn (approved):

orchagent/pfc_detect_mellanox.lua
stephenxs: Thanks.
We also verified 512-port SKUs and it's the same.
But we will do it if it becomes a bottleneck.

stephenxs:
Env issue. Retrying.
```
dpkg: dependency problems prevent configuration of libsaivs:
 libsaivs depends on libvppinfra; however:
  Package libvppinfra is not installed.
 libsaivs depends on vpp; however:
  Package vpp is not installed.

dpkg: error processing package libsaivs (--install):
 dependency problems - leaving unconfigured
Setting up libsairedis-dev (1.0.0) ...
Setting up libsaimetadata (1.0.0) ...
Setting up syncd-vs (1.0.0) ...

Configuration file '/etc/init.d/syncd', does not exist on system.
Installing new config file as you requested.
invoke-rc.d: could not determine current runlevel
invoke-rc.d: policy-rc.d denied execution of start.
dpkg: dependency problems prevent configuration of libsaivs-dev:
 libsaivs-dev depends on libsaivs (= 1.0.0); however:
  Package libsaivs is not configured yet.

dpkg: error processing package libsaivs-dev (--install):
 dependency problems - leaving unconfigured
Setting up libsaimetadata-dev (1.0.0) ...
Processing triggers for libc-bin (2.36-9+deb12u9) ...
Errors were encountered while processing:
 libsaivs
 libsaivs-dev
```

prsunny (approved):
approving to unblock merge.

r12f:
hi @kperumalbfn , do we need this for 202412? looks like a good change to add.

kperumalbfn:
> hi @kperumalbfn , do we need this for 202412? looks like a good change to add.

@r12f This change will be good for optimization.

PR #3587, opened 2025-04-01
Orchagent SAI error handling improvements

prabhataravind:
**What I did**
This change aims to reduce self-induced orchagent exit when any SAI API call fails (i.e returns anything other than SAI_STATUS_SUCCESS). This change is the first set of changes that does the following:

1.  handleSaiCreateStatus() / handleSaiSetStatus() changes
- Return 'task_success' for SAI_STATUS_ITEM_ALREADY_EXISTS, SAI_STATUS_ITEM_NOT_FOUND, SAI_STATUS_ADDR_NOT_FOUND and SAI_STATUS_OBJECT_IN_USE irrespective of the object type.
- Return 'task_need_retry' for SAI_STATUS_INSUFFICIENT_RESOURCES, SAI_STATUS_TABLE_FULL, SAI_STATUS_NO_MEMORY and 
SAI_STATUS_NV_STORAGE_FULL.
-  Call handleSaiFailure() and return 'task_failed' for other SAI errors. This will log a structured syslog via eventd and also take a SAI dump.

2. handleSaiRemoveStatus() changes
- Return 'task_success' for SAI_STATUS_ITEM_ALREADY_EXISTS, SAI_STATUS_ITEM_NOT_FOUND, SAI_STATUS_ADDR_NOT_FOUND, SAI_STATUS_INSUFFICIENT_RESOURCES,  SAI_STATUS_TABLE_FULL, SAI_STATUS_NO_MEMORY, SAI_STATUS_NV_STORAGE_FULL
- Return 'task_need_retry' for SAI_STATUS_OBJECT_IN_USE
- Call handleSaiFailure() and return 'task_failed' for other SAI errors. This will log a structured syslog via eventd and also take a SAI dump.

3. handleSaiGetStatus() changes
- Log a NOTICE message and return task_failed. This is similar to what is being done today for GET calls.

4. handleSaiFailure() changes
- Update handleSaiFailure() to take 3 arguments - namely the SAI API, operation type string and SAI API return status. This will be used in crafting a structured syslog error message when the failure happens. All callers of this function are updated accordingly.

5. Mock test changes
- Added new tests for coverage and updated existing tests that do "ASSERT_DEATH" assertions when SAI API calls fail.
- Fixed errors in existing portsorch_ut test cases

**What is not done**
There are changes needed to all orchs to handle scenarios where orchagent doesn't crash anymore when SAI API calls fail. There are also places in different orchs where an explicit exception is thrown in case of SAI errors. These and the remaining items in https://github.com/sonic-net/SONiC/pull/1698 will be handled in phase-2.
 
**Why I did it**
Crashing orchagent on every SAI error is an overkill. Instead, we follow the approach that is called out in the HLD above to handle these errors in a more graceful manner.

**How I verified it**
By adding mock tests to verify that orchagent no longer exits when there are SAI API call failures. 

**Details if related**
Sample error handling snippet showing eventd log and SAI dump invocation.
```
2025 Jun  6 17:53:24.448168 sonic ERR swss#orchagent: :- meta_sai_validate_route_entry: object key SAI_OBJECT_TYPE_ROUTE_ENTRY:{"dest":"10.1.0.32/32","switch_id":"oid:0x21000000000000","vr":"oid:0x3000000000023"} already exists
2025 Jun  6 17:53:24.448547 sonic ERR swss#orchagent: :- flush_creating_entries: EntityBulker.flush create entries failed, number of entries to create: 1, status: SAI_STATUS_ITEM_ALREADY_EXISTS
2025 Jun  6 17:53:24.448750 sonic ERR swss#orchagent: :- addRoutePost: Failed to create route 10.1.0.32/32 with next hop(s) 30.1.0.2@PortChannel101
2025 Jun  6 17:53:24.448933 sonic ERR swss#orchagent: :- handleSaiFailure: Encountered failure in create operation, SAI API: SAI_API_ROUTE, status: SAI_STATUS_NOT_EXECUTED
2025 Jun  6 17:53:24.449276 sonic NOTICE syncd#syncd: :- processNotifySyncd: Invoking SAI failure dump
2025 Jun  6 17:53:24.449276 sonic NOTICE swss#orchagent: :- publish: EVENT_PUBLISHED: {"sonic-events-swss:sai-operation-failure":{"api":"SAI_API_ROUTE","operation":"create","status":"SAI_STATUS_NOT_EXECUTED","timestamp":"2025-06-06T17:53:24.447963Z"}}
2025 Jun  6 17:53:24.449309 sonic NOTICE swss#orchagent: :- notifySyncd: sending syncd: SYNCD_INVOKE_DUMP
2025 Jun  6 17:53:24.465408 sonic NOTICE swss#orchagent: :- sai_redis_notify_syncd: invoked DUMP succeeded
```

### DIFF
--- a/orchagent/saihelper.cpp
+++ b/orchagent/saihelper.cpp
@@ -567,8 +567,8 @@ task_process_status handleSaiCreateStatus(sai_api_t api, sai_status_t status, vo
      *          in each orch.
      *       3. Take the type of sai api into consideration.
      */
-     switch (status)
-     {
+    switch (status)
+    {
         case SAI_STATUS_SUCCESS:
         case SAI_STATUS_ITEM_NOT_FOUND:
         case SAI_STATUS_ADDR_NOT_FOUND:
@@ -585,8 +585,8 @@ task_process_status handleSaiCreateStatus(sai_api_t api, sai_status_t status, vo
         default:
             handleSaiFailure(api, "create", status);
             break;
-      }
-      return task_failed;
+    }
+    return task_failed;
 }
 
 task_process_status handleSaiSetStatus(sai_api_t api, sai_status_t status, void *context)
--- a/orchagent/saihelper.cpp
+++ b/orchagent/saihelper.cpp
@@ -334,7 +334,7 @@ void initSaiRedis()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to set communication mode, rv:%d", status);
-        exit(EXIT_FAILURE);
+        return handleSaiFailure(SAI_API_SWITCH, "set", status);
     }
 
     auto record_filename = Recorder::Instance().sairedis.getFile();
@@ -352,7 +352,7 @@ void initSaiRedis()
         {
             SWSS_LOG_ERROR("Failed to set SAI Redis recording output folder to %s, rv:%d",
                 record_location.c_str(), status);
-            exit(EXIT_FAILURE);
+            return handleSaiFailure(SAI_API_SWITCH, "set", status);
         }
 
         attr.id = SAI_REDIS_SWITCH_ATTR_RECORDING_FILENAME;
@@ -364,7 +364,7 @@ void initSaiRedis()
         {
             SWSS_LOG_ERROR("Failed to set SAI Redis recording logfile to %s, rv:%d",
                 record_filename.c_str(), status);
-            exit(EXIT_FAILURE);
+            return handleSaiFailure(SAI_API_SWITCH, "set", status);
         }
 
     }
@@ -378,7 +378,7 @@ void initSaiRedis()
     {
         SWSS_LOG_ERROR("Failed to %s SAI Redis recording, rv:%d",
             Recorder::Instance().sairedis.isRecord() ? "enable" : "disable", status);
-        exit(EXIT_FAILURE);
+        return handleSaiFailure(SAI_API_SWITCH, "set", status);
     }
 
     if (gRedisCommunicationMode == SAI_REDIS_COMMUNICATION_MODE_REDIS_ASYNC)
@@ -391,7 +391,7 @@ void initSaiRedis()
         if (status != SAI_STATUS_SUCCESS)
         {
             SWSS_LOG_ERROR("Failed to enable redis pipeline, rv:%d", status);
-            exit(EXIT_FAILURE);
+            return handleSaiFailure(SAI_API_SWITCH, "set", status);
         }
     }
 
@@ -409,7 +409,7 @@ void initSaiRedis()
         if (status != SAI_STATUS_SUCCESS)
         {
             SWSS_LOG_ERROR("Failed to set SAI REDIS response timeout");
-            exit(EXIT_FAILURE);
+            return handleSaiFailure(SAI_API_SWITCH, "set", status);
         }
 
         SWSS_LOG_NOTICE("SAI REDIS response timeout set successfully to %" PRIu64 " ", attr.value.u64);
@@ -422,7 +422,7 @@ void initSaiRedis()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to notify syncd INIT_VIEW, rv:%d gSwitchId %" PRIx64, status, gSwitchId);
-        exit(EXIT_FAILURE);
+        return handleSaiFailure(SAI_API_SWITCH, "set", status);
     }
     SWSS_LOG_NOTICE("Notify syncd INIT_VIEW");
 
@@ -436,7 +436,7 @@ void initSaiRedis()
         if (status != SAI_STATUS_SUCCESS)
         {
             SWSS_LOG_ERROR("Failed to set SAI REDIS response timeout");
-            exit(EXIT_FAILURE);
+            return handleSaiFailure(SAI_API_SWITCH, "set", status);
         }
 
         SWSS_LOG_NOTICE("SAI REDIS response timeout set successfully to %" PRIu64 " ", attr.value.u64);
@@ -1044,7 +1044,7 @@ std::vector<sai_stat_id_t> queryAvailableCounterStats(const sai_object_type_t ob
 
     if (!info)
     {
-        SWSS_LOG_WARN("Metadata info query failed, invalid object: %d", object_type);
+        SWSS_LOG_ERROR("Metadata info query failed, invalid object: %d", object_type);
         return stat_list;
     }
 
--- a/orchagent/saihelper.cpp
+++ b/orchagent/saihelper.cpp
@@ -678,15 +678,21 @@ task_process_status handleSaiGetStatus(sai_api_t api, sai_status_t status, void 
      *          in each orch.
      *       3. Take the type of sai api into consideration.
      */
+    string s_api = sai_serialize_api(api);
+    string s_status = sai_serialize_status(status);
+
     switch (status)
     {
         case SAI_STATUS_SUCCESS:
             SWSS_LOG_WARN("SAI_STATUS_SUCCESS is not expected in handleSaiGetStatus");
             return task_success;
-        case SAI_STATUS_NOT_IMPLEMENTED:
-            throw std::logic_error("SAI get function not implemented");
         default:
-            handleSaiFailure(api, "get", status);
+            /*
+             * handleSaiFailure() is not called for GET failures as it might
+             * overwhelm the system if there are too many such calls
+             */
+            SWSS_LOG_NOTICE("Encountered failure in GET operation, SAI API: %s, status: %s",
+                        s_api.c_str(), s_status.c_str());
             break;
     }
     return task_failed;
--- a/orchagent/saihelper.cpp
+++ b/orchagent/saihelper.cpp
@@ -567,15 +567,22 @@ task_process_status handleSaiCreateStatus(sai_api_t api, sai_status_t status, vo
      *          in each orch.
      *       3. Take the type of sai api into consideration.
      */
+    string s_api = sai_serialize_api(api);
+    string s_status = sai_serialize_status(status);
+
     switch (status)
     {
         case SAI_STATUS_SUCCESS:
+            return task_success;
         case SAI_STATUS_ITEM_NOT_FOUND:
         case SAI_STATUS_ADDR_NOT_FOUND:
         case SAI_STATUS_OBJECT_IN_USE:
-            SWSS_LOG_WARN("Status %d is not expected in handleSaiCreateStatus", status);
+            SWSS_LOG_WARN("Status %s is not expected for create operation, SAI API: %s",
+                            s_status.c_str(), s_api.c_str());
             return task_success;
         case SAI_STATUS_ITEM_ALREADY_EXISTS:
+            SWSS_LOG_NOTICE("Returning success for create operation, SAI API: %s, status: %s",
+                                s_api.c_str(), s_status.c_str());
             return task_success;
         case SAI_STATUS_INSUFFICIENT_RESOURCES:
         case SAI_STATUS_TABLE_FULL:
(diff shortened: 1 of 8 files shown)


prsunny:
@dgsudharsan , this is a long-awaited feature to prevent unnecessary orchagent crashes for some scenarios and to handle it gracefully. Please review and signoff.

orchagent/saihelper.cpp
```diff
@@ -566,119 +567,26 @@ task_process_status handleSaiCreateStatus(sai_api_t api, sai_status_t status, vo
...
-                case SAI_STATUS_TABLE_FULL:
-                    return task_need_retry;
-                default:
-                    SWSS_LOG_ERROR("Encountered failure in create operation, exiting orchagent, SAI API: %s, status: %s",
-                                sai_serialize_api(api).c_str(), sai_serialize_status(status).c_str());
-                    handleSaiFailure(true);
-                    break;
-            }
-            break;
+     switch (status)
```
qiluo-msft: >      [](http://example.com/codeflow?start=0&length=5)

Could you keep 4-space indentation?

prabhataravind: Thanks for catching that. Fixed it.

qiluo-msft (dismissed):

orchagent/saihelper.cpp
```diff
@@ -566,119 +567,26 @@ task_process_status handleSaiCreateStatus(sai_api_t api, sai_status_t status, vo
...
-                                sai_serialize_api(api).c_str(), sai_serialize_status(status).c_str());
-                    handleSaiFailure(true);
-                    break;
-            }
-            break;
+        case SAI_STATUS_SUCCESS:
+        case SAI_STATUS_ITEM_NOT_FOUND:
+        case SAI_STATUS_ADDR_NOT_FOUND:
+        case SAI_STATUS_OBJECT_IN_USE:
+            SWSS_LOG_WARN("Status %d is not expected in handleSaiCreateStatus", status);
```
dgsudharsan: Why do we need this warning about SAI_STATUS_SUCCESS?

orchagent/saihelper.cpp
```diff
@@ -566,119 +567,26 @@ task_process_status handleSaiCreateStatus(sai_api_t api, sai_status_t status, vo
...
+            return task_success;
+        case SAI_STATUS_ITEM_ALREADY_EXISTS:
```
dgsudharsan, line 583: I would recommend a warning or notice log here

orchagent/saihelper.cpp
```diff
@@ -694,68 +602,31 @@ task_process_status handleSaiSetStatus(sai_api_t api, sai_status_t status, void
...
+        case SAI_STATUS_ITEM_NOT_FOUND:
+        case SAI_STATUS_ADDR_NOT_FOUND:
+            /* There are specific cases especially with dual-TORs where tunnel
+             * routes and non-tunnel routes could be create for the same prefix
+             * which can potentially lead to conditions where ITEM_NOT_FOUND can
+             * be returned. This needs special handling in muxorch/routeorch.
+             */
+            return task_success;
+        case SAI_STATUS_INSUFFICIENT_RESOURCES:
```
dgsudharsan, line 634: This may be problematic in buffer as previously we returned task_failed. Would it keep on retrying? @stephenxs  Can you review?

orchagent/saihelper.cpp
```diff
@@ -694,68 +602,31 @@ task_process_status handleSaiSetStatus(sai_api_t api, sai_status_t status, void
...
-                    SWSS_LOG_ERROR("Encountered failure in set operation, exiting orchagent, SAI API: %s, status: %s",
-                            sai_serialize_api(api).c_str(), sai_serialize_status(status).c_str());
-                    handleSaiFailure(true);
-                    break;
-            }
-            break;
-        case SAI_API_TUNNEL:
-            switch (status)
-            {
-                case SAI_STATUS_ATTR_NOT_SUPPORTED_0:
```
dgsudharsan, line 726: This was introduced by PR https://github.com/sonic-net/sonic-swss/pull/2591 However this might now crash with new flow. @stephenxs  can you please review?

orchagent/saihelper.cpp
```diff
@@ -772,57 +643,26 @@ task_process_status handleSaiRemoveStatus(sai_api_t api, sai_status_t status, vo
...
-                case SAI_STATUS_ITEM_NOT_FOUND:
-                    return task_success;
-                default:
-                    SWSS_LOG_ERROR("Encountered failure in remove operation, exiting orchagent, SAI API: %s, status: %s",
-                                sai_serialize_api(api).c_str(), sai_serialize_status(status).c_str());
-                    handleSaiFailure(true);
-                    break;
-            }
-            break;
+        case SAI_STATUS_SUCCESS:
```
dgsudharsan, line 665: Prefer avoiding warning log for SAI_STATUS_SUCCESS

orchagent/saihelper.cpp
```diff
@@ -772,57 +643,26 @@ task_process_status handleSaiRemoveStatus(sai_api_t api, sai_status_t status, vo
...
+    switch (status)
     {
-        case SAI_API_ROUTE:
-            switch (status)
-            {
-                case SAI_STATUS_SUCCESS:
-                    SWSS_LOG_WARN("SAI_STATUS_SUCCESS is not expected in handleSaiRemoveStatus");
-                    return task_success;
-                case SAI_STATUS_ITEM_NOT_FOUND:
-                case SAI_STATUS_NOT_EXECUTED:
```
dgsudharsan, line 784: SAI_STATUS_NOT_EXECUTED is not handled in the new flow below. Should we handle that?

https://github.com/sonic-net/sonic-swss/pull/2688

dgsudharsan (changes requested):

orchagent/saihelper.cpp
prabhataravind: @dgsudharsan this is just something I retained from current code. See https://github.com/sonic-net/sonic-swss/blob/master/orchagent/saihelper.cpp#L575. We don't expect anybody to call handleSai* functions passing in SAI_STATUS_SUCCESS.

orchagent/saihelper.cpp
prabhataravind: The current implementation does an infinite retry. There is plan to address this in the next phase to limit retries. Would that suffice or is this something that we need to handle right away? Is this condition likely to be hit frequently during normal operations?

orchagent/saihelper.cpp
prabhataravind: This is handled, We realized that SAI_STATUS_NOT_EXECUTED can be returned as a generic return value when using bulkers. A sample output for this scenario is shown in the PR description. I have also verified this by running test_duplicate_route.py in sonic-mgmt.

orchagent/saihelper.cpp
prabhataravind: I kept this as-is from legacy code, but I can remove it.

orchagent/saihelper.cpp
prabhataravind: In this case, we won't crash, as handleSaiFailure() does not call abort() anymore. We return task_failed which retains parity with current code.

prsunny (dismissed):
Overall, the approach lgtm and per HLD. Please address comments from @dgsudharsan for merge.

prsunny:
@prabhataravind , you may need to create a separate PR for 202505 after merging this since there is a cherry-pick conflict.

orchagent/saihelper.cpp
prabhataravind: Removed the warning log.

orchagent/saihelper.cpp
prabhataravind: Added a notice log.

orchagent/saihelper.cpp
prabhataravind: Fixed.

prabhataravind:
@dgsudharsan addressed your comments. Could you please check again?

orchagent/saihelper.cpp
prabhataravind: @stephenxs please review when you get a chance as per Sudarshan's comments.

prabhataravind:
> Overall, the approach lgtm and per HLD. Please address comments from @dgsudharsan for merge.

Fixed all of Sudarshan's comments regarding logs and responded to the others. Will wait for Sudarshan and Stephen to sign-off.

orchagent/saihelper.cpp
dgsudharsan: I don't think it is handled the same way as previously. Before this change it returned task_success. However after this change it will throw error log and take dump. The behavior should not change in my opinion

prabhataravind: @dgsudharsan, IMO the previous handling was not right. There are cases where SAI_STATUS_NOT_EXECUTED is returned for bulk calls when the actual error is something genuine that can't be treated as a success. Per the HLD referenced in the PR, we are treating SAI_STATUS_NOT_EXECUTED as an error case which warrants an error log and sai dump, however in this case we don't call abort() for orchagent.

prabhataravind:
> @prabhataravind , you may need to create a separate PR for 202505 after merging this since there is a cherry-pick conflict.

@prsunny, created PR https://github.com/sonic-net/sonic-swss/pull/3706 for 202505.

dgsudharsan:
Waiting for @stephenxs  to close on the [comment](https://github.com/sonic-net/sonic-swss/pull/3587#discussion_r2133237345) . Rest of the comments are addressed.

orchagent/saihelper.cpp
stephenxs: On our platform, `SAI_STATUS_INSUFFICIENT_RESOURCES` on buffer setting is usually caused by a wrong user configuration. In this case, the user should correct the configuration.
Originally, we returned task_fail on such error, so that the item could be removed from buffer orch's `m_toSync`, which caused inconsistency between user configuration and SAI/SDK/ASIC. After the user corrects the configuration, a new item will be added to `m_toSync` and then pushed to the ASIC. The user configuration and SAI/SDK/ASIC become consistent again.
With this PR, the problematic item stays in `m_toSync` on `SAI_STATUS_INSUFFICIENT_RESOURCES` and keeps retrying to apply to SAI until user corrects the problematic configuration.

It should be OK if there is a delay between retry. @prabhataravind  @dgsudharsan

stephenxs:
> Waiting for @stephenxs to close on the [comment](https://github.com/sonic-net/sonic-swss/pull/3587#discussion_r2133237345) . Rest of the comments are addressed.

Done.

orchagent/saihelper.cpp
dgsudharsan: @stephenxs how will that entry be removed? If the user adds correct configuration, will it override the already present entry in m_toSync? Will the retry overwhelm the orchagent? @prabhataravind  @prsunny

stephenxs: > @stephenxs how will that entry be removed? If the user adds correct configuration, will it override the already present entry in m_toSync? Will the retry overwhelm the orchagent? @prabhataravind @prsunny

@dgsudharsan after the user corrects the configuration, orchagnet won't retry anymore. The correct configuration will overwrite the problematic one.
I remember there is a 1-second delay between each retry but need to double check.

prabhataravind: @stephenxs @dgsudharsan @prsunny there is a 1000ms timeout for select in orchdaemon.cpp, so if there's no config change for these tables, it should wait until this timeout for each retry.

/* select() function timeout retry time */
#define SELECT_TIMEOUT 1000

prabhataravind:
@dgsudharsan could you pls approve if there are no outstanding comments/issues?

orchagent/saihelper.cpp
stephenxs: thanks. This is exactly what I referenced.

prabhataravind:
@qiluo-msft @prsunny could you please approve again? Earlier reviews got dismissed after addressing some review comments.

qiluo-msft (approved):

vrajeshe:
@prabhataravind can you please backport  this to 202411 branch too ?

deekshabhandary13:
@yejianquan can you please add the label for merging this commit to 202411 branch too?

yejianquan:
> @yejianquan can you please add the label for merging this commit to 202411 branch too?

@kperumalbfn (release manager of 202411) for 202411 requests

prsunny:
@deekshabhandary13 , 202411 is almost frozen. This is available from 202505.

Pterosaur:
Hi @prabhataravind , Due to the conflict, could you please create a PR for 202412?
https://github.com/Azure/sonic-swss.msft/tree/202412

PR #3588, opened 2025-04-02
[Dynamic Buffer][Mellanox] Fix an issue when handling 2-digit queue ID in the Lua plugin

stephenxs:
**What I did**

Fix an issue when handling 2-digit queue ID in the Lua plugin

**Why I did it**

Bug fix.

**How I verified it**

Manual and VS test

**Details if related**

### DIFF
--- a/tests/test_buffer_dynamic.py
+++ b/tests/test_buffer_dynamic.py
@@ -905,7 +905,7 @@ class TestBufferMgrDyn(object):
             dvs.runcmd("kill -s SIGCONT {}".format(oa_pid))
 
 
-    def test_bufferPoolPercentage(self, dvs, testlog):
+    def test_bufferPoolCalculation(self, dvs, testlog):
         self.setup_db(dvs)
 
         try:
@@ -914,6 +914,7 @@ class TestBufferMgrDyn(object):
             pass
 
         try:
+            # Test 1: Test the buffer pool calculation with a percentage
             percentage = 75
             margin = 1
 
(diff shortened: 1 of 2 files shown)


tests/test_buffer_dynamic.py
```diff
@@ -931,6 +932,45 @@ def test_bufferPoolPercentage(self, dvs, testlog):
...
+                                       'size': '16384'})
+
+            # Create buffer queue entries
+            self.config_db.update_entry('BUFFER_QUEUE', 'Ethernet0|0-7', {'profile': 'egress_test_profile'})
+            self.config_db.update_entry('BUFFER_QUEUE', 'Ethernet0|8-12', {'profile': 'egress_test_profile'})
+            self.config_db.update_entry('BUFFER_QUEUE', 'Ethernet0|13-19', {'profile': 'egress_test_profile'})
+
+            # Run lua script and check output
+            _, output = dvs.runcmd("redis-cli --eval /usr/share/swss/buffer_pool_vs.lua")
+            assert re.search(r"debug:BUFFER_PROFILE_TABLE:egress_test_profile:16384:20", output), "Profile reference count not found in output"
```
kperumalbfn, line 961: Can we also double check ASIC_DB?

stephenxs: Hi @kperumalbfn 
Understanding the mapping between ASIC_DB and APPL DB requires more complicated logic which we do not have in vs.
We have a lot of checks between ASIC/APPL DB in the sonic-mgmt test leveraging existing infra there
So, I don't want to add such logic here.
What do you think?

stephenxs:
same issue as that in other PRs. Retrying

prsunny:
@kperumalbfn , is it good to merge?

tests/test_buffer_dynamic.py
kperumalbfn: Can we use counter_db's QUEUE_NAME_MAP to get the queue OID based on the port+queueIndex and validate with ASIC_DB? This will make sure that profiles are attached to the queues 10-19

stephenxs: Hi @kperumalbfn 
We do not need to do it because we didn't change the queue mappings or whether the buffer profile is applied to queue 10-19. That is guaranteed by buffer orch.
We just corrected a scenario where there are two digits in BUFFER_QUEUE table and translated it to a correct number.
eg, if we have Ethernet8|7-15, it should be 9 queues. But it got the wrong result without this change.

PR #3596, opened 2025-04-08
[orchagent] Add value validation for RouteOrch, NhgOrch and Srv6Orch

GaladrielZhao:
Add null checks for variable assignments in RouteOrch, NhgOrch and Srv6Orch. The assignments will only proceed if the string is not empty.

**What I did**
  Add value validity checks to the assignment part of SET_COMMAND in the doTask methods for RouteOrch, NhgOrch and Srv6Orch, ensuring that the values assigned to fields are not empty.

**Why I did it**
  For the value obtained by orchagent, assigning an empty value without validation may cause an issue. 
  So add a validity check before assigning the value.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/nhgorch.cpp
+++ b/orchagent/nhgorch.cpp
@@ -70,29 +70,28 @@ void NhgOrch::doTask(Consumer& consumer)
             /* Get group's next hop IPs and aliases */
             for (auto i : kfvFieldsValues(t))
             {
-                if (fvField(i) == "nexthop")
+                if (fvField(i) == "nexthop" && fvValue(i) != "")
                     ips = fvValue(i);
 
-                if (fvField(i) == "ifname")
+                if (fvField(i) == "ifname" && fvValue(i) != "")
                     aliases = fvValue(i);
 
-                if (fvField(i) == "weight")
+                if (fvField(i) == "weight" && fvValue(i) != "")
                     weights = fvValue(i);
 
-                if (fvField(i) == "mpls_nh")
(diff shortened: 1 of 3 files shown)


prsunny:
@GaladrielZhao , could you explain how to repro this? is it a manual config error or due to any changes in fpmsyncd?

eddieruan-alibaba (approved):

GaladrielZhao:
> @GaladrielZhao , could you explain how to repro this? is it a manual config error or due to any changes in fpmsyncd?

@prsunny       In the current implementation, the Orch::doTask can handle the SET and DEL opertations for table entries, but it cannot handle the addition and deletion of specific fields. For example, if a route is updated from an SRv6 route to a regular route, its "segment" field will become "". If we follow the current implementation (as shown in the following code) by directly assigning an empty string to the "srv6_segments" and setting both "srv6_seg" and "srv6_nh" to true without any checks, the Orch will not be able to recognize the change that the route is no longer an SRv6 route. It will assume that it is an SRv6 route with an empty segment, which is incorrect.

```
if (fvField(i) == "segment") {
        srv6_segments = fvValue(i);
        srv6_seg = true;
        srv6_nh = true;
}
```

Therefore, adding a check for empty values can enable the Orch to know such changes in fields. As for those fields that must exist in table entries, validation is added for safety and consistency considerations.

prsunny (approved):
lgtm, @abdosi for viz

PR #3597, opened 2025-04-08
Local ARS (Adaptive Routing and Switching)

VladimirKuk:
Added support for local ARS (Adaptive Routing and Switching).

HLD: https://github.com/sonic-net/SONiC/pull/1958

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -11735,6 +11735,63 @@ bool PortsOrch::setPortPtTam(const Port& port, sai_object_id_t tam_id)
     return true;
 }
 
+bool PortsOrch::setPortArsEnable(const Port& port, bool is_enable)
+{
+    SWSS_LOG_ENTER();
+    sai_attribute_t attr;
+
+    attr.id = SAI_PORT_ATTR_ARS_ENABLE;
+    attr.value.booldata = is_enable;
+    sai_status_t status = sai_port_api->set_port_attribute(port.m_port_id, &attr);
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR(
+            "Failed to %s ARS on port %s (OID 0x%" PRIx64 "): SAI status = %d",
+            is_enable ? "enable" : "disable",
+            port.m_alias.c_str(),
+            port.m_port_id,
+            status);
+        task_process_status handle_status = handleSaiSetStatus(SAI_API_PORT, status);
+        if (handle_status != task_success)
+        {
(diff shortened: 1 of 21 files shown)


orchagent/portsorch.cpp
```diff
@@ -10173,6 +10173,66 @@ bool PortsOrch::setPortPtTam(const Port& port, sai_object_id_t tam_id)
...
+        task_process_status handle_status = handleSaiSetStatus(SAI_API_PORT, status);
+        if (handle_status != task_success)
+        {
+            return parseHandleSaiStatusFailure(handle_status);
+        }
+    }
+
+    return true;
+}
+
```
Yubin-Li, line 10221: syntax error

orchagent/routeorch.cpp
```diff
@@ -2998,3 +3429,74 @@ inline void RouteOrch::removeVipRouteSubnetDecapTerm(const IpPrefix &ipPrefix)
     m_appTunnelDecapTermProducer.del(key);
     m_SubnetDecapTermsCreated.erase(it);
 }
+
+
+bool RouteOrch::reconfigureNexthopGroupWithArsState(NextHopGroupKey nexthopGroupKey, const sai_object_id_t ars_object_id)
```
eddyk-nvidia, line 3434: I don't see where this method is called. Do you create a new ARS object for each NHG or re-use the same ARS object if it shares the same configuration ?

VladimirKuk: It is called in arsorch.cpp on NHG update.
ARS object is per NHG.

baxia-lan:
This PR has been stale for a while, any plan to merge it?

orchagent/portsorch.cpp
```diff
@@ -10173,6 +10173,52 @@ bool PortsOrch::setPortPtTam(const Port& port, sai_object_id_t tam_id)
...
+    sai_attribute_t attr;
+
+    if (scaling_factor == 0)
+    {
+        SWSS_LOG_ERROR("Invalid scaling factor %u for port %s", scaling_factor, port.m_alias.c_str());
+        return false;
+    }
+
+    attr.id = SAI_PORT_ATTR_ARS_PORT_LOAD_SCALING_FACTOR;
+    attr.value.u32 = port.m_speed / scaling_factor;
```
baxia-lan: This transition brings ambiguity. Why not just set the value of SAI_PORT_ATTR_ARS_PORT_LOAD_SCALING_FACTOR to be value of scaling_factor?

apannerselva: @baxia-lan i have addressed it. Will be using SAI values.

PR #3603, opened 2025-04-10
Fix a pipeline issue due to dependency in docker-sonic-vs

sdszhang:
Fix a pipeline issue due to dependency in docker-sonic-vs. 
similar change as https://github.com/sonic-net/sonic-swss-common/pull/999


sdszhang:
@saiarcot895 can you help to review?

saiarcot895 (approved):

prsunny (approved):

PR #3608, opened 2025-04-15
[DASH] Update dash orch

jimmyzhai:
**What I did**
- Check vnet exists for ROUTING_TYPE_VNET_DIRECT in `addOutboundRouting`
- Not remove outbound routing, route group if in use
- Keep adding OutboundCaToPa and PaValidation consistently

**Why I did it**
In updating dash vnet test cases - https://github.com/sonic-net/sonic-mgmt/pull/17994, find some DASH objects cannot be added/removed due to the dash object dependency.

**How I verified it**
All dash objects can be added/removed successfully in dash vnet test.

**Details if related**

### DIFF
--- a/orchagent/dash/dashrouteorch.cpp
+++ b/orchagent/dash/dashrouteorch.cpp
@@ -76,9 +76,16 @@ bool DashRouteOrch::addOutboundRouting(const string& key, OutboundRoutingBulkCon
         SWSS_LOG_INFO("Retry as route group %s not found", ctxt.route_group.c_str());
         return false;
     }
-    if (ctxt.metadata.has_vnet() && gVnetNameToId.find(ctxt.metadata.vnet()) == gVnetNameToId.end())
+    if (ctxt.metadata.routing_type() == dash::route_type::RoutingType::ROUTING_TYPE_VNET &&
+        ctxt.metadata.has_vnet() && gVnetNameToId.find(ctxt.metadata.vnet()) == gVnetNameToId.end())
     {
         SWSS_LOG_INFO("Retry as vnet %s not found", ctxt.metadata.vnet().c_str());
+        return false;
+    }
+    if (ctxt.metadata.routing_type() == dash::route_type::RoutingType::ROUTING_TYPE_VNET_DIRECT &&
+        ctxt.metadata.has_vnet_direct() && gVnetNameToId.find(ctxt.metadata.vnet_direct().vnet()) == gVnetNameToId.end())
+    {
+        SWSS_LOG_INFO("Retry as vnet %s not found", ctxt.metadata.vnet_direct().vnet().c_str());
         return false;
     }
 
@@ -220,7 +227,7 @@ bool DashRouteOrch::removeOutboundRouting(const string& key, OutboundRoutingBulk
     if (isRouteGroupBound(ctxt.route_group))
     {
         SWSS_LOG_WARN("Cannot remove route from route group %s as it is already bound", ctxt.route_group.c_str());
-        return true;
+        return false;
     }
 
     auto& object_statuses = ctxt.object_statuses;
@@ -727,6 +734,11 @@ bool DashRouteOrch::removeRouteGroup(const string& route_group)
     sai_status_t status = sai_dash_outbound_routing_api->remove_outbound_routing_group(route_group_oid);
     if (status != SAI_STATUS_SUCCESS)
     {
+        //Retry later if object is in use
+        if (status == SAI_STATUS_OBJECT_IN_USE)
+        {
+            return false;
+        }
         SWSS_LOG_ERROR("Failed to remove route group %s", route_group.c_str());
         task_process_status handle_status = handleSaiRemoveStatus((sai_api_t) SAI_API_DASH_OUTBOUND_ROUTING, status);
         if (handle_status != task_success)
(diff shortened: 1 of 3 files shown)


orchagent/dash/dashrouteorch.cpp
```diff
@@ -220,7 +227,7 @@ bool DashRouteOrch::removeOutboundRouting(const string& key, OutboundRoutingBulk
     if (isRouteGroupBound(ctxt.route_group))
     {
         SWSS_LOG_WARN("Cannot remove route from route group %s as it is already bound", ctxt.route_group.c_str());
-        return true;
+        return false;
```
prsunny, line 230: Why is this changed?

orchagent/dash/dashrouteorch.cpp
```diff
@@ -76,11 +76,18 @@ bool DashRouteOrch::addOutboundRouting(const string& key, OutboundRoutingBulkCon
...
+    if (ctxt.metadata.routing_type() == dash::route_type::RoutingType::ROUTING_TYPE_VNET &&
+        ctxt.metadata.has_vnet() && gVnetNameToId.find(ctxt.metadata.vnet()) == gVnetNameToId.end())
     {
         SWSS_LOG_INFO("Retry as vnet %s not found", ctxt.metadata.vnet().c_str());
         return false;
     }
+    if (ctxt.metadata.routing_type() == dash::route_type::RoutingType::ROUTING_TYPE_VNET_DIRECT &&
+        ctxt.metadata.has_vnet_direct() && gVnetNameToId.find(ctxt.metadata.vnet_direct().vnet()) == gVnetNameToId.end())
+    {
+        SWSS_LOG_INFO("Retry as vnet %s not found", ctxt.metadata.vnet_direct().vnet().c_str());
```
prsunny, line 88: Can you differentiate the logs for two conditions? We can just put a routing type string here.

orchagent/dash/dashvnetorch.cpp
```diff
@@ -446,11 +448,6 @@ bool DashVnetOrch::addVnetMap(const string& key, VnetMapBulkContext& ctxt)
         }
 
         remove_from_consumer = addOutboundCaToPa(key, ctxt);
-        // If addOutboundCaToPa fails, skip addPaValidation
```
prsunny, line 449: Why is this removed?

orchagent/dash/dashrouteorch.cpp
jimmyzhai: Updated the log to include routing type string

orchagent/dash/dashrouteorch.cpp
theasianpianist: IMO this should have been false to begin with, if route group is bound then we can keep trying to remove it, similar to removing objects w/ nonzero reference counts

orchagent/dash/dashvnetorch.cpp
theasianpianist: There was a logic error in the existing code, moving `addPaValidation` to inside `addOutboundCaToPa` should fix it. @prabhataravind for viz

theasianpianist (approved):

orchagent/dash/dashrouteorch.cpp
```diff
@@ -76,9 +76,22 @@ bool DashRouteOrch::addOutboundRouting(const string& key, OutboundRoutingBulkCon
         SWSS_LOG_INFO("Retry as route group %s not found", ctxt.route_group.c_str());
         return false;
     }
-    if (ctxt.metadata.has_vnet() && gVnetNameToId.find(ctxt.metadata.vnet()) == gVnetNameToId.end())
+
+    std::string routing_type_str = dash::route_type::RoutingType_Name(ctxt.metadata.routing_type());
+    if (ctxt.metadata.routing_type() == dash::route_type::RoutingType::ROUTING_TYPE_VNET &&
```
prabhataravind, line 81: What about the case for ROUTING_TYPE_PRIVATELINK? We will want to return false in that case too, right?

orchagent/dash/dashvnetorch.cpp
prabhataravind: Could you add a comment in addVnetMap that addPaValidation is being called in addOutboundCaToPa to avoid dependency issues?

orchagent/dash/dashrouteorch.cpp
jimmyzhai: ROUTING_TYPE_PRIVATELINK is not allowed here, caught by line 106 - 111:

```
        SWSS_LOG_WARN("Routing type %s for outbound routing entry %s not allowed", routing_type_str.c_str(), key.c_str());
        return false;
```

It is handled at vnet mapping case - `addOutboundCaToPa`

orchagent/dash/dashvnetorch.cpp
jimmyzhai: Added the comment in addVnetMap

orchagent/dash/dashrouteorch.cpp
jimmyzhai: Yes, 'false' means to not stop trying till route group is unbounded.

prsunny (approved):
lgtm,

PR #3615, opened 2025-04-17
dot3 Stats collection

bhouse-nexthop:
**What I did**

Implement RFC3635 dot3 statistics collection.

Used by https://github.com/sonic-net/sonic-snmpagent/pull/XXXX

**Why I did it**

RFC1284 defines dot3 stats that most switch vendors support.  This RFC was superseded by RFC3635 which includes 64bit "HC" counters. We need to collect these statistics for use by sonic_snmpagent.

**How I verified it**

Tested on Mellanox SN2201 physical hardware, where dot3 stats are implemented by the SAI.

**Details if related**

The Broadcom SAI does not currently implement these statistics.  A case has been raised with Broadcom to implement these stats.


prsunny:
@bhouse-nexthop , just want to confirm this will not have an impact if the underlying h/w implementation does not support the counters, right?

bhouse-nexthop:
@prsunny somehow I missed your reply on this.  Correct, I've tested it on Broadcom TH5 which currently isn't outputting these in their SAI implementation and it doesn't break anything.  However on Mellanox/Nvidia the stats are output correctly.

I just rebased this PR in order to fix the merge conflicts.

PR #3626, opened 2025-04-24
[voq][chassis][orchagent][lagid] Modified the lua script to fix orhagent crash during upgrade from 202205 to 202405

mlok-nokia:
**What I did**
Modify the lagids.lua script to check both SYSTEM_LAG_IDS_FREE_LIST and SYSTEM_LAG_ID_SET to make sure the LagId is really free before it is assigned.  If it is in the FREE_LIST and also in the SYSTEM_LAG_ID_SET, remove it from the FREE_LIST and check/use the next one in the FREE_LIST. This will avoid using the ID which has been using by the LC which is still running the 202205 image.  This will address the orchagent    

**Why I did it**
The orhcgent crash during upgrade from 202205 to 202405 is caused by the same lagid used by multiple PortChannels.  On chassis with mulitple LCs,  upgrade LCs image from 202205 to 202405.   orchagent crash could be occured when one LC is running with latest image while other LCs is still running with the old image. This PR fixes issue #https://github.com/Nokia-ION/ndk/issues/84

**How I verified it**
While chassis with multiple LCs are running 202205 image, Upgrade Supervisor first, then upgrade LCs.  Check there is not be any orchagent crashed in any LC.           

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [x] 202405
- [x] 202411

**Details if related**

### DIFF
--- a/orchagent/lagids.lua
+++ b/orchagent/lagids.lua
@@ -39,46 +39,50 @@ if op == "add" then
         -- the portchannel does not exist in the database
         -- If proposed lagid is not available, lpop the first availabe ID
         local index = redis.call("lpos", "SYSTEM_LAG_IDS_FREE_LIST", tostring(plagid))
-        if index == false then
-            if redis.call("llen", "SYSTEM_LAG_IDS_FREE_LIST") <= 0 then
-                return -1
-            end
-            local lagid = redis.call("lpop", "SYSTEM_LAG_IDS_FREE_LIST")
-            redis.call("hset", "SYSTEM_LAG_ID_TABLE", pcname, lagid)
-            redis.call("sadd", "SYSTEM_LAG_ID_SET", lagid)
-            if dblagid then
-                redis.call("srem", "SYSTEM_LAG_ID_SET", tostring(dblagid))
-                if redis.call("lpos", "SYSTEM_LAG_IDS_FREE_LIST", tostring(dblagid)) == false then
-                    redis.call("rpush", "SYSTEM_LAG_IDS_FREE_LIST", tostring(dblagid))
+        if index ~= false then
+            if redis.call("sismember", "SYSTEM_LAG_ID_SET", tostring(plagid)) == 0 then
+                redis.call("lrem", "SYSTEM_LAG_IDS_FREE_LIST", 1, tostring(plagid))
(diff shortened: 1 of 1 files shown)


mlok-nokia:
@Javier-Tan for viz

orchagent/lagids.lua
```diff
@@ -39,46 +39,50 @@ if op == "add" then
...
-        redis.call("sadd", "SYSTEM_LAG_ID_SET", lagid)
-        if dblagid then
-            if redis.call("lpos", "SYSTEM_LAG_IDS_FREE_LIST", tostring(dblagid)) == false then
-                redis.call("rpush", "SYSTEM_LAG_IDS_FREE_LIST", tostring(dblagid))
-            end
+        lagid = redis.call("lpop", "SYSTEM_LAG_IDS_FREE_LIST")
+    end
+
+    -- Remove it from free list in case there is duplicated one 
+    redis.call("lrem", "SYSTEM_LAG_IDS_FREE_LIST", 1, lagid)
```
Javier-Tan, line 76: Any reason not to use count = 0 to remove potential duplicates here?
`redis.call("lrem", "SYSTEM_LAG_IDS_FREE_LIST", 0, lagid)`

Javier-Tan (commented):
I've been able to recreate the scenario with previous 202405 version, and did not encounter the same problem when trying to recreate with this change (once).

Thanks @mlok-nokia, LGTM besides a small comment, will approve when addressed

orchagent/lagids.lua
Javier-Tan: Also applies to line 44 if we want to check for duplicates then

mlok-nokia: Thanks for the review and suggestion. Fixed as suggested.

Javier-Tan (approved):
LGTM

Javier-Tan:
@mlok-nokia could you rebase this PR so we can merge it? Thanks!

mlok-nokia:
> @mlok-nokia could you rebase this PR so we can merge it? Thanks!

Done. Thanks

PR #3635, opened 2025-05-08
[bfdorch] Must have mandatory attribute SAI_BFD_SESSION_ATTR_NEXT_HOP_ID in creating bfd session

jimmyzhai:
**What I did**

- Add a mandatory attribute SAI_BFD_SESSION_ATTR_NEXT_HOP_ID in creating bfd session
- Update AZP build to take artifacts from the sonic-saireds PR https://github.com/sonic-net/sonic-sairedis/pull/1589 build to allow bfd testcase pass in vstest with updating SAI headers.

**Why I did it**
In PR https://github.com/sonic-net/sonic-sairedis/pull/1589 check, vstest is being blocked due to bfd session creation failure, which is caused by the new mandatory attribute SAI_BFD_SESSION_ATTR_NEXT_HOP_ID not being added in the bfd session creation.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/bfdorch.cpp
+++ b/orchagent/bfdorch.cpp
@@ -348,6 +348,7 @@ bool BfdOrch::create_bfd_session(const string& key, const vector<FieldValueTuple
     MacAddress dst_mac;
     bool dst_mac_provided = false;
     bool src_ip_provided = false;
+    sai_object_id_t next_hop_id = SAI_NULL_OBJECT_ID;
 
     sai_attribute_t attr;
     vector<sai_attribute_t> attrs;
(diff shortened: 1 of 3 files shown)


kcudnik (approved):

orchagent/bfdorch.cpp
```diff
@@ -517,6 +518,14 @@ bool BfdOrch::create_bfd_session(const string& key, const vector<FieldValueTuple
         attr.id = SAI_BFD_SESSION_ATTR_DST_MAC_ADDRESS;
         memcpy(attr.value.mac, dst_mac.getMac(), sizeof(sai_mac_t));
         attrs.emplace_back(attr);
+
+        attr.id = SAI_BFD_SESSION_ATTR_USE_NEXT_HOP;
+        attr.value.booldata = true;
```
prsunny, line 523: i'm confused, default is  'false' and not use NEXTHOP attribute.

jimmyzhai: Here it is under the condition of `alias != "default"`.

jimmyzhai:
@prsunny, hold on, looks this PR is not needed.

jimmyzhai:
Not needed. Deprecate it.

PR #3636, opened 2025-05-09
Setting default nexthop weight to 1 in `fpmsyncd`

mramezani95:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -2388,14 +2388,9 @@ string RouteSync::getNextHopWt(struct rtnl_route *route_obj)
         struct rtnl_nexthop *nexthop = rtnl_route_nexthop_n(route_obj, i);
         /* Get the weight of next hop */
         uint8_t weight = rtnl_route_nh_get_weight(nexthop);
-        if (weight)
-        {
-            result += to_string(weight);
-        }
-        else
-        {
-            return "";
-        }
+        if (weight == 0)
+            weight = 1; // default weight is 1
+        result += to_string(weight);
 
         if (i + 1 < rtnl_route_get_nnexthops(route_obj))
         {
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -2390,10 +2390,7 @@ string RouteSync::getNextHopWt(struct rtnl_route *route_obj)
         uint8_t weight = rtnl_route_nh_get_weight(nexthop);
         if (weight == 0)
         {
-            nl_addr* nh_addr = rtnl_route_nh_get_gateway(nexthop);
-            char nh_addr_str[MAX_ADDR_SIZE + 1] = {0};
-            nl_addr2str(nh_addr, nh_addr_str, MAX_ADDR_SIZE);
-            SWSS_LOG_INFO("Using default weight of 1 for nexthop %s", nh_addr_str);
+            SWSS_LOG_INFO("Using default weight of 1 for nexthop");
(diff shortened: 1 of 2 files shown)


fpmsyncd/routesync.cpp
```diff
@@ -2388,14 +2388,9 @@ string RouteSync::getNextHopWt(struct rtnl_route *route_obj)
...
         uint8_t weight = rtnl_route_nh_get_weight(nexthop);
-        if (weight)
-        {
-            result += to_string(weight);
-        }
-        else
-        {
-            return "";
-        }
+        if (weight == 0)
```
prsunny, line 2391: Please use `{` to follow the coding style

prsunny (commented):
Add unit tests

fpmsyncd/routesync.cpp
mramezani95: Fixed.

fpmsyncd/routesync.cpp
```diff
@@ -2388,14 +2388,11 @@ string RouteSync::getNextHopWt(struct rtnl_route *route_obj)
         struct rtnl_nexthop *nexthop = rtnl_route_nexthop_n(route_obj, i);
         /* Get the weight of next hop */
         uint8_t weight = rtnl_route_nh_get_weight(nexthop);
-        if (weight)
+        if (weight == 0)
```
vaibhavhd, line 2391: Can you please add some log (DEBUG level?) here to mention what is being done?

I wanted it to be NOTICE, but Saikrishna mentions that this will lead to chatty syslog (for all routes this message will appear). Can we use NOTICE at least once somewhere to highlight that the default weight was set?

vaibhavhd (approved):
Please wait for Prince/Prabhat's approval too. Thanks!

fpmsyncd/routesync.cpp
prsunny: we can add an INFO level debug. NOTICE is going to be chatty as Sai mentioned

mramezani95: I have added an INFO log every time we use default weight for a nexthop and a NOTICE log for just the first time.

fpmsyncd/routesync.cpp
```diff
@@ -2381,21 +2381,28 @@ string RouteSync::getNextHopIf(struct rtnl_route *route_obj)
...
-        else
-        {
-            return "";
+            nl_addr* nh_addr = rtnl_route_nh_get_gateway(nexthop);
+            char nh_addr_str[MAX_ADDR_SIZE + 1] = {0};
+            nl_addr2str(nh_addr, nh_addr_str, MAX_ADDR_SIZE);
+            SWSS_LOG_INFO("Using default weight of 1 for nexthop %s", nh_addr_str);
+            if (!default_weight_logged)
+            {
+                SWSS_LOG_NOTICE("Using default weight of 1 for nexthop %s", nh_addr_str);
```
prabhataravind, line 2400: I would not recommend having LOG_NOTICE log here even if it is once per nexthop as this can also flood the syslog in case of link or bgp flaps. In addition, having an "if" check for every nexthop in a route programming path is not desirable from a performance point of view. We will know the weights that are being set from swss records anyway, so we won't miss out on knowing the weights that were passed down to orchagent.

mramezani95: I removed the NOTICE log, but this log was not generated once per nexthop. It was generated just once for one of the nexthops.

PR #3641, opened 2025-05-12
Change Log Level for BFD Offload Capability Implementation

ram25794:
**What I did**

Changed BFD offload capability Log Level to NOTICE Instead of error.

**Why I did it**

The BFD offload capability is optional for the product, so changing the log level from ERROR to NOTICE.

**How I verified it**

**Details if related**


dgsudharsan (approved):

bingwang-ms:
Retriggered PR test.

PR #3624, opened 2025-04-22
[orchagent] CoPP neighbor miss trap and enhancements

rminnikanti:
**What I did**

- Added neighbor_miss trap type support
- enum capability query for hostif trap type in copporch.cpp
- Added hw_status field to COPP_TRAP_TABLE table in state_db

**Why I did it**

HLD: [sonic-net/SONiC#1943](https://github.com/sonic-net/SONiC/pull/1943)

**How I verified it**

Verified with test_copp.py DVS test
Verified on Sonic switch with and without neighbor_miss trap support by vendorSAI.

**Details if related**

PR Dependencies:

- Depends on: sonic-net/sonic-swss-common#1010
- Depends on: sonic-net/sonic-sairedis#1586

Signed-off-by: Ravi Minnikanti <rminnikanti@marvell.com>

PR #3646, opened 2025-05-14
Add NEIGHBOR_MISS trap type and its own queue

msherif1234:
**What I did**
Add new trap type `SAI_HOSTIF_TRAP_TYPE_NEIGHBOR_MISS` and dedicated queue 
to better control those punt packets and not impact other critical punt traffic

**Why I did it**
As proposed in https://github.com/sonic-net/SONiC/blob/1fba16368a803bfeea57d93e8e39b3bcdcc3a67d/doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md we wanted to protect cpu from being hogged with neighbor miss packets

**How I verified it**

**Details if related**


dgsudharsan:
@msherif1234  we already have a PR for this. https://github.com/sonic-net/sonic-swss/pull/3624

msherif1234:
> @msherif1234 we already have a PR for this. #3624

yeah sorry didn't notice the open PR I will close mine Thanks

PR #3648, opened 2025-05-15
[202405] Setting default nexthop weight to 1 in `fpmsyncd`

mramezani95:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -1300,14 +1300,15 @@ string RouteSync::getNextHopWt(struct rtnl_route *route_obj)
         struct rtnl_nexthop *nexthop = rtnl_route_nexthop_n(route_obj, i);
         /* Get the weight of next hop */
         uint8_t weight = rtnl_route_nh_get_weight(nexthop);
-        if (weight)
+        if (weight == 0)
         {
-            result += to_string(weight);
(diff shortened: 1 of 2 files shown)


fpmsyncd/routesync.cpp
```diff
@@ -1300,14 +1300,15 @@ string RouteSync::getNextHopWt(struct rtnl_route *route_obj)
...
-            result += to_string(weight);
-        }
-        else
-        {
-            return "";
+            nl_addr* nh_addr = rtnl_route_nh_get_gateway(nexthop);
+            char nh_addr_str[MAX_ADDR_SIZE + 1] = {0};
+            nl_addr2str(nh_addr, nh_addr_str, MAX_ADDR_SIZE);
+            SWSS_LOG_INFO("Using default weight of 1 for nexthop %s", nh_addr_str);
+            weight = 1; // default weight is 1
```
prsunny, line 1309: Can you just keep this line and remove the above ones, we can do this for master since we have more bake time? @vaibhavhd for viz.

mramezani95: Removed the log.

prsunny (approved):

prsunny:
@bingwang-ms , please merge

PR #3649, opened 2025-05-15
[202411] Setting default nexthop weight to 1 in `fpmsyncd`

mramezani95:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


prsunny (approved):

prsunny:
@kperumalbfn , please merge this PR

prabhataravind (approved):
lgtm

PR #3652, opened 2025-05-19
[routeorch] Wait for the VRF to be created before route add

manoharan-nexthop:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/22642

**What I did**
With VRF configuration and route getting added in the startup, the orchangent crashes.

**Why I did it**
`addRoutePost()` assumes the VRF is already added and attempts to get the VRF ID for the same and used it to lookup the route entry. Since the VRF is not yet created, exception gets thrown.

When the VRF is not yet created, the route add can be postponed for the VRF to be created first (returning `false` in `RouteOrch::addRoutePost` keeps the entry in `consumer.m_toSync` and will be retried later).

**How I verified it**
Running the tests/vrf/test_vrf.py, the orchagent doesnt crash anymore and the route gets created in the ASIC

**Details if related**

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -2323,6 +2323,15 @@ bool RouteOrch::addRoutePost(const RouteBulkContext& ctx, const NextHopGroupKey 
         return false;
     }
 
+    // Ensure VRF exists in m_syncdRoutes
+    auto routeTableIter = m_syncdRoutes.find(vrf_id);
+    if (routeTableIter == m_syncdRoutes.end())
+    {
+        SWSS_LOG_INFO("VRF 0x%" PRIx64 " doesn't exist in syncd routes for route %s, will retry later",
+                      vrf_id, ipPrefix.to_string().c_str());
+        return false;
+    }
+
     if (m_fgNhgOrch->isRouteFineGrained(vrf_id, ipPrefix, nextHops))
     {
         /* Route is pointing to Fine Grained ECMP nexthop group */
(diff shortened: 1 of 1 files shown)


orchagent/routeorch.cpp
```diff
@@ -2383,11 +2392,11 @@ bool RouteOrch::addRoutePost(const RouteBulkContext& ctx, const NextHopGroupKey
     }
 
     auto it_status = object_statuses.begin();
-    auto it_route = m_syncdRoutes.at(vrf_id).find(ipPrefix);
+    auto it_route = routeTableIter->second.find(ipPrefix);
```
prsunny, line 2395: Prefer to keep the old code as-is

prsunny:
is this a transient issue or reproducible all the time. Seems to be a basic failure. My understanding is test will configure VRFs first and then add routes.

orchagent/routeorch.cpp
manoharan-nexthop: Thanks for the review. Since already a lookup done, felt there is no need for another lookup.

manoharan-nexthop:
> is this a transient issue or reproducible all the time. Seems to be a basic failure. My understanding is test will configure VRFs first and then add routes.

In a steady state, when the VRF is added and then the route, we dont see the issue. The issue is seen during reboot trigger and is reproducible most times.

prsunny (approved):

manoharan-nexthop:
@prsunny, the checks have passed, would it be possible to merge the PR. thanks.

PR #3655, opened 2025-05-21
[fpmsyncd]Fixing blackhole route to publish protocol field to APPL_DB

dgsudharsan:
**What I did**
Fixing blackhole route flow to publish protocol field to APPL_DB. When suppress fib pending is enabled, the protocol field is mandatory and not setting it results in fpmsyncd exiting in the route response path. 

**Why I did it**
To fix fpmsyncd from exiting on programming blackhole routes with suppress fib pending enabled.

**How I verified it**
Added UT to verify. Manually added blackhole route and verified as well.

**Details if related**

### DIFF
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -1636,6 +1636,9 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
     {
         sendOffloadReply(route_obj);
     }
+    auto proto_num = rtnl_route_get_protocol(route_obj);
+    auto proto_str = getProtocolString(proto_num);
+    FieldValueTuple proto("protocol", proto_str);
 
     switch (rtnl_route_get_type(route_obj))
     {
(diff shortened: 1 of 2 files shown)


vivekrnv (approved):

fpmsyncd/routesync.cpp
```diff
@@ -1983,13 +1980,18 @@ void RouteSync::onLabelRouteMsg(int nlmsg_type, struct nl_object *obj)
         return;
     }
```
BYGX-wcr, line 1982: Non-blocking comment: what does the label route refer to?

prsunny (dismissed):

fpmsyncd/routesync.cpp
dgsudharsan: Think it refers to mpls label. We enter it if family is AF_MPLS
    /* Check for Label route. */
    if (family == AF_MPLS)

prsunny (approved):

yejianquan:
@dgsudharsan could you fix the conflict and raise a new PR to 202505?

dgsudharsan:
> [yejianquan](/yejianquan)

Created it here https://github.com/sonic-net/sonic-swss/pull/3683

PR #3657, opened 2025-05-22
[fpmsyncd]Fixing fpmsyncd to handle routes without protocol

dgsudharsan:
**What I did**
When a route is programmed through swssconfig or through redis, there might be no protocol string. This would result in the response from orchagent to have emtpy protocol field. fpmsyncd would throw an exception which when handled through the try catch would exit fpmsyncd with exit code zero.

**Why I did it**
1) Handle empty protocol gracefully in fpmsyncd.
2) Remove the generic catch and exit 0 in fpmsyncd so that we will have a core dump and non zero exit on exceptions

**How I verified it**
Manually verified. Added UT to verify as well.

**Details if related**

### DIFF
--- a/fpmsyncd/fpmsyncd.cpp
+++ b/fpmsyncd/fpmsyncd.cpp
@@ -318,11 +318,6 @@ int main(int argc, char **argv)
         {
             cout << "Connection lost, reconnecting..." << endl;
         }
-        catch (const exception& e)
-        {
-            cout << "Exception \"" << e.what() << "\" had been thrown in daemon" << endl;
-            return 0;
-        }
     }
 
     return 1;
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -2529,6 +2529,14 @@ void RouteSync::onRouteResponse(const std::string& key, const std::vector<FieldV
         return;
     }
 
+    // When a route is programmed without FRR knowledge protocol will be empty.
+    if (protocol.empty())
+    {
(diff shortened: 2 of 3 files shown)


vivekrnv (approved):

fpmsyncd/routesync.cpp
```diff
@@ -2529,6 +2529,14 @@ void RouteSync::onRouteResponse(const std::string& key, const std::vector<FieldV
         return;
     }
 
+    // When a route is programmed without FRR knowledge protocol will be empty.
+    if (protocol.empty())
```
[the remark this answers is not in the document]
dgsudharsan: That's correct. This function onRouteResponse is called only in response path.

fpmsyncd/fpmsyncd.cpp
```diff
@@ -318,11 +318,6 @@ int main(int argc, char **argv)
         {
             cout << "Connection lost, reconnecting..." << endl;
         }
-        catch (const exception& e)
-        {
-            cout << "Exception \"" << e.what() << "\" had been thrown in daemon" << endl;
-            return 0;
```
prsunny, line 324: can we revert this change since protocol is now handled?

dgsudharsan: This is needed for better future debugging. With the catch statement when there is an exception we gracefully exit with 0 status and issues go unnoticed.
In order for better debugging it is important to have core generated. Hence I removed this catch statement

prsunny (approved):

dgsudharsan:
Created manual cherry-pick PR https://github.com/sonic-net/sonic-swss/pull/3690

PR #3658, opened 2025-05-22
Revert "Set Port UPDATE_DSCP attribute when TC_TO_DSCP map is attached"

kperumalbfn:
Reverts sonic-net/sonic-swss#3517

For some vendors this attribute is applicable at the ingress level, which should be evaluated by additional attribute in global metadata object. This enhancement will be taken later.


kperumalbfn:
Closing this PR for a new one.

PR #3661, opened 2025-05-27
Fix indentation

prabhataravind:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


prabhataravind:
@liushilongbuaa @yejianquan please add this for 202505 branch, otherwise pipeline runs will fail due to template issue

PR #3663, opened 2025-05-28
[portorch] Add capability to add unreliable LOS settings during a link down event

vdahiya12:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/port/porthlpr.cpp
+++ b/orchagent/port/porthlpr.cpp
@@ -411,9 +411,15 @@ bool PortHelper::parsePortUnreliableLos(PortConfig &port, const std::string &fie
         SWSS_LOG_ERROR("Failed to parse field(%s): invalid value(%s)", field.c_str(), value.c_str());
         return false;
     }
+    if (field == PORT_UNRELIABLE_LOS) {
+        port.serdes.unreliable_los.value = cit->second;
+        port.serdes.unreliable_los.is_set = true;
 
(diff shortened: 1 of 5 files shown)


prgeor (commented):
@vdahiya12 What will happen in a scenario where the unreliable los is already set and the link flaps?

prgeor:
@vdahiya12 in the PR description its not clear how is setting the field in port table:- apply_unreliable_los

orchagent/port/porthlpr.cpp
```diff
@@ -411,9 +411,15 @@ bool PortHelper::parsePortUnreliableLos(PortConfig &port, const std::string &fie
...
+    if (field == PORT_UNRELIABLE_LOS) {
+        port.serdes.unreliable_los.value = cit->second;
+        port.serdes.unreliable_los.is_set = true;
 
-    port.serdes.unreliable_los.value = cit->second;
-    port.serdes.unreliable_los.is_set = true;
+    } else if (field == PORT_APPLY_UNRELIABLE_LOS) {
+        port.apply_port_unreliable_los.value = cit->second;
+        port.apply_port_unreliable_los.is_set = true;
+        port.apply_port_unreliable_los.intf_type = SAI_PORT_INTERFACE_TYPE_SR4;
```
prgeor, line 421: @vdahiya12 looks to be a hack...can we set this outside OA? bcm config?

PR #3666, opened 2025-05-29
Revert "Set Port UPDATE_DSCP attribute when TC_TO_DSCP map is attache…

kperumalbfn:
This reverts commit 3a80d647a955cdd65160c81005b716804fa0d096.


stephenxs:
@r12f could you please take it into 202412? thanks.

PR #3667, opened 2025-05-29
[build] Support use other container regestry instead of DockerHub.

liushilongbuaa:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -376,7 +376,8 @@ class DockerVirtualSwitch:
 
         # Dynamically create a DVS container and servers
         else:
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -376,7 +376,8 @@ def __init__(
 
         # Dynamically create a DVS container and servers
         else:
-            self.ctn_sw = self.client.containers.run("debian:jessie",
+            cr_prefix = os.environ['DEFAULT_CONTAINER_REGISTRY'].rstrip("/")
```
saiarcot895, line 379: If this environment variable isn't set, then this will result in a KeyError.

PR #3668, opened 2025-05-29
[build] Support optionally using other container registries instead of DockerHub

saiarcot895:
Includes #3667

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


prabhataravind:
@liushilongbuaa @yejianquan please include this PR in 202505. @prsunny for viz.

PR #3690, opened 2025-06-06
[202505][fpmsyncd]Fixing fpmsyncd to handle routes without protocol

dgsudharsan:
Cherry-pick of PR https://github.com/sonic-net/sonic-swss/pull/3657
**What I did**
When a route is programmed through swssconfig or through redis, there might be no protocol string. This would result in the response from orchagent to have emtpy protocol field. fpmsyncd would throw an exception which when handled through the try catch would exit fpmsyncd with exit code zero.

**Why I did it**
Handle empty protocol gracefully in fpmsyncd.
Remove the generic catch and exit 0 in fpmsyncd so that we will have a core dump and non zero exit on exceptions

**How I verified it**
Manually verified. Added UT to verify as well.

**Details if related**

### DIFF
--- a/fpmsyncd/fpmsyncd.cpp
+++ b/fpmsyncd/fpmsyncd.cpp
@@ -318,11 +318,6 @@ int main(int argc, char **argv)
         {
             cout << "Connection lost, reconnecting..." << endl;
         }
-        catch (const exception& e)
-        {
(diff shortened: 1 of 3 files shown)


prsunny:
@yejianquan , please merge to 202505

PR #3698, opened 2025-06-09
[Portsorch[ Recreate any port(s) if the configured speed does not match

prgeor:
**What I did**
Delete the port if the speed mismatch with that in config DB and recreate it with the speed as per config DB

**Why I did it**

There could be scenarios where a platform creates default port with a speed/lane configuration other than in the config DB. If so, OA should delete those port during configuration ingestion and create them as per the speed specified in the 

**How I verified it**

Verified on a switch where the default speed is 400G(x4) and changed the speed in config DB to 100G(x4) and ensured the OA does not crash and the link is UP at 100G NRZ speed

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -495,7 +495,7 @@ static bool isPathTracingSupported()
         SWSS_LOG_INFO("Querying OBJECT_TYPE_LIST is not supported on this platform");
         return false;
     }
-    else 
+    else
     {
         SWSS_LOG_ERROR(
             "Failed to get a list of supported switch capabilities. Error=%d", status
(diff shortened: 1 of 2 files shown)


orchagent/portsorch.cpp
```diff
@@ -3784,7 +3784,7 @@ bool PortsOrch::initPort(const PortConfig &port)
                     port_buffer_drop_stat_manager.setCounterIdList(p.m_port_id, CounterType::PORT, port_buffer_drop_stats);
                 }
 
-		if (flex_counters_orch->getWredPortCountersState())
+		        if (flex_counters_orch->getWredPortCountersState())
```
nazariig: @prgeor is this a typo?

nazariig (changes requested):

orchagent/portsorch.cpp
```diff
@@ -4186,6 +4186,25 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+                                    it->second, 
+                                    pCfg.speed.value, 
+                                    speed,
+                                    swss::join(" ", it->first.cbegin(), it->first.cend()).c_str()
+                            );
+                            portsToRemoveList.push_back(it->second);
+                            it = m_portListLaneMap.erase(it);
+                            continue;
+                        }
+                    }
```
nazariig, line 4228: @prgeor port recreation due to invalid speed looks like an overhead. What's wrong with the speed configuration which takes place after port initialization? IMHO this looks like a W/A

nazariig (changes requested):

orchagent/portsorch.cpp
```diff
@@ -4186,6 +4186,25 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         it = m_portListLaneMap.erase(it);
                         continue;
                     }
+                    else
```
bingwang-ms, line 4189: If the ports are created by SDK and portorch doesn't have the info, will this still work?

prgeor: @bingwang-ms config DB should have all the port related information.

orchagent/portsorch.cpp
prgeor: @nazariig nope....space alignment to match the if block

orchagent/portsorch.cpp
prgeor: @nazariig there are platforms that have default ports created during the switch create at OA start ...Now when the config db port breakout configuration is pushed, SAI will change the speed one by one i.e set the speed for first subport 400G to 100G, the second subport is still 400G. So at some point you will have one subport at 100G and the other subport at 400G. This is not supported by HW and causes OA to crash. Due to hardware limitation, the port cannot operate in speed mismatch
subport 1 in 100G NRZ mode
subport 2 in 400G PAM4 mode

So, here OA is removing all the ports which don't match the speed set in the config DB and then recreate them with speed as per config DB. This does not impact platforms:-

1. Dont create switch ports at switch create itself. Eg Nvidia switches
2. Platforms where the config DB ports configuration speed matches with the default port speed

stepanblyschak (approved):

orchagent/portsorch.cpp
bingwang-ms: Can this logic handle incremental change?  For example, if we use config-apply patch to update speed for sub port one by one, then we may run into the issue again.
My understanding is to find out all the sub ports in one port group and update speed for all sub ports.

orchagent/portsorch.cpp
prsunny: Please revert this. Its not aligned.

orchagent/portsorch.cpp
theasianpianist, line 4228: suggest wrapping this code in a check to skip running on DPU devices

orchagent/portsorch.cpp
prgeor: reverted

orchagent/portsorch.cpp
```diff
@@ -4186,6 +4186,46 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+                                        pCfg.speed.value,
+                                        speed,
+                                        swss::join(" ", it->first.cbegin(), it->first.cend()).c_str()
+                                );
+                                portsToRemoveList.push_back(it->second);
+                                it = m_portListLaneMap.erase(it);
+                                continue;
+                            }
+                        }
+                    }
```
theasianpianist, line 4228: suggest wrapping this code in a check to skip running on DPU devices

PR #3699, opened 2025-06-09
[orchagent] Event-based Retry Strategy

a114j0y:
**What I did**
Implemented `RetryCache`

**Why I did it**
The original orchagent performs a brute-force retry strategy for all the previously failed tasks in its cache, which has the risk of blocking the normal workflow when the accumulated failures are scaled, since the event-handling loop and the retry logic is in the same thread. 

After observed the delay of normal event handling by previously accumulated failed tasks, we dived into the retry strategy and came to several conclusions. Firstly, not all failures deserve an immediate retry, since its constraint may have not been resolved yet. Secondly, it's worth analyzing the failure reason, and record it together with the failed task, as its constraint. Thirdly, we need a in-time notification mechanism to detect when the constraint is resolved and notify the related tasks to trigger their retry. Lastly, we should make sure the retry stage would not starve the normal event workflow, it can't last too long.
That means, we build up a event-based retry logic, instead of blindly repeating all the pending tasks in the cache in the main thread. The key is to skip unnecessary failure retries when you know they would fail again, and come up with these skipped failures in-time when they have a chance to succeed, instead of forgetting them forever.

**How I verified it**

**Details if related**


prabhataravind:
hi @a114j0y is there a design document for the changes implemented in this PR?

a114j0y:
> hi @a114j0y is there a design document for the changes implemented in this PR?

https://github.com/a114j0y/SONiC/blob/retry/doc/swss_retry_cache/retry_optimization.md

orchagent/orch.cpp
```diff
@@ -147,15 +147,102 @@ vector<Selectable *> Orch::getSelectables()
...
+
+        getConsumerBase(executorName)->addToSync(tasks, true);
+
+    }
+    return count;
+}
+
+void Orch::notifyRetry(Orch *retryOrch, const std::string &executorName, const Constraint &cst)
+{
+    retryOrch->getRetryCache(executorName)->add_resolution(cst);
```
selva-nexthop: might crash if getRetryCache returns nullptr

orchagent/orch.cpp
```diff
@@ -147,15 +147,102 @@ vector<Selectable *> Orch::getSelectables()
...
+
+    std::unordered_set<Constraint>& constraints = retryCache->getResolvedConstraints();
+
+    size_t count = 0;
+
+    for (auto it = constraints.begin(); it != constraints.end() && count < threshold;)
+    {
+        auto cst = *it++;
+
+        auto tasks = retryCache->resolve(cst, threshold - count);
```
selva-nexthop: resolve() calls m_resolvedConstraints.erase(cst) which can invalidate the iterator.
Current code is safe because iterator is advanced before calling resolve(), but this is fragile.

we can consider something like below;
```
while (!constraints.empty() && count < threshold) {
    auto cst = *constraints.begin();
    auto tasks = retryCache->resolve(cst, threshold - count);
    count += tasks->size();
    getConsumerBase(executorName)->addToSync(tasks, true);
}
```

orchagent/orch.cpp
```diff
@@ -147,15 +147,102 @@ vector<Selectable *> Orch::getSelectables()
...
+        {
+            if (it->second.second == entry) // skip duplicate task
+                return;
+            
+            auto cache = retryCache->erase_stale_cache(key);
+
+            Recorder::Instance().retry.record(dumpTuple(*cache).append(DECACHE));
+
+            if (op == SET_COMMAND)
+                m_toSync.emplace(key, std::move(*cache));
```
selva-nexthop: - might insert empty task.   
- the behaviour might be unexpected if  kfvKey(*cache) returns empty string
-  might crash as well

please consider to protect it like below;
```
auto cache = retryCache->erase_stale_cache(key);
if (op == SET_COMMAND && !kfvKey(*cache).empty())
    m_toSync.emplace(key, std::move(*cache));
```

orchagent/retrycache.h
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    /** Insert a failed task with its constraint to m_toRetry and m_retryKeys
+     * @param task the task that has failed
+     * @param cst constraint needs to be resolved for the task to succeed
+     */
+    void cache_failed_task(const Task &task, const Constraint &cst) {
+        const auto& key = kfvKey(task);
+        m_retryKeys[cst].insert(key);
+        m_toRetry.emplace(
```
selva-nexthop, line 103: m_toSync is actually a multimap.  if multiple operations exist for the same key in m_toSync, how it would be handled here?

orchagent/routeorch.cpp
```diff
@@ -1126,10 +1133,21 @@ void RouteOrch::doTask(ConsumerBase& consumer)
...
+            // - delete it from m_toSync after addRoutePost/removeRoutePost to avoid duplicates
+            if (ctx.retry_cst != DUMMY_CONSTRAINT)
+                consumer.addToRetry(it_prev->second, ctx.retry_cst);
+
             const auto& object_statuses = ctx.object_statuses;
             if (object_statuses.empty())
             {
-                it_prev++;
+                if (ctx.retry_cst != DUMMY_CONSTRAINT)
+                    it_prev = consumer.m_toSync.erase(it_prev);
```
selva-nexthop, line 1152: - tasks are added to retry cache BEFORE processing, then removed from m_toSync regardless of success.
- successful tasks remain in retry cache: If addRoutePost() succeeds but retry_cst != DUMMY_CONSTRAINT, task stays in retry cache unnecessarily
- Double processing: Task gets processed immediately AND cached for retry
- Inconsistent state: Task exists in both systems simultaneously

orchagent/nhgorch.cpp
```diff
@@ -266,6 +267,7 @@ void NhgOrch::doTask(Consumer& consumer)
                         if (nhg->sync())
                         {
                             m_syncdNextHopGroups.emplace(index, NhgEntry<NextHopGroup>(std::move(nhg)));
+                            notifyRetry(gRouteOrch, APP_ROUTE_TABLE_NAME, make_constraint(RETRY_CST_NHG, index));
```
selva-nexthop: how we handle the timing issue here? like route processing happens in the same iteration?  
e.g. 
- route fails due to missing NHG and add it to retry_cst
- nhg gets created and sends notification (with this logic)
- now, route tasks gets added to retry (after the previous notification)

We need to make sure that the route task never gets retried because the constraint is already marked as resolved.  This might lead to permanent task loss.

orchagent/orch.cpp
```diff
@@ -663,8 +746,14 @@ string Orch::objectReferenceInfo(
 
 void Orch::doTask()
 {
+
+    auto threshold = gBatchSize == 0 ? 30000 : gBatchSize;
```
selva-nexthop, line 840: how we are defining this threshold (also, it is hardcoded)?.  Most importantly, it seems to be a shared budget for all consumers in the m_consumerMap.  If one consumer consumes all the numbers, then tasks for other consumers will be lost.

Did we try with scaled number of tests like;
- more than 1,00,000 routes and NHG is not resolved
- then NHG is resolved

we need to measure who are all consumers in this case and how many retry tasks created by each consumer and how this threshold is being shared across all these consumers.   If we miss any task to be added in this retry cache, then it will lead to permanent traffic loss.

selva-nexthop: General comments and Testing
- Monitor memory usage impact in production environments
- Document the constraint system for future developers
- Unit Tests: Comprehensive test suite for RetryCache
- Stress Tests: High-load scenarios with many failures
- Edge Case Tests: Timing issues, memory limits, error conditions

orchagent/orch.cpp
a114j0y: It should not be an empty task here, since only a valid non-empty task would have a chance to be inserted into the retry cache.  But you are right,  it would be better to explicitly add one line to make sure only non-empty key could be cached in the function below, 
```
void cache_failed_task(const Task &task, const Constraint &cst) {
        const auto& key = kfvKey(task);
        if (key.empty())
            return;
...
```

But in orch.cpp,  if key is empty,
```
auto it = retryCache->getRetryMap().find(key);
```
would not meet the condition
```
if (it != retryCache->getRetryMap().end()) 
```
and even if key is empty ( (which equals to kfvKey(*cache))), emplace would not crash, since an empty string is also a valid string.

orchagent/retrycache.h
a114j0y: Thx for pointing out, i changed RetryMap to be multimap too !

a114j0y: modified! use iterator instead of key, 
```
m_toRetry.erase(it);
```

orchagent/routeorch.cpp
a114j0y: tasks are added to retry cache AFTER processing in addRoute function (not addRoutePost)
all tasks that encounter constraints would get its context `ctx` marked, and recording constraint data in field `ctx.retry_cst`, we didn't want to stop the workflow immediately after inserting the task into retrycache, to keep the logic same as before.
it's not removed from m_toSync regardless of success, if the ctx.retry_cst != DUMMY_CONSTRAINT, it means there is some failure detected.

when ctx.retry_cst is checked, addRoutePost hasn't been called.
```
ctx.retry_cst != DUMMY_CONSTRAINT
```

orchagent/nhgorch.cpp
a114j0y: `notifyRetry` would only add the resolved constraint into `m_resolvedConstraints`,  which is an unordered_set, indicating the TODO tasks for the retry cache.  Afterwards, the retry cache would check `m_resolvedConstraints` and start moving the failed tasks back to m_toSync related to the resolved constraint. Only if all affected tasks are moved out of retry cache, would the constraint deleted from m_resolvedConstraints (this avoids task loss).  If it's not deleted, retrycache would always know there is still affected tasks needs moving.

orchagent/orch.cpp
a114j0y: add this threshold just to make sure the retry stage would not starve the regular workflow

orchagent/main.cpp
```diff
@@ -363,10 +365,11 @@ int main(int argc, char **argv)
     string record_location = Recorder::DEFAULT_DIR;
     string swss_rec_filename = Recorder::SWSS_FNAME;
     string sairedis_rec_filename = Recorder::SAIREDIS_FNAME;
+    string retry_rec_filename = Recorder::RETRY_FNAME;
     string zmq_server_address = "";
     string vrf;
     string responsepublisher_rec_filename = Recorder::RESPPUB_FNAME;
-    int record_type = 3; // Only swss and sairedis recordings enabled by default.
+    int record_type = 11; // Only swss, retrycache and sairedis recordings enabled by default.
```
liuh-80: Suggestion set record_type with macros and bit operations:
int record_type = SAIREDIS_RECORD_ENABLE | SWSS_RECORD_ENABLE | RETRY_RECORD_ENABLE; #Closed

eddieruan-alibaba (approved):

eddieruan-alibaba:
This change has been used in Alibaba and Joy presented the result in this year's OCP SONiC workshop.

selva-nexthop (approved):

orchagent/retrycache.h
```diff
@@ -0,0 +1,213 @@
...
+    RETRY_CST_DUMMY,
+    RETRY_CST_NHG,          // nhg doesn't exist
+    RETRY_CST_NHG_REF,      // nhg refcnt nonzero
+    RETRY_CST_PIC,          // context doesn't exist
+    RETRY_CST_PIC_REF,      // context refcnt nonzero
+    RETRY_CST_ECMP          // ecmp resources exhausted
+};
+
+
+inline std::ostream& operator<<(std::ostream& os, ConstraintType t) {
```
qiluo-msft: >inline [](http://example.com/codeflow?start=0&length=6)

inline -> static inline
ref: [https://stackoverflow.com/a/47821267](https://stackoverflow.com/a/47821267) #Closed

orchagent/retrycache.h
```diff
@@ -0,0 +1,213 @@
...
+        default:           return os << "UNKNOWN";
+    }
+}
+
+using ConstraintData = std::string;
+using Constraint = std::pair<ConstraintType, ConstraintData>;
+
+const Constraint DUMMY_CONSTRAINT{RETRY_CST_DUMMY, ""};
+
+inline Constraint make_constraint(ConstraintType type, ConstraintData data = "") {
```
qiluo-msft: >inline [](http://example.com/codeflow?start=0&length=6)

the same #Closed

qiluo-msft:
Please minimize the force-push in future. We normally check between commits to understand new change, force-push will make code review iteration difficult.

a114j0y:
> Please minimize the force-push in future. We normally check between commits to understand new change, force-push will make code review iteration difficult.

Ok! But it seems that an update with rebase would trigger force-push automatically

prsunny:
@prabhataravind , can you review and signoff, thanks @a114j0y , looks to be an efficient strategy.

a114j0y:
> @prabhataravind , can you review and signoff, thanks @a114j0y , looks to be an efficient strategy.

Hi @prabhataravind could you help merge it? So we could include it in the 202511 release !

prabhataravind:
@a114j0y have you tested this with warm-boot? @vaibhavhd for viz.

orchagent/orch.cpp
```diff
@@ -663,8 +812,14 @@ string Orch::objectReferenceInfo(
...
+
+    size_t count = 0;
+
     for (auto &it : m_consumerMap)
     {
+        count += retryToSync(it.first, threshold - count);
```
prabhataravind, line 846: Please add a comment here.

orchagent/orch.cpp
```diff
@@ -147,15 +147,168 @@ vector<Selectable *> Orch::getSelectables()
...
+    if (retryCache)
+    {
+        Recorder::Instance().retry.record(getConsumerBase(executorName)->dumpTuple(task).append(CACHE));
+        retryCache->insert(task, cst);
+        return true;
+    }
+    return false;
+}
+
+size_t Orch::retryToSync(const std::string &executorName, size_t threshold)
```
prabhataravind: Please add documentation for this function explaining what threshold is.

orchagent/routeorch.cpp
```diff
@@ -1153,7 +1171,7 @@ void RouteOrch::doTask(ConsumerBase& consumer)
                 {
                     /* If any existing routes are updated to point to the
                      * above interfaces, remove them from the ASIC. */
-                    if (removeRoutePost(ctx))
+                     if (removeRoutePost(ctx) || rc_inserted)
```
prabhataravind: Fix stray space

orchagent/routeorch.cpp
```diff
@@ -1164,7 +1182,7 @@ void RouteOrch::doTask(ConsumerBase& consumer)
 
                 if (nhg.getSize() == 1 && nhg.hasIntfNextHop())
                 {
-                    if (addRoutePost(ctx, nhg))
+                     if (addRoutePost(ctx, nhg) || rc_inserted)
```
prabhataravind: Stray space

orchagent/orch.cpp
```diff
@@ -147,15 +147,168 @@ vector<Selectable *> Orch::getSelectables()
...
+
+        count += tasks->size();
+
+        getConsumerBase(executorName)->addToSync(tasks, true);
+
+    }
+    return count;
+}
+
+void Orch::notifyRetry(Orch *retryOrch, const std::string &executorName, const Constraint &cst)
```
prabhataravind, line 226: @a114j0y who is calling this function for other constraints like RETRY_CST_NHG, RETRY_CST_NHG_REF, RETRY_CST_ECMP etc?

a114j0y: These codes would be in a separate NHG-related PR, which employs RetryCache for optimization. I can move these definitions to that pr.

a114j0y:
> Azure Pipelines successfully started running 1 pipeline(s).
To pass DCO check: 
```
git rebase HEAD~11 --signoff
push --force-with-lease origin master
```

orchagent/orch.cpp
prabhataravind: This is fine. You can link that PR to this when you raise it.

prabhataravind:
> @a114j0y have you tested this with warm-boot? @vaibhavhd for viz.

@a114j0y can you confirm this? At a bare minimum, we shouldn't be seeing crashes when upgrading from an image with this change to another image when there are entries in retry cache.

a114j0y:
> > @a114j0y have you tested this with warm-boot? @vaibhavhd for viz.
> 
> @a114j0y can you confirm this? At a bare minimum, we shouldn't be seeing crashes when upgrading from an image with this change to another image when there are entries in retry cache.

Thx for the comment!

Add a commit to modify ConsumerBase::dumpPendingTasks(vector<string> &ts), so that it not only dumps m_toSync entries but also m_toRetry entries, which could validate for warm-reboot that no pending entries exist in either m_toSync or retry cache.

It's tested by Alibaba in warm-reboot scenario.

prsunny:
@vaibhavhd for viz. @prabhataravind , if you are good with comment, would you approve?

PR #3705, opened 2025-06-12
[trim]: Add Packet Trimming Asym DSCP to OA

nazariig:
Signed-off-by: Nazarii Hnydyn <nazariig@nvidia.com>

**DEPENDS:**
1. https://github.com/sonic-net/sonic-sairedis/pull/1610
2. https://github.com/sonic-net/sonic-swss/pull/3594

**HLD:** https://github.com/sonic-net/SONiC/pull/1988

**What I did**
* Implemented Packet Trimming Asymmetric DSCP feature

**Why I did it**
* Implementation is done according to the Packet Trimming Asymmetric DSCP HLD

**How I verified it**
1. Run Packet Trimming VS UTs

**Details if related**
* N/A

#### A picture of a cute animal (not mandatory but encouraged)
```
      .---.        .-----------
     /     \  __  /    ------
    / /     \(  )/    -----
   //////   ' \/ `   ---
  //// / // :    : ---
 // /   /  /`    '--
//          //..\\
       ====UU====UU====
           '//||\\`
             ''``
```


kperumalbfn:
@nazariig could you resolve the conflicts

orchagent/buffer/buffercontainer.h
```diff
@@ -0,0 +1,71 @@
...
+{
+public:
+    BufferProfileConfig() = default;
+    ~BufferProfileConfig() = default;
+
+    inline bool isTrimmingProhibited() const
+    {
+        return ((pgRefCount > 0) || (iBufProfListRefCount > 0) || (eBufProfListRefCount)) ? true : false;
+    }
+
```
kperumalbfn, line 29: we could keep all the counts as private and have methods to set/get. Would be clean to set/get by one method instead of direct access

orchagent/buffer/bufferhelper.cpp
```diff
@@ -0,0 +1,303 @@
...
+        cfg.profile.value = cit->second;
+        cfg.profile.is_set = true;
+    }
+}
+
+void BufferHelper::parseBufferConfig(IngressBufferProfileListConfig &cfg) const
+{
+    auto &map = cfg.fieldValueMap;
+
+    const auto &cit = map.find(BUFFER_PORT_INGRESS_PROFILE_LIST_PROFILE_LIST);
```
kperumalbfn, line 43: Does bufferorch use ingress/egress buffer profile list for trim eligible validation?

orchagent/buffer/buffercontainer.h
kperumalbfn: Currently not all egress buffer profiles are added to profile list. Could you please check?

https://github.com/sonic-net/sonic-buildimage/pull/22869/files

orchagent/buffer/buffercontainer.h
```diff
@@ -0,0 +1,71 @@
...
+class BufferProfileConfig final : public BufferContainer
+{
+public:
+    BufferProfileConfig() = default;
+    ~BufferProfileConfig() = default;
+
+    inline bool isTrimmingProhibited() const
+    {
+        return ((pgRefCount > 0) || (iBufProfListRefCount > 0) || (eBufProfListRefCount)) ? true : false;
```
kperumalbfn, line 27: Can we avoid this refcount to determine the trim eligibility?

kperumalbfn: if possible, buffer_profile-->buffer_pool(check ingress/egress)?

kperumalbfn: @stephenxs any other option?

orchagent/switchorch.cpp
```diff
@@ -981,24 +1010,104 @@ bool SwitchOrch::setSwitchTrimming(const SwitchTrimming &trim)
...
+                    SWSS_LOG_ERROR("Failed to validate switch trimming TC value: capability is not supported");
+                    return false;
+                }
+
+                if (!setSwitchTrimmingTcSai(trim))
+                {
+                    SWSS_LOG_ERROR("Failed to set switch trimming TC value in SAI");
+                    return false;
+                }
+
```
kperumalbfn, line 1089: nit: empty new lines, pls check

orchagent/switchorch.cpp
```diff
@@ -981,24 +1010,104 @@ bool SwitchOrch::setSwitchTrimming(const SwitchTrimming &trim)
...
+                tcSync = true;
+            }
+            else
+            {
+                SWSS_LOG_WARN("Skip setting switch trimming TC value for symmetric DSCP mode");
```
kperumalbfn, line 1094: return from here?

orchagent/buffer/buffercontainer.h
nazariig: > we could keep all the counts as private and have methods to set/get. Would be clean to set/get by one method instead of direct access

@kperumalbfn Getter/Setter is a way of doing business logic encapsulation. In this case it is redundant and won't provide any benefits. That was the reason for omitting the implementation

nazariig: > Currently not all egress buffer profiles are added to profile list. Could you please check?
> 
> https://github.com/sonic-net/sonic-buildimage/pull/22869/files

The current implementation relies on application db events, meaning all buffer profiles are processed in a single place - `bufferorch`. Please elaborate on how this relates to the current validation case ?

orchagent/buffer/bufferhelper.cpp
nazariig: @kperumalbfn validation is currently implemented for the next cases:

1. Buffer Profile is attached to either Ingress PG or Ingress/Egress Buffer Profile List and user tries to set trimming eligibility
2. Buffer Profile is configured as trimming eligible and user tries to attach it either to Ingress PG or Ingress/Egress Buffer Profile list

In both cases the error will be generated.
This guarantees that trimming eligible Buffer Profile can be used only with the Queue objects.

orchagent/buffer/buffercontainer.h
nazariig: > Can we avoid this refcount to determine the trim eligibility?

@kperumalbfn currently the most reliable way. Checking counter value is a low cost operation comparing to collection iteration/search

nazariig: > if possible, buffer_profile-->buffer_pool(check ingress/egress)?

@kperumalbfn Buffer Profile to Buffer Pool relation check (`resolveFieldRefArray`) is implemented as part of:
`task_process_status BufferOrch::processPriorityGroup(KeyOpFieldsValuesTuple &tuple)`
`task_process_status BufferOrch::processIngressBufferProfileList(KeyOpFieldsValuesTuple &tuple)`
`task_process_status BufferOrch::processEgressBufferProfileList(KeyOpFieldsValuesTuple &tuple)`

Trimming eligibility validation serves different purposes and takes place long after this check, so it is not relevant

orchagent/switchorch.cpp
nazariig: @kperumalbfn new line is added after if statement. No extra spaces are observed. Please elaborate on the issue

orchagent/switchorch.cpp
nazariig: @kperumalbfn the implementation is by design. If the user tries to set TC value while operating in symmetric DSCP mode, SWSS will save the new value to cache and print the warning message. Later on if the mode will be switched to asymmetric, the cached value will be set to SAI.

The warning message is printed only once - in case if a new (differs from the value stored in the cache) TC value is configured.
That helps to maintain the minimal informative level while avoiding the log pollution (in case user/controller sets the same value for multiple times).

nazariig:
> @nazariig could you resolve the conflicts

@kperumalbfn done

kperumalbfn (approved):
LGTM

kperumalbfn:
@nazariig Please rebase the branch

orchagent/buffer/buffercontainer.h
stephenxs: Why not using the pools that the buffer profiles reference? That will be much easier to maintain.

orchagent/bufferorch.cpp
```diff
@@ -741,6 +748,18 @@ task_process_status BufferOrch::processBufferProfile(KeyOpFieldsValuesTuple &tup
                 continue;
             }
         }
+
+        m_bufHlpr.parseBufferConfig(cfg);
+
+        if (cfg.isTrimmingEligible && cfg.isTrimmingProhibited())
```
stephenxs, line 754: `cfg.isTrimmingProhibited()` can be `false` if the profile is ingress but has not been referenced by any PG, which could happen as OA does not guarantee the order. To avoid it we have to add logic in the function to handle buffer PGs, and profile lists.
However, the issue can be resolved if it is tested using the reference to the buffer pool because the buffer profile can be handled only after the buffer pool is ready (please check the logic to handle `buffer_pool_field_name`).
By doing so we do not change the logic in the function to handle buffer PGs, and profile lists.

orchagent/bufferorch.cpp
```diff
@@ -741,6 +748,18 @@ task_process_status BufferOrch::processBufferProfile(KeyOpFieldsValuesTuple &tup
                 continue;
             }
         }
+
+        m_bufHlpr.parseBufferConfig(cfg);
```
stephenxs, line 752: What is the purpose of calling `parseBufferConfig`? From what I understood it's to parse whether the packet discard operation is `trim` or `discard` right?
It has been done when `SAI_BUFFER_PROFILE_ATTR_PACKET_ADMISSION_FAIL_ACTION` is translated from `BUFFER_PROFILE_PACKET_DISCARD_ACTION`

volodymyrsamotiy (approved):

prsunny (approved):
approving for merge. Code reviewed by SMEs

r12f:
hi @nazariig , do you mind to help create a manual cherry pick to 202412 for this PR?

dgsudharsan:
@r12f  Here is the PR https://github.com/Azure/sonic-swss.msft/pull/105

PR #3713, opened 2025-06-17
[routeorch] Handle SAI_STATUS_ITEM_NOT_FOUND when setting route entries

Ndancejic:
In dualtor systems, muxorch will program tunnel route entries for mux neighbors on standby ports. If a route with the same prefix as the mux tunnel route is learned, a cache entry will be created in routeorch. Then when the neighbor is moved to an active mux port, the tunnel route will be deleted.

In certain race-conditions the delete notification will be delayed, causing a mismatch in routeoch between the internal cache and syncd state. If a route update is processed during this time, routeorch will attempt to set the route entry in syncd. This will return an error code SAI_STATUS_ITEM_NOT_FOUND.

This adds error handling logic to remove the internal cache entry from routeorch, then attempt to retry the route set.

**What I did**
- Added error handling logic for SAI_STATUS_ITEM_NOT_FOUND in routeorch when performing a set operation on the route.
- Added unit test to verity the error handling logic performs as expected

**Why I did it**
A rare race condition could cause cache mismatch with the asic state. This caused routeorch to attemt to set a route entry that did not exist, leading to error code SAI_STATUS_ITEM_NOT_FOUND. This change adds error handling logic to remove the internal cache entry from routeorch, then attempt to retry the route set.

**How I verified it**
Added unit test that covers this path

**Details if related**
ado: #32780823

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -2495,6 +2495,14 @@ bool RouteOrch::addRoutePost(const RouteBulkContext& ctx, const NextHopGroupKey 
         status = *it_status++;
         if (status != SAI_STATUS_SUCCESS)
         {
+            if (status == SAI_STATUS_ITEM_NOT_FOUND)
+            {
+                // Routeorch internal cache has an entry, but it has already been removed in sai.
+                // This can happen in dualtor when a tunnel route is removed that matches a learned route
+                // remove the entry from the cache and retry route creation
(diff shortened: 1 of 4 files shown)


prabhataravind (approved):
lgtm

zjswhhh (approved):
lgtm

prsunny (approved):

PR #3719, opened 2025-06-20
[chassis] : LC reboot causing buffer_profile related errors

vmittal-msft:
Issue : https://github.com/sonic-net/sonic-buildimage/issues/20070

**What I did**
Updated processQueue and handleQueueTable function to do case insensitive comparison for asic name.

**Why I did it**
Since asic_name of the namespace could be lowercase or uppercase for chassis, this can cause comparison failure for local ports.

**How I verified it**
Verified on sonic chassis

**Details if related**

### DIFF
--- a/orchagent/bufferorch.cpp
+++ b/orchagent/bufferorch.cpp
@@ -8,6 +8,7 @@
 #include <inttypes.h>
 #include <sstream>
 #include <iostream>
+#include <boost/algorithm/string.hpp>
 
 #include "buffer/bufferschema.h"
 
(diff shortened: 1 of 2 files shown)


vmittal-msft:
@prsunny please help review.

prsunny (approved):

prsunny:
@vmittal-msft , possible to fix coverage?

vmittal-msft:
> @vmittal-msft , possible to fix coverage?

@prsunny  It's hard to cover this code. Can we push ?

prsunny:
Bypassing coverage since its 'lower-case' check

PR #3722, opened 2025-06-23
MUX nbrs to use no-host-route & use nh update

manamand2020:
**What I did**
MUX neighbors will not use host route and full prefix route is explicitly created.
During mux state switchover only update the nexthop in these full prefix route of the neighbors.

**Why I did it**
This avoids multiple SAI calls done during MUX state switchover and reduces the switchover time.

**How I verified it**
Verified on hardware and VS Tests.

**Details if related**


prsunny:
@manamand2020 , would you also run the sonic-mgmt dualtor io tests with this change and paste the results?

orchagent/muxorch.cpp
```diff
@@ -1359,6 +1443,44 @@ MuxCable* MuxOrch::findMuxCableInSubnet(IpAddress ip)
...
+
+    NextHopKey nh_key = NextHopKey(nbr, alias);
+    string curr_port = getNexthopMuxName(nh_key);
+    if (port.empty() && !curr_port.empty() && isMuxExists(curr_port))
+    {
+        return true;
+    }
+
+    // Check cached neighbors during warm boot
+    return isCachedMuxNeighbor(nbr, alias);
```
Ndancejic: Is there a way we can move this check to only happen once at the beginning of this function? if I understand this correctly this should only be during warmboot. If this evalutates as true is there a reason we need to run the other checks?

orchagent/muxorch.cpp
```diff
@@ -1533,7 +1667,8 @@ void MuxOrch::updateNeighbor(const NeighborUpdate& update)
         if (port.empty() || old_port == port)
         {
             addNexthop(update.entry, old_port);
-            return;
+            is_mux_neighbor = true;
+            goto handle_redis;
```
Ndancejic: While I don't personally mind using goto, we don't generally have goto statements in most orchagent code. Also I think this chunk of code is small enough to tolerate a couple of repeated lines and avoid goto in this case.
pref just having the following here:
```
if (update.add)
{
    saveNeighborToMuxTable(update.entry.ip_address, alias);
}
else
{
    removeNeighborFromMuxTable(update.entry.ip_address, alias);
}
return;
```

orchagent/muxorch.cpp
```diff
@@ -1927,6 +2089,106 @@ void MuxOrch::updateCachedNeighbors()
     }
 }
 
+void MuxOrch::saveMuxNeighbors()
```
Ndancejic: I'm not seeing this called anywhere, Is it currently used? If not, can you remove from this PR (as well as any supporting code) and it can be re-evaulated at a later point w/warmboot.

manamand2020: Removed.

orchagent/muxorch.cpp
manamand2020: Done.

orchagent/muxorch.cpp
manamand2020: Yes, done.

zjswhhh:
What's the plan on this? Do we need to backport it to 202505 or not?

manamand2020:
> What's the plan on this? Do we need to backport it to 202505 or not?
@Ndancejic can you approve this PR and can @prsunny help in merging this? We will need this change as well along with this:
https://github.com/sonic-net/sonic-utilities/pull/3931

zjswhhh (dismissed):

zjswhhh:

prsunny:
This PR cannot be merged as-is, the following are the gaps:
1. Missing link to HLD and Testplan
2. Too many changes that can impact backward compatibility
3. Support removed for bulk nexthops and neighbors which can be used for other usecases
4. Lack of capability check

prsunny:
@manamand2020 , @Ndancejic , closing this PR. let me know if this needs to be revisited.

PR #3723, opened 2025-06-23
Harden module build script by specifying the source version to get

saiarcot895:
**What I did**

Harden the module build script by specifying the source version of the kernel to get. This makes sure that either the correct version of the kernel module is built, or that the download itself fails, rather than causing a potential kernel panic.

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/fpmsyncd/Makefile.am
+++ b/fpmsyncd/Makefile.am
@@ -1,4 +1,4 @@
-INCLUDES = -I $(top_srcdir) -I $(top_srcdir)/warmrestart -I $(FPM_PATH)
+INCLUDES = -I $(top_srcdir) -I $(top_srcdir)/warmrestart -I $(FPM_PATH) -I $(top_srcdir)/lib
 
 bin_PROGRAMS = fpmsyncd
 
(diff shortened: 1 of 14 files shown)


.azure-pipelines/build_and_install_module.sh
```diff
@@ -35,24 +45,24 @@ function build_and_install_kmodule()
...
+    KERNEL_PACKAGE_SOURCE=$(trim $(apt-cache show linux-image-unsigned-${KERNEL_RELEASE} | grep ^Source: | cut -d':' -f 2))
+    KERNEL_PACKAGE_VERSION=$(trim $(apt-cache show linux-image-unsigned-${KERNEL_RELEASE} | grep ^Version: | cut -d':' -f 2))
+    SOURCE_PACKAGE_VERSION=$(apt-cache showsrc "${KERNEL_PACKAGE_SOURCE}" | grep ^Version: | cut -d':' -f 2 | tr '\n' ' ')
+    if ! echo "${SOURCE_PACKAGE_VERSION}" | grep "\b${KERNEL_PACKAGE_VERSION}\b"; then
+        echo "WARNING: the running kernel version (${KERNEL_PACKAGE_VERSION}) doesn't match any of the available source " \
+            "packge versions (${SOURCE_PACKAGE_VERSION}) being downloaded. There's no guarantee the module being downloaded " \
             "can be loaded into the kernel or function correctly. If possible, please update your kernel and reboot " \
             "your system so that it's running the matching kernel version." >&2
         echo "Continuing with the build anyways" >&2
     fi
```
prabhataravind, line 57: should we just exit out here rather than trying to proceed when no match is found?

prabhataravind: I guess if you are specifying KERNEL_PACKAGE_SOURCE, this condition is probably never hit. Is that correct?

saiarcot895: I suspect that with this current code, the subsequent `apt-get source` command will fail anyways, but I can change this to be a hard exit.

saiarcot895: Done

prabhataravind (approved):
lgtm

theasianpianist (approved):

PR #3726, opened 2025-06-24
[fpmsyncd]Fixing the blackhole route removal during warmboot

dgsudharsan:
**What I did**
Fixed blackhole route handling during warm reboot. Reorganized the code so that future flows don't miss this functionality.

**Why I did it**
During the warmboot, the blackhole route gets removed as a result of reconciliation. This is due to missing logic to handle blackhole routes during warm boot.
```
2025 Jun  9 21:06:38.240274 sonic NOTICE bgp#fpmsyncd: :- reconcile: Warm-Restart reconciliation: deleting stale entry 1000::/120 { blackhole: true | protocol: 0xc4 }
```

**How I verified it**
Added UT to verify

**Details if related**


vivekrnv (approved):

fpmsyncd/routesync.cpp
```diff
@@ -1789,34 +1722,18 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
         }
     }
 
-    if (!warmRestartInProgress)
+    if(nhg_id)
```
prsunny: space after if

fpmsyncd/routesync.cpp
```diff
@@ -1651,7 +1598,8 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
             FieldValueTuple fv("blackhole", "true");
             fvVector.push_back(fv);
             fvVector.push_back(proto);
-            m_routeTable->set(destipprefix, fvVector);
+            setRouteWithWarmRestart(destipprefix, fvVector, m_routeTable, SET_COMMAND);
```
prsunny, line 1601: Assuming this is the fix code and rest are reformatting, correct?

prsunny:
@vaibhavhd for viz

fpmsyncd/routesync.cpp
dgsudharsan: Correct. The reformatting will ensure future occurrences don't miss it. There were several places where this was missed and hence it is better to have the functionality under a API

fpmsyncd/routesync.cpp
dgsudharsan: Done

yejianquan:
Hi @dgsudharsan , please create a PR fixes the conflict to 202505 branch

dgsudharsan:
> Hi @dgsudharsan , please create a PR fixes the conflict to 202505 branch

@yejianquan Here is the PR https://github.com/sonic-net/sonic-swss/pull/3740

PR #3594, opened 2025-04-06
[trim]: Add Packet Trimming to OA

nazariig:
Signed-off-by: Nazarii Hnydyn <nazariig@nvidia.com>

**DEPENDS:**
1. https://github.com/sonic-net/sonic-swss-common/pull/1001
2. https://github.com/sonic-net/sonic-sairedis/pull/1575
3. https://github.com/sonic-net/sonic-sairedis/pull/1598

**HLD:** https://github.com/sonic-net/SONiC/pull/1898

**What I did**
* Implemented Packet Trimming feature

**Why I did it**
* Implementation is done according to the Packet Trimming HLD

**How I verified it**
1. Run Packet Trimming VS UTs

**Details if related**
* N/A

#### A picture of a cute animal (not mandatory but encouraged)
```
      .---.        .-----------
     /     \  __  /    ------
    / /     \(  )/    -----
   //////   ' \/ `   ---
  //// / // :    : ---
 // /   /  /`    '--
//          //..\\
       ====UU====UU====
           '//||\\`
             ''``
```

PR #3730, opened 2025-06-27
switch-trimming global config capability check

anilkpan:
Allow switch trimming global config even when some parameters capabilities are not supported.

**What I did**
Included changes from the following PR:
https://github.com/sonic-net/sonic-swss/pull/3594

- Added support to allow switch trimming global config even when some parameters capabilities are not supported.
- Pass down the configuration to SAI only if the capability is supported.

### DIFF
--- a/orchagent/switch/trimming/capabilities.cpp
+++ b/orchagent/switch/trimming/capabilities.cpp
@@ -1,0 +1,376 @@
+// includes -----------------------------------------------------------------------------------------------------------
+
+extern "C" {
+#include <saitypes.h>
+#include <saiobject.h>
+#include <saistatus.h>
+#include <saiswitch.h>
+}
+
+#include <cstdint>
+#include <cstdbool>
+#include <string>
+#include <vector>
+#include <set>
+#include <unordered_map>
+#include <algorithm>
+
+#include <sai_serialize.h>
+#include <stringutility.h>
+#include <dbconnector.h>
+#include <table.h>
+#include <schema.h>
+#include <logger.h>
+
+#include "schema.h"
+#include "capabilities.h"
+
+using namespace swss;
+
+// defines ------------------------------------------------------------------------------------------------------------
+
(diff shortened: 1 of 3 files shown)


adyeung:
@anilkpan pls regenerate a code PR, #3594 has merged

prsunny:
can you check on coverage?

orchagent/switch/trimming/capabilities.cpp
```diff
@@ -109,7 +109,27 @@ bool SwitchTrimmingCapabilities::isSwitchTrimmingSupported() const
...
+}
+
+bool SwitchTrimmingCapabilities::isSwitchTrimmingDscpSetSupported() const
+{
+    return capabilities.dscp.isAttrSupported;
+}
+
+bool SwitchTrimmingCapabilities::isSwitchTrimmingQueueSetSupported() const
+{
+    return capabilities.mode.isStaticModeSupported && capabilities.queue.isAttrSupported;
```
developfast, line 127: why does queue index setting require static mode support specifically?

orchagent/switchorch.cpp
```diff
@@ -959,94 +959,106 @@ bool SwitchOrch::setSwitchTrimming(const SwitchTrimming &trim)
         return true;
     }
 
-    if (trim.size.is_set)
+    if (trimCap.isSwitchTrimmingSizeSetSupported())
     {
-        if (!tObj.size.is_set || (tObj.size.value != trim.size.value))
+        if (trim.size.is_set)
```
developfast, line 964: this doesn't handle the case where `trim.queue.index.is_set` is false but `tObj.queue.index.is_set` is true. possibly add an else block with appropriate error handling for consistency.

nazariig:
These changes violate the original HLD proposal. I would prefer to avoid such a changes. In case some vendor needs to bypass it - we might need to think about the vendor specific capabilities check.

nazariig:
Please rebase the PR on top of: https://github.com/sonic-net/sonic-swss/pull/3705

PR #3743, opened 2025-07-03
[P4Orch] Temp reuse gUnderlayIfId for tunnel objects.

divyagayathri-hcl:
**What I did**
Updated the bulker API logic to ensure that bulk SAI operations for non-OID objects are executed in the order they were queued.
Temporarily reused the global **gUnderlayIfId** variable to program tunnel-related SAI objects that require a valid underlay router interface.

**Why I did it**
Some platforms and SAI implementations require strict ordering of non-OID operations due to dependencies.
The reuse of **gUnderlayIfId** provides a temporary workaround to support tunnel object creation when a proper tunnel interface or mapping is not yet available.

**How I verified it**
Created and applied a test configuration with dependent non-OID objects.

**Details if related**
The ordering fix only affects non-OID bulk operations.

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -1022,6 +1022,8 @@ void OrchDaemon::start(long heartBeatInterval)
  */
 bool OrchDaemon::warmRestoreAndSyncUp()
 {
+    SWSS_LOG_ENTER();
+
     WarmStart::setWarmStartState("orchagent", WarmStart::INITIALIZED);
 
     for (Orch *o : m_orchList)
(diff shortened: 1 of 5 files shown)


prsunny:
@divyagayathris , please have the PR descrption updated per template. You can keep the PR in 'draft' state until its ready for review

rkavitha-hcl (approved):

divyagayathri-hcl:
> Should the bulker change be done in a separate PR so we can keep a clean P4orch PR

Done.

rkavitha-hcl (approved):

PR #3749, opened 2025-07-08
Fix container crashes related to AP ports, ARP SVI, and loopback IP configuration

haojun291:
**What I did**
    1. Fixed the issue where the AP port with member container crashes during boot.
    2. Fixed the issue where the AP port located in an already learned ARP SVI crashes when deleting any of its AP member ports.
    3. Fixed the issue where configuring IP on a loopback port causes the container to crash after saving and restarting the configuration.

**Why I did it**
  NA

**How I verified it**
  NA

**Details if related**
  NA


dgsudharsan (changes requested):
Can you please add UT?

PR #3753, opened 2025-07-09
Skip ref counting standby mux neighbor NHs when added to NH group

manamand2020:
**What I did**
Skip ref counting the MUX neighbor NHs while getting added to NH group when MUX state is in standby.

**Why I did it**
This issue was causing ref counts of mux NHs to go wrong, observed this in DVS test. 
m_neighOrch->getNextHopId(it) this returns tunnel NH id when mux state is standby. However the code was incrementing the ref count of local neighbor NH based on the nh-key.

**How I verified it**
DVS test

**Details if related**

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -1172,7 +1172,7 @@ void RouteOrch::increaseNextHopRefCount(const NextHopGroupKey &nexthops)
     else
     {
         m_syncdNextHopGroups[nexthops].ref_count ++;
-        SWSS_LOG_INFO("Routeorch inc Ref count %u for next_hops", m_syncdNextHopGroups[nexthops].ref_count);
+        SWSS_LOG_INFO("Routeorch inc Ref count %u for next_hops: %s", m_syncdNextHopGroups[nexthops].ref_count, nexthops.to_string().c_str());
     }
 }
 
@@ -1194,7 +1194,7 @@ void RouteOrch::decreaseNextHopRefCount(const NextHopGroupKey &nexthops)
     else
     {
         m_syncdNextHopGroups[nexthops].ref_count --;
-        SWSS_LOG_INFO("Routeorch dec Ref count %u for next_hops", m_syncdNextHopGroups[nexthops].ref_count);
+        SWSS_LOG_INFO("Routeorch dec Ref count %u for next_hops: %s", m_syncdNextHopGroups[nexthops].ref_count, nexthops.to_string().c_str());
     }
 }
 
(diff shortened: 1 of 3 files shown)


orchagent/routeorch.cpp
```diff
@@ -1172,6 +1172,7 @@ void RouteOrch::increaseNextHopRefCount(const NextHopGroupKey &nexthops)
     else
     {
         m_syncdNextHopGroups[nexthops].ref_count ++;
+        SWSS_LOG_INFO("Routeorch inc Ref count %u for next_hops", m_syncdNextHopGroups[nexthops].ref_count);
```
Ndancejic: Could you also print out which nexthops are being incremented here?

orchagent/routeorch.cpp
```diff
@@ -1193,6 +1194,7 @@ void RouteOrch::decreaseNextHopRefCount(const NextHopGroupKey &nexthops)
     else
     {
         m_syncdNextHopGroups[nexthops].ref_count --;
+        SWSS_LOG_INFO("Routeorch dec Ref count %u for next_hops", m_syncdNextHopGroups[nexthops].ref_count);
```
Ndancejic: same here

orchagent/routeorch.cpp
```diff
@@ -1443,7 +1448,20 @@ bool RouteOrch::addNextHopGroup(const NextHopGroupKey &nexthops)
 
     /* Increment the ref_count for the next hops used by the next hop group. */
     for (auto it : next_hop_set)
-        m_neighOrch->increaseNextHopRefCount(it);
+    {
+        // Skip next hops that were filtered out (interface down)
+        if (m_neighOrch->isNextHopFlagSet(it, NHFLAGS_IFDOWN))
+        {
+            continue;
+        }
```
Ndancejic: does the refcount get incremented when the interface comes back up?

orchagent/routeorch.cpp
```diff
@@ -1527,11 +1545,19 @@ bool RouteOrch::removeNextHopGroup(const NextHopGroupKey &nexthops)
...
     m_nextHopGroupCount--;
     gCrmOrch->decCrmResUsedCounter(CrmResourceType::CRM_NEXTHOP_GROUP);
+    MuxOrch* mux_orch = gDirectory.get<MuxOrch*>();
+    sai_object_id_t mux_tunnel_nh_id = mux_orch->getTunnelNextHopId();
 
     set<NextHopKey> next_hop_set = nexthops.getNextHops();
     for (auto it : next_hop_set)
     {
-        m_neighOrch->decreaseNextHopRefCount(it);
+        // skip ref count update if the underlying nh from mux is tunnel nh
```
Ndancejic: Should the interface status be checked here as well? if the interface is down (and the refcount is decreased in the interface down path) we might decrease the refcount twice

prsunny:
@Ndancejic , @manamand2020 , is this needed for 202505 cherry-pick?

manamand2020:
> @Ndancejic , @manamand2020 , is this needed for 202505 cherry-pick?

I think we should pick this for 202505. This will be required before we can merge the no-host-route changes in 202505. @Ndancejic what is your opinion?

PR #3764, opened 2025-07-14
Support SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION

longhuan-cisco:
> [!NOTE]  
> This PR has dependency on latest OCP SAI attribute [SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION](https://github.com/opencomputeproject/SAI/commit/a08584b535c8715b15c9c02a200f0366672531f8), which hasn't been released to [sonic-sairedis/SAI](https://github.com/sonic-net/sonic-sairedis/tree/master). Thus, before its release, it's expected to see build error on this attribute in pre-commit pipeline.

sonic-platform-daemon side change: https://github.com/sonic-net/sonic-platform-daemons/pull/643

**What I did**
Support [SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION](https://github.com/opencomputeproject/SAI/pull/2156)

**Why I did it**

**How I verified it**
Tested end-to-end xcvrd->OA->SAI/SDK

**Details if related**

### DIFF
--- a/orchagent/port/porthlpr.cpp
+++ b/orchagent/port/porthlpr.cpp
@@ -20,7 +20,6 @@ using namespace swss;
 
 // types --------------------------------------------------------------------------------------------------------------
 
-typedef decltype(PortConfig::serdes) PortSerdes_t;
 typedef decltype(PortConfig::link_event_damping_config) PortDampingConfig_t;
 
 // constants ----------------------------------------------------------------------------------------------------------
@@ -697,6 +696,27 @@ bool PortHelper::parsePortLinkTraining(PortConfig &port, const std::string &fiel
     return true;
 }
 
+// custom_collection specialization
+bool PortHelper::parsePortSerdes(
+    decltype(PortSerdes_t::custom_collection)& serdes,
+    const std::string& field,
+    const std::string& value) const
+{
+    SWSS_LOG_ENTER();
+
+    if (value.empty())
+    {
+        SWSS_LOG_ERROR("Failed to parse field(%s): empty string is prohibited",
+                       field.c_str());
+        return false;
+    }
+
+    // directly assign the raw string
+    serdes.value  = value;
+    serdes.is_set = true;
+    return true;
+}
+
 template<typename T>
 bool PortHelper::parsePortSerdes(T &serdes, const std::string &field, const std::string &value) const
 {
@@ -1178,6 +1198,13 @@ bool PortHelper::parsePortConfig(PortConfig &port) const
         else if (field == PORT_REGN_BFM1N)
         {
             if (!this->parsePortSerdes(port.serdes.regn_bfm1n, field, value))
+            {
+                return false;
+            }
+        }
+        else if (field == PORT_CUSTOM_SERDES_ATTRS)
+        {
+            if (!this->parsePortSerdes(port.serdes.custom_collection, field, value))
             {
                 return false;
             }
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -4280,7 +4280,11 @@ void PortsOrch::doPortTask(Consumer &consumer)
             else
             {
                 PortSerdesAttrMap_t serdes_attr;
+                string custom_serdes_attrs_str;
                 getPortSerdesAttr(serdes_attr, pCfg);
+                if (pCfg.serdes.custom_collection.is_set) {
+                    custom_serdes_attrs_str = pCfg.serdes.custom_collection.value;
+                }
 
                 // Saved configured admin status
                 bool admin_status = p.m_admin_state_up;
@@ -4391,9 +4395,14 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         updatePortStatePoll(p, PORT_STATE_POLL_LT, pCfg.link_training.value);
 
                         // Restore pre-emphasis when LT is transitioned from ON to OFF
-                        if (!p.m_link_training && serdes_attr.empty())
+                        if (!p.m_link_training && (serdes_attr.empty() || custom_serdes_attrs_str.empty()))
                         {
-                            serdes_attr = p.m_preemphasis;
+                            if (serdes_attr.empty()) {
+                                serdes_attr = p.m_preemphasis;
+                            }
+                            if (custom_serdes_attrs_str.empty()) {
+                                custom_serdes_attrs_str = p.m_custom_serdes_attrs;
+                            }
                         }
 
                         SWSS_LOG_NOTICE(
@@ -4904,12 +4913,17 @@ void PortsOrch::doPortTask(Consumer &consumer)
                     }
                 }
 
-                if (!serdes_attr.empty())
+                if (!serdes_attr.empty() or !custom_serdes_attrs_str.empty())
                 {
                     if (p.m_link_training)
                     {
                         SWSS_LOG_NOTICE("Save port %s preemphasis for LT", p.m_alias.c_str());
-                        p.m_preemphasis = serdes_attr;
+                        if (!serdes_attr.empty()) {
+                            p.m_preemphasis = serdes_attr;
+                        }
+                        if (!custom_serdes_attrs_str.empty()) {
+                            p.m_custom_serdes_attrs = custom_serdes_attrs_str;
+                        }
                         m_portList[p.m_alias] = p;
                     }
                     else
@@ -4928,10 +4942,15 @@ void PortsOrch::doPortTask(Consumer &consumer)
                                 m_portList[p.m_alias] = p;
                         }
 
-                        if (setPortSerdesAttribute(p.m_port_id, gSwitchId, serdes_attr))
+                        if (setPortSerdesAttribute(p.m_port_id, gSwitchId, serdes_attr, custom_serdes_attrs_str))
                         {
                             SWSS_LOG_NOTICE("Set port %s SI settings is successful", p.m_alias.c_str());
-                            p.m_preemphasis = serdes_attr;
+                            if (!serdes_attr.empty()) {
+                                p.m_preemphasis = serdes_attr;
+                            }
+                            if (!custom_serdes_attrs_str.empty()) {
+                                p.m_custom_serdes_attrs = custom_serdes_attrs_str;
+                            }
                             m_portList[p.m_alias] = p;
                         }
                         else
@@ -9009,7 +9028,8 @@ bool PortsOrch::removeAclTableGroup(const Port &p)
 }
 
 bool PortsOrch::setPortSerdesAttribute(sai_object_id_t port_id, sai_object_id_t switch_id,
-                                       map<sai_port_serdes_attr_t, vector<uint32_t>> &serdes_attr)
+                                       map<sai_port_serdes_attr_t, vector<uint32_t>> &serdes_attr,
+                                       const std::string& custom_serdes_attrs_str)
 {
     SWSS_LOG_ENTER();
 
@@ -9060,8 +9080,17 @@ bool PortsOrch::setPortSerdesAttribute(sai_object_id_t port_id, sai_object_id_t 
         port_serdes_attr.value.u32list.list = it->second.data();
         attr_list.emplace_back(port_serdes_attr);
     }
+
+    if (!custom_serdes_attrs_str.empty())
+    {
+        port_serdes_attr.id = SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION;
+        port_serdes_attr.value.json.json.count = static_cast<uint32_t>(custom_serdes_attrs_str.size());
+        port_serdes_attr.value.json.json.list = reinterpret_cast<int8_t*>(const_cast<char*>(custom_serdes_attrs_str.data()));
+        attr_list.emplace_back(port_serdes_attr);
+    }
+
     status = sai_port_api->create_port_serdes(&port_serdes_id, switch_id,
-                                              static_cast<uint32_t>(serdes_attr.size()+1),
+                                              static_cast<uint32_t>(attr_list.size()),
                                               attr_list.data());
 
     if (status != SAI_STATUS_SUCCESS)
--- a/orchagent/port.h
+++ b/orchagent/port.h
@@ -243,6 +243,7 @@ public:
 
     /* pre-emphasis */
     std::map<sai_port_serdes_attr_t, std::vector<uint32_t>> m_preemphasis;
+    std::string m_custom_serdes_attrs;
 
     /* Force initial parameter configuration flags */
     bool m_an_cfg = false;        // Auto-negotiation (AN)
--- a/orchagent/port/portcnt.h
+++ b/orchagent/port/portcnt.h
@@ -190,6 +190,11 @@ public:
             bool is_set = false;
         } unreliable_los; // Port unreliable_los
 
+        struct {
+            std::string value;
+            bool is_set = false;
+        } custom_collection; // Port serdes custom_collection
+
     } serdes; // Port serdes
 
     struct {
--- a/orchagent/port/porthlpr.h
+++ b/orchagent/port/porthlpr.h
@@ -7,6 +7,9 @@
 
 #include "portcnt.h"
 
+typedef decltype(PortConfig::serdes) PortSerdes_t;
+typedef decltype(PortSerdes_t::custom_collection) CustomSerdes_t;
+
 class PortHelper final
 {
 public:
@@ -36,6 +39,7 @@ public:
 private:
     std::string getFieldValueStr(const PortConfig &port, const std::string &field) const;
 
+    bool parsePortSerdes(CustomSerdes_t& serdes, const std::string& field, const std::string& value) const;
     template<typename T>
     bool parsePortSerdes(T &serdes, const std::string &field, const std::string &value) const;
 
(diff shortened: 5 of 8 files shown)


longhuan-cisco:
@prgeor @mihirpat1  could you please review, thanks

orchagent/portsorch.cpp
```diff
@@ -9060,8 +9080,17 @@ bool PortsOrch::setPortSerdesAttribute(sai_object_id_t port_id, sai_object_id_t
...
+    {
+        port_serdes_attr.id = SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION;
+        port_serdes_attr.value.json.json.count = static_cast<uint32_t>(custom_serdes_attrs_str.size());
+        port_serdes_attr.value.json.json.list = reinterpret_cast<int8_t*>(const_cast<char*>(custom_serdes_attrs_str.data()));
+        attr_list.emplace_back(port_serdes_attr);
+    }
+
     status = sai_port_api->create_port_serdes(&port_serdes_id, switch_id,
-                                              static_cast<uint32_t>(serdes_attr.size()+1),
+                                              static_cast<uint32_t>(attr_list.size()),
```
prgeor, line 9485: @longhuan-cisco I am not sure why the attribute count was previously +1, but do you see any harm in maintaining the same this time as well?

orchagent/portsorch.cpp
```diff
@@ -9009,7 +9028,8 @@ bool PortsOrch::removeAclTableGroup(const Port &p)
 }
 
 bool PortsOrch::setPortSerdesAttribute(sai_object_id_t port_id, sai_object_id_t switch_id,
```
prgeor, line 9429: @longhuan-cisco please introduce `setPortCustomSerdesAttribute()` so that it does not affect old API usage and limits your change to custom API only

longhuan-cisco: enum [SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION](https://github.com/opencomputeproject/SAI/blob/b925fc0f0fc8c620eea7d9dead7fc99e33b7b44e/inc/saiport.h#L4185) is the same `sai_port_serdes_attr_t` type as today's `IDRIVER / PREEMPHASIS / etc`.
And the same SAI API `sai_port_api->create_port_serdes` is used for all of them.

```
typedef enum _sai_port_serdes_attr_t
{
    SAI_PORT_SERDES_ATTR_START,

    SAI_PORT_SERDES_ATTR_PORT_ID = SAI_PORT_SERDES_ATTR_START,

    SAI_PORT_SERDES_ATTR_PREEMPHASIS,

    SAI_PORT_SERDES_ATTR_IDRIVER,

...
    SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION,
...
} sai_port_serdes_attr_t;
```

The logic/code setting `IDRIVER / PREEMPHASIS / etc` is exactly the same as the one setting `CUSTOM_COLLECTION`. The only difference is their value type: the value of `IDRIVER / PREEMPHASIS / etc` is `u32list`,  the value of `CUSTOM_COLLECTION` is `json`.
If we have two separate `set` functions: one `setPortSerdesAttribute()` for `IDRIVER / PREEMPHASIS / etc`, another `setPortCustomSerdesAttribute()` for `CUSTOM_COLLECTION`. We will end up having lots of duplicate code in these functions (almost the entire function code will be copy/paste), which doesn't look good.

I feel the effect to old API usage is already fairly self-contained:
```
    if (!custom_serdes_attrs_str.empty())
    {
        port_serdes_attr.id = SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION;
        port_serdes_attr.value.json.json.count = static_cast<uint32_t>(custom_serdes_attrs_str.size());
        port_serdes_attr.value.json.json.list = reinterpret_cast<int8_t*>(const_cast<char*>(custom_serdes_attrs_str.data()));
        attr_list.emplace_back(port_serdes_attr);
    }
```
Only when `custom_serdes_attrs_str` is not empty, the above added code will get exercised.

orchagent/portsorch.cpp
longhuan-cisco: The previous code was using [`serdes_attr`](https://github.com/sonic-net/sonic-swss/blob/71219b0d41321093de633d5523b5e81eb7f34077/orchagent/portsorch.cpp#L9015)'s size +1 (not [`attr_list`](https://github.com/sonic-net/sonic-swss/blob/71219b0d41321093de633d5523b5e81eb7f34077/orchagent/portsorch.cpp#L9019)'s size + 1).
[`serdes_attr`](https://github.com/sonic-net/sonic-swss/blob/71219b0d41321093de633d5523b5e81eb7f34077/orchagent/portsorch.cpp#L9015) doesn't contain SAI_PORT_SERDES_ATTR_PORT_ID attribute (i.e. [`attr_list`'s first attribute](https://github.com/sonic-net/sonic-swss/blob/71219b0d41321093de633d5523b5e81eb7f34077/orchagent/portsorch.cpp#L9054-L9056)), that's why `+1`.


It would better just simply use `attr_list.size()` as attribute count, which is less bug prone.
https://github.com/opencomputeproject/SAI/blob/b925fc0f0fc8c620eea7d9dead7fc99e33b7b44e/inc/saiport.h#L4200-L4214
```
/**
 * @brief Create port serdes
 *
 * @param[out] port_serdes_id Port serdes id
 * @param[in] switch_id Switch id
 * @param[in] attr_count Number of attributes
 * @param[in] attr_list Array of attributes
 *
 * @return #SAI_STATUS_SUCCESS on success, failure status code on error
 */
typedef sai_status_t (*sai_create_port_serdes_fn)(
        _Out_ sai_object_id_t *port_serdes_id,
        _In_ uint32_t attr_count,
        _In_ const sai_attribute_t *attr_list);
```

Besides, here, if we maintain `serdes_attr.size()+1` as attribute count, issue will happen because it could end up truncating the last attribute of `attr_list` (i.e. SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION), which is not in `serdes_attr` since argument [`serdes_attr`](https://github.com/longhuan-cisco/sonic-swss/blob/fe4480665e2228d556f7ae252089b421239f869f/orchagent/portsorch.cpp#L9031) with type `std::map<sai_port_serdes_attr_t, std::vector<uint32_t>>` can only contain  `IDRIVER / PREEMPHASIS / etc` whose value type is `uint32_t` list, SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION is passed via argument [`custom_serdes_attrs_str`](https://github.com/longhuan-cisco/sonic-swss/blob/fe4480665e2228d556f7ae252089b421239f869f/orchagent/portsorch.cpp#L9032) separately.

orchagent/portsorch.cpp
prgeor: @longhuan-cisco there is no issue in the logic. This is just to keep the new code separate and less impactful for existing platforms which are not using custom attribute

orchagent/port/porthlpr.cpp
```diff
@@ -697,6 +696,27 @@ bool PortHelper::parsePortLinkTraining(PortConfig &port, const std::string &fiel
...
+
+    if (value.empty())
+    {
+        SWSS_LOG_ERROR("Failed to parse field(%s): empty string is prohibited",
+                       field.c_str());
+        return false;
+    }
+
+    // directly assign the raw string
+    serdes.value  = value;
```
prgeor: @longhuan-cisco I don't see a reason to NOT use existing parsePortSerdes() where serdes.value is a vector of uint32_t. This new function keeps the serdes.value as a string of values.

prgeor: @longhuan-cisco How would the std::string& value look like? Is this a json string? Any example?

orchagent/portsorch.cpp
longhuan-cisco: As discussed in the meeting, moved to variant type for `map<sai_port_serdes_attr_t, SerdesValue> &serdes_attr` to support both vector and string as value

```
using SerdesValue = boost::variant<std::vector<uint32_t>, std::string>;
```

orchagent/port/porthlpr.cpp
longhuan-cisco: > @longhuan-cisco I don't see a reason to NOT use existing parsePortSerdes() where serdes.value is a vector of uint32_t. This new function keeps the serdes.value as a string of values.

Keep only the existing parsePortSerdes() and use `std::enable_if` to achieve different implementations for different types of `serdes.value`

> @longhuan-cisco How would the std::string& value look like? Is this a json string? Any example?

It's a JSON,  e,g. `'{"attributes":[{"XYZ":{"value":[12,13]}},{"ABC":{"value":[3,4]}}]}'`

orchagent/port/porthlpr.cpp
```diff
@@ -708,6 +708,25 @@ bool PortHelper::parsePortSerdes(T &serdes, const std::string &field, const std:
         return false;
     }
 
+    // Use SFINAE with enable_if for extensible type handling for serdes.value
```
prgeor, line 711: @longhuan-cisco What is `SFINAE` ?

longhuan-cisco: SFINAE (Substitution Failure Is Not An Error) is a C++ template metaprogramming technique where a substitution failure for a template argument is not a compiler error, but rather a signal to discard that specific template overload from overload resolution. It allows developers to write generic code that adapts its behavior or even its existence based on the properties of the types involved, enabling advanced compile-time polymorphism and trait detection before the introduction of concepts. 

Examples of SFINAE Techniques
1. `std::enable_if`: A common tool that uses SFINAE to conditionally enable or disable template functions based on a type trait.
2. `if constexpr` (C++17): A more readable compile-time conditional that has become a simpler alternative to SFINAE in many cases.
3. Concepts (C++20): Provide a modern, more expressive, and user-friendly way to constrain templates, largely superseding SFINAE for type checking. 

Here, we use `enable_if`, because C++ compiler used in today's SONiC swss is < 17

prgeor:
@prsunny can you merge?

longhuan-cisco:
> @prsunny can you merge?

@prsunny kindly reminder, thanks

prsunny:
@longhuan-cisco , i see new boost library added, can you provide description of why its added?

longhuan-cisco:
> @longhuan-cisco , i see new boost library added, can you provide description of why its added?

@prsunny Sure, added explantation in PR description.

In today's orchagent code base, boost library has been used in multiple places, e.g.:
https://github.com/sonic-net/sonic-swss/blob/d0b25613fb41bfb5224a10c991368a3b28353138/orchagent/port/porthlpr.cpp#L10
https://github.com/sonic-net/sonic-swss/blob/d0b25613fb41bfb5224a10c991368a3b28353138/orchagent/bulker.h#L8

longhuan-cisco:
Looks like, [Azure.sonic-swss](https://github.com/sonic-net/sonic-swss/pull/3764/checks?check_run_id=55888029135) failure was due to base line issue, which was hit on other PRs also

longhuan-cisco:
I saw `Semgrep` pre-commit run got stuck in "Waiting for status to be reported" state for quite some time, not sure if it's due to some issue in Azure Pipelines

cc @prsunny

longhuan-cisco:
@prsunny kindly reminder for approval/merge. all pre-commit checks passed.

balanokia:
@prsunny @prgeor @lguohan : Pls shared review comments. We made this PR to introduce boost:variant type but it changes keywords in a non-functional way.
Hi @longhuan-cisco
Tx FIR Pre-emphasis is the key word for Tx FIR Equalization with certain number of taps and Pre-emphasis keyword is used in IEEE or all OEM datasheets in a functional meaning. Lets not change the keyword from pre-emphasis to generic keyword "serdes_attrs" and "serdes_attr" does not signify Tx FIR (tap) value set. Lets keep the code understanding in a functional meaning. Kindly share your views.
Thanks
Bala

longhuan-cisco:
> Hi @longhuan-cisco
> Tx FIR Pre-emphasis is the key word for Tx FIR Equalization with certain number of taps and Pre-emphasis keyword is used in IEEE or all OEM datasheets in a functional meaning. Lets not change the keyword from pre-emphasis to generic keyword "serdes_attrs" and "serdes_attr" does not signify Tx FIR (tap) value set. Lets keep the code understanding in a functional meaning. Kindly share your views.
> Thanks
> Bala

Hi @balanokia 
In the existing code before this PR, this variable had been used to cover many different serdes attributes for quite long time, including `SAI_PORT_SERDES_ATTR_PREEMPHASIS`, `SAI_PORT_SERDES_ATTR_IDRIVER`, `SAI_PORT_SERDES_ATTR_IPREDRIVER` and etc.
It's generic, not just for `SAI_PORT_SERDES_ATTR_PREEMPHASIS`

Please refer to below old code snapshot (before this PR merged):
https://github.com/sonic-net/sonic-swss/blob/5671e08be0ed467679de7fedf4a4cc2f0a636b0e/orchagent/portsorch.cpp#L435-L524
https://github.com/sonic-net/sonic-swss/blob/5671e08be0ed467679de7fedf4a4cc2f0a636b0e/orchagent/port/portschema.h#L75-L91
https://github.com/sonic-net/sonic-swss/blob/5671e08be0ed467679de7fedf4a4cc2f0a636b0e/orchagent/portsorch.cpp#L4492-L4493
https://github.com/sonic-net/sonic-swss/blob/5671e08be0ed467679de7fedf4a4cc2f0a636b0e/orchagent/portsorch.cpp#L5124

balanokia:
> > Hi @longhuan-cisco
> > Tx FIR Pre-emphasis is the key word for Tx FIR Equalization with certain number of taps and Pre-emphasis keyword is used in IEEE or all OEM datasheets in a functional meaning. Lets not change the keyword from pre-emphasis to generic keyword "serdes_attrs" and "serdes_attr" does not signify Tx FIR (tap) value set. Lets keep the code understanding in a functional meaning. Kindly share your views.
> > Thanks
> > Bala
> 
> Hi @balanokia In the existing code before this PR, this variable had been used to cover many different serdes attributes for quite long time, including `SAI_PORT_SERDES_ATTR_PREEMPHASIS`, `SAI_PORT_SERDES_ATTR_IDRIVER`, `SAI_PORT_SERDES_ATTR_IPREDRIVER` and etc. It's generic, not just for `SAI_PORT_SERDES_ATTR_PREEMPHASIS`
> 
> Please refer to below old code snapshot (before this PR merged):
> 
> https://github.com/sonic-net/sonic-swss/blob/5671e08be0ed467679de7fedf4a4cc2f0a636b0e/orchagent/portsorch.cpp#L435-L524
> 
> 
> https://github.com/sonic-net/sonic-swss/blob/5671e08be0ed467679de7fedf4a4cc2f0a636b0e/orchagent/port/portschema.h#L75-L91
> 
> 
> https://github.com/sonic-net/sonic-swss/blob/5671e08be0ed467679de7fedf4a4cc2f0a636b0e/orchagent/portsorch.cpp#L4492-L4493
> 
> 
> https://github.com/sonic-net/sonic-swss/blob/5671e08be0ed467679de7fedf4a4cc2f0a636b0e/orchagent/portsorch.cpp#L5124

Hi @longhuan-cisco 
Thanks for clarifying. Its good to have abstraction of serdes attributes under one bucket. When we invoke to apply a specific type of serdes attributes, we know what significant parameters that we update and hence it would be better to keep the variable names or atleast keep the log messages as-is in the past like "SWSS_LOG_NOTICE("Save port %s preemphasis for LT", p.m_alias.c_str())". It would help us to debug port bringup sequences (LT, FEC, Tx FIR, etc) especially external PHY stages. 

Thanks
Bala

longhuan-cisco:
> Hi @longhuan-cisco Thanks for clarifying. Its good to have abstraction of serdes attributes under one bucket. When we invoke to apply a specific type of serdes attributes, we know what significant parameters that we update and hence it would be better to keep the variable names or atleast keep the log messages as-is in the past like "SWSS_LOG_NOTICE("Save port %s preemphasis for LT", p.m_alias.c_str())". It would help us to debug port bringup sequences (LT, FEC, Tx FIR, etc) especially external PHY stages.
> 
> Thanks
> Bala

Hi @balanokia, thanks for the context on PHY terminology.

In general PHY literature, “pre‑emphasis” is indeed often used to refer to TX equalization and is sometimes implemented/expressed via TX‑FIR/FFE taps.

However, in SONiC orchagent this saved/restored bucket has historically been a generic SERDES attribute bucket, not strictly “pre‑emphasis only”. It can include multiple SAI_PORT_SERDES_ATTR_* knobs depending on platform media_settings.json/config (e.g., PREEMPHASIS, IDRIVER/IPREDRIVER, TX_FIR_* taps, PAM4 ratio/common-mode controls, precoding, etc.). SAI itself defines these as distinct SERDES attributes, where PREEMPHASIS is just one item in the list.

Because the bucket is not guaranteed to contain PREEMPHASIS on every platform/vendor/module, keeping “preemphasis” as the generic bucket/variable/log keyword is semantically misleading and can oversimplify what is actually being saved/restored. Renaming it to `serdes_attrs` improves correctness without changing behavior or reducing log informativeness—the previous logs did not actually confirm whether PREEMPHASIS was present either.

For debuggability, a more robust community-wide improvement would be to log which SERDES attributes are being set/saved/restored (e.g., list attr IDs/names at NOTICE, and optionally values at DEBUG), so anyone can grep for a stable tag and still see the exact contents. This makes the behavior transparent across all vendors/platforms/config combinations.

Thanks,
Longyin

longhuan-cisco:
@prgeor Could we cherry-pick this to 202511

balanokia:
@longhuan-cisco @prgeor Pls suggest if we plan to take this PR to 202511 branch. I already see @longhuan-cisco already raised it for merge to 202511.

PR #3769, opened 2025-07-17
SHlorch changes for EVPN MH

jayakj:
**What I did**
Implementation of SHL orch from EVPN VXLAN Multihoming HLD.
https://github.com/sonic-net/SONiC/blob/master/doc/vxlan/EVPN/EVPN_VxLAN_Multihoming.md#3310-ShlOrch

**Why I did it**

**How I verified it**
By SWSS UT

**Details if related**

### DIFF
--- a/orchagent/shlorch.cpp
+++ b/orchagent/shlorch.cpp
@@ -1,0 +1,604 @@
+#include "shlorch.h"
+#include "converter.h"
+#include "tokenize.h"
+#include "portsorch.h"
+#include "vxlanorch.h"
+#include "directory.h"
+
+extern sai_object_id_t gSwitchId;
+extern PortsOrch *gPortsOrch;
+extern sai_isolation_group_api_t*  sai_isolation_group_api;
+extern sai_bridge_api_t *sai_bridge_api;
+extern ShlOrch *gShlOrch;
+extern Directory<Orch*> gDirectory;
+
+ShlOrch::ShlOrch(vector<TableConnector> &connectors) : Orch(connectors)
+{
+    SWSS_LOG_ENTER();
+    gPortsOrch->attach(this);
+}
+
+ShlOrch::~ShlOrch()
+{
+    SWSS_LOG_ENTER();
+}
+
+shared_ptr<ShlIsolationGroup>
+ShlOrch::getIsolationGroup(string vtep_ip_addr)
+{
+    SWSS_LOG_ENTER();
+
+    shared_ptr<ShlIsolationGroup> ret = nullptr;
+
+    auto grp = m_isolationGrps.find(vtep_ip_addr);
+    if (grp != m_isolationGrps.end())
+    {
+        ret = grp->second;
+    }
+
+    return ret;
+}
+
+long unsigned int ShlOrch::getIsolationGroupCount()
+{
+    SWSS_LOG_ENTER();
+
+    return m_isolationGrps.size();
+}
+
+long unsigned int ShlOrch::getVtepsListCount()
+{
+    SWSS_LOG_ENTER();
+
+    return m_vtep_list.size();
+}
+
+void
+ShlOrch::doTask(Consumer &consumer)
+{
+    SWSS_LOG_ENTER();
+
+    if (!gPortsOrch->allPortsReady())
+    {
+        SWSS_LOG_ERROR("ports are not ready, exiting doTask");
+        return;
+    }
+
+    string table_name = consumer.getTableName();
+    if (table_name == APP_EVPN_SPLIT_HORIZON_TABLE_NAME)
+    {
+        doShlTblTask(consumer);
+    }
+    else
+    {
+        SWSS_LOG_ERROR("Invalid table %s", table_name.c_str());
+    }
+}
+
+void
+ShlOrch::doShlTblTask(Consumer &consumer)
+{
+    SWSS_LOG_ENTER();
+    shl_isolation_group_status_t status = SHL_ISO_GRP_STATUS_SUCCESS;
+
+    auto it = consumer.m_toSync.begin();
+    while (it != consumer.m_toSync.end())
+    {
+        KeyOpFieldsValuesTuple t = it->second;
+
+        /* format: <VLAN_name>:<Ifname> */
+        vector<string> keys = tokenize(kfvKey(t), ':', 1);
+        string op = kfvOp(t);
+
+        Port port;
+
+        if (!gPortsOrch->getPort(keys[1], port))
+        {
+            SWSS_LOG_ERROR("Port %s is not not yet created, delaying", keys[1].c_str());
+            it++;
+            continue;
+        }
+
+        string ifname(keys[1]);
+
+        if (op == SET_COMMAND)
+        {
+            string vteps_list("");
+
+            for (auto itp : kfvFieldsValues(t))
+            {
+                string attr_name = fvField(itp);
+                string attr_value = fvValue(itp);
+
+                if (attr_name == SHL_VTEPS_LIST)
+                {
+                    vteps_list = attr_value;
+
+                    auto old_list = m_vtep_list.find(ifname);
+                    auto new_list = tokenize(vteps_list, ',');
+                    vector<string> del_list, add_list;
+
+                    if (old_list != m_vtep_list.end()) {
+                        del_list = m_vtep_list[ifname];
+
+                        for (auto mem: new_list) {
+                            auto iter = find(del_list.begin(), del_list.end(), mem);
+                            if (iter != del_list.end()) {
+                                del_list.erase(iter);
+                            } else {
+                                add_list.push_back(mem);
+                            }
+                        }
+                    } else {
+                        add_list = new_list;
+                    }
+                    m_vtep_list[ifname] = new_list;
+                    addMemberToIsolationGroupPerVtep(add_list, port);
+                    delMemberFromIsolationGroupPerVtep(del_list, port);
+                }
+                else
+                    SWSS_LOG_ERROR("unknown Attr:%s", attr_name.c_str());
+            }
+        }
+        else
+        {
+            delMemberFromIsolationGroupPerVtep(m_vtep_list[ifname], port);
+            if (m_vtep_list.find(ifname) != m_vtep_list.end()) {
+                m_vtep_list.erase(ifname);
+            } else {
+                SWSS_LOG_ERROR("entry for ifname:%s does not exist", ifname.c_str());
+            }
+        }
+
+        if (status != SHL_ISO_GRP_STATUS_RETRY)
+        {
+            it = consumer.m_toSync.erase(it);
+        }
+        else
+        {
+            it++;
+        }
+    }
+}
+
+/*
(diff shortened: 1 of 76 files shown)


orchagent/shlorch.cpp
```diff
@@ -0,0 +1,604 @@
...
+}
+
+/*
+ * Walk through the Vtep list, do the following steps for each vtep
+ * step 1. check if isolation group exists, otherwise create isolation group
+ * stpe 2. add Port as member for this isolation group
+ * step 3. bind the tunnel port of vtep to this isolation group if its just created
+ */
+shl_isolation_group_status_t
+ShlOrch::addMemberToIsolationGroupPerVtep(vector<string> &addVtepList, Port &port) {
```
selva-nexthop, line 171: If we need logic to associate ports with multiple VTEPs or ES-IDs (for true multi-homing and split-horizon), we should ensure this mechanism is called appropriately for each group. If you want more detail or sample code, let me know.  Could you point to the code where it is being called?

orchagent/shlorch.h
```diff
@@ -0,0 +1,123 @@
...
+
+    map<string, sai_object_id_t> getMemberPorts()
+    {
+        return m_members;
+    }
+
+protected:
+    sai_object_id_t m_oid;  // sai isolation group object oid
+    string m_name;  // represents vtep ip address
+    map<string, sai_object_id_t> m_members; // Members {port Name, isolation group member OID}
```
selva-nexthop, line 79: suggestion:  use unordered_map if we really don't need ordered functionality in the map.  this will give better performance for scale.

orchagent/shlorch.h
```diff
@@ -0,0 +1,123 @@
...
+    vector<string> m_bind_ports; // Port to which this Isolation Group is applied.
```
selva-nexthop, line 80: Suggestion:
Consider std::unordered_set<std::string> for:
    - m_bind_ports
    - m_pending_members
    - m_pending_bind_ports
Reason: Fast insert, erase, and lookup.

orchagent/shlorch.h
```diff
@@ -0,0 +1,123 @@
...
+
+typedef enum ShlIsolationGroupStatus
+{
+    SHL_ISO_GRP_STATUS_RETRY = -100,
+    SHL_ISO_GRP_STATUS_FAIL,
+    SHL_ISO_GRP_STATUS_INVALID_PARAM,
+    SHL_ISO_GRP_STATUS_SUCCESS = 0
+} shl_isolation_group_status_t;
+
+class ShlIsolationGroup: public Observer, public Subject
```
selva-nexthop, line 18: The existing IsolationGroupOrch and IsolationGroup classes in orchagent/isolationgrouporch.h already provide nearly identical functionality.

The only meaningful difference is that ShlIsolationGroup omits the m_type field (since it's always BRIDGE_PORT) and adds VTEP-specific bind logic that resolves tunnel port names via VxlanTunnelOrch.

** Recommendation: ** ShlIsolationGroup should extend or compose IsolationGroup rather than duplicating ~400 lines of nearly identical SAI interaction code. This would also avoid divergence in bug fixes.

orchagent/shlorch.h
selva-nexthop: we can ignore this comment if we are considering to extend ShlIsolationGroup from IsolationGroup.

orchagent/shlorch.cpp
```diff
@@ -0,0 +1,604 @@
...
+    for (auto vtep: addVtepList) {
+        auto grp = m_isolationGrps.find(vtep);
+        if (grp == m_isolationGrps.end()) {
+            auto grp = make_shared<ShlIsolationGroup>(vtep);
+            status = grp->create();
+            if (SHL_ISO_GRP_STATUS_SUCCESS != status)
+            {
+                return status;
+            }
+            status = grp->addMember(port);
```
selva-nexthop, line 183: If addMember() fails after create() succeeds, the function returns immediately. The shared_ptr<ShlIsolationGroup> goes out of scope without ever being stored in m_isolationGrps, and the SAI isolation group object (created via sai_isolation_group_api->create_isolation_group) is never cleaned up. The same pattern applies if bind() fails — although its return value is entirely ignored.

** Recommendation: ** On any partial failure, call grp->destroy() before returning. Always check bind()'s return value. Consider a RAII wrapper or storing the group in the map immediately after create() and rolling back on subsequent failure.

orchagent/shlorch.cpp
```diff
@@ -0,0 +1,604 @@
...
+    {
+        SWSS_LOG_ERROR("Invalid table %s", table_name.c_str());
+    }
+}
+
+void
+ShlOrch::doShlTblTask(Consumer &consumer)
+{
+    SWSS_LOG_ENTER();
+    shl_isolation_group_status_t status = SHL_ISO_GRP_STATUS_SUCCESS;
```
selva-nexthop, line 82: The status variable is initialized to SUCCESS and never updated from the return values of addMemberToIsolationGroupPerVtep() or delMemberFromIsolationGroupPerVtep(). The retry branch (it++) is unreachable dead code. Furthermore, none of the called functions ever return SHL_ISO_GRP_STATUS_RETRY — the enum value exists but is never produced.

This means all entries are unconditionally erased from m_toSync regardless of whether the SAI operations succeeded or failed, making the system silently lose failed operations.

Recommendation: Capture return values and implement actual retry semantics, or remove the retry infrastructure if retry is not intended.

orchagent/shlorch.cpp
selva-nexthop, line 1: The HLD explicitly states ShlOrch must:

"Interact with EvpnMhOrch to check if the given access interface is associated with ES or not"
"Handle interface ES assignment and removal"

The PR's shlorch.cpp has zero references to EvpnMhOrch or gEvpnMhOrch. There is no validation that the port being added as an isolation group member is actually part of an Ethernet Segment. This means split-horizon groups could be applied to non-ES ports if the APP_DB table gets stale or incorrect data, violating the EVPN MH invariant.

Recommendation: Add a guard in doShlTblTask to verify the port is ES-associated via gEvpnMhOrch->isPortInterfaceAssociatedToEs() before processing.

orchagent/shlorch.cpp
```diff
@@ -0,0 +1,604 @@
...
+
+    auto it = consumer.m_toSync.begin();
+    while (it != consumer.m_toSync.end())
+    {
+        KeyOpFieldsValuesTuple t = it->second;
+
+        /* format: <VLAN_name>:<Ifname> */
+        vector<string> keys = tokenize(kfvKey(t), ':', 1);
```
selva-nexthop, line 90: The VLAN portion of the key is parsed and discarded. There is also no bounds check — if the key contains no : delimiter, keys will cause an out-of-bounds access.

Recommendation: Either use the VLAN name (for scoping isolation groups per-VLAN if needed) or validate key format and fail gracefully. At minimum, add if (keys.size() < 2) { SWSS_LOG_ERROR(...); continue; }.

orchagent/shlorch.cpp
```diff
@@ -0,0 +1,604 @@
...
+extern PortsOrch *gPortsOrch;
+extern sai_isolation_group_api_t*  sai_isolation_group_api;
+extern sai_bridge_api_t *sai_bridge_api;
+extern ShlOrch *gShlOrch;
+extern Directory<Orch*> gDirectory;
+
+ShlOrch::ShlOrch(vector<TableConnector> &connectors) : Orch(connectors)
+{
+    SWSS_LOG_ENTER();
+    gPortsOrch->attach(this);
```
selva-nexthop, line 18: The constructor attaches to PortsOrch, but the destructor is empty.

If ShlOrch is destroyed while PortsOrch still exists (e.g., during shutdown), PortsOrch will hold a dangling observer pointer. Additionally, the SAI isolation groups in m_isolationGrps are never cleaned up in the destructor.

Recommendation: Add gPortsOrch->detach(this) and iterate m_isolationGrps to call destroy() on each group.

orchagent/shlorch.cpp
selva-nexthop, line 1: There is no warm-boot handling anywhere in the code — no WarmStart checks, no reconciliation logic, no restore path. During a warm reboot, all SAI isolation group state will be lost and not re-programmed until FRR re-populates the EVPN_SPLIT_HORIZON_TABLE. This creates a window where split-horizon filtering is not enforced, potentially causing BUM traffic loops in the EVPN MH fabric.

Recommendation: Implement reconciliation similar to VxlanMgr::waitTillReadyToReconcile() / restoreVxlanNetDevices() pattern. At minimum, document this as a known limitation.

orchagent/shlorch.cpp
```diff
@@ -0,0 +1,604 @@
...
+
+    // Remove all bindings
+    attr.value.oid = SAI_NULL_OBJECT_ID;
+    for (auto p : m_bind_ports)
+    {
+        Port port;
+        gPortsOrch->getPort(p, port);
+
+        attr.id = SAI_BRIDGE_PORT_ATTR_ISOLATION_GROUP;
+        if (SAI_STATUS_SUCCESS != sai_bridge_api->set_bridge_port_attribute(port.m_bridge_port_id, &attr))
```
selva-nexthop, line 390: The destroy() method logs errors when SAI unbind/remove calls fail, but continues execution and always returns SHL_ISO_GRP_STATUS_SUCCESS. Callers (like delMemberFromIsolationGroupPerVtep) trust this return value to decide whether to erase the group from the map.

This can leave orphaned SAI objects that are removed from the ShlOrch's tracking but still exist in hardware.

Recommendation: Return SHL_ISO_GRP_STATUS_FAIL on SAI errors and let the caller decide whether to retry or proceed.

orchagent/shlorch.cpp
```diff
@@ -0,0 +1,604 @@
...
+                            m_oid,
+                            kv.first.c_str());
+        }
+    }
+    m_members.clear();
+
+    sai_status_t status = sai_isolation_group_api->remove_isolation_group(m_oid);
+    if (SAI_STATUS_SUCCESS != status)
+    {
+        SWSS_LOG_ERROR("Unable to delete issolation group %s with oid 0x%" PRIx64, m_name.c_str(), m_oid);
```
selva-nexthop, line 427: typo:  issolation should be isolation

orchagent/shlorch.cpp
```diff
@@ -0,0 +1,604 @@
...
+        else
+        {
+            m_bind_ports.push_back(port.m_alias);
+            SWSS_LOG_NOTICE("Bind function, bind port = %s, saved in m_bind_ports for iso grp: %s", port.m_alias.c_str(), m_name.c_str());
+        }
+    }
+    else
+    {
+        m_pending_bind_ports.push_back(port.m_alias);
+        SWSS_LOG_NOTICE("Port %s saved in pending bind ports for isolation group %s\n",
```
selva-nexthop, line 508: \n is not required in SWSS_LOG_XXXX

selva-nexthop (changes requested):

jayakj:
merged as part of https://github.com/sonic-net/SONiC/pull/2195

PR #3771, opened 2025-07-17
EVPNMhOrch changes for EVPN MH feature

jayakj:
**What I did**
Implementation of EVPNMH orch from EVPN VXLAN Multihoming HLD.
https://github.com/sonic-net/SONiC/blob/master/doc/vxlan/EVPN/EVPN_VxLAN_Multihoming.md#339-evpnmhorch

**Why I did it**

**How I verified it**
By SWSS UT

**Details if related**

### DIFF
--- a/orchagent/evpnmhorch.cpp
+++ b/orchagent/evpnmhorch.cpp
@@ -1,0 +1,324 @@
+#include "evpnmhorch.h"
+
+#include "portsorch.h"
+
+extern PortsOrch *gPortsOrch;
+
+extern sai_vlan_api_t *sai_vlan_api;
+
+#define VLAN_PREFIX "Vlan"
+
+EvpnMhOrch::EvpnMhOrch(vector<TableConnector> &connectors) : Orch(connectors)
+{
+    SWSS_LOG_ENTER();
+}
+
+EvpnMhOrch::~EvpnMhOrch()
+{
+    SWSS_LOG_ENTER();
+}
+
+struct EsCacheEntry *EvpnMhOrch::getEsCache(const std::string &key)
+{
+    std::map<std::string, struct EsCacheEntry *>::iterator entry_it = m_esDataMap.find(key);
+    struct EsCacheEntry *entry = nullptr;
+
+    if (entry_it == m_esDataMap.end())
+    {
+        entry = nullptr;
+    }
+    else
+    {
+        entry = (*entry_it).second;
+    }
+
+    return entry;
+}
+
+struct EsCacheEntry *EvpnMhOrch::getEsCacheForPort(const std::string &key)
+{
+    std::map<std::string, struct EsCacheEntry *>::iterator entry_it = m_esDataMap.begin();
+    struct EsCacheEntry *entry = nullptr;
+
+    while (entry_it != m_esDataMap.end() && entry == nullptr)
+    {
+        if (((*entry_it).first).find(key) != std::string::npos)
+        {
+            entry = (*entry_it).second;
+        }
+
+        entry_it++;
+    }
+
+    return entry;
+}
+
+std::string getPortFromEsKey(string &key)
+{
+    std::string ret_port;
+    std::size_t found = key.find(":");
+
+    if (found != std::string::npos)
+    {
+        ret_port = key.substr(found + 1, std::string::npos);
+    }
+    else
+    {
+        ret_port = "Unknown";
+    }
+
+    return ret_port;
+}
+
+std::string getVlanFromEsKey(string &key)
+{
+    std::string ret_vlan;
+    std::size_t found = key.find(":");
+
+    if (found != std::string::npos)
+    {
+        ret_vlan = key.substr(0, found);
+    }
+    else
+    {
+        ret_vlan = "Unknown";
+    }
+
+    return ret_vlan;
+}
+
+void EvpnMhOrch::updateEsCache(string &key, KeyOpFieldsValuesTuple &t)
+{
+    bool is_df = false;
+    struct EsCacheEntry *existing_entry = nullptr;
+
+    for (auto i : kfvFieldsValues(t))
+    {
+        if (fvField(i) == "df")
+        {
+            is_df = (fvValue(i) == "true");
+        }
+    }
+
+    existing_entry = getEsCache(key);
+    if (existing_entry)
+    {
+        existing_entry->is_df = is_df;
+    }
+    else
+    {
+        existing_entry = new EsCacheEntry(is_df);
+        m_esDataMap[key] = existing_entry;
+    }
+
+    if (existing_entry)
+    {
+        /*
+         * This is going to get slow at scale as each vlan in the bridge
+         * will trigger an attribute update against the main port.
+         *
+         * Might need to add a parent/child cache relationship to improve performance.
+         */
+        std::string port_name = getPortFromEsKey(key);
+        std::string vlan_id = getVlanFromEsKey(key);
+        Port port;
+        sai_object_id_t vlan_member_id;
+
+        SWSS_LOG_NOTICE("updateEsCache: SET oper: %s, vlan: %s, port_name: %s, is_df: %d", key.c_str(), vlan_id.c_str(), port_name.c_str(), existing_entry->is_df);
+
+        if (!gPortsOrch->getPort(vlan_id, port))
+        {
+            SWSS_LOG_ERROR("updateEsCache: interface: %s, Vlan is not not yet created, returning", key.c_str());
+            return;
+        }
+
+        if (gPortsOrch->getVlanMember(port_name, port, vlan_member_id))
+        {
+            /* TODO: Use proper attribute once its available in SAI */
+            sai_attribute_t attr;
+            attr.id = SAI_VLAN_MEMBER_ATTR_TUNNEL_TERM_BUM_TX_DROP;
+            attr.value.booldata = existing_entry->is_df;
+
+            auto status = sai_vlan_api->set_vlan_member_attribute(vlan_member_id, &attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                /* TODO: Error handling */
+            }
+        }
+        else
+        {
+            SWSS_LOG_ERROR("updateEsCache: interface %s vlan_member_id doesnt exit", key.c_str());
+            return;
+        }
+    }
+}
+
+void EvpnMhOrch::deleteEsCache(string &key)
+{
+    EsCacheEntry *entry = getEsCache(key);
+
+    if (entry)
+    {
+        SWSS_LOG_NOTICE("deleteEsCache: DEL oper: intf: %s, is_df: %d", key.c_str(), entry->is_df);
+        std::string port_name = getPortFromEsKey(key);
+        std::string vlan_id = getVlanFromEsKey(key);
+        Port port;
+        sai_object_id_t vlan_member_id;
+
+        if (!gPortsOrch->getPort(vlan_id, port))
+        {
+            SWSS_LOG_ERROR("deleteEsCache: interface: %s, Vlan is not not yet created, returning", key.c_str());
+            return;
+        }
+        if (gPortsOrch->getVlanMember(port_name, port, vlan_member_id))
+        {
+            /* TODO: Use proper attribute once its available in SAI */
+            sai_attribute_t attr;
+            attr.id = SAI_VLAN_MEMBER_ATTR_TUNNEL_TERM_BUM_TX_DROP;
+            attr.value.booldata = false;
+
+            auto status = sai_vlan_api->set_vlan_member_attribute(vlan_member_id, &attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                /* TODO: Error handling */
+            }
+        }
+        else
+        {
+            SWSS_LOG_ERROR("deleteEsCache: interface %s vlan_member_id doesnt exit\n", key.c_str());
+            return;
+        }
+        m_esDataMap.erase(key);
+        delete entry;
+        entry = nullptr;
+    }
+}
+
+void EvpnMhOrch::vlanMembersApplyNonDF(string port_name)
+{
+    vlan_members_t vlan_members;
+    Port port;
+    if (!gPortsOrch->getPort(port_name, port))
+    {
+        SWSS_LOG_ERROR("vlanMembersApplyNonDF: getPort() fails for port_name:%s", port_name.c_str());
+        return;
+    }
+    gPortsOrch->getPortVlanMembers(port, vlan_members);
+    for (const auto &member : vlan_members)
+    {
+        auto vlan_id = member.first;
+        auto vlan_mem_entry = member.second;
+        /* TODO: Use proper attribute once its available in SAI */
+        sai_attribute_t attr;
+        attr.id = SAI_VLAN_MEMBER_ATTR_TUNNEL_TERM_BUM_TX_DROP;
+        attr.value.booldata = isInterfaceDF(port_name, vlan_id);
+        SWSS_LOG_NOTICE("vlanMembersApplyNonDF: set Non-DF for port: %s, vlan: %d", port_name.c_str(), vlan_id);
+
+        auto status = sai_vlan_api->set_vlan_member_attribute(vlan_mem_entry.vlan_member_id, &attr);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            /* TODO: Error handling */
+        }
+    }
+    return;
+}
+void EvpnMhOrch::doEvpnEsIntfTask(Consumer &consumer)
+{
+    auto it = consumer.m_toSync.begin();
+
+    while (it != consumer.m_toSync.end())
+    {
+        KeyOpFieldsValuesTuple t = it->second;
+        string key = kfvKey(t);
+        string op = kfvOp(t);
+
+        if (op == SET_COMMAND)
+        {
+            SWSS_LOG_NOTICE("doEvpnEsIntfTask: SET oper: ESI intf: %s", key.c_str());
+            m_esIntfMap[key] = true;
+        }
+        else if (op == DEL_COMMAND)
+        {
+            SWSS_LOG_NOTICE("doEvpnEsIntfTask: DEL oper: ESI intf: %s", key.c_str());
+            m_esIntfMap.erase(key);
+        }
+        vlanMembersApplyNonDF(key);
+        it = consumer.m_toSync.erase(it);
+    }
+}
+
+void EvpnMhOrch::doEvpnEsDfTask(Consumer &consumer)
+{
+    auto it = consumer.m_toSync.begin();
+
+    while (it != consumer.m_toSync.end())
+    {
+        KeyOpFieldsValuesTuple t = it->second;
+        string key = kfvKey(t);
+        string op = kfvOp(t);
+
+        if (op == SET_COMMAND)
+        {
+            updateEsCache(key, t);
+        }
+        else if (op == DEL_COMMAND)
+        {
+            deleteEsCache(key);
+        }
+        it = consumer.m_toSync.erase(it);
+    }
+}
+
+void EvpnMhOrch::doTask(Consumer &consumer)
+{
+    SWSS_LOG_ENTER();
+
+    string table_name = consumer.getTableName();
+    if (table_name == "EVPN_DF_TABLE")
+    {
+        doEvpnEsDfTask(consumer);
+    }
+    else
+    {
+        if (table_name == "EVPN_ETHERNET_SEGMENT")
+        {
+            doEvpnEsIntfTask(consumer);
+        }
+    }
+}
+
+bool EvpnMhOrch::isInterfaceDF(const std::string port_name, const sai_vlan_id_t vlan_id)
+{
+    bool is_df = false;
+    string df_key = VLAN_PREFIX + to_string(vlan_id) + ":" + port_name;
+    EsCacheEntry *entry = getEsCache(df_key);
+
+    if (entry)
+    {
+        is_df = entry->is_df;
+    }
+
+    return (is_df);
+}
+
+bool EvpnMhOrch::isPortAndVlanAssociatedToEs(const std::string port_name, const sai_vlan_id_t vlan_id)
+{
+    if (isPortInterfaceAssociatedToEs(port_name))
+    {
+        return true;
+    }
+    string df_key = VLAN_PREFIX + to_string(vlan_id) + ":" + port_name;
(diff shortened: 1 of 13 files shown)


orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+    SWSS_LOG_ENTER();
+}
+
+EvpnMhOrch::~EvpnMhOrch()
+{
+    SWSS_LOG_ENTER();
+}
+
+struct EsCacheEntry *EvpnMhOrch::getEsCache(const std::string &key)
+{
```
selva-nexthop, line 22: this implementation can be simplified to 
```
struct EsCacheEntry *EvpnMhOrch::getEsCache(const std::string &key)
{
    auto entry_it = m_esDataMap.find(key);
    return (entry_it != m_esDataMap.end()) ? entry_it->second : nullptr;
}
```

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+    }
+    else
+    {
+        entry = (*entry_it).second;
+    }
+
+    return entry;
+}
+
+struct EsCacheEntry *EvpnMhOrch::getEsCacheForPort(const std::string &key)
```
selva-nexthop, line 38: if the key is Vlan<id>:<portname>, there can be multiple entries using same port name with different vlan ids.  In this lookup will just return the first matching entry.

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+            entry = (*entry_it).second;
+        }
+
+        entry_it++;
+    }
+
+    return entry;
+}
+
+std::string getPortFromEsKey(string &key)
```
selva-nexthop, line 56: can be optimized to 
```
std::string getPortFromEsKey(const std::string &key)
{
    auto pos = key.find(':');
    return (pos != std::string::npos) ? key.substr(pos + 1) : "Unknown";
}
```

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+    {
+        ret_port = "Unknown";
+    }
+
+    return ret_port;
+}
+
+std::string getVlanFromEsKey(string &key)
+{
+    std::string ret_vlan;
```
selva-nexthop, line 75: can be optimized to;
```
std::string getVlanFromEsKey(const std::string &key)
{
    auto pos = key.find(':');
    return (pos != std::string::npos) ? key.substr(0, pos) : "Unknown";
}
```

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+    else
+    {
+        existing_entry = new EsCacheEntry(is_df);
+        m_esDataMap[key] = existing_entry;
+    }
+
+    if (existing_entry)
+    {
+        /*
+         * This is going to get slow at scale as each vlan in the bridge
```
selva-nexthop, line 117: critical: the yang takes only the interface name.  But here the implementation is based on vlan_id + interfacename.  If we are considering vlan_id + interfacename, then it is better to consider different data structures for better lookups.

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+void EvpnMhOrch::updateEsCache(string &key, KeyOpFieldsValuesTuple &t)
+{
+    bool is_df = false;
+    struct EsCacheEntry *existing_entry = nullptr;
+
+    for (auto i : kfvFieldsValues(t))
+    {
+        if (fvField(i) == "df")
+        {
+            is_df = (fvValue(i) == "true");
```
selva-nexthop, line 99: can we break here when "df" is found?

selva-nexthop: can we think of better DS instead of KeyOpFieldsValuesTuple for quick lookup?

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+    if (entry)
+    {
+        is_df = entry->is_df;
+    }
+
+    return (is_df);
+}
+
+bool EvpnMhOrch::isPortAndVlanAssociatedToEs(const std::string port_name, const sai_vlan_id_t vlan_id)
```
selva-nexthop, line 304: can be optimized to;
bool EvpnMhOrch::isPortAndVlanAssociatedToEs(const std::string &port_name, sai_vlan_id_t vlan_id)
{
    if (isPortInterfaceAssociatedToEs(port_name))
        return true;

    std::string df_key = VLAN_PREFIX + std::to_string(vlan_id) + ":" + port_name;
    return getEsCache(df_key) != nullptr;
}

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+    else
+    {
+        if (table_name == "EVPN_ETHERNET_SEGMENT")
+        {
+            doEvpnEsIntfTask(consumer);
+        }
+    }
+}
+
+bool EvpnMhOrch::isInterfaceDF(const std::string port_name, const sai_vlan_id_t vlan_id)
```
selva-nexthop, line 290: can be optimized to;
```
bool EvpnMhOrch::isInterfaceDF(const std::string &port_name, sai_vlan_id_t vlan_id)
{
    std::string df_key = VLAN_PREFIX + std::to_string(vlan_id) + ":" + port_name;
    if (EsCacheEntry *entry = getEsCache(df_key))
        return entry->is_df;
    return false;
}
```

orchagent/evpnmhorch.h
```diff
@@ -0,0 +1,55 @@
...
+    ~EvpnMhOrch();
+
+    bool isPortInterfaceAssociatedToEs(const std::string port_name);
+    bool isPortAndVlanAssociatedToEs(const std::string port_name, const sai_vlan_id_t vlan_id);
+    bool isInterfaceDF(const std::string port_name, const sai_vlan_id_t vlan_id);
+
+private:
+    vector<Table *> m_appTables;
+
+    std::map<std::string, struct EsCacheEntry *> m_esDataMap;
```
selva-nexthop, line 40: This performs a full linear scan of the entire m_esDataMap, doing string::find() (substring match) on every key. At scale with thousands of VLAN-member entries, this is O(n) per call. 

Recommendation: Adopt a two-level data structure: std::unordered_map<std::string/*port*/, std::unordered_map<sai_vlan_id_t, EsCacheEntry>>. This gives O(1) lookups for both per-port and per-port-per-VLAN queries, and eliminates the dangerous substring match that could produce false positives (e.g., port "Ethernet1" matching "Ethernet10").

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+        return;
+    }
+    gPortsOrch->getPortVlanMembers(port, vlan_members);
+    for (const auto &member : vlan_members)
+    {
+        auto vlan_id = member.first;
+        auto vlan_mem_entry = member.second;
+        /* TODO: Use proper attribute once its available in SAI */
+        sai_attribute_t attr;
+        attr.id = SAI_VLAN_MEMBER_ATTR_TUNNEL_TERM_BUM_TX_DROP;
```
apannerselva, line 213: This seems to be workaround, shouldn't we use SAI_BRIDGE_PORT_ATTR_TUNNEL_TERM_BUM_TX_DROP  as mentioned in the HLD and stop iterating per VLAN.

tests/mock_tests/evpnmhorch_ut.cpp
```diff
@@ -0,0 +1,461 @@
...
+        });
+
+        std::deque<KeyOpFieldsValuesTuple> entries;
+        entries.clear();
+        entries.push_back({ "Vlan10:Ethernet1", "SET", { { "df", "true" } } });
+        consumer_df->addToSync(entries);
+        static_cast<Orch *>(gEvpnMhOrch)->doTask();
+
+        ASSERT_TRUE(gEvpnMhOrch->isInterfaceDF("Ethernet1", 10));
+        ASSERT_EQ(df_attr_received, SAI_VLAN_MEMBER_ATTR_TUNNEL_TERM_BUM_TX_DROP);
```
apannerselva, line 456: SAI_BRIDGE_PORT_ATTR_TUNNEL_TERM_BUM_TX_DROP  should be used here.

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+        attr.value.booldata = isInterfaceDF(port_name, vlan_id);
+        SWSS_LOG_NOTICE("vlanMembersApplyNonDF: set Non-DF for port: %s, vlan: %d", port_name.c_str(), vlan_id);
+
+        auto status = sai_vlan_api->set_vlan_member_attribute(vlan_mem_entry.vlan_member_id, &attr);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            /* TODO: Error handling */
```
selva-nexthop, line 220: There are four /* TODO: Error handling */ comments scattered through the SAI call sites in updateEsCache(), deleteEsCache(), and vlanMembersApplyNonDF():

```
auto status = sai_vlan_api->set_vlan_member_attribute(vlan_member_id, &attr);
{
    /* TODO: Error handling */
}
```
These are not just cosmetic — a SAI failure during DF programming means BUM traffic forwarding state is silently wrong, which can cause traffic duplication or black-holing in an EVPN MH environment.

Recommendation: At minimum, log SWSS_LOG_ERROR with the SAI status and return a failure indicator. 

For production readiness, implement retry logic by leaving the entry in m_toSync (return without erasing), consistent with the standard orchagent retry pattern.

selva-nexthop:
While checking the HLD vs PR changes, changes related EVPN_MH_GLOBAL and LAG_TABLE subscriptions are not available in this PR. 


A new EvpnMhOrch class will be introduced to perform following activities,
- Subscribe to EVPN_MH_GLOBAL config table and set the SAI switch attribute.
**[No relevant changes found]**
- Subscribe to EVPN_ETHERNET_SEGMENT config table updates and maintain internal cache.
[Available]
- Provide set of API to check if the given acces interface is ES associated or not.
[Available]
- Subscribe to LAG_TABLE oper status changes and trigger MAC update requests to FdbOrch.
**[No relevant changes found]**
- Subscribe to EVPN_DF_TABLE from APP-DB and set the SAI bridge port attribute SAI_BRIDGE_PORT_ATTR_TUNNEL_TERM_BUM_TX_DROP to relevant Bridgeport.

orchagent/portsorch.cpp
```diff
@@ -6494,6 +6507,14 @@ bool PortsOrch::addBridgePort(Port &port)
     attr.value.s32 = port.m_learn_mode;
     attrs.push_back(attr);
 
+    /* If the interface is associated with an ethernet segment, set the DF role */
+    if (gEvpnMhOrch->isPortInterfaceAssociatedToEs(port.m_alias)) {
+        /* TODO FIXME: sridsant : Use proper attribute once its available in SAI */
+        attr.id = SAI_BRIDGE_PORT_ATTR_EGRESS_FILTERING;
```
selva-nexthop, line 6513: **Critical**: This must be resolved before merge. Use SAI_BRIDGE_PORT_ATTR_TUNNEL_TERM_BUM_TX_DROP (or the community-agreed attribute) and query the correct VLAN context.  
SAI_BRIDGE_PORT_ATTR_EGRESS_FILTERING is a standard 802.1Q egress filtering attribute and has nothing to do with DF election or BUM TX drop.

orchagent/orchdaemon.cpp
```diff
@@ -468,8 +478,8 @@ bool OrchDaemon::init()
      * when iterating ConsumerMap. This is ensured implicitly by the order of keys in ordered map.
      * For cases when Orch has to process tables in specific order, like PortsOrch during warm start, it has to override Orch::doTask()
      */
-    m_orchList = { gSwitchOrch, gCrmOrch, gPortsOrch, gBufferOrch, gFlowCounterRouteOrch, gIntfsOrch, gNeighOrch, gNhgMapOrch, gNhgOrch, gCbfNhgOrch, gFgNhgOrch, gRouteOrch, gCoppOrch, gQosOrch, wm_orch, gPolicerOrch, gTunneldecapOrch, sflow_orch, gDebugCounterOrch, gMacsecOrch, bgp_global_state_orch, gBfdOrch, gIcmpOrch, gSrv6Orch, gMuxOrch, mux_cb_orch, gMonitorOrch, gStpOrch};
 
+    m_orchList = { gSwitchOrch, gCrmOrch, gPortsOrch, gBufferOrch, gFlowCounterRouteOrch, gIntfsOrch, gNeighOrch, gNhgMapOrch, gNhgOrch, gCbfNhgOrch, gFgNhgOrch, gRouteOrch, gCoppOrch, gQosOrch, wm_orch, gPolicerOrch, gTunneldecapOrch, sflow_orch, gDebugCounterOrch, gMacsecOrch, bgp_global_state_orch, gBfdOrch, gIcmpOrch, gSrv6Orch, gMuxOrch, mux_cb_orch, gMonitorOrch, gStpOrch, gEvpnMhOrch};
```
selva-nexthop, line 482: PortsOrch - to - EvpnMhOrch Ordering Race:

PortsOrch::addBridgePort() and PortsOrch::addVlanMember() now directly call gEvpnMhOrch->isPortInterfaceAssociatedToEs() and gEvpnMhOrch->isInterfaceDF(). This creates a startup ordering dependency:

gEvpnMhOrch must be constructed before gPortsOrch processes any ports.

In orchdaemon.cpp, gEvpnMhOrch is created after gPortsOrch and placed at the end of m_orchList. During warm boot, gPortsOrch processes ports first, calling into an gEvpnMhOrch whose EVPN_ETHERNET_SEGMENT data hasn't been consumed yet.

** Recommendation: ** Move gEvpnMhOrch creation before gPortsOrch in OrchDaemon::init(), or add a null-check guard in PortsOrch before calling EvpnMhOrch APIs. Additionally, consider the warm-boot reconciliation sequence — ES state must be restored before port/VLAN member state.

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+            SWSS_LOG_NOTICE("doEvpnEsIntfTask: SET oper: ESI intf: %s", key.c_str());
+            m_esIntfMap[key] = true;
+        }
+        else if (op == DEL_COMMAND)
+        {
+            SWSS_LOG_NOTICE("doEvpnEsIntfTask: DEL oper: ESI intf: %s", key.c_str());
+            m_esIntfMap.erase(key);
+        }
+        vlanMembersApplyNonDF(key);
+        it = consumer.m_toSync.erase(it);
```
selva-nexthop, line 246: Even when vlanMembersApplyNonDF() encounters failures (port not found, VLAN member not found, SAI failure), the entry is unconditionally erased from m_toSync. This means the task is lost and will never be retried.

**Recommendation: ** Check return values and only erase on success. On failure, increment the iterator to leave the entry for retry on the next doTask() cycle.

orchagent/evpnmhorch.cpp
selva-nexthop, line 1: ** Coding Style and Minor issues **

- const correctness: isInterfaceDF(), isPortInterfaceAssociatedToEs(), and isPortAndVlanAssociatedToEs() take std::string by value instead of const std::string &. This creates unnecessary copies on every call.

- #define private public in unit tests: evpnmhorch_ut.cpp uses #define private public to access internal members. This is a known SONiC test pattern but is technically undefined behavior in C++.

- Unused member: vector<Table *> m_appTables in the header is declared but never used.

- Helper functions in global scope: getPortFromEsKey() and getVlanFromEsKey() are free functions in the .cpp file but not declared static, potentially causing linker conflicts.

- Redundant null check: After updateEsCache() sets existing_entry, the outer if (existing_entry) is always true but adds visual noise.

orchagent/portsorch.cpp
```diff
@@ -6494,6 +6507,14 @@ bool PortsOrch::addBridgePort(Port &port)
...
+        attr.value.booldata = gEvpnMhOrch->isInterfaceDF(port.m_alias, DEFAULT_PORT_VLAN_ID);
```
vganesan-nokia, line 6514: Should negate the output of `gEvpnMhOrch->isInterfaceDF()`

orchagent/portsorch.cpp
```diff
@@ -6786,6 +6807,14 @@ bool PortsOrch::addVlanMember(Port &vlan, Port &port, string &tagging_mode, stri
     attr.value.s32 = sai_tagging_mode;
     attrs.push_back(attr);
 
+    /* If the interface is associated with an ethernet segment, send the SAI vlan DF attribute */
+    if (gEvpnMhOrch->isPortAndVlanAssociatedToEs(port.m_alias, vlan.m_vlan_info.vlan_id)) {
+        /* TODO: Use proper attribute once its available in SAI */
+        attr.id = SAI_VLAN_MEMBER_ATTR_TUNNEL_TERM_BUM_TX_DROP;
+        attr.value.booldata = gEvpnMhOrch->isInterfaceDF(port.m_alias, vlan.m_vlan_info.vlan_id);
```
vganesan-nokia, line 6814: Should set to "true" when non-DF. So should negate the output of `gEvpnMhOrch->isInterfaceDF()`

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+            SWSS_LOG_ERROR("updateEsCache: interface: %s, Vlan is not not yet created, returning", key.c_str());
+            return;
+        }
+
+        if (gPortsOrch->getVlanMember(port_name, port, vlan_member_id))
+        {
+            /* TODO: Use proper attribute once its available in SAI */
+            sai_attribute_t attr;
+            attr.id = SAI_VLAN_MEMBER_ATTR_TUNNEL_TERM_BUM_TX_DROP;
+            attr.value.booldata = existing_entry->is_df;
```
vganesan-nokia, line 140: Should be set to true when non-DF. Invert `existing_entry->is_df`?

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+        {
+            SWSS_LOG_ERROR("updateEsCache: interface: %s, Vlan is not not yet created, returning", key.c_str());
+            return;
+        }
+
+        if (gPortsOrch->getVlanMember(port_name, port, vlan_member_id))
+        {
+            /* TODO: Use proper attribute once its available in SAI */
+            sai_attribute_t attr;
+            attr.id = SAI_VLAN_MEMBER_ATTR_TUNNEL_TERM_BUM_TX_DROP;
```
apannerselva, line 139: This seems to be workaround, shouldn't we use SAI_BRIDGE_PORT_ATTR_TUNNEL_TERM_BUM_TX_DROP as mentioned in the HLD.

orchagent/portsorch.cpp
apannerselva, line 6513: SAI_BRIDGE_PORT_ATTR_EGRESS_FILTERING SAI attribute is not mentioned in the HLD. Please check

orchagent/evpnmhorch.h
apannerselva, line 40: EVPN DF election is Ethernet Segment scoped, not VLAN scoped.
Modeling DF state as Vlan:Port is wrong. 
One ES can program/call SAI multiple times to block the BUM traffic. 


struct EsDataEntry
{
    bool is_df;
    std::set<std::string> ports;
};

std::map<std::string /*ES key*/, EsDataEntry> m_esDataMap;

orchagent/evpnmhorch.cpp
```diff
@@ -0,0 +1,324 @@
...
+        else if (op == DEL_COMMAND)
+        {
+            deleteEsCache(key);
+        }
+        it = consumer.m_toSync.erase(it);
+    }
+}
+
+void EvpnMhOrch::doTask(Consumer &consumer)
+{
```
vganesan-nokia, line 273: Processing missing for EVPN_ES_BACKUP_NHG_TABLE entries. fpmsyncd writes this for backup up nhg message from zebra.

orchagent/evpnmhorch.cpp
pbrisset: done

orchagent/evpnmhorch.cpp
pbrisset: done

orchagent/evpnmhorch.cpp
pbrisset: Added a comment documenting that limitation and simplified the implementation using a range-based for loop. The behavior remains the same - returning the first match - but now it's documented and more readable

orchagent/evpnmhorch.cpp
pbrisset: done

orchagent/evpnmhorch.cpp
pbrisset: done

orchagent/evpnmhorch.cpp
pbrisset: Done. Added the break statement to exit the loop once "df" is found, and also changed to const auto & to avoid unnecessary copies. Added a comment explaining that [KeyOpFieldsValuesTuple] is the standard swss framework type for Redis table data and is part of the API contract, so it can't be easily changed.

orchagent/evpnmhorch.cpp
pbrisset: Added a comprehensive comment documenting the data structure design issue

orchagent/evpnmhorch.cpp
pbrisset: done

orchagent/evpnmhorch.cpp
pbrisset: fixed

orchagent/evpnmhorch.cpp
pbrisset: EVPN_ES_BACKUP_NHG_TABLE code is not available here yet. Coming in another PR.

orchagent/evpnmhorch.cpp
pbrisset: done

orchagent/evpnmhorch.cpp
pbrisset: done

orchagent/evpnmhorch.h
pbrisset: I agree, that is an optimization to come in a different PR.

orchagent/orchdaemon.cpp
pbrisset: done..good catch

jayakj:

PR #3776, opened 2025-07-22
[ssw][haorch] use kfv instead of serialized pb objects for internal tables

zjswhhh:
**What I did**
1. Adapt to recent sonic-dash-api changes
2. instead of expecting hamgrd to write serialized pb object to DPU_APPL_DB, now haorch will expect plain kfv. 

**Why I did it**
Otherwise integrated tests won't work. 

**How I verified it**
1. UTs.
2. Load binary to DPU (pending). 

**Details if related**

### DIFF
--- a/orchagent/dash/dashhaorch.cpp
+++ b/orchagent/dash/dashhaorch.cpp
@@ -9,6 +9,7 @@
 #include "table.h"
 #include "taskworker.h" 
 #include "pbutils.h"
+#include "converter.h"
 
 #include "chrono"
 
@@ -309,9 +310,22 @@ void DashHaOrch::doTaskHaSetTable(ConsumerBase &consumer)
         {
             dash::ha_set::HaSet entry;
 
-            if (!parsePbMessage(kfvFieldsValues(tuple), entry))
+            /*
+            * For HA internal tables, kfv format was used instead of serialized pb objects in the end.
+            * I decided to keep protobuf conversion still for:
+            *      - ensuring the data integrity.
+            *      - in case we need to switch to protobuf in the future.
+            */
+            // if (!parsePbMessage(kfvFieldsValues(tuple), entry))
+            // {
+            //     SWSS_LOG_WARN("Requires protobuf at HaSet :%s", key.c_str());
+            //     it = consumer.m_toSync.erase(it);
+            //     continue;
+            // }
+
+            if (!convertKfvToHaSetPb(kfvFieldsValues(tuple), entry))
             {
-                SWSS_LOG_WARN("Requires protobuf at HaSet :%s", key.c_str());
+                SWSS_LOG_WARN("Failed to convert KeyOpFieldsValuesTuple to HaSet entry for %s, invalid values probably.", key.c_str());
                 it = consumer.m_toSync.erase(it);
                 continue;
             }
@@ -426,7 +440,7 @@ bool DashHaOrch::addHaScopeEntry(const std::string &key, const dash::ha_scope::H
     SWSS_LOG_NOTICE("Created HA Scope object for %s", key.c_str());
 
     // set HA Scope ID to ENI
-    if (ha_set_it->second.metadata.scope() == dash::types::HaScope::SCOPE_ENI)
+    if (ha_set_it->second.metadata.scope() == dash::types::HaScope::HA_SCOPE_ENI)
     {
         auto eni_entry = m_dash_orch->getEni(key);
         if (eni_entry == nullptr)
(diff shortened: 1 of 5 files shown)


orchagent/dash/dashhaorch.cpp
```diff
@@ -789,3 +817,189 @@ void DashHaOrch::doTask(NotificationConsumer &consumer)
...
+                SWSS_LOG_ERROR("Invalid IPv6 address %s", value.c_str());
+                return false;
+            }
+            entry.mutable_vip_v6()->set_ipv6(value);
+        }
+        else if (field == "owner")
+        {
+            if (value == "switch")
+            {
+                entry.set_owner(dash::types::HaOwner::HA_OWNER_DPU);
```
yue-fred-gao: is the logic reversed? switch is mapped to dpu?

tests/mock_tests/dashhaorch_ut.cpp
```diff
@@ -74,15 +74,29 @@ namespace dashhaorch_ut
                 new swss::ConsumerStateTable(m_dpu_app_db.get(), APP_DASH_HA_SET_TABLE_NAME, 1, 1),
                 m_dashHaOrch, APP_DASH_HA_SET_TABLE_NAME));
 
-            dash::ha_set::HaSet ha_set = HaSetPbObject();
+            // dash::ha_set::HaSet ha_set = HaSetPbObject();
```
prsunny: please remove this line if not required

tests/mock_tests/dashhaorch_ut.cpp
```diff
@@ -97,8 +111,8 @@ namespace dashhaorch_ut
                 new swss::ConsumerStateTable(m_dpu_app_db.get(), APP_DASH_HA_SET_TABLE_NAME, 1, 1),
                 m_dashHaOrch, APP_DASH_HA_SET_TABLE_NAME));
 
-            dash::ha_set::HaSet ha_set = HaSetPbObject();
-            ha_set.set_scope(dash::types::SCOPE_ENI);
+            // dash::ha_set::HaSet ha_set = HaSetPbObject();
```
prsunny: Same as above and rest of PR

orchagent/dash/dashhaorch.cpp
```diff
@@ -789,3 +817,189 @@ void DashHaOrch::doTask(NotificationConsumer &consumer)
...
+        {
+            entry.set_version(value);
+        }
+        else if (field == "disabled")
+        {
+            entry.set_disabled(value == "true" || value == "1");
+        }
+        else if (field == "ha_role")
+        {
+            if (value == "active")
```
yue-fred-gao: might be extract it as a function from_string that returns dash::types::HaRole and add the function to pbutils along with other conversion functions

orchagent/dash/dashhaorch.cpp
```diff
@@ -789,3 +817,189 @@ void DashHaOrch::doTask(NotificationConsumer &consumer)
...
+            } else if (value == "standalone")
+            {
+                entry.set_ha_role(dash::types::HaRole::HA_ROLE_STANDALONE);
+            } else if (value == "switching_to_active")
+            {
+                entry.set_ha_role(dash::types::HaRole::HA_ROLE_SWITCHING_TO_ACTIVE);
+            }
+            else
+            {
+                SWSS_LOG_ERROR("Unknown HA Scope role %s, setting to dead", value.c_str());
```
yue-fred-gao: I see test maps "" to UNSPECIFIED but here it is to DEAD. Did I miss something?

zjswhhh: "" and UNSPECIFIED are the input in test, mapping to DEAD.

orchagent/dash/dashhaorch.cpp
zjswhhh: addressed

orchagent/dash/dashhaorch.cpp
zjswhhh: addressed

tests/mock_tests/dashhaorch_ut.cpp
zjswhhh: addressed

tests/mock_tests/dashhaorch_ut.cpp
zjswhhh: addressed

orchagent/dash/pbutils.cpp
```diff
@@ -185,16 +185,94 @@ dash::types::HaRole to_pb(const sai_dash_ha_role_t ha_role)
...
+    else if (ha_role == "standalone")
+    {
+        pb_ha_role = dash::types::HA_ROLE_STANDALONE;
+    }
+    else if (ha_role == "switching_to_active")
+    {
+        pb_ha_role = dash::types::HA_ROLE_SWITCHING_TO_ACTIVE;
+    }
+    else
+    {
```
yue-fred-gao, line 227: HA_ROLE_UNSPECIFIED is one of the enum value. We should not say it is unknown. When you pass it to SAI, you can map it to DEAD if SAI doesn't have unspecified ha_role.

zjswhhh: Yes we are mapping to DEAD today. I updated the log message and level a bit for clarification/

saiarcot895 (approved):

PR #3786, opened 2025-07-24
Add Basic 1.6t Support to sonic-swss

bobby-nexthop:
#### Why I did it
Two changes needed in sonic-swss to support 1.6T.
- The `parsePortSpeed` and `parsePortAdvSpeeds` functions validate speeds against min/max values
- Update FLR calculation for 200G SerDes rate

#### How I did it
- Changed max speed to 1.6T
- Added 200G lane_speed to the FLR calculation

#### How to verify it
We cannot yet as the hardware does not exist


prgeor:
@prsunny can you merge?

PR #3790, opened 2025-07-28
[ssw][ha] haorch consumes dpu_ appl_db to create bfd passive sessions on DPU

zjswhhh:
**What I did**
[ssw][ha] haorch consumes dpu_ appl_db to create bfd passive sessions on DPU

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/dash/dashhaorch.cpp
+++ b/orchagent/dash/dashhaorch.cpp
@@ -21,6 +21,8 @@ extern sai_dash_eni_api_t*  sai_dash_eni_api;
 extern sai_object_id_t      gSwitchId;
 extern sai_switch_api_t*    sai_switch_api;
 
+extern BfdOrch *gBfdOrch;
+
 static const map<sai_ha_set_event_t, string> sai_ha_set_event_type_name =
 {
     { SAI_HA_SET_EVENT_DP_CHANNEL_UP, "up" },
@@ -683,6 +685,37 @@ void DashHaOrch::doTaskHaScopeTable(ConsumerBase &consumer)
     }
 }
 
+void DashHaOrch::doTaskBfdSessionTable(ConsumerBase &consumer)
+{
+    SWSS_LOG_ENTER();
+
+    auto it = consumer.m_toSync.begin();
+    while (it != consumer.m_toSync.end())
+    {
+        KeyOpFieldsValuesTuple tuple = it->second;
+        const auto& key = kfvKey(tuple);
+        const auto& op = kfvOp(tuple);
+
+        SWSS_LOG_DEBUG("Processing BFD Session table");
+
+        if (op == SET_COMMAND)
+        {
+            gBfdOrch->createSoftwareBfdSession(key, kfvFieldsValues(tuple));
+            it = consumer.m_toSync.erase(it);
+        }
+        else if (op == DEL_COMMAND)
+        {
+            gBfdOrch->removeSoftwareBfdSession(key);
+            it = consumer.m_toSync.erase(it);
+        }
+        else
+        {
+            SWSS_LOG_ERROR("Unknown operation %s for BFD Session table", op.c_str());
+            it = consumer.m_toSync.erase(it);
+        }
+    }
+}
+
 void DashHaOrch::doTask(ConsumerBase &consumer)
 {
     SWSS_LOG_ENTER();
@@ -694,7 +727,12 @@ void DashHaOrch::doTask(ConsumerBase &consumer)
     else if (consumer.getTableName() == APP_DASH_HA_SCOPE_TABLE_NAME)
     {
         doTaskHaScopeTable(consumer);
-    } else
+    }
+    else if (consumer.getTableName() ==  APP_BFD_SESSION_TABLE_NAME)
+    {
+        doTaskBfdSessionTable(consumer);
+    }
+    else
     {
         SWSS_LOG_ERROR("Unknown table: %s", consumer.getTableName().c_str());
     }
--- a/orchagent/bfdorch.cpp
+++ b/orchagent/bfdorch.cpp
@@ -703,6 +703,18 @@ void BfdOrch::handleTsaStateChange(bool tsaState)
     }
 }
 
+void BfdOrch::createSoftwareBfdSession(const string &key, const vector<swss::FieldValueTuple>& data)
+{
+    m_stateSoftBfdSessionTable->set(createStateDBKey(key), data);
+    SWSS_LOG_NOTICE("Software BFD session created for %s", key.c_str());
+}
+
+void BfdOrch::removeSoftwareBfdSession(const string &key)
+{
+    m_stateSoftBfdSessionTable->del(createStateDBKey(key));
+    SWSS_LOG_NOTICE("Software BFD session removed for %s", key.c_str());
+}
+
 BgpGlobalStateOrch::BgpGlobalStateOrch(DBConnector *db, string tableName):
     Orch(db, tableName)
 {
--- a/orchagent/bfdorch.h
+++ b/orchagent/bfdorch.h
@@ -19,6 +19,13 @@ public:
     virtual ~BfdOrch(void);
     void handleTsaStateChange(bool tsaState);
 
+    /* APIs for HaOrch to create passive BFD sessions on DPU.*/
+    void createSoftwareBfdSession(
+        const std::string& key,
+        const std::vector<swss::FieldValueTuple>& data);
+    void removeSoftwareBfdSession(
+        const std::string& key);
+
 private:
     bool create_bfd_session(const std::string& key, const std::vector<swss::FieldValueTuple>& data);
     bool remove_bfd_session(const std::string& key);
--- a/orchagent/dash/dashhaorch.h
+++ b/orchagent/dash/dashhaorch.h
@@ -54,6 +54,7 @@ protected:
     void doTaskEniTable(ConsumerBase &consumer);
     void doTaskHaSetTable(ConsumerBase &consumer);
     void doTaskHaScopeTable(ConsumerBase &consumer);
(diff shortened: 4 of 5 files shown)


orchagent/dash/dashhaorch.cpp
```diff
@@ -683,6 +686,37 @@ void DashHaOrch::doTaskHaScopeTable(ConsumerBase &consumer)
     }
 }
 
+void DashHaOrch::doTaskBfdSessionTable(ConsumerBase &consumer)
```
yue-fred-gao, line 823: From previous discussion, my understanding is that DashHaOrch controls when to start/stop bfd session. For example, when HA state is activated or dead. Therefore, HaMgrd is changed to only provide parameters for DashHaOrch creating the bfd session at the right moment. Here it simply relays the request to BfdOrch. Is this a design change?

zjswhhh: Hi @yue-fred-gao - thought again about it. Theoretically ha_scope to passive bfd sessions are not one-to-one mapping. Even it is, how will haorch know what bfd sessions to remove in planned shutdown? The only way is to map the PA IPs, which doesn't feel nature. 

In the state activating path, we need to add the bfd session back, does that mean haorch needs to cache the parameters? Or hamgrd will write it again to DPU_APPL_DB?

yue-fred-gao: Hi @zjswhhh , I see your point. I think at this stage we can assume all sessions in the DPU_APPL_DB/BFD_SESSION_TABLE are for DPU software bfd sessions. So in planned shutdown or bring up, you can create bfd sessions through bfdorch for each session in the table. If we need to extend dpu software bfd session for other purposes, we need to refactor the table schema. For example, we can add a column for owner or purpose to identify HA bfd sessions.
For the second question, in planned shutdown, DashHaOrch remove sessions in BfdOrch but it doesn't remove the entries from BFD_SESSION_TABLE so you don't need to cache them. hamgrd will also write the table again during bring up. Currently it doesn't remove the entries but we might during HA re-pairing.

zjswhhh: Update from offline discussion: in `ENI` level HA, BFD_SESSION_TABLE removal will not be needed.

orchagent/dash/dashhaorch.cpp
```diff
@@ -806,6 +857,88 @@ void DashHaOrch::doTaskHaScopeTable(ConsumerBase &consumer)
...
+        KeyOpFieldsValuesTuple tuple = it->second;
+        const auto& key = kfvKey(tuple);
+        const auto& op = kfvOp(tuple);
+
+        SWSS_LOG_DEBUG("Processing BFD Session table");
+
+        if (op == SET_COMMAND)
+        {
+            bool has_eni_scope = false;
+            for (const auto& ha_set_entry : m_ha_set_entries)
```
yue-fred-gao: It might be better to create a function that returns dpu_scope or eni_scope because there are 3 places doing this check. The two scope should be mutually exclusive.

orchagent/dash/dashhaorch.cpp
```diff
@@ -806,6 +857,88 @@ void DashHaOrch::doTaskHaScopeTable(ConsumerBase &consumer)
...
+
+            bool has_dpu_scope = false;
+            bool has_dpu_scope_ha_role_active = false;
+            for (const auto& ha_set_entry : m_ha_set_entries)
+            {
+                if (ha_set_entry.second.metadata.scope() == dash::types::HA_SCOPE_DPU)
+                {
+                    has_dpu_scope = true;
+
+                    for (const auto& ha_scope_entry : m_ha_scope_entries)
```
yue-fred-gao: nested loop is not necessary

orchagent/dash/dashhaorch.cpp
```diff
@@ -806,6 +857,88 @@ void DashHaOrch::doTaskHaScopeTable(ConsumerBase &consumer)
...
+
+            if (!m_ha_set_entries.empty() && has_dpu_scope_ha_role_active)
+            {
+                m_bfd_orch->createSoftwareBfdSession(key, kfvFieldsValues(tuple));
+            }
+
+            /*
+                In case bfd sessions are programmed before HA, cache them until HA scopes are created.
+            */
+            if (m_ha_set_entries.empty() ||
```
yue-fred-gao: we should cache if it is not eni scope. Even if has_dpu_scope_ha_role_active, ha-scope can go down then up, at which time we need to create bfd sessions.

orchagent/dash/dashhaorch.cpp
zjswhhh: addressed.

orchagent/dash/dashhaorch.cpp
zjswhhh: addressed

orchagent/dash/dashhaorch.cpp
zjswhhh: previous code is same logic, but still updated for better readability

orchagent/dash/dashhaorch.cpp
```diff
@@ -806,6 +820,68 @@ void DashHaOrch::doTaskHaScopeTable(ConsumerBase &consumer)
...
+
+            if (has_dpu_scope_ha_state_active)
+            {
+                m_bfd_orch->createSoftwareBfdSession(key, kfvFieldsValues(tuple));
+            }
+
+            /*
+                In case bfd sessions are programmed before HA, cache them until HA scopes are created.
+            */
+            if ( (!has_eni_scope()) ||
```
yue-fred-gao: The condition is confusing. Since eni_scope and dpu_scope cannot mix, if  has_dpu_scope, !has_eni_scope will be true. The rest of the check is not needed. I think !has_eni_scope is enough.

zjswhhh: Hi @yue-fred-gao, the 3 lines are for scenarios that the bfd sessions need to be *cached and created later*:
1. No HA_SET created
2. HA_SET is created but no HA_SCOPE
3. HA_SCOPE is created but ha_state is not in (standby, standalone, active) yet.

yue-fred-gao: But !has_eni_scope() supersedes the other 2 conditions so they will never be used.  And my previous point is still valid. Even ha-set and ha-scope are created and ha_state in active/standby/standalone state, you still need to cache the bfd session because ha-scope can go down then up, at which time we need to recreate bfd sessions.

zjswhhh: Okay. 

But for re-creation from down to up, shouldn't hamgrd writes again? If not, dashhaorch shouldn't clear the cache after creation (as it's doing today).

yue-fred-gao: No, hamgrd doesn't write. It creates bfd session in dpu appl_db when dpu actor is started and removes the bfd sessions when the DPU entry is removed from config_db. ha state change doesn't change the bfd session parameters so it doesn't need to rewrite.

zjswhhh: Okay I see the gap of understanding here. In that case you are right, bfd sessions need to be cached regardless (when it's dpu scope ha).

zjswhhh: Addressed.

orchagent/dash/dashhaorch.cpp
```diff
@@ -806,6 +820,66 @@ void DashHaOrch::doTaskHaScopeTable(ConsumerBase &consumer)
...
+
+                m_bfd_session_pending_creation[key] = kfvFieldsValues(tuple);
+            }
+
+            it = consumer.m_toSync.erase(it);
+        }
+        else if (op == DEL_COMMAND)
+        {
+            m_bfd_orch->removeSoftwareBfdSession(key);
+            it = consumer.m_toSync.erase(it);
```
yue-fred-gao, line 879: Do we need to remove cached bfd session?

zjswhhh: Updated.

yue-fred-gao (approved):

zjswhhh:
Hi @qiluo-msft - please help merge.

orchagent/dash/dashhaorch.cpp
```diff
@@ -806,6 +820,67 @@ void DashHaOrch::doTaskHaScopeTable(ConsumerBase &consumer)
...
+                    {
+                        has_dpu_scope_ha_state_active = true;
+                        break;
+                    }
+                }
+            }
+
+            if (has_dpu_scope_ha_state_active)
+            {
+                m_bfd_orch->createSoftwareBfdSession(key, kfvFieldsValues(tuple));
```
kperumalbfn, line 861: Why BFD sessions only for DPU scope HA active?

zjswhhh: Hi @kperumalbfn - if it's eni scope, we go ahead and create the bfd session in line 840. 

If it's dpu scope ha, per design, bfd session is used for planned up & down: 
https://github.com/sonic-net/SONiC/blob/master/doc/smart-switch/high-availability/smart-switch-ha-dpu-scope-dpu-driven-setup.md#81-ha-set-creation

zjswhhh:
Hi @prsunny  - can you help merge?

orchagent/dash/dashhaorch.cpp
kperumalbfn: Thanks @zjswhhh please add a comment and rename the variable - has_dpu_scope_ha_state_active.

zjswhhh: Addressed.

kperumalbfn (approved):

prsunny:
@zjswhhh , please update description following the template

PR #3792, opened 2025-07-28
[202505][portsorch] Ports initialization speedup

stepanblyschak:
BACKPORT OF: https://github.com/sonic-net/sonic-swss/pull/3506

**What I did**

Optimized port initialization process using the following approach:
1. Attributes that are known at initialization (whether because their default values are defined by the SAI spec or because we created the port with a value we already know) like speed and MTU.
2. Heavy operations like querying list of PGs and queues are done in bulk to save on SAIRedis call overhead.

**Why I did it**

To improve boot time for systems with many ports.

**How I verified it**

Ran fast-reboot and measure. Observed 45% reduction in time.

**Details if related**

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -999,6 +999,8 @@ void OrchDaemon::start(long heartBeatInterval)
  */
 bool OrchDaemon::warmRestoreAndSyncUp()
 {
+    SWSS_LOG_ENTER();
+
     WarmStart::setWarmStartState("orchagent", WarmStart::INITIALIZED);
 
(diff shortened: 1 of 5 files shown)


bingwang-ms:
@prsunny Can you please help check this cherry-pick request?
cc @yejianquan

prsunny:
@stepanblyschak , @bingwang-ms , this is risky to backport to 202505. Suggest close the PR.
@yejianquan for viz

stepanblyschak:
@prsunny Master PR is opened since Feb, long before 202505 branch out. If this PR is not taken, the KPI for reboots are not going to be met

PR #3799, opened 2025-07-31
add implicit counter action to ACL table type

anilkpan:
add implicit counter action to ACL table type

**What I did**
Added implicit counter action in ACL table type.

**Why I did it**
Implicit counter creation is done during ACL rule creation in OA, so the counter action must be added to the ACL table type as well.

**How I verified it**
Verified that ACL rule creation is successful even if counter action is not specified in the ACL table type config.

**Details if related**
With the current code, if counter action is not specified in the ACL table type config, ACL rule creation fails from OA.
Following syslog errors are seen:

2025 Jul 24 03:40:35.116003 sonic ERR swss#orchagent: :- add: Failed to create ACL rule TRIM_RULE in table TRIM_TABLE
2025 Jul 24 03:40:35.855953 sonic ERR swss#orchagent: :- create: create status: SAI_STATUS_INVALID_PARAMETER
2025 Jul 24 03:40:35.855959 sonic ERR swss#orchagent: :- createCounter: Failed to create counter for the rule TRIM_RULE in table TRIM_TABLE

2025 Jul 24 03:41:03.857029 sonic ERR syncd#syncd: :- processQuadEvent: attr: SAI_ACL_COUNTER_ATTR_TABLE_ID: oid:0x7000000004eaf
2025 Jul 24 03:41:03.857084 sonic ERR syncd#syncd: :- processQuadEvent: attr: SAI_ACL_COUNTER_ATTR_ENABLE_PACKET_COUNT: true
2025 Jul 24 03:41:03.857084 sonic ERR syncd#syncd: :- processQuadEvent: attr: SAI_ACL_COUNTER_ATTR_ENABLE_BYTE_COUNT: true
2025 Jul 24 03:41:03.857286 sonic ERR swss#orchagent: :- create: create status: SAI_STATUS_INVALID_PARAMETER
2025 Jul 24 03:41:04.112059 sonic ERR syncd#syncd: [none] SAI_API_ACL:_brcm_sai_acl_xgs_validate_create_counter:6763 ACL table doesnt have counter action configured
2025 Jul 24 03:41:04.112059 sonic ERR syncd#syncd: [none] SAI_API_ACL:_brcm_sai_create_acl_counter_int:2491 pd validate create counter failed with error -5.
2025 Jul 24 03:41:04.112059 sonic ERR syncd#syncd: :- sendApiResponse: api SAI_COMMON_API_CREATE failed in syncd mode: SAI_STATUS_INVALID_PARAMETER

### DIFF
--- a/orchagent/Makefile.am
+++ b/orchagent/Makefile.am
@@ -159,6 +159,7 @@ orchagent_SOURCES += p4orch/p4orch.cpp \
                      p4orch/tunnel_decap_group_manager.cpp \
 		     p4orch/l3_admit_manager.cpp \
 		     p4orch/l3_multicast_manager.cpp \
+		     p4orch/ip_multicast_manager.cpp \
  		     p4orch/ext_tables_manager.cpp
 
 orchagent_CFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_SAI) $(CFLAGS_ASAN)
--- a/orchagent/p4orch/ip_multicast_manager.cpp
+++ b/orchagent/p4orch/ip_multicast_manager.cpp
@@ -1,0 +1,509 @@
+#include "p4orch/ip_multicast_manager.h"
+
+#include <memory>
+#include <nlohmann/json.hpp>
+#include <sstream>
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "SaiAttributeList.h"
+#include "converter.h"
+#include "crmorch.h"
+#include "dbconnector.h"
+#include "ipaddress.h"
+#include "logger.h"
+#include "p4orch/p4oidmapper.h"
+#include "p4orch/p4orch_util.h"
+#include "portsorch.h"
+#include "sai_serialize.h"
+#include "swssnet.h"
+#include "table.h"
+#include "vrforch.h"
+
+extern "C" {
+#include "sai.h"
+}
+
+using ::p4orch::kTableKeyDelimiter;
+
+extern sai_object_id_t gSwitchId;
+extern sai_object_id_t gVirtualRouterId;
+extern sai_ipmc_api_t* sai_ipmc_api;
+extern sai_router_interface_api_t* sai_router_intfs_api;
+extern sai_rpf_group_api_t* sai_rpf_group_api;
+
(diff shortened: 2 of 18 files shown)


prsunny:
@bingwang-ms to review

bingwang-ms:
This looks like a clean, well-targeted fix. The root cause is clearly correct  since OA always calls createCounter() implicitly during ACL rule creation, the SAI ACL table must always include SAI_ACL_ACTION_TYPE_COUNTER in its action type list. The syslog evidence in the PR description makes the issue obvious.

A couple of questions before approving:

**1. Does withAction() deduplicate?**

If a user explicitly includes COUNTER in their ACL_TABLE_TYPE_ACTIONS config, the counter action will now be added twice (once from the config, once implicitly). Does AclTableTypeBuilder::withAction() handle duplicates gracefully (e.g., set-semantics)? If not, this could cause a SAI error when the action type list contains duplicate entries.

**2. Does the fix apply when no ACTIONS field is present?**

The implicit counter is added inside the ACL_TABLE_TYPE_ACTIONS field handler. If a table type has **no actions configured at all**, this block is never reached and the bug would still occur. Is that a valid/possible configuration? If so, the counter should also be added as a default when actions are absent.

**Minor nit:**
A short comment explaining *why* counter is implicit here would help future maintainers:
`cpp
// Counter action is always implicitly added because OA creates ACL counters
// unconditionally during rule creation (createCounter()), so the ACL table
// must have SAI_ACL_ACTION_TYPE_COUNTER in its action type list.
builder.withAction(AclEntryActionToAclAction(SAI_ACL_ENTRY_ATTR_ACTION_COUNTER));
`

**On testing:** A DVS test verifying that ACL rule creation succeeds when COUNTER is absent from the table type config would be nice, but given this is a targeted one-line fix with clear syslog validation, it's not a blocker.

If the dedup and no-actions questions check out, this is good to merge.

anilkpan:
@bingwang-ms
Does withAction() deduplicate?
_ANIL>> Yes, it handles duplicate actions gracefully._
Does the fix apply when no ACTIONS field is present?
_ANIL> An ACL without an action is not expected to be configured. But, I will change the code to add counter action irrespective of whether an action is configured or not._
A short comment explaining why counter is implicit here would help future maintainers:
_ANIL>> I will add the comment as well._

PR #3809, opened 2025-08-05
Fix race condition between teamd starting and failing immediately and intfmgrd configuring an IP address

saiarcot895:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


cfgmgr/intfmgr.cpp
```diff
@@ -120,16 +120,18 @@ void IntfMgr::setIntfIp(const string &alias, const string &opCmd,
...
-                return;
+                return false;
             }
             ret = swss::exec(cmd.str(), res);
         }
 
         if (ret)
         {
             SWSS_LOG_ERROR("Command '%s' failed with rc %d", cmd.str().c_str(), ret);
+            return false;
```
prsunny: @saiarcot895 , this cannot be indefinitely retried. there can be cases where ip address is assigned and interface is deleted, like sub-interface etc. I suggest to remove the retry logic and fix the current issue with just teamsync changes

saiarcot895: Reverted the commit that added that logic, and the only changes present are for teamsyncd.

PR #3817, opened 2025-08-08
Adding Dynamic Port Stat counter creation support

maunnikr-cisco:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -3726,6 +3726,44 @@ string PortsOrch::getWredQueueFlexCounterTableKey(string key)
     return string(WRED_QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP) + ":" + key;
 }
 
+bool PortsOrch::querySupportedPortStats(sai_object_id_t port_id, vector<sai_port_stat_t>& stat_ids)
+ {
+     sai_attr_capability_t capability;
+     bool extendedCountersSupported = sai_query_attribute_capability(gSwitchId, SAI_OBJECT_TYPE_PORT,
+                                             SAI_PORT_ATTR_PORT_STAT_EXTENDED,
+                                             &capability) == SAI_STATUS_SUCCESS;
+
+     if (extendedCountersSupported)
+     {
+         sai_attribute_t attr;
+         vector<sai_port_stat_t> supported_counters(80); // Preallocate space
+         attr.value.s32list.count = static_cast<uint32_t>(supported_counters.size());
+         attr.value.s32list.list = reinterpret_cast<sai_int32_t*>(supported_counters.data());
+         attr.id = SAI_PORT_ATTR_PORT_STAT_EXTENDED;
+
+         sai_status_t status = sai_port_api->get_port_attribute(port_id, 1, &attr);
+         if (status == SAI_STATUS_SUCCESS)
+         {
+             supported_counters.resize(attr.value.s32list.count);
+             stat_ids = supported_counters;
+             SWSS_LOG_DEBUG("Queried %zu extended port stats for port 0x%" PRIx64, stat_ids.size(), port_id);
+             return true;
+         }
+         else
+         {
+             SWSS_LOG_ERROR("Failed to query extended port stats for port 0x%" PRIx64 ", status: %d", port_id, status);
+         }
+     }
+     else
+     {
+         SWSS_LOG_DEBUG("Extended port stats not supported, using default port stat counters");
+     }
+
+     // Fallback to default port stats
+     stat_ids = port_stat_ids;
+     return false;
+ }
+
 bool PortsOrch::initPort(const PortConfig &port)
 {
     SWSS_LOG_ENTER();
@@ -3767,7 +3805,10 @@ bool PortsOrch::initPort(const PortConfig &port)
                 /* Add port name map to counter table */
                 FieldValueTuple tuple(p.m_alias, sai_serialize_object_id(p.m_port_id));
                 vector<FieldValueTuple> fields;
-                fields.push_back(tuple);
+                fields.push_back(tuple);           
+				m_counterTable->set("", fields);
+                vector<sai_port_stat_t> stat_ids;
+                bool query_success = false;
                 // m_counterTable->set("", fields);
                 m_counterNameMapUpdater->setCounterNameMap(p.m_alias, p.m_port_id);
 
@@ -3777,9 +3818,16 @@ bool PortsOrch::initPort(const PortConfig &port)
                 If they are enabled, install the counters immediately */
                 if (flex_counters_orch->getPortCountersState())
                 {
-                    auto port_counter_stats = generateCounterStats(port_stat_ids, sai_serialize_port_stat);
-                    port_stat_manager.setCounterIdList(p.m_port_id,
-                            CounterType::PORT, port_counter_stats);
+                	vector<sai_port_stat_t> stat_ids;
+                    bool query_success = querySupportedPortStats(p.m_port_id, stat_ids);
+
+                    if (!query_success)
+                    {
+                        SWSS_LOG_DEBUG("Using default port stats for port 0x%" PRIx64, p.m_port_id);
+                    }
+
+                    auto port_counter_stats = generateCounterStats(stat_ids, sai_serialize_port_stat);                    
+					port_stat_manager.setCounterIdList(p.m_port_id, CounterType::PORT, port_counter_stats);
                     auto gbport_counter_stats = generateCounterStats(gbport_stat_ids, sai_serialize_port_stat);
                     if (p.m_system_side_id)
                         gb_port_stat_manager.setCounterIdList(p.m_system_side_id,
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -8364,8 +8364,6 @@ void PortsOrch::generatePortCounterMap()
         return;
     }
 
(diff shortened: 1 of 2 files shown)


orchagent/portsorch.cpp
```diff
@@ -3799,6 +3798,62 @@ string PortsOrch::getWredQueueFlexCounterTableKey(string key)
...
+    {
+        SWSS_LOG_ERROR("querySupportedPortStats: exception caught: %s", e.what());
+        extendedCountersSupported = true; // Default to true if query fails
+    }
+    
+    if (extendedCountersSupported)
+    {
+        SWSS_LOG_ERROR("querySupportedPortStats: extendedCountersSupported is true");
+        sai_attribute_t attr;
+        vector<sai_port_stat_t> supported_counters(80); // Preallocate space
```
dgsudharsan: Avoid using magic numbers. Check existing examples where we first call with 0 and get buffer overflow and based on it we get the size.

orchagent/portsorch.cpp
```diff
@@ -3799,6 +3798,62 @@ string PortsOrch::getWredQueueFlexCounterTableKey(string key)
     return string(WRED_QUEUE_STAT_COUNTER_FLEX_COUNTER_GROUP) + ":" + key;
 }
 
+bool PortsOrch::querySupportedPortStats(sai_object_id_t port_id, vector<sai_port_stat_t>& stat_ids)
```
dgsudharsan: querySupportedPortExtendedStats?

orchagent/portsorch.cpp
```diff
@@ -3799,6 +3798,62 @@ string PortsOrch::getWredQueueFlexCounterTableKey(string key)
...
+    bool extendedCountersSupported = true;
+    try
+    {
+        extendedCountersSupported = sai_query_attribute_capability(gSwitchId, SAI_OBJECT_TYPE_PORT,
+                                            SAI_PORT_ATTR_PORT_STAT_EXTENDED,
+                                            &capability) == SAI_STATUS_SUCCESS;
+        if (!extendedCountersSupported)
+        {
+            SWSS_LOG_ERROR("querySupportedPortStats: setting extendedCountersSupported to true by default");
+            extendedCountersSupported = true;
```
dgsudharsan: This is wrong. Why is this always true? Please revisit the logic

orchagent/portsorch.cpp
```diff
@@ -647,8 +647,7 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
 
     /* Initialize counter table */
     m_counter_db = shared_ptr<DBConnector>(new DBConnector("COUNTERS_DB", 0));
-    // m_counterTable = unique_ptr<Table>(new Table(m_counter_db.get(), COUNTERS_PORT_NAME_MAP));
-    m_counterNameMapUpdater = unique_ptr<CounterNameMapUpdater>(new CounterNameMapUpdater("COUNTERS_DB", COUNTERS_PORT_NAME_MAP));
```
dgsudharsan: Why did you change the existing logic?

orchagent/portsorch.cpp
```diff
@@ -3841,16 +3896,24 @@ bool PortsOrch::initPort(const PortConfig &port)
...
                 fields.push_back(tuple);
-                // m_counterTable->set("", fields);
-                m_counterNameMapUpdater->setCounterNameMap(p.m_alias, p.m_port_id);
+                m_counterTable->set("", fields);
 
                 // Install a flex counter for this port to track stats
                 auto flex_counters_orch = gDirectory.get<FlexCounterOrch*>();
                 /* Delay installing the counters if they are yet enabled
                 If they are enabled, install the counters immediately */
+                SWSS_LOG_ERROR("Flex counters :getPortCountersState");
```
dgsudharsan: Why is there an error?

orchagent/portsorch.cpp
```diff
@@ -3841,16 +3896,24 @@ bool PortsOrch::initPort(const PortConfig &port)
...
-                    auto port_counter_stats = generateCounterStats(port_stat_ids, sai_serialize_port_stat);
+                    vector<sai_port_stat_t> stat_ids;
+                    bool query_success = querySupportedPortStats(p.m_port_id, stat_ids);
+                    
+                    if (!query_success)
+                    {
+                        SWSS_LOG_ERROR("Using default port stats for port 0x%" PRIx64, p.m_port_id);
+                    }
+                    
+                    auto port_counter_stats = generateCounterStats(stat_ids, sai_serialize_port_stat);
```
dgsudharsan: The above API will break the existing stats. Your query function will just give extended stats and not the main stats. Please revisit the logic

orchagent/portsorch.cpp
```diff
@@ -3947,9 +4010,7 @@ void PortsOrch::deInitPort(string alias, sai_object_id_t port_id)
     }
 
     /* remove port name map from counter table */
-    // m_counterTable->hdel("", alias);
-    m_counterNameMapUpdater->delCounterNameMap(alias);
-
+    m_counterTable->hdel("", alias);
```
dgsudharsan: Why is this change needed?

orchagent/portsorch.cpp
```diff
@@ -4417,7 +4478,7 @@ void PortsOrch::doPortTask(Consumer &consumer)
                         p.m_an_cfg = true;
                         m_portList[p.m_alias] = p;
                         m_portStateTable.hdel(p.m_alias, "rmt_adv_speeds");
-                        updatePortStatePoll(p, PORT_STATE_POLL_AN, pCfg.autoneg.value);
+                        updatePortStatePoll(p, PORT_STATE_POLL_AN, pCfg.autoneg.value); //
```
dgsudharsan: Remove unnecessary code changes

orchagent/portsorch.cpp
```diff
@@ -8384,22 +8445,45 @@ void PortsOrch::removePortBufferPgCounters(const Port& port, string pgs)
 
 void PortsOrch::generatePortCounterMap()
 {
+    SWSS_LOG_ENTER();
     if (m_isPortCounterMapGenerated)
-    {
+    {   
         return;
     }
+    SWSS_LOG_ERROR("generatePortCounterMap Entered");
```
dgsudharsan: Why log error?

orchagent/portsorch.cpp
```diff
@@ -8384,22 +8445,45 @@ void PortsOrch::removePortBufferPgCounters(const Port& port, string pgs)
...
+        }
+        catch (const std::exception& e)
+        {
+            SWSS_LOG_ERROR("Failed to query supported port stats for port 0x%" PRIx64 ": %s",
+                            it.second.m_port_id, e.what());
+            SWSS_LOG_ERROR("Setting query success to true anyways");
+            stat_ids = port_stat_ids;
+        }
+
+        auto port_counter_stats = generateCounterStats(stat_ids, sai_serialize_port_stat);
```
dgsudharsan, line 8991: Same as previous. This will break the existing counters. Please revisit the logic

orchagent/portsorch.cpp
```diff
@@ -8384,22 +8445,45 @@ void PortsOrch::removePortBufferPgCounters(const Port& port, string pgs)
...
-        // Set counter stats only for PHY ports to ensure syncd will not try to query the counter statistics from the HW for non-PHY ports.
+        // Set counter stats only for PHY ports
         if (it.second.m_type != Port::Type::PHY)
         {
             continue;
         }
+
+        vector<sai_port_stat_t> stat_ids;
+        try{
+            bool query_success = querySupportedPortStats(it.second.m_port_id, stat_ids);
```
dgsudharsan: Isn't the fallback already present in  querySupportedPortStats. Why do we need it here?

orchagent/portsorch.cpp
maunnikr-cisco: Done.

orchagent/portsorch.cpp
maunnikr-cisco: I have tested this and made corrections such that we only go with the queried counters, if the function is a success, which also includes whether the attribute is supported.

orchagent/portsorch.cpp
maunnikr-cisco: Removed.

orchagent/portsorch.cpp
maunnikr-cisco: Removed. Was a diff issue.

orchagent/portsorch.cpp
maunnikr-cisco: Removed. Was a diff issue.

orchagent/portsorch.cpp
maunnikr-cisco: Removed, implemented a true false check instead.

orchagent/portsorch.cpp
maunnikr-cisco: Removed. Was a diff issue.

orchagent/portsorch.cpp
maunnikr-cisco: Removed. Was a diff issue.

orchagent/portsorch.cpp
maunnikr-cisco: Have defined a max value. constant.  We will have to introduce another attribute if we don't want to do this. The max value is big enough such that it should not cause buffer overflow.

orchagent/portsorch.cpp
maunnikr-cisco: Have fixed the logic. Was a an old diff issue.

PR #3818, opened 2025-08-09
[Dash] Only set relevant attributes for Floating NIC

vivekrnv:
**What I did**

Don't set SAI_DIRECTION_LOOKUP_ENTRY_ATTR_DASH_ENI_MAC_OVERRIDE_TYPE for Floating NIC i.e. when outbound_dir_lookup == "dst_mac"

**Why I did it**

Overriding MAC is implied in FNIC. This attribute holds no meaning for FNIC

**How I verified it**

UT and verified by loading and running the FNIC sonic-mgmt tests

**Details if related**

### DIFF
--- a/.azure-pipelines/test-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/test-docker-sonic-vs-template.yml
@@ -79,21 +79,6 @@ jobs:
     inputs:
       source: specific
       project: build
-      pipeline: Azure.sonic-buildimage.common_libs
-      runVersion: 'latestFromBranch'
-      runBranch: 'refs/heads/$(BUILD_BRANCH)'
-      path: $(Build.ArtifactStagingDirectory)/download
-      artifact: common-lib
-      patterns: |
-        target/debs/bookworm/libyang-*_1.0*.deb
-        target/debs/bookworm/libyang_1.0*.deb
-        target/debs/bookworm/libyang-cpp_*.deb
-        target/debs/bookworm/python3-yang_*.deb
-    displayName: "Download libyang from common lib"
-  - task: DownloadPipelineArtifact@2
-    inputs:
-      source: specific
-      project: build
       pipeline: sonic-net.sonic-buildimage-ubuntu22.04
       artifact: sonic-buildimage.amd64.ubuntu22_04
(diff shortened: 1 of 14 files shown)


prabhataravind:
@mukeshmv please review as well.

prabhataravind (approved):
lgtm. Please wait for Mukesh's approval as well.

mukeshmv:
lgtm. @prabhataravind it would be good to have Riff also take a look at this

prabhataravind:
> lgtm. @prabhataravind it would be good to have Riff also take a look at this

Ack. @r12f could you please sign-off as well?

theasianpianist (approved):

orchagent/dash/dashorch.cpp
```diff
@@ -202,16 +196,13 @@ bool DashOrch::addApplianceEntry(const string& appliance_id, const dash::applian
...
-    appliance_attr.id = SAI_DIRECTION_LOOKUP_ENTRY_ATTR_DASH_ENI_MAC_OVERRIDE_TYPE;
-    if (entry.has_outbound_direction_lookup())
-    {
-        appliance_attr.value.u32 = sMacOverride[entry.outbound_direction_lookup()];
-    }
-    else
+    // Don't set up this entry for dst_mac i.e. for Floating NIC.
+    if (!entry.has_outbound_direction_lookup() || (entry.outbound_direction_lookup() != "dst_mac"))
     {
+        appliance_attr.id = SAI_DIRECTION_LOOKUP_ENTRY_ATTR_DASH_ENI_MAC_OVERRIDE_TYPE;
```
r12f, line 202: the current design is that:
1. If outbound direction lookup is not set, we default to SRC MAC, no override needs to be provided.
2. If outbound direction lookup is set, we set to whatever the value is.

Here is the HLD link, just in case: https://github.com/sonic-net/DASH/blob/main/documentation/express-route-service/express-route-gateway-bypass.md#51-eni-mac-handling

r12f: Discussed this one in the email. Although we need to use ENI mode to determine the outbound direction lookup, but the team decided to have this as short-term solution. Hence considering this as non-blocking.

PR #3819, opened 2025-08-09
Support HFT on PG BUFFER and BUFFER_POOL

Pterosaur:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


Pterosaur:
Hi @prsunny , could you please help to review this PR?

prsunny:
@kperumalbfn , would you review the PR?

orchagent/high_frequency_telemetry/counternameupdater.cpp
```diff
@@ -35,21 +36,56 @@ void CounterNameMapUpdater::setCounterNameMap(const std::string &counter_name, s
     m_counters_table.hset("", counter_name, sai_serialize_object_id(oid));
 }
 
+void CounterNameMapUpdater::setCounterNameMap(const std::vector<swss::FieldValueTuple> &counter_name_maps)
+{
+    SWSS_LOG_ENTER();
+
+    if (gHFTOrch)
+    {
+        for (const auto& map : counter_name_maps)
```
kperumalbfn, line 45: Could you check if this is a one time effort? With large number of ports and queue/ports, this could add delay

orchagent/portsorch.h
```diff
@@ -260,18 +260,17 @@ class PortsOrch : public Orch, public Subject
...
 private:
     unique_ptr<CounterNameMapUpdater> m_counterNameMapUpdater;
-    // unique_ptr<Table> m_counterTable;
     unique_ptr<Table> m_counterSysPortTable;
     unique_ptr<Table> m_counterLagTable;
     unique_ptr<Table> m_portTable;
     unique_ptr<Table> m_sendToIngressPortTable;
     unique_ptr<Table> m_gearboxTable;
-    unique_ptr<Table> m_queueTable;
+    unique_ptr<CounterNameMapUpdater> m_queueCounterNameMapUpdater;
```
kperumalbfn, line 268: Where are these counterMaps for PG and Queue used?

orchagent/high_frequency_telemetry/counternameupdater.cpp
Pterosaur: I think it's just one time effort, because these objects, such as ports, buffer pool or PG, will only be created in the config load stage.
Meanwhile, as the line 43 shown, to these platforms without HFT supported, the delay is totally same as before.

orchagent/portsorch.h
Pterosaur: For the feature high frequency telemetry(https://github.com/sonic-net/SONiC/blob/master/doc/high-frequency-telemetry/high-frequency-telemetry-hld.md) , to these platforms with a large number of ports, such as 512 ports or above, the traditional telemetry collection, flex counter, will consume too much CPU, which is unacceptable.

kperumalbfn: @Pterosaur Do you have any performance measurement of all these counter query?

Pterosaur: Hi @kperumalbfn , I don't have profiling to the counter query.
I think this change will only happen on the create/delete stages instead of query operation. The query logic is totally same with before.

kperumalbfn (approved):

kperumalbfn:
@Pterosaur IS this already merged into 202412?

Pterosaur:
> @Pterosaur IS this already merged into 202412?

Haven't yet. This PR will not be merged into 202412 until we get 100% pass rate of 202412.

PR #3821, opened 2025-08-11
[portsorch] Fixed port comparison logic executed every doTask for every port

stepanblyschak:
**What I did**
- Removed TODO comment about issue that was causing initPort() to be called multiple times
- Fixed state checking logic to use explicit PORT_CONFIG_RECEIVED instead of != PORT_CONFIG_MISSING
- Added proper handling for PORT_CONFIG_DONE state to prevent duplicate port initialization
- Added error handling for port addition failures in the new state handling block

**Why I did it**

initPort is beeing called 17 965 328 times on a system with 514 ports during cold boot and shows up in profilers. This is because the port comparison logic is invoked every port update and loops for every port. For every port that hasn't changed its lanes initPort is called. This is an N^2 loop and causes inefficient PORT table handling in cold boot. Saved around ~6 sec in cold boot downtime test with this change.

Port comparison logic runs only once when PortConfigDone is received.

**How I verified it**
- Verified port initialization works correctly in both initial and runtime port split cases.
- Confirmed no duplicate port initialization occurs

**Details if related**

### DIFF
--- a/orchagent/switch/trimming/capabilities.cpp
+++ b/orchagent/switch/trimming/capabilities.cpp
@@ -115,6 +115,13 @@ static std::string toStr(const std::set<sai_packet_trim_queue_resolution_mode_t>
     return join(",", strList.cbegin(), strList.cend());
 }
 
+static std::string toStr(sai_status_t value)
+{
+    const auto *name = sai_metadata_get_status_name(value);
+
+    return name != nullptr ? name : "UNKNOWN";
+}
+
 static std::string toStr(sai_uint8_t value)
 {
     return std::to_string(value);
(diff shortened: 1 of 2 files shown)


stepanblyschak:
Coverage failed due to missing coverage on these lines:
image

but it is impossible to have them covered by UT, if I mock SAI create_ports to return any status code other then success it will throw exception instead of returning false. This code is unreachable, but removing it to pass coverage is also a bad idea.

volodymyrsamotiy:
@prsunny, could you please help to merge? (please check Stepan's comment for the test failure)

PR #3822, opened 2025-08-11
Avoid nhgroup update when mux state changes

manamand2020:
**What I did**
Avoided updating the NHGroup with mux neighbor nexthops during mux state updates. 

**Why I did it**
When a mux neighbor in a ECMP NexthopGroup changes state to standby we point the prefix route originally pointing to the ECMP nexthop group to any available active neighbor NH or tunnel NH and the original ECMP nexthopgroup remains unused, so this update is useless and uncessarily takes extra SAI calls. This fix also avoid creation of nexthop group with a mix of different types of NextHops which will allow platform that do not support such nexthop groups.

**How I verified it**
Verified DVS tests and on hardware setup.
**Details if related**

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -782,26 +782,6 @@ bool MuxNbrHandler::enable(bool update_rt)
         /* Increment ref count for new NHs */
         gNeighOrch->increaseNextHopRefCount(nh_key, num_routes);
 
-        /*
-         * Invalidate current nexthop group and update with new NH
-         * Ref count update is not required for tunnel NH IDs (nh_removed)
-         */
-        uint32_t nh_removed, nh_added;
-        if (!gRouteOrch->invalidnexthopinNextHopGroup(nh_key, nh_removed))
-        {
-            SWSS_LOG_ERROR("Removing existing NH failed for %s", nh_key.ip_address.to_string().c_str());
-            return false;
-        }
-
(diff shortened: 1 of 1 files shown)


Ndancejic (approved):
lgtm

zjswhhh:
@manamand2020  - please add proper labels for this PR.

zjswhhh:
Hi @prsunny - can you help review and merge this please?

prsunny (approved):

saiarcot895:
@manamand2020 I believe that this PR is causing dualtor tests on KVM single tor T0 to fail. See [this run](https://elastictest.org/scheduler/testplan/6920093d05c827bcee33e11a?testcase=dualtor%2Ftest_orchagent_active_tor_downstream.py&type=console) for an example. Can you take a look?

manamand2020:
> @manamand2020 I believe that this PR is causing dualtor tests on KVM single tor T0 to fail. See [this run](https://elastictest.org/scheduler/testplan/6920093d05c827bcee33e11a?testcase=dualtor%2Ftest_orchagent_active_tor_downstream.py&type=console) for an example. Can you take a look?

Did not get a chance to run the management tests to see the exact reason here for the failure. But I think this seems to be happening because the neighbor remains in use inside the ECMP group due to this change while we try to remove the entry from disableneighbors() which throws this error. We should have some DVS test to cover this.

PR #3825, opened 2025-08-13
[ssw][ha] consume new ha_scope fields

zjswhhh:
**What I did**
Address the minor changes in HA HLD:
1. reject ha_scope object createion if ha_set is not created first.
2.  get ha_set_id mapping from ha_scope table
3. get vip from ha_scope table 

Dependency:  https://github.com/sonic-net/sonic-dash-api/pull/44

**Why I did it**
To align with HLD change. 

**How I verified it**
* UTs. 
* load on DPU, program objects (pending)

**Details if related**


yue-fred-gao (approved):

prsunny (approved):

zjswhhh:

PR #3832, opened 2025-08-15
Revert "[swss] Install yang related debto  align with latest libswsscommon requirement"

wen587:
Reverts sonic-net/sonic-swss#3815


wen587:
image

PR #3836, opened 2025-08-18
[swss] Support FIPS MACSec POST

ysmanman:
Add SAI MACSec POST support in SWSS
- When FIPS is enabled in SONiC, enable MACSecPOST in switch creation.
- If MACSec POST is only supported in MACSec init,  create MACSec objects and enable POST when initializing MACSecOrch.
- Set SAI POST status in StateDB accordingly based on SAI POST status notificaiton.
- MACSecMgr does not process any MACSec configuration if SAI POST fails.

With the PR, the only case that Orchagent declares SAI MACSec POST to fail is:
- FIPS is enabled in SONiC; AND
- SAI supports MACSec POST; AND
- SAI returns failure on MACSec POST.

We particularly verified the following on switch with current BRCM SAI (13.2.1.0) that does not support MACSec POST yet:
- MACSec ports came up fine when FIPS is not enabled in SONiC;
- MACSec ports came up fine when FIPS is enabled in SONiC and SAI does not support MACSec POST.

### DIFF
--- a/cfgmgr/buffer_pool_mellanox.lua
+++ b/cfgmgr/buffer_pool_mellanox.lua
@@ -22,6 +22,7 @@ local total_port = 0
 
 local mgmt_pool_size = 256 * 1024
 local egress_mirror_headroom = 10 * 1024
+local modification_descriptors_pool_size = 0
 
 -- The set of ports with 8 lanes
 local port_set_8lanes = {}
@@ -185,6 +186,22 @@ end
 -- Main --
 -- Connect to CONFIG_DB
 redis.call('SELECT', config_db)
+
+-- Check if platform is SPC6 or later and set modification descriptors pool size
+-- Extract model number from platform string (e.g., "sn6600" -> 6600, "sn5800" -> 5800, "sn10600" -> 10600)
+-- Use (%d+) pattern to capture one or more digits for extensibility (handles future multi-digit series like sn10xxx, sn11xxx)
+local platform = redis.call('HGET', 'DEVICE_METADATA|localhost', 'platform')
+if platform then
+    local model_str = string.match(platform, "sn(%d+)")
+    if model_str then
+        local model_number = tonumber(model_str)
+        -- SPC6 or later models (>= 6000 excludes SPC5 models like 5400/5800, includes SPC6+ like 6600/7xxx/10xxx)
+        -- Reserve 32MB for modification descriptors pool
+        if model_number and model_number >= 6000 then
+            modification_descriptors_pool_size = 32 * 1024 * 1024
+        end
+    end
+end
 
 -- Parse all the pools and seperate them according to the direction
 local ipools = {}
@@ -384,7 +401,7 @@ accumulative_occupied_buffer = accumulative_occupied_buffer + accumulative_manag
 
 -- Accumulate sizes for egress mirror and management pool
 local accumulative_egress_mirror_overhead = admin_up_port * egress_mirror_headroom
-accumulative_occupied_buffer = accumulative_occupied_buffer + accumulative_egress_mirror_overhead + mgmt_pool_size
+accumulative_occupied_buffer = accumulative_occupied_buffer + accumulative_egress_mirror_overhead + mgmt_pool_size + modification_descriptors_pool_size
 
 -- Switch to CONFIG_DB
 redis.call('SELECT', config_db)
@@ -475,5 +492,6 @@ table.insert(result, "debug:shp_enabled:" .. tostring(shp_enabled))
 table.insert(result, "debug:shp_size:" .. shp_size)
 table.insert(result, "debug:total port:" .. total_port .. " ports with 8 lanes:" .. port_count_8lanes)
 table.insert(result, "debug:admin up port:" .. admin_up_port .. " admin up ports with 8 lanes:" .. admin_up_8lanes_port)
+table.insert(result, "debug:modification_descriptors_pool_size:" .. modification_descriptors_pool_size)
 
 return result


ysmanman:
@kenneth-arista @arlakshm @judyjoseph @wumiaont

rlhui:
@senthil-nexthop

orchagent/main.cpp
```diff
@@ -643,6 +644,34 @@ int main(int argc, char **argv)
         attrs.push_back(attr);
     }
 
+    // Enable FIPS POST if needed.
+    // TODO: MACSec POST needs to be eanbled in switch creation. So it's expected to
+    // to provide a config for orchagent to check if POST needs to be enabled or not.
+    bool macsec_post_enabled = false;
```
wumiaont: There's FIPS_STATS table in STATE_DB indicating if FIPS is enabled or not for the chassis. You can use this table to decide whether chassis needs POST or not. 
sonic-db-cli STATE_DB hgetall "FIPS_STATS|state"
{'enabled': 'True', 'enforced': 'True', 'timestamp': '2025-08-27T19:19:30.327287'}
Be aware of that if chassis is multi-asic then needs to create this table on namespace STATE_DB as well. So swss can access this table.

orchagent/main.cpp
```diff
@@ -756,6 +785,35 @@ int main(int argc, char **argv)
...
+                SWSS_LOG_NOTICE("MACSec POST enabled in switch init");
+            }
+            else if (sai_query_attribute_capability(gSwitchId, SAI_OBJECT_TYPE_MACSEC,
+                                                    SAI_MACSEC_ATTR_ENABLE_POST,
+                                                    &post_capability) == SAI_STATUS_SUCCESS &&
+                post_capability.create_implemented)
+            {
+                // POST is only supported in MACSec init. Set POST state to notify MACSecOrch
+                // to perform POST.
+                setMacsecPostState(&state_db, "macsec-level-post-in-progress");
```
wumiaont, line 811: Will there be possible race condition here? The POST notification will update the POST result which will come from syncd. Are we sure the notification handling is the last to update DB table? I am thinking about any scenario of that POST result notification comes early and then this part updates table later. It will be OK if those query happens before create_switch but sai_query_attribute_capability needs gSwitchId which is created after create_switch().
Recommend to add a check to the table.  Stop update if entry is "pass" or "fail".

orchagent/main.cpp
senthil-nexthop: After enabling FIPS the switch needs to be rebooted. procdockerstatsd, which updates FIPS_STATS, runs every two minutes. So there could be a possible window when the state is not available when Orchagent starts. I wonder if it is better to check /etc/fips/fips_enable or read /proc/cmdline and look for fips=1 or sonic_fips=1.

orchagent/main.cpp
senthil-nexthop: @wumiaont If the switch is not POST capable, do we expect to get a POST notification from syncd?

cfgmgr/macsecmgrd.cpp
```diff
@@ -61,6 +73,13 @@ int main(int argc, char **argv)
         swss::DBConnector cfgDb("CONFIG_DB", 0);
         swss::DBConnector stateDb("STATE_DB", 0);
 
+        /* Don't process any config until POST state is ready */
+        SWSS_LOG_NOTICE("checking FIPS POST state");
+        while (!isFipsPostStateReady(&stateDb))
+        {
+            sleep(1);
```
senthil-nexthop: Instead of a blocking while loop here, it may be safer to perform the post state readiness check inside the main event loop below. This will allow other events to be processed besides configuration - signals, timeouts, etc.

orchagent/macsecorch.cpp
```diff
@@ -649,6 +651,53 @@ MACsecOrch::MACsecOrch(
...
+
+    // Add handler for POST completion callback/notification.
+    string post_state = getMacsecPostState();
+    if (post_state == "switch-level-post-in-progress" ||
+        post_state == "macsec-level-post-in-progress" )
+    {
+        m_notificationsDb = make_shared<DBConnector>("ASIC_DB", 0);
+        m_postCompletionNotificationConsumer = new swss::NotificationConsumer(m_notificationsDb.get(), "NOTIFICATIONS");
+        auto postCompletionNotificatier = new Notifier(m_postCompletionNotificationConsumer, this, "POST_COMPLETION__NOTIFICATIONS");
+        Orch::addExecutor(postCompletionNotificatier);
```
senthil-nexthop, line 664: If the POST completion notifications come from Syncd before macsecorch is initialized, how do we want to handle the missed notifications? Is there a race condition here?

orchagent/macsecorch.cpp
```diff
@@ -649,6 +651,53 @@ MACsecOrch::MACsecOrch(
...
+                SWSS_LOG_NOTICE("Switch MACSec POST passed");
+            }
+            else if (attr.value.s32 == SAI_SWITCH_MACSEC_POST_STATUS_FAIL)
+            {
+                setMacsecPostState("fail");
+                SWSS_LOG_ERROR("Switch MACSec POST failed: oid %" PRIu64, gSwitchId);
+            }
+            else
+            {
+                SWSS_LOG_NOTICE("Switch MACSec POST status: %d", attr.value.s32);
```
senthil-nexthop, line 687: It may be useful add an "unknown" state for this condition and the one below, as otherwise the state will simply say in-progress and it will be hard to troubleshoot/distinguish if it is a missed notification or notification was never sent, etc.

orchagent/macsecorch.cpp
```diff
@@ -661,6 +710,137 @@ MACsecOrch::~MACsecOrch()
...
+
+    std::deque<KeyOpFieldsValuesTuple> entries;
+    consumer.pops(entries);
+    for (auto& entry : entries)
+    {
+        handleNotification(consumer, entry);
+    }
+}
+
+void MACsecOrch::setMacsecPostState(const std::string & postState)
```
senthil-nexthop: This and the getter below seems to be duplicate of the functions defined in macsecpost.cpp, reuse those here?

orchagent/macsecorch.cpp
```diff
@@ -661,6 +710,137 @@ MACsecOrch::~MACsecOrch()
...
+    SWSS_LOG_NOTICE("Received SAI notification: op %s, data %s", op.c_str(), data.c_str());
+
+    if (op == "switch_macsec_post_status")
+    {
+        sai_object_id_t switch_id;
+        sai_switch_macsec_post_status_t switch_macsec_post_status;
+        sai_deserialize_switch_macsec_post_status_ntf(data, switch_id, switch_macsec_post_status);
+
+        string post_state = getMacsecPostState();
+        if (post_state == "switch-level-post-in-progress")
```
senthil-nexthop, line 751: We don't expect the post_state to be "macsec-level-post-in-progress", but better to add an else and log the condition.

orchagent/main.cpp
wumiaont: > @wumiaont If the switch is not POST capable, do we expect to get a POST notification from syncd?

In this case, there will be no callback triggered by BCOM SDK and no notification from syncd.

ysmanman: @wumiaont POST notification is handled by MACSecOrch. In order to handle the case in which  POST is enabled in switch init and the status notification is sent before MAcSecOrch is initialized, MACSecOrch was enhanced to query SAI POST status when it's initialized. Please see the change in MaACSecOrch.

orchagent/macsecorch.cpp
ysmanman: @senthil-nexthop MACSecOrch will query SAI POST status when it's initialized, which is implemented to handle this case.

orchagent/macsecorch.cpp
ysmanman: As replied to above comments,  the current change should handle missing notification too. From SWSS perspective, `in-progress` is the same as `unknown` . Because it means POST is *not* passed yet, so no MACSec config can be processed.

orchagent/macsecorch.cpp
ysmanman: Sure, will do.

orchagent/main.cpp
```diff
@@ -643,6 +644,34 @@ int main(int argc, char **argv)
         attrs.push_back(attr);
     }
 
+    // Enable FIPS POST if needed.
+    // TODO: MACSec POST needs to be eanbled in switch creation. So it's expected to
```
wumiaont: typo "eanbled". Also, will there be another PR to set this macsec_post_enabled flag by checking if FIPS is enabled or not on system?

ysmanman: @wumiaont Once we have decided how to check FIPS is enabled (check STATE_DB or  /etc/fips/fips_enable ), I will update the PR accordingly.

orchagent/macsecorch.cpp
ysmanman: Fixed.

cfgmgr/macsecmgrd.cpp
```diff
@@ -61,6 +73,13 @@ int main(int argc, char **argv)
         swss::DBConnector cfgDb("CONFIG_DB", 0);
         swss::DBConnector stateDb("STATE_DB", 0);
 
+        /* Don't process any config until POST state is ready */
+        SWSS_LOG_NOTICE("checking FIPS POST state");
+        while (!isFipsPostStateReady(&stateDb))
```
vikram-nexthop: why continue with this loop if there is a POST failure (status='fail')?
isn't it appropriate to terminate the MACsec service, in that case?

ysmanman: If macsecmgrd is part of the crucial servers of macsec container, terminating it may lead to macsec container to restart. Given Orchagent already syslog POST failure. Macsec container restart may cause unnecessary alert. Furthermore POST failure should be really rare, so it may be rare to see this in field.

orchagent/main.cpp
ysmanman: Hi @wumiaont @senthil-nexthop ,  with master image, I observed the following:

`0 ` in `/etc/fips/fips_enable`
`sonic_fips=1` in `/proc/cmdline`
`enabled=True` in STATE_DB FIPS_STATS|state 

It seems the value in `/etc/fips/fips_enable`  is not consistent with other configs. Any idea?

rlhui:
please add PR description

ysmanman:
> please add PR description

Added.

orchagent/main.cpp
wumiaont: There are two ways to config fips on system. From cmd or from config. If from command (sudo sonic-installer set-fips --enable-fips" then result will be written to /proc/cmdline. If from config /etc/sonic/fips.json, then result will be updated into /etc/fips/fips_enable. 
We are not doing the config here and that /etc/fips/fips_enable still have 0 there. FIPS from command is currently what we are doing now.

senthil-nexthop: Updating CONFIG_DB does not seem to result in updating /etc/fips/fips_enable, because of a bug in hostcfgd:

```
2025 Sep 10 21:33:29.223950 dut INFO hostcfgd[240733]:   File "/usr/local/bin/hostcfgd", line 2269, in load
2025 Sep 10 21:33:29.224143 dut INFO hostcfgd[240733]:     self.fipscfg.load(fips_cfg)
2025 Sep 10 21:33:29.224162 dut INFO hostcfgd[240733]:   File "/usr/local/bin/hostcfgd", line 1778, in load
2025 Sep 10 21:33:29.224318 dut INFO hostcfgd[240733]:     self.read_config()
2025 Sep 10 21:33:29.224336 dut INFO hostcfgd[240733]:   File "/usr/local/bin/hostcfgd", line 1769, in read_config
2025 Sep 10 21:33:29.224482 dut INFO hostcfgd[240733]:     with open(PROC_CMDLINE) as f:
```
After I added a definition for PROC_CMDLINE, it seems to work correctly:
```
admin@dut:~$ cat /etc/fips/fips_enable
0
admin@dut:~$
admin@dut:~$ redis-cli -n 4 HSET "FIPS|global" "enable" "true"
(integer) 0
admin@dut:~$ cat /etc/fips/fips_enable
1
admin@dut:~$
```
Nevertheless, we can check for fips status in both /etc/fips/fips_enable and /proc/cmdline, and assume switch is in FIPS enabled mode if *either* is true?

cfgmgr/macsecmgrd.cpp
ysmanman: Fixed.

orchagent/main.cpp
ysmanman: Thanks for the clarification and input @senthil-nexthop @wumiaont . As suggested, I updated the change to check if FIPS is enabled either in /etc/fips/fips_enable or proc/cmdline.

cfgmgr/macsecmgrd.cpp
vikram-nexthop: the supervisord attempts to restart macsecmgrd a few times and finally gives up without any further action. the container is not restarted though. Only alert generated is that macsecmgrd has entered a FATAL state.

```
2025 Sep 11 05:14:26.036388 dut INFO macsec#supervisord 2025-09-11 05:14:26,035 WARN exited: macsecmgrd (exit status 1; not expected)
2025 Sep 11 05:14:26.038642 dut INFO macsec#supervisord 2025-09-11 05:14:26,037 INFO gave up: macsecmgrd entered FATAL state, too many start retries too quickly
2025 Sep 11 05:14:26.188364 dut INFO macsec#supervisord 2025-09-11 05:14:26,187 INFO exited: dependent-startup (exit status 0; expected)
```

orchagent/macsecpost.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "table.h"
+
+namespace swss {
+
+void setMacsecPostState(DBConnector *stateDb, string postState)
+{
+    Table macsecPostStateTable = Table(stateDb, STATE_FIPS_MACSEC_POST_TABLE_NAME);
+    vector<FieldValueTuple> fvts;
+    FieldValueTuple postStateFvt("post_state", postState);
+    fvts.push_back(postStateFvt);
```
vikram-nexthop, line 14: it will be useful to include the timestamp.

cfgmgr/macsecmgrd.cpp
ysmanman: Thanks for checking @vikram-nexthop .  In this case,  is it desired to see all these syslogs, without any details, or just one syslog that clearly says SAI POST failed.

orchagent/macsecpost.cpp
ysmanman: Added timestamp.

orchagent/macsecpost.h
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2025 Arista Networks, Inc.  All rights reserved.
```
judyjoseph: Could we remove these 2 lines both from .cpp and .h files.

orchagent/main.cpp
```diff
@@ -341,6 +342,38 @@ bool getSystemPortConfigList(DBConnector *cfgDb, DBConnector *appDb, vector<sai_
...
 }
 
+bool isFipsEnabled()
+{
+    // Check if FIPS was enabled via config file, i.e., /etc/sonic/fips.json
+    std::ifstream fips_enable_file("/etc/fips/fips_enable");
+    if (fips_enable_file.is_open())
+    {
+        bool fips_enabled;
+        fips_enable_file >> fips_enabled;
```
judyjoseph: Just to understand, I see we set the fips_anabled to 0 today here : https://github.com/sonic-net/sonic-buildimage/blob/7cadaa861901099390d5811e6f7f641e28d25f10/build_debian.sh#L691. How do we set to 1 - is it done per platform ? please add to PR description

orchagent/main.cpp
```diff
@@ -341,6 +342,38 @@ bool getSystemPortConfigList(DBConnector *cfgDb, DBConnector *appDb, vector<sai_
...
+        if (fips_enabled)
+        {
+            SWSS_LOG_NOTICE("FIPS enabled in /etc/fips/fips_enable");
+            return true;
+        }
+    }
+
+    // Check if FIPS was enabled via sonic-installer.
+    std::ifstream proc_cmdline_file("/proc/cmdline");
```
judyjoseph: Is this done via build time flag ENABLE_FIPS ? please add to PR description as well thanks

orchagent/main.cpp
wumiaont: It's done by the fips config in /etc/sonic/fips.json. FIPS can be enabled by command (sonic-installer set-fips --enable-fips" or by config.

orchagent/main.cpp
wumiaont: /proc/cmdline (sonic-fips=1) is set by sonic-install command. This requires a reboot after set.

orchagent/main.cpp
```diff
@@ -756,6 +815,35 @@ int main(int argc, char **argv)
...
+                post_capability.create_implemented)
+            {
+                // POST is only supported in MACSec init. Set POST state to notify MACSecOrch
+                // to perform POST.
+                setMacsecPostState(&state_db, "macsec-level-post-in-progress");
+                SWSS_LOG_NOTICE("MACSec POST will be enabled in MACSec init");
+            }
+            else
+            {
+                setMacsecPostState(&state_db, "fail");
```
judyjoseph: should we set it as "disabled" here i.s.o "fail". Also should we log this as ERROR ?

orchagent/main.cpp
```diff
@@ -643,6 +676,32 @@ int main(int argc, char **argv)
         attrs.push_back(attr);
     }
 
+    // Enable MACSec POST if FIPS is enabled.
+    bool macsec_post_enabled = isFipsEnabled();
+
+    string macsec_post_state;
+    if (gMySwitchType != "fabric" && macsec_post_enabled)
```
judyjoseph, line 657: Should we do a capability check first (currently we do it later line num: 818 below) before adding this as attribute in switchcreate ?

orchagent/macsecorch.cpp
```diff
@@ -649,6 +652,53 @@ MACsecOrch::MACsecOrch(
...
+                SWSS_LOG_NOTICE("Switch MACSec POST passed");
+            }
+            else if (attr.value.s32 == SAI_SWITCH_MACSEC_POST_STATUS_FAIL)
+            {
+                setMacsecPostState(m_state_db, "fail");
+                SWSS_LOG_ERROR("Switch MACSec POST failed: oid %" PRIu64, gSwitchId);
+            }
+            else
+            {
+                SWSS_LOG_NOTICE("Switch MACSec POST status: %d", attr.value.s32);
```
judyjoseph, line 687: Any case when switch level macsec POST is still going on, not completed

orchagent/main.cpp
wumiaont: capability check needs  gSwitchId as input which is created by create_switch at line 759. Before create_switch you cannot do capability check.

orchagent/macsecorch.cpp
```diff
@@ -661,6 +711,118 @@ MACsecOrch::~MACsecOrch()
...
+                if (macsec_obj->second.m_ingress_post_passed && macsec_obj->second.m_egress_post_passed)
+                {
+                    setMacsecPostState(m_state_db, "pass");
+                    SWSS_LOG_NOTICE("MACSec POST passed: oid %" PRIu64 ", direction %s", macsec_id, direction.c_str());
+                }
+            }
+            else if(macsec_post_status == SAI_MACSEC_POST_STATUS_FAIL)
+            {
+                // Consider POST failed since it failed on one MACSec object.
+                setMacsecPostState(m_state_db, "fail");
```
judyjoseph, line 828: when the POST fail .. does the SAI return failure for other macsec port related create commands which are being send by macsecorch via _void MACsecOrch::doTask(Consumer &consumer)? 
Thinking of whether we need to clean up if POST fail

orchagent/macsecpost.h
ysmanman: Thanks for catching this. Removed.

orchagent/macsecorch.cpp
ysmanman: @judyjoseph When POST fails, macsecmgrd blocks all macsec config, and thus no macsec related SAI call can be made at all.

orchagent/main.cpp
ysmanman: @judyjoseph `disabled` is a state indicating that FIPS is not enabled in SONiC and thus SAI POST is not required. This is a valid and successful state. However, this case may be a failure/mistake because FIPS is enabled but SAI does not support POST.  This was the reason why I used `fail` and also log it as error.

judyjoseph (dismissed):
LGTM, @ysmanman, can you update the branch + add test results in the PR summary - thx

rlhui:
can we please rebase

orchagent/main.cpp
```diff
@@ -341,6 +342,38 @@ bool getSystemPortConfigList(DBConnector *cfgDb, DBConnector *appDb, vector<sai_
     return true;
 }
 
+bool isFipsEnabled()
+{
+    // Check if FIPS was enabled via config file, i.e., /etc/sonic/fips.json
+    std::ifstream fips_enable_file("/etc/fips/fips_enable");
```
prsunny: is there any config_db entry that says FIPS enabled?

ysmanman:
Hi @judyjoseph @wumiaont @senthil-nexthop ,  `sonic_fips=1` is always set in /proc/cmdline in master. Is this expected? With this setting,  this PR will always enable POST in SAI switch creation, which fails. As a result, MACSec is not working.

kenneth-arista:
> can we please rebase

Based on the GitHub checks, the "changes can be cleanly merged" without having to rebase.

orchagent/main.cpp
ysmanman: Hi @prsunny , according to the previous discussion and comments, we don't have anything in config_db that indicates if FIPS is enabled or not today. @wumiaont @senthil-nexthop can confirm.

wumiaont: By default, there should have no FIPS config in CONFIG_DB.

senthil-nexthop:
> Hi @judyjoseph @wumiaont @senthil-nexthop , `sonic_fips=1` is always set in /proc/cmdline in master. Is this expected? With this setting, this PR will always enable POST in SAI switch creation, which fails. As a result, MACSec is not working.

My understanding is that the command "sonic-installer set-fips" will set sonic_fips=1 in kernel command line. Are you saying this is always set in the image during build? Is this a recent change?

ysmanman:
> > Hi @judyjoseph @wumiaont @senthil-nexthop , `sonic_fips=1` is always set in /proc/cmdline in master. Is this expected? With this setting, this PR will always enable POST in SAI switch creation, which fails. As a result, MACSec is not working.
> 
> My understanding is that the command "sonic-installer set-fips" will set sonic_fips=1 in kernel command line. Are you saying this is always set in the image during build? Is this a recent change?

Hi @senthil-nexthop , I noticed this with our latest internal image built with upstream master changes. I am not aware of any internal patch to enable FIPS. So this setting is likely from upstream.

ysmanman:
> > > Hi @judyjoseph @wumiaont @senthil-nexthop , `sonic_fips=1` is always set in /proc/cmdline in master. Is this expected? With this setting, this PR will always enable POST in SAI switch creation, which fails. As a result, MACSec is not working.
> > 
> > 
> > My understanding is that the command "sonic-installer set-fips" will set sonic_fips=1 in kernel command line. Are you saying this is always set in the image during build? Is this a recent change?
> 
> Hi @senthil-nexthop , I noticed this with our latest internal image built with upstream master changes. I am not aware of any internal patch to enable FIPS. So this setting is likely from upstream.

Hi @judyjoseph @wumiaont @senthil-nexthop ,  we checked internally and found that we do have internal patch that may enable FIPS in our SONiC build. This is probably the reason that we observed `sonic_fips=1` in /proc/cmdline in our internally built image. Sorry for the false alarm.  

Based on today's discussion,  we are going to revise the PR to follow what @rlhui suggested. That's Orchagent sets POST result based on SAI capabilities, but not global FIPS config (e.g., `sonic_fips` in /proc/cmdline). So in the same scenario here (FIPS is enabled but SAI does not support POST), Orchagent will set POST result to `disabled` (MACSec is not disabled),  instead of `failed`.  Orchagent sets POST result to `failed` only when SAI POST is supported and fails.

judyjoseph:
> > > > Hi @judyjoseph @wumiaont @senthil-nexthop , `sonic_fips=1` is always set in /proc/cmdline in master. Is this expected? With this setting, this PR will always enable POST in SAI switch creation, which fails. As a result, MACSec is not working.
> > > 
> > > 
> > > My understanding is that the command "sonic-installer set-fips" will set sonic_fips=1 in kernel command line. Are you saying this is always set in the image during build? Is this a recent change?
> > 
> > 
> > Hi @senthil-nexthop , I noticed this with our latest internal image built with upstream master changes. I am not aware of any internal patch to enable FIPS. So this setting is likely from upstream.
> 
> Hi @judyjoseph @wumiaont @senthil-nexthop , we checked internally and found that we do have internal patch that may enable FIPS in our SONiC build. This is probably the reason that we observed `sonic_fips=1` in /proc/cmdline in our internally built image. Sorry for the false alarm.
> 
> Based on today's discussion, we are going to revise the PR to follow what @rlhui suggested. That's Orchagent sets POST result based on SAI capabilities, but not global FIPS config (e.g., `sonic_fips` in /proc/cmdline). So in the same scenario here (FIPS is enabled but SAI does not support POST), Orchagent will set POST result to `disabled` (MACSec is not disabled), instead of `failed`. Orchagent sets POST result to `failed` only when SAI POST is supported and fails.

sure @ysmanman had a comment earlier on. But please add these details in the PR description.

orchagent/main.cpp
prsunny: @judyjoseph , can you chime in? Why there is no config_db entry for FIPS enabled or not?

ysmanman:
> > > > > Hi @judyjoseph @wumiaont @senthil-nexthop , `sonic_fips=1` is always set in /proc/cmdline in master. Is this expected? With this setting, this PR will always enable POST in SAI switch creation, which fails. As a result, MACSec is not working.
> > > > 
> > > > 
> > > > My understanding is that the command "sonic-installer set-fips" will set sonic_fips=1 in kernel command line. Are you saying this is always set in the image during build? Is this a recent change?
> > > 
> > > 
> > > Hi @senthil-nexthop , I noticed this with our latest internal image built with upstream master changes. I am not aware of any internal patch to enable FIPS. So this setting is likely from upstream.
> > 
> > 
> > Hi @judyjoseph @wumiaont @senthil-nexthop , we checked internally and found that we do have internal patch that may enable FIPS in our SONiC build. This is probably the reason that we observed `sonic_fips=1` in /proc/cmdline in our internally built image. Sorry for the false alarm.
> > Based on today's discussion, we are going to revise the PR to follow what @rlhui suggested. That's Orchagent sets POST result based on SAI capabilities, but not global FIPS config (e.g., `sonic_fips` in /proc/cmdline). So in the same scenario here (FIPS is enabled but SAI does not support POST), Orchagent will set POST result to `disabled` (MACSec is not disabled), instead of `failed`. Orchagent sets POST result to `failed` only when SAI POST is supported and fails.
> 
> sure @ysmanman had a comment earlier on. But please add these details in the PR description.

Thanks @judyjoseph . I added the details and also testing result in PR description.

orchagent/main.cpp
prsunny: @ysmanman , resolve the comment after its been replied/closed

judyjoseph: @prsunny there is a PR and HLD which tells about the config_db config for fips, https://github.com/sonic-net/sonic-buildimage/pull/15632. https://github.com/sonic-net/SONiC/pull/1372 

As per this PR https://github.com/sonic-net/sonic-host-services/pull/69, we need this file  /etc/sonic/fips.json to control enable/enforce FIPS. But there are questions like who creates it, Is it does per platform which supports FIPS byt platform sku vendor? @xumia @qiluo-msft to comment

judyjoseph:
> > Hi @judyjoseph @wumiaont @senthil-nexthop , `sonic_fips=1` is always set in /proc/cmdline in master. Is this expected? With this setting, this PR will always enable POST in SAI switch creation, which fails. As a result, MACSec is not working.
> 
> My understanding is that the command "sonic-installer set-fips" will set sonic_fips=1 in kernel command line. Are you saying this is always set in the image during build? Is this a recent change?

@ysmanman can we make it default disabled and set "sonic_fips=1" in kernel command line only if we call sonic-installer --set-fips

wumiaont:
> > > Hi @judyjoseph @wumiaont @senthil-nexthop , `sonic_fips=1` is always set in /proc/cmdline in master. Is this expected? With this setting, this PR will always enable POST in SAI switch creation, which fails. As a result, MACSec is not working.
> > 
> > 
> > My understanding is that the command "sonic-installer set-fips" will set sonic_fips=1 in kernel command line. Are you saying this is always set in the image during build? Is this a recent change?
> 
> @ysmanman can we make it default disabled and set "sonic_fips=1" in kernel command line only if we call sonic-installer --set-fips

@judyjoseph This is a false alarm. With master image build, FIPS is not enabled with image.

abdosi:
@judyjoseph to drive the merge of this PR.

orchagent/main.cpp
prsunny: @judyjoseph , can you try the approach of passing this option as startup param - Refer ` void usage() `

judyjoseph: @ysmanman the suggestion is to add a startup option here https://github.com/sonic-net/sonic-swss/blob/0e1558e31c1afd9847a1f77035c1f20577c843ab/orchagent/main.cpp#L80, to enable the sonic fips and pass this option here **dockers/docker-orchagent/orchagent.sh** when we start orchagent. This will avoid parsing of files in orchagent.

judyjoseph:
> > > > Hi @judyjoseph @wumiaont @senthil-nexthop , `sonic_fips=1` is always set in /proc/cmdline in master. Is this expected? With this setting, this PR will always enable POST in SAI switch creation, which fails. As a result, MACSec is not working.
> > > 
> > > 
> > > My understanding is that the command "sonic-installer set-fips" will set sonic_fips=1 in kernel command line. Are you saying this is always set in the image during build? Is this a recent change?
> > 
> > 
> > @ysmanman can we make it default disabled and set "sonic_fips=1" in kernel command line only if we call sonic-installer --set-fips
> 
> @judyjoseph This is a false alarm. With master image build, FIPS is not enabled with image.

@ysmanman another question. We too have internal builds with sonic_fips enabled, which will cause MACSEC sessions to not come up in our testbeds too. Do we have the SAI which supports this FIPS POST already merged in master ? cc. @vmittal-msft

ysmanman:
> > > > > Hi @judyjoseph @wumiaont @senthil-nexthop , `sonic_fips=1` is always set in /proc/cmdline in master. Is this expected? With this setting, this PR will always enable POST in SAI switch creation, which fails. As a result, MACSec is not working.
> > > > 
> > > > 
> > > > My understanding is that the command "sonic-installer set-fips" will set sonic_fips=1 in kernel command line. Are you saying this is always set in the image during build? Is this a recent change?
> > > 
> > > 
> > > @ysmanman can we make it default disabled and set "sonic_fips=1" in kernel command line only if we call sonic-installer --set-fips
> > 
> > 
> > @judyjoseph This is a false alarm. With master image build, FIPS is not enabled with image.
> 
> @ysmanman another question. We too have internal builds with sonic_fips enabled, which will cause MACSEC sessions to not come up in our testbeds too. Do we have the SAI which supports this FIPS POST already merged in master ? cc. @vmittal-msft

Hi @judyjoseph according to BRCM,  SAI 14.x and SDK 6.5.34 should support FIPS POST for MACSec.  But I don't think we have integrated them into master.

ysmanman:
> > > > > > Hi @judyjoseph @wumiaont @senthil-nexthop , `sonic_fips=1` is always set in /proc/cmdline in master. Is this expected? With this setting, this PR will always enable POST in SAI switch creation, which fails. As a result, MACSec is not working.
> > > > > 
> > > > > 
> > > > > My understanding is that the command "sonic-installer set-fips" will set sonic_fips=1 in kernel command line. Are you saying this is always set in the image during build? Is this a recent change?
> > > > 
> > > > 
> > > > @ysmanman can we make it default disabled and set "sonic_fips=1" in kernel command line only if we call sonic-installer --set-fips
> > > 
> > > 
> > > @judyjoseph This is a false alarm. With master image build, FIPS is not enabled with image.
> > 
> > 
> > @ysmanman another question. We too have internal builds with sonic_fips enabled, which will cause MACSEC sessions to not come up in our testbeds too. Do we have the SAI which supports this FIPS POST already merged in master ? cc. @vmittal-msft
> 
> Hi @judyjoseph according to BRCM, SAI 14.x and SDK 6.5.34 should support FIPS POST for MACSec. But I don't think we have integrated them into master.

Hi @judyjoseph One more clarification here. With the latest change in the PR,  Orchagent does not declare SAI POST to fail if SAI does not support POST. In this case, MACSec is still enabled. This allows to use MACSec with existing SAI, e.g., 13.x, that does not support SAI POST. In other words, the change is the PR is just no-op with existing SAI.

judyjoseph (approved):
LGTM - thanks @ysmanman

judyjoseph:
@prsunny can you review again and merge if ok

prsunny (approved):
lgtm, @judyjoseph please signoff

prsunny:
@ysmanman , thanks for addressing the comments

PR #3838, opened 2025-08-19
[muxorch] Fix handling mux neighbors learned after route

Ndancejic:
Mux neighbors that were learned in standby state when a route entry is present were not being disabled properly due to still being referenced by the route.

This change calls updateNextHopRoutes and decreases the nexthop refcount before disabling the neighbor, allowing the neighbor to be removed.

Also applies fix that passes in the port_name on initial fdb add.

**What I did**
- Update routes pointing to newly learned neighbor in standby state and update refcount
- Update refcount for routes updated when neighbor is learned in active state
- Moved update logic to after neighbor gets added to mux port
- Pass port name when adding Fdb entry
- Add vstest to test changes

**Why I did it**
Neighbors added while a route is present were not being disabled due to nonzero refcounts. This caused traffic to blackhole and dualtor_neighbor_check.py to fail due to mismatch between APPL and ASIC state.

**How I verified it**
Added vstest with test cases of adding neighbor while route exists in standby and active and toggling between the two states.

**Details if related**
ado: #34501086

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -652,7 +652,6 @@ void MuxCable::updateNeighbor(NextHopKey nh, bool add)
     SWSS_LOG_NOTICE("Processing update on neighbor %s for mux %s, add %d, state %d",
                      nh.ip_address.to_string().c_str(), mux_name_.c_str(), add, state_);
     sai_object_id_t tnh = mux_orch_->getNextHopTunnelId(MUX_TUNNEL, peer_ip4_);
-    nbr_handler_->update(nh, tnh, add, state_);
     if (add)
     {
         mux_orch_->addNexthop(nh, mux_name_);
@@ -661,6 +660,7 @@ void MuxCable::updateNeighbor(NextHopKey nh, bool add)
     {
         mux_orch_->removeNexthop(nh);
     }
+    nbr_handler_->update(nh, tnh, add, state_);
     updateRoutes();
 }
 
@@ -716,9 +716,12 @@ void MuxNbrHandler::update(NextHopKey nh, sai_object_id_t tunnelId, bool add, Mu
             neighbors_[nh.ip_address] = gNeighOrch->getLocalNextHopId(nh);
             gNeighOrch->enableNeighbor(nh);
             gRouteOrch->updateNextHopRoutes(nh, num_routes);
+            gNeighOrch->increaseNextHopRefCount(nh, num_routes);
             break;
         case MuxState::MUX_STATE_STANDBY:
             neighbors_[nh.ip_address] = tunnelId;
+            gRouteOrch->updateNextHopRoutes(nh, num_routes);
+            gNeighOrch->decreaseNextHopRefCount(nh, num_routes);
             gNeighOrch->disableNeighbor(nh);
             updateTunnelRoute(nh, true);
             create_route(pfx, tunnelId);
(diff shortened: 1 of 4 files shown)


prsunny:
@Ndancejic , i think the main issue is neighbor and route learned before any mux configuration. Is that handled? Say the following sequence:
1. Learn Nbr
2. Learn route->nbr
3. configure mux port
4. set to standby

Ndancejic:
I'll add a test case to check this. in the active issue we were seeing, I don't think the mux was configured after this sequence but I agree this should also be handled. I'll implement the fix for this case as well.

orchagent/neighorch.cpp
```diff
@@ -1005,6 +1005,31 @@ void NeighOrch::doTask(Consumer &consumer)
...
+    {
+        // Check if mux port exists for given neighbor entry
+        mux_port_name = "";
+        if (!mux_orch->getMuxPort(entry.second.mac, entry.first.alias, mux_port_name) || mux_port_name.empty())
+        {
+            continue;
+        }
+
+        // Add to m_neighbors if entry found
+        SWSS_LOG_INFO("mux_port: %s, port_name: %s", mux_port_name.c_str(), port_name.c_str());
```
prsunny: convert this to NOTICE

orchagent/muxorch.cpp
```diff
@@ -1529,11 +1532,20 @@ void MuxOrch::updateNeighbor(const NeighborUpdate& update)
         auto it = mux_nexthop_tb_.find(update.entry);
         if (it != mux_nexthop_tb_.end())
         {
+            SWSS_LOG_INFO("port %s, nexthop %s", port.c_str(), it->second.c_str());
             port = it->second;
             removeNexthop(update.entry);
         }
     }
 
+    SWSS_LOG_INFO("old port: %s, port %s, add %d, mac %s, alias %s, ip %s",
```
prsunny: Remove logs added for debug

tests/test_mux.py
```diff
@@ -1517,6 +1556,49 @@ def test_Peer(self, dvs, setup_peer_switch, setup_tunnel, setup, testlog):
 
         self.create_and_test_peer(asicdb, encap_tc_to_dscp_map_id, encap_tc_to_queue_map_id)
 
+    def test_neighbor_learned_before_mux_config(self, dvs, dvs_route, setup, setup_vlan, setup_peer_switch, setup_tunnel, testlog):
```
prsunny, line 1559: Give comment on the test steps:
1. Config Nbr, route x 2
2. Mux config 1-> Active 2->standby
3. Switchover from 1: A->S , 2: S->A

tests/test_mux.py
```diff
@@ -1517,6 +1556,49 @@ def test_Peer(self, dvs, setup_peer_switch, setup_tunnel, setup, testlog):
...
+                    self.check_neigh_in_asic_db(asicdb, test_ip_v6, expected=True)
+
+                    if create_route:
+                        self.add_route(dvs, "11.11.11.11/32", ["192.168.0.100"])
+
+                    fvs = {"server_ipv4": self.SERV2_IPV4+self.IPV4_MASK,
+                        "server_ipv6": self.SERV2_IPV6+self.IPV6_MASK}
+                    config_db.create_entry(self.CONFIG_MUX_CABLE, "Ethernet4", fvs)
+
+                    self.set_mux_state(appdb, "Ethernet4", state)
```
prsunny: Test by reversing the config

Ndancejic: reversing the config also works

PR #3839, opened 2025-08-21
Add mock test for IPv4/IPv6 route add/delete

meghtony:
**What I did**
I added a single GoogleTest RouteOrch_AddRemoveIPv4_And_DefaultRoute_State that covers both adding/removing 2.2.2.0/24 and the default-route. I also added small helpers to probe STATE_DB for the route state field and to wait for ok/na when available.

**Why I did it**
This mirrors the pytest intent while fitting the mock unit test harness that pre-seeds ports, interfaces, neighbors, and a default route in SetUp(). It keeps the test fast and deterministic by verifying behavior via SAI mocks/counters rather than relying on DVS/ASIC_DB.

**How I verified it**
Will be seeing if the pipeline fails.

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -923,7 +923,7 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
         }
         else if (attr_name == MATCH_TUNNEL_TERM)
         {
-            matchData.data.booldata = (attr_name == "true");
+            matchData.data.booldata = (to_upper(attr_value) == "TRUE");;
         }
         else if (attr_name == MATCH_INNER_DST_MAC || attr_name == MATCH_INNER_SRC_MAC)
         {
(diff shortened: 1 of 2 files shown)


tests/mock_tests/routeorch_ut.cpp
```diff
@@ -397,6 +429,213 @@ namespace routeorch_test
...
+            // Expect remove +1, create/set unchanged
+            ASSERT_EQ(base_create,         create_route_count);
+            ASSERT_EQ(base_set,            set_route_count);
+            ASSERT_EQ(base_remove + 1,     remove_route_count);
+        }
+
+        // PART B: Default route DEL -> state 'na' -> SET -> state 'ok'
+        {
+            const std::string def = "0.0.0.0/0";
+            const bool hasStateField = stateRouteStateFieldExists(m_state_db.get(), def);
```
prabhataravind: I think this should always exist. Could you add an assertion for this?

prsunny:
@meghtony , would you also mark the corresponding pytest skipped in this same PR? @prabhataravind to signoff as well.

prabhataravind (approved):

PR #3851, opened 2025-08-28
[ssw][ha] add `ha_term` field to ha_scope_state_db

zjswhhh:
**What I did**
Take `flow_version` field from ha scope event and write it into dpu_state_db entry `ha_term` per HLD.

https://github.com/opencomputeproject/SAI/blob/b925fc0f0fc8c620eea7d9dead7fc99e33b7b44e/experimental/saiswitchextensions.h#L113

**Why I did it**
Otherwise hamgrd won't be able to deserialize the state db table (due to missing fields).

**How I verified it**
UTs.

**Details if related**


zjswhhh:
Hi @qiluo-msft, can you help merge the PR? There is only one line uncovered in UTs, and honestly, I think it's impossible that it's uncovered...

{405EE955-265E-49DA-829A-84BF4C4683CD}

PR #3860, opened 2025-09-03
[fpmsyncd]: Fix SRv6 SID List sharing among prefixes

cscarpitta:
Currently, there is a one-to-one mapping between prefixes and SRv6 SID List. When fpmsyncd receives a new route, it creates a SID List and associates it with the route. The key for the SID List is the Prefix of route. If fpmsyncd receives a second route pointing to the same SRv6 SID List, fpmsyncd creates another copy of the SID List for the second route.

This approach results in unnecessary duplication of SID Lists for routes that could share the same SID List, and lead to significant scale issues. The number of SID Lists = the number of routes.

This commit fixes the issue in fpmsyncd by enabling it to reuse existing SID Lists.

### DIFF
--- a/cfgmgr/buffer_pool_mellanox.lua
+++ b/cfgmgr/buffer_pool_mellanox.lua
@@ -22,6 +22,7 @@ local total_port = 0
 
 local mgmt_pool_size = 256 * 1024
 local egress_mirror_headroom = 10 * 1024
+local modification_descriptors_pool_size = 0
 
 -- The set of ports with 8 lanes
 local port_set_8lanes = {}
@@ -185,6 +186,22 @@ end
 -- Main --
 -- Connect to CONFIG_DB
 redis.call('SELECT', config_db)
+
+-- Check if platform is SPC6 or later and set modification descriptors pool size
+-- Extract model number from platform string (e.g., "sn6600" -> 6600, "sn5800" -> 5800, "sn10600" -> 10600)
+-- Use (%d+) pattern to capture one or more digits for extensibility (handles future multi-digit series like sn10xxx, sn11xxx)
+local platform = redis.call('HGET', 'DEVICE_METADATA|localhost', 'platform')
+if platform then
+    local model_str = string.match(platform, "sn(%d+)")
+    if model_str then
+        local model_number = tonumber(model_str)
+        -- SPC6 or later models (>= 6000 excludes SPC5 models like 5400/5800, includes SPC6+ like 6600/7xxx/10xxx)
+        -- Reserve 32MB for modification descriptors pool
+        if model_number and model_number >= 6000 then
+            modification_descriptors_pool_size = 32 * 1024 * 1024
+        end
+    end
+end
 
 -- Parse all the pools and seperate them according to the direction
 local ipools = {}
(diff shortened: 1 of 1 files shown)


dgsudharsan:
@cscarpitta @ahsalam I see there are only cosmetic code changes and then UT changes. Is there any functionality change here?

dgsudharsan (approved):

fpmsyncd/routesync.cpp
```diff
@@ -1081,7 +1081,7 @@ void RouteSync::onSrv6SteerRouteMsg(struct nlmsghdr *h, int len)
     if (nlmsg_type == RTM_DELROUTE)
     {
         string routeTableKeyStr = string(routeTableKey);
-        string srv6SidListTableKey = routeTableKeyStr;
+        string srv6SidListTableKey = vpn_sid_str;
```
dgsudharsan, line 1223: With this new approach, should we maintain some reference count so that we don't delete the table when there are references?

fpmsyncd/routesync.cpp
```diff
@@ -1095,7 +1095,7 @@ void RouteSync::onSrv6SteerRouteMsg(struct nlmsghdr *h, int len)
         string routeTableKeyStr = string(routeTableKey);
         /* Write SID list to SRV6_SID_LIST_TABLE */
 
-        string srv6SidListTableKey = routeTableKeyStr;
+        string srv6SidListTableKey = vpn_sid_str;
```
guoguojia2021, line 1098: If there are multiple VPN SIDs, how to separate them?

guoguojia2021, line 1262: Hi Carmine,
I'm a bit confused about the SID field in this table.Does it represent the policy's SID(sid list) or the VPN's SID? Also, does the current code support only one SID?

fpmsyncd/routesync.cpp
cscarpitta: @dgsudharsan Thanks for the review.

I added the refcount for the SID lists.

fpmsyncd/routesync.cpp
cscarpitta: This is the SID list.

The current implementation can handle only one SID, but this is a different problem. I'm already working on another PR to fix this issue and allow fpmsyncd to properly handle SID lists with multiple SID. I will open a separate PR once this PR gets merged.

prsunny:
@dgsudharsan , @abdosi , please review/signoff

ahsalam (approved):
LGTM

prsunny:
@dgsudharsan , can you approve and merge?

ahsalam:
@dgsudharsan can you help to merge the PR ? it passed all CI checks.

prsunny:
@dgsudharsan , would you review/merge?

dgsudharsan (approved):

fpmsyncd/routesync.h
```diff
@@ -250,6 +250,8 @@ class RouteSync : public NetMsg
     /* nexthop group table */
     ProducerStateTable  m_nexthop_groupTable;
     map<uint32_t,NextHopGroup> m_nh_groups;
+    /* SID list to refcount */
+    map<string, uint32_t> m_srv6_sidlist_refcnt;
```
venkit-nexthop, line 254: shouldnt we be using std::unordered_map for quick lookups of the refcount?

PR #3866, opened 2025-09-05
Fix switch capabilities enum query with metadata validation

dhanasekar-arista:
**What I did**
Fix the queryEnumCapabilitiesSai function to use sai_metadata_get_attr_metadata for fetching the size of the buffer.

**Why I did it**
queryEnumCapabilitiesSai was using sai_query_attribute_enum_values_capability as a bait to fetch the actual count even though it was throwing an error SAI_STATUS_BUFFER_OVERFLOW. This was not logged as error in SAI until the latest release, hence it got overlooked.
The recent SAI upgrade explicitly throws this as an error, hence i've fixed it.

**How I verified it**
Ensuring these error logs are not thrown post fix.

**Details if related**

### DIFF
--- a/orchagent/switch/switch_capabilities.cpp
+++ b/orchagent/switch/switch_capabilities.cpp
@@ -5,6 +5,7 @@ extern "C" {
 #include <saiobject.h>
 #include <saistatus.h>
 #include <saihash.h>
+#include <saimetadata.h>
 #include <saiswitch.h>
 }
 
@@ -314,16 +315,20 @@ std::vector<FieldValueTuple> SwitchCapabilities::makeLagHashAlgorithmCapDbEntry(
 
 sai_status_t SwitchCapabilities::queryEnumCapabilitiesSai(std::vector<sai_int32_t> &capList, sai_object_type_t objType, sai_attr_id_t attrId) const
 {
-    sai_s32_list_t enumList = { .count = 0, .list = nullptr };
-
-    auto status = sai_query_attribute_enum_values_capability(gSwitchId, objType, attrId, &enumList);
-    if ((status != SAI_STATUS_SUCCESS) && (status != SAI_STATUS_BUFFER_OVERFLOW))
+    const auto* meta = sai_metadata_get_attr_metadata(objType, attrId);
+    if (!meta || !meta->isenum)
     {
-        return status;
+        SWSS_LOG_ERROR("Failed to get enum metadata for objType:%d attrId:%d", objType, attrId);
+        return SAI_STATUS_FAILURE;
     }
 
-    capList.resize(enumList.count);
-    enumList.list = capList.data();
+    SWSS_LOG_DEBUG("Capability size for objType:%d attrId:%d is count:%d", objType, attrId, meta->enummetadata->valuescount);
+    // Pre-allocate based on metadata to avoid buffer overflow
+    capList.resize(meta->enummetadata->valuescount);
+    sai_s32_list_t enumList = { 
+        .count = static_cast<uint32_t>(capList.size()),
+        .list = capList.data()
+    };
 
     return sai_query_attribute_enum_values_capability(gSwitchId, objType, attrId, &enumList);
 }


dhanasekar-arista:
[### ](https://dev.azure.com/mssonic/build/_build?definitionId=15)
these checkers have been failing since last week for all PRs, it looks like infra issue.

StormLiangMS:
hi @kperumalbfn  @prabhataravind could you take a look?

orchagent/switch/switch_capabilities.cpp
```diff
@@ -314,16 +315,20 @@ std::vector<FieldValueTuple> SwitchCapabilities::makeLagHashAlgorithmCapDbEntry(
...
-        return status;
+        SWSS_LOG_ERROR("Failed to get enum metadata for objType:%d attrId:%d", objType, attrId);
+        return SAI_STATUS_FAILURE;
     }
 
-    capList.resize(enumList.count);
-    enumList.list = capList.data();
+    SWSS_LOG_DEBUG("Capability size for objType:%d attrId:%d is count:%d", objType, attrId, meta->enummetadata->valuescount);
+    // Pre-allocate based on metadata to avoid buffer overflow
+    capList.resize(meta->enummetadata->valuescount);
```
kperumalbfn, line 327: @dhanasekar-arista what is the main difference between sai_query_attribute_enum_values_capability and sai_metadata_get_attr_metadata in getting the enum list? 

If there are any custom enums inside vendor SAI library, how does sai_metadata_get_attr_metadata() work?

dhanasekar-arista: The intention of this change is to not replace _sai_query_attribute_enum_values_capability_ with _sai_metadata_get_attr_metadata_ to get the enum list.  
_sai_query_attribute_enum_values_capability_ expects the maximum size of enum list to be provided while querying the asic, hence i have made a change to fetch the size using _sai_metadata_get_attr_metadata_.

I am not very sure about the impact custom enums would have here. do you believe that custom enums will not be returned in  sai_metadata_get_attr_metadata?

kperumalbfn: Could you pls verify that?

orchagent/switch/switch_capabilities.cpp
```diff
@@ -314,16 +315,20 @@ std::vector<FieldValueTuple> SwitchCapabilities::makeLagHashAlgorithmCapDbEntry(
...
 {
-    sai_s32_list_t enumList = { .count = 0, .list = nullptr };
-
-    auto status = sai_query_attribute_enum_values_capability(gSwitchId, objType, attrId, &enumList);
-    if ((status != SAI_STATUS_SUCCESS) && (status != SAI_STATUS_BUFFER_OVERFLOW))
+    const auto* meta = sai_metadata_get_attr_metadata(objType, attrId);
+    if (!meta || !meta->isenum)
     {
-        return status;
+        SWSS_LOG_ERROR("Failed to get enum metadata for objType:%d attrId:%d", objType, attrId);
```
yxieca, line 321: The if condition says NOT enum? Is the error message accurate?

r12f: I think this error message is correct. It means either the attribute is not found, or the attribute is found but not an enum.

r12f:
hi @dhanasekar-arista , do you mind to help check Kumaresh's comment? with this fixed, it will help us fixing 21 different test cases.

orchagent/switch/switch_capabilities.cpp
dhanasekar-arista: @kperumalbfn thanks for pointing this out, i couldn't find any reference which mandates the requirement of custom enums should be published or exposed in statically compiled metadata. Hence we should not proceed with this change.
I’ve raised a CSP(CS00012428372) to explore a more suitable approach for handling this scenario. If anyone is already aware of existing guidance or precedents, please do share.

dhanasekar-arista:
The fix posted is not generic enough to handle custom enums inside vendor SAI library. Hence im closing this issue for now. Will wait for update on the CS00012428372 from BRCM and fix it appropriately.

FYI, we already have a fix in loganalyzer to ignore this error https://github.com/sonic-net/sonic-mgmt/pull/20605/files

PR #3872, opened 2025-09-08
Make fpmsyncd capable of handling route replace semantics

venkit-nexthop:
ISSUE: https://github.com/sonic-net/sonic-buildimage/issues/23940

The FRR's zebra has two dataplane providers. One of them programs routes into the kernel, while the other one sends routes towards sonic. This one sends routes to fpmsyncd. Currently, this sonic dataplane provider does not make use of the route replace semantics. This means that any change to an existing route is first sent as a delete and then a new add notification is sent later. The ProducerStateTable mechanism is used by fpmsyncd to update APPL DB. This coalesces the delete+update into a single set operation and hence orchagent only sees a set. This checkin makes code changes to fpmsyncd so that it can handle a single set operation from the sonic dataplane provider. The crux of the change is to reset any field that is not being set very explicitly. This will have the effect of nullifying any previously set values in that field.
A followup PR would add support for use-route-replace semantics in the FRR's sonic dataplane provider.

**What I did**
Changed the fpmsyncd code to explicitly reset any fields in the ROUTE_TABLE is not being set. This is achieved using the RouteTableFieldValueTupleWrapper class.

**Why I did it**
This will allow fpmsyncd to support route replace semantics thus paving the way for using that in FRR and making things much more efficient. 
This also helps when there are ECMP changes (ecmp growth or shrink) and allows for zero traffic drops. The earlier code was not always coalescing the deletes+add effectively for all the routes.

**How I verified it**
Verified with ECMP routes with growth or shrink of ECMP.

**Details if related**
This change is safe because the ProducerStateTable mechanism used by fpmsyncd to update APPL DB already does coalescing of route changes. It was just not doing it effectively.

### DIFF
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -2895,7 +2895,6 @@ bool RouteOrch::removeRoutePost(const RouteBulkContext& ctx)
                         removeNextHopRoute(*nh, routekey);
                     }
                 }
-                mux_orch->updateRoute(ipPrefix);
             }
         }
         else if (ol_nextHops.is_overlay_nexthop())


fpmsyncd/routesync.cpp
```diff
@@ -1704,32 +1726,26 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
...
         }
         if (!weights.empty())
         {
-            FieldValueTuple wt("weight", weights);
-            fvVector.push_back(wt);
+            fvw.weight = std::move(weights);
         }
     }
 
+    setRouteWithWarmRestart(destipprefix, fvw.fieldValueTupleVector(), m_routeTable, SET_COMMAND);
```
eddieruan-alibaba: Can you add your current routes' key and values in app db? 

If you add these unused field in appdb, that would take redisdb write bw which would slow down routing downloading time.

eddieruan-alibaba: I would suggest to treat each incoming update as an atomic update for routes, a.k.a unspecified felids mean unused.

venkit-nexthop: Thanks for the review Eddie. Unfortunately the current interface does not treat the updates to a key as overwriting all the fields. It only updates the fields that are specified and leaves any previously written fields as is.
For example if the route was updated earlier with "nexthop-id" field, then fpmsyncd would have written that into redis (route-table in APP_DB). Later if the route sends "nexthop" field, we need to atomically remove "nexthop-id" and add "nexthop". But the current interface does not allow that (atleast not to my knowledge). This is the reason we need to explicitly SET empty strings for the other fields.

Regarding route download performance, I measured it and did not see any change. The significant overhead comes in serializing the routes and nexthops into strings (to be written into redis) and later deserializing them back again in orchagent. There is also a lot of overhead in the way ProducerStateTable mechanism uses lua scripts to update redis. This extra additions are quite small and does not contribute much to the overhead. Also the savings in route-update (which avoids the DEL + ADD) is somewhat big.

eddieruan-alibaba: Can you check it again? We saw performance penalty when we have extra fields to set as "" or "n/a" before. That is due to redis io limitation to update fields.

venkit-nexthop: Sure, I will check once again.

venkit-nexthop: @eddieruan-alibaba Thanks for your comment. You were right. There was a decrease in performance with setting empty strings. 

I found the right way to do this now. The solution is to use the batched version of set() for the table. This allows for atomically deleting the key and re-adding it with only the required fields. So no more empty strings (""). I have refactored the code to do it now.

Performance measurements
Image with original version of the code:
  ▎   Total routes injected: 100,000
  ▎   Total benchmark time: 39.800s

Image with empty strings:
  ▎  Total routes injected: 100,000
  ▎  Total benchmark time: 42.573s
  
Image with my latest code:
  ▎   Total routes injected: 100,000
  ▎   Total benchmark time: 38.676s

wen587:
pls pull lastest change and rerun.

venkit-nexthop:
> pls pull lastest change and rerun.

Thanks. I have pulled the latest changes now.

fpmsyncd/routesync.cpp
eddieruan-alibaba: My understanding is that you use the following vector to batch DEL and SET actions 


    vector<KeyOpFieldsValuesTuple> KeyOpFieldsValuesTupleVector() {
        vector<KeyOpFieldsValuesTuple> kfvVector;
        kfvVector.push_back(KeyOpFieldsValuesTuple {key.c_str(), "DEL", {}});
        auto fvVector = fieldValueTupleVector();
        kfvVector.push_back(KeyOpFieldsValuesTuple {key.c_str(), "SET", fvVector});
        return kfvVector;
    }

Then use ProducerStateTable's  batch set to make sure DEL and SET are handled together instead of current two actions. 

void ProducerStateTable::set(const std::vector<KeyOpFieldsValuesTuple>& values)

This way, you don't need to set extra fields.

Is that correct understanding?

venkit-nexthop: Hi Eddie,
Yes your understanding is correct.

eddieruan-alibaba:
@judyjoseph @prsunny 

This is the main PR.

venkit-nexthop:
> @judyjoseph @prsunny
> 
> This is the main PR.

This PR represents the fpmsyncd part of the change. There is one more PR that is in sonic-buildimage to change the sonic fpm dataplane provider code. It can only be done after this one merges.

venkit-nexthop:
@prsunny @dgsudharsan could you approve this PR if you think the changes are ok.
@eddieruan-alibaba has already approved this. The "Semgrep" check has not run yet. I think it needs one of the maintainers to approve it.

venkit-nexthop:
@dgsudharsan @prsunny gentle reminder to take a look

prsunny:
@vaibhavhd for viz

fpmsyncd/routesync.h
```diff
@@ -61,8 +186,50 @@ class RouteSync : public NetMsg
...
+        Srv6MySidTableFieldValueTupleWrapper & fvw,
+        ProducerStateTable & table);
+
+    void setTable(
+        Srv6SidListTableFieldValueTupleWrapper & fvw,
+        ProducerStateTable & table);
+
+    // Generic method for DEL operations with warm restart support
+    template<typename WrapperType>
+    void delWithWarmRestart(WrapperType && fvw, ProducerStateTable & table) {
```
dgsudharsan: Recommend to move implementation to routesync.cpp for better readability.

tests/mock_tests/fpmsyncd/test_routesync.cpp
```diff
@@ -596,11 +596,9 @@ TEST_F(FpmSyncdResponseTest, TestUpdateNextHopGroupDb)
         vector<FieldValueTuple> fieldValues;
         nexthop_group_table.get("5", fieldValues);
 
-        EXPECT_EQ(fieldValues.size(), 2);
+        EXPECT_EQ(fieldValues.size(), 1);
```
dgsudharsan, line 599: I believe this was changed because of the DEL and ADD operation. Can you please confirm if this won't affect any existing behavior?

fpmsyncd/routesync.h
```diff
@@ -42,6 +42,131 @@ struct NextHopGroup {
...
+    public:
+    FieldValueTupleWrapperBase(const string & _key) : key(_key) {}
+    FieldValueTupleWrapperBase(const string && _key) : key(std::move(_key)) {}
+    virtual ~FieldValueTupleWrapperBase() = default;
+
+    virtual vector<FieldValueTuple> fieldValueTupleVector() = 0;
+
+    vector<KeyOpFieldsValuesTuple> KeyOpFieldsValuesTupleVector() {
+        vector<KeyOpFieldsValuesTuple> kfvVector;
+        kfvVector.push_back(KeyOpFieldsValuesTuple {key.c_str(), "DEL", {}});
```
dgsudharsan, line 62: Please add comments in the code here explaining why DEL and SET is required in this flow.

fpmsyncd/routesync.cpp
```diff
@@ -163,31 +163,71 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
...
     }
     else
     {
-        const KeyOpFieldsValuesTuple kfv = std::make_tuple(key, cmd, fvVector);
-        m_warmStartHelper.insertRefreshMap(kfv);
+        m_warmStartHelper.insertRefreshMap(fvw.KeyOpFieldsValuesTupleVector()[1]);
     }
 }
 
+void RouteSync::setTable(VnetRouteTableFieldValueTupleWrapper & fvw,
```
dgsudharsan: Why do you need these many definitions of setTable. Can't we just have one API with the base class as the argument?

setTable(FieldValueTupleWrapper & fvw, ProducerStateTable & table )

fpmsyncd/routesync.cpp
```diff
@@ -163,31 +163,71 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
     rtnl_link_alloc_cache(m_nl_sock, AF_UNSPEC, &m_link_cache);
 }
 
-void RouteSync::setRouteWithWarmRestart(const std::string& key,
-                                      const std::vector<FieldValueTuple>& fvVector,
-                                      shared_ptr<ProducerStateTable> table,
-                                      const std::string& cmd)
+void RouteSync::setRouteWithWarmRestart(RouteTableFieldValueTupleWrapper & fvw,
+                                        ProducerStateTable & table )
```
dgsudharsan, line 167: Same as comment for setTable. Can't we just have one API with  FieldValueTupleWrapper & fvw, ProducerStateTable & table as arguments?

dgsudharsan (changes requested):

dgsudharsan:
@venkit-nexthop If we merge this PR first before the support in FRR PR, won't it break the exiting flow? Should these be merged together?

fpmsyncd/routesync.cpp
venkit-nexthop: Yes, we can have a single API. I left it like this after I refactored the code to pull out a base class because I thought new route-type support would be more explicit. But I think that is not very important. I will combine this into a single API.

fpmsyncd/routesync.cpp
venkit-nexthop: Like I mentioned in the earlier comment, I will combine this into a single API

fpmsyncd/routesync.h
venkit-nexthop: will do

fpmsyncd/routesync.h
venkit-nexthop: will do

tests/mock_tests/fpmsyncd/test_routesync.cpp
venkit-nexthop: Thats correct. This change was to handle the new code. It wont affect any existing behavior.

venkit-nexthop:
> @venkit-nexthop If we merge this PR first before the support in FRR PR, won't it break the exiting flow? Should these be merged together?

Thanks for the review @dgsudharsan 
No, this wont break the existing flow. This works without the corresponding FRR PR. We have been running tests with this code.

venkit-nexthop:
@dgsudharsan I have addressed all your comments. Please take a look when you get a chance. Once again appreciate your review.

venkit-nexthop:
@dgsudharsan could you help me with running "Semgrep". It says "Waiting for status to be reported".

venkit-nexthop:
@dgsudharsan @prsunny 
all the checks have passed now. But it says "Merging is blocked: You're not authorized to push to this branch."

PR #3885, opened 2025-09-16
Avoid notifying counter polling for duplicated priority groups or queues

stephenxs:
**What I did**

Avoid notifying counter polling for duplicated PG or queue when `create_only_config_db_buffers` is `False` in the flow:

1. Flex counter orchagent enables counter polling for PG or queue. It notifies counter polling for all PGs or queues since `create_only_config_db_buffers` is `False`
2. Particular items in BUFFER_PG or BUFFER_QUEUE is configured. It notifies counter polling for the items again.

In most cases, BUFFER_PG and BUFFER_QUEUE are handled ahead of counter polling (i.e., the order is 2, 1), and the issue won't occur. But in a rare scenario, it can be 1 and then 2.

**Why I did it**

**How I verified it**

Manual test, mock test

**Details if related**

### DIFF
--- a/orchagent/flexcounterorch.cpp
+++ b/orchagent/flexcounterorch.cpp
@@ -97,6 +97,24 @@ FlexCounterOrch::FlexCounterOrch(DBConnector *db, vector<string> &tableNames):
     m_deviceMetadataConfigTable(db, CFG_DEVICE_METADATA_TABLE_NAME)
 {
     SWSS_LOG_ENTER();
+
+    // Read create_only_config_db_buffers configuration once during initialization
+    std::string createOnlyConfigDbBuffersValue;
+    try
+    {
+        if (m_deviceMetadataConfigTable.hget("localhost", "create_only_config_db_buffers", createOnlyConfigDbBuffersValue))
+        {
+            if (createOnlyConfigDbBuffersValue == "true")
+            {
+                m_createOnlyConfigDbBuffers = true;
+            }
+        }
+    }
+    catch(const std::system_error& e)
+    {
+        SWSS_LOG_ERROR("System error reading create_only_config_db_buffers: %s", e.what());
+    }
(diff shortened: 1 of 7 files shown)


stephenxs:
The current failure is irrelevant to the PR itself. Same failures are seen in other PRs checker eg https://dev.azure.com/mssonic/build/_build/results?buildId=952049&view=logs&j=3f6395b2-1619-5ebe-f305-2aedcf353cb5&t=56fb64f8-b013-579d-11b0-5843431d20c9

orchagent/flexcounterorch.cpp
```diff
@@ -393,39 +418,59 @@ bool FlexCounterOrch::getWredPortCountersState() const
...
+        if (key == "localhost" && op == SET_COMMAND)
         {
-            if (createOnlyConfigDbBuffersValue == "true")
+            for (auto valuePair : data)
             {
-                return true;
+                const auto &field = fvField(valuePair);
+                const auto &value = fvValue(valuePair);
+
+                if (field == "create_only_config_db_buffers")
```
kperumalbfn, line 446: @stephenxs Do you have yang model changes for this new field("create_only_config_db_buffers") in DEVICE_METADATA?

kperumalbfn: If you already have the PR, please add that in description for reference.

stephenxs: @kperumalbfn now new. it's there for years :)
https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-device_metadata.yang#L251

kperumalbfn: ok, Thanks :)

kperumalbfn (approved):

PR #3888, opened 2025-09-17
[swss] Install yang related deb to align with latest libswsscommon requirement

wen587:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


yejianquan:
Prince is in OOF. Without this PR, swss new PR will all be blocked due to swsscommon change.

PR #3892, opened 2025-09-18
[P4Orch] Add new ACL match qualifiers vrf_id and ipmc_table_hit.

divyagayathri-hcl:
**What I did**
Added new ACL qualifiers (vrf_id, ipmc_table_hit) to enable more precise packet filtering based on VRF and multicast lookup status.
**Why I did it**

**How I verified it**
Verified this by creating and testing rules that used these new match criteria.
**Details if related**


rkavitha-hcl:
@StephenWangGoogle, can you pl. review

StephenWangGoogle (approved):

PR #3905, opened 2025-09-24
[Dash] Update ENI Forward Orch to handle new DPU Table schema

vivekrnv:
**What I did**

1) Update DashEniFwdOrch to handle the updated schema for DPU, VDPU, REMOTE_DPU & DASH_ENI_FORWARD_TABLE
2) Added REQ_T_STRING_LIST option to Request parser
3) Refactored macros
4) Updated UT's

**Why I did it**

Here is the latest schema:
https://github.com/sonic-net/SONiC/blob/91c3d94be67c96b2cba7fa8ce43bfff540638971/doc/smart-switch/high-availability/smart-switch-ha-detailed-design.md#2111-dpu--vdpu-definitions

https://github.com/sonic-net/SONiC/blob/91c3d94be67c96b2cba7fa8ce43bfff540638971/doc/smart-switch/high-availability/smart-switch-ha-detailed-design.md#2321-dash_eni_forward_table

**How I verified it**

Ran Tests

**Details if related**

### DIFF
--- a/orchagent/request_parser.h
+++ b/orchagent/request_parser.h
@@ -23,6 +23,7 @@ typedef enum _request_types_t
     REQ_T_IP_LIST,
     REQ_T_UINT_LIST,
     REQ_T_BOOL_LIST,
+    REQ_T_STRING_LIST,
 } request_types_t;
 
 typedef struct _request_description
@@ -175,6 +176,12 @@ public:
         return attr_item_bool_list_.at(attr_name);
     }
 
+    const std::vector<std::string>& getAttrStringList(const std::string& attr_name) const
+    {
+        assert(is_parsed_);
+        return attr_item_string_list_.at(attr_name);
+    }
+
 protected:
     Request(const request_description_t& request_description, const char key_separator)
         : request_description_(request_description),
@@ -200,6 +207,7 @@ private:
     std::vector<swss::MacAddress> parseMacAddressList(const std::string& str);
     std::vector<uint64_t> parseUintList(const std::string& str);
     std::vector<bool> parseBoolList(const std::string& str);
+    std::vector<std::string> parseStringList(const std::string& str);
 
     sai_packet_action_t parsePacketAction(const std::string& str);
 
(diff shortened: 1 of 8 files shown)


vivekkarri:
@prsunny, @zjswhhh, @prabhataravind Please help review and signoff

KrisNey-MSFT:
hi @vivekrnv , this is still in Draft.  Do reviews normally happen when a PR is in Draft state?

vivekkarri:
> hi @vivekrnv , this is still in Draft. Do reviews normally happen when a PR is in Draft state?

No, We recently had some design changes and i need to make further changes to this PR and so i moved it to draft

bingwang-ms:
@zjswhhh Can you help review?

prsunny:
@zjswhhh , could you review/signoff?

orchagent/request_parser.h
```diff
@@ -175,11 +176,18 @@ class Request
...
+        return attr_item_string_list_.at(attr_name);
+    }
+
 protected:
-    Request(const request_description_t& request_description, const char key_separator)
+    Request(const request_description_t& request_description, const char key_separator, bool relaxed_attr_parsing = false)
         : request_description_(request_description),
           key_separator_(key_separator),
           is_parsed_(false),
+          relaxed_attr_parsing_(relaxed_attr_parsing),
```
prsunny, line 190: Why is this attribute introduced? Can you explain this in description?

prsunny:
@vivekrnv , seeing a lot of changes from old implementation to this new one. Could you please give some details on why this is changed? what was the behavior before and why its changed?

orchagent/request_parser.h
vivekkarri:
```
This is needed because the Request parser expects a strict schema of field-value pairs. Anything extra and it would not parse the request. There might be cases where the orch might not need all the fv-pairs and so this option is introduced
```

Added

vivekkarri:
> @vivekrnv , seeing a lot of changes from old implementation to this new one. Could you please give some details on why this is changed? what was the behavior before and why its changed?

1. Firstly most changes are related to schema related refactoring. Schema widely changed when the first draft for first written. Refactoring ugly macros and Relaxed Attribute parsing and adding REQ_T_STRING_LIST is also part of it
2. FNIC design changes, Only 2 rules for ENI now. This is explained in HLD
3. Tunnel NH handling: Always pass in VNI while create tunnel NH, update ACL Orch to take in vni in the Redirect field. Remove ref counting which was previously there in DashEniFwdOrch which is redundant since VxLanOrch aleady has logic to handle this.

orchagent/request_parser.h
prsunny: ok, Vivek, I need an example to understand this. also, i suggest moving such infra change to a separate PR in future. Can you update description with changes on request parser with an example schema defintion

vivekkarri: Sure. Updated in the description

prsunny (approved):

PR #3907, opened 2025-09-25
ISS-3102 FIPS POST validation for MACsec control plane crypto.

vikram-nexthop:
**What I did**
- Query control plane (wpa_supplicant) crypto POST status check in macsecmgrd main, when FIPS mode is enabled
- Record status in FIPS_MACSEC_POST_TABLE.
- Implemented fail-secure behavior where macsecmgrd is forced to get into an infinite loop, if the POST validation fails.

**Why I did it**
MACsec control plane FIPS POST validation is required to ensure the cryptographic backend has passed self-tests before enabling MACsec operations. This is to ensure MACsec control plane fails securely if crypto backend POST validation fails in FIPS environments.

**How I verified it**
Enabled MACsec service with and without SymCrypt FIPS provider.

**Details if related**
related PR: https://github.com/sonic-net/sonic-wpa-supplicant/pull/99

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -10,9 +10,9 @@ parameters:
 - name: pool
   type: string
   values:
-  - sonicbld-1es
-  - sonicbld-armhf
-  - sonicbld-arm64
+  - sonicso1ES-amd64
+  - sonicso1ES-armhf
+  - sonicso1ES-arm64
   - default
   default: default
 
--- a/.azure-pipelines/gcov.yml
+++ b/.azure-pipelines/gcov.yml
@@ -8,7 +8,7 @@ parameters:
 - name: pool
   type: string
   values:
-  - sonicbld-1es
+  - sonicso1ES-amd64
   - default
   default: default
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -46,7 +46,7 @@ stages:
   - template: .azure-pipelines/build-template.yml
     parameters:
       arch: amd64
-      pool: sonicbld-1es
+      pool: sonicso1ES-amd64
       sonic_slave: sonic-slave-${{ parameters.debian_version }}
       common_lib_artifact_name: common-lib
       swss_common_artifact_name: sonic-swss-common-${{ parameters.debian_version }}
@@ -61,7 +61,7 @@ stages:
   - template: .azure-pipelines/build-template.yml
     parameters:
       arch: amd64
-      pool: sonicbld-1es
+      pool: sonicso1ES-amd64
       sonic_slave: sonic-slave-${{ parameters.debian_version }}
       common_lib_artifact_name: common-lib
       swss_common_artifact_name: sonic-swss-common-${{ parameters.debian_version }}
@@ -77,7 +77,7 @@ stages:
     parameters:
       arch: armhf
       timeout: 240
-      pool: sonicbld-armhf
+      pool: sonicso1ES-armhf
       sonic_slave: sonic-slave-${{ parameters.debian_version }}-armhf
       common_lib_artifact_name: common-lib.armhf
       swss_common_artifact_name: sonic-swss-common-${{ parameters.debian_version }}.armhf
@@ -89,7 +89,7 @@ stages:
     parameters:
       arch: arm64
       timeout: 240
-      pool: sonicbld-arm64
+      pool: sonicso1ES-arm64
       sonic_slave: sonic-slave-${{ parameters.debian_version }}-arm64
       common_lib_artifact_name: common-lib.arm64
       swss_common_artifact_name: sonic-swss-common-${{ parameters.debian_version }}.arm64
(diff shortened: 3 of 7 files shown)


cfgmgr/macsecmgrd.cpp
```diff
@@ -41,6 +42,89 @@ static void sig_handler(int signo)
...
 }
 
+/* Check if FIPS mode is enabled */
+bool fipsEnabled(DBConnector *stateDb)
+{
+    try
+    {
+        /* Check if FIPS was enabled via sonic-installer */
+        std::ifstream proc_cmdline_file("/proc/cmdline");
+        if (proc_cmdline_file.is_open())
```
wumiaont, line 53: There are 2 ways to enable FIPS. Through cfg or through cmd. /proc/cmdline is using cmd. /etc/fips/fips_enable is done through cfg. Should you also add check of /etc/fips/fips_enable for FIPS enabled or not?

vikram-nexthop: fixed.

vikram-nexthop:
@prsunny  Could you please review this PR for approval when convenient? All the review comments have been addressed.

vikram-nexthop:
> @prsunny Could you please review this PR for approval when convenient? All the review comments have been addressed.

gentle reminder to complete the review.

judyjoseph:
@vikram-nexthop we already stop the configs in macsecmgrd in PR : https://github.com/sonic-net/sonic-swss/pull/3836 ,  [macsecmgrd](https://github.com/sonic-net/sonic-swss/pull/3836/files#diff-4e9f2b8e0df5b106e8191a7e5f9f004f6c336dafbb6665b287d3aa83a459d6a7). Please share more details on this PR

vikram-nexthop:
@judyjoseph PR #[3836](https://github.com/sonic-net/sonic-swss/pull/3836) handles the data plane POST (SAI/hardware crypto engines), while this PR addresses control plane POST (wpa_supplicant crypto module for MKA protocol). Both are separate cryptographic modules and therefore require independent FIPS validation.

Complete MACsec FIPS compliance requires both POST validations to pass. So, both POST validations must pass before enabling MACsec configuration processing.

cfgmgr/macsecmgrd.cpp
```diff
@@ -82,6 +210,16 @@ int main(int argc, char **argv)
         SWSS_LOG_NOTICE("starting main loop");
         while (!received_sigterm)
         {
+            if (!isCpPostStateReady)
```
judyjoseph, line 213: Not seeing anyone updating this variable  isCpPostStateReady before checking ? This API getCpCryptoFipsPostStatus() is just called once I see ?

cfgmgr/macsecmgrd.cpp
judyjoseph, line 229: Can we do this check isCpPostStateReady() here. i.e check CpPost if dataplanePost is pass ? that way we need not check for fipsEnabled explicitly

We have removed the file open and check for fips here in orcahgent/main.cpp and replaced with an option "https://github.com/nexthop-ai/sonic-swss/blob/a3df4d5452fa30e52ac8205ae8b95d5bf3cf7b84/orchagent/main.cpp#L107"

cfgmgr/macsecmgrd.cpp
vikram-nexthop: The getCpCryptoFipsPostStatus() fetches the control plane POST status synchronously, and therefore invoked only once.

cfgmgr/macsecmgrd.cpp
vikram-nexthop: ok, will look into this.

wumiaont: > Can we do this check isCpPostStateReady() here. i.e check CpPost if dataplanePost is pass ? that way we need not check for fipsEnabled explicitly
> 
> We have removed the file open and check for fips here in orcahgent/main.cpp and replaced with an option "https://github.com/nexthop-ai/sonic-swss/blob/a3df4d5452fa30e52ac8205ae8b95d5bf3cf7b84/orchagent/main.cpp#L107"

I have the same feeling that we may be able to do those FIPS ready check in macsecmgrd. In this case there's no need to make code change into wpa_supplicant which is a third party open source.

vikram-nexthop: I think @judyjoseph's comment is to invoke isCpPostStateReady() after SAI/switch post ststus becomes available (within the if statement, line no 227).

Regarding you suggestion to move FIPS ready check to macsecmgrd, I have some concerns.  The crypto libraries (OpenSSL,etc) are linked to the wpa_supplicant. Macsecmgrd doesn't use any crypto libraries as such. 
The migration would require the following changes, as per my understanding,
1. change macsecmgrd to link/include all the crypto libraries.
2. since the wpa_supplicant supports multiple backend crypto libraries, keep track of the list of supported libraries and ensure that macsecmgrd validates the same library that is used by the wpa_supplicant.

The decision to make changes in the wpa_supplicant (which is directly connected to the crypto module) was based on these factors.

vikram-nexthop:
closing this PR, since control plane POST check is not required. Refer to https://github.com/sonic-net/SONiC/pull/2083 for more details.

PR #3910, opened 2025-09-30
[ZmqOrch] Optimize memory by popping batch size at a time

vivekrnv:
**What I did**

1. Used a reference instead of a unnecessary copy of a set object
2. Optimize memory by popping batch size at a time

**NOTE: Please merge only after the below two PR's are merged**

https://github.com/sonic-net/sonic-swss-common/pull/1084
https://github.com/sonic-net/sonic-sairedis/pull/1660

**Why I did it**

To reduce peak memory usage when applying high-volume dash configuration

**How I verified it**

**Details if related**


oleksandrivantsiv (approved):

prsunny (approved):

prsunny:
@liuh-80 , please review/signoff

prabhataravind (approved):

liuh-80 (approved):

venkit-nexthop:
Tagging @deepak-singhal0408 for viz
We found that this change cuts the route-download performance almost by half. 
I wanted to understand the change a bit more. By removing the do..while loop in zmqorch.cpp how does it save memory? The input queue would still fill up isnt it?

deepak-singhal0408:
> Tagging @deepak-singhal0408 for viz We found that this change cuts the route-download performance almost by half. I wanted to understand the change a bit more. By removing the do..while loop in zmqorch.cpp how does it save memory? The input queue would still fill up isnt it?

@venkit-nexthop since this PR is merged already, can you help create a new issue for the regression you are observing, while tagging this PR?

PR #3913, opened 2025-10-03
VxlanOrch changes for EVPN MH feature

jayakj:
**What I did**
 Implementation of VxlanOrch changes from EVPN VXLAN Multihoming HLD.
 https://github.com/sonic-net/SONiC/blob/master/doc/vxlan/EVPN/EVPN_VxLAN_Multihoming.md#336-vxlanorch
 
**Why I did it**

**How I verified it**
By SWSS UT

**Details if related**

### DIFF
--- a/orchagent/vxlanorch.cpp
+++ b/orchagent/vxlanorch.cpp
@@ -1063,10 +1063,8 @@ void VxlanTunnel::updateRemoteEndPointIpRef(const std::string remote_vtep, bool 
 
         SWSS_LOG_DEBUG("Decrementing remote end point %s reference to %d", remote_vtep.c_str(),
                        it->second.ip_refcnt);
-        if (it->second.ip_refcnt == 0)
-        {
-             tnl_users_.erase(remote_vtep);
-        }
+        // tnl_users_.erase should be done by the caller after checking
+        // related cleanup dependencies
     }
 }
 
--- a/orchagent/vxlanorch.cpp
+++ b/orchagent/vxlanorch.cpp
@@ -1166,6 +1166,29 @@ bool VxlanTunnel::deleteDynamicDIPTunnel(const std::string dip, tunnel_user_t us
     return true;
 }
 
+void VxlanTunnel::cleanupDynamicDIPTunnel(const std::string remote_vtep)
+{
+    Port tunnelPort;
+    string port_tunnel_name;
+    VxlanTunnelOrch* vxlan_orch = gDirectory.get<VxlanTunnelOrch*>();
+    int ref_cnt = this->getRemoteEndPointRefCnt(remote_vtep);
+
+    SWSS_LOG_DEBUG("DIP:%s diprefcnt = %d",
+                  remote_vtep.c_str(), ref_cnt);
+    if (ref_cnt == 0) {
+        port_tunnel_name =
+            vxlan_orch->getTunnelPortName(remote_vtep);
+        if (gPortsOrch->getPort(port_tunnel_name,tunnelPort) &&
+            tunnelPort.m_fdb_count == 0 &&
+            tunnelPort.m_type == Port::TUNNEL) {
+            vxlan_orch->deleteTunnelPort(tunnelPort);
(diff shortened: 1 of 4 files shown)


orchagent/vxlanorch.cpp
```diff
@@ -1166,6 +1166,29 @@ bool VxlanTunnel::deleteDynamicDIPTunnel(const std::string dip, tunnel_user_t us
     return true;
 }
 
+void VxlanTunnel::cleanupDynamicDIPTunnel(const std::string remote_vtep)
```
selva-nexthop, line 1167: where it is being used?

orchagent/vxlanorch.cpp
```diff
@@ -1063,10 +1063,8 @@ void VxlanTunnel::updateRemoteEndPointIpRef(const std::string remote_vtep, bool
 
         SWSS_LOG_DEBUG("Decrementing remote end point %s reference to %d", remote_vtep.c_str(),
                        it->second.ip_refcnt);
-        if (it->second.ip_refcnt == 0)
-        {
-             tnl_users_.erase(remote_vtep);
-        }
+        // tnl_users_.erase should be done by the caller after checking
```
selva-nexthop, line 1066: The comment says "caller should do it" but the callers are:
- delTunnelUser() — calls deleteDynamicDIPTunnel() which handles its own refcnt checks, but never erases tnl_users_
- addTunnelUser() — only increments, never erases

orchagent/vxlanorch.cpp
```diff
@@ -1657,9 +1681,9 @@ bool  VxlanTunnelOrch::delTunnelUser(const std::string remote_vtep, uint32_t vni
     }
 
     port_tunnel_name = getTunnelPortName(remote_vtep);
-    gPortsOrch->getPort(port_tunnel_name,tunnelPort);
-    if ((vtep_ptr->getRemoteEndPointRefCnt(remote_vtep) == 1) &&
-       tunnelPort.m_fdb_count == 0)
+    if (gPortsOrch->getPort(port_tunnel_name,tunnelPort) &&
```
selva-nexthop, line 1684: it is correct that the getPort return value is now checked, preventing use of an uninitialized.

But it seems to be incomplete. If getPort returns false (port doesn't exist), the code silently proceeds to deleteDynamicDIPTunnel() without removing the bridge port. But the tunnel SAI object may still reference a bridge port that's partially cleaned up. There's no error log or alternate cleanup path for this case.

orchagent/vxlanorch.cpp
selva-nexthop, line 1: Per the HLD, VxlanOrch changes should "maintain reference count of destination IP per L2 NHG." The L2nhgOrch (introduced in a separate PR) interacts with VxlanOrch to update refcnt per remote VTEP when NHG members are added/removed.

This PR doesn't add any L2NHG-aware reference counting. The cleanupDynamicDIPTunnel method doesn't consider NHG references. If a VTEP is referenced by both an IMR and an L2 NHG, the cleanup logic may prematurely delete the tunnel.

orchagent/vxlanorch.cpp
```diff
@@ -1166,6 +1164,29 @@ bool VxlanTunnel::deleteDynamicDIPTunnel(const std::string dip, tunnel_user_t us
     return true;
 }
 
+void VxlanTunnel::cleanupDynamicDIPTunnel(const std::string remote_vtep)
```
selva-nexthop, line 1167: Should be const std::string& (pass by reference).

orchagent/vxlanorch.cpp
selva-nexthop, line 1: No Warm-Boot Consideration
Tunnel user reference counts maintained in tnl_users_ are in-memory only. After a warm restart:
- tnl_users_ is empty
- Remote VTEPs with active FDB entries exist in SAI
- No reconciliation logic to rebuild tnl_users_ from SAI state

The PR should at least document this limitation.

jayakj:

PR #3914, opened 2025-10-03
FdbOrch changes for EVPN MH feature

jayakj:
**What I did**
Implementation of FdbOrch changes from EVPN VXLAN Multihoming HLD.
https://github.com/sonic-net/SONiC/blob/master/doc/vxlan/EVPN/EVPN_VxLAN_Multihoming.md#337-fdborch

**Why I did it**

**How I verified it**
By SWSS UT

**Details if related**


orchagent/neighorch.cpp
```diff
@@ -168,6 +279,12 @@ void NeighOrch::update(SubjectType type, void *cntx)
             updateNextHop (*update);
             break;
         }
+        case SUBJECT_TYPE_BFD_SESSION_DELETION:
```
prsunny: why is BFD notification added to this PR?

jayakj: This came as part of another fix. reverted now as this is not related to MH feature.

tests/mock_tests/fdborch/fdborch_vxlan_ut.cpp
```diff
@@ -0,0 +1,1233 @@
...
+
+            ASSERT_EQ(gIntfsOrch, nullptr);
+
+#if 0
+            vector<table_name_with_pri_t> intf_tables = {
+                { APP_INTF_TABLE_NAME,  IntfsOrch::intfsorch_pri},
+                { APP_SAG_TABLE_NAME,   IntfsOrch::intfsorch_pri}
+            };
+#endif
+            gIntfsOrch = new IntfsOrch(m_app_db.get(), APP_INTF_TABLE_NAME, gVrfOrch, m_chassis_app_db.get());
```
apannerselva, line 149: the second parameter is vector,  APP_INTF_TABLE_NAME macro may not work.
use as such - 
std::vector<table_name_with_pri_t> intfTables = {
    { APP_INTF_TABLE_NAME, 0 }
};

orchagent/fdborch.cpp
```diff
@@ -393,13 +449,59 @@ void FdbOrch::update(sai_fdb_event_t        type,
...
+                sai_attribute_t attr;
+                sai_ip_address_t ipaddr;
+
+                /* Since sai doesn't support remove one fdb attr, just set the ENDPOINT IP to 0 after move local */
+                attr.id = SAI_FDB_ENTRY_ATTR_ENDPOINT_IP;
+                ipaddr.addr_family = SAI_IP_ADDR_FAMILY_IPV6;
+                memcpy(ipaddr.addr.ip6, swss::IpAddress("0:0:0::0").getV6Addr(), sizeof(ipaddr.addr.ip6));
+                attr.value.ipaddr = ipaddr;
+                status = sai_fdb_api->set_fdb_entry_attribute(&fdb_entry, &attr);
+                if (status != SAI_STATUS_SUCCESS)
```
selva-nexthop, line 479: If the set_fdb_entry_attribute fails, the code continues to treat the MAC as moved locally. The old remote FDB entry's ENDPOINT_IP still points to the old VTEP, but the entry is overwritten in m_entries. This creates a stale SAI state that never gets cleaned up.

On failure, either retry or skip the move and leave the entry unchanged

orchagent/fdborch.cpp
```diff
@@ -1136,10 +1385,165 @@ void FdbOrch::flushFDBEntries(sai_object_id_t bridge_port_oid,
...
+            SWSS_LOG_WARN("FdbOrch flushAllFDB: Failed to locate port from bridge_port_id 0x%" PRIx64, bridge_port_oid);
+            return;
+        }
+
+        /* Using a workaround to flush all mac under the tunnel port */
+        if (port.m_type == Port::TUNNEL)
+        {
+            SWSS_LOG_NOTICE("Try to flushAllFDB for tunnel port %s bridge_port_id 0x%" PRIx64, port.m_alias.c_str(), bridge_port_oid);
+            /* Try to remove all remote FDB under this tunnel port one by one */
+            for (auto itr = m_entries.begin(); itr != m_entries.end();)
```
selva-nexthop, line 1459: For large-scale EVPN deployments with thousands of remote MACs per tunnel, this is a serial O(n) SAI call loop. Each SAI call has latency. The existing flushFDBEntries uses the efficient SAI bulk flush API (sai_fdb_api->flush_fdb_entries).

Extended convergence time during tunnel teardown or link failure events.

Any SAI bulk flush support for tunnel bridge ports, or at minimum use batched SAI operations?  It is worth to investigate.

orchagent/fdborch.h
```diff
@@ -98,25 +114,35 @@ class FdbOrch: public Orch, public Subject, public Observer
...
+    void flushAllFDBEntries(sai_object_id_t bridge_port_oid,
+                            sai_object_id_t vlan_oid);
     void flushFdbByVlan(const string &);
     void notifyObserversFDBFlush(Port &p, sai_object_id_t&);
 
 private:
     PortsOrch *m_portsOrch;
     map<FdbEntry, FdbData> m_entries;
-    fdb_entries_by_port_t saved_fdb_entries;
+    fdb_entries_by_port_t m_entries_by_port;
```
selva-nexthop, line 131: A new fdb_entries_by_port_t m_entries_by_port cache is added for per-port FDB tracking, but it's only populated in addFdbEntry() and cleaned up in removeFdbEntry() and flushAllFDBEntries().

- storeFdbEntryState() — handles LEARN events; modifies m_entries but never touches m_entries_by_port
- clearFdbEntry() — erases from m_entries via storeFdbEntryState but never cleans m_entries_by_port
- handleSyncdFlushNotif() — calls clearFdbEntry which doesn't clean the port cache

The m_entries_by_port will contain stale entries over time, leading to incorrect behavior in any code that relies on it.

orchagent/fdborch.cpp
```diff
@@ -1275,6 +1680,42 @@ void FdbOrch::updateVlanMember(const VlanMemberUpdate& update)
...
+ * is deleted or moved from one port to another to keep the cache consistent.
+ *
+ * @param entry The FDB entry to remove from the cache
+ * @param port The port from which to remove the FDB entry
+ */
+void FdbOrch::removeFdbEntryFromPortCache(const FdbEntry& entry, const Port& port)
+{
+    auto& entries_for_port = m_entries_by_port[port.m_alias];
+
+    for (auto it = entries_for_port.begin(); it != entries_for_port.end(); it++) {
```
selva-nexthop, line 1698: Linear search through a vector<FdbEntry> per port. With thousands of MACs per port (common in EVPN), this becomes expensive. Called frequently during MAC moves and deletes.

Please use unordered_set<FdbEntry> with a proper hash function, or maintain an index map.

orchagent/neighorch.cpp
```diff
@@ -107,6 +105,117 @@ void NeighOrch::clearResolvedNeighborEntry(const NeighborEntry &entry)
...
+                            from bv_id 0x%" PRIx64 ".", entry.bv_id);
+        return;
+    }
+    SWSS_LOG_NOTICE("processFDBResolve: Resolving ARP for mac: %s, port: %s, VLAN: %s",
+                    entry.mac.to_string().c_str(), entry.port_name.c_str(), vlan.m_alias.c_str());
+
+    // If the FDB entry MAC matches with neighbor/ARP entry MAC,
+    // and ARP entry incoming interface matches with VLAN name,
+    // flush neighbor/arp entry.
+    for (const auto &neighborEntry : m_syncdNeighbors)
```
selva-nexthop, line 208: The m_syncdNeighbors is keyed by NeighborEntry (IP+alias), not by MAC. Every FDB event triggers a full scan. In large-scale environments with 10K+ neighbors, this is a significant CPU burden.

Please add a reverse index map<MacAddress, vector<NeighborEntry>> for O(1) MAC-based lookups.

orchagent/neighorch.cpp
selva-nexthop: The notify(SUBJECT_TYPE_NEIGH_CHANGE) call from after m_syncdNeighbors.erase() is moved before the SAI remove call, protected by a new deletion_pending flag.

we will have issue like;
- Observers are notified the neighbor is gone before it's actually removed from SAI
- If SAI removal fails and deletion_pending is reset, observers have already acted on stale information (e.g., route withdrawals)
- The getNeighborEntry() check !entry.second.deletion_pending means the neighbor "disappears" from lookups before SAI confirms removal

It introduces the risk of route black-holes during transient SAI failures.

orchagent/fdborch.cpp
```diff
@@ -408,19 +510,45 @@ void FdbOrch::update(sai_fdb_event_t        type,
...
 
         break;
     }
     case SAI_FDB_EVENT_AGED:
     {
-        SWSS_LOG_INFO("Received AGE event for bvid=0x%" PRIx64 " mac=%s port=0x%" PRIx64,
+        SWSS_LOG_INFO("update: EVPN_MH_UC: Received AGE out event for bvid=0x%" PRIx64 " mac=%s port=0x%" PRIx64,
                        entry->bv_id, update.entry.mac.to_string().c_str(), bridge_port_id);
 
+        auto status = sai_fdb_api->remove_fdb_entry(entry);
```
selva-nexthop, line 545: The SAI remove is issued before checking if the entry exists in m_entries. For VXLAN-advertised MACs, the code then tries to re-add them (the "add-back" logic). But the FDB was already removed from SAI, creating a window where traffic is black-holed until the re-create succeeds.

orchagent/fdborch.cpp
selva-nexthop, line 556: Entry not in cache but already removed from SAI -- from above change at 545

jayakj:

PR #3915, opened 2025-10-06
PFCWD ACL recovery: Override getHwCounters to query acl counters

pinky-nexthop:
**What I did**
The current counters query queue stats which don't work for all the chips like Q3D, TH5. Since, acls are used to take action, acl counters should be queried to show pfcwd stats.

**Why I did it**
The current implementation doesn't work for all the broadcom chips. 

**How I verified it**
Ran the test test_pfcwd_cli.py successfully on TH5, Q3d, Trident3. 

**Details if related**

### DIFF
--- a/orchagent/pfcactionhandler.cpp
+++ b/orchagent/pfcactionhandler.cpp
@@ -30,6 +30,7 @@ extern AclOrch * gAclOrch;
 extern sai_port_api_t *sai_port_api;
 extern sai_queue_api_t *sai_queue_api;
 extern sai_buffer_api_t *sai_buffer_api;
+extern sai_acl_api_t *sai_acl_api;
 
 PfcWdActionHandler::PfcWdActionHandler(sai_object_id_t port, sai_object_id_t queue,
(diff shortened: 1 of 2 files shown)


pinky-nexthop:
@vmittal-msft 
Kindly review

rlhui:
please confirm if the issue is a Broadcom SAI regression, thanks.

pinky-nexthop:
> please confirm if the issue is a Broadcom SAI regression, thanks.

@rlhui No, this is not a regression. 
The thought behind this PR was this: 
On TH5, with s/w recovery enabled, the queue counters do not display the drop counters. Therefore, we need to query the acl counters. 
I noticed, that acl counters should work for all platforms because the drops will happen only when the acl hits. 
However, to stop this change from causing any regression, I will change the diff to only take effect for TH5. Thanks. 
Kindly give me sometime to update the PR.

pinky-nexthop:
Broadcom provided a sai patch to fix the issue in TH5. Closing this PR.

PR #3923, opened 2025-10-08
[routeorch] Remove updateRoute call to prevent set operations

Ndancejic:
This call to updateRoute is not needed because the route entry has already been removed. It is causing unneccesary route set operations and printing harmless (but verbose) error logs. Removing this call to prevent the side-effects.

**What I did**
Removed call to updateRoute from removeRoutePost path.

**Why I did it**
This call is not needed, and is causing error logs.

**How I verified it**
Small change, should be covered by current vstests and mgmt tests. Behavior will not change, just cause less ERR logs.


prsunny:
@zjswhhh , can you rebase to this commit?

PR #3926, opened 2025-10-09
[gearsyncd,macsec]: Deterministic MACsec backend selection for gearbox ports

rajshekhar-nexthop:
**What I did**
Introduce a platform capability flag in the gearbox config to determine, per PHY, whether MACsec is supported (applies to all ports mapped to that PHY). MACsec orchestration will:

- Use PHY switch by default on gearbox ports
- Use NPU/global switch only when the platform marks the PHY as not supporting MACsec

Have added three DVS testcases:

test_macsec_phy_switch_default: This tests the scenario when the macsec_supported field is absent in the gearbox_config.json
test_macsec_phy_switch_explicit: This tests the scenario when the macsec_supported field is set as true in the gearbox_config.json
test_macsec_npu_switch: This tests the scenario when the macsec_supported field is set as false in the gearbox_config.json

**Why I did it**
On gearbox ports, creating MACsec on the PHY switch fails (SAI_STATUS_NOT_IMPLEMENTED) if gearbox PHY does not have the MACsec engine.

**How I verified it**
Manually verified on DUT by adding macsec_supported=false in gearbox_config.json and configuring the macsec on the PHY port. Also ran the dvs testcase and made sure it is passing `sudo pytest -v tests/test_macsec_gearbox.py`

**Details if related**
HLD: https://github.com/sonic-net/SONiC/pull/2072
gearbox_config.json changes are posted here: https://github.com/sonic-net/sonic-buildimage/pull/24169/files#diff-737ea59a7eba8ea0ed71a15a052868815f7faad351fd353736ad196932bed57a

Co-authored by @shreyansh-nexthop


tests/test_macsec_gearbox.py
```diff
@@ -0,0 +1,133 @@
...
+
+            assert TestMacsecHelper.verify_macsec_in_asic_db(dvs, should_exist=False), (
+                "FAILED: MACsec objects should NOT exist in ASIC_DB "
+                "when using PHY backend"
+            )
+
+        finally:
+            TestMacsecHelper.cleanup_macsec(dvs, port_name)
+
+    def test_macsec_phy_switch_explicit(self, dvs, gearbox, gearbox_config):
```
judyjoseph, line 68: @rajshekhar-nexthop could you also add a negative case where the platform owner enables this flag "macsec_supported" only for a few GB phys in gearbox_config.json file ?

judyjoseph: @rajshekhar-nexthop I see comment from shreyansh. Yes it is ok to go with this approach --> Create a PR in sonic-buildimage to add a second PHY to brcm_gearbox_vs/gearbox_config.json. Once merged, I can add this testcase.

judyjoseph (approved):
LGTM

tests/test_macsec_gearbox.py
rajshekhar-nexthop: Hi @judyjoseph ,
Sure. Thanks for reviewing the code. Shreyansh is working on adding negative test case as per your suggestion. He will raise a separate PR soon.

rajshekhar-nexthop:
> Change LGTM, tested this against an Arista SKU which also has gearboxes that don't support macsec and the SAI call went to ASIC_DB as expected [sonic-net/sonic-buildimage#24761](https://github.com/sonic-net/sonic-buildimage/pull/24761)

Thank you, @arista-nwolfe, for testing and validating the changes on the Arista SKU.

rajshekhar-nexthop:
Hi @prsunny ,
Please enable workflows awaiting approval. Thanks

rajshekhar-nexthop:
Hi @prsunny ,
Please check and merge this PR. Thank you

tests/test_macsec_gearbox.py
shreyansh-nexthop: Hi @judyjoseph, I have raised the PR to add the negative case: https://github.com/sonic-net/sonic-swss/pull/4182. Can you please review?

PR #3934, opened 2025-10-11
Support checking capabilities of the mirror

stephenxs:
**What I did**

Support checking the capabilities of ingress/egress mirror before setting it to SAI.

1. Switch orchagent fetches the mirror capabilities from SAI during initialization and exposes to field `PORT_INGRESS_MIRROR_CAPABLE` and `PORT_EGRESS_MIRROR_CAPABLE` in `STATE_DB` table `SWITCH_CAPABILITY`.
2. Mirror orchagent check whether the direction is supported before calling the corresponding SAI API.

**Why I did it**

 This is to avoid SAI returning error on the platforms that the mirror is not supported on a direction.
It will collect SAI SDK dump on receiving a SAI error message, which is unnecessary.

**How I verified it**

Mock test + manual test

**Details if related**

### DIFF
--- a/orchagent/mirrororch.h
+++ b/orchagent/mirrororch.h
@@ -7,7 +7,6 @@
 #include "neighorch.h"
 #include "routeorch.h"
 #include "fdborch.h"
-//#include "switch_capabilities.h"
 #include "policerorch.h"
 
 #include "ipaddress.h"


stephenxs:
the failure is not relevant to the change. retrying.

stephenxs:
The vs test failure should be fixed by https://github.com/sonic-net/sonic-swss/pull/3931

keboliu (approved):

orchagent/switchorch.cpp
```diff
@@ -1859,6 +1860,63 @@ void SwitchOrch::querySwitchPortEgressSampleCapability()
...
+    sai_attr_capability_t capability;
+
+    // Check if SAI is capable of handling Port ingress mirror session
+    status = sai_query_attribute_capability(gSwitchId, SAI_OBJECT_TYPE_PORT,
+                            SAI_PORT_ATTR_INGRESS_MIRROR_SESSION, &capability);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_WARN("Could not query port ingress mirror capability %d", status);
+        fvVector.emplace_back(SWITCH_CAPABILITY_TABLE_PORT_INGRESS_MIRROR_CAPABLE, "false");
+        m_portIngressMirrorSupported = false;
```
bingwang-ms: This is not aligned with HLD. 
```
**Default Behavior**: When capability detection fails, system assumes full support for backward compatibility
```

orchagent/switchorch.cpp
```diff
@@ -1859,6 +1860,63 @@ void SwitchOrch::querySwitchPortEgressSampleCapability()
...
+    }
+
+    // Check if SAI is capable of handling Port egress mirror session
+    status = sai_query_attribute_capability(gSwitchId, SAI_OBJECT_TYPE_PORT,
+                            SAI_PORT_ATTR_EGRESS_MIRROR_SESSION, &capability);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_WARN("Could not query port egress mirror capability %d", status);
+        fvVector.emplace_back(SWITCH_CAPABILITY_TABLE_PORT_EGRESS_MIRROR_CAPABLE, "false");
+        m_portEgressMirrorSupported = false;
```
bingwang-ms: Same as above

orchagent/mirrororch.cpp
```diff
@@ -812,6 +813,18 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
                                     bool set,
                                     sai_object_id_t sessionId)
 {
+    // Check if the mirror direction is supported by the ASIC
+    if (ingress && !m_switchOrch->isPortIngressMirrorSupported())
```
bingwang-ms, line 817: If `set` is `false`, do we need this check?

stephenxs: Let's do it to keep the logic simple.
Set == false means to set a NULL OID on the attribute.
We won't reach such flow because we didn't set it if the capability is not supported. However, we can still check the capability to keep the logic simple

orchagent/switchorch.cpp
stephenxs: fixed.

orchagent/switchorch.cpp
stephenxs: fixed

bingwang-ms (dismissed):
LGTM now. @prsunny Would you mind helping review this PR?

orchagent/mirrororch.h
```diff
@@ -7,6 +7,7 @@
 #include "neighorch.h"
 #include "routeorch.h"
 #include "fdborch.h"
+//#include "switch_capabilities.h"
```
liat-grozovik: why commenting and not removing if this is not needed?

orchagent/mirrororch.cpp
```diff
@@ -812,6 +813,18 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
...
+    {
+        SWSS_LOG_ERROR("Port ingress mirror is not supported by the ASIC");
+        return false;
+    }
+    if (!ingress && !m_switchOrch->isPortEgressMirrorSupported())
```
liat-grozovik, line 822: anu reason why not using dedicated param for egress and instead using !ingress?

stephenxs: Either way works but I just want to follow the original style. It uses a Boolean to represent set/unset. So I add another Boolean to represent ingress/egress

orchagent/mirrororch.h
stephenxs: fixed.

stephenxs:
BuildAsan failed due to `##[error]The agent did not connect within the alloted time of 45 minute(s).`
Retrying.

prsunny (approved):
Overall lgtm, @bingwang-ms , could you signoff?

stephenxs:
@bingwang-ms  could you please sign-off and merge it? thanks
@prsunny  seems like the azure pipeline keeps failing.. Neither `Build amd64` nor `BuildAsan amd64` works due to the error `##[error]The agent did not connect within the alloted time of 45 minute(s).`
who can help to resolve this?
thanks

vmittal-msft:
@stephenxs can you please rebase ? we need this for updating sonic-utilities submodule in master.

stephenxs:
> @stephenxs can you please rebase ? we need this for updating sonic-utilities submodule in master.

@vmittal-msft @prsunny @bingwang-ms all test passed. could you please merge it? thanks

PR #3935, opened 2025-10-13
[Vnetorch] Ecmp vnet local routes

Leyza:
**What I did**
Implement Vnet local routes for single and ecmp case using RouteOrch's methods.

**Why I did it**
To enable ecmp vnet routes and the ability to update between single to ecmp vnet routes which is required for some use cases.

**How I verified it**
Tested on virtual and physical switches, and unit tests.

**Details if related**


orchagent/vnetorch.cpp
```diff
@@ -1595,57 +1601,105 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
...
+            RouteBulkContext ctx(key, (op == SET_COMMAND));
+            ctx.vrf_id = vr_id;
+            ctx.ip_prefix = ipPrefix;
+            ctx.nhg = nhg;
+
+            if (op == SET_COMMAND)
+            {
+                gRouteOrch->addRoute(ctx, nhg);
+                gRouteOrch->flushRouteBulker();
+                bulkNhgReducedRefCnt.clear();
```
anish-n: We will be touching global map of routeorch, but given things are single threaded in orchagent and we process the vnet route and then route orch route in serial this implementation should not cause issues.

tests/test_vnet.py
```diff
@@ -2969,6 +2969,121 @@ def test_vnet_orch_29(self, dvs, testlog):
...
+
+        # setup vnet neighbor
+        self.add_neighbor("Ethernet20", "10.10.0.9", "00:01:02:03:04:05")
+        self.add_neighbor("Ethernet16", "10.10.0.11", "00:01:02:03:04:06")
+
+        vnet_obj.fetch_exist_entries(dvs)
+
+        # create vnet local route
+        create_vnet_local_routes(dvs, "10.10.0.0/24", vnet_name, 'Ethernet20,Ethernet16', "10.10.0.9,10.10.0.11")
+        vnet_obj.check_vnet_local_routes(dvs, vnet_name)
```
anish-n: we should be checking for nexthop group as well in the ut

tests/test_vnet.py
```diff
@@ -2969,6 +2969,121 @@ def test_vnet_orch_29(self, dvs, testlog):
...
+        vnet_obj.check_del_router_interface(dvs, "Ethernet20")
+
+        delete_vnet_entry(dvs, vnet_name)
+        vnet_obj.check_del_vnet_entry(dvs, vnet_name)
+
+        delete_vxlan_tunnel(dvs, tunnel_name)
+        vnet_obj.check_del_vxlan_tunnel(dvs)
+
+    '''
+    Test 31 - Test ecmp vnet local route with multiple nexthops
```
anish-n, line 3036: would be nice to have a test case which checks for switching between single nh and ecmp group nh

tests/test_vnet.py
Leyza: Added check for next hops and next hop groups

tests/test_vnet.py
Leyza: Updated UT to switch between single and ecmp.

orchagent/vnetorch.cpp
```diff
@@ -1595,57 +1601,127 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
...
+                SWSS_LOG_INFO("Route del failed for %s", ipPrefix.to_string().c_str());
+                break;
+            }
         }
-        else if (op == DEL_COMMAND && !del_route(vr_id, pfx))
+        else
         {
-            SWSS_LOG_INFO("Route del failed for %s", ipPrefix.to_string().c_str());
-            break;
+            auto& bulkNhgReducedRefCnt = gRouteOrch->getBulkNhgReducedRefCnt();
```
prsunny: Would you please move this section to a separate function and give the comments on the handling? Otherwise, the changes looks good

Leyza: Moved new implementation to a separate function.

prsunny (approved):

PR #3944, opened 2025-10-15
Add check for empty m_priority_group_ids in initializePriorityGroupsBulk

arista-hpandya:
**What I did**

Fixes https://github.com/sonic-net/sonic-buildimage/issues/24268

Added a guard clause in `PortsOrch::initializePriorityGroupsBulk` to check if the port has any configured Priority Groups (`port.m_priority_group_ids.size() == 0`). If the size is zero, the function now skips the SAI calls to retrieve the Priority Group list.

**Why I did it**

This fixes a critical regression introduced by a previous PR discussed in detail here: https://github.com/sonic-net/sonic-buildimage/issues/24268 that replaced the individual `initializePriorityGroups` call with a bulk operation.

The original function included a check to ensure `SAI_PORT_ATTR_NUMBER_OF_INGRESS_PRIORITY_GROUPS` was non-zero before attempting to retrieve the list. This check was missing in the new `initializePriorityGroupsBulk` method.

**Impact of Regression:**
When processing ports without Priority Groups (e.g., recirculation ports like `Ethernet-Rec0`), the missing check led to a failed SAI operation (`SAI_STATUS_NOT_EXECUTED`) during initialization, causing a `std::runtime_error` and an immediate OrchAgent crash, which subsequently brings down the entire SWSS and Syncd stack.

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -6466,9 +6466,8 @@ void PortsOrch::initializePriorityGroupsBulk(std::vector<Port>& ports)
 
         // Track separate bulker index for verification
         size_t bulker_idx = 0;
-        for (size_t idx = 0; idx < portCount; idx++)
+        for (auto& port: ports)
         {
-            const auto& port = ports[idx];
             if (port.m_priority_group_ids.size() == 0)
             {
                 continue;


bingwang-ms:
@arista-hpandya Can you fix the code coverage issue?

arlakshm:
@arista-hpandya, can you please add UT to handle this case

abdosi:
@arista-hpandya  Please address UT

arista-hpandya:
Sure will add a UT soon

orchagent/portsorch.cpp
```diff
@@ -6464,10 +6464,16 @@ void PortsOrch::initializePriorityGroupsBulk(std::vector<Port>& ports)
 
         bulker.executeGet();
 
+        // Track separate bulker index for verification
+        size_t bulker_idx = 0;
         for (size_t idx = 0; idx < portCount; idx++)
```
arista-nwolfe: Let's iterate over ports instead of portCount.
I.E. `for (auto& port: ports)`
The `idx` was previously used to index into the `bulker.statuses` but it's no longer needed for that so iterating over the ports collection is better for consistency with the previous for loop.

vmittal-msft:
@arista-hpandya please rebase

Pavan-Nokia:
@arista-hpandya #3966 seems to be an enhancement to this PR that addresses additional use cases.
@abdosi @arlakshm @vmittal-msft 
@stepanblyschak

stepanblyschak:
@arista-hpandya @Pavan-Nokia Thanks for raising this PR and letting me know.
Right, In addition to PGs, I think the same issue is with queue and sched. groups.

Also, please have a look at changes I did in my PR in the UT regarding `PortHostIfCreateFailed` test.
I found this negative test case makes rest of the test cases hang for long time and cause timeouts ( I see here builds are canceled probably because of that ).

prsunny:
@arista-hpandya , can we close this in favor of #3966

arista-hpandya:
Thanks for the PR @stepanblyschak . Yes we can close this PR @prsunny

PR #3947, opened 2025-10-16
Allow the creation of port_serdes object if get fails for the SERDES

arpit-nexthop:
#### Why I did it

Orchagent does not create the port_serdes objects for the gearbox ports if it fails to find initial PORT_SERDES object associated with the port.

Current logic:

- call get with SAI_PORT_ATTR_PORT_SERDES_ID for a PORT object
- If the API returns error, stop every processing
- If the API returns NULL value for serdes object create new serdes object
- If API returns serdes object, delete and create new serdes object

If the object is not present, SAI SDK can return error as well. In such cases, we can assume that the serdes object is not present and proceed with creating it rather than bailing out.

#### How I did it

Allowed for creating serdes object if the API returns error.

#### How to verify it

Tx taps are applied correctly on the gearbox ports

### DIFF
--- a/orchagent/high_frequency_telemetry/hftelorch.cpp
+++ b/orchagent/high_frequency_telemetry/hftelorch.cpp
@@ -228,7 +228,7 @@ task_process_status HFTelOrch::profileTableDel(const std::string &profile_name)
         return task_process_status::task_need_retry;
     }
 
-    if (!profile_itr->second->isEmpty())
+    if (!profile_itr->second->isEmpty() || isProfileInUse(profile_itr->second))
     {
         return task_process_status::task_need_retry;
     }
@@ -353,6 +353,21 @@ std::shared_ptr<HFTelProfile> HFTelOrch::tryGetProfile(const std::string &profil
     }
 
     return std::shared_ptr<HFTelProfile>();
+}
+
+bool HFTelOrch::isProfileInUse(const std::shared_ptr<HFTelProfile> &profile) const
+{
+    SWSS_LOG_ENTER();
+
+    for (const auto &type_profiles : m_type_profile_mapping)
+    {
+        if (type_profiles.second.find(profile) != type_profiles.second.end())
+        {
+            return true;
+        }
+    }
+
+    return false;
 }
 
 void HFTelOrch::doTask(swss::NotificationConsumer &consumer)
--- a/orchagent/high_frequency_telemetry/hftelorch.h
+++ b/orchagent/high_frequency_telemetry/hftelorch.h
@@ -46,6 +46,7 @@ private:
     task_process_status groupTableDel(const std::string &profile_name, const std::string &group_name);
     std::shared_ptr<HFTelProfile> getProfile(const std::string &profile_name);
     std::shared_ptr<HFTelProfile> tryGetProfile(const std::string &profile_name);
+    bool isProfileInUse(const std::shared_ptr<HFTelProfile> &profile) const;
 
     void doTask(swss::NotificationConsumer &consumer);
     void doTask(Consumer &consumer);
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -63,6 +63,7 @@ Srv6Orch *gSrv6Orch;
 FlowCounterRouteOrch *gFlowCounterRouteOrch;
 DebugCounterOrch *gDebugCounterOrch;
 MonitorOrch *gMonitorOrch;
+BfdMonitorOrch *gBfdMonitorOrch;
 TunnelDecapOrch *gTunneldecapOrch;
 StpOrch *gStpOrch;
 MuxOrch *gMuxOrch;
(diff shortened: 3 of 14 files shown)


orchagent/portsorch.cpp
```diff
@@ -9473,18 +9473,12 @@ bool PortsOrch::setPortSerdesAttribute(sai_object_id_t port_id, sai_object_id_t
...
-                       port_attr.id, port_id);
-        task_process_status handle_status = handleSaiGetStatus(SAI_API_PORT, status);
-        if (handle_status != task_process_status::task_success)
-        {
-            return false;
-        }
-    }
 
-    if (port_attr.value.oid != SAI_NULL_OBJECT_ID)
+    /*
```
prgeor, line 9477: @arpit-nexthop In what scenario the port serdes object is NOT present and the OA is calling setPortSerdesAttribute() on that? Do you know the flow?

arpit-nexthop: It is in the initGearboxPort during the create of the port for the first time.

initGearboxPort:

```
            if (serdes_attr.size() != 0)
            {
                if (setPortSerdesAttribute(systemPort, phyOid, serdes_attr))
                {
                    SWSS_LOG_NOTICE("Set port %s system side preemphasis is success", port.m_alias.c_str());
                }
                else
                {
                    SWSS_LOG_ERROR("Failed to set port %s system side pre-emphasis", port.m_alias.c_str());
                    return false;
                }
            }
```
same for the linePort

This tries to fetch the serdes object and fails if we do not have the serdes object already created, which is the case since this is the first create.

arpit-nexthop: @prgeor, let me know if more clarifications are needed.

prgeor: @arpit-nexthop line 9760 the gearbox  system side created here: https://github.com/sonic-net/sonic-swss/blob/b88a5c20efb9296af946d49436d8487affe2dec1/orchagent/portsorch.cpp#L9760

And then at line 9868 line side is created

https://github.com/sonic-net/sonic-swss/blob/b88a5c20efb9296af946d49436d8487affe2dec1/orchagent/portsorch.cpp#L9868

And finally line 9931 setPortSerdesAttribute() is called. So at this point, both  system and line side ports serdes object should already be created.

https://github.com/sonic-net/sonic-swss/blob/b88a5c20efb9296af946d49436d8487affe2dec1/orchagent/portsorch.cpp#L9931

tjchadaga:
@arpit-nexthop - could you please update the branch?

orchagent/portsorch.cpp
arpit-nexthop: @prgeor 

This is based on the assumption that create_port API is required to create the port_serdes object.
I could not find any documentation in SAI that mentions that it is mandatory for sai_create_port_fn call to create the serdes object. We already have an explicit function to create the serdes object(sai_create_port_serdes_fn).

prgeor: @arpit-nexthop gearbox functionality is not new to sonic. Here is the design doc https://github.com/sonic-net/SONiC/blob/master/doc/gearbox/gearbox_mgr_design.md

```
     attr.id = SAI_PORT_CONNECTOR_ATTR_SYSTEM_SIDE_PORT_ID;
            attr.value.oid = systemPort;
            attrs.push_back(attr);
            attr.id = SAI_PORT_CONNECTOR_ATTR_LINE_SIDE_PORT_ID;
            attr.value.oid = linePort;
            attrs.push_back(attr);

            status = sai_port_api->create_port_connector(&connector, phyOid, static_cast<uint32_t>(attrs.size()), attrs.data());
```

The create port connector should created the appropriate line side and system side serdes object.

For non-gearbox platform, where do you see explicit serdes object creation? We only do port creation.

Going back to your logic we should see this issue in non-gearbox platform also where we need to set FIR settings on the host serdes?

arpit-nexthop: @prgeor my concern is with this segment of code.
https://github.com/sonic-net/sonic-swss/blob/master/orchagent/portsorch.cpp#L9474

```
    port_attr.id = SAI_PORT_ATTR_PORT_SERDES_ID;
    status = sai_port_api->get_port_attribute(port_id, 1, &port_attr);
    {
        SWSS_LOG_ERROR("Failed to get port attr serdes id %d to port pid:0x%" PRIx64,
                       port_attr.id, port_id);
        task_process_status handle_status = handleSaiGetStatus(SAI_API_PORT, status);
        if (handle_status != task_process_status::task_success)
        {
            return false;
        }
    }
```

If the get_port_attribute for the serdes object returns error it is a hard failure in orchagent. Next we check if it returned us a object. If it does we delete the serdes object.

```
if (port_attr.value.oid != SAI_NULL_OBJECT_ID)
    {
        status = sai_port_api->remove_port_serdes(port_attr.value.oid);
        {
            SWSS_LOG_ERROR("Failed to remove existing port serdes attr 0x%" PRIx64 " port 0x%" PRIx64,
                           port_attr.value.oid, port_id);
            task_process_status handle_status = handleSaiRemoveStatus(SAI_API_PORT, status);
            if (handle_status != task_success)
            {
                return parseHandleSaiStatusFailure(handle_status);
            }
        }
    }
```

However if it does not return a valid object, orchangent goes along with creating the new object. So the behavior is:

1. SAI sdk returns error for the getting serdes object: orchagent fails
2. SAI sdk returns success but gives a null serdes object: orchagent creates a new serdes object
3. SAI sdk returns success and gives a correct serdes object: orchagent deletes the old serdes object and creates a new one.

The points 1 and 2 behavior look conflicting to me IMHO. Either both should fail orchagent or both should just create a new serdes object.

My proposal here is to go with creating a new serdes object without failing. If we go with returning error in both 1 and 2 cases, existing SAI SDKs can fail.

arpit-nexthop: Conclusion after offline discussion

Since SAI is ambiguous on hard failure vs object not found failure here, we will preserve the existing behavior in orchagent.
SAI failure is hard failure.
SAI success and NULL pointer on serdes object --> Create serdes object

PR #3950, opened 2025-10-23
and l2mcorch for igmp snooping

jasper-hou-micas:
**What I did**

**Why I did it**

**How I verified it**

**Details if related**


orchagent/l2mcorch.cpp
```diff
@@ -0,0 +1,1508 @@
...
+    {
+        status = sai_l2mc_group_api->remove_l2mc_group_member(l2mc_group_member->second.l2mc_member_id);
+        
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("RemoveL2mcMrouterPort: Failed to remove l2mc mrouter group member %lx, from l2mc-grp:%lx rv:%d",
+                           l2mc_group_member->second.l2mc_member_id, l2mc_group->second.l2mc_group_id, status);
+            return false;
+        }
+        l2mc_group->second.l2mc_group_members.erase(l2mc_group_member);
```
aidan-gallagher: **Critical:** Use-after-erase. Line 622 erases `l2mc_group_member` from the map, invalidating the iterator. Lines 624-625 then dereference `l2mc_group_member->second.l2mc_member_id` -- undefined behavior.

Also, `status` is declared uninitialized on line 597 and used in the log at line 625 when the `if` block at line 612 is not entered.

orchagent/l2mcorch.cpp
```diff
@@ -0,0 +1,1508 @@
...
+
+    if (vlan.m_vlan_info.vlan_oid == SAI_NULL_OBJECT_ID)
+    {
+        SWSS_LOG_NOTICE("EnableIgmpSnooping: NULL vlan OID");
+        return false;
+    }
+
+    bzero(&attr, sizeof(attr));
+
+    attr.id = SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE;
```
aidan-gallagher, line 667: **Critical:** `SAI_VLAN_ATTR_CUSTOM_IGMP_SNOOPING_ENABLE` (also used at line 685) is not part of the standard SAI API. This will fail on any non-Broadcom platform. If this is a vendor extension, it would need to be conditionally compiled or handled via a platform-specific mechanism.

cfgmgr/l2mcmgr.h
```diff
@@ -0,0 +1,127 @@
...
+#include <set>
+#include <string>
+#include <unistd.h>
+#include <bitset>
+#include <iostream>
+#include <sstream>
+#include <thread>
+#include <vector>
+
+#include "../../sonic-l2mcd/include/l2mcd_ipc.h"
```
aidan-gallagher: **Critical:** This relative path (`../../sonic-l2mcd/include/l2mcd_ipc.h`) assumes a specific directory layout that only exists in the vendor's build environment. This will break any standard SONiC build. The include path should use a proper `-I` flag or install the header to a standard location.

cfgmgr/l2mcmgr.cpp
```diff
@@ -0,0 +1,1835 @@
...
+
+#define MLD_IP_IPV4_AFI 1
+#define MLD_IP_IPV6_AFI 2
+
+#define TAGGED 1
+#define UNTAGGED 0
+
+#define L2MCD_MAX_SIZE 16318
+
+static char buffer[L2MCD_MAX_SIZE] = {0};
```
aidan-gallagher, line 25: **Critical:** This static global 16KB buffer is shared across all message-building functions. Any future multi-threading or re-entrant call will corrupt data silently. Consider allocating locally or using a per-call buffer.

orchagent/l2mcorch.cpp
```diff
@@ -0,0 +1,1508 @@
...
+                    l2mc_GrpKey.vlan_alias.c_str(), l2mc_group->second.l2mc_group_id);
+    
+    auto l2mc_group_member = l2mc_group->second.l2mc_group_members.begin();
+    while(l2mc_group_member != l2mc_group->second.l2mc_group_members.end())
+    {
+        Port port;
+        if (!gPortsOrch->getPort(l2mc_group_member->first, port))
+        {
+            SWSS_LOG_NOTICE("RemoveL2mcGroupMembers: Failed to locate port %s", port.m_alias.c_str());
+            continue;
```
aidan-gallagher, line 271: **High:** `continue` here does not advance the iterator. Since `l2mc_group_member` is only advanced by the `erase()` call on the success path (line 279), a persistent `getPort()` failure on this member will cause an infinite loop.

The same pattern appears in `doL2mcMrouterTask` at lines 1254-1258 and 1260-1264, where both port and vlan lookup failures use `continue` without advancing `it`.

orchagent/l2mcorch.cpp
```diff
@@ -0,0 +1,1508 @@
...
+            mrouter_ports = &mrouter_ports_ipv4_per_vlan[vlan_alias];
+        }
+        else if (protocol == "V6")
+        {
+            mrouter_ports = &mrouter_ports_ipv6_per_vlan[vlan_alias];
+        }
+        else
+        {
+            SWSS_LOG_ERROR("Invalid protocol %s for VLAN %s", protocol.c_str(), vlan_alias.c_str());
+            return;
```
aidan-gallagher, line 1345: **High:** `return` here exits the entire function when an invalid protocol value is encountered, leaving all subsequent entries in the consumer unprocessed. They will be retried on the next event but the invalid entry will cause the same `return` each time.

The same issue exists in `l2mcmgr.cpp:919` where `doVlanUpdateTask` returns on a single invalid key, abandoning the rest of the batch. Both should use `continue` (with `it = consumer.m_toSync.erase(it)`) instead of `return`.

aidan-gallagher (commented):
This PR adds `l2mcorch` and `l2mcmgr` for L2 multicast orchestration and config management. There are several memory safety issues (use-after-erase, uninitialized values), infinite loop bugs, unchecked conversions, and a non-portable SAI attribute.

| Critical | High | Medium | Low | Total |
|----------|------|--------|-----|-------|
| 5 | 10 | 9 | 6 | **30** |

**Unchecked `stoi()` (Critical):** There are ~25 unchecked `stoi()` calls across `l2mcmgr.cpp` and `l2mcorch.cpp` (e.g., lines 189, 262, 317, 528, 583, 722, 788, 876, 923, 1283, 1404, 1559, 1631 in l2mcmgr.cpp). Only two locations (lines ~471, ~650) have try/catch. Any malformed Redis key or value will throw `std::invalid_argument` and crash the daemon.

Additional medium/low findings not called out inline: variable shadowing in `doL2mcSuppressTask` (lines 1318/1357 shadow outer `it`), no duplicate check when adding mrouter ports (line 1285 -- same port added to vector twice), `l2mcd_fd` not reset to -1 on `bind()` failure (line ~993), `using namespace std;` and `using namespace swss;` in `l2mcmgr.h:33-34`, and duplicate `extern PortsOrch *gPortsOrch` declarations (lines 23 and 26 of l2mcorch.cpp).

cfgmgr/l2mcmgr.cpp
```diff
@@ -0,0 +1,1835 @@
...
+    {
+		SWSS_LOG_ERROR("tx_msg send error type:%d len%d",msgType,(int)len);
+    }
+    SWSS_LOG_NOTICE("sendMsgL2Mcd tx_msg send type:%d len%d",msgType,(int)len);
+    free(tx_msg);
+    return rc;
+}
+
+/* To check the port init is done or not */
+int L2McMgr::isPortInitComplete(DBConnector *app_db)
```
aidan-gallagher, line 1084: **High:** This busy-wait loop has no timeout. `cnt` is incremented (line 1045) but never checked against a limit. If `PortInitDone` is never set, the daemon hangs here forever. The unconditional `sleep(5)` after the loop (line 1050) also runs even when the loop exits on first try.

orchagent/l2mcorch.cpp
```diff
@@ -0,0 +1,1508 @@
...
+L2mcOrch::L2mcOrch(DBConnector * appDb, vector<string> &tableNames) :
+    Orch(appDb, tableNames)
+{
+    SWSS_LOG_ENTER();
+    gPortsOrch->attach(this);
+};
+
+bool L2mcOrch::hasL2mcGroup(string vlan, const L2mcGroupKey &l2mckey)
+{
+    return m_syncdL2mcEntries.at(vlan).find(l2mckey) != m_syncdL2mcEntries.at(vlan).end();
```
aidan-gallagher: **High:** `.at()` throws `std::out_of_range` if `vlan` is not in `m_syncdL2mcEntries`. `AddL2mcEntry` at line 311 calls this function without first ensuring the vlan key exists. A `.find()` + check or `.count()` guard would be safer.

orchagent/l2mcorch.cpp
```diff
@@ -0,0 +1,1508 @@
...
+{
+    switch (type)
+    {
+        case L2mcSuppressType::UNKNOWN_IPV4:
+            attr.value.oid = enable ? info.ipv4_group_id : SAI_NULL_OBJECT_ID;
+            attr.id = SAI_VLAN_ATTR_UNKNOWN_IPV4_MCAST_OUTPUT_GROUP_ID;
+            break;
+        case L2mcSuppressType::IPV4_LINK_LOCAL:
+            attr.value.oid = enable ? info.ipv4_group_id : SAI_NULL_OBJECT_ID;
+            attr.id = SAI_VLAN_ATTR_UNKNOWN_LINKLOCAL_MCAST_OUTPUT_GROUP_ID;
```
aidan-gallagher, line 724: **High:** Both `IPV4_LINK_LOCAL` (line 710) and `IPV6_LINK_LOCAL` (line 718) set the same SAI attribute `SAI_VLAN_ATTR_UNKNOWN_LINKLOCAL_MCAST_OUTPUT_GROUP_ID`. Whichever runs second overwrites the first. IPv4 and IPv6 link-local suppression may need separate attributes or a combined group to work correctly.

orchagent/l2mcorch.cpp
```diff
@@ -0,0 +1,1508 @@
...
+        default:
+            return "UNKNOWN";
+    }
+}
+
+bool L2mcOrch::handleL2mcSuppression(const std::string &vlan_alias, bool enable, L2mcSuppressType type)
+{
+    SWSS_LOG_ENTER();
+
+    Port vlan, port;
```
aidan-gallagher, line 831: **High:** `port` is default-constructed here but never populated via `getPort()`. It is then passed to `createL2mcGroup` at lines 858 and 871. `Port`'s default constructor leaves `m_bridge_port_id` as 0, which is not the same as `SAI_NULL_OBJECT_ID`. The null-check at line 390 will pass, and SAI receives an invalid bridge port OID.

Related: `SuppressionGroupInfo` (l2mcorch.h:73-74) has `ipv4_group_id` and `ipv6_group_id` fields that are not initialized to `SAI_NULL_OBJECT_ID`, but the code checks them against `SAI_NULL_OBJECT_ID` at line 854.

orchagent/l2mcorch.cpp
```diff
@@ -0,0 +1,1508 @@
...
+
+    if (isSuppressionUnchanged(type,enable,info))
+    {
+        SWSS_LOG_NOTICE("L2MC suppression status unchanged for VLAN %s [%s]", 
+                        vlan_alias.c_str(),
+                        (type == L2mcSuppressType::UNKNOWN_IPV4 ? "Unknown IPv4" : "Link-local"));
+        return true;
+    }
+
+    updateSuppressionInfo(type, enable,info);
```
aidan-gallagher, line 869: **High:** `updateSuppressionInfo()` modifies the suppression state here, but the SAI call that depends on that state doesn't happen until line 884. If the SAI call fails (line 885), the function returns `false` but the state has already been modified and is not rolled back.

cfgmgr/l2mcmgr.cpp
```diff
@@ -0,0 +1,1835 @@
...
+    {
+        SWSS_LOG_ERROR(
+            "L2MCD_CFG:SNOOP payload too large: %u (max %u)",
+            msg_len, L2MCD_MAX_SIZE);
+        return;
+    }
+    memset(buffer, 0, msg_len);
+    msg = (L2MCD_CONFIG_MSG *)buffer;
+
+    size_t pos=key.find(key_seperator.c_str());
```
aidan-gallagher, line 1647: **Medium:** This chain of `find()` calls (lines 1547-1558) does not check any return value against `string::npos`. If the key has fewer separators than expected, `substr()` will produce garbage or throw `std::out_of_range`.

orchagent/l2mcorch.cpp
```diff
@@ -0,0 +1,1508 @@
...
+        SWSS_LOG_ERROR("RemoveL2mcEntry: Failed to remove l2mc group %lx, rv:%d", l2mc_group_id, status);
+        return false;
+    }
+
+    m_syncdL2mcEntries.at(vlan.m_alias).erase(key);
+
+    return true;
+}
+
+bool L2mcOrch::RemoveL2mcEntrys(Port &vlan)
```
aidan-gallagher, line 480: **Medium:** When a VLAN is removed in this function, `mrouter_ports_ipv4_per_vlan` and `mrouter_ports_ipv6_per_vlan` are not cleaned up for the removed VLAN. These maps will accumulate stale entries over time.

cfgmgr/l2mcmgr.h
```diff
@@ -0,0 +1,127 @@
...
+    L2McMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *staDb,
+            const vector<TableConnector> &tables);
+
+    using Orch::doTask;
+	void ipcInitL2McMgr();
+    int sendMsgL2Mcd(L2MCD_MSG_TYPE msgType, uint32_t msgLen, void *data);
+    int isPortInitComplete(DBConnector *app_db);
+    int getL2McPortList(DBConnector *state_db);
+    int getL2McCfgParams(DBConnector *cfgDb);
+    bool isRestoreDone();
```
aidan-gallagher: **Medium:** `isRestoreDone()` is declared here but has no definition in `l2mcmgr.cpp`. This will cause a linker error if anything calls it.

cfgmgr/l2mcmgr.cpp
```diff
@@ -0,0 +1,1835 @@
...
+                SWSS_LOG_NOTICE("L2MCMGRD warm reboot RECONCILED");
+                WarmStart::setWarmStartState("l2mcmgrd", WarmStart::RECONCILED);
+                m_warmstart = false;
+            }
+        }
+            
+    }
+}
+
+void L2McMgr::waitTillReadyToReconcile()
```
aidan-gallagher, line 1937: **Medium:** This `for (;;)` loop with `sleep(1)` has no timeout. If `vlanmgrd` never reaches `REPLAYED` or `RECONCILED` state, the daemon blocks here indefinitely. Consider adding a maximum iteration count or timeout.

orchagent/l2mcorch.cpp
jasper-hou-micas: Will further revise and optimize accordingly

orchagent/l2mcorch.cpp
jasper-hou-micas: In SAI.h, the naming is defined as such

cfgmgr/l2mcmgr.h
jasper-hou-micas: Will further revise and optimize accordingly

cfgmgr/l2mcmgr.cpp
jasper-hou-micas: Since there are no multi-threaded calls, this approach is simple and effective.

orchagent/l2mcorch.cpp
jasper-hou-micas: Will further revise and optimize accordingly

orchagent/l2mcorch.cpp
jasper-hou-micas: Will further revise and optimize accordingly

cfgmgr/l2mcmgr.cpp
jasper-hou-micas: Will further revise and optimize accordingly

orchagent/l2mcorch.cpp
jasper-hou-micas: Will further revise and optimize accordingly

orchagent/l2mcorch.cpp
jasper-hou-micas: SUPPRESS in the SAI definition does not distinguish between IPv4 and IPv6

orchagent/l2mcorch.cpp
jasper-hou-micas: Will further revise and optimize accordingly

orchagent/l2mcorch.cpp
jasper-hou-micas: Will further revise and optimize accordingly

cfgmgr/l2mcmgr.cpp
jasper-hou-micas: Will further revise and optimize accordingly

orchagent/l2mcorch.cpp
jasper-hou-micas: Will further revise and optimize accordingly

cfgmgr/l2mcmgr.h
jasper-hou-micas: Will further revise and optimize accordingly

cfgmgr/l2mcmgr.cpp
jasper-hou-micas: Reused from other sonic modules; it will be revised and optimized

aidan-gallagher:
I believe there is a memory-safety issue in `cfgmgr/l2mcmgr.cpp`: all ten IP-address copies use `memcpy(..., string.c_str(), L2MCD_IP_ADDR_STR_SIZE)`. `L2MCD_IP_ADDR_STR_SIZE` is 46, while `std::string::c_str()` only guarantees `size() + 1` readable bytes. Short IPv4/IPv6 strings therefore cause an out-of-bounds source read and copy adjacent memory into the l2mcd IPC message.

The most exposed instances are in `doL2McProcRemoteEntries`, reached through `L2MC_NOTIFICATIONS_REMOTE`, but the same pattern also occurs in the static-entry and L3-interface paths.

A uniform fix is to clear the destination and copy at most 45 bytes:

```cpp
static void copyIpAddress(char *destination, const string &address)
{
    const size_t copy_length = min(address.size(), static_cast<size_t>(L2MCD_IP_ADDR_STR_SIZE - 1));

    memset(destination, 0, L2MCD_IP_ADDR_STR_SIZE);
    memcpy(destination, address.c_str(), copy_length);
}
```

jasper-hou-micas:
@aidan-gallagher 
You are absolutely right. We overlooked this memory-safety issue during the initial development.

To address this, we will revise this part of the code using `strncpy(buffer, ip_str.c_str(), sizeof(buffer) - 1)`. We prefer this approach for the following reasons:

1. `strncpy` automatically pads the remaining destination buffer with null bytes `\0` if the source string is shorter than the specified length, which perfectly handles the short IPv4/IPv6 string scenario.
2. It keeps the modifications to the original codebase minimal and avoids introducing additional conditional logic or branching (such as `std::min`).

Thank you for catching this and providing the detailed analysis! I will push a new commit to fix this shortly.

PR #3956, opened 2025-10-28
[tests]: Update SRv6 test cases to use static SID configuration CLI

cscarpitta:
**What I did**

Previously, SRv6 test cases relied on a workaround to configure SIDs by creating them directly in the kernel. The kernel would then notify FRR, which in turn would program the SIDs in SONiC. This was the only available method before the introduction of the static SRv6 CLI in FRR.

With the introduction of the static SRv6 CLI, we can now configure SIDs directly through FRR, which is the intended and correct approach.

This commit updates the test cases to leverage the static SRv6 CLI.

**Why I did it**

This change aligns the test cases with the recommended and expected workflow for configuring SIDs in SONiC, which is also used by `bgpcfgd`.

**How I verified it**

Run the SRv6 test suite:
`pytest --dvsname=vs test_srv6.py`

### DIFF
--- a/orchagent/p4orch/next_hop_manager.cpp
+++ b/orchagent/p4orch/next_hop_manager.cpp
@@ -36,115 +36,201 @@ P4NextHopEntry::P4NextHopEntry(const std::string &next_hop_id, const std::string
     next_hop_key = KeyGenerator::generateNextHopKey(next_hop_id);
 }
 
-namespace
-{
+ReturnCode NextHopManager::validateAppDbEntry(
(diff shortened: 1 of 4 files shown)


ahsalam (approved):
LGTM. Thanks for the fix.

GaladrielZhao (approved):
Thanks for your update. Looks good to me.

prsunny (approved):
approved for merging

cscarpitta:
> approved for merging


@prsunny Many thanks for the review.

The PR passed the CI. Can you please help to merge?

PR #3957, opened 2025-10-28
Add PORT_PHY_ATTR flex counter support

dhanasekar-arista:
**What I did**
Added support for polling PORT PHY attributes using the flex counter infrastructure.
Following changes are added in swss to support this feature
 - define a new counter type PORT_PHY_ATTR
 - support added for polling following attributes, 
   - SAI_PORT_ATTR_RX_SIGNAL_DETECT, 
   - SAI_PORT_ATTR_FEC_ALIGNMENT_LOCK, 
   - SAI_PORT_ATTR_RX_SNR
 - Add port_phy_attr_manager to set counter id list in FLEX_COUNTERS_DB
 - Added unit tests to validate the enable/disable code flow, generatePortAttrCounterMap and QueryPortAttrCapabilitiesWithMockedSAI.

**Why I did it**
This helps to monitor the port serdes level attributes.
 
**How I verified it**
Verified by using sonic-mgmt tests (https://github.com/sonic-net/sonic-mgmt/pull/21601).

**Details if related**


prgeor:
@dhanasekar-arista PR not ready for review? I see its in draft

orchagent/portsorch.cpp
```diff
@@ -8807,6 +8836,158 @@ void PortsOrch::generatePortBufferDropCounterMap()
...
+        {
+            SWSS_LOG_NOTICE("PORT_PHY_ATTR: Attribute %s is NOT supported (status=%d, get_implemented=%d)",
+                            attr_name, status, capability.get_implemented);
+        }
+    }
+
+    m_phy_attr_capability_checked = true;
+}
+
+bool PortsOrch::verifyPortSupportsAllPhyAttr(sai_object_id_t port_id, const char* port_name)
```
prgeor, line 8982: @dhanasekar-arista there could be platforms that may support only few of these attributes and in that case the supported one won't be polled? is it all or nothing case?

orchagent/portsorch.cpp
```diff
@@ -8807,6 +8836,158 @@ void PortsOrch::generatePortBufferDropCounterMap()
...
+    {
+        SWSS_LOG_ERROR("PORT_PHY_ATTR: Port %s does not support FEC_ALIGNMENT_LOCK attribute (status=%d)",
+                      port_name, status);
+        return false;
+    }
+    SWSS_LOG_DEBUG("PORT_PHY_ATTR: Port %s supports FEC_ALIGNMENT_LOCK attribute (count=%d)",
+                   port_name, test_attr.value.portlanelatchstatuslist.count);
+
+    // Check RX_SNR
+    test_attr.id = SAI_PORT_ATTR_RX_SNR;
```
prgeor: @dhanasekar-arista can you rename ? `test_attr` -> `port_attr`? The test name in non test file does not look good.

orchagent/portsorch.cpp
dhanasekar-arista: yes, currently it is designed as a all or none case. At that time, I was facing some challenges in querying only supported attributes. I can take this up as a enhancement.

orchagent/flexcounterorch.h
```diff
@@ -57,6 +57,7 @@ class FlexCounterOrch: public Orch
     FlexCounterOrch(swss::DBConnector *db, std::vector<std::string> &tableNames);
     virtual ~FlexCounterOrch(void);
     bool getPortCountersState() const;
+    bool getPortAttrCountersState() const;
```
prgeor: @dhanasekar-arista can you rename to getPortPhyAttrCounterState()? Since in future there could be other set of port related counters and we want to distinguish PHY port attributes differently.

orchagent/flexcounterorch.cpp
```diff
@@ -322,6 +324,19 @@ void FlexCounterOrch::doTask(Consumer &consumer)
                     {
                         gSrv6Orch->setCountersState((value == "enable"));
                     }
+                    if (gPortsOrch && (key == PORT_PHY_ATTR_KEY))
+                    {
+                        if(value == "enable" && !m_port_phy_attr_enabled)
```
prgeor, line 328: @dhanasekar-arista is this counter NOT enabled by default? can you keep this enabled by default.

orchagent/portsorch.cpp
```diff
@@ -8807,6 +8836,158 @@ void PortsOrch::generatePortBufferDropCounterMap()
...
+        return false;
+    }
+    SWSS_LOG_DEBUG("PORT_PHY_ATTR: Port %s supports RX_SNR attribute (count=%d)",
+                   port_name, test_attr.value.portsnrlist.count);
+    return true;
+}
+
+void PortsOrch::generatePortAttrCounterMap()
+{
+    if (m_isPortAttrCounterMapGenerated)
```
prgeor: @dhanasekar-arista This function is called only once by flexcounterorch as there is already a check for `m_port_phy_attr_enabled` . so why is this variable required?

orchagent/flexcounterorch.cpp
dhanasekar-arista: No its not enabled by default, it is config driven.
If needed we can add this to init config to keep this enabled by default, https://github.com/sonic-net/sonic-buildimage/blob/master/files/build_templates/init_cfg.json.j2#L24.

If there is any other way to have this enabled by default, please let me know.

prsunny:
@prgeor , can you approve?

prsunny:
@dhanasekar-arista , please resolve conflict

prgeor (approved):

prgeor:
@dhanasekar-arista please raise the conflict PR for 202511

orchagent/portsorch.cpp
gordon-nexthop: Following up on this - for example TH6 does not support `fec_alignment_lock` but does support `rx_snr` and `rx_signal_detect`

r12f:
Pick to 202511: https://github.com/sonic-net/sonic-swss/pull/4214

PR #3959, opened 2025-10-28
Add test telemetry mock

meghtony:
**What I did**
Added a new mock test to the flexcounter_ut.cpp file. It is a conversion of:  test_telemetry_period.py.

**Why I did it**
This converts the relevant pytest behavior into fast, deterministic GoogleTests that don’t depend on DVS or ASIC_DB.

**How I verified it**
Will be seeing if the pipeline fails.

**Details if related**


tests/mock_tests/flexcounter_ut.cpp
```diff
@@ -996,6 +996,122 @@ namespace flexcounter_test
...
+        gp.operation.list           = reinterpret_cast<int8_t*>(const_cast<char*>(op.c_str()));
+
+        if (!plugin_key.empty())
+        {
+            gp.plugin_name.count    = static_cast<uint32_t>(plugin_key.size() + 1);
+            gp.plugin_name.list     = reinterpret_cast<int8_t*>(const_cast<char*>(plugin_key.c_str()));
+        }
+
+        sai_attribute_t attr{};
+        attr.id        = SAI_REDIS_SWITCH_ATTR_FLEX_COUNTER_GROUP;
```
prabhataravind, line 1081: these are probably not needed to be added again. You can use the existing functions which set these attributes during test setup. See _ut_stub_sai_set_switch_attribute for reference.

PR #3960, opened 2025-10-28
Allow state db to take modified entries made to the tunnel decap table

developfast:
**What I did**
Allow state db to take modified entries made to the tunnel decap table

**Why I did it**
Prevent DB inconsistency between state, asic and appl db 

**How I verified it**
redis-cli -n 6 HGETALL "TUNNEL_DECAP_TABLE|IPINIP_V6_TUNNEL"
1) "tunnel_type"
2) "IPINIP"
3) "dscp_mode"
4) "pipe"
5) "ecn_mode"
6) "standard"
7) "ttl_mode"
8) "pipe"

/var/log/syslog.1:2025 Nov 10 19:18:32.730385 bjw-can-2700-2 NOTICE swss#orchagent: :- doDecapTunnelTask: Fields for TUNNEL_DECAP_TABLE entry 'IPINIP_TUNNEL' have been synchronised in STATE_DB
/var/log/syslog.1:2025 Nov 10 19:18:32.740298 bjw-can-2700-2 NOTICE swss#orchagent: :- doDecapTunnelTask: Fields for TUNNEL_DECAP_TABLE entry 'IPINIP_V6_TUNNEL' have been synchronised in STATE_DB

Added comprehensive unit tests too which pass

```
make  check-TESTS
make[1]: Entering directory '/home/devojha/data/sonic-swss/tests/mock_tests'
make[2]: Entering directory '/home/devojha/data/sonic-swss/tests/mock_tests'
PASS: tests
PASS: tests_intfmgrd
PASS: tests_teammgrd
PASS: tests_portsyncd
PASS: tests_fpmsyncd
PASS: tests_response_publisher
============================================================================
Testsuite summary for sonic-swss 1.0
============================================================================
# TOTAL: 6
# PASS:  6
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
make[2]: Leaving directory '/home/devojha/data/sonic-swss/tests/mock_tests'
make[1]: Leaving directory '/home/devojha/data/sonic-swss/tests/mock_tests'
```

**Details if related**

### DIFF
--- a/orchagent/tunneldecaporch.cpp
+++ b/orchagent/tunneldecaporch.cpp
@@ -109,6 +109,8 @@ void TunnelDecapOrch::doDecapTunnelTask(Consumer &consumer)
 
         sai_object_id_t tunnel_id = SAI_NULL_OBJECT_ID;
 
+        // track whether any oper fields changed on this key
+        bool state_changed = false;
         // checking to see if the tunnel already exists
         bool exists = (tunnelTable.find(key) != tunnelTable.end());
         if (exists)
@@ -160,8 +162,10 @@ void TunnelDecapOrch::doDecapTunnelTask(Consumer &consumer)
                     }
                     if (exists)
                     {
+                        // Apply to SAI; only touch cache/flag on success
                         setTunnelAttribute(fvField(i), dscp_mode, tunnel_id);
                         tunnelTable[key].dscp_mode = dscp_mode;
+                        state_changed = true;
                     }
                 }
                 else if (fvField(i) == "ecn_mode")
@@ -207,7 +211,9 @@ void TunnelDecapOrch::doDecapTunnelTask(Consumer &consumer)
                     }
                     if (exists)
                     {
+                        // Apply to SAI; only touch cache/flag on success
                         setTunnelAttribute(fvField(i), ttl_mode, tunnel_id);
+                        state_changed = true;
                     }
                 }
                 else if (fvField(i) == decap_dscp_to_tc_field_name)
@@ -221,7 +227,9 @@ void TunnelDecapOrch::doDecapTunnelTask(Consumer &consumer)
                     }
                     if (exists)
                     {
+                        // Apply to SAI; only touch cache/flag on success
                         setTunnelAttribute(fvField(i), dscp_to_tc_map_id, tunnel_id);
+                        state_changed = true;
                     }
                 }
                 else if (fvField(i) == decap_tc_to_pg_field_name)
@@ -235,7 +243,9 @@ void TunnelDecapOrch::doDecapTunnelTask(Consumer &consumer)
                     }
                     if (exists)
                     {
+                        // Apply to SAI; only touch cache/flag on success
                         setTunnelAttribute(fvField(i), tc_to_pg_map_id, tunnel_id);
+                        state_changed = true;
                     }
                 }
                 else if (fvField(i) == encap_tc_to_dscp_field_name)
@@ -251,6 +261,7 @@ void TunnelDecapOrch::doDecapTunnelTask(Consumer &consumer)
                     {
                         // Record only
                         tunnelTable[key].encap_tc_to_dscp_map_id = tc_to_dscp_map_id;
+                        state_changed = true;
                     }
                 }
                 else if (fvField(i) == encap_tc_to_queue_field_name)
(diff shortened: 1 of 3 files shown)


orchagent/tunneldecaporch.cpp
```diff
@@ -207,7 +211,9 @@ void TunnelDecapOrch::doDecapTunnelTask(Consumer &consumer)
                     }
                     if (exists)
                     {
+                        // Apply to SAI; only touch cache/flag on success
                         setTunnelAttribute(fvField(i), ttl_mode, tunnel_id);
```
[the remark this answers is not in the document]
developfast: good catch - will take this up in a separate PR since I want to just keep this for state db syncing

developfast:
@kperumalbfn can you pls check why the checks won't run? The agents don't seem to be grabbing the workers before the 45 min timeout

developfast:
@prsunny @kperumalbfn can you pls review?

orchagent/tunneldecaporch.cpp
```diff
@@ -276,6 +288,12 @@ void TunnelDecapOrch::doDecapTunnelTask(Consumer &consumer)
                 }
             }
 
+            if (exists && state_changed) {
```
prsunny: Why you want an addition '`state_changed` variable? why not simply check 'exists' and set STATE_DB?

developfast: If none of the fields were updated, then there's no point updating the state db which this `state_changed` variable keeps track of

prsunny: Its fine, the current check is 'if exists' its anyways setting the SAI attribute.

developfast: not necessarily, some attributes like ecn are not set since they are not supported in SAI and still have the exists keyword check; the `state_changed` variable ensures that if the SAI attribute is not being changed, then the state db does not take in a false change

kperumalbfn: Could you add a comment that ECN_MODE SET attribute is not supported in SAI, so that we will revisit this check after updating SAI headers.

prsunny: also, this is not a frequent operation of setting the fields, so imo, if 'exists' its still ok to just update STATE_DB. it will be a no-op

tests/mock_tests/tunneldecaporch_ut.cpp
```diff
@@ -0,0 +1,375 @@
...
+        // Test 5: Test that getDscpMode works after setting up tunnel data
+        EXPECT_NO_THROW({
+            string dscp_mode = tunnelDecapOrch->getDscpMode("invalid_tunnel");
+            EXPECT_TRUE(dscp_mode.empty());
+        });
+
+        // Test 6: Test that getDstIpAddresses works with database entries
+        EXPECT_NO_THROW({
+            auto dst_ips = tunnelDecapOrch->getDstIpAddresses("empty_tunnel");
+            EXPECT_EQ(dst_ips.getSize(), 0);
```
kperumalbfn, line 371: Could you add UT for set attribute and check the state_db?

developfast: added

orchagent/tunneldecaporch.cpp
```diff
@@ -276,6 +280,12 @@ void TunnelDecapOrch::doDecapTunnelTask(Consumer &consumer)
                 }
             }
 
+            if (exists) {
```
prsunny: if` (valid && exists)`

prsunny: { -> new line

developfast: no need for valid since multiple attributes can be changed in a single call to this function. for instance, dscp and ecn value can be in SET mode where dscp set is allowed and SAI change is made to change asic attribute but ecn mode set is not allowed and the valid flag is set to false. in this case, I still want the dscp mode change to be reflected in the state db. Moreover, state db changes are only reflected if the cached tunnel table has changes to it which doesn't happen when valid = false.

prsunny: ok, i was thinking of this specific case -  `SWSS_LOG_ERROR("unknown decap tunnel table attribute '%s'.", fvField(i).c_str());` where valid will be false. But agree with you considering the current code flow of setting `valid `field

prsunny (approved):

kperumalbfn (approved):

ayurkiv-nvda:
Need to add this PR to 202511 branch

PR #3963, opened 2025-10-30
Explicitly setting the decap TTL mode for VxLAN tunnels

mramezani95:
**What I did**
Set the decap TTL mode of VxLAN tunnels according to the `ttl_mode` attribute.
`ttl_mode` is a new attribute that can be set for each VxLAN in the `VXLAN_TUNNEL` table in `CONFIG DB`. It can be set to `uniform` or `pipe` (similar to ordinary IP-in-IP tunnels). For each VxLAN tunnel,
1. If `ttl_mode` is set, then the corresponding SAI attribute is used to set the decap TTL mode for that VxLAN tunnel on the ASIC.
2. Otherwise, no SAI attribute is added for the decap TTL mode (i.e., we fall back to the default ASIC behavior).

**Why I did it**
For the decap scenario, the default expectation is to keep the inner TTL (i.e., the pipe model). I added support for the `ttl_mode` attribute to be able to control this behavior if needed.

**How I verified it**
1. Ran the modified VxLAN VS tests.
2. Tested on different platforms to ensure that the decap TTL mode is set correctly in SAI records.

**Details if related**
N/A


prsunny:
@mramezani95 , can you update the following for this PR
1. Enable do_task to process a 'ttl' mode of 'uniform' or 'pipe'
2. Corresponding yang change if not already present
3. Golden config change for setting 'ttl' to 'pipe'

orchagent/vxlanorch.cpp
```diff
@@ -360,6 +360,10 @@ create_tunnel(
         tunnel_attrs.push_back(attr);
     }
 
+    attr.id = SAI_TUNNEL_ATTR_DECAP_TTL_MODE;
+    attr.value.s32 = SAI_TUNNEL_TTL_MODE_PIPE_MODEL;
+    tunnel_attrs.push_back(attr);
```
bingwang-ms: Since this change is generic, please confirm the issue doesn't break warm-reboot. 
During warm-reboot, there could be a vxlan tunnel created for handling packet from CPA.

mramezani95:
> @mramezani95 , can you update the following for this PR
> 
> 1. Enable do_task to process a 'ttl' mode of 'uniform' or 'pipe'
> 2. Corresponding yang change if not already present
> 3. Golden config change for setting 'ttl' to 'pipe'

1. Done
2. https://github.com/sonic-net/sonic-buildimage/pull/24657
3. As discussed, no changed to golden config are needed.

prsunny:
@mramezani95 , would you also do a test with warmboot from an old version to this change? Expect no impact.

orchagent/vxlanorch.cpp
```diff
@@ -358,6 +372,13 @@ create_tunnel(
         attr.id = SAI_TUNNEL_ATTR_ENCAP_TTL_VAL;
         attr.value.u8 = encap_ttl;
         tunnel_attrs.push_back(attr);
+
+        if (decap_ttl_mode == VxlanTunnelTTLMode::NOT_SET)
```
prsunny: please add a comment here mentioning, if encap ttl is pipe mode, defaulting decap to pipe

mramezani95: Added a comment.

orchagent/vxlanorch.h
```diff
@@ -241,6 +249,7 @@ const request_description_t vxlan_tunnel_request_description = {
             {
                 { "src_ip", REQ_T_IP },
                 { "dst_ip", REQ_T_IP },
+                { "ttl_mode", REQ_T_STRING },
```
dgsudharsan, line 252: If we have an encap_ttl_mode in future how will we differentiate the attribute name? Should we consider renaming this to be more specific or fall inline with ip in ip tunnel to just imply that ttl_mode by default refers to decap_ttl_mode?

prsunny:
sonic-mgmt test - https://github.com/sonic-net/sonic-mgmt/pull/21156

PR #3966, opened 2025-10-30
[portsorch] fix crash when number of PGs returned 0

stepanblyschak:
**What I did**

Fix issue reported in https://github.com/sonic-net/sonic-swss/issues/3961.
Regression happened after move to bulk implementation. The error handling of no PGs, queues was incorrect.

Bulk array preparation and reading should use a consistent approach. For example, if a port is skipped in the input array, it should also be skipped when reading from the output array.

**Why I did it**

Fix issue reported in https://github.com/sonic-net/sonic-swss/issues/3961.

**How I verified it**

Implemented a UT to ensure no crash happens when number of PGs, queues or scheudler group is 0.

**Details if related**

### DIFF
--- a/tests/mock_tests/portsorch_ut.cpp
+++ b/tests/mock_tests/portsorch_ut.cpp
@@ -2925,45 +2925,6 @@ namespace portsorch_test
         }
     }
 
-    TEST_F(PortsOrchTest, PortHostIfCreateFailed)
-    {
-        Table portTable = Table(m_app_db.get(), APP_PORT_TABLE_NAME);
-
-        auto original_api = sai_hostif_api->create_hostif;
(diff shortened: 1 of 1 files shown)


stepanblyschak:
@sreeiyer-1 Could you please check ?

sreeiyer-1:
I raised https://github.com/sonic-net/sonic-swss/issues/3961 . Can confirm that after including the diff from this PR to my private view, the crash is not happening anymore.

stepanblyschak:
@prsunny @bingwang-ms @prgeor Could you please have a look?

kperumalbfn (commented):
LGTM

kperumalbfn (approved):

zhixzhu:
@stepanblyschak @prsunny Same crash observed in 202511 image, can you add label to do double commit to 202511?

PR #3974, opened 2025-11-05
[syncd] Fix the error log while running lua plugin

vivekrnv:
**What I did**

Fix https://github.com/sonic-net/sonic-buildimage/issues/24342

**Why I did it**

Variable is uninitialized which is causing the error logs mentioned in the issue during startup

**How I verified it**

Don't see the error logs after this fix

**Details if related**


vivekkarri:
@prabhataravind FYI

prsunny (approved):
@prgeor for viz

StormLiangMS:
hi @vmittal-msft  For 202511 cherrypick.

PR #3979, opened 2025-11-08
With ZMQ enabled between fpmsyncd to orchagent, default routes are se…

venkit-nexthop:
The following issues are resolved in this PR:

1. When ZMQ is enabled, default routes are set to DROP even when there is a valid default route present.

Issue# https://github.com/sonic-net/sonic-swss/issues/3978

2. ZMQ configuration does not work when mgmt VRF is configured

https://github.com/sonic-net/sonic-swss/issues/3985

**What I did**
Fixed orchagent to be able to handle DEL+SET. This was the root cause for the default route being set to DROP.
Fix fpmsyncd to not send DEL+SET and instead only send a SET for routes. This is achieved by including all the fields, even the ones not present, with a default empty value.
This is more efficient now.

**Why I did it**
Recently ZMQ was enabled in sonic-mgmt (refer to sonic-net/sonic-mgmt@111e635). fpmsyncd sends a DEL+SET for every route. This was coalesced by the producerStateTable infra. But with ZMQ enabled, this coalescing does not happen. As a result, orchagent gets a DEL+SET for every route. This works well normally. But for default routes, there is a bug in orchagent, where it adds a DROP action when it receives the DEL. But when the subsequent SET is received, it does not reset it to FORWARD action. This is due to a bug in its checking code.

**How I verified it**
The default route is set to the correct FORWARD entry.

**Details if related**


venkit-nexthop:
tagging @xwjiang-ms

xwjiang-ms:
Created test PR: https://github.com/sonic-net/sonic-buildimage/pull/24484

PR #3984, opened 2025-11-10
Connect to teamd before adding the lag to STATE_DB

saiarcot895:
**What I did**

It's possible for teamd to start up, create the port channel interface in the kernel, and then later exit because some failure condition was hit, or initialization took too long (and so the parent process killed it), or something else. When this happens, teamsyncd will get a notification from the kernel saying that the port channel interface has been created, and will start to add it into STATE_DB. If this happens before teamd goes down (and the port channel interface gets removed from the kernel), then anything depending on that STATE_DB entry will begin its processing, not realizing that that interface will get removed. This will result in dependent applications thinking everything has been processed and configs have been applied, but they haven't really been applied. (In the case of LAGs, this will be intfmgrd adding the IP address to the interface.)

This is functionally a race condition between teamd creating and then deleting the interface (due to a failure condition), teamsyncd acting too fast, and dependent applications assuming all setup is complete. This race condition is more visible on weaker systems.

Therefore, to try to prevent it, before adding an entry in STATE_DB, make sure that teamsyncd can get information from the kernel about the port channel interface, and then directly connect to teamd and make sure that the teamd is processing requests. If both of these succeed, then it can be assumed that all setup is done, and that teamd won't be exiting soon.

**Why I did it**

**How I verified it**

Perform several config reloads on a weaker hardware to make sure there are no cases of the LAG entry being added to STATE_DB, and then having teamd exit afterwards.

**Details if related**


teamsyncd/teamsync.cpp
```diff
@@ -279,19 +281,51 @@ TeamSync::TeamPortSync::TeamPortSync(const string &lagName, int ifindex,
...
+            {
+                team_free(m_team);
+                m_team = NULL;
+                teamdctl_free(m_teamdctl);
+                throw system_error(make_error_code(errc::connection_refused),
+                                   "Unable to connect to teamd");
+            }
+
+            char *response;
+            err = teamdctl_config_get_raw_direct(m_teamdctl, &response);
```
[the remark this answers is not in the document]
saiarcot895: https://github.com/jpirko/libteam/blob/master/libteamdctl/libteamdctl.c#L521-L532

teamsyncd/teamsync.cpp
```diff
@@ -12,6 +12,8 @@
 #include "warm_restart.h"
 #include "teamsync.h"
 
+#include <team.h>
+#include <teamdctl.h>
```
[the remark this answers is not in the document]
saiarcot895: This is a stylistic choice to explicitly include what I'm using in this file, so that if the header files change, this file doesn't require changing.

teamsyncd/teamsync.cpp
```diff
@@ -279,19 +281,51 @@ TeamSync::TeamPortSync::TeamPortSync(const string &lagName, int ifindex,
                                    "Unable to register port change event");
             }
 
+            struct teamdctl *m_teamdctl = teamdctl_alloc();
```
judyjoseph, line 284: If teamd exits, shouldn't we get netlink for interface down, as we registered earlier for netlink events in teamsyncd  ?

netlink.registerGroup(RTNLGRP_LINK);

saiarcot895: We will, but by the time we get the netlink event, portsorch may have already started processing the port channel interface creation.

saiarcot895:
@judyjoseph Please re-review.

teamsyncd/teamsync.cpp
```diff
@@ -279,19 +281,51 @@ TeamSync::TeamPortSync::TeamPortSync(const string &lagName, int ifindex,
...
+            if (!m_teamdctl)
+            {
+                team_free(m_team);
+                m_team = NULL;
+                throw system_error(make_error_code(errc::address_not_available),
+                                   "Unable to allocate teamdctl socket");
+            }
+
+            err = teamdctl_connect(m_teamdctl, lagName.c_str(), nullptr, "usock");
```
judyjoseph, line 293: LGTM .. so here we are adding check by connecting only in TeamPortSync constructor and not each time there is a change ( in function TeamSync::TeamPortSync::onChange())

saiarcot895: Correct, this is just to verify that teamd is fully up and running before we do anything else.

judyjoseph: One more suggestion if we want to verify if the teamd is running is to check for presence of these files, Will this help 

```
root@dev:/# ls -lrt /run/teamd/            
total 32
-rw-r--r-- 1 root root 3 Feb 27 00:09 PortChannel102.pid
srwx------ 1 root root 0 Feb 27 00:09 PortChannel102.sock
-rw-r--r-- 1 root root 3 Feb 27 00:09 PortChannel104.pid
srwx------ 1 root root 0 Feb 27 00:09 PortChannel104.sock
srwx------ 1 root root 0 Feb 27 00:09 PortChannel106.sock
```

saiarcot895: That would be implicitly required by `teamdctl_connect`. It connects via that socket file.

judyjoseph (dismissed):
LGTM

prsunny:
@judyjoseph , could you signoff?

PR #3997, opened 2025-11-14
[SmartSwitch-HA] Set pending flags back to false.

dypet:
**What I did**
Set flow_reconcile_pending, activate_role_pending, and brainsplit_recover_pending back to false after receiving the appropriate notification. For flow_reconcile_pending and activate_role_pending, this is after controller sends operation approval. For brainsplit, clearing the flag after DPU enters stable state again.

**Why I did it**
flow_reconcile_pending, activate_role_pending, and brainsplit_recover_pending were not being reset to false after being set true for the first time. 

**How I verified it**
Tested on HA Smartswitch testbed, checked activate_role_pending etc is false after sending operation approval. Tested brainsplit by shutting communication channel between DPUs and then un-shutting after both enter standalone state, causing brain split. Recovered from brainsplit by setting one DPU to admin down/dead and then back to admin up, after which the DPUs paired into active/standby again. 

**Details if related**


orchagent/dash/dashhaorch.cpp
```diff
@@ -1026,6 +1032,11 @@ void DashHaOrch::doTask(NotificationConsumer &consumer)
                             fvs.push_back({"activate_role_pending", "true"});
                             SWSS_LOG_NOTICE("DPU is pending on role activation for %s", key.c_str());
                         }
+                        else if (in(ha_scope_event[i].ha_state, {SAI_DASH_HA_STATE_ACTIVE,
+                                                                 SAI_DASH_HA_STATE_STANDBY}))
+                        {
+                            fvs.push_back({"brainsplit_recover_pending", "false"});
```
yue-fred-gao, line 1038: I am not sure what the ha_state is when it reports brain split. Have you tested brain split?

dypet: Brainsplit is reported when DPU-DPU connection recovers and both DPU are in standalone

yue-fred-gao: I see. So if they transition to active or standby, it means they have recovered from brain-split state.

PR #4002, opened 2025-11-17
Add secondary poll factor to flex counter infra

apoorv-arista:
HLD: [Configurable Port FEC FLR Poll Interval](https://github.com/sonic-net/SONiC/blob/master/doc/port_fec_flr/port_fec_flr.md)

  ## What I did

  Added secondary poll-factor support for the PORT flex-counter group. This allows FEC FLR statistics to be polled at a cadence derived
  from the normal port-counter interval.

  ## How it works

  The FEC FLR polling interval is calculated as:

  ```text
  FLR interval = port poll interval × secondary poll factor
  ```

  The default secondary poll factor is 120. With the default port polling interval of 1000 ms, FEC FLR statistics are therefore polled
  every 120 seconds.

  ## Changes

  - Initialize the PORT flex-counter group with a default secondary poll factor of 120.
  - Process SECONDARY_POLL_FACTOR updates from FLEX_COUNTER_TABLE|PORT.
  - Propagate runtime updates to both the switch and gearbox flex-counter groups.
  - Support traditional and nontraditional flex-counter implementations.
  - Validate secondary poll-factor values before forwarding them to syncd.
  - Update port_flr.lua to:
      - Accept the secondary factor as an optional fourth argument.
      - Fall back to 120 when the argument is missing, invalid, or nonpositive.
      - Calculate the FEC FLR interval using the configured port poll interval.

  - Add unit-test coverage for:
      - Default initialization.
      - Runtime factor updates.
      - Invalid and out-of-range values.
      - Traditional and nontraditional flex-counter modes.
      - Gearbox propagation.

  ## Backward compatibility

  The Lua plugin treats the secondary factor as optional and defaults to 120. This allows safe operation when used with components that
  do not yet supply the fourth plugin argument.

  Only the PORT flex-counter group processes SECONDARY_POLL_FACTOR; other counter groups are unaffected.

  ## Related changes

  - HLD: port_fec_flr.md : https://github.com/sonic-net/SONiC/blob/master/doc/port_fec_flr/port_fec_flr.md
  - sonic-swss-common: https://github.com/sonic-net/sonic-swss-common/pull/1110 — merged
  - sonic-sairedis: https://github.com/sonic-net/sonic-sairedis/pull/1699 — merged
  - sonic-utilities: https://github.com/sonic-net/sonic-utilities/pull/4126

  ## Verification

  - Verified default PORT secondary poll factor initialization.
  - Verified runtime ConfigDB updates.
  - Verified switch and gearbox propagation.
  - Verified fallback behavior for missing and invalid values.
  - Verified traditional and nontraditional flex-counter modes.
  - All affected unit tests pass.


prsunny:
@apoorv-arista , please add link to HLD to this PR

apoorv-arista:
The build pipelines for this PR are failing because they have interdependent changes.


I have updated the PR description with build order requirements

apoorv-arista:
Please Review and Merge  the following to unblock this PR and make the pipelines pass : 
- https://github.com/sonic-net/sonic-swss-common/pull/1110
- https://github.com/sonic-net/sonic-sairedis/pull/1699

StormLiangMS:
@prgeor @r12f could you help to review?

apoorv-arista:
Opened an qual issue tracking the merge of this PR : https://github.com/aristanetworks/sonic-qual.msft/issues/1338 
@prgeor @StormLiangMS

apoorv-arista:
https://github.com/sonic-net/sonic-swss/issues/4776
@prgeor @StormLiangMS

StormLiangMS:
@apoorv-arista Thanks for the update. I found two blocking functional regressions in the current head:

1. The non-gearbox `FlexCounterManager` delegating constructor declares `(enabled, secondary_poll_factor)` but passes `(secondary_poll_factor, enabled)`. For example, PFC watchdog's intended `enabled=true, factor=0` becomes `enabled=false, factor=1`. The mock expectation was changed to `disable` and `1`, which masks this regression rather than detecting it.
2. The PORT group initializes `SECONDARY_POLL_FACTOR` to `0`; sairedis supplies Lua `ARGV[4]` only when the factor is greater than zero, while `port_flr.lua` unconditionally evaluates `tonumber(ARGV[4])` and multiplies it. On a fresh system without explicit CLI configuration, this produces a nil arithmetic error instead of the HLD-required default factor of `120`.

Please correct the constructor delegation and restore the intended test expectations. The PORT path should also install or safely fall back to factor `120`, validate missing/invalid values, and include tests for default startup and runtime updates. This PR should remain blocked until sonic-sairedis#1699 also has a rollout-safe cross-version interface.

justin-wong-ce (approved):
LGTM

apoorv-arista:
> @apoorv-arista Thanks for the update. I found two blocking functional regressions in the current head:
> 
> 1. The non-gearbox `FlexCounterManager` delegating constructor declares `(enabled, secondary_poll_factor)` but passes `(secondary_poll_factor, enabled)`. For example, PFC watchdog's intended `enabled=true, factor=0` becomes `enabled=false, factor=1`. The mock expectation was changed to `disable` and `1`, which masks this regression rather than detecting it.
> 2. The PORT group initializes `SECONDARY_POLL_FACTOR` to `0`; sairedis supplies Lua `ARGV[4]` only when the factor is greater than zero, while `port_flr.lua` unconditionally evaluates `tonumber(ARGV[4])` and multiplies it. On a fresh system without explicit CLI configuration, this produces a nil arithmetic error instead of the HLD-required default factor of `120`.
> 
> Please correct the constructor delegation and restore the intended test expectations. The PORT path should also install or safely fall back to factor `120`, validate missing/invalid values, and include tests for default startup and runtime updates. This PR should remain blocked until sonic-sairedis#1699 also has a rollout-safe cross-version interface.

@StormLiangMS Thanks for the detailed review. Both regressions have been addressed in the latest revision:

  1. The `FlexCounterManager` constructor approach was removed, including the mock changes that masked the argument-order issue. The
  secondary factor is now configured through the dedicated `setFlexCounterGroupSecondaryPollFactor()` path, leaving the existing
  `enabled` behavior unchanged.

  2. The PORT group now installs the HLD-required default factor of `120` before registering the plugins. `port_flr.lua` treats `ARGV[4]`
  as optional and safely falls back to `120` when it is missing, invalid, or nonpositive.

  Validation was added for malformed and out-of-range values. Tests now cover default startup, runtime updates, invalid inputs, gearbox
  propagation, and both traditional and nontraditional flex-counter modes.

  sonic-sairedis [#1699](https://github.com/sonic-net/sonic-sairedis/pull/1699) has also been merged, and the optional Lua argument
  provides rollout compatibility.

  Could you please take another look?

justin-wong-ce (approved):
LGTM

PR #3964, opened 2025-10-30
[SmartSwitch-HA] Support peer_ip update in ha set.

dypet:
**What I did**
Supporting update to peer ip in ha set config.

**Why I did it**
In repairing process peer ip needs to be updated to the new DPU to be paired with.

**How I verified it**
Manually tested repairing process on HA topo.

**Details if related**

PR #4014, opened 2025-11-23
[DashHaOrch] Fix HA_SET PA address not updating on live changes

### LINKED ISSUE #3949: [ssw][ha] Issue found in DPU repairing: dashHaOrch missing support for live update PA address in HA SET table

Bojun-Feng:
Fix #3949

**What I did**
- Updated `DashHaOrch` to support live HA set updates.
- Added a unit test (`UpdateHaSetAttributes`) and a helper (`UpdateHaSet`) to the mock test harness.

`DashHaOrch` now detects repeat SET operations for existing HA sets, diffs the protobuf, calls `sai_dash_ha_api->set_ha_set_attribute` for changed PA parameters, and refreshes the cached metadata so `DashOrch`/`BfdOrch` see the new state.

**Why I did it**
During DPU repair, the controller updates the `HA_SET` entry to point at a new PA IP/port. Previously, `DashHaOrch` only handled the create path, so the ASIC and dependent components continued using stale values. Thus repairs are ineffective unless the operator manually deleted and recreated the HA set.

**How I verified it**
Covered the changes with the new `UpdateHaSetAttributes` mock test.

**Details if related**
I don't think this change would impact the behavior for HA scopes or ENI programming. They just read `DashHaOrch`’s cache, which now reflects live updates.

Fields like VIP/scope still require recreate semantics (unchanged from before); this patch strictly addresses PA endpoint updates.

### DIFF
--- a/orchagent/dash/dashhaorch.cpp
+++ b/orchagent/dash/dashhaorch.cpp
@@ -17,6 +17,15 @@
 using namespace std;
 using namespace swss;
 
+namespace
+{
+    template <typename T>
+    bool ProtoEquals(const T &lhs, const T &rhs)
+    {
+        return lhs.SerializeAsString() == rhs.SerializeAsString();
+    }
+}
+
 extern sai_dash_ha_api_t*   sai_dash_ha_api;
 extern sai_dash_eni_api_t*  sai_dash_eni_api;
 extern sai_object_id_t      gSwitchId;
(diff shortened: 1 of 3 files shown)


Bojun-Feng:
Not sure why `p4orch_tests` is failing - should be completely unrelated?
Repushing everything as a single commit to re-run the tests.

Bojun-Feng:
Closed as duplicate of #3964

PR #4015, opened 2025-11-23
[chassis][voq] ignore ref counting for buffer queue for voq systems

arlakshm:
**What I did**
This PR updates bufferorch logic for VOQ-based systems to:

Stop performing per-queue port reference counting when applying buffer queue configurations on VOQ switches (system ports are static and do not require dynamic ref tracking).
Correct the appl DB key transformation for VOQ buffer ready list initialization to use the standard delimiter consistently instead of mixing config_db_key_delimiter and delimiter.
**Why I did it**
To avoid ref count issues when there is port speed change

**How I verified it**

**Details if related**


orchagent/bufferorch.cpp
```diff
@@ -1130,30 +1130,34 @@ task_process_status BufferOrch::processQueuePost(const QueueTask& task)
              * so we added a map that will help us to know what was the last command for this port and priority -
              * if the last command was set command then it is a modify command and we dont need to increase the buffer counter
              * all other cases (no last command exist or del command was the last command) it means that we need to increase the ref counter */
-            if (op == SET_COMMAND)
+            /* for voq switches the buffer configuration is applied on the VOQ which is applicable on the system ports
+             * system ports are not dynamically created/deleted so no need to maintain ref counter */
+            if (gMySwitchType != "voq")
```
prsunny, line 1168: Do you want to use `isChassisDbInUse` instead of this check?

arlakshm: No, this is applicable of all the voq switch, both chassis and fixed box

arlakshm:
@ysmanman and @saksarav-nokia for viz...

vmittal-msft:
@arlakshm please rebase.

arlakshm:
@prsunny, can you please help signoff on this PR

orchagent/bufferorch.cpp
kperumalbfn, line 1168: Does VOQ switch support dynamic port breakout?

orchagent/bufferorch.cpp
```diff
@@ -1130,30 +1130,34 @@ task_process_status BufferOrch::processQueuePost(const QueueTask& task)
...
-                if (queue_port_flags[port_name][ind] == SET_COMMAND)
-		{
-                    /* we need to decrease ref counter only if the last operation was "SET_COMMAND" */
-                    gPortsOrch->decreasePortRefCount(port_name);
+                else if (op == DEL_COMMAND)
+                {
+                    if (queue_port_flags[port_name][ind] == SET_COMMAND)
+                    {
+                        /* we need to decrease ref counter only if the last operation was "SET_COMMAND" */
+                        gPortsOrch->decreasePortRefCount(port_name);
```
kperumalbfn, line 1183: If refcounts are not required for VOQ, should we bypass in processPriorityGroupPost as well?

orchagent/bufferorch.cpp
arlakshm: No we do not support dynamic port breakout

orchagent/bufferorch.cpp
arlakshm: PG is not configured on the system port, they are still on the local port. So better to keep it as-is

rlhui:
please rebase

PR #4016, opened 2025-11-23
[countersyncd]: Add benchmark suite for countersyncd and optimize otel actor

Janetxxx:
**What I did**
1. Add benchmark suite for countersyncd to evaluate the performances of actors and end to end.
2. Optimize otel actor to bulid otel message in batches.

**Why I did it**
1. Assess whether its performance meets our business requirement.
2. The performance of otel actor is too bad with small packets.

**How I verified it**
Test it locally via `cargo bench`
**Details if related**

### DIFF
--- a/gearsyncd/gearboxparser.cpp
+++ b/gearsyncd/gearboxparser.cpp
@@ -158,6 +158,12 @@ bool GearboxParser::parse()
                 attr = std::make_pair("macsec_ipg", std::to_string(val.get<int>()));
                 attrs.push_back(attr);
             }
+            if (phy.find("macsec_supported") != phy.end())
+            {
+                val = phy["macsec_supported"];
+                attr = std::make_pair("macsec_supported", val.get<bool>() ? "true" : "false");
+                attrs.push_back(attr);
+            }
             if (phy.find("hwinfo") == phy.end())
             {
                 SWSS_LOG_ERROR("missing 'hwinfo' field in 'phys' item %d in gearbox configuration", iter);
--- a/lib/gearboxutils.cpp
+++ b/lib/gearboxutils.cpp
@@ -137,6 +137,9 @@ std::map<int, gearbox_phy_t> GearboxUtils::loadPhyMap(Table *gearboxTable)
         {
             gearbox_phy_t phy = {};
 
+            // default capability if absent
+            phy.macsec_supported = true;
+
             gearboxTable->get(k, ovalues);
             for (auto &val : ovalues)
             {
@@ -192,6 +195,10 @@ std::map<int, gearbox_phy_t> GearboxUtils::loadPhyMap(Table *gearboxTable)
                 else if (val.first == "macsec_ipg")
                 {
                     phy.macsec_ipg = std::stoi(val.second);
+                }
+                else if (val.first == "macsec_supported")
+                {
+                    phy.macsec_supported = (val.second == "true");
                 }
             }
             gearboxPhyMap[phy.phy_id] = phy;
--- a/lib/gearboxutils.h
+++ b/lib/gearboxutils.h
@@ -64,6 +64,7 @@ typedef struct
     uint32_t bus_id;
     uint32_t context_id;
     uint32_t macsec_ipg;
+    bool macsec_supported;
 } gearbox_phy_t;
 
 typedef struct
--- a/orchagent/macsecorch.cpp
+++ b/orchagent/macsecorch.cpp
@@ -355,12 +355,25 @@ public:
             {
                 return nullptr;
             }
-            if (port->m_line_side_id != SAI_NULL_OBJECT_ID)
+
+            const auto *phy = get_gearbox_phy();
+            bool force_npu = true;
+            if (phy)
+            {
+                force_npu = !phy->macsec_supported;
+            }
+
+            if (!force_npu && port->m_line_side_id != SAI_NULL_OBJECT_ID)
             {
                 m_port_id = std::make_unique<sai_object_id_t>(port->m_line_side_id);
             }
             else
             {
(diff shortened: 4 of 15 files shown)


prsunny:
@Janetxxx , please fix the description per template and give details of the change

Pterosaur:
@Janetxxx Please revise the title and description

crates/countersyncd/benches/otel_actor_perf.rs
```diff
@@ -0,0 +1,193 @@
...
+    let start = std::time::Instant::now();
+
+    for tmpl in prepared.templates.iter() {
+        for msg_idx in 0..tmpl.records {
+            let msg = build_stats_message(tmpl.spec.counters, msg_idx as u64);
+            tx.send(msg).await.expect("send stats");
+        }
+    }
+
+    drop(tx); // close channel so actor exits after processing
```
r12f, line 111: should we drop the tx after getting the time?

crates/countersyncd/src/actor/ipfix.rs
```diff
@@ -749,11 +758,24 @@ impl IpfixActor {
...
                 if let ipfixrw::parser::Records::Data { set_id, data: _ } = set.records {
                     self.update_applied_template(set_id);
+                    if !self.is_template_known(set_id) {
+                        should_drop_message = true;
+                    }
                 }
             });
+
+            if should_drop_message {
+                debug!(
```
r12f, line 773: maybe this should be logged as warning.

I am also concerned on the number of messages, it is per ipfix message or per counter? if message, it might be ok, but the volume is too large per counter.

r12f: and on message level, it might be better to log a bit more info about the dropped message, so it will be easier for us to debug if any reasons from live site point of view.

crates/countersyncd/src/actor/otel.rs
```diff
@@ -86,13 +96,18 @@ impl OtelActor {
...
+        let flush_deadline = TokioInstant::now() + config.flush_timeout;
+
         Ok(OtelActor {
             stats_receiver,
             config,
             shutdown_notifier: Some(shutdown_notifier),
             client,
             resource,
             instrumentation_scope,
+            buffer: Vec::new(),
```
r12f, line 110: wow, this part of the change is better to be added into a new PR.

r12f: this is not adding perf test anymore.

crates/countersyncd/benches/end_to_end.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+}
+
+async fn run_end_to_end(prepared: PreparedDataset, endpoint: String, exports_counter: Arc<AtomicU64>) -> (Duration, usize, u64) {
+    let (template_tx, template_rx) = mpsc::channel::<IPFixTemplatesMessage>(prepared.template_messages.len() + 4);
+    let (buffer_tx, buffer_rx) = mpsc::channel::<SocketBufferMessage>(1024);
+    let (counter_tx, counter_rx) = mpsc::channel::<SAIStatsMessage>(1024);
+    let (otel_tx, otel_rx) = mpsc::channel::<SAIStatsMessage>(1024);
+    let (stats_tx, stats_rx) = mpsc::channel::<SAIStatsMessage>(1024);
+    let (otel_done_tx, otel_done_rx) = oneshot::channel();
```
r12f, line 122: do we have any log to monitor the actor queue length? this seems to be important from monitor point of view so we know how busy everything is easily.

crates/countersyncd/benches/otel_actor_perf.rs
Pterosaur: IMHO, there is not too much difference.

crates/countersyncd/src/actor/ipfix.rs
Pterosaur: It's definitely message level.
BTW, without IPFIX template, the message is just a range of meaningless binary. It means we don't know what the counters indicate to.
To the live site, if there are continuous drop messages, we can use CLI with debug level to investigate it. Because it already has the full binary hex strings.

crates/countersyncd/benches/end_to_end.rs
Pterosaur: We can add it later... In another PR.
Actually, I had a PR for that before. But I discarded and didn't maintain it because some components were refactored.
We can recover it later.

r12f (approved):
@Pterosaur  and @Janetxxx , can we add some description on the benchmark test design? like how many events, how many datapoint (timestamp + value) in a single event and etc.

prsunny (dismissed):

crates/countersyncd/benches/otel_actor_perf.rs
r12f: it does. in rust, drop is synchronous, so the message processing time is also including the channel closing time, which is not right.

crates/countersyncd/src/actor/ipfix.rs
r12f: > To the live site, if there are continuous drop messages, we can use CLI with debug level to investigate it. Because it already has the full binary hex strings.

this is exactly what I am worried about. we won't be able to know this issue until we enable the debug log. We should log this as warning with a bit more info, so we know what is happening. Since this is at message level, we should at least log the message size, so we can estimate how many messages are getting dropped.

crates/countersyncd/src/actor/otel.rs
r12f: As we discussed earlier, we need to update the PR title and description. But the title is not updated yet.

r12f:
Pick to 202412: https://github.com/Azure/sonic-swss.msft/pull/191

PR #4026, opened 2025-11-26
[fpmsyncd]: Program ECMP routes with multiple SRv6 nexthops

cscarpitta:
**What I did**

fpmsyncd was dropping ECMP routes with multiple SRv6 nexthops.

When FRR programs an ECMP route with multiple SRv6 nexthops, the FPM module encodes all nexthops in the netlink message. fpmsyncd parsed the message but returned `"Multipath SRv6 routes aren't supported"` and dropped the route, so the ECMP route was not installed in the ASIC.

This PR extends fpmsyncd to handle and install ECMP routes with multiple SRv6 nexthops.

The corresponding FPM changes to encode the required SRv6 nexthop information are included in this PR: 
https://github.com/sonic-net/sonic-buildimage/pull/28653

**Why I did it**

ECMP routes with multiple SRv6 nexthops were not being processed by fpmsyncd and were therefore not installed in the ASIC.

**How I verified it**

The SONiC VS test cases have been extended to include an ECMP route with multiple SRv6 nexthops. The extended tests verify that fpmsyncd processes the ECMP route correctly and that the route is successfully installed in the ASIC.


tests/test_srv6.py
```diff
@@ -1561,51 +1582,113 @@
...
+        tbl = swsscommon.Table(self.adb.db_connection, "ASIC_STATE:SAI_OBJECT_TYPE_NEXT_HOP_GROUP_MEMBER")
+        for nexthop_group_member_id in nexthop_group_member_ids:
+            (status, fvs) = tbl.get(nexthop_group_member_id)
+            assert status == True
+            for fv in fvs:
+                if fv[0] == "SAI_NEXT_HOP_GROUP_MEMBER_ATTR_NEXT_HOP_GROUP_ID":
+                    assert fv[1] == nexthop_group_id
+                if fv[0] == "SAI_NEXT_HOP_GROUP_MEMBER_ATTR_NEXT_HOP_ID":
+                    if fv[1] == nexthop_id_1:
+                        nexthop_group_member_id_1 = nexthop_group_member_id
```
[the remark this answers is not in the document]
cscarpitta: Fixed

tests/test_srv6.py
```diff
@@ -1561,51 +1582,113 @@
...
+                    elif fv[1] == nexthop_id_2:
+                        nexthop_group_member_id_2 = nexthop_group_member_id
```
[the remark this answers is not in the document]
cscarpitta: Fixed

tests/test_srv6.py
```diff
@@ -1665,39 +1761,218 @@
...
+        tbl = swsscommon.Table(self.adb.db_connection, "ASIC_STATE:SAI_OBJECT_TYPE_NEXT_HOP_GROUP_MEMBER")
+        for nexthop_group_member_id in nexthop_group_member_ids:
+            (status, fvs) = tbl.get(nexthop_group_member_id)
+            assert status == True
+            for fv in fvs:
+                if fv[0] == "SAI_NEXT_HOP_GROUP_MEMBER_ATTR_NEXT_HOP_GROUP_ID":
+                    assert fv[1] == nexthop_group_id
+                if fv[0] == "SAI_NEXT_HOP_GROUP_MEMBER_ATTR_NEXT_HOP_ID":
+                    if fv[1] == nexthop_id_1:
+                        nexthop_group_member_id_1 = nexthop_group_member_id
```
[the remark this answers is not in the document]
cscarpitta: Fixed

tests/test_srv6.py
```diff
@@ -1665,39 +1761,218 @@
...
+                    elif fv[1] == nexthop_id_2:
+                        nexthop_group_member_id_2 = nexthop_group_member_id
```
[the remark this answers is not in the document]
cscarpitta: Fixed

GaladrielZhao (dismissed):
Looks good to me

prsunny (commented):
@abdosi , @BYGX-wcr , would you review

ahsalam (approved):

fpmsyncd/routesync.cpp
```diff
@@ -167,6 +167,8 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
     m_nl_sock = nl_socket_alloc();
     nl_connect(m_nl_sock, NETLINK_ROUTE);
     rtnl_link_alloc_cache(m_nl_sock, AF_UNSPEC, &m_link_cache);
+
+    m_appDb = make_shared<DBConnector>("APPL_DB", 0);
```
selva-nexthop, line 290: The getSrv6SidListsFromRoute() function reads back the segment field from ROUTE_TABLE in APPL_DB to find the SID list to clean up on delete. 
But delWithWarmRestart() is called before this read-back, which means the route entry may already be gone from Redis by the time the read happens (especially over the pipeline). This silently results in vpn_sid_str being empty, SID list refcounts never being decremented, and a persistent SID list leak in SRV6_SID_LIST_TABLE.

The cleanest way would be to add a std::map<std::string, std::string> m_srv6_route_to_sidlists (mapping routeKey → sidlist_csv) populated at RTM_NEWROUTE time and consumed at RTM_DELROUTE time — exactly the same pattern as the existing m_srv6_sidlist_refcnt. This eliminates both the new DBConnector and the race condition.

fpmsyncd/routesync.cpp
```diff
@@ -1366,68 +1450,80 @@ void RouteSync::onSrv6SteerRouteMsg(struct nlmsghdr *h, int len)
...
-        string srv6SidListTableKey = vpn_sid_str;
 
         SWSS_LOG_INFO("SRV6 RouteTable del msg: %s", routeTableKeyStr.c_str());
         delWithWarmRestart(
             RouteTableFieldValueTupleWrapper{std::move(routeTableKeyStr), std::string(), isNbZmqEnabled()},
             *m_routeTable);
 
-        auto it = m_srv6_sidlist_refcnt.find(srv6SidListTableKey);
-        if (it != m_srv6_sidlist_refcnt.end())
+        getSrv6SidListsFromRoute(routeTableKeyStr, vpn_sid_str);
```
selva-nexthop: the route delete must happen after reading the SID list, not before.
```
// Read FIRST, then delete
getSrv6SidListsFromRoute(routeTableKeyStr, vpn_sid_str);
delWithWarmRestart(...);
// Now decrement refcounts
```

fpmsyncd/routesync.cpp
```diff
@@ -1366,68 +1450,80 @@ void RouteSync::onSrv6SteerRouteMsg(struct nlmsghdr *h, int len)
...
+        SWSS_LOG_INFO("Got Srv6 Sid list for route '%s' from route table: '%s'",
+                      routeTableKeyStr.c_str(), vpn_sid_str.c_str());
+
+        /* Delete SID lists from SRV6_SID_LIST_TABLE */
+        vector<string> sidlists = tokenize(vpn_sid_str, ',');
+        for (auto sidlist : sidlists)
```
selva-nexthop, line 1708: SID list key design:
Using the SID value (e.g., fc00:0:1:e000::) as the SRV6_SID_LIST_TABLE key works for sharing, but it diverges from the SONiC SRv6 HLD which uses user-defined segment names.

It also means a multi-segment SID list (e.g., fc00::1→fc00::2) has no natural single-SID key. Consider using FPM_ROUTE_ENCAP_SRV6_ENCAP_SIDLIST_NAME (already defined in dplane_fpm_sonic) to carry a stable name from FRR through to SONiC so the key is operator-visible and stable.

fpmsyncd/routesync.cpp
```diff
@@ -1366,68 +1450,80 @@ void RouteSync::onSrv6SteerRouteMsg(struct nlmsghdr *h, int len)
...
             *m_routeTable);
 
-        auto it = m_srv6_sidlist_refcnt.find(srv6SidListTableKey);
-        if (it != m_srv6_sidlist_refcnt.end())
+        getSrv6SidListsFromRoute(routeTableKeyStr, vpn_sid_str);
+        SWSS_LOG_INFO("Got Srv6 Sid list for route '%s' from route table: '%s'",
+                      routeTableKeyStr.c_str(), vpn_sid_str.c_str());
+
+        /* Delete SID lists from SRV6_SID_LIST_TABLE */
+        vector<string> sidlists = tokenize(vpn_sid_str, ',');
```
selva-nexthop, line 1707: Changing FPM_ROUTE_ENCAP_SRV6_VPN_SID to carry a variable-length SID blob — this changes the wire semantics of an existing attribute and breaks older consumers. 

FPM_ROUTE_ENCAP_SRV6_ENCAP_SIDLIST (already in the enum at value 7) was designed for exactly this purpose and carries structured segment entries. 

Keeping VPN_SID as the primary/backward-compat SID and adding ENCAP_SIDLIST as an optional extension for multi-SID routes is both safer and more aligned with the SONiC SRv6 data model.

selva-nexthop:
Thanks @cscarpitta  for the work on multipath SRv6 steer routes — this is a much-needed feature. I have a few design-level observations that I think are worth discussing before this merges, because there are some issues that could lead to correctness problems in production.

fpmsyncd/routesync.cpp
abdosi: @selva-nexthop : this map can have high memory footprint if the routescale is large.

fpmsyncd/routesync.cpp
```diff
@@ -1366,68 +1473,83 @@ void RouteSync::onSrv6SteerRouteMsg(struct nlmsghdr *h, int len)
...
-        auto it = m_srv6_sidlist_refcnt.find(srv6SidListTableKey);
-        if (it != m_srv6_sidlist_refcnt.end())
-        {
-            /* SID list already exists: just bump the refcount */
-            (it->second)++;
-            SWSS_LOG_INFO("Refcount for SID list'%s' increased to %u",
-                          srv6SidListTableKey.c_str(), it->second);
-        }
-        else
+        /* Write SID lists to SRV6_SID_LIST_TABLE */
```
selva-nexthop: When a multipath SRv6 steering route loses one nexthop, FRR sends RTM_NEWROUTE with the updated (smaller) nexthop set. The existing RTM_NEWROUTE handler only incremented refcounts for the new SID lists, but never decremented refcounts for the SID lists that belonged to the removed nexthop. This left stale entries in SRV6_SID_LIST_TABLE.

Specifically for test_AddRemoveSrv6SteeringRouteIpv4:

Route 192.0.3.0/24 is multipath: sids fc00:0:1:e000:: (refcnt=2, shared with 192.0.2.0/24) + fc00:0:3:e000:: (refcnt=1)
Removing 192.0.2.0/24 → RTM_DELROUTE → refcntfc00:0:1:e000:: = 1 ✓
Removing Ethernet104 from 192.0.3.0/24 → FRR sends RTM_NEWROUTE with only fc00:0:3:e000:: → refcntfc00:0:3:e000:: bumped to 2 (wrong!), fc00:0:1:e000:: stays at 1 (wrong!)
Final delete → refcntfc00:0:3:e000:: = 1 → both SID lists leaked, never deleted
wait_for_deleted_entry times out → test fails without teardown → sr0/Vrf13 persist → cascading failures into TestSrv6StaticSidFlavors
Fix: Before processing new SID lists on RTM_NEWROUTE, call the already-defined getSrv6SidListsFromRoute() to read the route's previous SID lists from APP_DB and decrement their refcounts. getSrv6SidListsFromRoute() was added in this commit precisely for this purpose but was never wired up.

we need to have below fix in this function.

```
      /* If this route already exists in APP_DB, it is an update rather than a new
       * route (e.g. a multipath route losing one of its nexthops). Decrement the
       * refcounts for the old SID lists so that SID lists no longer referenced by
       * any route are removed from SRV6_SID_LIST_TABLE.
       */
      string old_sidlists_str;
      if (getSrv6SidListsFromRoute(routeTableKeyStr, old_sidlists_str))
      {
          vector<string> old_sidlists = tokenize(old_sidlists_str, ',');
          for (auto& old_sidlist : old_sidlists)
          {
              auto it = m_srv6_sidlist_refcnt.find(old_sidlist);
              if (it != m_srv6_sidlist_refcnt.end())
              {
                  assert(it->second > 0);
                  (it->second)--;
                  SWSS_LOG_INFO("Route update: refcount for old SID list '%s' decreased to %u",
                                old_sidlist.c_str(), it->second);
                  if (it->second == 0)
                  {
                      m_srv6SidListTable.del(old_sidlist);
                      SWSS_LOG_INFO("Route update: SID list '%s' removed from ApplDB (refcount zero)",
                                    old_sidlist.c_str());
                      m_srv6_sidlist_refcnt.erase(it);
                  }
              }
              else
              {
                  SWSS_LOG_WARN("Route update: old SID list '%s' not found in refcount map",
                                old_sidlist.c_str());
              }
          }
      }
```

selva-nexthop:
@deepak-singhal0408  could you add this PR for routing dashboard for PR tracking.

deepak-singhal0408:
@GaladrielZhao could you relook at the change and sign-off if looks ok?

GaladrielZhao:
Looks like the coverage test is not passing in CI. Please address this.

PR #4028, opened 2025-12-01
Respect Cargo.lock for dependencies version

qiluo-msft:
**What I did**

**Why I did it**
Currently `cargo update` will update dependencies in the Cargo.lock file to the latest version on the specified branch. The sonic-swss-common Rust code is possibly not aligned with libswsscommon package, which will make sonic-buildimage build fail.

**How I verified it**
Verify local sonic-buildimage build passing, even there is sonic-swss-common changes on c-api.

**Details if related**

### DIFF
--- a/orchagent/bulker.h
+++ b/orchagent/bulker.h
@@ -368,8 +368,7 @@ struct SaiBulkerTraits<sai_next_hop_group_api_t>
     using set_entry_attribute_fn = sai_set_next_hop_group_member_attribute_fn;
     using bulk_create_entry_fn = sai_bulk_object_create_fn;
     using bulk_remove_entry_fn = sai_bulk_object_remove_fn;
-    // TODO: wait until available in SAI
-    //using bulk_set_entry_attribute_fn = sai_bulk_object_set_attribute_fn;
+    using bulk_set_entry_attribute_fn = sai_bulk_object_set_attribute_fn;
 };
 
 template<>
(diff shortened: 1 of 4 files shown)


saiarcot895:
We'd still somehow need to handle the locked version being significantly different from the swss-common version/hash being used in the pipeline (as build artifacts).

prabhataravind (approved):

yue-fred-gao:
with locked, we lose the ability to automatically upgrade dependencies to pick up fixes. Are we going to manually upgrade?

liushilongbuaa:
@qiluo-msft , How about using Cargo.lock in swss, dash-ha repo for UT. When UT passed, PR can be merged into swss repo.
When advance submodule ref in sonic-buildimage repo, we add a patch in src/sonic-swss.patch/series. The patch will change Cargo.toml file change dependency swss-common from github commit to local folder(../sonic-swss-common). In this scenario, we will use Cargo.toml instead of Cargo.lock. So, in debian/rules we can't add --lock option.

r12f:
hi @qiluo-msft , I wonder if you could help pick this PR to 202412 as well?

qiluo-msft:
> hi @qiluo-msft , I wonder if you could help pick this PR to 202412 as well?

ack

r12f:
Cherry pick here: https://github.com/Azure/sonic-swss.msft/pull/184

PR #4030, opened 2025-12-03
Revert "Avoid nhgroup update when mux state changes"

yxieca:
Reverts sonic-net/sonic-swss#3822

This change is blocking sonic-swss submodule advance for 2 weeks.

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -810,6 +810,26 @@ bool MuxNbrHandler::enable(bool update_rt)
         /* Increment ref count for new NHs */
         gNeighOrch->increaseNextHopRefCount(nh_key, num_routes);
 
+        /*
+         * Invalidate current nexthop group and update with new NH
+         * Ref count update is not required for tunnel NH IDs (nh_removed)
+         */
+        uint32_t nh_removed, nh_added;
+        if (!gRouteOrch->invalidnexthopinNextHopGroup(nh_key, nh_removed))
+        {
+            SWSS_LOG_ERROR("Removing existing NH failed for %s", nh_key.ip_address.to_string().c_str());
+            return false;
+        }
+
+        if (!gRouteOrch->validnexthopinNextHopGroup(nh_key, nh_added))
+        {
+            SWSS_LOG_ERROR("Adding NH failed for %s", nh_key.ip_address.to_string().c_str());
+            return false;
+        }
+
+        /* Increment ref count for ECMP NH members */
+        gNeighOrch->increaseNextHopRefCount(nh_key, nh_added);
+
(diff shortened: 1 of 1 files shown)


yxieca:
@manamand2020 Please chime in.

prsunny:
@Ndancejic, @zjswhhh , dualtor mock tests are failing . Can you check?

prsunny:
https://elastictest.org/scheduler/publictestplan/6920093d05c827bcee33e11a?testcase=dualtor%2Ftest_orchagent_active_tor_downstream.py&type=console

manamand2020:
> @manamand2020 Please chime in.
https://elastictest.org/scheduler/publictestplan/6920093d05c827bcee33e11a?testcase=dualtor%2Ftest_orchagent_active_tor_downstream.py&type=console

I think this seems to be happening because the neighbor remains in use inside the ECMP group due to the change in the original PR while we try to remove the entry from disableneighbors() which throws this error. We should have some DVS test to cover this. This change is fine though when we do not disable neighbors as part of no-host-route based mux neighbors.

prsunny:
> > @manamand2020 Please chime in.
> > https://elastictest.org/scheduler/publictestplan/6920093d05c827bcee33e11a?testcase=dualtor%2Ftest_orchagent_active_tor_downstream.py&type=console
> 
> I think this seems to be happening because the neighbor remains in use inside the ECMP group due to the change in the original PR while we try to remove the entry from disableneighbors() which throws this error. We should have some DVS test to cover this. This change is fine though when we do not disable neighbors as part of no-host-route based mux neighbors.

@manamand2020 , while the change is good with no-host-route, its failing currently on disableneighbors flow. May be we can introduce this as part of no-host-route change. agree?

manamand2020:
> > > @manamand2020 Please chime in.
> > > https://elastictest.org/scheduler/publictestplan/6920093d05c827bcee33e11a?testcase=dualtor%2Ftest_orchagent_active_tor_downstream.py&type=console
> > 
> > 
> > I think this seems to be happening because the neighbor remains in use inside the ECMP group due to the change in the original PR while we try to remove the entry from disableneighbors() which throws this error. We should have some DVS test to cover this. This change is fine though when we do not disable neighbors as part of no-host-route based mux neighbors.
> 
> @manamand2020 , while the change is good with no-host-route, its failing currently on disableneighbors flow. May be we can introduce this as part of no-host-route change. agree?

Agreed.

vmittal-msft:
@yxieca do we need this for 202511 ?  if not, can we remove these labels

yxieca:
@prsunny ?

prsunny:
yes, removed labels. @Ndancejic for viz and confirmation. if its needed for any issue, then a separate PR to be raised.

PR #4033, opened 2025-12-05
Disable test_Srv6MySidUNTunnelDscpMode and test_Srv6MySidUNTunnelDscpModeAmbiguity

BYGX-wcr:
**What I did**

Mark the two UTs with pytest.skip

**Why I did it**

Unblock the pipeline

**How I verified it**

**Details if related**


prsunny (commented):
please use pytest.skip.

prsunny (approved):

PR #4036, opened 2025-12-05
Cfgmgr/Teammgr changes for EVPN MH feature

jayakj:
**What I did**
Implementation of Cfgmgr/teammgr changes from EVPN VXLAN Multihoming HLD.
https://github.com/sonic-net/SONiC/blob/master/doc/vxlan/EVPN/EVPN_VxLAN_Multihoming.md#332-teammgr

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/cfgmgr/teammgr.cpp
+++ b/cfgmgr/teammgr.cpp
@@ -19,6 +19,7 @@
 #include <sys/wait.h>
 #include <sys/types.h>
 #include <signal.h>
+#include <netlink/route/link.h>
 
 
 using namespace std;
@@ -234,7 +235,6 @@ void TeamMgr::cleanTeamProcesses()
 void TeamMgr::doLagTask(Consumer &consumer)
 {
     SWSS_LOG_ENTER();
-
     auto it = consumer.m_toSync.begin();
     while (it != consumer.m_toSync.end())
     {
@@ -252,6 +252,7 @@ void TeamMgr::doLagTask(Consumer &consumer)
             string mtu = DEFAULT_MTU_STR;
             string learn_mode;
             string tpid;
+            string sys_mac;
 
             for (auto i : kfvFieldsValues(t))
             {
@@ -296,6 +297,11 @@ void TeamMgr::doLagTask(Consumer &consumer)
                     SWSS_LOG_INFO("Get fast_rate `%s`",
                                   fast_rate ? "true" : "false");
                 }
+                else if (fvField(i) == "system_mac")
+                {
+                    sys_mac = fvValue(i);
+                    SWSS_LOG_INFO("Get sys_mac %s.", sys_mac.c_str());
+                }
             }
 
             if (m_lagList.find(alias) == m_lagList.end())
@@ -322,6 +328,11 @@ void TeamMgr::doLagTask(Consumer &consumer)
             {
                 setLagTpid(alias, tpid);
                 SWSS_LOG_NOTICE("Configure %s TPID to %s", alias.c_str(), tpid.c_str());
+            }
+            if (!sys_mac.empty())
+            {
+                setLagSysmac(alias, sys_mac);
+                SWSS_LOG_NOTICE("Configure %s sys_mac to %s", alias.c_str(), sys_mac.c_str());
             }
         }
         else if (op == DEL_COMMAND)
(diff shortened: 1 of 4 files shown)


cfgmgr/teammgr.cpp
```diff
@@ -561,6 +572,92 @@ bool TeamMgr::setLagLearnMode(const string &alias, const string &learn_mode)
...
+    if (!link) {
+        SWSS_LOG_ERROR("Unable to create link");
+        nl_close(sockk);
+        nl_socket_free(sockk);
+        return -ENOMEM;
+    }
+    SWSS_LOG_NOTICE("ifindex %d, mac %02x:%02x:%02x:%02x:%02x:%02x, err %d",
+                ifindex, addr[0], addr[1], addr[2], addr[3], addr[4], addr[5], err);
+    void *mac = (void *)addr;
+    nl_addr = nl_addr_build(AF_UNSPEC, mac, ETHER_ADDR_LEN);
```
selva-nexthop, line 606: Critical:  nl_addr is built here but never freed using nl_addr_put

cfgmgr/teammgr.cpp
```diff
@@ -561,6 +572,92 @@ bool TeamMgr::setLagLearnMode(const string &alias, const string &learn_mode)
...
+        err = -ENOMEM;
+        rtnl_link_put(link);
+        nl_close(sockk);
+        nl_socket_free(sockk);
+        return err;
+    }
+
+    rtnl_link_set_addr(link, nl_addr);
+
+    if (rtnl_link_get_kernel(sockk, 0, alias.c_str(), &orig_link) < 0) {
```
selva-nexthop, line 619: orig_link obtained from rtnl_link_get_kernel() is also never released with rtnl_link_put()

cfgmgr/teammgr.cpp
```diff
@@ -561,6 +572,92 @@ bool TeamMgr::setLagLearnMode(const string &alias, const string &learn_mode)
...
+    FieldValueTuple fv("system_mac", sys_mac);
+    fvs.push_back(fv);
+    //Update in APP_DB
+    m_appLagTable.set(alias, fvs);
+    //update in kernel
+    int err = this->update_kernel(alias, sys_mac);
+    SWSS_LOG_NOTICE("kernel update returned with error %d", err);
+    //Update in State_DB
+    m_stateLagTable.set(alias, fvs);
+    return true;
```
selva-nexthop, line 658: setLagSysmac() calls update_kernel() and logs the returned error, but unconditionally returns true regardless of whether the kernel update succeeded

The fix should propagate the error, skip STATE_DB updates on failure, and return the entry to m_toSync for retry, consistent with how other TeamMgr operations like addLag() handle failures.

cfgmgr/teammgr.cpp
```diff
@@ -561,6 +572,92 @@ bool TeamMgr::setLagLearnMode(const string &alias, const string &learn_mode)
...
+{
+    vector<FieldValueTuple> fvs;
+    stringstream    cmd;
+    if (sys_mac == "None") {
+        sys_mac = m_mac.to_string();
+    }
+    FieldValueTuple fv("system_mac", sys_mac);
+    fvs.push_back(fv);
+    //Update in APP_DB
+    m_appLagTable.set(alias, fvs);
```
selva-nexthop, line 652: APP_DB and STATE_DB will be updated even when the kernel operation fails, creating database/kernel state divergence.

cfgmgr/teammgr.cpp
```diff
@@ -323,6 +329,11 @@ void TeamMgr::doLagTask(Consumer &consumer)
                 setLagTpid(alias, tpid);
                 SWSS_LOG_NOTICE("Configure %s TPID to %s", alias.c_str(), tpid.c_str());
             }
+            if (!sys_mac.empty())
+            {
+                setLagSysmac(alias, sys_mac);
+                SWSS_LOG_NOTICE("Configure %s sys_mac to %s", alias.c_str(), sys_mac.c_str());
```
selva-nexthop, line 335: What about the support for deleting the sytem_mac?  
Other attributes like MTU and TPID don't need explicit cleanup because the entire LAG entry is removed. However, if the APP_DB system_mac field has been set separately (as done in setLagSysmac()), there may be stale entries if the deletion path doesn't clean them up properly, or if the system_mac was written to tables using a different key structure.

More critically, if a LAG is deleted and re-created without a system_mac, the stale kernel MAC from the previous configuration would persist until the interface is fully recreated, since there is no code to restore the original MAC.

cfgmgr/teammgr.cpp
```diff
@@ -561,6 +572,92 @@ bool TeamMgr::setLagLearnMode(const string &alias, const string &learn_mode)
...
+    vector<FieldValueTuple> fvs;
+    stringstream    cmd;
+    if (sys_mac == "None") {
+        sys_mac = m_mac.to_string();
+    }
+    FieldValueTuple fv("system_mac", sys_mac);
+    fvs.push_back(fv);
+    //Update in APP_DB
+    m_appLagTable.set(alias, fvs);
+    //update in kernel
```
selva-nexthop, line 653: Critical: the existing TeamMgr has warm-boot awareness (it references warm_restart.h), but the new system_mac handling has zero warm-boot considerations
​
During warm-boot, the kernel MAC should already be set. Re-programming it via netlink could cause traffic disruption during the reconciliation window.

The setLagSysmac() function unconditionally writes to APP_DB and STATE_DB and programs the kernel — no check for whether the state is already correct.

cfgmgr/teammgr.h
```diff
@@ -19,6 +19,7 @@ class TeamMgr : public Orch
 
     using Orch::doTask;
     void cleanTeamProcesses();
+    bool setLagSysmac(const std::string &alias, std::string &sys_mac);
```
selva-nexthop, line 22: It should be declared private?

cfgmgr/teammgr.cpp
```diff
@@ -561,6 +572,92 @@ bool TeamMgr::setLagLearnMode(const string &alias, const string &learn_mode)
     return true;
 }
 
+int TeamMgr::update_kernel(const string &alias, const string &system_mac)
```
selva-nexthop, line 575: Naming convention.  All other methods in this class use camelCase not snake_case

cfgmgr/teammgr.cpp
```diff
@@ -561,6 +572,92 @@ bool TeamMgr::setLagLearnMode(const string &alias, const string &learn_mode)
...
+{
+    struct rtnl_link *link;
+    struct rtnl_link *orig_link;
+    int err = 0;
+    struct nl_addr *nl_addr;
+    MacAddress sys_mac(system_mac);
+    struct nl_sock * sockk = nl_socket_alloc();
```
selva-nexthop, line 582: The variable sockk (double-k) is an unusual spelling to avoid conflict with socket — a more conventional choice would be nl_sock or nlSock.

cfgmgr/teammgr.cpp
```diff
@@ -561,6 +572,92 @@ bool TeamMgr::setLagLearnMode(const string &alias, const string &learn_mode)
...
+    uint32_t ifindex = if_nametoindex(alias.c_str());
+    uint8_t *addr = const_cast<uint8_t *>(sys_mac.getMac());
+
+    if (sockk == NULL) {
+        SWSS_LOG_ERROR("Failed to allocate netlink socket.\n");
```
selva-nexthop, line 587: Log messages include trailing \n (e.g., "Failed to allocate netlink socket.\n") which is incorrect for SWSS_LOG_ERROR since the logging framework already appends newlines

cfgmgr/vxlanmgr.cpp
```diff
@@ -65,6 +65,7 @@ static int cmdCreateVxlan(const swss::VxlanMgr::VxlanInfo & info, std::string &
         cmd << " local " << shellquote(info.m_sourceIp);
     }
     cmd << " dstport 4789";
+    cmd << " udp6zerocsumrx";
```
selva-nexthop, line 68: Why this change is needed?

cfgmgr/teammgr.cpp
selva-nexthop, line 1: Critical: no unit tests or verification details available

cfgmgr/vxlanmgr.cpp
```diff
@@ -690,6 +692,12 @@ bool VxlanMgr::doVxlanEvpnNvoCreateTask(const KeyOpFieldsValuesTuple & t)
             SWSS_LOG_ERROR("NVO %s creation failed. VTEP not present",EvpnNvoName.c_str());
             return false;
         }
+        std::vector<FieldValueTuple> fv;
```
selva-nexthop, line 695: Race Condition: ProducerStateTable.set() writes data to a staging hash with an underscore prefix.  The data only lands in the final hash after orchagent's ConsumerStateTable pops and promotes it. Table.get() reads the final hash directly — so if orchagent hasn't consumed yet, the read returns empty, and NVO creation is falsely delayed.

The tunnel cache is already populated synchronously in doVxlanTunnelCreateTask() and the existing isTunnelActive() already checks it. The new Table.get() check is redundant and introduces the race.

selva-nexthop (commented):
Critical: no unit tests or verification details available

jayakj:

PR #4037, opened 2025-12-05
Neighsyncd changes for EVPN MH feature

jayakj:
**What I did**
Implementation of Neighsyncd changes from EVPN VXLAN Multihoming HLD.
https://github.com/sonic-net/SONiC/blob/master/doc/vxlan/EVPN/EVPN_VxLAN_Multihoming.md

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/neighsyncd/neighsync.cpp
+++ b/neighsyncd/neighsync.cpp
@@ -2,6 +2,7 @@
 #include <netinet/in.h>
 #include <netlink/route/link.h>
 #include <netlink/route/neighbour.h>
+#include <unistd.h>
 
 #include "logger.h"
 #include "dbconnector.h"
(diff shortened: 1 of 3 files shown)


neighsyncd/neighsync.cpp
```diff
@@ -139,9 +212,10 @@ void NeighSync::onMsg(int nlmsg_type, struct nl_object *obj)
         }
     }
     else if ((nlmsg_type == RTM_DELNEIGH) ||
-             (state == NUD_INCOMPLETE) || (state == NUD_FAILED))
+             (state == NUD_INCOMPLETE) || (state == NUD_FAILED) || (state == NUD_NOARP))
     {
-	    delete_key = true;
+        /* NUD_NOARP means this neighbor is moved to remote, we should delete the local neighbor */
```
selva-nexthop, line 217: The EVPN L3 learned neighbors will have ARP?  It will be a NO.  In this case, this change might delete all VXLAN-learned neighbors, breaking EVPN L3 forwarding entire?
The NUD_NOARP delete should only apply to neighbors that are NOT NTF_EXT_LEARNED, or use a different signal for local-to-remote moves?

neighsyncd/neighsync.cpp
```diff
@@ -185,6 +259,26 @@ void NeighSync::onMsg(int nlmsg_type, struct nl_object *obj)
...
+            hostRoute += ipStr;
+
+            SWSS_LOG_INFO("Remove host route before adding neighbor %s", hostRoute.c_str());
+            m_routeTable.del(hostRoute);
+            vector<FieldValueTuple> values;
+            while (m_routeCheckTable.get(hostRoute, values))
+            {
+                SWSS_LOG_DEBUG("Host route still exists: %s", hostRoute.c_str());
+                m_routeTable.del(hostRoute);
+				usleep(TENMS);
```
selva-nexthop, line 278: The host route deletion logic blocks the entire neighsyncd event loop.  During EVPN convergence events (many neighbors being learned simultaneously), each neighbor add for a VRF interface triggers a 10ms+ blocking poll. With 1,000 neighbors, that's 10+ seconds of blocking minimum.

We can remove the blocking loop entirely. Send a single del to the route table and proceed. If ordering matters, use orchagent's internal prioritization (routes vs. neighbors) or a dependency mechanism.

neighsyncd/neighsync.cpp
selva-nexthop, line 51: The constructor allocates
```
m_nl_sock = nl_socket_alloc();
nl_connect(m_nl_sock, NETLINK_ROUTE);
```
But, the destructor only cleans up m_AppRestartAssist — m_nl_sock and m_link_cache are never freed

jayakj:

PR #4047, opened 2025-12-08
[202505] Fix orchagent crash handling behavior

prabhataravind:
**What I did**
- Revert to old behavior for all SAI API calls in main.cpp. This implies reverting back to orchagent exiting behavior for critical failures like create switch failure

**Why I did it**
To make the behavior consistent with legacy behavior for critical failures like create switch failure and to make detection easier, given that there are no forced orchagent exits for most other cases.

**How I verified it**
By running mock tests

**Details if related**
Master PR - https://github.com/sonic-net/sonic-swss/pull/4040

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -838,7 +838,7 @@ void OrchDaemon::flush()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to flush redis pipeline %d", status);
-        handleSaiFailure(SAI_API_SWITCH, "set", status, false);
+        handleSaiFailure(SAI_API_SWITCH, "set", status, true);
     }
 
     for (auto* orch: m_orchList)
--- a/tests/mock_tests/orchdaemon_ut.cpp
+++ b/tests/mock_tests/orchdaemon_ut.cpp
@@ -86,7 +86,7 @@ namespace orchdaemon_test
     {
         orchd->enableRingBuffer();
 
-        // verify ring buffer is created  
+        // verify ring buffer is created
         EXPECT_TRUE(Executor::gRingBuffer != nullptr);
         EXPECT_TRUE(Executor::gRingBuffer == Orch::gRingBuffer);
 
@@ -169,12 +169,19 @@ namespace orchdaemon_test
 
     TEST_F(OrchDaemonTest, TestRedisFlushFailure)
     {
-        InSequence s;
 
-        EXPECT_CALL(mock_sai_switch_, set_switch_attribute( _, _)).WillOnce(Return(SAI_STATUS_FAILURE));
-        EXPECT_CALL(mock_sai_switch_, set_switch_attribute(_, _));
+        ASSERT_DEATH(
+            {
+                InSequence s;
 
-        orchd->flush();
+                EXPECT_CALL(mock_sai_switch_, set_switch_attribute(_, _))
+                .WillOnce(Return(SAI_STATUS_FAILURE));
+                EXPECT_CALL(mock_sai_switch_, set_switch_attribute(_, _));
+
+                orchd->flush();
+            },
+            ".*"
+        );
     }
 
 }


prabhataravind:
The failure in vstest needs cherry-pick of https://github.com/sonic-net/sonic-swss/pull/4010.

orchagent/orchdaemon.cpp
```diff
@@ -838,7 +838,7 @@ void OrchDaemon::flush()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to flush redis pipeline %d", status);
-        handleSaiFailure(SAI_API_SWITCH, "set", status);
+        handleSaiFailure(true);
```
prsunny: is this function called only once?

prabhataravind: no, this is periodic.

prsunny: ok, please address this part since its periodic

orchagent/saihelper.cpp
```diff
@@ -593,7 +593,7 @@ task_process_status handleSaiCreateStatus(sai_api_t api, sai_status_t status, vo
             handleSaiFailure(api, "create", status);
             break;
     }
-    return task_failed;
+    return task_need_retry;
```
lguohan: this could potentially repeatly generating a sai dump file, need to avoid that.

orchagent/saihelper.cpp
```diff
@@ -737,6 +737,30 @@ bool parseHandleSaiStatusFailure(task_process_status status)
     return true;
 }
 
+/*
+ * Optional sai failure handling to abort orchagent for some critical failures
+   like create switch failure.
+ */
+void handleSaiFailure(bool abort_on_failure)
```
lguohan: suggest to merge this function into the original function, avoid develop a new function.

lguohan:
i thought about it, we should separate this into two PR. one is change create_switch as well as other to crash behavior.

second, we should modify orchdeamon::start() function is check if there is failure, if yes, then print out error message periodically.

prsunny:
Review comment, split PR to two:
PR1:
1. Modify handleSaiFailure(sai_api_t api, string oper, sai_status_t status) to add one more attribute abort_on_failure
2. Initialization failure in syncd_apply_view, main and initSaiRedis and OrchDaemon::flush() to crash
3. Remove retry logic in handleSaiCreateStatus, Set or Remove functions and no sai dumps.

PR2: 
Log Message retries:
1. In handleSaiFailure, a new global variable introduced - gOrchUnhealthy set 'true' and gErrorString set to 'message'
2. In orchdeamon::start(), this variable is checked and logged periodically every 1 sec
3. Simulate failure and confirm log generation. 

@lguohan , @yxieca for viz

orchagent/orchdaemon.cpp
```diff
@@ -838,7 +838,7 @@ void OrchDaemon::flush()
     if (status != SAI_STATUS_SUCCESS)
     {
         SWSS_LOG_ERROR("Failed to flush redis pipeline %d", status);
-        handleSaiFailure(SAI_API_SWITCH, "set", status);
+        handleSaiFailure(SAI_API_SWITCH, "set", status, false);
```
prsunny: This must be true as well based on latest update - "Initialization failure in syncd_apply_view, main and initSaiRedis and OrchDaemon::flush() to crash"

orchagent/saihelper.cpp
prabhataravind: fixed

orchagent/saihelper.cpp
prabhataravind: fixed

orchagent/orchdaemon.cpp
prabhataravind: fixed

prabhataravind:
@prsunny please review. We might need a force merge as code coverage is not showing lines where handleSaiFailure() is called with abort_on_failure as covered as orchagent aborts and call doesn't return to the caller. This is expected.

prsunny (approved):

PR #4035, opened 2025-12-05
Revert "Orchagent changes needed to support single ASIC VOQ Fixed-System"

yxieca:
Reverts sonic-net/sonic-swss#3847

PR #4054, opened 2025-12-10
[orchagent] support single ASIC VOQ Fixed-System

saravanan-nexthop:
**What I did**
Orchagent changes needed to support single ASIC VOQ Fixed-System (#3847) was reverted via (#4035) because
vms-kvm-t2 tests were failing. 
That was because the `isChassisAppDbPresent()` was using an incorrect path to `database_config.json`
This PR brings back the original #3847 and fixes the `isChassisAppDbPresent()` to use the correct path. 

**Why I did it**
Single-asic VOQ systems do need #3847

**How I verified it**
Tests on vms-kvm-t2 and also single-asic voq systems
Verified on vms-kvm-t2 the test_bgp_fact is passing
```
bgp/test_bgp_fact.py::test_bgp_facts[vlab-t2-1-1-0] PASSED [ 25%]                                                                            
bgp/test_bgp_fact.py::test_bgp_facts[vlab-t2-1-1-1] PASSED [ 50%]                                                                            
bgp/test_bgp_fact.py::test_bgp_facts[lvlab-t2-1-2-0] PASSED [ 75%]                                                                            
bgp/test_bgp_fact.py::test_bgp_facts[vlab-t2-1-2-1] PASSED [100%]          
```

**Details if related**

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -647,7 +647,6 @@ int main(int argc, char **argv)
 
         //Connect to CHASSIS_APP_DB in redis-server in control/supervisor card as per
         //connection info in database_config.json
-        chassis_app_db = nullptr;
         if (isChassisAppDbPresent())
        	{
             gMultiAsicVoq = true;
@@ -909,12 +908,7 @@ int main(int argc, char **argv)
     }
 
     shared_ptr<OrchDaemon> orchDaemon;
-    DBConnector *chassis_db = nullptr;
-    if (chassis_app_db != nullptr)
-    {
-        chassis_db = chassis_app_db.get();
-    }
-
+    DBConnector *chassis_db = chassis_app_db.get();
     /*
      * Declare shared pointers for dpu specific databases.
      * These dpu databases exist on the npu for smartswitch.


orchagent/main.cpp
```diff
@@ -208,6 +216,18 @@ void getCfgSwitchType(DBConnector *cfgDb, string &switch_type, string &switch_su
 
 }
 
+bool isChassisAppDbPresent()
+{
+    std::ifstream file("/var/run/redis/sonic-db/database_config.json");
```
eswaran-nexthop: perhaps add a comment here that says this string is the value for DEFAULT_SONIC_DB_CONFIG_FILE but the constant is defined in sonic-swss-common and can't be used here?

orchagent/main.cpp
```diff
@@ -208,6 +216,19 @@ void getCfgSwitchType(DBConnector *cfgDb, string &switch_type, string &switch_su
 
 }
 
+bool isChassisAppDbPresent()
+{
+  //  constexpr const char *SonicDBConfig::DEFAULT_SONIC_DB_CONFIG_FILE;
```
eswaran-nexthop: nit: leftover comment?

saravanan-nexthop:
Original PR3847 reviewers: @mlok-nokia @vganesan-nokia @prsunny @arlakshm 
Please review this PR.
Thank you

.azure-pipelines/docker-sonic-vs/start.sh
```diff
@@ -61,10 +61,15 @@ else
         qos_cmd="-j /tmp/qos.json"
     fi
 
+    if [ -f /usr/share/sonic/single_asic_voq_fs/default_config.json ]; then
```
arlakshm, line 64: why do we need this change?

tests/mock_tests/mock_orchagent_main.cpp
```diff
@@ -24,3 +24,9 @@ sai_redis_communication_mode_t gRedisCommunicationMode = SAI_REDIS_COMMUNICATION
 VRFOrch *gVrfOrch;
 
 void syncd_apply_view() {}
+
+bool gMultiAsicVoq = false;
```
arlakshm, line 28: does this mean the chassis or multi asic voq will not tested?

tests/conftest.py
```diff
@@ -1911,7 +1926,13 @@ def update_dvs(log_path, new_dvs_env=[]):
...
                 dvs.destroy()
 
-            dvs = DockerVirtualSwitch(name, imgname, keeptb, new_dvs_env, log_path, max_cpu, forcedvs, buffer_model = buffer_model, enable_coverage=enable_coverage)
+            vol = {}
+            if switch_mode and switch_mode == SINGLE_ASIC_VOQ_FS:
+                cwd = os.getcwd()
+                voq_configs = cwd + "/single_asic_voq_fs"
+                vol[voq_configs] = {"bind": "/usr/share/sonic/single_asic_voq_fs", "mode": "ro"}
+
+            dvs = DockerVirtualSwitch(name, imgname, keeptb, new_dvs_env, log_path, max_cpu, forcedvs, buffer_model = buffer_model, enable_coverage=enable_coverage, ctnmounts=vol, switch_mode=switch_mode)
```
arlakshm, line 1935: how is the switch_mode passed?

.azure-pipelines/docker-sonic-vs/start.sh
saravanan-nexthop: This is a UT for single asic voq. That will require a new config file. This is doing that.

tests/mock_tests/mock_orchagent_main.cpp
saravanan-nexthop: Any existing mock test for chassis/mutli-asic will flip it to true.

tests/conftest.py
saravanan-nexthop: `switch-mode=single_asic_voq_fs` is passed from .azure-pipelines/test-docker-sonic-vs-template.yml

tests/mock_tests/mock_orchagent_main.cpp
arlakshm: this is handled in the current tests or we need to update the existing tests?

eswaran-nexthop: Looking at the mock test infra code, the switch type is set to "switch" and not "voq" anywhere. This makes me think voq mode is not covered in mock_tests at all. Hardcoding gMultiAsicVoq to false here is consistent with that. Please correct me if my understanding is incorrect. 

But it makes sense to open a new issue and fix the mock tests to cover the voq case because many UTs seem to have checks for voq mode. Perhaps this was running earlier but doesn't run anymore after some changes?

saravanan-nexthop: @arlakshm Please let us know if we can do the mock-test coverage for single vs multi-asic voq as a separate task ?
And get this PR to merge.

arlakshm: Looks like we do not use the mock infra for our existing chassis tests. so we should be ok here

arlakshm (approved):

arlakshm (approved):

orchagent/main.cpp
```diff
@@ -627,7 +647,19 @@ int main(int argc, char **argv)
...
+        if (isChassisAppDbPresent())
+       	{
+            gMultiAsicVoq = true;
+            try
+            {
+                chassis_app_db = make_shared<DBConnector>("CHASSIS_APP_DB", 0, true);
+            }
+            catch (const std::exception& e)
+            {
+                SWSS_LOG_NOTICE("CHASSIS_APP_DB not available, operating in standalone VOQ mode");
```
vganesan-nokia, line 659: Should not the **_gMultiAsicVoq_** be reset to **_false_** when this exception occurs? or set the **_gMultiAsicVoq_** to **_true_** only if the connection to CHASSIS_APP_DB is successful.

orchagent/main.cpp
```diff
@@ -627,7 +647,19 @@ int main(int argc, char **argv)
 
         //Connect to CHASSIS_APP_DB in redis-server in control/supervisor card as per
         //connection info in database_config.json
-        chassis_app_db = make_shared<DBConnector>("CHASSIS_APP_DB", 0, true);
+        chassis_app_db = nullptr;
```
vganesan-nokia: This initialization is redundant. Recommend removing this line.

orchagent/main.cpp
```diff
@@ -877,6 +909,11 @@ int main(int argc, char **argv)
     }
 
     shared_ptr<OrchDaemon> orchDaemon;
+    DBConnector *chassis_db = nullptr;
+    if (chassis_app_db != nullptr)
+    {
+        chassis_db = chassis_app_db.get();
+    }
```
vganesan-nokia: All these 5 lines can be safely replaced by

**_DBConnector *chassis_db = chassis_app_db.get()_**

Without any change in what is already achieved by these lines.

orchagent/neighorch.cpp
vganesan-nokia: Since we do not configure inband interface for single asic voq fixed systems, we should check **_isChassisDbInUse()_** here instead of **_gMySwitchType == "voq"_**. If the switch type is "voq" (which is true for single asic voq fixed systems also) and if inband interface is not configured, this may generate assert.

Recommend to avoid using **_gMySwitchType == "voq"_** check in all places where Inband interface info is accessed or created and instead use **_isChassisDbInUse()_**

.azure-pipelines/docker-sonic-vs/start.sh
```diff
@@ -61,10 +61,15 @@ else
...
+        sonic-cfggen -j /usr/share/sonic/single_asic_voq_fs/default_config.json --print-data > /tmp/voq.json
+        voq_cmd="-j /tmp/voq.json"
+    fi
+
```
vganesan-nokia, line 68: Would we need similar changes in platform/vs/docker-sonic-vs/start.sh also?

saravanan-nexthop: conftest.py changes is enough to make this work inside the docker-sonic-vs when `switch_mode == SINGLE_ASIC_VOQ_FS`

orchagent/main.cpp
saravanan-nexthop: `isChassisAppDbPresent()` being true means that `CHASSIS_APP_DB` needs to be running as per the database_config.json. So we want `gMultiAsicVoq` to be true as well. 
So the 2nd half of the LOG_NOTICE ("operating in standalone VOQ mode") is not required now. I will remove that.

orchagent/neighorch.cpp
saravanan-nexthop: seems like all inband related logic checks for `m_intfsOrch->isRemoteSystemPortIntf(nexthop.alias)`
Only this place doesn't have it.
@arlakshm do you think its worthy replacing the below check with `isRemoteSystemPortIntf(nexthop.alias)` ?
```
    if (gMySwitchType == "voq")                                                                                                             
    {
        gPortsOrch->getInbandPort(inbp);                                                                                                    
        assert(inbp.m_alias.length());
    }
    ```

orchagent/main.cpp
saravanan-nexthop: removed the try catch as that is how it was before.

arlakshm:
@saravanan-nexthop, can you please check the vs test failure

saravanan-nexthop:
> @saravanan-nexthop, can you please check the vs test failure

@arlakshm All checks are now passing

PR #4056, opened 2025-12-11
[portsorch] DOM config change causes interface link to flap

mihirpat1:
**What I did**
Fix issue where change in DOM configuration triggers unexpected link flap due to FEC reconfiguration, even when speed or FEC settings haven't changed.

**Why I did it**
There are 2 issues being fixed.
1. Update to DOM config causes gearbox speed to be reconfigured which eventually causes a link flap for platforms which have gearbox. This happens due to 
https://github.com/sonic-net/sonic-swss/blob/a7198d1a8ad30441bb3cf12cb94d9147a76c71c6/orchagent/portsorch.cpp#L4795-L4799

2. **Forced FEC config causes a link flap** - This issue was introduced by [PR #3529](https://github.com/sonic-net/sonic-swss/pull/3529) which refactored port configuration handling. The refactor inadvertently removed the logic to set `p.m_fec_cfg = true` after applying FEC configuration at runtime (in `doPortTask()`), while it was still being set during initial port creation (`addPortBulk()`).
On certain platforms, `addPortBuld` function is not called and hence, this causes the FEC configuration to be forcefully reapplied as part of handling DOM config change event since `p.m_fec_cfg` remains False.
https://github.com/sonic-net/sonic-swss/blob/a7198d1a8ad30441bb3cf12cb94d9147a76c71c6/orchagent/portsorch.cpp#L5034

On such platforms, the below is never called
https://github.com/sonic-net/sonic-swss/blob/a7198d1a8ad30441bb3cf12cb94d9147a76c71c6/orchagent/portsorch.cpp#L1253

**How I verified it**
- Verified that after applying FEC configuration at runtime, `m_fec_cfg` is properly set to `true`
- Confirmed that subsequent DOM configuration updates no longer trigger link flap

**Details if related**
MSFT ADO - 35980722

### DIFF
--- a/.azure-pipelines/docker-sonic-vs/start.sh
+++ b/.azure-pipelines/docker-sonic-vs/start.sh
@@ -61,10 +61,15 @@ else
         qos_cmd="-j /tmp/qos.json"
     fi
 
+    if [ -f /usr/share/sonic/single_asic_voq_fs/default_config.json ]; then
+        sonic-cfggen -j /usr/share/sonic/single_asic_voq_fs/default_config.json --print-data > /tmp/voq.json
+        voq_cmd="-j /tmp/voq.json"
+    fi
+
     sonic-cfggen -p /usr/share/sonic/device/$PLATFORM/$PLATFORM_CONF -k $HWSKU --print-data > /tmp/ports.json
     # change admin_status from up to down; Test cases dependent
     sed -i "s/up/down/g" /tmp/ports.json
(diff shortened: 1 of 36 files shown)


stepanblyschak (approved):

arpit-nexthop (approved):

longhuan-cisco (approved):

mihirpat1:
@prsunny Can you please help to merge this PR?

bingwang-ms:
Removed cherry-pick request to 202503 per offline discussion

PR #4058, opened 2025-12-12
Opportunistically use a matching kernel version for building the modules

saiarcot895:
**What I did**

When building the team kernel module (and others), if a matching source kernel version exists, then make sure we use that exact version. This avoids the case where we end up pulling the sources for a newer kernel version when a matching one exists.

**Why I did it**

**How I verified it**

**Details if related**


vmittal-msft:
@yxieca  @rlhui  please help merge

PR #4062, opened 2025-12-13
[fpmsyncd]: Handle SRv6 SID Lists with multiple SIDs

cscarpitta:
**What I did**

There is a bug in fpmsyncd related to the processing of SRv6 SID lists containing multiple SIDs.

When FRR attempts to program a SID list with multiple SRv6 SIDs, the FPM module correctly encodes all SIDs in the Netlink message. However, upon receiving this message, fpmsyncd processes only the first SID and installs only that SID into the ASIC.

As a result, the ASIC is programmed with an incomplete and incorrect SID list.

This pull request updates fpmsyncd to correctly handle and install SID lists containing multiple SRv6 SIDs.

**Why I did it**

Currently the ASIC is programmed with an incorrect SID list.

**How I verified it**

Added two test cases in `tests/test_srv6.py`

### DIFF
--- a/orchagent/p4orch/gre_tunnel_manager.cpp
+++ b/orchagent/p4orch/gre_tunnel_manager.cpp
@@ -104,15 +104,13 @@ ReturnCode GreTunnelManager::validateGreTunnelAppDbEntry(
 }
 
 ReturnCode GreTunnelManager::validateGreTunnelAppDbEntry(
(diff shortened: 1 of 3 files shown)


GaladrielZhao:
The code looks good to me, please fix the two new added tests.
- test_AddRemoveSrv6SteeringRouteIpv4MultipleSids
- test_AddRemoveSrv6SteeringRouteIpv6MultipleSids

PR #4072, opened 2025-12-18
[countersyncd]: Fix compiling warning of otel

Pterosaur:
**What I did**
Remove unused code to fix the compiling warning

**Why I did it**
Fix the compiling warning

**How I verified it**
Check the unit test

**Details if related**

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -91,7 +91,7 @@ jobs:
       vmImage: 'ubuntu-22.04'
 
   container:
-    image: sonicdev-microsoft.azurecr.io:443/${{ parameters.sonic_slave }}:latest
+    image: sonicdev-microsoft.azurecr.io:443/${{ parameters.sonic_slave }}:$(BUILD_BRANCH)
 
   steps:
   - checkout: self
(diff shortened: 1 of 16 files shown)


crates/countersyncd/src/main.rs
```diff
@@ -227,7 +227,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let (stats_report_sender, stats_report_receiver) = channel(args.stats_reporter_capacity);
     let (counter_db_sender, counter_db_receiver) = channel(args.counter_db_capacity);
     let (otel_sender, otel_receiver) = channel(args.otel_capacity);
-    let (otel_shutdown_sender, otel_shutdown_receiver) = tokio::sync::oneshot::channel();
+    let (otel_shutdown_sender, _otel_shutdown_receiver) = tokio::sync::oneshot::channel();
```
r12f, line 230: how is the shutdown being handled without receiver?

crates/countersyncd/src/actor/otel.rs
```diff
@@ -235,47 +234,6 @@ impl OtelActor {
         }
     }
 
-    pub fn print_conversion_report(sai_stats: &SAIStats, otel_metrics: &OtelMetrics) {
```
r12f, line 238: this report looks to be helpful, maybe we should keep it and do it in verbose level or under a debug log flag or something?

Pterosaur: This function is never used, and we already have other similar code for debugging information: https://github.com/Pterosaur/sonic-swss/blob/23842075ac49a4f739201d31d642ccfd40ec5f34/crates/countersyncd/src/actor/otel.rs#L144C14-L144C32

So, it's redundant.

crates/countersyncd/src/main.rs
Pterosaur: This channel is only used for testing code. To the regular workflow, actors will exit via deleting core communication channels.

prsunny (approved):

PR #4074, opened 2025-12-18
Use the correct slave container for the branch

saiarcot895:
**What I did**

Use the slave container that matches the branch that we are building for, to make sure the expected build environment is used.

**Why I did it**

**How I verified it**

**Details if related**


saiarcot895:
Included in 202505: https://github.com/sonic-net/sonic-swss/pull/4071

prsunny:
@vmittal-msft for viz

PR #4079, opened 2025-12-20
[RouteOrch] fix performance regression in ResponsePublisher

stepanblyschak:
**What I did**

- Fix performance regression caused by https://github.com/sonic-net/sonic-swss/pull/3066 : `setBuffered` wasn't using Redis pipeline correctly.
- Add a flag to directly write attributes passed by the caller of `publish` API avoiding Redis `HGET` which breaks Redis pipeline.

**Why I did it**

Fix performance degradation observed in BGP suppress test.

**How I verified it**

- t1-lag BGP route test. Starting BGP sessions `config bgp startup all` and profile.
- BGP suppress FIB sonic-mgmt test

Without this change:

image

With this change:

image

**Details if related**


stepanblyschak:
@dgsudharsan to take over this PR

PR #4103, opened 2026-01-02
Fix debug flexcounter uninstall to skip non-PHY ports

manish1-arista:
**What I did**
Updated uninstallDebugFlexCounters() to ensure stopFlexCounterPolling() is invoked only after verifying that the port type is Port::Type::PHY

**Why I did it**
To resolve a failure during the teardown of test_configurable_drop_counters.py where the ```config dropcounters delete``` command triggered a logAnalyzerError.
The system was logging an error: ```ERR syncd#syncd:: removeCounter: Object type for removal not supported, SAI_OBJECT_TYPE_NULL```.
It can easily be replicated by running these commands:
```{bash}
# Step 1: Install a debug counter 
sudo config dropcounters install TEST PORT_INGRESS_DROPS [L3_EGRESS_LINK_DOWN]

# Step 2: Delete the debug counter
sudo config dropcounters delete TEST

# Step 3: Check syslog for the error
sudo grep "Object type for removal not supported" /var/log/syslog
```
uninstallDebugFlexCounters() was attempting to stop polling for all ports including non-PHY ports before validating the port type, causing invalid DEL operations.

Fixes: 
https://github.com/aristanetworks/sonic-qual.msft/issues/1011
https://github.com/sonic-net/sonic-swss/issues/4229

**How I verified it**
Verified the fix by replicating the issue steps: 
Installing a debug counter (sudo config dropcounters install) and deleting it (sudo config dropcounters delete).
Tested the review image on a cluster and confirmed that the error is no longer generated.

**Details if related**

### DIFF
--- a/orchagent/p4orch/gre_tunnel_manager.cpp
+++ b/orchagent/p4orch/gre_tunnel_manager.cpp
@@ -104,15 +104,13 @@ ReturnCode GreTunnelManager::validateGreTunnelAppDbEntry(
 }
 
 ReturnCode GreTunnelManager::validateGreTunnelAppDbEntry(
-    const P4GreTunnelAppDbEntry& app_db_entry,
-    const std::string& operation) {
+    const P4GreTunnelAppDbEntry& app_db_entry, const std::string& operation) {
   SWSS_LOG_ENTER();
 
-  P4GreTunnelEntry entry = P4GreTunnelEntry(app_db_entry.tunnel_id,
-                                            app_db_entry.router_interface_id,
-                                            app_db_entry.encap_src_ip,
-                                            app_db_entry.encap_dst_ip,
-                                            app_db_entry.encap_dst_ip);
+  P4GreTunnelEntry entry =
+      P4GreTunnelEntry(app_db_entry.tunnel_id, app_db_entry.router_interface_id,
+                       app_db_entry.encap_src_ip, app_db_entry.encap_dst_ip,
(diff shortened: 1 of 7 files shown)


StormLiangMS:
@prsunny could you help to take a look?

StormLiangMS:
LGTM. The fix moves the `stopFlexCounterPolling` call below the port_id/counter_type filter checks, ensuring it only stops polling for the ports that actually had the debug counter installed. Previously it was unconditionally stopping flex counter polling for ALL ports before the filter, which caused `SAI_OBJECT_TYPE_NULL` errors when removing counters from ports that never had them. Clean surgical fix. CI is fully green. Ready to merge.

StormLiangMS:
@prsunny could you help to sign off?

manish1-arista:
@prsunny Needs to be cherry picked for 202511.
(Qual issue for reference https://github.com/aristanetworks/sonic-qual.msft/issues/1011)

StormLiangMS:
This PR is merged to master but needs a 202511 cherry-pick PR. @vmittal-msft could you help create the 202511 backport? Needed for Arista 202511 qualification (aristanetworks/sonic-qual.msft#1011).

StormLiangMS:
@vmittal-msft This PR is merged to master and related to aristanetworks/sonic-qual.msft#1011. Could you check if this is eligible for 202511 backport?

PR #4113, opened 2026-01-07
orchagent: Add support to program gearbox SI settings from APPL_DB

bgallagher-nexthop:
See [this PR](https://github.com/sonic-net/SONiC/pull/2174) for the HLD describing the high-level design behind this change.

**What I did**

This commit adds logic to program gearbox signal integrity settings published to APPL_DB by xcvrd via SAI in the doPortTask() hook in orchagent. The old code path that was previously responsible for programming gearbox signal integrity settings is left intact for backwards compatibility.

**Why I did it**

This is one PR in a series that adds support for configuring signal integrity settings for gearboxes via media_settings.json. This consolidates configuration of signal integrity settings into one place (gearbox or no gearbox) and provides a more flexible configuration mechanism than was previously available (it is now possible to configure different values for different physical media upfront).

**How I verified it**

Wrote unit-tests, which pass:
```
bgallagher@fe605fa931ca:/sonic/src/sonic-swss/tests/mock_tests$ ./tests --gtest_filter=PortsOrchTest.PortGearbox*
Running main() from ./googletest/src/gtest_main.cc
Note: Google Test filter = PortsOrchTest.PortGearbox*
[==========] Running 3 tests from 1 test suite.
[----------] 3 tests from PortsOrchTest
[ RUN      ] PortsOrchTest.PortGearboxSerdesConfig
[       OK ] PortsOrchTest.PortGearboxSerdesConfig (27 ms)
[ RUN      ] PortsOrchTest.PortGearboxSerdesConfigFailure
[       OK ] PortsOrchTest.PortGearboxSerdesConfigFailure (18 ms)
[ RUN      ] PortsOrchTest.PortGearboxSerdesConfigNoGearbox
[       OK ] PortsOrchTest.PortGearboxSerdesConfigNoGearbox (18 ms)
[----------] 3 tests from PortsOrchTest (64 ms total)

[==========] 3 tests from 1 test suite ran. (75 ms total)
[  PASSED  ] 3 tests.
```

- Installed [this media_settings.json](https://gist.github.com/bgallagher-nexthop/b4eaeb34bdcdfb09a51c66062098b2b7) file onto a nexthop network device.
- `docker exec`’d into the gbsyncd container and used the agera2sh CLI to verify that the hardware was successfully programmed with the SI settings in APPL_DB. Verified the gearbox hardware was programmed successfully with the appropriate values: https://gist.github.com/bgallagher-nexthop/dd89fd72b22735205a8dcf19c41d1801

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [x] 202511

I'd like to backport this change to 202511 in order to remove the need to support two different approaches for configuring gearbox TX taps across releases for Nexthop hardware.


arlakshm:
@prgeor @kenneth-arista can you please review this PR

arlakshm (dismissed):

orchagent/portsorch.cpp
```diff
@@ -561,6 +561,86 @@ static void getPortSerdesAttr(PortSerdesAttrMap_t &map, const PortConfig &port)
     }
 }
 
+static void getLinePortSerdesAttr(PortSerdesAttrMap_t &map, const PortConfig &port)
```
prgeor: @bgallagher-nexthop 
```suggestion
static void getGbPortLineSerdesAttr(PortSerdesAttrMap_t &map, const PortConfig &port)
```

orchagent/portsorch.cpp
```diff
@@ -561,6 +561,86 @@ static void getPortSerdesAttr(PortSerdesAttrMap_t &map, const PortConfig &port)
...
+    }
+
+    if (port.serdes.gb_line_post3.is_set)
+    {
+        map[SAI_PORT_SERDES_ATTR_TX_FIR_POST3] = port.serdes.gb_line_post3.value;
+    }
+
+}
+
+static void getSystemPortSerdesAttr(PortSerdesAttrMap_t &map, const PortConfig &port)
```
prgeor:
```suggestion
static void getGbPortSystemSerdesAttr(PortSerdesAttrMap_t &map, const PortConfig &port)
```

orchagent/port/portschema.h
```diff
@@ -82,6 +82,20 @@
 #define PORT_POST1                 "post1"
 #define PORT_POST2                 "post2"
 #define PORT_POST3                 "post3"
+#define PORT_GB_LINE_PRE1          "gb_line_pre1"
```
prgeor: @bgallagher-nexthop is there any specific reason why we cannot reuse the existing parameters "pre1", "pre2","main","post1", "post2" for system side?

orchagent/portsorch.cpp
arpit-nexthop: ACK

orchagent/portsorch.cpp
arpit-nexthop: ACK

orchagent/port/portschema.h
arpit-nexthop: @prgeor 
when we have a gearbox, we have 3 ports essentially:

ASIC port --> the older parameters work here
GB system port --> Going towards the ASIC
GB line port --> towards module

All 3 of these ports have independent settings and we need to mark it as such

prgeor: @arpit-nexthop I understand the difference, but the serdes SI setting SAI attribute is the same for ASIC port and GB system port.

rlhui:
@arista-hpandya please help review

orchagent/port/portschema.h
arpit-nexthop: @prgeor ACK, we will revisit and update this segment.
We can reuse the existing macros and add the extra gb prefix during the parse call

arpit-nexthop: @prgeor Updated the PR to remove duplicate entries in data structure and parsing. Each side has its own serdes object now

prgeor (approved):

prsunny:
@bgallagher-nexthop , could you check on the copilot comment and see if it must be addressed? We can raise a separate PR if needed

bgallagher-nexthop:
> @bgallagher-nexthop , could you check on the copilot comment and see if it must be addressed? We can raise a separate PR if needed

The suggested changes make sense. They are not high priority to merge (there are no correctness issues, just a minor understandability improvement + a test improvement), I will update in a separate PR

bgallagher-nexthop:
I have opened https://github.com/sonic-net/sonic-swss/pull/4285 to address the review comments from Copilot (cc @prsunny )

PR #4114, opened 2026-01-09
[ssw][ha][202511] cherry-pick `custom_bfd` changes

zjswhhh:
Cherry-pick: 
```
85412005 [vnetorch] missing handling of rx and tx interval of monitoring session  (#3878)
b88a5c20 [ssw][ha] introducing `custom_bfd` and BfdMonitorOrch to support primary/secondary NHG switchover (#3922)
```

What I did
Today vnetorch only supports:

custom monitoring sessions with primary & secondary NHG. bfd monitoring with primary NHG only.
To make HA work we need bfd monitoring to support primary+secondary. This PR is to introduce a new monitoring type custom_bfd for that purpose. It will:

create bfd sessions for both primary & secondary endpoints listen on state db for bfd session state update
ignore update call from bfdorch
switch and update nhg based on bfd session state
It's not an elegant design to introduce a new bfd monitoring type. But the intention is to not break current functionality of vnetorch. Refactor of the code will be needed in future.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/orchdaemon.cpp
+++ b/orchagent/orchdaemon.cpp
@@ -63,6 +63,7 @@ Srv6Orch *gSrv6Orch;
 FlowCounterRouteOrch *gFlowCounterRouteOrch;
 DebugCounterOrch *gDebugCounterOrch;
 MonitorOrch *gMonitorOrch;
+BfdMonitorOrch *gBfdMonitorOrch;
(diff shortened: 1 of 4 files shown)


yue-fred-gao (approved):

zjswhhh:
Hi @vmittal-msft - can you help merge this?

PR #4115, opened 2026-01-09
cherry-pick-202511: Disable bulk requests for PORT counters

rajkumar1-arista:
Add temporary fix for https://github.com/aristanetworks/sonic-qual.msft/issues/655, https://github.com/aristanetworks/sonic-qual.msft/issues/1031

This forces each port to be processed individually, avoiding capability mismatch between different ports in bulk requests
This is cherry-pick of https://github.com/sonic-net/sonic-swss/pull/3843 to 202511

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/flex_counter/flex_counter_manager.h
+++ b/orchagent/flex_counter/flex_counter_manager.h
@@ -194,14 +194,28 @@ struct CachedObjects
             auto counter_ids = FlexCounterManager::serializeCounterStats(counter_stats);
             auto counter_type_it = FlexCounterManager::counter_id_field_lookup.find(counter_type);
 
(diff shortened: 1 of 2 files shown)


prsunny:
I dont see a master PR but only cherry-pick to 202505/20511. This is not correct. Please raise master PR and we can mark labels for cherry-pick. @StormLiangMS  for viz

rajkumar1-arista:
@prsunny This is just a workaround, Actual fix is still work in progress. More info here: https://github.com/sonic-net/sonic-swss/pull/3843#issuecomment-3244624775

PR #4116, opened 2026-01-09
[P4Orch] Revert - Bulk support for wcmp group members & Add New WCMP group operations.

divyagayathri-hcl:
**What I did**
Added New WCMP group operations.

**Why I did it**
To enable WCMP group operations.
 
**How I verified it**
Verified with UT cases.

**Details if related**


divyagayathri-hcl:
> There are changes in tunnel manager, neighbor manager, and route manager are not included. Although they are just trial function name changes.

These changes are already taken in some other PRs.

PR #4117, opened 2026-01-09
[P4Orch] Update acl binding init.

divyagayathri-hcl:
What I did
Update acl binding init
The core of this PR involves shifting the responsibility of creating and managing default ACL Groups (Ingress, Egress, and Pre-ingress) from P4AclTableManager to SwitchOrch.

Why I did it
ACL Groups bound to the switch (Ingress/Egress/Pre-Ingress) are global switch resources. They should be managed by the SwitchOrch (which handles the SAI Switch Object) .

How I verified it
Verified with UT cases.

Details if related

### DIFF
--- a/orchagent/switchorch.cpp
+++ b/orchagent/switchorch.cpp
@@ -5,6 +5,7 @@
 
 #include "switchorch.h"
 #include "crmorch.h"
+#include "p4orch/p4orch.h"
 #include "converter.h"
 #include "notifier.h"
 #include "notificationproducer.h"
@@ -24,6 +25,7 @@ extern sai_acl_api_t *sai_acl_api;
 extern sai_hash_api_t *sai_hash_api;
 extern MacAddress gVxlanMacAddress;
 extern CrmOrch *gCrmOrch;
+extern P4Orch *gP4Orch;
 extern event_handle_t g_events_handle;
 extern string gMyAsicName;
 
@@ -172,6 +174,10 @@ SwitchOrch::SwitchOrch(DBConnector *db, vector<TableConnector>& connectors, Tabl
 
     auto executorT = new ExecutableTimer(m_sensorsPollerTimer, this, "ASIC_SENSORS_POLL_TIMER");
     Orch::addExecutor(executorT);
+    if (gP4Orch)
+    {
+        initAclGroupsBindToSwitch();
+    }
 }
 
 void SwitchOrch::generateSwitchCounterNameMap() const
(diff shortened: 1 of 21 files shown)


orchagent/vrforch.cpp
```diff
@@ -237,9 +237,9 @@ bool VRFOrch::updateVrfVNIMap(const std::string& vrf_name, uint32_t vni)
             {
                 /*call VE UP*/
                 error = gPortsOrch->updateL3VniStatus(vlan_id, true);
-                SWSS_LOG_INFO("addL3VniStatus vni %d vlan %d, status %d", vni, vlan_id, error);
+		SWSS_LOG_INFO("addL3VniStatus vni %d vlan %d, status %d", vni, vlan_id, error);
```
mint570: There is no change in this file. Undo it.

orchagent/switchorch.cpp
```diff
@@ -170,6 +170,7 @@ SwitchOrch::SwitchOrch(DBConnector *db, vector<TableConnector>& connectors, Tabl
 
     auto executorT = new ExecutableTimer(m_sensorsPollerTimer, this, "ASIC_SENSORS_POLL_TIMER");
     Orch::addExecutor(executorT);
+    initAclGroupsBindToSwitch();
```
mint570: This PR will need community review as it changes the orchagent ACL init.

orchagent/portsorch.cpp
```diff
@@ -919,8 +919,8 @@ PortsOrch::PortsOrch(DBConnector *db, DBConnector *stateDb, vector<table_name_wi
         }
     }
 
-    if (gSwitchOrch->querySwitchCapability(SAI_OBJECT_TYPE_HOSTIF, SAI_HOSTIF_ATTR_QUEUE))
```
mint570, line 933: Undo the change in this file

mint570:
@prsunny can you take a look at this PR to see how this can impact upstream SONiC?

mint570 (dismissed):

kishanps:
@prsunny Can you review this PR?
cc: @mint570

orchagent/switchorch.cpp
prsunny: What is this change for? @bingwang-ms to review

prsunny (changes requested):

orchagent/switchorch.cpp
bingwang-ms: This change looks risky to me. The behavior of switchorch is changed no matter p4orch is enabled or not.

mint570: Please add a check to only perform this switch binding when p4orch is enabled.

orchagent/switchorch.cpp
```diff
@@ -172,6 +174,10 @@ SwitchOrch::SwitchOrch(DBConnector *db, vector<TableConnector>& connectors, Tabl
 
     auto executorT = new ExecutableTimer(m_sensorsPollerTimer, this, "ASIC_SENSORS_POLL_TIMER");
     Orch::addExecutor(executorT);
+    if (gP4Orch)
```
prsunny, line 177: isnt this always true?

orchagent/switchorch.cpp
```diff
@@ -172,6 +174,10 @@ SwitchOrch::SwitchOrch(DBConnector *db, vector<TableConnector>& connectors, Tabl
...
+    {
+        initAclGroupsBindToSwitch();
```
bingwang-ms, line 179: Can you help me understand why we need this change? What if the device does not use ACL at all?
If this is necessary, can you move it to p4orch so that it's still self-contained?

PR #4119, opened 2026-01-09
Add metric field to vnet route tunnel request

Leyza:
**What I did**
Add metric to vnet route tunnel request, so vnet route tunnels can be created with a metric value.

Related yang model change: https://github.com/sonic-net/sonic-buildimage/pull/25019

**Why I did it**
Enable vnet route tunnels to be created with a metric value that can be used to identify type of route.

**How I verified it**
Tested with UT and on physical testbed.

**Details if related**


anish-n (approved):

prsunny:
Please link HLD

prabhataravind:
@vmittal-msft it looks like this PR didn't get cherry-picked to 202511 automatically. @prsunny for viz.

PR #3509, opened 2025-02-10
[Feature][debugcounterorch] Add support for configurable debug drop monitoring feature

arista-hpandya:
**Note: This change depends on https://github.com/sonic-net/sonic-swss-common/pull/971**
Fixes #3501 
HLD:  https://github.com/sonic-net/SONiC/pull/1912

**What I did**
- Added logic to read configuration from the DEBUG_DROP_MONITOR table.
- Added logic to register persistent alerts when the conditions are met.
- Added logic to toggle the feature off if desired on a per-counter level.

**Why I did it**
To implement the persistent drop counter monitoring feature which allows users to configure thresholds for drop counters and register alerts when persistent drops are detected.

**How I verified it**
- Existing unit tests were run using `make check` to ensure no functionality was affected.
- New unit tests have been added to verify the functionality.
- Manual testing was performed on a SONiC switch to verify that the orchagent correctly reads the configuration, generates alerts when thresholds are met, and can be toggled off/on.

**Details if related**
The configurable drop monitor feature reads he status (enabled/disabled) from the DEBUG_DROP_MONITOR table. The orchagent periodically (every minute) polls the drop counters and compares the counts against the configured thresholds. If the thresholds are exceeded, the orchagent generates a persistent alert which is stored in the `PERSISTENT_DROP_ALERTS` table in Counters DB. The feature can be toggled on/off for all drop counters or it can be turned on/off for a specific drop counter. The corresponding CLI to do so can be found in the CLI PR attached in the umbrella issue: https://github.com/sonic-net/SONiC/issues/1542

PR #4120, opened 2026-01-10
Add drop monitor attributes to supported debug counter attributes

manish1-arista:
**What I did**
Add the missing drop monitor attribute definitions to debug_counter.h and include them in the supported attributes set in debug_counter.cpp.

**Why I did it**
The drop monitor feature introduced new CONFIG_DB fields drop_monitor_status, drop_count_threshold, incident_count_threshold and window for drop monitoring configuration. However, these attributes were not added to the supported_debug_counter_attributes set, causing error logs during debug counter installation: ```"Unknown debug counter attribute 'drop_monitor_status'"```
It can replicated by running the ```test_configurable_drop_counters.py``` testcase which will fail in teardown with this error.

**How I verified it**
Verified the fix by running test_configurable_drop_counters.py which was previously failing in teardown with the error "Unknown debug counter attribute 'drop_monitor_status'". After applying the fix, the test passes without generating any error logs

**Details if related**

### DIFF
--- a/orchagent/debug_counter/debug_counter.cpp
+++ b/orchagent/debug_counter/debug_counter.cpp
@@ -27,7 +27,12 @@ const unordered_set<string> DebugCounter::supported_debug_counter_attributes =
     COUNTER_ALIAS,
     COUNTER_TYPE,
     COUNTER_DESCRIPTION,
-    COUNTER_GROUP
+    COUNTER_GROUP,
+    // Drop counter specific attributes for drop monitor feature
+    DROP_MONITOR_STATUS,
+    DROP_MONITOR_DROP_COUNT_THRESHOLD,
+    DROP_MONITOR_INCIDENT_COUNT_THRESHOLD,
+    DROP_MONITOR_WINDOW
 };
 
 const std::unordered_map<std::string, sai_debug_counter_type_t> DebugCounter::debug_counter_type_lookup =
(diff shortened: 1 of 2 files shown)


prsunny:
@manish1-arista , can you link the HLD here? Also, is it backward compatible if some platforms doesn't support this new attributes?

manish1-arista:
> @manish1-arista , can you link the HLD here? Also, is it backward compatible if some platforms doesn't support this new attributes?

This is a fix for a bug introduced in PR #3509, so there is no new HLD.
The attributes already exist in the Redis DB, but were missing from the validation list, causing Orchagent to log errors. This change simply fixes that logging issue; The functional logic is already working as expected.

StormLiangMS:
@prsunny  could you help to review again?

manish1-arista:
@prsunny  Needs to be cherry picked for 202511

StormLiangMS:
This PR is merged to master and has the 'Request for 202511 Branch' label but no 202511 cherry-pick PR exists yet. @vmittal-msft could you help create the 202511 backport? Needed for Arista 202511 qualification (aristanetworks/sonic-qual.msft#1011).

StormLiangMS:
@vmittal-msft This PR is merged to master and labeled 'Request for 202511 Branch'. Could you check if this is eligible for 202511 backport? Related to Arista qualification issue aristanetworks/sonic-qual.msft#1011.

PR #4121, opened 2026-01-12
Increase Vrf pool to 4k

ypcisco:
**What I did**

Increase VRF pool size from 999 to 4096 to support hardware platforms with higher VRF scale capabilities.

**Why I did it**

Current VRF pool is limited to 999 instances (VRF_TABLE_END=2000, VRF_TABLE_START=1001) which artificially restricts VRF scale even on hardware platforms that support more VRFs. 

**How I verified it**

**Details if related**


prsunny:
Is there an HLD and sonic-mgmt test for this scale scenario? Please link it here

PR #4150, opened 2026-01-22
[ha][vnetorch] relax attr parsing for vnet route table

zjswhhh:
**What I did**
Set the relax attr parsing flag to true for vnet route table, by default it's false. 

If not, exception will throw when there is an expected field. 

https://github.com/sonic-net/sonic-swss/blob/a8d968c9e1767b77573982b693aa17d562276c8b/orchagent/request_parser.cpp#L160-L167

**Why I did it**

We are adding new fields i.e.  `pinned_state`, the exception throwing adds unnecessary dependency on the ordering of PR merging and feature enabling. 

**How I verified it**
UTs. 

**Details if related**

### DIFF
--- a/orchagent/vnetorch.h
+++ b/orchagent/vnetorch.h
@@ -386,7 +386,7 @@ private:
 class VNetRouteRequest : public Request
 {
 public:
-    VNetRouteRequest() : Request(vnet_route_description, ':') { }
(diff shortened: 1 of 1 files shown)


vivekkarri:
FYI Direct backport to 202511 probably will not work. The ENI forwarding PR is not backported

zjswhhh:

PR #4152, opened 2026-01-23
Added mux neighbor_mode support with prefix-route

manamand2020:
**What I did**
Added support for mux neighbor-mode based on the following HLD:
https://github.com/sonic-net/SONiC/pull/2176

- configure mux neighbors with no-host-route flags when neighbor_mode is set to prefix_route
- Added SAI capability check for NO_HOST_ROUTE neighbor attribute
- By default the mux neighbors are treated with host-route neighbor_mode
- Added a new neighbor handler class overriding the existing neighbor handler class to support prefix-mode mux neighbors
- NHG creation with prefix-route mux neighbors will not add tunnel nexthop as its member
- NHG with prefix-route mux neighbors will not be replaced by tunnel nexthops when mux state goes to standby
- Modified and added new DVS tests for prefix-mode neighbors in test_mux_prefixroute.py

**Why I did it**
This improved switchover performance by reducing number of SAI calls.

**How I verified it**
DVS testing
Dataplane Community tests

**Details if related**
Management dualtor io tests run. 1 failure is related to baseline faliure and unrelated to the changes done in this PR.
image

### DIFF
--- a/.azure-pipelines/test-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/test-docker-sonic-vs-template.yml
@@ -119,7 +119,7 @@ jobs:
   - script: |
       set -ex
       # install packages for vs test
-      sudo pip3 install pytest flaky exabgp docker redis lcov_cobertura
+      sudo pip3 install pytest flaky exabgp docker lcov_cobertura
 
       # install other dependencies
       sudo apt-get -o DPkg::Lock::Timeout=600 install -y net-tools \
@@ -127,11 +127,9 @@ jobs:
           vlan \
           libzmq3-dev \
           libzmq5 \
-          libboost-serialization1.74.0 \
-          libboost1.74-dev \
-          libboost-dev \
           libhiredis0.14 \
-          libpcre3-dev
+          libpcre3-dev \
+          python3-redis
 
       sudo .azure-pipelines/build_and_install_module.sh
 
@@ -143,7 +141,7 @@ jobs:
 
       sudo dpkg -i $(Build.ArtifactStagingDirectory)/download/libprotobuf*_amd64.deb $(Build.ArtifactStagingDirectory)/download/libprotobuf-lite*_amd64.deb $(Build.ArtifactStagingDirectory)/download/python3-protobuf*_amd64.deb
       sudo dpkg -i $(Build.ArtifactStagingDirectory)/download/libdashapi*.deb
-      sudo dpkg -i --force-confask,confnew $(Build.ArtifactStagingDirectory)/download/libswsscommon_1.0.0_amd64.deb || apt-get install -f
+      sudo dpkg -i --force-confask,confnew $(Build.ArtifactStagingDirectory)/download/libswsscommon_1.0.0_amd64.deb || sudo apt-get install -y -f
       sudo dpkg -i $(Build.ArtifactStagingDirectory)/download/python3-swsscommon_1.0.0_amd64.deb
     displayName: "Install dependencies"
 


orchagent/muxorch.cpp
```diff
@@ -658,6 +679,27 @@ void MuxCable::updateNeighbor(NextHopKey nh, bool add)
...
+}
+
+void MuxCable::updateNeighborFromEvent(NextHopKey nh, bool add)
+{
+    SWSS_LOG_NOTICE("Processing neighbor event update on neighbor %s for mux %s, add %d, state %d",
+                     nh.ip_address.to_string().c_str(), mux_name_.c_str(), add, state_);
+    sai_object_id_t tnh = mux_orch_->getNextHopTunnelId(MUX_TUNNEL, peer_ip4_);
+    // For neighbor events, check prefix route to avoid updating neighbors that got added without prefix route.
+    nbr_handler_->update(nh, tnh, add, state_, true);
+
```
Ndancejic: Please add a comment why update is before the add/remove nexthop

orchagent/muxorch.cpp
```diff
@@ -1083,6 +1126,302 @@ void MuxNbrHandler::updateTunnelRoute(NextHopKey nh, bool add)
     }
 }
 
+bool MuxNbrHandler::setBulkRouteNH(std::list<MuxRouteBulkContext>& bulk_ctx_list)
```
Ndancejic, line 1117: Please add a comment for what this does

orchagent/muxorch.cpp
```diff
@@ -1379,6 +1718,75 @@ MuxCable* MuxOrch::findMuxCableInSubnet(IpAddress ip)
...
+    // If prefix nbrs are not supported, return false
+    if (!prefix_nbrs_supported_)
+    {
+        return false;
+    }
+
+    // Check if this is a skip neighbor and get its associated port
+    if (isSkipNeighbor(nbr))
+    {
+        std::string skip_port = getSkipNeighborPort(nbr);
```
Ndancejic, line 1689: Can skip neighbors change port?

orchagent/neighorch.cpp
```diff
@@ -70,6 +70,35 @@ NeighOrch::~NeighOrch()
     }
 }
 
+bool NeighOrch::isNoHostRouteSupported()
```
Ndancejic, line 78: Add a comment that mentions that this is related to prefix_based in muxorch

orchagent/neighorch.cpp
```diff
@@ -1133,82 +1190,171 @@ bool NeighOrch::addNeighbor(NeighborContext& ctx)
...
+                else
+                {
+                    gCrmOrch->decCrmResUsedCounter(CrmResourceType::CRM_IPV6_NEIGHBOR);
+                }
+
+                return false;
+            }
+
+            // full prefix route pointing to neighbor nh
+            if (prefix_route)
```
Ndancejic: Break out the prefix_route section into a separate function if it's feasable

orchagent/neighorch.cpp
```diff
@@ -2186,6 +2374,136 @@ bool NeighOrch::delInbandNeighbor(string alias, IpAddress ip_address)
...
+    auto neighbor_it = m_syncdNeighbors.find(neighborEntry);
+    if (neighbor_it == m_syncdNeighbors.end())
+    {
+        SWSS_LOG_ERROR("Neighbor %s on %s not found, cannot convert to MUX neighbor",
+                       neighborEntry.ip_address.to_string().c_str(), neighborEntry.alias.c_str());
+        return false;
+    }
+
+    // Check if neighbor is hardware configured before conversion
+    if (!neighbor_it->second.hw_configured)
```
Ndancejic, line 2421: Is it possible for a host-route standby neighbor being converted to prefix-based route as hw_configured=false?

tests/test_mux_prefixroute.py
```diff
@@ -0,0 +1,1988 @@
+import time
```
Ndancejic, line 1: Just adding comment for reference: this is the old test_mux.py tests (host route mux neighbors) now that the default behavior is prefix_based  mux neighbors

orchagent/muxorch.cpp
manamand2020: Skip neighbors do not change port as they are added based on config.

orchagent/neighorch.cpp
manamand2020: Since we do not allow dynamic re-configuration of neighbor_mode and we log error for such cases in handleMuxCfg() this case is not valid.
```
            // Check if neighbor_mode has changed - dynamic changes are not allowed
              auto mux_cable = getMuxCable(port_name);
              if (mux_cable && mux_cable->getNbrHandlerType() != nbr_handler_type)
              {
                  SWSS_LOG_ERROR("Neighbor mode change is not allowed for existing mux port '%s'. "
                                "Current: %s, Requested: %s. Please delete and recreate the mux port.",
                                port_name.c_str(),
                                mux_cable->getNbrHandlerType() == MuxNbrHandlerType::NBR_HANDLER_PREFIX_BASED ? "prefix-route" : "host-route",
                                nbr_handler_type == MuxNbrHandlerType::NBR_HANDLER_PREFIX_BASED ? "prefix-route" : "host-route");
                  return false;
              }

File: orchagent/muxorch.cpp (hunk 1 above)
Comment (manamand2020) [reply orchagent/muxorch.cpp]: This update is moved before updating the mux_nexthop_tb_ to make sure we only add/remove the NH from the table after the update is complete. As such there is no dependency but looks logically correct to me.

File: tests/test_mux.py [hunk 8]
```diff
@@ -1076,18 +1076,31 @@ def create_and_test_multi_nexthop_routes(self, dvs, dvs_route, appdb, macs, new_
             ### These tests create route: ###
 
             ## Testing route changes
+            dvs.add_log_marker()
```
Ndancejic: Please remove the markers

tests/test_mux.py
```diff
@@ -1892,6 +1939,7 @@ def test_NH(self, dvs, dvs_route, intf_fdb_map, setup, setup_mux_cable,
         self.create_and_test_NH_routes(appdb, asicdb, dvs, dvs_route, mac)
 
     def test_multi_nexthop(self, dvs, dvs_route, intf_fdb_map, neighbor_cleanup, testlog, setup):
+        dvs.runcmd("swssloglevel -l INFO -c orchagent")
```
Ndancejic: Please remove INFO log cmd

tests/test_mux_hostroute.py
```diff
@@ -1982,6 +1949,43 @@ def test_warm_boot_neighbor_restore(
             ]
         )
 
+    def test_warm_boot_mux_state(
```
Ndancejic: Please confirm if the sequence passes w/o changes

manamand2020: Good catch. The original warmboot sequence is not this and revering back to that sequence revealed a warmboot failure with following logs. Working on to resolve these failures. Looks like for some reason doTask() -- addOperation for MUX_CABLE_TABLE is not geting called and draining the MUX entries.
```
Feb  3 01:45:37.413345 0840af33886c NOTICE #orchagent: :- updateCachedNeighbors: Skip process cached neighbor updates, no peer switch addr is configured
Feb  3 01:45:37.413421 0840af33886c NOTICE #orchagent: :- warmRestoreValidation: There are pending consumer tasks after restore:
Feb  3 01:45:37.413431 0840af33886c NOTICE #orchagent: :- warmRestoreValidation: MUX_CABLE_TABLE:Ethernet0|SET|state:active
Feb  3 01:45:37.413439 0840af33886c NOTICE #orchagent: :- warmRestoreValidation: MUX_CABLE_TABLE:Ethernet4|SET|state:active
Feb  3 01:45:37.413448 0840af33886c NOTICE #orchagent: :- warmRestoreValidation: MUX_CABLE_TABLE:Ethernet8|SET|state:standby
Feb  3 01:45:37.413600 0840af33886c NOTICE #orchagent: :- setWarmStartState: orchagent warm start state changed to restored
Feb  3 01:45:37.413611 0840af33886c ERR #orchagent: :- warmRestoreAndSyncUp: Orchagent state restore failed
Feb  3 01:45:37.413634 0840af33886c ERR #orchagent: :- main: Failed to initialize orchestration daemon

File: tests/test_mux.py (hunk 9 above)
Comment (manamand2020) [reply tests/test_mux.py]: Done

File: tests/test_mux.py (hunk 8 above)
Comment (manamand2020) [reply tests/test_mux.py]: Done

File: tests/test_mux_hostroute.py (hunk 10 above)
Comment (manamand2020) [reply tests/test_mux_hostroute.py]: Resolved this, the bake() override now calls orch::bake() to execute the consumers of config tables.

**Review by zjswhhh — APPROVED** [review]
lgtm

File: orchagent/neighorch.cpp [hunk 11]
```diff
@@ -1064,18 +1187,47 @@ bool NeighOrch::addNeighbor(NeighborContext& ctx)
     memcpy(neighbor_attr.value.mac, macAddress.getMac(), 6);
     neighbor_attrs.push_back(neighbor_attr);
 
-    if ((ip_address.getAddrScope() == IpAddress::LINK_SCOPE) && (ip_address.isV4()))
+    MuxOrch* mux_orch = gDirectory.get<MuxOrch*>();
+    bool no_host_route = false;
+    bool prefix_route = false;
+
+    // Only set no_host_route if the capability is supported
+    if (isNoHostRouteSupported())
```
manamand2020: Since link local used the no-hot-route attribute without checking the capablity in the older code, keep the behavior same.

orchagent/neighorch.cpp
```diff
@@ -1133,82 +1284,106 @@ bool NeighOrch::addNeighbor(NeighborContext& ctx)
...
     }
 
-    if (!hw_config && mux_orch->isNeighborActive(ip_address, macAddress, alias))
+    if (!hw_config)
     {
-        // Using bulker, return and post-process later
-        if (bulk_op)
+        // When prefix route is not enabled (capability not supported),
+        // only add mux neighbors if they are active
+        if (!prefix_route && mux_orch && !mux_orch->isNeighborActive(ip_address, macAddress, alias))
```
manamand2020: Its better to separate out the prefix-route handling using a new handler which will avoid changing existing logic.

orchagent/neighorch.cpp
```diff
@@ -1280,91 +1468,128 @@ bool NeighOrch::removeNeighbor(NeighborContext& ctx, bool disable)
...
         sai_neighbor_entry_t neighbor_entry;
         neighbor_entry.rif_id = rif_id;
         neighbor_entry.switch_id = gSwitchId;
         copy(neighbor_entry.ip_address, ip_address);
 
         sai_object_id_t next_hop_id = m_syncdNextHops[nexthop].next_hop_id;
 
-        if (bulk_op)
+        // prefix-route neighbors do not use bulk_op
+        if (bulk_op && !m_syncdNeighbors[neighborEntry].prefix_route)
```
manamand2020: This check is not required because prefix_route mode will not use bulk. If we remove this change we will not have these changes below.

orchagent/neighorch.cpp
```diff
@@ -2186,6 +2411,136 @@ bool NeighOrch::delInbandNeighbor(string alias, IpAddress ip_address)
...
+    rt_attr.value.s32 = SAI_PACKET_ACTION_FORWARD;
+    rt_attrs.push_back(rt_attr);
+
+    // Set next hop (use tunnel nexthop if provided, otherwise use neighbor nexthop)
+    rt_attr.id = SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID;
+    rt_attr.value.oid = (tunnel_nexthop_id != SAI_NULL_OBJECT_ID) ? tunnel_nexthop_id : nexthop_it->second.next_hop_id;
+    rt_attrs.push_back(rt_attr);
+
+    // Create the prefix route
+    status = sai_route_api->create_route_entry(&route_entry, (uint32_t)rt_attrs.size(), rt_attrs.data());
```
manamand2020, line 2503: Check CRM for all other SAI APIs. Update the CRM here.

orchagent/muxorch.cpp
```diff
@@ -1944,6 +2538,74 @@ bool MuxOrch::isStandaloneTunnelRouteInstalled(const IpAddress& neighborIp)
     return standalone_tunnel_neighbors_.find(neighborIp) != standalone_tunnel_neighbors_.end();
 }
 
+void MuxOrch::restoreMuxNeighbors()
```
manamand2020: Keep warmboot changes out of this review and only test warmboot for host-route mode.

orchagent/muxorch.cpp
```diff
@@ -194,9 +194,20 @@ static sai_status_t set_route(const IpPrefix& pfx, sai_object_id_t next_hop_id)
     sai_status_t status = sai_route_api->set_route_entry_attribute(&route_entry, &route_attr);
     if (status != SAI_STATUS_SUCCESS)
     {
-        SWSS_LOG_ERROR("Failed to set route entry %s nh %" PRIx64 " rv:%d",
+        SWSS_LOG_ERROR("Failed to set route entry %s nh 0x%" PRIx64 " rv:%d",
                 pfx.to_string().c_str(), next_hop_id, status);
     }
+
```
manamand2020, line 200: Keep all changes limited to prefix-route.

manamand2020 (commented):
Address the comments. Mostly we want to keep the changes such that it does not alter the existing code for host-route mode niehgbors. Overall design looks good.

orchagent/neighorch.cpp
manamand2020: Done

orchagent/neighorch.cpp
manamand2020: Done

orchagent/neighorch.cpp
manamand2020: Done

orchagent/neighorch.cpp
manamand2020: Done

orchagent/muxorch.cpp
manamand2020: Removed.

orchagent/muxorch.cpp
manamand2020: Done.

orchagent/muxorch.cpp
```diff
@@ -711,12 +755,12 @@ void MuxCable::updateRoutesForNextHop(NextHopKey nh)
     }
 }
 
-void MuxNbrHandler::update(NextHopKey nh, sai_object_id_t tunnelId, bool add, MuxState state)
+void MuxNbrHandler::update(NextHopKey nh, sai_object_id_t tunnelId, bool add, MuxState state, bool check_prefix_route)
```
manamand2020: Check of we can remove the check_prefix_route.

orchagent/muxorch.cpp
```diff
@@ -1083,6 +1129,294 @@ void MuxNbrHandler::updateTunnelRoute(NextHopKey nh, bool add)
...
+        {
+        case MuxState::MUX_STATE_INIT:
+            neighbors_[nh.ip_address] = SAI_NULL_OBJECT_ID;
+            break;
+        case MuxState::MUX_STATE_ACTIVE:
+            neighbors_[nh.ip_address] = local_nhid;
+
+            // Update the prefix route with local nexthop
+            // Only update route if prefix route exists (when check_prefix_route is true)
+            if (!check_prefix_route || gNeighOrch->isPrefixNeighborNh(nh))
```
manamand2020: This check for prefixNeighborNh is not needed any more.

orchagent/muxorch.cpp
```diff
@@ -1083,6 +1129,294 @@ void MuxNbrHandler::updateTunnelRoute(NextHopKey nh, bool add)
...
+
+            gRouteOrch->updateNextHopRoutes(nh, num_routes);
+            gNeighOrch->increaseNextHopRefCount(nh, num_routes);
+            break;
+        case MuxState::MUX_STATE_STANDBY:
+            neighbors_[nh.ip_address] = tunnelId;
+
+            // Update the prefix route with tunnel nexthop
+            // Only update route if prefix route exists (when check_prefix_route is true)
+            if (!check_prefix_route || gNeighOrch->isPrefixNeighborNh(nh))
```
manamand2020: Ditto.

orchagent/muxorch.cpp
```diff
@@ -658,6 +682,27 @@ void MuxCable::updateNeighbor(NextHopKey nh, bool add)
...
+    updateRoutesForNextHop(nh);
+}
+
+void MuxCable::updateNeighborFromEvent(NextHopKey nh, bool add)
+{
+    SWSS_LOG_NOTICE("Processing neighbor event update on neighbor %s for mux %s, add %d, state %d",
+                     nh.ip_address.to_string().c_str(), mux_name_.c_str(), add, state_);
+    sai_object_id_t tnh = mux_orch_->getNextHopTunnelId(MUX_TUNNEL, peer_ip4_);
+    // For neighbor events, check prefix route to avoid updating neighbors that got added without prefix route.
+    nbr_handler_->update(nh, tnh, add, state_, true);
```
manamand2020: Can we revert to original logic? Since we now have a prefix based handler passing the prefix_route bool to the update handler is also not required.

orchagent/muxorch.cpp
```diff
@@ -1933,9 +2495,17 @@ void MuxOrch::createStandaloneTunnelRoute(IpAddress neighborIp)
...
 {
-    SWSS_LOG_INFO("Removing standalone tunnel route for neighbor %s", neighborIp.to_string().c_str());
-    IpPrefix pfx = neighborIp.to_string();
-    remove_route(pfx);
+    NeighborEntry neighbor;
+    MacAddress mac;
+
+    SWSS_LOG_NOTICE("Removing standalone tunnel route for neighbor %s", neighborIp.to_string().c_str());
+    // remove the route prefix if its not a mux neighbor
+    if (!gNeighOrch->getNeighborEntry(neighborIp, neighbor, mac) ||
```
manamand2020, line 2461: See why this change is needed. It seems this is not required and we should remove the route irrespective of the neigbor_mode used.

orchagent/routeorch.cpp
```diff
@@ -1697,6 +1704,20 @@ bool RouteOrch::removeNextHopGroup(const NextHopGroupKey &nexthops, const bool i
             continue;
         }
 
+        bool is_neighbor_nh = !nhop->first.isMplsNextHop() &&
```
manamand2020: Remove the extra NH checks.

Ndancejic:
Summary of offline discussion for PR:
- default behavior should be host_route mode for neighbors
- Remove changes to MuxNbrHandler::update() - not needed for host prefix mode
- keep updateNeighbor function call and change behavior only for prefix based neighbors - not needed due to inherited nbr_handler
- Keep standalone-tunnel-route code as-is (or confirm why check is needed)
- remove extra checks on is_neighbor_nh


will make final changes and look to merge late next week

orchagent/muxorch.cpp
manamand2020: This check is needed because of the following reasons:
Follow the sequence of events:
1. A zero MAC neighbor is found. --> This will install the tunnelroute using addZeroMacTunnelRoute and muxorch nbr update.
2. Now if MAC address is learnt and updated for this zero MAC nbr. --> this will call addNeighbor() -> addPrefixRouteForNeighbor() --> This code checks if the tunnelroute was installed and it will just update that prefix route to point to nbr nh. Follow this here:
```
    // if standalone mux route for this neighbor is created, then
      // set the new attributes
      if (mux_orch->isStandaloneTunnelRouteInstalled(ip_address))
      {
          for (auto& route_attr : rt_attrs)
          {
              status = sai_route_api->set_route_entry_attribute(&route_entry, &route_attr);
              {
                  SWSS_LOG_ERROR("Failed to set mux neigh route for %s with next_hop_id as 0x%" PRIx64 " rv:%d",
                          ip_address.to_string().c_str(), next_hop_id, status);
                  return false;
              }
          }
      }
 
```
The neighbor deletion will result in cleanup of this route eventually and removeStandaloneTunnelRoute() will avoid this route removal as the neighbor is in use of this route.

manamand2020: @prsunny please follow the above explanation

orchagent/muxorch.cpp
manamand2020: Good catch, done.

orchagent/routeorch.cpp
manamand2020: Done.

orchagent/muxorch.cpp
manamand2020: Removed, this is not needed.

orchagent/muxorch.cpp
manamand2020: Done.

orchagent/muxorch.cpp
manamand2020: Yes this is not needed anymore.

manamand2020:
> Summary of offline discussion for PR:
> 
> * default behavior should be host_route mode for neighbors
> * Remove changes to MuxNbrHandler::update() - not needed for host prefix mode
> * keep updateNeighbor function call and change behavior only for prefix based neighbors - not needed due to inherited nbr_handler
> * Keep standalone-tunnel-route code as-is (or confirm why check is needed)
> * remove extra checks on is_neighbor_nh
> 
> will make final changes and look to merge late next week

I have addressed the review comments, please review the latest diff.

tests/test_mux.py
```diff
@@ -5,6 +5,7 @@
 
 from ipaddress import ip_network, ip_address, IPv4Address
 from swsscommon import swsscommon
+from dvslib.dvs_common import PollingConfig
```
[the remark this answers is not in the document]
manamand2020: Used in this file now.

orchagent/muxorch.cpp
prsunny: what if there is only a 'zero mac' neighbor add and 'zero mac' neighbor delete. ie. no step 2 happening

manamand2020: There might be already a VS test to check that condition if not i can add a test case to check this condition. But here is what I understand the Neighbor Delete will delete the neighbor prefix route with removePrefixRouteForNeighbor() irrespective of zero mac neighbor.

Ndancejic (approved):

prsunny (approved):

prsunny:
@dgsudharsan, @lolyu  for viz/any comments. This PR is ensuring backward compatibility and prefix mode triggered only with config option

PR #4158, opened 2026-01-27
Fine Grained ECMP for Vnet Tunnel Routes

NavdhaJindal:
**What I did**
This PR adds support for fine-grained ECMP (FG-ECMP) for VNET tunnel routes using a new consistent_hashing_buckets attribute, wiring VNET route programming into the existing FG-NHG orchestration and extending tests accordingly.
Changed FG_ROUTE_TABLE schema to include Vnet name in the key. 

Changes:

Extend VNET route configuration to accept consistent_hashing_buckets and plumb it through to VNetRouteOrch::doRouteTask, which now invokes FgNhgOrch to create/update fine-grained ECMP next-hop groups for tunnel routes.
Enhance FgNhgOrch to support FG-ECMP for non-default VRFs (VNET VRFs), including per-VNET state-DB keys (vnet|prefix) and a new overload of setFgNhg that can work directly with tunnel next-hop IDs and consistent hashing bucket sizes.
Add and update unit tests in test_vnet.py, vnet_lib.py, and test_fgnhg.py to exercise FG-ECMP behavior for VNET routes and to validate new state-DB key formats.

**Why I did it**
To allow consistent ecmp for VXLAN tunnels according to HLD https://github.com/sonic-net/SONiC/pull/2099

**How I verified it**
Via unit tests and datapath testing

**Details if related**

### DIFF
--- a/orchagent/fgnhgorch.cpp
+++ b/orchagent/fgnhgorch.cpp
@@ -6,6 +6,8 @@
 #include "logger.h"
 #include "swssnet.h"
 #include "crmorch.h"
+#include "directory.h"
+#include "vnetorch.h"
 #include <array>
 #include <algorithm>
 
@@ -15,9 +17,11 @@
 extern sai_object_id_t gVirtualRouterId;
 extern sai_object_id_t gSwitchId;
 
+extern sai_switch_api_t*            sai_switch_api;
 extern sai_next_hop_group_api_t*    sai_next_hop_group_api;
 extern sai_route_api_t*             sai_route_api;
 
+extern Directory<Orch*> gDirectory;
 extern RouteOrch *gRouteOrch;
 extern CrmOrch *gCrmOrch;
 extern PortsOrch *gPortsOrch;
@@ -213,22 +217,30 @@ void FgNhgOrch::calculateBankHashBucketStartIndices(FgNhgEntry *fgNhgEntry)
 }
 
 
-void FgNhgOrch::setStateDbRouteEntry(const IpPrefix &ipPrefix, uint32_t index, NextHopKey nextHop)
+string FgNhgOrch::getWarmRebootStateDbKey(const string &vnet, const IpPrefix &ipPrefix)
+{
+    if (vnet.empty())
+    {
+        return ipPrefix.to_string();
+    }
+    return vnet + state_db_key_delimiter + ipPrefix.to_string();
+}
+
+void FgNhgOrch::setStateDbRouteEntry(const string &key, uint32_t index, NextHopKey nextHop)
 {
     SWSS_LOG_ENTER();
 
-    string key = ipPrefix.to_string();
     string field = std::to_string(index);
     string value = nextHop.to_string();
 
     m_stateWarmRestartRouteTable.hset(key, field, value);
 
-    SWSS_LOG_INFO("Set state db entry for ip prefix %s next hop %s with index %d",
+    SWSS_LOG_INFO("Set state db entry for key %s next hop %s with index %d",
                   key.c_str(), value.c_str(), index);
 }
 
 bool FgNhgOrch::writeHashBucketChange(FGNextHopGroupEntry *syncd_fg_route_entry, HashBucketIdx index, sai_object_id_t nh_oid,
-        const IpPrefix &ipPrefix, NextHopKey nextHop)
+        const string &key, NextHopKey nextHop)
 {
     SWSS_LOG_ENTER();
 
@@ -249,7 +261,59 @@ bool FgNhgOrch::writeHashBucketChange(FGNextHopGroupEntry *syncd_fg_route_entry,
         }
     }
 
-    setStateDbRouteEntry(ipPrefix, index, nextHop);
+    setStateDbRouteEntry(key, index, nextHop);
+    return true;
+}
+
+bool FgNhgOrch::getMaxEcmpMemberCount(uint32_t &maxEcmpMembers)
+{
+    SWSS_LOG_ENTER();
+
+    sai_attribute_t switch_attr;
+    switch_attr.id = SAI_SWITCH_ATTR_MAX_ECMP_MEMBER_COUNT;
+    switch_attr.value.u32 = 0;
+
+    sai_status_t status = sai_switch_api->get_switch_attribute(gSwitchId, 1, &switch_attr);
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_WARN("Failed to query SAI_SWITCH_ATTR_MAX_ECMP_MEMBER_COUNT, rv:%d", status);
+        return false;
+    }
+
+    maxEcmpMembers = switch_attr.value.u32;
+    return true;
+}
+
+bool FgNhgOrch::getEffectiveFgNhgMemberCapacity(uint32_t configuredBucketSize, uint32_t &memberCapacity)
+{
+    SWSS_LOG_ENTER();
+
+    memberCapacity = configuredBucketSize;
+
+    uint32_t hw_max_ecmp_members = 0;
+    if (!getMaxEcmpMemberCount(hw_max_ecmp_members))
+    {
+        return false;
+    }
+
+    memberCapacity = min(memberCapacity, hw_max_ecmp_members);
+    return true;
+}
+
+bool FgNhgOrch::validateFgNhgMemberCount(size_t memberCount, uint32_t configuredBucketSize, const string &logKey)
+{
+    SWSS_LOG_ENTER();
+
+    uint32_t member_capacity = 0;
+    getEffectiveFgNhgMemberCapacity(configuredBucketSize, member_capacity);
+
+    if (memberCount > member_capacity)
+    {
+        SWSS_LOG_ERROR("Number of next-hops %zu exceeds effective FG NHG member capacity %u for %s",
+                       memberCount, member_capacity, logKey.c_str());
+        return false;
+    }
+
     return true;
 }
 
@@ -261,6 +325,30 @@ bool FgNhgOrch::createFineGrainedNextHopGroup(FGNextHopGroupEntry &syncd_fg_rout
     string platform = getenv("platform") ? getenv("platform") : "";
     sai_attribute_t nhg_attr;
     vector<sai_attribute_t> nhg_attrs;
+
+    /* Query hardware maximum ECMP member count */
+    uint32_t effective_member_capacity = 0;
+    bool hw_limit_available = getEffectiveFgNhgMemberCapacity(fgNhgEntry->configured_bucket_size,
+                                                              effective_member_capacity);
+    
+    if (!hw_limit_available)
+    {
+        SWSS_LOG_WARN("Unable to verify hardware ECMP member limit. Proceeding with configured bucket size %d",
+                      fgNhgEntry->configured_bucket_size);
+    }
+    else
+    {
+        SWSS_LOG_INFO("Effective FG NHG member capacity: %d, Configured bucket size: %d",
+                      effective_member_capacity, fgNhgEntry->configured_bucket_size);
+        
+        /* If configured bucket size exceeds hardware maximum, adjust it */
+        if (fgNhgEntry->configured_bucket_size > effective_member_capacity)
+        {
+            SWSS_LOG_WARN("Configured bucket size %d exceeds effective FG NHG member capacity %d. Adjusting bucket size.",
+                          fgNhgEntry->configured_bucket_size, effective_member_capacity);
+            fgNhgEntry->configured_bucket_size = effective_member_capacity;
+        }
+    }
 
     nhg_attr.id = SAI_NEXT_HOP_GROUP_ATTR_TYPE;
     nhg_attr.value.s32 = SAI_NEXT_HOP_GROUP_TYPE_FINE_GRAIN_ECMP;
@@ -440,6 +528,10 @@ bool FgNhgOrch::validNextHopInNextHopGroup(const NextHopKey& nexthop)
                     nhs_to_add.push_back(nexthop);
             nhopgroup_members_set[nexthop] = m_neighOrch->getNextHopId(nexthop);
 
+            string vnet;
+            getVnetNameByVrfId(route_tables.first, vnet);
+            string key = getWarmRebootStateDbKey(vnet, route_table.first);
+
             if (syncd_fg_route_entry->points_to_rif)
             {
                 // RIF route is now neigh resolved: create Fine Grained ECMP
@@ -448,7 +540,7 @@ bool FgNhgOrch::validNextHopInNextHopGroup(const NextHopKey& nexthop)
                     return false;
                 }
 
-                if (!setNewNhgMembers(*syncd_fg_route_entry, fgNhgEntry, bank_member_changes, nhopgroup_members_set, route_table.first))
+                if (!setNewNhgMembers(*syncd_fg_route_entry, fgNhgEntry, bank_member_changes, nhopgroup_members_set, key))
                 {
                     return false;
                 }
@@ -468,7 +560,7 @@ bool FgNhgOrch::validNextHopInNextHopGroup(const NextHopKey& nexthop)
                 }
 
                 if (!computeAndSetHashBucketChanges(syncd_fg_route_entry, fgNhgEntry,
-                        bank_member_changes, nhopgroup_members_set, route_table.first))
+                        bank_member_changes, nhopgroup_members_set, key, route_table.first))
                 {
                     SWSS_LOG_ERROR("Failed to set fine grained next hop %s",
                         nexthop.to_string().c_str());
@@ -546,8 +638,12 @@ bool FgNhgOrch::invalidNextHopInNextHopGroup(const NextHopKey& nexthop)
             bank_member_changes[fgNhgEntry->next_hops[nexthop.ip_address].bank].
                     nhs_to_del.push_back(nexthop);
 
+            string vnet;
+            getVnetNameByVrfId(route_tables.first, vnet);
+            string key = getWarmRebootStateDbKey(vnet, route_table.first);
+
             if (!computeAndSetHashBucketChanges(syncd_fg_route_entry, fgNhgEntry,
-                    bank_member_changes, nhopgroup_members_set, route_table.first))
+                    bank_member_changes, nhopgroup_members_set, key, route_table.first))
             {
                 SWSS_LOG_ERROR("Failed to set fine grained next hop %s",
                     nexthop.to_string().c_str());
@@ -577,7 +673,7 @@ bool FgNhgOrch::invalidNextHopInNextHopGroup(const NextHopKey& nexthop)
  */
 bool FgNhgOrch::setActiveBankHashBucketChanges(FGNextHopGroupEntry *syncd_fg_route_entry, FgNhgEntry *fgNhgEntry,
         uint32_t syncd_bank, BankMemberChanges bank_member_change,
-        std::map<NextHopKey,sai_object_id_t> &nhopgroup_members_set, const IpPrefix &ipPrefix)
+        std::map<NextHopKey,sai_object_id_t> &nhopgroup_members_set, const string &key)
 {
     SWSS_LOG_ENTER();
 
@@ -595,7 +691,7 @@ bool FgNhgOrch::setActiveBankHashBucketChanges(FGNextHopGroupEntry *syncd_fg_rou
         {
             if (!writeHashBucketChange(syncd_fg_route_entry, hash_buckets->at(i),
                         nhopgroup_members_set[bank_member_change.nhs_to_add[add_idx]],
-                        ipPrefix, bank_member_change.nhs_to_add[add_idx]))
+                        key, bank_member_change.nhs_to_add[add_idx]))
             {
                 return false;
             }
@@ -643,7 +739,7 @@ bool FgNhgOrch::setActiveBankHashBucketChanges(FGNextHopGroupEntry *syncd_fg_rou
                     {
                         // this can neven happen
                         SWSS_LOG_ERROR("%s Unexpected no more active NHs before adding the %zu buckets, exp_bucket_size(%d)",
-                                       ipPrefix.to_string().c_str(),
+                                       key.c_str(),
                                        hash_buckets->size(), exp_bucket_size);
                         return false;
                     }
@@ -654,7 +750,19 @@ bool FgNhgOrch::setActiveBankHashBucketChanges(FGNextHopGroupEntry *syncd_fg_rou
                  * other available nhs, but for cases where # hash buckets is not
                  * divisible by # of nhs, simple round robin can make the hash bucket
                  * distribution non-ideal, thereby nhs can attract unequal traffic */
-                if (bank_fgnhg_map->at(*it).size() == exp_bucket_size)
+                if (bank_fgnhg_map->at(*it).size() > exp_bucket_size)
+                {
+                    // NH already has more than its fair share from a prior redistribution 
+                    if (num_nhs_with_one_more > 0)
+                    {
+                        num_nhs_with_one_more--;
+                    }
+                    SWSS_LOG_INFO("%s already has %zu buckets (> %d), skip assigning more, bkt_idx(%d)",
+                                  (*it).to_string().c_str(), bank_fgnhg_map->at(*it).size(),
+                                  exp_bucket_size, bkt_idx);
+                    remove_nh = true;
+                }
+                else if (bank_fgnhg_map->at(*it).size() == exp_bucket_size)
                 {
                     if (num_nhs_with_one_more == 0)
                     {
@@ -682,7 +790,7 @@ bool FgNhgOrch::setActiveBankHashBucketChanges(FGNextHopGroupEntry *syncd_fg_rou
                 if (move_bkt)
                 {
                     if (!writeHashBucketChange(syncd_fg_route_entry, hash_buckets->at(bkt_idx),
-                                               nhopgroup_members_set[*it], ipPrefix, *it))
+                                               nhopgroup_members_set[*it], key, *it))
                     {
                         return false;
                     }
@@ -794,7 +902,7 @@ bool FgNhgOrch::setActiveBankHashBucketChanges(FGNextHopGroupEntry *syncd_fg_rou
                     HashBucketIdx last_elem = map_entry->at((*map_entry).size() - 1);
                     if (!writeHashBucketChange(syncd_fg_route_entry, last_elem,
                                                nhopgroup_members_set[bank_member_change.nhs_to_add[add_idx]],
-                                               ipPrefix, bank_member_change.nhs_to_add[add_idx]))
+                                               key, bank_member_change.nhs_to_add[add_idx]))
                     {
                         return false;
                     }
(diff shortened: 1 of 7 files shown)


orchagent/fgnhgorch.cpp
```diff
@@ -213,22 +216,22 @@ void FgNhgOrch::calculateBankHashBucketStartIndices(FgNhgEntry *fgNhgEntry)
 }
 
 
-void FgNhgOrch::setStateDbRouteEntry(const IpPrefix &ipPrefix, uint32_t index, NextHopKey nextHop)
+void FgNhgOrch::setStateDbRouteEntry(const string &vnet, const IpPrefix &ipPrefix, uint32_t index, NextHopKey nextHop)
 {
     SWSS_LOG_ENTER();
 
-    string key = ipPrefix.to_string();
+    string key = vnet + '|' + ipPrefix.to_string();
```
anish-n: could we use orch.h: state_db_key_delimiter?

orchagent/fgnhgorch.cpp
```diff
@@ -213,22 +216,22 @@ void FgNhgOrch::calculateBankHashBucketStartIndices(FgNhgEntry *fgNhgEntry)
 }
 
 
-void FgNhgOrch::setStateDbRouteEntry(const IpPrefix &ipPrefix, uint32_t index, NextHopKey nextHop)
+void FgNhgOrch::setStateDbRouteEntry(const string &vnet, const IpPrefix &ipPrefix, uint32_t index, NextHopKey nextHop)
```
anish-n: could we rename vnet to vrf, vrf is often used as a generic representation of both vnet and vrf

anish-n: Could we pass in the key directly from the caller of setStateDbRouteEntry, I see that in some place the key is already present and we regenerate it later again

orchagent/fgnhgorch.cpp
```diff
@@ -1100,7 +1110,8 @@ bool FgNhgOrch::setNewNhgMembers(FGNextHopGroupEntry &syncd_fg_route_entry, FgNh
                 inactive_banks[i], ipPrefix.to_string().c_str());
     }
 
-    auto nexthopsMap = m_recoveryMap.find(ipPrefix.to_string());
+    string key = vnet + '|' + ipPrefix.to_string();
+    auto nexthopsMap = m_recoveryMap.find(key);
```
anish-n, line 1211: This change is not backwards compatible and would break warm reboot. Essentially if a switch goes through warmreboot, it reads the previously stored statedb entry(ie from a version which does not contain your changes, hence it would not contain <vnetname>|<ip-prefix>. The best way to fix this is to support no vnet name= default vrf, and if a vnet name is there then you add the vnet + | + ip prefix. That way the existing implementation for non vnet in default vrf space will keep working without introducing backwards incompatibility.

orchagent/fgnhgorch.cpp
```diff
@@ -1202,10 +1214,10 @@ bool FgNhgOrch::isRouteFineGrained(sai_object_id_t vrf_id, const IpPrefix &ipPre
 {
     SWSS_LOG_ENTER();
 
-    if (!isFineGrainedConfigured || (vrf_id != gVirtualRouterId))
+    if (!isFineGrainedConfigured)
```
anish-n: This is only used by routeorch, I believe we are not opening up support for route addition via routeorch for this feature, and is only for VNET_ROUTE_TUNNEL which comes via vnetorch. Can you keep this code the same/as it was before?

orchagent/fgnhgorch.cpp
```diff
@@ -1253,7 +1265,7 @@ bool FgNhgOrch::isRouteFineGrained(sai_object_id_t vrf_id, const IpPrefix &ipPre
 
 bool FgNhgOrch::syncdContainsFgNhg(sai_object_id_t vrf_id, const IpPrefix &ipPrefix)
 {
-    if (!isFineGrainedConfigured || (vrf_id != gVirtualRouterId))
+    if (!isFineGrainedConfigured)
```
anish-n, line 1364: same comment here: This is only used by routeorch, I believe we are not opening up support for route addition via routeorch for this feature, and is only for VNET_ROUTE_TUNNEL which comes via vnetorch. Can you keep this code the same/as it was before?

orchagent/fgnhgorch.cpp
```diff
@@ -1561,6 +1576,188 @@ bool FgNhgOrch::setFgNhg(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
...
+    {
+        m_FgNhgs[fg_nhg_name] = fgNhgEntry;
+    }
+
+    IpPrefix ip_prefix = ipPrefix;
+
+    if (m_fgNhgPrefixes.find(ip_prefix) == m_fgNhgPrefixes.end())
+    {
+        m_FgNhgs[fg_nhg_name].prefixes.push_back(ip_prefix);
+        m_fgNhgPrefixes[ip_prefix] = &m_FgNhgs[fg_nhg_name];
```
anish-n: This would cause issues for conflicting IP space,  ie 2 vnets with identical IP space containing consistent hashing. We may need to enhance this datastructure to be a 2 level map including vrf. OR what may be easier is that you can avoid putting anything into the m_fgNhgPrefixes, because this is primarily used by setFgNhg, validNextHopInNextHopGroup, and invalidNextHopInNextHopGroup

orchagent/fgnhgorch.cpp
```diff
@@ -1561,6 +1576,188 @@ bool FgNhgOrch::setFgNhg(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
...
+
+    std::string nhg_str;
+    for (const auto& member : nh_tunnel_ids)
+    {
+        if (!nhg_str.empty())
+        {
+            nhg_str += NHG_DELIMITER;
+        }
+        nhg_str += member.second.to_string(true, false);  // overlay_nh=true, srv6_nh=false
+        nhopgroup_members_set[member.second] = member.first;
```
anish-n: why are we passing in the reverse mapping as fn param map<sai_object_id_t, NextHopKey>& nh_tunnel_ids, and then converting it into the required mapping std::map<NextHopKey, sai_object_id_t> nhopgroup_members_set;. 

Why don't we just pass in nhopgroup_members_set directly?

orchagent/fgnhgorch.cpp
```diff
@@ -1561,6 +1576,188 @@ bool FgNhgOrch::setFgNhg(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
...
+    std::map<NextHopKey, sai_object_id_t> nhopgroup_members_set;
+
+    std::string nhg_str;
+    for (const auto& member : nh_tunnel_ids)
+    {
+        if (!nhg_str.empty())
+        {
+            nhg_str += NHG_DELIMITER;
+        }
+        nhg_str += member.second.to_string(true, false);  // overlay_nh=true, srv6_nh=false
```
anish-n: Why are we reconstructing the nhg_str/ nexthopgroupkey, why not just pass down the nexthopgroupkey because the caller has it already

orchagent/fgnhgorch.cpp
```diff
@@ -1561,6 +1576,188 @@ bool FgNhgOrch::setFgNhg(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
...
+        const auto& bucket_list = member.second;
+        if (!bucket_list.empty())
+        {
+            uint32_t bucket_idx = bucket_list[0];
+            if (bucket_idx < syncd_fg_entry.nhopgroup_members.size())
+            {
+                nhopgroup_member_ids[nhk] = syncd_fg_entry.nhopgroup_members[bucket_idx];
+            }
+        }
+    }
```
anish-n: why is this needed?

orchagent/fgnhgorch.cpp
```diff
@@ -1561,6 +1576,188 @@ bool FgNhgOrch::setFgNhg(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
...
+        {
+            if(m_FgNhgs[fg_nhg_name].next_hops.size() >= m_FgNhgs[fg_nhg_name].max_next_hops)
+            {
+                SWSS_LOG_WARN("Next-hop %s exceeds max_next_hops %d for prefix %s, skipping",
+                        nhk.to_string().c_str(), m_FgNhgs[fg_nhg_name].max_next_hops, ipPrefix.to_string().c_str());
+                continue;
+            }
+            FGNextHopInfo fg_nh_info = {0, "", LINK_UP}; // prefix_based match_mode uses single bank 0
+
+            m_FgNhgs[fg_nhg_name].next_hops[nhk.ip_address] = fg_nh_info;
```
anish-n: Future improvement, may not be needed in the current PR: We can have multiple nexthops containing the same nhk.ip_address, this is the case where we have VNET_ROUTE_TUNNEL with 2 nexthops containing the same nexthop IP address, but varying mac address/vni. I don't expect that it causes issues the way the code is written but this datastructure itself does not have much value in the vnet_route_tunnel case, given this we should probably clean it up and not use the next_hops table at all and substitute it with some type of check that makes the logic in the other functions where next_hops is used, work.

orchagent/fgnhgorch.cpp
```diff
@@ -1561,6 +1576,188 @@ bool FgNhgOrch::setFgNhg(sai_object_id_t vrf_id, const IpPrefix &ipPrefix, const
...
+            SWSS_LOG_INFO("Next-hop %s alias %s added to Fine Grained next-hop group member list for prefix %s",
+                    nhk.ip_address.to_string().c_str(), nhk.alias.c_str(), ipPrefix.to_string().c_str());
+        }
+
+        if (syncd_fg_route_entry_it == m_syncdFGRouteTables.at(vrf_id).end())
+        {
+            bank_member_changes[m_FgNhgs[fg_nhg_name].next_hops[nhk.ip_address].bank].
```
anish-n: no concept of multiple banks, so this ```m_FgNhgs[fg_nhg_name].next_hops[nhk.ip_address].bank``` can just be set to ```0```, this is a note for when we remove next_hops  population.

orchagent/vnetorch.cpp
```diff
@@ -933,6 +929,26 @@ bool VNetRouteOrch::removeNextHopGroup(const string& vnet, const NextHopGroupKey
     return true;
 }
 
+bool VNetRouteOrch::removeFgNextHopGroup(const string& vnet, const NextHopGroupKey &nexthops, const IpPrefix& ipPrefix, VNetVrfObject *vrf_obj)
+{
+    SWSS_LOG_ENTER();
+
+    sai_object_id_t vr_id = vrf_obj->getVRidIngress();
+    if (!gFgNhgOrch->removeFgNhg(vr_id, ipPrefix))
```
anish-n: There is a leak here, we are not cleaning up m_FgNhgs, m_fgNhgPrefixes etc for the vnet+ip-prefix

orchagent/vnetorch.cpp
```diff
@@ -3325,7 +3346,7 @@ bool VNetRouteOrch::handleTunnel(const Request& request)
     }
     if (vnet_orch_->isVnetExecVrf())
     {
-        return doRouteTask<VNetVrfObject>(vnet_name, ip_pfx, (has_priority_ep == true) ? nhg_primary : nhg, op, profile, monitoring, rx_monitor_timer, tx_monitor_timer, nhg_secondary, adv_prefix, monitors);
+        return doRouteTask<VNetVrfObject>(vnet_name, ip_pfx, (has_priority_ep == true) ? nhg_primary : nhg, op, profile, monitoring, rx_monitor_timer, tx_monitor_timer, nhg_secondary, adv_prefix, consistent_hashing_buckets, monitors);
```
anish-n: we don't intend to support the endpoint monitor, adv_prefix, secondary nhg etc features in conjunction with consistent_hashing, hence there is a choice to have a separate implementation of doRouteTask itself for consistent_hashing, and not introducing/reducing complexity for the very complicated existing doRouteTask function

orchagent/vnetorch.cpp
```diff
@@ -1268,73 +1329,75 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
...
-                    // but only remove the monitors for the set which has changed.
-                    if (it_route->second.primary != nexthops)
-                    {
-                        delEndpointMonitor(vnet, it_route->second.primary, ipPrefix);
-                    }
-                    if (it_route->second.secondary != nexthops_secondary)
+                    if (consistent_hashing_buckets > 0)
                     {
-                        delEndpointMonitor(vnet, it_route->second.secondary, ipPrefix);
+                        // only remove the inactive tunnel NHs as the FG NHG has already been updated
```
anish-n: does this support the case where an existing VNET_ROUTE_TUNNEL is updated to include consistent_hashing_buckets, so basically a tunnel route update asking for consistent hashing behavior?

anish-n (changes requested):

orchagent/vnetorch.cpp
NavdhaJindal: I chose not to have a separate implementation in order to avoid code repetition and allow the possibility of future features to coexist. However, after considering the added complexity of updating route tunnel to include consistent hashing, will defer to your suggestion and overload doroutetask 😃

orchagent/vnetorch.cpp
NavdhaJindal: this change was for update to the next hops of a route tunnel that already has consistent hashing, I didn't consider the case where route tunnel is updated to include consistent_hashing_buckets. 

Currently: have overloaded a new DoRouteTask for consistent hashing. This PR doesn't support route tunnel update asking for consistent hashing. Will handle that and also the edge case of an update that includes both 1) consistent hashing buckets, and 2) change in next hops, in a followup PR.

orchagent/fgnhgorch.cpp
NavdhaJindal: syncdContainsFgNhg is called in vnetorch to identify and remove fgnexthopgroup for delete case (since the del request won't contain consistent_hashing_buckets)

orchagent/fgnhgorch.cpp
NavdhaJindal: unclean code from when nhopgroup_members was constructed differently in vnetorch's addnexthopgroup. Removed, thanks!

orchagent/fgnhgorch.cpp
NavdhaJindal: passed as param instead

orchagent/fgnhgorch.cpp
NavdhaJindal: done, thanks

orchagent/fgnhgorch.cpp
NavdhaJindal: done, thanks

orchagent/fgnhgorch.cpp
NavdhaJindal: Was needed earlier to populate syncd_nexthop_groups_[vnet][nexthops].active_members as part of some checks in the existing DoRouteTask. This code is removed now that DoRouteTask has been overloaded

orchagent/vnetorch.cpp
NavdhaJindal: thanks so much for catching this! Fixed, cleanup done in removeFgNhgTunnel

orchagent/fgnhgorch.cpp
NavdhaJindal: next_hops is used in calculateBankHashBucketStartIndices and sprayBankNhgMembers. Will explore different key for next_hops instead of just ip address to avoid IP address conflict

orchagent/fgnhgorch.cpp
NavdhaJindal: thanks for catching! Changed the key to only be ip prefix when no vnet name (default vrf), and vnet + | + ip prefix otherwise

orchagent/fgnhgorch.cpp
NavdhaJindal: thanks for catching! Removed m_fgNhgPrefixes. Will add check in UT for conflicting IP space

orchagent/fgnhgorch.cpp
NavdhaJindal: removed vnet as param, passing key that is vnet | ipPrefix if non-default vnet, ipPrefix otherwise

orchagent/fgnhgorch.cpp
NavdhaJindal: verified case of routes for 2 vnets with conflicting IP space via UTs

orchagent/fgnhgorch.cpp
NavdhaJindal: used workaround logic wherever next_hops is used. Stopped populating next_hops in SetFgNhgTunnel

orchagent/vnetorch.cpp
```diff
@@ -1469,6 +1537,160 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
...
+        {
+            bool route_status = true;
+
+            if (it_route == syncd_tunnel_routes_[vnet].end())
+            {
+                route_status = add_route(vr_id, pfx, nh_id);
+            }
+            else
+            {
+                route_status = update_route(vr_id, pfx, nh_id);
```
anish-n: We should not be doing update_route as the nexthop id is unchanged in almost all cases, except the case where a route existed before and was non-fine grained ecmp and now became fine grained ecmp. Look at the behavior of isNextHopIdChanged for the non-vnet implementation as an example, we do not change the nexthop id if its just individual nexthops changing.

orchagent/vnetorch.cpp
```diff
@@ -1469,6 +1537,160 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
...
+            }
+        }
+        bool route_updated = false;
+        if (it_route != syncd_tunnel_routes_[vnet].end() && it_route->second.nhg_key != nexthops)
+        {
+            route_updated = true;
+            NextHopGroupKey nhg = it_route->second.nhg_key;
+
+            // In case of updating an existing route, decrease the reference count for the previous nexthop group
+            if (--syncd_nexthop_groups_[vnet][nhg].ref_count == 0)
```
anish-n: We will be decrementing the ref count even though the nexthopgroup still exists, resulting in unexpected behavior, this should tie into whether isNextHopIdChanged, if nexthopid did not change no ref count changes and no removeTunnelNextHop should occur.

NavdhaJindal: the SAI nexthopId will not change on nexthop modification, but the NextHopGroupKey used as key in `syncd_nexthop_groups_[vnet][nhg]` will change. Therefore,
1. old nhg's ref count must be decremented to 0
2. `get/createTunnelNextHop` was called for every nh in the new nexthopgroupkey, so `removeTunnelNextHop` must be called for every nh in the old nhg

orchagent/vnetorch.cpp
NavdhaJindal: added isNextHopIdChanged to only update route when nhid changes

orchagent/vnetorch.cpp
anish-n: Lets take this scenario as an example:
Create route with nexthop nh1, nh2, nh3
Update route with nexthop nh1, nh2

In this case you will do a removeTunnelNextHop for nh1 and nh2, and nh3 resulting in issues? I am surprised that the sonic-mgmt dataplane test does not run into an issue when this is tested

NavdhaJindal: - Create route with nexthop nh1, nh2, nh3 
  - selectFgNextHopGroup 
    - for nh in {nh1, nh2, nh3}:
      - vrf_obj->getTunnelNextHop(nh)
        - vxlan_orch->createNextHopTunnel
          - tunnel_obj->incNextHopRefCount
 **nh1, nh2 and nh3 refcount == 1**

- Update route with nexthop nh1, nh2
  - selectFgNextHopGroup
    - for each nh {nh1, nh2}:
      - vrf_obj->getTunnelNextHop(nh)
        - vxlan_orch->createNextHopTunnel
          - tunnel_obj->incNextHopRefCount
 **nh1, nh2 refcount == 2, nh3 refcount == 1**
  - if updating nhs of existing route:
    - for nh in prev nhg {nh1, nh2, nh3}:
      - vrf_obj->removeTunnelNextHop(nh)
  **nh1, nh2 refcount == 1, nh3 refcount == 0**

`createNextHopTunnel` is called for every nh in the nhg during create and update, therefore `removeTunnelNextHop` must be called for every nh in the old nhg, which then balances the ref count.

anish-n: From what I see, vrf_obj->removeTunnelNextHop does not actually check for ref count hence, it would blindly remove the nexthop and result in an issue? 
Code ref:
https://github.com/sonic-net/sonic-swss/blob/master/orchagent/vnetorch.cpp#L329C6-L329C19
https://github.com/sonic-net/sonic-swss/blob/master/orchagent/vxlanorch.cpp#L1450s

I agree that removeTunnelNextHop needs to occur for those nexthops which are removed, perhaps ref counting should be added to removeTunnelNextHop so that this is handled and it works as you described above.

orchagent/vnetorch.cpp
```diff
@@ -1469,6 +1536,160 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
...
+            vrf_obj->removeRoute(ipPrefix);
+        }
+        if (it_route == syncd_tunnel_routes_[vnet].end() || route_updated)
+        {
+            syncd_nexthop_groups_[vnet][nexthops].tunnel_routes.insert(ipPrefix);
+            VNetTunnelRouteEntry tunnel_route_entry;
+            tunnel_route_entry.nhg_key = nexthops;
+            tunnel_route_entry.primary = nexthops;
+            syncd_tunnel_routes_[vnet][ipPrefix] = tunnel_route_entry;
+            syncd_nexthop_groups_[vnet][nexthops].ref_count++;
```
anish-n: Lets say an update nexthop group occurs, you will decrement the ref count for old nexthop group to 0 and delete the individual nexthops, but you will not update the ref count for the new nexthopgroup at all because it will not come into this logic. Now 1) we should not delete individual nexthops, and 2) the nexthop group's ref count will become negative and will likely not get cleaned up. I think there are various ref count and update/delete issues I see, please can we enhance the ut to cover these as there might be other bug lurking in these areas, we need ASIC DB checks to make sure that the required cleanup for nexthops are occurring at the right point..

anish-n (changes requested):

orchagent/vnetorch.cpp
NavdhaJindal: vrf_obj->removeTunnelNextHop does check for ref count. 

Continuing the code ref: 
https://github.com/sonic-net/sonic-swss/blob/master/orchagent/vnetorch.cpp#L329C6-L329C19
https://github.com/sonic-net/sonic-swss/blob/master/orchagent/vxlanorch.cpp#L1450s
https://github.com/sonic-net/sonic-swss/blob/master/orchagent/vxlanorch.cpp#L634

orchagent/vnetorch.cpp
NavdhaJindal: if update nexthop group occurs,
route_updated = true at 
https://github.com/sonic-net/sonic-swss/pull/4158/changes#diff-9ec0e5825131dd3cf4689c3ace1822e193d39f18d3fd409313f9a76d7f3f8f24R1185-R1621

therefore it will come into this logic at https://github.com/sonic-net/sonic-swss/pull/4158/changes#diff-9ec0e5825131dd3cf4689c3ace1822e193d39f18d3fd409313f9a76d7f3f8f24R1639

ref count for new nhg will be incremented

Could you explain where the nhg's ref count will become negative?

orchagent/vnetorch.cpp
anish-n: Ok missed that, thanks for sharing that, looks good!

orchagent/vnetorch.cpp
anish-n: You are right, I cannot see this case upon another read of this code. Looks good to me.

orchagent/fgnhgorch.cpp
```diff
@@ -654,7 +712,19 @@ bool FgNhgOrch::setActiveBankHashBucketChanges(FGNextHopGroupEntry *syncd_fg_rou
                  * other available nhs, but for cases where # hash buckets is not
                  * divisible by # of nhs, simple round robin can make the hash bucket
                  * distribution non-ideal, thereby nhs can attract unequal traffic */
-                if (bank_fgnhg_map->at(*it).size() == exp_bucket_size)
+                if (bank_fgnhg_map->at(*it).size() > exp_bucket_size)
```
anish-n, line 753: Does the condition bank_fgnhg_map->at(*it).size() > exp_bucket_size occur because there are some  nexthops which exceed exp_bucket_size  due to some nhs acquiring more buckets due to num_nhs_with_one_more?

anish-n (approved):

prsunny:
@NavdhaJindal , can you update description as provided by copilot overview.

NavdhaJindal:
> @NavdhaJindal , can you update description as provided by copilot overview.

updated, thanks

orchagent/vnetorch.cpp
```diff
@@ -1472,6 +1543,160 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
     return true;
 }
 
+template<>
+bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipPrefix,
```
prsunny, line 1547: @NavdhaJindal , a lot of code got duplicated with the template addition. IMO, we can update the previous one to have this consistent hashing attribute and based on if its present, can diverge. The concern with this is if we fix some bug on original template, it may get missed here. @anish-n  for viz.

prsunny:
Overall, lgtm, just one comment on the template portion

orchagent/vnetorch.cpp
NavdhaJindal: thanks Prince, unified into a single DoRouteTask function

anish-n (approved):

orchagent/vnetorch.cpp
```diff
@@ -1186,187 +1344,370 @@ bool VNetRouteOrch::doRouteTask<VNetVrfObject>(const string& vnet, IpPrefix& ipP
...
+        bool isNextHopIdChanged = false;
+        NextHopGroupKey active_nhg("", true);
+
+        bool custom_monitor_ep_updated = false;
         std::map<NextHopKey, swss::IpAddress> origin_primary_monitors;
         std::map<NextHopKey, swss::IpAddress> origin_secondary_monitors;
-        if (custom_monitor_ep_updated)
+        bool is_custom_monitor_pinned_state_updated = false;
+
+        if (is_fg_route)
```
prsunny: Thanks for address the previous comment on consolidating the doRouteTask, @anish-n , @NavdhaJindal , but can we also move the fg changes to a separate function? I'm afraid too many changes here may impact other cases.

NavdhaJindal: have modularized the changes for fg routes into helper functions. It is hard to completely separate since there are different conditions based on if the route is changed from regular to fg and vice versa. Let me know if there's further improvements I can make

prsunny (dismissed):
lgtm, as there are common code changes for vnet routes, @zjswhhh for viz

zjswhhh (changes requested):
A same-key regular↔FG transition appears unsafe when the old NHG is shared (`ref_count > 1`). In `detectRouteTypeTransition`, the old map entry is erased; `selectFgNextHopGroup` can then create the FG entry under that same key; and `cleanupOldRouteNhg` restores the saved regular entry over it when another route still references the old group. This can orphan the FG SAI next-hop group, corrupt its CRM/accounting state, and leave tracking for the other route inconsistent. Please preserve both routes’ state correctly and add a test with two prefixes sharing an endpoint set while one transitions regular→FG.

prsunny:
@NavdhaJindal , could you address comments from Jing?

zjswhhh (approved):

anish-n (approved):

PR #4161, opened 2026-01-29
[202511] [portsorch] DOM config change causes interface link to flap (#4056)

mihirpat1:
202511 Cherry-pick for https://github.com/sonic-net/sonic-swss/pull/4056/changes

This PR fixes a bug where DOM (transceiver) configuration changes cause unexpected interface link flaps due to unnecessary FEC reconfiguration. The issue was introduced in PR #3529 which refactored port configuration handling but inadvertently removed the logic to set the m_fec_cfg flag after applying FEC configuration at runtime in doPortTask(), while it was still being set during initial port creation in addPortBulk().

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -4739,11 +4739,6 @@ void PortsOrch::doPortTask(Consumer &consumer)
                             p.m_alias.c_str(), pCfg.speed.value
                         );
                     }
-                    else
-                    {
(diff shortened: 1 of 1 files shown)


mihirpat1:
The changes are already merged via https://github.com/sonic-net/sonic-swss/pull/4108

PR #4164, opened 2026-01-29
[P4Orch] Update watchport to fetch port oper-status from PortsOrch instead of SAI

ksravani-hcl:
**What I did**

**Why I did it**

**How I verified it**
Verified with UT testcases.

**Details if related**


tests/p4rt/test_l3.py
yunang-c, line 1452: It seems to miss a few lines here

PR #4166, opened 2026-01-29
Port configuration not applied on non-autoneg capable ports

Keshavg-marvell:
**What I did**
Fix the port configuration not getting applied to non-autoneg capable port

Allow orchagent to continue applying remaining port configurations (e.g., admin state and speed) when auto‑negotiation is disabled on non‑auto‑negotiation‑capable ports. This ensures the ASIC is programmed correctly and the port can transition to the expected operational state.

**Why I did it**
When auto‑negotiation is configured off on a non‑auto‑negotiation‑capable port, orchagent incorrectly blocks processing of subsequent port attributes. As a result, other configurations applied alongside auto‑negotiation—such as admin state or speed—are not programmed into the ASIC. This leads to the port remaining down even when it should be operational.

**How I verified it**
Verified on the box that it is applying the port configuration correctly on ASIC and port is coming up.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -3366,6 +3366,11 @@ bool PortsOrch::setGearboxPortAttr(const Port &port, dest_port_type_t port_type,
                         attr.id = SAI_PORT_ATTR_SPEED;
                         attr.value.u32 = speed;
                     }
+                    else
+                    {
+                        SWSS_LOG_NOTICE("BOX: Speed %d not supported for port %s", speed, port.m_alias.c_str());
+                        return false;
+                    }
                     SWSS_LOG_NOTICE("BOX: Set %s lane %s %d", port.m_alias.c_str(), speed_attr.c_str(), speed);
                     break;
                 case SAI_PORT_ATTR_MTU:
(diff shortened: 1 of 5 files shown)


orchagent/portsorch.cpp
```diff
@@ -4585,10 +4585,22 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+                                it = taskMap.erase(it);
+                                continue;
+                            }
+                            else
+                            {
+                                // Auto-negotiation config "off" is valid for ports that do not support AN.
+                                // In this case, we should continue retrying other port configurations
+                                // (e.g., admin state and speed). Set m_an_cfg to prevent AN programming
+                                // attempts in subsequent retries.
+                                m_portList[p.m_alias].m_an_cfg = true;
```
prsunny, line 4675: IMO, this approach seems incorrect. @prgeor , please review

Keshavg-marvell: @prsunny @pregeor
We have a platform where both auto‑negotiation‑capable and non‑capable ports coexist. For auto‑negotiation‑capable ports, the default requirement is to keep auto‑negotiation disabled. During default configuration generation
(sonic-cfggen -H -k <> --preset t1), the auto‑negotiation config("autoneg": "off/on") is generated for all ports, which results in non‑auto‑negotiation‑capable ports also receiving an auto‑negotiation configuration.
While having auto‑negotiation set to off for non‑capable ports is logically valid, this scenario is not handled correctly in the current code and this change addresses this issue.

PR #4168, opened 2026-01-29
Increase Vrf pool to 4k

ypcisco:
**What I did**

Increase VRF pool size from 999 to 4096 to support hardware platforms with higher VRF scale capabilities.

**Why I did it**

Current VRF pool is limited to 999 instances (VRF_TABLE_END=2000, VRF_TABLE_START=1001) which artificially restricts VRF scale even on hardware platforms that support more VRFs.

**How I verified it**

**Details if related**

### DIFF
--- a/cfgmgr/vrfmgr.cpp
+++ b/cfgmgr/vrfmgr.cpp
@@ -10,9 +10,9 @@
 #include "warm_restart.h"
 
 #define VRF_TABLE_START 1001
-#define VRF_TABLE_END 2000
+#define VRF_TABLE_END 5097
 #define TABLE_LOCAL_PREF 1001 // after l3mdev-table
-#define MGMT_VRF_TABLE_ID 5000
+#define MGMT_VRF_TABLE_ID 6000
 #define MGMT_VRF          "mgmt"
 
 using namespace swss;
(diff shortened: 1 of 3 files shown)


anish-n (approved):

anish-n:
Please can we also backport to 202511?

cfgmgr/vrfmgr.cpp
```diff
@@ -10,9 +10,9 @@
 #include "warm_restart.h"
 
 #define VRF_TABLE_START 1001
-#define VRF_TABLE_END 2000
+#define VRF_TABLE_END 5097
 #define TABLE_LOCAL_PREF 1001 // after l3mdev-table
-#define MGMT_VRF_TABLE_ID 5000
```
venkatmahalingam: Hope you'll take care of updating the new mgmt vrf table id in the following code.
https://github.com/sonic-net/sonic-buildimage/blob/master/files/image_config/interfaces/interfaces.j2#L12

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/tests/sample_output/py3/mvrf_interfaces#L34

venkatmahalingam (commented):
Please address the comments.

cfgmgr/vrfmgr.cpp
ypcisco: Have updated the current PR and created following PRs to track the updated mgmt vrf table id : 
https://github.com/sonic-net/sonic-buildimage/pull/25504
https://github.com/sonic-net/sonic-utilities/pull/4279

Aravind-Subbaroyan:
@venkatmahalingam - could you please review the code? The comments are addressed

venkatmahalingam (approved):

PR #4180, opened 2026-01-30
[cfgmgr]: Add kernel interface and route for VNET_ROUTE_TUNNEL

bveerannamsft:
**What I did**
For Vxlan tunnel routes on a VNET (VNET_ROUTE_TUNNEL), the routes are programmed on NPU only and the kernel is not aware of the vxlan interface or the routes. Any CPU generated packets (like ping, bgp etc) to this VxLAN endpoint is dropped in the kernel. In this PR, a new flag called `install_on_kernel` is added on the VNET_ROUTE_TUNNEL object in CONFIG_DB. If this flag is set to true, the route and the vxlan interface are created on the kernel in addition to adding to the NPU.

**Why I did it**
Design reference: https://github.com/sonic-net/SONiC/pull/2092

For SONiC to establish BGP over VxLAN tunnel in a VNET, the kernel should be aware of the vxlan interface and routes. The routes added in VNET_ROUTE_TUNNEL should also be present in kernel for the CPU generated packets to be forwarded by the kernel. If these routes are not present, the kernel will drop any VxLAN packet to/from the CPU. 

Please refer the design PR for more information on the usecase.

**How I verified it**
Verified on Virtual Switch and Hardware device

**Details if related**

### DIFF
--- a/cfgmgr/vnetmgr.cpp
+++ b/cfgmgr/vnetmgr.cpp
@@ -1,0 +1,673 @@
+#include <unistd.h>
+#include <algorithm>
+#include <regex>
+#include <sstream>
+#include <string>
+#include <net/if.h>
+
+#include "logger.h"
+#include "producerstatetable.h"
+#include "macaddress.h"
+#include "vnetmgr.h"
+#include "exec.h"
+#include "tokenize.h"
+#include "shellcmd.h"
+#include "warm_restart.h"
+#include <swss/logger.h>
+
+using namespace std;
+using namespace swss;
+
+// Fields name
+#define VXLAN_TUNNEL "vxlan_tunnel"
+#define SOURCE_IP "src_ip"
+#define SOURCE_MAC "src_mac"
+#define MAC_ADDRESS "mac_address"
+#define ENDPOINT "endpoint"
+#define INSTALL_ON_KERNEL "install_on_kernel"
+#define VNI "vni"
+#define VNET "vnet"
+#define VXLAN "vxlan"
+#define VXLAN_NAME_PREFIX "Vxlan"
+#define VXLAN_SRC_PORT "vxlan_sport"
+#define SWITCH "switch"
+
+#define RET_SUCCESS 0
+
+// Commands
+
+static int cmdCreateVxlan(const swss::VNetMgr::VxlanKernelRouteInfo & info, std::string & res)
+{
+    // ip link add {{VXLAN}} [address {{SOURCE MAC}}] type vxlan id {{VNI}} [local {{SOURCE IP}}] [remote {{DEST IP}}] dstport 4789
+    ostringstream cmd;
+    cmd << IP_CMD " link add "
+        << shellquote(info.m_vxlanDevName);
+
+    if (!info.m_srcMac.empty())
+    {
+        cmd << " address " << shellquote(info.m_srcMac);
+    }
+
+    cmd << " type vxlan id "
+        << shellquote(info.m_vni);
+
+    if (!info.m_srcIp.empty())
+    {
+        cmd << " local " << shellquote(info.m_srcIp);
+    }
+
+    if (!info.m_dstIp.empty())
+    {
+        cmd << " remote " << shellquote(info.m_dstIp);
+    }
+    cmd << " dstport " << shellquote(info.m_vxlanSrcUdpPort);
+    return swss::exec(cmd.str(), res);
+}
+
+static int cmdDeleteVxlan(const swss::VNetMgr::VxlanKernelRouteInfo & info, std::string & res)
+{
+    // ip link del {{VXLAN}}
+    ostringstream cmd;
+    cmd << IP_CMD " link del "
+        << shellquote(info.m_vxlanDevName);
+    return swss::exec(cmd.str(), res);
+}
+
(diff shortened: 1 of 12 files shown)


prsunny:
@anish-n , please review, @bveerannamsft , could you address copilot comments

prsunny:
@bveerannamsft , please resolve conflicts

cfgmgr/vnetmgr.cpp
```diff
@@ -0,0 +1,673 @@
...
+#define VXLAN "vxlan"
+#define VXLAN_NAME_PREFIX "Vxlan"
+#define VXLAN_SRC_PORT "vxlan_sport"
+#define SWITCH "switch"
+
+#define RET_SUCCESS 0
+
+// Commands
+
+static int cmdCreateVxlan(const swss::VNetMgr::VxlanKernelRouteInfo & info, std::string & res)
```
anish-n, line 39: Could we test how scalable is this solution. This is likely to create a lot of netdevs and kernel config, such that there will be an ip link per vni+srcip+dstip combo and seems like this solution would not scale very well.

cfgmgr/vnetmgr.cpp
```diff
@@ -0,0 +1,673 @@
...
+std::string VNetMgr::getVxlanSourcePort()
+{
+    std::vector<FieldValueTuple> temp;
+
+    if (m_appSwitchTable.get(SWITCH, temp))
+    {
+        auto itr = std::find_if(
+            temp.begin(),
+            temp.end(),
+            [](const FieldValueTuple &fvt) { return fvt.first == VXLAN_SRC_PORT; });
```
anish-n, line 207: This should be vxlan_port and not vxlan_src_port

cfgmgr/vnetmgr.cpp
```diff
@@ -0,0 +1,673 @@
...
+    auto it = m_kernelRouteTunnelCache.find(vxlanRouteInfo.m_routeName);
+    if (it == m_kernelRouteTunnelCache.end())
+    {
+        SWSS_LOG_INFO("Vxlan route %s does not exists", vxlanRouteInfo.m_routeName.c_str());
+        return true;
+    }
+    
+    std::string vxlanDevName = VXLAN_NAME_PREFIX + vxlanRouteInfo.m_vni;
+    std::string res;
+    int ret = cmdDeleteVxlan(it->second, res);
```
anish-n, line 656: This logic will delete the vxlan netdev with name Vxlan<vni>. If there are 2+ routes on this netdev and we delete 1 out of those routes, it will prematurely delete the whole Vxlan<vni> netdev. Could you please address this scenario and add testing for it?

cfgmgr/vnetmgr.cpp
```diff
@@ -0,0 +1,673 @@
...
+        return false;
+    }
+
+    std::string vxlanDevName = VXLAN_NAME_PREFIX + vxlanRouteInfo.m_vni;
+
+    auto it = m_vxlanNetDevices.find(vxlanDevName);
+    if (it != m_vxlanNetDevices.end())
+    {
+        SWSS_LOG_INFO("Vxlan device %s already present", it->first.c_str());
+        //return false;
```
anish-n, line 576: Should return true here? Leaving it as commented out or returning false will cause the operation to retry forever without success. Commenting it out is an issue because the next operation is a create vxlan device/route which will fail with "RTNETLINK answers: File exists".

orchagent/vnetorch.cpp
```diff
@@ -1826,6 +1794,17 @@ bool VNetRouteOrch::handleRoutes(const Request& request)
     SWSS_LOG_INFO("VNET-RT '%s' op '%s' for ip %s", vnet_name.c_str(),
                    op.c_str(), ip_pfx.to_string().c_str());
 
+    auto nextHops = ip_addresses.getIpAddresses();
+    auto nextHop = nextHops.begin()->to_string();
```
anish-n, line 1798: I am seeing a crash here when testing, some locally connected routes do not have a nexthop.ipaddresses, and this dereferences a non-existing entry using to_string. Similarly I believe DEL calls will also cause a crash.

cfgmgr/vnetmgr.cpp
```diff
@@ -0,0 +1,673 @@
...
+    m_kernelRouteTunnelCache.erase(it);
+
+    auto dev = m_vxlanNetDevices.find(vxlanDevName);
+    if (dev != m_vxlanNetDevices.end())
+    {
+        m_vxlanNetDevices.erase(dev);
+    }
+
+    return true;
+}
```
anish-n, line 673: This is not config reload safe and can leak entires/if we want to change the kernel programming model it is not safe to change because it will not clear out the Kernel programmed interface upon reload, please check vxlanmgrd: https://github.com/sonic-net/sonic-swss/blob/master/cfgmgr/vxlanmgr.cpp#L214, https://github.com/sonic-net/sonic-swss/blob/master/cfgmgr/vxlanmgr.cpp#L221C5-L221C25

anish-n (changes requested):

anish-n:
There is no code coverage included in PR, needs unit test and python based e2e test with vs

PR #4184, opened 2026-02-02
[P4Orch] Enable multicast on ingress RIFs.

divyagayathri-hcl:
**What I did**
To enable multicast on ingress RIFs.

**Why I did it**

**How I verified it**
Verified with UT cases.

**Details if related**

### DIFF
--- a/orchagent/p4orch/tests/router_interface_manager_test.cpp
+++ b/orchagent/p4orch/tests/router_interface_manager_test.cpp
@@ -91,6 +91,12 @@ std::unordered_map<sai_attr_id_t, sai_attribute_value_t> CreateRouterInterfaceAt
     attr_value.u32 = mtu;
     attr_list[SAI_ROUTER_INTERFACE_ATTR_MTU] = attr_value;
 
+    attr_value.booldata = true;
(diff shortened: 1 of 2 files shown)


orchagent/p4orch/tests/router_interface_manager_test.cpp
```diff
@@ -272,17 +296,25 @@ class RouterInterfaceManagerTest : public ::testing::Test
     void AddRouterInterfaceEntry(P4RouterInterfaceEntry &router_intf_entry, const sai_object_id_t port_oid,
                                  const uint32_t mtu, const bool sub_port = false, const sai_object_id_t vlan_oid = 0)
     {
-        EXPECT_CALL(mock_sai_router_intf_,
-                    create_router_interface(
-                        ::testing::NotNull(), Eq(gSwitchId), sub_port ? Eq(6) : Eq(5),
```
yunang-c, line 277: Since we are only change the values in Eq(*), let's not re-format the entire function. Same for all changes below.

Also, why do we want to format this function? If possible, we should follow Microsoft C++ style in clang-format as the rest of the file?

divyagayathri-hcl: Done.

PR #4189, opened 2026-02-02
Gracefully handle case when speed is not supported in setGearboxPortAttr

arista-hpandya:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/25311

**What I did**
Fixed a bug in setGearboxPortAttr() where an uninitialized sai_attribute_t attr variable could be passed to the SAI API when isSpeedSupported() returns false for SAI_PORT_ATTR_SPEED operations.

The fix adds an else block that logs the unsupported speed and returns false early, preventing the SAI call with uninitialized data

**Why I did it**
This causes the following error logs to be printed:
```
	 E               2026 Jan 17 21:41:16.925578 ctn102 ERR gbsyncd#syncd: :- processQuadEvent: attr: SAI_PORT_ATTR_SPEED: 400000
	 E
	 E               2026 Jan 17 21:41:16.925694 ctn102 ERR swss#orchagent: :- setGearboxPortAttr: BOX: Failed to set Ethernet72 port attribute 31
	 E
	 E               2026 Jan 17 21:41:16.925694 ctn102 ERR swss#orchagent: :- handleSaiFailure: Encountered failure in set operation, SAI API: SAI_API_PORT, status: SAI_STATUS_INVALID_ATTR_VALUE_0
	 E
	 E               2026 Jan 17 21:41:16.932105 ctn102 ERR swss#orchagent: :- meta_generic_validation_set: unable to find attribute metadata SAI_OBJECT_TYPE_PORT:378307824
```

**How I verified it**
Verified that the fix prevents the uninitialized variable from being used in a SAI call

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -3352,6 +3352,11 @@ bool PortsOrch::setGearboxPortAttr(const Port &port, dest_port_type_t port_type,
                         attr.id = SAI_PORT_ATTR_SPEED;
                         attr.value.u32 = speed;
                     }
+                    else
+                    {
+                        SWSS_LOG_NOTICE("BOX: Speed %d not supported for port %s", speed, port.m_alias.c_str());
+                        return false;
+                    }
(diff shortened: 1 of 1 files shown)


rlhui:
@arpit-nexthop

PR #4193, opened 2026-02-03
[countersyncd] fix otel actor log level 

Janetxxx:
**What I did**
- Wrapping with log::log_enabled!(log::Level::Debug) avoids the cost of building log output unless Debug is enabled.
- Changing print_otel_metrics() to Debug level prevents noisy logs in normal operation; it only shows up when debugging.
- Removing print_to_console simplifies config and relies solely on log levels to control output.
**Why I did it**
Avoids the cost of building log output unless Debug is enabled.
**How I verified it**
Verified in internal setup.
**Details if related**

### DIFF
--- a/crates/countersyncd/src/main.rs
+++ b/crates/countersyncd/src/main.rs
@@ -176,14 +176,6 @@ struct Args {
     )]
     otel_endpoint: String,
 
-    /// Enable OpenTelemetry console output
-    #[arg(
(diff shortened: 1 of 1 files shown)


crates/countersyncd/src/main.rs
Pterosaur: Should we delete it as well?

PR #4199, opened 2026-02-05
[vnet/vxlan] Dualstack vxlan support on vnet

sridkulk:
**What I did**
This PR adds support for multiple vxlan tunnels on vnets. The tunnels can have either an IPv4 underlay or IPv6 underlay and so it enables vxlan tunnel routes over IPv4 and IPv6 underlay simultaneously

- Added a new parameter `Vxlan_tunnel_list` under Vnet configuration that takes a list of vxlan tunnel names
- Enhanced VnetOrch/VnetRouteOrch to handle tunnel next hop creation based on the destination type (IPv4/IPv6) provided in `VNET_ROUTE_TUNNEL` configurations
 
**Why I did it**

Currently, VNet supports only one vxlan tunnel and it can be either on IPv4 underlay or IPv6 underlay.  As a result, in a single vnet, only IPv4 traffic or IPv6 traffic is supported over Vxlan and not both. 

**How I verified it**
Verified on virtual switch and cisco hardware

**Details if related**

HLD: https://github.com/sonic-net/SONiC/pull/2093

### DIFF
--- a/cfgmgr/vxlanmgr.cpp
+++ b/cfgmgr/vxlanmgr.cpp
@@ -21,6 +21,7 @@ extern MacAddress gMacAddress;
 
 // Fields name
 #define VXLAN_TUNNEL "vxlan_tunnel"
+#define VXLAN_TUNNEL_LIST "vxlan_tunnel_list"
 #define SOURCE_IP "src_ip"
 #define VNI "vni"
 #define VNET "vnet"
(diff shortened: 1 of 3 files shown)


anish-n (changes requested):
There is no code coverage included in PR, needs unit test and python based e2e test with vs

PR #4204, opened 2026-02-09
[countersyncd]: Fix cargo bench compiling error

Pterosaur:
**What I did**
Remove deprecated parameters.

**Why I did it**
The parameter, print_to_console, has been removed, but the CI will not run any bench program, so this error was ignored.

**How I verified it**
Run cargo locally

**Details if related**


Pterosaur:
Retrigger 202511

PR #4205, opened 2026-02-09
[P4Orch] Add ACL support to redirect to L2 multicast group.

ksravani-hcl:
**What I did**

**Why I did it**

**How I verified it**
Verified with UT cases.

**Details if related**


yunang-c (approved):
The last commit 1cf3e1cb3f64c15e2cff06446bfee51fc191d74e LGTM.

PR #3814, opened 2025-08-07
[FC] Set FC delay in command line parameters

stepanblyschak:
**What I did**

I implemented a configurable delay mechanism for flex counter processing in the SONiC orchagent. The changes include:
- Flex counters can be delayed not just for warm or fast boot but for cold boot as well
- Added a new command line parameter -D to set the flex counter delay in seconds
- Replaced the hardcoded 60-second delay with a configurable global variable gFlexCounterDelaySec
- Modified the flex counter initialization logic to only create delay timers when the delay is greater than 0
- Updated unit tests to test both immediate processing (delay=0) and delayed processing (delay=120) scenarios

In previous implementation, the delay was applicable to warm and fast boot, but never done for cold boot. 
This PR addresses that to improve cold boot time as well.
I left 0 sec delay as a default mainly due to VS test suite expecting FC counters to be set immediately.

**Why I did it**

The flex counter delay feature improves system boot time by prioritizing data plane configuration over counter initialization. Systems with many ports, priority groups (PGs), and queues require significant time to generate counter maps, which is not immediately necessary during boot. 

**How I verified it**

- Updated unit tests to cover both immediate (delay=0) and delayed (delay=120 seconds) processing scenarios
- Ensured the command line parameter parsing works correctly with proper error handling for invalid values
- Verified that the delay timer is only created when delay > 0, maintaining backward compatibility
- Added logging to show the configured delay value for debugging purposes

**Details if related**

The feature is particularly beneficial for large-scale deployments where boot time optimization is critical for operational efficiency.

PR #4207, opened 2026-02-10
[orchagent] Fix getopt string for -R option (ring_thread_enabled regression)

Chiranjeevi-U-ML:
**What I did**
Fixed a typo in the getopt option string in `orchagent/main.cpp` where the `-R` option was incorrectly defined as requiring a mandatory argument (`R:`) instead of being a simple flag (`R`).

The fix changes line 403 from:
```cpp
while ((opt = getopt(argc, argv, "b:m:r:f:j:d:i:hsz:k:q:c:t:v:I:R:D:M")) != -1)
```
to:
```cpp
while ((opt = getopt(argc, argv, "b:m:r:f:j:d:i:hsz:k:q:c:t:v:I:RD:M")) != -1)
```

**Why I did it**

When ring_thread_enabled is set to true in DEVICE_METADATA, orchagent fails to start with the error:
`/usr/bin/orchagent: option requires an argument -- 'R'`

This bug was introduced in PR #3814 (specifically, commit 5de59222) when the -D option was added. 

The getopt string was changed from `I:R` to `I:R:D:` instead of the correct `I:RD:`. 

This accidentally added a colon after R, making it require a mandatory argument.

**How I verified it**

I did a code review and verified that the -R case handler does not use `optarg` and simply sets `gRingMode = true`. Also, I have checked the git history and traced the bug introduction to PR #3814 and confirmed the original implementation in PR #3242 was correct.

Check the code lines here:

Original PR link:
https://github.com/sonic-net/sonic-swss/pull/3242/changes#diff-3cba33a216c0ea2b942dc03b2638fa27a40a8b6eb81abf9f83f13e1ee7831ac4R360

Bug Link:
https://github.com/sonic-net/sonic-swss/pull/3814/changes#diff-3cba33a216c0ea2b942dc03b2638fa27a40a8b6eb81abf9f83f13e1ee7831ac4R373

**Details if related**

Fixes: sonic-net/sonic-buildimage#25366


dgsudharsan:
@Chiranjeevi-U-ML Don't we have a UT in buildimage or vstest to catch these failures? If possible can we add UT?

dgsudharsan (changes requested):
Since this is a basic change where we will not even start orchagent if the parameter is wrong, I suggest we need to add coverage through UT .

Chiranjeevi-U-ML:
@dgsudharsan Added an integration test `test_ring_thread` in `tests/test_zmq.py` to cover the -R flag.  It verifies that orchagent starts successfully with the -R flag (no argument required).

prsunny:
@dgsudharsan , can you sign off?

PR #4208, opened 2026-02-10
[P4Orch]Update P4RT RIF MTU when port MTU changes.

ksravani-hcl:
**What I did**

**Why I did it**

**How I verified it**
Verified with UT cases.

**Details if related**

### DIFF
--- a/tests/p4rt/test_l3.py
+++ b/tests/p4rt/test_l3.py
@@ -2129,6 +2129,123 @@ class TestP4RTL3(object):
         )
         self._clean_vrf(dvs)
 
+    @pytest.mark.skip(reason="sairedis vs MY MAC support is not ready")
+    def test_RouterInterfaceMtuUpdate(self, dvs, testlog):
+        self._set_up(dvs)
+        cdb = swsscommon.DBConnector(4, dvs.redis_sock, 0)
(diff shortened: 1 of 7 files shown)


tests/test_sub_port_intf.py
```diff
@@ -390,6 +390,10 @@ def get_newly_created_oid(self, table, old_oids):
 
     def get_default_vrf_oid(self):
         oids = self.get_oids(ASIC_VIRTUAL_ROUTER_TABLE)
+        print(f"before VRF OIDs: {oids}")
```
mint570: The internal CL does not include this file change

anbu1234: > The internal CL does not include this file change

This is added for debugging purpose but we will not upstream it.

tests/p4rt/test_l3.py
```diff
@@ -2129,6 +2129,123 @@ def test_P4rtNotification(self, dvs, testlog):
         )
         self._clean_vrf(dvs)
 
+    @pytest.mark.skip(reason="sairedis vs MY MAC support is not ready")
```
yunang-c, line 2132: Is there a TODO to track when we should enable this test?

PR #4217, opened 2026-02-14
[swssconfig] Add custom ZMQ endpoint for DPU Orchagent

vivekrnv:
**What I did**

On DPU, orchagent starts on eth0-midplane, need an option to provide a custom endpoint for swssconfig to work on DPU

**Why I did it**

This is needed to directly trigger the updates to DASH_HA_* and DASH_FLOW_SYNC* tables in orchagent. 

swssconfig already has the ability to write to Zmq endpoint, this change adds support to write to a custom endpoint. 

One use case is this CLI: https://github.com/sonic-net/sonic-utilities/pull/4278 which directly triggers Flow dump invocation from DPU directly

**How I verified it**

```
# cat /etc/sonic/ha/ha_config.json
[
    {
        "DASH_HA_SCOPE_TABLE:vdpu0_0": {
            "version": "1",
            "ha_role": "standby",
            "ha_set_id": "haset0_0"
        },
        "OP": "SET"
    }
]

# swssconfig -e tcp://169.254.200.1 /etc/sonic/ha/ha_config.json
#

# tail -f /var/log/swss/sairedis.rec
|c|SAI_OBJECT_TYPE_HA_SCOPE:oid:0x12008000000027|SAI_HA_SCOPE_ATTR_HA_SET_ID=oid:0x11008000000026|SAI_HA_SCOPE_ATTR_DASH_HA_ROLE=SAI_DASH_HA_ROLE_STANDBY
|s|SAI_OBJECT_TYPE_ENI:oid:0x7008000000025|SAI_ENI_ATTR_HA_SCOPE_ID=oid:0x12008000000027
```

**Details if related**


dgsudharsan (approved):

zjswhhh (approved):
lgtm

liat-grozovik (approved):

prsunny:
@vivekrnv , can you explain why this change is required? is it in the normal functional path? Please provide details in description

vivekkarri:
> @vivekrnv , can you explain why this change is required? is it in the normal functional path? Please provide details in description

Updated

swssconfig/swssconfig.cpp
```diff
@@ -66,9 +68,12 @@ shared_ptr<ProducerStateTable> get_table(unordered_map<string, shared_ptr<Produc
     return p_table;
 }
 
-bool write_db_data(vector<KeyOpFieldsValuesTuple> &db_items, set<string>  &zmq_tables, std::shared_ptr<ZmqClient> zmq_client)
+bool write_db_data(vector<KeyOpFieldsValuesTuple> &db_items, set<string>  &zmq_tables, std::shared_ptr<ZmqClient> zmq_client, bool use_custom_endpoint)
 {
-    DBConnector db("APPL_DB", 0, false);
+    // If custom endpoint is used, it's for DPU Orchagent - use DPU_APPL_DB
+    // Otherwise use APPL_DB
+    string db_name = use_custom_endpoint ? "DPU_APPL_DB" : "APPL_DB";
```
zjswhhh, line 75: @vivekrnv  - We should add option to select database container as there are more than one DPU?

Also NPU database config doesn't have DPU_* database config.

prsunny: @vivekrnv , can you check this comment? is this a blocker ?

vivekkarri: I don't think so.

- First to apply config from NPU to a specific DPU Orchagent use it's eth0-midplane IP. 
- Also, this tool was always meant to write APPL_DB and now DPU_APPL_DB, Why would we need the ability to write to other DB's

zjswhhh: It's not a blocker. 

But @vivekrnv - eth0-midplane IP is the ZMQ endpoint right? For each DPU there is a DPU_APPL_DB, all hosted on NPU (distinguished by port number), the tool doesn't know which one to write to.

vivekkarri: Ah, got you. that's correct. To support writing data from NPU, we need to specify the DPU so that DBConnector object can be initialized correctly.

Existing usecase of this tool is in Flow Dump CLI, which is run from DPU and directly uses DBConnector to initialize to DPU_APPL_DB.

Any specific use case why we need this ability to work from NPU

PR #4227, opened 2026-02-17
[P4Orch] Make room for new 'no_action' in the L3MulticastManager

bibhuprasad-hcl:
**What I did**
Make room for new 'no_action' in the L3MulticastManager

**Why I did it**
The use of 'no_action' is to bring in support for L2 multicast groups. The action no_action will signal that a bridge port object is going to be created to be used by a L2 multicast group.

**How I verified it**
Verified with UT cases.

**Details if related**

### DIFF
--- a/orchagent/p4orch/l3_multicast_manager.cpp
+++ b/orchagent/p4orch/l3_multicast_manager.cpp
@@ -432,7 +432,9 @@ L3MulticastManager::deserializeMulticastRouterInterfaceEntry(
     const auto& field = fvField(it);
     const auto& value = fvValue(it);
     if (field == p4orch::kAction) {
-      if (value != p4orch::kSetMulticastSrcMac) {
+      if (value == p4orch::kSetMulticastSrcMac || value == p4orch::kNoAction) {
+        router_interface_entry.action = value;
+      } else {
         return ReturnCode(StatusCode::SWSS_RC_INVALID_PARAM)
                << "Unexpected action " << QuotedVar(value) << " in "
                << APP_P4RT_MULTICAST_ROUTER_INTERFACE_TABLE_NAME;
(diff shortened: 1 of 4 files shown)


orchagent/p4orch/l3_multicast_manager.h
```diff
@@ -27,17 +27,20 @@ struct P4MulticastRouterInterfaceEntry {
...
   swss::MacAddress src_mac;
+  std::string action;
   std::string multicast_metadata;
   sai_object_id_t router_interface_oid = SAI_OBJECT_TYPE_NULL;
 
   P4MulticastRouterInterfaceEntry() = default;
   P4MulticastRouterInterfaceEntry(const std::string& port,
                                   const std::string& instance,
                                   const swss::MacAddress& mac,
+				  const std::string& action,
```
yunang-c: nit: please ensure the indent is consistent. same as L43

orchagent/p4orch/l3_multicast_manager.cpp
```diff
@@ -432,7 +432,9 @@ L3MulticastManager::deserializeMulticastRouterInterfaceEntry(
     const auto& field = fvField(it);
     const auto& value = fvValue(it);
     if (field == p4orch::kAction) {
-      if (value != p4orch::kSetSrcMac) {
+      if (value == p4orch::kSetSrcMac || value == p4orch::kNoAction) {
```
yunang-c: The Google CL uses kSetMulticastSrcMac ("set_multicast_src_mac") instead of kSetSrcMac ("set_src_mac"). We should probably upstream the CL 100490 first to add kSetMulticastSrcMac? @mint570

orchagent/p4orch/l3_multicast_manager.h
bibhuprasad-hcl: > nit: please ensure the indent is consistent. same as L43

Indent is fixed.

orchagent/p4orch/l3_multicast_manager.cpp
bibhuprasad-hcl: > The Google CL uses kSetMulticastSrcMac ("set_multicast_src_mac") instead of kSetSrcMac ("set_src_mac"). We should probably upstream the CL 100490 first to add kSetMulticastSrcMac? @mint570

kSetSrcMac  is replaced with kSetMulticastSrcMac . taken CL 83175 which has the original change.

PR #4234, opened 2026-02-19
copporch: trap port exclusion

bhouse-nexthop:
**What I did**

SONiC automatically installs all the traps listed in `copp_cfg.j2` for all interfaces.

This causes issues in special circumstances such as sonic-mgmt testbed fanouts where the traps will prevent LACP and LLDP packets flowing from the test server to the DUT as they are trapped rather than forwarded.

While in theory one could probably disable all the traps in `copp_cfg.j2`, traps may still be needed on other ports, especially if using VXLAN-EVPN between the fanout and root fanout.

**Why I did it**

The SAI supports an attribute of `SAI_HOSTIF_TRAP_ATTR_EXCLUDE_PORT_LIST` to be able to exclude ports. This uses a new `COPP_TRAP_EXCLUDE_PORTS` table that takes a list of ports to exclude from all traps. This is a fairly simple approach and didn't see a use-case to be able to specify an exclusion list per trap. It simply maintains and updates the list of excluded ports on change per hostif trap.

**How I verified it**

Config_db example:

```
"COPP_TRAP_EXCLUDE_PORTS": {
    "GLOBAL": {
        "ports": [ "Ethernet0", "Ethernet4" ]
    }
},
```

On broadcom platforms to verify, you can remove all ACLs so `bcmcmd show fp` lists only traps.  Then change the port exclusions and observe the port bitmap changes accordingly in the output.

**Details if related**

Needs https://github.com/sonic-net/sonic-buildimage/pull/25571

Fixes https://github.com/sonic-net/sonic-buildimage/issues/25570

HLD update https://github.com/sonic-net/SONiC/pull/2229

### DIFF
--- a/orchagent/copporch.cpp
+++ b/orchagent/copporch.cpp
@@ -510,10 +510,13 @@ bool CoppOrch::applyAttributesToTrapIds(sai_object_id_t trap_group_id,
         attr.value.s32 = trap_id;
         attrs.push_back(attr);
 
-        attr.id = SAI_HOSTIF_TRAP_ATTR_EXCLUDE_PORT_LIST;
-        attr.value.objlist.list = excluded_port_ids.data();
-        attr.value.objlist.count = static_cast<uint32_t>(excluded_port_ids.size());
-        attrs.push_back(attr);
+        if (!excluded_port_ids.empty())
+        {
+            attr.id = SAI_HOSTIF_TRAP_ATTR_EXCLUDE_PORT_LIST;
+            attr.value.objlist.list = excluded_port_ids.data();
+            attr.value.objlist.count = static_cast<uint32_t>(excluded_port_ids.size());
+            attrs.push_back(attr);
+        }
 
         attrs.insert(attrs.end(), trap_id_attribs.begin(), trap_id_attribs.end());
 
@@ -968,11 +971,13 @@ void CoppOrch::excludePortsSet(const std::string &ports)
     for (auto port : portList)
     {
         Port p;
-        if (!gPortsOrch->getPort(port, p)) {
+        if (!gPortsOrch->getPort(port, p))
+        {
             SWSS_LOG_ERROR("Port %s not found", port.c_str());
             continue;
         }
-        if (p.m_type != Port::PHY) {
+        if (p.m_type != Port::PHY)
+        {
             SWSS_LOG_ERROR("Port %s is not a physical port", port.c_str());
             continue;
         }
@@ -1008,9 +1013,9 @@ void CoppOrch::doConfigdbTask(Consumer &consumer)
         string tableAttr = kfvKey(t);
         string op = kfvOp(t);
 
-        if (tableAttr != "GLOBAL")
+        if (table_name != CFG_COPP_TRAP_EXCLUDE_PORTS_TABLE_NAME || tableAttr != "GLOBAL")
         {
-            SWSS_LOG_ERROR("COPP_TRAP_EXCLUDE_PORTS unknown attribute %s", tableAttr.c_str());
+            SWSS_LOG_ERROR("%s unknown attribute %s", table_name.c_cstr(), tableAttr.c_str());
             goto consume;
         }
 
@@ -1062,9 +1067,12 @@ void CoppOrch::doTask(Consumer &consumer)
         return;
     }
 
-    if (db_name == "CONFIG_DB") {
+    if (db_name == "CONFIG_DB")
+    {
         doConfigdbTask(consumer);
-    } else {
+    }
+    else
+    {
         doAppdbTask(consumer);
     }
 }


bhouse-nexthop:
This build will fail until https://github.com/sonic-net/sonic-buildimage/pull/25571 merges as it needs the auto-generated  define based on the yang models for `CFG_COPP_TRAP_EXCLUDE_PORTS_TABLE_NAME`

bhouse-nexthop:
@dgsudharsan @prsunny any issue with this implementation methodology ?

orchagent/copporch.cpp
```diff
@@ -933,6 +934,141 @@ void CoppOrch::doTask(Consumer &consumer)
...
+{
+    SWSS_LOG_ENTER();
+    string db_name = consumer.getDbName();
+
+    if (!gPortsOrch->allPortsReady())
+    {
+        return;
+    }
+
+    if (db_name == "CONFIG_DB") {
```
prsunny: { -> new line

prsunny:
@bhouse-nexthop , please update the copp hld with this implementation/design.

Another question, is the exclude list configurable? Meaning, can we dynamically add/remove ports in exclude list?

prsunny:
@dgsudharsan , please review

orchagent/copporch.cpp
bhouse-nexthop: ah, yeah, I see a couple formatting issues, I'll fix those up.

bhouse-nexthop:
> @bhouse-nexthop , please update the copp hld with this implementation/design.
> 
> Another question, is the exclude list configurable? Meaning, can we dynamically add/remove ports in exclude list?

Yes, the exclude list is dynamically reconfigurable with GCU.

orchagent/copporch.cpp
rminnikanti: How about doing table specific check, instead of db_name check, for consistency with other Orch's?
I think most of the places checks are table name specific but not db_name.

orchagent/copporch.cpp
```diff
@@ -509,6 +510,11 @@ bool CoppOrch::applyAttributesToTrapIds(sai_object_id_t trap_group_id,
         attr.value.s32 = trap_id;
         attrs.push_back(attr);
 
+        attr.id = SAI_HOSTIF_TRAP_ATTR_EXCLUDE_PORT_LIST;
```
rminnikanti: Can you help understand why to set SAI_HOSTIF_TRAP_ATTR_EXCLUDE_PORT_LIST unconditionally here?

1. applyAttributesToTrapIds() gets called also during init.  excluded_port_ids would be empty in this case.
2. The logic assumes that all vendors support SAI_HOSTIF_TRAP_ATTR_EXCLUDE_PORT_LIST.
If a particular vendor’s SAI does not implement this attribute, then even the trap creation will fail.

In my view, SAI_HOSTIF_TRAP_ATTR_EXCLUDE_PORT_LIST must be set only if CONFIG_DB entry is pushed, but if there are any other cases it is expected to be set without CONFIG_DB entry then, vendor capability check of this attribute is required (I would appreciate input from others on this).

orchagent/copporch.cpp
rminnikanti, line 500: Applying exclude port list on traps across feature disable/enable must be validated.

I suggest adding a DVS test to verify this flow; including validation of the new CONFIG_DB table flow in general:
https://github.com/sonic-net/sonic-swss/blob/master/tests/test_copp.py

orchagent/copporch.cpp
bradh352: copporch listens on both APPDB and CONFIGDB, so this check is to determine which db the event is for. Since in theory table names could overlap between APPDB and CONFIGDB I didn't think checking the table name was sufficient (even though in this case it would be since I know there is no overlap).

orchagent/copporch.cpp
bradh352: You're right, it should only be set if there are entries instead of setting it blank.  i'll update that logic.  i'm not sure if there are any scenarios where the port list may be populated before this is called, especially if traps themselves get modified at runtime, so I think I still need to set it here, just not if its blank.

orchagent/copporch.cpp
bradh352: fixed

bhouse-nexthop: I added an explicit table name check in doConfigdbTask() in case additional tables are added in the future.

orchagent/copporch.cpp
bhouse-nexthop: modified to not set it if the list is blank.

bhouse-nexthop:
> @bhouse-nexthop , please update the copp hld with this implementation/design.

HLD updated here: https://github.com/sonic-net/SONiC/pull/2229

orchagent/p4orch/tests/wcmp_manager_test.cpp
```diff
@@ -267,7 +268,13 @@ class WcmpManagerTest : public ::testing::Test
         EXPECT_CALL(mock_sai_hostif_, create_hostif_table_entry(_, _, _, _)).WillRepeatedly(Return(SAI_STATUS_SUCCESS));
         EXPECT_CALL(mock_sai_hostif_, create_hostif_trap(_, _, _, _)).WillOnce(Return(SAI_STATUS_SUCCESS));
         EXPECT_CALL(mock_sai_switch_, get_switch_attribute(_, _, _)).WillOnce(Return(SAI_STATUS_SUCCESS));
-        copp_orch_ = new CoppOrch(gAppDb, APP_COPP_TABLE_NAME);
+        std::vector<std::string> appCoppOrchTables = {
+            APP_COPP_TABLE_NAME
+        };
+        std::vector<std::string> cfgCoppOrchTables = {
```
prsunny, line 274: IMO, this must be handled by coppmgr and pass it to orchagent via APP_DB based on coppmgr validation. There are also WB implications of CoPP tables in APP_DB, so, keeping copporch to just consume APP_DB would be better.

bhouse-nexthop: @prsunny can you point me to a reference so I can better understand why this would need to flow through APP_DB first to mitigate potential warm reboot issues?  I'm not intimately familiar with warm reboot, but reading through https://github.com/sonic-net/SONiC/blob/master/doc/warm-reboot/swss_warm_restart.md and https://github.com/sonic-net/SONiC/blob/master/doc/warm-reboot/sai_redis_api_idempotence.md leads me to believe APP_DB is mostly needed if there are external influencers outside of CONFIG_DB that might need to record changes (such as from FRR or the linux kernel itself) to be applied ... 

If the request is just because a lot of other copporch stuff relies on APP_DB and is a request simply from a code consistency standpoint, I'd argue that it would be a more complex implementation and would opt for the more simplistic solution (this).

But if there is an actual implication, then I definitely need to understand the underlying nature so I can make sure any code changes I make properly address that.   If there's a guide of some sort on how to ensure any changes are warm reboot capable, I'd love to read up on it.

Thanks!

orchagent/p4orch/tests/acl_manager_test.cpp
```diff
@@ -904,7 +904,13 @@ class AclManagerTest : public ::testing::Test
         EXPECT_CALL(mock_sai_hostif_, create_hostif_table_entry(_, _, _, _)).WillRepeatedly(Return(SAI_STATUS_SUCCESS));
         EXPECT_CALL(mock_sai_hostif_, create_hostif_trap(_, _, _, _)).WillOnce(Return(SAI_STATUS_SUCCESS));
         EXPECT_CALL(mock_sai_switch_, get_switch_attribute(_, _, _)).WillRepeatedly(Return(SAI_STATUS_SUCCESS));
-        copp_orch_ = new CoppOrch(gAppDb, APP_COPP_TABLE_NAME);
+        std::vector<std::string> appCoppOrchTables = {
+            APP_COPP_TABLE_NAME
+        };
+        std::vector<std::string> cfgCoppOrchTables = {
```
dgsudharsan, line 910: As suggested by others please  use appl_db table instead of directly subscribing to cfg_db.

orchagent/copporch.cpp
```diff
@@ -933,6 +937,146 @@ void CoppOrch::doTask(Consumer &consumer)
     }
 }
 
+void CoppOrch::excludedPortsUpdate()
```
dgsudharsan, line 940: Why do we want to do it globally, instead of having it per trap whe SAI supports it per trap. I would recommend adding this field to existing COPP_TRAP or COPP_GROUP table if we want to set it for list of traps rather than introducing a different table. We can reuse the existing flow to set it as an attribute in existing flow.

orchagent/p4orch/tests/acl_manager_test.cpp
bhouse-nexthop: ok, I'll make that conversion

orchagent/copporch.cpp
bhouse-nexthop: Can you think of an actual use-case where you'd want some traps enabled and some disabled in sonic on a per-port basis?  I thought it would simply cruft up the configuration too much to give that kind of flexibility.

I'd probably do the inverse of what you suggest though.  I would make each member of the copp trap exclude list be able to take a list of traps to exclude, if none are provided, default to all.  That way the most common use case is easy ...

dgsudharsan: The problem I see here is we are tailoring the definition of a new table for specific fanout based use case which is not how a feature is supposed to be defined. When SAI exposes per trap exclusion it makes sense to provide such capability in config_db. 
While there may not be a scenario right now, think of how we can handle when such a requirement arises. It would mean that we have to make code changes again.
Instead if you design it like per trap exclude list, you can achieve your use case by defining the appropriate config_db entries (although it may be more now). This design will be future proof as well.
I will request @prsunny  to provide his thoughts on this.

prsunny: I tend to agree with @dgsudharsan as this can be useful in future.  For e.g, if I dont want DHCP trap to be enabled on some ports, I can have this excluded for those if the exclude list is per COPP_GROUP.

PR #4236, opened 2026-02-19
[P4Orch] Support update, validate, and verify for no_action action entries.

bibhuprasad-hcl:
What I did
Support update, validate, and verify for no_action action entries.

Why I did it
Support update, validate, and verify for no_action action entries.

How I verified it
Verified with UT cases.


bibhuprasad-hcl:
> Let's also combine it into PR #4230

This PR branched out from #4231. shall we merge it with #4231, as that is the parent branch?

PR #4254, opened 2026-02-24
Add Trixie build

saiarcot895:
**What I did**

Build swss for Trixie, and fix compilation errors that have popped up (largely due to armhf being compiled with 64-bit file offset and time support in Debian Trixie).

The changes to the log statements in `natorch.cpp` fall under the above category; however, with the default configuration, there are no tests that specifically test that code path.

**Why I did it**

This will be needed for moving containers in sonic-buildimage to Trixie.

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/port/portcnt.h
+++ b/orchagent/port/portcnt.h
@@ -210,7 +210,7 @@ public:
             bool is_set = false;
         } custom_collection; // Port serdes custom_collection
 
-    } serdes; // Port serdes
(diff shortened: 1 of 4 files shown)


prsunny:
@saiarcot895 , can you provide details as to why code coverage cannot be met for this PR?

saiarcot895:
For `natorch.cpp`, the code coverage for the file is low (37%), which suggests there aren't many existing tests that are testing the functionality in this file (beyond some initialization). In addition, the log statements in question appear to be related to creating NAT entries and updating how often they are hit by the ASIC, which requires some infra work on the testing side.

PR #4264, opened 2026-02-26
[P4Orch] Enable copp trap group for all queues

ksravani-hcl:
**What I did**

**Why I did it**

**How I verified it**
   Verified with UT test Cases

**Details if related**

### DIFF
--- a/orchagent/p4orch/tests/acl_manager_test.cpp
+++ b/orchagent/p4orch/tests/acl_manager_test.cpp
@@ -1,6 +1,5 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
-
 #include <map>
 #include <memory>
 #include <nlohmann/json.hpp>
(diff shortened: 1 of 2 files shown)


orchagent/p4orch/tests/acl_manager_test.cpp
```diff
@@ -3157,7 +3156,7 @@ TEST_F(AclManagerTest, AclRuleWithColorPacketActionsButWithRateLimit) {
   auto acl_rule = GetAclRule(kAclIngressTableName, acl_rule_key);
   ASSERT_NE(nullptr, acl_rule);
   // Check action field value
-  EXPECT_EQ(gUserDefinedTrapStartOid + queue_num - P4_CPU_QUEUE_MIN_NUM + 1,
+  EXPECT_EQ(gUserDefinedTrapStartOid + queue_num - P4_CPU_QUEUE_MIN_NUM ,
```
mint570, line 3159: Why are there changes in the test?
The internal cl does not have test change.

anbu1234: We updated the test case based on the change in value of P4_CPU_QUEUE_MIN_NUM  from 1 to 0. Since the value is changed from 1 to 0, the additional plus one in the test case is not required and results in incorrect value.

PR #4265, opened 2026-02-26
Temporarily moving the PORT_PHY_ATTR ERR messages to NOTICE level.

dhanasekar-arista:
**What I did**
Suppressed the noisy ERR logs by moving them to NOTICE level logs.

**Why I did it**
For vendors who didn't implement the new attributes yet, the following log errors will be generated for all ports: ERR swss#orchagent: :- verifyPortSupportsAllPhyAttr: PORT_PHY_ATTR: Port Ethernet0 does not support RX_SIGNAL_DETECT attribute (status=-196608)

The ideal fix is to query only supported set of attributes in m_supported_phy_attrs. We are working on this , in the meantime to unblock others , we are moving this to NOTICE level log.

**How I verified it**

**Details if related**

Fixes: 
https://github.com/sonic-net/sonic-buildimage/issues/25619
https://github.com/sonic-net/sonic-swss/issues/4242

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -8993,7 +8993,7 @@ bool PortsOrch::verifyPortSupportsAllPhyAttr(sai_object_id_t port_id, const char
     sai_status_t status = sai_port_api->get_port_attribute(port_id, 1, &port_attr);
     if (status != SAI_STATUS_BUFFER_OVERFLOW)
     {
-        SWSS_LOG_ERROR("PORT_PHY_ATTR: Port %s does not support RX_SIGNAL_DETECT attribute (status=%d)",
+        SWSS_LOG_NOTICE("PORT_PHY_ATTR: Port %s does not support RX_SIGNAL_DETECT attribute (status=%d)",
                       port_name, status);
         return false;
     }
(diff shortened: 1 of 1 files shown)


dhanasekar-arista:
FYI, [noaOrMlnx](https://github.com/noaOrMlnx), [rustiqly](https://github.com/rustiqly) , @prgeor

dhanasekar-arista:
this is a simple change of ERR to NOTICE, the PR check failures are irrelevant to the change.

noaOrMlnx:
@dhanasekar-arista have you considered updating the code to check if the vendor supports the feature using sai_query_attribute_capability()?

dhanasekar-arista:
> @dhanasekar-arista have you considered updating the code to check if the vendor supports the feature using sai_query_attribute_capability()?

@noaOrMlnx `sai_query_attribute_capability` is already used in `void PortsOrch::queryPortPhyAttrCapabilities` to build m_supported_phy_attrs, but the code is not robust enough to handle this scenario.
We are working on enhancing this to be more robust, in the meantime if this we wanted to unblock others by reducing the log severity.

prsunny:
@dhanasekar-arista , please title the PR reflecting the changes.

dhanasekar-arista:
This is a temporary change of log level from ERR to NOTICE level until fix is added. Can we bypass the checks for code-coverage and get this merged?

orchagent/portsorch.cpp
```diff
@@ -8993,7 +8993,7 @@ bool PortsOrch::verifyPortSupportsAllPhyAttr(sai_object_id_t port_id, const char
     sai_status_t status = sai_port_api->get_port_attribute(port_id, 1, &port_attr);
     if (status != SAI_STATUS_BUFFER_OVERFLOW)
```
prgeor, line 8994: @dhanasekar-arista I dont understand why we check for buffer overflow status

dhanasekar-arista: AFAIK, this is the only known way to validate if the port supports this attribute or not.

PR #4274, opened 2026-02-28
Fix orchagent 100% CPU spin when gOrchUnhealthy is set

### LINKED ISSUE #4273: orchagent busy-wait loop in cSONiC containers when PortInitDone is never set
## Description

On the SONiC Virtual Switch (VS) platform, `orchagent` occasionally enters a tight CPU spin loop in its CRM (Counter Resource Monitoring) thread, consuming 100% of a CPU core indefinitely. The issue is observed in `docker-sonic-vs` containers used as cSONiC neighbors in sonic-mgmt testbeds, but likely affects any VS deployment.

## Symptoms

- `orchagent` process pegged at ~100% CPU
- `rsyslogd` also elevated (~55% CPU) due to log flooding from orchagent
- Container overall CPU usage ~158% (orchagent + rsyslogd combined)
- Occurs on some containers but not all, even in the same testbed (suggests a timing/race condition at startup)

## Root Cause Analysis

The CRM thread in orchagent continuously polls SAI resource availability via `object_type_get_availability()`. On the VS SAI implementation, many of these queries return `SAI_STATUS_NOT_SUPPORTED`. The CRM thread appears to retry without any backoff or rate-limiting, resulting in a tight poll loop.

Evidence from `sairedis.rec` on an affected container shows continuous rapid-fire queries:

```
2026-02-28.08:28:59.277910|q|object_type_get_availability|SAI_OBJECT_TYPE_SWITCH:oid:0x21000000000000|SAI_ROUTE_ENTRY_ATTR_IP_ADDR_FAMILY=SAI_IP_ADDR_FAMILY_IPV4
2026-02-28.08:28:59.278251|Q|object_type_get_availability|SAI_STATUS_NOT_SUPPORTED|COUNT=0
2026-02-28.08:28:59.278499|q|object_type_get_availability|SAI_OBJECT_TYPE_SWITCH:oid:0x21000000000000|SAI_ROUTE_ENTRY_ATTR_IP_ADDR_FAMILY=SAI_IP_ADDR_FAMILY_IPV6
2026-02-28.08:28:59.278704|Q|object_type_get_availability|SAI_STATUS_NOT_SUPPORTED|COUNT=0
... (repeating continuously)
```

## Reproduction

1. Deploy a sonic-mgmt KVM testbed with cSONiC neighbors (`-k csonic`)
2. After `add-topo` completes, check CPU usage: `docker stats --no-stream`
3. Some cSONiC containers will show ~158% CPU; inside, `orchagent` is at 100%

In our test, VM0100 and VM0101 were healthy (0.2% CPU), while VM0102 and VM0103 were affected (158% CPU). The only difference is startup timing.

## CPU Evidence

```
VM0100 orchagent:  0.0% CPU,  00:00:01 cumulative
VM0101 orchagent:  0.0% CPU,  00:00:01 cumulative
VM0102 orchagent: 99.9% CPU,  00:45:00 cumulative (in ~45 min uptime)
VM0103 orchagent: 99.9% CPU,  00:45:00 cumulative (in ~45 min uptime)
```

## Environment

- Platform: VS (docker-sonic-vs)
- Usage: cSONiC neighbors in sonic-mgmt KVM testbed
- Host OS: Ubuntu with KVM/Docker

## Suggested Fix

The CRM polling loop should:
1. Skip or cache results for resource types that return `SAI_STATUS_NOT_SUPPORTED`
2. Add a backoff/sleep between polling cycles to prevent tight loops
3. Optionally, disable CRM polling entirely on VS platform where resource counters are not meaningful

## Impact

- Wastes ~2-3 CPU cores per testbed (when multiple neighbors are affected)
- Floods syslog, making it harder to debug real issues
- Functional impact is low — BGP and data plane still work correctly

## Related

- sonic-net/sonic-swss#189 (different orchagent CPU issue — unresolved next hops)

securely1g:
## Description

Fix orchagent consuming 100% CPU when `gOrchUnhealthy` is permanently set after a SAI API failure. Two root causes addressed:

1. **Syslog spam**: The error log was called every select loop iteration (~80K/sec), causing rsyslogd to also consume ~55% CPU. Fixed by moving the log into the existing `SELECT_TIMEOUT` periodic block (once per second).

2. **epoll_wait spin**: When `gOrchUnhealthy` is set, internal `Selectable` objects remain in a "ready" state, causing `epoll_wait()` to return immediately with timeout=0 instead of blocking for `SELECT_TIMEOUT` (1000ms). This causes the select loop to spin at ~160K iterations/sec. Fixed by adding an explicit `sleep_for(SELECT_TIMEOUT)` when both `gOrchUnhealthy` is set and select returns `TIMEOUT`.

### Root Cause
PR #4040 introduced `gOrchUnhealthy` flag that gets set permanently when SAI hostif creation fails during startup. On SONiC VS (docker-sonic-vs) used as cSONiC neighbors in testbeds, a race condition between orchagent and syncd causes this failure reliably on some containers.

### Impact Before Fix
- orchagent: **100% CPU** (epoll_wait spin at 160K calls/sec)
- rsyslogd: **~55% CPU** (processing ~37K syslog messages/sec)
- System effectively unusable for its intended purpose

### Impact After Fix
- orchagent: **0% CPU** (select loop blocks properly for 1 second intervals)
- rsyslogd: **0% CPU** (1 log message per second)

### Validation
- Tested on SONiC KVM testbed with 4 cSONiC neighbors (docker-sonic-vs)
- Before fix: VM0102/VM0103 at 100% CPU
- After fix: All 4 neighbors at 0% CPU
- Confirmed via `strace`: `epoll_wait` now blocks with 1000ms timeout
- All 4 BGP sessions Established and stable

### Changes
- `orchagent/orchdaemon.cpp`: Add `sleep_for(SELECT_TIMEOUT)` when `gOrchUnhealthy && ret == TIMEOUT`
- `orchagent/orchdaemon.cpp`: Move `gOrchUnhealthy` error log into `SELECT_TIMEOUT` periodic block
- `orchagent/orchdaemon.cpp`: Add `#include <thread>` for `std::this_thread::sleep_for`
- `tests/mock_tests/test_failure_handling.cpp`: Unit test for `gOrchUnhealthy` flag behavior

Fixes: #4273


securely1g:
cc @prabhataravind — this fixes a rate-limiting gap in the `gOrchUnhealthy` logging you introduced in #4040. The intent was periodic logging but it fired on every select iteration. Would appreciate your review.

orchagent/orchdaemon.cpp
```diff
@@ -931,12 +932,14 @@ void OrchDaemon::start(long heartBeatInterval)
         ret = m_select->select(&s, SELECT_TIMEOUT);
 
         /*
-         * Log an error message periodically if a previous SAI API call failed with
-         * an unrecoverable error.
+         * When gOrchUnhealthy is set, the Select event loop may return
+         * immediately with timeout=0 (epoll_wait spin) because internal
+         * Selectable objects remain in a "ready" state after the SAI failure.
```
lguohan: why is selectable objects will remain in a "ready" state after the SAI failure?

securely1g: Good question — I investigated further with strace.

When `gOrchUnhealthy` is set, the orchagent `epoll_wait` is called with **timeout=0** (non-blocking) on every iteration, returning immediately ~160K times/sec:

```
epoll_wait(44, [], 161, 0) = 0   // repeated endlessly
```

A healthy orchagent on the same host shows the expected behavior:
```
epoll_wait(44, [{events=EPOLLIN, ...}], 161, 1000) = 1   // blocks up to 1000ms
```

Looking at `swss-common/common/select.cpp`, `Select::select()` first calls `poll_descriptors(c, 0)` (non-blocking check), and only falls through to `poll_descriptors(c, timeout)` (blocking with 1000ms) if the first call returns `TIMEOUT`. 

The root cause appears to be that after the SAI failure, the select loop processes the error path and returns before reaching the blocking `poll_descriptors` call. The exact mechanism of why `poll_descriptors(c, 0)` does not return `TIMEOUT` in the unhealthy case requires further investigation in the Selectable state machine, but the observed behavior is consistent and reproducible.

The `sleep_for(SELECT_TIMEOUT)` is a defensive fix to prevent CPU spin regardless of the underlying cause — it ensures the loop sleeps for 1 second when the system is already in an unrecoverable error state (`gOrchUnhealthy`), where tight polling provides no benefit.

lguohan: i still do not understand. can you find why select always return immediately without timeout?

securely1g:
**Validated on cSONiC testbed (docker-sonic-vs) — all 4 neighbors now under 1% CPU**

Before fix (VM0102 and VM0103 had `gOrchUnhealthy=true`):
```
CONTAINER ID   NAME                   CPU %     MEM USAGE / LIMIT     MEM %     NET I/O   BLOCK I/O         PIDS
csonic_vms6-1_VM0100   0.21%     ...
csonic_vms6-1_VM0101   0.21%     ...
csonic_vms6-1_VM0102   100.40%   ...
csonic_vms6-1_VM0103   100.68%   ...
```

After fix (commit `f6565a54`):
```
CONTAINER ID   NAME                   CPU %     MEM USAGE / LIMIT     MEM %     NET I/O           BLOCK I/O         PIDS
f400d44de4c5   csonic_vms6-1_VM0103   1.24%     131.5MiB / 62.51GiB   0.21%     0B / 0B           77.8kB / 12.2MB   106
3e4a5dc74e86   csonic_vms6-1_VM0102   0.99%     131.5MiB / 62.51GiB   0.21%     0B / 0B           217kB / 12.2MB    106
0dff16a184da   csonic_vms6-1_VM0101   1.02%     131.6MiB / 62.51GiB   0.21%     0B / 0B           77.8kB / 12.2MB   104
e5f781c2fed0   csonic_vms6-1_VM0100   1.06%     135.7MiB / 62.51GiB   0.21%     0B / 0B           3.87MB / 12.2MB   104
cfb301d4d5f9   ptf_vms6-1             13.04%    506.2MiB / 32GiB      1.54%     0B / 0B           65.5kB / 35.4MB   24
62904555b5b0   net_vms6-1_VM0103      0.00%     1.078MiB / 62.51GiB   0.00%     0B / 0B           0B / 0B           1
9eb6b932df53   net_vms6-1_VM0102      0.00%     1.082MiB / 62.51GiB   0.00%     0B / 0B           0B / 0B           1
c940e14c1eba   net_vms6-1_VM0101      0.00%     1.078MiB / 62.51GiB   0.00%     0B / 0B           0B / 0B           1
8ed7ccc3d61a   net_vms6-1_VM0100      0.00%     1.086MiB / 62.51GiB   0.00%     0B / 0B           307kB / 0B        1
15eeb3ba3395   sonic-mgmt             0.00%     31.96MiB / 62.51GiB   0.05%     23.4MB / 1.27GB   55.4MB / 226MB    2
```

**Note:** Initial version had `if (gOrchUnhealthy && ret == Select::TIMEOUT)` which never triggered because `Select::select()` returns `Select::OBJECT` (not `TIMEOUT`) — `poll_descriptors(c, 0)` phase 1 always finds ready Selectables after the SAI failure. Updated to unconditional `if (gOrchUnhealthy)` in commit `f6565a54`.

orchagent/orchdaemon.cpp
securely1g: I traced through the code — here is the root cause of the spin:

**`Select::select()` in `swss-common/common/select.cpp:167`:**
```cpp
int Select::select(Selectable **c, int timeout, ...) {
    ret = poll_descriptors(c, 0);        // phase 1: non-blocking
    if (ret != Select::TIMEOUT || timeout == 0)
        return ret;                       // exits here if m_ready not empty
    ret = poll_descriptors(c, timeout);   // phase 2: blocking (never reached)
}
```

**`poll_descriptors()` at line 155:**
```cpp
if (sel->hasCachedData()) {
    m_ready.insert(sel);   // re-insert if selectable has buffered data
}
```

So `m_ready` stays non-empty → phase 1 always returns `OBJECT` → the blocking phase 2 (`epoll_wait` with 1000ms timeout) is never reached.

**Why `m_ready` stays non-empty after SAI failure:**

When `handleSaiFailure()` is called (`saihelper.cpp:762`), it:
1. Sets `gOrchUnhealthy = true`
2. Calls `sai_redis_switch_attr_notify_syncd` (`SAI_REDIS_NOTIFY_SYNCD_INVOKE_DUMP`)
3. This writes to the Redis notification channel

The orchagent main loop processes the `Select::OBJECT` return and calls `Orch::execute()`, but the orch does not fully consume the selectable data in the error path. The selectable retains cached/buffered data (Redis consumer has unconsumed entries), so `hasCachedData()` returns true, and `m_ready` is perpetually non-empty.

**The proper fix would be in `Select::select()` itself** — using the blocking phase 2 even when `m_ready` is non-empty. However, that would change behavior for all selectables across all daemons. The defensive sleep in `orchdaemon.cpp` is scoped to the already-unrecoverable `gOrchUnhealthy` state where tight polling has no benefit.

Would you prefer a fix in `Select::select()` instead?

securely1g:
@lguohan — gentle ping. The detailed root cause analysis is in the inline thread: `Select::poll()` uses `timeout=0` when `m_ready` is non-empty, and failed SAI operations leave selectables in the ready set permanently. This causes 100% CPU on any VS/cSONiC container with `gOrchUnhealthy=true`.

The fix rate-limits the retry loop by yielding back to the select with proper timeout when unhealthy. Happy to discuss further or adjust the approach.

orchagent/orchdaemon.cpp
```diff
@@ -931,12 +932,15 @@ void OrchDaemon::start(long heartBeatInterval)
...
         /*
-         * Log an error message periodically if a previous SAI API call failed with
-         * an unrecoverable error.
+         * When gOrchUnhealthy is set, the Select event loop spins with
+         * epoll_wait(timeout=0) because poll_descriptors phase 1 always
+         * finds "ready" Selectables (the notification channel stays readable
+         * after the SAI failure), so the blocking phase 2 poll is never
+         * reached. This causes 100% CPU. Sleep to throttle the loop.
          */
         if (gOrchUnhealthy)
```
prabhataravind: we could additionally check here if ret also did not return timeout("&& ret != Select::TIMEOUT") to avoid a sleep of additional 1000ms when select actually went to phase2 (though that is probably not a common scenario).

prabhataravind (dismissed):

orchagent/orchdaemon.cpp
securely1g: Good call — updated. Now checks `ret != Select::TIMEOUT` so we only sleep when select didn't already block in phase 2. Thanks for the review and approval!

prabhataravind (approved):
lgtm

PR #4276, opened 2026-03-01
[P4Orch] Add support to set a VLAN ID in the router_interface_table.

divyagayathri-hcl:
**What I did**
To Add support to set a VLAN ID in the router_interface_table.

**Why I did it**

**How I verified it**
Verified with UT cases.

**Details if related**


orchagent/response_publisher.cpp
yunang-c, line 1: What is changed in this file? It seems to be a pure formatting change and it's not related to the PR.

mint570: +1
Please undo the changes on the two response publisher files.

mint570 (approved):

PR #4280, opened 2026-03-02
[P4Orch] Change log message level for setLabel() when the label already exists & Increase the warmboot timer for teamsyncd from 70s to 120s.

ksravani-hcl:
**What I did**

**Why I did it**

**How I verified it**
Verified with UT cases

**Details if related**

### DIFF
--- a/teamsyncd/teamsync.h
+++ b/teamsyncd/teamsync.h
@@ -13,7 +13,7 @@
 
 #define TEAMSYNCD_APP_NAME  "teamsyncd"
 // seconds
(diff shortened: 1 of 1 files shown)


teamsyncd/teamsync.h
```diff
@@ -13,7 +13,7 @@
 
 #define TEAMSYNCD_APP_NAME  "teamsyncd"
 // seconds
-const uint32_t DEFAULT_WR_PENDING_TIMEOUT = 70;
+const uint32_t DEFAULT_WR_PENDING_TIMEOUT = 120;
```
mint570, line 16: This change will need community review

PR #4285, opened 2026-03-02
Extract serdes programming logic into dedicated function

bgallagher-nexthop:
In the below pull request, Copilot suggested two improvements that were not resolved before the PR was merged.

Addressing those two suggestions with this commit.

https://github.com/sonic-net/sonic-swss/pull/4113

**What I did**
- Refactored programming of signal integrity serdes settings for a port into a lambda helper function so the same logic can be re-used for both gearbox line-side, gearbox system-side and non-gearbox ports.
- Augmented a unit test mock stub to correctly clear state.

**Why I did it**

Copilot's comments were valid.

**How I verified it**

Ran the portsorch unit tests in `tests/mock_tests/portsorch_ut.cpp`.

### DIFF
--- a/cfgmgr/vrfmgr.cpp
+++ b/cfgmgr/vrfmgr.cpp
@@ -10,9 +10,9 @@
 #include "warm_restart.h"
 
 #define VRF_TABLE_START 1001
-#define VRF_TABLE_END 2000
+#define VRF_TABLE_END 5097
 #define TABLE_LOCAL_PREF 1001 // after l3mdev-table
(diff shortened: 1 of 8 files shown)


prgeor (approved):

prsunny:
@bgallagher-nexthop , could you provide title to reflect the changes in PR? you can add copilot review etc in description.

bgallagher-nexthop:
@prsunny do you think this is good to merge?

PR #4289, opened 2026-03-03
[P4Orch] Verify state for L2 multicast groups and members and No longer share bridge ports for L2 multicast.

bibhuprasad-hcl:
**What I did**
Verify state for L2 multicast groups and members.
No longer share RIFs across multicast router interface entries and bridge ports for L2 multicast.

**Why I did it**
To Verify state for L2 multicast groups and members.

**How I verified it**
UT Cases.

**Details if related**


orchagent/p4orch/p4orch_util.cpp
```diff
@@ -196,20 +196,6 @@ std::string KeyGenerator::generateMulticastReplicationKey(
   return generateKey(fv_map);
 }
 
-std::string KeyGenerator::generateMulticastRouterInterfaceRifKey(
```
mint570, line 199: This seems to be new change from the internal cl.

mint570 (approved):

PR #4290, opened 2026-03-03
[202511] Fix for sonic-net/sonic-buildimage#25397

venkit-nexthop:
What I did
Fix for https://github.com/sonic-net/sonic-buildimage/issues/25397
Improve performance in fpmsyncd by writing only the non-empty fields into redis. This is done only when zmq is not enabled.

Why I did it
To improve performance when ZMQ is not enabled in the northbound direction.

How I verified it
Ran some perf tests.


venkit-nexthop:
@prsunny all the checks have passed. Could you approve this?

prsunny (approved):

PR #4295, opened 2026-03-04
[AclOrch] Updates Mirror and MirrorV6 Ingress Stage to Accept Egress Action

AndoniSanguesa:
**What I did**
Added Mirror Egress Action to set of available actions for Ingress Stage Acl tables of type Mirror and MirrorV6 

**Why I did it**
Broadcom has specified that for TH5, they only support egress on the ingress ACL pipeline. There is no ASIC support for any mirroring actions on the egress ACL pipeline.

**How I verified it**
Installed on 7060x6 and verified that acl rule was accepted after change. Installed on a mellanox device as well to verify the rule is still rejected on a different sku. Mirroring occurred as expected on 7060x6, no regressions were observed.

**Details if related**


bingwang-ms:
The change is good for testing, but it may cause combability issue if we added both actions in any direction blindly.
The suggestion is to add the action based on the capability check result.
For example, if EGRESS MIRROR is supported in INGRESS ACL action capability check result, then add it.

bingwang-ms:
@prsunny Could you help merge?

prsunny:
> The change is good for testing, but it may cause combability issue if we added both actions in any direction blindly. The suggestion is to add the action based on the capability check result. For example, if EGRESS MIRROR is supported in INGRESS ACL action capability check result, then add it.

I dont see this addressed or replied - @bingwang-ms , @AndoniSanguesa 
@dgsudharsan for viz

bingwang-ms:
> > The change is good for testing, but it may cause combability issue if we added both actions in any direction blindly. The suggestion is to add the action based on the capability check result. For example, if EGRESS MIRROR is supported in INGRESS ACL action capability check result, then add it.
> 
> I dont see this addressed or replied - @bingwang-ms , @AndoniSanguesa @dgsudharsan for viz

@AndoniSanguesa Would you mind adding the offline discussion into this PR so that this discussion is closed?

AndoniSanguesa:
@prsunny, sorry for the late reply on this thread. The compatibility check already exists. The Default ACl Action List defines all the of the rules that the orchagent will attempt to apply but those will be filtered by the SAI' capabilities. See below code snippet:

https://github.com/AndoniSanguesa/sonic-swss/blob/493b7936002073631ed53fde60cbffdff1df5c49/orchagent/aclorch.cpp#L2588-L2601

bingwang-ms:
@prsunny Could you help check please?

PR #4304, opened 2026-03-06
[hft]: Remove IPFIX template size estimation, query actual size from SAI instead

Pterosaur:
**What I did**

Removed the IPFIX template buffer size estimation logic in `HFTelProfile::updateTemplates()` and replaced it with the standard SAI two-phase query pattern: first query the required size with `count=0` / `list=nullptr`, then allocate and fetch.

**Why I did it**

The previous estimation logic used hardcoded constants to predict the IPFIX template buffer size, but could underestimate (e.g., estimated 65535 bytes vs. actual 119352 bytes). This caused the SDK to log ERR-level messages on the first attempt:

```
ERR syncd#SDK: [SAI_TAM.ERR] mlnx_generate_ipfix_templates: Buffer size is too small to hold IPFIX template [size:65535, required:119352].
ERR syncd#SDK: [SAI_TAM.ERR] mlnx_tam_tel_type_get_ipfix_templates: Failed to generate IPFIX templates.
ERR syncd#SDK: [SAI_TAM.ERR] mlnx_tam_tel_type_attrib_get: Failed to get attribute.
ERR syncd#SDK: [SAI_UTILS.ERR] get_dispatch_attribs_handler: Failed Get #0, IPFIX_TEMPLATES, key:TAM_TEL_TYPE
```

Although the retry with `SAI_STATUS_BUFFER_OVERFLOW` worked correctly, these error logs caused `test_hft_full_queue_counters` to fail in LogAnalyzer teardown.

By always querying the size first (count=0, list=nullptr), we avoid the unreliable estimation entirely and follow the idiomatic SAI pattern for variable-length attributes.

**How I verified it**

- `test_hft_full_queue_counters` passes without LogAnalyzer catching IPFIX template size mismatch errors
- IPFIX template retrieval works correctly regardless of template size

**Details if related**

- A separate request has been filed with NVIDIA SDK team to suppress ERR-level logs when `count=0` / `list=nullptr` is passed, since this is the standard way to query required buffer size. Until that is addressed, this change avoids triggering the error path altogether.

### DIFF
--- a/orchagent/p4orch/p4orch_util.cpp
+++ b/orchagent/p4orch/p4orch_util.cpp
@@ -259,10 +259,11 @@ std::string KeyGenerator::generateTunnelKey(const std::string &tunnel_id)
 }
 
 std::string KeyGenerator::generateIpv6TunnelTermKey(
-  const swss::IpAddress &dst_ipv6_ip,
-  const swss::IpAddress &dst_ipv6_mask,
-  const std::string& vrf_id) {
+    const swss::IpAddress& src_ipv6_ip, const swss::IpAddress& src_ipv6_mask,
+    const swss::IpAddress& dst_ipv6_ip, const swss::IpAddress& dst_ipv6_mask) {
   std::map<std::string, std::string> fv_map = {
+      {p4orch::kDecapSrcIpv6Ip, src_ipv6_ip.to_string()},
+      {p4orch::kDecapSrcIpv6Mask, src_ipv6_mask.to_string()},
       {p4orch::kDecapDstIpv6Ip, dst_ipv6_ip.to_string()},
       {p4orch::kDecapDstIpv6Mask, dst_ipv6_mask.to_string()}};
   return generateKey(fv_map);
--- a/orchagent/p4orch/p4orch_util.h
+++ b/orchagent/p4orch/p4orch_util.h
@@ -103,7 +103,10 @@ constexpr char* kDisableSrcMacRewrite = "disable_src_mac_rewrite";
 constexpr char* kDisableDstMacRewrite = "disable_dst_mac_rewrite";
 constexpr char* kDisableVlanRewrite = "disable_vlan_rewrite";
 constexpr char* kIpv6TunnelTermAction = "mark_for_tunnel_decap_and_set_vrf";
+constexpr char* kDecapSrcIpv6 = "src_ipv6";
 constexpr char* kDecapDstIpv6 = "dst_ipv6";
+constexpr char* kDecapSrcIpv6Ip = "src_ipv6_ip";
+constexpr char* kDecapSrcIpv6Mask = "src_ipv6_mask";
 constexpr char* kDecapDstIpv6Ip = "dst_ipv6_ip";
 constexpr char* kDecapDstIpv6Mask = "dst_ipv6_mask";
 
@@ -318,6 +321,8 @@ struct P4AclRuleAppDbEntry
 struct Ipv6TunnelTermAppDbEntry
 {
   // Match
+  swss::IpAddress src_ipv6_ip;
+  swss::IpAddress src_ipv6_mask;
   swss::IpAddress dst_ipv6_ip;
   swss::IpAddress dst_ipv6_mask;
   // Action
@@ -412,9 +417,10 @@ class KeyGenerator
 
     static std::string generateTunnelKey(const std::string &tunnel_id);
 
-    static std::string generateIpv6TunnelTermKey(const swss::IpAddress &dst_ipv6_ip,
-                                                 const swss::IpAddress &dst_ipv6_mask,
-                                                 const std::string& vrf_id);
+    static std::string generateIpv6TunnelTermKey(const swss::IpAddress& src_ipv6_ip,
+                                                 const swss::IpAddress& src_ipv6_mask,
+                                                 const swss::IpAddress& dst_ipv6_ip,
+                                                 const swss::IpAddress& dst_ipv6_mask);
 
     static std::string generateExtTableKey(const std::string &table_name, const std::string &table_key);
 
--- a/orchagent/p4orch/tests/p4orch_util_test.cpp
+++ b/orchagent/p4orch/tests/p4orch_util_test.cpp
@@ -51,9 +51,11 @@ TEST(P4OrchUtilTest, KeyGeneratorTest)
 
     auto ipv6_tunnel_term_key =
         KeyGenerator::generateIpv6TunnelTermKey(
-          swss::IpAddress("::1"), swss::IpAddress("::1"), "vrf_id");
+        swss::IpAddress("::1"), swss::IpAddress("::1"), swss::IpAddress("::2"),
+        swss::IpAddress("::2"));
     EXPECT_EQ(
-      "dst_ipv6_ip=::1:dst_ipv6_mask=::1", ipv6_tunnel_term_key);
+        "dst_ipv6_ip=::2:dst_ipv6_mask=::2:src_ipv6_ip=::1:src_ipv6_mask=::1",
+        ipv6_tunnel_term_key);
 }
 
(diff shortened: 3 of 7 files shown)


Pterosaur:
202511 backport PR: #4305

tests/mock_tests/hftelprofile_ut.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+                /* Placement-new the containers and strings that may be
+                 * accessed (directly or via logging) by updateTemplates(). */
+                new (const_cast<string*>(&p->m_profile_name)) string();
+                new (&p->m_sai_tam_tel_type_objs)
+                    decay_t<decltype(p->m_sai_tam_tel_type_objs)>();
+                new (&p->m_sai_tam_tel_type_templates)
+                    decay_t<decltype(p->m_sai_tam_tel_type_templates)>();
+
+                p->m_sai_tam_tel_type_objs[SAI_OBJECT_TYPE_PORT] = guard;
```
banidoru, line 142: Nit: This Stub pattern with `memset` + selective placement-new is fragile — if `updateTemplates()` ever accesses additional members (e.g., via logging or new logic), this will silently hit UB on uninitialized objects. Consider adding a brief comment listing which members `updateTemplates()` actually touches, so future maintainers know what to placement-new if the function evolves.

orchagent/high_frequency_telemetry/hftelprofile.cpp
```diff
@@ -955,47 +955,35 @@ void HFTelProfile::updateTemplates(sai_object_id_t tam_tel_type_obj)
...
-    const size_t IPFIX_TEMPLATE_METADATA_SIZE (12LLU);
-    const size_t IPFIX_TEMPLATE_MAX_STATS_COUNT (((IPFIX_TEMPLATE_MAX_SIZE - IPFIX_HEADER_SIZE - IPFIX_TEMPLATE_METADATA_SIZE) / COUNTER_SIZE) - 1LLU);
-    size_t estimated_template_size = (counters_count / IPFIX_TEMPLATE_MAX_STATS_COUNT + 1) * IPFIX_TEMPLATE_MAX_SIZE;
-
-    vector<uint8_t> buffer(estimated_template_size, 0);
-
+    // Query the required buffer size first by passing count=0 and list=nullptr,
+    // then allocate and fetch the actual data.
     sai_attribute_t attr;
     attr.id = SAI_TAM_TEL_TYPE_ATTR_IPFIX_TEMPLATES;
```
banidoru, line 961: Nit: `sai_attribute_t attr;` leaves padding/unused union members uninitialized. Consider `sai_attribute_t attr{};` for defensive zero-initialization — trivial cost and avoids any future surprises if the struct layout changes.

tests/mock_tests/hftelprofile_ut.cpp
```diff
@@ -0,0 +1,217 @@
...
+// the #define private public hack (they use 'private' internally).
+#include <sstream>
+#include <string>
+#include <vector>
+#include <map>
+#include <unordered_map>
+#include <set>
+#include <memory>
+
+#define private public
```
banidoru, line 11: The `#define private public` / `#define protected public` hack is used elsewhere in this test suite, so it's consistent. Just flagging for awareness: this is technically undefined behavior per the C++ standard (redefining keywords). If the project ever moves to a test-friendlier pattern (e.g., friend test classes or a `_test_access` method), these tests should be migrated.

tests/mock_tests/hftelprofile_ut.cpp
```diff
@@ -0,0 +1,217 @@
...
+        g_mock.first_call_count  = 0;
+
+        ASSERT_NO_THROW(s.p->updateTemplates(fake_tel_type_oid));
+
+        auto &tpl = s.p->m_sai_tam_tel_type_templates[SAI_OBJECT_TYPE_PORT];
+        EXPECT_TRUE(tpl.empty());
+    }
+
+    /* ---- First query fails with unexpected status ---- */
+    TEST_F(UpdateTemplatesTest, FirstCall_UnexpectedFailure)
```
banidoru, line 207: Nit: Consider adding a test for the edge case where SAI returns `SAI_STATUS_BUFFER_OVERFLOW` with `count = 0`. The production code handles it correctly (the `count > 0` guard skips the second call and stores an empty buffer), but having an explicit test would document that this branch is intentional and prevent regressions.

orchagent/high_frequency_telemetry/hftelprofile.cpp
```diff
@@ -955,47 +955,35 @@ void HFTelProfile::updateTemplates(sai_object_id_t tam_tel_type_obj)
...
+        attr.value.u8list.list = buffer.data();
+        status = sai_tam_api->get_tam_tel_type_attribute(tam_tel_type_obj, 1, &attr);
 
-    buffer.resize(attr.value.u8list.count);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_THROW("Failed to get the TAM telemetry type object %s attributes: %d",
+                           sai_serialize_object_id(tam_tel_type_obj).c_str(), status);
+        }
+
```
banidoru, line 984: Minor: The second `buffer.resize(attr.value.u8list.count)` is a nice defensive touch in case SAI returns fewer bytes than requested. Worth a brief inline comment (e.g., `// SAI may return fewer bytes than allocated`) so future readers know this isn't redundant with the earlier resize.

tests/mock_tests/hftelprofile_ut.cpp
Pterosaur: Good point. I added a comment documenting which members the partial Stub currently constructs/touches so future updates to updateTemplates() are less likely to silently drift into UB territory. Commit: ca07d7b

orchagent/high_frequency_telemetry/hftelprofile.cpp
Pterosaur: Agreed. I changed this to `sai_attribute_t attr{}` for defensive zero-initialization. Commit: 18257be

tests/mock_tests/hftelprofile_ut.cpp
Pterosaur: Agreed. This pattern already exists in multiple tests in this area, so I am leaving it as-is in this PR and will revisit it later if we move to a cleaner test-access pattern.

tests/mock_tests/hftelprofile_ut.cpp
Pterosaur: Good suggestion. I added an explicit unit test for the `BUFFER_OVERFLOW` + `count == 0` edge case to document the intended behavior and protect it from regressions. Commit: cde3f65

orchagent/high_frequency_telemetry/hftelprofile.cpp
Pterosaur: Makes sense. I added a brief comment explaining that the second resize is intentional because SAI may return fewer bytes than initially requested. Commit: 2c68f63

banidoru (approved):
All prior review feedback addressed. The two-phase SAI query pattern is correct (accepts both SUCCESS and BUFFER_OVERFLOW), zero-initialization is applied, defensive resize comment added, and test coverage is thorough including the BUFFER_OVERFLOW+count=0 edge case. LGTM.

banidoru (commented):
Iteration 2 re-review: All prior feedback has been addressed. The two-phase SAI query pattern (count=0/nullptr probe → allocate → fetch) is correctly implemented, handles both SUCCESS and BUFFER_OVERFLOW on the initial call, and guards against count=0. Test coverage is thorough — happy path, empty templates (both SUCCESS and BUFFER_OVERFLOW variants), first-call failure, second-call failure, and unknown OID. No new issues found. LGTM.

r12f (approved):

PR #4309, opened 2026-03-09
Support deleting/updating entries with new multicast actions

bibhuprasad-hcl:
**What I did**
Add functionality to deleting and updating entries with new multicast actions.

**Why I did it**
To support deleting and updating entries with new multicast actions

**How I verified it**
UT cases

**Details if related**


orchagent/p4orch/tests/l3_multicast_manager_test.cpp
```diff
@@ -2509,11 +2518,6 @@ TEST_F(L3MulticastManagerTest,
                               NextHopAttrArrayEq(exp_nh_attrs)))
       .WillOnce(Return(SAI_STATUS_FAILURE));
 
-  // Expect to enter critical state.
```
yunang-c: why remove these in this commit? If ReportComponentState has not been upstreamed yet, can you remove it from the commit that added these lines?

orchagent/p4orch/l3_multicast_manager.cpp
```diff
@@ -1880,7 +1961,7 @@ ReturnCode L3MulticastManager::deleteL3MulticastRouterInterfaceEntry(
   }
 
   // For re-factoring purposes, only the new actions will setup the next hop.
-  if (entry->action != p4orch::kSetMulticastSrcMac) {
+  if (entry->action != p4orch::kSetSrcMac) {
```
yunang-c, line 1964: why change the action name here? It doesn't seem to be part of the original CL.

PR #4317, opened 2026-03-10
[countersyncd]: align multi-message netlink parsing

Pterosaur:
## What I did
- advance to the next netlink message using 4-byte alignment instead of raw `nlmsg_len`
- keep payload extraction bounded by the original message length (exclude padding)
- add a unit test that covers multiple messages in one recv buffer where the first message requires netlink padding

## Why
Linux netlink messages are 4-byte aligned, and the canonical `NLMSG_NEXT()` macro advances by `NLMSG_ALIGN(nlmsg_len)`. Advancing by raw `nlmsg_len` can misread the next header when a recv buffer contains multiple messages and one message length is not already aligned.

## Validation
Ran in CI-like `sonic-slave-bookworm` container with Azure Pipelines deb artifacts installed:
- `cargo test --locked --manifest-path crates/countersyncd/Cargo.toml test_multiple_aligned_messages_in_buffer -- --nocapture`
- `cargo test --locked --manifest-path crates/countersyncd/Cargo.toml test_multiple_messages_in_buffer -- --nocapture`
- `cargo test --locked --manifest-path crates/countersyncd/Cargo.toml test_fragmented_message -- --nocapture`

### DIFF
--- a/crates/countersyncd/src/actor/data_netlink.rs
+++ b/crates/countersyncd/src/actor/data_netlink.rs
@@ -76,6 +76,7 @@ impl NetlinkMessageParser {
         }
     }
 
+    /// Mirrors `NLMSG_ALIGN` from `linux/netlink.h` by rounding lengths up to the next 4-byte boundary.
     fn nlmsg_align(len: usize) -> usize {
         (len + 3) & !3
     }


Pterosaur:
@banidoru follow-up fixes for the latest review feedback are now pushed. Please take another look when convenient.

crates/countersyncd/src/actor/data_netlink.rs
```diff
@@ -76,6 +76,10 @@ impl NetlinkMessageParser {
         }
     }
 
+    fn nlmsg_align(len: usize) -> usize {
+        (len + 3) & !3
+    }
```
banidoru, line 82: Nit (non-blocking): A one-line doc comment like `/// Mirrors NLMSG_ALIGN from linux/netlink.h — rounds up to 4-byte boundary.` would help readers unfamiliar with the kernel convention.

banidoru: Addressed in commit 2217d857 — the doc comment now reads `/// Mirrors NLMSG_ALIGN from linux/netlink.h by rounding lengths up to the next 4-byte boundary.` which is clear and informative. Thanks!

banidoru: Confirmed — the doc comment added in 2217d857 looks good. This concern is addressed.

banidoru (approved):
Re-review (iteration 2): Previous concerns addressed — doc comment added for nlmsg_align, offset clamping applied, test helper uses the parser's own alignment method. The alignment fix correctly mirrors NLMSG_ALIGN/NLMSG_NEXT semantics. Test coverage for padded multi-message buffers looks solid. No new issues found. LGTM.

50n1c-rnsft:
## Code Review

### PR Changes — Looks Good Overall ✅

The core `NLMSG_ALIGN` fix is correct and matches the Linux kernel `NLMSG_NEXT()` semantics:

- `nlmsg_align()` — `(len + 3) & !3` correctly mirrors `linux/netlink.h`.
- Payload extraction still uses raw `nl_len` (not aligned) — correct, no padding bytes leak into payload.
- `offset += min(aligned_nl_len, remaining)` — correctly prevents out-of-bounds when the last message in the buffer has no trailing padding.
- The new test covers the primary multi-message alignment scenario.

**Verified scenario:** Two messages `[msg1(21B)][pad 3B][msg2(30B)]` — old code `offset+=21` would read padding as header (🐛), new code `offset+=24` correctly lands on msg2. End-of-buffer case (`min(24,21)=21`) also works correctly.

---

### Minor Suggestion for the PR

**Test accuracy:** In `test_multiple_aligned_messages_in_buffer`, `append_aligned_mock_netlink_message` adds trailing padding after the **last** message too. In real kernel buffers, the last message has no trailing padding. Consider not padding the final message (or adding an additional test case without it) to better mirror real-world recv buffers:

```rust
// More realistic: last message has no trailing padding
append_aligned_mock_netlink_message(&mut combined_buffer, b"A");
let msg2 = create_mock_netlink_message(b"SECOND");
let msg2_len = 20 + b"SECOND".len();
combined_buffer.extend_from_slice(&msg2[..msg2_len]); // no padding
```

This doesn't affect correctness (the `min` clamp handles it), but improves test fidelity.

---

### Pre-existing Issues Found in `data_netlink.rs` (not introduced by this PR)

#### 🔴 Bug: Parser buffer not cleared on reconnect (Significant)

**Location:** `disconnect()` (L629) and `connect()` (L576)

When `parse_buffer` encounters corrupted data and returns `Err(InvalidData)`, the `incomplete_buffer` is **not cleared**. The `run()` loop then calls `disconnect()` + `connect()`, but `message_parser` state is never reset. Data from the new socket gets **appended to the old corrupted buffer**, causing:
- Persistent parse failures → repeated reconnects → failure loop
- Even external reconnect commands from `ControlNetlinkActor` cannot recover the parser

**Suggested fix:** Clear parser state on disconnect:
```rust
fn disconnect(&mut self) {
    if self.socket.is_some() {
        self.socket = None;
        self.nl_resolver = None;
        self.message_parser = NetlinkMessageParser::new(); // ← add this
    }
}
```
(And similarly in `reset()`.)

#### 🟡 Blocking sleep in async context

**Location:** L944
```rust
sleep(Duration::from_millis(SOCKET_READINESS_TIMEOUT_MS)); // std::thread::sleep
```
This is inside `async fn run()` and blocks the tokio runtime thread. Should be `tokio::time::sleep(...).await`.

#### 🟡 Byte order assumption

**Location:** L110-115, L192-193, L207-211

All netlink header fields use `u32::from_le_bytes()` / `u16::from_le_bytes()`, but netlink headers use **host byte order**. This works on little-endian platforms (x86, most ARM) but would break on big-endian. Should use `from_ne_bytes()`.

#### 🟢 Dead code

**Location:** L699-708 — `if size == 0` check inside a `size > 0` match arm is unreachable.

Pterosaur:
Created 202412 backport PR: https://github.com/Azure/sonic-swss.msft/pull/212

PR #4319, opened 2026-03-10
[countersyncd]: avoid blocking async actor loops

Pterosaur:
## What I changed
- moved `SwssActor` table polling onto a dedicated blocking reader thread
- replaced async-task `std::thread::sleep(...)` loops in `control_netlink.rs` with `tokio::time::interval(...).tick().await`
- reworked `DataNetlinkActor`'s main loop to use async `select!` + interval-driven non-blocking recv instead of blocking sleep-based polling
- kept existing netlink/socket semantics intact (`WouldBlock`, `ENOBUFS`, reconnect flow)

## Why
These actors are spawned on Tokio, but they mixed in blocking waits / `std::thread::sleep`, which can pin Tokio worker threads and cause scheduling jitter under load.

The riskiest case was `swss.rs`, where synchronous SWSS/Redis table polling could block for long periods. That is now isolated onto its own dedicated thread.

## Validation
Validated in CI-like `sonic-slave-bookworm` container with Azure Pipelines deb artifacts:
- `cargo test --locked --manifest-path crates/countersyncd/Cargo.toml test_control_netlink_actor -- --nocapture`
- `cargo test --locked --manifest-path crates/countersyncd/Cargo.toml test_data_netlink_actor_run -- --nocapture`
- `cargo test --locked --manifest-path crates/countersyncd/Cargo.toml test_swss_actor_processes_existing_data -- --nocapture`


crates/countersyncd/src/actor/swss.rs
```diff
@@ -58,86 +64,148 @@ impl SwssActor {
...
+
+                    match Self::blocking_collect_events(&mut session_table, timeout) {
+                        Ok(events) => {
+                            for event in events {
+                                if event_sender.blocking_send(event).is_err() {
+                                    debug!("SwssActor event receiver dropped, terminating reader thread");
+                                    return;
                                 }
                             }
                         }
```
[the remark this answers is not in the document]
Pterosaur: Fixed in 1276dd5e. The SWSS reader thread now checks `event_sender.is_closed()` at the top of each polling iteration and exits cleanly even if the receiver is dropped while the table is idle.

crates/countersyncd/src/actor/control_netlink.rs
```diff
@@ -386,8 +386,11 @@ impl ControlNetlinkActor {
         let mut heartbeat_counter = 0u32;
         let mut last_periodic_reconnect_counter = 0u32;
         let mut family_was_available = true; // Assume family starts available
+        let mut poll_interval = interval(Duration::from_millis(10));
+        poll_interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
```
banidoru, line 391: `control_netlink` still uses `poll_interval.tick()` + `try_recv()` polling, while `data_netlink` was refactored to use `tokio::select!` over the command channel. This is inconsistent: commands here are still delayed up to 10ms. Consider applying the same `select!` pattern for uniformity and to make command handling truly event-driven.

crates/countersyncd/src/actor/swss.rs
```diff
@@ -162,48 +235,75 @@ impl SwssActor {
     /// # Arguments
     /// * `key` - Session key (e.g., "test|PORT")  
     /// * `field_values` - HashMap of field-value pairs from the state DB
```
banidoru, line 241: Gating `handle_session_update`, `handle_session_delete`, and `validate_and_process_session` with `#[cfg(test)]` means tests exercise a different code path than production (`process_session_update` / `process_session_delete` / `validate_and_send_session`). If the production methods diverge or have bugs, tests won't catch them. Consider having the test-only wrappers delegate to the production static methods (which `validate_and_process_session` does), but also ensure `process_session_update` and `process_session_delete` themselves have test coverage — or remove the test-only wrappers entirely and test the production path directly.

crates/countersyncd/src/actor/swss.rs
```diff
@@ -58,86 +64,153 @@ impl SwssActor {
...
-                        MAX_TEST_ITERATIONS
-                    );
-                    break;
-                }
-            }
+        // Keep the SWSS table polling on a dedicated blocking thread so we don't park a Tokio worker.
+        let SwssActor {
+            mut session_table,
+            template_recipient,
+        } = actor;
```
banidoru, line 78: The `JoinHandle` is stored in `_reader_thread` but never joined. When `run()` returns (after `event_receiver` stream ends), the thread is detached. In the normal shutdown path, the thread will detect `is_closed()` — but only after the next `read_data` timeout (up to 10 seconds in production). Consider joining the thread (with a timeout) after the event loop exits, or using a secondary shutdown signal the reader checks more frequently, so shutdown is prompt rather than waiting up to 10s for the table poll to time out.

crates/countersyncd/src/actor/data_netlink.rs
```diff
@@ -786,73 +748,69 @@ impl DataNetlinkActor {
         );
         let mut heartbeat_counter = 0u32;
         let mut consecutive_failures = 0u32;
+        let mut poll_interval = interval(Duration::from_millis(SOCKET_READINESS_TIMEOUT_MS));
```
banidoru: Nit: `SOCKET_READINESS_TIMEOUT_MS` is typed `u64` but `interval()` takes `Duration`. Worth adding a comment noting this is the poll interval for non-blocking recv attempts, since the constant name ("socket readiness timeout") now describes a slightly different semantic — it's no longer a readiness check timeout, it's a poll cadence.

banidoru (commented):
Good improvement — moving blocking I/O off Tokio workers is the right call. A few concerns below, mostly around consistency and test fidelity.

crates/countersyncd/src/actor/swss.rs
```diff
@@ -58,86 +64,153 @@ impl SwssActor {
...
-                    debug!(
-                        "SwssActor test mode reached maximum iterations ({}), terminating",
-                        MAX_TEST_ITERATIONS
-                    );
-                    break;
-                }
-            }
+        // Keep the SWSS table polling on a dedicated blocking thread so we don't park a Tokio worker.
+        let SwssActor {
+            mut session_table,
```
banidoru, line 76: Nit: consider extracting `32` into a named constant (e.g., `const SWSS_EVENT_CHANNEL_CAPACITY: usize = 32;`). This makes the bound discoverable and tunable without digging into `run()`.

crates/countersyncd/src/actor/data_netlink.rs
```diff
@@ -786,73 +748,69 @@ impl DataNetlinkActor {
...
-                    NetlinkCommand::SocketConnect(SocketConnect { family, group }) => {
-                        actor.reset(&family, &group);
-                        consecutive_failures = 0; // Reset failure counter on reconnect command
-                    }
-                    NetlinkCommand::Reconnect => {
-                        actor.connect(true); // Force reconnection
-                        consecutive_failures = 0; // Reset failure counter on reconnect command
+            select! {
+                command = actor.command_recipient.recv() => {
+                    match command {
```
banidoru, line 792: Minor: after processing a `Reconnect` or `SocketConnect` command, the new socket won't be read until the next `poll_interval` tick. With a 10ms interval this is negligible, but if the interval is ever increased, consider resetting the interval (or doing an immediate `try_recv` pass) after socket state changes to avoid unnecessary latency on reconnect.

banidoru (approved):
Good refactoring. The key improvements are solid:

- Replacing `std::thread::sleep` with `tokio::time::interval` in control/data netlink actors correctly unblocks the async runtime.
- Moving `SwssActor` SWSS table polling to a dedicated OS thread via `thread::Builder::spawn` + mpsc channel is the right pattern for blocking FFI calls in an async context.
- The `tokio::select!` in `DataNetlinkActor` makes command handling truly event-driven.

A couple of minor observations below, but nothing blocking.

crates/countersyncd/src/actor/swss.rs
```diff
@@ -58,86 +64,153 @@ impl SwssActor {
...
+                    return;
+                }
+            };
+
+        while let Some(event) = event_receiver.recv().await {
+            match event {
+                SwssEvent::Update { key, session_data } => {
+                    Self::process_session_update(&template_recipient, &key, &session_data).await;
+                }
+                SwssEvent::Delete { key } => {
```
banidoru, line 141: `blocking_collect_events` returning `Err` on a `pops()` failure causes the caller to sleep 100ms and retry. In the previous code, a `pops()` error was logged but the loop continued immediately without delay. More importantly, a `pops()` error after `read_data` returns `Data` means that batch of signaled data is silently lost — the next `read_data` call won't re-signal it.

Consider separating `read_data` errors (where a retry delay makes sense) from `pops()` errors (where the data is already lost and delaying doesn't help). At minimum, log at `error!` level in the caller for `pops()` failures to maintain parity with the old code's explicit `error!` log.

crates/countersyncd/src/actor/data_netlink.rs
```diff
@@ -786,73 +748,69 @@ impl DataNetlinkActor {
...
+                                    consecutive_failures = 0;
+                                }
+                                NetlinkCommand::Reconnect => {
+                                    actor.connect(true);
+                                    consecutive_failures = 0;
+                                }
+                                NetlinkCommand::SoftReconnect => {
+                                    actor.connect(false);
+                                    consecutive_failures = 0;
+                                }
```
banidoru, line 810: `try_recv` performs a synchronous `socket.recv()` call directly on a tokio worker thread inside this `select!` arm. This is safe only because the netlink socket is set to non-blocking mode (confirmed by the `WouldBlock` handling below). Worth adding a brief comment here documenting the non-blocking invariant, e.g. `// Safety: socket is non-blocking, recv() returns immediately with WouldBlock if no data`, so future maintainers don't accidentally break this assumption.

banidoru (commented):
Overall this is a solid improvement — replacing blocking `sleep()` with `tokio::select!` and `interval` in the data path, and offloading the SWSS blocking poll to a dedicated thread, are the right architectural moves.

Key findings:
1. **`blocking_collect_events` error semantic change**: A `pops()` failure after a successful `read_data(Data)` now causes a 100ms sleep+retry (same as `read_data` failures), whereas previously only an error log was emitted and the loop continued immediately. This means transient `pops()` errors now introduce latency and silently drop the data batch that was signaled as ready.
2. **`try_recv` called synchronously inside async `select!` arm**: The netlink socket `recv()` is called directly on a tokio worker thread. This is safe only if the socket is guaranteed non-blocking. A brief comment documenting this invariant would help future maintainers.
3. Minor: `control_netlink` inconsistency with `data_netlink` (poll+try_recv vs select!) was noted by other reviewers — agree this should be unified.

banidoru (approved):
All reviewers approved. LGTM.

crates/countersyncd/src/actor/swss.rs
banidoru: Fixed — `#[cfg(test)]` removed from all three methods. Tests now exercise the same code path as production. ✅

crates/countersyncd/src/actor/data_netlink.rs
banidoru: Fixed — renamed to `SOCKET_POLL_INTERVAL_MS` with an accurate comment describing its purpose. ✅

crates/countersyncd/src/actor/swss.rs
banidoru: Fixed — `SWSS_EVENT_CHANNEL_CAPACITY` constant added and used. ✅

crates/countersyncd/src/actor/swss.rs
banidoru: Fixed — `pops()` failure now logs at `error!` level and returns the (empty) events vec instead of propagating an error to the retry-with-delay path. This preserves log parity and avoids the unnecessary 100ms delay on data-loss errors. ✅

crates/countersyncd/src/actor/data_netlink.rs
banidoru: Fixed — safety comment added documenting the non-blocking invariant. ✅

crates/countersyncd/src/actor/control_netlink.rs
banidoru: Still open — `control_netlink` continues to use `poll_interval.tick()` + `try_recv()` rather than `tokio::select!` over the command channel. The inconsistency with `data_netlink` remains.

crates/countersyncd/src/actor/swss.rs
banidoru: Still open — `_reader_thread` JoinHandle is still unused. Shutdown can be delayed up to 10s waiting for the `read_data` timeout. Consider joining the thread after the event loop exits.

crates/countersyncd/src/actor/data_netlink.rs
banidoru: Still open — no change to the reconnect path; the new socket still waits for the next `poll_interval` tick before being read.

banidoru (commented):
Re-review of iteration 2. Five prior concerns addressed in `a8f9d8c4` and `1276dd5e`:

✅ `#[cfg(test)]` gating removed — tests now exercise production code paths
✅ `SOCKET_READINESS_TIMEOUT_MS` renamed to `SOCKET_POLL_INTERVAL_MS` with accurate comment
✅ Magic `32` extracted to `SWSS_EVENT_CHANNEL_CAPACITY`
✅ `pops()` errors now log at `error!` level and avoid unnecessary retry delay
✅ Non-blocking safety comment added to `try_recv`

Three prior concerns remain open (replies posted on each thread):
- `control_netlink` still uses poll+try_recv rather than `select!`
- `_reader_thread` JoinHandle not joined — shutdown delayed up to 10s
- No interval reset after reconnect in data_netlink

No new issues found in this iteration. The follow-up commit is clean and well-scoped.

crates/countersyncd/src/actor/swss.rs
banidoru: Confirmed fixed — `#[cfg(test)]` removed. The instance methods now delegate to the same `validate_and_send_session` static method used by the production `run()` path, so tests and production converge.

crates/countersyncd/src/actor/data_netlink.rs
banidoru: Confirmed fixed — renamed to `SOCKET_POLL_INTERVAL_MS` with an accurate doc comment.

crates/countersyncd/src/actor/swss.rs
banidoru: Confirmed fixed — `SWSS_EVENT_CHANNEL_CAPACITY` constant extracted and used.

crates/countersyncd/src/actor/swss.rs
banidoru: Confirmed fixed — `pops()` failure now logs at `error!` and returns the empty events vec, avoiding the unnecessary 100ms retry delay.

crates/countersyncd/src/actor/data_netlink.rs
banidoru: Confirmed fixed — safety comment added at the `recv` call site documenting the non-blocking fd invariant.

banidoru (commented):
Re-review (iteration 2): The fixes for the previous round of feedback look good — `#[cfg(test)]` removal, `SOCKET_POLL_INTERVAL_MS` rename, `SWSS_EVENT_CHANNEL_CAPACITY` extraction, `pops()` error handling fix, and the non-blocking safety comment are all confirmed addressed.

Three prior concerns remain open:
1. **`control_netlink` still uses `poll_interval.tick()` + `try_recv()` polling** instead of `select!` (inconsistent with `data_netlink`).
2. **`_reader_thread` JoinHandle is never joined** — shutdown may be delayed up to 10s.
3. **Post-reconnect socket read latency** in `data_netlink` — new socket waits for next `poll_interval` tick.

No new issues found in this iteration. The core refactoring (dedicated blocking thread for SWSS, `select!` in data_netlink, `interval` replacing `sleep`) is solid.

crates/countersyncd/src/actor/swss.rs
banidoru: Verified — `#[cfg(test)]` gates are removed. The instance methods now delegate to the same static `validate_and_send_session` / `process_session_delete` used by the production `run()` path. Tests and production converge.

crates/countersyncd/src/actor/data_netlink.rs
banidoru: Verified — renamed to `SOCKET_POLL_INTERVAL_MS` with an accurate doc comment describing the poll cadence semantic.

crates/countersyncd/src/actor/swss.rs
banidoru: Verified — `SWSS_EVENT_CHANNEL_CAPACITY` constant extracted and used in the `mpsc::channel` call.

crates/countersyncd/src/actor/swss.rs
banidoru: Verified — `pops()` failure now logs at `error!` and returns `Ok(events)` (empty vec), avoiding the unnecessary 100ms retry delay. Good fix.

crates/countersyncd/src/actor/data_netlink.rs
banidoru: Verified — safety comment added at the `recv` call site documenting the non-blocking fd invariant.

crates/countersyncd/src/actor/control_netlink.rs
banidoru: Still open — `control_netlink` continues to use `poll_interval.tick()` + `try_recv()` rather than `tokio::select!`. The inconsistency with `data_netlink` remains unaddressed.

crates/countersyncd/src/actor/swss.rs
banidoru: Still open — `_reader_thread` JoinHandle is stored but never joined. Shutdown can still be delayed up to 10s waiting for the blocking `read_data` timeout.

crates/countersyncd/src/actor/data_netlink.rs
banidoru: Still open — after `Reconnect`/`SocketConnect`, the new socket waits for the next `poll_interval` tick before being read. No change in this iteration.

banidoru (commented):
Iteration 2 re-review: 5 of 8 prior concerns addressed (cfg(test) removal, constant rename, channel capacity extraction, pops() error handling, non-blocking safety comment). 3 remain open: (1) control_netlink still uses tick+try_recv vs data_netlink's select!, (2) _reader_thread JoinHandle never joined — up to 10s shutdown delay, (3) no immediate recv after reconnect. New code is well-structured — dedicated blocking thread for SWSS polling is the right call, SwssEvent enum cleanly separates blocking I/O from async processing, and parse_session_data extraction reduces duplication. No new issues found in this iteration.

banidoru (approved):
All reviewers approved. LGTM. Three non-blocking nits remain as open threads: control_netlink polling consistency, JoinHandle join on shutdown, and interval reset after reconnect.

crates/countersyncd/src/actor/control_netlink.rs
Pterosaur: Not changing this in the current PR. `ControlNetlinkActor` does not have an inbound command channel like `DataNetlinkActor`; it has a single periodic responsibility and already uses non-blocking socket reads. Keeping the 10ms tick loop here keeps the control path simple while still removing the actual blocking work from Tokio workers, which was the goal of this PR.

crates/countersyncd/src/actor/swss.rs
Pterosaur: I am leaving this as-is in this PR. The reader thread is intentionally fire-and-forget and now terminates once the async side drops the receiver or the process shuts down. Explicitly joining it would require adding a separate shutdown handshake or blocking an async shutdown path. If we decide we want a tighter graceful-shutdown bound than the current `read_data` timeout, I would rather do that as a focused follow-up.

crates/countersyncd/src/actor/data_netlink.rs
Pterosaur: Keeping the current behavior. The extra latency here is bounded by a single `SOCKET_POLL_INTERVAL_MS` tick (10ms), which is negligible relative to the reconnect path. I prefer that small bounded delay over duplicating the receive path inside every reconnect command arm in this PR.

banidoru (approved):
Re-review (iteration 3): All prior feedback has been addressed or reasonably deferred. Key fixes confirmed: `#[cfg(test)]` gates removed, `SWSS_EVENT_CHANNEL_CAPACITY` extracted, `SOCKET_POLL_INTERVAL_MS` renamed, `pops()` error handling improved, non-blocking safety comment added, `is_closed()` shutdown check added. The two deferred items (control_netlink select! pattern, JoinHandle joining) have clear justification for follow-up work. The async-blocking separation is clean and correct. LGTM.

banidoru (approved):
Iteration 3 re-review: all prior concerns addressed or reasonably deferred. The blocking-to-async migration is clean — SWSS polling correctly isolated on a dedicated thread with proper shutdown detection, data_netlink refactored to event-driven select!, and control_netlink converted from thread::sleep to tokio interval. No new issues found. LGTM.

banidoru (approved):
Clean refactor. The blocking SWSS table polling is correctly moved to a dedicated thread with proper shutdown signaling via `is_closed()`. The `data_netlink` actor now uses `tokio::select!` for event-driven command handling, eliminating the old `try_recv` polling. The `control_netlink` `sleep` → `interval` change removes the last `std::thread::sleep` from async context. Previously raised concerns (channel capacity constant, `#[cfg(test)]` removal, `pops()` error handling, non-blocking safety comment, constant rename) are all addressed. Deferred items (JoinHandle join, control_netlink select!, reconnect immediate read) are reasonable scope boundaries for this PR. No new issues found.

banidoru (approved):
All reviewers approved. LGTM.

Pterosaur:
Created 202412 backport PR: https://github.com/Azure/sonic-swss.msft/pull/214

prsunny:
@r12f , could you approve?

banidoru (approved):
Iteration 4 re-review: all prior threads resolved. New diff adds correct `NLMSG_ALIGN` handling for netlink message parsing with proper test coverage. No new issues found. LGTM.

banidoru (approved):
- All prior review threads resolved; fixes confirmed for cfg(test) removal, constant extraction, error handling, and safety comments
- Three deferred items (control_netlink polling, JoinHandle join, reconnect immediate-read) have reasonable justifications for follow-up
- Core refactor cleanly moves blocking SWSS table polling and netlink recv off Tokio workers
- `select!`-based command handling in data_netlink is a solid improvement over try_recv polling
- No new issues found in this iteration

banidoru (approved):
All reviewers approved. LGTM.

prsunny (approved):

prsunny:
@r12f for viz

vmittal-msft:
@prsunny @r12f is it a must fix for 202511 ?

r12f:
@Pterosaur Do you mind to help check this?

PR #4333, opened 2026-03-12
[swss]:Skip APPL_STATE_DB and pub/sub writes when FIB suppression is disabled

mike-dubrovsky:
**Why I did it**

Route download speed degraded after BGP prefix suppression was introduced. RouteOrch::publishRouteState() unconditionally writes to APPL_STATE_DB and sends a pub/sub notification on every route add/remove, even when FIB suppression is disabled and no consumer needs the data.

**What I did**

Made publishRouteState() a no-op until FIB suppression is enabled.

Made "suppress-fib-pending" config changes take effect only after swss restart or config reload, 
not dynamically at runtime. This eliminates multi-process race conditions between orchagent, fpmsyncd, and FRR.

Detailed description:

The CLI (config suppress-fib-pending) writes to DEVICE_METADATA and prints a message telling the user to perform config save and config reload  for the change to take effect. This is in the sonic-utilities PR (#4361).

Detailed Changes in this PR:
- orchagent: add -F command-line flag (gEnableFibSuppress) to control FIB suppression at startup; guard publishRouteState() to return early when suppression is disabled
- fpmsyncd: read orchagent's FIB suppression state from CONFIG_DB at startup instead of using CONFIG_DB subscription; remove dynamic CONFIG_DB subscription.

**How I verified it**
UT and sonic-mgmt  script.

**Details if related**
This change needs to go together with 
https://github.com/sonic-net/sonic-buildimage/pull/26151
https://github.com/sonic-net/sonic-utilities/pull/4361
https://github.com/sonic-net/sonic-mgmt/pull/22916
https://github.com/sonic-net/SONiC/pull/2335

### DIFF
--- a/fpmsyncd/fpmsyncd.cpp
+++ b/fpmsyncd/fpmsyncd.cpp
@@ -8,7 +8,6 @@
 #include "netdispatcher.h"
 #include "netlink.h"
 #include "notificationconsumer.h"
-#include "subscriberstatetable.h"
 #include "warmRestartHelper.h"
 #include "fpmsyncd/fpmlink.h"
 #include "fpmsyncd/fpmsyncd.h"
@@ -79,7 +78,6 @@ int main(int argc, char **argv)
 
     DBConnector db("APPL_DB", 0);
     DBConnector cfgDb("CONFIG_DB", 0);
-    SubscriberStateTable deviceMetadataTableSubscriber(&cfgDb, CFG_DEVICE_METADATA_TABLE_NAME);
     Table deviceMetadataTable(&cfgDb, CFG_DEVICE_METADATA_TABLE_NAME);
     DBConnector applStateDb("APPL_STATE_DB", 0);
     std::unique_ptr<NotificationConsumer> routeResponseChannel;
@@ -118,6 +116,7 @@ int main(int argc, char **argv)
         routeResponseChannel = std::make_unique<NotificationConsumer>(&applStateDb, routeResponseChannelName);
         sync.setSuppressionEnabled(true);
     }
+    SWSS_LOG_NOTICE("FIB suppression state: %s", suppressionEnabledStr.c_str());
 
     while (true)
     {
@@ -144,8 +143,6 @@ int main(int argc, char **argv)
 
             s.addSelectable(&fpm);
             s.addSelectable(&netlink);
-            s.addSelectable(&deviceMetadataTableSubscriber);
-
             if (sync.isSuppressionEnabled())
             {
                 s.addSelectable(routeResponseChannel.get());
@@ -250,61 +247,6 @@ int main(int argc, char **argv)
                     {
                         s.removeSelectable(&eoiuCheckTimer);
                     }
-                }
-                else if (temps == &deviceMetadataTableSubscriber)
-                {
-                    std::deque<KeyOpFieldsValuesTuple> keyOpFvsQueue;
-                    deviceMetadataTableSubscriber.pops(keyOpFvsQueue);
-
-                    for (const auto& keyOpFvs: keyOpFvsQueue)
-                    {
-                        const auto& key = kfvKey(keyOpFvs);
-                        const auto& op = kfvOp(keyOpFvs);
-                        const auto& fvs = kfvFieldsValues(keyOpFvs);
-
-                        if (op != SET_COMMAND)
-                        {
-                            continue;
-                        }
-
-                        if (key != "localhost")
-                        {
-                            continue;
-                        }
-
-                        for (const auto& fv: fvs)
-                        {
-                            const auto& field = fvField(fv);
-                            const auto& value = fvValue(fv);
-
-                            if (field != "suppress-fib-pending")
-                            {
-                                continue;
-                            }
-
-                            bool shouldEnable = (value == "enabled");
-
-                            if (shouldEnable && !sync.isSuppressionEnabled())
-                            {
-                                routeResponseChannel = std::make_unique<NotificationConsumer>(&applStateDb, routeResponseChannelName);
-                                sync.setSuppressionEnabled(true);
-                                s.addSelectable(routeResponseChannel.get());
-                            }
-                            else if (!shouldEnable && sync.isSuppressionEnabled())
-                            {
-                                /* When disabling suppression we mark all existing routes offloaded in zebra
-                                 * as there could be some transient routes which are pending response from
-                                 * orchagent, thus such updates might be missing. Since we are disabling suppression
-                                 * we no longer care about real HW offload status and can mark all routes as offloaded
-                                 * to avoid routes stuck in suppressed state after transition. */
-                                sync.markRoutesOffloaded(db);
-
-                                sync.setSuppressionEnabled(false);
-                                s.removeSelectable(routeResponseChannel.get());
-                                routeResponseChannel.reset();
-                            }
-                        } // end for fvs
-                    } // end for keyOpFvsQueue
                 }
                 else if (routeResponseChannel && (temps == routeResponseChannel.get()))
                 {
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -59,6 +59,7 @@ extern int gBatchSize;
 
 bool gRingMode = false;
 bool gSyncMode = false;
+bool gEnableFibSuppress = false;
 sai_redis_communication_mode_t gRedisCommunicationMode = SAI_REDIS_COMMUNICATION_MODE_REDIS_ASYNC;
 string gAsicInstance;
 
@@ -91,7 +92,7 @@ bool isChassisDbInUse()
 
 void usage()
 {
-    cout << "usage: orchagent [-h] [-r record_type] [-A] [-d record_location] [-f swss_rec_filename] [-j sairedis_rec_filename] [-b batch_size] [-m MAC] [-i INST_ID] [-s] [-z mode] [-k bulk_size] [-q zmq_server_address] [-c mode] [-t create_switch_timeout] [-v VRF] [-I heart_beat_interval] [-R] [-M]" << endl;
+    cout << "usage: orchagent [-h] [-r record_type] [-A] [-d record_location] [-f swss_rec_filename] [-j sairedis_rec_filename] [-b batch_size] [-m MAC] [-i INST_ID] [-s] [-z mode] [-k bulk_size] [-q zmq_server_address] [-c mode] [-t create_switch_timeout] [-v VRF] [-I heart_beat_interval] [-R] [-M] [-F]" << endl;
     cout << "    -h: display this message" << endl;
     cout << "    -r record_type: record orchagent logs with type (default 3)" << endl;
     cout << "                    Bit 0: sairedis.rec, Bit 1: swss.rec, Bit 2: responsepublisher.rec. For example:" << endl;
@@ -117,6 +118,7 @@ void usage()
     cout << "    -I heart_beat_interval: Heart beat interval in millisecond (default 10)" << endl;
     cout << "    -R enable the ring thread feature" << endl;
     cout << "    -M enable SAI MACSec POST" << endl;
+    cout << "    -F enable BGP FIB suppression" << endl;
 }
 
 void sighup_handler(int signo)
@@ -468,7 +470,7 @@ int main(int argc, char **argv)
     // Disable SAI MACSec POST by default. Use option -M to enable it.
     bool macsec_post_enabled = false;
 
-    while ((opt = getopt(argc, argv, "b:m:r:Af:j:d:i:hsz:k:q:c:t:v:I:RM")) != -1)
+    while ((opt = getopt(argc, argv, "b:m:r:Af:j:d:i:hsz:k:q:c:t:v:I:RMF")) != -1)
     {
         switch (opt)
         {
@@ -594,6 +596,9 @@ int main(int argc, char **argv)
          case 'M':
             macsec_post_enabled = true;
             break;
+        case 'F': // LCOV_EXCL_LINE
+            gEnableFibSuppress = true; // LCOV_EXCL_LINE
+            break; // LCOV_EXCL_LINE
         default: /* '?' */
             exit(EXIT_FAILURE);
         }
--- a/orchagent/routeorch.cpp
+++ b/orchagent/routeorch.cpp
@@ -34,6 +34,7 @@ extern TunnelDecapOrch *gTunneldecapOrch;
 extern size_t gMaxBulkSize;
 extern string gMySwitchType;
 extern bool gRouteStateAsyncPublish;
+extern bool gEnableFibSuppress;
 
 /* Default maximum number of next hop groups */
 #define DEFAULT_NUMBER_OF_ECMP_GROUPS   128
@@ -3201,6 +3202,11 @@ void RouteOrch::decNhgRefCount(const std::string &nhg_index, const std::string &
 void RouteOrch::publishRouteState(const RouteBulkContext& ctx, const ReturnCode& status)
 {
     SWSS_LOG_ENTER();
+
+    if (!gEnableFibSuppress)
+    {
+        return;
+    }
 
     std::vector<FieldValueTuple> fvs;
 
--- a/tests/mock_tests/routeorch_ut.cpp
+++ b/tests/mock_tests/routeorch_ut.cpp
@@ -12,6 +12,7 @@
 #include "bulker.h"
 
 extern string gMySwitchType;
+extern bool gEnableFibSuppress;
 
 extern std::unique_ptr<MockResponsePublisher> gMockResponsePublisher;
 
@@ -411,6 +412,8 @@ namespace routeorch_test
 
         void TearDown() override
         {
+            gEnableFibSuppress = false;
+
             RestoreSaiApis();
             DEINIT_SAI_API_MOCK(route);
             DEINIT_SAI_API_MOCK(next_hop_group);
@@ -876,6 +879,7 @@ namespace routeorch_test
 
     TEST_F(RouteOrchTest, RouteOrchTestSetDelResponse)
     {
+        gEnableFibSuppress = true;
         gMockResponsePublisher = std::make_unique<MockResponsePublisher>();
 
         std::deque<KeyOpFieldsValuesTuple> entries;
@@ -912,6 +916,7 @@ namespace routeorch_test
 
     TEST_F(RouteOrchTest, RouteOrchSetFullMaskSubnetPrefix)
     {
+        gEnableFibSuppress = true;
         gMockResponsePublisher = std::make_unique<MockResponsePublisher>();
 
         std::deque<KeyOpFieldsValuesTuple> entries;
@@ -930,6 +935,7 @@ namespace routeorch_test
 
     TEST_F(RouteOrchTest, RouteOrchLoopbackRoute)
     {
+        gEnableFibSuppress = true;
         gMockResponsePublisher = std::make_unique<MockResponsePublisher>();
 
         std::deque<KeyOpFieldsValuesTuple> entries;
--- a/tests/test_route.py
+++ b/tests/test_route.py
@@ -6,7 +6,7 @@ import pytest
 import ipaddress
 
 from swsscommon import swsscommon
-from dvslib.dvs_common import wait_for_result
+from dvslib.dvs_common import wait_for_result, PollingConfig
 
 class TestRouteBase(object):
     def setup_db(self, dvs):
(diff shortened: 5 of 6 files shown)


vpandian-nokia:
Hi, thank you for the work on adding FIB‑suppression support across orchagent and fpmsyncd.
While integrating this PR (along with sonic‑buildimage PR 26151) into our internal branch, I ran into several issues that required additional fixes. I’m documenting the details here so we can clarify whether these changes were intentionally omitted or simply missed.

1. Missing definition for STATE_FIB_SUPPRESS_TABLE_NAME
fpmsyncd.cpp uses: STATE_FIB_SUPPRESS_TABLE_NAME but this constant is not defined anywhere. I resolved the compile error by adding:
**#define STATE_FIB_SUPPRESS_TABLE_NAME  "FIB_SUPPRESS_TABLE"**
to **sonic-swss-common/common/schema.h**. 

2. Missing definition of gEnableFibSuppress in the test harness
After fixing the schema, the SWSS unit tests failed to compile/link because RouteOrch references: 
extern bool gEnableFibSuppress;
but the test environment never defines it. I fixed the issue by adding the missing global to tests/mock_tests/mock_orchagent_main.cpp:
**bool gEnableFibSuppress = true;**

This is required for the RouteOrch tests to build and appears to be another file that the PR may have unintentionally missed?

After defining gEnableFibSuppress, the following failing RouteOrch tests all passed:

- RouteOrchTestSetDelResponse
- RouteOrchSetFullMaskSubnetPrefix
- RouteOrchLoopbackRoute

Question: Did the upstream CI pipeline succeed or did I miss something from the PR?

prsunny:
@mike-dubrovsky , please follow the description template "why i did", "how tested" etc. Also update results from sonic-mgmt tests. @dgsudharsan for viz.

fpmsyncd/fpmsyncd.cpp
```diff
@@ -249,61 +247,6 @@ int main(int argc, char **argv)
                         s.removeSelectable(&eoiuCheckTimer);
                     }
                 }
-                else if (temps == &deviceMetadataTableSubscriber)
```
prsunny, line 254: Why is this removed? seems unrelated.

mike-dubrovsky:
@vpandian-nokia, I’m sorry — I initially posted an incomplete diff. I think this version is better and tested. I will address review comments in a day or two.

fpmsyncd/fpmsyncd.cpp
mike-dubrovsky: This fpmsyncd code subscribed to CONFIG_DB changes to dynamically toggle FIB suppression at runtime. After this fix, changes to the suppress-fib-pending configuration take effect only after the switch is restarted, so there is no longer any need to listen for CONFIG_DB changes.

fpmsyncd/fpmsyncd.cpp
```diff
@@ -110,12 +108,14 @@ int main(int argc, char **argv)
...
     std::string suppressionEnabledStr;
-    deviceMetadataTable.hget("localhost", "suppress-fib-pending", suppressionEnabledStr);
+    Table fibSuppressTable(&stateDb, STATE_FIB_SUPPRESS_TABLE_NAME);
+    fibSuppressTable.hget("system", "oper_state", suppressionEnabledStr);
     if (suppressionEnabledStr == "enabled")
     {
         routeResponseChannel = std::make_unique<NotificationConsumer>(&applStateDb, routeResponseChannelName);
         sync.setSuppressionEnabled(true);
     }
+    SWSS_LOG_NOTICE("FIB suppression oper_state: %s", suppressionEnabledStr.c_str());
```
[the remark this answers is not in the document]
mike-dubrovsky: The FIB_SUPPRESS_TABLE|system oper_state is written to STATE_DB by swss.sh before the swss container starts, as part of the companion PR [sonic-buildimage #26151](https://github.com/sonic-net/sonic-buildimage/pull/26151). Both orchagent and fpmsyncd read this value at startup.

orchagent/main.cpp
```diff
@@ -521,6 +523,9 @@ int main(int argc, char **argv)
          case 'M':
             macsec_post_enabled = true;
             break;
+        case 'F':
+            gEnableFibSuppress = true;
+            break;
```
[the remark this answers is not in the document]
mike-dubrovsky: The operational state is published to STATE_DB by swss.sh (in companion PR [sonic-buildimage #26151](https://github.com/sonic-net/sonic-buildimage/pull/26151)) before the swss container starts. orchagent does not need to write it — it receives the -F flag from orchagent.sh, which reads the same STATE_DB entry. This keeps the coordination in a single place (swss.sh) and avoids duplication.

orchagent/routeorch.cpp
```diff
@@ -3180,6 +3197,11 @@ void RouteOrch::publishRouteState(const RouteBulkContext& ctx, const ReturnCode&
 {
     SWSS_LOG_ENTER();
 
+    if (!gEnableFibSuppress)
+    {
+        return;
+    }
```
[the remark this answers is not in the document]
mike-dubrovsky: The existing tests already cover the enabled path where publishing actually matters. Adding a negative test to assert zero publish() calls would be testing a trivial early-return guard. I'll consider adding only if requested by human reviewer.

tests/mock_tests/routeorch_ut.cpp
```diff
@@ -12,6 +12,7 @@
 #include "bulker.h"
 
 extern string gMySwitchType;
+extern bool gEnableFibSuppress;
```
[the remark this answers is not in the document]
mike-dubrovsky: fixed

tests/test_route.py
```diff
@@ -1116,6 +1121,10 @@ def check_offloaded():
 
             # make sure route suppression is disabled
             dvs.runcmd("config suppress-fib-pending disabled")
+            dvs.stop_fpmsyncd()
+            dvs.stop_swss()
+            dvs.start_swss()
+            dvs.start_fpmsyncd()
```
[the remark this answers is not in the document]
mike-dubrovsky: The missing rc assertion on this runcmd call is pre-existing -- it was already like this before my change. I only added the swss/fpmsyncd restart sequence after it. Fixing the pre-existing cleanup style is out of scope for this PR.

mike-dubrovsky:
@dgsudharsan Could you please help me with a code review?

fpmsyncd/fpmsyncd.cpp
```diff
@@ -110,12 +108,14 @@ int main(int argc, char **argv)
     nlmsg_set_default_size(FPM_MAX_MSG_LEN);
 
     std::string suppressionEnabledStr;
-    deviceMetadataTable.hget("localhost", "suppress-fib-pending", suppressionEnabledStr);
+    Table fibSuppressTable(&stateDb, STATE_FIB_SUPPRESS_TABLE_NAME);
```
dgsudharsan: I recommend having the device-metadata table and if someone configures during runtime, lets throw a message in CLI saying please perform config save and config reload to take effect.

orchagent/main.cpp
```diff
@@ -91,7 +92,7 @@ bool isChassisDbInUse()
 
 void usage()
 {
-    cout << "usage: orchagent [-h] [-r record_type] [-d record_location] [-f swss_rec_filename] [-j sairedis_rec_filename] [-b batch_size] [-m MAC] [-i INST_ID] [-s] [-z mode] [-k bulk_size] [-q zmq_server_address] [-c mode] [-t create_switch_timeout] [-v VRF] [-I heart_beat_interval] [-R] [-M]" << endl;
+    cout << "usage: orchagent [-h] [-r record_type] [-d record_location] [-f swss_rec_filename] [-j sairedis_rec_filename] [-b batch_size] [-m MAC] [-i INST_ID] [-s] [-z mode] [-k bulk_size] [-q zmq_server_address] [-c mode] [-t create_switch_timeout] [-v VRF] [-I heart_beat_interval] [-R] [-M] [-F]" << endl;
```
dgsudharsan: Lets continue to use the config_db knob but make sure we add relevant message in CLI and documentation that they need to config save and reboot to take effect.

dgsudharsan (changes requested):

fpmsyncd/fpmsyncd.cpp
mike-dubrovsky: @dgsudharsan 
Thank you for the review; sorry for the delay with the reply.

I do keep the suppress-fib-pending config state in the DEVICE_METADATA table. The CLI (config suppress-fib-pending) writes to DEVICE_METADATA and prints a message telling the user to perform config save and config reload (or systemctl restart swss) for the change to take effect. This is in the sonic-utilities PR (#4361).

The reason I also introduced an oper state in STATE_DB is the following: between the time the user changes CONFIG_DB and actually restarts swss, if the BGP container restarts independently, fpmsyncd would read the new CONFIG_DB value while orchagent is still running with the old flag. If fpmsyncd thinks suppression is enabled but orchagent isn't sending route confirmations, BGP route propagation stalls (traffic black hole).

To solve this, swss.sh reads CONFIG_DB and writes the current operational FIB suppression state to STATE_DB (FIB_SUPPRESS_TABLE|system / oper_state) before orchagent and fpmsyncd start. All processes (orchagent, fpmsyncd, route_check) read the oper state from STATE_DB, so they always stay in sync regardless of CONFIG_DB changes.

orchagent/main.cpp
mike-dubrovsky: @dgsudharsan
Same as my reply above: the proposed fix does continue to use CONFIG_DB, and we do print a message in the CLI. 
However, if the user does not restart swss (or perform config save / config reload) right away, there is a vulnerability: if BGP crashes or the BGP container restarts, fpmsyncd and orchagent can get out of sync. So I introduced an oper state in STATE_DB to keep them in sync.

vganesan-nokia:
There is a PR #4437 that does async update of APPL_STATE_DB, notification and recording instead of completely skipping the update of APPL_STATE_DB. With PR #4437 we observed almost same performance with both suppress-fib-pending enabled and disabled while keeping APPL_STATE_DB contents and recordings intact. 
When compared to skipping completely, the async update has about 1.9% lower performance gain. However with  APPL_STATE_DB retained with both suppress-fib-pending enabled and disabled, this reduction in performance can be acceptable. I think we can have both of these approaches and choose based on use case. 
@deepak-singhal0408, any comments?

deepak-singhal0408:
Hi @mike-dubrovsky ,  This PR changes FIB suppression from a **runtime dynamic toggle** to a **startup-time flag** (`config reload` required). 
Conceptually, I like the intention, but would recommend getting sign-off from @stepanblyschak (the original author) as well. Also recommend to update the original HLD if we converge on this.

deepak-singhal0408:
Also @mike-dubrovsky, I don't think swss is the right backend for BGP config — it should not own the STATE_DB population.

Here are two potential options:

Design Proposal: FIB Suppress Config Backend & Conditional FRR Config

Currently bgpd.main.conf.j2 always includes bgp suppress-fib-pending — even when the feature is disabled. The current code works as a workaround to support runtime config changes, but if we get rid of that requirement, we should make this conditional as well.

Proposal: Make the FRR config conditional on CONFIG_DB, and ensure the right component owns the config lifecycle.

-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Common to both options: Conditional FRR template

 {% if DEVICE_METADATA['localhost']['suppress-fib-pending'] == 'enabled' %}
  bgp suppress-fib-pending
 {% endif %}

This eliminates the FRR ↔ feedback loop mismatch. config reload regenerates the template — this is why config reload (not just swss restart) should be the only option given to the user.

-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Option A: Direct CONFIG_DB read (Recommended — simplest)

Each consumer reads CONFIG_DB at startup. No new STATE_DB table, no intermediary.

 config reload
   ├─→ bgpd.main.conf.j2: conditional "bgp suppress-fib-pending"
   ├─→ fpmsyncd (BGP container, startup): reads CONFIG_DB → subscribes to response channel if enabled
   └─→ orchagent.sh (SWSS container, startup): reads CONFIG_DB → passes -F if enabled

Pros: Minimal code, no new tables, CONFIG_DB is populated before containers start (no race).

Option B: bgpcfgd → STATE_DB intermediary

bgpcfgd owns the config lifecycle, writes operational state to STATE_DB.

 config reload
   ├─→ bgpd.main.conf.j2: conditional "bgp suppress-fib-pending"
   ├─→ bgpcfgd (BGP container, startup): reads CONFIG_DB → writes STATE_DB FIB_SUPPRESS_TABLE
   ├─→ fpmsyncd (BGP container): subscribes to STATE_DB (SubscriberStateTable for startup race)
   └─→ orchagent (SWSS container): reads STATE_DB at startup

Pros: Follows config manager pattern, STATE_DB reflects operational truth. Cons: bgpcfgd/fpmsyncd startup race (same container, parallel supervisord), extra indirection.

let me know what you think..

cc @stepanblyschak @StormLiangMS @prsunny @dgsudharsan @vganesan-nokia

mike-dubrovsky:
Hi @vganesan-nokia

This set of PRs blocks both pub/sub and APPL_STATE_DB population when FIB suppression is disabled. 
We see a 2K to 4K improvement on one platform (all IPC is Redis) and 15K to 20K on another (with north and south ZMQ).

P.S. Plus the current Sonic memory consumption is about 4,852 bytes per route ...

deepak-singhal0408:
> Hi @vganesan-nokia
> 
> This set of PRs blocks both pub/sub and APPL_STATE_DB population when FIB suppression is disabled. We see a 2K to 4K improvement on one platform (all IPC is Redis) and 15K to 20K on another (with north and south ZMQ).
> 
> P.S. Plus the current Sonic memory consumption is about 4,852 bytes per route ...

agree.. we need not populate some DB for which there is no consumer.. That extra thread processing capacity populating APPL_STATE_DB(in async update case), could be used for other tasks.

mike-dubrovsky:
Hi @deepak-singhal0408,
Thank you for the detailed feedback and the design proposals.

Regarding @stepanblyschak sign-off : 
@dgsudharsan is also from NVIDIA and he said he will help with the review.

Why STATE_DB instead of Option A (direct CONFIG_DB read)

The problem is the window between the CLI command and the actual reload. When the user runs `config suppress-fib-pending enabled/disabled`, CONFIG_DB changes immediately, but orchagent is still running with the old flag. Until the user performs config reload, we are vulnerable to:
- Independent BGP container restart (crash, OOM, systemctl restart bgp): fpmsyncd starts fresh, reads the new CONFIG_DB value, but orchagent still has the old flag. If fpmsyncd expects route confirmations that orchagent isn't sending, BGP route propagation stalls (traffic black hole).
- route_check.py: If it reads CONFIG_DB directly, it could apply mitigation logic based on a config that doesn't match orchagent's actual running state.

One alternative would be to restart swss container immediately when the CLI command is entered -- then we could use Option A because there would be no vulnerability window. But an immediate orchagent restart would be disruptive (data plane impact), so I hesitated to take that approach.

STATE_DB avoids this entire class of problems: swss.sh writes the oper_state when swss restarts 
(whether via systemctl restart swss, config reload, or reboot).
Then swss.sh starts the swss and bgp containers. All consumers (orchagent, fpmsyncd, FRR template, route_check) read the same oper_state that reflects orchagent's actual running flag, regardless of what CONFIG_DB says.

I am happy to do immediate orchagent restart if you / Sudarshan okay with it. It would simplify things.

Why not Option B (bgpcfgd → STATE_DB)

bgpcfgd is not the only FRR config manager — if frr_mgmt_framework_config is set, the container runs frrcfgd instead. So bgpcfgd is not a universal source of truth. Additionally, bgpcfgd and fpmsyncd both run inside the BGP container and are started in parallel by supervisord, creating a startup race for who writes/reads STATE_DB first.

dgsudharsan:
@mike-dubrovsky  Lets discuss this in routing subgroup to understand feedback from others whether if restarting orchagent is preferred or we can stay with the current approach.

@eddieruan-alibaba @pbrisset  FYI

dgsudharsan:
@deepak-singhal0408 We discussed this briefly in AM routing subgroup on 5th May. The reason for this complexity is to handle a rare scenario where config is saved by services have not restarted and FRR docker crashes and restarts (or restarted by the user). This is a highly unlikely scenario and we can add this is not supported in the HLD. This would simply the design to just using the existing field in the config_db and we can avoid the new state_db table altogether. @mike-dubrovsky is fine with the approach.

Please let us know if you have any concerns.

mike-dubrovsky:
@dgsudharsan In the PM meeting, Deepak asked: what's the reason we implemented the SONiC FIB-suppression CLI so that it takes effect instantly, instead of waiting for the next config reload? Do you happen to know the history?

dgsudharsan:
> @dgsudharsan In the PM meeting, Deepak asked: what's the reason we implemented the SONiC FIB-suppression CLI so that it takes effect instantly, instead of waiting for the next config reload? Do you happen to know the history?

@prsunny @StormLiangMS Are you aware of the reason for the CLI. I believe it should be as part of the initial requirements for the suppress fib pending feature.

prsunny:
> > @dgsudharsan In the PM meeting, Deepak asked: what's the reason we implemented the SONiC FIB-suppression CLI so that it takes effect instantly, instead of waiting for the next config reload? Do you happen to know the history?
> 
> @prsunny @StormLiangMS Are you aware of the reason for the CLI. I believe it should be as part of the initial requirements for the suppress fib pending feature.

i need to check, @StormLiangMS , do you recollect?

mike-dubrovsky:
@deepak-singhal0408 @dgsudharsan Could you please take a look when you have time? I re-did the PR as per our discussion.

tests/mock_tests/routeorch_ut.cpp
```diff
@@ -750,10 +752,12 @@ namespace routeorch_test
         static_cast<Orch *>(gRouteOrch)->doTask();
 
         gMockResponsePublisher.reset();
+        gEnableFibSuppress = false;
     }
```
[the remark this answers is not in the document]
mike-dubrovsky: fixed

mike-dubrovsky:
@deepak-singhal0408 @dgsudharsan  
I updated all four PRs. They are ready for your review.

deepak-singhal0408:
@mike-dubrovsky , please fix DCO on all PRs.

mike-dubrovsky:
> @mike-dubrovsky , please fix DCO on all PRs.

@deepak-singhal0408 

done

deepak-singhal0408:
For the context: This was re-discussed in routing-wg on 07/09, and we have decided to go ahead with making suppress-fib-pending command static instead of runtime toggle.

Also to keep the design simple, both swss and bgp(fpmsyncd) will listen to this config knob during init and apply the behavior accordingly. Any edge case will be captured accordingly in https://github.com/sonic-net/SONiC/pull/2335 HLD.

cc: @StormLiangMS @prsunny for viz.

deepak-singhal0408:
@mike-dubrovsky , please fix the CI failure.

deepak-singhal0408:
**`Test vstest` is failing on this PR's own de-xfailed `test_offload`**

`tests/test_route.py::TestFpmSyncResponse::test_offload[enabled]` and `[disabled]` both fail deterministically (reran with identical result), at the route-injection step *before* any offload assertion runs:

```
test_route.py:1102  dvs.runcmd(f"ip route add {route} via 10.0.0.1 proto bgp")
test_route.py:1103  assert rc == 0, "Failed to configure route"
E   AssertionError: Failed to configure route
E   assert 2 == 0
# kernel: "Error: Nexthop has invalid gateway."
```

**Root cause** — timing against the newly added restart. The fixture programs `Ethernet0 = 10.0.0.0/31` (line 1050) via CONFIG_DB → intfmgrd before the test body. The test then restarts swss+fpmsyncd to apply the static config (lines 1092–1095) and waits a fixed `time.sleep(5)` (1096). That restart re-initializes `Ethernet0`, and 5 s is not reliably enough for intfmgrd to reprogram the interface IP back into the kernel. So when line 1102 injects `1.1.1.0/24 via 10.0.0.1`, `10.0.0.1` is not yet on-link and the kernel rejects the nexthop (`rc=2`). The test never reaches `is_offloaded()`, so it currently validates nothing.

**Potential fix** — after the restart, wait for `Ethernet0`'s address / nexthop to be reachable again instead of a fixed sleep, e.g. poll until the server responds before injecting the route (the fixture already establishes this at line 1058):

```python
# after start_fpmsyncd(), replace the bare time.sleep(5):
def _nexthop_ready():
    rc, _ = dvs.runcmd("ping -c 1 -W 1 10.0.0.1")
    return rc == 0
# poll _nexthop_ready() up to ~30s (e.g. wait_for_result / retry loop) before `ip route add ... via 10.0.0.1`
```

The same reprogram-wait is needed after the teardown `disabled` restart (lines 1126–1129) so the next parametrization starts clean. With that, `test_offload` will actually exercise the offload handshake and `vstest` should go green.

(Note: `TestAsan vstest` happens to pass on this commit, but the standard `Test vstest` shard fails deterministically — this is a real test-timing bug, not infra.)

mike-dubrovsky:
@deepak-singhal0408 thanks for your patience 

I've updated `TestFpmSyncResponse::test_offload` with  proposed ping wait loop
also added  `dvs.check_swss_ready()` — waits until orchagent has finished initializing ports

deepak-singhal0408:
@mike-dubrovsky, please rebase this PR. Need https://github.com/sonic-net/sonic-swss/pull/4761 change to be picked up for Ci to pass.

And also resolve addressed comments.

dgsudharsan (dismissed):

deepak-singhal0408:
**CI: `Test vstest` red — `test_route.py::TestFpmSyncResponse::test_offload[enabled]` and `[disabled]`** ([failed job](https://github.com/sonic-net/sonic-swss/runs/88800374661), failed all 3 retries).

Both params fail at the **precondition**, before any suppression is asserted:

```
self.wait_for_nexthop_reachable(dvs, "10.0.0.1")
AssertionError: Nexthop 10.0.0.1 not reachable after swss/fpmsyncd restart
ping: connect: Network is unreachable   (rc=2, for the full 60s poll)
```

`10.0.0.1` is the directly-connected neighbor reachable via Ethernet0's `10.0.0.0/31`. *"Network is unreachable"* means Ethernet0's interface IP isn't back in the kernel after the `stop_swss`/`start_swss` — i.e. `intfmgrd` hasn't reprogrammed the interface within the 60s window. Since it fails identically for `enabled` **and** `disabled`, it's independent of the suppression code path — looks like a readiness gap in the new restart scaffolding introduced by this PR.

tests/test_route.py
```diff
@@ -1077,22 +1077,52 @@ def is_offloaded(self, dvs, route):
...
 
-        time.sleep(5)
+        self.configure_orchagent_fib_suppress(dvs, suppress_state)
+
+        dvs.runcmd("config save -y")
+        dvs.restart()
+
+        self.wait_for_nexthop_reachable(dvs, "10.0.0.1")
 
         try:
```
deepak-singhal0408: The persistent setup changes above (`suppress-fib-pending`, the patched `/usr/bin/orchagent.sh`, and the DVS restart) occur before this `try`, so a restart or readiness failure can bypass `finally` and leave the DVS modified for later tests. Consider starting the protected block before the first mutation and checking the script-edit and cleanup return codes.

orchagent/main.cpp
```diff
@@ -594,6 +596,9 @@ int main(int argc, char **argv)
          case 'M':
             macsec_post_enabled = true;
             break;
+        case 'F':
```
deepak-singhal0408: The diff-coverage check reports 75% because the two uncovered changed lines are this `case` label and its `break`; the assignment between them is already marked `LCOV_EXCL_LINE`. Since the DVS test exercises orchagent with `-F`, consider applying the exclusion consistently to the full parser branch (lines 599–601), or otherwise adding parser-level coverage, so the required 80% gate passes.

mike-dubrovsky: fixed

tests/test_route.py
mike-dubrovsky: fixed

deepak-singhal0408 (approved):

deepak-singhal0408:
@dgsudharsan @prsunny please help on this with the re-review/sign-off/merge? Thanks,

dgsudharsan (approved):

PR #4337, opened 2026-03-13
[Port-Breakout] Fix warmboot-failure due to port count mismatch

Keshavg-marvell:
What is issue:

SONiC does not update the port count after DPB is done. Due to this warm-boot fails if it is done after port breakout is done with error like below.

sonic ERR swss#orchagent: :- bake: Invalid port table: portCount, expecting 66, got 67

portsyncd write the port count in PortConfigDone only at the time of init. but this count value is not updated with port delete/add events which happens during port breakout.

Fix:
Adjusting the port count based on netdev event in portsyncd with dynamic port addition and deleting trigger through dynamic port breakout.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


radha-danda:
@prsunny, can you please help review the PR?

prsunny:
Is this a valid scenario to have port breakout and then WB? Seems to be the deployment changed and then requiring hitless. @vaibhavhd for viz

Keshavg-marvell:
> Is this a valid scenario to have port breakout and then WB? Seems to be the deployment changed and then requiring hitless. @vaibhavhd for viz

 @prsunny This is a valid scenario. DPB may be performed, followed by a warm boot or warm upgrade. This scenario has been verified on Marvell switches, and there are no technical blockers to supporting this functionality.

radha-danda:
@prsunny, can you please review the PR and share your feedback?

Issue #4110, opened 2026-01-06
Azure Pipeline Build Failure - Repeated failure of error ""../../../config/test-driver: line 112: Segmentation fault "


rkavitha-hcl:
I'm not able to re-open Issue PR https://github.com/sonic-net/sonic-swss/issues/4032 and hence opening a new PR again. 

We are again getting the following error in PR https://github.com/sonic-net/sonic-swss/pull/3862. Kindly help to fix this issue permanently. 

2026-01-06T14:17:15.3573539Z ../../../config/test-driver: line 112: 15125 Segmentation fault      (core dumped) "$@" >> "$log_file" 2>&1
2026-01-06T14:17:15.3574852Z FAIL: p4orch_tests
2026-01-06T14:17:15.3894970Z ===========================================================
2026-01-06T14:17:15.3895469Z    sonic-swss 1.0: orchagent/p4orch/tests/test-suite.log
2026-01-06T14:17:15.3895855Z ===========================================================
2026-01-06T14:17:15.3896031Z 
2026-01-06T14:17:15.3896307Z # TOTAL: 1
2026-01-06T14:17:15.3896551Z # PASS:  0
2026-01-06T14:17:15.3896802Z # SKIP:  0
2026-01-06T14:17:15.3897070Z # XFAIL: 0
2026-01-06T14:17:15.3897339Z # FAIL:  1
2026-01-06T14:17:15.3897634Z # XPASS: 0
2026-01-06T14:17:15.3897891Z # ERROR: 0

@prsunny , can you please help here to close on this issue ASAP. 

Thanks and Warm Regards
Kavitha Ramalingam

prsunny:
Closing in favor of the newly tracking issue - https://github.com/sonic-net/sonic-swss/issues/4345

PR #1964, opened 2021-10-15
[SRV6] Sonic-swss changes for SRV6

kperumalbfn:
SONiC swss support for SRV6
- RouteOrch changes to trigger SRV6 nexthops and update route entries
- SRV6Orch changes to create SRV6 nexthops, tunnel and SRV6 MY_SID_ENTRY objects
- CRMOrch changes to include SRV6 MY_SID

sonic-swss-common changes for SRV6 APPL_DB tables: https://github.com/Azure/sonic-swss-common/pull/538
sonic-utiliities changes for SRV6 CRM: https://github.com/Azure/sonic-utilities/pull/1883
sonic-buildimage changes for SRV6 CRM: https://github.com/Azure/sonic-buildimage/pull/9004

SRV6 HLD: https://github.com/Azure/SONiC/pull/795

PR #4341, opened 2026-03-15
orchagent: Fix memory leak and UB in srv6orch sidlist handling

uramanath:
**What I did**

Fixed a memory leak and undefined behavior in `Srv6Orch::createUpdateSidList()` by replacing raw `new[]`/`delete` with `std::unique_ptr`.

**Why I did it**

`segment_list.list` is allocated with `new sai_ip6_t[count]` but on two SAI error paths (`create_srv6_sidlist` and `set_srv6_sidlist_attribute` failures) the function returns `false` without freeing the allocation. Additionally, the success path uses plain `delete` instead of `delete[]` for an array allocation, which is undefined behavior per the C++ standard.

**How I fixed it**

Replaced the raw `new[]` allocation with `std::unique_ptr<sai_ip6_t[]>` (RAII). The buffer is now freed automatically on every exit path — no manual `delete` calls needed. This is leak-proof by construction, even if new error paths are added in the future.

**How I verified it**

Code inspection — the fix is mechanical. Pass all existing test cases. No behavioral change on the happy path.

### DIFF
--- a/orchagent/srv6orch.cpp
+++ b/orchagent/srv6orch.cpp
@@ -2,6 +2,7 @@
 #include <sstream>
 #include <inttypes.h>
 #include <iterator>
+#include <memory>
 
 #include "routeorch.h"
 #include "logger.h"
@@ -1055,7 +1056,8 @@ bool Srv6Orch::createUpdateSidList(const string sid_name, const string sid_list,
         return true;
     }
     SWSS_LOG_INFO("Segment count %d", segment_list.count);
-    segment_list.list = new sai_ip6_t[segment_list.count];
+    auto segment_buf = std::make_unique<sai_ip6_t[]>(segment_list.count);
+    segment_list.list = segment_buf.get();
     uint32_t index = 0;
 
     for (string ip_str : sid_ips)
@@ -1112,7 +1114,6 @@ bool Srv6Orch::createUpdateSidList(const string sid_name, const string sid_list,
             return false;
         }
     }
-    delete segment_list.list;
     return true;
 }
 


uramanath:
The BuildAsan failure is unrelated to this change — it's a segfault in `p4orch_tests` (same as #4110). This PR only modifies `orchagent/srv6orch.cpp`. All other builds (amd64, arm64, armhf, Trixie, Docker) and vstest pass. Could a maintainer please re-trigger with `/azp run`? Thank you.

skthodupunoori-upscaleai (commented):
Good catch on both the leak and the `delete` vs `delete[]` UB — these are real bugs. The fix is correct as-is, but here are a few suggestions to make it more robust and land stronger:

### 1. Consider RAII instead of manual `delete[]`

The current fix is correct today, but it's fragile: the next person who adds an error path to this function could easily forget the `delete[]` and recreate the exact same bug. You can eliminate the entire class of problem by letting `std::unique_ptr` own the allocation:

```cpp
#include <memory>
// ...
auto segment_buf = std::unique_ptr<sai_ip6_t[]>(new sai_ip6_t[segment_list.count]);
segment_list.list = segment_buf.get();
```

Then remove **all** `delete`/`delete[]` calls — the destructor handles every exit path automatically. No `delete[]` before early returns, no risk of wrong `delete` form, zero chance of future leaks from this allocation. The SAI APIs only borrow the pointer during the call, so lifetime is fine.

### 2. The `delete` vs `delete[]` UB on the success path deserves more emphasis

The PR description frames the `delete` → `delete[]` fix as secondary, but it's actually the more dangerous issue — it fires on **every successful call**, not just error paths. For POD types like `sai_ip6_t` many allocators happen to tolerate this, but it's formally undefined behavior and could corrupt the heap on certain platforms/allocators. Worth calling this out as the primary fix.

### 3. Test coverage for error paths

Since existing tests didn't catch either bug, it would be worth adding a vstest that mocks `sai_srv6_api->create_srv6_sidlist` to return `SAI_STATUS_FAILURE`, confirming the error path doesn't leak. Even a basic test that exercises the failure branch would prevent regression. (I know BuildAsan is currently broken per #4110 — but that's an argument for fixing it alongside this, not skipping validation.)

### 4. Audit similar patterns in orchagent

There are likely other functions in orchagent that follow the same `new[]` → SAI call → early return pattern. A quick `grep -rn 'new sai_' orchagent/` would surface them. If this is systemic, it's worth noting in the PR ("I audited X other sites and found none / N others I'll fix in follow-ups"). Reviewers love seeing that.

---

Overall this is a solid find — just think the RAII approach would make it land cleaner and set a better pattern for the codebase. Nice work 👍

orchagent/srv6orch.cpp
```diff
@@ -1093,6 +1093,7 @@ bool Srv6Orch::createUpdateSidList(const string sid_name, const string sid_list,
         if (status != SAI_STATUS_SUCCESS)
         {
             SWSS_LOG_ERROR("Failed to create srv6 sidlist object, rv %d", status);
+            delete[] segment_list.list;
```
skthodupunoori-upscaleai: This works, but with RAII (`std::unique_ptr<sai_ip6_t[]>` owning the buffer at the allocation site), this line and the matching one in the `set` path wouldn't be needed at all — the destructor handles cleanup on all exit paths.

Concretely, replace the allocation with:
```cpp
auto segment_buf = std::unique_ptr<sai_ip6_t[]>(new sai_ip6_t[segment_list.count]);
segment_list.list = segment_buf.get();
```
Then delete all three `delete`/`delete[]` calls. The function becomes leak-proof by construction, even if someone adds new error paths in the future.

(`<memory>` isn't currently included in this file — you'd need to add it.)

orchagent/srv6orch.cpp
```diff
@@ -1110,10 +1111,11 @@ bool Srv6Orch::createUpdateSidList(const string sid_name, const string sid_list,
         if (status != SAI_STATUS_SUCCESS)
         {
             SWSS_LOG_ERROR("Failed to set srv6 sidlist object with new segments, rv %d", status);
+            delete[] segment_list.list;
             return false;
         }
     }
-    delete segment_list.list;
+    delete[] segment_list.list;
```
skthodupunoori-upscaleai: nit: Worth noting in the commit message that this was UB on **every successful call** (not just a style issue). `delete` on a `new[]` allocation is undefined behavior per the C++ standard, even for POD types. Some allocators silently tolerate it for trivially-destructible types, but others will corrupt the heap.

With the `unique_ptr` approach suggested above, this line disappears entirely.

skthodupunoori-upscaleai:
**Follow-up on my review above** — after looking more closely at the broader codebase, I want to add some context to my earlier suggestions:

The RAII (`unique_ptr`) suggestion I made is technically sound, but I should note that **every SAI list buffer in orchagent uses the same raw `new[]`/`delete[]` pattern** — `qosorch.cpp` alone has 10+ instances, and `nhgmaporch.cpp` has it too. The QoS code even has a dedicated `freeAttribResources()` method to centralize cleanup rather than using smart pointers. So the pattern in this PR is consistent with the rest of the codebase, not an oversight specific to this change.

Asking this bugfix to be the one that introduces RAII for SAI buffers is scope creep for what is a correct, minimal, 3-line fix. The `unique_ptr` conversion and a codebase-wide audit would be better as a separate follow-up effort.

**What stands from my original review:** the `delete` to `delete[]` fix on the success path is the higher-severity item (UB on every successful call, not just error paths) and is worth emphasizing in the commit message.

Nice catch on a bug that has been in the codebase since the original SRV6 commit (#1964, Nov 2021). 👍

uramanath:
@abdosi Did you get a chance to review?

prsunny (approved):

PR #4346, opened 2026-03-17
orchagent: Fix off-by-one in ASIC instance_id length check

uramanath:
**What I did**

Fixed an off-by-one error in the ASIC instance ID length validation in `orchagent/main.cpp`.

**Why I did it**

`strnlen(optarg, SAI_MAX_HARDWARE_ID_LEN)` returns at most `SAI_MAX_HARDWARE_ID_LEN`. When the input exceeds this limit, `optarg[len]` is the first character beyond the scanned range — this is the character that should be checked to detect truncation. The current code checks `optarg[len+1]`, which skips one character and can read past a null terminator.

**How I verified it**

Code inspection. The fix changes `optarg[len+1]` to `optarg[len]` — a single-character index correction. The truncation warning now triggers correctly when the input exceeds `SAI_MAX_HARDWARE_ID_LEN`. No behavioral change for inputs within the limit.


dgsudharsan (dismissed):

uramanath:
@prsunny Please take a look

prsunny:
@uramanath , possible to fix coverage by adding a unit test?

PR #4352, opened 2026-03-17
[AclOrch] Updates Mirror and MirrorV6 Ingress Stage to Accept Egress Action

AndoniSanguesa:
What I did
Added Mirror Egress Action to set of available actions for Ingress Stage Acl tables of type Mirror and MirrorV6

Why I did it
Broadcom has specified that for TH5, they only support egress on the ingress ACL pipeline. There is no ASIC support for any mirroring actions on the egress ACL pipeline.

How I verified it
Installed on 7060x6 and verified that acl rule was accepted after change. Installed on a mellanox device as well to verify the rule is still rejected on a different sku. Mirroring occurred as expected on 7060x6, no regressions were observed.

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -260,7 +260,8 @@ static acl_table_action_list_lookup_t defaultAclActionList =
             {
                 ACL_STAGE_INGRESS,
                 {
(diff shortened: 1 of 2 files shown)


vmittal-msft:
@AndoniSanguesa please share master pr

vmittal-msft (approved):

PR #4355, opened 2026-03-17
Create new V6ACL tables with upper V6 words

kperumalbfn:
- Added new ACL fields to match V6 SIP/DIP WORD0/1/2/3.
- Created new V6 ACL 'lite' tables to match only few fields of IPV6 header.
- Added tests for the new ACL tables


kperumalbfn:
Will open a new one after validation

PR #4356, opened 2026-03-17
Fix unnecessary route retry due to premature VRF table eviction in bu…

ypcisco:
…lk processing

**What I did**

- Added cleanupVrfTable(vrf_id) for targeted per-VRF cleanup in RouteOrch
- Added cleanupEmptyVrfTables() to use iterator-based erase.

**Why I did it**

- In bulk processing, with DEL and SET operations on the same VRF.
- If DEL empties a VRF table, the VRF entry is evicted from m_syncdRoutes
- The subsequent addRoutePost() for a SET in the same VRF fails to find the VRF in cache and returns false — the routes are programmed in ASIC.
- The subsequent bulk retry fails as entry already exists and the status is returned as SAI_STATUS_NOT_EXECUTED for remaining routes, these routes are removed and never tracked in m_syncdRoutes.
- When a DEL arrives later for that route, removeRoute() finds no VRF entry in m_syncdRoutes and returns true (no-op), leaving a stale route permanently in ASIC

**How I verified it**

Added testcase test_RouteVrfTableCleanupRaceCondition for the same.

**Details if related**


orchagent/routeorch.cpp
```diff
@@ -3001,16 +2991,44 @@ bool RouteOrch::removeRoutePost(const RouteBulkContext& ctx)
...
+    }
+
+    return true;
+}
+
+void RouteOrch::cleanupEmptyVrfTables()
+{
+    SWSS_LOG_ENTER();
+
+    for (auto it_vrf = m_syncdRoutes.begin(); it_vrf != m_syncdRoutes.end();)
```
anish-n: This adds an all vrf iteration on every route operation so there will be a high performance impact of this change, especially for cases where there are 1-2k vrfs on the device. 

To address this, I suggest that we create an unordered_set<sai_object_id_t> m_pendingEmptyVrfCleanup; and update this set in removeRoutePost and process it once at end of  doTask, that way we only touch the ones which actually need the delete and generate the delete list dynamically

anish-n (changes requested):

orchagent/vnetorch.cpp
```diff
@@ -1682,6 +1682,8 @@ bool VNetRouteOrch::setAndDeleteRoutesWithRouteOrch(const sai_object_id_t vr_id,
         if (gRouteOrch->removeRoutePost(ctx))
         {
             SWSS_LOG_NOTICE("Route %s removed via routeorch for vnet %s", ipPrefix.to_string().c_str(), vnet_name.c_str());
+            // Cleanup VRF if empty.
+            gRouteOrch->cleanupVrfTable(vr_id);
```
anish-n, line 1686: As of today vnetorch does not actually bulk multiple operations together so vnetorch routes are actually not subject to the problem

orchagent/routeorch.cpp
anish-n: Or alternatively we do not fail the call in addRoutePost when the vrf is absent(because a prior removeRoutePost removed the vrf because it found 0 routes), but we just simply add the vrf back in in the addRoutePost call

orchagent/vnetorch.cpp
ypcisco: thats true but keeping in sync with routeorch in case of bulker add to vnetorch

orchagent/routeorch.cpp
ypcisco: updated to use a vector to track the pending vrfs for cleanup, the alternate method is susceptible to vrf refcount leaks in case of addRoutPost premature failures.

PR #4358, opened 2026-03-18
[Hw PFCWD feature] Refactor existing pfcwd code

abhishek-nexthop:
**What I did**

- Created new base class PfcWdBaseOrch in pfcwdorch.cpp containing common functionality shared between software and hardware PFC watchdog implementations
- Moved software-specific PFC watchdog implementation (PfcWdSwOrch) to new file pfcwdsworch.cpp with corresponding header pfcwdsworch.h
- Renamed template class PfcWdOrch to PfcWdBaseOrch to better reflect its role as a base class
- Updated Makefile.am to include new pfcwdsworch.cpp source file
- Updated includes in orchdaemon.h and test files to reference the new header structure
- Maintained all existing functionality including Big Red Switch mode, storm detection/restoration, and warm-reboot support

**Why I did it**
Refactor PFC Watchdog code to improve maintainability by splitting base class and software implementation in different file. Base file now can be extended for hardware based implementation.

**How I verified it**
Build the orchagent and ran existing pfc test to check any change in behavior.

HLD: [2159](https://github.com/sonic-net/SONiC/pull/2159)

### DIFF
--- a/orchagent/pfcwdorch.cpp
+++ b/orchagent/pfcwdorch.cpp
@@ -38,8 +38,7 @@ extern event_handle_t g_events_handle;
 extern SwitchOrch *gSwitchOrch;
 extern PortsOrch *gPortsOrch;
 
-template <typename DropHandler, typename ForwardHandler>
-PfcWdOrch<DropHandler, ForwardHandler>::PfcWdOrch(DBConnector *db, vector<string> &tableNames):
+PfcWdBaseOrch::PfcWdBaseOrch(DBConnector *db, vector<string> &tableNames):
     Orch(db, tableNames),
     m_countersDb(new DBConnector("COUNTERS_DB", 0)),
     m_countersTable(new Table(m_countersDb.get(), COUNTERS_TABLE)),
@@ -54,14 +53,43 @@ PfcWdOrch<DropHandler, ForwardHandler>::PfcWdOrch(DBConnector *db, vector<string
 }
 
 
-template <typename DropHandler, typename ForwardHandler>
-PfcWdOrch<DropHandler, ForwardHandler>::~PfcWdOrch(void)
+PfcWdBaseOrch::~PfcWdBaseOrch(void)
 {
     SWSS_LOG_ENTER();
 }
 
-template <typename DropHandler, typename ForwardHandler>
-void PfcWdOrch<DropHandler, ForwardHandler>::doTask(Consumer& consumer)
+bool PfcWdBaseOrch::getLosslessTcsForPort(const Port& port, set<uint8_t>& losslessTc)
+{
+    SWSS_LOG_ENTER();
+
+    // Get PFC mask to identify lossless queues
+    uint8_t pfcMask = 0;
+    if (!gPortsOrch->getPortPfcWatchdogStatus(port.m_port_id, &pfcMask))
+    {
+        SWSS_LOG_ERROR("Failed to get PFC mask on port %s", port.m_alias.c_str());
+        return false;
+    }
+
+    // Collect lossless TCs (traffic classes with PFC enabled)
+    losslessTc.clear();
+    for (uint8_t i = 0; i < PFC_WD_TC_MAX; i++)
+    {
+        if ((pfcMask & (1 << i)) != 0)
+        {
+            losslessTc.insert(i);
+        }
+    }
+
+    if (losslessTc.empty())
+    {
+        return false;
+    }
+
+    SWSS_LOG_INFO("Found %zu lossless TCs on port %s", losslessTc.size(), port.m_alias.c_str());
+    return true;
+}
+
+void PfcWdBaseOrch::doTask(Consumer& consumer)
 {
     SWSS_LOG_ENTER();
 
@@ -121,24 +149,7 @@ void PfcWdOrch<DropHandler, ForwardHandler>::doTask(Consumer& consumer)
     }
 }
 
-template <typename DropHandler, typename ForwardHandler>
-template <typename T>
-unordered_set<string> PfcWdSwOrch<DropHandler, ForwardHandler>::counterIdsToStr(
-    const vector<T> ids, string (*convert)(T))
-{
-    SWSS_LOG_ENTER();
-    unordered_set<string> counterIdSet;
-
-    for (const auto& i: ids)
-    {
-        counterIdSet.emplace(convert(i));
-    }
-
-    return counterIdSet;
-}
-
-template <typename DropHandler, typename ForwardHandler>
-PfcWdAction PfcWdOrch<DropHandler, ForwardHandler>::deserializeAction(const string& key)
+PfcWdAction PfcWdBaseOrch::deserializeAction(const string& key)
 {
     SWSS_LOG_ENTER();
 
@@ -157,8 +168,7 @@ PfcWdAction PfcWdOrch<DropHandler, ForwardHandler>::deserializeAction(const stri
     return actionMap.at(key);
 }
 
-template <typename DropHandler, typename ForwardHandler>
-string PfcWdOrch<DropHandler, ForwardHandler>::serializeAction(const PfcWdAction& action)
+string PfcWdBaseOrch::serializeAction(const PfcWdAction& action)
 {
     SWSS_LOG_ENTER();
 
@@ -178,8 +188,7 @@ string PfcWdOrch<DropHandler, ForwardHandler>::serializeAction(const PfcWdAction
 }
 
 
-template <typename DropHandler, typename ForwardHandler>
-task_process_status PfcWdOrch<DropHandler, ForwardHandler>::createEntry(const string& key,
+task_process_status PfcWdBaseOrch::createEntry(const string& key,
         const vector<FieldValueTuple>& data)
 {
     SWSS_LOG_ENTER();
@@ -319,8 +328,7 @@ task_process_status PfcWdOrch<DropHandler, ForwardHandler>::createEntry(const st
     return task_process_status::task_success;
 }
 
-template <typename DropHandler, typename ForwardHandler>
-task_process_status PfcWdOrch<DropHandler, ForwardHandler>::deleteEntry(const string& name)
+task_process_status PfcWdBaseOrch::deleteEntry(const string& name)
 {
     SWSS_LOG_ENTER();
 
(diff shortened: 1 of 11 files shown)


orchagent/pfcwdorch.cpp
```diff
@@ -54,14 +53,43 @@ PfcWdOrch<DropHandler, ForwardHandler>::PfcWdOrch(DBConnector *db, vector<string
...
-template <typename DropHandler, typename ForwardHandler>
-PfcWdOrch<DropHandler, ForwardHandler>::~PfcWdOrch(void)
+PfcWdBaseOrch::~PfcWdBaseOrch(void)
 {
     SWSS_LOG_ENTER();
 }
 
-template <typename DropHandler, typename ForwardHandler>
-void PfcWdOrch<DropHandler, ForwardHandler>::doTask(Consumer& consumer)
+bool PfcWdBaseOrch::getLosslessTcsForPort(const Port& port, set<uint8_t>& losslessTc)
```
pinky-nexthop, line 62: Please create a section of "helper functions" grouped together that will be used by both hw/sw recovery. 

// Helper functions used by both classes
.... <function1>
... < function 2> 
....
// End helper functions.

pinky-nexthop:
@radha-danda
 @kperumalbfn 
 Please review this PR. The next set of PRs will add the h/w recovery code but this is a pre-requisite to them. 
 This PR mostly refactors the code as per the implementation plan discussed in the HLD [2159](https://github.com/sonic-net/SONiC/pull/2159)

orchagent/pfcwdorch.cpp
abhishek-nexthop: ACK

vmittal-msft:
@saksarav-nokia @ysmanman  please help review

vmittal-msft (commented):
please check below. 
Image

vmittal-msft:
@abhishek-nexthop @pinky-nexthop 

Since this is important feature as well as touching critical code base: please check following for all HW PFCWD PRs -
1. Make sure all sonic-mgmt/snappi tests pass after these changes related to PFCWD
2. Make sure these tests pass for XGS/DNX both type of SKUs
3. Make sure above tests work fine in both SW and HW based model.
4. Can you please help me remember if there is any knob available to switch backto SW mode in case of any issues ?

vmittal-msft:
@saksarav-nokia @ysmanman @arista-nwolfe

vmittal-msft:
@abhishek-nexthop please resolve conflicts.

prsunny:
Im not clear on the overall approach on this feature, this PR plus  https://github.com/sonic-net/sonic-swss/pull/4412, https://github.com/sonic-net/SONiC/pull/2159.

@pinky-nexthop , @abhishek-nexthop , let me know how this can be resolved. @kperumalbfn , @vmittal-msft, @rlhui @eddieruan-alibaba  for viz

orchagent/pfcwdsworch.cpp
```diff
@@ -0,0 +1,764 @@
...
+#include "sai_serialize.h"
+#include "portsorch.h"
+#include "converter.h"
+#include "redisapi.h"
+#include "select.h"
+#include "notifier.h"
+#include "schema.h"
+#include "subscriberstatetable.h"
+
+#define PFC_WD_GLOBAL                   "GLOBAL"
```
saksarav-nokia: Some of these #define's are already defined in pfcwdorch.cpp. can't we reuse?

orchagent/pfcwdorch.h
```diff
@@ -14,6 +14,7 @@ extern "C" {
 }
 
 #define PFC_WD_FLEX_COUNTER_GROUP       "PFC_WD"
+#define PFC_WD_TC_MAX                   8
```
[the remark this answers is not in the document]
abhishek-nexthop: PFC_WD_TC_MAX is intentionally defined in the base header pfcwdorch.h because it's a constant shared by all PFCWD translation units: the base (pfcwdorch.cpp), the software orch (pfcwdsworch.cpp), and the hardware orch (pfcwdhworch.cpp) added in the follow-up PRs. Centralizing it here keeps a single source of truth rather than duplicating it per file. There's no -Werror redefinition break since every definition expands to the same value (8), which is a benign redefinition. I've removed the leftover #define in pfcwdorch.cpp so this header is the only definition. (countercheckorch.h keeps its own pre-existing copy, same value.)

orchagent/pfcwdsworch.cpp
```diff
@@ -0,0 +1,764 @@
...
+#define PFC_WD_IN_STORM                 "storm"
+
+#define PFC_WD_DETECTION_TIME_MAX       (5 * 1000)
+#define PFC_WD_DETECTION_TIME_MIN       100
+#define PFC_WD_RESTORATION_TIME_MAX     (60 * 1000)
+#define PFC_WD_RESTORATION_TIME_MIN     100
+#define PFC_WD_POLL_TIMEOUT             5000
+#define SAI_PORT_STAT_PFC_PREFIX        "SAI_PORT_STAT_PFC_"
+#define COUNTER_CHECK_POLL_TIMEOUT_SEC  1
+
```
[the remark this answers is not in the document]
abhishek-nexthop: As stated in above comment we should keep def in base instead of sw as it is common for both sw and hw orch.

orchagent/pfcwdsworch.cpp
```diff
@@ -0,0 +1,764 @@
...
+                "" :
+                to_string(restorationTime * 1000));
+        countersFieldValues.emplace_back("PFC_WD_ACTION", this->serializeAction(action));
+        countersFieldValues.emplace_back("PFC_STAT_HISTORY", pfcStatHistory);
+
+        this->getCountersTable()->set(queueIdStr, countersFieldValues);
+
+        // We register our queues in PFC_WD table so that syncd will know that it must poll them
+        string key = getFlexCounterTableKey(queueIdStr);
+
```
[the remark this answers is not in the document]
abhishek-nexthop: This is day 0 implementation of pfcwdsworch. We just moved it to different file.

orchagent/pfcwdsworch.cpp
```diff
@@ -0,0 +1,764 @@
...
+    m_applDb(make_shared<DBConnector>("APPL_DB", 0)),
+    m_applTable(make_shared<Table>(m_applDb.get(), APP_PFC_WD_TABLE_NAME "_INSTORM"))
+{
+    SWSS_LOG_ENTER();
+
+    string detectSha, restoreSha;
+    string detectPluginName = "pfc_detect_" + this->m_platform + ".lua";
+    string restorePluginName;
+    string pollIntervalStr = to_string(m_pollInterval);
+    string plugins;
```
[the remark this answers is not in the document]
abhishek-nexthop: This is day 0 implementation of pfcwdsworch. We just moved it to different file.

abhishek-nexthop:
> please check below. Image

1. report_pfc_storm: This function simply print the log for storm detection as we want it to be common for both sw and hw moved it to base class. But there is no concept of entry in case of HwOrch so we unpacked the data that needed to be printed.
2. getLosslessTcsForPort: Similar to point 1. This will be common for both hw and sw thats why created a function in base class instead.
3. doTask: That is the case and intention as APPL_DB are only related to sw orch(lua script entries) so it will be only used in pfcwdsworch.
4. warmReboot: This is day 0 behavior of sw orch. Logic is just moved to different file.
5. Template: Same as 4 day 0 code.
6. report_pfc_restored: Verified works correctly
7. Big red switch: Day 0 nothing changed in this logic.

orchagent/pfcwdsworch.cpp
abhishek-nexthop: @saksarav-nokia Yes, we have moved only the software-specific implementation to PfcWdSwOrch. The common functionality will remain in the PfcWdOrch (base) class, while the hardware-specific macro will reside in its respective class.

vmittal-msft:
> @abhishek-nexthop @pinky-nexthop
> 
> Since this is important feature as well as touching critical code base: please check following for all HW PFCWD PRs -
> 
> 1. Make sure all sonic-mgmt/snappi tests pass after these changes related to PFCWD
> 2. Make sure these tests pass for XGS/DNX both type of SKUs
> 3. Make sure above tests work fine in both SW and HW based model.
> 4. Can you please help me remember if there is any knob available to switch backto SW mode in case of any issues ?

@abhishek-nexthop  @pinky-nexthop please share result for these.

vmittal-msft (commented):
## Code Review: PFCWD Refactor

### 🔴 Critical (Build-Breaking)

**1. `pollIntervalStr` unused variable — `pfcwdsworch.cpp:386`**
```cpp
string pollIntervalStr = to_string(m_pollInterval);  // declared but never used
```
The old code passed this to FlexCounter; the new `FlexCounterTaggedCachedManager` constructor takes `m_pollInterval` directly. This dead variable **will fail `-Werror=unused-variable`**.
**Fix:** Remove the variable.

---

**2. `key` unused variable — `pfcwdsworch.cpp:266`**
```cpp
string key = getFlexCounterTableKey(queueIdStr);  // computed but never used
```
The old FlexCounter API used `key`; the new API takes `queueId` directly. The variable and stale comment above it are dead code.
**Fix:** Remove both the variable and its stale comment.

---

**3. `PFC_WD_TC_MAX` triple-definition — build failure**

`orchdaemon.h` (modified by this PR) now includes:
- Line 27: `pfcwdorch.h` → defines `PFC_WD_TC_MAX 8`
- Line 36: `countercheckorch.h` → also defines `PFC_WD_TC_MAX 8`

Additionally, `pfcwdorch.cpp` still has `#define PFC_WD_TC_MAX 8` at line 32, so including `pfcwdorch.h` in `pfcwdorch.cpp` causes a redefinition within the same TU.

With `-Wmacro-redefined -Werror`, any TU including `orchdaemon.h` will fail to compile.
**Fix:** Remove `#define PFC_WD_TC_MAX 8` from `pfcwdorch.cpp` (it's now in the header). Then coordinate with `countercheckorch.h` — it should include `pfcwdorch.h` instead of redefining the constant, or both should pull from a shared `pfcwdcommon.h`.

---

### 🟠 High (Design / Correctness)

**4. All 14 macros duplicated verbatim in both `.cpp` files**

Every `#define` in `pfcwdorch.cpp` lines 15–29 is copy-pasted identically into `pfcwdsworch.cpp` lines 15–29 (e.g., `PFC_WD_DETECTION_TIME_MIN/MAX`, `PFC_WD_RESTORATION_TIME_MIN/MAX`, `BIG_RED_SWITCH_FIELD`, `PFC_WD_IN_STORM`, etc.). These were **not duplicated before** — `pfcwdsworch.cpp` is new, so this is freshly introduced by this PR.

This defeats the refactor's goal: if bounds or string keys change, they must be updated in two places. They belong in a shared internal header or all in the base `.cpp`.

> The response that this is "day 0 code, just moved" is inaccurate for this specific issue — the macros only existed in `pfcwdorch.cpp` before and are newly duplicated here.

---

**5. Global statics defined in the header — `pfcwdorch.h`**
```cpp
const string pfc_wd_flex_counter_group = PFC_WD_FLEX_COUNTER_GROUP;
static const map<string, sai_packet_action_t> packet_action_map = {...};
```
Both are defined in the header, so each including TU gets its own copy (code bloat, ODR risk for the `const string`). They should be `extern const` declared in the header and defined once in `pfcwdorch.cpp`.

---

**6. Validation bounds macros duplicated across both `.cpp` files**

`PFC_WD_DETECTION_TIME_MIN/MAX` and `PFC_WD_RESTORATION_TIME_MIN/MAX` are used in `PfcWdBaseOrch::createEntry` (in `pfcwdorch.cpp`) but duplicated in `pfcwdsworch.cpp` too. If bounds need to differ for HW PFCWD they'll silently diverge. These should live only where they're used.

---

### 🟡 Medium (Code Quality / Consistency)

**7. `PfcDlrPacketAction` member naming — `pfcwdorch.h`**
```cpp
PfcWdAction PfcDlrPacketAction = PfcWdAction::PFC_WD_ACTION_UNKNOWN;
```
Every other private member uses `m_` prefix (`m_countersDb`, `m_countersTable`, `m_pfcwd_ports`, `m_platform`). This looks like a public member and is inconsistent.
**Fix:** Rename to `m_pfcDlrPacketAction`.

---

**8. `report_pfc_restored` doesn't publish an event**

`report_pfc_storm` calls `event_publish(g_events_handle, "pfc-storm", &params)` but `report_pfc_restored` only logs. If this asymmetry is intentional (pre-existing behavior), please add a comment explaining why. If it's an oversight, a `"pfc-storm-restored"` event should be published symmetrically.

---

**9. Redundant `allPortsReady()` check — `pfcwdsworch.cpp`**

`PfcWdSwOrch::doTask(Consumer&)` calls `PfcWdBaseOrch::doTask(consumer)` (which already guards with `allPortsReady()`) then checks `allPortsReady()` again. The redundant check is harmless but confusing — a reader must trace into the parent to understand why the early-return case can't be reached at that point.

---

### 🔵 Low / Nit

**10. `//XXX Add port/queue state change event handlers` — `pfcwdsworch.h`**
`XXX` comments in a public header should be converted to tracked GitHub issues or removed.

---

### ✅ What's Done Well
- Base class extraction is logically clean — shared `doTask`, `createEntry`/`deleteEntry`, `deserializeAction`/`serializeAction`, and helper functions are correctly placed.
- Template explicit instantiations are correct and complete at the bottom of `pfcwdsworch.cpp`.
- Warm-reboot `bake()` logic correctly moved to `PfcWdSwOrch`.
- Build system (`Makefile.am`) correctly updated for both orchagent and mock tests.

---

### Priority fixes before merge
1. Remove `pollIntervalStr` — build failure (`-Werror`)
2. Remove unused `key` variable — build failure (`-Werror`)
3. Fix `PFC_WD_TC_MAX` redefinition in `pfcwdorch.cpp` — build failure (`-Werror`)
4. Consolidate shared macros — avoids silent divergence between SW and HW implementations

vmittal-msft:
> Im not clear on the overall approach on this feature, this PR plus #4412, [sonic-net/SONiC#2159](https://github.com/sonic-net/SONiC/pull/2159).
> 
> @pinky-nexthop , @abhishek-nexthop , let me know how this can be resolved. @kperumalbfn , @vmittal-msft, @rlhui @eddieruan-alibaba for viz

@prsunny i have approved HLD for this feature and we are working on reviewing PR list. If @kperumalbfn @eddyk-nvidia can approve HLD then we can focus on reviewing PRs in the order of their dependency.

vmittal-msft:
@abhishek-nexthop @pinky-nexthop please let me know if all comments has been taken care ?

pinky-nexthop:
> report_pfc_restored

@vmittal-msft Latest commit fixes the comments. Kindly have a look. Thanks!

pinky-nexthop (approved):

vmittal-msft (commented):
Reviewed this as a pure move-refactor (splitting the templated `PfcWdOrch` into a non-template base `PfcWdBaseOrch` in `pfcwdorch.cpp/h` and moving the software implementation `PfcWdSwOrch` into new `pfcwdsworch.cpp/h`). Verified faithfulness at the line level, not just by reading the diff.

**LGTM** — no behavioral changes, structurally sound.

### What I verified
- **Function inventory** — every function from the original `pfcwdorch.cpp` maps to a location in the new files; nothing dropped or duplicated. `counterIdsToStr` moved to the base (now an inline static template), `report_pfc_storm` moved to the base with a field-based signature, and `report_pfc_restored` / `getLosslessTcsForPort` were extracted from previously-inline code.
- **Line-level equivalence** — after normalizing class names and doing a set-diff of old vs. new (base + sw combined), the only differences are the expected transformations. All other function bodies are byte-identical (indentation preserved). The `report_pfc_storm` event params still map 1:1 (`entry->first→queueId`, `entry->second.portId→portId`, `entry->second.index→queueIndex`, `entry->second.portAlias→portAlias`), and `report_pfc_restored` preserves the exact restore log.
- **Macros** — `PFC_WD_TC_MAX` now has a single definition in `orch.h` (removed from `countercheckorch.h`, which still includes `orch.h`); config/timer macros moved to `pfcwdorch.h`; SW-only macros correctly live in `pfcwdsworch.cpp`. No duplicate/missing definitions.
- **Templates & linkage** — all four explicit `PfcWdSwOrch<...>` instantiations are preserved at the tail of `pfcwdsworch.cpp`. The base `.cpp` include set is byte-identical to the original, so transitive types (`SwitchOrch`, handlers, flex-counter manager) resolve as before. No dangling `PfcWdOrch` / `PfcDlrPacketAction` references anywhere.
- **Tests** — `mock_orchagent_main.h` adds `#include "pfcwdsworch.h"` inside both the `#define private public` and `#define protected public` blocks, so `portsorch_ut` / `flexcounter_ut` access to `m_pfcwd_ports`, `m_entryMap`, and `m_platform` still works; `gPfcwdOrch` remains typed as the still-templated `PfcWdSwOrch<...>*`. Makefiles updated.
- **Nice cleanup**: moving `packet_action_map` out of the header into `pfcwdorch.cpp` removes an unused internal-linkage copy from every TU that included the header.

### Minor nits (non-blocking)
1. `PFC_WD_POLL_TIMEOUT` (`pfcwdsworch.cpp`) is defined but unused. It was dead in the original too, but since one of the commits is "remove unused locals and dead code," it would be consistent to drop it here as well.
2. Tiny log-noise change: if `getPortPfcWatchdogStatus` fails inside the new `getLosslessTcsForPort`, `registerInWdDb` now additionally logs `NOTICE "No lossless TC found..."` after the helper's `ERROR`. The original returned immediately after the error without that NOTICE. Purely cosmetic.
3. `getLosslessTcsForPort` isn't reused in the two similar `pfcMask` loops in `enableBigRedSwitchMode` — acceptable since their logic differs slightly, just noting it for a potential follow-up.

vmittal-msft:
> @abhishek-nexthop @pinky-nexthop
> 
> Since this is important feature as well as touching critical code base: please check following for all HW PFCWD PRs -
> 
> 1. Make sure all sonic-mgmt/snappi tests pass after these changes related to PFCWD
> 2. Make sure these tests pass for XGS/DNX both type of SKUs
> 3. Make sure above tests work fine in both SW and HW based model.
> 4. Can you please help me remember if there is any knob available to switch backto SW mode in case of any issues ?

@abhishek-nexthop @pinky-nexthop  anything we can handle here. Also, please add tests for coverage shortage

pinky-nexthop:
> > @abhishek-nexthop @pinky-nexthop
> > Since this is important feature as well as touching critical code base: please check following for all HW PFCWD PRs -
> > 
> > 1. Make sure all sonic-mgmt/snappi tests pass after these changes related to PFCWD
> > 2. Make sure these tests pass for XGS/DNX both type of SKUs
> > 3. Make sure above tests work fine in both SW and HW based model.
> > 4. Can you please help me remember if there is any knob available to switch backto SW mode in case of any issues ?
> 
> @abhishek-nexthop @pinky-nexthop anything we can handle here. Also, please add tests for coverage shortage

@vmittal-msft We have requested broadcom for a knob to disable h/w recovery mode. Currently s/w can be selected using 
`sai_pfc_dlr_init_capability=0`. 
we have another PR https://github.com/sonic-net/sonic-swss/pull/4760 that can fix the missing coverage required here, so waiting for that to merge and then rebase this.

vmittal-msft:
> > > @abhishek-nexthop @pinky-nexthop
> > > Since this is important feature as well as touching critical code base: please check following for all HW PFCWD PRs -
> > > 
> > > 1. Make sure all sonic-mgmt/snappi tests pass after these changes related to PFCWD
> > > 2. Make sure these tests pass for XGS/DNX both type of SKUs
> > > 3. Make sure above tests work fine in both SW and HW based model.
> > > 4. Can you please help me remember if there is any knob available to switch backto SW mode in case of any issues ?
> > 
> > 
> > @abhishek-nexthop @pinky-nexthop anything we can handle here. Also, please add tests for coverage shortage
> 
> @vmittal-msft We have requested broadcom for a knob to disable h/w recovery mode. Currently s/w can be selected using `sai_pfc_dlr_init_capability=0`. we have another PR #4760 that can fix the missing coverage required here, so waiting for that to merge and then rebase this.

Could you guys do any tests and share results here ?

abhishek-nexthop:
> > > > @abhishek-nexthop @pinky-nexthop
> > > > Since this is important feature as well as touching critical code base: please check following for all HW PFCWD PRs -
> > > > 
> > > > 1. Make sure all sonic-mgmt/snappi tests pass after these changes related to PFCWD
> > > > 2. Make sure these tests pass for XGS/DNX both type of SKUs
> > > > 3. Make sure above tests work fine in both SW and HW based model.
> > > > 4. Can you please help me remember if there is any knob available to switch backto SW mode in case of any issues ?
> > > 
> > > 
> > > @abhishek-nexthop @pinky-nexthop anything we can handle here. Also, please add tests for coverage shortage
> > 
> > 
> > @vmittal-msft We have requested broadcom for a knob to disable h/w recovery mode. Currently s/w can be selected using `sai_pfc_dlr_init_capability=0`. we have another PR #4760 that can fix the missing coverage required here, so waiting for that to merge and then rebase this.
> 
> Could you guys do any tests and share results here ?

@vmittal-msft  This PR is a pure code move. PfcWdOrch is renamed to PfcWdBaseOrch and the software implementation is split into pfcwdsworch.{h,cpp} so the base can be shared with a hardware implementation. No functional change to software PFC watchdog behavior.
Refactored s/w based recovery code is currently running on both Q3d and TH6 which passes all pfcwd test suite.

vmittal-msft:
> > > > > @abhishek-nexthop @pinky-nexthop
> > > > > Since this is important feature as well as touching critical code base: please check following for all HW PFCWD PRs -
> > > > > 
> > > > > 1. Make sure all sonic-mgmt/snappi tests pass after these changes related to PFCWD
> > > > > 2. Make sure these tests pass for XGS/DNX both type of SKUs
> > > > > 3. Make sure above tests work fine in both SW and HW based model.
> > > > > 4. Can you please help me remember if there is any knob available to switch backto SW mode in case of any issues ?
> > > > 
> > > > 
> > > > @abhishek-nexthop @pinky-nexthop anything we can handle here. Also, please add tests for coverage shortage
> > > 
> > > 
> > > @vmittal-msft We have requested broadcom for a knob to disable h/w recovery mode. Currently s/w can be selected using `sai_pfc_dlr_init_capability=0`. we have another PR #4760 that can fix the missing coverage required here, so waiting for that to merge and then rebase this.
> > 
> > 
> > Could you guys do any tests and share results here ?
> 
> @vmittal-msft This PR is a pure code move. PfcWdOrch is renamed to PfcWdBaseOrch and the software implementation is split into pfcwdsworch.{h,cpp} so the base can be shared with a hardware implementation. No functional change to software PFC watchdog behavior. Refactored s/w based recovery code is currently running on Q3d which passes all pfcwd test suite.

@abhishek-nexthop  @pinky-nexthop  Thanks for confirming for DNX. Is this the case for XGS based platforms as well ?

abhishek-nexthop:
> > > > > > @abhishek-nexthop @pinky-nexthop
> > > > > > Since this is important feature as well as touching critical code base: please check following for all HW PFCWD PRs -
> > > > > > 
> > > > > > 1. Make sure all sonic-mgmt/snappi tests pass after these changes related to PFCWD
> > > > > > 2. Make sure these tests pass for XGS/DNX both type of SKUs
> > > > > > 3. Make sure above tests work fine in both SW and HW based model.
> > > > > > 4. Can you please help me remember if there is any knob available to switch backto SW mode in case of any issues ?
> > > > > 
> > > > > 
> > > > > @abhishek-nexthop @pinky-nexthop anything we can handle here. Also, please add tests for coverage shortage
> > > > 
> > > > 
> > > > @vmittal-msft We have requested broadcom for a knob to disable h/w recovery mode. Currently s/w can be selected using `sai_pfc_dlr_init_capability=0`. we have another PR #4760 that can fix the missing coverage required here, so waiting for that to merge and then rebase this.
> > > 
> > > 
> > > Could you guys do any tests and share results here ?
> > 
> > 
> > @vmittal-msft This PR is a pure code move. PfcWdOrch is renamed to PfcWdBaseOrch and the software implementation is split into pfcwdsworch.{h,cpp} so the base can be shared with a hardware implementation. No functional change to software PFC watchdog behavior. Refactored s/w based recovery code is currently running on Q3d which passes all pfcwd test suite.
> 
> @abhishek-nexthop @pinky-nexthop Thanks for confirming for DNX. Is this the case for XGS based platforms as well ?

@vmittal-msft Refactored s/w based is also running on TH6 which passes all pfcwd test suite, so we are good in that as well.

vmittal-msft (dismissed):

vmittal-msft:
@abhishek-nexthop approved. please attach test results from XGS as well as DNX in SW mode.

abhishek-nexthop:
> @abhishek-nexthop approved. please attach test results from XGS as well as DNX in SW mode.

Result for TH6(XGS):
```
-- tests/pfcwd/test_pfcwd_all_port_storm.py  ->  PASS  (total=2 pass=2 fail=0 err=0 skip=0)
     PASS   test_all_port_storm_restore[IPv4-dut]
     PASS   test_all_port_storm_restore[IPv6-dut]

-- tests/pfcwd/test_pfcwd_cli.py  ->  PASS  (total=2 pass=2 fail=0 err=0 skip=0)
     PASS   test_pfcwd_show_stat[IPv4-dut]
     PASS   test_pfcwd_show_stat[IPv6-dut]

-- tests/pfcwd/test_pfcwd_function.py  ->  PASS  (total=14 pass=8 fail=0 err=0 skip=6)
     PASS   test_pfcwd_actions[IPv4-dut]
     PASS   test_pfcwd_actions[IPv6-dut]
     PASS   test_pfcwd_mmu_change[IPv4-dut]
     PASS   test_pfcwd_mmu_change[IPv6-dut]
     PASS   test_pfcwd_port_toggle[IPv4-dut]
     PASS   test_pfcwd_port_toggle[IPv6-dut]
     PASS   test_pfcwd_storm_during_traffic[IPv4-dut]
     PASS   test_pfcwd_storm_during_traffic[IPv6-dut]
     SKIP   test_pfcwd_config_fail_with_egress_acl[IPv4-dut]   <-- PFCWD egress-ACL create did not fail (status='configured') - no egress-ACL/PFCWD bank contention or HW/DLR recovery on this platform
     SKIP   test_pfcwd_config_fail_with_egress_acl[IPv6-dut]   <-- PFCWD egress-ACL create did not fail (status='configured') - no egress-ACL/PFCWD bank contention or HW/DLR recovery on this platform
     SKIP   test_pfcwd_multi_port[IPv4-dut]   <-- Pfcwd multi port test needs at least 2 ports
     SKIP   test_pfcwd_multi_port[IPv6-dut]   <-- Pfcwd multi port test needs at least 2 ports
     SKIP   test_pfcwd_no_traffic[IPv4-dut]   <-- This test is applicable only for cisco-8000 / Pfcwd tests skipped on M* testbed.
     SKIP   test_pfcwd_no_traffic[IPv6-dut]   <-- This test is applicable only for cisco-8000 / Pfcwd tests skipped on M* testbed.

-- tests/pfcwd/test_pfcwd_timer_accuracy.py  ->  PASS  (total=2 pass=2 fail=0 err=0 skip=0)
     PASS   test_pfcwd_timer_accuracy[IPv4-dut]
     PASS   test_pfcwd_timer_accuracy[IPv6-dut]
```

Result for Q3D(DNX):
```
-- tests/pfcwd/test_pfcwd_all_port_storm.py  ->  PASS  (total=2 pass=2 fail=0 err=0 skip=0)
     PASS   test_all_port_storm_restore[IPv4-dut]
     PASS   test_all_port_storm_restore[IPv6-dut]

-- tests/pfcwd/test_pfcwd_cli.py  ->  FAIL  (total=2 pass=0 fail=2 err=0 skip=0)
     FAIL   test_pfcwd_show_stat[IPv4-dut]   <-- Failed: PFC storm Tx drop count not correct
     FAIL   test_pfcwd_show_stat[IPv6-dut]   <-- Failed: PFC storm Tx drop count not correct

-- tests/pfcwd/test_pfcwd_function.py  ->  FAIL  (total=14 pass=6 fail=2 err=0 skip=6)
     FAIL   test_pfcwd_storm_during_traffic[IPv4-dut]   <-- Failed: tx_drop_count not incrementing after storm on port Ethernet52 queue 4
     FAIL   test_pfcwd_storm_during_traffic[IPv6-dut]   <-- Failed: tx_drop_count not incrementing after storm on port Ethernet52 queue 4
     PASS   test_pfcwd_actions[IPv4-dut]
     PASS   test_pfcwd_actions[IPv6-dut]
     PASS   test_pfcwd_mmu_change[IPv4-dut]
     PASS   test_pfcwd_mmu_change[IPv6-dut]
     PASS   test_pfcwd_port_toggle[IPv4-dut]
     PASS   test_pfcwd_port_toggle[IPv6-dut]
     SKIP   test_pfcwd_config_fail_with_egress_acl[IPv4-dut]   <-- PFCWD egress-ACL create did not fail (status='configured') - no egress-ACL/PFCWD bank contention or HW/DLR recovery on this platform
     SKIP   test_pfcwd_config_fail_with_egress_acl[IPv6-dut]   <-- PFCWD egress-ACL create did not fail (status='configured') - no egress-ACL/PFCWD bank contention or HW/DLR recovery on this platform
     SKIP   test_pfcwd_multi_port[IPv4-dut]   <-- Pfcwd multi port test needs at least 2 ports
     SKIP   test_pfcwd_multi_port[IPv6-dut]   <-- Pfcwd multi port test needs at least 2 ports
     SKIP   test_pfcwd_no_traffic[IPv4-dut]   <-- This test is applicable only for cisco-8000 / Pfcwd tests skipped on M* testbed.
     SKIP   test_pfcwd_no_traffic[IPv6-dut]   <-- This test is applicable only for cisco-8000 / Pfcwd tests skipped on M* testbed.


-- tests/pfcwd/test_pfcwd_timer_accuracy.py  ->  PASS  (total=2 pass=2 fail=0 err=0 skip=0)
     PASS   test_pfcwd_timer_accuracy[IPv4-dut]
     PASS   test_pfcwd_timer_accuracy[IPv6-dut]
```

For Q3d all counters related tcs are failing due to regression(https://github.com/sonic-net/sonic-buildimage/issues/28470) but functionality is working.

vmittal-msft:
@saksarav-nokia @ysmanman @alpeshspatel @bingwang-ms @abdosi, HW PFCWD changes will impact all vendors, please help review/check, it's not breaking anything on respective HWSKU. This is the first PR in the list, there will be more coming. Will tag for those as well.

vmittal-msft:
@yxieca @rlhui  please help merge

kperumalbfn (commented):
@stephenxs Could you approve this refactoring. This is the initial changes to support hardware based PFCWD.

kperumalbfn (dismissed):

stephenxs:
> @stephenxs Could you approve this refactoring. This is the initial changes to support hardware based PFCWD.

ack. will provide feedback by the end of week

orchagent/pfcwdorch.h
```diff
@@ -13,9 +13,24 @@ extern "C" {
...
 #define PFC_WD_FLEX_COUNTER_GROUP       "PFC_WD"
+#define PFC_WD_GLOBAL                   "GLOBAL"
+
+#define PFC_WD_ACTION                   "action"
+#define PFC_WD_DETECTION_TIME           "detection_time"
+#define PFC_WD_RESTORATION_TIME         "restoration_time"
+#define PFC_STAT_HISTORY                "pfc_stat_history"
 
-const string pfc_wd_flex_counter_group = PFC_WD_FLEX_COUNTER_GROUP;
+// Timer limits in milliseconds
```
stephenxs: the range of PFC watchdog detection and restoration time can be vendor-specific.
in case a vendor has a larger or different range, how should they be overridden in the PFC WD hardware orch? macros can not be overridden in derived class, right?
maybe change them to class variables

abhishek-nexthop: Yes, for the hardware-specific configuration, we retrieve it from SAI and store it in a class variable. The macro already exists for pfcwdsw, so we left it unchanged.

vmittal-msft:
@rlhui  @yxieca please help merge

abdosi:
## Code Review

Overall, the refactoring is structurally sound — behavior is faithfully preserved across the base/derived class split. Two concrete issues found:

---

### Issue 1 — `pfc_wd_flex_counter_group` removed from public header (Medium)

**`orchagent/pfcwdorch.h`**

The declaration:
```cpp
const string pfc_wd_flex_counter_group = PFC_WD_FLEX_COUNTER_GROUP;
```
was removed from `pfcwdorch.h` with no replacement. Any translation unit that uses this symbol by name (rather than the raw macro `PFC_WD_FLEX_COUNTER_GROUP`) will fail to compile.

**Fix:** grep the repo for `pfc_wd_flex_counter_group` — if any files reference it, keep the declaration or migrate them to the macro before removing it.

---

### Issue 2 — Misleading log message on SAI failure in `registerInWdDb` (Medium)

**`orchagent/pfcwdsworch.cpp`**

The old code had two distinct failure paths in `registerInWdDb`:
- `getPortPfcWatchdogStatus` SAI call fails → `SWSS_LOG_ERROR("Failed to get PFC mask on port %s")`
- No PFC-enabled TCs → `SWSS_LOG_NOTICE("No lossless TC found on port %s")`

After refactoring, both paths return `false` from `getLosslessTcsForPort`, but `registerInWdDb` unconditionally logs `SWSS_LOG_NOTICE("No lossless TC found on port %s")` for both. When a real SAI failure occurs, the log shows an ERROR from `getLosslessTcsForPort` **followed by** a misleading NOTICE saying "No lossless TC found" — confusing in production triage.

**Fix:** Propagate the failure reason from `getLosslessTcsForPort` so `registerInWdDb` can distinguish the two cases and log appropriately.

---

### Everything else looks correct ✅
- Virtual dispatch on `startWdOnPort`/`stopWdOnPort` — clean
- `packet_action_map` moved from static-in-header to static-in-.cpp — correct
- Warm-reboot `bake()` path — preserved
- BRS mode enable/disable loops — preserved
- Makefile additions for `pfcwdsworch.cpp` — correct
- Test additions in `test_pfcwd.py` — adequate for the refactor scope

abdosi:
can you check on above comments

orchagent/pfcwdorch.h
stephenxs: I suggest moving them to variables so that they can be overridden by derived class. otherwise we have a lot of code to change for vendors with different range:

## Summary

If a vendor supports PFC watchdog detection or restoration timer ranges that differ from the legacy SONiC ranges, only timer-range validation should need to change.

In the current design, however, the fixed ranges are embedded inside `PfcWdBaseOrch::createEntry()`. A derived implementation therefore has to override almost the entire per-port `createEntry()` transaction merely to use different timer limits.

The legacy software ranges are:

| Timer | Minimum | Maximum |
|---|---:|---:|
| Detection | 100 ms | 5,000 ms |
| Restoration | 100 ms | 60,000 ms |

For example, a hardware implementation may obtain different ranges from these SAI switch attributes:

- `SAI_SWITCH_ATTR_PFC_DLD_INTERVAL_RANGE`
- `SAI_SWITCH_ATTR_PFC_DLR_INTERVAL_RANGE`

## Why `createEntry()` Has To Be Rewritten Today

The common implementation validates the timers directly against compile-time constants:

```cpp
detectionTime = to_uint<uint32_t>(
        value,
        PFC_WD_DETECTION_TIME_MIN,
        PFC_WD_DETECTION_TIME_MAX);

restorationTime = to_uint<uint32_t>(
        value,
        PFC_WD_RESTORATION_TIME_MIN,
        PFC_WD_RESTORATION_TIME_MAX);
```

There is no virtual range-provider or timer-validation hook. Consequently, an implementation with different ranges must override `createEntry()`.

Because field parsing, common validation, and watchdog activation are all contained in the same method, overriding the timer checks also duplicates all the unrelated common behavior described below.

## Common Logic Duplicated by the Vendor Implementation

### 1. Initialize the Parsed Configuration

The replacement method must reproduce the common defaults:

```cpp
uint32_t detectionTime = 0;
uint32_t restorationTime = 0;
PfcWdAction action = PfcWdAction::PFC_WD_ACTION_DROP;
string pfcStatHistory = "disable";
```

This establishes:

- Detection time as a required field, represented initially by zero.
- Restoration time as optional.
- `DROP` as the default watchdog action.
- Disabled PFC statistics history as the default.

### 2. Resolve and Validate the Port

The replacement method must:

- Resolve the configuration key using `gPortsOrch->getPort()`.
- Reject an unknown interface.
- Verify that the resolved object is a physical port.
- Reject VLANs, LAGs and other non-physical interfaces.

This logic is independent of the timer ranges.

### 3. Iterate and Recognize Configuration Fields

The replacement method must iterate through every `FieldValueTuple` and recognize:

- `detection_time`
- `restoration_time`
- `action`
- `pfc_stat_history`

It must also reject unknown fields. This parsing structure is common to software and hardware watchdog implementations.

### 4. Parse Numeric Values Safely

The common implementation uses `to_uint<uint32_t>()`, which combines:

- String-to-integer conversion.
- Unsigned 32-bit type validation.
- Minimum-range validation.
- Maximum-range validation.

PR #4412 instead uses:

```cpp
static_cast<uint32_t>(stoul(value))
```

This is not equivalent. On a 64-bit system, `stoul()` can accept a value greater than `UINT32_MAX`; the subsequent cast to `uint32_t` can wrap the value before the vendor-range check runs. A sufficiently large invalid input could therefore become a smaller, apparently valid timer.

This difference is an example of generic behavior drifting because common parsing was copied into a derived implementation.

### 5. Deserialize and Validate the Action

The replacement method must:

- Call `deserializeAction()`.
- Detect `PFC_WD_ACTION_UNKNOWN`.
- Reject an invalid action.

Basic action parsing is common. Platform-specific action restrictions should be handled separately through an action-policy hook.

### 6. Parse and Validate PFC Statistics History

The replacement method must:

- Read `pfc_stat_history`.
- Preserve `disable` as its default.
- Accept only `enable` or `disable`.
- Reject every other value.

This has no relationship to the supported timer ranges.

### 7. Handle Parsing Failures

The replacement method must catch numeric conversion and other parsing exceptions, log the failed field, and return `task_invalid_entry`.

Duplicated implementations may differ in:

- Which exception types are caught.
- The reported error text.
- Whether overflow is detected before or after narrowing.
- Whether an invalid entry is rejected consistently.

### 8. Enforce Required Fields

After parsing, the replacement method must verify that `detection_time` was supplied:

```cpp
if (detectionTime == 0)
{
}
```

This is common schema validation, not vendor range validation.

### 9. Start the Watchdog

The replacement method must call the already-virtual implementation hook:

```cpp
startWdOnPort(port, detectionTime, restorationTime, action, pfcStatHistory)
```

It must then:

- Return `task_need_retry` if activation fails.
- Log successful activation.
- Return `task_success` on success.

The hardware or software programming belongs in `startWdOnPort()`; the surrounding transaction handling can remain common.

## Logic That Does Not Need To Be Rewritten for a Different Range

### `PfcWdBaseOrch::doTask(Consumer&)`

The consumer task logic remains common. It:

1. Reads pending database operations.
2. Dispatches `SET` to virtual `createEntry()`.
3. Dispatches `DEL` to virtual `deleteEntry()`.
4. Erases successfully processed or invalid entries.
5. Retains entries that should be retried.

Different ranges do not require a different consumer loop. Virtual dispatch already selects the appropriate implementation.

### Entry Deletion

A different timer range does not by itself require overriding `deleteEntry()`.

Hardware watchdog implementations may need different SAI teardown behavior, but that is a hardware programming difference, not a timer-range difference.

### Hardware Event and Counter Handling

The following are also independent of range validation:

- SAI deadlock notifications.
- Queue storm-state tracking.
- Hardware counter collection.
- Flex counter registration.
- Detection and restoration event processing.

## Genuinely Hardware-Specific Logic in PR #4412

PR #4412 contains additional behavior that reasonably belongs to the hardware implementation:

- Rejecting unsupported `GLOBAL` configuration.
- Enforcing a switch-wide PFC DLR packet action.
- Preventing reconfiguration while a port has a stormed queue.
- Stopping an existing hardware watchdog before applying a replacement configuration when in-place updates are unsupported.
- Tracking ports configured for hardware watchdog operation.
- Programming hardware through SAI.

These differences should not be confused with the duplicated common parsing caused by vendor-specific timer ranges.

## Recommended Design

Keep the per-port `createEntry()` flow common and expose a timer-range provider:

```cpp
struct PfcWdTimerRange
{
    uint32_t detectionMin;
    uint32_t detectionMax;
    uint32_t restorationMin;
    uint32_t restorationMax;
};

virtual std::optional<PfcWdTimerRange>
getTimerRange(const Port& port) const;
```

The common parser can then use the selected ranges without changing its conversion behavior:

```cpp
const auto range = getTimerRange(port);
if (!range)
{
    // Range information is temporarily unavailable; fail or retry explicitly.
}

detectionTime = to_uint<uint32_t>(
        value,
        range->detectionMin,
        range->detectionMax);

restorationTime = to_uint<uint32_t>(
        value,
        range->restorationMin,
        range->restorationMax);
```

The implementations would then be small:

- Software watchdog returns the legacy SONiC ranges.
- NVIDIA hardware watchdog returns the ranges queried from SAI.
- Another vendor can return its own SAI or platform ranges.

If a vendor needs additional action or reconfiguration policies, those should be separate hooks rather than reasons to replace the parser:

```cpp
virtual task_process_status validateAction(
        const Port& port,
        PfcWdAction action);

virtual task_process_status preparePortUpdate(
        const Port& port);
```

## Recommended Responsibility Split

| Responsibility | Common base | Vendor/HW implementation |
|---|:---:|:---:|
| Resolve port | Yes | No |
| Require physical port | Yes | No |
| Parse known fields | Yes | No |
| Reject unknown fields | Yes | No |
| Safe `uint32_t` conversion | Yes | No |
| Require detection time | Yes | No |
| Validate statistics-history value | Yes | No |
| Supply timer ranges | No | Yes |
| Enforce platform action policy | Framework | Policy hook |
| Prepare for non-in-place update | Framework | Policy hook |
| Program SW or HW watchdog | No | `startWdOnPort()` |
| Handle database retry status | Yes | No |

## Conclusion

A vendor with a different supported range should need to provide only four values—or one timer-range object.

Under the current design, that vendor must replace nearly the entire `createEntry()` method, duplicating port validation, field parsing, action parsing, statistics-history validation, exception handling, required-field checks, activation and task-status handling.

This duplication is not functionally necessary. It increases review and maintenance cost and can introduce generic behavioral differences, as demonstrated by the change from `to_uint<uint32_t>()` to `static_cast<uint32_t>(stoul())` in the hardware implementation.

arawat-nexthop: createEntry() no longer validates against the compile-time constants. The base
now exposes a range provider
` virtual bool getTimerRange(PfcWdTimerRange& range) const;`
The default returns the legacy software limits, so SW watchdog behaviour is
unchanged.
On the action-policy hook: agreed, but it moves live SW code paths and we are trying not to make many changes in sw code path in this PR. This will be taken in #4412 where the file is being changed and gets HW tested

arawat-nexthop:
> ## Code Review
> Overall, the refactoring is structurally sound — behavior is faithfully preserved across the base/derived class split. Two concrete issues found:
> 
> ### Issue 1 — `pfc_wd_flex_counter_group` removed from public header (Medium)
> **`orchagent/pfcwdorch.h`**
> 
> The declaration:
> 
> ```c++
> const string pfc_wd_flex_counter_group = PFC_WD_FLEX_COUNTER_GROUP;
> ```
> 
> was removed from `pfcwdorch.h` with no replacement. Any translation unit that uses this symbol by name (rather than the raw macro `PFC_WD_FLEX_COUNTER_GROUP`) will fail to compile.
> 
> **Fix:** grep the repo for `pfc_wd_flex_counter_group` — if any files reference it, keep the declaration or migrate them to the macro before removing it.
> 
> ### Issue 2 — Misleading log message on SAI failure in `registerInWdDb` (Medium)
> **`orchagent/pfcwdsworch.cpp`**
> 
> The old code had two distinct failure paths in `registerInWdDb`:
> 
> * `getPortPfcWatchdogStatus` SAI call fails → `SWSS_LOG_ERROR("Failed to get PFC mask on port %s")`
> * No PFC-enabled TCs → `SWSS_LOG_NOTICE("No lossless TC found on port %s")`
> 
> After refactoring, both paths return `false` from `getLosslessTcsForPort`, but `registerInWdDb` unconditionally logs `SWSS_LOG_NOTICE("No lossless TC found on port %s")` for both. When a real SAI failure occurs, the log shows an ERROR from `getLosslessTcsForPort` **followed by** a misleading NOTICE saying "No lossless TC found" — confusing in production triage.
> 
> **Fix:** Propagate the failure reason from `getLosslessTcsForPort` so `registerInWdDb` can distinguish the two cases and log appropriately.
> 
> ### Everything else looks correct ✅
> * Virtual dispatch on `startWdOnPort`/`stopWdOnPort` — clean
> * `packet_action_map` moved from static-in-header to static-in-.cpp — correct
> * Warm-reboot `bake()` path — preserved
> * BRS mode enable/disable loops — preserved
> * Makefile additions for `pfcwdsworch.cpp` — correct
> * Test additions in `test_pfcwd.py` — adequate for the refactor scope

Issue 1: All call sites use the PFC_WD_FLEX_COUNTER_GROUP macro directly, so this drops a
dead header definition 
Issue 2: Thanks for the insight. Fixed in d81e3073

arawat-nexthop:
@stephenxs @vmittal-msft @abdosi can you please approve if there are no pending items

vmittal-msft (approved):

orchagent/pfcwdorch.h
rlhui: @stephenxs , could you check above response and if approve this PR? (btw, the review comments from AI were a bit too long, good to shorten or key comment only... thanks)

stephenxs: ok. then this PR LGTM.

stephenxs (approved):

pinky-nexthop:
@rlhui  @yxieca kindly merge this change. Thank you!

abdosi (approved):

PR #4363, opened 2026-03-18
Refactor IPv6 address handling to use temporary objects

kperumalbfn:
Store temporary objects for IPv6 address and mask to prevent dangling pointers.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**


prsunny:
@kperumalbfn , suggest create branch to forked ones. For this one, please delete the branch after use.

PR #4368, opened 2026-03-19
[P4Orch] p4orch combined PRs.

bibhuprasad-hcl:
**What I did**
p4orch combined commits

**Why I did it**
combined all p4orch commits to a single PR.

**How I verified it**
UT Cases

**Details if related**


tests/test_warm_reboot.py
```diff
@@ -2492,6 +2492,9 @@ def test_FpmsyncdWarmRestart(self, dvs):
         # Start fpmsyncd
         dvs.start_fpmsyncd()
 
+        # Disable state verification during warmboot as a part of cleanup.
+        self.configure_state_verification(dvs, "false")
```
mint570: Do we need this change?

bibhuprasad-hcl: > Do we need this change?

This change is removed.

PR #4370, opened 2026-03-19
TAM Mirror-on-drop feature implementation

prashanth-nexthop:
**What I did**
This is an implementation of TAM MOD feature.  

**Why I did it**
Customer requirement

**How I verified it**
Verified with both unit tests and sonic-mgmt tests

**Details if related**

### DIFF
--- a/orchagent/tamorch.cpp
+++ b/orchagent/tamorch.cpp
@@ -1,0 +1,3375 @@
+#include "tamorch.h"
+
+#include "saiextensions.h"
+#include "logger.h"
+#include "schema.h"
+#include "converter.h"
+#include "ipprefix.h"
+#include "swssnet.h"
+#include "directory.h"
+#include "saihelper.h"
+#include "aclorch.h"
+#include "observer.h"
+#include "portsorch.h"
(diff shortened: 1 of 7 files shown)


orchagent/tamorch.cpp
```diff
@@ -0,0 +1,3375 @@
...
+}
+
+bool TamOrch::tamTransportDelete(sai_object_id_t transportObjId)
+{
+    auto status = sai_tam_api->remove_tam_transport(transportObjId);
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("TAM ERROR: Error deleting TAM transport object");
+        auto taskStatus = handleSaiCreateStatus(SAI_API_TAM, status);
```
GarfieldZzj: When deleting a TAM Transport, the function `handleSaiCreateStatus` (used for handling create operation status) is invoked, but I think `handleSaiRemoveStatus` should be called instead.

prashanth-nexthop: Thanks for catching this.  I fixed it.

PR #4371, opened 2026-03-20
SRv6: the decap_dscp_mode configuration is now optional

Yakiv-Huryk:
**What I did**
The SRv6 MySID decap_dscp_mode configuration is now optional. When not configured, the tunnel-related configuration is not created.

**Why I did it**
To add an ability to have the SRv6 configuration without the tunnel.

**How I verified it**
Manual tests and new unit tests.

**Details if related**

### DIFF
--- a/cfgmgr/nbrmgr.cpp
+++ b/cfgmgr/nbrmgr.cpp
@@ -65,7 +65,10 @@ NbrMgr::NbrMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, con
                               TableConsumable::DEFAULT_POP_BATCH_SIZE, default_orch_pri);
     auto consumer = new Consumer(consumerStateTable, this, APP_NEIGH_RESOLVE_TABLE_NAME);
     Orch::addExecutor(consumer);
-          
+
+    /* Reconcile any pending entries in NEIGH_RESOLVE_TABLE from before restart */
+    reconcileNeighResolveTable(appDb);
+
     string swtype;
     Table cfgDeviceMetaDataTable(cfgDb, CFG_DEVICE_METADATA_TABLE_NAME);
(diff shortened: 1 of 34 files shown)


kperumalbfn:
@abdosi @BYGX-wcr please review

liat-grozovik:
@r12f please share you view so we can have this part of 202511 with prio.

orchagent/srv6orch.cpp
```diff
@@ -1425,7 +1424,8 @@ bool Srv6Orch::mySidTunnelRequired(const string& my_sid_addr, const sai_my_sid_e
 
     auto locator_cfg = getMySidEntryLocatorCfg(sai_entry);
 
-    return getMySidEntryDscpMode(my_sid_addr, locator_cfg, dscp_mode);
+    auto status = getMySidEntryDscpMode(my_sid_addr, locator_cfg, dscp_mode);
+    return status && dscp_mode.has_value();
```
BYGX-wcr, line 1428: With this change, this function returns false now when the dscp mode is missing and that will make the caller's code not set the endpoint flavor. In that scenario, the SDK will get NONE as the endpoint flavor. This doesn't sound right. Maybe we should add an extra config db knob for end point flavor.

Yakiv-Huryk: This is intended. W/o the tunnel config, USD (and PSP_AND_USD) flavors are not applicable since we don't do the decap.

kperumalbfn:
@rick-arista FYI

abdosi:
@dgsudharsan : can we review this PR also : https://github.com/sonic-net/sonic-swss/pull/3797.

Falls in similar context

dgsudharsan:
@prsunny @r12f  Please let me know if it is good to be merged.

prsunny:
lgtm, @BYGX-wcr and @abdosi to signoff as well.

PR #4380, opened 2026-03-23
[countersyncd]: Make it possible to configure the poll intervals

### LINKED ISSUE #26246: Bug: countersyncd: Netlink receive buffer overflows with very large HFT counter set
### Is it platform specific

generic

### Importance or Severity

Critical

### Description of the bug

Under high HFT load with a very large number of counters (e.g. full queue counters, as in test_hft_full_queue_counters), countersyncd can hit ENOBUFS (“No buffer space available”) on the netlink socket. 

The kernel reports that the netlink socket’s receive buffer is full (“No buffer space available”), so the overflow is at the kernel → userspace boundary: data is arriving faster than the process is reading it.

So the kernel is telling us its netlink receive buffer is full and it had to drop data. The problem is before our processing: we are not draining the socket often enough.

The code waits a fixed interval (e.g. 10 ms) before each socket readiness check, then does one read. So we only try to drain the kernel buffer at most once per interval. If HFT samples are sent every 10 ms. With a very large number of counters, each 10 ms burst is large. If we only read once per 10 ms, the kernel can get the next burst before we’ve read the previous one, so its buffer fills and we see the “buffer full” error. Because comm stats show downstream is keeping up, the only remaining explanation is that we’re not draining the kernel buffer often enough for this data rate and burst size.

I have created a debug version with reduced interval of 5ms - which has caused this issue to disappear under the same load 

### Steps to Reproduce

1. Configure SONiC with HFT enabled and a profile that creates a very large number of counters (e.g. full queue counters, as covered by test_hft_full_queue_counters on a switch with 512 ports, and 8 QUEUES defined per port), with 10 ms sample interval.
2. Start countersyncd with:
    countersyncd -e --stats-interval 10 
3. Run under steady traffic so HFT samples are sent every 10 ms with the full counter set.
4. Observe logs for ENOBUFS.

### Actual Behavior and Expected Behavior

**Expected behavior**
No ENOBUFS when running with a very large number of HFT counters (e.g. full queue counters as in test_hft_full_queue_counters), as long as the poll interval is appropriate for the data rate.

**Actual behavior**
With a very large number of counters, logs repeatedly show:

`[WARN] Netlink receive buffer full (ENOBUFS). Consider increasing buffer size or processing messages faster. Error: Os { code: 105, kind: Uncategorized, message: "No buffer space available" }
`
Some HFT samples are dropped; stats can be incomplete or lag during bursts.
With a smaller number of counters, the same setup typically does not show ENOBUFS.

### Relevant log output

```shell
With 10 ms poll interval (reproduces the bug):

[crates/countersyncd/src/actor/data_netlink.rs:892] [WARN] Netlink receive buffer full (ENOBUFS). Consider increasing buffer size or processing messages faster. Error: Os { code: 105, kind: Uncategorized, message: "No buffer space available" }[2026-03-18 12:10:46.408] [crates/countersyncd/src/actor/data_netlink.rs:892] [WARN] Netlink receive buffer full (ENOBUFS). ...[2026-03-18 12:10:46.508] [crates/countersyncd/src/actor/data_netlink.rs:892] [WARN] Netlink receive buffer full (ENOBUFS). ...

Comm stats (data path itself is not backlogged; bottleneck is kernel→userspace):

[crates/countersyncd/src/utilities/mod.rs:140] [INFO] Comm stats [data_netlink.ipfix_records]: count=47754, avg_len=0.00, peak_len=0, min_len=0, last_len=0, rms_len=0.00, nonzero_count=0, capacity=1024, avg_util=0.00, peak_util=0.00
[crates/countersyncd/src/utilities/mod.rs:140] [INFO] Comm stats [ipfix.stats_reporter]: count=95507, avg_len=0.50, peak_len=3, min_len=0, last_len=1, rms_len=0.71, nonzero_count=47741, capacity=1024, avg_util=0.00, peak_util=0.00
[crates/countersyncd/src/utilities/mod.rs:140] [INFO] Comm stats [control_netlink.data_netlink_cmd]: count=10, avg_len=0.00, peak_len=0, min_len=0, last_len=0, rms_len=0.00, nonzero_count=0, capacity=10, avg_util=0.00, peak_util=0.00
```

### Output of `show version`, `show techsupport`

```shell

```

### Attach files (if any)

_No response_

DavidZagury:
**What I did**
- Made the socket readiness poll interval configurable via --socket-readiness-timeout-ms (default 5 ms; was fixed at 10 ms). The data_netlink actor uses this value when waiting before checking the netlink socket.
- Made the netlink socket receive buffer configurable via --netlink-rcvbuf (default 4 MB; 0 = OS default). Set SO_RCVBUF on the data netlink socket at connect/reconnect and log requested vs actual size.
- Made the comm stats log interval configurable via --comm-stats-interval (default 600 s) so queue-depth stats can be emitted more often when debugging.

**Why I did it**
To make sure we can empty the buffer before the next samples of counters is received.
And to fix https://github.com/sonic-net/sonic-buildimage/issues/26246

**How I verified it**
Configure HFT with a large amount of counters and ports and verified the amount of messages received per second.
Use the test test_hft_full_queue_counters to validate such secnario.

**Details if related**

### DIFF
--- a/crates/countersyncd/src/actor/data_netlink.rs
+++ b/crates/countersyncd/src/actor/data_netlink.rs
@@ -348,8 +348,33 @@ impl DataNetlinkActor {
                 bytes,
                 std::io::Error::last_os_error()
             );
+            return;
+        }
+
+        // Read back the actual value — Linux may cap it at net.core.rmem_max and doubles it internally.
+        let mut actual: libc::c_int = 0;
+        let mut len = std::mem::size_of::<libc::c_int>() as libc::socklen_t;
+        let ret = unsafe {
+            libc::getsockopt(
+                fd,
+                libc::SOL_SOCKET,
+                libc::SO_RCVBUF,
+                &mut actual as *mut _ as *mut libc::c_void,
+                &mut len,
+            )
+        };
+        if ret != 0 {
+            warn!(
+                "Failed to read back SO_RCVBUF: {:?}",
+                std::io::Error::last_os_error()
+            );
         } else {
-            info!("Set netlink socket SO_RCVBUF to {} bytes", bytes);
+            info!(
+                "Netlink SO_RCVBUF: requested={} bytes, actual={} bytes{}",
+                bytes,
+                actual,
+                if (actual as usize) < bytes { " (capped by net.core.rmem_max — consider raising it)" } else { "" }
+            );
         }
     }
 
--- a/crates/countersyncd/src/actor/data_netlink.rs
+++ b/crates/countersyncd/src/actor/data_netlink.rs
@@ -56,8 +56,8 @@ const DEBUG_LOG_INTERVAL: u32 = 3000; // 3000 * 10ms = 30 seconds
 /// WouldBlock debug logging interval (in iterations) - log WouldBlock every minute
 const WOULDBLOCK_LOG_INTERVAL: u32 = 6000; // 6000 * 10ms = 1 minute
 
-/// Socket readiness check timeout in milliseconds
-const SOCKET_READINESS_TIMEOUT_MS: u64 = 10;
+/// Default socket readiness poll interval (ms) when not configured via CLI.
+const DEFAULT_SOCKET_READINESS_TIMEOUT_MS: u64 = 5;
 
 /// Maximum size for buffering incomplete messages (1MB)
 const MAX_INCOMPLETE_MESSAGE_SIZE: usize = 1024 * 1024;
@@ -251,6 +251,8 @@ pub struct DataNetlinkActor {
     message_parser: NetlinkMessageParser,
     /// Netlink socket receive buffer size in bytes (0 = OS default). Reduces ENOBUFS when set.
     netlink_rcvbuf_bytes: usize,
+    /// Socket readiness poll interval in milliseconds. Shorter than HFT interval reduces ENOBUFS.
+    socket_readiness_timeout_ms: u64,
 }
 
 impl DataNetlinkActor {
@@ -262,6 +264,7 @@ impl DataNetlinkActor {
     /// * `group` - The multicast group name
     /// * `command_recipient` - Channel for receiving control commands
     /// * `netlink_rcvbuf_bytes` - Socket SO_RCVBUF size in bytes (0 = OS default). Larger values reduce ENOBUFS under high HFT load.
+    /// * `socket_readiness_timeout_ms` - Poll interval in ms for socket readiness. Shorter than HFT interval (e.g. 10 ms) reduces ENOBUFS.
     ///
     /// # Returns
     ///
@@ -271,6 +274,7 @@ impl DataNetlinkActor {
         group: &str,
         command_recipient: Receiver<NetlinkCommand>,
         netlink_rcvbuf_bytes: usize,
+        socket_readiness_timeout_ms: u64,
     ) -> Self {
         let nl_resolver = Self::create_nl_resolver();
         let mut actor = DataNetlinkActor {
@@ -283,6 +287,7 @@ impl DataNetlinkActor {
             command_recipient,
             message_parser: NetlinkMessageParser::new(),
             netlink_rcvbuf_bytes,
+            socket_readiness_timeout_ms,
         };
 
         // Use instance method for initial connection
@@ -880,7 +885,7 @@ impl DataNetlinkActor {
             }
 
             // Check socket readiness with configurable timeout to allow periodic checks
-            match Self::check_socket_readiness(SOCKET_READINESS_TIMEOUT_MS).await {
+            match Self::check_socket_readiness(actor.socket_readiness_timeout_ms).await {
                 Ok(data_ready) => {
                     // Only try to receive data if we have a socket and data is ready
                     if actor.socket.is_some() && data_ready {
@@ -934,7 +939,10 @@ impl DataNetlinkActor {
                                 // Handle specific errors
                                 if let Some(os_error) = e.raw_os_error() {
                                     if os_error == ENOBUFS {
-                                        warn!("Netlink receive buffer full (ENOBUFS). Consider increasing buffer size or processing messages faster. Error: {:?}", e);
+                                        warn!(
+                                            "Netlink receive buffer full (ENOBUFS). poll_interval_ms={}. Consider reducing --socket-readiness-timeout-ms or increasing buffer. Error: {:?}",
+                                            actor.socket_readiness_timeout_ms, e
+                                        );
                                         // Don't disconnect on ENOBUFS, just continue
                                         continue;
                                     }
@@ -975,7 +983,7 @@ impl DataNetlinkActor {
                 Err(e) => {
                     warn!("Poll error: {:?}", e);
                     // Wait a bit before retrying to avoid busy loop on persistent poll errors
-                    sleep(Duration::from_millis(SOCKET_READINESS_TIMEOUT_MS));
+                    sleep(Duration::from_millis(actor.socket_readiness_timeout_ms));
                 }
             }
         }
@@ -1168,7 +1176,7 @@ pub mod test {
         let (command_sender, command_receiver) = channel(1);
         let (buffer_sender, mut buffer_receiver) = channel(1);
 
-        let mut actor = DataNetlinkActor::new("family", "group", command_receiver, 0);
+        let mut actor = DataNetlinkActor::new("family", "group", command_receiver, 0, DEFAULT_SOCKET_READINESS_TIMEOUT_MS);
         actor.add_recipient(buffer_sender);
 
         let task = spawn(DataNetlinkActor::run(actor));
--- a/crates/countersyncd/src/actor/data_netlink.rs
+++ b/crates/countersyncd/src/actor/data_netlink.rs
@@ -322,7 +322,17 @@ impl DataNetlinkActor {
             return;
         }
         let fd = socket.as_raw_fd();
-        let v: libc::c_int = bytes.try_into().unwrap_or(libc::c_int::MAX);
+        let v: libc::c_int = match bytes.try_into() {
+            Ok(v) => v,
+            Err(_) => {
+                warn!(
+                    "netlink_rcvbuf {} exceeds c_int::MAX, clamping to {}",
+                    bytes,
+                    libc::c_int::MAX
+                );
+                libc::c_int::MAX
+            }
+        };
         let ret = unsafe {
             libc::setsockopt(
                 fd,
--- a/crates/countersyncd/src/actor/data_netlink.rs
+++ b/crates/countersyncd/src/actor/data_netlink.rs
@@ -1462,6 +1462,22 @@ pub mod test {
             assert!(!group.is_empty());
         }
     }
+
+    #[test]
+    fn test_netlink_rcvbuf_stored_on_construction() {
+        let (_, command_receiver) = channel(1);
+        let actor = DataNetlinkActor::new("family", "group", command_receiver, 4194304, 5);
+        assert_eq!(actor.netlink_rcvbuf_bytes, 4194304);
+    }
+
+    #[test]
+    fn test_log_interval_cadence_at_default_poll_ms() {
+        const DEFAULT_POLL_MS: u64 = 5;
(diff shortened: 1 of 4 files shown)


crates/countersyncd/src/actor/data_netlink.rs
```diff
@@ -303,6 +317,34 @@ impl DataNetlinkActor {
...
+                fd,
+                libc::SOL_SOCKET,
+                libc::SO_RCVBUF,
+                &v as *const _ as *const libc::c_void,
+                std::mem::size_of::<libc::c_int>() as libc::socklen_t,
+            )
+        };
+        if ret != 0 {
+            warn!(
+                "Failed to set netlink SO_RCVBUF to {}: {:?}",
```
50n1c-rnsft: **Verify actual SO_RCVBUF value after setsockopt**

Linux caps `SO_RCVBUF` at `net.core.rmem_max` (often ~208KB by default) and doubles it internally. With the default 4MB, the actual buffer may be far smaller with no indication.

Suggest calling `getsockopt(SO_RCVBUF)` after setting and logging the actual value:
```rust
let mut actual: libc::c_int = 0;
let mut len = std::mem::size_of::<libc::c_int>() as libc::socklen_t;
unsafe { libc::getsockopt(fd, libc::SOL_SOCKET, libc::SO_RCVBUF, &mut actual as *mut _ as *mut libc::c_void, &mut len) };
info!("Requested SO_RCVBUF={}, actual={}", bytes, actual);
```

Also consider trying `SO_RCVBUFFORCE` first (requires `CAP_NET_ADMIN`, typically available in containers), falling back to `SO_RCVBUF`.

crates/countersyncd/src/actor/data_netlink.rs
```diff
@@ -303,6 +317,34 @@ impl DataNetlinkActor {
...
     }
 
+    /// Sets SO_RCVBUF on the netlink socket to reduce ENOBUFS under high HFT load.
+    #[cfg(not(test))]
+    fn set_socket_rcvbuf(socket: &mut Socket, bytes: usize) {
+        if bytes == 0 {
+            return;
+        }
+        let fd = socket.as_raw_fd();
+        let v: libc::c_int = bytes.try_into().unwrap_or(libc::c_int::MAX);
```
50n1c-rnsft: **Silent truncation on `usize → c_int` overflow**

If a user passes a very large value, `unwrap_or(libc::c_int::MAX)` silently clamps to ~2GB. Consider logging a warning when truncation occurs:
```rust
let v: libc::c_int = match bytes.try_into() {
    Ok(v) => v,
    Err(_) => {
        warn!("netlink_rcvbuf {} exceeds c_int::MAX, clamping to {}", bytes, libc::c_int::MAX);
        libc::c_int::MAX
    }
};
```

crates/countersyncd/src/main.rs
```diff
@@ -171,6 +171,30 @@ struct Args {
...
+    )]
+    netlink_rcvbuf: usize,
+
+    /// Socket readiness poll interval in milliseconds. Shorter than HFT sample interval (e.g. 10 ms) reduces ENOBUFS.
+    #[arg(
+        long,
+        default_value = "5",
+        help = "Poll interval in ms for netlink socket readiness. Default 5"
+    )]
+    socket_readiness_timeout_ms: u64,
```
50n1c-rnsft, line 198: **Guard against `socket_readiness_timeout_ms = 0` (busy loop)**

A value of 0 would cause a tight busy loop in the poll loop. Consider adding a range constraint:
```rust
#[arg(
    long,
    default_value = "5",
    value_parser = clap::value_parser!(u64).range(1..),
    help = "..."
)]
```
Or a runtime check: `max(1, value)`.

crates/countersyncd/src/actor/data_netlink.rs
```diff
@@ -56,8 +56,8 @@ const DEBUG_LOG_INTERVAL: u32 = 3000; // 3000 * 10ms = 30 seconds
 /// WouldBlock debug logging interval (in iterations) - log WouldBlock every minute
 const WOULDBLOCK_LOG_INTERVAL: u32 = 6000; // 6000 * 10ms = 1 minute
```
50n1c-rnsft: **Stale comments**

`WOULDBLOCK_LOG_INTERVAL` and `DEBUG_LOG_INTERVAL` comments still reference 10ms (e.g. `6000 * 10ms = 1 minute`), but the default is now 5ms. Either update the comments to reflect the new default, or make these intervals dynamically computed from the actual `socket_readiness_timeout_ms`.

crates/countersyncd/src/utilities/mod.rs
```diff
@@ -30,8 +31,16 @@ pub fn format_hex_lines(buffer: &[u8]) -> String {
     lines.join("\n")
 }
 
-/// Log interval for communication stats.
-const COMM_LOG_INTERVAL: Duration = Duration::from_secs(600);
+/// Configurable log interval for communication stats (seconds).
+/// Set via set_comm_log_interval_secs() at startup (e.g. from CLI); default 600.
+static COMM_LOG_INTERVAL_SECS: AtomicU64 = AtomicU64::new(600);
+
```
50n1c-rnsft, line 37: **Global `AtomicU64` for comm stats interval**

Using a global mutable `AtomicU64` works for a single-process daemon, but could become problematic if multi-instance scenarios arise. Consider passing the interval through the actor structs instead — not blocking for this PR, just worth noting for future refactoring.

DavidZagury: No action needed for this

50n1c-rnsft (approved):
All review comments have been addressed. LGTM — nice improvements on the SO_RCVBUF validation, overflow warning, busy-loop guard, and updated interval comments with tests.

crates/countersyncd/src/actor/data_netlink.rs
```diff
@@ -47,17 +47,15 @@ const MAX_LOCAL_RECONNECT_ATTEMPTS: u32 = 3;
 /// Socket health check timeout - if no data received for this duration, socket is considered unhealthy
 const SOCKET_HEALTH_TIMEOUT_SECS: u64 = 60;
 
-/// Heartbeat logging interval (in iterations) - log every 5 minutes at 10ms per iteration
-const HEARTBEAT_LOG_INTERVAL: u32 = 30000; // 30000 * 10ms = 5 minutes
+/// Heartbeat logging interval (in iterations) - log every 5 minutes at 5ms per iteration
+const HEARTBEAT_LOG_INTERVAL: u32 = 60000; // 60000 * 5ms = 5 minutes
```
Pterosaur: **Bug: Log interval constants are hardcoded for the 5ms default poll interval**

These constants (`HEARTBEAT_LOG_INTERVAL`, `DEBUG_LOG_INTERVAL`, `WOULDBLOCK_LOG_INTERVAL`) assume `socket_readiness_timeout_ms = 5`. Since the poll interval is now configurable via `--socket-readiness-timeout-ms`, using a non-default value will silently skew all log cadences.

For example, with `--socket-readiness-timeout-ms 50`:
- Heartbeat: `60000 * 50ms = 50 minutes` (expected 5 minutes)
- Debug: `6000 * 50ms = 5 minutes` (expected 30 seconds)
- WouldBlock: `12000 * 50ms = 10 minutes` (expected 1 minute)

The test `test_log_interval_cadence_at_default_poll_ms` only validates the default, masking this issue.

Suggested fix: define target durations as constants and compute iteration counts at runtime:
```rust
const HEARTBEAT_TARGET_MS: u64 = 5 * 60 * 1000; // 5 minutes
const DEBUG_TARGET_MS: u64 = 30 * 1000;          // 30 seconds
const WOULDBLOCK_TARGET_MS: u64 = 60 * 1000;     // 1 minute

// in DataNetlinkActor or run():
let heartbeat_interval = (HEARTBEAT_TARGET_MS / socket_readiness_timeout_ms) as u32;
let debug_interval = (DEBUG_TARGET_MS / socket_readiness_timeout_ms) as u32;
let wouldblock_interval = (WOULDBLOCK_TARGET_MS / socket_readiness_timeout_ms) as u32;
```

crates/countersyncd/src/actor/data_netlink.rs
```diff
@@ -303,6 +315,69 @@ impl DataNetlinkActor {
...
+                    libc::c_int::MAX
+                );
+                libc::c_int::MAX
+            }
+        };
+        let ret = unsafe {
+            libc::setsockopt(
+                fd,
+                libc::SOL_SOCKET,
+                libc::SO_RCVBUF,
```
Pterosaur: **Suggestion: Try `SO_RCVBUFFORCE` before falling back to `SO_RCVBUF`**

Linux caps `SO_RCVBUF` at `net.core.rmem_max` (often ~208 KB by default). With the default `--netlink-rcvbuf 4194304`, the actual buffer may end up far smaller than requested.

`SO_RCVBUFFORCE` bypasses the `rmem_max` cap when the process has `CAP_NET_ADMIN`, which is typically available in SONiC containers. Consider trying it first and falling back to `SO_RCVBUF`:
```rust
let mut ret = unsafe {
    libc::setsockopt(fd, libc::SOL_SOCKET, libc::SO_RCVBUFFORCE,
                     &v as *const _ as *const libc::c_void,
                     std::mem::size_of::<libc::c_int>() as libc::socklen_t)
};
if ret != 0 {
    // Fallback to SO_RCVBUF (subject to rmem_max cap)
    ret = unsafe {
        libc::setsockopt(fd, libc::SOL_SOCKET, libc::SO_RCVBUF,
                         &v as *const _ as *const libc::c_void,
                         std::mem::size_of::<libc::c_int>() as libc::socklen_t)
    };
}
```

crates/countersyncd/src/main.rs
```diff
@@ -171,6 +171,31 @@ struct Args {
...
     log_format: String,
 
+    /// Interval (seconds) between periodic comm stats log lines (channel queue stats)
+    #[arg(
+        long,
+        default_value = "600",
+        help = "Interval in seconds for logging comm stats (channel lengths). Use a shorter value (e.g. 60) when verifying HFT processing slowness"
+    )]
+    comm_stats_interval: u64,
+
```
Pterosaur, line 182: **Missing minimum-value constraint on `comm_stats_interval`**

`socket_readiness_timeout_ms` has `value_parser = clap::value_parser!(u64).range(1..)` to prevent 0, but `comm_stats_interval` has no such guard. A value of 0 would cause comm stats to be logged on every single iteration, which could be extremely noisy.

Suggested fix:
```rust
#[arg(
    long,
    default_value = "600",
    value_parser = clap::value_parser!(u64).range(1..),
    help = "..."
)]
comm_stats_interval: u64,
```

crates/countersyncd/src/actor/data_netlink.rs
```diff
@@ -314,6 +315,69 @@ impl DataNetlinkActor {
         netlink_utils::create_nl_resolver()
     }
 
+    /// Sets SO_RCVBUF on the netlink socket to reduce ENOBUFS under high HFT load.
+    #[cfg(not(test))]
+    fn set_socket_rcvbuf(socket: &mut Socket, bytes: usize) {
```
nazariig: @DavidZagury this is not a `DataNetlinkActor` dedicated implementation, consider moving it out to some `utils` module.

Also, please double check whether it is possible to use native API:
```
// Set the receive buffer size (e.g., 1MB)
if let Err(e) = socket.set_rx_buf_sz(1024 * 1024) {
    warn!("Failed to set netlink receive buffer size: {:?}", e);
    // Usually, we continue even if this fails, but you can return None if it's critical
}
```

nazariig (changes requested):

crates/countersyncd/src/actor/data_netlink.rs
DavidZagury: Done

crates/countersyncd/src/main.rs
DavidZagury: @Pterosaur done

nazariig (approved):

Pterosaur (approved):

Pterosaur:
Hi @prsunny , could you please help to merge this PR?

PR #4388, opened 2026-03-25
Fix port breakout failure when port is member of port-channel

Keshavg-marvell:
**What I did**

In portmgr, defer port deletion from APP_DB if the port is still a member of LAG in APP_DB LAG_MEMBER_TABLE. The deletion is retried on subsequent doTask() cycles. Once teamsyncd removes the LAG member entry, portmgrd proceeds with the port deletion, ensuring proper sequencing and successful port breakout.

**Why I did it**

During port breakout of a port-channel member (in force mode), the CLI removes both the port and port-channel member config from CONFIG_DB, then waits up to 60 seconds for the port to be deleted from ASIC_DB.

A race condition between portmgrd (swss docker) and teammgrd (teamd docker) causes breakout failure:
1. portmgrd processes the port DEL and removes the port from APP_DB
2. teammgrd's removeLagMember() writes admin_status/mtu to APP_DB, inadvertently re-creating the port entry
3. orchagent sees the port still in APP_DB and never deletes it from ASIC_DB, causing the 60-second timeout and breakout failure as shown below.

$config interface breakout Ethernet176 "2x400G[200G,100G]" -y -f Running Breakout Mode : 1x800G[400G,200G]
Target Breakout Mode : 2x400G[200G,100G]
Ports to be deleted : { "Ethernet176": "800000" }
Ports to be added : {
      "Ethernet176": "400000", "Ethernet180": "400000"
}
!!!  Critical Failure, Ports are not Deleted from ASIC DB, Bail Out !!! Ports are present in ASIC DB after 60 secs
Ports are present in ASIC DB after 60 secs
[ERROR] Port breakout Failed!!! Opting Out
Failed to break out Port. Error:}

So here main issue cfgmgrs writing to same APP_PORT_TABLE then ordering of event plays role here and need special handling

Root cause - two cfgmgrs writing to the SAME APP_DB table: The core issue is that both portmgrd and teammgrd write to the SAME APP_PORT_TABLE in APP_DB. When two processes operate on the same table, the ordering of operations becomes critical - a SET from teammgrd can undo a DEL from portmgrd, effectively re-creating a port entry that was just deleted.

Contrast this with QoS/buffer configuration, where there is no such ordering problem: buffermgrd writes to SEPARATE APP_DB tables such as APP_BUFFER_PG_TABLE, APP_BUFFER_QUEUE_TABLE, APP_BUFFER_POOL_TABLE, and APP_BUFFER_PROFILE_TABLE - all different from APP_PORT_TABLE. When a port is deleted, it does not matter whether buffermgrd removes the buffer/QoS entries before or after portmgrd deletes the port from APP_PORT_TABLE. This is because orchagent internally maintains reference counts and
 will not delete the port from ASIC_DB until all dependent
objects (buffer PGs, queues, QoS maps etc.) referencing that port are removed first. The cross-table dependency resolution is handled entirely by orchagent'\''s reference counting mechanism, making the ordering of DEL operations between different cfgmgrs irrelevant.

**How I verified it**

Verified it on Hardware by successful port breakout when port is part of port-channel.

**Details if related**

### DIFF
--- a/cfgmgr/portmgr.cpp
+++ b/cfgmgr/portmgr.cpp
@@ -17,6 +17,7 @@ PortMgr::PortMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, c
         m_cfgSendToIngressPortTable(cfgDb, CFG_SEND_TO_INGRESS_PORT_TABLE_NAME),
         m_cfgLagMemberTable(cfgDb, CFG_LAG_MEMBER_TABLE_NAME),
         m_statePortTable(stateDb, STATE_PORT_TABLE_NAME),
+        m_appLagMemberTable(appDb, APP_LAG_MEMBER_TABLE_NAME),
         m_appSendToIngressPortTable(appDb, APP_SEND_TO_INGRESS_PORT_TABLE_NAME),
         m_appPortTable(appDb, APP_PORT_TABLE_NAME)
 {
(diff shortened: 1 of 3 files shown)


prsunny:
Why not check this at CLI level?

Keshavg-marvell:
@prsunny During breakout force mode, the CLI (config_mgmt) computes a diff between the existing configuration and what the configuration would look like after removing all references to the breakout interface. It then applies this entire diff set to CONFIG_DB in bulk, so the CLI does not have fine-grained control.

PR #4390, opened 2026-03-25
Protection Nexthop group support in nhgorch

manamand2020:
**What I did**
Extended support for  protection nexthop group in nhgorch. 
Added support for protection next hop groups to NhgOrch : a SAI next hop group with exactly two members, a primary and a standby, where the ASIC switches traffic between them either autonomously (via a monitored/health object) or under software control. This is the ECMP-adjacent counterpart to today's NhgOrch, which only manages SAI_NEXT_HOP_GROUP_TYPE_DYNAMIC* groups; protection NHGs use SAI_NEXT_HOP_GROUP_TYPE_PROTECTION.

*ProtNhg / ProtNhgMember (orchagent/protnhg.h, .cpp)* : ProtNhgMember wraps one primary/standby member with a configured role, an optional monitored object (e.g. a BFD/ICMP session OID driving hardware failover), and dynamic NH resolution (individual via NeighOrch, recursive via NhgOrch). ProtNhg owns the pair, supports both individual-NextHopKey and nested-NextHopGroupKey membership, and handles sync/remove with proper cleanup on failure.

NhgOrch protection API (tracked in new m_protNhgs map): createProtNhg() (4 overloads: explicit/auto key × NH-pair/NHG-pair, all idempotent), removeProtNhg(), buildProtNhgKey(), hasProtNhg()/getProtNhg()/getProtNhgId(), setProtNhgAdminRole()/setProtNhgSwitchover(), attachProtNhgMonitoredObject()/detachProtNhgMonitoredObject() (toggles between SW-driven and HW-autonomous mode), getProtNhgMemberObservedRole()/getProtNhgAllObservedRoles(), and incProtNhgRefCount()/decProtNhgRefCount(). NeighOrch's validateNextHop()/invalidateNextHop() were extended to also walk protection groups.

Capability probing distinguishes three independent things published to STATE_DB|SWITCH_CAPABILITY: whether the ASIC supports SAI_NEXT_HOP_GROUP_TYPE_PROTECTION at all, whether HW switchover is supported (probed via SAI_SWITCH_ATTR_SUPPORTED_PROTECTED_OBJECT_TYPE, since that's a property of the monitored object, not the NHG type), and whether SAI_NEXT_HOP_GROUP_TYPE_HW_PROTECTION is supported purely as a "backup-group hint" for the standby leg (never used to actually create a protection group).

Query: hasProtNhg(), getProtNhg(), getProtNhgId().
Control: setProtNhgAdminRole() (HW_PROTECTION only), setProtNhgSwitchover() (PROTECTION only), attachProtNhgMonitoredObject()/detachProtNhgMonitoredObject().
Observability: getProtNhgMemberObservedRole(), getProtNhgAllObservedRoles().
Ref counting: incProtNhgRefCount() / decProtNhgRefCount() for route/consumer ownership.

SAI specs for protection NHG: 
https://github.com/opencomputeproject/SAI/blob/master/doc/SAI-Proposal-FRR.md
https://github.com/opencomputeproject/SAI/blob/master/doc/SAI-Proposal-HW-FRR.md

**Why I did it**
This is needed to support hardware protection failover of mux port.
HLD: 
https://github.com/sonic-net/SONiC/pull/2251
https://github.com/sonic-net/SONiC/pull/2346

**How I verified it**
Added mock Gtests
**Details if related**


orchagent/protnhg.cpp
```diff
@@ -0,0 +1,439 @@
...
+    gCrmOrch->incCrmResUsedCounter(CrmResourceType::CRM_NEXTHOP_GROUP);
+    incSyncedCount();
+
+    set<NextHopKey> member_keys;
+    for (const auto &mbr : m_members)
+    {
+        member_keys.insert(mbr.first);
+    }
+
+    if (!syncMembers(member_keys))
```
Ndancejic, line 456: If syncMembers fails, does the next hop group get cleaned up?

orchagent/protnhg.h
```diff
@@ -0,0 +1,101 @@
...
+                               sai_next_hop_group_member_observed_role_t &observed_role) const;
+
+    /* Query observed roles for all synced members at once. */
+    bool getAllMemberObservedRoles(
+        map<NextHopKey, sai_next_hop_group_member_observed_role_t> &observed_roles) const;
+
+    string to_string() const override { return m_key; }
+
+private:
+    bool syncMembers(const set<NextHopKey> &member_keys) override;
```
Ndancejic: Does removeMembers need an override as well?

orchagent/protnhg.cpp
manamand2020: No it does not cleanup, I was expecting the caller will destroy the protnhg object which will cleanup the nhg. I will add the remove(); call when syncMembers() fail.

orchagent/protnhg.h
manamand2020: No not needed because the base implementation calls the removeMembers() which calls member.remove() on success which is already has override.

orchagent/protnhg.cpp
manamand2020: Added the remove() call when syncMembers() fails.

orchagent/nhgorch.cpp
```diff
@@ -1158,3 +1158,247 @@ bool NextHopGroup::invalidateNextHop(const NextHopKey& nh_key)
...
+        SWSS_LOG_NOTICE("Cannot query NHG type enum metadata");
+        return false;
+    }
+
+    vector<int32_t> values_list(meta->enummetadata->valuescount);
+    sai_s32_list_t values;
+    values.count = static_cast<uint32_t>(values_list.size());
+    values.list = values_list.data();
+
+    sai_status_t status = sai_query_attribute_enum_values_capability(
```
prashanth-nexthop: Couldn't you just query the attribute SAI_NEXT_HOP_GROUP_TYPE_HW_PROTECTION instead ?

orchagent/nhgorch.cpp
```diff
@@ -1158,3 +1158,247 @@ bool NextHopGroup::invalidateNextHop(const NextHopKey& nh_key)
...
+    }
+
+    SWSS_LOG_NOTICE("SAI_NEXT_HOP_GROUP_TYPE_HW_PROTECTION is %s",
+                    supported ? "supported" : "not supported");
+    return supported;
+}
+
+bool NhgOrch::createProtNhg(const string &key,
+                             const vector<NextHopKey> &primary_nhs,
+                             const NextHopKey &standby_nh,
```
prashanth-nexthop: Is this making an assumption that there is only one standby member?  Isn't that a HW limitation ?

manamand2020: This is not about hardware limitation. Protection and hardware protection NHG will always contain 2 members as per SAI definition. If we need multiple standby-nh then control plane can create another nhgroup and use that as standby.
Copying from SAI doc https://github.com/opencomputeproject/SAI/blob/master/doc/SAI-Proposal-FRR.md :
```
Firstly, a new type of Next Hop Group is specified – SAI_NEXT_HOP_GROUP_TYPE_PROTECTION. This indicates that the group does not represent ECMP next hops but a primary-backup pair.

It is expected that the Control Plane stack will never attempt to program more than two next hops within a single Protection Next Hop Group. It is outside the scope of this proposal to specify how the Adapter or Adapter Host should enforce this condition.
```

orchagent/nhgorch.cpp
manamand2020: Note that nexthop type is an enum and not an attribute. Attribute of nhg is SAI_NEXT_HOP_GROUP_ATTR_TYPE. I do not think we have a SAI API which can take single enum value and return a bool.

orchagent/nhgorch.cpp
manamand2020: Have used a vector for primary members for future extensions.

prashanth-nexthop: Thanks for the pointer to the HLD.  Like you said, I do see that the HLD limits primary and standby paths to just one nh each.  

But creating another group for standby members like you suggested, won't work well with this API. Right ?  

Alternatively, since you are using a vector for primary members, can't you do the same for standby as well ?

orchagent/nhgorch.cpp
prashanth-nexthop: Sorry.  You are right.  Please ignore.

orchagent/nhgorch.cpp
manamand2020: The future use case for multiple primary is slightly different. I will either remove the vector of primary since the SAI spec is missing it or I will add a comment about this.

manamand2020: For using Nexthop groups as primary and standby I will overload the createProtNhg() to accept NextHopGroupKey. Let me add that version.

tahmed-dev:
## Bug Fix: ProtNhg /32 route orphaning from neighbor churn

**File:** `orchagent/evpnmhorch.cpp`

Three bugs cause ASIC transit forwarding failures:

1. **Observer tears down ProtNhg on every neighbor DEL.** EVPN `extern_learn` neighbors from zebra are periodically reinstalled (`NUD_NOARP`), causing a create/destroy cycle every ~2 minutes. ProtNhg lifecycle should be tied to ES config, not individual neighbor events.

2. **`create_route_entry` returns `ITEM_ALREADY_EXISTS` treated as error.** When re-creating the /32 route after a churn cycle, the route still exists from neighorch. The ProtNhg is created but the route still points to the old NH OID. Fix: fall back to `set_route_entry_attribute` to update the NH.

3. **`remove_route_entry` return value unchecked.** If the ASIC fails to remove the /32 route, it remains with a stale NHG OID while the ProtNhg is destroyed.

```diff
+    if (status == SAI_STATUS_ITEM_ALREADY_EXISTS)
+    {
+        status = sai_route_api->set_route_entry_attribute(&route_entry, &route_attr);
+        // Update existing route NH → new ProtNhg
+    }
```

The orphaned /32 routes (more specific than /24 connected) capture all server-destined transit traffic and forward to invalid NHG OIDs → 100% packet loss on return path.

orchagent/nhgorch.cpp
manamand2020: Added API to accept Nhgrp key.

zjswhhh (dismissed):

prsunny:
> ## Bug Fix: ProtNhg /32 route orphaning from neighbor churn
> **File:** `orchagent/evpnmhorch.cpp`
> 
> Three bugs cause ASIC transit forwarding failures:
> 
> 1. **Observer tears down ProtNhg on every neighbor DEL.** EVPN `extern_learn` neighbors from zebra are periodically reinstalled (`NUD_NOARP`), causing a create/destroy cycle every ~2 minutes. ProtNhg lifecycle should be tied to ES config, not individual neighbor events.
> 2. **`create_route_entry` returns `ITEM_ALREADY_EXISTS` treated as error.** When re-creating the /32 route after a churn cycle, the route still exists from neighorch. The ProtNhg is created but the route still points to the old NH OID. Fix: fall back to `set_route_entry_attribute` to update the NH.
> 3. **`remove_route_entry` return value unchecked.** If the ASIC fails to remove the /32 route, it remains with a stale NHG OID while the ProtNhg is destroyed.
> 
> ```diff
> +    if (status == SAI_STATUS_ITEM_ALREADY_EXISTS)
> +    {
> +        status = sai_route_api->set_route_entry_attribute(&route_entry, &route_attr);
> +        // Update existing route NH → new ProtNhg
> +    }
> ```
> 
> The orphaned /32 routes (more specific than /24 connected) capture all server-destined transit traffic and forward to invalid NHG OIDs → 100% packet loss on return path.

@tahmed-dev , I see a reference to evpnmhorch.cpp. Could you clarify what needs to be addressed here?

prsunny:
@manamand2020 , please provide all details in the description.

prsunny:
@prashanth-nexthop , could you re-review/signoff?

manamand2020:
> @prashanth-nexthop , could you re-review/signoff?

@prashanth-nexthop Do we have more comments on the latest set of changes?

orchagent/nhgorch.cpp
```diff
@@ -1158,3 +1158,374 @@ bool NextHopGroup::invalidateNextHop(const NextHopKey& nh_key)
...
+    if (primary_nhs.empty())
+    {
+        SWSS_LOG_ERROR("Protection NHG %s requires at least one primary NH", key.c_str());
+        return false;
+    }
+
+    if (m_protNhgs.find(key) != m_protNhgs.end())
+    {
+        SWSS_LOG_ERROR("Protection NHG %s already exists", key.c_str());
+        return false;
```
prashanth-nexthop, line 1402: Should this be false ?  If the group already exists, isn't it better to return true for idompotency ?

orchagent/nhgorch.cpp
```diff
@@ -1158,3 +1158,374 @@ bool NextHopGroup::invalidateNextHop(const NextHopKey& nh_key)
...
+    if (!nhg->sync())
+    {
+        SWSS_LOG_ERROR("Failed to sync protection NHG %s", key.c_str());
+        return false;
+    }
+
+    m_protNhgs.emplace(key, NhgEntry<ProtNhg>(move(nhg)));
+
+    string primary_str;
+    for (const auto &nh : primary_nhs)
```
prashanth-nexthop: I see that none of the ProtNhg code is being used by any code.  So, I had a couple of questions.

1. I think there is an expectation that both primary and standby exist as NHGs.  So, if the ECMP members go down, they are required to be handled by existing NHG orchestrator ?  I mean if there were multiple primaries.
2. If regular ECMP is taking care of groups shrinking, for us to do switch over we need to detect all primaries being down.  So, we may have to prevent primary ECMP groups shrinking to 0 members before switch over.  

Have you thought about how that code flow is going to be ?

manamand2020: With primary as NHG there are 2 use cases:
1. Hardware protection: Note there is single monitored object for the hardware protection NHG. Based on state of monitored object switching will be done in hardware. 
2. Software protection: In this case switching will not be handled by ProtNhg. Software switching needs to be done by the user of protnhg using the exposed APIs.

This code will be modified as more requirements come. We plan to introduce 2 more APIs to accept different combination of primary and secondary member to support the following:
1. Primary NH | secondary NHG
2. Primary NHG | secondary NH

orchagent/protnhg.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+    sai_status_t status =
+        sai_next_hop_group_api->set_next_hop_group_member_attribute(m_gm_id, &attr);
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to update monitored object for member %s, rv: %d",
+                       to_string().c_str(), status);
+        return false;
+    }
```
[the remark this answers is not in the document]
manamand2020: Fixed.

orchagent/protnhg.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+        if (nhgm.isSynced())
+        {
+            continue;
+        }
+
+        if (nhgm.getNhId() == SAI_NULL_OBJECT_ID)
+        {
+            SWSS_LOG_WARN("Next hop %s not resolved for protection NHG %s",
+                          nh_key.to_string().c_str(), m_key.c_str());
```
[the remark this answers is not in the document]
manamand2020: Skipping is fine to add the other member. User of the API should manage the dependency when NH finally gets resolves by calling sync again. This allows to have unresolved members in the NHGroup.

prashanth-nexthop (approved):

orchagent/nhgorch.cpp
manamand2020: Done.

orchagent/muxorch.cpp
```diff
@@ -1521,6 +1521,11 @@ sai_object_id_t MuxOrch::createNextHopTunnel(std::string tunnelKey, swss::IpAddr
     if (SAI_NULL_OBJECT_ID != nh)
     {
         mux_tunnel_nh_[ipAddr] = { nh, 1 };
+        NextHopKey nhKey(ipAddr, tunnelKey, true /*tunnel_nh*/, 0 /*tag*/);
+        if (gNeighOrch)
+        {
```
abdosi: can this be moved to seprate PR ? Does not seems correct in context of this PR which is more on infrastructure change to bring Protection Nexthop Group

orchagent/neighorch.h
```diff
@@ -100,6 +100,9 @@ class NeighOrch : public Orch, public Subject, public Observer
     sai_object_id_t addTunnelNextHop(const NextHopKey&);
     bool removeTunnelNextHop(const NextHopKey&);
 
+    bool addIpinipTunnelNextHop(const NextHopKey& nh, sai_object_id_t nh_id);
+    bool removeIpinipTunnelNextHop(const NextHopKey& nh);
```
abdosi: same as above

orchagent/neighorch.cpp
```diff
@@ -2045,6 +2045,63 @@ bool NeighOrch::removeTunnelNextHop(const NextHopKey& nh)
...
+
+bool NeighOrch::removeIpinipTunnelNextHop(const NextHopKey& nh)
+{
+    SWSS_LOG_ENTER();
+
+    auto it = m_syncdNextHops.find(nh);
+    if (it == m_syncdNextHops.end())
+    {
+        SWSS_LOG_ERROR("IPinIP tunnel NH not found: %s", nh.to_string().c_str());
+        return false;
```
abdosi: same as above

orchagent/nexthopkey.h
```diff
@@ -177,6 +203,12 @@ struct NextHopKey
         return (srv6_vpn_sid != "");
     }
 
+    bool isTunnelNextHop() const
+    {
+        return (!tunnel_name.empty());
+    }
+
```
abdosi: same as above

orchagent/nexthopkey.h
```diff
@@ -33,6 +34,7 @@ struct NextHopKey
     string              srv6_segment;   // SRV6 segment string
     string              srv6_source;    // SRV6 source address
     string              srv6_vpn_sid;   // SRV6 vpn sid
+    string              tunnel_name;    // IPinIP tunnel name
```
abdosi: same as above

orchagent/nhgorch.cpp
```diff
@@ -1158,3 +1158,377 @@ bool NextHopGroup::invalidateNextHop(const NextHopKey& nh_key)
...
+
+    return prefix + primary_str + "|" + standby_nh.to_string();
+}
+
+string NhgOrch::buildProtNhgKey(const NextHopGroupKey &primary_nhg_key,
+                                 const NextHopGroupKey &standby_nhg_key,
+                                 bool hw_protection)
+{
+    string prefix = hw_protection ? "prot:hw:" : "prot:sw:";
+    return prefix + primary_nhg_key.to_string() + "|" + standby_nhg_key.to_string();
```
abdosi: @prashanth-nexthop do we don't need NHG_DELIMITER same as being done in above API ?

orchagent/nhgorch.cpp
```diff
@@ -1158,3 +1158,377 @@ bool NextHopGroup::invalidateNextHop(const NextHopKey& nh_key)
...
+    {
+        if (values.list[i] == SAI_NEXT_HOP_GROUP_TYPE_HW_PROTECTION)
+        {
+            supported = true;
+            break;
+        }
+    }
+
+    SWSS_LOG_NOTICE("SAI_NEXT_HOP_GROUP_TYPE_HW_PROTECTION is %s",
+                    supported ? "supported" : "not supported");
```
abdosi: can we update state db to keep the track of this capability ?

orchagent/nhgorch.cpp
```diff
@@ -1158,3 +1158,377 @@ bool NextHopGroup::invalidateNextHop(const NextHopKey& nh_key)
...
+
+    SWSS_LOG_NOTICE("Created protection NHG %s (primaries: [%s], standby: %s)",
+                    key.c_str(),
+                    primary_str.c_str(),
+                    standby_nh.to_string().c_str());
+
+    return true;
+}
+
+string NhgOrch::buildProtNhgKey(const vector<NextHopKey> &primary_nhs,
```
abdosi: why can't this be const NextHopGroupKey &primary_nhg_key ?  Do we need explicit sorting here ? NexthopGroupKey is self sorted

orchagent/protnhg.h
```diff
@@ -0,0 +1,119 @@
...
+    ProtNhgRole m_role;
+    sai_object_id_t m_monitored_oid;
+    string m_nhg_key;   /* Non-empty when member is a recursive NHG (resolved via NhgOrch). */
+};
+
+/*
+ * ProtNhg represents a SAI_NEXT_HOP_GROUP_TYPE_HW_PROTECTION group.
+ * It has one or more primary next hops and exactly one standby next hop.
+ * Hardware toggles traffic between the primary set and the standby based on
+ * the monitored object state. Administrative override is supported via
```
abdosi: can we update comment as this infrastructure is not just only Hw based protection

orchagent/protnhg.cpp
```diff
@@ -0,0 +1,525 @@
...
+            continue;
+        }
+
+        vector<sai_attribute_t> attrs = createNhgmAttrs(nhgm);
+        bulker.create_entry(&syncing[nh_key],
```
abdosi: where are we incrementing CrmResourceType::CRM_NEXTHOP_GROUP_MEMBER ?

orchagent/protnhg.cpp
```diff
@@ -0,0 +1,525 @@
...
+{
+    SWSS_LOG_ENTER();
+
+    if (isSynced())
+    {
+        return true;
+    }
+
+    if (m_members.size() < 2)
+    {
```
abdosi, line 404: this might not be true .. we can have 2 primary members but no stand by 

@prashanth-nexthop what do you think ?

orchagent/protnhg.h
```diff
@@ -0,0 +1,119 @@
...
+    void remove() override;
+
+    inline ProtNhgRole getRole() const { return m_role; }
+    sai_object_id_t getNhId() const;
+
+    /* True when this member represents a recursive NHG rather than an individual NH. */
+    inline bool isRecursive() const { return !m_nhg_key.empty(); }
+
+    inline sai_object_id_t getMonitoredObject() const { return m_monitored_oid; }
+    void setMonitoredObject(sai_object_id_t oid) { m_monitored_oid = oid; }
```
abdosi, line 100: can we make sure here that it's hw-protected protection group before setting it ?

orchagent/protnhg.h
```diff
@@ -0,0 +1,119 @@
...
+
+    /* Query the hardware-observed role (active/inactive) from SAI. */
+    bool getObservedRole(sai_next_hop_group_member_observed_role_t &observed_role) const;
+
+    string to_string() const override;
+
+private:
+    ProtNhgRole m_role;
+    sai_object_id_t m_monitored_oid;
+    string m_nhg_key;   /* Non-empty when member is a recursive NHG (resolved via NhgOrch). */
```
abdosi: assumption here is we support 1 level of recursive ?

orchagent/nhgorch.cpp
manamand2020: This API was added for some planned SAI enhancement where multiple primary NHs are actually part of the single hw protection NHG. Since we have not yet ironed out the SAI model for this I can remove this API at this point to avoid confusion,

orchagent/muxorch.cpp
manamand2020: Does it seem confusing? I will like to keep this together to keep the context related to this infra.

abdosi: yes, lot of API's are will get used from muxorch . I would tie up the change as seprate PR or add them as part of muxorch

orchagent/nhgorch.cpp
abdosi: what change is needed SAI ? @prashanth-nexthop please confirm ?

orchagent/protnhg.h
manamand2020: Sure will update this.

orchagent/protnhg.cpp
manamand2020: As per SAI spec documentation protection groups will have only 2 members. Copying from SAI doc https://github.com/opencomputeproject/SAI/blob/master/doc/SAI-Proposal-FRR.md :
```
Firstly, a new type of Next Hop Group is specified – SAI_NEXT_HOP_GROUP_TYPE_PROTECTION. This indicates that the group does not represent ECMP next hops but a primary-backup pair.

It is expected that the Control Plane stack will never attempt to program more than two next hops within a single Protection Next Hop Group. It is outside the scope of this proposal to specify how the Adapter or Adapter Host should enforce this condition.
```

orchagent/muxorch.cpp
manamand2020: Done, moved to new PR: https://github.com/sonic-net/sonic-swss/pull/4646

orchagent/nhgorch.cpp
manamand2020: Done

orchagent/protnhg.cpp
manamand2020: CRM_NEXTHOP_GROUP_MEMBER is already incremented/decremented for protection NHG members through the inherited NhgMember::sync()/remove() path — identical to how NextHopGroup does it. 

Increment: ProtNhg::syncMembers() calls member.sync() → ProtNhgMember::sync() → NhgMember::sync() (nhgbase.h:132).
Decrement: ProtNhg::remove() → NhgCommon::removeMembers() → member.remove() → ProtNhgMember::remove() → NhgMember::remove() (nhgbase.h:151).
This is the same pattern NextHopGroup uses (nhgorch.cpp:983), which has no explicit CRM_NEXTHOP_GROUP_MEMBER call either.

I will update a comment here.

orchagent/protnhg.h
manamand2020: Actually SAI allows setting a monitored object even for normal SW based protection NHG.

orchagent/nhgorch.cpp
manamand2020: Removed APIs supporting multiple primary NH as this is not yet specified by SAI.

orchagent/muxorch.cpp
abdosi: thanks

abdosi:
@prashanth-nexthop : please review the comments so that we can do sign-off again

orchagent/protnhg.h
manamand2020: yes

zjswhhh (dismissed):

manamand2020:
@prsunny @dgsudharsan do you have any comments on this? You approvals will help merge this.

tahmed-dev (commented):
Reviewed for correctness and for interaction with FPM/fdbsyncd work landing in parallel — no overlap there apart from `tests/mock_tests/Makefile.am`, and that's a different hunk (`tests_SOURCES` vs `tests_fpmsyncd_SOURCES`).

CRM accounting, move semantics and the failed-`sync()` cleanup path all check out. Two issues below look blocking; the rest are minor.

orchagent/nhgorch.cpp
```diff
@@ -1161,3 +1163,375 @@ bool NextHopGroup::invalidateNextHop(const NextHopKey& nh_key)
 
     return true;
 }
+
+/* ----------------------------------------------------------------------- */
+/* Protection NHG management APIs                                          */
+/* ----------------------------------------------------------------------- */
+
+void NhgOrch::probeProtectionCapabilities()
```
tahmed-dev, line 1223: `probeProtectionCapabilities()` is only reachable via `isSwProtectionSupported()`/`isHwProtectionSupported()`, and neither is called outside `protnhg_ut.cpp`. Two consequences:

- `SW_/HW_NHG_PROTECTION_CAPABLE` is never written to `STATE_DB` on a real switch, so the discovery mechanism doesn't work in production.
- `createProtNhg()` never consults the result, so on a platform without the type it goes straight to `create_next_hop_group()` and fails in SAI.

Suggest probing once at init and rejecting in `createProtNhg()` when the requested type is unsupported. FWIW on the Broadcom SAI we're running, an unsupported attribute of this kind produces a retrying create rather than a terminal error, so the failure mode is noisy rather than obvious.

orchagent/protnhg.cpp
```diff
@@ -0,0 +1,529 @@
...
+    for (const auto &nh_key : member_keys)
+    {
+        ProtNhgMember &nhgm = m_members.at(nh_key);
+
+        if (nhgm.isSynced())
+        {
+            continue;
+        }
+
+        if (nhgm.getNhId() == SAI_NULL_OBJECT_ID)
```
tahmed-dev, line 843: Unresolved members `continue` without being added to `syncing`. If neither next hop is resolved, the loop over `syncing` never executes, `success` stays `true`, and `sync()` returns success having created a group with zero members.

There's also no recovery path — `NhgOrch::validateNextHop()`/`invalidateNextHop()` (called from `neighorch.cpp:609/641`) iterate `m_syncedNhgs` only and aren't extended to `m_protNhgs`, so members skipped here are never synced later. Creating a protection NHG before ARP resolution leaves it permanently degraded while reporting success.

orchagent/protnhg.cpp
```diff
@@ -0,0 +1,529 @@
...
+ProtNhg::ProtNhg(const string &key,
+                  const NextHopGroupKey &primary_nhg_key,
+                  const NextHopGroupKey &standby_nhg_key,
+                  bool hw_protection) :
+    NhgCommon(key),
+    m_hw_protection(hw_protection)
+{
+    SWSS_LOG_ENTER();
+
+    const NextHopKey &primary_rep = *primary_nhg_key.getNextHops().begin();
```
tahmed-dev: Members are keyed by `*nhg_key.getNextHops().begin()`. If the primary and standby groups share that next hop, the second `m_members.emplace()` is a no-op and the group ends up with a single member; `sync()` then reports "must have at least 2 members", which points away from the actual cause.

Either reject overlapping representatives explicitly, or key members on the NHG key rather than a representative next hop.

orchagent/protnhg.h
```diff
@@ -0,0 +1,128 @@
...
+};
+
+/*
+ * ProtNhg represents a SAI protection next hop group of either type, selected
+ * at construction via the hw_protection flag:
+ *   - SAI_NEXT_HOP_GROUP_TYPE_PROTECTION    (software protection), or
+ *   - SAI_NEXT_HOP_GROUP_TYPE_HW_PROTECTION (hardware protection).
+ * It is a strict pair: exactly one primary next hop and exactly one standby
```
tahmed-dev, line 147: Comment says "strict pair: exactly one primary next hop and exactly one standby", but `sync()` checks `m_members.size() < 2`, `getPrimaryMembers()` returns a vector, and `getStandbyMember()` returns the first match. Worth aligning the comment with the intended cardinality, or enforcing `== 2`.

tests/mock_tests/protnhg_ut.cpp
```diff
@@ -0,0 +1,1100 @@
...
+        EXPECT_NE(standby->getNhId(), SAI_NULL_OBJECT_ID);
+
+        ASSERT_TRUE(gNhgOrch->removeProtNhg(key));
+        removeEcmpNhg(primary_nhg_key.to_string());
+        removeEcmpNhg(standby_nhg_key.to_string());
+    }
+
+    /* --- Protection capabilities are published to STATE_DB --- */
+
+    TEST_F(ProtNhgTest, ProtectionCapabilitiesPublishedToStateDb)
```
tahmed-dev, line 1693: This passes because the test calls `isHwProtectionSupported()`/`isSwProtectionSupported()` directly; nothing in orchagent does, so the production path isn't covered.

Two cases that would catch the issues flagged in `nhgorch.cpp` and `syncMembers()`: create with an unresolved next hop (assert member count, and that a later resolution syncs it), and create when the group type is unsupported.

orchagent/nhgorch.cpp
manamand2020: - createProtNhg() now checks isHwProtectionSupported()/isSwProtectionSupported() before calling SAI, rejecting unsupported types up front instead of hitting a noisy SAI retry.
- This call is what triggers probeProtectionCapabilities() in production, so STATE_DB|SWITCH_CAPABILITY now actually gets populated on a real switch (previously only exercised by tests).
- Added tests asserting create_next_hop_group is never called when the capability is unsupported.

orchagent/protnhg.cpp
manamand2020: - syncMembers() now sets success = false for unresolved members instead of silently skipping them, so a group with zero synced members no longer reports success.
- Added ProtNhg::validateNextHop()/invalidateNextHop(), and extended NhgOrch::validateNextHop()/invalidateNextHop() to also iterate m_protNhgs (previously only m_syncdNextHopGroups).
- NeighOrch::addNextHop()/processBulkAddNextHop() now call gNhgOrch->validateNextHop() on resolution, so a previously-unresolved member gets synced later instead of staying permanently degraded.
- createProtNhg() keeps the group registered even on partial sync (self-heals via the above) rather than discarding it.
- Added tests: UnresolvedMemberSyncedLaterViaValidateNextHop, InvalidateNextHopRemovesMemberAndValidateRestoresIt, ValidateNextHopNoOpWhenMemberNotFound.

orchagent/protnhg.cpp
manamand2020: - ProtNhg::sync() now counts members by role (primary/standby) instead of m_members.size() < 2, so a representative-next-hop collision (this constructor) collapses to 1 primary/0 standby and fails with a clear "must have exactly one primary and one standby" error instead of the misleading "at least 2 members" message.
- This is the same fix as the direct-NextHopKey overload's collision case — one check in sync() covers both constructors since they share m_members.

orchagent/protnhg.h
manamand2020: - Enforced the cardinality in sync(): replaced m_members.size() < 2 with a role-based count, requiring exactly one primary and one standby member.
- Renamed getPrimaryMembers() (vector) → getPrimaryMember() (single pointer), matching getStandbyMember().
- Updated the class comment to note the strict pair is enforced in sync(), not just convention.

tests/mock_tests/protnhg_ut.cpp
manamand2020: - Added CreateFailsWhenHwProtectionUnsupported / CreateFailsWhenSwProtectionUnsupported — create with an unsupported group type, assert create_next_hop_group is never called and the group isn't registered.
- Added UnresolvedMemberSyncedLaterViaValidateNextHop — create with one next hop unresolved, assert member count stays 2 and the standby syncs once validateNextHop() fires.

orchagent/nhgorch.cpp
tahmed-dev: Thanks, the guard in both `createProtNhg()` overloads does make `probeProtectionCapabilities()` reachable in production, which was the main point.

One part is still open though: `SW_/HW_NHG_PROTECTION_CAPABLE` now reaches STATE_DB only as a side effect of the first `createProtNhg()` call. On a switch where no protection NHG is ever created the capability never appears, so anything polling STATE_DB at startup to discover support still sees nothing. If STATE_DB is meant as the discovery surface, probing once at init would close that. If it is only meant as a diagnostic for groups that were actually attempted, then this is fine as is and worth a comment saying so.

orchagent/protnhg.cpp
```diff
@@ -0,0 +1,578 @@
...
+bool ProtNhg::validateNextHop(const NextHopKey &nh_key)
+{
+    SWSS_LOG_ENTER();
+
+    if (!isSynced())
+    {
+        return true;
+    }
+
+    return syncMembers({nh_key});
```
tahmed-dev: `validateNextHop()` and `invalidateNextHop()` don't check `hasMember(nh_key)` before calling `syncMembers()`/`removeMembers()`. Both reach `m_members.at(nh_key)`, which throws `std::out_of_range` on a key that isn't a member. `ProtNhg::syncMembers()` does that lookup directly, and `removeMembers()` does it through the inherited `NhgCommon` implementation. These are public, so a direct caller takes orchagent down. Today it is safe only because `NhgOrch::validateNextHop()` guards with `hasMember()` at the call site.

`ValidateNextHopNoOpWhenMemberNotFound` doesn't cover this. It calls `gNhgOrch->validateNextHop()`, i.e. the guarded caller, so it passes without reaching the path its name describes. A `hasMember()` check in each function would make the API safe on its own and let that test exercise it directly.

manamand2020: Addressed and added hasMember() check.

orchagent/nhgorch.cpp
manamand2020: We will add doTask() control later and will make this init time setting.

orchagent/protnhg.cpp
tahmed-dev: Thanks, the unresolved-member path now behaves as expected, and `NhgOrch::validateNextHop()`/`invalidateNextHop()` iterating `m_protNhgs` closes the recovery gap.

One thing that surfaced while reading it: `syncMembers()` now returns `false` for two cases that need different handling.

1. Next hop unresolved. This self-heals later through `validateNextHop()`.
2. `create_next_hop_group_member` failed in SAI, the `SWSS_LOG_ERROR("Failed to create member %s of protection NHG %s", ...)` branch. There is no retry path for this one, because `validateNextHop()` only fires when a next hop transitions to resolved.

`createProtNhg()` treats both identically:

```cpp
if (!synced)
{
    SWSS_LOG_WARN("Protection NHG %s created with unresolved member(s); "
                  "will complete once the next hop(s) are validated", ...);
}
m_protNhgs.emplace(key, NhgEntry<ProtNhg>(move(nhg)));
```

So in case 2 the group is registered while permanently missing a member, and the warning attributes it to an unresolved next hop. That is the same "silently incomplete group" shape as the original comment, reached through the SAI failure path rather than the unresolved path.

Would it work to separate the two, for example have `syncMembers()` report unresolved counts separately from SAI failures, and have `createProtNhg()` retain the group only when every failure was "unresolved"? A genuine SAI failure would then fail creation, as it did before.

Small bookkeeping note: the earlier thread about adding `remove()` when `syncMembers()` fails is resolved with "Added the remove() call when syncMembers() fails", but the current code deliberately keeps the group so it can self-heal. Worth updating that thread so the history is preserved.

tahmed-dev:
Thanks for this. The header documentation and the explicit primary/standby enforcement in `sync()` made the design easy to follow.

A few questions to make sure I understand the intended behaviour before going deeper:

**1. Member lifecycle on next-hop invalidation.** `NhgOrch::invalidateNextHop()` now also walks `m_protNhgs` and calls `ProtNhg::invalidateNextHop()`, which removes that member from the group. What is the intended state of the protection group at that point? `sync()` requires exactly one primary and one standby, so I was not sure whether a group left with a single member is a supported steady state or expected to be transient, and what the ASIC does with traffic while it is in that state.

**2. Recursive members.** For the `NextHopGroupKey` pair constructor, the member is keyed on `*primary_nhg_key.getNextHops().begin()`, while ref counting goes through `incNhgRefCount(m_nhg_key)`. Is a recursive member's validity intended to track that one representative next hop, or the nested group as a whole? I could not tell which lifecycle is authoritative.

**3. HW-autonomous groups.** When a monitored object is attached and the hardware owns the switchover decision, is orchagent still expected to react to the primary next hop going down, or is that deliberately left to the hardware? I was not sure how the two mechanisms are meant to interact.

**4. `isHwAutonomous()` scope.** It keys off the primary member's monitored object only. Is attaching a monitored object to the standby leg alone a configuration you expect, and if so which mode should the group be in?

manamand2020:
> Thanks for this. The header documentation and the explicit primary/standby enforcement in `sync()` made the design easy to follow.
> 
> A few questions to make sure I understand the intended behaviour before going deeper:
> 
> **1. Member lifecycle on next-hop invalidation.** `NhgOrch::invalidateNextHop()` now also walks `m_protNhgs` and calls `ProtNhg::invalidateNextHop()`, which removes that member from the group. What is the intended state of the protection group at that point? `sync()` requires exactly one primary and one standby, so I was not sure whether a group left with a single member is a supported steady state or expected to be transient, and what the ASIC does with traffic while it is in that state.
> 
> **2. Recursive members.** For the `NextHopGroupKey` pair constructor, the member is keyed on `*primary_nhg_key.getNextHops().begin()`, while ref counting goes through `incNhgRefCount(m_nhg_key)`. Is a recursive member's validity intended to track that one representative next hop, or the nested group as a whole? I could not tell which lifecycle is authoritative.
> 
> **3. HW-autonomous groups.** When a monitored object is attached and the hardware owns the switchover decision, is orchagent still expected to react to the primary next hop going down, or is that deliberately left to the hardware? I was not sure how the two mechanisms are meant to interact.
> 
> **4. `isHwAutonomous()` scope.** It keys off the primary member's monitored object only. Is attaching a monitored object to the standby leg alone a configuration you expect, and if so which mode should the group be in?

**# 1.** You're right that it's not a supported steady state, and digging in, the invalidate walk shouldn't be there at all. Its only trigger is interface oper-down (`setNextHopFlag(NHFLAGS_IFDOWN)`); neighbor deletion never reaches it. What I needed was the validate side for deferred member resolution, and I mirrored invalidate for symmetry with the ECMP path — but member pruning is ECMP's way to steer off a dead port, while a protection group has SET_SWITCHOVER and the monitored object. Worse, the monitored object is a member attribute, so deleting the primary in HW-autonomous mode removes the monitoring for the very event hardware handles, and `isHwAutonomous()` stays true so `setSwitchover()` is refused.

I will fix by keeping the walk in `validateNextHop()`, drop it from `invalidateNextHop()`, and leave per-mode IFDOWN handling to ProtNhgOrch. Only the UT changes.

orchagent/nhgorch.cpp
```diff
@@ -543,22 +551,13 @@ bool NhgOrch::invalidateNextHop(const NextHopKey& nh_key)
         }
     }
 
-    /* Also invalidate the next hop in any protection groups containing it. */
```
prashanth-nexthop: Was there a reason for removing this ?

manamand2020: The invalidate walk shouldn't be there at all. Its only trigger is interface oper-down (setNextHopFlag(NHFLAGS_IFDOWN)); neighbor deletion never reaches it. What I needed was the validate side for deferred member resolution, and I mirrored invalidate for symmetry with the ECMP path, but member pruning is ECMP's way to steer off a dead port, while a protection group has SET_SWITCHOVER and the monitored object. Worse, the monitored object is a member attribute, so deleting the primary in HW-autonomous mode removes the monitoring for the very event hardware handles, and isHwAutonomous() stays true so setSwitchover() is refused.
Fixed by keeping the walk in validateNextHop(), drop it from invalidateNextHop(), and leave per-mode IFDOWN handling to ProtNhgOrch.

manamand2020: @prashanth-nexthop I had added this comment:
```
    /*
     * Protection NHGs are deliberately skipped: both legs stay programmed and
     * switching to the standby is the owning application's decision, applied
     * with SAI_NEXT_HOP_GROUP_ATTR_SET_SWITCHOVER or left to the hardware via
     * the monitored object.  Dropping a member here would pre-empt that, and
     * would delete the primary's monitored object with it.
     */
```
We can add logic to find out the member role and do the switchover when protnhg is not operating in hardware autonomous mode(no monitored object). We would need the reverse logic in the validate case as well then. But I'm not in favor of adding it here and want applications to decide that. If we really need this here then application must set a contract/rule via another protnhg API. Please note the keys have changed as well, there is no notion of sw/hw in the key in itself anymore.

PR #4395, opened 2026-03-26
[fpmsyncd]: add  nhgmgr to support RIB/FIB design

LARLSN:
**What I did**

Adding nhg mgr to realize fib block in RIB/FIB design.

**Why I did it**

In the RIB/FIB design, we need to add a fib block to fpmsyncd to process the NHG information received from zebra, and create corresponding Sonic NHG objects according to zebra information. Through fib block, NHG ID processing and PIC (prefix-independent convergence) can be introduced.In this way, the goal of improving routing convergence performance in RIB/FIB design can be achieved.

More details about RIB/FIB please refer RIB/FIB HLD pr:  https://github.com/sonic-net/SONiC/pull/2060
More details about nhg mgr please refer nhg mgr LLD pr: https://github.com/sonic-net/SONiC/pull/2270

**How I verified it**

1. Unit tests in code
2. Daily vsonic tests

**Details if related**

### DIFF
--- a/fpmsyncd/nhgmgr.cpp
+++ b/fpmsyncd/nhgmgr.cpp
@@ -2,307 +2,518 @@
 #include "logger.h"
 #include <string.h>
 
-#define NHG_DELIMITER ','
 
 using namespace std;
 using namespace swss;
 
-NHGMgr::NHGMgr(RedisPipeline *pipeline, const std::string &tableName, bool isStateTable) {
-	m_rib_nhg_table = new RIBNHGTable(pipeline, tableName, isStateTable);
-	m_sonic_id_manager.recoverSonicIDMapFromDB();
+NHGMgr::NHGMgr(RedisPipeline *pipeline, const std::string &nexthopTableName, const std::string &picTableName, bool isStateTable) {
+    m_rib_nhg_table = new RIBNHGTable(pipeline, nexthopTableName, isStateTable);
+    m_sonic_nhg_table = new SonicNHGTable(pipeline, picTableName, isStateTable);
+    m_sonic_id_manager.init({SONIC_NHG_OBJ_TYPE_NHG_SRV6_GATEWAY, SONIC_NHG_OBJ_TYPE_NHG_NORMAL});
 }
 
 int NHGMgr::addNHGFull(NextHopGroupFull nhg) {
-	int ret = 0;
-	if (m_rib_nhg_table->isNHGExist(nhg.id)) {
-		ret = m_rib_nhg_table->updateEntry(nhg);
-	} else {
-		ret = m_rib_nhg_table->addEntry(nhg);
-		RIBNHGEntry *entry = m_rib_nhg_table->getEntry(nhg.id);
-		if (entry == nullptr) {
-			return -1;
-		}
-		uint32_t sonic_id = m_sonic_id_manager.allocateID();
-		entry->setSonicObjId(sonic_id);
-		if (m_rib_nhg_table->writeToDB(entry) != 0) {
-			m_rib_nhg_table->delEntry(nhg.id);
-			m_sonic_id_manager.freeID(sonic_id);
-			SWSS_LOG_ERROR("Failed to write to DB for %d", nhg.id);
-			return -1;
-		}
-	}
+    int ret = 0;
 
+    // insert rib nhg in rib table
+    RIBNHGEntry *entry;
+    bool updated = false;
+    if (m_rib_nhg_table->isNHGExist(nhg.id)) {
+        entry = m_rib_nhg_table->getEntry(nhg.id);
+        if (entry == nullptr) {
+            SWSS_LOG_ERROR("Failed to get NHG %d", nhg.id);
+            return ret;
+        }
+        bool hasPreviousSonicObj = entry->hasSonicObj();
+        uint32_t previousSonicGatewayObjID = entry->getSonicGatewayObjID();
+        sonicNhgObjType previousType = entry->getSonicObjType();
+        ret = m_rib_nhg_table->updateEntry(nhg, updated);
+        if (ret != 0) {
+            SWSS_LOG_ERROR("Failed to update NHG %d", nhg.id);
+            return ret;
+        }
+        entry = m_rib_nhg_table->getEntry(nhg.id);
+        if (entry == nullptr) {
+            return ret;
+        }
+        if (updated) {
+            ret = m_rib_nhg_table->writeToDB(entry);
+            if (ret != 0) {
+                SWSS_LOG_ERROR("Failed to write to DB for %d", nhg.id);
+                return ret;
+            }
+            if (entry->hasSonicObj()) {
+                SonicGateWayNHGObjectKey key;
+                ret = SonicGateWayNHGObjectKey::createSonicGateWayNHGObjectKey(entry, key);
+                if (ret != 0) {
+                    SWSS_LOG_ERROR("Failed to create sonic nhg object key for %d", nhg.id);
+                    return ret;
+                }
+                if (m_sonic_nhg_table->getEntry(key) != nullptr) {
+                    SWSS_LOG_WARN("Sonic NHG Object with key already exists");
+                    return 0;
+                }
+                ret = createSonicNhgObject(entry);
+                if (ret != 0) {
+                    m_rib_nhg_table->removeFromDB(entry);
+                    m_rib_nhg_table->delEntry(nhg.id);
+                    SWSS_LOG_ERROR("Failed to create sonic nhg %d", nhg.id);
+                    return ret;
+                }
+                if (hasPreviousSonicObj) {
+                    m_sonic_nhg_table->delEntry(previousType, previousSonicGatewayObjID);
+                    m_sonic_id_manager.freeID(previousType, previousSonicGatewayObjID);
+                }
 
-	// TODO: srv6 sonic nhg create
-	return ret;
+            } else {
+                if (hasPreviousSonicObj) {
+                    m_sonic_nhg_table->delEntry(previousType, previousSonicGatewayObjID);
+                    m_sonic_id_manager.freeID(previousType, previousSonicGatewayObjID);
+                    entry->setSonicGatewayObjId(0);
+                }
+            }
+        }
+    } else {
+        ret = m_rib_nhg_table->addEntry(nhg);
+        if (ret != 0) {
+            SWSS_LOG_ERROR("Failed to add NHG %d", nhg.id);
+            return ret;
+        }
+        entry = m_rib_nhg_table->getEntry(nhg.id);
+        if (entry == nullptr) {
+            return ret;
+        }
+
+        if (entry->isEntryNeedOffload()) {
+            uint32_t sonicId = m_sonic_id_manager.allocateID(SONIC_NHG_OBJ_TYPE_NHG_NORMAL);
+            if (sonicId == 0) {
+                SWSS_LOG_ERROR("Failed to allocate sonic nhg id");
+                m_rib_nhg_table->delEntry(nhg.id);
+                return -1;
+            }
+
+            entry->setSonicObjId(sonicId);
+            ret = m_rib_nhg_table->writeToDB(entry);
+            if (ret != 0) {
+                m_rib_nhg_table->delEntry(nhg.id);
+                m_sonic_id_manager.freeID(SONIC_NHG_OBJ_TYPE_NHG_NORMAL, sonicId);
+                SWSS_LOG_ERROR("Failed to write to DB for %d", nhg.id);
+                return ret;
+            }
+        }
+
+        if (entry->hasSonicObj()) {
+            SonicGateWayNHGObjectKey key;
+
+            ret = SonicGateWayNHGObjectKey::createSonicGateWayNHGObjectKey(entry, key);
+            if (ret != 0) {
+                SWSS_LOG_ERROR("Failed to create sonic nhg object key for %d", nhg.id);
+                return ret;
+            }
+            if (m_sonic_nhg_table->getEntry(key) != nullptr) {
+                SWSS_LOG_WARN("Sonic NHG Object with key  already exists");
+                return 0;
+            }
+
+            ret = createSonicNhgObject(entry);
+            if (ret != 0) {
+                m_rib_nhg_table->removeFromDB(entry);
+                m_rib_nhg_table->delEntry(nhg.id);
+                SWSS_LOG_ERROR("Failed to create sonic nhg %d", nhg.id);
+                return ret;
+            }
+        }
+    }
+    return ret;
+}
+
+int NHGMgr::createSonicNhgObject(RIBNHGEntry *entry) {
+    // srv6 sonic nhg create
+    SonicGateWayNHGObject sonicObj;
+    uint32_t sonicGatewayNHGID;
+    int ret = 0;
+    sonicNhgObjType sType = entry->getSonicObjType();
+
+    sonicGatewayNHGID = m_sonic_id_manager.allocateID(sType);
+    if (sonicGatewayNHGID == 0) {
+        SWSS_LOG_ERROR("Failed to allocate sonic nhg id");
+        return -1;
+    }
+
+    switch (entry->getSonicObjType()) {
+        case SONIC_NHG_OBJ_TYPE_NHG_SRV6_GATEWAY: {
+
+            entry->createSRv6GatewayObjFromRIBEntry(sonicObj);
+            sonicObj.id = sonicGatewayNHGID;
+            break;
+        }
+        default: {
+            m_sonic_id_manager.freeID(sType, sonicGatewayNHGID);
+            SWSS_LOG_ERROR("Unsupported sonic nhg type %d", sType);
+            return -1;
+        }
+    }
+
+    ret = m_sonic_nhg_table->addEntry(sonicObj);
+    if (ret != 0) {
+        SWSS_LOG_ERROR("Failed to add sonic nhg %d", sonicGatewayNHGID);
+        m_sonic_id_manager.freeID(sType, sonicGatewayNHGID);
+        return ret;
+    }
+
+    SonicGateWayNHGEntry *sonicEntry = m_sonic_nhg_table->getEntry(sType, sonicGatewayNHGID);
+    if (sonicEntry == nullptr) {
+        SWSS_LOG_ERROR("Failed to get sonic nhg %d", sonicGatewayNHGID);
+        m_sonic_id_manager.freeID(sType, sonicGatewayNHGID);
+        return ret;
+    }
+    ret = m_sonic_nhg_table->writeToDB(sonicEntry);
+    if (ret != 0) {
+        SWSS_LOG_ERROR("Failed to write to DB for %d", sonicGatewayNHGID);
+        m_sonic_id_manager.freeID(sType, sonicGatewayNHGID);
+        m_sonic_nhg_table->delEntry(sType, sonicGatewayNHGID);
+        return ret;
+    }
+    entry->setSonicGatewayObjId(sonicGatewayNHGID);
+    return 0;
+}
+
+int NHGMgr::updateSonicNhgObject(RIBNHGEntry *entry, uint32_t previousSonicGatewayObjID) {
+    SonicGateWayNHGObject sonicObj;
+    uint32_t sonicGatewayNHGID;
+    int ret = 0;
+    sonicNhgObjType sType = entry->getSonicObjType();
+    if (previousSonicGatewayObjID == 0) {
+
+        SWSS_LOG_ERROR("Failed to allocate sonic nhg id");
+        return -1;
+    }
+    sonicGatewayNHGID = previousSonicGatewayObjID;
+
+    switch (entry->getSonicObjType()) {
+        case SONIC_NHG_OBJ_TYPE_NHG_SRV6_GATEWAY: {
+            entry->createSRv6GatewayObjFromRIBEntry(sonicObj);
+            sonicObj.id = sonicGatewayNHGID;
+            break;
+        }
+        default: {
+            SWSS_LOG_ERROR("Unsupported sonic nhg type %d", sType);
+            return -1;
+        }
+    }
+
+    ret = m_sonic_nhg_table->updateEntry(sonicObj);
+    if (ret != 0) {
+        SWSS_LOG_ERROR("Failed to add sonic nhg %d", sonicGatewayNHGID);
+        return ret;
+    }
+
+    SonicGateWayNHGEntry *sonicEntry = m_sonic_nhg_table->getEntry(sType, sonicGatewayNHGID);
+    if (sonicEntry == nullptr) {
+        SWSS_LOG_ERROR("Failed to get sonic nhg %d", sonicGatewayNHGID);
+        return ret;
+    }
+    ret = m_sonic_nhg_table->writeToDB(sonicEntry);
+    if (ret != 0) {
+        SWSS_LOG_ERROR("Failed to write to DB for %d", sonicGatewayNHGID);
+        return ret;
+    }
+    entry->setSonicGatewayObjId(sonicGatewayNHGID);
+    return 0;
 }
 
 int NHGMgr::delNHGFull(uint32_t id) {
-	if (!m_rib_nhg_table->isNHGExist(id)) {
-		SWSS_LOG_ERROR("NextHop group id %d not found.", id);
-		return 0;
-	}
 
-	RIBNHGEntry *entry = m_rib_nhg_table->getEntry(id);
-	if (entry == nullptr) {
-		return 0;
-	}
-	uint32_t sonic_id = entry->getSoincObjID();
-	if (m_rib_nhg_table->delEntry(id) != 0) {
-		return -1;
-	}
-	m_sonic_id_manager.freeID(sonic_id);
-	return 0;
+    if (!m_rib_nhg_table->isNHGExist(id)) {
+        SWSS_LOG_ERROR("NextHop group id %d not found.", id);
+        return 0;
+    }
+
+
+    RIBNHGEntry *entry = m_rib_nhg_table->getEntry(id);
+    if (entry == nullptr) {
+        return 0;
+    }
+    uint32_t sonicObjID = entry->getSonicObjID();
+
+    // del the sonic gateway nhg first
+    if (entry->hasSonicObj()) {
+        uint32_t sonicGatewayNHGObjID = entry->getSonicGatewayObjID();
+        m_sonic_nhg_table->delEntry(entry->getSonicObjType(), sonicGatewayNHGObjID);
+        m_sonic_id_manager.freeID(entry->getSonicObjType(), sonicGatewayNHGObjID);
+    }
+
+    if (m_rib_nhg_table->delEntry(id) != 0) {
+        return -1;
+    }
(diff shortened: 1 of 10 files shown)


theasianpianist:
FYI @LARLSN we are reverting the change to add libnexthopgroup as a build dependency for SWSS: https://github.com/sonic-net/sonic-swss/pull/4657 (original change https://github.com/sonic-net/sonic-swss/pull/4394) as it is causing issues w/ CI pipelines.

You will need to include re-include the changes in this PR prior to merging. You will also need to update sonic-buildimage so that libnexthopgroup gets explicitly built as part of the common-libs pipeline and is available to download by the SWSS PR pipeline.

Please note that you do not need to explicitly download/purge/install the libnexthopgroup library in the docker-sonic-vs Dockerfile, since this library is already included in default DVS image.

LARLSN:
> FYI @LARLSN we are reverting the change to add libnexthopgroup as a build dependency for SWSS: #4657 (original change #4394) as it is causing issues w/ CI pipelines.
> 
> You will need to include re-include the changes in this PR prior to merging. You will also need to update sonic-buildimage so that libnexthopgroup gets explicitly built as part of the common-libs pipeline and is available to download by the SWSS PR pipeline.
> 
> Please note that you do not need to explicitly download/purge/install the libnexthopgroup library in the docker-sonic-vs Dockerfile, since this library is already included in default DVS image.

Hi Lawrence, thank you for remind. I've submitted PR in sonic-buildimage repo https://github.com/sonic-net/sonic-buildimage/pull/27807, but i'm not very sure whether it works. Could you help me to review this PR. Any feedback would be super helpful. Thanks a lot!

fpmsyncd/nhgmgr.h
```diff
@@ -0,0 +1,891 @@
...
+#define CHECK_FLAG(V,F)      ((V) & (F))
+
+
+/* Forward declaration for unit test friend access */
+namespace ut_fpmsyncd { struct FpmSyncdNhgMgr; }
+
+namespace swss {
+
+using namespace std;
+
```
venkit-nexthop, line 32: Can you keep the indentation level at 0 under the namespace block?

fpmsyncd/nhgmgr.h
```diff
@@ -0,0 +1,891 @@
...
+        /* PIC_CONTEXT_TABLE type */
+        SONIC_NHG_OBJ_TYPE_NHG_WITH_SRV6_PIC = 1,
+
+        SONIC_NHG_OBJ_TYPE_MAX = 255,
+    };
+
+    /* Key Object used to hash Sonic PIC nexthop object */
+    struct SonicNHGObjectKey {
+        vector<std::pair<uint32_t, uint32_t>> groupMember;
```
venkit-nexthop: Add some comments about what these uint32_t represents. If possible create struct types (with uint32_t members) so as to have type safety and not allow any other integer to be passed.

fpmsyncd/nhgmgr.h
```diff
@@ -0,0 +1,891 @@
...
+        string nexthop;
+        string vpnSid;
+        string segSrc;
+        string ifName;
+        sonicNhgObjType type;
+
+        std::vector<std::pair<uint32_t, uint32_t>> getSortedGroupMember() const {
+            std::vector<std::pair<uint32_t, uint32_t>> sorted = groupMember;
+            sort(sorted.begin(), sorted.end());
```
venkit-nexthop: Could we store this in std::set (which is sorted already) to avoid this constant sorting when operator< or operator== is called?

fpmsyncd/nhgmgr.h
```diff
@@ -0,0 +1,891 @@
...
+        }
+
+        static SonicNHGObjectKey createSonicPICContentObjectKey(SonicPICContentObject obj);
+
+        static int createSonicPICContentObjectKey(RIBNHGEntry *entry, SonicNHGObjectKey &key_out);
+
+        static void createSonicNormalNHGObjectKey(RIBNHGEntry *entry, SonicNHGObjectKey &key_out);
+    };
+
+    /* Sonic PIC object */
```
venkit-nexthop: what is Sonic PIC object. We need detailed comments to understand the code.
Is this covered in the HLD?

fpmsyncd/nhgmgr.h
```diff
@@ -0,0 +1,891 @@
...
+         * currently support three types:
+         * SONIC_NHG_OBJ_TYPE_NHG_NORMAL: normal NHG Objects
+         * SONIC_NHG_OBJ_TYPE_NHG_WITH_SRV6_PIC: SRv6 VPN PIC Contexts
+         */
+        sonicNhgObjType type;
+
+        /*
+         * contain the <ribID, weight> pairs of Sonic PIC object
+         */
+        vector<std::pair<uint32_t, uint32_t>> groupMember;
```
venkit-nexthop: can we create types for ribID and weight and not use uint32_t?

fpmsyncd/nhgmgr.h
```diff
@@ -0,0 +1,891 @@
...
+    };
+
+    /* Type of sonic nexthop object */
+    enum sonicNhgObjType {
+
+        /* NEXTHOP_GROUP_TABLE type */
+        SONIC_NHG_OBJ_TYPE_NHG_NORMAL = 0,
+
+        /* PIC_CONTEXT_TABLE type */
+        SONIC_NHG_OBJ_TYPE_NHG_WITH_SRV6_PIC = 1,
```
venkit-nexthop: Given that you are using the term PIC_CONTEXT everywhere, can you rename this as
`SONIC_NHG_OBJ_TYPE_NHG_WITH_SRV6_PIC_CONTEXT` ?

fpmsyncd/nhgmgr.cpp
```diff
@@ -0,0 +1,1576 @@
...
+        } else {
+            SWSS_LOG_ERROR("sync fv vector failed, unsupported address family %d", m_af);
+            return -1;
+        }
+    }
+
+    FieldValueTuple nh("nexthop", m_nexthop.c_str());
+    m_fvVector.push_back(nh);
+
+    if (!m_ifName.empty()) {
```
venkit-nexthop, line 923: Please handle the case for when nbZmqEnabled is enabled.

fpmsyncd/nhgmgr.h
```diff
@@ -0,0 +1,891 @@
...
+#include <nexthopgroup/nexthopgroupfull.h>
+#include <nexthopgroup/nexthopgroupfull_json.h>
+#include <nexthopgroup/nexthopgroup_debug.h>
+
+#include <string.h>
+#include <algorithm>
+#include <set>
+
+#define NHG_DELIMITER ','
+#define NEXTHOP_GROUP_RECEIVED_FLAG (1 << 10)
```
venkit-nexthop: https://github.com/FRRouting/frr/blob/master/zebra/zebra_nhg.h#L175

seems to define this as `NEXTHOP_GROUP_RECEIVED_FROM_EXTERNAL` 
Is this representing the same thing?

fpmsyncd/nhgmgr.cpp
```diff
@@ -0,0 +1,1576 @@
...
+     * Process the PICContent offload
+     */
+    if (entry->hasSonicPICObj()) {
+        SWSS_LOG_DEBUG("Create sonic PICContent object for NHG %d", nhg.id);
+        ret = createSonicPICObject(entry);
+        if (ret != 0) {
+            m_rib_nhg_table->removeFromDB(entry->getSonicObjID());
+            // sub the ref for the shared sonic nhg object
+            m_rib_nhg_table->subSonicNHGObjectRef(entry->getSonicNHGObjectKey());
+            m_rib_nhg_table->delEntry(nhg.id);
```
venkit-nexthop, line 238: Is the `removeFromDB()` in line 234 necessary? `delEntry` calls the same.

fpmsyncd/nhgmgr.h
```diff
@@ -0,0 +1,891 @@
...
+        friend struct ut_fpmsyncd::FpmSyncdNhgMgr;
+
+    public:
+        RIBNHGTable(RedisPipeline *pipeline, const std::string &tableName, bool isStateTable);
+        ~RIBNHGTable(){
+
+        };
+
+        // add entry into table
+        int addEntry(NextHopGroupFull nhg, uint8_t af);
```
venkit-nexthop: I see that in a lot of places `NextHopGroupFull` is copied by value instead of taking by reference. What is the overhead of calling these functions? Is it possible to convert all these places to take `NextHopGroupFull` by reference?

fpmsyncd/nhgmgr.cpp
```diff
@@ -0,0 +1,1576 @@
...
+    vector<FieldValueTuple> fvVector = entry->getFvVector();
+    if (fvVector.size() == 0) {
+        SWSS_LOG_ERROR("Failed to sync fvVector for %d, empty fvVector", entry->getNHG().id);
+        return -1;
+    }
+    m_nexthop_groupTable.set(std::to_string(entry->getSonicObjID()), fvVector);
+
+    // At present, the specification of NHG is not large.
+    // Use NOTICE to facilitate debugging during converge.
+    SWSS_LOG_NOTICE("writeToDB NEXTHOP_GROUP_TABLE : RIB_id %d sonic_id %d",
```
venkit-nexthop: This will be very verbose with scale. Lets not have it as LOG_NOTICE. It seems to dump all the fields as well.

fpmsyncd/nhgmgr.cpp
```diff
@@ -0,0 +1,1576 @@
...
+        }
+
+        SWSS_LOG_NOTICE("Create sonic NHG for %d, sonic id %d", nhg.id, sonicId);
+    } else if(previousKey != entry->getSonicNHGObjectKey()){
+
+        // For the shared NHG, maybe changed to another existing shared NHG, sub the previous ref
+        if (entry->isSharedSonicNHG()){
+            m_rib_nhg_table->subSonicNHGObjectRef(previousKey);
+        }
+    }
```
venkit-nexthop, line 347: On update, the old SONiC object is only released when the new entry still needs an object (first branch) or when the new entry is shared (this `else if`). The multi→single transition — entry previously owned a non-shared object but now needs none — hits neither branch, so `NEXTHOP_GROUP_TABLE:<old_id>` is never deleted and the id is never freed (stale APP_DB entry + id leak, and `m_sonic_nhg_id_2_rib_nhg_id_map` is left dangling). Same problem when a non-shared object transitions to shared: the `else if` sub-refs `previousKey`, which was never in the shared map.

`previousKey` is the only previous state captured, so the old ownership (was it shared? what id?) is lost by the time we get here. Suggest snapshotting the previous id/shared flag before `updateEntry()` and doing an explicit release-old-then-acquire-new — when the new entry owns no non-shared object, `removeFromDB(prevId)` + `freeID` + erase the id→rib map + `setSonicNHGObjId(0)`.

LARLSN: Thank you for your review. We have had this logic before. However, as I mentioned in the working group, at the zabra level, an NHG from single nexthop-multinexthop, or from non SRv6 NHG(non shared NHG) -> SRv6 NHG(shared NHG), zebra will issue an NHG with new zebra ID to fpmsyncd. So now we don't consider this update change here.

LARLSN:
@venkit-nexthop  hi venkit, thanks for your review, they are very helpful. I've updated the code according to your comments. However, our sanity check can't pass now, because the PR on which we compiled into libnexthopgroup has been revert and needs to be resubmitted. I'll ask you to have a look again after sanity check is ok.

fpmsyncd/nhgmgr.cpp
LARLSN: You are right, updated

fpmsyncd/nhgmgr.cpp
LARLSN: Currently, the ZMQ channel is not enabled in SONiC. We could collaborate together to complete the ZMQ adaptation in the next phase of rib/fib.

fpmsyncd/nhgmgr.h
LARLSN: Finished.

fpmsyncd/nhgmgr.h
LARLSN: Done. Created struct ribID, sonicObjectID and nhgWeight.

fpmsyncd/nhgmgr.h
LARLSN: Done

fpmsyncd/nhgmgr.h
LARLSN: Added the comments in the code. This object is designed to convey the corresponding information of pic object to Sonic PIC table

fpmsyncd/nhgmgr.h
LARLSN: Done, Created struct ribID, sonicObjectID and nhgWeight.

fpmsyncd/nhgmgr.h
LARLSN: Yes, they are same. I've updated NEXTHOP_GROUP_RECEIVED_FLAG to NEXTHOP_GROUP_RECEIVED_FROM_EXTERNAL and added comments in the code

fpmsyncd/nhgmgr.h
LARLSN: Updated. Main functions have changed to use reference.

fpmsyncd/nhgmgr.cpp
LARLSN: Done.

LARLSN:
Summary of changes addressing previous reviews:

| Comment | Resolution |
| --- | --- |
| [https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392412447](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392412447)<br/> Can you keep the indentation level at 0 under the namespace block?<br/> | Fixed. |
| [https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392420054](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392420054)<br/> Add some comments about what these uint32_t represents. If possible create struct types (with uint32_t members) so as to have type safety and not allow any other integer to be passed.<br/> | Updated. Created struct ribID, sonicObjectID, BaseID and nhgWeight structs. |
| [https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392438452](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392438452)<br/> Could we store this in std::set (which is sorted already) to avoid this constant sorting when operator< or operator== is called?<br/> | Fixed. Use set instead of map or unordered map |
| [https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392450548](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392450548)<br/> what is Sonic PIC object. We need detailed comments to understand the code.   Is this covered in the HLD?<br/> | Added the comments in the code. This object is designed to convey the corresponding information of pic object to Sonic PIC table. LLD has details. |
| [https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392461895](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392461895)<br/> can we create types for ribID and weight and not use uint32_t?<br/> | Updated. Created struct ribID, sonicObjectID, BaseID and nhgWeight structs. |
| [https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392520572](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392520572)<br/> Given that you are using the term PIC_CONTEXT everywhere, can you rename this as   `SONIC_NHG_OBJ_TYPE_NHG_WITH_SRV6_PIC_CONTEXT` ?<br/> | Updated. Rename SONIC_NHG_OBJ_TYPE_NHG_WITH_SRV6_PIC to SONIC_NHG_OBJ_TYPE_NHG_WITH_SRV6_PIC_CONTEXT |
| [https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392643936](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392643936)<br/> Please handle the case for when nbZmqEnabled is enabled.<br/> | Kept as-is. Currently, the ZMQ channel is not enabled in SONiC. We could collaborate together to complete the ZMQ adaptation in the next phase of rib/fib. |
| [https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392741116](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392741116)<br/> [https://github.com/FRRouting/frr/blob/master/zebra/zebra_nhg.h#L175](https://github.com/FRRouting/frr/blob/master/zebra/zebra_nhg.h#L175)<br/> seems to define this as NEXTHOP_GROUP_RECEIVED_FROM_EXTERNAL   Is this representing the same thing?<br/> | Updated. They are same. Rename NEXTHOP_GROUP_RECEIVED_FLAG to NEXTHOP_GROUP_RECEIVED_FROM_EXTERNAL and added comments in the code. |
| [https://github.com/sonic-net/sonic-swss/pull/4395/changes#diff-d3a37f1b63e4445ffcc9cc663a1dcde06829ef0a35dd5f7eca4aa53173b3da42R237](https://github.com/sonic-net/sonic-swss/pull/4395/changes#diff-d3a37f1b63e4445ffcc9cc663a1dcde06829ef0a35dd5f7eca4aa53173b3da42R237)<br/> Is the `removeFromDB()` in line 234 necessary? `delEntry` calls the same.<br/> | Fixed. Good catch. I've removed the removeFromDB() in line 234  |
| [https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392784050](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392784050)<br/> I see that in a lot of places `NextHopGroupFull` is copied by value instead of taking by reference. What is the overhead of calling these functions? Is it possible to convert all these places to take `NextHopGroupFull` by reference?<br/> | Fixed. Use reference in main functions in nhg mgr now. |
| [https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392798452](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3392798452)<br/> This will be very verbose with scale. Lets not have it as LOG_NOTICE. It seems to dump all the fields as well.<br/> | Fixed. Use DEBUG level instead of NOTICE. |
| [https://github.com/sonic-net/sonic-swss/pull/4395/changes#r3392865626](https://github.com/sonic-net/sonic-swss/pull/4395/changes#r3392865626)<br/> On update, the old SONiC object is only released when the new entry still needs an object (first branch) or when the new entry is shared (this else if). The multi→single transition — entry previously owned a non-shared object but now needs none — hits neither branch, so NEXTHOP_GROUP_TABLE:<old_id> is never deleted and the id is never freed (stale APP_DB entry + id leak, and m_sonic_nhg_id_2_rib_nhg_id_map is left dangling). Same problem when a non-shared object transitions to shared: the else if sub-refs previousKey, which was never in the shared map.<br/> previousKey is the only previous state captured, so the old ownership (was it shared? what id?) is lost by the time we get here. Suggest snapshotting the previous id/shared flag before updateEntry() and doing an explicit release-old-then-acquire-new — when the new entry owns no non-shared object, removeFromDB(prevId) + freeID + erase the id→rib map + setSonicNHGObjId(0).<br/> | Kept as-is. We have had this logic before. However, as I mentioned in the working group, at the zabra level, an NHG from single nexthop-multinexthop, or from non SRv6 NHG(non shared NHG) -> SRv6 NHG(shared NHG), zebra will issue an NHG with new zebra ID to fpmsyncd. So now we don't consider this update change here. |

@venkit-nexthop  PTAL when you have time. Thanks!

eddieruan-alibaba:
@venkit-nexthop can you help to review it again?

fpmsyncd/nhgmgr.cpp
```diff
@@ -0,0 +1,1578 @@
...
+    for (auto it = nhg.dependents.begin(); it != nhg.dependents.end(); it++) {
+        m_dependents.insert(*it);
+        SWSS_LOG_DEBUG("NextHop id %d add dependents %d.", m_rib_id.id, *it);
+    }
+
+    // check the full list NHG entry and update m_group set
+    for (auto it = nhg.nh_grp_full_list.begin(); it != nhg.nh_grp_full_list.end(); it++) {
+        // validate group member
+        if (!m_table->isNHGExist(it->id)) {
+            SWSS_LOG_ERROR("NextHop id %d in group not found.", it->id);
```
selva-nexthop, line 862: We fail here if a member NHG isn't in the table yet. FPM doesn't really promise parent-after-child ordering — how do we handle messages arriving out of order? Do we buffer and retry, or is zebra guaranteeing the order somehow?

fpmsyncd/nhgmgr.cpp
```diff
@@ -0,0 +1,1578 @@
...
+            SonicNHGObjectKey key;
+            SonicNHGObjectKey::createSonicNormalNHGObjectKey(entry, key);
+            m_rib_nhg_table->insertCreatedSharedNHGObject(key, sonicId);
+        }else{
+            // non shared NHG track by sonic nhg id, map the sonic id to rib nhg id
+            m_rib_nhg_table->insertCreatedNhgObject(sonicId, ribId);
+        }
+
+        SWSS_LOG_NOTICE("Create sonic NHG for %d, sonic id %d", nhg.id, sonicId.id);
+    } else if(previousKey != entry->getSonicNHGObjectKey()){
```
selva-nexthop, line 336: On the shared NHG update path above we always `allocateID()` when we need a new object. That's fine for a brand new key. For this branch — key changed to another shared key that already exists — we only drop the old ref. Can you double-check we always pick up the existing id through `checkNeedCreateSonicNHGObj`, and don't leave the entry hanging onto a stale sonic id?

fpmsyncd/nhgmgr.cpp
```diff
@@ -0,0 +1,1578 @@
...
+
+void RIBNHGTable::addSonicNHGObjectRef(SonicNHGObjectKey key) {
+    m_created_shared_nhg_map[key].refCount++;
+}
+
+void RIBNHGTable::subSonicNHGObjectRef(SonicNHGObjectKey key) {
+    if (m_created_shared_nhg_map.find(key) == m_created_shared_nhg_map.end()){
+        return ;
+    }
+    if (m_created_shared_nhg_map[key].refCount == 0) {
```
selva-nexthop, line 739: If `refCount` is already 0 we log an error and free/delete the object anyway. That feels like it's covering up a real accounting bug. Might be better to keep the entry and fail loudly instead of deleting on underflow.

fpmsyncd/nhgmgr.cpp
```diff
@@ -0,0 +1,1578 @@
...
+        if (sonicId.id == 0) {
+            SWSS_LOG_ERROR("Failed to allocate sonic nhg id");
+            return -1;
+        }
+        entry->setSonicNHGObjId(sonicId);
+        ret = m_rib_nhg_table->writeToDB(entry);
+        if (ret != 0) {
+            m_rib_nhg_table->delEntry(nhg.id);
+            // This sonic nhg id is reused or new allocated, so free it
+            m_sonic_id_manager.freeID(SONIC_NHG_OBJ_TYPE_NHG_NORMAL, sonicId.id);
```
selva-nexthop: Minor consistency thing — some `freeID()` call sites pass a `sonicObjectID`, others pass `.id` like here. It works because of the constructor, but it'd read cleaner if we always passed `sonicObjectID`.

fpmsyncd/nhgmgr.cpp
```diff
@@ -0,0 +1,1578 @@
...
+    }
+    return 0;
+}
+
+/*
+ * process of update NHG full
+ */
+int NHGMgr::updateExistingNHGFull(const NextHopGroupFull& nhg, uint8_t af) {
+    /*
+     * Not support update of NHG fields:
```
selva-nexthop, line 251: Venkit already called out the multi→single object leak and the ZMQ path. From the reply, zebra always issues a new NHG id for those transitions — worth writing that assumption down in the LLD (near this "not supported on update" note) so we don't accidentally regress it later.

fpmsyncd/nhgmgr.cpp
venkit-nexthop: Microsoft and Nexthop are testing with ZMQ enabled and we have fixed a few bugs in this area. This is present in the code. So I think any new code should accomodate it.

eddieruan-alibaba: My team is unfamiliar with this approach, and our current deployment does not use ZMQ for routing. Additionally, the Cisco team has submitted a new proposal for the centralized VoQ chassis to address excessive variations. 

Moving forward, we can phase in ZMQ in a future commit. Once the foundational RIB/FIB is in place, Songnan will collaborate with the Nexthop team on this ZMQ enhancement.

fpmsyncd/nhgmgr.cpp
LARLSN: Done. Documented in the comment block at the top of updateExistingNHGFull(), and also updated the LLD accordingly — added a "Note on update assumptions" section under the update flow in doc/ribfib/nhgmgr_lld.md.

fpmsyncd/nhgmgr.cpp
LARLSN: Good point, agreed. Changed the underflow path to keep the entry and the sonic object as-is and only log an error, so an accounting bug fails loudly instead of being masked by deleting the object.

LARLSN:
@selva-nexthop hi selva, Thanks a lot for the detailed review — really appreciate you taking the time to go through this carefully. I've addressed all of your comments in the latest revision:
| # | Comment | Reply |
|---|---------|------------|
| 1 | [Member NHG not in table fails; how are out-of-order FPM messages handled?](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3681032605) | Zebra guarantees the ordering here: in zebra's dataplane, an NHG's dependencies (singleton nexthops) are always installed and pushed through FPM before the group that references them, and fpmsyncd processes FPM messages in order. So a parent NHG can never legitimately arrive before its members. If it does happen, it indicates a zebra-side bug, and failing loudly here (rather than buffering) is the intended behavior. |
| 2 | [Shared key changed to another existing shared key — stale sonic id?](https://github.com/sonic-net/sonic swss/pull/4395#discussion_r3681043692) | Verified: `updateEntry()` internally calls `setEntry()` → `checkNeedCreateSonicNHGObj()`, which looks up the existing shared object by the new key, sets the entry's sonic id to the existing one and adds a reference. So by the time we reach this branch, the entry already holds the correct id; we only need to drop the reference on the previous key. Added a comment in the code to clarify this. |
| 3 | [refCount underflow still frees/deletes the object, masking accounting bugs](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3681060635) | Code changed: fail loudly, keep object | Good point, agreed. Changed the underflow path to keep the entry and the sonic object as-is and only log an error, so an accounting bug fails loudly instead of being masked by deleting the object. |
| 4 | [Inconsistent freeID() arguments (`.id` vs `sonicObjectID`)](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3681093460) | Done, all `freeID()` call sites now pass `sonicObjectID` consistently. |
| 5 | [Write the zebra new-NHG-id assumption down in the LLD](https://github.com/sonic-net/sonic-swss/pull/4395#discussion_r3681121948) | Done. Documented in the comment block at the top of `updateExistingNHGFull()`, and also updated the LLD accordingly — added a "Note on update assumptions" section under the update flow in`doc/ribfib/nhgmgr_lld.md`. |

When you have a moment, could you take another look?

PR #4400, opened 2026-03-26
[orchagent] Async swss.rec task recording

vpandian-nokia:
**What I did**

This change moves `swss.rec` work off the consumer-ingress path in `orchagent`, which is part of the route programming path.

APP DB tasks are still recorded at the same consumer-ingress point, but the raw `KeyOpFieldsValuesTuple` is now queued together with its ingress timestamp and serialized later by a background worker. The async queue/worker implementation now sits under the recorder infrastructure rather than as a local implementation in `orch.cpp`.

Scope of the change:
- Preserve the existing `swss.rec` line format.
- Preserve the logical recording point at consumer ingress.
- Preserve per-entry consumer-ingress timestamps.
- Move the async queue/worker and `swss.rec` serialization under recorder infrastructure.
- Add fatal-signal diagnostics for `pending`, `high_watermark`, `enqueued`, and `drained`.
- Add graceful `SIGTERM`/`SIGINT` drain support with a single `INFO` shutdown summary.
- Keep `retry.rec` behavior unchanged.
- Do not change route programming or APP DB task semantics.

**Why I did it**

The synchronous `swss.rec` path performs tuple formatting and file writes during APP task ingestion in `orchagent`. During large route bursts, that adds avoidable work to the route programming path.

This change shifts record serialization and file I/O to a dedicated worker thread so that recorder overhead is reduced without changing the observed `swss.rec` output.

**How I verified it**

- Added unit test `tests/quoted_ut.cpp::preservesProvidedTimestamp`.
- Added unit test `tests/mock_tests/consumer_ut.cpp::AsyncSwssRecorderWritesBatchRecords`.
- Added unit test `tests/quoted_ut.cpp::swssrec.asyncUpdatesStatsAndWritesRecord`.
- Added unit test `tests/quoted_ut.cpp::swssrec.syncFallbackWhenAsyncDisabled`.
- Added unit test `tests/quoted_ut.cpp::swssrec.signalSafeStatsDumpFormatsOutput`.
- Confirmed that timestamps are captured at consumer ingress and formatted later by the recorder worker.
- Validated on DUT that `swss.rec` timestamps reflect per-entry ingress timing and the line format remains unchanged.
- Validated fatal-signal diagnostics and graceful shutdown drain behavior.
- Compared route convergence with recorder enabled against the previous synchronous path.

**Details if related**

- This change only affects how `swss.rec` entries are produced.
- `retry.rec` remains synchronous and unchanged.
- A single worker drains the shared queue and emits records in queue order.
- Fatal-signal handling is limited to signal-safe recorder diagnostics rather than queue flush.
- Pending records are drained during orderly shutdown before worker exit.
- Companion config-knob change: [`sonic-net/sonic-buildimage#26726`](https://github.com/sonic-net/sonic-buildimage/pull/26726)


vpandian-nokia:
To clarify the intended safety of this PR:

This change does not alter route programming behavior, APP DB task semantics, or retry handling. It only moves `swss.rec` tuple serialization and file writes out of `ConsumerBase::addToSync()`, which is part of the route programming path.

The logical recording point remains consumer ingress, the `swss.rec` line format is preserved, and each record still carries its original per-entry ingress timestamp captured before enqueue. A single worker drains the shared queue and emits records in queue order, and pending records are drained during shutdown.

So, the scope here is intentionally narrow: reduce recorder overhead on the route programming path while preserving existing recording behavior.

deepak-singhal0408:
@venkit-nexthop FYI.

venkit-nexthop:
> @venkit-nexthop FYI.

Thanks for keeping me in the loop. 

I dont know how tolerant we are to missed record entries. But one thing we could do is to keep a circular buffer and produce stuff from the main thread and read from the worker thread and record. The downside is that the circular buffer can wrap around if the worker thread is slow.

venkit-nexthop (commented):
Thanks. This looks good to me.
There is also a sairedis.rec. Maybe doing something similar for that would help as well?

vpandian-nokia:
@venkit-nexthop: Thanks for the review.

The circular-buffer approach makes sense if we are willing to trade strict recorder completeness for lower overhead under backpressure. For this PR, I intentionally kept the implementation lossless and ordered, so entries are not overwritten if the worker falls behind. A ring/circular buffer could be explored later, but that would need an explicit decision on whether missed `swss.rec` entries are acceptable.

I also evaluated whether a similar approach for `sairedis.rec` would be worthwhile. In our testing, we did not observe significant performance improvement there, so I do not plan to extend this change in that direction for now. I would prefer to keep this PR focused on the `swss.rec` path where the gain was measurable.

venkit-nexthop:
> @venkit-nexthop: Thanks for the review.
> 
> The circular-buffer approach makes sense if we are willing to trade strict recorder completeness for lower overhead under backpressure. For this PR, I intentionally kept the implementation lossless and ordered, so entries are not overwritten if the worker falls behind. A ring/circular buffer could be explored later, but that would need an explicit decision on whether missed `swss.rec` entries are acceptable.
> 
> I also evaluated whether a similar approach for `sairedis.rec` would be worthwhile. In our testing, we did not observe significant performance improvement there, so I do not plan to extend this change in that direction for now. I would prefer to keep this PR focused on the `swss.rec` path where the gain was measurable.

Thanks, makes sense @vpandian-nokia,

deepak-singhal0408:
@vpandian-nokia ,  what will be the impact due to this change on process Crash(SIGSEGV, SIGABRT etc). we might lose some critical data that may otherwise be useful for debugging.. right?

venkit-nexthop:
> @vpandian-nokia , what will be the impact due to this change on process Crash(SIGSEGV, SIGABRT etc). we might lose some critical data that may otherwise be useful for debugging.. right?

@deepak-singhal0408 We had documented the same issue for the Northbound ZMQ change as well 
Refer here => https://github.com/nexthop-ai/SONiC/blob/f0c993dcb613a9f654b8df9c5013f0df5757ca08/doc/route_programming_optimization_zmq/HLD%20Northbound%20ZMQ.md#the-restart-issue

In general this is a problem for all the asynchronous state update changes, including in Southbound ZMQ as well.
Maybe we should come up with a common solution that would equally apply for all these cases.

vpandian-nokia:
@deepak-singhal0408:
Good point. The pending-record drain applies only to an orderly shutdown, not to crash scenarios such as `SIGSEGV` or `SIGABRT`.

So, in an abnormal termination, it is possible to lose the most recent `swss.rec` entries that are still queued in memory and have not yet been written by the worker. Compared with the previous synchronous path, that is the tradeoff with this async approach.

I agree with @venkit-nexthop that this is a broader consideration for asynchronous state-update paths in SONiC, and not unique to this specific change.

deepak-singhal0408:
> > @vpandian-nokia , what will be the impact due to this change on process Crash(SIGSEGV, SIGABRT etc). we might lose some critical data that may otherwise be useful for debugging.. right?
> 
> @deepak-singhal0408 We had documented the same issue for the Northbound ZMQ change as well Refer here => https://github.com/nexthop-ai/SONiC/blob/f0c993dcb613a9f654b8df9c5013f0df5757ca08/doc/route_programming_optimization_zmq/HLD%20Northbound%20ZMQ.md#the-restart-issue
> 
> In general this is a problem for all the asynchronous state update changes, including in Southbound ZMQ as well. Maybe we should come up with a common solution that would equally apply for all these cases.

@venkit-nexthop, thanks for pointing to the doc.. I view the problem identified in the  doc differently, and classify in the bucket of other process crashes(orchagent/syncd) that my HLD doc(https://github.com/deepak-singhal0408/SONiC/blob/route-download-performance-strategies/doc/route-performance/strategies/large-scale-datacenter-clos-fabric.md#zmq-specific-risks) covers which requires eventual consistency (if backups are available).

For 202605, I am not sure if we can achieve full warm reboot support with this feature (that could be worked in parallel as a separate task).. For this scenario, and the other two cases listed in my doc(orchagent crash/syncd crash), looks like simplest way to solve it is build dependency between bgp/swss and syncd containers. we already have partially dependencies built in the existing code, this can be enhanced to solve this problem.  
Main objective for the network is to achieve fast convergence, which will be dependent on how quickly we notify on our neighbors that we cant handle the traffic (on this, need to identify gaps and fix it if any).  


> @deepak-singhal0408: Good point. The pending-record drain applies only to an orderly shutdown, not to crash scenarios such as `SIGSEGV` or `SIGABRT`.
> 
> So, in an abnormal termination, it is possible to lose the most recent `swss.rec` entries that are still queued in memory and have not yet been written by the worker. Compared with the previous synchronous path, that is the tradeoff with this async approach.
> 
> I agree with @venkit-nexthop that this is a broader consideration for asynchronous state-update paths in SONiC, and not unique to this specific change.

@vpandian-nokia, what is the %age gain with this? I am afraid, this solution at the moment looks incomplete  (comes with side effect: debugging regression from what we have currently). Can we try reproducing this on local setup, and explore potential solutions?

vpandian-nokia:
@deepak-singhal0408: agreed that this should be evaluated with data before we conclude on the tradeoff.

To enable a clean A/B comparison on the same SONiC image, I am adding a new orchagent CLI option -a to enable ASYNC_SWSS_REC, while keeping the current synchronous swss.rec path as the default.

I am now measuring the % gain on DUT and, in parallel, reproducing abnormal termination locally to quantify any swss.rec tail loss and assess whether mitigation is needed.

I will update this thread once I have both sets of results.

venkit-nexthop:
> @venkit-nexthop, thanks for pointing to the doc.. I view the problem identified in the doc differently, and classify in the bucket of other process crashes(orchagent/syncd) that my HLD doc(https://github.com/deepak-singhal0408/SONiC/blob/route-download-performance-strategies/doc/route-performance/strategies/large-scale-datacenter-clos-fabric.md#zmq-specific-risks) covers which requires eventual consistency (if backups are available).
> 
> For 202605, I am not sure if we can achieve full warm reboot support with this feature (that could be worked in parallel as a separate task).. For this scenario, and the other two cases listed in my doc(orchagent crash/syncd crash), looks like simplest way to solve it is build dependency between bgp/swss and syncd containers. we already have partially dependencies built in the existing code, this can be enhanced to solve this problem. Main objective for the network is to achieve fast convergence, which will be dependent on how quickly we notify on our neighbors that we cant handle the traffic (on this, need to identify gaps and fix it if any).

@deepak-singhal0408 just to be clear, you are recommending that we would restart all the associated containers and also reset the ASIC? If not the scenario we have described in our docs will not result in eventual consistency.
It can lead to a scenario where the ASIC may have some extra routes, while the rest of the state will not have the route.

orchagent/orch.cpp
```diff
@@ -15,6 +16,134 @@
...
+{
+
+struct AsyncSwssRecordEntry
+{
+    struct timeval received_time;
+    std::string prefix;
+    KeyOpFieldsValuesTuple tuple;
+};
+
+class AsyncSwssRecorder
```
vivekkarri: AsyncRecorder is a generic infra, can it be moved into Recorder.cpp

We can then derive AsyncSwssRecorder from it.

orchagent/orch.cpp
```diff
@@ -15,6 +16,134 @@
...
+            s += "|" + fvField(fv) + ":" + fvValue(fv);
+        }
+        return s;
+    }
+
+    void drain()
+    {
+        while (true)
+        {
+            std::deque<AsyncSwssRecordEntry> pending;
```
vivekkarri: Can this logic be moved into Recorder itself? Orch class should not be bothered with all these details.

deepak-singhal0408:
> > @venkit-nexthop, thanks for pointing to the doc.. I view the problem identified in the doc differently, and classify in the bucket of other process crashes(orchagent/syncd) that my HLD doc(https://github.com/deepak-singhal0408/SONiC/blob/route-download-performance-strategies/doc/route-performance/strategies/large-scale-datacenter-clos-fabric.md#zmq-specific-risks) covers which requires eventual consistency (if backups are available).
> > For 202605, I am not sure if we can achieve full warm reboot support with this feature (that could be worked in parallel as a separate task).. For this scenario, and the other two cases listed in my doc(orchagent crash/syncd crash), looks like simplest way to solve it is build dependency between bgp/swss and syncd containers. we already have partially dependencies built in the existing code, this can be enhanced to solve this problem. Main objective for the network is to achieve fast convergence, which will be dependent on how quickly we notify on our neighbors that we cant handle the traffic (on this, need to identify gaps and fix it if any).
> 
> @deepak-singhal0408 just to be clear, you are recommending that we would restart all the associated containers and also reset the ASIC? If not the scenario we have described in our docs will not result in eventual consistency. It can lead to a scenario where the ASIC may have some extra routes, while the rest of the state will not have the route.
[DS]: yes, thats right. restart the associated containers and reset asic too.. fpmsyncd/orchagent/syncd processes are all critical processes in their respective containers.. and hence that means, the parent container would go down (if bgp container goes down, connection gets reset, if swss goes down->bgp and syncd also go down) etc.. So anyways, with the current design, we have some dependencies built.., So, In order to avoid inconsistencies in the network (due to edge cases etc), this seems like a simpler and better approach for 202605..

orchagent/orch.cpp
vpandian-nokia: I agree this can be refactored further, but I would prefer to keep this PR focused on the `swss.rec` optimization.

This change intentionally defers not only the file write, but also the `swss.rec` tuple serialization itself to the worker thread by enqueueing the raw tuple plus ingress timestamp. Since `Recorder` today mostly handles already-serialized strings, moving this fully into `Recorder` would require a broader interface refactor.

To keep this PR narrow and easier to review, I kept the async logic local to the `swss.rec` path in `orch.cpp`.

orchagent/orch.cpp
vpandian-nokia: The reason `orch.cpp` is involved today is that the optimization is anchored at the consumer-ingress recording point, where the raw `KeyOpFieldsValuesTuple` and its ingress timestamp are still available before serialization.

Since this enhancement shifts both serialization and file write work to the worker thread, the logic is still tightly coupled to that `swss.rec` ingress path rather than the current `Recorder` interface.

orchagent/orch.cpp
vivekkarri: Even then I don't think we need to put all the logic here.  From orchagent PoV, there only should be a couple of API's for recorder. 

Something on the lines of 

```
enableAsync(), 
record()
recordAsync()
```

enableAsync() can start a thread to handle writing, other 2 are obvious  recordAsync can take the current timestamp. Internally you can have any logic to Queue/deque and have separate thread or however you want.  Feel free to update Recorder class/library however you want to fit your design. 

The whole point of moving all this into a separate lib is defeated if we introduce all these changes back.

orchagent/orch.cpp
```diff
@@ -15,6 +16,134 @@
 
 using namespace swss;
 
+namespace
+{
+
+struct AsyncSwssRecordEntry
```
vivekkarri: You could serialize all this into a string and add it to the queue, making the Recorder API simple. 

Main overhead i think is writing to a file, which is parallelized anyway

vivekkarri:
> @deepak-singhal0408: Good point. The pending-record drain applies only to an orderly shutdown, not to crash scenarios such as `SIGSEGV` or `SIGABRT`.
> 
> So, in an abnormal termination, it is possible to lose the most recent `swss.rec` entries that are still queued in memory and have not yet been written by the worker. Compared with the previous synchronous path, that is the tradeoff with this async approach.
> 
> I agree with @venkit-nexthop that this is a broader consideration for asynchronous state-update paths in SONiC, and not unique to this specific change.

signals other than SIGKILL should be handled gracefully. Ideally, Writer thread should upon getting this signal should dump everything that is in the queue to a file and exit. 

For SIGKILL, and process crash i don't think we can do much

orchagent/orch.cpp
vpandian-nokia: We tried that earlier in this work. In our DUT measurements, moving only the file write off the main path while still doing `swss.rec` string serialization before enqueue did not help, and in some cases performance was worse.

The measurable gain came only when we also deferred tuple serialization/formatting to the worker by enqueueing the raw `KeyOpFieldsValuesTuple` together with its ingress timestamp. That is why the current version does not serialize to a string before enqueue.

orchagent/orch.cpp
vpandian-nokia: I agree that, from `orchagent`'s point of view, a small `Recorder` API would be cleaner.

The difficulty here is that the measured gain depends on handing off the raw `KeyOpFieldsValuesTuple` plus ingress timestamp and deferring the `swss.rec` serialization itself to the worker. So, this is not just a matter of moving queue/thread logic behind `recordAsync()`: it would also require Recorder to own the `swss.rec`-specific tuple serialization semantics rather than only handling already-serialized strings.

I wanted to keep this PR focused on the measured optimization path. Broadening `Recorder` to own that richer async contract would be a larger refactor than I was aiming to mix into this change.

vpandian-nokia:
> > @deepak-singhal0408: Good point. The pending-record drain applies only to an orderly shutdown, not to crash scenarios such as `SIGSEGV` or `SIGABRT`.
> > So, in an abnormal termination, it is possible to lose the most recent `swss.rec` entries that are still queued in memory and have not yet been written by the worker. Compared with the previous synchronous path, that is the tradeoff with this async approach.
> > I agree with @venkit-nexthop that this is a broader consideration for asynchronous state-update paths in SONiC, and not unique to this specific change.
> 
> signals other than SIGKILL should be handled gracefully. Ideally, Writer thread should upon getting this signal should dump everything that is in the queue to a file and exit.
> 
> For SIGKILL, and process crash i don't think we can do much

In the current implementation, pending-record drain applies only to normal process teardown where the recorder destructor runs and joins the worker after draining the queue. It does not currently cover generic signal-driven termination, and it does not apply to abnormal crash signals such as `SIGSEGV` or `SIGABRT`.

If we extend this further, graceful drain is realistic for signals like `SIGTERM` and `SIGINT`, and possibly `SIGQUIT`, where the process is still in a valid state. `SIGHUP` is already used in `orchagent` for log rotation and should likely remain separate. For crash signals such as `SIGSEGV`, `SIGABRT`, `SIGBUS`, `SIGILL`, and `SIGFPE`, I would not rely on flushing the queue from the signal handler; only minimal crash diagnostics are safe there. `SIGKILL` and `SIGSTOP` cannot be handled at all.

orchagent/orch.cpp
vivekrnv: I see, Thanks. 

This should be fine then

orchagent/orch.cpp
vivekrnv: It is called AsyncSwssRecorder, I think it can own the logic of tuple serialization semantics. 

I won't call it a refactor, this PR introduces async contract. This change is in the scope of it

vganesan-nokia:
@deepak-singhal0408, would you please mark this PR for Sonic Routing Dashboard?

deepak-singhal0408:
> @deepak-singhal0408, would you please mark this PR for Sonic Routing Dashboard?

its already added to "SONiC Routing Dashboard project".. could you check "Open Code PRs" tab(I could see its present there)?

vpandian-nokia:
> > > @vpandian-nokia , what will be the impact due to this change on process Crash(SIGSEGV, SIGABRT etc). we might lose some critical data that may otherwise be useful for debugging.. right?
> > 
> > 
> > @deepak-singhal0408 We had documented the same issue for the Northbound ZMQ change as well Refer here => https://github.com/nexthop-ai/SONiC/blob/f0c993dcb613a9f654b8df9c5013f0df5757ca08/doc/route_programming_optimization_zmq/HLD%20Northbound%20ZMQ.md#the-restart-issue
> > In general this is a problem for all the asynchronous state update changes, including in Southbound ZMQ as well. Maybe we should come up with a common solution that would equally apply for all these cases.
> 
> @venkit-nexthop, thanks for pointing to the doc.. I view the problem identified in the doc differently, and classify in the bucket of other process crashes(orchagent/syncd) that my HLD doc(https://github.com/deepak-singhal0408/SONiC/blob/route-download-performance-strategies/doc/route-performance/strategies/large-scale-datacenter-clos-fabric.md#zmq-specific-risks) covers which requires eventual consistency (if backups are available).
> 
> For 202605, I am not sure if we can achieve full warm reboot support with this feature (that could be worked in parallel as a separate task).. For this scenario, and the other two cases listed in my doc(orchagent crash/syncd crash), looks like simplest way to solve it is build dependency between bgp/swss and syncd containers. we already have partially dependencies built in the existing code, this can be enhanced to solve this problem. Main objective for the network is to achieve fast convergence, which will be dependent on how quickly we notify on our neighbors that we cant handle the traffic (on this, need to identify gaps and fix it if any).
> 
> > @deepak-singhal0408: Good point. The pending-record drain applies only to an orderly shutdown, not to crash scenarios such as `SIGSEGV` or `SIGABRT`.
> > So, in an abnormal termination, it is possible to lose the most recent `swss.rec` entries that are still queued in memory and have not yet been written by the worker. Compared with the previous synchronous path, that is the tradeoff with this async approach.
> > I agree with @venkit-nexthop that this is a broader consideration for asynchronous state-update paths in SONiC, and not unique to this specific change.
> 
> @vpandian-nokia, what is the %age gain with this? I am afraid, this solution at the moment looks incomplete (comes with side effect: debugging regression from what we have currently). Can we try reproducing this on local setup, and explore potential solutions?

@deepak-singhal0408 : 
The measured gain with this change is in the range of about 12% to 16% on a DUT with AMD EPYC 3251 8-Core 2.5GHz and 16 GB DRAM, using a route-download workload of 250K IPv4 + 250K IPv6.

The scenario was reproduced locally by injecting fatal signals into orchagent during active route download, with additional async-recorder instrumentation enabled to capture pending, high_watermark, enqueued, and drained counts at crash time.

So far, 15 runs have been collected across IPv4 250K, IPv6 250K, and mixed IPv4+IPv6 250K cases, using SIGABRT, SIGSEGV, SIGBUS, SIGILL, and SIGFPE, and with different batch sizes (1024, 1000, 5000, 10000).

In 14 out of 15 runs, the async backlog at crash time was 0. In the one run where a non-zero backlog was observed, the pending count was 758, with enqueued=684879 and drained=684121, i.e., the gap matched the pending backlog exactly. The recorder high_watermark also consistently tracked the configured batch size, which suggests the backlog is typically short-lived and roughly bounded to one consumer batch in this workload rather than accumulating over time.

The fatal-signal timestamp from /var/log/syslog was also compared with the last persisted line in /var/log/swss/swss.rec. In these runs, the gap was typically small, ranging from about 4 ms to about 581 ms, with the median around 100 ms. That suggests the worker is usually keeping up closely with the workload, and that the crash-time loss window is real but narrow and timing-dependent in this setup.

Regarding potential solutions: graceful drain looks reasonable for orderly termination signals such as SIGTERM / SIGINT, but flushing the queue from fatal crash-signal handlers is not something to rely on, because that path is not robust once the process may already be in a corrupted state. A safer mitigation for crash scenarios is to improve crash-time diagnostics and visibility. As one concrete step, crash-time async-recorder stats (pending, high_watermark, enqueued, drained) were added to the fatal-signal path by capturing those counters in the recorder and emitting them from the fatal-signal handler using a signal-safe write() to stderr. In SONiC, that output is visible via syslog, which makes it possible to correlate the crash timestamp with both the recorder backlog at the time of crash and the last persisted line in /var/log/swss/swss.rec.

orchagent/orch.cpp
vpandian-nokia: Fair point. Keeping `orchagent` on a smaller recorder-facing API is a reasonable direction here.

If there is strong preference for that layering, the implementation can be reworked so `AsyncSwssRecorder` sits under `Recorder` and `orchagent` only uses a narrow recorder API, while preserving the same behavior of enqueueing the raw tuple plus ingress timestamp and deferring serialization to the worker.

deepak-singhal0408:
> e. A safer mitigation for crash scenarios is to improve crash-time diagnostics and visibility. As one concrete step, crash-time async-recorder stats (pending, high_watermark, enqueued, drained) were added to the fatal-signal path by capturing those counters in the recorder and emitting them from the fatal-signal handler using a signal-safe write() to stderr. In SONiC, that output is visible via syslog, which makes it possible to correlate the crash timestamp with both the recorder backlog at the time of crash and the last persisted line in /var/log/swss/swss.rec.

thanks @vpandian-nokia for the detailed analysis.. Looks like probability of losing swss records is quite low. And the perf gain is significant.. Moreover, on typical crash scenarios, we will have core dump, orchagent logs(syslog), APPL_DB, ASIC_DB etc, to correlate/debug..  At this point, I am thinking about having a config_knob(something like this: DEVICE_METADATA|localhost|async_swss_rec) to make it operationally manageable. DEFAULT could be OFF. We could turn it on in deployments where perf is indeed important.
As a next step on this PR, Please help address @vivekrnv's concerns on the code refactoring/alignment and lets have a config knob for this enhancement. With these in place, its a good to go from my side.. Thanks,
cc: @prsunny @prabhataravind if any concerns.

orchagent/orch.cpp
vivekrnv: yeah, that'd be great. Thanks

vpandian-nokia: Updated in the latest commit.

The async `swss.rec` queue/worker path has now been moved under the recorder infrastructure in `lib/recorder.h` and `lib/recorder.cpp`, including the raw tuple handoff, worker-side serialization, shutdown handling, and debug stats. `orch.cpp` is now reduced to thin dispatch through the `SwSSRec` APIs rather than carrying the queue/thread implementation locally.

orchagent/orch.cpp
vpandian-nokia: Updated in the latest commit.

The async `swss.rec` queue/worker implementation has now been moved under the recorder infrastructure in `lib/recorder.h` and `lib/recorder.cpp`, including the raw tuple handoff, worker-side serialization, shutdown handling, and debug stats. `orch.cpp` is now reduced to thin dispatch through the `SwSSRec` APIs rather than carrying those queue/thread details locally.

vpandian-nokia:
> > e. A safer mitigation for crash scenarios is to improve crash-time diagnostics and visibility. As one concrete step, crash-time async-recorder stats (pending, high_watermark, enqueued, drained) were added to the fatal-signal path by capturing those counters in the recorder and emitting them from the fatal-signal handler using a signal-safe write() to stderr. In SONiC, that output is visible via syslog, which makes it possible to correlate the crash timestamp with both the recorder backlog at the time of crash and the last persisted line in /var/log/swss/swss.rec.
> 
> thanks @vpandian-nokia for the detailed analysis.. Looks like probability of losing swss records is quite low. And the perf gain is significant.. Moreover, on typical crash scenarios, we will have core dump, orchagent logs(syslog), APPL_DB, ASIC_DB etc, to correlate/debug.. At this point, I am thinking about having a config_knob(something like this: DEVICE_METADATA|localhost|async_swss_rec) to make it operationally manageable. DEFAULT could be OFF. We could turn it on in deployments where perf is indeed important. As a next step on this PR, Please help address @vivekrnv's concerns on the code refactoring/alignment and lets have a config knob for this enhancement. With these in place, its a good to go from my side.. Thanks, cc: @prsunny @prabhataravind if any concerns.

@deepak-singhal0408:
Thanks, both follow-ups are now addressed.

The latest `sonic-swss` update moves the async `swss.rec` queue/worker path under the recorder infrastructure (`lib/recorder.h` / `lib/recorder.cpp`) and keeps `orch.cpp` limited to thin recorder dispatch.

For operational manageability, the companion [`sonic-net/sonic-buildimage#26726`](https://github.com/sonic-net/sonic-buildimage/pull/26726) adds `DEVICE_METADATA|localhost|async_swss_rec` support in `orchagent.sh`, with the knob modeled in `sonic-device_metadata.yang`. The intended behavior is default off unless explicitly configured to `enabled`.

deepak-singhal0408:
@vpandian-nokia could you add more UT coverage to satisfy threshold?

vpandian-nokia:
> @vpandian-nokia could you add more UT coverage to satisfy threshold?

@deepak-singhal0408 Certainly. I will add a few focused UTs around the new recorder paths to cover the uncovered branches and bring the PR coverage back above the required threshold.

lib/recorder.cpp
```diff
@@ -40,6 +52,249 @@ SwSSRec::SwSSRec()
...
+
+    struct timeval received_time;
+    gettimeofday(&received_time, nullptr);
+
+    {
+        std::lock_guard<std::mutex> lock(m_mutex);
+        m_queue.push_back({received_time, prefix, tuple});
+    }
+
+    onEnqueue(1);
```
deepak-singhal0408: Suggestion: move onEnqueue() inside the lock scope

onEnqueue() is called after the mutex is released in both recordTupleAsync()  and recordTuplesAsync() 
This creates a small window where the worker thread can drain the entry and call onDrain() before onEnqueue() runs — causing m_pendingCount to momentarily underflow (uint64_t wraps to ~UINT64_MAX).

Recording correctness is not affected — this only impacts the crash diagnostics output from dumpAsyncSignalSafeStats(). If fatal_signal_handler fires during that window, pending shows a bogus value.

the lock Extension cost should be negligible for this operation.

vpandian-nokia: Good catch. Updated the code so `onEnqueue()` now runs within the same lock scope as queue insertion in both `recordTupleAsync()` and `recordTuplesAsync()`, which closes the transient pending-count race for the crash-diagnostic stats.

vpandian-nokia:
> > @vpandian-nokia could you add more UT coverage to satisfy threshold?
> 
> @deepak-singhal0408 Certainly. I will add a few focused UTs around the new recorder paths to cover the uncovered branches and bring the PR coverage back above the required threshold.

@deepak-singhal0408 Added a few focused UTs in the latest update around the new recorder paths to cover the uncovered branches and help bring the PR coverage back above the required threshold.

deepak-singhal0408:
@vpandian-nokia , please resolve merge conflicts,

vpandian-nokia:
> @vpandian-nokia , please resolve merge conflicts,

@deepak-singhal0408 Merged the latest master into this branch and resolved the resulting conflict in tests/mock_tests/consumer_ut.cpp; the PR is mergeable again and CI is running on the updated branch.

deepak-singhal0408 (dismissed):
LGTM..

request @prsunny and @prabhataravind to also help review/sign-off. thanks,

lib/recorder.cpp
```diff
@@ -40,6 +52,249 @@ SwSSRec::SwSSRec()
...
+    }
+
+    ensureAsyncWorker();
+
+    {
+        std::lock_guard<std::mutex> lock(m_mutex);
+        for (const auto& entry : entries)
+        {
+            struct timeval received_time;
+            gettimeofday(&received_time, nullptr);
```
prabhataravind, line 121: The mutex is held for the entire batch loop. This can keep the lock held across potentially hundreds of string copies and heap allocations, blocking the worker thread from draining..

lib/recorder.h
```diff
@@ -53,9 +63,56 @@ class RetryRec : public RecWriter {
...
+    bool m_asyncEnabled = false;
+    bool m_shutdown = false;
+    bool m_workerStarted = false;
+    mutable std::atomic<uint64_t> m_pendingCount{0};
+    mutable std::atomic<uint64_t> m_highWatermark{0};
+    mutable std::atomic<uint64_t> m_enqueuedTotal{0};
+    mutable std::atomic<uint64_t> m_drainedTotal{0};
+    std::mutex m_mutex;
+    std::condition_variable m_signal;
+    std::deque<AsyncSwssRecordEntry> m_queue;
```
prabhataravind, line 114: This unbounded queue can potentially grow very large quickly if the worker can't keep up. We should at least have a logic to log a warning when say m_pendingCount  becomes higher than a threshold or something like that..

orchagent/orchdaemon.cpp
```diff
@@ -931,8 +934,20 @@ void OrchDaemon::start(long heartBeatInterval)
...
 
+        if (gOrchShutdownRequested != 0)
+        {
+            SWSS_LOG_NOTICE("Received signal %d, shutting down orchagent gracefully", gOrchShutdownRequested);
+            break;
+        }
+
         ret = m_select->select(&s, SELECT_TIMEOUT);
 
+        if (gOrchShutdownRequested != 0)
```
prabhataravind, line 945: The pre-select() check and the EINTR check are sufficient I think. the post-select() check is redundant.

tests/mock_tests/consumer_ut.cpp
```diff
@@ -169,6 +177,33 @@ namespace consumer_test
             }
             ASSERT_EQ(sync.size(), exp_sz-1);
         }
+
+        vector<string> waitForRecordedLines(const string& path, const string& marker, size_t expected)
```
prabhataravind, line 181: this seems to be duplicated.. Can we reuse using a shared util header?

tests/mock_tests/consumer_ut.cpp
```diff
@@ -449,6 +484,47 @@ namespace consumer_test
         ASSERT_EQ(test_orch.m_notification_count, consumer_pops_batch_size*2);
     }
 
+    TEST_F(ConsumerTest, AsyncSwssRecorderWritesBatchRecords)
+    {
+        char dir_template[] = "/tmp/swss-consumer-ut-XXXXXX";
```
prabhataravind, line 489: Can we cleanup these temp dirs at the end of the tests?

lib/recorder.cpp
```diff
@@ -40,6 +52,249 @@ SwSSRec::SwSSRec()
...
+
+    {
+        std::lock_guard<std::mutex> lock(m_mutex);
+        m_shutdown = true;
+    }
+
+    m_signal.notify_one();
+    m_worker.join();
+    m_workerStarted = false;
+}
```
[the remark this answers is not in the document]
vpandian-nokia: Good point. The current INFO log is emitted before the worker has fully drained, so it does not reflect the final post-shutdown counters. While this does not affect recording correctness, the shutdown summary will be adjusted so the logged counters reflect the final drained/pending state after shutdown completes.

orchagent/main.cpp
```diff
@@ -130,6 +132,51 @@ void sighup_handler(int signo)
...
 }
 
+void fatal_signal_handler(int signo)
+{
+    /*
+     * Do not use SWSS logging here since it takes locks. Dump only the
+     * async recorder counters with async-signal-safe write() and then let
+     * the process terminate with the original signal to preserve core dumps.
+     */
+    dumpAsyncSwssRecorderSignalSafeStats(STDERR_FILENO, signo);
```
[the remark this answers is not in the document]
vpandian-nokia: Agree. The current fatal-signal path should not rely on `kill()` followed by `_exit()` to preserve the default fatal action. This will be updated so the handler dumps the async-recorder stats, restores the default disposition for the signal, and re-raises it in a way that preserves the normal core-dump behavior.

orchagent/main.cpp
```diff
@@ -130,6 +132,51 @@ void sighup_handler(int signo)
...
+    kill(getpid(), signo);
+    _exit(128 + signo);
```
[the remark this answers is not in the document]
vpandian-nokia: Agree. The fatal-signal path should not be the first place that triggers function-local static initialization. This will be hardened by forcing `Recorder::Instance()` construction during normal startup before the fatal handlers are registered.

lib/recorder.cpp
```diff
@@ -40,6 +52,249 @@ SwSSRec::SwSSRec()
...
+{
+    if (m_workerStarted)
+    {
+        return;
+    }
+
+    m_shutdown = false;
+    m_worker = std::thread(&SwSSRec::drain, this);
+    m_workerStarted = true;
+}
```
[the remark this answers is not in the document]
vpandian-nokia: Good catch. `ensureAsyncWorker()` currently relies on unsynchronized worker-start state, so worker creation should be serialized rather than depending on a single-caller assumption. This will be fixed by serializing worker startup/shutdown state so the worker is created exactly once.

lib/recorder.cpp
```diff
@@ -40,6 +52,249 @@ SwSSRec::SwSSRec()
...
 }
 
+SwSSRec::~SwSSRec()
+{
+    stopAsyncWorker();
+}
+
+void SwSSRec::setAsync(bool enabled)
+{
+    m_asyncEnabled = enabled;
```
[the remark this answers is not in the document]
vpandian-nokia: Agree. `setAsync(false)` should not leave the worker alive while later calls can take the synchronous path. This will be updated so disabling async stops and drains the worker, avoiding mixed-mode interleaving and making the API behavior consistent with the current tests.

lib/recorder.cpp
vpandian-nokia: Good point. Holding the `mutex` across the full batch enqueue can delay draining when the batch is large. This is a lock-hold/performance concern rather than a recording-correctness issue, so I’d prefer to keep the immediate follow-up focused on the current correctness and signal-safety fixes and revisit reducing the batch lock scope separately.

lib/recorder.h
vpandian-nokia: Agree that additional observability on backlog growth would be useful. In the current measurements, the observed high-watermark stayed aligned with the orchagent batch size (`-b`), which suggests the queue depth is currently tracking ingress batch shape rather than showing unbounded growth in the tested scenarios. I’d prefer to keep the immediate follow-up focused on the current correctness and signal-safety fixes, and handle queue-depth warning/high-watermark threshold logging as a separate observability improvement.

orchagent/orchdaemon.cpp
vpandian-nokia: I think the `post-select()` shutdown check is still useful. It covers the case where shutdown is requested while blocked in `select()` and lets the loop exit before continuing normal `post-select()` work; the later `EINTR` path only covers one subset of wakeups. I’d prefer to keep this extra guard in place.

tests/mock_tests/consumer_ut.cpp
vpandian-nokia: This helper is currently local to `consumer_ut.cpp` and only used by this fixture, so moving it into a shared test util header feels like more churn than value right now. I’d prefer to keep it file-local unless we end up reusing the same helper across multiple test files.

tests/mock_tests/consumer_ut.cpp
vpandian-nokia: Agree. The temp directory created by this test should be cleaned up before exit. I’ll add cleanup at the end of the test so the unit test does not leave `/tmp` artifacts behind.

lib/recorder.cpp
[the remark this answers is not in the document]
vpandian-nokia: Addressed in follow-up commit `b25d8440`.

`stopAsyncWorker()` now logs the shutdown summary after signaling shutdown and joining the worker, so the INFO line reflects the final post-shutdown `pending`/`drained` counters rather than a pre-drain snapshot.

orchagent/main.cpp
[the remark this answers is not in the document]
vpandian-nokia: Addressed in follow-up commit `b25d8440`.

The fatal signal handler no longer calls `_exit()`. The handler is registered with `SA_RESETHAND`, so the disposition is already reset to the default fatal action on entry; the follow-up change unblocks the handled signal and re-raises it with `kill(getpid(), signo)` so the normal core-dump path can run.

orchagent/main.cpp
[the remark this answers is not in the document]
vpandian-nokia: Addressed in follow-up commit `b25d8440`.

`Recorder::Instance()` is now forced during normal startup in `main()` before the fatal handlers are registered, so the signal path no longer depends on first-time function-local static initialization.

lib/recorder.cpp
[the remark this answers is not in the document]
vpandian-nokia: Addressed in follow-up commit `b25d8440`.

Worker lifecycle is now serialized by protecting the async state transitions with `m_stateMutex`, re-checking async mode under that lock, and starting the worker only from the locked path, so concurrent producers do not race worker creation.

lib/recorder.cpp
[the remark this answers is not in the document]
vpandian-nokia: Addressed in follow-up commit `b25d8440`.

`setAsync(false)` now stops and drains the worker on the enabled-to-disabled transition, which avoids leaving the async worker alive while later calls take the synchronous path.

tests/mock_tests/consumer_ut.cpp
vpandian-nokia: Addressed in follow-up commit `b25d8440`.

The test now removes the temporary recorder file and the temporary directory during cleanup, so it no longer leaves `/tmp` artifacts behind.

orchagent/orchdaemon.cpp
vpandian-nokia: @prabhataravind : Leaving as-is per rationale above; happy to revisit in a follow-up if you feel strongly.

lib/recorder.cpp
vpandian-nokia: No further change planned for this PR. Tracking this as a follow-up improvement.

lib/recorder.h
vpandian-nokia: No further change planned for this PR. Tracking this as a follow-up improvement.

tests/mock_tests/consumer_ut.cpp
vpandian-nokia: No further change planned for this PR. Tracking this as a follow-up improvement.

deepak-singhal0408 (approved):
LGTM

tests/mock_tests/consumer_ut.cpp
prabhataravind: @vpandian-nokia can you please track all the items that are planned as follow-up improvements via a github issue so they are not missed?

tests/quoted_ut.cpp
```diff
@@ -38,3 +99,243 @@ TEST(quoted, copy1_v6)
...
+    string last_line;
+    while (getline(ifs, line))
+    {
+        last_line = line;
+    }
+
+    EXPECT_EQ(last_line, timestamp + "|" + payload);
+}
+
+TEST(swssrec, asyncUpdatesStatsAndWritesRecord)
```
prabhataravind, line 135: @vpandian-nokia several tests in this file still don't cleanup the temp directories. Please fix them.

prabhataravind (approved):
Left a couple of follow-up comments to address. Otherwise lgtm.

prsunny:
Approving for 202605

PR #4403, opened 2026-03-28
[Hw PFCWD feature] Skeleton hw pfcwd

abhishek-nexthop:
**What I did**
Added hardware-based PFC watchdog support with SKU-specific enablement. Created PfcWdHwOrch skeleton class that inherits from PfcWdBaseOrch, refactored shared functionality into the base class, and implemented runtime SKU detection to automatically choose between hardware and software watchdog implementations.

**Why I did it**
Modern switch ASICs support hardware-based PFC deadlock detection/recovery which is more efficient than software-based approaches. Need infrastructure to:
    • Enable hardware watchdog only on supported SKUs
    • Share common code between software and hardware implementations
    • Gracefully fall back to software watchdog on unsupported platforms

**How I verified it**
1. Code compiles successfully with new files added to Makefile
2. SKU detection logic correctly identifies supported hardware:
       • Check logs: grep "PFC hardware watchdog" /var/log/syslog
       • Supported SKU shows: "HWSKU 'nh-4010' supports PFC hardware watchdog"
       • Unsupported SKU shows: "does not support PFC hardware watchdog"
3. Correct watchdog instance created based on SKU and DLR init flag


vmittal-msft:
@abhishek-nexthop please help resolve conflicts

prsunny:
What is this PR about when there are other hw pfcwd PRs.? Folks, if this is experimental, please keep this as "draft".

abhishek-nexthop:
> What is this PR about when there are other hw pfcwd PRs.? Folks, if this is experimental, please keep this as "draft".

@prsunny We have 3 PRs for pfcwd h/w orch. 
1. Refactor the existing code so it can be modularized for sw and hw classes,
2. Boilerplate code for pfcwd hw orch(this PR)
3. Implementation of pfcwd hw orch.
There are other PR which are related to cli/sonic-mgmt/sonic-buildimage Please refer https://github.com/sonic-net/SONiC/pull/2159 for more details on PRs.

vmittal-msft:
@abhishek-nexthop please take care of copilot comments. once depending PRs are in. we can do further review.

vmittal-msft:
@abhishek-nexthop @pinky-nexthop please help resolve conflicts

orchagent/switchorch.cpp
```diff
@@ -127,6 +127,30 @@ static std::unordered_set<std::string> serializeSwitchCounterStats(const std::ve
 
 // Switch OA ----------------------------------------------------------------------------------------------------------
 
+void SwitchOrch::set_switch_pfc_dldr_capability()
+{
+    vector<FieldValueTuple> fvVector;
+
+    /* Query PFC DLDR capability. SAI_QUEUE_ATTR_ENABLE_PFC_DLDR covers both
+     * deadlock detection and recovery in hardware, so it distinguishes a
```
vmittal-msft, line 135: @abhishek-nexthop i believe there a BRCM CSP open to support this ? Can we still check in this until that is fixed ?

vmittal-msft:
@abhishek-nexthop can you please share if SW PFCWD still working fine with these changes? please share which platforms it is verified on ?

orchagent/pfcwdorch.h
```diff
@@ -112,13 +128,12 @@ class PfcWdBaseOrch: public Orch
...
 
-    string m_platform = "";
-    shared_ptr<FlexCounterTaggedCachedManager<sai_object_type_t>> m_pfcwdFlexCounterManager;
-
 private:
 
     shared_ptr<DBConnector> m_countersDb = nullptr;
     shared_ptr<Table> m_countersTable = nullptr;
+    shared_ptr<DBConnector> m_stateDb = nullptr;
+    shared_ptr<Table> m_stateTable = nullptr;
```
vmittal-msft: These new `m_stateDb`/`m_stateTable` members default to `nullptr` and are never initialized — the constructor at `pfcwdorch.cpp:30-42` only initializes `m_countersDb`/`m_countersTable`. But the new inline helper `updateStateTable()` (L88-92) dereferences both (`m_stateDb->hset(...)` and `m_stateTable->getTableName()`), so the first caller (the implementation PR) will hit a null-deref.

Since the members are landing here, simplest is to initialize them in the constructor now, e.g.:

```cpp
m_stateDb(new DBConnector("STATE_DB", 0)),
m_stateTable(new Table(m_stateDb.get(), <state table name>)),
```

Alternatively, defer `updateStateTable()`/`updateDlrPacketActionInStateTable()` to the implementation PR. Either way, avoid landing added members plus dereferencing helpers that aren't wired up.

vmittal-msft (commented):
Reviewing this as the intentional skeleton PR (#2 of 3: refactor → skeleton → implementation), so the never-instantiated `pfcwdhworch.h` and the log-only `checkPfcDldrEnable()` in orchdaemon are expected — no concern there.

**One substantive item** (inline on `pfcwdorch.h`): the newly added `m_stateDb`/`m_stateTable` are never initialized, but the new inline helpers dereference them, so the first caller hits a null-deref.

**Stale/addressed:** the earlier Copilot comments about the stub `createEntry`/`deleteEntry` returning `task_success`, the `memset`, and orchdaemon instantiating the stub no longer apply — the `.cpp` was removed and orchdaemon no longer constructs `PfcWdHwOrch`.

**Good:** making `deleteEntry` virtual is exactly what enables the future HW override (signatures match the base pure-virtuals), and detecting capability via the SAI DLDR attribute instead of an HWSKU regex is the right, scalable choice.

**Minor/optional nits:** `updateStateTable()` could just be `m_stateTable->hset("PFC_WD", field, value)`; `pfcwdhworch.h` relies on transitive includes for `<unordered_map>/<set>/<functional>/<memory>`; `doTask(SelectableTimer&)`/`onQueuePfcDeadlock` aren't marked `override`; and ensure the add-then-revert `.cpp`/test commits get squashed on merge.

orchagent/pfcwdhworch.cpp
```diff
@@ -0,0 +1,256 @@
...
+    // TODO: Implementation
+    return false;
+}
+
+task_process_status PfcWdHwOrch::createEntry(const string& key, const vector<FieldValueTuple>& data)
+{
+    SWSS_LOG_ENTER();
+    // TODO: Implementation
+    return task_process_status::task_success;
+}
```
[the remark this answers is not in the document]
arawat-nexthop: Addressed. createEntry was first changed to return task_invalid_entry with an error log, and orchagent/pfcwdhworch.cpp has since been removed from this PR entirely — the class declaration lands here and the definitions arrive with the implementation in #4412. Nothing constructs PfcWdHwOrch, so no CONFIG_DB update can reach this path.

orchagent/pfcwdhworch.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+task_process_status PfcWdHwOrch::deleteEntry(const string& key)
+{
+    SWSS_LOG_ENTER();
+    // TODO: Implementation
+    return task_process_status::task_success;
+}
```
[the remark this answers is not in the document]
arawat-nexthop: Addressed, same as the createEntry comment: deleteEntry returned task_invalid_entry with an error log, and orchagent/pfcwdhworch.cpp is no longer part of this PR.

orchagent/pfcwdhworch.cpp
```diff
@@ -0,0 +1,256 @@
...
+    // TODO: Implementation
+}
+
+PfcWdHwOrch::PfcWdQueueStats PfcWdHwOrch::getQueueStats(const string &queueIdStr)
+{
+    SWSS_LOG_ENTER();
+    // TODO: Implementation
+    PfcWdQueueStats stats;
+    memset(&stats, 0, sizeof(PfcWdQueueStats));
+    return stats;
```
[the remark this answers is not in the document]
arawat-nexthop: No longer applies — getQueueStats() was one of the unused private helpers removed from this PR. When the implementation lands in #4412 the struct will be value-initialised rather than memset.

orchagent/orchdaemon.cpp
```diff
@@ -781,8 +781,31 @@ bool OrchDaemon::init()
...
+        if (pfcDlrInit && pfcHwWdSupportedSku)
+        {
+            SWSS_LOG_NOTICE("Starting hardware-based pfc watchdog");
+            m_orchList.push_back(new PfcWdHwOrch(
+                        m_configDb,
+                        pfc_wd_tables,
+                        portStatIds,
+                        queueStatIds,
+                        queueAttrIds));
+        }
```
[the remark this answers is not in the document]
arawat-nexthop: Addressed. orchdaemon no longer constructs PfcWdHwOrch at all; it does not reference the class. The SKU-matching selection this comment describes was also removed — the platform check is now a runtime SAI capability query (SAI_QUEUE_ATTR_ENABLE_PFC_DLDR) rather than an HWSKU regex, and it only publishes a capability value. Nothing selects the hardware watchdog until the implementation lands.

orchagent/switchorch.cpp
```diff
@@ -160,6 +178,12 @@ SwitchOrch::SwitchOrch(DBConnector *db, vector<TableConnector>& connectors, Tabl
     auto restartCheckNotifier = new Notifier(m_restartCheckNotificationConsumer, this, "RESTARTCHECK");
     Orch::addExecutor(restartCheckNotifier);
 
+    // Read HWSKU from CONFIG_DB
+    DBConnector configDb("CONFIG_DB", 0);
+    Table deviceMetadataTable(&configDb, CFG_DEVICE_METADATA_TABLE_NAME);
+    deviceMetadataTable.hget("localhost", "hwsku", m_hwSku);
+    SWSS_LOG_NOTICE("HWSKU: %s", m_hwSku.c_str());
+
```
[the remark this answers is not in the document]
arawat-nexthop: No longer applies — m_hwSku and the HWSKU-based selection were removed. switchorch now queries the DLDR capability from SAI and publishes PFC_DLDR_CAPABLE to the switch capability table, so there is no hwsku read to guard.

orchagent/pfcwdorch.h
arawat-nexthop: Correct on all three. Deferred both members and helpers to #4412 rather than initialising here — initialising in PfcWdBaseOrch's constructor would open a STATE_DB connection for every PfcWdSwOrch instance too, for a field only the hardware path writes. #4412 already initialises them

orchagent/orchdaemon.cpp
```diff
@@ -825,8 +825,17 @@ bool OrchDaemon::init()
...
         }
 
+        // Complete hardware recovery needs deadlock detection and recovery in
+        // hardware, which SAI_QUEUE_ATTR_ENABLE_PFC_DLDR reports. A hybrid
+        // platform only advertises SAI_QUEUE_ATTR_PFC_DLR_INIT.
+        if (gSwitchOrch->checkPfcDldrEnable())
+        {
+            SWSS_LOG_NOTICE("Switch supports PFC hardware watchdog");
+        }
+
```
kperumalbfn, line 835: @abhishek-nexthop DLDR should take precedence over DLR as some vendor SAI can return DLRInit as true when DLDR is also supported in the hardware. We need to handle that check and avoid any s/w handler for DLDR

orchagent/orchdaemon.cpp
```diff
@@ -825,8 +825,17 @@ bool OrchDaemon::init()
...
         if(pfcDlrInit)
         {
+            SWSS_LOG_NOTICE("Starting dlr init handler for pfc watchdog");
```
kperumalbfn, line 838: Could we add a tests to check {DLR, DLDR} enable/disable combination?

orchagent/switchorch.cpp
arawat-nexthop: Yes, Vineet. This PR doesn't construct PfcWdHwOrch. I have tested internally with the changes in current comment. 
`SONiC.arawat.pfcwd-dldr-testbuild.Nexthop.164541-1eb0b871f7`

SW PFCWD was armed without any issues.

orchagent/orchdaemon.cpp
arawat-nexthop: the DLDR check here only publishes the capability, and PfcWdHwOrch isn't constructed yet. Gating the SW handler on DLDR now would leave those platforms with no watchdog at all.

The selection happens in #4412

on the box that actually has both flags true(TH5) reported PFC_DLDR_CAPABLE: true and PFC_DLR_INIT_CAPABLE: true, and SW PFCWD still came up:

PORT          ACTION   DETECTION TIME   RESTORATION TIME
Ethernet448   drop     200              200
Ethernet456   drop     200              200

On the {DLR, DLDR} combination tests — this is planned in the implementation PR along with the other coverage tests

kperumalbfn: Makes sense for this initial conditional checks. With #4412, we should make sure DLDR has higher precedence over DLR and no other handles likePfcWdAclhandlers are created.

kperumalbfn (approved):
LGTM

pinky-nexthop:
@yxieca @rlhui Kindly help merge. CI is all green. Thanks.

PR #4404, opened 2026-03-28
[acl] Support arbitrary IP masks for ACL match fields

ashutosh-agrawal:
**What I did**

Added `SRC_IP_MASK`, `DST_IP_MASK`, `SRC_IPV6_MASK`, `DST_IPV6_MASK` match fields to support non-contiguous (arbitrary) IPv4/IPv6 masks that cannot be expressed in CIDR notation. CIDR notation continues to work unchanged.

**Why I did it**

Some network policies require matching on non-contiguous bit patterns (e.g. `255.0.255.0`) that are impossible to express as a prefix length. This adds the flexibility to specify the mask independently of the IP address.

**How I verified it**

- Unit tests: `AclRule_ArbitraryIpv4Mask`, `AclRule_ArbitraryIpv6Mask` in `aclorch_ut.cpp`
- DVS tests: `test_AclRuleArbitraryIpv4Mask`, `test_V6AclRuleArbitraryIpv6Mask` in `test_acl.py`

**Details if related**

Side-effect fix: `setMatch()` now correctly removes a rejected SAI attribute from `m_matches` on validation failure (pre-existing bug).

Yang Model PR: https://github.com/sonic-net/sonic-buildimage/pull/27046

### DIFF
--- a/doc/swss-schema.md
+++ b/doc/swss-schema.md
@@ -661,14 +661,31 @@ Stores rules associated with a specific ACL table on the switch.
     src_ip        = ipv4_prefix                ; options of the source ipv4
                                                ; address (and mask) field
 
+    src_ip_mask   = ipv4_address               ; optional non-contiguous IPv4 mask for src_ip.
+                                               ; Used when the mask cannot be expressed as a
+                                               ; CIDR prefix length (e.g. "255.0.255.0").
+                                               ; Requires src_ip to be a plain host address
+                                               ; (no "/" prefix notation).
+
     dst_ip        = ipv4_prefix                ; options of the destination ipv4
                                                ; address (and mask) field
+
+    dst_ip_mask   = ipv4_address               ; optional non-contiguous IPv4 mask for dst_ip.
+                                               ; See src_ip_mask for usage.
 
     src_ipv6      = ipv6_prefix                ; options of the source ipv6
                                                ; address (and mask) field
 
+    src_ipv6_mask = ipv6_address               ; optional non-contiguous IPv6 mask for src_ipv6.
+                                               ; Used when the mask cannot be expressed as a
+                                               ; CIDR prefix length (e.g. "ffff::ffff").
+                                               ; Requires src_ipv6 to be a plain host address.
+
     dst_ipv6      = ipv6_prefix                ; options of the destination ipv6
                                                ; address (and mask) field
+
+    dst_ipv6_mask = ipv6_address               ; optional non-contiguous IPv6 mask for dst_ipv6.
+                                               ; See src_ipv6_mask for usage.
 
     l4_src_port   = port_num                   ; source L4 port or the
     l4_dst_port   = port_num                   ; destination L4 port
@@ -700,6 +717,8 @@ Stores rules associated with a specific ACL table on the switch.
     h16         = 1*4HEXDIG
     ls32        = ( h16 ":" h16 ) / IPv4address
     ipv4_prefix = dec-octet "." dec-octet "." dec-octet "." dec-octet “/” %d1-32
+    ipv4_address = dec-octet "." dec-octet "." dec-octet "." dec-octet
+    ipv6_address = ipv6 host address without prefix-length (e.g. "ffff::ffff")
     dec-octet   = DIGIT                     ; 0-9
                     / %x31-39 DIGIT         ; 10-99
                     / "1" 2DIGIT            ; 100-199
--- a/doc/swss-schema.md
+++ b/doc/swss-schema.md
@@ -658,8 +658,8 @@ Stores rules associated with a specific ACL table on the switch.
 
     ip_protocol   = h8                         ; options of the l3_protocol_type field
 
-    src_ip        = ipv4_prefix                ; options of the source ipv4
-                                               ; address (and mask) field
+    src_ip        = ipv4_prefix / ipv4_address ; source IPv4 match: CIDR prefix (e.g. "10.0.0.0/8")
+                                               ; or plain host address when paired with src_ip_mask
 
     src_ip_mask   = ipv4_address               ; optional non-contiguous IPv4 mask for src_ip.
                                                ; Used when the mask cannot be expressed as a
@@ -667,22 +667,22 @@ Stores rules associated with a specific ACL table on the switch.
                                                ; Requires src_ip to be a plain host address
                                                ; (no "/" prefix notation).
 
-    dst_ip        = ipv4_prefix                ; options of the destination ipv4
-                                               ; address (and mask) field
+    dst_ip        = ipv4_prefix / ipv4_address ; destination IPv4 match: CIDR prefix or plain
+                                               ; host address when paired with dst_ip_mask
 
     dst_ip_mask   = ipv4_address               ; optional non-contiguous IPv4 mask for dst_ip.
                                                ; See src_ip_mask for usage.
 
-    src_ipv6      = ipv6_prefix                ; options of the source ipv6
-                                               ; address (and mask) field
+    src_ipv6      = ipv6_prefix / ipv6_address ; source IPv6 match: prefix (e.g. "2001:db8::/32")
+                                               ; or plain host address when paired with src_ipv6_mask
 
     src_ipv6_mask = ipv6_address               ; optional non-contiguous IPv6 mask for src_ipv6.
                                                ; Used when the mask cannot be expressed as a
-                                               ; CIDR prefix length (e.g. "ffff::ffff").
+                                               ; prefix length (e.g. "ffff::ffff").
                                                ; Requires src_ipv6 to be a plain host address.
 
-    dst_ipv6      = ipv6_prefix                ; options of the destination ipv6
-                                               ; address (and mask) field
+    dst_ipv6      = ipv6_prefix / ipv6_address ; destination IPv6 match: prefix or plain host
+                                               ; address when paired with dst_ipv6_mask
 
     dst_ipv6_mask = ipv6_address               ; optional non-contiguous IPv6 mask for dst_ipv6.
                                                ; See src_ipv6_mask for usage.
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -61,8 +61,12 @@ acl_rule_attr_lookup_t aclMatchLookup =
     { MATCH_OUT_PORTS,         SAI_ACL_ENTRY_ATTR_FIELD_OUT_PORTS },
     { MATCH_SRC_IP,            SAI_ACL_ENTRY_ATTR_FIELD_SRC_IP },
     { MATCH_DST_IP,            SAI_ACL_ENTRY_ATTR_FIELD_DST_IP },
+    { MATCH_SRC_IP_MASK,       SAI_ACL_ENTRY_ATTR_FIELD_SRC_IP },
+    { MATCH_DST_IP_MASK,       SAI_ACL_ENTRY_ATTR_FIELD_DST_IP },
     { MATCH_SRC_IPV6,          SAI_ACL_ENTRY_ATTR_FIELD_SRC_IPV6 },
     { MATCH_DST_IPV6,          SAI_ACL_ENTRY_ATTR_FIELD_DST_IPV6 },
+    { MATCH_SRC_IPV6_MASK,     SAI_ACL_ENTRY_ATTR_FIELD_SRC_IPV6 },
+    { MATCH_DST_IPV6_MASK,     SAI_ACL_ENTRY_ATTR_FIELD_DST_IPV6 },
     { MATCH_L4_SRC_PORT,       SAI_ACL_ENTRY_ATTR_FIELD_L4_SRC_PORT },
     { MATCH_L4_DST_PORT,       SAI_ACL_ENTRY_ATTR_FIELD_L4_DST_PORT },
     { MATCH_ETHER_TYPE,        SAI_ACL_ENTRY_ATTR_FIELD_ETHER_TYPE },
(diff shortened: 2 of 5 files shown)


prsunny:
@bingwang-ms , @kperumalbfn , please review and signoff for merge.

ashutosh-agrawal:
@kperumalbfn @bingwang-ms Friendly reminder for review.

bingwang-ms:
⚠️ **This review was performed by an AI agent** on behalf of @bingwang-ms. Please treat this as a preliminary review and verify independently.

---

## PR Review: [acl] Support arbitrary IP masks and L3V6 IN_PORTS/OUT_PORTS

### What the PR Does
1. **Arbitrary IP masks** — Adds `SRC_IP_MASK`, `DST_IP_MASK`, `SRC_IPV6_MASK`, `DST_IPV6_MASK` match fields using a deferred-join approach. CIDR notation continues to work unchanged.
2. **L3V6 IN_PORTS/OUT_PORTS** — Adds port-specific matching for L3V6 ACL tables.
3. **Side-effect bug fix** — Fixes a pre-existing bug where a rejected SAI attribute could remain in `m_matches`.

### ⚠️ Issues Found

#### Medium/Critical

1. **Double-call to `processPendingIpFields()`** — Called both in `doAclRuleTask` and inside `create()`. The second call is a no-op (pending maps are cleared after first call) but is confusing and inconsistent. Recommend removing the internal call from `create()` or the explicit call in `doAclRuleTask`.

2. **Orphaned `_MASK` silently ignored** — If `SRC_IP_MASK` (or `DST_IP_MASK` etc.) is specified without a corresponding `SRC_IP`, the mask is silently dropped with no warning or log. At minimum a `SWSS_LOG_WARN` should be emitted for unused masks after the loop in `processPendingIpFields()`.

3. **Dead code in `aclMatchLookup`** — The new entries `MATCH_SRC_IP_MASK → SAI_ACL_ENTRY_ATTR_FIELD_SRC_IP` and `MATCH_DST_IP_MASK → SAI_ACL_ENTRY_ATTR_FIELD_DST_IP` are never actually reached via that lookup path (mask branches return early). These are dead code and could be misleading. Recommend removing or adding a comment explaining why they exist.

4. **L3V6 IN_PORTS/OUT_PORTS in `stageMandatoryMatchFields`** — Adding these to "mandatory" fields means ALL L3V6 tables will request `SAI_ACL_TABLE_ATTR_FIELD_IN_PORTS`/`OUT_PORTS` at creation time, even if no rule uses them. This could cause table creation failures on hardware that doesn't support this combination. Please verify SAI compatibility across all target platforms, or consider making these optional/capability-gated.

#### Minor

5. **Deferred IPv4 validation** — `validateAddMatch` returns `true` for non-CIDR IP fields without validating the address format, deferring to `processPendingIpFields()`. A comment explaining this is intentional deferred validation would help readability.

6. **`INNER_SRC_IP` asymmetry** — `MATCH_INNER_SRC_IP` goes through CIDR-only logic with no mask equivalent. Worth a comment if this is intentional.

7. **`processPendingIpFields()` visibility** — Currently `public` but is an internal implementation detail. Could be `private` with `friend` test access.

### Overall Assessment
The core logic is sound and well-structured. Items **1** and **4** in particular warrant discussion before merge. None are showstoppers, but they should be addressed or explicitly acknowledged.

orchagent/aclorch.cpp
```diff
@@ -1099,26 +1101,59 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
         }
         else if (attr_name == MATCH_SRC_IP || attr_name == MATCH_DST_IP || attr_name == MATCH_INNER_SRC_IP)
         {
-            IpPrefix ip(attr_value);
-
-            if (!ip.isV4())
+            if (attr_value.find('/') != string::npos)
```
kperumalbfn, line 1104: Please create a separate method to check for subnet in the config

orchagent/aclorch.h
```diff
@@ -28,8 +28,12 @@
 #define MATCH_OUT_PORTS         "OUT_PORTS"
 #define MATCH_SRC_IP            "SRC_IP"
 #define MATCH_DST_IP            "DST_IP"
+#define MATCH_SRC_IP_MASK       "SRC_IP_MASK"
```
kperumalbfn, line 31: any change required in acl-loader to add mask?

kperumalbfn: please update swss-schema.md for the new set of mask fields

orchagent/aclorch.cpp
```diff
@@ -472,13 +472,15 @@ static acl_table_match_field_lookup_t stageMandatoryMatchFields =
...
+                    SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE,
+                    SAI_ACL_TABLE_ATTR_FIELD_IN_PORTS
                 }
             },
             {
                 ACL_STAGE_EGRESS,
                 {
-                    SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE
+                    SAI_ACL_TABLE_ATTR_FIELD_ACL_RANGE_TYPE,
+                    SAI_ACL_TABLE_ATTR_FIELD_OUT_PORTS
```
kperumalbfn, line 483: Why do we need IN_PORTS and OUT_PORTS part of mandatory? This might increase the tcam key width if IN_PORT/OUT_PORT is not used in rules part of these tables.

tests/mock_tests/aclorch_ut.cpp
```diff
@@ -2365,6 +2364,203 @@ namespace aclorch_test
...
+
+        // SRC_IP plain address + SRC_IP_MASK
+        addRule({ { MATCH_SRC_IP, "192.168.1.1" }, { MATCH_SRC_IP_MASK, "255.255.255.0" } });
+        ASSERT_NE(acl_table.rules.find(acl_rule_id), acl_table.rules.end());
+        ASSERT_EQ(getAclRuleSaiAttribute(*acl_table.rules.at(acl_rule_id), SAI_ACL_ENTRY_ATTR_FIELD_SRC_IP),
+                  "192.168.1.1&mask:255.255.255.0");
+        delRule();
+
+        // DST_IP plain address + DST_IP_MASK
+        addRule({ { MATCH_DST_IP, "10.0.0.1" }, { MATCH_DST_IP_MASK, "255.0.0.0" } });
```
kperumalbfn, line 2423: could you also add sonic-mgmt tests for the mask config?

orchagent/aclorch.h
kperumalbfn: could you also check for ACL yang model?

orchagent/aclorch.cpp
ashutosh-agrawal: @kperumalbfn I've removed IN_PORTS/OUT_PORTS related changes out of this PR and will handle them separately.

orchagent/aclorch.h
ashutosh-agrawal: > any change required in acl-loader to add mask?

acl-loader is a OpenConfig to Config-DB translator and there is no *_Mask field in  OpencConfig schema, so acl-loader is not expected to support this new capability.

ashutosh-agrawal: > please update swss-schema.md for the new set of mask fields

Updated swss-schema.md as requested.

ashutosh-agrawal: > could you also check for ACL yang model?

Created https://github.com/sonic-net/sonic-buildimage/pull/27046/ for yang changes. Please help with review.

tests/mock_tests/aclorch_ut.cpp
ashutosh-agrawal: Yes, I’m working on a sonic-mgmt test and should have a PR out in the next few days. Hopefully that will not block this PR from merging.

orchagent/aclorch.cpp
ashutosh-agrawal: Added a method as requested.

ashutosh-agrawal:
@kperumalbfn Please trigger ASAN sanity for this CI run - https://dev.azure.com/mssonic/build/_build/results?buildId=1101170&view=results

tests/test_acl.py
```diff
@@ -779,6 +779,80 @@ def test_AclTableMandatoryMatchFields(self, dvs, pfcwd_acl_table):
         else:
             assert not match_in_ports
 
+    def test_AclRuleArbitraryIpv4Mask(self, dvs_acl, l3_acl_table):
+        """Verify SRC_IP/DST_IP with a separate *_MASK field creates an ACL rule
+        with a non-contiguous IPv4 mask that cannot be expressed as a CIDR prefix.
+        """
+        # SRC_IP + SRC_IP_MASK — non-contiguous mask (cannot be expressed as /prefix)
```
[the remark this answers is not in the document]
ashutosh-agrawal: Added negative tests for malformed combinations.

orchagent/aclorch.cpp
```diff
@@ -940,6 +971,17 @@ bool AclRule::validateAddMatch(string attr_name, string attr_value)
 
     matchData.enable = true;
 
+    // *_MASK fields have no direct SAI entry attribute mapping; they are stored
+    // here and combined with the companion IP field in processPendingIpFields().
+    // This check must come before the aclMatchLookup guard below.
+    if (attr_name == MATCH_SRC_IP_MASK || attr_name == MATCH_DST_IP_MASK ||
+        attr_name == MATCH_SRC_IPV6_MASK || attr_name == MATCH_DST_IPV6_MASK)
+    {
+        string ip_field = attr_name.substr(0, attr_name.length() - 5);
```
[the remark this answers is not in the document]
ashutosh-agrawal: Added explicit check to ensure that SRC/DST_IP field is present. This check is also enforced by the yang-model.

ashutosh-agrawal (commented):
@kperumalbfn I've resolved all the copilot comments.

orchagent/aclorch.cpp
```diff
@@ -1253,8 +1310,89 @@ bool AclRule::processIpType(string type, sai_uint32_t &ip_type)
...
+        }
+    }
+
+    m_pendingIpFields.clear();
+
+    // Warn about any mask fields that had no corresponding IP field
+    for (const auto& entry : m_pendingIpMasks)
+    {
+        SWSS_LOG_WARN("IP mask field %s_MASK specified without a corresponding %s address field; ignoring",
+            entry.first.c_str(), entry.first.c_str());
```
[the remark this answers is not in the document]
ashutosh-agrawal: Fixed.

prsunny:
@ashutosh-agrawal , does SAI support such arbitrary mask? Do we need any capability check here?

ashutosh-agrawal:
> @ashutosh-agrawal , does SAI support such arbitrary mask? Do we need any capability check here?

@prsunny SAI supports arbitrary masks natively for all ACL key fields. The sai_acl_field_data_t struct has separate .data and .mask fields as raw IP address bitmasks. Any IP address with CIDR notation like 10.0.0.0/8 was already being converted to 255.0.0.0 before being written to matchData.mask.ip4. This PR essentially lets you specify that bitmask directly from Config DB.

Since, we are not introducing any new SAI attributes, there is nothing additional to query for. If a platform's ASIC doesn't support a non-contiguous mask pattern, sai_acl_api->create_acl_entry() will return an error, the rule will be marked Inactive in STATE_DB, and an error will be logged — same as any other unsupported ACL config.

prsunny:
> > @ashutosh-agrawal , does SAI support such arbitrary mask? Do we need any capability check here?
> 
> @prsunny SAI supports arbitrary masks natively for all ACL key fields. The sai_acl_field_data_t struct has separate .data and .mask fields as raw IP address bitmasks. Any IP address with CIDR notation like 10.0.0.0/8 was already being converted to 255.0.0.0 before being written to matchData.mask.ip4. This PR essentially lets you specify that bitmask directly from Config DB.
> 
> Since, we are not introducing any new SAI attributes, there is nothing additional to query for. If a platform's ASIC doesn't support a non-contiguous mask pattern, sai_acl_api->create_acl_entry() will return an error, the rule will be marked Inactive in STATE_DB, and an error will be logged — same as any other unsupported ACL config.

ok, looks good then. @kperumalbfn , @bingwang-ms , would you sign off

ashutosh-agrawal:
@bingwang-ms @kperumalbfn Gentle reminder for review/approval.

doc/swss-schema.md
```diff
@@ -658,17 +658,34 @@ Stores rules associated with a specific ACL table on the switch.
 
     ip_protocol   = h8                         ; options of the l3_protocol_type field
 
-    src_ip        = ipv4_prefix                ; options of the source ipv4
-                                               ; address (and mask) field
+    src_ip        = ipv4_prefix / ipv4_address ; source IPv4 match: CIDR prefix (e.g. "10.0.0.0/8")
```
bingwang-ms, line 661: Can you show a config_db sample for an ACL rule with SRC_IP_MASK? Trying to understand how the src_ip_mask is mapped to src_ip.

orchagent/aclorch.cpp
```diff
@@ -1253,8 +1310,97 @@ bool AclRule::processIpType(string type, sai_uint32_t &ip_type)
...
+                matchData.data.ip4 = ip.getV4Addr();
+                if (maskIt != m_pendingIpMasks.end())
+                {
+                    IpAddress mask(maskIt->second);
+                    matchData.mask.ip4 = mask.getV4Addr();
+                    m_pendingIpMasks.erase(maskIt);
+                }
+                else
+                {
+                    matchData.mask.ip4 = 0xFFFFFFFF;
```
bingwang-ms, line 1343: Does this mean if there is no mask present, then 0xFFFFFFFF is applied by default?
This changed the default behavior and may cause compatible issue.

ashutosh-agrawal: @bingwang-ms Yes, there is a minor behavior difference here. Earlier, CIDR notation was mandatory and a plain IP like 10.1.2.3 would have been rejected. Now it's accepted and treated as a host-exact match (equivalent to /32). This is intentional — the *_MASK feature requires plain-address form, so we had to accept plain addresses, and defaulting to host-exact is the natural behavior when no mask is provided. 

All existing configs using CIDR notation are unaffected.

doc/swss-schema.md
ashutosh-agrawal: @bingwang-ms There is an example entry in sample_config_db.json of the associated yang-model PR
https://github.com/sonic-net/sonic-buildimage/pull/27046/changes#diff-35a7f6a55896179bc30a6aa51e5eaf7bc06cf3f541b6c77cfeef0e270f59fd7f

There are also examples in DVS test test_acl.py of this PR that show the config-DB snippet as well as the corresponding SAI ACL entry.

prsunny:
@bingwang-ms , could you check again on this PR for comments reply?

ashutosh-agrawal:
@kperumalbfn @bingwang-ms @prsunny Friendly reminder for review/approval.

prsunny:
Changes lgtm, @bingwang-ms to approve for merge

bingwang-ms (approved):

bingwang-ms:
LGTM. Thanks for addressing the comments

ashutosh-agrawal:
@bingwang-ms Thanks for approving this PR. Can you also please review the companion PR for yang model - https://github.com/sonic-net/sonic-buildimage/pull/27046

PR #4410, opened 2026-03-30
orchagent: Add missing end() check in TunnelDecapOrch::removeDecapTunnel

uramanath:
**What I did**

Added an `end()` check before dereferencing the `tunnelTable.find(key)` result in `TunnelDecapOrch::removeDecapTunnel()`.

**Why I did it**

`tunnelTable.find(key)->second` is dereferenced without verifying the iterator against `end()`. If the key is absent, this is undefined behavior and can crash orchagent. While current callers guard with an existence check, the function itself should be defensive against invalid keys.

**How I verified it**

Code inspection — the fix resolves the iterator first and returns early with an error log when the tunnel is not found. No behavioral change for valid tunnels. All existing tests pass.


prsunny:
can you add coverage?

PR #4411, opened 2026-03-30
orchagent: Add null check after dynamic_cast in DebugCounterOrch

uramanath:
**What I did**

Added null checks after `dynamic_cast<DropCounter*>` in `DebugCounterOrch::addDropReason()` and `DebugCounterOrch::removeDropReason()`.

**Why I did it**

Both functions cast the stored debug counter to `DropCounter*` via `dynamic_cast` but dereference the result on the very next line without verifying it is non-null. If the cast fails (e.g. a future counter type that isn't a `DropCounter`), the unchecked dereference is undefined behavior and will crash orchagent.

**How I verified it**

Code inspection — the fix adds a null check and returns `task_failed` with an error log when the cast fails. No behavioral change when the cast succeeds. All existing tests pass.


theasianpianist:
Just curious, is this an issue that was actually observed?

orchagent/debugcounterorch.cpp
```diff
@@ -481,7 +486,7 @@ task_process_status DebugCounterOrch::removeDropReason(const string& counter_nam
 
     if (!isDropReasonValid(drop_reason))
     {
-        return task_failed;
+        return task_process_status::task_failed;
```
theasianpianist, line 489: Is this change needed? Let's keep things consistent within this function

PR #4412, opened 2026-03-30
[Hw PFCWD feature] Implementation for H/w based PFCWD

abhishek-nexthop:
**What I did**
Implemented complete hardware-based PFC watchdog support for SONiC with ASIC-native storm detection and recovery. The implementation includes:

- Integrated into orchagent daemon with platform capability detection, FlexCounter manager initialization, and SAI notification callback registration (SAI_SWITCH_ATTR_QUEUE_PFC_DEADLOCK_NOTIFY)
- Implemented full configuration workflow: parameter validation, hardware timer range checking, action consistency enforcement, SAI attribute programming (switch-level DLR action, port-level timer intervals/granularity), per-queue
- DLDR enablement, FlexCounter registration, and COUNTERS_DB initialization
- Implemented cleanup workflow: storm-state validation, DLDR disablement, timer interval clearing, FlexCounter deregistration, STATE_DB cleanup, and global action reset
- Implemented counter logic: SAI notification handling for storm detection/recovery, baseline counter initialization, periodic counter updates during storms, TX/RX statistics collection from queues and priority groups, delta calculation with underflow protection, and detect/restore count tracking
- Added warm reboot recovery to restore configurations from CONFIG_DB
- Updated FlexCounter configuration to collect queue dropped packets and priority group packets

**Why I did it**
Modern ASICs have native hardware PFC deadlock detection providing microsecond-latency storm detection vs millisecond software polling, with zero CPU overhead and line-rate accuracy. This implementation leverages these hardware capabilities while maintaining operational consistency with existing software watchdog through a shared base class architecture.

**How I verified it**
Ran multiple testcases for pfcwd with h/w based recovery enabled and they worked as expected.

**Details if related**


orchagent/switchorch.cpp
```diff
@@ -144,6 +145,23 @@ void SwitchOrch::set_switch_pfc_dlr_init_capability()
     set_switch_capability(fvVector);
 }
 
+bool SwitchOrch::isHwPfcWdSupportedSku() const
+{
+    static const std::vector<std::regex> patterns = {
+        std::regex("nh-4010.*", std::regex_constants::icase)
```
kperumalbfn, line 151: Can we use capability query instead of SKU list?

abhishek-nexthop: @kperumalbfn This is a temporary implementation for now. We’re keeping a fallback option in place for certain SKUs if needed. Over time, we’ll move it to the default approach, removing the whitelist logic and relying entirely on capabilities.

kperumalbfn: @abhishek-nexthop there are lot of SKUs that support hardware based PFCWD, so it will be better to use the capability model.

orchagent/switchorch.cpp
```diff
@@ -144,6 +145,23 @@ void SwitchOrch::set_switch_pfc_dlr_init_capability()
...
+    };
+
```
kperumalbfn: @pinky-nexthop @abhishek-nexthop also for global PFCWD timer, which module sets SAI_SWITCH_ATTR_PFC_TC_DLD_INTERVAL and SAI_SWITCH_ATTR_PFC_TC_DLR_INTERVAL. There are high radix switches where global DLD/DLR makes sense instead of per-port config.

kperumalbfn: @eddyk-nvidia FYI

orchagent/switchorch.cpp
abhishek-nexthop: @kperumalbfn We are doing this intentionally so that it is not enabled for SKUs that haven’t been tested yet. Once a SKU qualifies the workflow, the whitelist can be updated. Using regex helps qualify a large number of chips with a single rule. After we have a sufficient number of qualified SKUs, we can enable it by default based on capability alone. 
There is already a branch based on DLR init capability, which would be disabled if we removed this check. 
We received the same suggestion from @vmittal-msft [here](https://github.com/sonic-net/SONiC/pull/2159#discussion_r2743100301)

orchagent/switchorch.cpp
abhishek-nexthop: @kperumalbfn Right now, we are not using global PFCWD timer. Is there any CLI support currently available for configuring global counters as part of a global workflow? Our current implementation is based on port-level timer support. While it is possible to extend this to switch-level support, we do not have the required chip available to validate it.
If switch-level/global support is needed, this can be used as a reference and extended further based on specific workflow requirements.

orchagent/switchorch.cpp
prsunny: @abhishek-nexthop , @vmittal-msft , responded to the comment on HLD, this should not be sku check

orchagent/switchorch.cpp
kperumalbfn: @abhishek-nexthop We could also go per-capability. If some ASICs don't have per-port timer, then we should go with global timer. There is a global scope in PFCWD config

kperumalbfn: @eddyk-nvidia What is the model in Spectrum ASICs? Does it support both port and global mode??

orchagent/switchorch.cpp
eddyk-nvidia: Can just add to what Microsoft folks mentioned and I commented in HLD that use of SKU in the SONIC code is a wrong approach

orchagent/switchorch.cpp
eddyk-nvidia: > [@kperumalbfn](https://github.com/kperumalbfn) Right now, we are not using global PFCWD timer. Is there any CLI support currently available for configuring global counters as part of a global workflow? Our current implementation is based on port-level timer support. While it is possible to extend this to switch-level support, we do not have the required chip available to validate it. If switch-level/global support is needed, this can be used as a reference and extended further based on specific workflow requirements.


@abhishek-nexthop - It should not be neccessary the HW capability. If for some deployment it is decided to use a single DLD and single DLR intervals (even if the HW platform supports the port-level granularity) we should allow it. Like in QOS configuration where we have some configuration global but then it can be overriden with per-port/per-tunnel settings

orchagent/pfcwdhworch.cpp
```diff
@@ -0,0 +1,1336 @@
...
+                this->report_pfc_storm(notification.queue_id, port_id,
+                                      queue_index, port_alias, "");
+            }
+            else
+            {
+                SWSS_LOG_WARN("PFC deadlock DETECTED on queue 0x%" PRIx64 " (no port info)", notification.queue_id);
+            }
+
+            // Let SAI/SDK manage recovery automatically
+            notification.app_managed_recovery = false;
```
eddyk-nvidia, line 242: Is there any place where "true" is set ? I recall from HLD that "true" is supposed to trigger SAI_QUEUE_ATTR_PFC_DLR_INIT

Is this field set here later read by vendor SAI to understand which mode of work is expected ?

orchagent/switchorch.cpp
```diff
@@ -135,7 +136,7 @@ void SwitchOrch::set_switch_pfc_dlr_init_capability()
         m_PfcDlrInitEnable = false;
         fvVector.emplace_back(SWITCH_CAPABILITY_TABLE_PFC_DLR_INIT_CAPABLE, "false");
```
eddyk-nvidia, line 164: This SAI attribute might be supported ONLY if SAI_QUEUE_ATTR_PFC_DLR_INIT is used by the vendor SAI for triggering the recovery. It happens in case of "app_managed_recovery=true". When "app_managed_recovery=false" this SAI attribute might be not used. So, its check for capability query of PFC WD DLDR support is not ideal

eddyk-nvidia (changes requested):

orchagent/switchorch.cpp
abhishek-nexthop: @eddyk-nvidia This is a valid point. However, the current goal is to bring hardware-based recovery to parity with software-based recovery and support the workflows that exist today. We can track this separately by opening an issue and extend the logic later to support global-level configuration.

orchagent/pfcwdhworch.cpp
abhishek-nexthop: @eddyk-nvidia This is to ensure we do not enter the manual recovery workflow. Currently, only auto-recovery is supported, where both detection and recovery are handled by the hardware. SAI_QUEUE_ATTR_PFC_DLR_INIT is required only for manual recovery workflows, where it must be set to true for the packet action to take effect, otherwise, no action occurs. It also needs to be reset back to false to allow recovery to proceed.

orchagent/switchorch.cpp
abhishek-nexthop: @eddyk-nvidia @prsunny ACK. We are currently discussing the best way to handle this and plan to remove it accordingly.

orchagent/switchorch.cpp
abhishek-nexthop: @eddyk-nvidia This workflow is currently applicable only to Broadcom chips. For Broadcom, SAI_QUEUE_ATTR_PFC_DLR_INIT indicates whether the PFCWD hardware is enabled. The workflow can be extended in the future to support other chipsets based on their respective use cases.

orchagent/switchorch.cpp
eddyk-nvidia: @abhishek-nexthop - if you refer to SW-based recovery there is no such mechanism using DLD/R today in SONIC. So, parity is a but vague term here. The only PFC WD solution in SONIC today is using the Lua script and it is using the global timers (same for all ports/queues)

IMO - the design should support both Global and Per-Queue (overriding the global one) setting of intervals

orchagent/switchorch.cpp
eddyk-nvidia: @abhishek-nexthop - I am not sure it is the right approach. You mentioned earlier that SAI_QUEUE_ATTR_PFC_DLR_INIT is required only for "manual" triggering of recovery and now you say that it is the BRCM way to work. I think the design in generic SONIC should be vendor-agnostic

@kperumalbfn - FYI

orchagent/switchorch.cpp
abhishek-nexthop: > The only PFC WD solution in SONIC today is using the Lua script and it is using the global timers (same for all ports/queues)

@eddyk-nvidia Could we double-check this before we build on it? My understanding is that the software watchdog sets detection and restoration times per port rather than globally, is that not the case here? It seems like the only thing shared across all ports is the polling interval, which governs how often we check rather than the actual detect/restore timing. Could these two be getting mixed up?
Along the same lines, is there actually a way today to set a single global timer across all interfaces? As far as I can tell there's no command that does that, and the one command that applies to every port at once(start_default) doesn't take a timer at all. If that's right, can "parity with the software solution" really mean a global timer, given the software side never offered one?
If so, wouldn't a global timer on the hardware side be a new feature rather than parity? It might be worth framing it that way so we're aligned on expectations. Happy to go through it together if that would help.

cc: @kperumalbfn

orchagent/switchorch.cpp
eddyk-nvidia: @abhishek-nexthop - did you have any progress with changes related to this feature ?

noaOrMlnx: @abhishek-nexthop any update?

abhishek-nexthop: @eddyk-nvidia @noaOrMlnx We’re working with Broadcom to determine how to remove it. It won’t be part of the final PR. We’ll provide an update once we have a fix.

noaOrMlnx: thanks @abhishek-nexthop 
do you have an ETA for the fix?

arawat-nexthop:
@kperumalbfn @vmittal-msft 

Adding STATE_DB and `sairedis.rec` dumps as requested.

Captured on a Broadcom platform running a Nexthop build that carries this
implementation (`main.Nexthop.179117-fe96002c80`, `asic_type: broadcom`), on a port
with `pfc_enable: 3,4`. Config applied with
`pfcwd start --action drop Ethernet160 200 --restoration-time 200`.

### 1. Init: capability query and timer ranges

```
|g|SAI_OBJECT_TYPE_SWITCH:oid:0x21000000000000|SAI_SWITCH_ATTR_PFC_TC_DLD_INTERVAL_RANGE=0,0
|G|SAI_STATUS_SUCCESS|SAI_SWITCH_ATTR_PFC_TC_DLD_INTERVAL_RANGE=1,1500
|g|SAI_OBJECT_TYPE_SWITCH:oid:0x21000000000000|SAI_SWITCH_ATTR_PFC_TC_DLR_INTERVAL_RANGE=...
|G|SAI_STATUS_SUCCESS|SAI_SWITCH_ATTR_PFC_TC_DLR_INTERVAL_RANGE=100,1000
|q|attribute_capability|SAI_OBJECT_TYPE_SWITCH:...|ATTR_ID=SAI_SWITCH_ATTR_QUEUE_PFC_DEADLOCK_NOTIFY
|Q|attribute_capability|SAI_STATUS_SUCCESS|...|CREATE_IMP=true|SET_IMP=true|GET_IMP=false
|s|SAI_OBJECT_TYPE_SWITCH:oid:0x21000000000000|SAI_SWITCH_ATTR_QUEUE_PFC_DEADLOCK_NOTIFY=0x56101400f040
```

The ranges read back from the ASIC are what STATE_DB then advertises, and the
deadlock notification is only registered after the capability query reports it
settable.

### 2. STATE_DB, global scope

Before any port is configured:

```
STATE_DB PFC_WD_STATE_TABLE|PFC_WD
{ DLR_PACKET_ACTION: unknown, RECOVERY_MECHANISM: HARDWARE,
  DETECTION_TIME_MIN: 1, DETECTION_TIME_MAX: 1500,
  RESTORATION_TIME_MIN: 100, RESTORATION_TIME_MAX: 1000 }
```

`DETECTION_TIME_MIN/MAX` and `RESTORATION_TIME_MIN/MAX` match the SAI ranges above,
and are what the configuration path range-checks against.

### 3. Configuring a port

```
|s|SAI_OBJECT_TYPE_QUEUE:oid:0x1500000000030f|SAI_QUEUE_ATTR_ENABLE_PFC_DLDR=false
|s|SAI_OBJECT_TYPE_QUEUE:oid:0x15000000000310|SAI_QUEUE_ATTR_ENABLE_PFC_DLDR=false
|s|SAI_OBJECT_TYPE_PORT:oid:0x1000000000011|SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL={"count":0,"list":null}
|s|SAI_OBJECT_TYPE_PORT:oid:0x1000000000011|SAI_PORT_ATTR_PFC_TC_DLR_INTERVAL={"count":0,"list":null}
|s|SAI_OBJECT_TYPE_SWITCH:oid:0x21000000000000|SAI_SWITCH_ATTR_PFC_DLR_PACKET_ACTION=SAI_PACKET_ACTION_DROP
|s|SAI_OBJECT_TYPE_PORT:oid:0x1000000000011|SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL={"count":2,"list":[{"key":3,"value":200},{"key":4,"value":200}]}
|s|SAI_OBJECT_TYPE_PORT:oid:0x1000000000011|SAI_PORT_ATTR_PFC_TC_DLR_INTERVAL={"count":2,"list":[{"key":3,"value":200},{"key":4,"value":200}]}
|s|SAI_OBJECT_TYPE_QUEUE:oid:0x1500000000030f|SAI_QUEUE_ATTR_ENABLE_PFC_DLDR=true
|s|SAI_OBJECT_TYPE_QUEUE:oid:0x15000000000310|SAI_QUEUE_ATTR_ENABLE_PFC_DLDR=true
```

The watchdog is stopped before it is (re)started, the switch level packet action is
programmed once, the intervals are set only for the lossless TCs (3 and 4, matching
`pfc_enable`), and DLDR is enabled only on those two queues.

### 4. Read-back, and the resulting per-port STATE_DB

```
|g|...|SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL={"count":8,...}
|G|SAI_STATUS_SUCCESS|SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL={... {"key":3,"value":200},{"key":4,"value":200} ...}
|g|...|SAI_PORT_ATTR_PFC_TC_DLR_INTERVAL={"count":8,...}
|G|SAI_STATUS_SUCCESS|SAI_PORT_ATTR_PFC_TC_DLR_INTERVAL={... {"key":3,"value":200},{"key":4,"value":200} ...}
```

```
STATE_DB PFC_WD_HW_STATE|Ethernet160
{ recovery_type: hardware, status: configured, action: drop,
  configured_detection_time: 200,  hw_detection_time: 200,
  configured_restoration_time: 200, hw_restoration_time: 200 }
```

`hw_*` are the values read back from the ASIC rather than an echo of the request,
which is what lets an operator see when hardware quantises a requested value.
Global scope after configuration shows `DLR_PACKET_ACTION: drop`.

### 5. Teardown

```
|s|SAI_OBJECT_TYPE_QUEUE:oid:0x1500000000030f|SAI_QUEUE_ATTR_ENABLE_PFC_DLDR=false
|s|SAI_OBJECT_TYPE_QUEUE:oid:0x15000000000310|SAI_QUEUE_ATTR_ENABLE_PFC_DLDR=false
|s|SAI_OBJECT_TYPE_PORT:oid:0x1000000000011|SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL={"count":0,"list":null}
|s|SAI_OBJECT_TYPE_PORT:oid:0x1000000000011|SAI_PORT_ATTR_PFC_TC_DLR_INTERVAL={"count":0,"list":null}
```

After `pfcwd stop`, every `PFC_WD_HW_STATE|*` key is removed and the global scope
returns to `DLR_PACKET_ACTION: unknown`, so no per-port state and no switch level
action are left behind.

orchagent/pfcwdhworch.cpp
```diff
@@ -0,0 +1,975 @@
...
+extern PortsOrch *gPortsOrch;
+
+// Global instance pointer for SAI callback
+
+PfcWdHwOrch::PfcWdHwOrch(DBConnector *db, vector<string> &tableNames,
+                         const vector<sai_port_stat_t> &portStatIds,
+                         const vector<sai_queue_stat_t> &queueStatIds,
+                         const vector<sai_queue_attr_t> &queueAttrIds):
+    PfcWdBaseOrch(db, tableNames),
+    c_portStatIds(portStatIds),
```
vmittal-msft, line 25: These stat/attr id vectors (`c_portStatIds`/`c_queueStatIds`/`c_queueAttrIds`) are stored but never referenced anywhere else in the file, and the HW orch never registers a FlexCounter or writes the `PFC_WD` COUNTERS_DB group. If HW-mode counters aren't planned, please drop these constructor params + members (and the matching args passed in `orchdaemon.cpp`); otherwise this is where the registration is missing, and `show pfcwd stats` won't populate for the hardware path.

vmittal-msft (commented):
Thanks @abhishek-nexthop for the iterations — good to see the SKU whitelist replaced by the `SAI_QUEUE_ATTR_ENABLE_PFC_DLDR` capability query in `switchorch`, which addresses the earlier feedback from @kperumalbfn / @prsunny / @eddyk-nvidia. The single-threaded design (forwarding the SAI deadlock callback through ASIC_DB and consuming it on the orchagent thread) is clean and well-commented, the storm-transition de-dup via `m_stormedQueues` looks correct, and the teardown path is nicely defensive (shrunk lossless-TC set, unreadable PFC mask, action reset on last port). Test coverage is strong (25 cases incl. SAI-failure paths).

A few substantive items in addition to the still-open design discussion:

**Counters / stats parity (medium).** The constructor takes `portStatIds`/`queueStatIds`/`queueAttrIds` but they're stored and never used, and there's no FlexCounter or COUNTERS_DB registration anywhere in the HW orch. `report_pfc_storm`/`report_pfc_restored` only log + publish the event; they don't bump per-queue detected/restored counters. So `show pfcwd stats` will likely stay empty for the HW path, diverging from the software watchdog and from this PR's description ("FlexCounter registration, COUNTERS_DB initialization, TX/RX statistics, delta calculation"). Was the counter piece intentionally dropped? If so, please remove the dead constructor args/members and trim the description; if not, it needs wiring up.

**Virtual-switch default (please verify).** Because vs advertises DLDR, `pfcDldrEnable` is true by default, so `PfcWdHwOrch` is now instantiated on the virtual switch unless `PFC_DLR_INIT_ENABLE=0`. Existing software-PFCWD vs / sonic-mgmt tests that don't set that env may now exercise the HW path — worth confirming they still pass.

**Design thread still open.** I don't want to duplicate @eddyk-nvidia's thread — the global-vs-per-queue DLD/DLR interval question (and keeping config vendor-agnostic) still looks like the main gate, so I'm leaving this as a comment rather than a block.

Inline notes below. Nits: tabs vs spaces in `deleteEntry`/`startWdOnPort`, leftover empty "global instance pointer" comments, and a redundant `getPort` in `deleteEntry` (the base `deleteEntry` looks it up again).

orchagent/pfcwdhworch.cpp
```diff
@@ -0,0 +1,975 @@
...
+            else if (field == PFC_WD_ACTION)
+            {
+                action = deserializeAction(value);
+                if (action == PfcWdAction::PFC_WD_ACTION_UNKNOWN)
+                {
+                    SWSS_LOG_ERROR("Invalid PFC Watchdog action %s", value.c_str());
+                    return task_process_status::task_invalid_entry;
+                }
+            }
+            else if (field == PFC_STAT_HISTORY)
```
vmittal-msft, line 330: `pfc_stat_history` is parsed and validated here and passed into `startWdOnPort`, but then dropped — `configureHwWatchdog` never consumes it, so `pfc_stat_history=enable` silently has no effect. Please either implement it or reject it explicitly as unsupported so operators aren't misled.

orchagent/pfcwdhworch.cpp
```diff
@@ -0,0 +1,975 @@
...
+                                          queue_index, port_alias, "");
+                }
+            }
+            else
+            {
+                SWSS_LOG_WARN("PFC deadlock DETECTED on queue 0x%" PRIx64 " (no port info)", notification.queue_id);
+            }
+
+            // Let SAI/SDK manage recovery automatically
+            notification.app_managed_recovery = false;
```
vmittal-msft, line 242: This assignment has no effect: `notification` is a locally deserialized copy that `sai_deserialize_free_queue_deadlock_ntf` frees right after `onQueuePfcDeadlock` returns, and because the event was forwarded through the ASIC_DB NOTIFICATIONS channel the value can't propagate back to SAI/SDK. If hardware-managed recovery must be selected, it needs to be programmed via a SAI attribute at config time. As written it reads as though it controls recovery mode but doesn't — suggest removing it or replacing it with the real mechanism plus a clarifying comment.

orchagent/pfcwdhworch.cpp
```diff
@@ -0,0 +1,975 @@
...
+        return handleStartWdOnPortFailure(port);
+    }
+
+    clearPfcWdPending(port);
+    SWSS_LOG_NOTICE("Started PFC Watchdog on port %s", port.m_alias.c_str());
+    // Port is tracked in m_pfcwd_ports by configureHwWatchdog
+    return task_process_status::task_success;
+}
+
+task_process_status PfcWdHwOrch::deleteEntry(const string& key)
```
vmittal-msft, line 451: Unlike `PfcWdBaseOrch::deleteEntry`, this override doesn't short-circuit `PFC_WD|GLOBAL`, so a `DEL` of the GLOBAL key falls through to `getPort("GLOBAL")` and logs `SWSS_LOG_ERROR("Invalid port interface GLOBAL")` on a routine removal. `createEntry` already special-cases GLOBAL — please add `if (key == PFC_WD_GLOBAL) return task_process_status::task_success;` at the top here to match the base behavior.

orchagent/pfcwdhworch.cpp
arawat-nexthop: These will be used by the counters PR

orchagent/pfcwdhworch.cpp
arawat-nexthop: Addressed, went with warn-and-accept rather than rejecting the entry: pfc_stat_history can be set for all ports by a default config, and failing the entry there would disable the watchdog on those ports entirely. Now createEntry logs that it is not supported in hardware mode and is ignored, naming the port, so the config stays valid but nothing is silently swallowed. If you'd rather it be a hard rejection, that's a one-line change

arawat-nexthop:
16/18 checks pass, and the 2 red ones are the same failure (an Azure PublishTestResults EACCES on a root-owned container log file), not a code failure. The commit compiled clean under -Werror on six arch/toolchain combos, mock_tests were 1399 PASS / 0 FAIL, and test_pfcwd.py passed 10/10 first try including the three GLOBAL-delete no-crash tests.

vmittal-msft:
@arawat-nexthop can i review again if you handled all the comments ?

arawat-nexthop:
> @arawat-nexthop can i review again if you handled all the comments ?

I have taken care of all the comments, you can review again

PR #4419, opened 2026-03-31
fpmsyncd: Add RIB/FIB processing support

GaladrielZhao:
**What I did**
This commit introduces support for processing NextHopGroupFull messages with sonic-fib in fpmsyncd when RIB/FIB functionality is enabled.

The implementation adds a new device metadata configuration option "nhg_fib" which can be set to "enabled" to activate the RIB/FIB feature. When this option is enabled, fpmsyncd will handle the new netlink message types RTM_NEWNHGFIB and RTM_DELNHGFIB from zebra.

For decoding the NextHopGroupFull JSON data received from zebra, the code utilizes the sonic-fib library's from_json function to parse the JSON string and construct NextHopGroupFull objects. These objects are then forwarded to the NHG Manager for further processing.

The route handling logic has also been adapted to work correctly when RIB/FIB is enabled. When processing routes that reference next hop groups by ID, fpmsyncd now queries the NHG Manager's RIB NHG table to retrieve the corresponding RIBNHGEntry. This allows proper mapping between zebra's NHG IDs and SONiC's internal NHG IDs, and handles both single next-hop and multi-next-hop scenarios appropriately.

**Why I did it**
To support for processing NextHopGroupFull messages with sonic-fib in fpmsyncd when RIB/FIB functionality is enabled.

**How I verified it**
- PhoenixWing test
- UT will be added later


fpmsyncd/routesync.cpp
```diff
@@ -2638,6 +2689,85 @@ void RouteSync::onPicContextMsg(struct nlmsghdr *h, int len)
...
+            "Nexthop group without an ID received from the zebra");
+        return;
+    }
+
+    /* We use the ID as a key for nhg table */
+    id = *((uint32_t *)RTA_DATA(tb[NHA_ID]));
+
+    addr_family = nhm->nh_family;
+
+    if (nlmsg_type == RTM_NEWNEXTHOP)
```
selva-nexthop, line 2731: The branches here are checking `RTM_NEWNEXTHOP` / `RTM_DELNEXTHOP`, but we only call into this function for `RTM_NEWNHGFIB` / `RTM_DELNHGFIB`. As written those branches never match, so the NHG full messages would just get ignored. Can we switch these over to the FIB message types?

fpmsyncd/routesync.cpp
```diff
@@ -2199,34 +2213,71 @@ void RouteSync::onRouteMsg(int nlmsg_type, struct nl_object *obj, char *vrf)
...
-                fvw.weight = std::move(weights);
+            if (!entry)
+            {
+                SWSS_LOG_ERROR("NextHop group id %d not found in m_rib_nhg_table. Dropping the route %s", nhg_id, destipprefix);
+                return;
+            }
+            SWSS_LOG_INFO("Get NHG with id %d", nhg_id);
 
-            SWSS_LOG_DEBUG("NextHop group id %d is a single nexthop address. Filling the route table %s with nexthop and ifname", nhg_id, destipprefix);
+            sonic_nhg_id = entry->getSonicObjID();
```
selva-nexthop, line 2230: Small type issue — `sonic_nhg_id` is a `uint32_t`, but `getSonicObjID()` returns a `sonicObjectID`. I don't think this compiles as-is. Should this be `getSonicObjIDNum()` (or `.id`)?

fpmsyncd/routesync.cpp
```diff
@@ -2638,6 +2689,85 @@ void RouteSync::onPicContextMsg(struct nlmsghdr *h, int len)
...
+    if (nlmsg_type == RTM_NEWNEXTHOP)
+    {
+        SWSS_LOG_INFO("New nexthop group full message!");
+
+        /* Get NextHopGroupFull JSON string */
+        json_str = (char *)RTA_DATA(tb[NHA_JSON_STR]);
+        SWSS_LOG_NOTICE("Received NHGFULL %d JSON string: %s", id, json_str);
+
+        /* Conver JSON to NextHopGroupFull object */
+        nlohmann::ordered_json j = nlohmann::ordered_json::parse(json_str);
```
selva-nexthop, line 2740: We're parsing the JSON attribute here without first checking that `tb[NHA_JSON_STR]` is present, and without a try/catch around the parse. If zebra sends a bad or empty attribute, that can take fpmsyncd down. Can we add a guard?

fpmsyncd/routesync.cpp
```diff
@@ -2638,6 +2689,85 @@ void RouteSync::onPicContextMsg(struct nlmsghdr *h, int len)
...
+
+    addr_family = nhm->nh_family;
+
+    if (nlmsg_type == RTM_NEWNEXTHOP)
+    {
+        SWSS_LOG_INFO("New nexthop group full message!");
+
+        /* Get NextHopGroupFull JSON string */
+        json_str = (char *)RTA_DATA(tb[NHA_JSON_STR]);
+        SWSS_LOG_NOTICE("Received NHGFULL %d JSON string: %s", id, json_str);
```
selva-nexthop, line 2737: Logging the full JSON string at NOTICE on every NHG is going to get very noisy once we have scale. Would DEBUG be enough here?

PR #4420, opened 2026-03-31
[fpmsyncd]: Add Support for SRv6 processing based on RIB/FIB

GaladrielZhao:
**What I did**
This commit implements SRv6 VPN Route processing that supports the RIB/FIB project.
We've removed the PIC context message handling, because there is no need to use it after RIB/FIB's implementation.

**Why I did it**
To support SRv6 VPN Route processing for RIB/FIB.

**How I verified it**
- Phoenixwing test
- UT will be added later


fpmsyncd/fpmlink.h
```diff
@@ -17,10 +17,8 @@
 
 #define RTM_NEWSRV6LOCALSID		1000
 #define RTM_DELSRV6LOCALSID		1001
-#define RTM_NEWPICCONTEXT		2000
-#define RTM_DELPICCONTEXT		2001
-#define RTM_NEWSRV6VPNROUTE		3000
-#define RTM_DELSRV6VPNROUTE		3001
+#define RTM_NEWSRV6VPNROUTE		2000
```
selva-nexthop, line 20: One thing that stood out — on the fpmsyncd side we're moving `RTM_NEWSRV6VPNROUTE` from 3000 to 2000, but `dplane_fpm_sonic` still sends 3000 today (and 2000 is still used for PICCONTEXT over there). Unless the FRR side is changing at the same time, these won't match up. If PICCONTEXT is going away, I'd rather leave the VPN numbers alone and just drop the PIC ones. Is the renumber intentional?

fpmsyncd/routesync.cpp
```diff
@@ -2013,17 +2013,20 @@ void RouteSync::onMsgRaw(struct nlmsghdr *h)
         onNextHopMsg(h, len);
         return;
     }
-    if(h->nlmsg_type == RTM_NEWPICCONTEXT || h->nlmsg_type == RTM_DELPICCONTEXT)
+    if ((h->nlmsg_type == RTM_NEWSRV6VPNROUTE)
```
selva-nexthop, line 2016: We're routing SRv6 VPN through this path now and the old PICCONTEXT handling is gone. When `nhg_fib` is disabled, do we still need a fallback for SRv6 PIC, or is the plan that this path only exists with the new feature on?

PR #4421, opened 2026-03-31
[fpmsyncd]: Support fpmsyncd to use APP_STATE DB

GaladrielZhao:
**What I did**
- Introduce an additional RedisPipeline for APPL_STATE_DB in fpmsyncd.
- Add a NHG_FULL_STATE_TABLE (state-only) backed by this pipeline.
- No behavioral change to existing route/NHG handling.

</br>

**Why I did it**
In the upcoming RIBFIB module, we will use it to store NHG information pushed down by FRR. 
Since there is no consumer for this data, APPL_STATE_DB is a more appropriate place than APPL_DB.

</br>

**How I verified it**
- PR sanity: Updated existing fpmsyncd mock tests
- Phoenixwing test

### DIFF
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -150,7 +150,7 @@ static decltype(auto) makeNlAddr(const T& ip)
 }
 
 
-RouteSync::RouteSync(RedisPipeline *pipeline) :
+RouteSync::RouteSync(RedisPipeline *pipeline, RedisPipeline *app_state_pipeline) :
     // When the feature ORCH_NORTHBOND_ROUTE_ZMQ_ENABLED is enabled, route events must be sent to orchagent via the ZMQ channel.
     m_zmqClient(create_local_zmq_client(ORCH_NORTHBOND_ROUTE_ZMQ_ENABLED, false)),
     m_routeTable(createProducerStateTable(pipeline, APP_ROUTE_TABLE_NAME, true, m_zmqClient)),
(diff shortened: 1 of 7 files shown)


fpmsyncd/routesync.cpp
```diff
@@ -162,7 +162,9 @@ RouteSync::RouteSync(RedisPipeline *pipeline) :
     m_warmStartHelper(pipeline, m_routeTable.get(), APP_ROUTE_TABLE_NAME, "bgp", "bgp"),
     m_srv6MySidTable(pipeline, APP_SRV6_MY_SID_TABLE_NAME, true),
     m_srv6SidListTable(pipeline, APP_SRV6_SID_LIST_TABLE_NAME, true),
-    m_nl_sock(NULL), m_link_cache(NULL)
+    m_nl_sock(NULL), m_link_cache(NULL),
+    m_app_state_pipeline(app_state_pipeline),
+    m_nhgFullStateTable(app_state_pipeline, "NHG_FULL_STATE_TABLE", true)
```
selva-nexthop, line 167: Good to see the app_state pipeline wired in. One question — does NHGMgr / RIBNHGTable actually use this for warm restart reconciliation yet, or is that coming in a follow-up (or in #4395)?

fpmsyncd/routesync.h
```diff
@@ -219,7 +219,7 @@ class RouteSync : public NetMsg
 public:
     enum { MAX_ADDR_SIZE = 64 };
 
-    RouteSync(RedisPipeline *pipeline);
+    RouteSync(RedisPipeline *pipeline, RedisPipeline *app_state_pipeline);
```
selva-nexthop, line 222: `RouteSync` now takes two pipelines and there's no default. Looks like the in-tree call sites are updated — just flagging it in case anyone out of tree is constructing `RouteSync` directly.

PR #4424, opened 2026-03-31
Cascade PortChannel admin shutdown to member ports

### LINKED ISSUE #26494: Enhancement: Cascade PortChannel admin shutdown to member ports
### Is it platform specific

generic

### Importance or Severity

Medium

### Description of the enhancement

When a PortChannel (LAG) is administratively shut down (config interface shutdown <PortChannel>), the member links' transceivers remain physically active. The peer device must rely on LACP timeout (which could go upto 90s) to detect that the PortChannel is down, which introduces unnecessary delay. This is especially problematic in production environments where fast failover is critical. (A customer reported this as a gap)

When a LAG is admin-shut, the member ports shut go operationally down immediately such that the peer devices can detect the links going down and bring PC down.

### Current Behavior

When a PortChannel is admin-shut (config interface shutdown PortChannelXXXX):
- The PortChannel interface goes admin-down and oper-down
- Teamsyncd disables LAG member collection/distribution at the ASIC level
- However, the individual member ports remain physically UP - their transceivers stay active, lasers stay on, and SAI_PORT_ATTR_ADMIN_STATE remains true
- The peer device continues to see the physical links as UP and must wait for LACP timeout to bring its own PortChannel down
- show interfaces status shows member ports as Admin: Up, Oper: Up even though the parent PortChannel is shut

### Proposed Behavior

- When a PortChannel is admin-shut all member ports are cascaded admin-down at the SAI level, physically shutting the transceivers (laser off)
- When the PortChannel is admin-restored (no shutdown), member ports are restored to their individually configured admin state - a member that was already individually admin-down before the PortChannel shutdown stays down
- When a new member is added to an admin-down PortChannel, it is immediately forced admin-down
- When a member is removed from an admin-down PortChannel, its configured admin state is restored
- When an admin-down PortChannel is deleted, member ports are restored to their configured admin state

spandan-nexthop:
When a PortChannel is admin-shut, force all member ports admin-down. This allows the peer device to immediately detect link-down instead of waiting for LACP timeout.
On PortChannel admin-up (or member removal/LAG deletion), restore each member to its configured admin state from APP_DB.

**What I did**
Added `cascadeLagAdminStatusToMembers()` in `portsorch` which propagates LAG admin-down to member ports via `setPortAdminStatus()`.
Added a guard in `doPortTask()` to prevent APP_DB from overriding the forced state.
On LAG admin-up (or member removal/LAG deletion), each member's configured admin status is restored to the intended state.

**Why I did it**
When a PortChannel is admin-shut, its member links' transceivers remain physically active.
The peer device must rely on LACP timeout to detect the PortChannel is down, which introduces unnecessary delay (up to 90s with long timeout). 

**How I verified it**
- New DVS tests: `test_lag_admin_down_cascades_to_members`, `test_lag_deletion_restores_member_admin_state`
- On hardware: `config interface shutdown PortChannelXXXX` → verify member port LEDs go off and peer detects link-down immediately

**Details if related**
Fixes: https://github.com/sonic-net/sonic-buildimage/issues/26494

cc: @venkit-nexthop @kalash-nexthop

### DIFF
--- a/orchagent/port.h
+++ b/orchagent/port.h
@@ -212,6 +212,8 @@ public:
     sai_object_id_t     m_hif_id = 0;
     sai_object_id_t     m_lag_id = 0;
     sai_object_id_t     m_lag_member_id = 0;
+    /* PHY port admin state is overriden by parent LAG admin-down */
+    bool                m_lag_forced_admin_down = false;
     sai_object_id_t     m_tunnel_id = 0;
     sai_object_id_t     m_ingress_acl_table_group_id = 0;
     sai_object_id_t     m_egress_acl_table_group_id = 0;
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -5515,6 +5515,15 @@ void PortsOrch::doPortTask(Consumer &consumer)
                     pCfg.admin_status.value = admin_status;
                 }
 
+                /* This is a guard to suppress admin-up of port if parent LAG is admin-down */
+                if (pCfg.admin_status.is_set && pCfg.admin_status.value && p.m_lag_forced_admin_down)
+                {
+                    SWSS_LOG_NOTICE("Port %s: suppressing admin-up from APP_DB "
+                                    "(forced admin-down by parent LAG)",
+                                    p.m_alias.c_str());
+                    pCfg.admin_status.is_set = false;
+                }
+
                 /* Last step set port admin status */
                 if (pCfg.admin_status.is_set)
                 {
@@ -6076,6 +6085,7 @@ void PortsOrch::doLagTask(Consumer &consumer)
             int32_t switch_id = -1;
             string tpid_string;
             uint16_t tpid = 0;
+            string admin_status_str;
 
             for (auto i : kfvFieldsValues(t))
             {
@@ -6123,6 +6133,10 @@ void PortsOrch::doLagTask(Consumer &consumer)
                     tpid = (uint16_t)stoi(tpid_string, 0, 16);
                     SWSS_LOG_DEBUG("reading TPID string:%s to uint16: 0x%x", tpid_string.c_str(), tpid);
                  }
+                else if (fvField(i) == "admin_status")
+                {
+                    admin_status_str = fvValue(i);
+                }
             }
 
             if (table_name == CHASSIS_APP_LAG_TABLE_NAME)
@@ -6220,6 +6234,22 @@ void PortsOrch::doLagTask(Consumer &consumer)
                         SWSS_LOG_NOTICE("Saved to set port %s learn mode %s", alias.c_str(), learn_mode_str.c_str());
                     }
                 }
+
+                /* Detect admin status transition and propagate to members */
+                if (!admin_status_str.empty())
+                {
+                    bool newAdminUp = (admin_status_str == "up");
+
+                    if (l.m_admin_state_up != newAdminUp)
+                    {
+                        SWSS_LOG_NOTICE("LAG %s admin status changed to %s",
+                                        alias.c_str(),
+                                        newAdminUp ? "up" : "down");
+                        l.m_admin_state_up = newAdminUp;
+                        cascadeLagAdminStatusToMembers(l, newAdminUp);
+                        m_portList[alias] = l;
+                    }
+                }
             }
 
             it = consumer.m_toSync.erase(it);
@@ -6234,6 +6264,12 @@ void PortsOrch::doLagTask(Consumer &consumer)
                 continue;
             }
 
+            /* If LAG is admin-down, restore member states before deleting the LAG */
+            if (!lag.m_admin_state_up)
+            {
+                cascadeLagAdminStatusToMembers(lag, true);
+            }
+
             if (removeLag(lag))
                 it = consumer.m_toSync.erase(it);
             else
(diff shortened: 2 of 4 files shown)


prsunny:
I dont think this is an expected behavior. Parent portchannel 'admin' down doesn't mean individual members needs to be kept down. We may have usecase for shutting down the logical portchannel but the member ports (physical ports) have to be up from a datacenter deployment point of view.

bhouse-nexthop:
> I dont think this is an expected behavior. Parent portchannel 'admin' down doesn't mean individual members needs to be kept down. We may have usecase for shutting down the logical portchannel but the member ports (physical ports) have to be up from a datacenter deployment point of view.

@prsunny do you have any other suggestions for how to handle the root issue then?

prsunny:
> > I dont think this is an expected behavior. Parent portchannel 'admin' down doesn't mean individual members needs to be kept down. We may have usecase for shutting down the logical portchannel but the member ports (physical ports) have to be up from a datacenter deployment point of view.
> 
> @prsunny do you have any other suggestions for how to handle the root issue then?

@bhouse-nexthop , IMO, this has to be handled outside of orchagent. User/Controller doing this operation can explicitely trigger individual member shut and/or bgp shut.

prsunny (changes requested):
Suggest handle outside of orchagent.

venkit-nexthop:
> > > I dont think this is an expected behavior. Parent portchannel 'admin' down doesn't mean individual members needs to be kept down. We may have usecase for shutting down the logical portchannel but the member ports (physical ports) have to be up from a datacenter deployment point of view.
> > 
> > 
> > @prsunny do you have any other suggestions for how to handle the root issue then?
> 
> @bhouse-nexthop , IMO, this has to be handled outside of orchagent. User/Controller doing this operation can explicitely trigger individual member shut and/or bgp shut.

@prsunny maybe I am misunderstanding your concern. But this PR is only doing an admin-down of the member ports of the LAG when the operator admin-downs the LAG itself. If BGP session was going over the LAG port in question it will get affected anyways? When a member port is part of the LAG (aka port-channel), it wouldnt be carrying any other traffic outside of what is going through the LAG. So I am trying to understand your concern better.

prsunny:
> > > > I dont think this is an expected behavior. Parent portchannel 'admin' down doesn't mean individual members needs to be kept down. We may have usecase for shutting down the logical portchannel but the member ports (physical ports) have to be up from a datacenter deployment point of view.
> > > 
> > > 
> > > @prsunny do you have any other suggestions for how to handle the root issue then?
> > 
> > 
> > @bhouse-nexthop , IMO, this has to be handled outside of orchagent. User/Controller doing this operation can explicitely trigger individual member shut and/or bgp shut.
> 
> @prsunny maybe I am misunderstanding your concern. But this PR is only doing an admin-down of the member ports of the LAG when the operator admin-downs the LAG itself. If BGP session was going over the LAG port in question it will get affected anyways? When a member port is part of the LAG (aka port-channel), it wouldnt be carrying any other traffic outside of what is going through the LAG. So I am trying to understand your concern better.

Didn't quite get the point. What is the ultimate goal? if member ports have to be brought down when lag is down, for what exact usecase?

venkit-nexthop:
> > > > > I dont think this is an expected behavior. Parent portchannel 'admin' down doesn't mean individual members needs to be kept down. We may have usecase for shutting down the logical portchannel but the member ports (physical ports) have to be up from a datacenter deployment point of view.
> > > > 
> > > > 
> > > > @prsunny do you have any other suggestions for how to handle the root issue then?
> > > 
> > > 
> > > @bhouse-nexthop , IMO, this has to be handled outside of orchagent. User/Controller doing this operation can explicitely trigger individual member shut and/or bgp shut.
> > 
> > 
> > @prsunny maybe I am misunderstanding your concern. But this PR is only doing an admin-down of the member ports of the LAG when the operator admin-downs the LAG itself. If BGP session was going over the LAG port in question it will get affected anyways? When a member port is part of the LAG (aka port-channel), it wouldnt be carrying any other traffic outside of what is going through the LAG. So I am trying to understand your concern better.
> 
> Didn't quite get the point. What is the ultimate goal? if member ports have to be brought down when lag is down, for what exact usecase?

The usecase is that when PortChannel is admin shut on this device (sudo config interface shutdown <PortChannel>), its member links should also be automatically brought administratively down. This will help peer device to detect the link down and bring Portchannel down and not rely on LACP timeout.
What is the concern with this? Which use-case would break when the member ports are brought down?

kalash-nexthop:
> > > > > I dont think this is an expected behavior. Parent portchannel 'admin' down doesn't mean individual members needs to be kept down. We may have usecase for shutting down the logical portchannel but the member ports (physical ports) have to be up from a datacenter deployment point of view.
> > > > 
> > > > 
> > > > @prsunny do you have any other suggestions for how to handle the root issue then?
> > > 
> > > 
> > > @bhouse-nexthop , IMO, this has to be handled outside of orchagent. User/Controller doing this operation can explicitely trigger individual member shut and/or bgp shut.
> > 
> > 
> > @prsunny maybe I am misunderstanding your concern. But this PR is only doing an admin-down of the member ports of the LAG when the operator admin-downs the LAG itself. If BGP session was going over the LAG port in question it will get affected anyways? When a member port is part of the LAG (aka port-channel), it wouldnt be carrying any other traffic outside of what is going through the LAG. So I am trying to understand your concern better.
> 
> Didn't quite get the point. What is the ultimate goal? if member ports have to be brought down when lag is down, for what exact usecase?

FWIW, something similar is already done to handle MTU changes to port-channels. Port channel MTU value is forced down on member ports even if member ports have other values of their own configured in config_db. And when member ports leave LAG, only then their configured MTU is assigned back to them.

PR #4430, opened 2026-04-02
[fdborch] Fix VLAN FDB flush failure caused by missing VLAN prefix

haojun291:
Fixes #4428

**What I did**
Fix VLAN-based FDB flush in `fdborch`.
For VLAN flush requests, `data` contains the VLAN ID only (for example, `1000`), but `getPort()` expects the VLAN interface name format (`Vlan1000`).  
This change prepends `Vlan` to the VLAN ID before calling `getPort()` in the VLAN FDB flush path.

**Why I did it**
Before this change, running `fdbclear -v <vid>` could report success on CLI, but the dynamic MAC entry was not removed.  
Meanwhile, `orchagent` logged an error like:
```text
doTask: Get Port from vlan(1000) failed!
```

**How I verified it**
- Learned a dynamic MAC entry in VLAN 1000
- Ran `fdbclear -v 1000`
- Verified that before the fix the entry remained and `orchagent` logged `Get Port from vlan(1000) failed!`
- Verified that after the fix the entry was removed successfully and no such error was logged

**Details if related**

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -15,6 +15,8 @@
 #include "vxlanorch.h"
 #include "directory.h"
 
+#define VLAN_PREFIX         "Vlan"
+
 extern sai_fdb_api_t    *sai_fdb_api;
 
 extern sai_object_id_t  gSwitchId;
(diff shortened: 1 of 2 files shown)


prsunny:
@haojun291 , thanks for the fix. Can you add unit tests for coverage?

haojun291:
> @haojun291 , thanks for the fix. Can you add unit tests for coverage?

Thanks for the review. I'm looking into the existing test framework. I will update the PR once I am done.

haojun291:
> @haojun291 , thanks for the fix. Can you add unit tests for coverage?

Hi @prsunny, I have added the unit tests to cover both VLAN and PORTVLAN FDB flush in flush_syncd_notif_ut.cpp.

prsunny (approved):

PR #4437, opened 2026-04-03
[routeorch] Asynchronous route state publish  in db update thread.

vganesan-nokia:
**What I did**

Implemented asynchoronous update of route state publication which includes APPL_STATE_DB update, notifications and recordings. 

Route state updates (APPL_STATE_DB notification, APPL_STATE_DB state, response recording and db write recording) can be queued on the ResponsePublisher DB update thread instead of running synchronously on the routeorch's route programming hot path.

- Add publishAsync() and publishAsyncBatch() so full publish (notification + state DB write + recorder) runs on m_update_thread and m_ntf_pipe flush on flush entries also done in the worker thread to avoid cross-thread notifications wrongly flushed.
- publishAsync() adds all the route states to the batch without taking lock. After completing gRouteBulker.flus(), all the batched states are queued as single item to the worker queue. The lock is taken only during this queuing. The assumption is the batching of route states is done only in the RouteOrch.
- RouteOrch uses a dedicated ResponsePublisher and publishAsync() from publishRouteState(); override flushResponses() to flush it before the base implementation.
- This async route state publish  is disabled by default. To enable, a knob gRouteStateAsyncPublish is used and this can be enabled by config db "SYSTEM_DEFAULTS|async_rec:status" set to "enabled". Setting this adds orchagent option -A which enables async route state publish (also enabling async swss.rec recording)
- For recording, to ensuare correct timestamp, the timestamp is passed while enqueuing the state and RecWrite::record() method is called with additional paramter of timestamp. (Depends on sonic-swss PR https://github.com/sonic-net/sonic-swss/pull/4400)
- Adjusted mock ResponsePublisher for unit tests.

Ref: HLD: https://github.com/sonic-net/SONiC/pull/2285

**Why I did it**

To improve route convergence. Solution for Issue: https://github.com/sonic-net/sonic-swss/issues/4436

**How I verified it**

Advertised 250k routes from a neighbor, measured the time required for FIB programming of all 250k routes. The convergence is measured as per RFC7747 rib-in-convergence methodology. 

With suppress-fib-pending enabled, with async route state published, the average rib-in-convergence improvement is 45%+ (when compared with sync hot path route state publish)

**Details if related**

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -274,6 +274,35 @@ void getCfgSwitchType(DBConnector *cfgDb, string &switch_type, string &switch_su
 
 }
 
+/*
+ * DEVICE_METADATA|localhost route_state_async_publish: if value is "disabled", turn off async route state publish.
+ * Otherwise keep default gRouteStateAsyncPublish == true. Must run before OrchDaemon::init() constructs RouteOrch.
+ */
+void getCfgRouteStateAsyncPublish(DBConnector *cfgDb)
+{
+    Table cfgDeviceMetaDataTable(cfgDb, CFG_DEVICE_METADATA_TABLE_NAME);
+    string val;
+
+    try
+    {
+        if (!cfgDeviceMetaDataTable.hget("localhost", "route_state_async_publish", val))
+        {
+            return;
+        }
+    }
+    catch (const std::system_error &e)
+    {
+        SWSS_LOG_WARN("Could not read route_state_async_publish from CONFIG_DB: %s; default enabled", e.what());
+        return;
+    }
+
+    if (val == "disabled")
+    {
+        gRouteStateAsyncPublish = false;
+        SWSS_LOG_NOTICE("route_state_async_publish is disabled in DEVICE_METADATA|localhost; synchronous route state publish");
+    }
+}
+
 bool isChassisAppDbPresent()
 {
     std::ifstream file(SonicDBConfig::DEFAULT_SONIC_DB_CONFIG_FILE);
@@ -1016,6 +1045,8 @@ int main(int argc, char **argv)
         /* Initialize the ring before OrchDaemon initializing Orchs */
         orchDaemon->enableRingBuffer();
     }
+
+    getCfgRouteStateAsyncPublish(&config_db);
 
     if (!orchDaemon->init())
     {
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -275,31 +275,31 @@ void getCfgSwitchType(DBConnector *cfgDb, string &switch_type, string &switch_su
 }
 
 /*
- * DEVICE_METADATA|localhost route_state_async_publish: if value is "disabled", turn off async route state publish.
+ * SYSTEM_DEFAULTS|async_rec status: if value is "disabled", turn off async route state publish.
  * Otherwise keep default gRouteStateAsyncPublish == true. Must run before OrchDaemon::init() constructs RouteOrch.
  */
 void getCfgRouteStateAsyncPublish(DBConnector *cfgDb)
 {
-    Table cfgDeviceMetaDataTable(cfgDb, CFG_DEVICE_METADATA_TABLE_NAME);
+    Table cfgSystemDefaultsTable(cfgDb, CFG_SYSTEM_DEFAULTS_TABLE_NAME);
     string val;
 
     try
     {
-        if (!cfgDeviceMetaDataTable.hget("localhost", "route_state_async_publish", val))
+        if (!cfgSystemDefaultsTable.hget("async_rec", "status", val))
         {
             return;
         }
     }
     catch (const std::system_error &e)
     {
-        SWSS_LOG_WARN("Could not read route_state_async_publish from CONFIG_DB: %s; default enabled", e.what());
+        SWSS_LOG_WARN("Could not read SYSTEM_DEFAULTS|async_rec status from CONFIG_DB: %s; default enabled", e.what());
         return;
     }
 
     if (val == "disabled")
     {
         gRouteStateAsyncPublish = false;
-        SWSS_LOG_NOTICE("route_state_async_publish is disabled in DEVICE_METADATA|localhost; synchronous route state publish");
+        SWSS_LOG_NOTICE("SYSTEM_DEFAULTS|async_rec status is disabled; synchronous route state publish");
     }
 }
 
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -80,6 +80,7 @@ uint32_t gCfgSystemPorts = 0;
 string gMyHostName = "";
 string gMyAsicName = "";
 bool gTraditionalFlexCounter = false;
+bool gRouteStateAsyncPublish = true;
 uint32_t create_switch_timeout = 0;
 bool gMultiAsicVoq = false;
 
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -80,7 +80,7 @@ uint32_t gCfgSystemPorts = 0;
 string gMyHostName = "";
 string gMyAsicName = "";
 bool gTraditionalFlexCounter = false;
-bool gRouteStateAsyncPublish = true;
+bool gRouteStateAsyncPublish = false;
 uint32_t create_switch_timeout = 0;
 bool gMultiAsicVoq = false;
 
@@ -104,7 +104,7 @@ void usage()
     cout << "    -b batch_size: set consumer table pop operation batch size (default 128)" << endl;
     cout << "    -m MAC: set switch MAC address" << endl;
     cout << "    -i INST_ID: set the ASIC instance_id in multi-asic platform" << endl;
-    cout << "    -A: enable async swss.rec recording path" << endl;
+    cout << "    -A: enable async swss.rec recording and async route state publish path" << endl;
     cout << "    -s enable synchronous mode (deprecated, use -z)" << endl;
     cout << "    -z redis communication mode (redis_async|redis_sync|zmq_sync), default: redis_async" << endl;
     cout << "    -f swss_rec_filename: swss record log filename(default 'swss.rec')" << endl;
@@ -272,35 +272,6 @@ void getCfgSwitchType(DBConnector *cfgDb, string &switch_type, string &switch_su
         SWSS_LOG_ERROR("System error in parsing switch subtype: %s", e.what());
     }
 
-}
-
-/*
- * SYSTEM_DEFAULTS|async_rec status: if value is "disabled", turn off async route state publish.
- * Otherwise keep default gRouteStateAsyncPublish == true. Must run before OrchDaemon::init() constructs RouteOrch.
- */
-void getCfgRouteStateAsyncPublish(DBConnector *cfgDb)
-{
-    Table cfgSystemDefaultsTable(cfgDb, CFG_SYSTEM_DEFAULTS_TABLE_NAME);
-    string val;
-
-    try
-    {
-        if (!cfgSystemDefaultsTable.hget("async_rec", "status", val))
-        {
-            return;
-        }
-    }
-    catch (const std::system_error &e)
-    {
-        SWSS_LOG_WARN("Could not read SYSTEM_DEFAULTS|async_rec status from CONFIG_DB: %s; default enabled", e.what());
-        return;
-    }
-
-    if (val == "disabled")
-    {
-        gRouteStateAsyncPublish = false;
-        SWSS_LOG_NOTICE("SYSTEM_DEFAULTS|async_rec status is disabled; synchronous route state publish");
-    }
 }
 
 bool isChassisAppDbPresent()
@@ -532,7 +503,8 @@ int main(int argc, char **argv)
             break;
         case 'A':
             Recorder::Instance().swss.setAsync(true);
-            SWSS_LOG_NOTICE("Async swss recorder enabled");
+            gRouteStateAsyncPublish = true;
+            SWSS_LOG_NOTICE("Async swss recorder and async route state publish enabled");
             break;
         case 'd':
             record_location = optarg;
@@ -1045,8 +1017,6 @@ int main(int argc, char **argv)
         /* Initialize the ring before OrchDaemon initializing Orchs */
         orchDaemon->enableRingBuffer();
     }
-
-    getCfgRouteStateAsyncPublish(&config_db);
 
     if (!orchDaemon->init())
     {
--- a/orchagent/response_publisher.cpp
+++ b/orchagent/response_publisher.cpp
@@ -4,6 +4,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include "timestamp.h"
 
 namespace
 {
@@ -29,7 +30,7 @@ std::string PrependedComponent(const ReturnCode &status)
 }
 
 void RecordDBWrite(const std::string &table, const std::string &key, const std::vector<swss::FieldValueTuple> &attrs,
-                   const std::string &op)
+                   const std::string &op, const std::string *record_ts = nullptr)
 {
     if (!swss::Recorder::Instance().respub.isRecord())
     {
@@ -42,11 +43,19 @@ void RecordDBWrite(const std::string &table, const std::string &key, const std::
         s += "|" + fvField(attr) + ":" + fvValue(attr);
     }
 
-    swss::Recorder::Instance().respub.record(s);
+    if (record_ts != nullptr && !record_ts->empty())
+    {
+        swss::Recorder::Instance().respub.record(*record_ts, s);
+    }
+    else
+    {
+        swss::Recorder::Instance().respub.record(s);
+    }
 }
 
 void RecordResponse(const std::string &response_channel, const std::string &key,
-                    const std::vector<swss::FieldValueTuple> &attrs, const std::string &status)
+                    const std::vector<swss::FieldValueTuple> &attrs, const std::string &status,
+                    const std::string *record_ts = nullptr)
 {
     if (!swss::Recorder::Instance().respub.isRecord())
     {
@@ -59,7 +68,14 @@ void RecordResponse(const std::string &response_channel, const std::string &key,
         s += "|" + fvField(attr) + ":" + fvValue(attr);
     }
 
-    swss::Recorder::Instance().respub.record(s);
+    if (record_ts != nullptr && !record_ts->empty())
+    {
+        swss::Recorder::Instance().respub.record(*record_ts, s);
+    }
+    else
+    {
+        swss::Recorder::Instance().respub.record(s);
+    }
 }
 
 } // namespace
@@ -149,6 +165,61 @@ void ResponsePublisher::publish(const std::string &table, const std::string &key
     publish(table, key, intent_attrs, status, state_attrs, replace);
 }
 
+void ResponsePublisher::setAsyncFullPublish(bool enable)
+{
+    m_async_full_publish = enable;
+}
+
+void ResponsePublisher::publishAsync(const std::string &table, const std::string &key,
+                                     const std::vector<swss::FieldValueTuple> &intent_attrs, const ReturnCode &status,
+                                     bool replace)
+{
+    if (m_update_thread == nullptr)
+    {
+        publish(table, key, intent_attrs, status, replace);
+        return;
+    }
+    {
+        std::lock_guard<std::mutex> lock(m_lock);
+        std::string ts = swss::getTimestamp();
+        entry e(table, key, std::vector<swss::FieldValueTuple>{}, "", replace, /*flush=*/false, /*shutdown=*/false);
+        e.fullPublish = true;
+        e.intent_attrs = intent_attrs;
+        e.status = status;
+        e.record_ts = std::move(ts);
+        m_queue.push(std::move(e));
+    }
+    m_signal.notify_one();
+}
+
+void ResponsePublisher::publishFullFromThread(const std::string &table, const std::string &key,
+                                              const std::vector<swss::FieldValueTuple> &intent_attrs,
+                                              const ReturnCode &status, bool replace, const std::string &record_ts)
+{
+    std::vector<swss::FieldValueTuple> state_attrs;
+    if (status.ok())
+    {
+        state_attrs = intent_attrs;
+    }
+    std::string response_channel = "APPL_DB_" + table + "_RESPONSE_CHANNEL";
+    swss::NotificationProducer notificationProducer{m_ntf_pipe.get(), response_channel, m_buffered};
+
+    auto intent_attrs_copy = intent_attrs;
+    swss::FieldValueTuple err_str("err_str", PrependedComponent(status) + status.message());
+    intent_attrs_copy.insert(intent_attrs_copy.begin(), err_str);
+    notificationProducer.send(status.codeStr(), key, intent_attrs_copy);
+    // Same enqueue-time stamp for both lines (parity with sync publish recorder timing).
+    const std::string *ts_ptr = record_ts.empty() ? nullptr : &record_ts;
+    RecordResponse(response_channel, key, intent_attrs_copy, status.codeStr(), ts_ptr);
+
+    if ((intent_attrs.size() && state_attrs.size()) || (status.ok() && !intent_attrs.size()))
+    {
+        std::string op = intent_attrs.size() ? SET_COMMAND : DEL_COMMAND;
+        writeToDBInternal(table, key, state_attrs, op, replace);
+        RecordDBWrite(table, key, state_attrs, op, ts_ptr);
+    }
+}
+
 void ResponsePublisher::writeToDB(const std::string &table, const std::string &key,
                                   const std::vector<swss::FieldValueTuple> &values, const std::string &op, bool replace)
 {
@@ -229,19 +300,24 @@ void ResponsePublisher::flush() {
   } else {
     m_ntf_pipe->flush();
   }
-  if (m_update_thread != nullptr)
-  {
-      {
-          std::lock_guard<std::mutex> lock(m_lock);
-          m_queue.emplace(/*table=*/"", /*key=*/"", /*values =*/std::vector<swss::FieldValueTuple>{}, /*op=*/"",
-                          /*replace=*/false, /*flush=*/true, /*shutdown=*/false);
-      }
-      m_signal.notify_one();
-  }
-  else
-  {
-      m_db_pipe->flush();
-  }
+    if (m_update_thread != nullptr)
+    {
+        if (!m_async_full_publish)
+        {
+            m_ntf_pipe->flush();
+        }
+        {
+            std::lock_guard<std::mutex> lock(m_lock);
+            m_queue.emplace(/*table=*/"", /*key=*/"", /*values =*/std::vector<swss::FieldValueTuple>{}, /*op=*/"",
+                            /*replace=*/false, /*flush=*/true, /*shutdown=*/false);
+        }
+        m_signal.notify_one();
+    }
+    else
+    {
+        m_ntf_pipe->flush();
+        m_db_pipe->flush();
+    }
 }
 
 void ResponsePublisher::setBuffered(bool buffered)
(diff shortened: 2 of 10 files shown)


prsunny:
@mint570 for viz

deepak-singhal0408:
@venkit-nexthop for viz.

orchagent/response_publisher.cpp
```diff
@@ -232,6 +303,10 @@ void ResponsePublisher::flush(bool warmboot) {
   }
     if (m_update_thread != nullptr)
     {
```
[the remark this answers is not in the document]
vganesan-nokia: Fixed.

orchagent/response_publisher.cpp
```diff
@@ -146,6 +162,61 @@ void ResponsePublisher::publish(const std::string &table, const std::string &key
...
+        std::lock_guard<std::mutex> lock(m_lock);
+        std::string ts = swss::getTimestamp();
+        entry e(table, key, std::vector<swss::FieldValueTuple>{}, "", replace, /*flush=*/false, /*shutdown=*/false);
+        e.fullPublish = true;
+        e.intent_attrs = intent_attrs;
+        e.status = status;
+        e.record_ts = std::move(ts);
+        m_queue.push(std::move(e));
+    }
+    m_signal.notify_one();
```
vivekkarri, line 215: Currently, RouteOrchs batch the updates upto the batch size (1024 by default) and then it calls flush() on the redis notification consumer pipe

From the current change, that ability seems to be removed, IMO there should be a batch AsyncPublish.

orchagent/response_publisher.cpp
```diff
@@ -241,15 +316,18 @@ void ResponsePublisher::flush(bool warmboot) {
...
+  }
 }
 
 void ResponsePublisher::setBuffered(bool buffered)
 {
     m_buffered = buffered;
 }
 
+// Runs on m_update_thread (response publisher db update thread).
 void ResponsePublisher::dbUpdateThread()
```
vivekkarri: Tis is no longer a async dbUpdateThread, worth renaming IMO

orchagent/routeorch.cpp
```diff
@@ -3198,7 +3200,7 @@ void RouteOrch::publishRouteState(const RouteBulkContext& ctx, const ReturnCode&
 
     const bool replace = false;
 
-    m_publisher.publish(APP_ROUTE_TABLE_NAME, ctx.key, fvs, status, replace);
+    m_publisher.publishAsync(APP_ROUTE_TABLE_NAME, ctx.key, fvs, status, replace);
```
vivekkarri: Can you benchmark the results, preferably on a device with weak CPU. 

I think there are tests that measure how long route updates are taking like test_bgp_update_timer_single_route

Please make sure to run with suppress-fib-pending pending enabled

vivekkarri:
@prsunny IMO This fix is similar on the lines of RingBuffer added here. 
https://github.com/sonic-net/sonic-swss/pull/3242 

Async write to swss record in other PR is definitely useful

@vganesan-nokia Can you check if this is solving anything extra that is not solved by having a Ring Buffer

orchagent/response_publisher.cpp
vganesan-nokia: The batching ability is not removed but there might be slight difference in number of entries flushed in a flush. With async publish, the flush is queued (the sequence of flush marking is same orcdaemon flush->route orch flush -> response publisher flush  and it is queued instead of flushing immediately) after processing the number of entries as per the batch size. But after flush marked, if the main thread is interleaved (might occure under heavy load) before the update thread is executed to drain the queue, and routeorch::dotask() is called, there is a chance some additional entries might be in the redis piple line.

orchagent/response_publisher.cpp
vganesan-nokia: Yes, this async thread is for  db update and also for notification and recording. We'll rename to stateUpdateThread.

orchagent/routeorch.cpp
vganesan-nokia: We tested with both suppress-fib-pending enabled and disabled.
**System:** AMD EPYC   8-Core  2.5Ghz    16GDRAM
**Test setting**: APPL_DB async update (ZMQ northbound),  async ASIC DB update (w/ South bound ZMQ) , swss.rec async write; default bulk and batch size;  250k IPv4 routes; prefix len 24; Convergence time measured using RFC7747 rib-in-convergence test guidelines with 0% tolerance.
**Results**
With suppress-fib-pending disabled: 23.94 seconds
With suppress-fib-pending enabled: 24.33 seconds
About 1.6% less with suppress-fib-pending is enabled when compared to convergence with suppress-fib -pending disabled.
More test results will be added [here](https://github.com/deepak-singhal0408/SONiC/tree/route-download-performance-strategies/doc/route-performance/benchmarks)

vganesan-nokia:
> @prsunny IMO This fix is similar on the lines of RingBuffer added here. #3242
> 
> Async write to swss record in other PR is definitely useful
> 
> @vganesan-nokia Can you check if this is solving anything extra that is not solved by having a Ring Buffer

(1) We want to enable ZMQ northbound and south bound. (Ref: [SONiC Route Download Performance: Large-Scale Datacenter CLOS Fabric Strategy](https://github.com/deepak-singhal0408/SONiC/blob/route-download-performance-strategies/doc/route-performance/strategies/large-scale-datacenter-clos-fabric.md#6-long-term-why-path-a-zmq-is-the-recommended-direction)  With north bound ZMQ enabled, the ring buffer is not effective. The addToSync() and drain() are still executed in the main thread but not in the ring buffer thread.
(2) Even if we make ring buffer effective with Zmq enabled, the high latency adding redis access for db write and notification and file i/o access for recording are still in the route throughput path.

Ring buffer only solves the problem of the select() is not available enough in orchdaemon main thread i.e., it only solves the problem of scheduling of consumer execute/doTask vs main thread select() but not the problem of high latency by redis access.

Splitting the redis access from the route programming hot path gives mugh higher performance gain. By async route publish the response publish which involves redis access is split from the route programming. This is extra and is not solved by having ring buffer.

orchagent/response_publisher.cpp
```diff
@@ -42,11 +43,19 @@ void RecordDBWrite(const std::string &table, const std::string &key, const std::
...
 
-    swss::Recorder::Instance().respub.record(s);
+    if (record_ts != nullptr && !record_ts->empty())
+    {
+        swss::Recorder::Instance().respub.record(*record_ts, s);
+    }
+    else
+    {
+        swss::Recorder::Instance().respub.record(s);
+    }
```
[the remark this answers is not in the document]
vganesan-nokia: Depends on PR #4400. Once this PR is merged, this will be avaiable.

orchagent/routeorch.cpp
vivekkarri: I meant the measurements with and without this change (with suppress-fib-pending enabled)

vivekkarri:
> > @prsunny IMO This fix is similar on the lines of RingBuffer added here. #3242
> > Async write to swss record in other PR is definitely useful
> > @vganesan-nokia Can you check if this is solving anything extra that is not solved by having a Ring Buffer
> 
> (1) We want to enable ZMQ northbound and south bound. (Ref: [SONiC Route Download Performance: Large-Scale Datacenter CLOS Fabric Strategy](https://github.com/deepak-singhal0408/SONiC/blob/route-download-performance-strategies/doc/route-performance/strategies/large-scale-datacenter-clos-fabric.md#6-long-term-why-path-a-zmq-is-the-recommended-direction) With north bound ZMQ enabled, the ring buffer is not effective. The addToSync() and drain() are still executed in the main thread but not in the ring buffer thread. (2) Even if we make ring buffer effective with Zmq enabled, the high latency adding redis access for db write and notification and file i/o access for recording are still in the route throughput path.
> 
> Ring buffer only solves the problem of the select() is not available enough in orchdaemon main thread i.e., it only solves the problem of scheduling of consumer execute/doTask vs main thread select() but not the problem of high latency by redis access.
> 
> Splitting the redis access from the route programming hot path gives mugh higher performance gain. By async route publish the response publish which involves redis access is split from the route programming. This is extra and is not solved by having ring buffer.

I see, Thanks for the clarification. I see a ResponseThread in the HLD. https://github.com/a114j0y/SONiC/blob/db2b6a77b22a475581b58d42929c26c813df3d8d/doc/bgp_loading_optimization/bgp-loading-optimization-hld.md

Thus i thought something was added in the Response path as well

orchagent/response_publisher.cpp
vivekkarri: Yes, that's the problem. Under load main thread will be fight with writer thread for the same lock. 

RouteOrch knows, it has to flush responses to K entries, that is why we have RedisPipeline added instead of DBConnector one redis call in the end. 

We should leverage that knowledge even  with async updates, we should batch all the updates and in the end, a flush() should send a notification to writer thread to dump everything.

vganesan-nokia: > Yes, that's the problem. Under load main thread will be fight with writer thread for the same lock.
> 
> RouteOrch knows, it has to flush responses to K entries, that is why we have RedisPipeline added instead of DBConnector one redis call in the end.
> 
> We should leverage that knowledge even with async updates, we should batch all the updates and in the end, a flush() should send a notification to writer thread to dump everything.

This problem should have been there with the sync publsh flow also. The number of entries flushed is not always guaranteed to be equal to the K entries that RouteOrch knows to flush in any doTask() invocation. Depending on when the orchdaemon flush() is called the main thread, the pipleline may have more than K. The flush() is done periodically at the cadence of select() but the pipleline is loaded to K entries during the RouteOrch::doTask(). For example, if flush() is called after two RouteOrch::doTask(), then there is chance the pipleline may have entries from two batches (K1 + K2)

IMHO, the problem mentioned is an existing old problem. The async publish still retains the batching in the same way how it was before but it just moves where the flush happens (from manin thread to update thread).

vivekrnv: `flush()` is explicitly called in `doTask() `after RouteOrch finished processing the SAI updates. Change was recently made: https://github.com/sonic-net/sonic-swss/pull/4389. The fix was made to reduce the 1 sec delay in waiting for `OrchDaemon::select()`

With Async, we of course don't have this 1 sec wait problem. 

However, in case of async when there are K updates, in the case of current design, both the RouterOrch and Updater would fight to acquire lock for each read/write. That IMHO is inefficient, When the daemon has knowledge that there are K entries, it should do batchAsyncUpdate.

vganesan-nokia: @vivekrnv, thanks for the clarification. Yes. Agree that main thread and update thread contesting to acquire lock is inefficient. I'll implement batched async publish.

vganesan-nokia: Implemented publishAsyncBatch() to address the inefficiency in contenstion acquiring lock by main thread and sate update thread.

orchagent/routeorch.cpp
vganesan-nokia: With suppresss-fib-pending enabled:

**System:** AMD EPYC 8-Core 2.5Ghz 16GDRAM
**Test setting**: APPL_DB async update (ZMQ northbound), async ASIC DB update (ZMQ southbound) , swss.rec async write ; default bulk and batch size; 250k IPv4 routes; prefix len 24; Convergence time measured using RFC7747 rib-in-convergence test guidelines with 5% tolerance.
**Results**:
With async route state publish **_disabled_**: 33.47 seconds (7469 routes/second)
With async route state publish **_enabled_**: 22.7 seconds (11013 routes/second) (about 47.44% improvement)

orchagent/response_publisher.cpp
vganesan-nokia: @vivekrnv, implemented batched async route state publish and added bench mark results with these changes. Please review.

vganesan-nokia:
@deepak-singhal0408 would you please help restrating azp run? The build error I am seeing (link errors) is not related to the changes of this PR.

deepak-singhal0408:
@vganesan-nokia add more UT? code coverage check is failing..

vganesan-nokia:
Hi @deepak-singhal0408, would you please help restarting the azp run? Thanks in advance for your help.

orchagent/response_publisher.h
```diff
@@ -52,6 +52,20 @@ class ResponsePublisher : public ResponsePublisherInterface
...
+
+    void publishAsync(const std::string &table, const std::string &key,
+                      const std::vector<swss::FieldValueTuple> &intent_attrs, const ReturnCode &status,
+                      bool replace = false);
+
+    // When true and the response publisher db update thread is used, all notifications for this publisher 
+    // are sent from that thread (publishAsync path). flush() then flushes the notification pipeline on the 
+    // response publisher db update thread as well, avoiding concurrent use of the notification RedisPipeline 
+    // from two threads.
+    void setAsyncFullPublish(bool enable);
```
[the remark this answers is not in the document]
vganesan-nokia: Added unit tests for code coverage.

deepak-singhal0408 (dismissed):
overall LGTM..

One observation: with setAsyncFullPublish(true), flush() becomes non-blocking — it enqueues a flush marker to the worker and returns before fpmsyncd actually receives the notifications. This means there's a brief window where notifications are in-flight on the worker thread but not yet delivered to Redis.

Acceptable for production — if orchagent crashes in that window, the swss→bgp cascade restart replays all state

vganesan-nokia:
@deepak-singhal0408, all checks have passed. would you please help getting more reviews/approval and possibly getting this merged? Thanks.

deepak-singhal0408:
> @deepak-singhal0408, all checks have passed. would you please help getting more reviews/approval and possibly getting this merged? Thanks.

@prsunny could you help on this? has been reviewed by Vivek Earlier. his comments are addressed, Changes LGTM as well.
cc @prabhataravind as well in case there are any other comments.

deepak-singhal0408:
Hi @vivekrnv could you check the latest diff to see if all your comments are addressed? if so, please sign-off as well? thanks,

deepak-singhal0408:
@vganesan-nokia, please resolve merge conflict. thanks,

prsunny:
@vganesan-nokia , could you provide a timestamp measurement and update in description?

1. Time to program 6k, 16k and 32k routes with and without this async channel
2. Time to update 6k, 16k and 32k routes in APPL_STATE_DB with and without this async channel

vganesan-nokia:
@prsunny, I can get the convergence time for **programming** 6k, 16k and 32k routes with and without async publish. For both cases (with and without async publish) the routes will be in APPL_STATE_DB. 
Would you please clarify for the second data? By "update" do you mean re-sending the same routes with some different attributes while these routes are already in the DUT?

prsunny:
> @prsunny, I can get the convergence time for **programming** 6k, 16k and 32k routes with and without async publish. For both cases (with and without async publish) the routes will be in APPL_STATE_DB. Would you please clarify for the second data? By "update" do you mean re-sending the same routes with some different attributes while these routes are already in the DUT?

Basically looking for the delay in APPL_STATE_DB with the async write vs inline write

vganesan-nokia:
Here is the convergence times for programming 6k, 16k and 32k routes with and without this async route publish.
**Without async route publish (route publish inline):**
6k: 0.913 s
16k: 2.17 s
32k: 4.2 s
**With async route publish enabled**
6k: 0.62 s
16k: 1.503 s
32k: 2.74 s

deepak-singhal0408:
> Here is the convergence times for programming 6k, 16k and 32k routes with and without this async route publish. **Without async route publish (route publish inline):** 6k: 0.913 s 16k: 2.17 s 32k: 4.2 s **With async route publish enabled** 6k: 0.62 s 16k: 1.503 s 32k: 2.74 s

@vganesan-nokia, the ask here is to check the APPL_STATE_DB write timestamps before the change and after (if you could provide the comparison data for 1st entry and last entry for 6k, 16k and 32k routes case)

orchagent/routeorch.h
```diff
@@ -313,6 +317,9 @@ class RouteOrch : public ZmqRouteOrch, public Subject
     EntityBulker<sai_mpls_api_t>            gLabelRouteBulker;
     ObjectBulker<sai_next_hop_group_api_t>  gNextHopGroupMemberBulker;
 
+    // Shadows Orch::m_publisher: APPL_STATE_DB + optional async thread for route state publish (publishAsync).
```
prabhataravind: With the shadowing, anyone operating on Orch m_publisher silently bypasses the RouteOrch's async publisher. It looks like by design, but consider using a different name.

orchagent/response_publisher.cpp
```diff
@@ -149,6 +173,100 @@ void ResponsePublisher::publish(const std::string &table, const std::string &key
...
+    item.table = table;
+    item.key = key;
+    item.intent_attrs = intent_attrs;
+    item.status = status;
+    item.replace = replace;
+    item.record_ts = swss::getTimestamp();
+    m_async_publish_pending.push_back(std::move(item));
+}
+
+void ResponsePublisher::publishFullBatchFromThread(std::vector<asyncPublishItem> &&items)
```
prabhataravind, line 237: A lot of the core logic overlaps with publish(). If publish() changes, the async path must be updated in tandem. Can we extract common logic to a private helper?

orchagent/routeorch.cpp
```diff
@@ -3271,3 +3276,9 @@ inline void RouteOrch::removeVipRouteSubnetDecapTerm(const IpPrefix &ipPrefix)
     m_appTunnelDecapTermProducer.del(key);
     m_SubnetDecapTermsCreated.erase(it);
 }
+
+void RouteOrch::flushResponses()
```
prabhataravind, line 3284: Add m_publisher.publishAsyncBatch() here before flush?

vganesan-nokia:
> > Here is the convergence times for programming 6k, 16k and 32k routes with and without this async route publish. **Without async route publish (route publish inline):** 6k: 0.913 s 16k: 2.17 s 32k: 4.2 s **With async route publish enabled** 6k: 0.62 s 16k: 1.503 s 32k: 2.74 s
> 
> @vganesan-nokia, the ask here is to check the APPL_STATE_DB write timestamps before the change and after (if you could provide the comparison data for 1st entry and last entry for 6k, 16k and 32k routes case)

@deepak-singhal0408, thanks for the clarification. The entries written to the APPL_STATE_DB do not have write stamps. Would you please let me know which write timestamp are you referring to?

orchagent/response_publisher.cpp
vganesan-nokia: Agreed. I'll make the publish write logic as common function and call from both publish and async publish.

orchagent/routeorch.cpp
vganesan-nokia: Yes. I"ll fix this.

orchagent/routeorch.h
vganesan-nokia: Yes. Agreed. I'll rename the publisher in the routeorch appropriately to avoid chance of bypassing async publisher.

orchagent/main.cpp
```diff
@@ -275,6 +276,35 @@ void getCfgSwitchType(DBConnector *cfgDb, string &switch_type, string &switch_su
...
+ * Otherwise keep default gRouteStateAsyncPublish == true. Must run before OrchDaemon::init() constructs RouteOrch.
+ */
+void getCfgRouteStateAsyncPublish(DBConnector *cfgDb)
+{
+    Table cfgDeviceMetaDataTable(cfgDb, CFG_DEVICE_METADATA_TABLE_NAME);
+    string val;
+
+    try
+    {
+        if (!cfgDeviceMetaDataTable.hget("localhost", "route_state_async_publish", val))
```
deepak-singhal0408: @vganesan-nokia, we are consolidating these config knobs (as they are too many and difficult to manage and really not METADATA of the Device) and also moving out of DEVICE_METADATA to SYSTEMS_DEFAULT.. Could you adjust your PR accordingly..
 
Please refer https://github.com/sonic-net/sonic-buildimage/pull/27674 and related PRs.

vganesan-nokia: Sure @deepak-singhal0408, I'll do the update for this and push the changes along with the changes to incorporate the review comments given above.

orchagent/routeorch.h
vganesan-nokia: Fixed

orchagent/response_publisher.cpp
vganesan-nokia: Fixed

orchagent/routeorch.cpp
vganesan-nokia: Fixed

orchagent/main.cpp

vganesan-nokia:
> > Here is the convergence times for programming 6k, 16k and 32k routes with and without this async route publish. **Without async route publish (route publish inline):** 6k: 0.913 s 16k: 2.17 s 32k: 4.2 s **With async route publish enabled** 6k: 0.62 s 16k: 1.503 s 32k: 2.74 s
> 
> @vganesan-nokia, the ask here is to check the APPL_STATE_DB write timestamps before the change and after (if you could provide the comparison data for 1st entry and last entry for 6k, 16k and 32k routes case)

Since the redis records do not have write timestamps, I used the timestamps from responsepublisher.rec. Here is the ts info. 

Write timestamps | | | |
 --- | --- | --- | --- |
  | First entry | Last entry | Delta (sec)
APPL_STATE_DB   write w/ inline route state publish |   |   |   |
6k | 2026-06-05.11:29:45.390310 | 2026-06-05.11:29:46.401781 | 1.011471
16k | 2026-06-05.11:32:21.378812 | 2026-06-05.11:32:23.908471 | 2.529659
32k | 2026-06-05.11:36:00.962471 | 2026-06-05.11:36:06.009874 | 5.047403
  |   |   |  
APPL_STATE_DB   write w/ async route state publish |   |   |   | 
6k | 2026-06-05.04:21:48.045484 | 2026-06-05.04:21:48.663988 | 0.618504
16k | 2026-06-05.04:41:44.306897 | 2026-06-05.04:41:45.772176 | 1.465279
32k | 2026-06-05.04:54:56.391418 | 2026-06-05.04:54:58.996115 | 2.604697

vganesan-nokia:
@deepak-singhal0408, would you please re-review and approve? Thanks.

orchagent/main.cpp
```diff
@@ -80,6 +80,7 @@ uint32_t gCfgSystemPorts = 0;
 string gMyHostName = "";
 string gMyAsicName = "";
 bool gTraditionalFlexCounter = false;
+bool gRouteStateAsyncPublish = true;
```
deepak-singhal0408: **Finding 1 (HIGH/DESIGN — blocking):** Knob design conflicts with the established consolidation pattern.

**Problem:** This PR adds `gRouteStateAsyncPublish = true` (default **enabled**) and a `getCfgRouteStateAsyncPublish()` that reads `SYSTEM_DEFAULTS|async_rec` in C++. This conflicts with PR #4630 + sonic-buildimage #27674 where:
- `orchagent.sh` already reads `SYSTEM_DEFAULTS|async_rec`
- If `== "enabled"` → passes `-A` flag to orchagent
- Default when key absent = **disabled** (opt-in)

**Issues:**
1. **Polarity inverted** — existing pattern defaults to disabled (opt-in). This defaults to enabled (opt-out). A fresh install with no CONFIG_DB entry → async route state ON but async swss.rec OFF — inconsistent.
2. **Redundant CONFIG_DB access** — shell already reads this knob.
3. **Two systems controlling one knob** — fragile.

**Recommended fix:** Drop `getCfgRouteStateAsyncPublish()`. Extend the existing `-A` flag:

```cpp
// main.cpp getopt loop:
case 'A':
    Recorder::Instance().swss.setAsync(true);  // existing
    gRouteStateAsyncPublish = true;             // new: async APPL_STATE_DB
    break;
```

Default `gRouteStateAsyncPublish` to `false`. One knob → one shell read → one flag → both effects.

orchagent/response_publisher.cpp
```diff
@@ -220,36 +329,49 @@ void ResponsePublisher::writeToDBInternal(const std::string &table, const std::s
...
-      m_zmqServer->sendMsg("APPL_DB", response.first, response.second);
+void ResponsePublisher::flush()
+{
+    if (m_update_thread != nullptr)
+    {
+        // When m_async_full_publish, only the worker may use m_ntf_pipe (batch publish);
+        // flushing it here would race with the worker and can drop state/notification Redis ops.
+        if (!m_async_full_publish)
+        {
+            m_ntf_pipe->flush();
```
deepak-singhal0408: **Finding 2 (MEDIUM):** `m_ntf_pipe` shared between threads when `m_async_full_publish=false`.

When `m_async_full_publish=false` but `db_write_thread=true`:
- Worker thread processes `fullPublishBatch` entries → `publishWrite()` → creates `NotificationProducer` with `m_ntf_pipe`
- Main thread arrives here → `m_ntf_pipe->flush()`

These run concurrently. `RedisPipeline` is not thread-safe.

**Mitigated for RouteOrch** (always calls `setAsyncFullPublish(true)` at routeorch.cpp:61), but the API is fragile for future consumers.

**Suggestion:** Either force `m_async_full_publish=true` when `db_write_thread=true` (constructor assert), or add a prominent comment in the header documenting the requirement.

orchagent/response_publisher.cpp
```diff
@@ -149,6 +192,72 @@ void ResponsePublisher::publish(const std::string &table, const std::string &key
     publish(table, key, intent_attrs, status, state_attrs, replace);
```
deepak-singhal0408, line 194: **Finding 3 (LOW):** `responses` map thread ownership assumption.

When `m_async_full_publish=true` + ZMQ, `responses` is only accessed from the worker thread (safe). But if synchronous `publish()` were called on this same instance from the main thread, it would also write to `responses` — creating a race.

Currently safe because RouteOrch uses a dedicated `m_routeStatePublisher`. Consider adding a comment or assert that synchronous `publish()` must not be called on an instance using `m_async_full_publish=true`.

deepak-singhal0408 (changes requested):
Good work on the async batch mechanism and threading model. The `publishAsync` → `publishAsyncBatch` → worker thread design minimizes lock contention effectively.

However, the knob design conflicts with the established consolidation pattern (PR #4630 + sonic-buildimage #27674). See inline comments for details.

deepak-singhal0408:
**Finding 4 (LOW/DOC):** PR description references `DEVICE_METADATA|localhost:route_state_publish_async` but the code uses `SYSTEM_DEFAULTS|async_rec` (the correct consolidated key from PR #4630). Please update the PR description to reflect the current implementation.

**Finding 5 (LOW):** No test exercises the synchronous fallback path (`gRouteStateAsyncPublish=false`). The fallback at response_publisher.cpp:225-228 is trivial, but a test would improve confidence in the disable path.

deepak-singhal0408:
also @vganesan-nokia please resolve comments that are already addressed,

orchagent/main.cpp
vganesan-nokia: Fixed.

orchagent/response_publisher.cpp
vganesan-nokia: Fixed

orchagent/response_publisher.cpp
vganesan-nokia: Fixed. Added comment

vganesan-nokia:
> **Finding 4 (LOW/DOC):** PR description references `DEVICE_METADATA|localhost:route_state_publish_async` but the code uses `SYSTEM_DEFAULTS|async_rec` (the correct consolidated key from PR #4630). Please update the PR description to reflect the current implementation.
> 
> **Finding 5 (LOW):** No test exercises the synchronous fallback path (`gRouteStateAsyncPublish=false`). The fallback at response_publisher.cpp:225-228 is trivial, but a test would improve confidence in the disable path.

Fixed both

vganesan-nokia:
> also @vganesan-nokia please resolve comments that are already addressed,

Done.

deepak-singhal0408 (approved):
LGTM

deepak-singhal0408:
@prabhataravind @prsunny could you help re-review this? the async APPL_STATE_DB functionality is disabled by default now and part of "SYSTEM_DEFAULTS|async_rec" knob.

ResponsePublisher Rec Timestamps have been posted as well  above for both(w/o async writing and with async writing).

prabhataravind (approved):

prsunny:
@mint570 , @dgsudharsan for viz

PR #4451, opened 2026-04-07
Runtime knob for Southbound ZMQ

inder-nexthop:
#### Why I did it
Enable ZMQ communication between orchagent and syncd (southbound).

#### How I did it
Added -g flag to orchagent to specify the context GUID `gContextGuid`. This tells orchagent which context should use ZMQ for communication with syncd.

Updated saihelper.cpp to set `SAI_REDIS_SWITCH_ATTR_ZMQ_ENABLED_GUID` when both ZMQ mode (-z flag) and a specific GUID (-g flag) are provided. This configures the SAI Redis layer to use ZMQ for the specified context.

#### How to verify it
Ran route benchmark test with both configurations to confirm:

- Default behavior (knob disabled) remains unchanged
- ZMQ mode activates correctly when knob is enabled
- Routes program successfully through the full pipeline in both modes

#### Related PRs:

https://github.com/sonic-net/sonic-sairedis/pull/1835
https://github.com/sonic-net/sonic-buildimage/pull/26638

### DIFF
--- a/orchagent/main.cpp
+++ b/orchagent/main.cpp
@@ -62,6 +62,9 @@ bool gRingMode = false;
 bool gSyncMode = false;
 sai_redis_communication_mode_t gRedisCommunicationMode = SAI_REDIS_COMMUNICATION_MODE_REDIS_ASYNC;
 string gAsicInstance;
+// Context guid used to select a specific context from context_config.json
+// Set via -g command line flag. UINT32_MAX means "not set" and applies to all contexts
+uint32_t gContextGuid = UINT32_MAX;
 
 extern bool gIsNatSupported;
 
@@ -113,6 +116,7 @@ void usage()
     cout << "    -c counter mode (traditional|asic_db), default: asic_db" << endl;
     cout << "    -t Override create switch timeout, in sec" << endl;
     cout << "    -v vrf: VRF name (default empty)" << endl;
+    cout << "    -g guid: context guid (default: all contexts)" << endl;
     cout << "    -I heart_beat_interval: Heart beat interval in millisecond (default 10)" << endl;
     cout << "    -R enable the ring thread feature" << endl;
     cout << "    -M enable SAI MACSec POST" << endl;
@@ -400,7 +404,7 @@ int main(int argc, char **argv)
     // Disable SAI MACSec POST by default. Use option -M to enable it.
     bool macsec_post_enabled = false;
 
-    while ((opt = getopt(argc, argv, "b:m:r:f:j:d:i:hsz:k:q:c:t:v:I:RD:M")) != -1)
+    while ((opt = getopt(argc, argv, "b:m:r:f:j:d:i:hsz:k:q:c:t:v:g:I:RD:M")) != -1)
     {
         switch (opt)
         {
@@ -522,6 +526,10 @@ int main(int argc, char **argv)
             macsec_post_enabled = true;
             break;
         case 'D': { gFlexCounterDelaySec = swss::to_int<int>(optarg); } break;
+        case 'g':
+            gContextGuid = (uint32_t)std::stoul(optarg);
+            SWSS_LOG_NOTICE("Context guid set to: %u", gContextGuid);
+            break;
         default: /* '?' */
             exit(EXIT_FAILURE);
         }
--- a/orchagent/saihelper.cpp
+++ b/orchagent/saihelper.cpp
@@ -106,6 +106,7 @@ extern sai_object_id_t gSwitchId;
 extern bool gTraditionalFlexCounter;
 extern bool gSyncMode;
 extern sai_redis_communication_mode_t gRedisCommunicationMode;
+extern uint32_t gContextGuid;
 extern event_handle_t g_events_handle;
 extern bool gOrchUnhealthy;
 extern string gSaiErrorString;
@@ -348,6 +349,21 @@ void initSaiRedis()
 
     sai_attribute_t attr;
     sai_status_t status;
+    // Set ZMQ enabled guid if ZMQ mode and a specific guid is requested
+    if (gContextGuid != UINT32_MAX &&
+        gRedisCommunicationMode == SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC)
+    {
+        attr.id = SAI_REDIS_SWITCH_ATTR_ZMQ_ENABLED_GUID;
+        attr.value.u32 = gContextGuid;
+
+        status = sai_switch_api->set_switch_attribute(gSwitchId, &attr);
+        if (status != SAI_STATUS_SUCCESS)
+        {
+            SWSS_LOG_ERROR("Failed to set ZMQ enabled guid: rv:%d", status);
+            return handleSaiFailure(SAI_API_SWITCH, "set", status, true);
+        }
+        SWSS_LOG_NOTICE("ZMQ enabled guid set successfully to %u", gContextGuid);
+    }
 
     attr.id = SAI_REDIS_SWITCH_ATTR_REDIS_COMMUNICATION_MODE;
     attr.value.s32 = gRedisCommunicationMode;
--- a/tests/mock_tests/mock_orchagent_main.cpp
+++ b/tests/mock_tests/mock_orchagent_main.cpp
@@ -4,6 +4,7 @@ extern "C" {
 }
 
 #include "orchdaemon.h"
+#include <cstdint>
 
 /* Global variables */
 sai_object_id_t gVirtualRouterId;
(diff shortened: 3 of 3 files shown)


inder-nexthop:
@deepak-singhal0408 Southbound ZMQ config knob PR is ready for review.

prsunny:
@inder-nexthop , would you check on this - https://github.com/sonic-net/sonic-swss/pull/3547 and associated PRs. @prabhataravind to review. Not fully aligned on this PR.

prsunny:
@dgsudharsan , @vivekrnv for viz

venkit-nexthop:
> @inder-nexthop , would you check on this - #3547 and associated PRs. @prabhataravind to review. Not fully aligned on this PR.

@prsunny Inder and I synced up with @deepak-singhal0408 on this. We will take a look at the PR you pointed to.
But the background is that we are working on improving route download speed and the southbound ZMQ is giving us some good perf boost. This PR is to enable/disable it via a config knob.

vivekrnv:
Please note that notifications will not work with ZMQ enabled https://github.com/sonic-net/sonic-swss/issues/1145

deepak-singhal0408:
> > @inder-nexthop , would you check on this - #3547 and associated PRs. @prabhataravind to review. Not fully aligned on this PR.
> 
> @prsunny Inder and I synced up with @deepak-singhal0408 on this. We will take a look at the PR you pointed to. But the background is that we are working on improving route download speed and the southbound ZMQ is giving us some good perf boost. This PR is to enable/disable it via a config knob.

Thanks @prsunny for highlighting the concern.. yes, as @venkit-nexthop mentioned, we see decent gains with southbound zmq, and would like to enable this for routes etc..

however, I dont think I am fully aligned on the approach..

Fundamentally, I consider zmq as Infra/PI piece, and coupling this with platforms/HWSKUs creates duplication/complex code.

I was going through of this PR set and the ASIC_DB persistence PRs (https://github.com/sonic-net/sonic-sairedis/pull/1801, https://github.com/sonic-net/sonic-sairedis/pull/1819). I think there's an opportunity to significantly simplify the design. Wanted to share my thoughts.. up for discussion.


Core observation: The current design couples three independent decisions into one trigger:

   1. ZMQ transport ON/OFF
   2. ASIC_DB persistence ON/OFF
   3. context_config.json required

On DPU, coupling these made sense — DPUs don't need ASIC_DB, and they already ship context_config.json. But for non-DPU switches, this coupling creates the problems we're solving with  -g, ASIC_DB restore PRs, and per-hwsku JSON files.. we have bunch of platforms which dont ship with context_config.json.. Also this PR set had a need to add context_config.json for vs..(JFYI, in the past, we had built some similar platform dependency in Chassis for Midplane-IPs and later had to move it out of platform specific code)..

 Proposed simplification (4 changes):

  1. ContextConfig.cpp — Change defaults from ipc:///tmp/zmq_ep to tcp://127.0.0.1:5555. Every platform
   that ships context_config.json already uses TCP localhost. The IPC default looks like dead code (please correct me if I am wrong). This eliminates the need for context_config.json just to get ZMQ transport working.
  
2. Syncd.cpp — Tighten the DisabledRedisClient guard from if (zmq_enable && !VS) to if 
  (zmq_enable && !VS && isDpu). Non-DPU switches get ZMQ transport with ASIC_DB preserved — tooling 
  (routeCheck, SNMP, show commands) works immediately without any ASIC_DB restore mechanism. To also 
  remove syncd's Redis write from the hot path, above PRs for async persistence is still valuable but 
  becomes a pure performance optimization, not a correctness requirement. Much cleaner scope for those PRs.

 3. orchagent.sh / syncd_init_common.sh — Read a CONFIG_DB knob (zmq_southbound or similar). If enabled, pass -z zmq_sync. Keep the existing elif switch_type == dpu as fallback for backward compatibility — DPU vendors can migrate to the CONFIG_DB knob at their own pace.

   4. No -g flag needed — I assume this is to keep gearbox communication channel via redis?  if so, Gearbox is naturally handled by process separation (syncd gets -z, gbsyncd doesn't — they're separate processes with independent command lines). Orchagent still reads context_config.json for gearbox topology (guids), and selectively enables zmq for the required context.

Result: Any switch  — can enable ZMQ with one CONFIG_DB key. No context_config.json needed. No ASIC_DB breakage. DPUs continue working unchanged. And https://github.com/sonic-net/sonic-sairedis/pull/1801, https://github.com/sonic-net/sonic-sairedis/pull/1819 can focus purely on async perf optimization rather than correctness restoration.

let me know what you think.. @prabhataravind @prsunny @venkit-nexthop @inder-nexthop 
I will setup a call to sync on this offline as well..

inder-nexthop:
> > > @inder-nexthop , would you check on this - #3547 and associated PRs. @prabhataravind to review. Not fully aligned on this PR.
> > 
> > 
> > @prsunny Inder and I synced up with @deepak-singhal0408 on this. We will take a look at the PR you pointed to. But the background is that we are working on improving route download speed and the southbound ZMQ is giving us some good perf boost. This PR is to enable/disable it via a config knob.
> 
> Thanks @prsunny for highlighting the concern.. yes, as @venkit-nexthop mentioned, we see decent gains with southbound zmq, and would like to enable this for routes etc..
> 
> however, I dont think I am fully aligned on the approach..
> 
> Fundamentally, I consider zmq as Infra/PI piece, and coupling this with platforms/HWSKUs creates duplication/complex code.
> 
> I was going through of this PR set and the ASIC_DB persistence PRs ([sonic-net/sonic-sairedis#1801](https://github.com/sonic-net/sonic-sairedis/pull/1801), [sonic-net/sonic-sairedis#1819](https://github.com/sonic-net/sonic-sairedis/pull/1819)). I think there's an opportunity to significantly simplify the design. Wanted to share my thoughts.. up for discussion.
> 
> Core observation: The current design couples three independent decisions into one trigger:
> 
> 1. ZMQ transport ON/OFF
> 2. ASIC_DB persistence ON/OFF
> 3. context_config.json required
> 
> On DPU, coupling these made sense — DPUs don't need ASIC_DB, and they already ship context_config.json. But for non-DPU switches, this coupling creates the problems we're solving with -g, ASIC_DB restore PRs, and per-hwsku JSON files.. we have bunch of platforms which dont ship with context_config.json.. Also this PR set had a need to add context_config.json for vs..(JFYI, in the past, we had built some similar platform dependency in Chassis for Midplane-IPs and later had to move it out of platform specific code)..
> 
> Proposed simplification (4 changes):
> 
> 1. ContextConfig.cpp — Change defaults from ipc:///tmp/zmq_ep to tcp://127.0.0.1:5555. Every platform
>    that ships context_config.json already uses TCP localhost. The IPC default looks like dead code (please correct me if I am wrong). This eliminates the need for context_config.json just to get ZMQ transport working.
> 2. Syncd.cpp — Tighten the DisabledRedisClient guard from if (zmq_enable && !VS) to if
>    (zmq_enable && !VS && isDpu). Non-DPU switches get ZMQ transport with ASIC_DB preserved — tooling
>    (routeCheck, SNMP, show commands) works immediately without any ASIC_DB restore mechanism. To also
>    remove syncd's Redis write from the hot path, above PRs for async persistence is still valuable but
>    becomes a pure performance optimization, not a correctness requirement. Much cleaner scope for those PRs.
> 3. orchagent.sh / syncd_init_common.sh — Read a CONFIG_DB knob (zmq_southbound or similar). If enabled, pass -z zmq_sync. Keep the existing elif switch_type == dpu as fallback for backward compatibility — DPU vendors can migrate to the CONFIG_DB knob at their own pace.
> 4. No -g flag needed — I assume this is to keep gearbox communication channel via redis?  if so, Gearbox is naturally handled by process separation (syncd gets -z, gbsyncd doesn't — they're separate processes with independent command lines). Orchagent still reads context_config.json for gearbox topology (guids), and selectively enables zmq for the required context.
> 
> Result: Any switch — can enable ZMQ with one CONFIG_DB key. No context_config.json needed. No ASIC_DB breakage. DPUs continue working unchanged. And [sonic-net/sonic-sairedis#1801](https://github.com/sonic-net/sonic-sairedis/pull/1801), [sonic-net/sonic-sairedis#1819](https://github.com/sonic-net/sonic-sairedis/pull/1819) can focus purely on async perf optimization rather than correctness restoration.
> 
> let me know what you think.. @prabhataravind @prsunny @venkit-nexthop @inder-nexthop I will setup a call to sync on this offline as well..

@deepak-singhal0408 Thanks for the detailed feedback, I agree with the direction. Decoupling ZMQ transport from ASIC_DB persistence will simplify things significantly. I'm working on the proposed changes and will post an updated diff.

We are more than happy to jump on a call!

CC @venkit-nexthop

PR #4461, opened 2026-04-09
Clean up wcmp manager watchport code to align with multicast fallback group.     In P4Orch wcmp manager, update internal prune status together with SAI calls.     Process port_state_change notification in p4orch after portsorch updated its cache to avoid race condition.

ksravani-hcl:
**What I did**

**Why I did it**

**How I verified it**
verified with UT cases.

**Details if related**

### DIFF
--- a/orchagent/p4orch/tests/wcmp_manager_test.cpp
+++ b/orchagent/p4orch/tests/wcmp_manager_test.cpp
@@ -255,36 +255,37 @@ class WcmpManagerTest : public ::testing::Test
       return wcmp_group_manager_->updateWcmpGroups(entries);
     }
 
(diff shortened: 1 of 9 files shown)


bibhuprasad-hcl (approved):

orchagent/p4orch/tests/wcmp_manager_test.cpp
yunang-c, line 1: The original CL has a VerifyStateAsicDbWithWatchportEventTest, is this test not upstreamable?

yunang-c (approved):

StephenWangGoogle (approved):

PR #4469, opened 2026-04-11
Handle orchagent exceptions without crashing

prabhataravind:
**What I did**
 * Add a catchall for exceptions in orchagent similar to how it is done for Orch2::doTask().
 * Made exception error log string consistent for easier detection and parsing.

**Why I did it**
Unhandled exceptions can cause orchagent to terminate which is not necessary.

**How I verified it**
By adding mocktests to verify exception handling functionality.

**Details if related**

### DIFF
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -125,7 +125,7 @@ Orch::Orch(DBConnector *db, const vector<table_name_with_pri_t> &tableNames_with
     }
 }
 
-Orch::Orch(const vector<TableConnector>& tables)ca
+Orch::Orch(const vector<TableConnector>& tables)
 {
     for (auto it : tables)
     {
--- a/orchagent/orch.cpp
+++ b/orchagent/orch.cpp
@@ -125,7 +125,7 @@ Orch::Orch(DBConnector *db, const vector<table_name_with_pri_t> &tableNames_with
     }
 }
 
-Orch::Orch(const vector<TableConnector>& tables)
+Orch::Orch(const vector<TableConnector>& tables)ca
 {
     for (auto it : tables)
     {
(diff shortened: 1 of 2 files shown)


orchagent/orch.cpp
```diff
@@ -1157,19 +1205,23 @@ void Orch2::doTask(Consumer &consumer)
         }
         catch (const std::invalid_argument& e)
         {
-            SWSS_LOG_ERROR("Parse error in %s: %s", typeid(*this).name(), e.what());
```
prsunny, line 1160: @prabhataravind , since we are not really adding anything new to be logged, suggest keeping the existing syslogs and copy it over to orch class as well. I'm not sure if someone may have alerting on these logs

prabhataravind: @prsunny I thought about this but then realized it would be useful to have one standard signature to alert on for all such orchagent exceptions. The current exception handling logs in Orch2 are all distinct and not following a pattern that can be easily grepped on.

prsunny: ok.

orchagent/orch.cpp
```diff
@@ -1157,19 +1205,23 @@ void Orch2::doTask(Consumer &consumer)
...
         catch (const std::logic_error& e)
         {
-            SWSS_LOG_ERROR("Logic error in %s: %s", typeid(*this).name(), e.what());
+            SWSS_LOG_ERROR("Exception caught: type=logic_error, orch=%s, error=%s",
+                           typeid(*this).name(), e.what());
         }
         catch (const std::exception& e)
         {
-            SWSS_LOG_ERROR("Exception was caught in the request parser in %s: %s", typeid(*this).name(), e.what());
+            SWSS_LOG_ERROR("Exception caught: type=std::exception, orch=%s, error=%s",
```
prsunny: can we have the type consistent? you can remove `std::`

tests/mock_tests/consumer_ut.cpp
```diff
@@ -29,6 +29,84 @@ namespace consumer_test
...
+            switch (m_throwType)
+            {
+                case ThrowType::InvalidArgument:
+                    throw std::invalid_argument("retryToSync invalid argument");
+                case ThrowType::LogicError:
+                    throw std::logic_error("retryToSync logic error");
+                case ThrowType::RuntimeError:
+                    throw std::runtime_error("retryToSync runtime error");
+                case ThrowType::UnknownException:
+                    throw 42;
```
prsunny: what is this magic number. could you define this?

orchagent/orch.cpp
prabhataravind: Fixed

tests/mock_tests/consumer_ut.cpp
prabhataravind: It was just meant to be some non-standard exception for test. I replaced it with a named constant.

PR #4473, opened 2026-04-13
[P4Orch] Remove group member mapper check in multicast group validation.

bibhuprasad-hcl:
**What I did**
1.Allow mcast ritf to be created with CPU port.
2.Move bridge port create and delete from mcast manager to portsorch.
3.Disallow empty multicast group to be program in P4Orch.
4.Update APPL_DB schema for multicast fallback group.

**Why I did it**

**How I verified it**
Verified with UT cases.

**Details if related**


orchagent/portsorch.cpp
```diff
@@ -5945,7 +5948,7 @@ void PortsOrch::doVlanMemberTask(Consumer &consumer)
             {
                 if (removeVlanMember(vlan, port))
                 {
-                    if (m_portVlanMember[port.m_alias].empty())
+		    if (getBridgePortReferenceCount(port) == 0)
```
mint570, line 5951: Indent

orchagent/portsorch.cpp
```diff
@@ -10993,6 +11002,10 @@ bool PortsOrch::addSystemPorts()
             {
                 m_port_ref_count[port.m_alias] = 0;
             }
+	    if(m_bridge_port_ref_count.find(port.m_alias) == m_bridge_port_ref_count.end())
```
mint570, line 11005: Indent

mint570 (approved):

PR #4475, opened 2026-04-13
[aclorch] Disable combined mirror V6 table for Tomahawk6 platforms

dgodwin-nokia:
Add isSeparateMirrorV6TablePlatform() to check the device platform name from CONFIG_DB DEVICE_METADATA and disable the combined mirror V6 table for platforms that require separate V4/V6 mirror ACL tables. Currently this applies to Nokia IXR7220-H6 (Tomahawk6) platforms.

**Why I did it**

Currently, combined everflow IPv4/IPv6 table creation fails on TH6 platform. This is due to the qualifier set being too large for TH6 field processor. There is currently an open Broadcom CSP (Case CS00012433094) to address this. For now, to pass everflow tests, tell orchagent to create seperate IPv4 and IPv6 mirror tables similar to broadcom DNX platform.

**How I verified it**

Validate that everflow tests now pass.

**Details if related**

### DIFF
--- a/orchagent/aclorch.cpp
+++ b/orchagent/aclorch.cpp
@@ -3470,6 +3470,33 @@ bool AclRange::remove()
     return true;
 }
 
+static bool isSeparateMirrorV6TablePlatform()
+{
+    static const vector<string> separateMirrorV6Platforms = {
+        "nokia_ixr7220_h6",   // Tomahawk6
+    };
+
+    DBConnector cfgDb("CONFIG_DB", 0);
+    Table cfgDeviceMetaDataTable(&cfgDb, CFG_DEVICE_METADATA_TABLE_NAME);
+    string specific_platform;
+
+    if (!cfgDeviceMetaDataTable.hget("localhost", "platform", specific_platform))
+    {
+        return false;
+    }
+
+    for (const auto& p : separateMirrorV6Platforms)
+    {
+        if (specific_platform.find(p) != string::npos)
+        {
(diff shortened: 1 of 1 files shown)


bingwang-ms (commented):
Do we know which qualifier is talking too many bits?
I would like to hold this change as only a few slices are available. Separating EVERFLOW and EVERFLOWV6 means hogging 3 additional IFP slices, which results in no resources for other Ingress ACL application.

dgodwin-nokia:
> Do we know which qualifier is talking too many bits? I would like to hold this change as only a few slices are available. Separating EVERFLOW and EVERFLOWV6 means hogging 3 additional IFP slices, which results in no resources for other Ingress ACL application.

Hi Bing, 

In the BCM CSP I opened, they said that the primary issue is that TH6 has a larger inPorts qualifier due to increased number of logical ports. Even though this is also the case on TH5-512, they had said that there were other TH6 specific IFP constraints that may need to be worked around to allow this group to fit in a triple (with range check being include). I will also follow up on the case to see if there is any progress. 

I noticed that L4Range check was removed from the host trap group for TH6, presumably to make that qset fit in a double-wide group. 

If I remove L4Range check (64 bits for src/dst range) from the combined V4/V6 mirror table, the qualifier set fits in a triple wide group. (instead of creating seperate groups which occupy a double and a triple). Do you think this change is OK to make? If so, we can close this PR and I will open a new one to disable L4Range check for TH6 platforms. 

Dylan

bingwang-ms:
Hi @dgodwin-nokia, the issue will be addressed from SAI and SDK. We can probably close this PR now.

PR #4487, opened 2026-04-16
[teamsyncd] Filter out spurious millisecond LAG member status flaps

### LINKED ISSUE #26531: [FDB] [orchagent] Stale bridge port OID after LAG member transition causes FDB state divergence and traffic loss
#### Description

We had a production incident where all servers behind a ToR became unreachable for about 75 minutes. The switch looked fine - ports up, management reachable - but servers couldn't be reached. Only a `config reload` recovered it.

It looks like a race condition between FDB event notifications and bridge port OID lifecycle during a LAG member state transition. When the remote side removed and re-added a LAG member, the switch got a burst of FDB MOVE/DELETE events referencing a bridge port OID that was already marked as removed in syncd. orchagent couldn't resolve it and silently dropped 1,046 FDB events over 11 minutes, causing its FDB cache to go out of sync with the ASIC.

Same family of issues as:
- sonic-net/sonic-swss#421, #6936, #7538, #10605, #10629, #13069
- sonic-net/sonic-swss#2913

Prior reports describe this as a log error or stale Redis entry. In our
case it caused a full traffic blackhole. Code is identical across 202405,
202411, and 202511.

#### Steps to reproduce:

1. SONiC switch with Broadcom ASIC, PortChannel uplink with 2 members
2. Large FDB table (~800+ MACs) learned on the LAG
3. On remote side, remove one LAG member from the aggregate
4. Wait a few seconds, re-add it
5. Check syslog for "Failed to get port by bridge port ID"
6. Remove the LAG member again

#### What we saw in the logs

**FDB storm** at the moment the LAG member was re-added (thousands of
events at the same sub-second):
```
syncd: fdbEvent: delete (0) for mac AA:BB:CC:DD:EE:01 vid:0x8063, port:0x0 lagid:0x1 flags:0x120080 lag:true
syncd: fdbEvent: move event (5) for mac AA:BB:CC:DD:EE:01 vid:0x8063, port:0xe lagid:0x0 flags:0x120400 lag:false
syncd: fdbEvent: delete (0) for mac AA:BB:CC:DD:EE:02 vid:0x8063, port:0x0 lagid:0x1 flags:0x120080 lag:true
syncd: fdbEvent: move event (5) for mac AA:BB:CC:DD:EE:02 vid:0x8063, port:0xe lagid:0x0 flags:0x120400 lag:false
```

**syncd translating a removed RID** - the bridge port RID is in
`m_removedRid2vid` (the graveyard in `VirtualOidTranslator.cpp`), not
the active map. `eraseRidAndVid()` puts it there when an object is
deleted. syncd still translates it and passes the stale VID to orchagent:
```
syncd: :- checkRidExists: removed RID oid:0x1003a0000000e exists
syncd: :- translateRidToVid: translating removed RID oid:0x1003a0000000e, to VID oid:0x3a00000000009b
```

**orchagent dropping FDB events** - `FdbOrch::update()` calls
`getPortByBridgePortId()` to look up the VID in `saiOidToAlias`. Fails.
For non-FLUSH events it logs ERROR and returns, dropping the event.
1,046 times over 11 minutes:
```
ERR orchagent: :- update: Failed to get port by bridge port ID 0x3a00000000009b.
ERR orchagent: message repeated 16 times: [ :- update: Failed to get port by bridge port ID 0x3a00000000009b.]
...
ERR orchagent: message repeated 1046 times: [ :- update: Failed to get port by bridge port ID 0x3a00000000009b.]
```

Same thing reproduced during the config reload with different OIDs:
```
ERR orchagent: message repeated 310 times: [ :- update: Failed to get port by bridge port ID 0x3a0000000000a9.]
ERR orchagent: message repeated 117 times: [ :- update: Failed to get port by bridge port ID 0x3a00000000008f.]
```

**Neighbor-to-MAC resolution failures** - at 05:58 (when traffic dropped),
938 of these in a burst:
```
syncd: SAI_API_NEIGHBOR:brcm_sai_xgs_common_neighbor_mac_db_update:118 Mac entry not found - skip nbr processing. add 0, dir 3
```
The ASIC is trying to resolve neighbor entries to MAC entries in the FDB but the MACs aren't found. When this happens, L3 forwarding breaks because the ASIC doesn't know which port to send packets to.

#### Key observation: orchagent didn't initiate the bridge port removal

There is no `removeBridgePort` log during the incident. The only one is
from the config reload at 07:15:
```
NOTICE orchagent: :- removeDefaultBridgePorts: Remove bridge ports from default 1Q bridge
```

This matters because `removeBridgePort()` in `portsorch.cpp` calls `flushFDBEntries()` before removing the bridge port, which cleans up orchagent's `m_entries` cache. If that had run, the stale FDB events arriving afterward probably wouldn't have mattered because there'd be nothing in `m_entries` to go out of sync.

But since the bridge port appears to have been invalidated below orchagent (possibly inside the SAI during the LAG transition), orchagent never got a chance to flush. Its `m_entries` went stale, 1,046 FDB update events were dropped, and the software FDB state diverged from the ASIC.

Bouncing the interface didn't help. Only `config reload` recovered it, which suggests the software state was broken in a way that couldn't self-heal.

#### Output of `show version`:

```
SONiC Software Version: SONiC.202405.762
Platform: x86_64-micas_m2-w6520-48c8qc-r0
HwSKU: M2-W6520-48C8QC
ASIC: broadcom
Kernel: 6.1.0-29-2-amd64
```

haojun291:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/26531.

**What I did**
Introduces a debounce timer in teamsyncd to delay the state transition of LAG members from enabled to disabled.
- Implemented a 10ms delay window that triggers only when a member transitions from enabled to disabled.
- Spurious flaps occurring within this window are suppressed before reaching APP_DB, ensuring downstream components (like orchagent and SAI) only process stabilized state changes.

**Why I did it**
Rapid LACP renegotiation during remote LAG member re-addition can induce millisecond status jitter, which is a spurious artifact that should not occur in stable protocol operations.
img_v3_0210j_ed39a4e8-ae4d-47ab-b694-f8b1ca412b6g

Source-side suppression: Filtering these flaps at the source prevents unnecessary and wasteful state propagation from teamsyncd to APPL_DB and the SAI layer.
Mitigate cascading failures: It specifically prevents cascading effects such as unexpected MAC move storms(as reported in #26531).

**How I verified it**
  1. monitored APP_DB to confirm ms flaps no longer trigger redundant writes.
  2. verified that persistent link-down events are still correctly propagated after the 10ms delay.
  3. confirmed traffic forwarding remains stable during LAG member transitions with ~800 MACs.

**Details if related**

### DIFF
--- a/teamsyncd/teamsync.cpp
+++ b/teamsyncd/teamsync.cpp
@@ -23,6 +23,8 @@ using namespace swss;
 /* Taken from drivers/net/team/team.c */
 #define TEAM_DRV_NAME "team"
 
+static const timespec LAG_MEMBER_DISABLE_DEBOUNCE = {0, 10 * 1000 * 1000};
+
 TeamSync::TeamSync(DBConnector *db, DBConnector *stateDb, Select *select) :
     m_select(select),
     m_lagTable(db, APP_LAG_TABLE_NAME),
@@ -59,6 +61,36 @@ void TeamSync::periodic()
     }
 
     doSelectableTask();
+}
+
+bool TeamSync::processTimerSelectable(Selectable *selectable)
+{
+    auto timerIt = m_timerToMember.find(selectable);
+    if (timerIt == m_timerToMember.end())
+    {
+        SWSS_LOG_WARN("Detect untracked debounce timer selectable %p!",
+                selectable);
+        return false;
+    }
+
+    auto lag = timerIt->second.first;
+    auto memberName = timerIt->second.second;
+    auto stateIt = lag->m_memberDebounceStates.find(memberName);
+    if (stateIt == lag->m_memberDebounceStates.end() || !stateIt->second.pendingDisable)
+    {
+        return true;
+    }
+
+    auto timer = stateIt->second.timer;
+
+    timer->stop();
+    stateIt->second.pendingDisable = false;
+
+    applyLagMemberStatus(lag, memberName, false);
+    SWSS_LOG_NOTICE("Debounce timeout expired for LAG %s member %s, apply disabled",
+            lag->getLagName().c_str(), memberName.c_str());
+
+    return true;
 }
 
 void TeamSync::doSelectableTask()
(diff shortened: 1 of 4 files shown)


prsunny:
Seems like a hacky fix. @saiarcot895 , please check the PR.

saiarcot895 (changes requested):
This doesn't seem right. If there are spurious flaps being reported by teamd (which seems to be the case, because that's the source of events for teamsyncd), then the cause of the flaps in teamd needs to be investigated, and the fix needs to be in teamd. SWSS shouldn't be the one filtering out such events.

haojun291:
> This doesn't seem right. If there are spurious flaps being reported by teamd (which seems to be the case, because that's the source of events for teamsyncd), then the cause of the flaps in teamd needs to be investigated, and the fix needs to be in teamd. SWSS shouldn't be the one filtering out such events.

Thanks for the feedback @saiarcot895. The attached packet trace indicates that the flap results from rapid LACP convergence during a remote member re-addition.

At .887136, an 'Expired' state packet is received, triggering an 'enabled → disabled' event. Less than 1ms later (.888012), the state transitions to 'Collecting/Distributing', triggering a 'disabled → enabled' event.
image
image


Teamd faithfully reflect every state machine transition. As a generic protocol daemon, i think it should not be throttled to accommodate hardware-specific constraints. teamsyncd acts as an impedance matcher here. By introducing a small debounce window, we filter high-frequency software events into a stable state that the ASIC can reliably consume.

This approach preserves system stability without compromising LACP protocol integrity within teamd. What do you think?

PR #4488, opened 2026-04-16
[orchagent][MirrorOrch] Add sampled port mirroring support to MirrorOrch

Janetxxx:
**What I did**
Depends on: https://github.com/sonic-net/sonic-swss/pull/4502

Adds sampled port mirroring support to MirrorOrch (orchagent), extending mirror session configuration to optionally create and attach a SAI SamplePacket object and to program ports using sampled-mirroring attributes instead of full-mirror attributes.

- Extend MirrorEntry and CONFIG_DB parsing to support sample_rate and truncate_size.
- Create/remove SAI SamplePacket objects when sample_rate > 0 and wire them into session activation/deactivation.
- Update per-port mirroring programming to use SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE + SAI_PORT_ATTR_INGRESS_SAMPLE_MIRROR_SESSION for sampled mirroring.

**Why I did it**
Added sampled port mirroring with truncation support on ERSPAN sessions

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -31,6 +31,8 @@
 #define MIRROR_SESSION_DST_PORT             "dst_port"
 #define MIRROR_SESSION_DIRECTION            "direction"
 #define MIRROR_SESSION_TYPE                 "type"
+#define MIRROR_SESSION_SAMPLE_RATE          "sample_rate"
+#define MIRROR_SESSION_TRUNCATE_SIZE        "truncate_size"
 
 #define MIRROR_SESSION_DEFAULT_VLAN_PRI 0
 #define MIRROR_SESSION_DEFAULT_VLAN_CFI 0
@@ -47,6 +49,7 @@
 extern sai_switch_api_t *sai_switch_api;
 extern sai_mirror_api_t *sai_mirror_api;
 extern sai_port_api_t *sai_port_api;
+extern sai_samplepacket_api_t *sai_samplepacket_api;
 
 extern sai_object_id_t  gSwitchId;
 extern PortsOrch*       gPortsOrch;
@@ -59,6 +62,9 @@ MirrorEntry::MirrorEntry(const string& platform) :
         dscp(8),
         ttl(255),
         queue(0),
+        sample_rate(0),
+        truncate_size(0),
+        samplepacketId(SAI_NULL_OBJECT_ID),
         sessionId(0),
         refCount(0)
 {
@@ -473,6 +479,14 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
             {
                 entry.type = fvValue(i);
             }
+            else if (fvField(i) == MIRROR_SESSION_SAMPLE_RATE)
+            {
+                entry.sample_rate = to_uint<uint32_t>(fvValue(i));
+            }
+            else if (fvField(i) == MIRROR_SESSION_TRUNCATE_SIZE)
+            {
+                entry.truncate_size = to_uint<uint32_t>(fvValue(i));
+            }
             else
             {
                 SWSS_LOG_ERROR("Failed to parse session %s configuration. Unknown attribute %s", key.c_str(), fvField(i).c_str());
@@ -494,6 +508,22 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
     if (src_ip_initialized && dst_ip_initialized && entry.srcIp.getIp().family != entry.dstIp.getIp().family)
     {
         SWSS_LOG_ERROR("Address family of source and destination IPs is different");
+        return task_process_status::task_invalid_entry;
+    }
+
+    // Truncate size requires sampled mirroring to be enabled
+    if (entry.truncate_size > 0 && entry.sample_rate == 0)
+    {
+        SWSS_LOG_ERROR("Truncate size requires sampled mirroring to be enabled for session %s",
+                       key.c_str());
+        return task_process_status::task_invalid_entry;
+    }
+
+    // Sampled mirroring requires RX direction
+    if (entry.sample_rate > 0 && entry.direction != MIRROR_RX_DIRECTION)
+    {
+        SWSS_LOG_ERROR("Sampled mirroring requires RX direction for session %s",
+                       key.c_str());
         return task_process_status::task_invalid_entry;
     }
 
@@ -811,7 +841,9 @@ bool MirrorOrch::updateSession(const string& name, MirrorEntry& session)
 bool MirrorOrch::setUnsetPortMirror(Port port,
                                     bool ingress,
                                     bool set,
-                                    sai_object_id_t sessionId)
+                                    sai_object_id_t sessionId,
+                                    sai_object_id_t samplepacketId,
+                                    uint32_t sample_rate)
 {
     // Check if the mirror direction is supported by the ASIC
     if (ingress && !m_switchOrch->isPortIngressMirrorSupported())
@@ -826,6 +858,70 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
     }
 
     sai_status_t status;
+
+    if (sample_rate > 0 && ingress)
+    {
+        // Sampled mirroring path: use SAMPLEPACKET_ENABLE + SAMPLE_MIRROR_SESSION
+        sai_attribute_t sp_attr;
+        sp_attr.id = SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE;
+        sp_attr.value.oid = set ? samplepacketId : SAI_NULL_OBJECT_ID;
+
+        sai_attribute_t mirror_attr;
+        mirror_attr.id = SAI_PORT_ATTR_INGRESS_SAMPLE_MIRROR_SESSION;
+        if (set)
+        {
+            mirror_attr.value.objlist.count = 1;
+            mirror_attr.value.objlist.list = &sessionId;
+        }
+        else
+        {
+            mirror_attr.value.objlist.count = 0;
+        }
+
+        // Set/clear in correct order
+        if (set)
+        {
+            // Set: SAMPLEPACKET_ENABLE first, then SAMPLE_MIRROR_SESSION
+            status = sai_port_api->set_port_attribute(port.m_port_id, &sp_attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to set SAMPLEPACKET_ENABLE on port %s, status %d",
+                                port.m_alias.c_str(), status);
+                return false;
+            }
+            status = sai_port_api->set_port_attribute(port.m_port_id, &mirror_attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to set SAMPLE_MIRROR_SESSION on port %s, status %d",
+                                port.m_alias.c_str(), status);
+                // Rollback: clear SAMPLEPACKET_ENABLE
+                sp_attr.value.oid = SAI_NULL_OBJECT_ID;
+                sai_port_api->set_port_attribute(port.m_port_id, &sp_attr);
+                return false;
+            }
+        }
+        else
+        {
+            // Clear: SAMPLE_MIRROR_SESSION first, then SAMPLEPACKET_ENABLE (reverse order)
+            status = sai_port_api->set_port_attribute(port.m_port_id, &mirror_attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to clear SAMPLE_MIRROR_SESSION on port %s, status %d",
+                                port.m_alias.c_str(), status);
+                return false;
+            }
+            status = sai_port_api->set_port_attribute(port.m_port_id, &sp_attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to clear SAMPLEPACKET_ENABLE on port %s, status %d",
+                                port.m_alias.c_str(), status);
+                return false;
+            }
+        }
+        return true;
+    }
+
+    // Full mirror path (existing behavior)
     sai_attribute_t port_attr;
     port_attr.id = ingress ? SAI_PORT_ATTR_INGRESS_MIRROR_SESSION:
                            SAI_PORT_ATTR_EGRESS_MIRROR_SESSION;
@@ -896,7 +992,7 @@ bool MirrorOrch::configurePortMirrorSession(const string& name, MirrorEntry& ses
             }
             if (session.direction == MIRROR_RX_DIRECTION  || session.direction == MIRROR_BOTH_DIRECTION)
             {
-                if (!setUnsetPortMirror(port, true, set, session.sessionId))
+                if (!setUnsetPortMirror(port, true, set, session.sessionId, session.samplepacketId, session.sample_rate))
                 {
                     SWSS_LOG_ERROR("Failed to configure mirror session %s port %s",
                         name.c_str(), port.m_alias.c_str());
@@ -905,7 +1001,7 @@ bool MirrorOrch::configurePortMirrorSession(const string& name, MirrorEntry& ses
             }
             if (session.direction == MIRROR_TX_DIRECTION || session.direction == MIRROR_BOTH_DIRECTION)
             {
-                if (!setUnsetPortMirror(port, false, set, session.sessionId))
+                if (!setUnsetPortMirror(port, false, set, session.sessionId, session.samplepacketId, session.sample_rate))
                 {
                     SWSS_LOG_ERROR("Failed to configure mirror session %s port %s",
                         name.c_str(), port.m_alias.c_str());
@@ -1079,6 +1175,18 @@ bool MirrorOrch::activateSession(const string& name, MirrorEntry& session)
 
     session.status = true;
 
+    // Create SamplePacket if sample_rate > 0
+    if (session.sample_rate > 0)
+    {
+        if (!createSamplePacket(name, session))
+        {
+            SWSS_LOG_ERROR("Failed to create samplepacket, removing mirror session %s", name.c_str());
+            sai_mirror_api->remove_mirror_session(session.sessionId);
+            session.status = false;
+            return false;
+        }
+    }
+
     if (!session.src_port.empty() && !session.direction.empty())
     {
         status = configurePortMirrorSession(name, session, true);
@@ -1116,6 +1224,16 @@ bool MirrorOrch::deactivateSession(const string& name, MirrorEntry& session)
         if (status == false)
         {
             SWSS_LOG_ERROR("Failed to deactivate port mirror session %s", name.c_str());
+            return false;
+        }
+    }
+
+    // Remove SamplePacket if it exists
+    if (session.samplepacketId != SAI_NULL_OBJECT_ID)
+    {
+        if (!removeSamplePacket(name, session))
+        {
+            SWSS_LOG_ERROR("Failed to remove samplepacket for session %s", name.c_str());
             return false;
         }
     }
@@ -1175,6 +1293,79 @@ bool MirrorOrch::updateSessionDstMac(const string& name, MirrorEntry& session)
 
     setSessionState(name, session, MIRROR_SESSION_DST_MAC_ADDRESS);
 
+    return true;
+}
+
+
+bool MirrorOrch::createSamplePacket(const string& name, MirrorEntry& session)
+{
+    SWSS_LOG_ENTER();
+
+    sai_attribute_t attrs[4];
+    uint32_t attr_count = 0;
+
+    attrs[attr_count].id = SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE;
+    attrs[attr_count].value.u32 = session.sample_rate;
+    attr_count++;
+
+    attrs[attr_count].id = SAI_SAMPLEPACKET_ATTR_TYPE;
+    attrs[attr_count].value.s32 = SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION;
+    attr_count++;
+
+    if (session.truncate_size > 0)
+    {
+        attrs[attr_count].id = SAI_SAMPLEPACKET_ATTR_TRUNCATE_ENABLE;
+        attrs[attr_count].value.booldata = true;
+        attr_count++;
+
+        attrs[attr_count].id = SAI_SAMPLEPACKET_ATTR_TRUNCATE_SIZE;
+        attrs[attr_count].value.u32 = session.truncate_size;
+        attr_count++;
+    }
+
+    sai_status_t status = sai_samplepacket_api->create_samplepacket(
+        &session.samplepacketId, gSwitchId, attr_count, attrs);
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to create samplepacket for session %s, status %d",
+                       name.c_str(), status);
+        task_process_status handle_status = handleSaiCreateStatus(SAI_API_SAMPLEPACKET, status);
+        if (handle_status != task_success)
+        {
+            return parseHandleSaiStatusFailure(handle_status);
+        }
+    }
+
+    SWSS_LOG_NOTICE("Created samplepacket for session %s, rate %u, truncate %u",
+                    name.c_str(), session.sample_rate, session.truncate_size);
+    return true;
+}
+
+bool MirrorOrch::removeSamplePacket(const string& name, MirrorEntry& session)
+{
+    SWSS_LOG_ENTER();
+
+    if (session.samplepacketId == SAI_NULL_OBJECT_ID)
+    {
+        return true;
+    }
+
+    sai_status_t status = sai_samplepacket_api->remove_samplepacket(session.samplepacketId);
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to remove samplepacket for session %s, status %d",
+                       name.c_str(), status);
+        task_process_status handle_status = handleSaiRemoveStatus(SAI_API_SAMPLEPACKET, status);
+        if (handle_status != task_success)
+        {
+            return parseHandleSaiStatusFailure(handle_status);
+        }
+    }
+
+    session.samplepacketId = SAI_NULL_OBJECT_ID;
+    SWSS_LOG_NOTICE("Removed samplepacket for session %s", name.c_str());
     return true;
 }
 
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -388,6 +388,13 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
 {
     SWSS_LOG_ENTER();
 
+
+    if (m_syncdMirrors.find(key) != m_syncdMirrors.end())
+    {
+        SWSS_LOG_ERROR("Mirror session %s already exists", key.c_str());
+        return task_process_status::task_invalid_entry;
+    }
+
     bool src_ip_initialized = false;
     bool dst_ip_initialized = false;
 
@@ -476,11 +483,21 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
             }
             else if (fvField(i) == MIRROR_SESSION_SAMPLE_RATE)
             {
-                entry.sample_rate = to_uint<uint32_t>(fvValue(i));
+                try {
+                    entry.sample_rate = to_uint<uint32_t>(fvValue(i));
+                } catch (const exception& e) {
+                    SWSS_LOG_ERROR("Invalid sample_rate for session %s: %s", key.c_str(), e.what());
+                    return task_process_status::task_invalid_entry;
+                }
             }
             else if (fvField(i) == MIRROR_SESSION_TRUNCATE_SIZE)
             {
-                entry.truncate_size = to_uint<uint32_t>(fvValue(i));
+                try {
+                    entry.truncate_size = to_uint<uint32_t>(fvValue(i));
+                } catch (const exception& e) {
+                    SWSS_LOG_ERROR("Invalid truncate_size for session %s: %s", key.c_str(), e.what());
+                    return task_process_status::task_invalid_entry;
+                }
             }
             else
             {
@@ -1031,8 +1048,15 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
 
     sai_status_t status;
 
-    if (sample_rate > 0 && ingress)
+    if (sample_rate > 0)
     {
+        if (!ingress)
+        {
+            SWSS_LOG_ERROR("Sampled mirroring on egress is not supported for port %s",
+                            port.m_alias.c_str());
+            return false;
+        }
+
         // Sampled mirroring path: use SAMPLEPACKET_ENABLE + SAMPLE_MIRROR_SESSION
         sai_attribute_t sp_attr;
         sp_attr.id = SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE;
@@ -1350,7 +1374,14 @@ bool MirrorOrch::activateSession(const string& name, MirrorEntry& session)
     // Create SamplePacket if sample_rate > 0
     if (session.sample_rate > 0)
     {
-        if (!createSamplePacket(name, session))
+        if (!m_switchOrch->isPortIngressSampleMirrorSupported())
+        {
+            SWSS_LOG_WARN("Sampled mirroring not supported on this platform, "
+                          "falling back to full mirror for session %s", name.c_str());
+            session.sample_rate = 0;
+            session.truncate_size = 0;
+        }
+        else if (!createSamplePacket(name, session))
         {
             SWSS_LOG_ERROR("Failed to create samplepacket, removing mirror session %s", name.c_str());
             sai_mirror_api->remove_mirror_session(session.sessionId);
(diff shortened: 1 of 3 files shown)


orchagent/mirrororch.cpp
```diff
@@ -1079,6 +1175,18 @@ bool MirrorOrch::activateSession(const string& name, MirrorEntry& session)
...
+            SWSS_LOG_ERROR("Failed to create samplepacket, removing mirror session %s", name.c_str());
+            sai_mirror_api->remove_mirror_session(session.sessionId);
+            session.status = false;
+            return false;
+        }
+    }
+
     if (!session.src_port.empty() && !session.direction.empty())
     {
         status = configurePortMirrorSession(name, session, true);
```
banidoru, line 1192: **Resource leak on partial failure:** If `createSamplePacket` succeeds but `configurePortMirrorSession` (a few lines below) fails and returns `false`, `activateSession` returns failure but never cleans up the samplepacket object. The mirror session removal on line 1181 only runs if samplepacket creation fails — there's no equivalent cleanup for the reverse scenario. Consider adding `removeSamplePacket()` in the failure path after `configurePortMirrorSession` returns false, or restructuring so `deactivateSession` is called on any activation failure.

orchagent/mirrororch.cpp
```diff
@@ -826,6 +858,70 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
...
     sai_status_t status;
+
+    if (sample_rate > 0 && ingress)
+    {
+        // Sampled mirroring path: use SAMPLEPACKET_ENABLE + SAMPLE_MIRROR_SESSION
+        sai_attribute_t sp_attr;
+        sp_attr.id = SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE;
+        sp_attr.value.oid = set ? samplepacketId : SAI_NULL_OBJECT_ID;
+
+        sai_attribute_t mirror_attr;
```
banidoru, line 869: **Silent fallthrough for egress + sampled mirroring:** When `sample_rate > 0` but `ingress` is false, this condition is not met and the code silently falls through to the full-mirror path. While `createEntry` currently blocks non-RX sampled sessions, `setUnsetPortMirror` is a general-purpose function. Add a defensive guard here (e.g., `if (sample_rate > 0 && !ingress) { SWSS_LOG_ERROR(...); return false; }`) to prevent subtle bugs if callers change in the future.

orchagent/mirrororch.cpp
```diff
@@ -1178,6 +1296,79 @@ bool MirrorOrch::updateSessionDstMac(const string& name, MirrorEntry& session)
...
+        SWSS_LOG_ERROR("Failed to create samplepacket for session %s, status %d",
+                       name.c_str(), status);
+        task_process_status handle_status = handleSaiCreateStatus(SAI_API_SAMPLEPACKET, status);
+        if (handle_status != task_success)
+        {
+            return parseHandleSaiStatusFailure(handle_status);
+        }
+    }
+
+    SWSS_LOG_NOTICE("Created samplepacket for session %s, rate %u, truncate %u",
```
banidoru, line 1340: **Misleading success log after SAI failure:** If `create_samplepacket` fails but `handleSaiCreateStatus` returns `task_success` (e.g., item already exists), execution falls through to the `SWSS_LOG_NOTICE("Created samplepacket...")` log and returns true. However, the value of `session.samplepacketId` after a failed create call is undefined — SAI doesn't guarantee the output OID is valid on failure. This could lead to using a garbage OID downstream. Wrap the success log and `return true` in the `status == SAI_STATUS_SUCCESS` branch, and handle the `handleSaiCreateStatus == task_success` case separately.

orchagent/mirrororch.h
```diff
@@ -57,6 +57,11 @@ struct MirrorEntry
         sai_object_id_t portId;
     } neighborInfo;
 
+    // Sampled mirroring fields
+    uint32_t sample_rate;                        // 0 = full mirror (default)
+    uint32_t truncate_size;                      // 0 = no truncation (default)
+    sai_object_id_t samplepacketId;              // SAI_NULL_OBJECT_ID if not sampled
+
```
banidoru, line 64: **Naming inconsistency:** `sample_rate` and `truncate_size` use snake_case while all other `MirrorEntry` fields use camelCase (e.g., `sessionId`, `refCount`, `greType`). Consider renaming to `sampleRate` and `truncateSize` for consistency.

banidoru (commented):
- **Resource leak**: samplepacket not cleaned up if port mirror config fails after samplepacket creation succeeds in `activateSession`
- **Silent fallthrough**: egress + sample_rate > 0 silently uses full-mirror path in `setUnsetPortMirror` — needs defensive guard
- **Undefined OID usage**: `createSamplePacket` may use garbage samplepacketId after SAI failure if handleSaiCreateStatus returns task_success
- **Naming inconsistency**: new fields use snake_case vs existing camelCase convention

orchagent/mirrororch.cpp
banidoru, line 869: **Silent fallthrough for egress + sampled mirroring:** When `sample_rate > 0` but `ingress` is false, this condition is not met and the code silently falls through to the full-mirror path. While `createEntry` currently blocks non-RX sampled sessions, `setUnsetPortMirror` is a general-purpose function. Add a defensive guard here (e.g., `if (sample_rate > 0 && !ingress) { SWSS_LOG_ERROR(...); return false; }`) to prevent subtle bugs if callers change in the future.

orchagent/mirrororch.cpp
banidoru, line 1192: **Resource leak on partial failure:** If `createSamplePacket` succeeds but `configurePortMirrorSession` (a few lines below) fails and returns `false`, `activateSession` returns failure but never cleans up the samplepacket object. The mirror session removal on line 1181 only runs if samplepacket creation fails — there is no equivalent cleanup for the reverse scenario. Consider adding `removeSamplePacket()` in the failure path after `configurePortMirrorSession` returns false, or restructuring so `deactivateSession` is called on any activation failure.

orchagent/mirrororch.cpp
banidoru, line 1340: **Misleading success log after SAI failure:** If `create_samplepacket` fails but `handleSaiCreateStatus` returns `task_success` (e.g., item already exists), execution falls through to the `SWSS_LOG_NOTICE("Created samplepacket...")` log and returns true. However, the value of `session.samplepacketId` after a failed create call is undefined — SAI does not guarantee the output OID is valid on failure. This could lead to using a garbage OID downstream. Wrap the success log and `return true` in the `status == SAI_STATUS_SUCCESS` branch, and handle the `handleSaiCreateStatus == task_success` case separately.

orchagent/mirrororch.cpp
```diff
@@ -1120,6 +1228,16 @@ bool MirrorOrch::deactivateSession(const string& name, MirrorEntry& session)
         }
     }
```
banidoru, line 1229: If `removeSamplePacket()` fails during `deactivateSession()`, the code logs an error but proceeds to remove the mirror session anyway. This can leak the samplepacket object. The function should return `false` to allow for a retry.

banidoru (changes requested):
- In `setUnsetPortMirror`, the `sample_rate > 0 && ingress` check is insufficient. If `ingress` is false, the function silently falls through to the non-sampled path. Add a guard to explicitly reject sampled egress mirroring.
- In `activateSession`, a resource leak can occur. If `createSamplePacket` succeeds but `configurePortMirrorSession` fails, the created samplepacket is not cleaned up.
- In `createSamplePacket`, if `create_samplepacket` fails, the code may log a success message and return true if `handleSaiCreateStatus` returns `task_success`, leading to the use of an uninitialized `samplepacketId`.
- In `deactivateSession`, if `removeSamplePacket` fails, the error is logged but the function proceeds. This can lead to a resource leak. The function should return `false` on failure.

orchagent/mirrororch.cpp
```diff
@@ -473,6 +479,14 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
             {
                 entry.type = fvValue(i);
             }
+            else if (fvField(i) == MIRROR_SESSION_SAMPLE_RATE)
+            {
+                entry.sample_rate = to_uint<uint32_t>(fvValue(i));
+            }
```
banidoru, line 485: **Uncaught exception on malformed input:** `to_uint<uint32_t>` throws `std::invalid_argument` or `std::out_of_range` on non-numeric or overflowing values. Since this is config-DB driven, malformed input is plausible. Wrap in a try/catch and return `task_invalid_entry`, consistent with how other parse failures are handled in this function.

```cpp
try {
    entry.sample_rate = to_uint<uint32_t>(fvValue(i));
} catch (const std::exception& e) {
    SWSS_LOG_ERROR("Invalid sample_rate for session %s: %s", key.c_str(), e.what());
}
```
Same applies to `truncate_size` below.

orchagent/mirrororch.cpp
```diff
@@ -1120,6 +1228,16 @@ bool MirrorOrch::deactivateSession(const string& name, MirrorEntry& session)
...
+    // Remove SamplePacket if it exists
+    if (session.samplepacketId != SAI_NULL_OBJECT_ID)
+    {
+        if (!removeSamplePacket(name, session))
+        {
+            SWSS_LOG_ERROR("Failed to remove samplepacket for session %s", name.c_str());
+            return false;
+        }
+    }
+
```
banidoru, line 1240: **Half-deactivated state on samplepacket removal failure:** At this point, `configurePortMirrorSession(name, session, false)` has already succeeded — ports are unconfigured. If `removeSamplePacket` fails and returns `false`, the mirror session and samplepacket SAI objects remain, but the ports no longer reference them. A subsequent retry of `deactivateSession` will attempt `configurePortMirrorSession(false)` again on already-cleared ports.

Consider restructuring so samplepacket removal failure still proceeds to remove the mirror session (log and continue), or skip port unconfiguration on retry if ports are already cleared.

orchagent/mirrororch.cpp
```diff
@@ -1079,6 +1175,18 @@ bool MirrorOrch::activateSession(const string& name, MirrorEntry& session)
...
+    if (session.sample_rate > 0)
+    {
+        if (!createSamplePacket(name, session))
+        {
+            SWSS_LOG_ERROR("Failed to create samplepacket, removing mirror session %s", name.c_str());
+            sai_mirror_api->remove_mirror_session(session.sessionId);
+            session.status = false;
+            return false;
+        }
+    }
```
banidoru, line 1188: **Resource leak: samplepacket not cleaned up on port config failure.** If `createSamplePacket` succeeds (line 1181) but `configurePortMirrorSession` fails a few lines below (line 1195), `activateSession` removes the mirror session but does not call `removeSamplePacket`. The samplepacket SAI object is leaked.

Add `removeSamplePacket(name, session)` in the failure path after `configurePortMirrorSession` returns false.

orchagent/mirrororch.cpp
```diff
@@ -1178,6 +1296,79 @@ bool MirrorOrch::updateSessionDstMac(const string& name, MirrorEntry& session)
...
+    sai_status_t status = sai_samplepacket_api->remove_samplepacket(session.samplepacketId);
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to remove samplepacket for session %s, status %d",
+                       name.c_str(), status);
+        task_process_status handle_status = handleSaiRemoveStatus(SAI_API_SAMPLEPACKET, status);
+        if (handle_status != task_success)
+        {
+            return parseHandleSaiStatusFailure(handle_status);
```
banidoru, line 1363: **Success log/return reached after SAI failure:** When `create_samplepacket` fails but `handleSaiCreateStatus` returns `task_success`, execution falls through to this log and `return true`. The `samplepacketId` output parameter is undefined after a failed SAI create call — using it downstream is undefined behavior.

```cpp
{
    // ... existing error handling ...
}
else
{
    SWSS_LOG_NOTICE("Created samplepacket ...");
}
return (status == SAI_STATUS_SUCCESS);
```

banidoru (commented):
- `to_uint<uint32_t>` can throw on malformed config input, crashing orchagent instead of returning `task_invalid_entry`
- `deactivateSession` clears ports before removing samplepacket — if samplepacket removal fails, state is half-torn-down with no recovery path
- `createSamplePacket` logs success and returns true even when SAI create failed (if handleSaiCreateStatus returns task_success) — OID may be garbage
- Resource leak: samplepacket not cleaned up when `configurePortMirrorSession` fails after successful `createSamplePacket`

orchagent/mirrororch.cpp
```diff
@@ -473,6 +479,14 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
                 entry.type = fvValue(i);
             }
+            else if (fvField(i) == MIRROR_SESSION_SAMPLE_RATE)
+            {
+                entry.sample_rate = to_uint<uint32_t>(fvValue(i));
+            }
+            else if (fvField(i) == MIRROR_SESSION_TRUNCATE_SIZE)
+            {
+                entry.truncate_size = to_uint<uint32_t>(fvValue(i));
+            }
```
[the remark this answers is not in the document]
banidoru: Still open — this concern (STATE_DB persistence) is not addressed in the new commits.

orchagent/mirrororch.cpp
```diff
@@ -826,6 +851,70 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
...
+        {
+            // Set: SAMPLEPACKET_ENABLE first, then SAMPLE_MIRROR_SESSION
+            status = sai_port_api->set_port_attribute(port.m_port_id, &sp_attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to set SAMPLEPACKET_ENABLE on port %s, status %d",
+                                port.m_alias.c_str(), status);
+                return false;
+            }
+            status = sai_port_api->set_port_attribute(port.m_port_id, &mirror_attr);
```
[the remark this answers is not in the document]
banidoru: Still open — this concern (unit tests) is not addressed in the new commits.

orchagent/mirrororch.cpp
```diff
@@ -826,6 +851,70 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
...
+            status = sai_port_api->set_port_attribute(port.m_port_id, &sp_attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to clear SAMPLEPACKET_ENABLE on port %s, status %d",
+                                port.m_alias.c_str(), status);
+                return false;
+            }
+        }
+        return true;
+    }
```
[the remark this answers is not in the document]
banidoru: Still open — this concern (LAG handling) is not addressed in the new commits.

orchagent/mirrororch.cpp
```diff
@@ -826,6 +851,70 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
...
+            {
+                SWSS_LOG_ERROR("Failed to clear SAMPLE_MIRROR_SESSION on port %s, status %d",
+                                port.m_alias.c_str(), status);
+                return false;
+            }
+            status = sai_port_api->set_port_attribute(port.m_port_id, &sp_attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to clear SAMPLEPACKET_ENABLE on port %s, status %d",
+                                port.m_alias.c_str(), status);
```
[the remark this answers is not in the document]
banidoru: Still open — this concern (rollback on clear path) is not addressed in the new commits.

orchagent/mirrororch.cpp
banidoru: Still open — this concern (resource leak on partial failure) is not addressed in the new commits.

orchagent/mirrororch.cpp
banidoru: Still open — this concern (silent fallthrough egress+sampled) is not addressed in the new commits.

orchagent/mirrororch.cpp
banidoru: Still open — this concern (misleading success log) is not addressed in the new commits.

orchagent/mirrororch.h
banidoru: Still open — this concern (naming inconsistency) is not addressed in the new commits.

orchagent/mirrororch.cpp
banidoru: Still open — this concern (silent fallthrough (dup)) is not addressed in the new commits.

orchagent/mirrororch.cpp
banidoru: Still open — this concern (resource leak (dup)) is not addressed in the new commits.

orchagent/mirrororch.cpp
banidoru: Still open — this concern (misleading success log (dup)) is not addressed in the new commits.

orchagent/mirrororch.cpp
banidoru: Still open — this concern (removeSamplePacket failure) is not addressed in the new commits.

orchagent/mirrororch.cpp
banidoru: Still open — this concern (uncaught exception on malformed input) is not addressed in the new commits.

orchagent/mirrororch.cpp
banidoru: Still open — this concern (half-deactivated state) is not addressed in the new commits.

orchagent/mirrororch.cpp
banidoru: Still open — this concern (resource leak samplepacket) is not addressed in the new commits.

orchagent/mirrororch.cpp
banidoru: Still open — this concern (success log/return after SAI failure) is not addressed in the new commits.

orchagent/mirrororch.cpp
```diff
@@ -522,6 +547,22 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+
+    SWSS_LOG_NOTICE("Updating mirror session %s", key.c_str());
+
+    auto task_status = deleteEntry(key);
+    if (task_status != task_process_status::task_success)
+    {
+        SWSS_LOG_ERROR("Failed to delete existing mirror session %s during update", key.c_str());
+        return task_status;
+    }
+
```
banidoru, line 635: **Delete+recreate is disruptive for active sessions.** `updateEntry` unconditionally deletes and recreates the mirror session, which tears down all port mirroring configuration (SAI objects, port attributes) and rebuilds from scratch. This causes a window where mirrored traffic is lost.

For fields that are mutable on an active session (e.g., `sample_rate`, `truncate_size`), consider updating the existing session in-place. Reserve delete+recreate only for immutable fields (e.g., src/dst IP changes). At minimum, document this limitation.

orchagent/mirrororch.cpp
```diff
@@ -522,6 +547,22 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
     return task_process_status::task_success;
 }
 
+task_process_status MirrorOrch::updateEntry(const string& key, const vector<FieldValueTuple>& data)
+{
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_NOTICE("Updating mirror session %s", key.c_str());
+
+    auto task_status = deleteEntry(key);
```
banidoru: **No rollback if `createEntry` fails after `deleteEntry` succeeds.** If `deleteEntry` succeeds but `createEntry` fails (e.g., resource exhaustion, invalid new config), the original session is gone with no way to restore it. Consider saving the old config and attempting to restore on failure, or at least logging a clear error about the lost session.

orchagent/mirrororch.cpp
```diff
@@ -385,12 +391,7 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
     bool src_ip_initialized = false;
     bool dst_ip_initialized = false;
```
banidoru, line 400: **Removed duplicate-entry guard from `createEntry`.** The deleted check (`m_syncdMirrors.find(key) != end → task_duplicated`) was a safety net. Now `createEntry` assumes the caller checked `sessionExists()`. If any code path calls `createEntry` without the guard, it will insert a new `MirrorEntry` into the map, leaking the old SAI objects.

Consider keeping a defensive assertion or early-return here.

banidoru (commented):
Re-review (iteration 2) — new `updateEntry` commit reviewed.

- **`updateEntry` is delete+recreate**: This works but is disruptive — it tears down the active mirror session (deactivates ports, removes SAI objects) and recreates from scratch, causing a traffic-mirroring gap. For mutable fields, consider updating in-place; for immutable fields, the delete+recreate approach is acceptable but should be documented.
- **Duplicate entry guard removed from `createEntry`**: The old duplicate check (`task_duplicated`) was removed. Now `createEntry` no longer guards against duplicates — it relies on the caller (`doTask`) to check `sessionExists()` first. If any other caller invokes `createEntry` directly on an existing session, it will silently overwrite state. Consider keeping a defensive check or adding an assert.
- **`updateEntry` doesn't validate delete side-effects**: If `deleteEntry` partially succeeds (e.g., deactivation fails mid-way but entry is removed from map), `createEntry` will proceed on potentially inconsistent SAI state.
- **All prior review concerns remain unaddressed** — resource leaks, rollback gaps, STATE_DB persistence, LAG handling, naming, input validation, unit tests, etc. are still open.

banidoru (changes requested):
- **STATE_DB Persistence**:  and  are still not persisted to STATE_DB, which is necessary for warm-restart scenarios.
- **Unit Tests**: No new unit tests have been added to cover the new sampled mirroring logic, including success, failure, and rollback paths.
- **LAG Port Handling**: The sampled mirroring implementation does not correctly handle LAG ports; it attempts to configure the LAG object itself rather than its members.
- **Resource Leaks**: There are multiple paths where  objects can be leaked upon partial failures during session activation or deactivation.
- **Error Handling**: Several error paths have insufficient or incorrect error handling, such as silent fallthroughs and uncaught exceptions from string-to-integer conversions.

banidoru (changes requested):
- **STATE_DB Persistence**:  and  are still not persisted to STATE_DB, which is necessary for warm-restart scenarios.
- **Unit Tests**: No new unit tests have been added to cover the new sampled mirroring logic, including success, failure, and rollback paths.
- **LAG Port Handling**: The sampled mirroring implementation does not correctly handle LAG ports; it attempts to configure the LAG object itself rather than its members.
- **Resource Leaks**: There are multiple paths where  objects can be leaked upon partial failures during session activation or deactivation.
- **Error Handling**: Several error paths have insufficient or incorrect error handling, such as silent fallthroughs and uncaught exceptions from string-to-integer conversions.

banidoru (changes requested):
The previous reviewer's comments have been addressed, but several key issues remain unresolved. The new commits have not fixed the most critical problems.

- **STATE_DB Persistence**:  and  are still not persisted to STATE_DB. This is required for warm-restart.
- **Unit Tests**: No new unit tests have been added to cover the new sampled mirroring logic.
- **LAG Port Handling**: The sampled mirroring implementation does not correctly handle LAG ports.
- **Resource Leaks**: Multiple paths can lead to  objects being leaked on partial failures.
- **Error Handling**: Several error paths have insufficient or incorrect error handling.

banidoru (changes requested):
- **No prior concerns addressed:** All previously raised issues (resource leaks, LAG handling, STATE_DB persistence, uncaught exceptions, misleading success log, naming inconsistency, missing unit tests, rollback gaps) remain unaddressed in this iteration.
- **New `updateEntry` adds risk:** Delete+recreate tears down active sessions causing traffic loss; no rollback if recreate fails after successful delete.
- **Removed duplicate-entry guard:** Defensive check removed from `createEntry` — callers must now remember `sessionExists()` or leak SAI objects.
- **Validation improvements landed:** truncate_size/sample_rate cross-validation and RX-direction enforcement are good additions.
- **Verdict:** Requesting changes — too many open correctness and safety issues remain.

banidoru (changes requested):
Re-review iteration 2 — new `updateEntry` commit reviewed. **No prior concerns were addressed.** Key outstanding issues:

- **Resource leaks**: samplepacket not cleaned up when `configurePortMirrorSession` fails after `createSamplePacket` succeeds (multiple call sites)
- **`updateEntry` is unsafe**: delete+recreate tears down active sessions with no rollback if `createEntry` fails after `deleteEntry` succeeds; removed duplicate-entry guard from `createEntry` risks SAI object leaks
- **LAG not handled**: sampled mirroring path programs the LAG OID directly instead of iterating PHY members — will fail on platforms requiring per-member programming
- **Crash on malformed input**: `to_uint<uint32_t>` throws on non-numeric config values; needs try/catch
- **Missing STATE_DB persistence, unit tests, naming consistency, and defensive guards** — all prior threads remain open

orchagent/mirrororch.cpp
banidoru: Addressed — the new code now updates mutable fields (sample_rate, truncate_size) in-place via SAI set attributes, reserving delete+recreate only for immutable fields. Good improvement.

orchagent/mirrororch.cpp
banidoru: Partially addressed — immutable field changes still do delete+recreate with no rollback on createEntry failure. The scope is reduced since mutable fields now update in-place, but the concern remains for immutable field changes.

orchagent/mirrororch.cpp
```diff
@@ -522,6 +547,131 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+                SWSS_LOG_ERROR("Failed to update sample_rate for session %s, status %d", key.c_str(), status);
+                return task_process_status::task_failed;
+            }
+            session.sample_rate = new_sample_rate;
+            SWSS_LOG_NOTICE("Updated sample_rate to %u for session %s", new_sample_rate, key.c_str());
+        }
+
+        if (truncate_size_changed)
+        {
+            sai_attribute_t attr;
```
banidoru, line 682: **Partial rollback on sequential attribute updates:** If `sample_rate` was successfully updated (line 623) but `truncate_size` update fails here, `session.sample_rate` has already been mutated to `new_sample_rate` with no rollback. The session is now in an inconsistent state — SAI has the new sample_rate but the old truncate_size, yet a retry will see `sample_rate_changed == false` and skip it.

Consider: (1) apply both updates before committing to `session`, or (2) rollback sample_rate on truncate_size failure.

orchagent/mirrororch.cpp
```diff
@@ -522,6 +547,131 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+    uint32_t new_truncate_size = session.truncate_size;
+
+    for (auto& fv : data)
+    {
+        const auto& field = fvField(fv);
+        const auto& value = fvValue(fv);
+
+        if (field == MIRROR_SESSION_SAMPLE_RATE)
+        {
+            new_sample_rate = to_uint<uint32_t>(value);
```
banidoru: **Uncaught exception on malformed input:** `to_uint<uint32_t>(value)` can throw `std::invalid_argument` / `std::out_of_range` on non-numeric input from CONFIG_DB. Same issue flagged in `createEntry` (still open). Wrap in try/catch and return `task_invalid_entry`.

orchagent/mirrororch.cpp
```diff
@@ -522,6 +547,131 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+            return task_status;
+        }
+        return createEntry(key, data);
+    }
+
+    // Only mutable fields (sample_rate, truncate_size) changed - update in-place
+    bool sample_rate_changed = (new_sample_rate != session.sample_rate);
+    bool truncate_size_changed = (new_truncate_size != session.truncate_size);
+
+    if (!sample_rate_changed && !truncate_size_changed)
```
banidoru, line 655: **Sampled→full mirror transition not handled:** When `new_sample_rate` becomes 0 (disabling sampling), this path updates the samplepacket's rate attribute to 0 via SAI instead of removing the samplepacket and transitioning to the full-mirror path. A sample_rate of 0 on a samplepacket object has undefined/vendor-specific behavior. This should fall through to the delete+recreate path when transitioning between sampled and full mirror modes.

banidoru (changes requested):
Re-review (iteration 3) — new `updateEntry` in-place update logic is a solid improvement over unconditional delete+recreate. Key remaining issues:

- **Partial rollback gap in updateEntry:** if sample_rate update succeeds but truncate_size fails, sample_rate is committed with no rollback
- **Uncaught exception in updateEntry:** `to_uint<uint32_t>` can throw on malformed config input (same issue as createEntry, still open)
- **Sampled→full transition not handled:** if sample_rate changes to 0, the in-place path updates the samplepacket rate to 0 instead of tearing down the samplepacket and switching to full mirror
- **Prior concerns still unaddressed:** resource leak on partial activation failure, LAG handling, defensive guard for egress+sampled, naming inconsistency, STATE_DB persistence in createEntry, unit tests

banidoru (changes requested):
**Iteration 3 re-review:**

- **Good:** `updateEntry()` addresses the disruptive delete+recreate for mutable fields. Clean separation of mutable vs immutable update paths.
- **Still open (critical):** Resource leak — `createSamplePacket` success + `configurePortMirrorSession` failure leaks the samplepacket object. No cleanup in that path.
- **Still open (critical):** `createSamplePacket` success log and `return true` reached when SAI create fails but `handleSaiCreateStatus` returns `task_success` — undefined `samplepacketId` used downstream.
- **Still open:** LAG handling missing in sampled-mirroring path, uncaught exceptions on `to_uint` in both `createEntry`/`updateEntry`, no unit tests, STATE_DB persistence gap, naming inconsistency.
- **New (iter 3):** `updateEntry` has partial rollback issue — if `sample_rate` update succeeds but `truncate_size` fails, session state is inconsistent. Sampled→full transition (rate→0) updates samplepacket with rate=0 instead of removing it.

orchagent/mirrororch.cpp
banidoru, line 682: **Partial rollback on sequential attribute updates:** If `sample_rate` was successfully updated (line 623) but `truncate_size` update fails here, `session.sample_rate` has already been mutated to `new_sample_rate` with no rollback. The session is now in an inconsistent state — SAI has the new sample_rate but the old truncate_size, yet a retry will see `sample_rate_changed == false` and skip it.

Consider: (1) apply both updates before committing to `session`, or (2) rollback sample_rate on truncate_size failure.

orchagent/mirrororch.cpp
banidoru: **Uncaught exception on malformed input:** `to_uint<uint32_t>(value)` can throw `std::invalid_argument` / `std::out_of_range` on non-numeric input from CONFIG_DB. Same issue flagged in `createEntry` (still open). Wrap in try/catch and return `task_invalid_entry`.

orchagent/mirrororch.cpp
banidoru, line 655: **Sampled→full mirror transition not handled:** When `new_sample_rate` becomes 0 (disabling sampling), this path updates the samplepacket's rate attribute to 0 via SAI instead of removing the samplepacket and transitioning to the full-mirror path. A sample_rate of 0 on a samplepacket object has undefined/vendor-specific behavior. This should fall through to the delete+recreate path when transitioning between sampled and full mirror modes.

orchagent/mirrororch.cpp
```diff
@@ -522,6 +547,131 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+    {
+        SWSS_LOG_NOTICE("No field changes detected for session %s", key.c_str());
+        return task_process_status::task_success;
+    }
+
+    // If samplepacket exists, update its attributes in-place
+    if (session.samplepacketId != SAI_NULL_OBJECT_ID)
+    {
+        if (sample_rate_changed)
```
banidoru, line 665: **Sampled→full mirror transition not handled in-place path.** When `session.samplepacketId != SAI_NULL_OBJECT_ID` and `new_sample_rate` drops to 0, this code sets the SAI samplepacket rate attribute to 0 instead of removing the samplepacket and transitioning to the full-mirror path. A samplepacket with rate=0 has undefined vendor behavior.

This path should detect the transition (`new_sample_rate == 0 && session.sample_rate > 0`) and fall through to delete+recreate, or explicitly remove the samplepacket and reconfigure port attributes.

orchagent/mirrororch.cpp
banidoru: **Uncaught exception on malformed input (still open from iter 1).** `to_uint<uint32_t>(value)` throws on non-numeric CONFIG_DB input. Wrap in try/catch → `task_invalid_entry`. Same issue exists in `createEntry` at the equivalent parse lines.

banidoru (changes requested):
- **Still 12+ open concerns from prior iterations unaddressed** — resource leaks, LAG handling, uncaught exceptions, STATE_DB persistence, unit tests, naming inconsistency.
- **New: `updateEntry` sampled→full transition bug** — when `samplepacketId != SAI_NULL_OBJECT_ID` and `new_sample_rate` drops to 0, the code updates the samplepacket rate to 0 instead of removing it and transitioning to full mirror.
- **New: `updateEntry` missing input validation** — no `truncate_size`/`direction` cross-validation, no try/catch on `to_uint`.
- **Good progress:** `updateEntry` with in-place mutable updates and `sessionExists()` routing are solid improvements.

orchagent/mirrororch.cpp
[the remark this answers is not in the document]
Janetxxx: Clearing SAMPLEPACKET_ENABLE (setting to SAI_NULL_OBJECT_ID) is unlikely to fail in practice. Adding rollback in the clear path would mean re-applying the bindings we're trying to remove, which adds
 complexity without clear benefit. The function returns false on failure, allowing the caller to retry the full deactivation.

Janetxxx:
Closing this PR and merging contents into https://github.com/sonic-net/sonic-swss/pull/4502

PR #4491, opened 2026-04-16
[muxorch] Fix FDB-after-neighbor conversion for host-route mode

Ndancejic:
## What I did
Fix MuxOrch to properly convert neighbors to MUX neighbors when FDB is learned after the neighbor on host-route mode mux ports (202511 backport).

## How I did it
When a secondary neighbor (IP not in the mux cable configured subnet) is learned before its FDB entry is available, `updateNeighbor()` adds it to `mux_nexthop_tb_` with an empty port name. Later when the FDB entry arrives, `updateFdb()` finds the existing entry and processes it through the normal port-change path. However, if the neighbor was never added to `mux_nexthop_tb_` for any reason (e.g., `getNeighborEntry` resolves differently at FDB time), there is no fallback to discover and convert it.

This fix adds FDB-after-neighbor conversion logic to `updateFdb()`, matching the approach used on master (PR #4459):

1. **`updateFdb()`**: Track whether any existing mux neighbor matched the FDB MAC. If not, scan NeighOrch's neighbor table for neighbors with matching MAC and convert them to mux neighbors via `convertNeighborToMux()`.

2. **`convertNeighborToMux()`**: New helper that validates the mux cable, checks for duplicates via `containsNextHop()`, adds the neighbor to `mux_nexthop_tb_`, and calls `MuxCable::updateNeighbor()` to manage the neighbor based on mux state.

3. **`NeighOrch::getNeighborTable()`**: New const accessor exposing the synced neighbor table for scanning in `updateFdb()`.

## How to verify it
Two new VS tests exercise the FDB-after-neighbor race condition:
- `test_fdb_after_neighbor_on_standby_mux`: Verifies neighbor gets tunnel route after FDB arrives on standby
- `test_fdb_after_neighbor_on_active_mux`: Verifies neighbor stays direct on active mux

These tests add a neighbor with IP outside the mux subnet before FDB is learned, then add the FDB entry and verify proper mux neighbor management.

## Which release branch to backport
N/A (this is the 202511 backport of the master fix in #4459)

## Related PRs
- Master fix: https://github.com/sonic-net/sonic-swss/pull/4459

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -1459,6 +1459,7 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
     NeighborEntry neigh;
     MacAddress mac;
     MuxCable* ptr;
+    bool found_existing_mux_neighbor = false;
     for (auto nh = mux_nexthop_tb_.begin(); nh != mux_nexthop_tb_.end(); ++nh)
     {
         auto res = neigh_orch_->getNeighborEntry(nh->first, neigh, mac);
@@ -1466,6 +1467,8 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
         {
             continue;
         }
+
+        found_existing_mux_neighbor = true;
 
         if (nh->second != update.entry.port_name)
         {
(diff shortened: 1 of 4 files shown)


Ndancejic:
@manamand2020 for reference

prsunny:
@Ndancejic , why is it closed? You can move to draft if that helps.

Ndancejic:
> @Ndancejic , why is it closed? You can move to draft if that helps.

When I was working on this, I discovered it's only needed on master branch due to some changes made in the prefix-route mode PR. I need to re-evaluate this

PR #4492, opened 2026-04-17
Replay VxLAN port settings from device metadata

aakashtiwari2630-max:
**What I did**
As part of Sonic CLI support for VxLAN port configuration:
https://github.com/sonic-net/sonic-utilities/pull/4467

Added support to replay VxLAN port settings from DEVICE_METADATA to APP_DB. Whenever Vxlanmgrd comes up, it will read the configured port values from DEVICE_METADATA and replay it to SWITCH_TABLE in APP_DB.

**Why I did it**
Currently there is no way for the VxLAN port setting to persist across config reload or reboot. This change makes it persistence.

**How I verified it**
Tested manually.
[vxlan_test.txt](https://github.com/user-attachments/files/26809525/vxlan_test.txt)

**Details if related**

### DIFF
--- a/cfgmgr/vxlanmgrd.cpp
+++ b/cfgmgr/vxlanmgrd.cpp
@@ -74,6 +74,33 @@ int main(int argc, char **argv)
         }
         gMacAddress = MacAddress(it->second);
 
+        /*
+         * Replay VXLAN port settings from DEVICE_METADATA to APP_DB SWITCH_TABLE
+         * so that orchagent/SAI picks them up on restart/reload.
+         * These fields (vxlan_port, vxlan_sport, vxlan_mask) are persisted in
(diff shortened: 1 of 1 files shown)


anish-n (commented):
DEVICE_METADATA does not look like the right place to add this. Additionally, SWITCH_TABLE can be used as-is to place in the config for vxlan. If there is a usecase to make it persistent via config_db then a new table may need to be introduced.

aakashtiwari2630-max:
> DEVICE_METADATA does not look like the right place to add this. Additionally, SWITCH_TABLE can be used as-is to place in the config for vxlan. If there is a usecase to make it persistent via config_db then a new table may need to be introduced.

Yes @anish-n . We need to make it persistent. Let me update this.

PR #4495, opened 2026-04-20
Skip usb0 in neighbor/route/interface orchagent

stephenxs:
**What I did**

- Treated the BMC **usb0** host interface like other non-ASIC host links (e.g. **eth0** / **docker0** / **lo**):
  - **neighorch**: drop **NEIGH** work for **usb0** before `getPort` so we do not log *"Port usb0 doesn't exist"* or retry on a non-routable port.
  - **neighsyncd**: do not publish **adds** for neighbors on **usb0** to **APPL_DB**; still process **deletes** and warm-restart housekeeping.
  - **intfsorch**: ignore **INTF** entries for **usb0**.
  - **routeorch**: skip routes whose outgoing interface is **usb0**.
- Added / adjusted **mock_tests**: **NeighOrchTest.SkipHostInterfaceUsb0** uses **`NeighborEntry(TEST_IP, std::string("usb0"))`** so **`NextHopKey`** resolves to the `(ip, interface)` overload (avoids **`Error converting … to NextHop`** from the **`(string, bool)`** overload).

**Why I did it**

On platforms where the BMC exposes **usb0**, kernel/neighbor events can generate SWSS/APPL_DB churn and noisy orchagent errors for an interface that is not a switch front-panel port. Filtering **usb0** consistently avoids pointless orchagent work and misleading logs without affecting dataplane routing on ASIC ports.

**How I verified it**

- Built **sonic-swss** mock tests in the slave docker environment (deps from **`/sonic/target/debs/bookworm`** as needed).
- Ran **`./tests --gtest_filter=NeighOrchTest.SkipHostInterfaceUsb0`** → **PASSED**.

**Details if related**

- Aligns behavior with documentation that treats **usb0** as a host-side skip alongside **eth0**/**docker0** where applicable.
- **gtest note**: when **`NeighborEntry`** aliases **`NextHopKey`**, passing a string literal as the second constructor argument can bind to **`(const std::string &, bool)`**; use **`std::string("usb0")`** (or equivalent) for explicit interface-name overload selection.

### DIFF
--- a/orchagent/intfsorch.cpp
+++ b/orchagent/intfsorch.cpp
@@ -814,7 +814,7 @@ void IntfsOrch::doTask(Consumer &consumer)
             }
         }
 
-        if (alias == "eth0" || alias == "docker0")
+        if (alias == "eth0" || alias == "docker0" || alias == "usb0")
         {
             it = consumer.m_toSync.erase(it);
             continue;
--- a/orchagent/neighorch.cpp
+++ b/orchagent/neighorch.cpp
@@ -908,7 +908,7 @@ void NeighOrch::doTask(Consumer &consumer)
 
         string alias = key.substr(0, found);
 
-        if (alias == "eth0" || alias == "lo" || alias == "docker0"
+        if (alias == "eth0" || alias == "lo" || alias == "docker0" || alias == "usb0"
             || ((op == SET_COMMAND) && m_intfsOrch->isInbandIntfInMgmtVrf(alias)))
         {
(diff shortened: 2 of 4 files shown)


Junchao-Mellanox (approved):

keboliu (approved):

stephenxs:
@prsunny could you please review this PR? this is to skip yet another non-ASIC interface. thanks

liat-grozovik (approved):

keboliu:
@prsunny would you please help to review?

PR #799, opened 2019-02-19
Fix vlan incremental config and add vs test cases

jipanyang:
Signed-off-by: Jipan Yang <jipan.yang@alibaba-inc.com>

**What I did**
Fix vlan incremental config update.

**Why I did it**

**How I verified it**
VS.
**Details if related**

PR #4497, opened 2026-04-20
[orchagent] SI settings sync protocol between xcvrd and orchagent

arpit-nexthop:
**What I did**

Implements https://github.com/sonic-net/SONiC/pull/2301

#### Why I did it

Implements the orchagent side of a redesigned SI settings synchronization protocol between xcvrd and orchagent. The previous design used a single NPU_SI_SETTINGS_SYNC_STATUS field in STATE_DB with values NPU_SI_SETTINGS_DEFAULT / _NOTIFIED / _DONE, with xcvrd writing directly to STATE_DB. This is replaced by a two-field, counter-based protocol.

#### How I did it
  Serdes programming is now gated explicitly on xcvrd writing si_settings_notification = SI_SETTINGS_NOTIFIED:<N> to APPL_DB PORT_TABLE, rather than triggering on any port config update carrying serdes fields. A monotonic counter <N> is embedded in both the notification and acknowledgment so xcvrd can reject stale ACKs from prior insertions.

  On the orchagent side, a new applyPortSerdesConfig function handles the full SI settings flow: on SI_SETTINGS_DEFAULT it mirrors the value to STATE_DB without programming serdes (handles transceiver removal); on SI_SETTINGS_NOTIFIED it brings the port down, programs serdes, then writes SI_SYNC_DONE:<N> back to STATE_DB. The admin-state bounce during serdes
  programming no longer updates host_tx_ready, so xcvrd is not triggered until the normal initHostTxReadyState path runs after the port comes back up.

  Orchagent also now initializes si_settings_ack = SI_SETTINGS_DEFAULT:0 in STATE_DB for all ports during port initialization, replacing the equivalent initialization that xcvrd previously did.

#### How to verify it

  Unit tests in portsorch_ut.cpp:
  - New test verifies that SI_SETTINGS_DEFAULT:<N> is mirrored to STATE_DB without serdes programming.
  - All existing serdes tests updated to include si_settings_notification = SI_SETTINGS_NOTIFIED:1 in the port config, routing serdes programming through the new explicit flow.

Tests done on device:
 - shut / no shut tests: The notification count does not increase after 1.
 - xcvr removal and insertion tests: Checks for the programming of the SI settings and increase in the notification number
 - xcvrd restart test: The transceiver was not programmed again.

**Details if related**


prsunny:
@prgeor , @judyjoseph , please review the design and signoff before code PR merge.

snider-nokia:
@arpit-nexthop, all,
   Strongly suggest considering a slight alteration to this refactoring whereby orchagent does not actually bring the ASIC port admin up until such time as the xcvrd provided SI settings have actually arrived at Redis for orchagent use.   Even if SI settings are not in the mix for a given configuration on a given port, orchagent can still wait until the Redis notification flag explicitly indicates that the extant SI configuration provided by xcvrd (or not) is good-to-go.
   
Introducing this type of gate for orchagent, in advance of orchagent actually bringing the ASIC port admin up will prevent the churn that is currently seen when an ASIC port has already been enabled but then SI settings arrive at Redis thus causing orchagent to modify ASIC SerDes settings and consequently flap the ASIC port.

Waiting at orchagent until such time as the SI settings are confirmed to have arrived from xcvrd (even if there were no changes actually provided by xcvrd with the exception of Redis confirmation flag) would prevent ASIC port from being brought up while only partially configured (in advance of SI settings arrival).  Preventing this ASIC port churn/flap will help to clean up the critical enable path such that the ASIC port is only brought up once all provisioning has necessarily been completed and thus commensurately serve to minimize SI related flux within the machinery (as ASIC port(s) are transitioning down-then-up during provisioning time).

arpit-nexthop:
> @arpit-nexthop, all, Strongly suggest considering a slight alteration to this refactoring whereby orchagent does not actually bring the ASIC port admin up until such time as the xcvrd provided SI settings have actually arrived at Redis for orchagent use. Even if SI settings are not in the mix for a given configuration on a given port, orchagent can still wait until the Redis notification flag explicitly indicates that the extant SI configuration provided by xcvrd (or not) is good-to-go.
> 
> Introducing this type of gate for orchagent, in advance of orchagent actually bringing the ASIC port admin up will prevent the churn that is currently seen when an ASIC port has already been enabled but then SI settings arrive at Redis thus causing orchagent to modify ASIC SerDes settings and consequently flap the ASIC port.
> 
> Waiting at orchagent until such time as the SI settings are confirmed to have arrived from xcvrd (even if there were no changes actually provided by xcvrd with the exception of Redis confirmation flag) would prevent ASIC port from being brought up while only partially configured (in advance of SI settings arrival). Preventing this ASIC port churn/flap will help to clean up the critical enable path such that the ASIC port is only brought up once all provisioning has necessarily been completed and thus commensurately serve to minimize SI related flux within the machinery (as ASIC port(s) are transitioning down-then-up during provisioning time).

@snider-nokia I agree with this comment, I am in process of making these changes to test out.

@prsunny @prgeor please comment if this is alright or you need any alteration is required.

orchagent/portsorch.cpp
```diff
@@ -6681,6 +6766,20 @@ bool PortsOrch::initializePorts(std::vector<Port>& ports)
...
 
+void PortsOrch::initializePortSiSettingsSyncStatusBulk(std::vector<Port>& ports)
+{
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_TIMER(__FUNCTION__);
+
+    for (auto& port : ports)
+    {
+        setPortSiSettingsSyncStatus(port, std::string(PORT_SI_SETTINGS_DEFAULT) + ":0");
```
longhuan-cisco, line 7061: This unconditionally overwrites si_settings_ack with SI_SETTINGS_DEFAULT:0 for every port during port initialization.

The HLD relies on retained STATE_DB si_settings_ack=SI_SYNC_DONE:<N> across warm reboot / xcvrd restart to avoid re-applying SI settings. If this initialization runs in a warm-restart/warm-reboot path where STATE_DB was preserved, it can erase the retained ACK and make xcvrd think SI is not applied.

Should this only initialize the field when absent, instead of always overwriting it?

orchagent/portsorch.cpp
```diff
@@ -4494,6 +4508,106 @@ void PortsOrch::doSendToIngressPortTask(Consumer &consumer)
...
+    /* Default settings requested from xcvrd, cleanup happens in next programming
+     * Update the sync_status to default for warm restart scenarios
+     */
+    if (si_settings_notification == PORT_SI_SETTINGS_DEFAULT) 
+    {
+        setPortSiSettingsSyncStatus(port, std::string(PORT_SI_SETTINGS_DEFAULT) + ":" + std::to_string(si_sync_count));
+        return true;
+    }
+
+    if (serdes_attr.empty() && line_serdes_attr.empty() && system_serdes_attr.empty())
```
longhuan-cisco: If OA receives SI_SETTINGS_NOTIFIED:<N> but the parsed update has no recognized ASIC/line/system SerDes attributes, this returns true without setting serdes_settings_changed. Later, si_settings_ack=SI_SYNC_DONE:<N> is only written when serdes_settings_changed is true.

That means xcvrd can wait in SI_SETTINGS_WAIT until timeout even though OA consumed the notification. This can happen if the media_settings payload contains only non-SerDes fields, or if fields are filtered/unsupported by OA. Should OA still ACK SI_SYNC_DONE:<N> after consuming a NOTIFIED message, even when there is nothing to program?

orchagent/portsorch.cpp
arpit-nexthop: ACK, can move it to be done during cold start only or if the data is not present.

orchagent/portsorch.cpp
arpit-nexthop: Ideally this case should not happen as 

```
    if (!pCfg.serdes_settings_sync_status.is_set)
    {
        return false;
    }
```
should have something set, I will check and update this check

arpit-nexthop:
@snider-nokia Have updated the approach to gate admin_state call on the SI values. 
This introduces a dependency on xcvrd being present and notifying the SI values. 
The behavior remains unchanged if xcvrd is not running or if it has no values to publish.

abdosi:
@ymd-arista

abdosi:
@mlok-nokia : can you please help sign-off from Nokia.
@longhuan-cisco : please help sign-off from Cisco
@ymd-arista : please help sign-off from Arista.
@prgeor / @mihirpat1 : please sign-off from MSFT

tests/mock_tests/portsorch_ut.cpp
```diff
@@ -685,6 +685,10 @@ namespace portsorch_test
 
             gEvpnMhOrch = new EvpnMhOrch(evpn_df_es_table_connectors);
 
+            // Default the SI-settings admin gate OFF for tests; the dedicated gate tests
+            // enable it explicitly. This keeps legacy admin/serdes tests behaving as before.
+            gPortsOrch->m_xcvrdSiSyncExpected = false;
```
anamehra, line 690: This default makes most of the old tests run with the new gate disabled:

```cpp
// Default the SI-settings admin gate OFF for tests; the dedicated gate tests
// enable it explicitly. This keeps legacy admin/serdes tests behaving as before.
gPortsOrch->m_xcvrdSiSyncExpected = false;
```

That is fine for keeping old tests stable, but it means we still need explicit tests with the gate enabled for the paths this PR changes: chassis-packet/VOQ behavior and gearbox line/system-side serdes programming. I do not see `gbsyncd` directly in these two PRs, but the gearbox serdes path is changed here and should be covered with the production gate setting.

orchagent/portsorch.cpp
```diff
@@ -4652,20 +4806,6 @@ bool PortsOrch::programSerdes(
         return false;
     }
 
-    if (port.m_admin_state_up)
```
ymd-arista, line 4655: To confirm, are we removing this block because it's stale code?

Since, we call `programSerdes` on a port only when we are programming it for the first time and it can't be up then?

arpit-nexthop: Yes, this is stale code, programSerdes will happen the first time on a down port

tests/mock_tests/portsorch_ut.cpp
arpit-nexthop: We would not want to edit all test cases that just rely on link up to have some SI values.
I will add gearbox SI testcases to enable the sync as well

snider-nokia:
@arpit-nexthop, We are attempting to run OC on master image after incorporating this PR alongside its sister PR sonic-platform-daemons [#799](https://github.com/sonic-net/sonic-platform-daemons/pull/799), but running into trouble.  Please see PR comment there for details: 
https://github.com/sonic-net/sonic-platform-daemons/pull/799#issuecomment-5241785223

arpit-nexthop:
> @arpit-nexthop, We are attempting to run OC on master image after incorporating this PR alongside its sister PR sonic-platform-daemons [#799](https://github.com/sonic-net/sonic-platform-daemons/pull/799), but running into trouble. Please see PR comment there for details: [sonic-net/sonic-platform-daemons#799 (comment)](https://github.com/sonic-net/sonic-platform-daemons/pull/799#issuecomment-5241785223)

Problem is the following:

  bool gateAdminOnSi = m_xcvrdSiSyncExpected && isFrontPanelPort(p);
  and isFrontPanelPort is only port.m_type == Port::PHY (line 1915). An inband port like Ethernet-IB0 is Port::PHY, so it gets gated:

  xcvrd never manages inband/recycle ports (no transceiver), so it never publishes anything → si_settings_ack stays at the SI_SETTINGS_DEFAULT:0

I will update the fix in PR

orchagent/port/portschema.h
```diff
@@ -109,4 +115,5 @@
 #define PORT_MODE                  "mode"
 #define PORT_UNRELIABLE_LOS        "unreliable_los"
 #define PORT_MEDIA_TYPE            "media_type"
+#define PORT_SI_SYNC_SETTINGS      "si_settings_notification"
 #define PORT_FAST_LINKUP           "fast_linkup"
```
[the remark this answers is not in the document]
arpit-nexthop: Updated PR description

orchagent/portsorch.cpp
```diff
@@ -2367,6 +2409,20 @@ void PortsOrch::setHostTxReady(Port port, const std::string &status)
...
+void PortsOrch::setPortSiSettingsSyncStatus(const Port& port, const std::string& status)
+{
+    SWSS_LOG_ENTER();
+
+    vector<FieldValueTuple> tuples;
+    FieldValueTuple tuple("si_settings_ack", status);
+    tuples.push_back(tuple);
+
+    m_portStateTable.set(port.m_alias, tuples);
+
```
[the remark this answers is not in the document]
arpit-nexthop: Updated the PR description

tests/mock_tests/portsorch_ut.cpp
```diff
@@ -1767,9 +1774,26 @@ namespace portsorch_test
...
+        std::vector<swss::FieldValueTuple> ackValues;
+        statePortTable.get("Ethernet0", ackValues);
+        bool found_sync_done = false;
+        for (const auto &fv : ackValues)
+        {
+            if (fvField(fv) == "si_settings_ack")
+            {
+                found_sync_done = true;
+                ASSERT_EQ(fvValue(fv), "SI_SYNC_DONE:1");
+                break;
```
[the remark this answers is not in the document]
arpit-nexthop: Updated the PR description to match code

orchagent/portsorch.cpp
```diff
@@ -5613,18 +5734,81 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+                 * shapes pCfg.admin_status BEFORE serdes is applied. */
+                auto &siGate = m_portAdminSiGate[p.m_alias];
+                bool gateAdminOnSi = m_xcvrdSiSyncExpected && isFrontPanelPort(p);
+
+                if (pCfg.serdes_settings_sync_status.is_set)
                 {
-                    if (!programSerdes(p, p.m_line_side_id, p.m_switch_id, line_serdes_attr))
+                    /* If we get UNAVIL or NOTIFIED, we can allow admin_up to proceed */
+                    siGate.si_notified =
+                        (pCfg.serdes_settings_sync_status.type != PORT_SI_SETTINGS_DEFAULT);
```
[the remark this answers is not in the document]
arpit-nexthop: Done

tests/mock_tests/portsorch_ut.cpp
```diff
@@ -1483,7 +1487,8 @@ namespace portsorch_test
                 { "regn_bfm1p",    "0x1e,0x20,0x1f,0x21"         },
                 { "regn_bfm1n",    "0xaa,0xac,0xab,0xad"         },
                 { "custom_serdes_attrs", custom_serdes_attrs     },
-                { "media_type",    "backplane"                   }
+                { "media_type",    "backplane"                   },
+                { "si_settings_notification","SI_SETTINGS_NOTIFIED:1"      }
             }
```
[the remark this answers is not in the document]
arpit-nexthop: Updated the PR description

orchagent/portsorch.cpp
```diff
@@ -5613,18 +5734,84 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+                /* SI-settings admin gate: when xcvrd is expected to provide SI settings for
+                 * this front-panel port, hold the NPU MAC admin-up until the settings are
+                 * notified (so serdes is applied on a down port - never bouncing the link),
+                 * then replay the deferred admin-up so the link comes up exactly once. This
+                 * shapes pCfg.admin_status BEFORE serdes is applied. */
+                auto &siGate = m_portAdminSiGate[p.m_alias];
+                // Only external (transceiver) ports are SI-notified by xcvrd; internal-role
+                // ports (inband/recirc/internal/DPU-connect) would otherwise defer forever.
+                bool gateAdminOnSi = m_xcvrdSiSyncExpected && isFrontPanelPort(p)
+                                     && p.m_role == Port::Role::Ext;
```
longhuan-cisco, line 5746: **High: the external-port gate depends on an indeterminate role for normal front-panel ports**

The `Ext` filter is necessary: internal, inband, recirculation, and DPU-connect ports do not receive xcvrd SI notifications and would otherwise be held forever. However, a missing `role` is currently not normalized to `Ext`:

- `PortConfig::role.value` has no initializer and `role` is optional (`validatePortConfig()` only requires lanes and speed).
- `Port::m_role` also has no initializer.
- `addPortBulk()` and `initExistingPort()` copy `port.role.value` into `m_role` without checking `role.is_set`.

Therefore a valid ordinary front-panel PORT entry with no explicit `role` reaches this comparison with an indeterminate enum value:

```text
Ethernet0 has no role field
        -> m_role is indeterminate
        -> m_role == Ext may be false
        -> gateAdminOnSi is false
        -> admin-up can proceed before SerDes is applied
```

This turns a latent initialization issue into a safety decision in the new handshake. Could we normalize an absent role to `Port::Role::Ext` (ideally in both the parsed config and stored Port object), while preserving explicit `Int`/`Inb`/`Rec`/`Dpc` values? Please also add gate-enabled tests for a front-panel port with no role, explicit `Ext`, and each non-external role. Removing the `Ext` filter instead would reintroduce the internal-port deadlock.

orchagent/portsorch.cpp
```diff
@@ -5613,18 +5734,84 @@ void PortsOrch::doPortTask(Consumer &consumer)
...
+                   (the SI-only APPL_DB update carries no admin_status of its own). */
+                if (gateAdminOnSi && siGate.si_notified && siGate.admin_up_deferred && !pCfg.admin_status.is_set)
                 {
-                    if (!programSerdes(p, p.m_system_side_id, p.m_switch_id, system_serdes_attr))
+                    pCfg.admin_status.is_set = true;
+                    pCfg.admin_status.value  = true;   // we only ever defer admin-up
+                    // Keep the raw config string in sync so getAdminStatusStr() (used for
+                    // logging) reflects the replayed admin-up instead of an empty value.
+                    pCfg.fieldValueMap[PORT_ADMIN_STATUS] = "up";
+                    siGate.admin_up_deferred = false;
```
longhuan-cisco: **High: clearing `admin_up_deferred` here can strand a configured-up port down after a retry**

This clears the only remembered admin-up intent before either SerDes programming or the final admin-up operation has succeeded. The SI update is normally an SI-only `ProducerStateTable` update, so it carries no `admin_status` (as the comment above notes). On failure, `it++; continue` retains that same task for OA retry; xcvrd does not need to republish it.

A transient failure can therefore run as follows:

```text
ConfigDB desired state is UP
OA defers admin-up, or downs an already-up MAC for OIR
NOTIFIED arrives -> this block injects admin_status=up
admin_up_deferred is cleared here
applyPortSerdesConfig() fails -> SI-only task remains pending
next OA pass reconstructs the same task without admin_status
SerDes now succeeds and OA writes SI_SYNC_DONE:N
but the port is already down and admin_up_deferred is false
final admin restoration is not requested -> hardware remains down
```

The same problem occurs if SerDes succeeds but the final `setPortAdminStatus(true)` call fails: `SI_SYNC_DONE:N` has already been written, the task retries, but the deferred-up intent is gone. CMIS may then advance after seeing `DONE:N` while the NPU port remains down even though ConfigDB still says up. Another insertion alone is not guaranteed to recover it because another SI-only update still carries no configured admin status.

Would it be better to keep `admin_up_deferred` set until `setPortAdminStatus(true)` actually succeeds? Tests should inject (1) SerDes failure after an initial deferred-up, (2) SerDes failure after an already-up OIR port is downed, and (3) final admin-up failure after SI success, then verify retry eventually restores the port. Staying down while SI is unsafe is correct; forgetting the pending configured-up intent is not.

arpit-nexthop:
Sister PR: https://github.com/sonic-net/sonic-platform-daemons/pull/799 
All the tests have been done with these changes, 
@prsunny can we get these merged into master for further soaking and test?

snider-nokia (approved):
Now with battery of OC tests passing successfully, suggest merging these changes to master so that multiple vendors can integrate/pull easily into their lab environments for soak.

ymd-arista (approved):

anamehra (approved):

orchagent/port/porthlpr.cpp
```diff
@@ -973,6 +973,68 @@ bool PortHelper::parsePortPtIntfId(PortConfig &port, const std::string &field, c
...
+        port.serdes_settings_sync_status.count = count;
+        port.serdes_settings_sync_status.is_set = true;
+    }
+    catch (const std::exception &e)
+    {
+        SWSS_LOG_ERROR("Failed to parse field(%s): %s", field.c_str(), e.what());
+        return false;
+    }
+
+    return true;
```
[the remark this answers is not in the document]
anamehra: please check this one.

prsunny:
@prgeor , @judyjoseph , could you review/signoff?

orchagent/port/porthlpr.cpp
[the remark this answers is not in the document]
arpit-nexthop: I am inclined not to include it as the input field is generated from xcvrd, the logic there will not lead to malformed generation. The extra step look parsing steps makes code unreadable.

PR #4498, opened 2026-04-20
[orchagent] Add host_tx_ready_count to track state transitions

arpit-nexthop:
### Summary

Safer fix in 202511 for https://github.com/sonic-net/sonic-swss/pull/4497

Add persistent counter in STATE_DB PORT_TABLE that increments on every host_tx_ready state transition. Uses stateless read-before-write pattern with STATE_DB as single source of truth. Enables xcvrd to detect state changes even when boolean status remains unchanged across restarts.

#### Why I did it

When a port transitions through host_tx_ready states (e.g., false->true->false->true),
downstream consumers like xcvrd may miss transitions if they only monitor the boolean
status value. Specifically, if a port goes down and comes back up while xcvrd is
processing other events, the final state appears unchanged (true->true), causing xcvrd
to skip necessary re-initialization of the transceiver.

This is problematic for CMIS and SFF transceivers that need to re-run initialization
sequences whenever the host interface is reset, regardless of whether the final state
looks the same.

#### How I did it

- Added `host_tx_ready_count` field to PORT_TABLE in STATE_DB
- Modified `initHostTxReadyState()` to initialize the counter to 0 if not present
- Implemented stateless logic in `setHostTxReady()`:
  - Read current status and count from STATE_DB (single source of truth)
  - Compare current status with new status
  - If different, increment counter and write both fields atomically
  - If same, skip write (no state change)
- Coalesced writes of both host_tx_ready and host_tx_ready_count in single set()

The counter increments on every true state transition:
- false->true: count++
- true->false: count++
- true->true (same): count unchanged

#### How to verify it

Check state db for count, after interface create on a config reload, we apply serdes settings, which sets it to false and then true.
```
2026 Mar 27 21:33:21.634547 humm212 NOTICE swss#orchagent: :- initHostTxReadyState: initialize host_tx_ready as false for port Ethernet0
2026 Mar 27 21:33:21.673139 humm212 NOTICE swss#orchagent: :- setHostTxReady: Setting host_tx_ready status = true, count = 1, alias = Ethernet0, port_id = 0x1000000000001
2026 Mar 27 21:33:21.673259 humm212 NOTICE swss#orchagent: :- setPortAdminStatus: Set admin status UP host_tx_ready to true for port Ethernet0
2026 Mar 27 21:33:23.248422 humm212 NOTICE swss#orchagent: :- setHostTxReady: Setting host_tx_ready status = false, count = 2, alias = Ethernet0, port_id = 0x1000000000001
2026 Mar 27 21:33:23.248495 humm212 NOTICE swss#orchagent: :- setPortAdminStatus: Set admin status DOWN host_tx_ready to false for port Ethernet0
2026 Mar 27 21:33:23.358714 humm212 NOTICE swss#orchagent: :- setHostTxReady: Setting host_tx_ready status = true, count = 3, alias = Ethernet0, port_id = 0x1000000000001
2026 Mar 27 21:33:23.358826 humm212 NOTICE swss#orchagent: :- setPortAdminStatus: Set admin status UP host_tx_ready to true for port Ethernet0

  "PORT_TABLE|Ethernet0": {
    "expireat": 1774653533.1311154,
    "ttl": -0.001,
    "type": "hash",
    "value": {
      "NPU_SI_SETTINGS_SYNC_STATUS": "NPU_SI_SETTINGS_NOTIFIED",
      "admin_status": "up",
      "fec": "rs",
      "host_tx_ready": "true",
      "host_tx_ready_count": "3",
      "mtu": "9100",
      "netdev_oper_status": "up",
      "speed": "400000",
      "state": "ok",
      "supported_fecs": "rs",
      "supported_speeds": "100000,200000,400000"
    }
  },
```

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -2165,9 +2165,12 @@ void PortsOrch::initHostTxReadyState(Port &port)
 
     if (hostTxReady.empty())
     {
-        setHostTxReady(port, "false");
-        SWSS_LOG_NOTICE("initialize host_tx_ready as false for port %s",
-                        port.m_alias.c_str());
+        /* Initialize host_tx_ready to false and count to 0 */
+        vector<FieldValueTuple> fvs;
+        fvs.emplace_back("host_tx_ready", "false");
+        fvs.emplace_back("host_tx_ready_count", "0");
+        m_portStateTable.set(port.m_alias, fvs);
+        SWSS_LOG_NOTICE("initialize host_tx_ready as false for port %s", port.m_alias.c_str());
     }
 }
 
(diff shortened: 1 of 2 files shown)


arpit-nexthop:
Will post a new workaround for 202511 only in xcvrd

PR #4499, opened 2026-04-20
[vnetorch][vrfmgrd][fpmsyncd]: Defer VNET kernel device deletion until orchagent cleanup completes

nsoma26:
Cherry-picked these fixes into `202511` branch through https://github.com/sonic-net/sonic-swss/pull/4552

**What I did**

Coordinate VNET kernel device deletion with orchagent route cleanup so that fpmsyncd can resolve the VRF ifindex for any in-flight `RTM_DELROUTE` messages from FRR. Three commits:

1. `[fpmsyncd]: Log the correct error and reduce severity for VRF route delete log`
   - Check the `getIfName()` return value in `RouteSync::onMsg()` and return early on failure with a clear message including the `ifindex` and route prefix (previously the empty `master_name` produced a misleading `Invalid VRF name` log in `onRouteMsg()`).
   - Use `ERROR` severity for route adds (unexpected) and `INFO` for route deletes (benign post-deletion cleanup).
   - Log VRF/VNET link events (`RTM_NEWLINK`/`RTM_DELLINK`) with device name and ifindex at `NOTICE`, consistent with portsyncd/fdbsyncd, so operators can correlate VRF name to ifindex from logs.

2. `[vnetorch][vrfmgrd]: Defer VNET kernel device deletion until orchagent cleanup completes`
   - `VNetOrch`: write to `STATE_VRF_OBJECT_TABLE` on VNET creation (`hset`) and delete on VNET removal (`del`), mirroring `VRFOrch` behavior.
   - `vrfmgrd`: apply the same `isVrfObjExist()` polling logic in the VNET delete path, deferring `delLink()` until `VNetOrch` signals completion — closing the gap called out by the existing `"VNETOrch is not do so now."` comment.
   - `vrfmgrd`: add a per-`vrfName` retry counter and a rate-limited (every `VNET_DEL_DEFER_WARN_THRESHOLD = 20` deferrals) `SWSS_LOG_WARN` on both defer paths, so a stuck handshake (state row never populated by `VNetOrch`, or never cleared) fails loudly instead of silently looping forever. The counter is reset on the success path.
   - `orchdaemon`: pass `stateDb` to the `VNetOrch` constructor. `stateDb` is now a **required** parameter (no default `nullptr`); `m_stateVrfObjectTable` is held as a value `Table` member constructed in the initializer list (matches `VRFOrch`/`VrfMgr`). This eliminates a silent fallback where a future caller forgetting to wire `stateDb` would re-introduce the race with no log trail. The three mock-test callers were updated accordingly.
   - Adds `tests/mock_tests/vnetorch_ut.cpp` covering the `STATE_VRF_OBJECT_TABLE` lifecycle (`hset` on add, `del` on remove) in `VNetOrch::addOperation` / `delOperation`.

3. `[test_vnet]: Add VS test for VNET delete with routes present`
   - New `test_vnet_orch_37` exercises the create -> add routes -> delete VNET while routes are present -> delete routes -> verify cleanup flow and asserts `STATE_VRF_OBJECT_TABLE` lifecycle.
   - Adds `check_vnet_obj_in_state_db`, `wait_for_vnet_obj_in_state_db`, and `wait_for_vnet_obj_removed_from_state_db` helpers in `vnet_lib.py`.

**Why I did it**

Without this coordination, `vrfmgrd` was deleting the kernel VRF device (`ip link del`) immediately on VNET removal, before orchagent had a chance to remove the corresponding routes from APP_DB and the kernel. This left fpmsyncd unable to resolve the VRF ifindex for any pending `RTM_DELROUTE` messages, which:

- produced the misleading `Invalid VRF name` error in the syslog, and
- left stale `VNET_ROUTE_TUNNEL_TABLE` entries in APP_DB after the VNET was gone.

`VRFOrch`/`vrfmgrd` already implemented exactly this two-phase coordination via `STATE_VRF_OBJECT_TABLE`; `VNetOrch` simply did not participate, as called out by the existing comment in the code (`"VNETOrch is not do so now."`). This change closes that gap and reuses the existing mechanism.

**How I verified it**

- Ran the new VS test `tests/test_vnet.py::TestVnetOrch::test_vnet_orch_37` and confirmed:
  - `STATE_VRF_OBJECT_TABLE` is populated for the VNET on creation.
  - VNET delete is deferred while routes are still present and orchagent is cleaning up.
  - All `VNET_ROUTE_TUNNEL_TABLE` entries for the deleted VNET are removed from APP_DB after route cleanup completes (no stale entries).
  - `STATE_VRF_OBJECT_TABLE` entry is removed once cleanup is complete.
- Added a focused mock test `tests/mock_tests/vnetorch_ut.cpp::VNetOrchTest.VnetAddDelStateVrfObjectTableLifecycle` driving `APP_VNET_TABLE` SET/DEL and asserting the `STATE_VRF_OBJECT_TABLE` row appears on add and disappears on del.
- Added two `fpmsyncd` mock tests in `tests/mock_tests/fpmsyncd/test_routesync.cpp` (`RouteDeleteOnMissingVrfDeviceLogsInfoAndSkips`, `RouteAddOnMissingVrfDeviceLogsErrorAndSkips`) covering the new `getIfName()`-failure error/info paths and asserting the early-return contract (no APP_DB write on failure).
- Verified existing `test_vnet.py` and orchagent mock-test suites still pass.
- On a system reproducing the original issue, confirmed the `Invalid VRF name` log is no longer emitted on VNET deletion and that no stale routes are left behind.

**Details if related**

- Reuses the `STATE_VRF_OBJECT_TABLE` schema already documented in `doc/swss-schema.md` (minor wording update to clarify both `VRFOrch` and `VNetOrch` write to it).
- No SAI behavior changes; only an additional STATE_DB hand-off between `VNetOrch` and `vrfmgrd`.

### DIFF
--- a/fpmsyncd/routesync.cpp
+++ b/fpmsyncd/routesync.cpp
@@ -2522,6 +2522,22 @@ void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
 {
     if (nlmsg_type == RTM_NEWLINK || nlmsg_type == RTM_DELLINK)
     {
+        struct rtnl_link *link_obj = (struct rtnl_link *)obj;
+        const char *link_name = rtnl_link_get_name(link_obj);
+        int link_ifindex = rtnl_link_get_ifindex(link_obj);
+
+        if (link_name)
+        {
+            if (strncmp(link_name, VRF_PREFIX, strlen(VRF_PREFIX)) == 0 ||
+                strncmp(link_name, VNET_PREFIX, strlen(VNET_PREFIX)) == 0 ||
+                strncmp(link_name, MGMT_VRF_PREFIX, strlen(MGMT_VRF_PREFIX)) == 0)
+            {
+                SWSS_LOG_NOTICE("VRF link event: %s dev %s ifindex %d",
+                    (nlmsg_type == RTM_NEWLINK) ? "RTM_NEWLINK" : "RTM_DELLINK",
+                    link_name, link_ifindex);
+            }
+        }
+
         nl_cache_refill(m_nl_sock, m_link_cache);
         return;
     }
@@ -2550,8 +2566,30 @@ void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
     if (master_index)
     {
         /* Get the name of the master device */
-        getIfName(master_index, master_name, IFNAMSIZ);
-    
+        if (!getIfName(master_index, master_name, IFNAMSIZ))
+        {
+            char prefix_buf[MAX_ADDR_SIZE + 1] = {0};
+            struct nl_addr *dip = rtnl_route_get_dst(route_obj);
+            if (dip)
+            {
+                nl_addr2str(dip, prefix_buf, MAX_ADDR_SIZE);
+            }
+
+            if (nlmsg_type == RTM_DELROUTE)
+            {
+                SWSS_LOG_INFO("Failed to get interface name for ifindex %u prefix %s, "
+                    "skipping route delete (VRF device may have been deleted)",
+                    master_index, prefix_buf);
+            }
+            else
+            {
+                SWSS_LOG_ERROR("Failed to get interface name for ifindex %u prefix %s, "
+                    "skipping route add (VRF device may have been deleted)",
+                    master_index, prefix_buf);
+            }
+            return;
+        }
+
         /* If the master device name starts with VNET_PREFIX, it is a VNET route.
            The VNET name is exactly the name of the associated master device. */
         if (string(master_name).find(VNET_PREFIX) == 0)
--- a/cfgmgr/vrfmgr.cpp
+++ b/cfgmgr/vrfmgr.cpp
@@ -15,6 +15,13 @@
 #define MGMT_VRF_TABLE_ID 6000
 #define MGMT_VRF          "mgmt"
 
+/*
+ * Emit a SWSS_LOG_WARN every N deferrals of a VNET delete waiting on
+ * VNetOrch's STATE_VRF_OBJECT_TABLE handshake. Chosen to be visible
+ * (sustained backlog) without flooding logs on transient races.
+ */
+#define VNET_DEL_DEFER_WARN_THRESHOLD 20
+
 using namespace swss;
 
 VrfMgr::VrfMgr(DBConnector *cfgDb, DBConnector *appDb, DBConnector *stateDb, const vector<string> &tableNames) :
@@ -314,7 +321,7 @@ void VrfMgr::doTask(Consumer &consumer)
             /*
              * Delay delLink until vrf object deleted in orchagent to ensure fpmsyncd can get vrf ifname.
              * Now state VRF_TABLE|Vrf represent vrf exist in appDB, if it exist vrf device is always effective.
-             * VRFOrch add/del state VRF_OBJECT_TABLE|Vrf to represent object existence. VNETOrch is not do so now.
+             * VRFOrch/VNETOrch add/del state VRF_OBJECT_TABLE|Vrf to represent object existence.
              */
             if (consumer.getTableName() == CFG_VXLAN_EVPN_NVO_TABLE_NAME)
             {
@@ -347,8 +354,60 @@ void VrfMgr::doTask(Consumer &consumer)
             }
             else
             {
-                m_appVnetTableProducer.del(vrfName);
-                m_stateVrfTable.del(vrfName);
+                vector<FieldValueTuple> temp;
+
+                if (m_stateVrfTable.get(vrfName, temp))
+                {
+                    /*
+                     * Invariant: when m_stateVrfTable says the VNET device
+                     * exists, VNetOrch must have observed our APP_DB add and
+                     * populated STATE_VRF_OBJECT_TABLE. If isVrfObjExist() is
+                     * false here, either VNetOrch has not yet processed the
+                     * addOperation, or the state row was never written. In
+                     * either case, defer until VNetOrch catches up; emit a
+                     * rate-limited warn after a sustained backlog so a
+                     * silent infinite-defer (e.g. orchagent stuck or the
+                     * STATE_VRF_OBJECT_TABLE write was skipped) does not
+                     * go unnoticed.
+                     */
+                    if (!isVrfObjExist(vrfName))
+                    {
+                        auto& retries = m_vnetDelRetries[vrfName];
+                        if (++retries % VNET_DEL_DEFER_WARN_THRESHOLD == 0)
+                        {
+                            SWSS_LOG_WARN("VNET '%s' delete deferred %u times "
+                                "waiting for VNetOrch to populate "
+                                "STATE_VRF_OBJECT_TABLE",
+                                vrfName.c_str(), retries);
+                        }
+                        it++;
+                        continue;
+                    }
+
+                    m_appVnetTableProducer.del(vrfName);
+                    m_stateVrfTable.del(vrfName);
+                }
+
+                /*
+                 * Wait for VNetOrch to clear STATE_VRF_OBJECT_TABLE after it
+                 * processes the APP_DB delete above. Same rate-limited warn
+                 * to surface a stuck handshake.
+                 */
+                if (isVrfObjExist(vrfName))
+                {
+                    auto& retries = m_vnetDelRetries[vrfName];
+                    if (++retries % VNET_DEL_DEFER_WARN_THRESHOLD == 0)
+                    {
+                        SWSS_LOG_WARN("VNET '%s' delete deferred %u times "
+                            "waiting for VNetOrch to clear "
+                            "STATE_VRF_OBJECT_TABLE",
+                            vrfName.c_str(), retries);
+                    }
+                    it++;
+                    continue;
+                }
+
+                m_vnetDelRetries.erase(vrfName);
             }
 
             if (consumer.getTableName() != CFG_VXLAN_EVPN_NVO_TABLE_NAME)
--- a/cfgmgr/vrfmgr.h
+++ b/cfgmgr/vrfmgr.h
@@ -42,6 +42,14 @@ private:
     std::set<uint32_t> m_freeTables;
     VRFNameVNIMapTable m_vrfVniMapTable;
 
+    /*
+     * Per-VNET retry counter for the delete-time wait on VNetOrch
+     * STATE_VRF_OBJECT_TABLE handshake. Used to emit a rate-limited
+     * SWSS_LOG_WARN if a VNET delete is deferred for an unreasonably long
+     * time, so a stuck handshake fails loudly instead of hanging silently.
+     */
+    std::unordered_map<std::string, uint32_t> m_vnetDelRetries;
+
     Table m_stateVrfTable, m_stateVrfObjectTable;
     ProducerStateTable m_appVrfTableProducer, m_appVnetTableProducer, m_appVxlanVrfTableProducer;
 };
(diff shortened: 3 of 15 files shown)


fpmsyncd/routesync.cpp
```diff
@@ -2054,6 +2054,22 @@ void RouteSync::onMsg(int nlmsg_type, struct nl_object *obj)
...
     {
+        struct rtnl_link *link_obj = (struct rtnl_link *)obj;
+        const char *link_name = rtnl_link_get_name(link_obj);
+        int link_ifindex = rtnl_link_get_ifindex(link_obj);
+
+        if (link_name)
+        {
+            if (strncmp(link_name, VRF_PREFIX, strlen(VRF_PREFIX)) == 0 ||
+                strncmp(link_name, VNET_PREFIX, strlen(VNET_PREFIX)) == 0 ||
+                strncmp(link_name, MGMT_VRF_PREFIX, strlen(MGMT_VRF_PREFIX)) == 0)
```
wsycqyz, line 2533: Minor (no change requested): `MGMT_VRF_PREFIX` is `"mgmt"` and `strncmp(link_name, MGMT_VRF_PREFIX, strlen(MGMT_VRF_PREFIX)) == 0` will match any link name that starts with `mgmt` (e.g. `mgmt0`, `mgmt-foo`). The actual mgmt VRF device name is exactly `mgmt`, so `strcmp(link_name, MGMT_VRF_PREFIX) == 0` would be a tighter check. Not blocking — just noting it for a future cleanup.

nsoma26: @wsycqyz Thank you for the comment. Agree with you completely. There are other instance where we can do this. Perhaps we can fix all at once. If you want me, I can fix this instance now.

nsoma26:
@mssonicbld  There is a check pending for a long time. Could you please check?
Semgrep Expected — Waiting for status to be reported

nsoma26:
@wsycqyz All checks have passed. Can we merge this PR if there are no further comments?

nsoma26:
@prsunny Could you please review the changes?

StormLiangMS:
@nsoma-cisco 
A few review comments on the coordination logic — overall the approach (mirroring the existing `VRFOrch`/`vrfmgrd` two-phase handshake) is sound. Three things worth addressing before merge:

### 1. Silent fallback when `stateDb == nullptr` in `VNetOrch` constructor

In `orchagent/vnetorch.cpp`:

```cpp
if (stateDb)
{
    m_stateVrfObjectTable = std::unique_ptr<Table>(new Table(stateDb, STATE_VRF_OBJECT_TABLE_NAME));
}
```

Production (`OrchDaemon::init()`) always passes `m_stateDb`, so today this is fine. But the silent fallback is fragile: if any future caller forgets to pass `stateDb` (or passes `nullptr`), `m_stateVrfObjectTable` stays null → `addOperation` skips the `hset` → `vrfmgrd::isVrfObjExist()` always returns false → `vrfmgrd` skips the deferral and calls `delLink()` immediately. That silently re-introduces the exact race this PR is trying to fix, with no log trail to debug it.

Suggest one of:
- Make `stateDb` a required parameter and update the three test callers (`flowcounterrouteorch_ut.cpp`, `intfsorch_ut.cpp`, `mock_orch_test.cpp`).
- At minimum, add a `SWSS_LOG_WARN` in the constructor when `stateDb` is null so the misconfiguration is visible.

### 2. Pre-existing infinite-defer trap is now duplicated in `vrfmgr.cpp`

The new VNet branch (~lines 348–370) mirrors the VRF branch shape:

```cpp
if (m_stateVrfTable.get(vrfName, temp))
{
    if (!isVrfObjExist(vrfName)) { it++; continue; }
    m_appVnetTableProducer.del(vrfName);
    m_stateVrfTable.del(vrfName);
}
```

If `isVrfObjExist` returns false on the **first** iteration (e.g., the state row was never written — see #1 above, or orchagent hasn't yet processed `addOperation` when delete arrives), we `it++; continue;` and never publish the `del` to APP_DB. orchagent then has nothing to process, so `isVrfObjExist` stays false forever and the delete is deferred indefinitely (and the misleading error chain returns).

This trap is pre-existing on the VRF path, so I'm not asking you to fix the original here, but duplicating it does propagate it. Worth at least a comment describing the assumed invariant ("if `m_stateVrfTable` says the device exists, orchagent must have observed the create and written `STATE_VRF_OBJECT_TABLE`"), and ideally a rate-limited warn after N retries so this fails loudly instead of hanging silently.

### 3. Document the retry-safety of `m_stateVrfObjectTable->del` placement in `delOperation`

```cpp
vnet_table_.erase(vnet_name);

if (m_stateVrfObjectTable)
{
    m_stateVrfObjectTable->del(vnet_name);
}
```

Today this is correct: the `del` happens only on the success path (after `vnet_table_.erase`), so when `getRouteCount() != 0` or `removeVxlanTunnelMap` fails (returning `false`) the state row is intentionally kept and the framework retries. But that's load-bearing for the whole handshake to work, and it's not obvious from the code. A short comment near the `del` saying something like "must remain on the success path so vrfmgrd keeps deferring `delLink()` on retry" would prevent a future refactor from accidentally lifting it earlier or wrapping it in the wrong branch.

nsoma26:
@StormLiangMS Appreciate the meticulous review. All three points have been addressed and amended the commits.

1. Silent fallback when stateDb == nullptr in VNetOrch constructor
Made stateDb a required parameter and changed m_stateVrfObjectTable from std::unique_ptr to a value Table member, constructed in the initializer list. Any future caller that forgets to wire stateDb will fail to compile rather than silently re-introduce the race. Also, updated the three test call sites.

2. Pre-existing infinite-defer trap propagated into the VNET branch
Added the invariant comment you suggested (paraphrased: "if m_stateVrfTable says the device exists, VNetOrch must have observed the create and written STATE_VRF_OBJECT_TABLE") at both it++; continue; sites in the new VNET delete path. Also added a small per-vrfName retry counter (std::unordered_map<std::string, uint32_t> m_vnetDelRetries) and a rate-limited SWSS_LOG_WARN every 20 deferrals, so a stuck handshake fails loudly:
**VNET 'Vnet_5037' delete deferred 20 times waiting for VNetOrch to populate STATE_VRF_OBJECT_TABLE**
The counter is reset on the success path. Per your note, I scoped the change to the new VNET branch only and didn't touch the pre-existing VRF path.

3. Document the retry-safety of m_stateVrfObjectTable.del placement in delOperation
Added a comment immediately above the del call calling out that the placement is load-bearing, so vrfmgrd keeps deferring delLink() on retry when a partial cleanup failure leaves us with getRouteCount() != 0 or removeVxlanTunnelMap() == false.

nsoma26:
Pushed an update addressing the `coverage.Azure.sonic-swss.vstest` failure.

The amend to @StormLiangMS's review comments added executable lines to `cfgmgr/vrfmgr.cpp` (the per-`vrfName` retry counter and rate-limited `SWSS_LOG_WARN`) that fall in a tight VNET delete-time race window unreachable from the existing test suite, plus three executable lines in `orchagent/vnetorch.cpp` (the `STATE_VRF_OBJECT_TABLE` `hset` on add, `del` on remove, and the trailing `SWSS_LOG_NOTICE`) — `vnetorch.cpp` is at ~5% overall coverage today because no existing mock test exercises `VNetOrch::addOperation` / `delOperation`. Diff coverage on the prior head dropped from 94% to 76%.

Added a focused mock test `tests/mock_tests/vnetorch_ut.cpp` that drives `APP_VNET_TABLE` SET/DEL through the existing `MockOrchTest` scaffolding (already wires `VxlanTunnelOrch` + `VNetOrch` against the SAI VS backend) and asserts the `STATE_VRF_OBJECT_TABLE['Vnet_ut']` row appears on add and disappears on del — covering all three uncovered `vnetorch.cpp` lines. Folded into the **\`[vnetorch][vrfmgrd]: Defer VNET kernel device deletion until orchagent cleanup completes\`** commit so each commit stays self-contained — bisect-friendly and per-commit reviewable.

Projected diff coverage now ~83%, comfortably above the 80% gate. The remaining uncovered lines are the race-window-only block in `cfgmgr/vrfmgr.cpp` (the first-iteration deferral, only reachable when `vrfmgrd` polls before `VNetOrch` has populated `STATE_VRF_OBJECT_TABLE`), which can't be deterministically forced from black-box VS or mock harnesses; if anyone wants those covered, the right vehicle is a small `vrfmgrd` mock harness in a follow-up PR rather than a flaky timing test here.

wsycqyz:
@StormLiangMS @prsunny @nsoma-cisco 

Manual cherry-pick is here https://github.com/sonic-net/sonic-swss/pull/4567

nsoma26:
@StormLiangMS Following up on point #2.

The invariant comment + rate-limited warn you asked for are in. The defer-forever path still defers forever — just loudly now — and I'd like to address the real issue with your guidance, if it's alright with you.

Proposal: in `vrfmgrd::doTask`, drop the first-stage `if (!isVrfObjExist) { it++; continue; }` block on the VNET branch; publish the APP_DB DEL unconditionally; keep the second-stage wait. The framework on the orchagent side handles the rest:

- `Consumer::addToSync` is keyed by object name, so in the `ADD(pending) → DEL(came in)` race the DEL overwrites the still-pending ADD in `m_toSync`. No new logic needed in `VNetOrch` to cancel it.
- `VNetOrch::delOperation` already returns `true` (no-op) when the VNET isn't in the live table, so the framework drops the m_toSync row. The modify case (ADD finished → ADD pending modify → DEL) still goes through the normal cleanup path because the object is still live.
- The second-stage wait in `vrfmgrd` is already correct for both: state row absent → fall through immediately; state row present → wait for `delOperation` to clear it.

The same trap exists on the VRF branch — the handshake is identical. Would you be open to us applying the same fix on the VRF side in this PR or a new PR? I don't have full visibility into the existing VRF code paths and the implications there, so happy to defer to you on whether that's safe to do here or better left for a separate change.

One question before I touch anything: anything in this reasoning you'd push back on? I'm specifically nervous about `m_toSync` edge cases you may have hit in `VRFOrch` history that we haven't seen.

nsoma26:
@wsycqyz @prsunny @StormLiangMS Please let me know if there is anything I need to follow up to proceed further. All tests have passed and review comments have been addressed. Thank you.

prsunny:
IMO, this is introducing some unnecessary cross-dependencies. Can you please share how this issue was found? if its during sonic-mgmt test, isn't it already addressed? Please share how we can simulate this condition.

nsoma26:
> IMO, this is introducing some unnecessary cross-dependencies. Can you please share how this issue was found? if its during sonic-mgmt test, isn't it already addressed? Please share how we can simulate this condition.

@prsunny We have encountered "bgp#fpmsyncd: :- onRouteMsg: Invalid VRF name" error while running vxlan/test_vxlan_route_advertisement.py test suite. When we run into this error, stale routes will be left in APPL_DB and ASIC. This solution is very similar to what we do in VRF path. I have applied the same logic to VNET path. There are existing issues prior to this PR with this approach (applicable to VRF path) as @StormLiangMS pointed out in review comments and we will fix them in a follow up PR for both VRF and VNET paths.

nsoma26:
@prsunny 

> > IMO, this is introducing some unnecessary cross-dependencies. Can you please share how this issue was found? if its during sonic-mgmt test, isn't it already addressed? Please share how we can simulate this condition.
> 
> @prsunny We have encountered "bgp#fpmsyncd: :- onRouteMsg: Invalid VRF name" error while running vxlan/test_vxlan_route_advertisement.py test suite. When we run into this error, stale routes will be left in APPL_DB and ASIC. This solution is very similar to what we do in VRF path. I have applied the same logic to VNET path. There are existing issues prior to this PR with this approach (applicable to VRF path) as @StormLiangMS pointed out in review comments and we will fix them in a follow up PR for both VRF and VNET paths.

@prsunny How do we move forward? As mentioned above, I have not introduced any new unnecessary cross-dependencies. They are the existing ones in VRF path. And we can fix the existing issues in both VRF and VNET paths in a follow up PR.

PR #4502, opened 2026-04-21
[switchorch][mirrororch]: Add sampled port mirroring with truncation support

Janetxxx:
**What I did**
Adds sampled port mirroring with packet truncation support. This PR targets and validates the ERSPAN use case, which is the primary motivation for the feature in the HLD (https://github.com/sonic-net/SONiC/pull/2296).
 
 ##### SwitchOrch — capability discovery
During init, queries SAI via `sai_query_attribute_capability()` and publishes to `STATE_DB` under `SWITCH_CAPABILITY|switch`:
 - `PORT_INGRESS_SAMPLE_MIRROR_CAPABLE` (SAI_PORT_ATTR_INGRESS_SAMPLE_MIRROR_SESSION)
 - `PORT_EGRESS_SAMPLE_MIRROR_CAPABLE`  (SAI_PORT_ATTR_EGRESS_SAMPLE_MIRROR_SESSION)
 - `SAMPLEPACKET_TRUNCATION_CAPABLE`    (SAI_SAMPLEPACKET_ATTR_TRUNCATE_ENABLE)
 
 ##### MirrorOrch — SamplePacket lifecycle
 - Creates/binds a single SAMPLEPACKET object when `sample_rate` is set. Direction-aware binding: RX uses `INGRESS_SAMPLEPACKET_ENABLE` + `INGRESS_SAMPLE_MIRROR_SESSION`, TX uses the `EGRESS_*` equivalents, and BOTH binds ingress and egress with one SAMPLEPACKET.
 - Truncation implies sampling: `truncate_size` without an explicit `sample_rate` defaults the rate to 1 (sample every packet).
 - Capability gating: rejects the session when the required per-direction sampled mirroring (ingress/egress) is not supported; skips truncation if unsupported.
 - Session update is delete-and-recreate only; in-place SET on an existing key is treated as a duplicate and ignored.
 - Cross-feature guard with sFlow: get-before-set on `INGRESS_SAMPLEPACKET_ENABLE` prevents silent overwrite.
 
 ##### SflowOrch 
 `sflowAddPort()` / `sflowUpdateSampleDirection()`check `INGRESS/EGRESS_SAMPLEPACKET_ENABLE` before binding and reject if the port's samplepacket slot is already owned by a non-sFlow feature
 
 **Why I did it**
This is part of the sampled port mirroring with truncation feature. It enables per-port hardware-based sampling on ERSPAN mirror sessions, reducing mirror bandwidth by only mirroring 1 out of every N packets with optional packet truncation.
 
 **How I verified it**
 ##### Unit (mock) tests 
 `switchorch_ut.cpp`, `mirrororch_ut.cpp`:
 - Capability fields written to STATE_DB
 - Sampled vs full mirror path selection
 - SAMPLEPACKET create/remove, MirrorEntry defaults
 - Per-direction capability gating; rejection when ingress/egress sampled mirroring or truncation is unsupported (booleans toggled directly, which a VS test cannot do)
 - Input validation, non-existent session handling, fault-injection on SAMPLEPACKET create/remove
 
 ##### VS integration tests
 - SAMPLEPACKET lifecycle + port attribute binding/cleanup, truncation attrs
 - RX, TX (egress) and BOTH direction binding
 - `truncate_size` without `sample_rate` defaults to `sample_rate=1`
 - Delete-and-recreate updates a session; duplicate SET on an existing key is ignored (no SAI/STATE_DB mutation)
 - Route lifecycle: session stays unprogrammed while inactive, and is torn down / reprogrammed across route withdrawal and restore
 - Backward compat: no SAMPLEPACKET created when `sample_rate` absent
 - STATE_DB field verification, switch capability values, multi-port binding, LAG members, sFlow conflict rejection

**Details if related**

### DIFF
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -31,6 +31,8 @@
 #define MIRROR_SESSION_DST_PORT             "dst_port"
 #define MIRROR_SESSION_DIRECTION            "direction"
 #define MIRROR_SESSION_TYPE                 "type"
+#define MIRROR_SESSION_SAMPLE_RATE          "sample_rate"
+#define MIRROR_SESSION_TRUNCATE_SIZE        "truncate_size"
 
 #define MIRROR_SESSION_DEFAULT_VLAN_PRI 0
 #define MIRROR_SESSION_DEFAULT_VLAN_CFI 0
@@ -47,6 +49,7 @@
 extern sai_switch_api_t *sai_switch_api;
 extern sai_mirror_api_t *sai_mirror_api;
 extern sai_port_api_t *sai_port_api;
+extern sai_samplepacket_api_t *sai_samplepacket_api;
 
 extern sai_object_id_t  gSwitchId;
 extern PortsOrch*       gPortsOrch;
@@ -59,6 +62,9 @@ MirrorEntry::MirrorEntry(const string& platform) :
         dscp(8),
         ttl(255),
         queue(0),
+        sample_rate(0),
+        truncate_size(0),
+        samplepacketId(SAI_NULL_OBJECT_ID),
         sessionId(0),
         refCount(0)
 {
@@ -473,6 +479,14 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
             {
                 entry.type = fvValue(i);
             }
+            else if (fvField(i) == MIRROR_SESSION_SAMPLE_RATE)
+            {
+                entry.sample_rate = to_uint<uint32_t>(fvValue(i));
+            }
+            else if (fvField(i) == MIRROR_SESSION_TRUNCATE_SIZE)
+            {
+                entry.truncate_size = to_uint<uint32_t>(fvValue(i));
+            }
             else
             {
                 SWSS_LOG_ERROR("Failed to parse session %s configuration. Unknown attribute %s", key.c_str(), fvField(i).c_str());
@@ -494,6 +508,22 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
     if (src_ip_initialized && dst_ip_initialized && entry.srcIp.getIp().family != entry.dstIp.getIp().family)
     {
         SWSS_LOG_ERROR("Address family of source and destination IPs is different");
+        return task_process_status::task_invalid_entry;
+    }
+
+    // Truncate size requires sampled mirroring to be enabled
+    if (entry.truncate_size > 0 && entry.sample_rate == 0)
+    {
+        SWSS_LOG_ERROR("Truncate size requires sampled mirroring to be enabled for session %s",
+                       key.c_str());
+        return task_process_status::task_invalid_entry;
+    }
+
+    // Sampled mirroring requires RX direction
+    if (entry.sample_rate > 0 && entry.direction != MIRROR_RX_DIRECTION)
+    {
+        SWSS_LOG_ERROR("Sampled mirroring requires RX direction for session %s",
+                       key.c_str());
         return task_process_status::task_invalid_entry;
     }
 
@@ -811,7 +841,9 @@ bool MirrorOrch::updateSession(const string& name, MirrorEntry& session)
 bool MirrorOrch::setUnsetPortMirror(Port port,
                                     bool ingress,
                                     bool set,
-                                    sai_object_id_t sessionId)
+                                    sai_object_id_t sessionId,
+                                    sai_object_id_t samplepacketId,
+                                    uint32_t sample_rate)
 {
     // Check if the mirror direction is supported by the ASIC
     if (ingress && !m_switchOrch->isPortIngressMirrorSupported())
@@ -826,6 +858,70 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
     }
 
     sai_status_t status;
+
+    if (sample_rate > 0 && ingress)
+    {
+        // Sampled mirroring path: use SAMPLEPACKET_ENABLE + SAMPLE_MIRROR_SESSION
+        sai_attribute_t sp_attr;
+        sp_attr.id = SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE;
+        sp_attr.value.oid = set ? samplepacketId : SAI_NULL_OBJECT_ID;
+
+        sai_attribute_t mirror_attr;
+        mirror_attr.id = SAI_PORT_ATTR_INGRESS_SAMPLE_MIRROR_SESSION;
+        if (set)
+        {
+            mirror_attr.value.objlist.count = 1;
+            mirror_attr.value.objlist.list = &sessionId;
+        }
+        else
+        {
+            mirror_attr.value.objlist.count = 0;
+        }
+
+        // Set/clear in correct order
+        if (set)
+        {
+            // Set: SAMPLEPACKET_ENABLE first, then SAMPLE_MIRROR_SESSION
+            status = sai_port_api->set_port_attribute(port.m_port_id, &sp_attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to set SAMPLEPACKET_ENABLE on port %s, status %d",
+                                port.m_alias.c_str(), status);
+                return false;
+            }
+            status = sai_port_api->set_port_attribute(port.m_port_id, &mirror_attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to set SAMPLE_MIRROR_SESSION on port %s, status %d",
+                                port.m_alias.c_str(), status);
+                // Rollback: clear SAMPLEPACKET_ENABLE
+                sp_attr.value.oid = SAI_NULL_OBJECT_ID;
+                sai_port_api->set_port_attribute(port.m_port_id, &sp_attr);
+                return false;
+            }
+        }
+        else
+        {
+            // Clear: SAMPLE_MIRROR_SESSION first, then SAMPLEPACKET_ENABLE (reverse order)
+            status = sai_port_api->set_port_attribute(port.m_port_id, &mirror_attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to clear SAMPLE_MIRROR_SESSION on port %s, status %d",
+                                port.m_alias.c_str(), status);
+                return false;
+            }
+            status = sai_port_api->set_port_attribute(port.m_port_id, &sp_attr);
+            if (status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to clear SAMPLEPACKET_ENABLE on port %s, status %d",
+                                port.m_alias.c_str(), status);
+                return false;
+            }
+        }
+        return true;
+    }
+
+    // Full mirror path (existing behavior)
     sai_attribute_t port_attr;
     port_attr.id = ingress ? SAI_PORT_ATTR_INGRESS_MIRROR_SESSION:
                            SAI_PORT_ATTR_EGRESS_MIRROR_SESSION;
@@ -896,7 +992,7 @@ bool MirrorOrch::configurePortMirrorSession(const string& name, MirrorEntry& ses
             }
             if (session.direction == MIRROR_RX_DIRECTION  || session.direction == MIRROR_BOTH_DIRECTION)
             {
-                if (!setUnsetPortMirror(port, true, set, session.sessionId))
+                if (!setUnsetPortMirror(port, true, set, session.sessionId, session.samplepacketId, session.sample_rate))
                 {
                     SWSS_LOG_ERROR("Failed to configure mirror session %s port %s",
                         name.c_str(), port.m_alias.c_str());
@@ -905,7 +1001,7 @@ bool MirrorOrch::configurePortMirrorSession(const string& name, MirrorEntry& ses
             }
             if (session.direction == MIRROR_TX_DIRECTION || session.direction == MIRROR_BOTH_DIRECTION)
             {
-                if (!setUnsetPortMirror(port, false, set, session.sessionId))
+                if (!setUnsetPortMirror(port, false, set, session.sessionId, session.samplepacketId, session.sample_rate))
                 {
                     SWSS_LOG_ERROR("Failed to configure mirror session %s port %s",
                         name.c_str(), port.m_alias.c_str());
@@ -1079,6 +1175,18 @@ bool MirrorOrch::activateSession(const string& name, MirrorEntry& session)
 
     session.status = true;
 
+    // Create SamplePacket if sample_rate > 0
+    if (session.sample_rate > 0)
+    {
+        if (!createSamplePacket(name, session))
+        {
+            SWSS_LOG_ERROR("Failed to create samplepacket, removing mirror session %s", name.c_str());
+            sai_mirror_api->remove_mirror_session(session.sessionId);
+            session.status = false;
+            return false;
+        }
+    }
+
     if (!session.src_port.empty() && !session.direction.empty())
     {
         status = configurePortMirrorSession(name, session, true);
@@ -1116,6 +1224,16 @@ bool MirrorOrch::deactivateSession(const string& name, MirrorEntry& session)
         if (status == false)
         {
             SWSS_LOG_ERROR("Failed to deactivate port mirror session %s", name.c_str());
+            return false;
+        }
+    }
+
+    // Remove SamplePacket if it exists
+    if (session.samplepacketId != SAI_NULL_OBJECT_ID)
+    {
+        if (!removeSamplePacket(name, session))
+        {
+            SWSS_LOG_ERROR("Failed to remove samplepacket for session %s", name.c_str());
             return false;
         }
     }
@@ -1175,6 +1293,79 @@ bool MirrorOrch::updateSessionDstMac(const string& name, MirrorEntry& session)
 
     setSessionState(name, session, MIRROR_SESSION_DST_MAC_ADDRESS);
 
+    return true;
+}
+
+
+bool MirrorOrch::createSamplePacket(const string& name, MirrorEntry& session)
+{
+    SWSS_LOG_ENTER();
+
+    sai_attribute_t attrs[4];
+    uint32_t attr_count = 0;
+
+    attrs[attr_count].id = SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE;
+    attrs[attr_count].value.u32 = session.sample_rate;
+    attr_count++;
+
+    attrs[attr_count].id = SAI_SAMPLEPACKET_ATTR_TYPE;
+    attrs[attr_count].value.s32 = SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION;
+    attr_count++;
+
+    if (session.truncate_size > 0)
+    {
+        attrs[attr_count].id = SAI_SAMPLEPACKET_ATTR_TRUNCATE_ENABLE;
+        attrs[attr_count].value.booldata = true;
+        attr_count++;
+
+        attrs[attr_count].id = SAI_SAMPLEPACKET_ATTR_TRUNCATE_SIZE;
+        attrs[attr_count].value.u32 = session.truncate_size;
+        attr_count++;
+    }
+
+    sai_status_t status = sai_samplepacket_api->create_samplepacket(
+        &session.samplepacketId, gSwitchId, attr_count, attrs);
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to create samplepacket for session %s, status %d",
+                       name.c_str(), status);
+        task_process_status handle_status = handleSaiCreateStatus(SAI_API_SAMPLEPACKET, status);
+        if (handle_status != task_success)
+        {
+            return parseHandleSaiStatusFailure(handle_status);
+        }
+    }
+
+    SWSS_LOG_NOTICE("Created samplepacket for session %s, rate %u, truncate %u",
+                    name.c_str(), session.sample_rate, session.truncate_size);
+    return true;
+}
+
+bool MirrorOrch::removeSamplePacket(const string& name, MirrorEntry& session)
+{
+    SWSS_LOG_ENTER();
+
+    if (session.samplepacketId == SAI_NULL_OBJECT_ID)
+    {
+        return true;
+    }
+
+    sai_status_t status = sai_samplepacket_api->remove_samplepacket(session.samplepacketId);
+
+    if (status != SAI_STATUS_SUCCESS)
+    {
+        SWSS_LOG_ERROR("Failed to remove samplepacket for session %s, status %d",
+                       name.c_str(), status);
+        task_process_status handle_status = handleSaiRemoveStatus(SAI_API_SAMPLEPACKET, status);
+        if (handle_status != task_success)
+        {
+            return parseHandleSaiStatusFailure(handle_status);
+        }
+    }
+
+    session.samplepacketId = SAI_NULL_OBJECT_ID;
+    SWSS_LOG_NOTICE("Removed samplepacket for session %s", name.c_str());
     return true;
 }
 
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -784,6 +784,22 @@ void MirrorOrch::setSessionState(const string& name, const MirrorEntry& session,
      fvVector.emplace_back(MIRROR_SESSION_NEXT_HOP_IP, value);
     }
 
+    if (attr.empty() || attr == MIRROR_SESSION_SAMPLE_RATE)
+    {
+        if (session.sample_rate > 0)
+        {
+            fvVector.emplace_back(MIRROR_SESSION_SAMPLE_RATE, to_string(session.sample_rate));
+        }
+    }
+
+    if (attr.empty() || attr == MIRROR_SESSION_TRUNCATE_SIZE)
+    {
+        if (session.truncate_size > 0)
+        {
+            fvVector.emplace_back(MIRROR_SESSION_TRUNCATE_SIZE, to_string(session.truncate_size));
+        }
+    }
+
     m_mirrorTable.set(name, fvVector);
 }
 
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -391,12 +391,7 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
     bool src_ip_initialized = false;
     bool dst_ip_initialized = false;
 
-    auto session = m_syncdMirrors.find(key);
-    if (session != m_syncdMirrors.end())
-    {
-        SWSS_LOG_NOTICE("Failed to create session %s: object already exists", key.c_str());
-        return task_process_status::task_duplicated;
-    }
+
 
     string platform = getenv("platform") ? getenv("platform") : "";
     MirrorEntry entry(platform);
@@ -550,6 +545,22 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
     SWSS_LOG_NOTICE("Created mirror session %s", key.c_str());
 
     return task_process_status::task_success;
+}
+
+task_process_status MirrorOrch::updateEntry(const string& key, const vector<FieldValueTuple>& data)
+{
+    SWSS_LOG_ENTER();
+
+    SWSS_LOG_NOTICE("Updating mirror session %s", key.c_str());
+
+    auto task_status = deleteEntry(key);
+    if (task_status != task_process_status::task_success)
+    {
+        SWSS_LOG_ERROR("Failed to delete existing mirror session %s during update", key.c_str());
+        return task_status;
+    }
+
+    return createEntry(key, data);
 }
 
 task_process_status MirrorOrch::deleteEntry(const string& name)
@@ -1775,7 +1786,14 @@ void MirrorOrch::doTask(Consumer& consumer)
 
         if (op == SET_COMMAND)
         {
-            task_status = createEntry(key, kfvFieldsValues(t));
+            if (sessionExists(key))
+            {
+                task_status = updateEntry(key, kfvFieldsValues(t));
+            }
+            else
+            {
+                task_status = createEntry(key, kfvFieldsValues(t));
+            }
         }
         else if (op == DEL_COMMAND)
         {
--- a/orchagent/mirrororch.cpp
+++ b/orchagent/mirrororch.cpp
@@ -553,14 +553,123 @@ task_process_status MirrorOrch::updateEntry(const string& key, const vector<Fiel
 
     SWSS_LOG_NOTICE("Updating mirror session %s", key.c_str());
 
-    auto task_status = deleteEntry(key);
-    if (task_status != task_process_status::task_success)
+    auto sessionIter = m_syncdMirrors.find(key);
+    if (sessionIter == m_syncdMirrors.end())
     {
-        SWSS_LOG_ERROR("Failed to delete existing mirror session %s during update", key.c_str());
-        return task_status;
+        SWSS_LOG_ERROR("Failed to update non-existent mirror session %s", key.c_str());
+        return task_process_status::task_invalid_entry;
     }
 
-    return createEntry(key, data);
+    auto& session = sessionIter->second;
+
+    // Determine if any immutable fields have changed
+    bool immutable_changed = false;
+    uint32_t new_sample_rate = session.sample_rate;
+    uint32_t new_truncate_size = session.truncate_size;
+
+    for (auto& fv : data)
+    {
+        const auto& field = fvField(fv);
+        const auto& value = fvValue(fv);
+
+        if (field == MIRROR_SESSION_SAMPLE_RATE)
+        {
+            new_sample_rate = to_uint<uint32_t>(value);
+        }
+        else if (field == MIRROR_SESSION_TRUNCATE_SIZE)
+        {
+            new_truncate_size = to_uint<uint32_t>(value);
+        }
+        else if (field == MIRROR_SESSION_SRC_IP ||
+                 field == MIRROR_SESSION_DST_IP ||
+                 field == MIRROR_SESSION_GRE_TYPE ||
+                 field == MIRROR_SESSION_DSCP ||
+                 field == MIRROR_SESSION_TTL ||
+                 field == MIRROR_SESSION_QUEUE ||
+                 field == MIRROR_SESSION_SRC_PORT ||
+                 field == MIRROR_SESSION_DIRECTION ||
+                 field == MIRROR_SESSION_POLICER)
+        {
+            immutable_changed = true;
+        }
+    }
+
+    // If any immutable fields changed, must delete and recreate
+    if (immutable_changed)
+    {
+        SWSS_LOG_NOTICE("Immutable fields changed for session %s, performing delete+recreate", key.c_str());
+        auto task_status = deleteEntry(key);
+        if (task_status != task_process_status::task_success)
+        {
+            SWSS_LOG_ERROR("Failed to delete mirror session %s during update", key.c_str());
+            return task_status;
+        }
+        return createEntry(key, data);
+    }
+
+    // Only mutable fields (sample_rate, truncate_size) changed - update in-place
+    bool sample_rate_changed = (new_sample_rate != session.sample_rate);
+    bool truncate_size_changed = (new_truncate_size != session.truncate_size);
+
+    if (!sample_rate_changed && !truncate_size_changed)
+    {
+        SWSS_LOG_NOTICE("No field changes detected for session %s", key.c_str());
+        return task_process_status::task_success;
+    }
+
+    // If samplepacket exists, update its attributes in-place
+    if (session.samplepacketId != SAI_NULL_OBJECT_ID)
+    {
+        if (sample_rate_changed)
+        {
+            sai_attribute_t attr;
+            attr.id = SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE;
+            attr.value.u32 = new_sample_rate;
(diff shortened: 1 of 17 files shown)


orchagent/mirrororch.cpp
```diff
@@ -522,6 +586,167 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+    if (immutable_changed)
+    {
+        SWSS_LOG_NOTICE("Immutable fields changed for session %s, performing delete+recreate", key.c_str());
+        auto task_status = deleteEntry(key);
+        if (task_status != task_process_status::task_success)
+        {
+            SWSS_LOG_ERROR("Failed to delete mirror session %s during update", key.c_str());
+            return task_status;
+        }
+        return createEntry(key, data);
```
Pterosaur: This recreate path uses only the current SET payload. Partial CONFIG_DB updates can delete a valid session and recreate it with missing fields; please rebuild from the full row or merge with the existing session before delete.

orchagent/mirrororch.cpp
```diff
@@ -522,6 +586,167 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+                return task_process_status::task_failed; // LCOV_EXCL_LINE
+            }
+            session.sample_rate = new_sample_rate;
+            SWSS_LOG_NOTICE("Updated sample_rate to %u for session %s", new_sample_rate, key.c_str());
+        }
+
+        if (truncate_size_changed)
+        {
+            sai_attribute_t attr;
+            attr.id = SAI_SAMPLEPACKET_ATTR_TRUNCATE_SIZE;
```
Pterosaur: Updating truncate_size should also maintain SAI_SAMPLEPACKET_ATTR_TRUNCATE_ENABLE. 0->nonzero may leave truncation disabled, and nonzero->0 may leave it enabled.

orchagent/mirrororch.cpp
```diff
@@ -634,6 +859,22 @@ void MirrorOrch::setSessionState(const string& name, const MirrorEntry& session,
      fvVector.emplace_back(MIRROR_SESSION_NEXT_HOP_IP, value);
     }
 
+    if (attr.empty() || attr == MIRROR_SESSION_SAMPLE_RATE)
+    {
+        if (session.sample_rate > 0)
```
Pterosaur, line 711: When this value becomes 0 the field is omitted, but Table::set merges and does not delete old fields. Please hdel the optional fields or write 0 to avoid stale STATE_DB.

orchagent/sfloworch.cpp
```diff
@@ -115,7 +125,47 @@ bool SflowOrch::sflowAddPort(sai_object_id_t sample_id, sai_object_id_t port_id,
     SWSS_LOG_DEBUG("sflowAddPort  %" PRIx64 " portOid %" PRIx64 " dir %s",
                            sample_id, port_id, direction.c_str());
 
-    if (direction == "both" || direction == "rx")
+    bool need_ingress = (direction == "both" || direction == "rx");
```
Pterosaur, line 148: This protects sflowAddPort, but sflowUpdateSampleDirection still writes SAMPLEPACKET_ENABLE directly and can overwrite another owner during direction changes. Please apply the same pre-check there.

orchagent/sfloworch.cpp
```diff
@@ -141,6 +196,20 @@ bool SflowOrch::sflowAddPort(sai_object_id_t sample_id, sai_object_id_t port_id,
         if (sai_rc != SAI_STATUS_SUCCESS)
         {
             SWSS_LOG_ERROR("Failed to set session %" PRIx64 " on port %" PRIx64, sample_id, port_id);
+
+            if (ingress_applied)
+            {
+                sai_attribute_t rollback_attr;
+                rollback_attr.id = SAI_PORT_ATTR_INGRESS_SAMPLEPACKET_ENABLE;
+                rollback_attr.value.oid = SAI_NULL_OBJECT_ID;
+                sai_status_t rb_rc = sai_port_api->set_port_attribute(port_id, &rollback_attr);
```
Pterosaur: Rollback should restore the previous ingress samplepacket OID, not always clear it. During a rate update this can remove the old valid sFlow binding.

tests/test_mirror_sampled_erspan.py
```diff
@@ -0,0 +1,443 @@
...
+        dvs.state_db.wait_for_field_match("MIRROR_SESSION_TABLE", session, {"status": "active"})
+
+        # Verify full mirror path - no SAMPLEPACKET
+        dvs.asic_db.wait_for_n_keys("ASIC_STATE:SAI_OBJECT_TYPE_SAMPLEPACKET", 0)
+        fvs = dict(dvs.counters_db.get_entry("COUNTERS_PORT_NAME_MAP", ""))
+        port_oid = fvs.get("Ethernet12")
+        port_entry = dvs.asic_db.wait_for_entry("ASIC_STATE:SAI_OBJECT_TYPE_PORT", port_oid)
+        assert "SAI_PORT_ATTR_INGRESS_MIRROR_SESSION" in port_entry
+
+        # Transition to sampled mirror: delete and recreate with sample_rate
```
Pterosaur: This test still uses delete+recreate, so it does not cover the updateEntry sampled/full transition path. Please update CONFIG_DB in place to exercise the new logic.

tests/test_mirror_sampled_erspan.py
```diff
@@ -0,0 +1,443 @@
...
+
+        dvs.setup_db()
+        # Create sampled session with TX direction - should be rejected
+        self.dvs_mirror.create_erspan_session_sampled(
+            session, "1.1.1.1", "2.2.2.2", "0x8949", "8", "64", "0",
+            src_ports="Ethernet12", direction="TX",
+            sample_rate="50000")
+
+        # Session should not be created in STATE_DB
+        dvs.state_db.wait_for_deleted_entry("MIRROR_SESSION_TABLE", session)
```
Pterosaur: This can pass before orchagent processes the invalid config because the STATE_DB entry is already absent. Please add a synchronization point proving the config was consumed/rejected.

Pterosaur (dismissed):
lgtm

Pterosaur (dismissed):

orchagent/mirrororch.cpp
```diff
@@ -382,16 +389,24 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+        }
         SWSS_LOG_NOTICE("Failed to create session %s: object already exists", key.c_str());
         return task_process_status::task_duplicated;
     }
 
+    bool src_ip_initialized = false;
+    bool dst_ip_initialized = false;
+
+
+
```
r12f: duplicated empty lines.

orchagent/mirrororch.cpp
```diff
@@ -473,6 +488,24 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
             {
                 entry.type = fvValue(i);
             }
+            else if (fvField(i) == MIRROR_SESSION_SAMPLE_RATE)
+            {
+                try {
+                    entry.sample_rate = to_uint<uint32_t>(fvValue(i));
+                } catch (const exception& e) { // LCOV_EXCL_LINE: YANG validates input
```
r12f: can we cover this by unit test? this should not have YANG dependency here.

orchagent/mirrororch.cpp
```diff
@@ -497,6 +530,38 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
         return task_process_status::task_invalid_entry;
     }
 
+    // Truncate size requires sampled mirroring to be enabled
+    if (entry.truncate_size > 0 && entry.sample_rate == 0)
+    {
+        SWSS_LOG_ERROR("Truncate size requires sampled mirroring to be enabled for session %s",
```
r12f: I don't think this is true. They are do idempotent attributes.

orchagent/mirrororch.cpp
```diff
@@ -497,6 +530,38 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+                       key.c_str());
+        return task_process_status::task_invalid_entry;
+    }
+
+    // Sampled mirroring requires RX direction
+    if (entry.sample_rate > 0 && entry.direction != MIRROR_RX_DIRECTION)
+    {
+        SWSS_LOG_ERROR("Sampled mirroring requires RX direction for session %s",
```
r12f: A better way for this log is - Failed to enable session x: Only RX direction is supported by sampled mirror session.

orchagent/mirrororch.cpp
```diff
@@ -382,16 +389,24 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
 
-    auto session = m_syncdMirrors.find(key);
-    if (session != m_syncdMirrors.end())
+    if (m_syncdMirrors.find(key) != m_syncdMirrors.end())
     {
+        // Only support update for ERSPAN sessions
+        auto& existing = m_syncdMirrors.find(key)->second;
+        if (existing.type == MIRROR_SESSION_ERSPAN)
+        {
+            return updateEntry(key, data);
```
bingwang-ms: This change is confusing to me. Does this mean we also support updating the regular mirror session (without sample rate or truncated size)?

orchagent/mirrororch.cpp
r12f: also can this be a capability check?

orchagent/mirrororch.cpp
```diff
@@ -497,6 +530,38 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+    // Sampled mirroring requires RX direction
+    if (entry.sample_rate > 0 && entry.direction != MIRROR_RX_DIRECTION)
+    {
+        SWSS_LOG_ERROR("Sampled mirroring requires RX direction for session %s",
+                       key.c_str());
+        return task_process_status::task_invalid_entry;
+    }
+
+    // Platform capability: reject early if sampled mirroring is not supported
+    if (entry.sample_rate > 0 && !m_switchOrch->isPortIngressSampleMirrorSupported())
```
bingwang-ms: Do we also need to check egress support here?

orchagent/mirrororch.cpp
```diff
@@ -522,6 +587,194 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+    if (sessionIter == m_syncdMirrors.end())
+    {
+        SWSS_LOG_ERROR("Failed to update non-existent mirror session %s", key.c_str());
+        return task_process_status::task_invalid_entry;
+    }
+
+    auto& session = sessionIter->second;
+
+    // Determine if any immutable fields have changed
+    bool immutable_changed = false;
```
r12f: immutable_property_changed

orchagent/mirrororch.cpp
```diff
@@ -522,6 +587,194 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+        else if (field == MIRROR_SESSION_POLICER)
+        {
+            if (value != session.policer)
+                immutable_changed = true;
+        }
+    }
+
+    // Sampled <-> full mode switch needs delete+recreate; in-place would
+    // strand the samplepacket OID and port sampled-attr bindings.
+    if ((session.sample_rate == 0) != (new_sample_rate == 0))
```
r12f: does sample rate support sets?

orchagent/mirrororch.cpp
```diff
@@ -1178,6 +1532,168 @@ bool MirrorOrch::updateSessionDstMac(const string& name, MirrorEntry& session)
...
+            SWSS_LOG_ERROR("Failed to clear SAMPLEPACKET_ENABLE on port %s, status %d", // LCOV_EXCL_LINE: SAI VS set always succeeds
+                            phy_port_alias.c_str(), status); // LCOV_EXCL_LINE
+            return false; // LCOV_EXCL_LINE
+        }
+    }
+    return true;
+}
+
+bool MirrorOrch::createSamplePacket(const string& name, MirrorEntry& session)
+{
```
bingwang-ms, line 1444: Can you help me understand what's the purpose of  this function?

orchagent/mirrororch.cpp
```diff
@@ -522,6 +587,194 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+
+            session.truncate_size = new_truncate_size;
+            SWSS_LOG_NOTICE("Updated truncate_size to %u for session %s", new_truncate_size, key.c_str());
+        }
+    }
+    else
+    {
+        // No samplepacket exists (full mirror path) - cannot transition in-place.
+        // User must delete and recreate the session to change mirroring mode.
+        SWSS_LOG_NOTICE("Session %s requires delete and recreate to change mirroring mode", key.c_str());
```
r12f: this handling is incorrect. if we are in this branch, this means previous creation failed. the user may try updating the config to fix this, however, we are now end with a dead-end.

If session is not created, I think we should try recreate it.

orchagent/mirrororch.cpp
```diff
@@ -522,6 +587,194 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+    bool truncate_size_changed = (new_truncate_size != session.truncate_size);
+
+    if (!sample_rate_changed && !truncate_size_changed)
+    {
+        SWSS_LOG_NOTICE("No field changes detected for session %s", key.c_str());
+        return task_process_status::task_success;
+    }
+
+    // If samplepacket exists, update its attributes in-place
+    if (session.samplepacketId != SAI_NULL_OBJECT_ID)
```
r12f: what if sample rate is 0 here?

bingwang-ms:
@Janetxxx Is there a design doc for this change?

orchagent/mirrororch.cpp
```diff
@@ -31,6 +31,8 @@
 #define MIRROR_SESSION_DST_PORT             "dst_port"
 #define MIRROR_SESSION_DIRECTION            "direction"
 #define MIRROR_SESSION_TYPE                 "type"
+#define MIRROR_SESSION_SAMPLE_RATE          "sample_rate"
+#define MIRROR_SESSION_TRUNCATE_SIZE        "truncate_size"
```
kperumalbfn, line 36: This truncation applies to mirror session, so it should be applicable to any mirror session(erspan or port). please update the description.

orchagent/mirrororch.cpp
```diff
@@ -382,16 +389,24 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
-    bool src_ip_initialized = false;
-    bool dst_ip_initialized = false;
 
-    auto session = m_syncdMirrors.find(key);
-    if (session != m_syncdMirrors.end())
+    if (m_syncdMirrors.find(key) != m_syncdMirrors.end())
     {
+        // Only support update for ERSPAN sessions
+        auto& existing = m_syncdMirrors.find(key)->second;
+        if (existing.type == MIRROR_SESSION_ERSPAN)
```
kperumalbfn: why only erspan?

orchagent/mirrororch.cpp
```diff
@@ -473,6 +488,24 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+            else if (fvField(i) == MIRROR_SESSION_SAMPLE_RATE)
+            {
+                try {
+                    entry.sample_rate = to_uint<uint32_t>(fvValue(i));
+                } catch (const exception& e) { // LCOV_EXCL_LINE: YANG validates input
+                    SWSS_LOG_ERROR("Invalid sample_rate for session %s: %s", key.c_str(), e.what()); // LCOV_EXCL_LINE: YANG validates input
+                    return task_process_status::task_invalid_entry; // LCOV_EXCL_LINE
+                } // LCOV_EXCL_LINE
+            }
+            else if (fvField(i) == MIRROR_SESSION_TRUNCATE_SIZE)
```
kperumalbfn, line 491: What if the configured truncation size is not supported by ASIC vendor? Do they reduce it to minimum supported by ASIC?

orchagent/mirrororch.cpp
```diff
@@ -497,6 +530,38 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+    // Platform capability: reject early if sampled mirroring is not supported
+    if (entry.sample_rate > 0 && !m_switchOrch->isPortIngressSampleMirrorSupported())
+    {
+        SWSS_LOG_ERROR("Sampled mirroring not supported on this platform, "
+                       "rejecting session %s", key.c_str());
+        return task_process_status::task_invalid_entry;
+    }
+
+    // Platform capability: reject early if samplepacket truncation is not supported
+    if (entry.truncate_size > 0 && !m_switchOrch->isSamplepacketTruncationSupported())
```
kperumalbfn, line 565: Mirror truncation and sflow sample rate should be independent.

orchagent/mirrororch.cpp
```diff
@@ -522,6 +587,194 @@ task_process_status MirrorOrch::createEntry(const string& key, const vector<Fiel
...
+        }
+
+        auto task_status = deleteEntry(key);
+        if (task_status != task_process_status::task_success)
+        {
+            SWSS_LOG_ERROR("Failed to delete mirror session %s during update", key.c_str());
+            return task_status;
+        }
+        return createEntry(key, fullRow);
+    }
```
kperumalbfn: any reason why the mirror session has to be deleted and created? SAI attributes support both CREATE_AND_SET https://github.com/opencomputeproject/SAI/blob/f60c11f2da3b3d940451c1cefcaa0e4d6c3164e1/inc/saimirror.h#L267

Janetxxx:
> @Janetxxx Is there a design doc for this change?

@bingwang-ms Here is the HLD doc for this change: https://github.com/sonic-net/SONiC/pull/2296

orchagent/mirrororch.cpp
Janetxxx: This is intentional and specified in the HLD (§3.3 / §3.6.2) https://github.com/sonic-net/SONiC/blob/master/doc/port-mirroring/SONiC_Port_Mirroring_HLD.md#sampled-port-mirroring-extension  For any existing ERSPAN session, mutable fields (sample_rate, truncate_size) update in-place and immutable fields trigger delete+recreate. Regular sessions are included: re-applying the same config does nothing (returns success without touching SAI), an immutable change rebuilds.
Is there a specific concern you have with this applying to regular sessions? I'm happy to adjust if needed.

orchagent/mirrororch.cpp
Janetxxx: Currently only RX direction is supported by sampled mirror session, so any session with sample_rate > 0 and a non-RX direction is rejected up front.

orchagent/mirrororch.cpp
Janetxxx: At a high level, this is because SAI models sampling/truncation as a separate SamplePacket object, not as attributes on the mirror session itself. 
`createSamplePacket()` creates the SAI SamplePacket object that backs a sampled mirror session. It builds the attribute list: SAMPLE_RATE, TYPE = SAI_SAMPLEPACKET_TYPE_MIRROR_SESSION, and (when truncate_size > 0) TRUNCATE_ENABLE + TRUNCATE_SIZE. Then calls create_samplepacket() and stores the resulting OID in session.samplepacketId. That OID is later bound to the source port (via INGRESS_SAMPLEPACKET_ENABLE + INGRESS_SAMPLE_MIRROR_SESSION) to activate sampled mirroring.

orchagent/mirrororch.cpp
Janetxxx: @r12f The truncate attributes are defined by SAI on the `SamplePacket` object, not on the mirror session, and SAI describes them as truncation of sampled packets: https://github.com/opencomputeproject/SAI/blob/master/inc/saisamplepacket.h#L117-L137 
 
 In other words, without `sample_rate`, no `SamplePacket` object is created; since `TRUNCATE_ENABLE/TRUNCATE_SIZE` are attributes of that object, there's nothing for `truncate_size` to attach to, so it can't be set without sampling.

orchagent/mirrororch.cpp
Janetxxx: Since we'd previously scoped this to RX-only, my current logic checks the direction first and then the hardware capability - it only accepts RX-direction sessions, and then verifies the platform supports ingress sampled mirroring.
For egress, I do query the capability (`SAI_PORT_ATTR_EGRESS_SAMPLE_MIRROR_SESSION`) and publish it to `STATE_DB`, but `MirrorOrch` currently has no logic that reads or acts on it, and there's no egress creation path implemented.
Would you prefer I accept both RX/TX, check the per-direction capability, and create an egress sample packet when egress is supported?

orchagent/mirrororch.cpp
Janetxxx: Yes, `SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE` is `MANDATORY_ON_CREATE | CREATE_AND_SET` in saisamplepacket.h, so it supports in-place SET after creation. https://github.com/opencomputeproject/SAI/blob/master/inc/saisamplepacket.h#L86-L97

orchagent/mirrororch.cpp
Janetxxx: Thanks, fixed. When there's no SamplePacket object, recreates the session from the full CONFIG_DB row (deleteEntry + createEntry), reusing the same delete+recreate pattern as the immutable-field path. This should cover both a failed activation and a not-yet-activated session.

orchagent/mirrororch.cpp
Janetxxx: new_sample_rate cannot be 0 here. A non-NULL samplepacketId implies a sampled session (session.sample_rate > 0), and reaching the in-place path requires immutable_property_change == false, i.e. the mode-transition guard at [line 669](https://github.com/Janetxxx/sonic-swss/blob/d85eb87883e2578d4473cfa8fa2bcec7cb8a1357/orchagent/mirrororch.cpp#L669) did not fire, so (session.sample_rate==0) == (new_sample_rate==0); since session.sample_rate > 0, this forces new_sample_rate > 0. A sample_rate→0 (disable sampling) is treated as a mode switch and handled by delete+recreate above.

orchagent/mirrororch.cpp
Janetxxx: Updated in the PR description.

orchagent/mirrororch.cpp
Janetxxx: `sample_rate` and `truncate_size` are only valid on ERSPAN sessions, this is enforced at the YANG model layer via when "current()/../type = 'ERSPAN'" on both leaves in https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-mirror-session.yang#L194, and the CLI exposes these fields for ERSPAN only (CLI PR (https://github.com/sonic-net/sonic-utilities/pull/4459)). A SPAN session can therefore never carry these fields in CONFIG_DB.

orchagent/mirrororch.cpp
Janetxxx: The orchagent does not clamp in my current implementation, it passes the value through and relies on SAI to return a non-success status for an unsupported value, which then log and surface as a failure.

orchagent/mirrororch.cpp
Janetxxx: Thanks, I've updated `updateEntry` to apply these attributes in-place via `set_mirror_session_attribute`.
In-place now: src_ip, gre_type, dscp, ttl, queue and policer.

I kept delete+recreate only for cases that need it:
  - type is SAI CREATE_ONLY.
  - dst_ip drives neighbor-resolved DST_MAC / monitor port / VLAN, so a bare SET would leave stale L2 info.
  - src_port / direction are port-object attributes (SAI_PORT_ATTR_*_MIRROR_SESSION) that need rebinding.
  - sampled ↔ full transitions, since the samplepacket object lifecycle changes.

orchagent/mirrororch.cpp
bingwang-ms: Yes, it's mainly about the test scope as we never enabled updating regular mirror session before. This need to be covered by sonic-mgmt test if the change is intentional. Moreover, we also need to consider if the mirror session is already referenced and in use, can we update it?

orchagent/mirrororch.cpp
r12f: it is also defined on mirror:

https://github.com/opencomputeproject/SAI/blob/master/inc/saimirror.h

Image

r12f: sample packet will be used only when sample rate is set.

if sample rate is not set, it falls back to mirror session only.

orchagent/mirrororch.cpp
bingwang-ms: Can you help me understand what's the major difference between sampled mirror session and regular mirror session / sflow / port mirroring? I'm a little confused here

orchagent/mirrororch.cpp
```diff
@@ -1178,6 +1692,168 @@ bool MirrorOrch::updateSessionDstMac(const string& name, MirrorEntry& session)
...
+    if (session.truncate_size > 0)
+    {
+        attrs[attr_count].id = SAI_SAMPLEPACKET_ATTR_TRUNCATE_ENABLE;
+        attrs[attr_count].value.booldata = true;
+        attr_count++;
+
+        attrs[attr_count].id = SAI_SAMPLEPACKET_ATTR_TRUNCATE_SIZE;
+        attrs[attr_count].value.u32 = session.truncate_size;
+        attr_count++;
+    }
```
kperumalbfn, line 1468: @Janetxxx Both Mirror session(SAI_MIRROR_SESSION_ATTR_TRUNCATE_SIZE) and SAI_SAMPLEPACKET_ATTR_TRUNCATE_SIZE has the truncation size. Which one will be given high priority?

orchagent/mirrororch.cpp
```diff
@@ -1178,6 +1692,168 @@ bool MirrorOrch::updateSessionDstMac(const string& name, MirrorEntry& session)
...
+
+    sai_status_t status = sai_samplepacket_api->create_samplepacket(
```
kperumalbfn, line 1470: Also, is it possible to share the mirror session by both regular everflow and sample_session? If not, do we restrict?

orchagent/mirrororch.cpp
Janetxxx: @kperumalbfn , in my current implementation, it can only support truncation on sample packet object.

orchagent/mirrororch.cpp
Janetxxx: agree, thanks, in my current implementation, it only supports truncation size `SAI_SAMPLEPACKET_ATTR_TRUNCATE_ENABLE` on sample packet object. When truncation exists, but sample rate = 0, it will set sample rate to 1 and use sample packet object to do full mirror.

orchagent/mirrororch.cpp
Janetxxx: IMHO they can share the mirror session, because getSessionOid doesn't check sample_rate, so everflow can reference a sampled session.
In my current implementation, sampling is applied at the port binding, not on the mirror session object itself. When everflow also references the session, traffic matched by the ACL is mirrored at full rate, while traffic on the sampled source ports is sampled.

Pterosaur (dismissed):

orchagent/mirrororch.cpp
```diff
@@ -826,6 +924,48 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
...
     sai_status_t status;
+
+    if (sample_rate > 0)
+    {
+        // Sampled mirroring path: dispatch to per-PHY helper; LAG must be
+        // expanded to its member ports because SAMPLEPACKET/SAMPLE_MIRROR_SESSION
+        // are port-level SAI attributes and a LAG OID is not a valid target.
+        if (port.m_type == Port::LAG)
+        {
+            std::vector<Port> portv;
```
kperumalbfn, line 935: How about new LAG members after the sample session is created? How do we update the new ports?

orchagent/mirrororch.cpp
```diff
@@ -826,6 +924,48 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
...
+            m_portsOrch->getLagMember(port, portv);
+            for (const auto &p : portv)
+            {
+                if (p.m_type != Port::PHY)
+                {
+                    SWSS_LOG_ERROR("Failed to locate PHY member port %s of LAG %s",
+                                   p.m_alias.c_str(), port.m_alias.c_str());
+                    return false;
+                }
+                if (!setUnsetSampledMirrorOnPhyPort(p.m_port_id, p.m_alias, set,
```
kperumalbfn, line 945: SAme for LAG member removal. How do we remove the sample session from the old member?

orchagent/mirrororch.cpp
```diff
@@ -1120,6 +1278,16 @@ bool MirrorOrch::deactivateSession(const string& name, MirrorEntry& session)
         }
     }
 
+    // Remove SamplePacket if it exists
+    if (session.samplepacketId != SAI_NULL_OBJECT_ID)
```
kperumalbfn, line 1282: it will be good to remove the mirror session before removing the sample and disable the sample from the port. could you pls check the order.

orchagent/switchorch.cpp
```diff
@@ -1957,6 +1958,87 @@ void SwitchOrch::querySwitchPortMirrorCapability()
...
+        else
+        {
+            fvVector.emplace_back(SWITCH_CAPABILITY_TABLE_PORT_INGRESS_SAMPLE_MIRROR_CAPABLE, "false");
+            m_portIngressSampleMirrorSupported = false;
+        }
+        SWSS_LOG_NOTICE("port ingress sample mirror capability %d", capability.set_implemented);
+    }
+
+    // Check if SAI is capable of handling Port egress sample mirror session
+    status = sai_query_attribute_capability(gSwitchId, SAI_OBJECT_TYPE_PORT,
```
kperumalbfn, line 1992: any CRM for this sample session? If so, please add them

orchagent/mirrororch.cpp
Janetxxx: It will go into the else statement (a few lines below) to bind only the new port:
```
        else if (port.m_type == Port::PHY)
        {
            return setUnsetSampledMirrorOnPhyPort(port.m_port_id, port.m_alias, set,
                        ingress ? MirrorBindDirection::Ingress : MirrorBindDirection::Egress,
                        sessionId, samplepacketId);
        }
```

orchagent/mirrororch.cpp
Janetxxx: It will go to the same logic but with parameter `set=false`, 
```
        else if (port.m_type == Port::PHY)
        {
            return setUnsetSampledMirrorOnPhyPort(port.m_port_id, port.m_alias, set,
                      ...;
        }
```
Then it will clear SAMPLE_MIRROR_SESSION and SAMPLEPACKET_ENABLE. The shared SAMPLEPACKET object survives and is freed only on `deactivateSession`.

orchagent/mirrororch.cpp
Janetxxx: I checked the existing full port mirror teardown order is 1. unbind the port attributes, 2. remove the mirror session object.
I kept the order unchanged and added a step to remove sample packet object:
So the current order is: 1. unbind the port attributes 2. remove sample packet object. 3. remove mirror session object.  
The sample packet and mirror-session objects are independent; do you see any concerns of this removal order?

orchagent/switchorch.cpp
Janetxxx: Thanks @kperumalbfn, do you mean in https://github.com/sonic-net/sonic-swss/blob/master/orchagent/crmorch.cpp? I couldn't find any port mirror (for existing full mirror) or sample packet related resources in it.

orchagent/mirrororch.cpp
```diff
@@ -634,6 +706,30 @@ void MirrorOrch::setSessionState(const string& name, const MirrorEntry& session,
...
+        if (session.truncate_size > 0)
+        {
+            fvVector.emplace_back(MIRROR_SESSION_TRUNCATE_SIZE, to_string(session.truncate_size));
+        }
+        else
+        {
+            m_mirrorTable.hdel(name, MIRROR_SESSION_TRUNCATE_SIZE);
+        }
+    }
+
```
bingwang-ms, line 732: How about directions? Does that also need to be handled here?

orchagent/mirrororch.cpp
```diff
@@ -826,6 +924,48 @@ bool MirrorOrch::setUnsetPortMirror(Port port,
     }
 
     sai_status_t status;
+
+    if (sample_rate > 0)
+    {
+        // Sampled mirroring path: dispatch to per-PHY helper; LAG must be
+        // expanded to its member ports because SAMPLEPACKET/SAMPLE_MIRROR_SESSION
+        // are port-level SAI attributes and a LAG OID is not a valid target.
+        if (port.m_type == Port::LAG)
```
bingwang-ms, line 933: 1. For sampled session, how is the `src_port` and `dst_port` determined?
2. Per our discussion offline, the mirror session is monitoring on a specific port. How is that port specified?
3. How to guarantee the port being monitored is not the same one that the truncated packet egressed?

orchagent/mirrororch.cpp
```diff
@@ -1178,6 +1346,177 @@ bool MirrorOrch::updateSessionDstMac(const string& name, MirrorEntry& session)
...
+            SWSS_LOG_ERROR("Failed to clear SAMPLEPACKET_ENABLE on port %s, status %d",
+                            phy_port_alias.c_str(), status);
+            return false;
+        }
+    }
+    return true;
+}
+
+bool MirrorOrch::createSamplePacket(const string& name, MirrorEntry& session)
+{
```
bingwang-ms, line 1444: Help me understand this
1. Which class determines what packet being mirrored?
2. Which class determines where the destination of the mirroring packet is?
3. Which class determines the sample rate and truncated size?

orchagent/sfloworch.cpp
```diff
@@ -115,8 +125,56 @@ bool SflowOrch::sflowAddPort(sai_object_id_t sample_id, sai_object_id_t port_id,
...
+            && check_attr.value.oid != SAI_NULL_OBJECT_ID
+            && check_attr.value.oid != sample_id
+            && !isSflowSamplePacket(check_attr.value.oid))
+        {
+            SWSS_LOG_ERROR("Port %" PRIx64 " INGRESS_SAMPLEPACKET_ENABLE already bound to "
+                           "OID 0x%" PRIx64 ", cannot bind sFlow",
+                           port_id, check_attr.value.oid);
+            return false;
+        }
+    }
```
bingwang-ms, line 156: Suggest putting these code into a function instead of copying it for Ingress/Egress. The only difference is the API and log string

Janetxxx: Agreed. I will add a helper function to avoid redundancy.

orchagent/mirrororch.cpp
Janetxxx: If direction changes, user will need to remove and recreate the mirror session.

orchagent/mirrororch.cpp
Janetxxx: 1. User can define `src_port` in CLI, then it will be written into CONFIG_DB.
For `dst_port`, do you mean `dst_ip`? User can define collector IP as `dst_ip` in CLI, then RouteOrch will calculate `neighborInfo.portId` which is the `monitor_port`. Then `activateSession` function will write `monitor_port` into SAI.

Janetxxx: 2. User can input a list or a single `src_port` through CLI, then it will be tokenized and looped over in `configurePortMirrorSession` function in MirrorOrch, then it binds the session to each port by writing the session OID to that port's SAI attributes.

Janetxxx: 3. As we discussed offline, it is difficult to handle it statically in the software level because route can be changed and be re-calculated by RouteOrch dynamically, monitor port will be changed correspondingly, so it will be discussed with vendor to make sure it can be handled in the ASIC level.

orchagent/mirrororch.cpp
Janetxxx: 1. User config (`src_port`, `direction`, `sample_rate`), `MirrorOrch` binds the session via `configurePortMirrorSession()` and creates the sampler via `createSamplePacket()`; the ASIC then mirrors packets matching those criteria.

Janetxxx: 2. `MirrorOrch::activateSession` sets `dst_ip` (from user config) and `monitor_port` (resolved via routing) in the SAI `MIRROR_SESSION` object, defining where ERSPAN packets are sent.

Janetxxx: 3. It be handled in `MirrorOrch::createSamplePacket()`, it reads user config (`sample_rate', 'truncate_size`), then creates SAI `SAMPLEPACKET` object with these attributes `SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE` and `SAI_SAMPLEPACKET_ATTR_TRUNCATE_SIZE`.

Pterosaur (approved):

bingwang-ms:
Discussed and addressed comments offline. Approving the PR

bingwang-ms (approved):

prsunny:
@Janetxxx , this is pretty big for 202605 backport. Wouldn't prefer a CP - @r12f , @vaibhavhd , @yxieca for viz

prsunny:
@Janetxxx , would you asses the PR for cherry-pick?
1. No backward compatibility issues
2. Are these changes enabled under a specific configuration?

Janetxxx:
Thanks for your comment @prsunny, I manually built an image with this PR included and ran the existing port mirroring tests `span/test_port_mirroring.py` on a SN5640 T0 testbed to verify that it didn't cause any backward compatibility issues.
image
image

Janetxxx:
@prsunny 2. Yes, these changes are opt-in. Sampled mirroring only activates when users specify `sample_rate` or `truncate_size` in the mirror session. Existing sessions without these optional fields continue to work as full mirrors, no behavior change.

image

Janetxxx:
Manually cherry-picked into 202607 https://github.com/Azure/sonic-swss.msft/pull/254

PR #4503, opened 2026-04-21
ECMP cleanup on member add failure

ypcisco:
**What I did**

- Added proper rollback in routeorch when ECMP NHG member bulk-create fails partway: remove already-created members, delete the NHG SAI object, and correct CRM counters and NHG count. Also clean up zero-refcount MPLS NHs at all failure paths.
- Same cleanup on partial NHG member create failure in vnetorch: remove already-created members, delete the NHG SAI object, and decrement the tunnel NH ref count for each member — the tunnel NH SAI object is only removed if its ref count reaches zero

**Why I did it**

- Partial NHG creation failures left SAI objects orphaned in hardware with no software reference, causing CRM counter drift and NHG count mismatches.

**How I verified it**

Added three mock unit tests:

- routeorch_ut — ECMP route with 2 NHs where the second NHG member create fails; verifies the partial state (created member + NHG object) is fully rolled back and no route or NHG entry is left behind.
- routeorch_ut (MPLS variant) — Same scenario with MPLS-labeled NHs; additionally asserts the non-surviving MPLS NH is cleaned up while the surviving one retains its ref count.
- vnetorch_ut — ECMP VNET route with local endpoints where the second member create fails; verifies the created member and NHG are rolled back.
- vnetorch_ut (Tunnel variant) — ECMP VNET tunnel route with remote endpoints where a member create fails; verifies full rollback with all tunnel next hops removed and no orphaned NHG remaining.

**Details if related**


tests/mock_tests/vnetorch_ut.cpp
```diff
@@ -0,0 +1,572 @@
...
+            static_cast<Orch *>(gNeighOrch)->doTask();
+        }
+
+        void TearDown() override
+        {
+            RestoreSaiApis();
+            DEINIT_SAI_API_MOCK(next_hop);
+            DEINIT_SAI_API_MOCK(next_hop_group);
+
+            gDirectory.m_values.clear();
```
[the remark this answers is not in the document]
ypcisco: updated

tests/mock_tests/routeorch_ut.cpp
```diff
@@ -412,6 +413,7 @@ namespace routeorch_test
         void TearDown() override
         {
             RestoreSaiApis();
+            DEINIT_SAI_API_MOCK(next_hop_group);
             DEINIT_SAI_API_MOCK(route);
             DEINIT_SAI_API_MOCK(next_hop_group);
```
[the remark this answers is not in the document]
ypcisco: updated

orchagent/routeorch.cpp
```diff
@@ -1629,9 +1643,46 @@ bool RouteOrch::addNextHopGroup(const NextHopGroupKey &nexthops)
         auto nhgm_id = nhgm_ids[i];
         if (nhgm_id == SAI_NULL_OBJECT_ID)
         {
-            // TODO: do we need to clean up?
             SWSS_LOG_ERROR("Failed to create next hop group %" PRIx64 " member %" PRIx64 ": %d\n",
                            next_hop_group_id, nhgm_ids[i], status);
```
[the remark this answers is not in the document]
ypcisco: updated

orchagent/routeorch.cpp
```diff
@@ -1629,9 +1643,46 @@ bool RouteOrch::addNextHopGroup(const NextHopGroupKey &nexthops)
...
+            for (size_t j = 0; j < i; j++)
+            {
+                if (statuses[j] != SAI_STATUS_SUCCESS)
+                {
+                    SWSS_LOG_ERROR("Failed to remove next hop group member[%zu] %" PRIx64 ", rv:%d",
+                                   j, nhgm_ids[j], statuses[j]);
+                    task_process_status handle_status = handleSaiRemoveStatus(SAI_API_NEXT_HOP_GROUP, statuses[j]);
+                    if (handle_status != task_success)
+                    {
+                        return parseHandleSaiStatusFailure(handle_status);
```
prabhataravind, line 1664: If member removal fails during rollback, returning here orphans the NHG SAI object, any remaining members, and skips MPLS NH cleanup — the same leak this PR fixes. Make the rollback best-effort: log the failure, continue the loop, and still attempt NHG removal (lines 1670-1681) and MPLS cleanup (line 1684).

orchagent/vnetorch.cpp
```diff
@@ -869,6 +869,40 @@ bool VNetRouteOrch::addNextHopGroup(const string& vnet, const NextHopGroupKey &n
...
                            next_hop_group_id, next_hop_group_member_id, status);
+            
+            for (auto& nhop : next_hop_group_entry.active_members)
+            {
+                sai_status_t cleanup_status = sai_next_hop_group_api->remove_next_hop_group_member(nhop.second);
+                if (cleanup_status != SAI_STATUS_SUCCESS)
+                {
+                    SWSS_LOG_ERROR("Failed to cleanup next hop group member %" PRIx64 ", rv:%d",
+                                   nhop.second, cleanup_status);
+                    return false;
```
prabhataravind, line 880: Same pattern as routeorch: if member removal fails, the function returns without removing the NHG object (line 896) or cleaning up tunnel next hops (lines 887-893). Make cleanup best-effort rather than aborting on the first failure.

orchagent/routeorch.cpp
ypcisco: If member removal fails we cant remove the NHG referencing that member also.
The `handleSaiRemoveStatus` already provide a coarse handling for sai remove failures.
This is in line with `RouteOrch::removeNextHopGroup` implementation also.

orchagent/vnetorch.cpp
ypcisco: similar to routeorch, this is inline with `VNetRouteOrch::removeNextHopGroup`

orchagent/vnetorch.cpp
```diff
@@ -869,6 +869,40 @@ bool VNetRouteOrch::addNextHopGroup(const string& vnet, const NextHopGroupKey &n
         {
             SWSS_LOG_ERROR("Failed to create next hop group %" PRIx64 " member %" PRIx64 ": %d\n",
                            next_hop_group_id, next_hop_group_member_id, status);
+            
+            for (auto& nhop : next_hop_group_entry.active_members)
```
anish-n, line 873: Agree with the changes being made here, but it seems like we will never retry a failed tunnel creation based on the logic here : https://github.com/sonic-net/sonic-swss/blob/master/orchagent/vnetorch.cpp#L1252

I wonder if we should return false so that we can retry the call, if we believe that nexthopgroup creation failure is a retriable error, that could be another change which should be made. But if its not retriable then this is fine.

anish-n (approved):
Approving vnetorch changes, with 1 suggestion about failure retry

prsunny:
@prabhataravind , could you re-revieiw/signoff? @ypcisco , please resolve conflict

orchagent/vnetorch.cpp
```diff
@@ -869,6 +869,40 @@ bool VNetRouteOrch::addNextHopGroup(const string& vnet, const NextHopGroupKey &n
...
+                gCrmOrch->decCrmResUsedCounter(CrmResourceType::CRM_NEXTHOP_GROUP_MEMBER);
+            }
+
+            /* Remove all created tunnel next hops to avoid a refcount leak. */
+            for (auto& member : nhopgroup_members_set)
+            {
+                if (!isLocalEndpoint(vnet, member.second.ip_address))
+                {
+                    NextHopKey nexthop = member.second;
+                    vrf_obj->removeTunnelNextHop(nexthop);
```
prabhataravind: removeTunnelNextHop() can fail here, but its return value is ignored. The NHG cleanup then continues and the function returns false, potentially leaving a tunnel next-hop object/refcount behind. Could we capture and log this failure?

orchagent/routeorch.cpp
prabhataravind: Before this change, NHG member creation failure always returned false. In this new rollback path, parseHandleSaiStatusFailure(task_failed) returns true, which addNextHopGroup() exposes as successful NHG creation even though the group was never inserted into m_syncdNextHopGroups. Could this path always return false after handling cleanup failure?

orchagent/routeorch.cpp
```diff
@@ -1629,9 +1643,46 @@ bool RouteOrch::addNextHopGroup(const NextHopGroupKey &nexthops)
...
+            }
+            
+            sai_status_t remove_status = sai_next_hop_group_api->remove_next_hop_group(next_hop_group_id);
+            if (remove_status != SAI_STATUS_SUCCESS)
+            {
+                SWSS_LOG_ERROR("Failed to remove next hop group %" PRIx64 ", rv:%d", next_hop_group_id, remove_status);
+                task_process_status handle_status = handleSaiRemoveStatus(SAI_API_NEXT_HOP_GROUP, remove_status);
+                if (handle_status != task_success)
+                {
+                    return parseHandleSaiStatusFailure(handle_status);
```
prabhataravind: This has the same issue: parseHandleSaiStatusFailure(task_failed) can return true, causing addNextHopGroup() to report success after rollback failed. Please ensure this member-create failure path always returns false, regardless of the cleanup handler result.

PR #4507, opened 2026-04-22
muxorch: fix shared-MAC neighbor recovery in updateFdb

Ndancejic:
## What I did

Fixes a shared-MAC neighbor recovery bug in `MuxOrch::updateFdb()` introduced by #4152.

When two mux neighbors (e.g. standby ToR + SoC behind the same server) share a MAC, the function-scoped `bool found_existing_mux_neighbor` was set by *any* same-MAC hit in the loop over `mux_nexthop_tb_`. That gated off the fallback recovery block globally, so whenever the first loop re-bound *one* of the neighbors, the *other* — sitting as a plain NeighOrch entry after FDB age-out — stayed stranded and never got promoted back into `mux_nexthop_tb_`.

Replace the boolean with `std::set<NeighborEntry> handled` that tracks the specific IPs the first loop already re-bound. The fallback now always runs when the FDB port is a mux, using `handled` as a skip filter to avoid re-processing.

## Why

Reproduces like this on master:

1. FDB primed → neighbor A added, gets MUX-managed
2. FDB ages out (del_fdb)
3. Neighbor B added while FDB is absent → stranded as plain NeighOrch entry
4. FDB re-learned → `updateFdb` fires, loop (A) re-binds A, sets the flag, fallback skipped
5. B stays stranded: no tunnel route, no ASIC neighbor entry

Added `tests/test_mux.py::test_fdb_aging_after_neighbor_shared_mac` which reproduces this end-to-end and asserts both A and B are MUX-managed with tunnel routes.

## Notes for reviewers

- The outer `for (auto vlan_alias : vlan_ports)` loop in the fallback block is technically redundant — `getMuxPort()` already rejects non-VLAN aliases, so the explicit alias filter and `getAllVlans()` iteration aren't needed. I left it alone to keep this PR focused on the bugfix; happy to do a follow-up cleanup if reviewers agree.
- No change to the first loop's logic — only the gating bool was replaced.

## How I tested

- `test_fdb_aging_after_neighbor_shared_mac` — new, fails on master, passes with fix
- Existing mux tests pass locally

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -1,6 +1,7 @@
 #include <cassert>
 #include <string>
 #include <vector>
+#include <set>
 #include <unordered_map>
 #include <unordered_set>
 #include <stdexcept>
@@ -1818,7 +1819,7 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
     NeighborEntry neigh;
     MacAddress mac;
     MuxCable* ptr;
-    bool found_existing_mux_neighbor = false;
+    std::set<NeighborEntry> handled;
 
     // Handle existing MUX neighbors that might be moving between ports
     for (auto nh = mux_nexthop_tb_.begin(); nh != mux_nexthop_tb_.end(); ++nh)
@@ -1829,7 +1830,7 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
             continue;
         }
 
-        found_existing_mux_neighbor = true;
+        handled.insert(nh->first);
 
         if (nh->second != update.entry.port_name)
         {
@@ -1852,15 +1853,17 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
         }
     }
 
-    // Handle case where neighbor exists but is not yet a MUX neighbor
-    // This can happen when FDB entry is learned after neighbor is added
-    if (!found_existing_mux_neighbor && isMuxExists(update.entry.port_name))
+    // Fallback: convert any NeighOrch entry that shares this MAC on the FDB's
+    // mux port but is not yet in mux_nexthop_tb_. Always runs when the FDB
+    // port is a mux; the `handled` set prevents re-processing neighbors that
+    // the loop above already re-bound.
+    if (isMuxExists(update.entry.port_name))
     {
         // Check if there's an existing neighbor with this MAC on any VLAN interface
         // that could be converted to a MUX neighbor
         auto vlan_ports = gPortsOrch->getAllVlans();
-        SWSS_LOG_INFO("FDB update without mux neighbor on mux port %s, mac %s", update.entry.port_name.c_str(),
-                update.entry.mac.to_string().c_str());
+        SWSS_LOG_INFO("FDB update on mux port %s, mac %s — scanning for stranded neighbors",
+                update.entry.port_name.c_str(), update.entry.mac.to_string().c_str());
 
         for (auto vlan_alias : vlan_ports)
         {
@@ -1870,6 +1873,12 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
             {
                 const NeighborEntry& neighbor_entry = neighbor_pair.first;
                 const auto& neighbor_data = neighbor_pair.second;
+
+                // Already re-bound by the loop above — skip.
+                if (handled.count(neighbor_entry))
+                {
+                    continue;
+                }
 
                 // Skip prefix_route neighbors that are not skip neighbors
                 // soc neighbors will get added with prefix_route but
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -1859,50 +1859,44 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
     // the loop above already re-bound.
     if (isMuxExists(update.entry.port_name))
     {
-        // Check if there's an existing neighbor with this MAC on any VLAN interface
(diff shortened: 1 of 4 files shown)


Ndancejic:
@manamand2020 for review + visibility

prsunny:
@manamand2020 , signoff?

prsunny:
@Ndancejic , the link in description points to a closed PR - Fixes a shared-MAC neighbor recovery bug in MuxOrch::updateFdb() introduced by https://github.com/sonic-net/sonic-swss/pull/4491.

Ndancejic:
> @Ndancejic , the link in description points to a closed PR - Fixes a shared-MAC neighbor recovery bug in MuxOrch::updateFdb() introduced by #4491.

Updated, linked PR should have been the original prefix-route PR #4152

tests/test_mux.py
```diff
@@ -2117,6 +2117,113 @@ def test_fdb_after_neighbor_on_active_mux(
...
+
+        Scenario (two neighbors sharing one MAC on a standby mux port, e.g.
+        IPv6 link-local + global on the same server NIC):
+
+        1. FDB for MAC X is present on Ethernet0 (standby).
+        2. Neighbor A (IP_A / MAC X) is learned -> fully converted to a MUX
+           neighbor, present in mux_nexthop_tb_.
+        3. FDB for MAC X ages out.
+        4. Neighbor B (IP_B / MAC X) is learned while FDB is absent.
+           On master (post-#4459) updateNeighbor returns early because
```
manamand2020: Fix the prefix route PR number here as well.

orchagent/muxorch.cpp
```diff
@@ -1852,15 +1853,17 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
...
     {
         // Check if there's an existing neighbor with this MAC on any VLAN interface
         // that could be converted to a MUX neighbor
         auto vlan_ports = gPortsOrch->getAllVlans();
-        SWSS_LOG_INFO("FDB update without mux neighbor on mux port %s, mac %s", update.entry.port_name.c_str(),
-                update.entry.mac.to_string().c_str());
+        SWSS_LOG_INFO("FDB update on mux port %s, mac %s — scanning for stranded neighbors",
+                update.entry.port_name.c_str(), update.entry.mac.to_string().c_str());
 
         for (auto vlan_alias : vlan_ports)
```
manamand2020: Looks like this vlan loop is not really needed even for the original code.

manamand2020 (commented):
LGTM

prsunny:
@Ndancejic , can you address copilot comments

orchagent/muxorch.cpp
```diff
@@ -1871,6 +1874,12 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
                 const NeighborEntry& neighbor_entry = neighbor_pair.first;
                 const auto& neighbor_data = neighbor_pair.second;
 
+                // Already re-bound by the loop above — skip.
+                if (handled.count(neighbor_entry))
```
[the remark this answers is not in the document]
Ndancejic: Fixed in latest

orchagent/muxorch.cpp
```diff
@@ -1852,15 +1853,17 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
...
-    if (!found_existing_mux_neighbor && isMuxExists(update.entry.port_name))
+    // Fallback: convert any NeighOrch entry that shares this MAC on the FDB's
+    // mux port but is not yet in mux_nexthop_tb_. Always runs when the FDB
+    // port is a mux; the `handled` set prevents re-processing neighbors that
+    // the loop above already re-bound.
+    if (isMuxExists(update.entry.port_name))
     {
         // Check if there's an existing neighbor with this MAC on any VLAN interface
         // that could be converted to a MUX neighbor
         auto vlan_ports = gPortsOrch->getAllVlans();
```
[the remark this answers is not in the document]
Ndancejic: removed vlan loop

tests/test_mux.py
Ndancejic: fixed

orchagent/muxorch.cpp
Ndancejic: remove the vlan loop

manamand2020 (approved):

prabhataravind:
@Ndancejic could you address coverage?

prsunny:
@Ndancejic , would you double check? i do see other PRs test passing

prsunny:
Coverage is shown 10%:

https://dev.azure.com/mssonic/build/_build/results?buildId=1124128&view=codecoverage-tab

Ndancejic:
Pushed a test-only reorder: moved the three new FDB tests (`test_fdb_after_neighbor_on_standby_mux`, `test_fdb_after_neighbor_on_active_mux`, `test_fdb_aging_after_neighbor_shared_mac`) to run right after `test_NH` and before `test_multi_nexthop`. Pure relocation, no content changes.

Context: the diff-coverage check reports 0 hits on the new `updateFdb` paths (lines 1827-1899). Locally all three tests exercise every branch (verified with debug instrumentation), and they also pass in the Test stage on the first try. However, `test_mux.py` has been getting retried in CI when `test_multi_nexthop` flakes, so coverage data from the failed run may be getting lost on teardown. Running these tests earlier in the file should let them complete on a clean DVS before any retry happens.

prsunny (approved):

PR #4508, opened 2026-04-23
[202511][muxorch] Backport FDB-after-neighbor conversion for host-route mode

Ndancejic:
## Summary

Backport of master PRs #4459 and #4507 to `202511` (host-route only). Fixes a `dualtor_neighbor_check` failure case where neighbors stranded on a standby mux port are never converted to tunnel next hops.

## Problem

Neighbors learned on a mux Vlan port before the FDB learns their MAC are installed as direct Vlan next hops and never reconciled to tunnel next hops when the FDB entry finally arrives, leaving them stranded as black-hole routes on standby ports.

Specifically:
- `NeighOrch::addNeighbor` calls `MuxOrch::isNeighborActive`, which defaults to `true` when the FDB lookup fails — so the neighbor is installed as a direct Vlan next hop (active-side path).
- `MuxOrch::updateNeighbor` is called next, but `getMuxPort` also fails the FDB lookup and returns early without a port, so no reconciliation happens.
- When the FDB entry arrives later, `MuxOrch::updateFdb` only walks `mux_nexthop_tb_` — which does not contain these stranded neighbors — so they are never converted.

## Fix

`MuxOrch::updateFdb` gains a fallback scan: on any FDB add for a mux port, walk `NeighOrch`'s synced neighbors and convert any same-MAC entry whose Vlan resolves via `getMuxPort` to this port. A per-IP `handled` set tracks neighbors already processed by the existing `mux_nexthop_tb_` loop, preventing the shared-MAC masking bug that #4507 addressed.

## What's included

- **#4459** — [muxorch] Fix FDB-after-neighbor conversion for host-route mode
- **#4507** — muxorch: fix shared-MAC neighbor recovery in updateFdb

Master's prefix-route branching (`convertNeighborToMux()` helper, `prefix_route && !isSkipNeighbor` skip) is omitted — `202511` has no prefix-route infrastructure (no #4152 / `MuxNbrHandlerType` / `isMuxCablePrefixBased`). The fix is adapted for host-route only via a direct `MuxCable::updateNeighbor(nh, true)` call.

Master's outer `vlan_ports` loop is also omitted — `getMuxPort` already filters non-Vlan interfaces, making the outer loop redundant.

## Tests

Adds three unit tests to `tests/test_mux.py` ported from master+#4507:
- `test_fdb_after_neighbor_on_standby_mux` — neighbor-then-FDB on standby port converts to tunnel NH
- `test_fdb_after_neighbor_on_active_mux` — neighbor-then-FDB on active port installs direct NH
- `test_fdb_aging_after_neighbor_shared_mac` — shared-MAC regression guard (#4507)

### DIFF
--- a/orchagent/muxorch.cpp
+++ b/orchagent/muxorch.cpp
@@ -1,4 +1,5 @@
 #include <cassert>
+#include <set>
 #include <string>
 #include <vector>
 #include <unordered_map>
@@ -1459,6 +1460,7 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
     NeighborEntry neigh;
     MacAddress mac;
     MuxCable* ptr;
+    std::set<NeighborEntry> handled;
     for (auto nh = mux_nexthop_tb_.begin(); nh != mux_nexthop_tb_.end(); ++nh)
     {
         auto res = neigh_orch_->getNeighborEntry(nh->first, neigh, mac);
@@ -1466,6 +1468,8 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
         {
             continue;
         }
+
+        handled.insert(nh->first);
 
         if (nh->second != update.entry.port_name)
         {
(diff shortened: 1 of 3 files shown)


Ndancejic:
@manamand2020 backport of fix to 202511
@prsunny had to reopen with the fix from the issue I mentioned once I understood it. This is needed to prevent dualtor_neighbor_check failures in 202511. Please help review + merge once pipelines pass

prsunny:
@manamand2020 , please review/signoff

@Ndancejic , can you check on the coverage?

orchagent/muxorch.cpp
```diff
@@ -1487,6 +1491,51 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
...
+    // This happens when a neighbor resolves on a mux port before its FDB entry is available:
+    // NeighOrch::addNeighbor installs it as a direct Vlan next hop (active-side path) because
+    // the FDB lookup fails and MuxOrch::isNeighborActive defaults to true. When the FDB entry
+    // finally arrives here, we need to locate any such stranded neighbors by MAC and convert
+    // them via MuxCable::updateNeighbor so the standby path installs a tunnel next hop.
+    if (isMuxExists(update.entry.port_name))
+    {
+        ptr = getMuxCable(update.entry.port_name);
+        const NeighborTable& neighbor_table = gNeighOrch->getNeighborTable();
+        for (const auto& neighbor_pair : neighbor_table)
```
manamand2020, line 1505: Can we have a logic where we do not execute this for loop? Maybe marking the neighbor in nh_tbl with some state when fdb entry is not found while adding them in muxorch.

orchagent/muxorch.cpp
```diff
@@ -1487,6 +1491,51 @@ void MuxOrch::updateFdb(const FdbUpdate& update)
...
+        {
+            const NeighborEntry& neighbor_entry = neighbor_pair.first;
+            const auto& neighbor_data = neighbor_pair.second;
+
+            if (handled.count(neighbor_entry))
+            {
+                continue;
+            }
+
+            if (neighbor_data.mac != update.entry.mac)
```
manamand2020, line 1515: Neighbor mac will not be set to anything because the FDB was not learnt, so this might continue. Check if there is a non-zero mac then continue.

PR #4511, opened 2026-04-23
[bfd]: Honour software_bfd config to force software BFD over HW offload

### LINKED ISSUE #4513: [bfd]: SYSTEM_DEFAULTS|software_bfd should force software BFD over HW offload
### Description

Operator intent expressed via `CONFIG_DB SYSTEM_DEFAULTS|software_bfd` is currently ignored by `BgpGlobalStateOrch`, which decides `bfd_offload` purely from the SAI capability query. When an operator explicitly asks for software BFD, SONiC still programs BFD sessions to `ASIC_DB` (HW offload) as long as SAI reports support.

### Impact

On platforms where SAI reports BFD offload support but SONiC does not have a working HW-offload integration — notably **Broadcom DNX** — the operator is left with no functional BFD:

- The HW path is effectively unusable because SONiC's HW BFD offload support on DNX in sustenance mode is missing.
- The software path cannot be selected today, because `BgpGlobalStateOrch` overrides the operator's `software_bfd=enabled` setting with the SAI flag.

The operator has clearly expressed their choice; SONiC should honour it.

### Current behaviour

`orchagent/bfdorch.cpp` in `BgpGlobalStateOrch::BgpGlobalStateOrch(...)`:

```cpp
bool ipv6 = true;
bfd_offload = (offload_supported(!ipv6) && offload_supported(ipv6));
```

`bfd_offload` is derived purely from SAI capability queries. `SYSTEM_DEFAULTS|software_bfd` is not consulted.

### Expected behaviour

If `CONFIG_DB SYSTEM_DEFAULTS|software_bfd` is set to `enabled`, `BgpGlobalStateOrch` should treat HW offload as disabled so that BFD sessions take the software path (`BFD_SOFTWARE_SESSION_TABLE` in `STATE_DB`) and are handled by `bfdd` in FRR.

One-way override: setting the flag forces software BFD; leaving it unset preserves today's behaviour.

### Proposed change

In `BgpGlobalStateOrch`, after computing `bfd_offload` from SAI, read `SYSTEM_DEFAULTS|software_bfd` from CONFIG_DB and clear `bfd_offload` if `status == "enabled"`.

### Related PR

https://github.com/sonic-net/sonic-swss/pull/4511

rajshekhar-nexthop:
**What I did**

When `CONFIG_DB SYSTEM_DEFAULTS|software_bfd` is `enabled`, `BgpGlobalStateOrch` clears the SAI-derived `bfd_offload` flag so BFD sessions take the software path (FRR `bfdd` via `BFD_SOFTWARE_SESSION_TABLE`) instead of being programmed into `ASIC_DB`. No change to default behaviour when the flag is unset.

**Why I did it**

Fixes #4513. The operator's explicit preference for software BFD (`SYSTEM_DEFAULTS|software_bfd=enabled`) is currently overridden by a pure SAI capability check. On platforms like Broadcom DNX, SAI reports BFD offload as supported but SONiC's HW BFD offload support in sustenance mode is missing — so the HW path is unusable. Today the operator ends up with no working BFD at all: HW can't be used, and the software path can't be selected because the SAI flag wins. Honouring `software_bfd=enabled` is the right fix: it's an explicit operator opt-in that should not be silently overridden.

**How I verified it**

- New DVS test `tests/test_soft_bfd_override.py` (IPv4 + IPv6) with `DVS_ENV=["BFDOFFLOAD=true"]`: sessions land in `BFD_SOFTWARE_SESSION_TABLE` and do **not** appear in `ASIC_DB:SAI_OBJECT_TYPE_BFD_SESSION`; cleanup removes them.
- Live DUT verification of the internal backport: BFD sessions created/removed with `SYSTEM_DEFAULTS|software_bfd=enabled` on hardware that reports HW offload; sessions flowed only through the software path and FRR `show bfd peers` matched expectations.

### DIFF
--- a/orchagent/bfdorch.cpp
+++ b/orchagent/bfdorch.cpp
@@ -733,6 +733,26 @@ BgpGlobalStateOrch::BgpGlobalStateOrch(DBConnector *db, string tableName):
     tsa_enabled = false;
     bool ipv6 = true;
     bfd_offload = (offload_supported(!ipv6) && offload_supported(ipv6));
+
+    /* When the operator explicitly sets SYSTEM_DEFAULTS|software_bfd to
+     * "enabled" in CONFIG_DB, honour that intent and use software BFD
+     * (FRR bfdd) regardless of what the ASIC/SAI reports for HW offload
+     * capability.  Operators who want HW-offloaded BFD should not set
+     * the software_bfd flag in CONFIG_DB.
+     *
+     * This is evaluated once at construction time.  A runtime change to
+     * SYSTEM_DEFAULTS requires a swss restart to take effect.
+     */
+    if (bfd_offload)
+    {
+        string value;
+        Table systemDefaultsTable(db, "SYSTEM_DEFAULTS");
+        if (systemDefaultsTable.hget("software_bfd", "status", value) && value == "enabled")
+        {
+            SWSS_LOG_NOTICE("software_bfd is enabled in SYSTEM_DEFAULTS, overriding HW BFD offload");
+            bfd_offload = false;
+        }
+    }
 }
 
 BgpGlobalStateOrch::~BgpGlobalStateOrch(void)
--- a/orchagent/bfdorch.cpp
+++ b/orchagent/bfdorch.cpp
@@ -734,24 +734,13 @@ BgpGlobalStateOrch::BgpGlobalStateOrch(DBConnector *db, string tableName):
     bool ipv6 = true;
     bfd_offload = (offload_supported(!ipv6) && offload_supported(ipv6));
 
-    /* When the operator explicitly sets SYSTEM_DEFAULTS|software_bfd to
-     * "enabled" in CONFIG_DB, honour that intent and use software BFD
-     * (FRR bfdd) regardless of what the ASIC/SAI reports for HW offload
-     * capability.  Operators who want HW-offloaded BFD should not set
-     * the software_bfd flag in CONFIG_DB.
-     *
-     * This is evaluated once at construction time.  A runtime change to
-     * SYSTEM_DEFAULTS requires a swss restart to take effect.
-     */
-    if (bfd_offload)
+    string value;
+    Table systemDefaultsTable(db, "SYSTEM_DEFAULTS");
+    software_bfd_forced = (systemDefaultsTable.hget("software_bfd", "status", value)
+                           && value == "enabled");
+    if (software_bfd_forced)
     {
-        string value;
-        Table systemDefaultsTable(db, "SYSTEM_DEFAULTS");
-        if (systemDefaultsTable.hget("software_bfd", "status", value) && value == "enabled")
-        {
-            SWSS_LOG_NOTICE("software_bfd is enabled in SYSTEM_DEFAULTS, overriding HW BFD offload");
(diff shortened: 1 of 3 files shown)


rajshekhar-nexthop:
@prsunny Please review this PR. Thanks

orchagent/bfdorch.cpp
```diff
@@ -733,6 +733,26 @@ BgpGlobalStateOrch::BgpGlobalStateOrch(DBConnector *db, string tableName):
...
+    /* When the operator explicitly sets SYSTEM_DEFAULTS|software_bfd to
+     * "enabled" in CONFIG_DB, honour that intent and use software BFD
+     * (FRR bfdd) regardless of what the ASIC/SAI reports for HW offload
+     * capability.  Operators who want HW-offloaded BFD should not set
+     * the software_bfd flag in CONFIG_DB.
+     *
+     * This is evaluated once at construction time.  A runtime change to
+     * SYSTEM_DEFAULTS requires a swss restart to take effect.
+     */
+    if (bfd_offload)
```
prsunny: I'm confused, this is specifically SAI offload capability being reported and not user configuration. If upper layer needs to distinguish, then it should read both tables and make the choice.

prsunny (changes requested):

orchagent/bfdorch.cpp
rajshekhar-nexthop: Thanks for the review! Good point — pushed a refactor. `bfd_offload` is now strictly the SAI capability and the policy lives in `getSoftwareBfd()`, which combines the SAI flag and `SYSTEM_DEFAULTS|software_bfd` and returns the resolved decision (`software_bfd_forced || !bfd_offload`). External behavior is unchanged; the four cases in `test_soft_bfd_override.py` produce identical results.

orchagent/bfdorch.cpp
```diff
@@ -733,6 +733,15 @@ BgpGlobalStateOrch::BgpGlobalStateOrch(DBConnector *db, string tableName):
     tsa_enabled = false;
     bool ipv6 = true;
     bfd_offload = (offload_supported(!ipv6) && offload_supported(ipv6));
+
+    string value;
+    Table systemDefaultsTable(db, "SYSTEM_DEFAULTS");
+    software_bfd_forced = (systemDefaultsTable.hget("software_bfd", "status", value)
```
prsunny: can you rename to "`bfd_config`" to convey that this is a configured bfd state

rajshekhar-nexthop: Thanks. `bfd_config` reads as generic, doesn't capture that this is specifically the software-BFD configured state. Would `bfd_software` work? Stays specific to this flag alongside `bfd_offload` (the SAI capability). Happy to go with `bfd_config` if you'd prefer.

rajshekhar-nexthop: Gentle reminder @prsunny. Please check above comment.

rajshekhar-nexthop: Done — renamed to `bfd_config` in both `bfdorch.h` and `bfdorch.cpp`. Pushed as an amend to the same commit, so the only delta from the previous revision is the rename plus a continuation-line realignment. Thanks for the review.

PR #4514, opened 2026-04-24
FlexCounter support for policer-queue stats

rajath-nexthop:
**What I did**
Added support to query for policer-queue statistics (RED, GREEN and YELLOW) packets - support for which was already present in `sai` using an API - `sai_policer_api->get_policer_stats`. Enable policer binding to `FlexCounterOrch` in swss and to be able to call the SAI get APIs.

**Why I did it**
There was no way to check the policer stats for `COPP` on SONiC at the queue level, only `flowcnt-trap` had statistics about packets hitting specific traps.

**How I verified it**
`show copp stats` should be able to show all the copp statistics per queue. 

```bash
admin@gold221:~$ show copp stats
Trap Group       Total Pkts    Total Bytes    Green Pkts    Green Bytes    Yellow Pkts    Yellow Bytes    Red Pkts    Red Bytes
-------------  ------------  -------------  ------------  -------------  -------------  --------------  ----------  -----------
default                   5            438             5            438              0               0           0            0
queue1_group1             0              0             0              0              0               0           0            0
queue1_group3             0              0             0              0              0               0           0            0
queue4_group1           470         515344           470         515344              0               0           0            0
queue4_group2            16           1232            16           1232              0               0           0            0
queue4_group3             0              0             0              0              0               0           0            0
admin@gold221:~$
```

**Details if related**

Related PRs - 

[sonic-buildimage changes](https://github.com/sonic-net/sonic-buildimage/pull/26972)
[sonic-utilities CLI changes](https://github.com/sonic-net/sonic-utilities/pull/4490)
[sonic-mgmt test changes](https://github.com/sonic-net/sonic-mgmt/pull/24182)


orchagent/flexcounterorch.cpp
```diff
@@ -96,6 +97,7 @@ unordered_map<string, string> flexCounterGroupMap =
     {SRV6_KEY, SRV6_STAT_COUNTER_FLEX_COUNTER_GROUP},
     {SWITCH_KEY, SWITCH_STAT_COUNTER_FLEX_COUNTER_GROUP},
     {HA_SET_KEY, HA_SET_STAT_COUNTER_FLEX_COUNTER_GROUP}
```
Bojun-Feng: Seems like there is a missing comma here?

```suggestion
    {HA_SET_KEY, HA_SET_STAT_COUNTER_FLEX_COUNTER_GROUP},
```

prsunny:
@rajath-nexthop , could you add test to fix coverage?

prsunny:
@kperumalbfn for viz

prsunny:
This is indeed a good feature to have, can you ensure this will not break for those implementations that doesn't support copp stats?

orchagent/flexcounterorch.cpp
```diff
@@ -321,6 +324,28 @@ void FlexCounterOrch::doTask(Consumer &consumer)
                             m_hostif_trap_counter_enabled = false;
                         }
                     }
+                    if (gCoppOrch && (key == COPP_STATS_KEY))
+                    {
+                        // Gate to broadcom-xgs only.
+                        string platform     = getenv("platform")     ? getenv("platform")     : "";
+                        string sub_platform = getenv("sub_platform") ? getenv("sub_platform") : "";
+
+                        if (platform != BRCM_PLATFORM_SUBSTRING || sub_platform == BRCM_DNX_PLATFORM_SUBSTRING)
```
prsunny: We should not be checking platform here instead use SAI capability

tests/test_flex_counters.py
```diff
@@ -142,6 +147,8 @@ def set_only_config_db_buffers_field(self, value):
 
     @pytest.mark.parametrize("counter_type", counter_group_meta.keys())
     def test_flex_counters(self, dvs, counter_type):
+        if counter_type == 'copp_policer_counter':
```
prsunny: Will need to enhance tests to get coverage and not limited to a specific platform.

orchagent/copporch.cpp
```diff
@@ -1525,3 +1554,255 @@ void CoppOrch::clearHostIfTrapCounterIdList()
...
+        return true;
+    }
+
+    // Update COUNTERS_POLICER_NAME_MAP
+    vector<FieldValueTuple> nameMapFvs;
+    nameMapFvs.emplace_back(trap_group_name, sai_serialize_object_id(policer_id));
+    m_policerCounterTable->set("", nameMapFvs);
+
+    // Register policer with FlexCounter — only the SAI-reported subset
+    m_policer_counter_manager.setCounterIdList(policer_id, CounterType::POLICER, supported_stats);
```
Bojun-Feng: With `-c traditional` and async sairedis, this registration can reach syncd before the policer create. Syncd converts the unmapped SET to DEL and doesn't retry (see [`Syncd.cpp:4194`](https://github.com/sonic-net/sonic-sairedis/blob/64d22caca933a2ef9841df192a7860ebea1c84c8/syncd/Syncd.cpp#L4182-L4195)). The name-map check above blocks re-issue, so polling never starts. Consider reusing the `m_pendingAddToFlexCntr` timer and `VIDTORID` guard, similar to `bindTrapCounter`.

orchagent/copporch.cpp
```diff
@@ -1525,3 +1554,255 @@ void CoppOrch::clearHostIfTrapCounterIdList()
...
+        return false;
+    }
+
+    // Check if already bound (avoid duplicate binding)
+    std::string existing_oid;
+    if (m_policerCounterTable->hget("", trap_group_name, existing_oid))
+    {
+        SWSS_LOG_DEBUG("Policer counter already bound for trap group %s", trap_group_name.c_str());
+        return true;
+    }
```
Bojun-Feng, line 1665: After a warm reboot, `COUNTERS_POLICER_NAME_MAP` survives; the replayed policer gets a new VID. This early return skips [`setCounterIdList`](https://github.com/sonic-net/sonic-swss/blob/e905d4d42785fc77313fd773f5a3ce2dff481819/orchagent/copporch.cpp#L1657-L1658), so polling never resumes, and the CLI reads frozen values through the stale OID. Consider process-local dedup like [`m_trap_obj_name_map`](https://github.com/sonic-net/sonic-swss/blob/e905d4d42785fc77313fd773f5a3ce2dff481819/orchagent/copporch.cpp#L1495), or drop the check.

orchagent/copporch.cpp
```diff
@@ -1525,3 +1554,255 @@ void CoppOrch::clearHostIfTrapCounterIdList()
...
+            m_policer_stats_caps_queried = true;
+            return false;
+        }
+        m_policer_stats_caps_queried = true;
+    }
+
+    return std::any_of(m_policer_stats_caps.cbegin(), m_policer_stats_caps.cend(),
+        [stat](const sai_stat_capability_t& c) {
+            return static_cast<sai_policer_stat_t>(c.stat_enum) == stat;
+        });
```
Bojun-Feng, line 1761: Nit: This match ignores `stat_modes`, but the [group polls in `READ` mode](https://github.com/sonic-net/sonic-swss/blob/e905d4d42785fc77313fd773f5a3ce2dff481819/orchagent/copporch.cpp#L217). Syncd filters out a `READ_AND_CLEAR`-only stat, so it's advertised in `STATE_DB` yet never appears in `COUNTERS_DB`. Consider also requiring `(c.stat_modes & SAI_STATS_MODE_READ)`.

PR #4516, opened 2026-04-24
Swss stats use componentstats

yutongzhang-microsoft:
Delegate the generic plumbing (atomic counters, dirty tracking, writer thread, deferred DB connect) to swss::ComponentStats in sonic-swss-common, so other containers (gnmi, bmp, telemetry, ...) can reuse it. SwssStats keeps only the swss-specific metric vocabulary (SET/DEL/COMPLETE/ERROR) and the SWSS_STATS Redis key prefix. Public API and orch.cpp hooks are unchanged.

**What I did**

**Why I did it**

**How I verified it**

**Details if related**

### DIFF
--- a/orchagent/swssstats.h
+++ b/orchagent/swssstats.h
@@ -1,0 +1,54 @@
+#pragma once
+
+#include <string>
+#include <memory>
+#include <atomic>
+#include <cstdint>
+
+// Forward-declare the common library type so users of this header don't need
+// to pull in swss-common transitively.
+namespace swss { class ComponentStats; }
+
+/// Global enable flag. Setting it to false disables recording on the hot path
+/// without tearing down the singleton or its writer thread.
+extern std::atomic<bool> gSwssStatsRecord;
+
+/**
+ * SwssStats — a thin orchagent-specific facade over swss::ComponentStats.
+ *
+ * The generic plumbing (atomic counters, version-based dirty tracking, writer
+ * thread, deferred DB connect, cv-based shutdown) lives in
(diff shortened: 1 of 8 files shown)


orchagent/swssstats.h
```diff
@@ -0,0 +1,73 @@
...
+    /// contribute to ERROR again if they fail again.
+    void recordError(const std::string &table_name, uint64_t count = 1);
+
+    CounterSnapshot getCounters(const std::string &table_name);
+
+private:
+    SwssStats();
+    ~SwssStats() = default;
+
+    std::shared_ptr<swss::ComponentStats> m_impl;
```
r12f: shared_ptr is not multi-thread safe.

yutongzhang-microsoft: You're right that shared_ptr does not by itself make multi-threaded access safe. In this specific case the access pattern is safe because:

1. m_impl is initialised exactly once in SwssStats::SwssStats(), which only runs at the first call to getInstance() under the C++11 thread-safe static-local-initialisation guarantee.
2. It is never reassigned afterwards — only m_impl->... reads happen, never m_impl = ....
3. The pointed-to ComponentStats is itself internally thread-safe (shared_timed_mutex for the structural map + atomics for counter/version values).

The dangerous case — concurrent read+write on the same shared_ptr instance — does not occur here. But the code did not previously *enforce* invariant (2). Pushed 55ca44e that adds the const qualifier to m_impl so the compiler rejects any future attempt to mutate it, plus a doc comment explaining the reasoning above. Behaviour is unchanged at runtime.

PR #4520, opened 2026-04-26
orchagent: Add null check after dynamic_cast in PfcWdOrch

uramanath:
**What I did**

Added a null check after `dynamic_cast<FlexCounterManager*>` in PfcWdOrch queue counter setup.

**Why I did it**

The `dynamic_cast` result is dereferenced directly with `->` on the same line without checking for nullptr. If the cast fails, this is undefined behavior and will crash orchagent.

**How I verified it**

Code inspection — the fix splits the cast into a variable with a null guard before calling `setCounterIdList`. No behavioral change when the cast succeeds. All existing tests pass.


anarasimhan-upscale:
@kperumalbfn, @prsunny, can you please review and merge this PR?

anarasimhan-upscale (approved):
lgtm

PR #4522, opened 2026-04-26
cfgmgr: Fix unchecked npos and wrong substr in NbrMgr::parseAliasIp

uramanath:
**What I did**

Added a `string::npos` guard and fixed an incorrect `substr` length in `NbrMgr::parseAliasIp()`.

**Why I did it**

`find(delimiter)` is called without checking the result against `string::npos`. When the delimiter is absent, `found` is `npos` and the subsequent `substr(found + 1, ...)` call produces wrong results (`npos + 1` wraps to 0 on most platforms).

Additionally, the `ip_address` substr uses `size() - 1` as the length argument instead of the correct tail length. Using single-argument `substr(found + 1)` extracts to end-of-string, which is the intended behavior.

**How I verified it**

Code inspection — the fix adds a npos guard that returns an empty vector with an error log, and corrects the substr call. No behavioral change for well-formed keys. All existing tests pass.


prsunny (dismissed):

tests/mock_tests/nbrmgrd/nbrmgr_ut.cpp
```diff
@@ -201,4 +201,24 @@ namespace nbrmgr_ut
...
+        auto result = nbrmgr.parseAliasIp("NoDelimiterHere", ":");
+        EXPECT_TRUE(result.empty());
+    }
+
+    TEST_F(NbrMgrTest, ParseAliasIpValidKey)
+    {
+        std::vector<std::string> cfg_nbr_tables = {CFG_NEIGH_TABLE_NAME};
+        swss::NbrMgr nbrmgr(m_config_db.get(), m_app_db.get(), m_state_db.get(), cfg_nbr_tables);
+
+        auto result = nbrmgr.parseAliasIp("Ethernet0:10.0.0.1", ":");
```
prsunny, line 219: can you add a test with IPv6 entry as well?

PR #4525, opened 2026-04-28
[vlanmgr] apply VLAN mtu to host netdev

kalash-nexthop:
**What I did**

`VlanMgr::doVlanTask` parsed the mtu field from `CONFIG_DB` `VLAN` but never pushed it through to the kernel via `ip link set Vlan<id> mtu <v>` — only propagated it to APPL_DB.

The 2017 TODO ("support host VLAN mtu setting") is now addressed: after `addHostVlan / updateVlanAdminState`, call the existing `setHostVlanMtu` helper and only write mtu to `APPL_DB` if the kernel accepted it.

A VLAN interface's MTU is independent of its bridge members (unlike team/bond), so the "member configured" half of the TODO doesn't correspond to a real kernel constraint and is dropped. As opposed to vlan, this is handled in portchannel case by forcing port-channel's mtu on it's member interfaces. We can't do this for vlan as the member interfaces could be part of multiple vlans, and do not have 1x1 relationship like portchannel and their members.

DVS test test_VlanHostNetdevMtu covers both the apply and reject paths.

closes# https://github.com/sonic-net/sonic-buildimage/issues/27017

**Why I did it**

`VlanMgr::doVlanTask `parsed the mtu field from `CONFIG_DB` `VLAN` but never pushed it through to the kernel via `ip link set Vlan<id> mtu <v>`

**How I verified it**

Added a DVS test (test_VlanHostNetdevMtu) to verify actual mtu programming in kernel, and that passes:
```
test_vlan.py::TestVlan::test_VlanHostNetdevMtu PASSED                    [  9%]
```

One can manually verify this by doing:
```
redis-cli -n 4 HSET 'VLAN|Vlan100' mtu 1500
ip link show Vlan100
```

**Details if related**

### DIFF
--- a/cfgmgr/vlanmgr.cpp
+++ b/cfgmgr/vlanmgr.cpp
@@ -1,4 +1,5 @@
 #include <string.h>
+#include <limits>
 #include "logger.h"
 #include "producerstatetable.h"
 #include "macaddress.h"
@@ -398,12 +399,6 @@ void VlanMgr::doVlanTask(Consumer &consumer)
                 else if (fvField(i) == "mtu")
                 {
                     mtu = fvValue(i);
-                    /*
-                     * TODO: support host VLAN mtu setting.
-                     * Host VLAN mtu should be set only after member configured
-                     * and VLAN state is not UNKNOWN.
-                     */
-                    SWSS_LOG_DEBUG("%s mtu %s: Host VLAN mtu setting to be supported.", key.c_str(), mtu.c_str());
                 }
                 else if (fvField(i) == "members@") {
                     members = fvValue(i);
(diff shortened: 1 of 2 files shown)


prsunny:
@kalash-nexthop , does this intended to fix this issue? - https://github.com/sonic-net/sonic-buildimage/issues/7134. iirc, there were some limitations on cases like if a member port is added after mtu is applied or not. Can you check the history of the issue and see if those are addressed? @dgsudharsan for viz

kalash-nexthop:
> @kalash-nexthop , does this intended to fix this issue? - https://github.com/sonic-net/sonic-buildimage/issues/7134. iirc, there were some limitations on cases like if a member port is added after mtu is applied or not. Can you check the history of the issue and see if those are addressed? @dgsudharsan for viz

@prsunny I am not handling all that. Automatically adjusting member ports MTU based on vlan interface is not feasible, simply because the same ethernet ports could be part of multiple vlans. And silently adjusting vlan mtu when user requested a different value is confusing (plus we would need to automatially change it back when members mtu changes). Covering all such cases is cumbersome and unnecessarily complex and confusing

And current code is no better- it doesn't even allow user to adjust the mtu on vlan interface, and vlan interfaces' mtu are always set to 9100. With this change, and operator can at least manually adjust that to match the min of it's member ports.

PR #4527, opened 2026-04-28
[orchagent]: Fix sai_fdb_type bleeding across events in FDB notification batch

xq9mend:
**What I did**

Moved `sai_fdb_type` declaration inside the per-event loop in `FdbOrch::doTask(NotificationConsumer&)`. Added a regression test.

**Why I did it**

Previously the variable was declared outside the loop and only updated when `SAI_FDB_ENTRY_ATTR_TYPE` was present in an event's attributes. If one event in a batch set it to `STATIC` and the next event lacked the attribute, the second event would incorrectly inherit `STATIC` type from the first. This could cause `handleSyncdFlushNotif()` to skip flushing entries that should have matched, since it filters on `sai_fdb_type` equality.

**How I verified it**

Added `FdbTypeDoesNotBleedAcrossBatchEvents` unit test to `tests/mock_tests/fdborch/flush_syncd_notif_ut.cpp`. Injects a two-event batch notification via the `NotificationConsumer` queue, verifies the second event (no type attribute) is stored as `DYNAMIC`. Built and ran all unit tests locally: 532 passed, 0 failed.

**Details if related**

N/A

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -1049,13 +1049,12 @@ void FdbOrch::doTask(NotificationConsumer& consumer)
     {
         uint32_t count;
         sai_fdb_event_notification_data_t *fdbevent = nullptr;
-        sai_fdb_entry_type_t sai_fdb_type = SAI_FDB_ENTRY_TYPE_DYNAMIC;
-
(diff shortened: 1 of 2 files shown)


xq9mend:
@prsunny @Ndancejic @saiarcot895

theasianpianist (approved):

PR #4533, opened 2026-04-29
[orchagent]: Consume FDB LEARN/AGED/MOVE events from ProducerStateTable

xq9mend:
**What I did**

Add ConsumerStateTable consumer for FDB_EVENT_STATE table in FdbOrch. LEARN/AGED/MOVE events are now dispatched via doTask(Consumer&) with key-based dedup, so MAC flap storms collapse to one entry per MAC+BVID.

The existing NotificationConsumer path (PUBLISH/SUBSCRIBE) now handles only FLUSH events plus a fallback safety-net for any non-FLUSH events that arrive during transition or in ZeroMQ mode.

Also renames the local sai_fdb_type variable to fdb_type in the NotificationConsumer loop to avoid shadowing.

**Why I did it**

MAC flap storms generate a large number of FDB notifications for the same MAC+BVID in quick succession. With the previous PUBLISH path, each event was queued independently, causing memory pressure and processing delays. ProducerStateTable provides key-based deduplication: multiple events for the same MAC+BVID are merged so only the latest state is processed.

**How I verified it**

VS tests: all 15 FDB tests pass (test_fdb.py, test_fdbsync.py, test_fdb_update.py).

**Details if related**

Depends on sonic-sairedis change: https://github.com/sonic-net/sonic-sairedis/pull/1875

### DIFF
--- a/orchagent/fdborch.cpp
+++ b/orchagent/fdborch.cpp
@@ -12,6 +12,7 @@
 #include "macmoveguard.h"
 #include "crmorch.h"
 #include "notifier.h"
+#include "consumerstatetable.h"
 #include "sai_serialize.h"
 #include "mlagorch.h"
 #include "vxlanorch.h"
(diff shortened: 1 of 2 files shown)


xq9mend:
@prsunny @Ndancejic @saiarcot895

xq9mend:
**CI status:** Same systemic vstest/p4rt infra failure as affecting all sonic-swss-common and related PRs since June 8. Not caused by this change.

This PR consumes FDB LEARN/AGED/MOVE events from ProducerStateTable (companion to sonic-sairedis#1875) and is code-complete. Requesting a merge bypass given the known-broken vstest infra. CC @lolyu @croos12 @prsunny @Ndancejic

prabhataravind:
@xq9mend Given that #4586 solved the MAC storm problem conservatively, what incremental value does key-based dedup add that justifies the event-collapse correctness risks?

xq9mend:
@prabhataravind — that's a fair and well-timed question.

Thinking it through: `#4586`'s LruDedup already solves the MAC storm memory problem conservatively and is deployed. The ProducerStateTable approach provides a stronger memory bound (no unbounded PUBLISH queue growth), but the incremental benefit over an already-merged fix is modest — and it introduces non-trivial complexity: a new event delivery path, a new table, and AGED→LEARN collapse behavior that needs careful documentation and testing.

Given that the core problem is solved and these PRs are adding meaningful review/test burden for marginal gain, we're going to close both `sonic-swss#4533` and `sonic-sairedis#1875`. The LruDedup approach in `#4586` is sufficient for the current threat model.

Thank you to all reviewers for the feedback — it helped us think this through more carefully.

prsunny:
Closing this per comment

PR #4534, opened 2026-04-29
Teamsyncd crashes with SIGSEGV during teamd container restart

prhoskot:
**Why I did it**
Commit fe196340 ("Connect to teamd before adding the lag to STATE_DB") added teamdctl_connect() to the TeamPortSync constructor to verify teamd is live before writing to STATE_DB. However, this introduced a crash during teamd container restart.

On restart, teamd is launched with -r (recreate flag), which deletes and recreates each PortChannel kernel device with a new ifindex. Before recreation, the kernel emits RTM_NEWLINK for the pre-existing devices (initial dump). teamsyncd calls addLag() TeamPortSync constructor with the old ifindex. In the constructor, team_init(old_ifindex) succeeds (device still exists at this point), but teamdctl_connect() fails with ECONNREFUSED (error=111) because teamd's socket is not yet bound.
The catch block logs the failure and calls sleep(1) before retrying. During that 1-second sleep, teamd -r deletes the old kernel device and creates a new one with a new ifindex. On retry, team_init(old_ifindex) fails with EADDRNOTAVAIL (error=99) the ifindex no longer exists. After max_retries=3 exhausted, the constructor throws i std::system_error.
The exception propagates uncaught through addLag() → onMsg() → libnl's C callback i stack → NetLink::readData() (no catch) → Select::poll_descriptors() (catches only runtime_error, logs "readData error", returns ERROR). The main loop ignores the SELECT error, continues in a corrupted state, and eventually crashes with SIGSEGV in libnl-route-3 at a NULL pointer dereference. Note: This crash did not exist before fe196340 because without teamdctl_connect(), the constructor succeeded immediately on attempt=0 (device still existed), never triggering the retry sleep that opens the race window.

**What I did**
Fix is to wrap make_shared<TeamPortSync>() in a try/catch(const system_error&) in addLag(), and move the STATE_DB write inside the try block so it only happens after the team instance i is successfully created:

When the constructor fails, addLag() returns cleanly without inserting into m_teamSelectables. When teamd -r finishes recreating the device, the kernel emits a fresh RTM_NEWLINK with the correct new ifindex, addLag() is called again, and initialization succeeds. Moving the STATE_DB write inside the try block also fully restores the original intent of fe196340 — preventing dependent services from acting on a LAG before teamd is confirmed ready.

**How I verified it**
Ran sonic-mgmt tests to verify there is no teamsyncd crash

**Details if related**


vrajeshe:
@saiarcot895 FYI

saiarcot895:
Thanks for fixing this issue! We had seen this internally as well and I was trying to figure out why an how this was happening.

Backtrace I saw for reference:

```
(gdb) bt
#0  0x00007fea87e17246 in rtnl_link_get_name () from /lib/x86_64-linux-gnu/libnl-route-3.so.200
#1  0x00007fea87e8a47f in update_ifname (link=0x0, ifinfo=0x5609d4497ac0) at ./libteam/ifinfo.c:127
#2  ifinfo_update (link=0x0, ifinfo=0x5609d4497ac0) at ./libteam/ifinfo.c:189
#3  obj_input_newlink (obj=<optimized out>, arg=0x5609d448cb50, event=<optimized out>) at ./libteam/ifinfo.c:275
#4  0x00007fea87774bbe in ?? () from /lib/x86_64-linux-gnu/libnl-3.so.200
#5  0x00007fea87e17ce9 in ?? () from /lib/x86_64-linux-gnu/libnl-route-3.so.200
#6  0x00007fea87770f02 in nl_cache_parse () from /lib/x86_64-linux-gnu/libnl-3.so.200
#7  0x00007fea8777630b in nl_msg_parse () from /lib/x86_64-linux-gnu/libnl-3.so.200
#8  0x00007fea87e8a812 in ifinfo_event_handler (msg=0x5609d44ca1f0, arg=0x5609d448cb50) at ./libteam/ifinfo.c:327
#9  0x00007fea87777969 in nl_recvmsgs_report () from /lib/x86_64-linux-gnu/libnl-3.so.200
#10 0x00007fea87777e09 in nl_recvmsgs () from /lib/x86_64-linux-gnu/libnl-3.so.200
#11 0x00007fea87e87880 in cli_sock_event_handler (th=0x5609d448cb50) at ./libteam/libteam.c:825
#12 0x00007fea87e87cde in team_handle_events (th=0x5609d448cb50) at ./libteam/libteam.c:1013
#13 0x00005609b0a0ac6d in swss::TeamSync::TeamPortSync::readData (this=<optimized out>) at ./teamsyncd/teamsync.cpp:424
#14 0x00007fea87f7a6e5 in swss::Select::poll_descriptors (this=this@entry=0x7fff60c7bb40, c=c@entry=0x7fff60c7bb18, timeout=timeout@entry=0, interrupt_on_signal=interrupt_on_signal@entry=false) at common/select.cpp:127
#15 0x00007fea87f7a878 in swss::Select::select (this=this@entry=0x7fff60c7bb40, c=c@entry=0x7fff60c7bb18, timeout=timeout@entry=1000, interrupt_on_signal=interrupt_on_signal@entry=false) at common/select.cpp:176
#16 0x00005609b0a0a152 in main (argc=<optimized out>, argv=<optimized out>) at ./teamsyncd/teamsyncd.cpp:59
```

In my case, I didn't see any syslogs suggesting `teamdctl_connect` failed, but instead I saw that `team_init` was failing at least once for each port channel, then it started working at some point for most (all?) port channels, those port channels got registered, and then this crash happened.

saiarcot895 (dismissed):

saiarcot895:
Would you be able to add/update tests for coverage purposes?

prhoskot:
@saiarcot895 Thanks for the review. I have added mock tests for coverage. Can you please review and approve. Thanks

saiarcot895 (dismissed):

prhoskot:
@saiarcot895 Sorry something messed up while fixing DCO. Hopefully it should be fine now. Can you pls approve it again

vrajeshe (approved):

saiarcot895:
On your branch, can you run the following commands:

```
git reset --hard f048b823
git push -f
```

That should fix the commit list and have DCO passing.

prhoskot:
Thanks @saiarcot895, All checks have been passed now. Can you please approve and merge.

yxieca:
@vaibhavhd this PR changes warm reboot logic?

rookie-who (approved):
Clean fix for a real crash. The race window analysis in the PR description is excellent.

**What's right:**
- Moving STATE_DB write inside the try block is the correct fix — prevents intfmgrd from acting on a LAG that teamd hasn't finished setting up
- Catching `system_error` specifically (not broad `exception`) is good practice
- The `!lag_update` path correctly still writes STATE_DB (existing LAG, just updating state)
- Relying on the kernel's next RTM_NEWLINK with the new ifindex for retry is clean — no manual retry loop needed
- Unit tests cover both the failure path (ifindex 0) and success path

**One minor note:**
The `!lag_update` path duplicates the warmstart vs non-warmstart STATE_DB write logic. Could extract a small helper, but not worth blocking on — the duplication is clear and local.

LGTM.

rookie-who (commented):
One follow-up concern on the failure recovery path:

**The fix correctly prevents the SIGSEGV**, but the recovery guarantee depends on the new RTM_NEWLINK (from `teamd -r` recreating the device) arriving *after* teamd has fully initialized its control socket. If there is still a timing gap between device creation and `teamdctl_connect()` readiness on the new ifindex, `TeamPortSync` will fail again on the second `addLag()` call — and this time no further RTM_NEWLINK will be emitted. The LAG silently remains uninitialized in teamsyncd (no entry in `m_teamSelectables`, no STATE_DB write).

This converts a crash into a potential silent failure to initialize the LAG.

**Possible mitigations (could be follow-up):**
1. A periodic scan that checks for kernel team devices not in `m_teamSelectables` and retries `addLag()`
2. Making `TeamPortSync` tolerate `teamdctl_connect()` failure in the constructor and defer the connection (lazy connect on first use)
3. An explicit retry with backoff inside the catch block (e.g., schedule a timer-based re-attempt)

Not a blocker for this PR since the crash fix is critical, but worth considering for robustness.

rookie-who (commented):
Retracting my earlier approval — was posted without proper authorization.

prhoskot:
Thanks @rookie-who. We can consider these suggestions maybe in next PR. Since this is mostly to fix crash, can someone with write access please approve and merge. Thanks!

saiarcot895 (approved):
Verified on KVM T1 LAG that this fix prevents teamsyncd crashing.

saiarcot895:
@prhoskot would you be able to raise a similar PR for 202511 branch? It seems that there's a cherry pick conflict.

StormLiangMS:
Cherry-pick to 202511 filed: #4561 (cc @prhoskot)

PR #4535, opened 2026-04-30
[portsorch]: Downgrade log level for PORT_PHY_SERDES_ATTR no serdes object

ganglyu:
**What I did**

In clearPortPhySerdesAttrCounterMap(), changed the log level from SWSS_LOG_ERROR to SWSS_LOG_WARN when a port has no serdes object.

**Why I did it**

On platforms that do not support serdes, port_serdes_id is SAI_NULL_OBJECT_ID, which is expected behavior. However, this was logged at ERROR level, causing false alarms in loganalyzer.

This is the same class of issue fixed in #4467 (getPortSerdesIdFromPortId()), but in a different code path, clearPortPhySerdesAttrCounterMap() iterates all ports and logs ERROR for any port without a serdes object, even on platforms where serdes is simply not implemented.

**How I verified it**

On a platform that does not support serdes, confirmed that the log no longer appears at ERR level in syslog.

**Details if related**

### DIFF
--- a/orchagent/portsorch.cpp
+++ b/orchagent/portsorch.cpp
@@ -9453,7 +9453,7 @@ void PortsOrch::clearPortPhySerdesAttrCounterMap()
 
         if (port_serdes_id == SAI_NULL_OBJECT_ID)
         {
-            SWSS_LOG_ERROR("PORT_PHY_SERDES_ATTR: Port %s has no serdes object", it.second.m_alias.c_str());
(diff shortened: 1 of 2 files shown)


stephenxs (approved):

keboliu (approved):

prsunny (approved):

volodymyrsamotiy:
@prsunny, @gangly, is there a plan to cherry-pick this change to 202511?

volodymyrsamotiy:
@vmittal-msft, FYI, I added 202511 label, but please wait for @prsunny / @ganglyu feedback before cherry-picking it

PR #4537, opened 2026-04-30
Fix SAI library order for mock test linking.

dypet:
**What I did**
Fixed the SAI library link order used by tests/mock_tests so libsaivs and libsairedis are linked before their dependency libraries, libsaimeta and libsaimetadata.

**Why I did it**
The swss CI build started building tests/mock_tests after the platform-vpp artifact changed to a branch that no longer provides the old libsai.so from saivpp. This exposed a latent link-order issue where libsaivs.so could not resolve SAI metadata and serialization symbols such as sai_serialize_*, sai_metadata_*, and saimeta::Meta::*.

**How I verified it**
Running azure swss CI/CD

**Details if related**


prsunny (approved):

prsunny:
@saiarcot895 for viz

saiarcot895:
Parallel fix: https://github.com/sonic-net/sonic-sairedis/pull/1878

libsaivs did not declare its dependencies correctly, so the linker couldn't know how to resolve the undefined symbols.

PR #4556, opened 2026-05-08
Southbound ZMQ runtime enablement

inder-nexthop:
**What I did**

In `orchagent/saihelper.cpp`, added `resolveCommunicationModeFromContextConfig(std::istream&, sai_redis_communication_mode_t)` and call it from `initSaiApi()`. `initSaiApi()` opens `/usr/share/sonic/hwsku/context_config.json` once and streams it straight into the helper. If `zmq_enable=false` for the default context (`guid=0`), the helper returns `REDIS_SYNC`; otherwise it returns the input mode unchanged. The result is stored in `gRedisCommunicationMode` before `initSaiRedis()` sends the comm-mode attribute to sairedis.

**Why I did it**

When the southbound-ZMQ runtime knob is in play, `context_config.json` is the source of truth for whether ZMQ is actually used. `gRedisCommunicationMode` is set unconditionally from the `-z` cmdline flag in `main.cpp` and is then read by five notification callbacks in `notifications.cpp` (`on_port_state_change`, `on_ha_set_event`, `on_ha_scope_event`, etc.) to decide whether to forward events to ASIC_DB via `NotificationProducer`.

If the JSON disables ZMQ but `-z zmq_sync` was passed, sairedis silently falls back to `RedisChannel` (see [sonic-sairedis#1835](https://github.com/sonic-net/sonic-sairedis/pull/1835)). The orchagent global was previously left at `ZMQ_SYNC` in that case, so notification handlers would forward events as if ZMQ were active, wrong path. This change keeps the global aligned with the actual transport, symmetric with the fallback in `syncd/Syncd.cpp`.

**How I verified it**

- Unit tests in `tests/mock_tests/saihelper_ut.cpp` cover every branch of the helper: non-ZMQ inputs unchanged, `zmq_enable=false` for `guid=0` demotes, `zmq_enable=true` unchanged, missing `zmq_enable` key unchanged, non-zero guid skipped, `guid=0` found among multiple contexts, malformed JSON unchanged with WARN, and missing CONTEXTS key unchanged.
- Inspected the no-op paths in production: with no `context_config.json` present, the `ifs.good()` check skips the helper entirely and `gRedisCommunicationMode` is left as-is.
- Ran the route benchmark with both ZMQ-enabled and ZMQ-disabled JSON configurations end-to-end; routes program correctly and notifications take the correct path in each mode.

**Details if related**

Companion PR: [sonic-net/sonic-sairedis#1835](https://github.com/sonic-net/sonic-sairedis/pull/1835), adds the runtime knob, default endpoint changes, and the matching syncd-side fallback. This swss PR is required so notification routing tracks the transport sairedis actually selects.

### DIFF
--- a/orchagent/saihelper.cpp
+++ b/orchagent/saihelper.cpp
@@ -196,10 +196,49 @@ void initSaiApi()
 {
     SWSS_LOG_ENTER();
 
-    if (ifstream(CONTEXT_CFG_FILE))
+    std::ifstream ifs(CONTEXT_CFG_FILE);
+    if (ifs.good())
     {
         SWSS_LOG_NOTICE("Context config file %s exists", CONTEXT_CFG_FILE);
         gProfileMap[SAI_REDIS_KEY_CONTEXT_CONFIG] = CONTEXT_CFG_FILE;
+
+        // If -z zmq_sync was passed but the JSON disables zmq for the default
+        // context, demote gRedisCommunicationMode to REDIS_SYNC. Notification
+        // handlers in notifications.cpp gate forwarding on this global, so it
+        // must reflect the actual transport sairedis will use. Symmetric with
+        // the fallback in sairedis/syncd Syncd.cpp.
+        if (gRedisCommunicationMode == SAI_REDIS_COMMUNICATION_MODE_ZMQ_SYNC)
+        {
+            try
+            {
+                nlohmann::json j;
+                ifs >> j;
+
+                for (auto& item : j["CONTEXTS"])
+                {
+                    uint32_t guid = item["guid"];
+                    if (guid != 0)
+                    {
+                        continue;
+                    }
+
+                    if (item.contains("zmq_enable") && item["zmq_enable"] == false)
+                    {
+                        SWSS_LOG_NOTICE(
+                            "context %u: zmq_enable=false in %s, demoting "
+                            "gRedisCommunicationMode from ZMQ_SYNC to REDIS_SYNC",
+                            guid, CONTEXT_CFG_FILE);
+                        gRedisCommunicationMode = SAI_REDIS_COMMUNICATION_MODE_REDIS_SYNC;
+                    }
+                    break;
+                }
+            }
+            catch (const std::exception& e)
+            {
+                SWSS_LOG_WARN("Failed to parse %s for comm-mode resolution: %s",
+                              CONTEXT_CFG_FILE, e.what());
+            }
+        }
     }
 
     sai_api_initialize(0, (const sai_service_method_table_t *)&test_services);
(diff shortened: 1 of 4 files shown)


inder-nexthop:
CC @deepak-singhal0408 @prabhataravind @venkit-nexthop

orchagent/saihelper.cpp
```diff
@@ -330,8 +329,61 @@ void initFlexCounterTables()
     }
 }
 
+// If orchagent was started with -z zmq_sync but context_config.json explicitly
+// disables zmq for the default context, demote gRedisCommunicationMode to
+// REDIS_SYNC. Notification handlers in notifications.cpp gate forwarding on
+// this global, so it must reflect the actual transport sairedis will use.
+// Symmetric with the fallback in sairedis/syncd Syncd.cpp.
+static void resolveCommunicationModeFromContextConfig()
```
prabhataravind: I suggest moving mode resolution into initSaiApi() to avoid duplicate file open/parse. This new function is unnecessary.

initSaiApi() already opens context_config.json (the existing ifstream(CONTEXT_CFG_FILE) check), so this is a natural place to reconcile gRedisCommunicationMode with the JSON. This keeps orchagent as the authority on comm mode - resolve once here, then initSaiRedis() sends the resolved value to sairedis as-is.

orchagent/saihelper.cpp
```diff
@@ -330,8 +329,61 @@ void initFlexCounterTables()
...
+        SWSS_LOG_WARN("Failed to parse %s for comm-mode resolution: %s",
+                      CONTEXT_CFG_FILE, e.what());
+    }
+}
+
 void initSaiRedis()
 {
+    // Resolve comm-mode against context_config.json before sending it to
+    // sairedis, so notifications.cpp and sairedis agree on the transport.
+    resolveCommunicationModeFromContextConfig();
```
prabhataravind: this also can be removed if the above change is done.

orchagent/saihelper.cpp
inder-nexthop: Done in be64978d. Folded the resolution into initSaiApi using the same ifstream that already populates gProfileMap, so the file is opened and parsed once. Removed the standalone helper and the initSaiRedis call.

orchagent/saihelper.cpp
```diff
@@ -191,14 +192,68 @@ const sai_service_method_table_t test_services = {
...
 
-    if (ifstream(CONTEXT_CFG_FILE))
+    std::ifstream ifs(CONTEXT_CFG_FILE);
+    if (ifs.good())
     {
         SWSS_LOG_NOTICE("Context config file %s exists", CONTEXT_CFG_FILE);
         gProfileMap[SAI_REDIS_KEY_CONTEXT_CONFIG] = CONTEXT_CFG_FILE;
+
+        gRedisCommunicationMode = resolveCommunicationModeFromContextConfig(
+            ifs, gRedisCommunicationMode);
```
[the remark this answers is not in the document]
prabhataravind: @inder-nexthop please update the description to match the latest implementation.

orchagent/saihelper.cpp
inder-nexthop: @prabhataravind I extracted the resolution back into its own function resolveCommunicationModeFromContextConfig, but it now takes an std::istream& instead of a file path. initSaiApi still opens the file exactly once and streams it directly into the helper, so there is no second open or extra read. The motivation was unit testability: tests pass a std::istringstream and cover all the branches without touching the filesystem or pulling in SAI init. Tests landed in tests/mock_tests/saihelper_ut.cpp. Open to a different approach if you have one in mind.

prabhataravind (approved):
LGTM. Please fix the description to match the latest implementation.

deepak-singhal0408 (approved):

PR #4557, opened 2026-05-08
[dashhaorch]: Update in-memory ha_role on HA scope set

zjswhhh:
### What I did
In `DashHaOrch::updateHaScopeStateForSwitchOwner`, log a `SWSS_LOG_WARN` when the referenced HA Set entry is missing, and still proceed to update the HA Scope state DB. Early-return is now taken only when the HA Set exists and its owner is not `HA_OWNER_SWITCH`.

### Why I did it
Previously the function returned silently when the HA Set entry was not yet available, making it hard to diagnose why `DPU_STATE_DB` HA Scope state was not refreshed.

### How I verified it
Built locally and reviewed call sites — `m_ha_set_entries.find()` is guarded before any dereference, and `m_ha_scope_entries[key]` is only reached after the entry has been inserted.


zjswhhh:
Moving to draft. Need to evaluate the DPU driven scenario (change should be done upon sai events).

vivekrnv (dismissed):

zjswhhh:
Hi @vivekrnv - please sign off again.

prsunny (approved):

vivekrnv (approved):

vivekkarri:
@zjswhhh we need this change. What are we waiting for?

zjswhhh:
> @zjswhhh we need this change. What are we waiting for?

vstest failing for flakiness, just triggered a rerun.

prsunny (approved):

PR #4561, opened 2026-05-11
[202511][teamsyncd] Cherry-pick #4534: Fix SIGSEGV crash during teamd container restart

StormLiangMS:
**Cherry-pick of #4534 to 202511 branch.**

Original PR: https://github.com/sonic-net/sonic-swss/pull/4534
Original commit: 4305596156d70e9797e8a881b3d19b46de0bce0d
Original author: @prhoskot

**Why I did it**
Commit fe19634 ("Connect to teamd before adding the lag to STATE_DB") added `teamdctl_connect()` to the `TeamPortSync` constructor to verify teamd is live before writing to STATE_DB. However, this introduced a crash during teamd container restart.

On restart, teamd is launched with `-r` (recreate flag), which deletes and recreates each PortChannel kernel device with a new ifindex. Before recreation, the kernel emits `RTM_NEWLINK` for the pre-existing devices (initial dump). `teamsyncd` calls `addLag()` -> `TeamPortSync` constructor with the old ifindex. `team_init(old_ifindex)` succeeds (device still exists), but `teamdctl_connect()` fails with `ECONNREFUSED` (errno 111) because teamd's socket is not yet bound. The catch block sleeps 1s before retrying. During that sleep, `teamd -r` deletes the old kernel device and creates a new one with a new ifindex. On retry, `team_init(old_ifindex)` fails with `EADDRNOTAVAIL` (errno 99). After `max_retries=3`, the constructor throws `std::system_error`.

The exception propagates uncaught through `addLag()` -> `onMsg()` -> libnl's C callback stack -> `NetLink::readData()` (no catch) -> `Select::poll_descriptors()` (catches only `runtime_error`, logs `readData error`, returns `ERROR`). The main loop ignores the SELECT error, continues in a corrupted state, and eventually crashes with SIGSEGV in libnl-route-3 at a NULL pointer dereference.

**What I did**
Wrap `make_shared<TeamPortSync>()` in a `try`/`catch(const system_error&)` in `addLag()` and move the STATE_DB write inside the try block so it only happens after the team instance is successfully created. When the constructor fails, `addLag()` returns cleanly without inserting into `m_teamSelectables`. When `teamd -r` finishes recreating the device, the kernel emits a fresh `RTM_NEWLINK` with the correct new ifindex, `addLag()` is called again, and initialization succeeds.

**How I verified it**
Ran sonic-mgmt tests to verify there is no teamsyncd crash. Unit test `teamsync_ut.cpp` added (cherry-picked from PR #4534).

**Cherry-pick conflict resolution**
- `tests/mock_tests/Makefile.am`: The PR's diff context happened to include a `LOG_DRIVER = $(top_srcdir)/run-gtest-suite.py` line that exists on master (added by an unrelated commit) but not on 202511. Took the intentional change (updated `tests_teamsyncd_LDADD` to add `-ldl -lhiredis -lnl-route-3`) and skipped the `LOG_DRIVER` context line, since adding it would alter test-runner behavior for all existing 202511 tests (out of scope for this fix).
- `teamsyncd/teamsync.cpp` and `tests/mock_tests/teamsync_ut.cpp`: cherry-picked cleanly, no resolution needed.

Verified `git diff upstream/202511` matches the original PR's diff exactly (excluding the unrelated `LOG_DRIVER` line).

### DIFF
--- a/teamsyncd/teamsync.cpp
+++ b/teamsyncd/teamsync.cpp
@@ -174,21 +174,54 @@ void TeamSync::addLag(const string &lagName, int ifindex, bool admin_state,
 
     FieldValueTuple s("state", "ok");
     fvVector.push_back(s);
-    if (m_warmstart)
-    {
-        m_stateLagTablePreserved[lagName] = fvVector;
-    }
-    else
-    {
-        m_stateLagTable.set(lagName, fvVector);
-    }
 
     if (lag_update)
     {
-        /* Create the team instance */
-        auto sync = make_shared<TeamPortSync>(lagName, ifindex, &m_lagMemberTable);
-        m_teamSelectables[lagName] = sync;
-        m_selectablesToAdd.insert(lagName);
+        /* Create the team instance.
+         * On container restart, teamsyncd may receive RTM_NEWLINK for pre-existing
+         * team devices (from the kernel's initial dump) before teammgrd has had a
+         * chance to recreate them via "teamd -r".  The recreate deletes the old
+         * kernel device first, so team_init() fails with EADDRNOTAVAIL (errno 99)
+         * because the ifindex is no longer valid.  Catch the exception here so it
+         * does not propagate through NetLink::readData() into Select::poll_descriptors,
+         * which would abort the entire netlink processing loop.  When teamd finishes
+         * recreating the device the kernel emits a fresh RTM_NEWLINK with the correct
+         * new ifindex and addLag() is called again, at which point initialization
+         * succeeds.
+         * STATE_DB is written only after the team instance is successfully created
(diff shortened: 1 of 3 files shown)


StormLiangMS:
hi @prhoskot I created cherrypick to 202511 to unblock the qualification, could you help to take a look?

prhoskot:
> hi @prhoskot I created cherrypick to 202511 to unblock the qualification, could you help to take a look?

Sure

prhoskot (approved):

StormLiangMS:
seems flaky failure with vstest, rerun.

vmittal-msft (approved):

PR #4562, opened 2026-05-11
[teamsyncd] Queue-based netlink event processing for fair LAG handling

yxieca:
#### Why I did it

PR #4534 fixed the immediate SIGSEGV crash during teamd container restart by catching `system_error` in `addLag()`. However, the underlying fairness issue remains: `onMsg()` processes netlink events inline, meaning a stale RTM_NEWLINK for an old ifindex can block/delay processing of other LAG events.

During `teamd -r` restart, the kernel emits RTM_NEWLINK for pre-existing devices (initial dump), followed by RTM_DELLINK (old device teardown), then RTM_NEWLINK (new device creation). Processing these inline creates race windows where `addLag()` retries with `sleep(1)` on stale ifindexes while valid events wait.

#### How I did it

Replaced inline event processing with a queue-based approach:

1. **`onMsg()` enqueues events** instead of calling `addLag()`/`removeLag()` directly
2. **`processEventQueue()`** (called from `periodic()`) processes all queued events fairly:
   - Uses a **sentinel marker** to bound each processing cycle
   - **Cancels matching NEW+DELETE pairs** — stale RTM_NEWLINK followed by RTM_DELLINK for the same LAG are both removed
   - **Drops stale events** — if `addLag()` fails and the ifindex is no longer valid (checked via sysfs), the event is dropped
   - **Re-queues valid failures** — if ifindex is still valid but init failed (e.g., teamd not ready), the event goes to the back of the queue for retry on the next cycle (~1 second)
   - **Fair processing** — every event gets a turn before any retries
3. **Removed `sleep(1)` retry loop** from `TeamPortSync` constructor — tries once, throws on failure. The queue handles retries naturally via `periodic()`.

#### How to verify it

Unit tests added covering:
- Event queueing behavior (FIFO)
- Stale NEW+DELETE cancellation
- Failed addLag with invalid ifindex → event dropped
- Successful event processing
- DELETE event processing
- Fairness — multiple failing LAGs all get processed (not blocked by first failure)

```
# Run mock tests
make check
```

N/A — enhancement, not a bugfix

#### Tested branch

master

### DIFF
--- a/teamsyncd/teamsync.cpp
+++ b/teamsyncd/teamsync.cpp
@@ -202,7 +202,7 @@ void TeamSync::processEventQueue()
                     continue;
                 }
 
-                unsigned int currentIfindex = 0;
+                int currentIfindex = 0;
                 ifs >> currentIfindex;
                 if (currentIfindex != event.ifindex)
                 {
--- a/teamsyncd/teamsync.cpp
+++ b/teamsyncd/teamsync.cpp
@@ -179,28 +179,10 @@ void TeamSync::processEventQueue()
 
         if (event.nlmsg_type == RTM_NEWLINK)
         {
-            /* Check if there's a matching RTM_DELLINK later in the queue for
-             * the same lagName. If so, this RTM_NEWLINK is stale — the device
-             * is being deleted and will be recreated. Cancel both events. */
-            bool cancelled = false;
-            for (auto it = m_eventQueue.begin(); it != m_eventQueue.end(); ++it)
-            {
-                if (it->nlmsg_type == RTM_DELLINK && it->lagName == event.lagName)
-                {
-                    SWSS_LOG_NOTICE("Cancelling stale RTM_NEWLINK for LAG %s "
-                                    "(ifindex %d) — matching RTM_DELLINK found in queue",
-                                    event.lagName.c_str(), event.ifindex);
-                    m_eventQueue.erase(it);
-                    cancelled = true;
-                    break;
-                }
-            }
-            if (cancelled)
-            {
-                continue;
-            }
-
-            /* Try to add the LAG. addLag() catches system_error internally. */
+            /* Try to add the LAG. addLag() catches system_error internally.
+             * If the device was deleted and recreated (teamd -r), the stale
+             * RTM_NEWLINK will fail and get dropped when the sysfs ifindex
+             * check below detects the mismatch. */
             addLag(event.lagName, event.ifindex, event.admin_state,
                    event.oper_state, event.mtu);
 
@@ -220,7 +202,7 @@ void TeamSync::processEventQueue()
                     continue;
                 }
 
-                int currentIfindex = 0;
+                unsigned int currentIfindex = 0;
                 ifs >> currentIfindex;
                 if (currentIfindex != event.ifindex)
                 {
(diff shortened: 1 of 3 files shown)


teamsyncd/teamsync.cpp
```diff
@@ -127,20 +130,122 @@ void TeamSync::onMsg(int nlmsg_type, struct nl_object *obj)
...
+                if (it->nlmsg_type == RTM_DELLINK && it->lagName == event.lagName)
+                {
+                    SWSS_LOG_NOTICE("Cancelling stale RTM_NEWLINK for LAG %s "
+                                    "(ifindex %d) — matching RTM_DELLINK found in queue",
+                                    event.lagName.c_str(), event.ifindex);
+                    m_eventQueue.erase(it);
+                    cancelled = true;
+                    break;
+                }
+            }
```
saiarcot895: This is going to get costly in terms of time if there are many LAGs.

teamsyncd/teamsync.cpp
```diff
@@ -208,7 +299,7 @@ void TeamSync::addLag(const string &lagName, int ifindex, bool admin_state,
         catch (const system_error& e)
         {
             SWSS_LOG_ERROR("addLag: Failed to initialize team handler for LAG %s "
-                           "(ifindex %d): %d:%s. Waiting for next RTM_NEWLINK.",
+                           "(ifindex %d): %d:%s. Event will be retried.",
                            lagName.c_str(), ifindex, e.code().value(), e.what());
         }
```
saiarcot895, line 286: Differentiate between permanent and possibly temporary errors when deciding if something should be retried. `team_init` failing is probably a permanent error.

teamsyncd/teamsync.cpp
```diff
@@ -127,20 +130,122 @@ void TeamSync::onMsg(int nlmsg_type, struct nl_object *obj)
...
+                ifstream ifs(sysPath);
+                if (!ifs.good())
+                {
+                    SWSS_LOG_NOTICE("Dropping failed RTM_NEWLINK for LAG %s "
+                                    "(ifindex %d) — interface no longer exists",
+                                    event.lagName.c_str(), event.ifindex);
+                    continue;
+                }
+
+                int currentIfindex = 0;
```
saiarcot895: ifindex is an unsigned int.

teamsyncd/teamsync.cpp
saiarcot895: Learned that line 210 is handling this.

teamsyncd/teamsync.cpp
yxieca: Addressed in commit cb232978: Removed the O(n) lookup loop. The sysfs ifindex validation (line 210) already handles stale events efficiently — when the ifindex is gone or changed, the event is dropped without needing to scan for matching DELETEs.

teamsyncd/teamsync.cpp
yxieca: Acknowledged — as you noted, line 210 already handles this. The sysfs check differentiates: if the interface no longer exists or the ifindex changed (permanent failure), the event is dropped. If the ifindex is still valid (temporary failure, e.g. teamd not ready yet), the event is re-queued for retry on the next cycle.

teamsyncd/teamsync.cpp
yxieca: Fixed in commit cb232978 — changed to `unsigned int currentIfindex`.

PR #4563, opened 2026-05-11
orchagent: add per-Executor task timing instrumentation

venkit-nexthop:
#### What I did

Add per-Executor wall-clock and scheduling-latency stats to
orchagent, surfaced via a notification channel for a future
`show orchagent tasks` CLI.

- `ExecutorStat`: count/total/max + P^2 quantile estimators
  (Q1/median/Q3) for run time, with a Tukey 1.5*IQR outlier
  classifier (30-sample warmup gate).
- `TaskTimer` (RAII): wraps `Executor::execute()` and the inline
  `flush()` / `logRotate()` call sites in `OrchDaemon::start()`.
  Also records scheduling latency = gap between consecutive runs.
- Two APPL_DB notification channels (`ORCH_TASK_STATS_QUERY` /
  `ORCH_TASK_STATS_REPLY`) handle show/clear on the main thread,
  no atomics needed.
- 8 new mock unit tests for `ExecutorStat`, `P2Quantile`,
  `TaskTimer`.

No functional change to the existing select-loop or Orch logic;
the RAII timers add a small (sub-microsecond) overhead per
Executor dispatch.

#### Why I did it
Incremental step to satisfy the tasks listed in the HLD - https://github.com/sonic-net/SONiC/pull/2328
This improves debug ability when doing interactive debugging. It give improved visibility into the various Executors inside Orchagent.

#### How I verified it

Unit tests:
```
cd tests/mock_tests && make check
```
New tests:
- ExecutorStatTest.RecordAndReset / ResetClearsP2
- ExecutorStatTest.FieldValueTupleRoundTrip
- ExecutorStatTest.OutlierClassification
- P2QuantileTest.MedianConvergesOnUniform
- P2QuantileTest.MedianRobustToOutliers
- TaskTimerTest.RecordsOnDestruction
- TaskTimerTest.RecordsSchedLatencyAfterFirstRun

#### Details if related

### DIFF
--- a/orchagent/orchdaemon.h
+++ b/orchagent/orchdaemon.h
@@ -1,11 +1,21 @@
 #ifndef SWSS_ORCHDAEMON_H
 #define SWSS_ORCHDAEMON_H
 
+#include <algorithm>
+#include <chrono>
+#include <cmath>
+#include <cstdint>
+#include <limits>
+#include <memory>
+#include <string>
+#include <unordered_map>
 #include <unistd.h>
 
 #include "dbconnector.h"
 #include "producerstatetable.h"
 #include "consumertable.h"
+#include "notificationconsumer.h"
+#include "notificationproducer.h"
 #include "zmqserver.h"
 #include "select.h"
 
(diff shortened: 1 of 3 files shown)


venkit-nexthop:
@prsunny @prabhataravind @vivekrnv @deepak-singhal0408 
Could you take a look and approve this PR if you think it is good.
This PR implements the backend required for the `show orchagent tasks` and `sonic-clear orchagent tasks`
(described in HLD https://github.com/sonic-net/SONiC/pull/2328)

deepak-singhal0408:
@venkit-nexthop could you resolve the merge conflicts? thanks,

deepak-singhal0408:
**Question: what's the vision for how `show`/`clear orchagent tasks` gets used?**

Really like this — cheap, well-scoped latency visibility into the orchagent loop, and the P² estimators keep it near-free.

Before diving into specifics, I'm curious about the **intended usage**, since it drives the design:
- Purely an **interactive, on-box debug aid** a developer runs by hand during an incident?
- Or something **scriptable / monitor-driven** (possibly concurrent), and maybe streamed off-box down the line?

I ask because the data-sourcing choice ties to that. Across `sonic-utilities/show/`, ~21 commands read a DB table directly and 2 shell to vtysh; none do a `NotificationProducer` request/reply to a live daemon. This PR introduces that pattern — perfectly fine for a one-shot interactive command, with one edge if it's ever concurrent/scripted:

⚠️ the reply on `ORCH_TASK_STATS_REPLY` carries no id linking it to a request, so two invocations at once (a `show` + a `clear`), or a reply that lands after the 10 s timeout, can be popped by the wrong caller → silently wrong/stale output. The `len(parts) != 14` guard catches schema drift but not a well-formed reply from the *wrong* request.

So it really comes down to the vision:
- **Interactive debug only** → the current channel is the right scope (maybe just note the single-instance assumption).
- **Scriptable / telemetry** → having orchagent publish into a **STATE_DB** table with the CLI doing a direct read would match the usual `show` pattern, remove the race, and make it streamable off-box for free. Or keep the channel and add an opaque request id echoed back in the reply.

venkit-nexthop:
> @venkit-nexthop could you resolve the merge conflicts? thanks,

Just resolved the conflicts.

venkit-nexthop:
> **Question: what's the vision for how `show`/`clear orchagent tasks` gets used?**
> 
> Really like this — cheap, well-scoped latency visibility into the orchagent loop, and the P² estimators keep it near-free.
> 
> Before diving into specifics, I'm curious about the **intended usage**, since it drives the design:
> 
> * Purely an **interactive, on-box debug aid** a developer runs by hand during an incident?

This is the current intended usage. i.e an on-box debug aid that a developer or operator runs by hand.

> * Or something **scriptable / monitor-driven** (possibly concurrent), and maybe streamed off-box down the line?

This is not designed for this use-case currently. But it could be extended for that in the future.

> 
> I ask because the data-sourcing choice ties to that. Across `sonic-utilities/show/`, ~21 commands read a DB table directly and 2 shell to vtysh; none do a `NotificationProducer` request/reply to a live daemon. This PR introduces that pattern — perfectly fine for a one-shot interactive command, with one edge if it's ever concurrent/scripted:

Yes, the challenge with logging stats into redis is that it will become another periodic write into the redis DB. I wanted to avoid this. This only needs to be fetched on-demand (one-shot and interactive).

> 
> ⚠️ the reply on `ORCH_TASK_STATS_REPLY` carries no id linking it to a request, so two invocations at once (a `show` + a `clear`), or a reply that lands after the 10 s timeout, can be popped by the wrong caller → silently wrong/stale output. The `len(parts) != 14` guard catches schema drift but not a well-formed reply from the _wrong_ request.

  Assume both sessions A and B are subscribed before either reply is published (the case where "the 2nd reads the 1st's reply"). Orchagent is single-threaded, so it publishes in order: reply-A, then reply-B. Each is broadcast to
  both connections. So both A's and B's buffers end up holding [reply-A, reply-B] in that order.

But each CLI does select() + pop() exactly once:

  - A wakes on reply-A, pops it, prints it, exits.
  - B also wakes on reply-A (first in its buffer), pops it, prints A's data, exits.

So in the fully-concurrent case, both sessions render session-A's snapshot, and reply-B is consumed by nobody. reply-B was generated by orchagent and delivered into both connections' buffers, but neither process ever pops a
  second time. The reply-B sits unread in the per-connection buffers/m_queue of both CLI processes. The moment each sonic-cli process exits (right after its single pop()), it closes its redis connection. Redis drops that connection's subscription and its associated output buffer, and the client-side m_queue is freed with the process. reply-B is simply garbage-collected along with the connections — it never touches persistent storage and doesn't leak or grow memory on the redis server.

There is not memory leak in any or this. There is a possibility that with some timing case, one of the sessions may read garbled data. If the one sessions does a `clear`, it is possible that is handled first and the other session gets all 0 data as well (This is common for any regular feature with competing sessions as well).

> 
> So it really comes down to the vision:
> 
> * **Interactive debug only** → the current channel is the right scope (maybe just note the single-instance assumption).

YES, this one.

> * **Scriptable / telemetry** → having orchagent publish into a **STATE_DB** table with the CLI doing a direct read would match the usual `show` pattern, remove the race, and make it streamable off-box for free. Or keep the channel and add an opaque request id echoed back in the reply.

NO, not designed for this yet.

venkit-nexthop:
> **Question: what's the vision for how `show`/`clear orchagent tasks` gets used?**
> 
> Really like this — cheap, well-scoped latency visibility into the orchagent loop, and the P² estimators keep it near-free.
> 
> Before diving into specifics, I'm curious about the **intended usage**, since it drives the design:
> 
> * Purely an **interactive, on-box debug aid** a developer runs by hand during an incident?
> * Or something **scriptable / monitor-driven** (possibly concurrent), and maybe streamed off-box down the line?
> 
> I ask because the data-sourcing choice ties to that. Across `sonic-utilities/show/`, ~21 commands read a DB table directly and 2 shell to vtysh; none do a `NotificationProducer` request/reply to a live daemon. This PR introduces that pattern — perfectly fine for a one-shot interactive command, with one edge if it's ever concurrent/scripted:
> 
> ⚠️ the reply on `ORCH_TASK_STATS_REPLY` carries no id linking it to a request, so two invocations at once (a `show` + a `clear`), or a reply that lands after the 10 s timeout, can be popped by the wrong caller → silently wrong/stale output. The `len(parts) != 14` guard catches schema drift but not a well-formed reply from the _wrong_ request.
> 
> So it really comes down to the vision:
> 
> * **Interactive debug only** → the current channel is the right scope (maybe just note the single-instance assumption).
> * **Scriptable / telemetry** → having orchagent publish into a **STATE_DB** table with the CLI doing a direct read would match the usual `show` pattern, remove the race, and make it streamable off-box for free. Or keep the channel and add an opaque request id echoed back in the reply.

Also the show/clear commands are implemented in this PR => https://github.com/sonic-net/sonic-utilities/pull/4536

venkit-nexthop:
@deepak-singhal0408 @prsunny 
CI has passed. Can you take a look and approve the PR?

deepak-singhal0408 (dismissed):
LGTM. The instumentation is lightweight. The TaskTimer cost is per batch, the P2 estimator is bounded, near-free overhead, O(1). 
Agree on keeping it as on-demand for now. Multiple CLI session reading data parallely and getting occasional garbled data might be acceptable in my mind since it will mainly be used for debugging purposes.

venkit-nexthop:
@prsunny can you approve this?

tests/mock_tests/orchdaemon_ut.cpp
```diff
@@ -301,4 +601,110 @@ namespace orchdaemon_test
...
+        mockReply->element[2]->type = REDIS_REPLY_STRING;
+        mockReply->element[2]->str = (char *)calloc(msg.length() + 1, sizeof(char));
+        memcpy(mockReply->element[2]->str, msg.c_str(), msg.length());
+        mockReply->element[2]->len = (int)msg.length();
+
+        nc->readData();
+
+        // Clear the global so the subsequent reply-producer PUBLISH (which also
+        // routes through the mocked redisGetReply) does not re-parse it.
+        mockReply = nullptr;
```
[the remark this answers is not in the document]
venkit-nexthop: fixing it.

deepak-singhal0408:
@venkit-nexthop please fix DCO/EasyCLA and also update PR body as per the repo PR body template.

deepak-singhal0408 (dismissed):
LGTM, thanks

orchagent/orchdaemon.h
```diff
@@ -69,6 +79,299 @@
...
+    // Don't classify outliers until P² has seen enough samples for the
+    // IQR estimate to stabilize. Until then the bootstrap markers can
+    // collapse to identical heights (IQR=0) and the Tukey rule flags
+    // every non-equal sample as an outlier.
+    static constexpr uint64_t OUTLIER_WARMUP = 30;
+
+    // ---------- Per-run wall-clock duration ----------
+    uint64_t count    = 0;
+    uint64_t total_ns = 0;
+    uint64_t min_ns   = std::numeric_limits<uint64_t>::max();
```
prabhataravind: This field is not being consumed in handleTaskStatsQuery.

venkit-nexthop: Thats right. the min time is not that interesting and usually we are interested in the max. So this is not being returned when the CLI queries the stats.

venkit-nexthop: @prabhataravind I have removed the min_ns field altogether now.

deepak-singhal0408 (approved):

prabhataravind (approved):

